PR #223, opened 2022-11-30
Soil and Crop redesign

morrowcj:
This pull request will incorporate the soil and crop redesign of 2022-2023 into the main branch. 

The PR will remain a draft until all Crop and Soil code has been refactored and tested. 

Anyone looking to review updates should look exclusively in the `SC_redesign/` folder. This folder is effectively a brand new Soil and Crop module and contains all the code that will eventually be the new SC module. The will remain separate from `RUFAS` until we are sure that it has full functionality and works properly. After such time we will replace the code that resides in the `RUFAS/routines/field/` directory. 

To be very clear and explicit: **do not** bother reviewing the Soil and Crop code in `RUFAS/routines/field/`, as it will not be updated until the redesign is complete. The code in this folder is outdated.

The only code that needs to be reviewed are contained in files not covered by sub-PRs: `biomass_allocation.py`, `growth_constraints.py`, `heat_units.py`, `nitrogen_incorporation.py`, and `water_dynamics.py` (and associated test files) were all changed from the master branch **prior** to starting the sub-PR process.

PR #224, opened 2022-11-30
Canopy growth/LAI refactor

morrowcj:
This PR, is a part of the Soil and Crop redesign. It is the first sub-PR of a series that aims to make PR #223 more manageable to review. Each sub-PR will make small changes to the `SC_redesign` branch to allow reviewers an easier time to understand, comment, and make suggestions on each change. 

This PR focuses specifically on the restructuring of the leaf area crop processes, contained in [leaf_area_index.py](SC_redesign/Crop_and_Soil/crop/leaf_area_index.py). The test suite for these processes is implemented in [test_leaf_area_index.py](SC_redesign/Crop_and_Soil/tests_SC/test_leaf_area_index.py)

`grow_canopy()` is the primary process function that will be called externally. This function updates leaf attributes (primarily `leaf_area_index`). The function follows one set of methods during a plant's normal growth and a different set for a plant that is undergoing senescence.

### DIFF
--- a/SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
+++ b/SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
@@ -4,11 +4,185 @@ from math import exp, log, sqrt
 This module is based off of the 'Canopy Cover and Height' section of SWAT
 """
 
+class LeafAreaIndex:
+    def __init__(self):
+        # fixed attributes (unchanged during simulations)
+        self.first_heat_fraction_point = 0.15
+        self.second_heat_fraction_point = 0.50
+        self.first_leaf_fraction_point = 0.01
+        self.second_leaf_fraction_point = 0.95
+        self.max_canopy_height = 2.5  # m
+        self.growth_factor = 1.0
+        self.max_leaf_area_index = 3.0
+        self.senescent_heat_fraction = 0.9
+        # variable attributes (change throughout simulations)
+        self.leaf_area_index = 0
+        self.heat_fraction = 0.73
+        self.is_in_senescence = False
+        # empty variables
+        self._lai_shapes = None
+        self.optimal_leaf_area_fraction = None
+        self.canopy_height = None
+        self.leaf_area_added = None
+        self.max_leaf_area_change = None
+        self.previous_leaf_area_index = None
+        self.previous_optimal_leaf_area_fraction = None
+
+    def grow_canopy(self):
+        """main leaf area index function"""
+        self.determine_optimal_leaf_area_fraction()
+        self.determine_canopy_height()
+        self.check_for_senescence()
+        if self.is_in_senescence:  # senescence
+            self.senesce_leaf_area()
+        else:  # normal growth
+            self.check_previous_leaf_area_values()
+            self.determine_max_leaf_area_change()
+            self.determine_leaf_area_added()
+            self.add_leaf_area()
+        self.shift_leaf_area_time()
+
+    def _determine_lai_shapes(self):
+        """sets lai shape parameters for optimal leaf area fraction calculations"""
+        self._lai_shapes = calc_shape_parameters(self.first_heat_fraction_point, self.second_heat_fraction_point,
+                                                 self.first_leaf_fraction_point, self.second_leaf_fraction_point)
+
+    def determine_optimal_leaf_area_fraction(self):
+        """sets optimal leaf area fraction
+        SWAT Reference: 5:2.1.10"""
+        self._determine_lai_shapes()
+        self.optimal_leaf_area_fraction = calc_optimal_leaf_area_fraction(self.heat_fraction,
+                                                                          self._lai_shapes[0],
+                                                                          self._lai_shapes[1])
+
+    def shift_leaf_area_time(self):
+        """shifts the time window by one step for leaf area attributes"""
+        self.previous_leaf_area_index = self.leaf_area_index
+        self.previous_optimal_leaf_area_fraction = self.optimal_leaf_area_fraction
+
+    def determine_canopy_height(self):
+        """sets the current height of the canopy, in meters
+        SWAT Reference: 5:2.1.14"""
+        self.canopy_height = min(self.max_canopy_height,
+                                 self.max_canopy_height * sqrt(self.optimal_leaf_area_fraction))
+
+    def check_previous_leaf_area_values(self):
+        """check for previous LAI values and set them to 0 if none are present. This function handles the
+        initial time point in the simulation"""
+        if self.previous_optimal_leaf_area_fraction is None:
+            self.previous_optimal_leaf_area_fraction = 0
+        if self.previous_leaf_area_index is None:
+            self.previous_leaf_area_index = 0
+
+    def determine_max_leaf_area_change(self):
+        """sets the maximum amount that leaf area can change under ideal circumstances
+        SWAT Reference: 5:2.1.16"""
+        self.max_leaf_area_change = calc_max_leaf_area_change(self.optimal_leaf_area_fraction,
+                                                              self.previous_optimal_leaf_area_fraction,
+                                                              self.max_leaf_area_index,
+                                                              self.previous_leaf_area_index)
+
+    def determine_leaf_area_added(self):
+        """sets actual leaf area added, by adjusting for the plant growth factor
+        SWAT Reference: 5:3.2.2
+        """
+        self.leaf_area_added = min(self.max_leaf_area_change * sqrt(self.growth_factor), self.max_leaf_area_change)
+
+    def add_leaf_area(self):
+        """add new leaf area to the plant
+        SWAT Reference: 5:2.1.18"""
+        self.leaf_area_index = max(0, self.previous_leaf_area_index + self.leaf_area_added)
+
+    def check_for_senescence(self):
+        """check if the plant is in senescence"""
+        self.is_in_senescence = self.heat_fraction > self.senescent_heat_fraction
+
+    def senesce_leaf_area(self):
+        """determines the leaf area index during senescence
+        SWAT Reference: 5:2.1.19
+        """
+        self.leaf_area_index = calc_senescent_leaf_area_index(self.heat_fraction, self.senescent_heat_fraction,
+                                                              self.optimal_leaf_area_fraction)
 
 
+def calc_senescent_leaf_area_index(heat_fraction, senescent_heat_fraction, optimal_leaf_area_fraction):
+    """calculates a plant's leaf area index during senescence
+
+    Args:
+        heat_fraction: the current fraction of potential heat units
+        senescent_heat_fraction: the fraction of potential heat units at which senescence begins
+        optimal_leaf_area_fraction: the optimal leaf area fraction for the plant
+
+    Returns: the plant's leaf area index
+    """
+    if senescent_heat_fraction == 1:
+        prop = 1 - heat_fraction
+    else:
+        prop = (1 - heat_fraction) / (1 - senescent_heat_fraction)
+
+    return max(prop * optimal_leaf_area_fraction, 0)
+
+# ---- helper functions ----
+
+def calc_shape_log(heat_fraction: float, leaf_area_fraction: float) -> float:
+    """calculates the log term of LAI shape parameter function
+
+    Args:
+        heat_fraction: fraction of heat units accumulated; must be greater than zero
+        leaf_area_fraction: fraction of max leaf area; must be greater than zero, less than one, and not
+            equal to heat_fraction
+
+    Details: used by calc_shape_parameters, where errors are handled
+    """
+    return log((heat_fraction/leaf_area_fraction) - heat_fraction)
+
+def calc_shape_parameters(first_heat_fraction: float, second_heat_fraction: float,
+                          first_leaf_fraction: float, second_leaf_fraction: float) -> list[float]:
+    """calculates the shape coefficients for optimal LAI formula"""
+    if not 0 < first_heat_fraction:
+        raise ValueError("first_heat_fraction must be greater than 0")
+    if not 0 < second_heat_fraction:
+        raise ValueError("second_heat_fraction must be greater than 0")
+    if not 0 < first_leaf_fraction < 1:
+        raise ValueError("first_leaf_fraction must not be greater than 0 or less than 1")
+    if not 0 < second_leaf_fraction < 1:
+        raise ValueError("second_leaf_fraction must not be greater than 0 or less than 1")
+    if first_heat_fraction == second_heat_fraction:
+        # TODO: perhaps a way to handle this instead of throwing an error would be better
+        #   something like: second_heat_fraction += 1e-9
+        raise ValueError("first_heat_fraction cannot be exactly equal to second_heat_fractions")
 
 
+    # TODO: add error handling (see test_error_calc_shape_log() for conditions that fail)
 
+    first_log = calc_shape_log(first_heat_fraction, first_leaf_fraction)
+    second_log = calc_shape_log(second_heat_fraction, second_leaf_fraction)
+
+    second_shape = (first_log - second_log) / (second_heat_fraction - first_heat_fraction)
+    first_shape = first_log + (second_shape * first_heat_fraction)
+
+    return [first_shape, second_shape]
+
+def calc_max_leaf_area_change(leaf_area_fraction, previous_leaf_area_fraction,
+                              max_leaf_area_index, previous_leaf_area_index):
+    """
+    calculates the maximum leaf area added during the day
+
+    Args:
+        leaf_area_fraction: optimal leaf area fraction for the day
+        previous_leaf_area_fraction: previous day's optimal leaf area fraction
+        max_leaf_area_index: the maximum leaf area index achievable by the plant
+        previous_leaf_area_index: the previous day's leaf area index
+
+    Returns:
+        the maximum leaf area added during the day
+
+    Details: because actual leaf area index (LAI) is corrected for growth constraints, the previous
+    day's optimal leaf area fraction may not be the same as the previous day's LAI divided by the
+    max LAI.
+    """
+    return (leaf_area_fraction - previous_leaf_area_fraction) * max_leaf_area_index * \
+           (1 - exp(5 * previous_leaf_area_index - max_leaf_area_index))
 
 def calc_optimal_leaf_area_fraction(heat_fraction: float, shape1: float, shape2: float) -> float:
     """calculates leaf area index fraction, from the optimal leaf area development curve, for the initial period of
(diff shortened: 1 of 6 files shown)


SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
```diff
@@ -0,0 +1,204 @@
...
+            equal to heat_fraction
+
+    Details: used by calc_shape_parameters, where errors are handled
+    """
+    return log((heat_fraction/leaf_area_fraction) - heat_fraction)
+
+def calc_shape_parameters(first_heat_fraction: float, second_heat_fraction: float,
+                          first_leaf_fraction: float, second_leaf_fraction: float) -> list[float]:
+    """calculates the shape coefficients for optimal LAI formula"""
+    if not 0 < first_heat_fraction:
```
ew3361zh: I'm curious if you could use `if first_heat_fraction <= 0:` instead here for easier comprehension vs how it's written? Or is there a particular intention with how this is written?

ew3361zh (approved):
I think this looks good overall. Well organized, good doc strings that made this accessible and understandable to someone without a crop and soil background. I also appreciated that you broke up some of the equations into more digestible chunks so that not too much was happening in any one function. Good test suite. I think most of the additional testing I thought would be helpful is listed in your TODOs. I had one comment about the readability of how a few if statements were written but otherwise this is great!

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
morrowcj: Haha, no good reason at all! I have no clue why I did this. I'll fix it

morrowcj: @ew3361zh, how do you feel about line 148? In words I'd say "if second leaf fraction is not between 0 and 1, do X":
```python
if not 0 < second_leaf_fraction < 1:
```

but I guess it could be:
```python
if second_leaf_fraction <= 0 or second_leaf_fraction >=1:
```

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
```diff
@@ -0,0 +1,204 @@
...
+        self.second_heat_fraction_point = 0.50
+        self.first_leaf_fraction_point = 0.01
+        self.second_leaf_fraction_point = 0.95
+        self.max_canopy_height = 2.5  # m
+        self.growth_factor = 1.0
+        self.max_leaf_area_index = 3.0
+        self.senescent_heat_fraction = 0.9
+        # variable attributes (change throughout simulations)
+        self.leaf_area_index = 0
+        self.heat_fraction = 0.73
```
morrowcj, line 21: @ew3361zh, one question here. I've gone back and forth about documenting attributes in each of the inherited parent classes or in the eventual child class `Crop`. On one hand, documenting the arguments in parents would be very convenient and would prevent the need to re-declare all attributes in the child. On the other hand, some arguments (e.g., `heat_fraction`) are used by multiple parent classes, so the risk of duplicating documentation and having conflicting documentation for the same attribute is higher.

```suggestion
        self.heat_fraction = 0.73
        """total heat units accumulated by the plant, as a proportion of potential heat units"""
```

What do you think?

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
ew3361zh: My personal preference is to try to avoid using `not` in these types of statements just because my brain has to always do a second step but I understand that second line is a lot longer. If I were writing that line I'd prefer the second (i.e. `if second_leaf_fraction <= 0 or second_leaf_fraction >=1:`) because I can look at it and instantly know what it's asking without a second thought.

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
ew3361zh: I believe there are ways to cross reference docstrings between parent and child classes with Sphinx documentation and there are also ways to write decorator functions that you can use to manually accomplish the same thing but I don't have a lot of experience with either unfortunately and I'm not sure how far down that path you want to go. Maybe a good question for Pooya?

This is the example of the decorator function I've seen pop up a few times.
https://softwareengineering.stackexchange.com/a/386758

README.md
```diff
@@ -1,21 +1,30 @@
...
-|:---:|:---:|
-|Jan 31st - Feb 6th|[Lesson 1](https://www.youtube.com/watch?v=7EmboKQH8lM&list=PLs4sTjbm8kLhbJy-rT4DILg-kKw3ZCwDx)|
-|Feb 7th - Feb 13th|[Lesson 2](https://www.youtube.com/watch?v=2a_ytyt9sf8&list=PLs4sTjbm8kLhbJy-rT4DILg-kKw3ZCwDx)|
-|Feb 14th - Feb 20th|[Lesson 3](https://www.youtube.com/watch?v=Qjywrq2gM8o&list=PLs4sTjbm8kLhbJy-rT4DILg-kKw3ZCwDx)|
-|Feb 21st - Feb 27th|[Lesson 4](https://www.youtube.com/watch?v=58jGpV2Cg50&list=PLs4sTjbm8kLhbJy-rT4DILg-kKw3ZCwDx)|
-|Feb 28th - Mar 6th|[Lesson 5](https://www.youtube.com/watch?v=sn0aFEMVTpA&list=PLs4sTjbm8kLhbJy-rT4DILg-kKw3ZCwDx)|
-|Mar 7th - Mar 13th|[Lesson 6](https://www.youtube.com/watch?v=l-gF0vDhJVI&list=PLs4sTjbm8kLhbJy-rT4DILg-kKw3ZCwDx)|
-
-# 2022 Fixathon
-[Fixathon tasks](https://docs.google.com/spreadsheets/d/1SnY3c4vBXrD9ybHoswJW28FrkG4DWx4fyvotIXA85BM/edit?usp=sharing)
```
morrowcj: I have no idea why this file shows as edited. I think these are the changes implemented by @PooyaHekmati that must have tracked poorly when I remerged with master.

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
ltp23: Hi Clay, can you clarify on what you meant by 'some arguments (e.g., heat_fraction) are used by multiple parent classes'?

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
```diff
@@ -0,0 +1,205 @@
...
+    def __init__(self):
+        # fixed attributes (unchanged during simulations)
+        self.first_heat_fraction_point = 0.15
+        self.second_heat_fraction_point = 0.50
+        self.first_leaf_fraction_point = 0.01
+        self.second_leaf_fraction_point = 0.95
+        self.max_canopy_height = 2.5  # m
+        self.growth_factor = 1.0
+        self.max_leaf_area_index = 3.0
+        self.senescent_heat_fraction = 0.9
```
ltp23, line 18: You can put these hardcoded values into an object of a class called `LeafAreaIndexConfig`.  Then, you can pass one such object into the `__init__()`. This allows external customization beyond the predefined default values. 

```
def __init__(config: LeafAreaIndexConfig = some_default_config):
    self.first_heat_fraction_point = config.first_heat_fraction_point
    ...
```

Then, somewhere else if you want to pass in a custom config object, you can.

```
lai = LeafAreaIndex(config=some_custom_config)
```

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
```diff
@@ -0,0 +1,205 @@
...
+        self.leaf_area_added = min(self.max_leaf_area_change * sqrt(self.growth_factor), self.max_leaf_area_change)
+
+    def add_leaf_area(self):
+        """add new leaf area to the plant
+        SWAT Reference: 5:2.1.18"""
+        self.leaf_area_index = max(0, self.previous_leaf_area_index + self.leaf_area_added)
+
+    def check_for_senescence(self):
+        """check if the plant is in senescence"""
+        self.is_in_senescence = self.heat_fraction > self.senescent_heat_fraction
```
ltp23, line 99: You can convert this function into a `property` to make the check more dynamic (i.e., always using the current states). A property can be accessed as though it were an instance variable. 

```
@property
```

This would allow you to use `self.is_in_senescence` anytime without having to call `self.check_for_senescence()` first.

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
```diff
@@ -0,0 +1,205 @@
...
+    def shift_leaf_area_time(self):
+        """shifts the time window by one step for leaf area attributes"""
+        self.previous_leaf_area_index = self.leaf_area_index
+        self.previous_optimal_leaf_area_fraction = self.optimal_leaf_area_fraction
+
+    def determine_canopy_height(self):
+        """sets the current height of the canopy, in meters
+        SWAT Reference: 5:2.1.14"""
+        self.canopy_height = min(self.max_canopy_height,
+                                 self.max_canopy_height * sqrt(self.optimal_leaf_area_fraction))
```
ltp23, line 68: I see most of the methods here follow the above pattern. So, I just use this as an example.

An alternative pattern is to make it into a method that returns a value. This method can be an instance method or a class method. It can also be either public or 'private' (prefixed with an underscore). I believe you discussed this in the annual meeting also. 

```
@classmethod
def determine_canopy_height2(cls, max_canopy_height: float, optimal_leaf_area_fraction: float) -> float:
    return min(max_canopy_height, max_canopy_height * sqrt(optimal_leaf_area_fraction))
```

Then we assign the return value to `self.canopy_height` in `grow_canopy()`

```
def grow_canopy(self):
    ...
    self.canopy_height = self.determine_canopy_height2(self.max_canopy_height, self.optimal_leaf_area_fraction)
    ...
```

There are some pros and cons to this approach. 
Pros:
- Most or all of the state changes will be in one place (i.e., `grow_canopy()`). The helper functions don't change states. 
- The second advantage derives from the first one. Because the helper functions don't change states, you can call them in a more flexible order. 
- One other good thing is that it is easier to test the helper functions (less mocking and patching). 

Cons:
- The orchestrating function (i.e., `grow_canopy()`) can look busy because of the explicit state changes and spelled-out parameter lists.
- The helper functions have longer parameter lists. This can be thought of as a pro since you know exactly what a function needs to do its job.

ltp23:
Hi Clay, overall I really enjoyed reviewing your code thanks to all the descriptive names of variables and functions as well as all the comments!

Some additional comments:
- It took me a while to realize the SC redesign module is located outside the RUFAS module. Is this your intended location for the module?
- The standalone helper functions near the end of `leaf_area_index.py` can be converted into class methods under the current class or under some other classes. 
- Some unit tests may need more mocking and patching. Otherwise, they look more like integration tests where multiple functions are being tested together. But this can be done in another PR.

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
```diff
@@ -0,0 +1,205 @@
...
+    if second_heat_fraction <= 0:
+        raise ValueError("second_heat_fraction must be greater than 0")
+    if not 0 < first_leaf_fraction < 1:
+        raise ValueError("first_leaf_fraction must not be greater than 0 or less than 1")
+    if not 0 < second_leaf_fraction < 1:
+        raise ValueError("second_leaf_fraction must not be greater than 0 or less than 1")
+    if first_heat_fraction == second_heat_fraction:
+        # TODO: perhaps a way to handle this instead of throwing an error would be better
+        #   something like: second_heat_fraction += 1e-9
+        raise ValueError("first_heat_fraction cannot be exactly equal to second_heat_fractions")
```
ltp23, line 154: I think the `math.isclose()` function may help.

```
if math.isclose(first_heat_fraction, second_heat_fraction, rel_tol=1e-09):
    raise ValueError("first_heat_fraction cannot be exactly equal to second_heat_fractions")
```

ltp23 (approved):

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
morrowcj: Definitely: The main class `Crop` will inherit from `LeafAreaIndex` and another class `HeatUnits` (in fact, more than just these 2). Both parent classes contain processes (methods) that use the `heat_fraction` attribute. Neither of the parent classes are likely to be used by anything - the child `Crop` will almost always be used instead. So the question becomes: where do we document `heat_fraction`? In all parents? In just one of the parent classes, with a cross-reference in the others? Or in the main `Crop` class?

Here's a short example:

```python
# ---- leaf_area_index.py ----
class LeafAreaIndex:
  def __init__(self):
    self.heat_fraction = 0.2
    """fraction of potential heat units accumulated by a plant to date"""
  def leaf_method(self):
    print(f"do something with {self.heat_fraction}")
  # ... more code ...

# ---- heat_units.py ----
class HeatUnits:
  def __init__(self):
    self.heat_fraction = 0.2
    """fraction of potential heat units accumulated by a plant to date"""
  def heat_method(self):
    print(f"do something different with {self.heat_fraction}")
  # ... more code ...

# ---- crop.py ----
class Crop(LeafAreaIndex, HeatUnits):
  def __init__(self, heat_fraction):
    LeafAreaIndex.__init__(self)
    HeatUnits.__init__(self)

    self.heat_fraction = heat_fraction
    """fraction of potential heat units accumulated by a plant to date"""
```

It seems to me like documenting the attribute 3+ times is bad and will lead to contradictions.

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
morrowcj: This is already similar to the plan. These values are simply place-holders. The child class that actually uses all of these attributes gets data set via config file. Here's an simplified example: 

``` python
# ---- crop.py ----
class Crop(LeafAreaIndex):
  def __init__(self, config: dict):
    LeafAreaIndex.__init__(self):
    # set attributes from config file
    for key, val in config.items():
      setattr(self, key, val) 
  # ... more code ...

# ---- config_crop.py ----    
CORN = {
  "heat_fraction": 0.21
  # ... other default values ...
}      

# ---- usage ----
corn = Crop(CORN)
```

This functionality isn't in this PR, but does exist in `SC_redesign` branch and will be implemented in a future PR.

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
morrowcj: That is excellent! Thank you, @ltp23. I've seen some of these but didn't quite understand them. I'll go ahead and implement that.

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
morrowcj: @ltp23, I actually really like this idea. I might play around with using this pattern instead. The main reason I didn't do something like this originally is because of the whole "do one thing" - but I think "update all state variables" can be considered 1 thing. 

One clarification: do you mean `@staticmethod` instead of `@classmethod`? My understanding was that class methods are constructor functions (i.e., return an instance of the class). Is that wrong?

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
morrowcj: I think that as long as the difference is greater than the machine tolerance, there is no real problem (i.e., `abs(first_heat_fraction - second_heat_fraction) > numpy.finfo(float).eps`. Is checking the equality as above not sufficient?

ltp23: This works too.

morrowcj: is that not what `==` is doing?

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
ltp23: Both class and static methods are very similar. One main difference is when you write a class method, you want the first parameter to be the class itself. 

```
class SomeClass:
    def __init__(self):
        self.var1 = 42
        self.var2 = 43

    @classmethod
    def from_dict(cls, data: dict):
        # Is this what you meant by 'constructor'?
        obj = cls()
        obj.var1 = data['var1']
        obj.var2 = data['var2']
        return obj

    @staticmethod
    def some_static_method(a, b):
        # Works like a standalone function, but for organizational purposes,
        # I'd prefer to put in inside a class if the context makes sense.

        # You can call a class method with classname.
        s = SomeClass.some_class_method(a, b)

        return s

    @classmethod
    def some_class_method(cls, a, b):
        # Because we mark this as a class method, the first parameter is the class itself,
        # and is recommended, but not required, to be named 'cls'.
        # It's preferred to use 'cls' in some way, but not required.

        # You can call a static method with 'cls'.
        s = cls.some_static_method(a, b)

        return s

    def some_instance_method(self):
        # It's preferred to use 'self' in some way, but not required.
        # If the method doesn't use 'self', it's probably better to make it a class method.

        # You can call a class method with 'self'.
        sum1 = self.some_class_method(self.var1, self.var2)

        # You can call a static method with 'self' too.
        sum2 = self.some_static_method(self.var1, self.var2)

        return sum1 + sum2


if __name__ == '__main__':
    foo = SomeClass.from_dict({'var1': 1, 'var2': 2})
    print(foo.some_instance_method())  # 6

    s1 = SomeClass.some_static_method(1, 2)  # 3
    s2 = SomeClass.some_class_method(3, 4)  # 7
    print(s1, s2)  # 3 7

```

ltp23: I don't strictly follow one pattern over the other just because consistency is hard. But when there are a lot of helper functions that work as silent setters, it becomes difficult to see what variables are being changed.

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
ltp23: That's great!

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
ltp23: It is unpredictable when comparing two floating-point numbers for equality, even though you know for sure they are equal. The way they are represented internally is not perfect and only good up to a finite number of decimal places. So, comparing the absolute difference to a tolerance level is usually sufficient.

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
morrowcj: This is interesting! Thanks. Also - yes, the first `@classmethod` is what I meant by a constructor (`C++` terminology)

morrowcj:
I merged this PR into SC_redesign in its current state. Ed is working to implement the design pattern that @ltp23 suggested. We will submit another pull request once that is complete.

PR #225, opened 2022-11-30
root development refactor

morrowcj:
This pull request involves the refactoring of the crop root development processes. It is a small PR and involves only a handful of functions.

The main change in this PR is [root_development.py](SC_redesign/Crop_and_Soil/crop/root_development.py) and associated tests: [test_root_development.py](SC_redesign/Crop_and_Soil/tests_SC/test_root_development.py). With this PR, I've used the pattern suggested to me by @ltp23, which I think works well. 

This PR follows #224. Review of this PR should probably wait until #224 is merged into SC_redesign.


SC_redesign/Crop_and_Soil/crop/crop.py
```diff
@@ -4,18 +4,25 @@
 from SC_redesign.Crop_and_Soil.crop.water_dynamics import WaterDynamics
 from SC_redesign.Crop_and_Soil.crop.heat_units import HeatUnits
 from SC_redesign.Crop_and_Soil.crop.leaf_area_index import LeafAreaIndex
-from SC_redesign.Crop_and_Soil.soil.soil import Soil
+from SC_redesign.Crop_and_Soil.crop.root_development import RootDevelopment
+
+from typing import List
 
 # TODO: Should use an ENUM class to represent the supported species??
```
Isgotkowitz, line 11: Sounds like a good idea, but this require extra code to update the ENUM every time a new crop species is added right?

SC_redesign/Crop_and_Soil/crop/root_development.py
```diff
@@ -0,0 +1,58 @@
...
+            return 0.4 - 0.2 * heat_fraction
+
+    @staticmethod
+    def determine_root_depth(max_depth: float, heat_fraction: float) -> float:
+        """calculates a plant's root depth on a given day
+
+        Args:
+            max_depth: maximum possible root depth (mm)
+            heat_fraction: fraction of potential heat units
+
```
Isgotkowitz, line 52:
```suggestion

SWAT Reference: 5:2.1.23, 24

```

Isgotkowitz (commented):
Adding the SWAT reference in `root_development.py` is the only thing I think should be changed, otherwise LGTM

SC_redesign/Crop_and_Soil/crop/crop.py
morrowcj: I'm not sure. I don't fully understand ENUMs yet. It is something I want to learn more about (perhaps you can enlighten me a bit) before moving forward. I did a small amount of reading on the pattern and it seemed useful for what I'm planning. We'll need to re-evaluate later when we start working on the details of the `Crop` class.

SC_redesign/Crop_and_Soil/crop/root_development.py
morrowcj: Good catch, thanks!

Isgotkowitz (approved):

SC_redesign/Crop_and_Soil/crop/crop.py
```diff
@@ -57,6 +62,13 @@ def grow_crop(self, layer_nitrates: list[float], layer_depths: list[float], soil
         self.allocate_biomass(incoming_light)
         self.cycle_water(evaporation, transpiration, max_evapotranspiration)
 
+    def list_all_var_names(self):
```
Isgotkowitz: This function needs a return type

SC_redesign/Crop_and_Soil/crop/crop.py
Isgotkowitz: Ok, I'm not sure how ENUMs work in Python so I will have to do some learning too.

Isgotkowitz (changes requested):

SC_redesign/Crop_and_Soil/crop/crop.py
morrowcj: Done. Thanks

Isgotkowitz (approved):

SC_redesign/Crop_and_Soil/crop/crop.py
```diff
@@ -57,6 +62,13 @@ def grow_crop(self, layer_nitrates: list[float], layer_depths: list[float], soil
         self.allocate_biomass(incoming_light)
         self.cycle_water(evaporation, transpiration, max_evapotranspiration)
 
+    def list_all_var_names(self) -> list:
+        """list all variables used by Crop"""  # TODO: check for duplicates or conflicts among parents
+        return vars(self)
```
ltp23: I believe `vars(obj)` returns a dictionary, so if you want a list, then use `list(vars(obj).keys())` or `list(vars(obj).values())` depending on what makes sense to you.

ltp23 (approved):
Everything looks great! Thanks for your work.

SC_redesign/Crop_and_Soil/crop/crop.py
morrowcj: No, I think returning a dictionary makes more sense. I'll fix it and then merge.

PR #226, opened 2022-12-01
Output manager parameters

### LINKED ISSUE #227: Rename OutputManager params and ensure json data entered is properly syntaxed
This issue serves to address some small fixes and adjustments to the OutputManager class and instances that paired well together for addressing at the same time.

## Expected Behavior
Two main OutputManager params need to be renamed in the OM class constructor and in every instance of the OM in the model so far.
"caller_class" -> "class"
"caller_function" -> "function"
The OutputManager will function using json formatting so all data input should use json syntax which means changing every single quote ' to a double quote ".

## Current Behavior
The current params for OutputManager and all of its methods are "caller_class" and "caller_function". There are currently a number of single quotes in the instances of OutputManager and within its constructor class. 

## Possible Solution
Through search and replace of the params for all the instances of the OM in the model so far including in the class constructor for OutputManager. Additionally doing search and replace for single quotes when used in context of OM instantiation and usage to ensure json data is entered with double quotes.

## Steps to Reproduce
In RUFAS > classes.py

the info_map param is instantiated on line 89:
```
info_map = {'caller_class': self.__class__.__name__, 
                    'caller_function': self.__init__.__name__,
                    'data': data,
                    'weather_file': weather_file,}
```
this should be rewritten as:
```
                    "data": data,
                    "weather_file": weather_file,}
```

1. OutputManager and params are searched for in the model.
2. caller_class and caller_function are rewritten as class and function respectively.
3. All single quotes when entering data to OutputManager are changed to double quotes.

## Context (Environment)
OutputManager requires the class and function names where it is being instantiated and called in the code. Shortening these to more manageable names helps with code clarity and conciseness.
Additionally, single quotes are not acceptable syntax in json files and so need to be changed to double quotes.

ew3361zh:
**What:** 
This PR makes some name-adjustments to OutputManager params ("caller_function" to "function" and "caller_class" to "class") as well as addressing some issues with proper json syntax (single quotes to double quotes).

**Why**
The renaming of the OutputManager params is meant to make them easier to enter while not losing the intent or understanding behind them. The json syntax adjustments are due to json only accepting double quotes around strings and parameter names.

**How**
Through search and replace of the params for all the instances of the OM in the model so far including in the class constructor for OutputManager. Additionally doing search and replace for single quotes when used in context of OM instantiation and usage to ensure json data is entered with double quotes.

### DIFF
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -210,7 +210,7 @@ class OutputManager (object):
 
         return f"{prefix}{name}{suffix}"
 
-    def _get_prefix(self, class: str, function: str) -> str:
+    def _get_prefix(self, class_: str, function: str) -> str:
         """
         Returns the prefix for a key in the pool.
 
@@ -226,7 +226,7 @@ class OutputManager (object):
         str
             {class}.{function}
         """
-        return f"{class}.{function}"
+        return f"{class_}.{function}"
 
     def _get_time_based_suffix(self) -> str:
         """
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -210,13 +210,13 @@ class OutputManager (object):
 
         return f"{prefix}{name}{suffix}"
 
-    def _get_prefix(self, class_: str, function: str) -> str:
+    def _get_prefix(self, caller_class: str, function: str) -> str:
         """
         Returns the prefix for a key in the pool.
 
         Parameters
         ----------
-        class : str
+        caller_class : str
             The name of the class in which this key-value pair is originated
         function : str
             The name of the function in which this key-value pair is originated
@@ -226,7 +226,7 @@ class OutputManager (object):
         str
             {class}.{function}
         """
-        return f"{class_}.{function}"
+        return f"{caller_class}.{function}"
 
     def _get_time_based_suffix(self) -> str:
         """
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -40,8 +40,8 @@ class OutputManager (object):
             self.errors_pool: Dict[str, Any] = {}
             self.logs_pool: Dict[str, Any] = {}
             self.add_log("init_log", "Output Manager instantiated.",
-                         info_map={'caller_class': self.__class__.__name__,
-                                   'caller_function': self.__init__.__name__})
+                         info_map={"class": self.__class__.__name__,
+                                   "function": self.__init__.__name__})
 
     def add_variable(self, name: str, value: Any,
                      info_map: Dict[str, Union[str, bool]]) -> None:
@@ -56,24 +56,24 @@ class OutputManager (object):
             The value of the variable
         info_map : Dict[str, Union[str,bool]]
             Additional args, some are non-optional
-        info_map['caller_class'] : str
+        info_map["class"] : str
             The name of the class which called this function
-        info_map['caller_function'] : str
+        info_map["function"] : str
             The name of the function which called this function 
-        info_map['prefix'] : str, optional
+        info_map["prefix"] : str, optional
             If present, overrides the automated prefix
-        info_map['suffix'] : str, optional
+        info_map["suffix"] : str, optional
             If present, overrides the automated suffix
-        info_map['suppress_prefix'] : bool, optional
+        info_map["suppress_prefix"] : bool, optional
             If present and True, suppresses the automated prefix generation.
             Has no effect on manual prefix overrides.
-        info_map['suppress_suffix'] : bool, optional
+        info_map["suppress_suffix"] : bool, optional
             If present and True, suppresses the automated suffix generation.
             Has no effect on manual suffix overrides.
-        info_map['enforce_override'] : bool, optional
+        info_map["enforce_override"] : bool, optional
             If present and True, overrides the existing value in the pool if a
             key collision happens.
-        info_map['fail_silently'] : bool, optional
+        info_map["fail_silently"] : bool, optional
             If present and True, suppresses raising error if a key collision
             happens.
 
@@ -81,7 +81,7 @@ class OutputManager (object):
         ------
         ValueError
             If a key collision happens in the pool and either
-            info_map['fail_silently'] or info_map['enforce_override'] 
+            info_map["fail_silently"] or info_map["enforce_override"] 
             are not set to True.
         """
         key = self._generate_key(name, info_map)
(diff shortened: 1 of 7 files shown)


morrowcj (approved):
I can't find any problems and completely agree with all the changes. 

One thing I'd like to see implemented in the future, if possible is the ability to automatically detect the "class" and "function" from the caller. So, if `crop.add_output()` calls `add_variable()`, it could automatically set `info_map = {"class": "crop", "function": "add_output"}`

ltp23:
I haven't had a chance to review this PR yet, but I think I can help with implementing Clay's suggestion if Niko is interested in adding that to this PR. I was experimenting with this last week.
- It can be somewhat cumbersome to include an info-map for every call to the OutputManager. One alternative is to make info-map optional and only include it when there is a need to suppress something or use custom prefixes or suffixes. 
- It looks like the only required keys in an info-map are `caller_class` and `caller_function` while the other keys are optional. So, to get those two pieces of information, we could do some call stack inspection as follows.

```
    import inspect

    def _get_caller_class_and_function(self) -> Dict[str, str]:
        stack = inspect.stack()
        f_locals = stack[2].frame.f_locals  # 0 is this function, 1 is the caller, 2 is the caller's caller and so on
        if 'self' in f_locals:
            caller_class = stack[2].frame.f_locals['self'].__class__.__name__
        elif 'cls' in f_locals:
            caller_class = stack[2].frame.f_locals['cls'].__name__
        else:
            raise ValueError("Cannot find caller class. "
                             "If static methods or standalone functions are used, "
                             "please provide caller_class and caller_function in info_map.")
        caller_function = stack[2].frame.f_code.co_name
        return {'caller_class': caller_class, 'caller_function': caller_function}


    def add_variable(self, name: str, value: Any,
                     info_map: Optional[Dict[str, Union[str, bool]]] = None) -> None:
        if not info_map:
            info_map = self._get_caller_class_and_function()
        elif 'caller_class' not in info_map or 'caller_function' not in info_map:
            info_map.update(self._get_caller_class_and_function())

        # ...

```

RUFAS/output_manager.py
```diff
@@ -167,66 +167,66 @@ def add_error(self, name: str, msg: str, info_map: Dict[str, Any]) -> None:
...
-        elif not info_map.get('suppress_suffix', False):
+        if info_map.get("suffix") is not None:
+            suffix = "." + info_map.get("suffix")
+        elif not info_map.get("suppress_suffix", False):
             suffix = "." + self._get_time_based_suffix()
 
         return f"{prefix}{name}{suffix}"
 
-    def _get_prefix(self, caller_class: str, caller_function: str) -> str:
+    def _get_prefix(self, class_: str, function: str) -> str:
```
```suggestion
    def _get_prefix(self, class: str, function: str) -> str:
```

RUFAS/output_manager.py
```diff
@@ -167,66 +167,66 @@ def add_error(self, name: str, msg: str, info_map: Dict[str, Any]) -> None:
...
             The name of the function in which this key-value pair is originated
 
         Returns
         -------
         str
-            {caller_class}.{caller_function}
+            {class}.{function}
         """
-        return f"{caller_class}.{caller_function}"
+        return f"{class_}.{function}"
```
```suggestion
        return f"{class}.{function}"
```

PooyaHekmati (approved):
Thanks Niko for taking this on. There was one little type, please fix it before merging.

PooyaHekmati:
> I haven't had a chance to review this PR yet, but I think I can help with implementing Clay's suggestion if Niko is interested in adding that to this PR. I was experimenting with this last week.
> 
> * It can be somewhat cumbersome to include an info-map for every call to the OutputManager. One alternative is to make info-map optional and only include it when there is a need to suppress something or use custom prefixes or suffixes.
> * It looks like the only required keys in an info-map are `caller_class` and `caller_function` while the other keys are optional. So, to get those two pieces of information, we could do some call stack inspection as follows.
> 
> ```
>     import inspect
> 
>     def _get_caller_class_and_function(self) -> Dict[str, str]:
>         stack = inspect.stack()
>         f_locals = stack[2].frame.f_locals  # 0 is this function, 1 is the caller, 2 is the caller's caller and so on
>         if 'self' in f_locals:
>             caller_class = stack[2].frame.f_locals['self'].__class__.__name__
>         elif 'cls' in f_locals:
>             caller_class = stack[2].frame.f_locals['cls'].__name__
>         else:
>             raise ValueError("Cannot find caller class. "
>                              "If static methods or standalone functions are used, "
>                              "please provide caller_class and caller_function in info_map.")
>         caller_function = stack[2].frame.f_code.co_name
>         return {'caller_class': caller_class, 'caller_function': caller_function}
> 
> 
>     def add_variable(self, name: str, value: Any,
>                      info_map: Optional[Dict[str, Union[str, bool]]] = None) -> None:
>         if not info_map:
>             info_map = self._get_caller_class_and_function()
>         elif 'caller_class' not in info_map or 'caller_function' not in info_map:
>             info_map.update(self._get_caller_class_and_function())
> 
>         # ...
> ```

Thanks, Loi for taking the time and writing this code snippet. I think we should avoid doing this because:

1. It is out of the scope of this PR.
2. info_map needs to be passed anyways, excluding class and function doesn't make it optional. As explained before, info_map is intended to carry runtime env settings so the outputs are more meaningful and reproducible. As shown in the code, info map can be initialized at the beginning of a function with class and function, then new elements can be added to it as necessary. I don't see how and why this is a problem.
3. From a design perspective, this doesn't sound right. Output manager is, and should be, agnostic about things that have happened upstream. In other words, as its name suggests, the output manager's job is to manage the output rather figuring out the dependency and inspect the call stack.
4. This adds one more thing to maintain and fail.
5. This has the potential for problems when we want have multi-thread processing.

RUFAS/output_manager.py
ew3361zh: This was the issue I brought up in Slack. If I use the Python reserved word "class" as an argument name for this function (or as a variable or file name or any capacity outside its main reserved definition), I get a syntax error. This is the error I get when I make the switch:

> E     File "/Users/niko/Desktop/RuFaS Home/MASM/RUFAS/output_manager.py", line 213
> E       def _get_prefix(self, class: str, function: str) -> str:
> E                             ^
> E   SyntaxError: invalid syntax
> 

I believe the go-to adaption if we want to use the word class as a variable/argument/param is "class_":

https://docs.python.org/3/reference/lexical_analysis.html#keywords
https://www.crummy.com/software/BeautifulSoup/bs4/doc/#searching-by-css-class

RUFAS/output_manager.py
ew3361zh: Please see above comment for line 213 suggested change.

PR #228, opened 2022-12-02
Output handler 2/3

PooyaHekmati:
This is the second PR for the output manager, out of 3.
This one implements basic saving functionalities. Implements a flush function to reset pools. 

Some other minor improvements are done as well.


Fjordrunner (approved):

tests/test_misc.py
```diff
@@ -408,5 +403,101 @@ def test_output_manager_singleton(mocker: MockerFixture) -> None:
...
+    mock_output_manager._generate_file_name = MagicMock(
+        return_value="dummy_name")
+    mock_output_manager._dict_to_file_json = MagicMock()
+
+    mock_output_manager.save_errors("dummy_path")
+
+    mock_output_manager._generate_file_name.assert_called_once_with(
+        "errors", "json")
+    mock_output_manager._dict_to_file_json.assert_called_once_with(
+        mock_output_manager.errors_pool, os.path.join("dummy_path", "dummy_name"))
```
morrowcj, line 546: Can we move tests of the output manager to a dedicated file?

PooyaHekmati: Yes, we can, would you elaborate more on why?

ltp23 (approved):

PR #95, opened 2021-05-11
Heifer ration

cvankerkhove:
[Pen Allocation Flow Chart.pdf](https://github.com/RuminantFarmSystems/MASM/files/6480760/Pen.Allocation.Flow.Chart.pdf)
This pull request includes major development from 2 branches (heifer_raiton and v_patches). The Heifer ration development mainly concerns updates to the ration module and optimization process, along with an overhaul of the requirements structure. The v_patches development mainly contributes an updated pen allocation method. A document is attached with a more in depth analysis of the changes made in this pull request. 
[Heifer Ration Pull Request.docx](https://github.com/RuminantFarmSystems/MASM/files/6461773/Heifer.Ration.Pull.Request.docx)

PR #229, opened 2022-12-05
Output pool populate pt2

ew3361zh:
What:
This PR adds some additional warnings, variables, and errors to their respective Output Manager pools.

Why:
The idea is to continue building each of these pools with their respective outputs throughout the model so that Output Manager is capturing a majority of the output the model is producing for use after simulations have run. I somewhat arbitrarily put a stopping point on this branch to make sure the PR is manageable for review.

How:
This group of warnings, variables, and errors continued to focus on the replacement of print statements with Output Manager add methods. So I primarily searched for print statements throughout the model.

### DIFF
--- a/RUFAS/classes.py
+++ b/RUFAS/classes.py
@@ -208,14 +208,16 @@ class Config:
             if self.start_day > year_length:
                 start_date_error_3 = "Start date invalid. Starting simulation on " \
                     f"{self.start_year}:{year_length}"
-                om.add_error("invalid_start_date_3", start_date_error_3, info_map)
+                om.add_error("invalid_start_date_3",
+                             start_date_error_3, info_map)
                 self.start_day = year_length
                 om.add_variable("start_day", self.start_day, info_map)
         else:
             if self.start_day > leap_year_length:
                 start_date_error_4 = "Start date invalid. Starting simulation on " \
                     f"{self.start_year}:{leap_year_length}"
-                om.add_error("invalid_start_date_4", start_date_error_4, info_map)
+                om.add_error("invalid_start_date_4",
+                             start_date_error_4, info_map)
                 self.start_day = leap_year_length
                 om.add_variable("start_day", self.start_day, info_map)
 
@@ -456,7 +458,8 @@ class Weather:
             if skips > 5:
                 weather_csv_error_pt3 = "Only printing first 5 invalid rows, there are" \
                     f" {skips} total invalid rows."
-                om.add_error("weather_csv_error_pt3", weather_csv_error_pt3, info_map)
+                om.add_error("weather_csv_error_pt3",
+                             weather_csv_error_pt3, info_map)
 
             # calculates T_avg_annual for each year
             for i in range(len(years)):
--- a/RUFAS/classes.py
+++ b/RUFAS/classes.py
@@ -178,9 +178,9 @@ class Config:
         # special error statements for start and end years
         if self.start_year == w_start_year and self.start_day < w_start_day \
                 or self.start_year < w_start_year:
-            start_date_error = "Start date invalid. Starting simulation on " \
+            start_date_warning = "Start date invalid. Starting simulation on " \
                 f"{w_start_year}:{w_start_day}"
-            om.add_error("invalid_start_date", start_date_error, info_map)
+            om.add_warning("invalid_start_date", start_date_warning, info_map)
             self.start_day = w_start_day
             om.add_variable("start_day", self.start_day, info_map)
             self.start_year = w_start_year
@@ -188,9 +188,9 @@ class Config:
 
         if self.end_year == w_end_year and self.end_day > w_end_day \
                 or self.end_year > w_end_year:
-            end_date_error = "End date invalid. Ending simulation on " \
+            end_date_warning = "End date invalid. Ending simulation on " \
                 f"{w_end_year}:{w_end_day}"
-            om.add_error("invalid_end_date", end_date_error, info_map)
+            om.add_warning("invalid_end_date", end_date_warning, info_map)
             self.end_day = w_end_day
             om.add_variable("end_day", self.end_day, info_map)
             self.end_year = w_end_year
@@ -199,57 +199,54 @@ class Config:
         # start date errors if the simulation starts before day 1 or after
         # the last possible day of the year
         if self.start_day < 1:
-            start_date_error_2 = "Start date invalid. Starting simulation on " \
+            start_date_warning_2 = "Start date invalid. Starting simulation on " \
                 f"{self.start_year}:1"
-            om.add_error("invalid_start_date_2", start_date_error_2, info_map)
+            om.add_warning("invalid_start_date_2", start_date_warning_2, info_map)
             self.start_day = 1
             om.add_variable("start_day", self.start_day, info_map)
         if not is_leap_year(self.start_year):
             if self.start_day > year_length:
-                start_date_error_3 = "Start date invalid. Starting simulation on " \
+                start_date_warning_3 = "Start date invalid. Starting simulation on " \
                     f"{self.start_year}:{year_length}"
-                om.add_error("invalid_start_date_3",
-                             start_date_error_3, info_map)
+                om.add_warning("invalid_start_date_3",
+                             start_date_warning_3, info_map)
                 self.start_day = year_length
                 om.add_variable("start_day", self.start_day, info_map)
         else:
             if self.start_day > leap_year_length:
-                start_date_error_4 = "Start date invalid. Starting simulation on " \
+                start_date_warning_4 = "Start date invalid. Starting simulation on " \
                     f"{self.start_year}:{leap_year_length}"
-                om.add_error("invalid_start_date_4",
-                             start_date_error_4, info_map)
+                om.add_warning("invalid_start_date_4",
+                             start_date_warning_4, info_map)
                 self.start_day = leap_year_length
                 om.add_variable("start_day", self.start_day, info_map)
 
         # end date errors if the simulation ends before day 1 or after
         # the last possible day of the year
         if self.end_day < 1:
-            end_date_error_2 = "End date invalid. Ending simulation on " \
+            end_date_warning_2 = "End date invalid. Ending simulation on " \
                 f"{self.end_year}:1"
-            om.add_error("invalid_end_date_2", end_date_error_2, info_map)
+            om.add_warning("invalid_end_date_2", end_date_warning_2, info_map)
             self.end_day = 1
             om.add_variable("end_day", self.end_day, info_map)
         if not is_leap_year(self.end_year):
             if self.end_day > year_length:
-                end_date_error_3 = "End date invalid. Ending simulation on " \
+                end_date_warning_3 = "End date invalid. Ending simulation on " \
                     f"{self.end_year}:{year_length}"
-                om.add_error("invalid_end_date_3", end_date_error_3, info_map)
+                om.add_warning("invalid_end_date_3", end_date_warning_3, info_map)
                 self.end_day = year_length
                 om.add_variable("end_day", self.end_day, info_map)
         else:
             if self.end_day > leap_year_length:
-                end_date_error_4 = "End date invalid. Ending simulation on " \
+                end_date_warning_4 = "End date invalid. Ending simulation on " \
                     f"{self.end_year}:{leap_year_length}"
-                om.add_error("invalid_end_date_4", end_date_error_4, info_map)
+                om.add_warning("invalid_end_date_4", end_date_warning_4, info_map)
                 self.end_day = leap_year_length
                 om.add_variable("end_day", self.end_day, info_map)
 
         # checks that start date is not after end date
         if self.start_year > self.end_year \
                 or (self.start_year == self.end_year and self.start_day > self.end_day):
-            # TODO determine if this below raised error should be rewritten to reassign 
-            # either the the start or end year to then allow the simulation to continue 
-            # instead of crashing.
             raise errors.JSONfileData(
                 "CONFIG", "\tThe start date must be before the end date")
 
--- a/RUFAS/classes.py
+++ b/RUFAS/classes.py
@@ -153,20 +153,20 @@ class Config:
 
             # compares actual size of the file to expected size
             if file_line - 1 != expected_w_file_size + 1:
-                dates_not_matching_size_warning = "Start and end dates of the Weather CSV file do not match the size."
+                dates_not_matching_size = "Start and end dates of the Weather CSV file do not match the size."
                 om.add_warning("dates_not_matching_size",
-                               dates_not_matching_size_warning,
+                               dates_not_matching_size,
                                info_map)
                 if file_line - 1 > expected_w_file_size + 1:
                     info_map["weather_full_path.name"] = weather_full_path.name
-                    duplicate_days_warning = f"There may be duplicate days in {weather_full_path.name}"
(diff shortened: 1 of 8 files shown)


RUFAS/routines/animal/ration/pyomo_solver.py
```diff
@@ -167,9 +167,7 @@ def create_model(feeds_data, req_data, feeds):
                 "req_data": req_data,
                 "feeds": feeds,}
 
-    om.add_variable("m1", 
-                    m1,
-                    info_map)  
+    om.add_variable("m1", m1, info_map)
```
JoeWaddell, line 171: This seems like a variable name we might need to modify to be more descriptive (that said, the info_map provides additional information). 

As general practice, if we're unsure what the more descriptive name should be, should we add a # TODO, or should it be a direct question to reviewers?

JoeWaddell (approved):
Looks good to me, only left a single comment with a fairly general question. I really appreciate how straightforward the method for adding variables/errors/warnings is.

RUFAS/routines/animal/ration/pyomo_solver.py
ew3361zh: I think this is a great point, Joe. I think the latter would be my preferred option so that we're addressing it before it gets merged (and likely forgotten about). I'll take a look and see if there's an easy adjustment to the name that would better explain it or if you have a minute and could look I'd welcome that.

ew3361zh: ok, one issue with this particular instance is that the pyomo_solver module and its functions aren't used anywhere else in the model. It looks like it's involved with ration_driver.py but the imported code and function references have all been commented out ([and was done so to allow the model to successfully run back in March 2021](https://github.com/RuminantFarmSystems/MASM/pull/95/commits/7f07d7b8b4c3c73305bdd8b2e8082d9aedbcee84))

@JoeWaddell do you have any sense if any part of pyomo_solver is still in use? Has it come up with any of the work you've been doing or could you check with Edward to see if he is using it at all?

I'm wondering if I may go back on my above preference and say in this instance maybe we leave it in as m1 but add a TODO. Definitely open to discussion if others have thoughts.

JoeWaddell: Exactly! I'm working on the Ration module, and I think the pyomo stuff has been superseded by the use of non-linear programming. I'll be talking to Edward and Kristan in detail on some ongoing questions I have about this section of the code today and tomorrow (we're implementing a user-defined ration). 
This isn't the only "dead end" in this part of code that requires some careful thought - because sometimes I don't know if it's a placeholder for future work, or simply something that was part of the development of the current code. Simon is leading a nice effort on this end to use GitHub to comment and ask questions about parts of the code like this, so that should be a nice tool moving forward.

ew3361zh: Maybe I'll just add a TODO statement since it seems like this block could go a lot of different ways and it's hard to get a sense of what it is or how it might be used in its current state.

RUFAS/classes.py
```diff
@@ -152,14 +153,21 @@ def __init__(self, data, weather_file):
 
             # compares actual size of the file to expected size
             if file_line - 1 != expected_w_file_size + 1:
-                #TODO add warnings to warning pool for each of these
-                print("Start and end dates of the Weather CSV file do not match the size.")
+                dates_not_matching_size = "Start and end dates of the Weather CSV file" \
+                    " do not match the size."
```
morrowcj, line 157: Question for my own learning: Is this preferred way to break up along string? I thought you needed a `+` to concatenate two strings with python, as in:
```suggestion
                dates_not_matching_size = "Start and end dates of the Weather CSV file" + \
                    " do not match the size."
```

RUFAS/classes.py
```diff
@@ -86,11 +87,11 @@ def __init__(self, data, weather_file):
             weather_file: path to the weather file specified in the json file
         """
 
-        info_map = {"class": self.__class__.__name__, 
+        info_map = {"class": self.__class__.__name__,
                     "function": self.__init__.__name__,
                     "data": data,
```
morrowcj, line 92: It seems to me, based on how I understand `info_map` (which may still be poorly), that a data structure like `data` should not be included. Is that not right? Info *about* the data makes more sense to me than the actual data.

RUFAS/classes.py
```diff
@@ -86,11 +87,11 @@ def __init__(self, data, weather_file):
...
-                    "weather_file": weather_file,}
-        
+                    "weather_file": weather_file, }
```
morrowcj: It seems like `weather_file` is not actually a file, but a path string. Can we change the mapping key (and the name of the value object, if possible)?
```suggestion
                    "weather_path": weather_file, }
```

RUFAS/classes.py
morrowcj: Or is this a naming issue only? Is `data` just a dictionary containing file path *pointing* to the data? If so, I think a name change is in order.

RUFAS/classes.py
JoeWaddell: Great question. I've been using it as a line continuation character like in Niko's code here (since it seems equivalent to using the '+' operator), but it would be good to know if there's edge cases where it could lead to problems.

ew3361zh: I was unsure about the + being necessary for the newline continuation of a string but it appears like it makes no difference:
Screenshot 2022-12-06 at 1 36 27 PM

Quick addendum via Pooya -

"+ only works for strings, \ is used to breakdown lines that cannot be broken down using other techniques. Using + \ adds unnecessary computation overhead and should be avoided. Using one of them is enough"

So I think we should keep it with \ alone.

RUFAS/classes.py
```diff
@@ -170,18 +178,19 @@ def __init__(self, data, weather_file):
...
-            print("Start date invalid. Starting simulation on "
-                  + str(w_start_year) + ":" + str(w_start_day))
+            start_date_error = "Start date invalid. Starting simulation on " \
+                f"{w_start_year}:{w_start_day}"
+            om.add_error("invalid_start_date", start_date_error, info_map)
             self.start_day = w_start_day
-
             om.add_variable("start_day", self.start_day, info_map)
             self.start_year = w_start_year
             om.add_variable("start_year", self.start_year, info_map)
```
morrowcj, line 187: I know this isn't part of why this PR was created, but going through this gave me some additional thoughts about the output manager. 

Here is a good example of how `info_map` leads to unnecessarily long and redundant pools. The entirety of  `info_map` is copied into **each element** of **each pool**. This could result in frustratingly long output. I've included an example below to demonstrate this. It also isn't helpful that a long string of digits is appended to each key (`suppress_sufix=False` by default).

Here's an example:
```python
# Initialize output manager to track warnings within an arbitrary namespace
om = OutputManager()
# create the info map for this namespace
info_map = {"class": "This", "function": "That",
            "optional_mapping": "something annoyingly long.... like... super annoying"}
# add three separate warnings, generated by the namespace
om.add_warning("strike one", "This is your first warning. Please don't do what you did again.", info_map)
om.add_warning("strike two", "This is your second warning. Don't make me warn you agian.", info_map)
om.add_warning("strike three", "That's it, you're outta here!", info_map)
# Demonstrate that the extra mapping is present for every single warning (and variable, etc.) in this namespace
tmp = [element[1]["info_map"]["optional_mapping"] for element in om.warnings_pool.items()]
print(tmp)
# print full warning pool
print(om.warnings_pool)
```

RUFAS/classes.py
```diff
@@ -170,18 +178,19 @@ def __init__(self, data, weather_file):
...
 
         if self.end_year == w_end_year and self.end_day > w_end_day \
                 or self.end_year > w_end_year:
-            print("End date invalid. Ending simulation on "
-                  + str(w_end_year) + ":" + str(w_end_day))
+            end_date_error = "End date invalid. Ending simulation on " \
+                f"{w_end_year}:{w_end_day}"
+            om.add_error("invalid_end_date", end_date_error, info_map)
```
morrowcj: Should this be a warning, since it does not halt the operation?

RUFAS/classes.py
```diff
@@ -190,47 +199,56 @@ def __init__(self, data, weather_file):
         # start date errors if the simulation starts before day 1 or after
         # the last possible day of the year
         if self.start_day < 1:
-            print("Start date invalid. Starting simulation on "
-                  + str(self.start_year) + ":" + str(1))
+            start_date_error_2 = "Start date invalid. Starting simulation on " \
+                f"{self.start_year}:1"
+            om.add_error("invalid_start_date_2", start_date_error_2, info_map)
```
morrowcj: Same question here and on the other start and end date errors.

RUFAS/routines/animal/animal_management.py
```diff
@@ -201,7 +196,11 @@ def init_pens(self, all_pens_data, herd_data):
         manure_separator = "sedimentation"
         manure_storage = "storage_pit"
         animal_combination = None
-        info_map["all_pens"] = self.all_pens
+        info_map = {"class": self.__class__.__name__, 
+                    "function": self.init_pens.__name__,
+                    "all_pens_data": all_pens_data,
+                    "herd_data": herd_data,
```
morrowcj, line 202: are these data or paths?

RUFAS/classes.py
```diff
@@ -190,47 +199,56 @@ def __init__(self, data, weather_file):
...
+                om.add_error("invalid_end_date_4", end_date_error_4, info_map)
                 self.end_day = leap_year_length
                 om.add_variable("end_day", self.end_day, info_map)
 
         # checks that start date is not after end date
         if self.start_year > self.end_year \
                 or (self.start_year == self.end_year and self.start_day > self.end_day):
-            raise errors.JSONfileData("CONFIG", "\tThe start date must be before the end date")
+            raise errors.JSONfileData(
+                "CONFIG", "\tThe start date must be before the end date")
```
morrowcj, line 254: Should this error be logged in the output manager?

morrowcj (approved):
Overall, I think this looks pretty good. I have a few comments about naming and a few others that may be beyond the scope of this PR about the OM more generally. 

Also, don't bother adding OM stuff to anything in the `field/` folder (recursively) right now. Most things will not resemble their current form (for example `crop.py` will be completely rewritten and individual crop species files like `tall_fescue.py` are obsolete) - sorry you looked through them. I should have been more clear before you embarked.

RUFAS/classes.py
morrowcj: Yeah, that makes good sense to me and is useful to know! Thanks for the clarification.

RUFAS/classes.py
ew3361zh: One of the guidelines that was laid out to me with info_map was to capture the arguments for methods with the idea is that the developer might want all the contextual information possible to see what's happening in the system at that point. In this case the "data" variable is holding the dictionary info from the json file selected. So maybe if the developer is looking back at what's going on and why they're getting certain results at this point in the model, knowing the json file selected by the user and what data is in that json file could be helpful to them. 

Once the third part of the Output Manager (what Pooya lovingly refers to as "swimming" in the pools of data we've put together, I think the presentation of the data back to the developer may make it more useful). 

I also think Output Manager will continue to develop as we start to use it and we may decide to rethink what goes into the info_maps. This first time through though I think it's a good idea to start with what might later be considered overkill.

ew3361zh: As far as name changes, it is a reference to that data argument sent to the __init__ method but we could certainly talk about giving it a different key that could help explain what it is. But also the developer will have the caller function and caller class so they could easily reference where this data came from and at what point in the model it was added to the pool.

RUFAS/classes.py
ew3361zh: How about "weather_file_path"?

RUFAS/classes.py
morrowcj: This all makes sense to me. Thanks for elaborating.

RUFAS/classes.py
morrowcj: even better!

ew3361zh: I will make that change. Can you say a little more about changing the variable name for the value as well? Were you thinking for the name in the argument or just for this one instance to be stored under that name in the info_map?

RUFAS/classes.py
ew3361zh: Hopefully this is resolved by our discussion up above.

RUFAS/classes.py
ew3361zh: Yea, I think I'm still learning how best to classify these types of things. I was thinking this would be an error because it's an invalid date that would otherwise break the program if we hadn't had a way to resolve it. With the errors that actually do break the program (like below where there's code to raise an error, Pooya said specifically not to enter those into the pool because the message will show to the user and there's nothing we or they can do about it in the moment).

Whereas I think a warning might be something that wouldn't otherwise break the simulation but is still a heads up that something is off from the normal expected way the model is supposed to run.

Really this one came down to the word "invalid" to me. Do you think you'd switch this to an error? @JoeWaddell what about you?

RUFAS/routines/animal/animal_management.py
ew3361zh: These are the method arguments from line 158 which Pooya said should be included as baseline info_map data. I believe they are both dictionaries (according to the doc strings for the init_pens method).

RUFAS/classes.py
ew3361zh: I originally was adding these but Pooya asked that they not be added. To me this reads like some of the above errors but I think if this error is raised, it's not something that the previous developers felt like the simulation could recover from so it would essentially end the current simulation there. I'm not totally sure why in this case there wasn't a decision made like earlier in the script where if there was an invalid start or end date that one would be adjusted in the moment so the simulation could continue. I think this might be something to put in as an issue if you agree?

RUFAS/classes.py
morrowcj: Sure, I think that it is OK to rename only the dictionary key. Ideally the entire `weather_file` object would be renamed to `weather_file_path` (unless it actually is a connection to the file and not a string), but that would require changing it throughout the entire repo. I'm not sure if such a minor gripe is worth your time (or anyone else's) right now.

RUFAS/classes.py
morrowcj: I think it is, yes! I'll resolve this comment.

RUFAS/classes.py
morrowcj: I think your point is a good one. Can we add a TODO and github issue to address this in the future?

RUFAS/classes.py
morrowcj: Personally, I'd switch it to a warning. But, if only program-breaking situations are considered "errors", then the error pool will never have more than 1 entry, which would make the error pool moot, right? 

I'm also curious to know what @JoeWaddell.

JoeWaddell: Great question, because you're right: if an "error" is strictly something that causes the program to end...by definition there's only one in a given attempt. 

In this case, it's an incorrect input that we're resolving in code: so I think it would qualify as a warning. 

As for the pool being moot: my interpretation is that the _hypothetical/potential_ error pool is large, but in practice an error report will only ever be populated by a single event (if we use the strict definition as above). So that's just restating your question, but I think it makes sense to have a separate final output storing the error message/condition/metadata, despite it being a single event (and ideally not being necessary).

ew3361zh: Thanks guys! I've changed these to warnings and they will also now be added to the warnings pool instead of the errors pool.

PR #203, opened 2022-10-21
Manure management pr

ltp23:

PR #230, opened 2022-12-07
Manure management PR

ltp23:
Hi reviewers,

The purpose of this PR is to update the existing files in the manure module in the master branch to match those in the manure development branch [(link)](https://github.com/RuminantFarmSystems/MASM/pull/203), before I add any more files.

Notable changes: 
- Adding a JSON file that lists all the different manure management scenarios (`manure_management.json`). 
- Moving constants into separate files (the `constants` folder).
- Adjusting the manure data received from the animal module to fit the usage in the manure module (`pen_manure.py`). This step will likely be iterated several times in future PRs until the manure excretion equations from the animal module [(link)](https://github.com/RuminantFarmSystems/MASM/tree/manure_management_pr/RUFAS/routines/animal/manure) are more settled.
- Refactoring the bedding classes and the milking parlor class.

Notes:
- Currently, the `manure_storage` module is still there because it is mainly used in two places - output handler and field management. The goal is to replace it totally with the `manure` module, which is what this PR is part of. 
- The fairly large number of lines of code changed comes mostly from unit tests. 

Estimated review time: 20-30 minutes
Preferred PR closing time: Friday, Dec 9


RUFAS/routines/manure/constants/general_constants.py
```diff
@@ -0,0 +1,29 @@
...
+    LITERS_TO_CUBIC_METERS = 0.001
+    KG_TO_CUBIC_METERS = 0.001
+
+    # Time-related
+    DAYS_PER_YEAR = 365
+    SECONDS_PER_DAY = 86400
+
+    # Density-related
+    WATER_DENSITY_KG_PER_LITER = 0.997
+    WATER_DENSITY_KG_PER_M3 = WATER_DENSITY_KG_PER_LITER * LITERS_TO_CUBIC_METERS
```
morrowcj, line 29: These constants (some of) are likely useful for the whole model. Put them in `MASM/config/`?

RUFAS/routines/manure/manure/pen_manure.py
```diff
@@ -0,0 +1,89 @@
...
+    manure_volume: float = field(init=False)
+    TS: float = 0.0
+    VSd: float = 0.0
+    VSnd: float = 0.0
+    WIP_frac: float = 0.0
+    WOP_frac: float = 0.0
+    P: float = 0.0
+    P_frac: float = 0.0
+    K: float = 0.0
+    CH4: float = 0.0
```
morrowcj: Can we improve these names and group attribute documentation with their declarations (google's sphynx style allows for this)?

Here's my suggestion. I've added brackets to some docstrings asking for clarification and offer a few alternative names on some lines (both shown with `#` comments).

```suggestion
    """
    
    urea: float = 0.0
    """urea concentration [in the manure?], g/L"""   # brackets added
    urine: float = 0.0
    """amount of urine, kg"""
    total_ammoniacal_nitrogen: float = 0.0
    total ammoniacal nitrogen concentration in the manure slurry, g/L
    urine_ammoniacal_nitrogen: float = 0.0
    """[ammoniacal nitrogen concentration in the urine]"""  # brackets - new docstring
    nitrogen: float = 0.0
    """nitrogen in liquid and solid manure, kg"""
    manure_mass: float = 0.0
    """amount of manure, kg"""
    manure_volume: float = field(init=False)
    """volume of manure, m^3"""
    solids_mass: float = 0.0  # alternatively: total_solids: float = 0.0
    """total [manure] solids, kg"""  # brackets added
    degradable_solids: float = 0.0  # alternatively: degradable_volatiles: float = 0.0
    """degradable volatile solids, kg"""
    stable_solids: float = 0.0  # alternatively: stable_volatiles: float = 0.0
    """non-degradable volatile solids, kg"""    
    inorganic_phosphorus_fraction: float = 0.0
    """water-extractable inorganic P fraction, [unitless]"""  # brackets: I don't know if "dimensionless" means what is implied.
    organic_phosphorus_fraction: float = 0.0
    """water-extractable organic P fraction, [unitless]"""  # brackets
    phosphorus: float = 0.0
    """manure phosphorus excretion for manure module input, kg"""
    phosphorus_fraction: float = 0.0
    """phosphorus fraction of manure, [unitless]"""  # brackets
    potassium: float = 0.0
    """potassium in manure, kg"""
    methane: float = 0.0
    """methane emission, g/day"""
```


What do you think, @KFosterReed, @JoeWaddell, and @varrma?

RUFAS/routines/manure/manure/pen_manure.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def __post_init__(self):
+        """Performs any necessary unit conversion after initialization."""
+        self.manure_volume = self.manure_mass / ManureConstants.MANURE_DENSITY
```
morrowcj: Why initialize with `manure_volume` above if it is getting updated no matter what?

RUFAS/routines/manure/manure/pen_manure.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @staticmethod
```
morrowcj: why not `@classmethod`?

RUFAS/routines/manure/beddings/bedding_classes.py
```diff
@@ -25,8 +24,8 @@ class BaseBedding(ABC):
     Attributes:
         bedding_mass_per_day: Amount of bedding needed for each animal per day, kg/animal/day.
         bedding_density: Density of the bedding, kg/m^3.
-        bedding_dry_matter_mass: Dry matter mass of the bedding, kg.
-        bedding_cleaned_frac: Fraction of bedding removed from the barn (0.0 - 1.0).
+        bedding_dry_matter_content: Dry matter mass of the bedding, [0.7 - 1.0].
+        bedding_cleaned_frac: Fraction of bedding removed from the barn [0.7 - 1.0].
```
morrowcj: double-check the units here. Dry matter mass can only be between 0.7 and 1? I'm guessing this should be kg instead (as in previous version)

morrowcj (approved):
# Main comments
I think this looks PR great. Nice job breaking up #203 , by the way. I do have a few minor comments that I'd like you to address, pending feed back from the others that I linked. But I'm going to approve, since all the critical things look good to me. 

# General comments

* I would do something like `mv RUFAS/routines/manure/manure `RUFAS/routines/manure/pen_manure` so that the dir doesn't
have the same name as its parent. It looks like the dir only contains `pen_manure.py`.

* I think the refactoring of the various functions and classes looks great!

* The test suite looks good to me, though I can't really evaluate the different test cases as well as someone who understands the animal stuff a bit better.

* I don't understand the enum classes, I'd be curious to learn more about what they do.

RUFAS/routines/manure/manure/pen_manure.py
morrowcj: Also, why use class variables instead of instance variables?

RUFAS/routines/manure/beddings/bedding_classes.py
ltp23: I will change the comment to say dry matter content instead of mass.

morrowcj: or "fraction"?

RUFAS/routines/manure/manure/pen_manure.py
ltp23: Here, either way works, but I will change it to a class method. Class methods work better with the current OutputManager class.

RUFAS/routines/manure/beddings/bedding_classes.py
ltp23: I think the two terms are equivalent here, so I'm not sure what the preference is. But dry matter content sounds a little better.

RUFAS/routines/manure/manure/pen_manure.py
ltp23: `manure_volume: float = field(init=False)` just means that we declare there is a `manure_volume` attribute but its value won't be assigned/calculated until after all the other attributes have been given a value (hence the name 'post' init). Syntactically, we can't pass in a `manure_volume` value either when creating an object, even if we want to.

RUFAS/routines/manure/beddings/bedding_classes.py
KFosterReed: Dry matter content is a common term in the animal science world. It is a fraction or percent of the total mass but I think it is generally understood. In the definition, we could say something like 'dry matter content of the bedding as a fraction of the total mass'

RUFAS/routines/manure/manure/pen_manure.py
ltp23: Those are in fact instance variables. This is just the syntax for creating a dataclass, which specializes in storing data objects. A dataclass still works as a normal class, though it has some extra advantages.

morrowcj: Oh, OK! Thanks for filling me in.

RUFAS/routines/manure/manure/pen_manure.py
morrowcj: Interesting. Is that a data class thing?

RUFAS/routines/manure/beddings/bedding_classes.py
morrowcj: Ok, this all makes sense. Thanks for filling me in.

RUFAS/routines/manure/manure/pen_manure.py
KFosterReed: Hi Clay and Loi, I had shared these updates to the manure excretion values with Varma back in August shortly before I left for Sardinia and am realizing that I never followed up on them and that they were not incorporated into the current manure composition values. See this commit: [](https://github.com/RuminantFarmSystems/MASM/commit/741c67df81a57435331edee05077b03177641c44)

I didn't go quite as far as Clay in spelling out all of the element abbreviations but I did alter the names for clarity. I think having more clear names is worthwhile and I go back and forth on whether element abbreviations are OK or not. Definitely not for single letter variables but maybe something like manure_P and manure_K? manure_phosphorus is also super clear so I think that is also a good choice

ltp23: Thanks, Clay and Kristan, I'm happy to change the attribute names to make them more readable, but I'd prefer to do that in the next PR, so the changes will propagate down to other parts of the manure module. 

I also like the way you document each attribute after each line of code. I didn't know this would work with Sphinx too. But then there is also the question of styling: do we want all the docstrings in one place and code in another, or alternating between docstring and code? But I will check how both styles look.

RUFAS/routines/manure/constants/general_constants.py
ltp23: I'm flexible on this. Since these constants are only relevant to the `routines` module, we can move the file there. Do you have similar constants in the crop and soil module?

ltp23:
Thank you Clay for your detailed review. 

I have changed the `manure` subfolder to `pen_manure`. 

Enums are pretty easy to understand. Normally, you would just inherit from the standard, built-in Enum class and then list the member values, and you are done. 

Here, I only added one extra intermediate class called DefaultEnum. If you declare a member called DEFAULT, and then when you search for a member by string and it fails, then the DEFAULT member will be returned.

ew3361zh (approved):
This looks great, Loi and I appreaciate you taking the time to break the larger manure module refactoring into more digestible size portions.

I agree with the attribute naming adjustments discussion and defer to Kristan and Clay for their best practices there.

I think the test suite looks comprehensive and the structure is set up well to make testing pretty straight-forward.

I'm curious to know more about how the DefaultBeddingConfigFactory will be used in simulations outside of testing if that's in the plans.

RUFAS/routines/manure/constants/general_constants.py
morrowcj: I don't know if we use all of these, but I do know there are some similar ones that we'll use (pound to kg, for example). 

Maybe it isn't worth doing until we're sure there will be enough need for a global constants file.

ltp23:
Thank you, Niko, for reviewing my code. 

The default config factories are there just in case nothing is specified in the JSON config file. There are two kinds of factories: config factories and component factories. Component factories call config factories to get default config values when none is given externally. I have an input handler to parse the JSON file but it is not included here.

PR #233, opened 2022-12-07
Crop yield refactor

morrowcj:
This sub-PR of #223 refactors the code involving crop yield and production. 

# Main changes

* The `Yields` class was fully re-designed from the old version. It consists of a main method (`obtain_yields()`), which calls numerous static sub-methods. 
* unit tests for the `Yields` class have been updated
* `crop.py` was updated to include inheritance from `Yields`. 

# Minor changes

* added missing SWAT reference to `biomass_allocation.py`
* crop maturity checking has been moved from `heat_units.py` into `yields.py`

# Other notes

* A number of github issues were created as reminders for future PRs. 
* `crop.py` contains unimplemented methods for `cut()` and `kill()` as well as a comment block with the old implementation. 
This functionallity will be addresesed/implemented in a future PR (liekly #238).

---- 
Edit: 
* This PR also includes some changes present in #241. So `heat_units.py` and `test_heat_units.py` can be ignored until #241 is merged. 
* I'd like to get this merged before Jan, if possible. With the holidays, I understand if that can't happen.


SC_redesign/Crop_and_Soil/crop/yields.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def determine_potential_harvest_index(self):
+        """updates the potential harvest index for the day"""
+        self.potential_harvest_index = self.calc_potential_harvest_index(self.heat_fraction, self.optimal_harvest_index)
+
+    def adjust_harvest_index(self):
+        """adjusts the plant's harvest index for """
+        # TODO: how to pass collect warnings from calc_actual_harvest_index and pass them to output manager?
+        self.harvest_index = self.calc_actual_harvest_index(self.potential_harvest_index, self.min_harvest_index,
+                                                            self.water_deficiency)
```
morrowcj: @ew3361zh, @ltp23, or @PooyaHekmati: the static function `calc_actual_harvest_index()` has some input control warnings. I want to collect those warnings when `adjust_harvest_index()` is called (and only then) and pass the message to the output manager. Do any of you know how I can do this?

I want to keep the `warnings.warn()` functionality in the calc function.

PooyaHekmati: would adding a boolean flag work? it controls whether or not the warnings should be reported to om; it has the default value of False, and only adjust_harvest_index sets it to True.

SC_redesign/Crop_and_Soil/crop/yields.py
```diff
@@ -0,0 +1,248 @@
...
+        # TODO: residue also needs to be accumulated in the soil (via Soil class) - GitHub Issue #245
+
+        # TODO: are above- and below-ground lignin residue (percent) needed?
+        #   in the old version, they were both hard-coded to 17 - GitHub Issue #163
+
+    # ---- Sub-Methods ----
+    # TBD
+
+    # ---- Other Methods ----
+    # TBD
```
morrowcj:
```suggestion
```

SC_redesign/Crop_and_Soil/crop/yields.py
morrowcj: The variables passed to output manager will all be handled by `FieldManager` (PR #238)

Isgotkowitz (changes requested):
Everything looks good to me, but before approving I would like to confirm that the is_mature property in PR #241 should be removed, so when both are merged they will stay consistent

Isgotkowitz (approved):

morrowcj:
> Everything looks good to me, but before approving I would like to confirm that the is_mature property in PR #241 should be removed, so when both are merged they will stay consistent

@Isgotkowitz, I pulled the heat units branch into this one and fixed the conflicts. In general, I kept things in the heat units branch, but I did remove references to `is_mature`, which is included in yields now instead. Unless any major updates to the heat units branch occur, we should be good to go now. Feel free to look over my conflict resolution if you'd like.

SC_redesign/Crop_and_Soil/crop/crop.py
```diff
@@ -20,6 +21,7 @@ def __init__(self):
         HeatUnits.__init__(self)
         LeafAreaIndex.__init__(self)
         RootDevelopment.__init__(self)
+        Yields.__init__(self)
```
ltp23, line 25: The design of this part looks quite complex. I guess the intention here is to parcel out the variables into meaningful units corresponding to those smaller classes. 

Instead of using multiple inheritance as a way to pool all the components together, a probably simpler approach is to use composition. 

```
# In crop.py
class Crop:
  def __init__(self):
      self.heat_units = Heat_Units(...)

  def grow_crop(...):
      self.heat_units.absorb_heat_units(...)  # compared to self.absorb_heat_units(...), helpful in tracking down where things are and giving semantic clues as well

```

ltp23 (approved):
Overall, everything looks great. Very well-documented. It definitely motivates me to be more thorough in writing docstrings.

SC_redesign/Crop_and_Soil/crop/crop.py
morrowcj: I don't know enough about composition (is that where objects have components?) but am very interested to learn more. I think it is important to separate each process into their own classes but multiple inheritance has always seemed like a less-than-ideal solution.

PR #234, opened 2022-12-12
LAI rewrote helper methods as static functions inside class

Isgotkowitz:
Moved helper functions from outside the LeafAreaIndex class inside and made them `@staticmethod`'s, and changed tests accordingly

Edit: changed senescence flag to get rid of redundancy

### DIFF
--- a/SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
+++ b/SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
@@ -88,8 +88,6 @@ class LeafAreaIndex:
         """calculates leaf area index fraction, from the optimal leaf area development curve, for the initial period of
         plant growth
 
-        replaces method calc_optimal_leaf_area_fraction
-
         Args:
             heat_fraction: fraction of potential heat units
             shape1: first shape coefficient
--- a/SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
+++ b/SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
@@ -30,7 +30,11 @@ class LeafAreaIndex:
 
     def grow_canopy(self):
         """main leaf area index function"""
-        self.determine_optimal_leaf_area_fraction()
+        # self.determine_optimal_leaf_area_fraction()
+        self._determine_lai_shapes()
+        self.optimal_leaf_area_fraction = calc_optimal_leaf_area_fraction(self.heat_fraction,
+                                                                          self._lai_shapes[0],
+                                                                          self._lai_shapes[1])
         self.determine_canopy_height()
         self.check_for_senescence()
         if self.is_in_senescence:  # senescence
@@ -47,13 +51,25 @@ class LeafAreaIndex:
         self._lai_shapes = calc_shape_parameters(self.first_heat_fraction_point, self.second_heat_fraction_point,
                                                  self.first_leaf_fraction_point, self.second_leaf_fraction_point)
 
-    def determine_optimal_leaf_area_fraction(self):
-        """sets optimal leaf area fraction
-        SWAT Reference: 5:2.1.10"""
-        self._determine_lai_shapes()
-        self.optimal_leaf_area_fraction = calc_optimal_leaf_area_fraction(self.heat_fraction,
-                                                                          self._lai_shapes[0],
-                                                                          self._lai_shapes[1])
+    @staticmethod
+    def determine_optimal_leaf_area_fraction(heat_fraction: float, shape1: float, shape2: float) -> float:
+        """calculates leaf area index fraction, from the optimal leaf area development curve, for the initial period of
+        plant growth
+
+        Args:
+            heat_fraction: fraction of potential heat units
+            shape1: first shape coefficient
+            shape2: second shape coefficient
+
+        Returns:
+            fraction of the plant's maximum leaf area index
+
+        Details: specifically, the calculated value is the 'fraction of the plant's maximum leaf area index corresponding
+        to a given fraction of potential heat units for the plant' (heat_fraction), constrained to be bounded at zero.
+
+        SWAT Reference: 5:2.1.10
+        """
+        return max(heat_fraction / (heat_fraction + exp(shape1 - (shape2 * heat_fraction))), 0)
 
     def shift_leaf_area_time(self):
         """shifts the time window by one step for leaf area attributes"""
@@ -184,21 +200,3 @@ def calc_max_leaf_area_change(leaf_area_fraction, previous_leaf_area_fraction,
     return (leaf_area_fraction - previous_leaf_area_fraction) * max_leaf_area_index * \
            (1 - exp(5 * previous_leaf_area_index - max_leaf_area_index))
 
-def calc_optimal_leaf_area_fraction(heat_fraction: float, shape1: float, shape2: float) -> float:
-    """calculates leaf area index fraction, from the optimal leaf area development curve, for the initial period of
-    plant growth
-
-    Args:
-        heat_fraction: fraction of potential heat units
-        shape1: first shape coefficient
-        shape2: second shape coefficient
-
-    Returns:
-        fraction of the plant's maximum leaf area index
-
-    Details: specifically, the calculated value is the 'fraction of the plant's maximum leaf area index corresponding
-    to a given fraction of potential heat units for the plant' (heat_fraction), constrained to be bounded at zero.
-
-    SWAT Reference: 5:2.1.10
-    """
-    return max(heat_fraction / (heat_fraction + exp(shape1 - (shape2 * heat_fraction))), 0)
--- a/SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
+++ b/SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
@@ -20,7 +20,6 @@ class LeafAreaIndex:
         # variable attributes (change throughout simulations)
         self.leaf_area_index = 0
         self.heat_fraction = 0.73
-        self.is_in_senescence = False
         # empty variables
         self._lai_shapes = None
         self.optimal_leaf_area_fraction = None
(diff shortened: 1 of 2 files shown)


SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
```diff
@@ -94,112 +163,44 @@ def add_leaf_area(self):
         SWAT Reference: 5:2.1.18"""
         self.leaf_area_index = max(0, self.previous_leaf_area_index + self.leaf_area_added)
 
-    def check_for_senescence(self):
+    # TODO CHECKME: having this method with an @property decorator and having the is_in_senescence field seems redundant
+    @property
+    def check_for_senescence(self) -> bool:
```
morrowcj: How about this?
```suggestion
    def is_in_senescence(self) -> bool:
```

morrowcj: Also, consider changing line 48. Is the default starting value `False` actually needed?

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
```diff
@@ -31,30 +32,80 @@ def __init__(self):
...
+        self._lai_shapes = self.determine_lai_shapes(self.first_heat_fraction_point, self.second_heat_fraction_point,
+                                                     self.first_leaf_fraction_point, self.second_leaf_fraction_point)
+
+        self.optimal_leaf_area_fraction = self.determine_optimal_leaf_area_fraction(self.heat_fraction,
+                                                                                    self._lai_shapes[0],
+                                                                                    self._lai_shapes[1])
+
         self.determine_canopy_height()
-        self.check_for_senescence()
+        self.is_in_senescence = self.check_for_senescence
```
morrowcj: I believe that `@property` functions don't need to be assigned to variables.
```suggestion
        self.is_in_senescence
```

SC_redesign/Crop_and_Soil/tests_SC/test_leaf_area_index.py
```diff
@@ -273,4 +175,4 @@ def test_determine_leaf_area(heatfrac):
         assert lai.previous_optimal_leaf_area_fraction == optimal_lai
     else:  # senescence
         assert lai.is_in_senescence is True
```
morrowcj, line 203: might need to change this line (and 166 above) to reflect my suggestions in previous comments.

morrowcj (changes requested):
There are a few small issues with the is_in_senescence property that I think need to be addressed.

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
morrowcj: I actually think I was wrong - after looking at the code in its entirety, I think this line can just be deleted. The next line (45) contains the code `if self.is_in_senescence:`, which will automatically check via the `@property` (I think).

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
```diff
@@ -31,30 +32,80 @@ def __init__(self):
...
             self.add_leaf_area()
         self.shift_leaf_area_time()
 
-    def _determine_lai_shapes(self):
-        """sets lai shape parameters for optimal leaf area fraction calculations"""
-        self._lai_shapes = calc_shape_parameters(self.first_heat_fraction_point, self.second_heat_fraction_point,
-                                                 self.first_leaf_fraction_point, self.second_leaf_fraction_point)
+    @staticmethod
+    def determine_lai_shapes(first_heat_fraction: float, second_heat_fraction: float,
+                             first_leaf_fraction: float, second_leaf_fraction: float) -> List[float]:
```
morrowcj: Should this be a "private" function (preceded with an underscore)? I don't think LAI shapes will ever be used outside of the class.

SC_redesign/Crop_and_Soil/tests_SC/test_leaf_area_index.py
Isgotkowitz: Made suggested changes to `is_in_senescence` and all tests passing

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
```diff
@@ -31,30 +32,80 @@ def __init__(self):
...
+            raise ValueError("first_leaf_fraction must not be greater than 0 or less than 1")
+        if not 0 < second_leaf_fraction < 1:
+            raise ValueError("second_leaf_fraction must not be greater than 0 or less than 1")
+        if first_heat_fraction == second_heat_fraction:
+            # TODO: perhaps a way to handle this instead of throwing an error would be better
+            #   something like: second_heat_fraction += 1e-9
+            raise ValueError("first_heat_fraction cannot be exactly equal to second_heat_fractions")
+
+        # TODO: add error handling (see test_error_calc_shape_log() for conditions that fail)
+
```
morrowcj: I'd like to update this todo
```suggestion
        # TODO: need to add any of these errors that get thrown when RuFaS runs to the  `OutputManager`. 
        #    This should probably be done in the `grow_canopy()` function 
        #    I'm still unsure how to do this effectively with warnings raised by static functions. - morrowcj
```

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
```diff
@@ -31,30 +31,79 @@ def __init__(self):
...
-                                                                          self._lai_shapes[0],
-                                                                          self._lai_shapes[1])
+        return [first_shape, second_shape]
+
+    @staticmethod
+    def determine_optimal_leaf_area_fraction(heat_fraction: float, shape1: float, shape2: float) -> float:
+        """calculates leaf area index fraction, from the optimal leaf area development curve, for the initial period of
+        plant growth
+
+        replaces method calc_optimal_leaf_area_fraction
```
morrowcj: No need for this during the redesign, but I appreciate the thoughtfulness.
```suggestion
```

go ahead and remove similar strings throughout.

SC_redesign/Crop_and_Soil/crop/leaf_area_index.py
```diff
@@ -31,30 +31,79 @@ def __init__(self):
...
-    def _determine_lai_shapes(self):
-        """sets lai shape parameters for optimal leaf area fraction calculations"""
-        self._lai_shapes = calc_shape_parameters(self.first_heat_fraction_point, self.second_heat_fraction_point,
-                                                 self.first_leaf_fraction_point, self.second_leaf_fraction_point)
+    @staticmethod
+    def determine_lai_shapes(first_heat_fraction: float, second_heat_fraction: float,
+                             first_leaf_fraction: float, second_leaf_fraction: float) -> List[float]:
+        """
+        calculates the shape coefficients for optimal LAI formula
+        replaces method calc_shape_parameters
```
morrowcj:
```suggestion
```

morrowcj (changes requested):
This is a more thorough review, after the example review that we walked through together. I've left a few more in-line comments on the code and have a few more general comments below. Once these issues are addressed, I'll approve the PR. 

## General comments:

* I think that `determine_canopy_height()` should be converted to a static function. `shift_leaf_area_time()` and
`check_previous_leaf_area_values()`can remain as regular methods. 

* Type declarations:  on line 33, `grow_canopy()` needs a function type declaration (i.e., void function: `-> None`). 
Check the remaining functions as well - `shift_leaf_area_time()`, `determine_canopy_height()`, and
`check_previous_leaf_area_values()` all need typing. Caution: they won't all be void functions if you address the
previous bullet first.

* I know this is nit-picky, but can we organize the code so that the static functions are grouped together at the bottom of
the class?

morrowcj:
I've requested @ltp23 to look over this PR, since it implements some comments that you made on the last PR concerning these files.

morrowcj:
Before we merge this in, I'd also like to get @PooyaHekmati's opinion about the pattern that we've employed in this PR.

Edit: To be clear for Pooya's sake, the pattern I'm referring to is the use of static methods like so: 

```python
from math import exp

class SomeClass:
    def __init__(self):
        self.x = 1
        self.y = None

    def main_function(self) -> None:
        """update attributes and apply any logic by applying static methods"""
        self.x = self.change_x(self.x, 5)
        self.y = self.find_y(0.23)

    @staticmethod
    def change_x(x: float, plus: float) -> float:
        """performs a calculation and outputs a value"""
        return x + plus

    @staticmethod
    def find_y(power: float) -> float:
        """performs a different calculation"""
        return exp(power)
```

morrowcj:
To clarify and expand upon: 
>* Type declarations:  on line 33, `grow_canopy()` needs a function type declaration (i.e., void function: `-> None`).
>Check the remaining functions as well - `shift_leaf_area_time()`, `determine_canopy_height()`, and
>`check_previous_leaf_area_values()` all need typing. Caution: they won't all be void functions if you address the
>previous bullet first.

I simply meant that you'll need to make sure that the static function returns a float (i.e., `determine_canopy_height() -> float:`), while the others are void functions (they return nothing). Changing the typing like this should not affect the testing at all, as far as I know. 

As I'm sure you know, Python is a dynamically typed language by default but we want to suggest stricter typing, which is done by adding these type annotations to function declarations:

```python
def void_function(float_arg: float, str_arg: str, int_arg: int, bool_arg: bool, default_float_arg: float = 1.0) -> None:
  return None
```

Isgotkowitz:
> To clarify and expand upon:
> 
> > * Type declarations:  on line 33, `grow_canopy()` needs a function type declaration (i.e., void function: `-> None`).
> >   Check the remaining functions as well - `shift_leaf_area_time()`, `determine_canopy_height()`, and
> >   `check_previous_leaf_area_values()` all need typing. Caution: they won't all be void functions if you address the
> >   previous bullet first.
> 
> I simply meant that you'll need to make sure that the static function returns a float (i.e., `determine_canopy_height() -> float:`), while the others are void functions (they return nothing). Changing the typing like this should not affect the testing at all, as far as I know.
> 
> As I'm sure you know, Python is a dynamically typed language by default but we want to suggest stricter typing, which is done by adding these type annotations to function declarations:
> 
> ```python
> def void_function(float_arg: float, str_arg: str, int_arg: int, bool_arg: bool, default_float_arg: float = 1.0) -> None:
>   return NULL
> ```

Ok, I understand

morrowcj (approved):
This looks excellent! Great first PR! Thanks for your patience and let me know if you have any feedback about my reviewing process (now or in the future).

ltp23 (approved):
Overall, it looks very good and clean.

PR #236, opened 2022-12-12
Manure management

ltp23:
Main changes:
- Changing manure attribute names. In this PR only those in the manure module are changed, and those in the animal module will be a separate PR.
- `gas_emissions.py`: A list of class methods for gas equations. 
- `liquid_manure_portion_protocol` and `manure_handler_daily_output`: A protocol is just a list of core, expected attributes, and any class that follows such protocol needs to have all those attributes at a minimum. I went back and forth between using inheritance or using protocol, but I opted for the latter mainly for flexibility reasons. 
- `manure_handler_classes.py`: An enum class (`ManureHandlerType`) that lists all the handler types; `BaseManureHandler` and all its subclasses; `ManureHandlerConfig` class to store handler config data; and `ManureHandlerFactory` class to create handler objects.

---

One note about the general design of the three main manure components - handlers (this PR), separators, and treatments (later PRs). 
- To create an object of a specific component class, the preferred way is to call the corresponding factory's  `get_instance()` method. 
```
flush_system = ManureHandlerFactory.get_instance(ManureHandlerType.FLUSH_SYSTEM, other_args)
screw_press = ManureSeparatorFactory.get_instance(ManureSeparatorType.SCREW_PRESS, other_args)
slurry_storage_outdoor = ManureTreatmentFactory.get_instance(ManureTreatmentType.SLURRY_STORAGE_OUTDOOR, other_args)

# More generally,
handler = ManureHandlerFactory.get_instance(handler_type, other_args)
separator = ManureSeparatorFactory.get_instance(separator_type, other_args)
treatment = ManureTreatmentFactory.get_instance(treatment_type, other_args)
```

- Alternatively, you could create the same object by calling the specific class' constructor, but it is preferable to hide or delegate the specifics of object creation like above.
```
flush_system = FlushSystem(other_args)
```

---

Estimated review time: 45-60 minutes
Preferred PR closing time: Friday, Dec 16

### DIFF
--- a/RUFAS/routines/manure/gas_emissions/gas_emissions.py
+++ b/RUFAS/routines/manure/gas_emissions/gas_emissions.py
@@ -1,0 +1,276 @@
+import math
+
+from RUFAS.routines.manure.constants.gas_emission_constants import GasEmissionConstants
+from RUFAS.routines.manure.constants.general_constants import GeneralConstants
+from RUFAS.routines.manure.constants.manure_constants import ManureConstants
+
+
+class GasEmissions:
+    @classmethod
+    def calc_E_CH4_slurry_storage(cls,
+                                  TS: float,
+                                  enclosed=False,
+                                  tempC=15.0,
+                                  VS_frac=GasEmissionConstants.DEFAULT_VOLATILE_SOLIDS_FRACTION,
+                                  n_eff=0.99) -> float:
+        """Calculates methane emissions from manure storage using total solids and manure kg.
+
+        Args:
+            TS: Total solids, kg.
+            enclosed: True if manure storage is enclosed, and False if manure storage is open to air.
+            tempC: temperature, C.
+            VS_frac: Fraction (0-1) volatile solids.
+            n_eff: efficiency of process              TODO: confirm n_eff meaning
+
+        Returns:
+            CH4 emissions from storage, kg CH4/day.
+
+        """
+        c = 0.024
+        VS_tot = TS * VS_frac
+
+        constants = GasEmissionConstants
+        b1 = constants.b1
+        b2 = constants.b2
+        lnA = constants.lnA
+        E = constants.E
+        R = constants.R
+        Bo = constants.Bo
+        E_CH4_pot = constants.POTENTIAL_METHANE_YIELD_OF_MANURE
+
+        tempK = cls._convert_tempC_to_tempK(tempC)
+        ex = math.exp(lnA - (E / (R * tempK)))
+
+        VSd = (Bo / E_CH4_pot)
+        VSnd = 1 - VSd
+        E_CH4_open_air = c * VS_tot * (VSd * b1 + VSnd * b2) * ex
+
+        if not enclosed:
+            return E_CH4_open_air
+        else:
+            return E_CH4_open_air * (1 - n_eff)
+
+    @classmethod
+    def _calc_modified_hours(cls, hours: float) -> float:
+        """Calculates modified hours."""
+
+        if hours > 14:
+            modified_hours = - math.tanh(hours - 21.5) / 3.5
+        elif hours > 4:
+            modified_hours = math.tanh(hours - 9.5) / 2.5
+        else:
+            modified_hours = - math.tanh(hours + 3.5) / 3.5
+
+        return modified_hours
+
+    @classmethod
+    def _calc_ambient_temp(cls, hours: float, t_min: float, t_max: float) -> float:
+        """Calculates ambient temperature.
+
+        Args:
+            hours: hours of the day from 1 to 24.
+            t_min: Minimum barn temperature, C.
+            t_max: Maximum barn temperature, C.
+
+        Returns:
+            Ambient temperature, °C.
+
+        """
+        modified_hours = cls._calc_modified_hours(hours)
+        t_ambient = modified_hours * (t_max - t_min) / 2 + (t_max + t_min) / 2
+        return t_ambient
+
+    @classmethod
+    def calc_E_CH4_housing(cls, num_animals: int, barn_area: float, hours=24, t_min=20.0, t_max=25.0) -> float:
+        """Calculates methane housing emissions.
+
+        Args:
+            num_animals: Number of animals in the pen.
+            barn_area: Area of the barn based on housing type, m^2.
+            hours: hours of the day from 1 to 24.
+            t_min: Minimum barn temperature, C.
+            t_max: Maximum barn temperature, C.
+
+        Returns:
+            Methane floor emissions, kg CH4/day.
+
+        """
+        t_ambient = cls._calc_ambient_temp(hours, t_min, t_max)
+        t = max(-5.0, 0.63 * t_ambient + 6.0)
+        return num_animals * max(0.0, 0.13 * t) * barn_area / 1000
+
+    @classmethod
+    def calc_E_CO2_housing(cls, num_animals: int, barn_area: float, hours=24, t_min=20.0, t_max=25.0) -> float:
+        """Calculates carbon dioxide housing emissions.
+
+        Args:
+            num_animals: Number of animals in the pen.
+            barn_area: Area of the barn based on housing type, m^2.
+            hours: hours of the day from 1 to 24.
+            t_min: Minimum barn temperature, C.
+            t_max: Maximum barn temperature, C.
+
+        Returns:
+            Carbon dioxide floor emissions, kg CO2/day.
+
+        """
+        t_ambient = cls._calc_ambient_temp(hours, t_min, t_max)
+        t = max(-5.0, 0.63 * t_ambient + 6.0)
+        return num_animals * max(0.0, 0.0065 + 0.0192 * t) * barn_area / 1000
+
+    @classmethod
+    def calc_E_NH3_emission(cls, num_animals: int,
+                            barn_area: float,
+                            urine_TAN: float,
+                            urine: float,
+                            tempC: float,
+                            hsc=GasEmissionConstants.DEFAULT_HOUSING_SPECIFIC_CONSTANT) -> float:
+        """Calculates NH3 storage emissions.
+
+        Args:
+            num_animals: Number of animals in the pen.
+            barn_area: surface area for treatment, m^2.
+            urine_TAN: total ammoniacal nitrogen in manure urine, kg N.
+            urine: total amount of manure urine in exposed surface area, kg.
+            tempC: temperature, C.
+            hsc: housing specific constant, s/m.
+
+        Returns:
+            NH3 storage emissions, kg N/day.
+
+        """
+        p = ManureConstants.MANURE_DENSITY  # kg/m^3
+        pH = 7.5
+        c = GeneralConstants.SECONDS_PER_DAY  # s/day
+        tempK = cls._convert_tempC_to_tempK(tempC)  # K
+        r = cls._calc_r_barn(tempC, hsc)
+        M = urine / barn_area  # manure per area of exposed surface, kg/m^2
+        Q = cls._calc_Q(tempK, pH)
+        if r * M * Q > 0:
+            return num_animals * barn_area * ((urine_TAN / barn_area) * c * p) / (r * M * Q)
+        else:
+            return 0.0
+
+    @classmethod
+    def _calc_r_barn(cls, tempC: float, hsc=GasEmissionConstants.DEFAULT_HOUSING_SPECIFIC_CONSTANT) -> float:
+        """Calculates barn resistance.
+
+        Args:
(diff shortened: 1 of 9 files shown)


RUFAS/routines/manure/pen_manure/pen_manure.py
```diff
@@ -20,39 +20,41 @@ class PenManure:
...
     Attributes:
-        urea: urea concentration in the manure, g/L.
-        urine: amount of urine, kg.
-        TAN: total ammoniacal nitrogen concentration in the manure slurry, g/L.
-        N: nitrogen in liquid and solid manure, kg.
+        manure_urea: manure_urea concentration in the manure, g/L.
+        manure_urine: amount of urine, kg.
+        manure_urine_ammoniacal_nitrogen: ammoniacal nitrogen concentration in the urine, g/L.
+        manure_total_ammoniacal_nitrogen: total ammoniacal nitrogen concentration in the manure slurry, g/L.
+        manure_nitrogen: nitrogen in liquid and solid manure, kg.
```
morrowcj, line 27: I really appreciate the effort that you put in to give the attributes more descriptive names. I think you could drop "manure" from some of these, though. I think that `urea`, `urine`, `total_ammoniacal_nitrogen`, etc. are sufficient since they already come from the "manure" module.

`manure_volume` and `manure_mass` should probably stay as they are, since `volume` and `mass` may be a bit vague, but most of the other variables throughout this module would be clear without it.

This is just a minor suggestion, I don't have strong feelings about it. It is already so much better!

RUFAS/routines/manure/gas_emissions/gas_emissions.py
```diff
@@ -0,0 +1,292 @@
...
+
+        Args:
+            manure_total_solids: Total solids, kg.
+            is_enclosed: True if manure storage is is_enclosed, and False if manure storage is open to air.
+            tempC: temperature in Celsius, C.
+            manure_volatile_solids_fraction: Fraction (0-1) volatile solids.
+            efficiency_fraction: efficiency of process, unitless.
+
+        Returns:
+            CH4 emissions from storage, kg CH4/day.
```
morrowcj, line 27: I think the documentation could be improved for many of these functions. Some descriptions are a bit unclear, and argument names could be also be slightly improved. I'll use this function as an example:

```suggestion
    def calc_methane_emission_for_slurry_storage(cls,
                                                 total_solids: float,
                                                 is_enclosed=False,
                                                 temperature=15.0,   # if units are required in name: temperature_celsius?
                                                 volatile_solids_fraction=(GasEmissionConstants.DEFAULT_VOLATILE_SOLIDS_FRACTION),
                                                 efficiency_fraction=0.99) -> float:
        """Calculates methane emissions from manure storage using total solids.

        Args:
            total_solids: Total solids in the manure, kg.
            is_enclosed: True if manure storage is is_enclosed, and False if manure storage is open to air.
            temperature: temperature in degrees Celsius, C.
            volatile_solids_fraction: Fraction (0-1) of volatile solids **(fraction of the manure or fraction of total solids?)**.
            efficiency_fraction: efficiency of  process **(which process? "methane retention efficiency"?)**, unitless.

        Returns:
            methane emissions from storage, kg/day.
```

RUFAS/routines/manure/gas_emissions/gas_emissions.py
```diff
@@ -0,0 +1,292 @@
...
+        E_CH4_open_air = c * VS_tot * (VSd * b1 + VSnd * b2) * ex
+
+        if not is_enclosed:
+            return E_CH4_open_air
+        else:
+            return E_CH4_open_air * (1 - efficiency_fraction)
+
+    @classmethod
+    def _calc_modified_hours(cls, hours: float) -> float:
+        """Calculates modified hours.
```
morrowcj, line 56: What **are** modified hours?

RUFAS/routines/manure/gas_emissions/gas_emissions.py
```diff
@@ -0,0 +1,292 @@
...
+        elif hours > 4:
+            modified_hours = math.tanh(hours - 9.5) / 2.5
+        else:
+            modified_hours = - math.tanh(hours + 3.5) / 3.5
+
+        return modified_hours
+
+    @classmethod
+    def _calc_ambient_temp(cls, hours: float, t_min: float, t_max: float) -> float:
+        """Calculates ambient temperature.
```
morrowcj, line 77: Could this be more descriptive? From the code below "ambient temperature" is some sort of corrected average.

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
```diff
@@ -0,0 +1,247 @@
...
+        self.milking_parlor = MilkingParlor()
+
+    def _get_current_day_avg_tempC(self) -> float:
+        """Gets the average temperature of the day, in Celsius.
+
+        Returns:
+            The average temperature of the day, in Celsius.
+
+        """
+        return self.weather.T_avg[self.time.year - 1][self.time.day - 1]
```
morrowcj, line 69: Soil and Crop does something just like this, and I suspect the other modules do as well. I wonder if we should add some methods to the Weather and Time (etc.) classes to facilitate these types of operations. Things like 

```python
time_instance.get_current_day()
time_instance.get_current_year()
time_instance.is_leap_year()
weather_instance.get_current_average_temperature()
weather_instance.get_current_min_temperature()
weather_instance.get_current_precipitation()
# etc.
```
and similar functionality would all be very useful.

I don't think you need to do anything for this PR, but maybe we should bring this up to the whole team if you (and others) agree.

morrowcj (approved):
I think all the code looks good. It is clear and well structured. I'm approving this PR.

I did have a few minor comments about documentation: There are a few places where the docstrings could be clearer. Perhaps a subject matter expert should go through and update them. This might be good tasks for the fixathon.

RUFAS/routines/manure/gas_emissions/gas_emissions.py
```diff
@@ -0,0 +1,292 @@
...
+from RUFAS.routines.manure.constants.general_constants import GeneralConstants
+from RUFAS.routines.manure.constants.manure_constants import ManureConstants
+
+
+class GasEmissions:
+    @classmethod
+    def calc_methane_emission_for_slurry_storage(cls,
+                                                 manure_total_solids: float,
+                                                 is_enclosed=False,
+                                                 tempC=15.0,
```
ew3361zh, line 13: Should the value for tempC here be considered a constant instead of hard coded in?

RUFAS/routines/manure/gas_emissions/gas_emissions.py
```diff
@@ -0,0 +1,292 @@
...
+                                                 manure_volatile_solids_fraction=(
+                                                         GasEmissionConstants.DEFAULT_VOLATILE_SOLIDS_FRACTION),
+                                                 efficiency_fraction=0.99) -> float:
+        """Calculates methane emissions from manure storage using total solids.
+
+        Args:
+            manure_total_solids: Total solids, kg.
+            is_enclosed: True if manure storage is is_enclosed, and False if manure storage is open to air.
```
ew3361zh, line 21:
```suggestion
            is_enclosed: True if manure storage is enclosed and False if manure storage is open to air.
```

RUFAS/routines/manure/gas_emissions/gas_emissions.py
ew3361zh: I agree with Clay here, I think some further description of what modified hours is and why it's relevant would be helpful.

RUFAS/routines/manure/gas_emissions/gas_emissions.py
```diff
@@ -0,0 +1,292 @@
...
+        """
+        modified_hours = cls._calc_modified_hours(hours)
+        t_ambient = modified_hours * (t_max - t_min) / 2 + (t_max + t_min) / 2
+        return t_ambient
+
+    @classmethod
+    def calc_methane_housing_emission(cls, num_animals: int, barn_area: float, hours=24, t_min=20.0,
+                                      t_max=25.0) -> float:
+        """Calculates methane housing emission.
+
```
ew3361zh, line 96: I think a little further detailed description on what methane housing emission is would be helpful

RUFAS/routines/manure/gas_emissions/gas_emissions.py
```diff
@@ -0,0 +1,292 @@
...
+        return num_animals * max(0.0, 0.13 * t) * barn_area / 1000
+
+    @classmethod
+    def calc_carbon_dioxide_housing_emission(cls,
+                                             num_animals: int,
+                                             barn_area: float,
+                                             hours=24,
+                                             t_min=20.0,
+                                             t_max=25.0) -> float:
+        """Calculates carbon dioxide housing emissions.
```
ew3361zh, line 119: Similar to the methane housing emission comment, more description of what this is would be helpful

RUFAS/routines/manure/gas_emissions/gas_emissions.py
```diff
@@ -0,0 +1,292 @@
...
+        if r * M * Q > 0:
+            return num_animals * barn_area * ((manure_urine_total_ammoniacal_nitrogen / barn_area) * c * p) / (
+                    r * M * Q)
+        else:
+            return 0.0
+
+    @classmethod
+    def _calc_barn_resistance(cls, tempC: float, hsc=GasEmissionConstants.DEFAULT_HOUSING_SPECIFIC_CONSTANT) -> float:
+        """Calculates barn resistance.
+
```
ew3361zh, line 173: More descriptive explanation of what this is calculating as well.

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
```diff
@@ -0,0 +1,247 @@
...
+
+    Attributes:
+        All inherited from BaseManureHandler.
+
+    """
+    pass
+
+
+class AlleyScraper(BaseManureHandler):
+    """A class that handles calculations related to an alley scraper.
```
ew3361zh, line 154: I'm curious what each of these child classes is doing longer term? Does this add clarity within in the model of whate type of manure handling is happening or lay groundwork for future specific functionality for each child class?

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
```diff
@@ -0,0 +1,247 @@
...
+        cleanings_per_day: Number of cleanings per day.
+
+    """
+    cleaning_water_use_rate: float = 0.0
+    minutes_per_cleaning: int = 8
+    cleanings_per_day: int = 2
+
+
+class DefaultManureHandlerConfigFactory:
+    """Class for creating default manure handler configurations."""
```
ew3361zh, line 179: This is great, Loi, I like the instance factories design pattern and implementation here. Really clearly laid out and I appreciate your note in the PR body about choosing this approach as well.

ew3361zh (approved):
This looks really great, Loi. I added some comments in a few places for clarity and follow up questions but I am approving this as it is right now.

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
ltp23: Currently, the child classes differ from each other only in terms of config parameters, so that can be safely handled in the base class. However, the child classes are still there to handle future specific functionalities.

ltp23:
Hi Clay and Niko, thank you both very much for your reviews. 

Regarding the documentation for the gas equations, I will temporarily add TODOs in those places until I know more to fill those in. Unless I state in the summary description in the next couple PRs, please assume I'm still working on those details with Varma.

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
ltp23: Great suggestion. Can you help me create a git issue for this so we address it? Thanks, Clay.

RUFAS/routines/manure/gas_emissions/gas_emissions.py
ltp23: I will add a constant to the GasEmissionConstant file for this.

RUFAS/routines/manure/gas_emissions/gas_emissions.py
ltp23: One thing I can do right away is to spell out all the abbreviated parameters.

RUFAS/routines/manure/pen_manure/pen_manure.py
ltp23: This sounds good. I will remove the manure- prefix in the attribute names.

PR #241, opened 2022-12-14
Heat units SC redesign

Isgotkowitz:
Redesign of Heat Units and associated tests


SC_redesign/Crop_and_Soil/tests_SC/test_heat_units.py
```diff
@@ -154,9 +109,9 @@ def test_assign_new_heat_units(use_alt, temp):
                      minimum_temperature=15)
     heat.assign_new_heat_units(temp)
     if use_alt or (temp is None):
-        assert heat.new_heat_units == calc_new_heat_units(25, 15)
+        assert heat.new_heat_units == HeatUnits.determine_new_heat_units(25, 15)
     else:
-        assert heat.new_heat_units == calc_new_heat_units(temp, 15)
+        assert heat.new_heat_units == HeatUnits.determine_new_heat_units(temp, 15)
```
morrowcj, line 111: I think you should add assertions for all of the attributes that get changed in `accumulate_heat_units()`

For example, what should the values of `is_growing`, `heat_fraction`, `previous_heat_fraction`, `is_mature` etc. be? you can test for them by adding assertions:

```python
assert heat.new_heat_units == calc_new_heat_units(25, 15)
assert heat.is_growing = x
assert ...
```

morrowcj (changes requested):
Only one small request: add a few assertions to `test_accumulate_heat_units()`

SC_redesign/Crop_and_Soil/tests_SC/test_heat_units.py
Isgotkowitz: `is_growing`, `heat_fraction`, etc. are not touched by `accumulate_heat_units()`, they are only touched in `absorb_heat_units()` which was not being tested , so I wrote a separate test for just that method.

SC_redesign/Crop_and_Soil/tests_SC/test_heat_units.py
```diff
@@ -106,38 +89,22 @@ def test_check_absorb_heat_for_input_errors(use_alt, meant, min_t, max_t):
...
     patch_a = mocker.patch("SC_redesign.Crop_and_Soil.crop.heat_units.HeatUnits.assign_new_heat_units")
     patch_b = mocker.patch("SC_redesign.Crop_and_Soil.crop.heat_units.HeatUnits.add_heat_units")
     heat = init_heat()
+    expect = init_heat()
     heat.accumulate_heat_units(temp)
+    expect.assign_new_heat_units(temp)
+    expect.add_heat_units()
+    assert heat.accumulated_heat_units == expect.accumulated_heat_units
     assert patch_a.assert_called_once
     assert patch_b.assert_called_once
```
morrowcj, line 90: Good thinking for how to solve this test problem. You don't need to change anything, but for the future, I think it would've been sufficient to do:

```suggestion
def test_accumulate_heat_units(temp, mocker: MockerFixture):
    """check that accumulate_heat_units() calls the right functions"""
    patch_a = mocker.patch("SC_redesign.Crop_and_Soil.crop.heat_units.HeatUnits.assign_new_heat_units")
    patch_b = mocker.patch("SC_redesign.Crop_and_Soil.crop.heat_units.HeatUnits.add_heat_units")
    heat = init_heat()
    heat.accumulate_heat_units(temp)
    assert patch_a.assert_called_once
    assert patch_b.assert_called_once
```

since `assign_new_heat_units()` and `add_heat_units()` are tested elsewhere. Over-testing is always better than under-testing in my view, though. So, good job!

morrowcj (approved):
Looks good to me.

morrowcj:
I pushed a `#TODO` to remind our future selves to look here for output manager logging.

SC_redesign/Crop_and_Soil/tests_SC/test_heat_units.py
Isgotkowitz: Thanks Clay, I will keep that in mind when looking at future tests

ltp23:
Overall, everything looks great. 

Some minor comments:
- I really like how you document the `accumulate_heat_units()`.
- Some of the one-liner functions can just be used directly, but I understand you may want to provide more context to the code.

SC_redesign/Crop_and_Soil/crop/heat_units.py
```diff
@@ -31,23 +31,24 @@ def absorb_heat_units(self, mean_air_temperature: float = None,
...
-            use_temp = air_temperature
+            use_temp = mean_air_temperature
         self.is_growing = self.minimum_temperature <= use_temp <= self.maximum_temperature
+        self.accumulate_heat_units(mean_air_temperature)
+        self.previous_heat_fraction = self.heat_fraction
+        self.heat_fraction = self.accumulated_heat_units / self.potential_heat_units
+        self.is_mature = self.heat_fraction >= 1.0
 
-    def decide_to_use_heat_unit_temperature(self, use_heat_unit_temperature: bool):
+    def decide_to_use_heat_unit_temperature(self, use_heat_unit_temperature: bool) -> None:
```
ltp23: This method is not really necessary. In Python, a normal instance variable can be accessed and modified directly (i.e., no need for getters and setters).

ltp23 (approved):

Issue #240, opened 2022-12-14
Double check cow and heiferIII calc_all_p_comp assignments


ew3361zh:
Wanted to confirm a couple lines of code from a closed PR [Animal Management Refactor #1](https://github.com/RuminantFarmSystems/MASM/pull/138/files).

## Expected Behavior
in RUFAS/routines/animal/animal_management.py, the function "calc_all_p_comp" calculates each animal class's P concentration. There is a 1 to 1 self.p_comp['animal_class'] = self._calc_p_comp(self.animal_class) relationship for Calf, HeifierI, and HeiferII.

## Current Behavior
When the method reaches the HeiferIII class, it skips HeiferIII and sets the self.p_comp for the Cow class and sets that by using the self._cal_p_comp method on the HeiferIII class. This doesn't follow the pattern up to this point so I wanted to make sure it is correct before entering these calculations into the OutputManager.
self.p_comp['calf'] = self._calc_p_comp(self.calves)
self.p_comp['heiferI'] = self._calc_p_comp(self.heiferIs)
self.p_comp['heiferII'] = self._calc_p_comp(self.heiferIIs)
**self.p_comp['cow'] = self._calc_p_comp(self.heiferIIIs)**

## Possible Solution
Prior to the refactor, this pattern was followed:
self.calf_p_comp = self.p_comp(self.calves)
self.heiferI_p_comp = self.p_comp(self.heiferIs)
self.heiferII_p_comp = self.p_comp(self.heiferIIs)
self.heiferIII_p_comp = self.p_comp(self.heiferIIIs)
self.cow_p_comp = self.p_comp(self.cows)

If this is correct, it is an easy fix to return to this.

## Steps to Reproduce
image of change from Animal Management Refactor #1 below:
![Screenshot 2022-12-13 at 2 23 27 PM](https://user-images.githubusercontent.com/70217952/207711970-2514400d-ab13-4903-9ef3-f2f83431e2de.png)

JoeWaddell:
Fixed in  Phosphorus Concentration Naming Convention rework #251 , see comment: https://github.com/RuminantFarmSystems/MASM/pull/251#pullrequestreview-1229130061

@JoeWaddell also reviewed the changes. Only remaining instances of p_comp in the code is a TODO in animal_management.py (Which is undergoing refactor)

PR #242, opened 2022-12-15
Output Manager Animal Module body_weight

### LINKED ISSUE #244: OutputManager Unexepected Key Collision Error on Windows OS
Several PC Windows users are reporting duplicate keys when trying to add data to the OutputManager. There are specific safeguards in place to specifically prevent this from happening so it is unexpected and causes the simulation to crash.

## Expected Behavior
When running the simulation and entering a piece of data into one of the OutputManager (OM) pools, the <_get_time_based_suffix()> method in that class is supposed to get a current time stamp down to the nano-second and append it as a suffix to the key for the value of each piece of data entered. This should allow the OM to generate unique keys and not run into key collision errors where the same key is entered for a different value thereby overwriting the previous data in the pool.

## Current Behavior
When adding variables to the OutputManager, user running the simulation on a Windows OS encountered a Key Collision Error where the OutputManager <_get_time_based_suffix()> method is generating the same time suffix for multiple calls to that same method at sequential moments in the simulation. This has not been able to be replicated on a Mac by 2 separate Mac users.

## Possible Solution
One potential solution is eliminating the unique time stamp suffix and adding values from the same caller class and caller function to a list.

## Steps to Reproduce
Adding print(om._get_time_based_suffix()) above the om.add_variable() call on line 468 of life_cycle.py (in the daily_update() method adding the variable "daily_milk_production") results in two equivalent time stamps printing right before the crash.

## Context (Environment)
This unexpected key collision is crashing the program for Windows users and doesn't allow the user to run a full simulation. The variables added are important to the output needed by the SMEs for reporting so figuring out a way to add these data is important.

ew3361zh:
What:
This PR is focused on OutputManager additions in the Animal Module specifically around animal body_weight.

Why:
This will be the first of 4 main target areas to collect data across the animal module for Output Manager.

How:
Primarily focused on searches for the word "weight" across the animal module. I've run the simulation several times with these changes implented to look at the OutputManager output json files and additionally this has passed all current tests.

A couple notes to keep in mind:
1. It was challenging to try not to get entangled with existing pathways for extracting information via the output_handler. If you notice something entangled with output_handler or the existing reports, please note it.
2. I'm interested to hear from SMEs on whether they feel this captures the data around body_weight they're looking for. Unfortunately right now after running the simulation the json file generated for variables is too large for me to open in vs_code and you may encounter some issues loading it in your IDE or other application as well. Hopefully that is not too big an issue in reviewing the PR.

**Preferred PR closing time: Friday, Dec 16**

### DIFF
--- a/RUFAS/routines/animal/animal_management.py
+++ b/RUFAS/routines/animal/animal_management.py
@@ -27,6 +27,7 @@ from statistics import mean
 
 om = OutputManager()
 
+
 def daily_animal_routine(animal_management, feed, weather, time):
     """
     Executes daily routines relating to Animals. This method is called every day
@@ -148,7 +149,8 @@ class AnimalManagement:
         AnimalBase.set_config(animal_config)
         AnimalBase.set_nutrient_list(feed.nutrient_rqmts)
         self.init_pens(data['pen_information'], data['herd_information'])
-        self.init_animals(data['herd_information'], self.all_pens, weather, time, config, feed)
+        self.init_animals(data['herd_information'],
+                          self.all_pens, weather, time, config, feed)
         self.housing = data['housing']
         self.pasture_concentrate = data['pasture_concentrate']
         self.ration_user_input = data['ration']['user_input']
@@ -196,16 +198,16 @@ class AnimalManagement:
         manure_separator = "sedimentation"
         manure_storage = "storage_pit"
         animal_combination = None
-        info_map = {"class": self.__class__.__name__, 
+        info_map = {"class": self.__class__.__name__,
                     "function": self.init_pens.__name__,
                     "all_pens_data": all_pens_data,
                     "herd_data": herd_data,
-                    "all_pens": self.all_pens,}
+                    "all_pens": self.all_pens, }
         if (len(self.all_pens) == 0) and (herd_num > 0):
-            om.add_warning("invalid_pen_num_warning", 
-                            "Warning: herd_num > 0, but pen_num = 0."
-                            + " Initilizing 3 default pens.",
-                             info_map)
+            om.add_warning("invalid_pen_num_warning",
+                           "Warning: herd_num > 0, but pen_num = 0."
+                           + " Initilizing 3 default pens.",
+                           info_map)
             pen_1 = Pen(0, 0.1, 1.6, 100, 'open air barn', 'sand', 'freestall',
                         manure_handling, manure_separator, manure_storage,
                         animal_combination, 1.2)
@@ -219,10 +221,10 @@ class AnimalManagement:
             self.all_pens.append(pen_2)
             self.all_pens.append(pen_3)
         elif (len(self.all_pens) == 1) and (herd_num > 0):
-            om.add_warning("invalid_pen_num_warning", 
-                            "Warning: herd_num > 0, but pen_num = 1."
-                            + " Initilizing 2 default pens.",
-                             info_map)
+            om.add_warning("invalid_pen_num_warning",
+                           "Warning: herd_num > 0, but pen_num = 1."
+                           + " Initilizing 2 default pens.",
+                           info_map)
             pen_2 = Pen(1, 0.1, 1.6, 300, 'open air barn', 'sawdust', 'freestall',
                         manure_handling, manure_separator, manure_storage,
                         animal_combination, 1.2)
@@ -232,10 +234,10 @@ class AnimalManagement:
             self.all_pens.append(pen_2)
             self.all_pens.append(pen_3)
         elif (len(self.all_pens) == 2) and (herd_num > 0):
-            om.add_warning("invalid_pen_num_warning", 
-                            "Warning: herd_num > 0, but pen_num = 2." 
-                            + " Initilizing 1 default pen.",
-                             info_map)
+            om.add_warning("invalid_pen_num_warning",
+                           "Warning: herd_num > 0, but pen_num = 2."
+                           + " Initilizing 1 default pen.",
+                           info_map)
             pen_3 = Pen(2, 0.1, 1.6, 300, 'open air barn', 'straw', 'tiestall',
                         manure_handling, manure_separator, manure_storage,
                         animal_combination, 1.2)
@@ -259,14 +261,14 @@ class AnimalManagement:
             weather: instance of the Weather class defined in classes.py
             time: instance of the Time class defined in classes.py
         """
-        info_map = {"class": self.__class__.__name__, 
+        info_map = {"class": self.__class__.__name__,
                     "function": self.init_animals.__name__,
                     "herd_data": herd_data,
                     "pen_data": pen_data,
                     "weather": weather,
                     "time": time,
                     "config": config,
-                    "feed": feed,}
+                    "feed": feed, }
         calf_num = herd_data['calf_num']
         heiferI_num = herd_data['heiferI_num']
         heiferII_num = herd_data['heiferII_num']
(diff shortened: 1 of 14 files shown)


RUFAS/routines/animal/animal_management.py
```diff
@@ -838,11 +876,20 @@ def calc_all_p_comp(self):
...
+                    "function": self.calc_all_p_comp.__name__, }
         # TODO: see if there is a better way to do this using dictionary comprehension
         self.p_comp['calf'] = self._calc_p_comp(self.calves)
+        om.add_variable("calf_p_comp", self.p_comp["calf"], info_map)
         self.p_comp['heiferI'] = self._calc_p_comp(self.heiferIs)
+        om.add_variable("heiferI_p_comp", self.p_comp["heiferI"], info_map)
         self.p_comp['heiferII'] = self._calc_p_comp(self.heiferIIs)
+        om.add_variable("heiferII_p_comp", self.p_comp["heiferII"], info_map)
         self.p_comp['cow'] = self._calc_p_comp(self.heiferIIIs)
+        # TODO check if this is set up correctly. Currently p_comp for the cow class is
```
JoeWaddell, line 876: This seems incorrect, since it's not actually setting/storing p_comp for heiferIIIs. As you noted, p_comp for cows is being calculated using HeiferIII animals, which doesn't seem correct either, unless there's something I'm missing about the p_comp calcs.

RUFAS/routines/animal/life_cycle/calf.py
```diff
@@ -196,4 +215,11 @@ def update(self, sim_day):
 
         self.daily_growth = self.body_weight - prev_weight
 
+        info_map = {"class": self.__class__.__name__,
+                    "function": self.update.__name__,
+                    "sim_day": sim_day, }
+        om.add_variable("calf_body_weight_after_update",
+                        self.body_weight, info_map)
+        om.add_variable("calf_daily_growth", self.daily_growth, info_map)
```
JoeWaddell: This is great: I waffled on whether we need the `daily_growth` value, since it's trivial to calculate from the `body_weight` variable, but if it's only adding literal nanoseconds to store, it will certainly make reports easier to generate.

RUFAS/routines/animal/life_cycle/calf.py
```diff
@@ -196,4 +215,11 @@ def update(self, sim_day):
 
         self.daily_growth = self.body_weight - prev_weight
 
+        info_map = {"class": self.__class__.__name__,
+                    "function": self.update.__name__,
+                    "sim_day": sim_day, }
+        om.add_variable("calf_body_weight_after_update",
```
JoeWaddell: For my own clarification on variable naming here: since info map tells us the sim_day for each body_weight value, is the idea that "after_update" in the name clarifies that it's the "end of day" weight?

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -134,6 +137,11 @@ def __init__(self, args):
             self.milking = self.days_in_milk != 0
             self.calves = args['parity']
             self.CI = args['calving_interval']
+        
+        info_map = {"class": self.__class__.__name__,
+                    "function": self.__init__.__name__,
+                    "args": args, }
+        om.add_variable("cow_body_weight_at_init", args["body_weight"], info_map)
```
JoeWaddell: This initial weight was handled slightly differently for calves: where I think it's only being stored in the info_map of the mature_body_weight_calf om variable. Should there be a calf_body_weight_at_init variable for standardization, or was your reasoning that calves are distinct enough from the other animal types that they should be handled differently?

JoeWaddell (commented):
First quick pass through to leave comments before a more thorough review tomorrow (Friday), in the interests of meeting your requested PR deadline.

RUFAS/routines/animal/animal_management.py
ew3361zh: I spoke with Pooya and he asked me to create an issue and tag him and Anchey in it so it's [here](https://github.com/RuminantFarmSystems/MASM/issues/240). I thought I had commented this out as well but I will go back and do that for the time being since I agree it seems incorrect.

RUFAS/routines/animal/life_cycle/calf.py
ew3361zh: I was thinking that as well but including it may be part of what is leading to the variables json file generated by these variables being added to be unmanageably large (1.55GB if left as is). It's not necessarily that each individual variable added is that large in and of itself but that the large volume of variables added together seems to add up quicker than I'd thought.

This is helpful feedback from your perspective though as to what is useful and would make report generation on the other side easier for you.

RUFAS/routines/animal/life_cycle/calf.py
ew3361zh: Correct. So this is the calf's weight after the update method has been executed on a particular day in the simulation. I could definitely change it to "end_of_day_calf_weight" or something like that if that's clearer. What do you think?

RUFAS/routines/animal/life_cycle/cow.py
ew3361zh: That's a great point - do you think it would be helpful to have each of the following as variables or is there a way they could be in the info map that might be helpful? I may also not be understanding each of these - is the body_weight the only one that changes and the others are guideposts? 
-calf_body_weight_at_init
-calf_birth_weight_at_init
-calf_wean_weight_at_init
-calf_mature_body_weight_at_init

RUFAS/routines/animal/life_cycle/calf.py
JoeWaddell: OK then one of those variables is a good candidate for removing from the output manager, since there is no information loss. I'd think body_weight is a better candidate for storing, but it can still be reconstructed from init_weight and daily_growth, if other folks think it's more important to directly track those.

Going back to the discussion of how for the graphics printing of individual animals is handled: it's limited to a variable number per animal type that the user can define, to avoid plotting tens of thousands of jpgs. So for the final step of saving these data to CSVs, I imagine there may need to be a similar process in place - but in any case they all still need to be added to the om

RUFAS/routines/animal/life_cycle/calf.py
JoeWaddell: "end_of_day_calf_weight" sounds perfect

RUFAS/routines/animal/life_cycle/cow.py
ew3361zh: I think I'm going to put the birth_weight, wean_weight, and mature_body_weight figures in the info_map as they appear to be fixed and the body_weight will be the variable added. Let me know what you think.

RUFAS/routines/animal/life_cycle/calf.py
```diff
@@ -87,11 +91,20 @@ def init_values(self, args):
...
+                                                AnimalBase.config['mature_body_weight_std'])
         self.wean_weight = 0
         self.p_animal = args['p_init']
 
+        info_map = {"class": self.__class__.__name__,
+                    "function": self.init_values.__name__,
+                    "args": args,
+                    "birth_weight": self.birth_weight,
+                    "body_weight": self.body_weight,
+                    "wean_weight": self.wean_weight}
```
ltp23: Hi Niko, can you tell me the expected behavior for those extra key-value pairs like `args`, `birth-weight`, etc. in the `info_map`? When and where are they being used?

ew3361zh: Hi Loi, good question. I believe the extra variables logged into info_map are contextual and would hopefully be helpful to determining the state of the model and class at the moment something was added to one of the OutputManager pools. I have not yet seen how they might be used in practice though.

ltp23 (approved):
Overall, it looks good to me. There may be a little too many existing files being changed, but the changes are minor enough, so I think it's OK. Thanks Niko!

RUFAS/routines/animal/life_cycle/calf.py
ltp23: Ok, Niko. I think we could find a way to incorporate those extra bits of information somewhere later.

morrowcj:
Something is wrong with the time based suffix in OutputManager where the same suffix is given at multiple time points (on my Windows machine, at least).

Take, for example the script 

```python
from time import sleep

om = OutputManager()

old_time = None
for i in range (20):
    sleep(5)
    new_time = om._get_time_based_suffix()
    print(new_time == old_time)
    old_time = new_time
```

which prints `True` on all iterations except the first. 

---- 

Edit: This is all to say that RuFaS will not successfully run on my computer in this branch. I get key collision errors.

JoeWaddell:
Re: @morrowcj 

Can confirm. I've tried it on my laptop and two others in the lab (all Windows machines), and we're getting those key collisions. Will keep trying to figure out why (currently using t he debug function to walk through the process).

morrowcj:
Actually, after resetting my python session and running the above script, I get `False`s. But removing the `sleep()` call results in `True`s.

PooyaHekmati:
@morrowcj and @JoeWaddell can you please re-run the model and confirm the fix?

morrowcj:
@PooyaHekmati, it is still throwing an error. 

Adding `print(om._get_time_based_suffix())` above the `om.add_variable()` call on line 468 of `life_cycle.py` results in two equivalent time stamps printing right before the crash. 

I ran `python main.py input/animal_management.json --no-graphics` to produce the error, which is:

```
ValueError: Key LifeCycleManager.daily_update.average_cow_body_weight.1671676435820309100 already exists in the variables_pool.Consider using different name, prefix/suffix;or turn the automated prefix/suffix generation on.info_map is: {'class': 'OutputManager', 'function': 'add_variable', 'date': 1, 'key': 'LifeCycleManager.daily_update.average_cow_body_weight.1671676435820309100'}

```

PooyaHekmati:
Thanks, @morrowcj . This new commit removes usage of time stamps in the key completely.

morrowcj:
No problem. I just don't understand why it was only an issue on the windows machines.

JoeWaddell (approved):
The fixes from last night are working on Windows machines in the lab as well. Looks good to me!

PR #249, opened 2022-12-22
Manure management

ltp23:
Hi reviewers,

The manure treatment submodule is somewhat more complex than the other submodules, so I think it would be more manageable to add/modify one or two files at a time, but three for this one. 

New files: 
- `manure_treatment_daily_output.py`: Each daily update should generate a separate object so they can be aggregated in different ways later on. There is an `__add__()` function that enables the use of the standard `+` operation to add the corresponding attributes of two objects with similar types. Some attributes don't make logical sense when adding together two `pen_id`s, but for the sake of simplicity, I don't make such exclusions at the moment. 
- `manure_treatment_types.py`: List of available treatment types. 
- `manure_treatment_configs.py`: mainly deals with providing default config parameters for different treatment types. 

Thank you everyone as always for your patience.

---

Preferred closing time: Friday, Dec 23


RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
```diff
@@ -0,0 +1,69 @@
...
+        pen_id: ID of the pen that this output is associated with.
+        simulation_day: Number of days into the simulation.
+        liquid_manure_total_ammoniacal_nitrogen: Total ammoniacal nitrogen, kg.
+        liquid_manure_nitrogen: Amount of nitrogen in manure, kg.
+        liquid_manure_total_solids: Total amount of solids from the manure and the bedding, kg.
+        liquid_manure_total_volatile_solids: Total amount of volatile solids, kg.
+        liquid_manure_phosphorus: Amount of phosphorus excreted in manure, kg.
+        liquid_manure_potassium: Amount of potassium in manure, kg.
+        final_manure_volume: Final manure volume after treatment, m^3.
+
```
morrowcj, line 23: I think you're missing a few attribute docstrings. is there a reason you excluded them? Perhaps they're documented in the parent class (it doesn't look like it, though).

```suggestion
        liquid_manure_potassium: Amount of potassium in manure, kg.
        final_manure_volume: Final manure volume after treatment, m^3.
        liquid_manure_daily_volume: [description of variable]
        storage_methane: [description]
        ...
        accumulated_final_manure_volume: [description]
```

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
```diff
@@ -0,0 +1,69 @@
...
+        Returns:
+            ManureTreatmentDailyOutput with corresponding attributes summed.
+
+        """
+        if not isinstance(other, ManureTreatmentDailyOutput):
+            raise TypeError('Other must be of type ManureTreatmentDailyOutput.')
+
+        return ManureTreatmentDailyOutput(*[
+            attr1 + attr2 for attr1, attr2 in zip(astuple(self), astuple(other))
+        ])
```
morrowcj, line 69: Does this syntax (`*[]`) simply indicate that the list can be of any length?

RUFAS/routines/manure/manure_treatments/manure_treatment_configs.py
```diff
@@ -0,0 +1,162 @@
...
+from RUFAS.routines.manure.manure_treatments.manure_treatment_types import ManureTreatmentType
+
+
+@dataclass
+class ManureTreatmentConfig:
+    """Class for storing manure treatment configuration data.
+
+    Attributes:
+        total_solids_removal_efficiency_for_treatment: Percent of total solids removed from manure during treatment.
+        volatile_solids_removal_efficiency_for_treatment: Percent of volatile solids removed from manure during
```
morrowcj, line 14: I really appreciate the attention to variable naming that you're giving. Names are becoming much more descriptive and useful now. In my opinion, these ones are right on the edge of "too long" for attribute names.

I don't think you need to change anything about this PR, but in the future, I think something like `total_solids_removal_efficiency`, `volatile_solids_removal_efficiency`, etc. would have been sufficiently descriptive. It is a tough balance, but I love the progress we're making!

morrowcj (approved):
## Main comments
This Looks great to me. Very intuitive and clean. I left a few minor comments about documentation.

## Additional comments: 
* Is there a github issue associated with the TODO on line 122 of `manure_treatment_configs.py`? If not, should we create one?
* I'm thinking that the pattern you use here will be really useful for instanciating a `Crop` class.
Would you be willing to consult on that, when we get to it?

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
```diff
@@ -0,0 +1,69 @@
...
+    sludge_manure_potassium: float = 0.0
+    sludge_manure_daily_volume: float = 0.0
+
+    # Temporary variables used for making plots
+    accumulated_sludge_volume: float = 0.0
+    accumulated_final_manure_volume: float = 0.0
+
+    def __post_init__(self):
+        """Ensures that the daily volume is set to the final manure volume."""
+        self.liquid_manure_daily_volume = self.final_manure_volume
```
ew3361zh, line 52: Can you explain the rationale behind doing this in the __post_init__ vs at the start of the dataclass when initializing each of the variables? I have mostly seen __post_init__ used for when one variable value depends on the value of another variable at initialization to determine its own value but here it seems like final_manure_volume (and therefore the liquid_manure_daily_vaolume) will always be set to 0.0?

ew3361zh (approved):
I appreciate the consistency in form and organization across these manure module PRs. My only concern here is that there are a LOT of scientific equations and constants that I'm not a good resource to confirm so I think it would be very helpful for a SME to review all those aspects across these manure module PRs to ensure they are accurate. Otherwise, this is really well done, Loi!

varrma (approved):

PR #258, opened 2023-01-05
Example fix

morrowcj:
added doscstrings for optimal harvest index.


morrowcj:
closing PR, since this was just an example for fixathon 2023

PR #260, opened 2023-01-06
water dynamics composition (SC)

morrowcj:


SC_redesign/Crop_and_Soil/crop/water_dynamics.py
```diff
@@ -6,20 +6,16 @@
 class WaterDynamics:
     def __init__(self, crop_data: Optional[CropData] = None):
         data = crop_data or CropData()  # initialize with defaults, if not given
```
ltp23, line 8: This may be redundant, right?

ltp23 (approved):
Everything looks good to me!

PR #266, opened 2023-01-07
crop yields wiki

morrowcj:
markdown documentation of the `Yields` class. This will eventually be added to the wiki page for yields.

### DIFF
--- a/SC_redesign/Crop_and_Soil/docs/crop_yields.md
+++ b/SC_redesign/Crop_and_Soil/docs/crop_yields.md
@@ -1,6 +1,7 @@
-## Summary
+## Crop Yields
 
-The `Yields` class contains the methods that determine the amount and composition of harvestable crop production. 
+Crop yields in `RuFaS` are calculated by the `Yields` class.  `Yields` contains the methods that 
+determine the amount and composition of harvestable crop production. 
 Its main method function `obtain_yields()` executes all routines to determine production at the time of harvest. As 
 such, this function is only called once throughout a crop's growing season. This is typically done at cut/harvest time.
 
--- a/SC_redesign/Crop_and_Soil/docs/crop_yields.md
+++ b/SC_redesign/Crop_and_Soil/docs/crop_yields.md
@@ -1,0 +1,95 @@
+## Summary
+
+The `Yields` class contains the methods that determine the amount and composition of harvestable crop production. 
+Its main method function `obtain_yields()` executes all routines to determine production at the time of harvest. As 
+such, this function is only called once throughout a crop's growing season. This is typically done at cut/harvest time.
+
+## Reference
+
+The `Yields` class is primarily an implementation of the methods described in 
+[SWAT](https://swat.tamu.edu/media/99192/swat2009-theory.pdf) section 5:2.4, with some minor modifications.
+
+## Details
+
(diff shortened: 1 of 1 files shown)


SC_redesign/Crop_and_Soil/docs/crop_yields.md
```diff
@@ -0,0 +1,95 @@
...
+The `Yields` class calculates the following attributes with its methods. Unless otherwise specified, these attributes
+are updated by `obtain_yields()`.
+
+* `potential_harvest_index`: potential harvest index for a given day. 
+* `harvest_index`: harvest index for a given day; fraction of above-ground plant biomass that is harvestable 
+economic yield (unitless)
+* `crop_yield`: total amount (kg/ha) of the desired crop product
+* `yield_collected`: amount (kg/ha) of the desired crop product to be removed from the field
+* `yield_residue`: amount of residue created; unharvested yield (kg/ha)
+* `collected_nitrogen`: nitrogen contained in the harvested yield (kg/ha)
```
LaplacetteCM, line 53: Just a question regarding units for N and P: kg/ha or kg N/ha and kg P/ha as used in SWAT?

LaplacetteCM (approved):
Very clear Clay, easy to understand and follow with that structure.

SC_redesign/Crop_and_Soil/docs/crop_yields.md
morrowcj: I don't agree conceptually with "kg X" as a unit, so I've written it as just "kg" here and elsewhere. 

I could be wrong though. If there's a good reason to do the former, I'm open to learning.

LaplacetteCM: I don't see a good reason to change it: the description gives that information. I left the comment because I didn't know the unit criteria you are using and realized the SWAT theory was using the other way. Just in case you were missing it.

PooyaHekmati:
The markdown file compiles and looks nice. However, the content looks like a compilation of docstrings. We do not want a wiki page for such information. Wiki is a place for high-level information. Details like this PR have to go to docstrings so we can compile html, pdf, etc.. from them. Also, having detailed documentation in a file other than the code itself leads to doc and code drifting and getting out of sync, causing confusion.

morrowcj:
@PooyaHekmati can you elaborate and give an example? This is what I recall our discussions of model documentation encompassing.

PooyaHekmati:
> @PooyaHekmati can you elaborate and give an example? This is what I recall our discussions of model documentation encompassing.

The answer is within your sentence: "model documentation"; this PR is documenting the code not the model. And as I said in the last comment, code documentation should reside with the code itself.

A good rule of thumb, which also exists in our onboarding document, is to answer the what-why-how questions. Generally speaking, the answers to what question are code documentation, and the answers to why and how questions are high-level/model documentation (which goes to wikis)
-What are the inputs of this code? -- docstring
-What are the outputs of this code? -- docstring
-What are the units? --docstring
-What are the methods, attributes, exceptions, etc...? --docstring
-Why is this code designed this way? --wiki
-How the output of this code is validated? --wiki
-Why is this code the way it is? --wiki
-How should this code run/called? --wiki

PR #270, opened 2023-01-09
growth constraints composition restructure

morrowcj:
## Summary
This PR addresses a few of the fixation 2023 tasks, since they were easier to do together. I've tried making it so that individual commits address individual tasks as best as possible. The main things this PR does:

* [x] Restructures `yields.py` to use composition design pattern.
* [x] adds docstrings to the variables used by `yields.py` (in the `crop_data.py` file)
* [x] Restructures `growth_constraints.py` to use composition design pattern.
* [x] adds docstrings to the variables used by `growth_constraints.py` (in the `crop_data.py` file)
* [x] **Refactors** `growth_constraints.py` to match other Crop Module files
* [x] Restructures `water_dynamics.py` to use composition design pattern.
* [x] adds docstrings to the variables used by `water_dynamics.py` (in the `crop_data.py` file)
* [x] adds docstrings for root development attributes (missed previously)

## Other Changes
changed static functions for all crop classes to private by prefixing with `_`.


SC_redesign/Crop_and_Soil/crop/crop.py
```diff
@@ -77,7 +77,7 @@ def grow_crop(self, layer_nitrates: List[float], layer_depths: List[float],
         """
         self.absorb_heat_units(mean_air_temperature, min_air_temperature,
                                max_air_temperature)
-        self.root_development.develop_roots()
+        self.root_development._develop_roots()
```
ltp23: We probably want to use a public method here since we are outside of the RootDevelopment class.

ltp23 (approved):
I'm wondering if later on, you want to decompose the CropData class into smaller classes as well, just so that it becomes more manageable as you add on more attributes. After reviewing your code, I also realize I can do that for the manure module, but for a different reason, to minimize similar-looking code. 

Overall, everything looks really great! Great work!

SC_redesign/Crop_and_Soil/crop/crop.py
morrowcj: Oh, good point. This was a mistake. I'll fix it.

morrowcj:
> I'm wondering if later on, you want to decompose the CropData class into smaller classes as well, just so that it becomes more manageable as you add on more attributes. 

I see. That might be a good idea in the future. I don't foresee adding too many more variables. Do you think the class is too big currently?

One issue with doing so is that many of the `CropData` attributes are updated by multiple process classes. It is nice to to have a single class where all the variables are stored so that we can be sure that we're always using the most up-to-date version of a variable.

PR #271, opened 2023-01-09
Manure management

ltp23:
The main focus of this PR is to refactor the anaerobic lagoon class (`anaerobic_lagoon.py`) and update related unit tests. 

Some changes in other files had to do variable name changes. 

Please provide the amount of review time it took as well.

### DIFF
--- a/RUFAS/routines/manure/manure_treatments/anaerobic_lagoon.py
+++ b/RUFAS/routines/manure/manure_treatments/anaerobic_lagoon.py
@@ -125,7 +125,7 @@ class AnaerobicLagoon(BaseManureTreatment):
 
         return daily_output
 
-    def _adjust_final_manure_volume(self, current_day_final_manure_volume: float):
+    def _adjust_final_manure_volume(self, current_day_final_manure_volume: float) -> float:
         """Adjusts the final manure volume to account for the precipitation and the storage time period.
 
         Args:
--- a/RUFAS/routines/manure/manure_treatments/anaerobic_lagoon.py
+++ b/RUFAS/routines/manure/manure_treatments/anaerobic_lagoon.py
@@ -7,7 +7,6 @@ from RUFAS.routines.manure.gas_emissions.gas_emissions import GasEmissions
 from RUFAS.routines.manure.manure_treatments.base_manure_treatment import BaseManureTreatment
 from RUFAS.routines.manure.manure_treatments.manure_treatment_configs import ManureTreatmentConfig
 from RUFAS.routines.manure.manure_treatments.manure_treatment_daily_output import ManureTreatmentDailyOutput
-from RUFAS.routines.manure.manure_treatments.manure_treatment_daily_output import SludgeOutput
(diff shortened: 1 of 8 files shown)


RUFAS/routines/manure/manure_treatments/anaerobic_lagoon.py
```diff
@@ -16,79 +15,30 @@ def __init__(self, weather, time, manure_treatment_config: ManureTreatmentConfig
         super().__init__(weather, time, manure_treatment_config)
         self.storage_time_period = self.config.storage_time_period
         self.freeboard_input = self.config.freeboard_input
-        self._accumulated_sludge_output = SludgeOutput()
         self._accumulated_precipitation_volume = 0.0
 
     def _calc_daily_sludge_output(self, daily_output: ManureTreatmentDailyOutput,
-                                  manure_treatment_daily_input: LiquidManurePortionProtocol) -> None:
+                                  manure_treatment_daily_input: LiquidManurePortionProtocol) \
+            -> ManureTreatmentDailyOutput:
```
ew3361zh, line 22: What issue was happening that led to needing these refactors (i.e. changing this method to have a return vs altering the object attribute)? Just curious to see the issue and how it relates to the design pattern you chose to address it with.

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
```diff
@@ -69,6 +74,15 @@ def __add__(self, other: ManureTreatmentDailyOutput) -> ManureTreatmentDailyOutp
             attr1 + attr2 for attr1, attr2 in zip(astuple(self), astuple(other))
         ])
 
+    def clone(self) -> ManureTreatmentDailyOutput:
+        """Returns a clone of this object.
```
ew3361zh, line 78: Can you explain the purpose behind this .clone() method? Is it to leave a version of daily_output that is not altered by the _calculated_daily_sludge_output method? And is there a reason to create this method vs using deepcopy() vs. "="?

ew3361zh (approved):
Very nice work. I have a couple comments for clarity on the design pattern approach but otherwise this looks great!

review took ~30 minutes

RUFAS/routines/manure/manure_treatments/anaerobic_lagoon.py
ltp23: I don't think there is really a strict rule to follow here. Whenever I can, I would prefer returning something to making changes silently to the arguments being passed in. The good thing is the argument objects remain intact but this way may not always be efficient since I'm making extra copies.

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
ltp23: Using shallow copying or deep copying would work too because these data objects only contain primitive values like floats or ints or strings. I did think about that, but I decided to write a custom clone() method just so that I have control over the copying process.

morrowcj (approved):
Looks good to me. My only comments were in line with @ew3361zh's and your clarification was helpful.

PR #272, opened 2023-01-10
temporary fix of heifer repro protocol not assigned correctly

YijingGong:


input/animal/animal_management_animal.json
```diff
@@ -42,8 +42,8 @@
 				},
 				"TAI_related":{
 					"heifer_TAI_protocol": "d5CG2P",
-					"m5dCG2P_conception_rate": 0.6,
-					"m5dCGP_conception_rate": 0.48,
+					"md5CG2P_conception_rate": 0.6,
+					"md5CGP_conception_rate": 0.48,
```
JoeWaddell, line 46: So in master on line 455 in [heiferII.py ](https://github.com/RuminantFarmSystems/MASM/blob/da1ae3cf7f1d753018b0239b0a84db6af9b2b813/RUFAS/routines/animal/life_cycle/heiferII.py#L455) it's looking for `m5dCGP_conception_rate`, which you're renaming (correctly) here. 

I think that if you change that line in heiferII.py, it should resolve the issue for now, until the larger refactor is in place.

JoeWaddell (changes requested):

JoeWaddell (approved):
Great! Seems like a good temporary fix.

PR #281, opened 2023-01-11
Manure management

ltp23:
The main focus of this PR is to refactor and update unit tests for the anaerobic digestion class (`anaerobic_digestion.py`).


tests/test_manure_module/test_manure_treatments.py
```diff
@@ -2032,3 +2083,123 @@ def test_bound_sludge_accumulation_volume(mocker: MockFixture) -> None:
...
     assert patch_for_sludge_accumulation_volume_property.call_count == 2
     assert actual_bound_sludge_accumulation_volume == approx(expected_bound_sludge_accumulation_volume)
+
+
+# Test AnaerobicDigestion-specific methods
+# ========================================
+
+def test_calc_anaerobic_digestion_daily_output(mocker: MockFixture) -> None:
+    # Arrange
+    pass
```
morrowcj: Did you mean to leave this empty? If so, maybe add a `TODO`?

tests/test_manure_module/test_manure_treatments.py
```diff
@@ -2032,3 +2083,123 @@ def test_bound_sludge_accumulation_volume(mocker: MockFixture) -> None:
...
+
+
+@pytest.mark.parametrize(
+        'total_daily_mass',
+        [
+            0.0,
+            100.0,
```
morrowcj: What happens if a negative value is given (incorrectly), or a small number [0, 1]?

My advice for tests like this would be to throw a bunch of cases, including edge cases, at your function to see how/if it handles them. I think it makes the tests more valuable to know that they are robust. I'd err on the side of too many cases rather than too few.

tests/test_manure_module/test_manure_treatments.py
```diff
@@ -2032,3 +2083,123 @@ def test_bound_sludge_accumulation_volume(mocker: MockFixture) -> None:
...
+        ])
+def test_calc_moisture_content(total_daily_mass: float) -> None:
+    """Unit test for _calc_moisture_content() in anaerobic_digestion.py."""
```
morrowcj, line 2239: I'd also suggest making these test docstrings a bit more descriptive. Rather than saying is a unit test, say what it is meant to check. For example:
```suggestion
    """check that moisture content is properly calculated by  _calc_moisture_content()."""
```
It's still 1 line but it tells the reader what you intended the test to do.

morrowcj (approved):
I think this looks good. I did have a few small suggestions about the tests, but I'm approving.

tests/test_manure_module/test_manure_treatments.py
ltp23: Updated now.

tests/test_manure_module/test_manure_treatments.py
ltp23: I agree with you on this. The test values I use for unit tests tend to follow somewhat too closely to the implementation. I would say overall it is not very rigorous but arguably sufficient in catching the main failed cases.

tests/test_manure_module/test_manure_treatments.py
ltp23: I think it's not that important to have a descriptive docstring under the name of each unit test. Though, the two good places for that are when you have multiple edge cases inside the unit test and when you want to annotate each case in the parameterized list above the unit test. But I do get your point of the more descriptive the better as opposed to something generic.

RUFAS/routines/manure/manure_treatments/anaerobic_digestion.py
```diff
@@ -20,183 +16,140 @@ class AnaerobicDigestion(BaseManureTreatment):
...
-
-    def _accumulate_daily_sludge_output(self, daily_sludge_output: SludgeOutput) -> None:
-        """Accumulates the daily sludge output from anaerobic digestion.
-
-        Args:
-            daily_sludge_output: The daily sludge output from anaerobic digestion.
+    def _calc_daily_sludge_output(self, daily_output: ManureTreatmentDailyOutput,
+                                  manure_treatment_daily_input: LiquidManurePortionProtocol) \
+            -> ManureTreatmentDailyOutput:
+        """Calculates the daily sludge output for the current day.
```
ew3361zh, line 22: Why did you drop the args/returns docstrings here?

RUFAS/routines/manure/manure_treatments/anaerobic_digestion.py
```diff
@@ -20,183 +16,140 @@ class AnaerobicDigestion(BaseManureTreatment):
...
-        self._assign_extra_output_variables(daily_output, daily_sludge_output)
         return daily_output
 
-    def _create_anaerobic_digestion_daily_output(self, final_manure_volume: float) -> AnaerobicDigestionOutput:
+    def _calc_anaerobic_digestion_daily_output(self, manure_treatment_daily_output: ManureTreatmentDailyOutput) \
+            -> ManureTreatmentDailyOutput:
         """Returns the daily output from anaerobic digestion.
 
         Args:
-            final_manure_volume: The final manure volume of manure treatment daily output.
```
ew3361zh, line 60: are you planning to replace one arg description with the new arg description?

tests/test_manure_module/test_manure_treatments.py
ew3361zh: Second Clay's feedback on the wider array of test case data

ew3361zh (approved):
Looks good! A couple of comments/questions I had for you but nothing to necessarily prevent merging

PR #282, opened 2023-01-12
Restructure LAI and HeatUnits

morrowcj:
## Summary
This PR will restructure, and document the `LeafAreaIndex` and `HeatUnits` class. It addresses the following tasks:
* [x] switch `LeafAreaIndex` to composition design
* [x] update `LeafAreaIndex` docstrings
* [x] switch `HeatUnits` to composition design
* [x] update `HeatUnits` docstrings

## Notes:
This PR also contains changes that are in #274, so `biomass_allocation.py` and `test_biomass_allocation.py` can be ignored.

### DIFF
--- a/SC_redesign/Crop_and_Soil/tests_SC/test_heat_units.py
+++ b/SC_redesign/Crop_and_Soil/tests_SC/test_heat_units.py
@@ -156,23 +156,23 @@ def test_absorb_heat_units(mean, mint, maxt, use_heat_unit_temp):
         expect_heat_unit_temp = (expect_min_heat_unit_temp / 2) + (expect_max_heat_unit_temp / 2)
         assert expect_heat_unit_temp == data.heat_unit_temperature
 
-    # if use_heat_unit_temp or mean is None:
-    #     use_temp = expect_heat_unit_temp
-    # else:
-    #     use_temp = mean
-    #
-    # if 20 <= use_temp <= 38:
-    #     expect_is_growing = True
-    # else:
-    #     expect_is_growing = False
-    # assert expect_is_growing == data.is_growing
-    #
-    # if use_heat_unit_temp or (mean is None):
-    #     expect_new_heat_units = HeatUnits._determine_new_heat_units(expect_heat_unit_temp, 20)
-    # else:
-    #     expect_new_heat_units = HeatUnits._determine_new_heat_units(mean, 20)
-    # assert data.use_heat_unit_temperature == use_heat_unit_temp
-    # assert expect_new_heat_units == data.accumulated_heat_units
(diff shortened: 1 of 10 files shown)


SC_redesign/Crop_and_Soil/tests_SC/test_heat_units.py
```diff
@@ -138,36 +141,38 @@ def test_add_heat_units(start, new):
...
-        expect_new_heat_units = HeatUnits._determine_new_heat_units(mean, 20)
-    assert heat.use_heat_unit_temperature == use_heat_unit_temp
-    assert expect_new_heat_units == heat.accumulated_heat_units
-    assert heat.previous_heat_fraction is None
-    expect_heat_fraction = expect_new_heat_units / 800
-    assert expect_heat_fraction == heat.heat_fraction
+        assert expect_heat_unit_temp == data.heat_unit_temperature
+
+    # if use_heat_unit_temp or mean is None:
+    #     use_temp = expect_heat_unit_temp
```
ew3361zh: Is this section to delete? Or what's the plan with it?

ew3361zh (approved):
I like the redesign. Does it seem better organized and make more sense to you as someone that's been working in this module for a while? I don't know a lot about the composite design pattern so can't offer detailed feedback on your implementation of it here but perhaps Loi can as he suggested it in the first place.

Docstrings for the CropData class are very helpful and descriptive while still being succinct. Tests were all passing when I ran the code locally. Nice work. This looks like it took a lot of detail and extended attention span to follow through!

SC_redesign/Crop_and_Soil/tests_SC/test_heat_units.py
morrowcj: Oh! No, this was me testing a bug and then me forgetting to  uncomment when I was finished. I'll do so now and make sure changes are still working.

morrowcj:
> I like the redesign. Does it seem better organized and make more sense to you as someone that's been working in this module for a while? I don't know a lot about the composite design pattern so can't offer detailed feedback on your implementation of it here but perhaps Loi can as he suggested it in the first place.

Yes, It does seem much better now. I wasn't hugely familiar with the pattern until recently either. I've been teaching myself Unity (and `C#`) in my spare time lately and the composite pattern is used heavily there. I think it suits this problem perfectly, since a `Crop` **has** all of these difference processes. 

> Docstrings for the CropData class are very helpful and descriptive while still being succinct. Tests were all passing when I ran the code locally. Nice work. This looks like it took a lot of detail and extended attention span to follow through!

Thanks! It has definitely been a long and iterative process, but I think it is settling into something good.

Issue #189, opened 2022-08-25
Animals in Pen only updated with ration formulation interval


KFosterReed:
## Expected Behavior

animals should change pens daily if their biological state changes (i.e. their animal class changes or they change from not milking to milking). 
## Current Behavior

Currently, animals are only added or removed from pens on a daily basis if they are purchased or born using the daily_update_id_pen() function in animal_management.py. 

Animals that change classes or start or stop milking are only moved to a new pen when pen_allocation() is called at the end of a ration formulation interval (also in animal_management.py)

## Possible Solution

both daily_update_id_pen and pen_allocation probably need to be refactored/started from scratch but other than that I don't have any good suggestions. 

## Steps to Reproduce

1. Run any simulation with animals and set the ration formulation interval to something above 1 (a larger number will make the issue more apparent)
2.
3.
4.

## Context (Environment)

Tracking the animals in the pen is an essential part of the animal module. Not having the correct number of animals in the pen messes up all other pen related estimates like this enteric methane estimate I was looking at: 
![image](https://user-images.githubusercontent.com/41019131/186549178-f0dafefc-ee24-41fa-a20a-8d690c2bc582.png)

## Detailed Description

You can see this behavior in the following plot of the different pens. Pens 3 & 4 are lactating cow pens. Pen 0 is the calf pen that is only having animals added to it through birth on a daily basis but they are only 'graduating' to Pen1, the growing pen, once every 30 day ration interval. Similarly, Pen 2 is the Close-up pen which has newly purchased animals added to the herd on a daily basis but only updates to have the animals leave that pen and move to the lactating cow pens on the 30 day ration interval. 
![image](https://user-images.githubusercontent.com/41019131/186547478-0832c308-e796-4e74-b32d-ad96f56d158c.png)
The challenge with adding an animal to a pen on a day when it is not a ration formulation interval, is that we will need to update the animal attributes related to the ration so it has the same ration information as all the other animals in that pen. 

## Possible Implementation

I think we need to track which animals are changing classes and which animals are moving from milking to non milking and back again and include those animals in the daily_pen_id_update. I also don't se a reason why we need both the pen_allocation and the daily_pen_id_update. They do very similar things.

JoeWaddell:
Working on a review of PR #322 by @ltp23 , and there is significant progress regarding this bug.

[...]

Issue #74, opened 2020-07-15
Pen change Frequency


KFosterReed:
Cows change pens very frequently. Is this what we want? What is causing them to change so frequently?

JoeWaddell:
Related to ongoing changes in animal_management.py by @Fjordrunner and @ancheypeng 
This issue and related issues #96, #188, and #189 will likely be resolved with forthcoming refactor.

See branches: https://github.com/RuminantFarmSystems/MASM/tree/animal_management_comments 
and associated
https://github.com/RuminantFarmSystems/MASM/tree/animal_management_refactor_fa22

Functions associated with this issue include: daily_update_id_pen, and also allocate_all_pens inside function daily_updates (which happens at the end of the ration interval).

PR #283, opened 2023-01-12
Animal management comments

Fjordrunner:
Final commit for daily_update_id_pen() refactor. This pull request includes the addition of three new functions, remove_animal_from_herd(), track_former_pen_population(), and calculate_pen_rations(). Additionally, daily_update_id_pen() was renamed to daily_update_id_map() to better describe its purpose. This rework also fixed the docstrings across animal_management.py and life_cycle.py to reflect that we are working with Animal objects and not directly with their IDs, which is another new change introduced in this pull request. Unit tests for the 4 functions mentioned above will be introduced in the next pull request so as to not make this one too large.


RUFAS/routines/animal/animal_management.py
```diff
@@ -1022,4 +1021,4 @@ def get_initialize_db_summary(self):
         Returns: a dictionary which is the summary of the animal initialization
         database
         """
-        return self.life_cycle_manager.initialize_db_summary
+        return self.life_cycle_manager.initialize_db_summary
```
JoeWaddell: autoformatter wants a final blank line here

JoeWaddell (commented):
This is looking good. The use of animal objects instead of ids is great! Streamlines things nicely.

I can't find any errors, but I'll have to take another round through the codebase...seems like the variable/attribute names were all updated correctly. Runs on my end just fine, though I do notice that some cows are still moving between pens every few days: likely related to other parts of the code since it could be what's referenced in git issue #74

RUFAS/routines/animal/animal_management.py
```diff
@@ -331,17 +330,68 @@ def calc_nutrient_rqmts(self, feed, temp):
...
-                self.id_pen[animal.id] = pen.id
+                if animal.id == 378168:
+                    print("fully update hit")
+                self.animal_to_pen_id_map[animal.id] = pen.id
+
+    def remove_animal_from_herd(self, animals_removed):
+        """
+        Removes the animal IDs from the animal_to_pen_id_map dictionary pertaining to the
+        animals that have been removed from the herd, and updates the stocking density
+        of the pens they were in. This function is void.
```
KFosterReed: What does 'This function is void' mean? Is there a missing part of this doc string? 

Also, in light of a conversation we had during the fixathon checkin maybe a more succinct definition for this function would be: 

Deletes the IDs of animals from animal_to_pen_id_map dictionary when the animal was removed from the herd; updates the relevant pen's stocking density.

JoeWaddell: I think "This function is void" is explicitly noting that it doesn't return anything.

ancheypeng: It might be more clear to change that sentence to something like, "This function returns nothing" or "This function is a void type function".

RUFAS/routines/animal/animal_management.py
```diff
@@ -331,17 +330,66 @@ def calc_nutrient_rqmts(self, feed, temp):
...
+            pen_population_before_additions[index] = len(pen.animals_in_pen)
+
+        return pen_population_before_additions
+
+    def calculate_pen_rations(self, prior_pen_populations):
+        """
+        Calculates the ration of every pen in the simulated farm. This is done by
+        updating the values of each Pen object's ration dictionary, where the keys
+        pertain to certain food items in a ration. This function is void.
+
```
KFosterReed, line 413: I am trying to think of a more succinct way to say this. What about: 

Adjusts the amount of each feed within a ration that is delivered to a pen when the number of animals in the pen is changed

RUFAS/routines/animal/animal_management.py
```diff
@@ -108,10 +108,9 @@ def __init__(self, data, config, feed, weather, time):
         self.cows_culled = []
 
         # list of all the pens on the farm
-        self.all_pens = []
-
+        self.all_pens_ids = []
```
ancheypeng: This might be nitpicky or just in my head, but it seems to me like `all_pens_ids` is less clear than `all_pen_ids` because the list contains all of the "pen_ids" not all of the "pens_ids"

```suggestion
        self.all_pen_ids = []
```

RUFAS/output_handler/reports/pen_report.py
```diff
@@ -13,7 +13,7 @@
 class PensReport(BaseReportDriver):
     def __init__(self, data, state):
         super().__init__(data)
-        for pen in state.animal_management.all_pens:
+        for pen in state.animal_management.all_pens_ids:
             self.reports['pen_' + str(pen.id)] = PenReport(data, state.feed, pen, pen.id)
```
PooyaHekmati: if we are looping through all_pens_ids then the variable in the loop should be pen_id rather than pen. This means that in the next line, we will have pen_id.id which doesn't sounds right. There is something wrong in here.

input/animal_management.json
```diff
@@ -2,12 +2,12 @@
...
     {
         "start_date" : "2009:244",
-        "end_date" : "2009:264",
+        "end_date" : "2012:245",
         "csv_dir": "output/CSVs/",
         "graphic_dir": "output/graphics/",
         "run_tests": true,
         "set_seed": true,
-        "seed": 42,
+        "seed": 41,
```
PooyaHekmati: Per convention, the seed has to be 42.

PooyaHekmati (changes requested):

RUFAS/routines/animal/animal_management.py
PooyaHekmati: The fact that this function doesn't return anything has to be communicated using type annotation:
_def foo(args) -> None:_
Also, if it is important to reiterate this in docstring as well, it has to be done like:
Returns
\--------
            None

RUFAS/routines/animal/animal_management.py
```diff
@@ -331,17 +330,66 @@ def calc_nutrient_rqmts(self, feed, temp):
...
         associated with the pen that animal is in.
         """
-        for pen in self.all_pens:
+        for pen in self.all_pens_ids:
             animals_in_pen = pen.animals_in_pen
             for animal in animals_in_pen:
-                self.id_pen[animal.id] = pen.id
+                self.animal_to_pen_id_map[animal.id] = pen.id
+
+    def remove_animal_from_herd(self, animals_removed):
```
ancheypeng:
```suggestion
    def remove_animals_from_herd(self, animals_removed):
```

Would this make more sense to be `remove_animals_from_herd` because its removing multiple animals?

PooyaHekmati:
Newly developed functions need to be type annotated and have unit tests.

input/animal_management.json
Fjordrunner: Just switched this, thanks for letting me know!

RUFAS/routines/animal/animal_management.py
Fjordrunner: Thanks for catching that, good point!

RUFAS/routines/animal/animal_management.py
Fjordrunner: Good catch, just changed that!

RUFAS/routines/animal/animal_management.py
Fjordrunner: I added this change, pushing a commit with it right now!

Fjordrunner:
> Newly developed functions need to be type annotated and have unit tests.

I was planning on adding those in a separate PR to immediately follow this one, since this PR is already on the larger side, but if you think that it is appropriate to add them here then I will do so.

RUFAS/output_handler/reports/pen_report.py
Fjordrunner: Good catch. I've done some more investigating and all_pens_ids should have never referred to pens, as its contents are the Pen objects on a farm. However, that being said, there are some places in the code where this seems inconsistent, so this will probably require a lengthier conversation with @KFosterReed

RUFAS/routines/animal/animal_management.py
Fjordrunner: Good point. However, we're migrating away from the old name entirely as the class variable was changed to `all_pens` . In actuality, the variable contains a list of Pen objects and not their IDs, so this change was definitely needed.

RUFAS/output_handler/reports/pen_report.py
Fjordrunner: As per our meeting, all instances of `all_pens_ids` are now more appropriately named `all_pens` . Once again, thanks for catching this!

RUFAS/routines/animal/animal_management.py
Fjordrunner: I removed the sentences indicating that a given function was void, and added type annotations to the 4 functions I created/edited. Should be good to go now, let me know if I can make anything clearer!

RUFAS/routines/animal/animal_management.py
```diff
@@ -354,126 +402,53 @@ def daily_update_id_pen(self, animals_added, ids_removed, calves_born, feed, tem
         they were assigned.
```
KFosterReed, line 429: How about: 

Updates the dictionary that maps animal objects to the pen they are housed in when new animals are born or purchased and when animals leave the herd due to death or culling

Fjordrunner: I like this, this sounds good! I think that we are operating off of IDs so I will make that distinction once confirming, but I'll be pushing this now! Only have unit tests and this PR should be good to go once that is done!

KFosterReed (approved):
This seems to work as expected! It doesn't fix our overall challenge of moving animals between pens but the animals added and removed are working as expected and the clarity is hugely improved :)

ancheypeng (approved):
Looks good to me!

JoeWaddell (approved):
Looks good to me! Nice work, and I think the issues raised in previous reviews have been worked out in the revision process.

RUFAS/routines/animal/animal_management.py
```diff
@@ -355,153 +363,122 @@ def calc_nutrient_rqmts(self, feed, temp):
         for cow in self.cows:
             cow.set_nutrient_rqmts()
 
-    def fully_update_id_pen(self):
+    def fully_update_animal_to_pen_id_map(self):
```
PooyaHekmati:
```suggestion
    def fully_update_animal_to_pen_id_map(self) -> None:
```

RUFAS/routines/animal/pen.py
```diff
@@ -318,7 +311,7 @@ def set_milk_avgs(self, avg_milk: float, avg_CP_milk: float) -> None:
         self.avg_milk = avg_milk
         self.avg_CP_milk = avg_CP_milk
 
-    def add_new_animals(self, new_animals: List[Calf | Cow | HeiferI | HeiferII | HeiferIII]) -> None:
+    def add_new_animals(self, new_animals: List[Any]) -> None:
```
PooyaHekmati: Why is type Any used here? That is not a good practice and should only be used when no other solution is possible.

RUFAS/routines/animal/pen.py
```diff
@@ -675,8 +668,27 @@ def set_up_new_animal(self, animal, p_conc, feed, temp, num_animals_before_addit
         animal.p_intake = self.avg_p_intake
 
         self.animals_in_pen.append(animal)
-        # updating stocking density
-        self.stocking_density = len(self.animals_in_pen) / self.num_stalls * 100
+
+    def remove_animals_by_ids(self, animal_ids: List[int]):
```
PooyaHekmati:
```suggestion
    def remove_animals_by_ids(self, animal_ids: List[int]) -> None:
```

PooyaHekmati (changes requested):
LGTM overall.
As we discussed and agreed in the meeting, I was expecting a git issue linked to this PR both in the PR description/comments and as a comment in the code to document the faulty logic. Since I don't see it, I'd like to know the reason.
Sending back to fix those two minor comments.

Fjordrunner:
> LGTM overall. As we discussed and agreed in the meeting, I was expecting a git issue linked to this PR both in the PR description/comments and as a comment in the code to document the faulty logic. Since I don't see it, I'd like to know the reason. Sending back to fix those two minor comments.

The issue we mentioned in the former meeting is seemingly fixed via some new code that Loi and I wrote. The problem that still persists is the one elaborated upon in [https://github.com/RuminantFarmSystems/MASM/issues/189](url). That will be the next step of making sure the animal-to-pen allocation is being done correctly.

PooyaHekmati (approved):

PooyaHekmati:
Please fix merge conflicts.

PR #221, opened 2022-11-29
Nasem_update

JoeWaddell:
Hi folks: buckle up for a long explanation ahead. So I've been helping Edward get up to speed in Python, and we've been going through this part of the code together. As he presented in the RuFaS meeting, the new standards for calculating various nutrient requirements are the "NASEM" equations, so we're updating this section accordingly. Since RuFaS currently uses NRC equations, I'm "simply" adding an option to use either NASEM or NRC equations.

Hopefully it's clear from the changes made, but for more detail: currently, `calc_rqmts` calculates each of the values in the return statement in one long sequence. What I propose is that we'll instead first decide which set of equations to use, and then call an appropriate series of functions. The current calculations for each return value will act as the basis for the series of NRC functions; Edward and I will continue to write up the NASEM equations from the literature. This should allow for a series of shorter functions that each handle a single nutrient requirement, each with a detailed docstring. 

What you'll note is that I've only changed a section towards the end of the `calc_rqmts` function. My reasoning is that I'd like to ask for input in this draft PR about the overall structure/logic of the proposed change. Right now, selecting which set of equations to use follows simple "if-else" logic, but I'm assuming there is a more elegant way to do this. Likewise, I'd appreciate any input on the functions themselves, before writing functions for the rest of the equations in the file. This will also help the User-defined Ration functionality that I'm working on - but more to come on that later.

Comments and assorted workflow questions: 

- Important note: the NRC option will be removed at some (unknown) point in the future, so this is a temporary step. The option to switch between the two has a few uses in the meantime: as some users may not have yet switched to NASEM. Also, it allows us to check that the NASEM equations are comparable to NRC equations across many scenarios. This is especially helpful for someone like Edward who wants to tinker with inputs to see how things change under wildly different conditions. 

- I've put the option in the "management_decisions" part of the `animal_management_animal.json`, and call it by importing `animal_base `into `animal_requirements.py`. Question: Should the value be assigned to a variable, or "checked" again at each stage in the decision tree? Or should this be done at an individual animal level, or using an entirely different method/logic?

- Elsewhere in the code it seems that Body Condition Scores are being provided as 3 in most (if not all) cases (see e.g. `ration_driver.py`), but this means that some individuals (non heifers) are left with "None." That didn't affect the equations before, as it was only checking the score for heifers, but now it requires that input for all individuals, instead of changing code elsewhere, I'm assigning 3 where values return None. Is there a better way to set 3 as the default if None?

[...]

PR #202, opened 2022-10-19
Animal life cycle refactoring

ltp23:
Some notes about the main changes in this PR:
- `life_cycle.py`: This refactoring focuses on breaking up the two main functions `initialize_herd` and `daily_update` into smaller, more manageable, and testable private functions. 
- `animal_typed_dicts.py`: list of dictionaries that mirror what we have in the animal config json. The main purpose is to enable type-checking and make sure we don't use an undeclared key.
- `clustering_pen_grouping`: there seems to be an occasional numerical error when comparing percentages. For example, 1.00000000000002 is supposed to be equal to 1.0, but due to rounding errors or inherent imperfect representation of floating point numbers, they are unequal. So one way to fix that is to check the absolute difference between the two numbers and compare it to a tolerance level. There is a helper function for this: `math.isclose(x, y, rel_tol=1e-09)`

PR #284, opened 2023-01-12
Animal tests

JoeWaddell:
Started to remove tests that are no longer needed (methods removed from code), but holding off on that until refactor is complete on associated modules.

In the meantime, I've begun to add test coverage, and have tests here - and some fixtures - that should be useful in writing additional tests.


JoeWaddell:
I need to resolve the conflicts on this branch before I can push it as a proper PR (there's some cleaning up I need to do), but for now: we can use the google doc  to track what tests are needed in the short term.  I've also added it to the "test coverage" item in the Fixathon 2023 Live document on MASM. The rankings in the sheet are based off an older analysis, but is a good rough estimate of the importance of different functions. Just note that many have changed names since, and/or are involved in refactors.

https://docs.google.com/spreadsheets/d/1E0wHZEfw3KpRt28H81drblE2wYN1fXLE/edit?usp=sharing&ouid=111242254510123751441&rtpof=true&sd=true

The main refactors to worry about as we add test coverage are:
My ongoing requirements update: [#221](https://github.com/RuminantFarmSystems/MASM/pull/221)
Loi's temporarily closed refactor: https://github.com/RuminantFarmSystems/MASM/pull/202
and the animal_management stuff: https://github.com/RuminantFarmSystems/MASM/pull/283

tests/test_animal.py
```diff
@@ -41,11 +43,6 @@ def test_calc_avg_nutrient_rqmts():
     pass
 
 
-def test_calc_avg_stats():
```
JoeWaddell, line 533: Removed this test because it's no longer in pen.py. Checking the remaining tests in this file is an ongoing task for this branch.

JoeWaddell:
Adding a few animal tests and associated fixtures. They aren't all terribly elegant, but pass in pytest on my machine, and - if approved - should be useful in writing more tests. Previous comment touches on ongoing plan for this branch, to improve test coverage in the animal module.

tests/test_animal.py
```diff
@@ -131,14 +170,58 @@ def test_set_p_purchased():
...
+    # Case 1
+    # update hist with designated vals, using the time and the obj itself
+    cow_fixture.update_pen_history(
+        curr_pen, curr_day, classes_in_pen)
+    assert cow_fixture.pen_history[-1].pen == 3
+    assert cow_fixture.pen_history[-1].classes_in_pen == ['Cow']
+    assert cow_fixture.pen_history[-1].start_date == 2
+    assert cow_fixture.pen_history[-1].end_date == 2
+
+    # Case 2
```
ancheypeng, line 655: Would it be possible to parameterize the test case using pytest's [mark.parameterize](https://docs.pytest.org/en/6.2.x/parametrize.html) functionality?

JoeWaddell: I think so: I'll update this test (and others) to take advantage of how concise that functionality makes the tests.

JoeWaddell: I updated the others to use mark.parametrize, but for this specific test I couldn't figure out a way to elegantly (and easily read by later devs) how to test the third case: whereby we're checking another event was appended to an already existing list, because it required running the test twice (and in the first two cases is only called once).

ancheypeng: I'm not sure if this would help the readability overall, but a potential idea would be to have a list of arguments for the calls to `update_pen_history` as one of the parametrization parameters, for example, `[(3, 2, ['Cow'])]` for case 1, `[(3, 2, ['Cow']), (4, 3, ['Cow'])]` for case 2, and `[(3, 2, ['Cow']), (4, 3, ['Cow']), (4, 4, ['Cow'])]` for case 3. Then `update_pen_history` would be called once for every element in the list -- for case 3, `update_pen_history` would be called 3 times. (I'm also not sure if I'm interpreting the problem with parameterization correctly, so this suggestion might not be helpful)

tests/test_animal.py
```diff
@@ -620,24 +639,91 @@ def test_set_p_purchased():
...
-def test_update_body_weight_history():
-    """Unit test for function update_body_weight_history in file routines/animal/life_cycle/animal_base.py"""
-    pass
+def test_update_pen_history(cow_fixture: AnimalBase) -> None:
+    """Unit test for update_pen_history in file routines/animal/life_cycle/animal_base.py"""
+
+    # Case 1
+    # update hist with designated vals, using the time and the obj itself
+    cow_fixture.update_pen_history(
+        3, 2, ['Cow'])
```
ltp23, line 648: It would be easier to read if we include the name of each argument. 

```
cow_fixture.update_pen_history(
    curr_pen=3,
    curr_day=2,
    classes_in_pen=['Cow']
)
```

tests/test_animal.py
ltp23: I think if you want to use that pytest feature, each tuple should probably contain 6 elements - prev pen, prev day, prev classes in pen, curr pen, curr day, curr classes in pen. Next, call the update_pen_history() twice, one for each set of 3 elements. Then, make assertions.

ltp23 (approved):

ancheypeng (approved):
Looks good!

PR #285, opened 2023-01-12
Nitrogen composition

morrowcj:
This PR is meant to update the nitrogen_incorporation processes. It adds docstrings and switches the class to the composition pattern.

### DIFF
--- a/SC_redesign/Crop_and_Soil/crop/crop_data.py
+++ b/SC_redesign/Crop_and_Soil/crop/crop_data.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Optional
+from typing import Optional, List
 
 
 @dataclass
@@ -142,6 +142,7 @@ class CropData:
     root_depth: float = 1  # arbitrary
     """current depth of the plant roots in the soil (mm)"""
     # empty declarations
+    _nitrogen_shapes: Optional[List[float]] = None
     previous_nitrogen: Optional[float] = None
     shapes_nitrogen_uptake: Optional[float] = None
     optimal_nitrogen_fraction: Optional[float] = None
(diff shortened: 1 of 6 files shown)


SC_redesign/Crop_and_Soil/crop/crop_data.py
```diff
@@ -142,23 +149,37 @@ class CropData:
...
-    inaccessible_soil_layers: Optional[float] = None
+    """potential nitrogen to be taken up by the plant under ideal circumstances for the current day (kg/ha)"""
+    total_soil_layers: Optional[int] = None
+    """total number of layers in the soil profile (unitless)"""
+    accessible_soil_layers: Optional[int] = None
+    """number of layers in the soil profile that the plant roots have access to (unitless)"""
+    inaccessible_soil_layers: Optional[int] = None
+    """number of layers in the soil profile that the plant roots do not have access to (unitless)"""
     layer_nitrogen_potentials: Optional[float] = None
+    """potential nitrogen uptake from each soil layer (unitless)"""
```
ew3361zh: I guess I don't understand why these (`layer_nitrogen_potentials` and `unmet_nitrogen_demands`) are unitless. Is the former a measure of number of potential layers? The name to me reads like it is looking at a quantity of nitrogen at a particular layer. The latter as well seems like a measure of nitrogen which would seem like it would need a unit.

ew3361zh (approved):
Looks consistent with the other refactoring using the composite design but again, maybe good to have Loi look at it since he has more familiarity with the pattern. Testing worked as expected. Nice work.

SC_redesign/Crop_and_Soil/crop/crop_data.py
morrowcj: Oh, you know what. They should not be unitless. Good catch.

ltp23 (commented):
Everything looks really good here!

PR #286, opened 2023-01-12
Init sphinx

PooyaHekmati:
Compiled HTML pages from the current docstrings.

There were some errors, their list is in: https://pastebin.com/LqdgWw1K (password: kyfXLtutf4)


JoeWaddell:
So I'm seeing that it's not generating html for modules in my area, which is "deep" inside nested folders in the routines portion of the codebase. It goes as far as RUFAS.routines.animal, and the .pys in that dir, but it's not generating html for anything inside the folders RUFAS.routines.animal.life_cycle, RUFAS.routines.animal.manure, or RUFAS.routines.animal.ration

Is that defined by a "tree depth" type variable in initialization, or is it something we'll have to manually write in a .rst (or other type) file?

ew3361zh:
Is one of the goals for this PR to address the docstring errors?

PooyaHekmati:
> Is one of the goals for this PR to address the docstring errors?

No Niko, this is just to compile HTML from docstrings.

PooyaHekmati:
> So I'm seeing that it's not generating html for modules in my area, which is "deep" inside nested folders in the routines portion of the codebase. It goes as far as RUFAS.routines.animal, and the .pys in that dir, but it's not generating html for anything inside the folders RUFAS.routines.animal.life_cycle, RUFAS.routines.animal.manure, or RUFAS.routines.animal.ration
> 
> Is that defined by a "tree depth" type variable in initialization, or is it something we'll have to manually write in a .rst (or other type) file?

I think that's due to depth, we can change it later on. This pr is just to set the ball rolling.

JoeWaddell:
The other question is how to handle the CodeFactor errors. Since they're mostly Complex Code/Method complaints, with some empty files, is there a way to temporarily ignore those across the board? For example, if we are able to simply ignore errors that stem from the docs folder. 

Because otherwise, it looks good to me as a start! Nice work

PooyaHekmati:
> The other question is how to handle the CodeFactor errors. Since they're mostly Complex Code/Method complaints, with some empty files, is there a way to temporarily ignore those across the board? For example, if we are able to simply ignore errors that stem from the docs folder.
> 
> Because otherwise, it looks good to me as a start! Nice work

Yes, there used to be a way to tell code factor that it should ignore certain files. We may end up removing code factor from our pipeline altogether. Anyways, I wouldn't worry much about it right now.

ew3361zh (approved):
Very nice. HTML pages look incredible and are a testament to the great work so far by everyone on docstrings and for how wonderful Sphinx is.

JoeWaddell (approved):
Excellent! With all the comments in mind, this is looking great.

PR #287, opened 2023-01-13
Manure management

ltp23:
This PR is mainly about reviewing the design of the anaerobic digestion and lagoon combo class and adding unit tests for it. The type annotation of the config object for this class is a little clunky, so I may have to revisit this later on. I don't want to complicate other existing treatment classes just to accommodate this extra combo class.

### DIFF
--- a/RUFAS/routines/manure/manure_treatments/anaerobic_digestion_and_lagoon.py
+++ b/RUFAS/routines/manure/manure_treatments/anaerobic_digestion_and_lagoon.py
@@ -1,5 +1,8 @@
 from __future__ import annotations
 
+from typing import Tuple
+from typing import Union
+
 from RUFAS.routines.manure.manure_treatments.anaerobic_digestion import AnaerobicDigestion
 from RUFAS.routines.manure.manure_treatments.anaerobic_lagoon import AnaerobicLagoon
 from RUFAS.routines.manure.manure_treatments.base_manure_treatment import BaseManureTreatment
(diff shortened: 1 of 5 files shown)


RUFAS/routines/manure/manure_treatments/anaerobic_digestion_and_lagoon.py
```diff
@@ -19,11 +22,14 @@ class AnaerobicDigestionAndLagoon(BaseManureTreatment):
...
+                 manure_treatment_config: Union[ManureTreatmentConfig,
+                                                Tuple[ManureTreatmentConfig, ManureTreatmentConfig]]
+                 ) -> None:
         super().__init__(weather, time, manure_treatment_config)
-        self._anaerobic_digestion = AnaerobicDigestion(weather, time, manure_treatment_config)
-        self._anaerobic_digestion_daily_output = None
-        self._anaerobic_lagoon = AnaerobicLagoon(weather, time, manure_treatment_config)
+        self._anaerobic_digestion = AnaerobicDigestion(weather, time, manure_treatment_config[0])
+        self.anaerobic_digestion_daily_output = None
+        self._anaerobic_lagoon = AnaerobicLagoon(weather, time, manure_treatment_config[1])
```
morrowcj, line 32: How does this [line 32] work if the `manure_treatment_config` is only a single `ManureTreatmentConfig` object instead of a tuple? I'm unclear what the effect of indexing an empty value would be in this context.

morrowcj (approved):
This looks pretty good to me. I am a bit confused by the typing issue that you mention in your initial PR comment, but I think it can be addressed at a later time. 

The tests all run successfully and everything else make sense to me.

varrma (approved):

PR #288, opened 2023-01-17
Output handler 2.5 / 3

PooyaHekmati:
This PR is mainly about restructuring pools in the output manager. Now, each pool is a dictionary, which has a list for values and a list for info_maps. Items with the same indices in these two lists are from the same report. For example, the info_map of the 10th item in the values list is the 10th item in the info_maps list. Also, to reduce the pool size, class and function keys are not being stored in the info_map anymore, as they are already present in the key. 

Things remaining to un-draft the PR:
* fix type alias (add List[any])
* add unit tests


morrowcj (commented):
This looks pretty good to me. I think it was a good idea to make a custom type, it looks clean.

----

I do have one suggestion about future functionality. I think it would be really valuable to have some additional options associated with `add_variable`. Here's an example of what I would consider useful behavior in the various modules: 

```python
# add variable "x" once, with value 5
om.add_variable("x", 5, description="some details about x")
print(om.variables_pool["x"].value)  # 5
print(om.variables_pool["x"].description)  # "some details about x"

# overwrite "x" with 7
om.add_variable("x", 7, overwrite=True)
print(om.variables_pool["x"].value)  # 7

# create variable "z" and append to it
for i in range(5):
    om.add_variable("z", i, append=True, description=["details about z", "tracked through time"])
print(om.variables_pool["z"].value)  # [0, 1, 2, 3, 4]
print(om.variables_pool["z"].description)  # ["details about z", "tracked through time"]
```

I think that `description` may be what `info.map` does, but hopefully my intention is understood.

morrowcj (approved):

varrma:
Looks good to me

RUFAS/output_manager.py
```diff
@@ -36,17 +37,36 @@ def __new__(cls):
...
-
-    def add_variable(self, name: str, value: Any,
-                     info_map: Dict[str, Union[str, bool]]) -> None:
+                                   "function": self.__init__.__name__})
+
+    def _pool_element_factory(self) -> pool_element_type:
+        """Factory for elements added to pools"""
+        info_maps: List[Dict[str, Any]] = []
+        values: List[Any] = []
+        return {'info_maps': info_maps, 'values': values}
```
JoeWaddell, line 54: This (and a few other cases further below) might be the source of the problem I'm struggling with right now: I ran the simulation, and it generates the output JSONs, but since they're using single quotes in many/all(?) positions, I'm not able to read in the data in Python using json.load (and Notepad++ can't parse it either), so I'm unable to fully explore how the data are structured in their final forms, in practice. That said, I have been checking variables/warnings/logs in the debugger, and things look good there.

I tried the approach of replacing single quotes with double quotes on the files, but there's an unmatched '{' (or '}') somewhere, so it's not happy with that solution either...I'm assuming this is something that will be handled in the final part of the Output Handler, but wanted to draw attention to it at this stage in case it's a valid concern.

KFosterReed:
So I started to review this last week and realized that my python data skills are not good enough to try to parse the json files into a meaningful output. Rather than waiting for me to learn that skill I thought I might just comment that it seems like the info maps for the life cycle information could use some work. Specifically, the date is repeated several times for each set of values without any description of what the replications on each day means.

image

PR #289, opened 2023-01-17
Manure management

ltp23:
Adding a class that parses the config json file for the manure management module. This class acts as a stop-gap that will probably and eventually be replaced by a proper input handler. 

Thank you all for reviewing!


ew3361zh (approved):
Looks really good, Loi! Testing all worked and passed as intended locally.

RUFAS/routines/manure/input_handler/manure_management_config_handler.py
```diff
@@ -0,0 +1,228 @@
...
+
+        """
+        bedding_config_by_bedding_type: Dict[BeddingType, BeddingConfig] = {}
+        for json_bedding_config in bedding_json_configs:
+            bedding_type = BeddingType.get_type(json_bedding_config['bedding_type'])
+            del json_bedding_config['bedding_type']
+            bedding_config_by_bedding_type[bedding_type] = BeddingConfig(
+                    **json_bedding_config,
+                    bedding_type=bedding_type
+            )
```
morrowcj, line 130: Out of curiosity, why do you need to delete the "bedding_type" element from `json_bedding_config`?

tests/test_manure_module/test_manure_management_config_handler.py
```diff
@@ -0,0 +1,752 @@
...
+        ),
+        mocker.call(
+                bedding_mass_per_day=25.0,
+                bedding_density=1500.0,
+                bedding_dry_matter_content=0.9,
+                bedding_cleaned_fraction=1.0,
+                sand_removal_efficiency=1.0,
+                bedding_type=BeddingType.SAND
+        )
+    ]
```
morrowcj, line 87: I didn't know you could do this - I'm going to steal this pattern. I only knew about `assert_called_with()`, which doesn't like multiple callings.

morrowcj (approved):
This looks good to me. The class is very clear and well documented and the tests are comprehensive. Nice work. 

I had just a couple of questions about the code.

varrma (approved):

tests/test_manure_module/test_manure_management_config_handler.py
```diff
@@ -0,0 +1,752 @@
...
+
+    patch_for_bedding_type_get_type = mocker.patch(
+            'RUFAS.routines.manure.input_handler.manure_management_config_handler.BeddingType.get_type',
+            side_effect=[
+                BeddingType.SAWDUST,
+                BeddingType.MANURE_SOLIDS,
+                BeddingType.SAND
+            ]
+    )
```
morrowcj, line 45: Could you explain what the `side_effect` argument does? Are you indicating that those attributes will get used/modified?

The documentation just says
> side_effect allows you to perform side effects, including raising an exception when a mock is called

Which is very unhelpful.

PR #290, opened 2023-01-18
Phosphorus morrowcj

morrowcj:
## Summary
This PR refactors the old crop methods for `PhosphorusIncorporation` (`phosphorus_incorporation.py`). The methods used to calculate phosphorus uptake are basically identical to those used to calculate nitrogen uptake. Therefore, `PhosphorusIncorporation` uses many of the static methods of `NitrogenIncorporation`. 

Tests were also implemented for the new code. Many of the phosphorus tests are also extremely similar (or identical) to those of nitrogen.

## Other changes
* moved the script which finds project TODOs

## Notes 
`PhosphorusIncorporation` and `NitrogenIncorporation` should probably be combined into a single `NutrientIncorporation` class at some point, but this will have to be a a future project.


SC_redesign/Crop_and_Soil/crop/crop.py
```diff
@@ -36,6 +37,8 @@ def __init__(self, crop_data: Optional[CropData] = None):
         """Process component controlling plant water dynamics"""
         self.nitrogen_incorporation = NitrogenIncorporation(data)
         """Process component controlling plant nitrogen incorporation, including uptake and fixation"""
+        self.phosphorus_incorporation = PhosphorusIncorporation(data)
+        """Process component controlling plant nitrogen incorporation, including uptake and fixation"""
```
morrowcj:
```suggestion
        """Process component controlling plant phosphorus uptake and incorporation"""
```

SC_redesign/Crop_and_Soil/crop/nitrogen_incorporation.py
```diff
@@ -191,41 +174,41 @@ def fix_nitrogen(self, water_factor: float) -> None:
...
-            emergence_nitrogen_fraction: nitrogen fraction at emergence
-            half_mature_nitrogen_fraction: nitrogen fraction at half-maturity
-            near_mature_nitrogen_fraction: nitrogen fraction *near* maturity
-            mature_nitrogen_fraction: nitrogen fraction *at* maturity
+            emergence_nutrient_fraction: nitrogen fraction at emergence
+            half_mature_nutrient_fraction: nitrogen fraction at half-maturity
+            near_mature_nutrient_fraction: nitrogen fraction *near* maturity
+            mature_nutrient_fraction: nitrogen fraction *at* maturity
 
         Returns: list of the first and second shape coefficients, respectively
```
Isgotkowitz, line 195: Could you a SWAT reference here

SC_redesign/Crop_and_Soil/crop/phosphorus_incorporation.py
```diff
@@ -1 +1,137 @@
...
\ No newline at end of file
+from typing import Optional, List
+from SC_redesign.Crop_and_Soil.crop.crop_data import CropData
+from SC_redesign.Crop_and_Soil.crop.nitrogen_incorporation import NitrogenIncorporation
+
+# TODO: Phosphorus has identical methods to NitrogenIncorporation, excluding nitrogen fixation. They
+#  should be combined into one class (NutrientIncorporation) and simplified.
+
+"""
+This module is based upon the 'Phosphorus Uptake" section of of the SWAT model documentation
```
Isgotkowitz: Can you add the SWAT section numbers here

SC_redesign/Crop_and_Soil/crop/nitrogen_incorporation.py
```diff
@@ -407,43 +390,43 @@ def _determine_nitrogen_uptake_to_depth(demand: float, depth: float, root_depth:
...
+        if len(layer_uptake_potentials) != len(layer_demands) or len(layer_uptake_potentials) != len(layer_nutrient):
             raise ValueError("layer_potential, layer_demand, and layer_nitrate must be the same length")
         # calculate results
         layer_desired = [potential + demand for potential, demand in zip(layer_uptake_potentials, layer_demands)]
-        return [min(desired, nitrate) for desired, nitrate in zip(layer_desired, layer_nitrates)]
+        return [min(desired, nitrate) for desired, nitrate in zip(layer_desired, layer_nutrient)]
 
     @staticmethod
-    def _determine_layer_extracted_resource(requests: List[float], sources: List[float]) -> List[float]:
+    def determine_layer_extracted_resource(requests: List[float], sources: List[float]) -> List[float]:
```
Isgotkowitz, line 443: This method needs a comment

Isgotkowitz (changes requested):
Looks good, only work that needs to be done is adding SWAT references to methods

SC_redesign/Crop_and_Soil/crop/nitrogen_incorporation.py
morrowcj: Done.

SC_redesign/Crop_and_Soil/crop/phosphorus_incorporation.py
morrowcj: Done

SC_redesign/Crop_and_Soil/crop/nitrogen_incorporation.py
morrowcj: And done.

SC_redesign/Crop_and_Soil/crop/nitrogen_incorporation.py
Isgotkowitz: I think this method should also say that it references SWAT 5:2.3.2 and 3

SC_redesign/Crop_and_Soil/crop/nitrogen_incorporation.py
```diff
@@ -311,23 +296,23 @@ def _determine_optimal_nitrogen(fraction: float, whole: float) -> float:  # pseu
...
         Args:
             demand: maximum/optimal nitrogen uptake of the plant on a given day
-            nitrogen_start: nitrogen biomass at the end of the previous day
-            mature_nitrogen_fraction: nitrogen fraction at maturity
+            nutrient_start: nitrogen biomass at the end of the previous day
+            mature_nutrient_fraction: nitrogen fraction at maturity
             max_growth: maximum potential biomass the plant can gain on a given day
 
         Returns: the potential nitrogen uptake for the day
         """
```
Isgotkowitz, line 319: This is SWAT 5:2.3.5, I believe

Isgotkowitz (changes requested):
I don't think I was clear enough before. I think it is a good idea for *every* method that is associated with a specific equation in SWAT to have a reference to that equation in its description.

SC_redesign/Crop_and_Soil/crop/nitrogen_incorporation.py
morrowcj: My bad, I thought they were. Thanks for the added clarity.

morrowcj:
OK, @Isgotkowitz. I think that all the methods that can be linked to SWAT now contain references. Thanks for your patience.

Isgotkowitz (approved):
Looks gucci to me

ltp23 (approved):
Great job, Clay!

If you think both nitrogen and phosphorus incorporation classes are similar enough to warrant using inheritance, then I think you should do it to make things more maintainable for you. 

For those static methods, you can simplify by using `self` like `self.some_static_method()` since they reside in the same class.

morrowcj:
> If you think both nitrogen and phosphorus incorporation classes are similar enough to warrant using inheritance, then I think you should do it to make things more maintainable for you.

That isn't a bad idea. I think that might be a job for a different PR though. 

I'm going to merge this.

PR #292, opened 2023-01-18
leaf_area_index wiki

LaplacetteCM:
This is the first draft of the leaf_area_index.md wiki file

### DIFF
--- a/SC_redesign/Crop_and_Soil/docs/leaf_area_index.md
+++ b/SC_redesign/Crop_and_Soil/docs/leaf_area_index.md
@@ -1,0 +1,100 @@
+### __Leaf area index__
+
+The `LeafAreaIndex`class handles the seasonal changes in canopy height and
+leaf area for each crop during the growing season. The **leaf area index**
+(*LAI*) is defined as the area of green leaf per unit area of land, having a 
+maximum value that, once reached, remains constant until leaf senescence 
+becomes the dominant growth process.
+
+The main function, `grow_canopy()`, returns on a daily basis the updated 
+**optimal leaf area fraction**, **shape coefficients**, **canopy height**,
+**maximum leaf area change** and **senescent leaf area index** to finally 
+determine the **leaf area added** every day.
+
+The **optimal leaf area fraction** controls the canopy height and leaf area
+until the maximum *LAI* is reached and represents the fraction of the 
+plant’s maximum *LAI* corresponding to a given fraction of potential heat 
+units. To calculate this fraction on a given day, RuFaS needs to first 
+calculate two **shape coefficients** and the **fraction of potential heat 
+units** accumulated. Then, the **optimal leaf area fraction** is calculated 
+as follows:
+
+```math
+  $$ fr_{\text{LAI, opt}} = \frac{h}{h+exp(l_{\text{1}}-l_{\text{2}}\times h)} 
+       
+$$
+```
+where $fr_{\text{LAI, opt}}$ is the `optimal_leaf_area_fraction`, $l_{\text{1}}$ 
+is `first_shape`, $l_{\text{2}}$ is `second_shape` and $h$ is `heat_fraction`, 
+the fraction of potential heat units accumulated for the plant on a given day.
+
+**Canopy height** represents the canopy height on a given day, in meters, 
+and is calculated using the following equation:
+
+```math
+  $$ h_{\text{c}} = 
+  h_{\text{c,max}}\times \sqrt{fr_{\text{LAI,opt}}}     
+$$
+```
+
+where $h_{\text{c}}$ is `canopy_height`, $h_{\text{c,max}}$ is `max_canopy_height`, 
+and $fr_{\text{LAI,opt}}$ is `optimal_leaf_area_fraction`. Once $h_{\text{c,max}}$ 
+is reached, $h_{\text{c}}$ remains constant. 
+
+The **maximum leaf area change** calculates the maximum or potential leaf 
+area added during the day using the following equation:
+
+```math
+  $$ LAI_{\text{pot}} = (fr_{\text{LAI, opt}}-fr_{\text{LAI, opt i-1}}) \times LAI_{\text{max}} \times (1-exp(5\times (LAI_{\text{i-1}}-LAI_{\text{max}})))
+       
+$$
+```
+
+where $LAI_{\text{pot}}$ is the `maximum_leaf_area_change` on a given day, $fr_{\text{LA, opt}}$  
+is the `optimal_leaf_area_fraction` on a given day, $fr_{\text{LA, opt i-1}}$ 
+is `previous_optimal_leaf_area_fraction`, the optimal leaf area fraction of the previous day, $LAI_{\text{max}}$ 
+is `max_leaf_area_index`, the maximum leaf area index achievable by the plant, 
+and $LAI_{\text{i-1}}$ is `previous_leaf_area`, the leaf area index on the previous day.
+
+When **senescence** starts, *LAI* is calculated as follows: 
+
+```math
+  $$ LAI_{\text{i}} = LAI_{\text{max}} \times \frac {(1-h)}{(1-h_{\text{sen}})}
+$$
+```
+where $LAI_{\text{i}}$ is the `leaf_area_index` on a given day, $LAI_{\text{max}}$ is `max_leaf_area_index`,
+the maximum leaf area index achievable by the plant, $h$ is `heat_fraction`,
+the fraction of potential heat units accumulated for the plant on a given 
+day and $h_{\text{sen}}$ is `senescent_heat_fraction`, the fraction of 
+potential heat units at which senescence begins.
+
+The **actual leaf area added** on a given day is based upon the potential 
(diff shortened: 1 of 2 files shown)


SC_redesign/Crop_and_Soil/docs/leaf_area_index.md
```diff
@@ -0,0 +1,100 @@
...
+units** accumulated. Then, the **optimal leaf area fraction** is calculated 
+as follows:
+
+```math
+  $$ fr_{\text{LAI, opt}} = \frac{h}{h+exp(l_{\text{1}}-l_{\text{2}}\times h)} 
+       
+$$
+```
+where $fr_{\text{LAI, opt}}$ is the `optimal_leaf_area_fraction`, $l_{\text{1}}$ 
+is `first_shape`, $l_{\text{2}}$ is `second_shape` and $h$ is `heat_fraction`,
```
morrowcj: To better match the code, can we do this:

```suggestion
is the first of two shape parameters (`_lai_shapes`), $l_{\text{2}}$ is the second shape parameter, and $h$ is `heat_fraction`, 

```

SC_redesign/Crop_and_Soil/docs/leaf_area_index.md
```diff
@@ -0,0 +1,100 @@
...
+The **optimal leaf area fraction** controls the canopy height and leaf area
+until the maximum *LAI* is reached and represents the fraction of the 
+plant’s maximum *LAI* corresponding to a given fraction of potential heat 
+units. To calculate this fraction on a given day, RuFaS needs to first 
+calculate two **shape coefficients** and the **fraction of potential heat 
+units** accumulated. Then, the **optimal leaf area fraction** is calculated 
+as follows:
+
+```math
+  $$ fr_{\text{LAI, opt}} = \frac{h}{h+exp(l_{\text{1}}-l_{\text{2}}\times h)}
```
morrowcj: I'm not a huge fan of the "fr" notation that SWAT uses. Would it look better if we replace "fr" with a different variable? What about something like a lowercase "p" for "proportion"?

SC_redesign/Crop_and_Soil/docs/leaf_area_index.md
```diff
@@ -0,0 +1,100 @@
...
+where $LAI_{\text{pot}}$ is the `maximum_leaf_area_change` on a given day, $fr_{\text{LA, opt}}$  
+is the `optimal_leaf_area_fraction` on a given day, $fr_{\text{LA, opt i-1}}$ 
+is `previous_optimal_leaf_area_fraction`, the optimal leaf area fraction of the previous day, $LAI_{\text{max}}$ 
+is `max_leaf_area_index`, the maximum leaf area index achievable by the plant, 
+and $LAI_{\text{i-1}}$ is `previous_leaf_area`, the leaf area index on the previous day.
+
+When **senescence** starts, *LAI* is calculated as follows: 
+
+```math
+  $$ LAI_{\text{i}} = LAI_{\text{max}} \times \frac {(1-h)}{(1-h_{\text{sen}})}
```
morrowcj: Similar to my above comment, I'm also not really a fan of using abbreviations for a single symbol (this could be confused for $L \times A \times I$). What if we use $A_{\text{leaf}}$ to represent leaf **area**?

I don't feel strongly about this or my previous notation comment, so, if you think it is more clear to use "LAI", feel free to say so and leave it as is.

morrowcj (approved):
I think this looks really good. The documentation is extremely clear and concise (less than 100 lines!). I have a few minor comments about notation, but nothing big.

SC_redesign/Crop_and_Soil/docs/leaf_area_index.md
LaplacetteCM: True. I think the use of LAI is very popular within the ag world and probably everyone would understand it, but in terms of the code I agree with your suggestion of using clearer notation. However, I think that we must keep the name of the variables the way it is now here and in the code (*e.g.*, Leaf area is not the same as leaf area index). 
So I would say: " $A_{\text{leaf, i}}$ is the `leaf_area_index` on a given day". Otherwise, we would be changing the meaning of the equation. Does it make sense? 

Note: I've changed both the fraction and LAI notation to A+subscripts

SC_redesign/Crop_and_Soil/docs/leaf_area_index.md
LaplacetteCM: perfect!

SC_redesign/Crop_and_Soil/docs/leaf_area_index.md
morrowcj: Oh, good point! I meant to say "leaf area index". I think your solution is great.

morrowcj:
Looks great! I don't think we need to wait for another review, so I'm going to go ahead and force the merge, if you're ready - just let me know.

SC_redesign/Crop_and_Soil/docs/leaf_area_index.md
```diff
@@ -0,0 +1,78 @@
+### __Leaf area index__
+
+The `LeafAreaIndex`class handles the seasonal changes in canopy height and leaf area for each crop during the growing season. The **leaf area index** (*LAI*) is 
+defined as the area of green leaf per unit area of land, having a maximum value that, once reached, remains constant until leaf senescence becomes the dominant 
+growth process.
+
+The main function, `grow_canopy()`, returns on a daily basis the updated **optimal leaf area fraction**, **shape coefficients**, **canopy height**, **maximum leaf 
+area change** and **senescent leaf area index** to finally determine the **leaf area added** every day.
+
+The **optimal leaf area fraction** controls the canopy height and leaf area until the maximum *LAI* is reached and represents the fraction of the plant’s maximum
```
ew3361zh: It seems like it could be helpful to the reader if you have bolded a term previously, to continue bold it throughout. e.g. **canopy height** being bolded on line 10 to be consistent with its bolding on line 7.

SC_redesign/Crop_and_Soil/docs/leaf_area_index.md
```diff
@@ -0,0 +1,78 @@
...
+
+The `LeafAreaIndex`class handles the seasonal changes in canopy height and leaf area for each crop during the growing season. The **leaf area index** (*LAI*) is 
+defined as the area of green leaf per unit area of land, having a maximum value that, once reached, remains constant until leaf senescence becomes the dominant 
+growth process.
+
+The main function, `grow_canopy()`, returns on a daily basis the updated **optimal leaf area fraction**, **shape coefficients**, **canopy height**, **maximum leaf 
+area change** and **senescent leaf area index** to finally determine the **leaf area added** every day.
+
+The **optimal leaf area fraction** controls the canopy height and leaf area until the maximum *LAI* is reached and represents the fraction of the plant’s maximum 
+*LAI* corresponding to a given fraction of potential heat units. To calculate this fraction on a given day, RuFaS needs to first calculate two **shape
```
ew3361zh: What do you think about rewording "RuFaS needs to first calculate..." to "RuFaS first calculates"? The former sounds a little like RuFaS may not currently be doing the calculating and the latter makes it sound like a description of a process that is happening. Very nitpicky so if you like it the way it's written please leave it.

SC_redesign/Crop_and_Soil/docs/leaf_area_index.md
```diff
@@ -0,0 +1,78 @@
...
+coefficients** and the **fraction of potential heat units** accumulated. Then, the **optimal leaf area fraction** is calculated as follows:
+
+```math
+  $$ A_{\text{leaf, opt}} = \frac{h}{h+exp(l_{\text{1}}-l_{\text{2}}\times h)}
+$$
+```
+where $A_{\text{leaf, opt}}$ is the `optimal_leaf_area_fraction`, $l_{\text{1}}$ is the first of two shape parameters (`_lai_shapes`), $l_{\text{2}}$ is the second 
+shape parameter, and $h$ is `heat_fraction`, the fraction of potential heat units accumulated for the plant on a given day.
+
+**Canopy height** represents the canopy height on a given day, in meters, and is calculated using the following equation:
```
ew3361zh, line 21: The second canopy height would not need to be bolded in my mind (despite my earlier comment) since it's not a reference to the variable.

ew3361zh (approved):
I think this looks great and is a helpful model for trying to create the wiki for other parts of RuFaS. I think it's helpful that you've done this at the class level. I had a couple of small comments and suggestions but otherwise really nice work and thanks for bringing me in on it!

SC_redesign/Crop_and_Soil/docs/leaf_area_index.md
LaplacetteCM: Totally agree, I tried to do that at the beginning of each paragraph but I should have also done that every time I used them.

SC_redesign/Crop_and_Soil/docs/leaf_area_index.md
LaplacetteCM: Yes, I agree!

PR #298, opened 2023-01-23
Biomass_allocation.md wiki

LaplacetteCM:
This is the proposed wiki for biomass_allocation.py

Notes:
1. According to the comments in the code, the notation of the .py file will be updated to match the other files or to better indicate what the code is doing, so this wiki might need a future update (e.g., photosynthesize, intercept radiation). 
2. The code does not include the total biomass calculation yet but total biomass is used for 'partioned_biomass'. It is included  in the wiki. 
3. I'm making a list of the notation we are using because the idea is to be consistent between wikis. If there's a correction to make, I'll correct the other files as well.


morrowcj:
>     1. According to the comments in the code, the notation of the .py file will be updated to match the other files

The file already has been updated, actually. I think I forgot to remove the #TODO. I'll fix that and push it here.

Great!

>     2. The code does not include the total biomass calculation yet

Is this true? `photosynthesize()` calculates total `biomass` (line 29), by adding the maximum accumulation value (`biomass_growth_max`) to the previous `biomass`.

SC_redesign/Crop_and_Soil/docs/biomass_allocation.md
```diff
@@ -0,0 +1,87 @@
...
+
+**Intercept radiation** represents the amount of daily photosynthetically active 
+radiation intercepted by the leaf area of the crop according to:
+
+```math
+  $$ R_{\text{int}} = 0.5\times R_{\text{inc}}\times (1-exp\times (-k_{\text{l}}\times A_{\text{leaf, i}}))
+       
+$$
+```
+where $R_{\text{int}}$ is `intercept_radiation`, $R_{\text{inc}}$ is`incident_radiation`,
```
morrowcj: The code calls the intercepted radiation `usable_light`, not `intercept_radiation`. Also, `incident_radiation` is not used in the code and I think we can use a simpler term like "total incoming solar radiation" or something similar.

SC_redesign/Crop_and_Soil/docs/biomass_allocation.md
```diff
@@ -0,0 +1,87 @@
...
+on a given day that results from **intercept_radiation** and the crop-specific radiation-use 
+efficiency, which is the amount of dry biomass produced per unit of intercepted solar 
+radiation. It is calculated using the following equation:
+
+```math
+  $$ Biom_{\text{max}} = R_{\text{int}}\times Eff_{\text{rad}}
+       
+$$
+```
+where $Biom_{\text{max}}$ is `max_accumulation`, $R_{\text{int}}$ is `intercept_radiation`,
```
morrowcj: the max accumulation variable is `biomass_growth_max` in the code.

SC_redesign/Crop_and_Soil/docs/biomass_allocation.md
```diff
@@ -0,0 +1,87 @@
...
+and $Eff_{\text{rad}}$ is `efficiency`, the radiation-use efficiency of the crop.
```
morrowcj: the efficiency is called `light_conversion` factor, in the code. I could be easily persuaded to change this to something like `light_use_efficiency`, though.

SC_redesign/Crop_and_Soil/docs/biomass_allocation.md
```diff
@@ -0,0 +1,87 @@
...
+
+To obtain the **daily accumulated biomass**, **max accumulation** is adjusted according to 
+the daily crop growth factor:
+
+```math
+  $$ Biom_{\text{i}} = Biom_{\text{max}}\times \gamma_{\text{reg}}
+       
+$$
+```
+where $Biom_{\text{i}}$ is `daily_accumulated_biomass`, $Biom_{\text{max}}$ is `max_accumulation`,
```
morrowcj:
```suggestion
where $Biom_{\text{i}}$ is daily accumulated `biomass`, $Biom_{\text{max}}$ is the maximum biomass accumulation,
```

morrowcj (changes requested):
This looks pretty good overall. A few variables names do need clarification, though.

SC_redesign/Crop_and_Soil/docs/biomass_allocation.md
```diff
@@ -0,0 +1,87 @@
+### __Biomass allocation__
+
+The `BiomassAllocation`class manages the crop biomass accumulation through the photosynthesis 
+process and its partition between above and below ground organs during the growing season.
+
+The central function, `allocate_biomass()`, returns on a daily basis the updated 
+**photosynthesize** and **partition biomass**.
```
ew3361zh: I don't know if I'd use the word "returns" here since `allocate_biomass()` is not really returning (in the coding sense) anything. I think something along the lines of "The central method, `allocate_biomass()`, calls on the **photsynthesize** and **partition biomass** methods to make daily updates on crop biomass allocation."

I don't know how consistently we're using python method vs python function but I think generally any functions associated with classes or objects are referred to as methods (a good rule of thumb is a method is passed a reference of self and is called with self.method_name). I also don't know how much of a stickler most people are on this point but here's some more info on it: https://www.geeksforgeeks.org/difference-method-function-python/

ew3361zh (approved):
From my pov this looks good. I had one suggested change but it's not a dealbreaker for me. I do think it would be helpful to get Pooya to look at one of these wiki files to make sure they're matching up with what he had in mind (and maybe that's already happened).

SC_redesign/Crop_and_Soil/docs/biomass_allocation.md
morrowcj: I did not know this about functions vs. methods. Thanks for this info. I'll try to internalize it and be consistent.

LaplacetteCM: Thank you Niko, l still need to get more familiar with the terms and this helps a lot. I didn't know about this difference so I'm gonna take a look through the previous wikis to check the way I used them.

SC_redesign/Crop_and_Soil/docs/biomass_allocation.md
LaplacetteCM: Between the 2 options, I like "light_use_efficiency" here. It's closer to RUE and also describes more clearly what RUE is doing. But I'll leave it up to you!

morrowcj: You're right. OK. Go with `light_use_efficiency` and I'll update it in the code and push to this branch.

----
Edit: Done

SC_redesign/Crop_and_Soil/docs/biomass_allocation.md
morrowcj: @ew3361zh, would you consider things decorated with `@staticmethod` to be "functions" or "methods"? My intuition says they are functions (because they don't use a `self` reference and don't require the class to be initialized).

SC_redesign/Crop_and_Soil/docs/biomass_allocation.md
LaplacetteCM: Done

LaplacetteCM:
Thanks for your comments, I've included all the corrections you mentioned. 

I agree with Niko about asking Pooya if we are on the right track with this approach, maybe this is not what he had on mind for the wikis or he might have some suggestions for improvement.

SC_redesign/Crop_and_Soil/docs/biomass_allocation.md
ew3361zh: @morrowcj good question - I think the name would indicate it's more considered a method than a function and static methods are still bound to classes so even though they don't take the self argument and aren't bound the object, they still operate just within the class where they're defined which seems to be a major distinguisher for a method vs a function. https://pynative.com/python-static-method/

morrowcj (approved):
This looks great! You can go ahead and merge it, or if you'd like to get @PooyaHekmati's opinion, you can request him as a reviewer and await his response. 

Great work and thank you!

morrowcj:
I spoke with Pooya, and he doesn't feel the need to look over this. Go ahead and merge when you are ready.

PR #299, opened 2023-01-24
Info map objects fix

ew3361zh:
**What:**
This PR seeks to address an issue in the Output Manager (OM) converting the the info_map and data collected by the OM throughout a simulation run into a JSON object using the json.dumps() method. 

**Why:**
Certain data types including object instances and file paths are not "serializable" by the json library method dumps().

**How:**
These object instances and file paths need to be removed from info_maps and not added as variables.

### DIFF
--- a/RUFAS/classes.py
+++ b/RUFAS/classes.py
@@ -88,8 +88,8 @@ class Config:
         """
 
         info_map = {"class": self.__class__.__name__,
-                    "function": self.__init__.__name__, 
-                    "data": data, 
+                    "function": self.__init__.__name__,
+                    "config_data": data,
                     "weather_file_path": weather_file, }
 
         # gets a start/end date in the format year:julian-day. That way the program
--- a/RUFAS/classes.py
+++ b/RUFAS/classes.py
@@ -88,8 +88,8 @@ class Config:
         """
 
         info_map = {"class": self.__class__.__name__,
-                    "function": self.__init__.__name__,
-                    "data": data,
+                    "function": self.__init__.__name__, 
+                    "data": data, 
                     "weather_file_path": weather_file, }
 
         # gets a start/end date in the format year:julian-day. That way the program
--- a/RUFAS/classes.py
+++ b/RUFAS/classes.py
@@ -333,8 +333,7 @@ class Weather:
 
         info_map = {"class": self.__class__.__name__,
                     "function": self.__init__.__name__,
-                    "weather_file": weather_file,
-                    "config": config, }
+                    "weather_file": weather_file, }
 
         years = config.years
         w_start_year = config.w_start_year
--- a/RUFAS/routines/animal/ration/pyomo_solver.py
+++ b/RUFAS/routines/animal/ration/pyomo_solver.py
@@ -165,7 +165,7 @@ def create_model(feeds_data, req_data, feeds):
                 "function": create_model.__name__,
                 "feeds_data": feeds_data,
                 "req_data": req_data,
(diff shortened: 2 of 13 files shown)


RUFAS/classes.py
```diff
@@ -88,8 +88,8 @@ def __init__(self, data, weather_file):
         """
 
         info_map = {"class": self.__class__.__name__,
-                    "function": self.__init__.__name__,
-                    "data": data,
+                    "function": self.__init__.__name__, 
+                    "data": data,
```
PooyaHekmati: what data type is data?

RUFAS/routines/animal/ration/pyomo_solver.py
```diff
@@ -161,14 +161,8 @@ def create_model(feeds_data, req_data, feeds):
...
 
-    info_map = {"class": "no_caller_class",
-                "function": create_model.__name__,
-                "feeds_data": feeds_data,
-                "req_data": req_data,
-                "feeds": feeds,}
-
     # TODO update "m1" variable to a more descriptive name.
-    om.add_variable("m1", m1, info_map)  
+
```
PooyaHekmati, line 165: are we missing an add_variable call here?

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
```diff
@@ -59,11 +59,7 @@ def __init__(self, weather, time, manure_handler_config: ManureHandlerConfig):
 
         """
         info_map = {"class": self.__class__.__name__,
-                    "function": self.__init__.__name__,
-                    "weather": vars(weather),
-                    "time": vars(time),
-                    "config": vars(manure_handler_config),
-                    }
+                    "function": self.__init__.__name__, }
```
PooyaHekmati, line 62: vars was used here, I'm curious to know if that can solve the object issue; i.e., why did you need to remove vars output? did json.dumps fail to handle it?

PooyaHekmati (commented):
in some cases, vars() were used but they are being removed in this PR. I think json.dumps can handle it; if so, we don't need to delete them.

RUFAS/classes.py
ew3361zh: the docstrings describe this as "dictionary containing information from the json file specified under 'config'". What do you think about "config_data" as a rename?

RUFAS/routines/animal/ration/pyomo_solver.py
ew3361zh: I believe "m1" is an object instance so I removed it from being added to the variables pool.

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
ew3361zh: As the manure module is not really connected with any of the input files currently it's difficult to run a simulation and have it add the info_maps and variables from that particular module (same with the crop and soil redesign).

I did, however, attempt to use the vars() function on some of the objects that were added from the animal module during simulations such as instances of the feed and available_feeds classes and it errored for me (see attached photo).
Screenshot 2023-01-24 at 3 17 59 PM

RUFAS/classes.py
PooyaHekmati: I like it

PooyaHekmati (approved):
[LGTM](https://lgtm.lol/i/438)

JoeWaddell (approved):
Runs in good time on my work Windows machine, and no problems I can see in the changes you've made.

PR #302, opened 2023-01-25
workflow to add labeled tasks to proeject

morrowcj:
## Summary 
This only purpose of this PR is to add a GitHub Action that adds all issues and pull requests tagged with the `SC_redesign` label to the newly created ["Soil and Crop Redesign" Project](https://github.com/orgs/RuminantFarmSystems/projects/1), which tracks all tasks associated with this project. 

I think [GitHub Projects](https://docs.github.com/en/issues/planning-and-tracking-with-projects/learning-about-projects/about-projects) could be a very useful organizational tool for tracking tasks. It can be setup just like our fixathon task sheets but entirely within github. so I'm going to try using it for a while. 

This particular action may not be particularly necessary, since you can associate a project with a PR and issue right below where you add a label,  but I think it will provide a fail-safe if someone forgets to add the project and label. There are other actions associated with Projects, that may be added to this file in the future.

## Complications
One complication is that, for this action to work, it needs access to a personal access token, which can only be generated by users (not organizations). Furthermore only the owner (@PooyaHekmati, I think your Admin status counts) of a repository can associate one of these tokens with the repo (I was unable to generate a token for use with the project). Therefore, someone with owner access to repo needs to [generate a personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token), associate it with `MASM`, and then add the token to the "value" field of the MASM [secret variable](https://github.com/Azure/actions-workflow-samples/blob/master/assets/create-secrets-for-GitHub-workflows.md) `ADD_TO_PROJECT_PAT`

## Additional Changes
* added link to project task sheet to main readme file


morrowcj:
I'm going to keep adding and removing the label to test this functionality. For some reason, it doesn't like the version reference I'm using.

---
Edit: The version reference is now fixed and the action is stalling on the credentials (as expected).

morrowcj:
At Pooya's recommendation, I'm going to test this on a fork. I'll switch it to a draft and will mark it for review once I am finished testing.

morrowcj:
I'm going to close this PR. After further exploration, this seems unnecessary.

PR #238, opened 2022-12-14
Draft of field management structure

morrowcj:

Issue #248, opened 2022-12-22
Implement cut, kill, and harvest Crop methods (SC_redesign)


morrowcj:
before the refactor, `kill` and `cut` were implemented in `yields.py`. They are better suited for `Crop` class, but need reworked implementation. 

This issue is to serve as a reminder to move the functionality of these two methods (and an additional `harvest()` method) into `Crop` during the redesign. 

This will likely be fixed in PR #238

morrowcj:
This was implemented in #346

Issue #246, opened 2022-12-22
C&S should handle harvest index override (SC_redesign)


morrowcj:
the file `SC_redesign/Crop_and_Soil/crop/yields.py` accepts an attribute `user_harvest_index`, which, when provided, triggers a harvest-index override and the user-defined value is used in place of the calculated value. 

This issue will serve as a reminder to add that option to the crop and soil input. The logical place to handle this is the as-yet non-existent `FieldManager` class.

morrowcj:
This was implemented in #346

PR #305, opened 2023-01-25
Field Management Design (part 1)

morrowcj:
## Summary

This PR is the first of $N$ that will determine how individual `Field`s are managed within Soil and Crop (SC). This initial PR is meant to establish a design foundation for the relevant classes, how they are interconnected, and general ideas of what they should do.

The design presented herein is simply a draft. I want this review to help push it to where it needs to be. I fully expect this to be an iterative process.

## Details

* Generally speaking, this PR is not meant to implement specific methods, rather it treats methods as conceptual for the time being. As such, methods remain largely unimplemented (and untested).
* This PR centers around the `Field` class (field.py) and its `FieldData` component (field_data.py). 
* The following configuration classes are new and will be utilized by `Field` methods. These classes are bare bones placeholders for now.
  - `CropPlantingConfig` contains specifications for how to plant crops in the field
  - `CurrentWeather` gathers weather attributes for the current day
*  The `Soil`, `LayerData`, `Crop`, and `CropData` classes have been slightly altered to include attributes that `Field` method needs.

## Specific requests for Reviewers

* As a design PR, I'd like reviewers to focus primarily on structure of the altered files. Does this design work? If not, how would you change it? I want any and all feedback at this stage.

* While most methods remain unimplemented, I've tried to write docstrings such that they can be used to guide programmers towards the desired functionality. Please make sure that these are clear enough towards that end.

## Other Notes

* This is a continuation of the erroneously closed #238 
* This branch should (eventually) try to address #246 and #248
* This version of `Field` and `Soil` likely conflict with versions that @matthew7838 and @Isgotkowitz are working on in other Issues/PRs.


SC_redesign/Crop_and_Soil/crop/crop.py
```diff
@@ -14,6 +14,7 @@
 
 # TODO: Should use an ENUM class to represent the supported species??
```
Isgotkowitz: That sounds like a good idea to me. Would it be possible/a good idea to write the enum inside `CropData` and then import it to this module?

SC_redesign/Crop_and_Soil/crop/crop.py
```diff
@@ -56,26 +57,25 @@ def __init__(self, crop_data: Optional[CropData] = None):
...
                   incoming_light: float,
                   evaporation: float, transpiration: float,
                   max_evapotranspiration: float,
                   mean_air_temperature: float, min_air_temperature: float,
-                  max_air_temperature: float) -> None:
+                  max_air_temperature: float, adjusted_potential_evapotranspiration: float) -> None:
         """main function for growing the crop on a daily basis
 
         Args:
+            adjusted_potential_evapotranspiration:
```
Isgotkowitz:
```suggestion
            adjusted_potential_evapotranspiration: potential evapotranspiration adjusted for evaporation of free water in canopy in mm
```
This is how it is currently defined in `soil\evapotranspiration.py`.

SC_redesign/Crop_and_Soil/crop/crop_data.py
```diff
@@ -32,6 +32,10 @@ class CropData:
     """is the plant perennial?"""
     is_nitrogen_fixer: bool = False
     """is the plant a nitrogen fixer?"""
+    priority: int = 1
```
Isgotkowitz, line 35: Is this inherent to the species? Like, if corn and wheat are planted in the same field together, corn always sucks up water from the soil faster than wheat and therefore have a higher priority? Or would it be user defined?

Also, if it is possible for one crop to have a higher priority than another for resource A, but a lower priority for resource B, then maybe this should be a list of priorities

SC_redesign/Crop_and_Soil/field/crop_planting_config.py
```diff
@@ -0,0 +1,21 @@
...
+class CropPlantingConfig:
+    """Configuration class that specifies how crops in a field should be initialized by the Field.plant_crops()
+    method
+
+    Details: this class should contain any and all attributes needed to setup and create a series of Crop objects,
+    including their respective CropData objects.
+    """
+    species: Optional[List[CropSpecies]] = None
+    """all the species to be planted in the field"""
+    planting_day: int = 0  # perhaps a list option to specify each crop being planted separately in the field?
```
Isgotkowitz, line 17: Yes

SC_redesign/Crop_and_Soil/field/crop_planting_config.py
```diff
@@ -0,0 +1,21 @@
...
+    """the (sequential) day of the simulation on which the crops should be planted in their field"""
+    planting_year: int = 0  # perhaps a list option to specify each crop being planted separately in the field?
```
Isgotkowitz, line 19: Also yes. Maybe these should be a tuple (i.e. `(day, year)`), and should be stored as an attribute within the `CropSpecies` object it is associated with? 

And does the year even need to be recorded? If the sequential day it was planted in the simulation is kept, then wouldn't it be trivial to calculate which year it was planted in (for example, if it was planted on day 700, then it was planted in the second year)?

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -0,0 +1,228 @@
...
+manager classes. The field class, as laid out here, handles the management actions and scenarios that can be performed
+in an agricultural field. 
+
+Note that some of the field-level attributes will be tracked by the FieldData class
+"""
+
+
+class Field:
+    """object representing an agricultural field"""
+    def __init__(self, field_data: Optional[FieldData] = None, soil: Optional[Soil] = None):
```
Isgotkowitz, line 20: Should this have an optional parameter for a list of crops?

Isgotkowitz (changes requested):
This is a really good start to the Field module. I have two main issues (at this stage) that I would like to discuss:
1. How are we going to effectively manage a field that has more than one crop planted in it? Having multiple crops in a single field seems (to me) like it will be prone to inaccuracies, assuming that the field is partitioned into multiple subsections, each containing a single crop. This is because different crops can have significantly different characteristics, and when applying the simulation to a specific crop it seems like it would be difficult (if not impossible) to restrict the effects of those characteristics to that crop's section of the field and not have the effects be averaged with all the other crops' effects, then applied to the entire field. 

    For example, water uptake. I believe that currently, SWAT (and therefore, RuFaS) can only treat water in the soil as uniformly distributed across the entire field. If a field is equally partitioned into two sections, one with only crop A and one with only crop B, and crop A uptakes significantly more water than crop B, then at the end of the day RuFaS will think that there are x millimeters of water uniformly distributed across the entire soil profile of the field. But in reality, the soil water content in crop A's section is << x, and in crop B's section it is >> x, causing crop A to wilt and crop B to have strong growth, leading to inaccurate simulation results. 

2. Is it possible/helpful to define a heuristic for deciding the order in which processes occur? For example, we try to adhere as closely as possible to the following order:
    1. weather processes
    2. farmer-controlled processes (tilling, fertilizing, cutting, killing, ...)
    3. animal processes (grazing, manuring, ...)
    4. plant processes (growing, uptaking water/nutrients, ...)
    5. soil processes (erosion, water infiltration/percolation, ...)

     when we are writing the `Field` method that will actually execute all these processes (because they have to happen sequentially).

SC_redesign/Crop_and_Soil/crop/crop.py
morrowcj: This is actually an old TODO that I missed. This feature was already implemented in #314

SC_redesign/Crop_and_Soil/crop/crop_data.py
morrowcj: Great question. I think it is somewhat inherent to the species. Some plants are, for example, better at acquiring water than others, so on any given day they would have "priority" in accessing that resource. In my mind it isn't an entirely fixed value, though and would depend upon the composition of the multi-crop system (more on that later).

Yes, I think you're right conceptually that there might be resource-specific priorities, but this tremendously more complicated to implement in the code: Since each crop conducts their entire grow routine one after another, it is simply to re-arrange the order in which that occurs. It is much more difficult to re-arrange specific components separately. Not impossible, but it would require some restructuring - definitely not something needed for the MVP.

SC_redesign/Crop_and_Soil/field/crop_planting_config.py
morrowcj: I'm not entirely sure that is needed yet. I don't know of common real-world scenarios in which different crops would be planted at different times in the same field. The use case that multi-cropping is intended to address is farmers planting a mix of seeds in their fields (random/uniformly distributed) with no intention of harvesting them (i.e., cover cropping).

SC_redesign/Crop_and_Soil/field/crop_planting_config.py
morrowcj: Not a terrible idea, but similar to my comment above: I think the most common use case would be that the entire field is planted at once and harvested at once (with machinery). At least in the economic systems we're trying to model. I definitely think this is worth continued discussion and thought, though!

Technically multi-cropping isn't a required feature of v1/MVP, but is something we need to plan for.

SC_redesign/Crop_and_Soil/field/field.py
morrowcj: Yes, I suppose it should. Maybe it should be a `@dataclass` to simplify syntax.

SC_redesign/Crop_and_Soil/crop/crop_data.py
ltp23: If this turns out to be more complicated, it could be modeled as a graph problem to resolve the relative ordering.

morrowcj: I'm not sure what is meant by a graph problem. Could you provide a reference, @ltp23?

morrowcj:
Thanks for the great feedback, @Isgotkowitz. It was really helpful. In addition to the direct responses to your smaller comments, I wanted to reply to these more general points in a few separate comments.

> 1. How are we going to effectively manage a field that has more than one crop planted in it?

Great question, I should have explained this. One of the desired features of the model (e.g., farmer shareholders) is the ability to plant assemblages of cover crops, which are planted to protect and restore soil and to allow a field to "rest". Cover crops are commonly grasses and one strategy is to spread a mix of grass and other prairie/rangeland plants. 

The most common scenario, by far, will be to only plant 1 crop in a field, but we need the option (in the future) to handle more than one. But the management activities will be (as I see it) at the field level: till the field, plant the field, harvest the field, etc.

One crop's methods are independent of an other's and only affect each other indirectly by changing the resources available. On a given day, The first crop in the field (e.g., highest priority) will conduct its growth method, uptaking the resources it needs. Then, the second crop does the same (with access to the reduced resource pool), and so on. The main (but relatively small) problem here would be that, on a day that the resource pools are reduced below the cumulative demands, those with highest priorities will meet their needs before the others. And those with lowest priorities might not have *any* needs met for the day. A solution that I've considered to this is dividing the resource pool in to fractions every day (higher priorities get larger fractions) and allowing the plants to draw from that proportion only. Either method would be sufficient in my mind.  A model isn't expected to replicate natural processes exactly and estimating growth daily is a very fine-grained resolution which means that some amount of that error will be absorbed.

morrowcj:
> For example, water uptake. I believe that currently, SWAT (and therefore, RuFaS) can only treat water in the soil as 
uniformly distributed across the entire field. If a field is equally partitioned into two sections, one with only crop A and one with only crop B, and crop A uptakes significantly more water than crop B, then at the end of the day RuFaS will think that there are x millimeters of water uniformly distributed across the entire soil profile of the field. But in reality, the soil water content in crop A's section is << x, and in crop B's section it is >> x, causing crop A to wilt and crop B to have strong growth, leading to inaccurate simulation results.

As mentioned above, I think the more common situation would be a uniform mixture of plants. However, I have though about partitioning a field (left half corn, right half alfalfa for example). A way to solve this would simply be to, again, divide the resource pools into fractions based on the partition sizes. However, in this scenario, those pools would be independent from each other. But at that point, just treat each partition as its own field, because there really isn't a difference.

Again, though these are just estimations of processes. In real life, individuals of a single species don't uptake nutrients at the same rate and there can be substantial spatial variation. For that matter the available nutrients and soil attributes aren't uniform across the field either. But the central limit theorem dictates that, with large enough samples (field area), averages are good estimators.

SC_redesign/Crop_and_Soil/field/field.py
ltp23: I think it's better to have parameters like these required, instead of being optional, and in some other similar places also. This allows the implementer of this class to not have to know how to initialize those objects and leaves that job to the user of the class instead.

morrowcj:
>  2. Is it possible/helpful to define a heuristic for deciding the order in which processes occur? For example, we try to adhere as closely as possible to the following order:        
>        1. weather processes
>        2. farmer-controlled processes (tilling, fertilizing, cutting, killing, ...)
>        3. animal processes (grazing, manuring, ...)
>        4. plant processes (growing, uptaking water/nutrients, ...)
>        5. soil processes (erosion, water infiltration/percolation, ...)
>        
>        when we are writing the `Field` method that will actually execute all these processes (because they have to happen sequentially).

I'm not entirely sure what you mean by this but if I'm understanding correctly, yes, it is important that we generalize the order in which processes occur. I tried to do that with my heading blocks within `manage_field()`. Was that not clear/sufficient, or do you think there's a better way?

SC_redesign/Crop_and_Soil/soil/soil.py
```diff
@@ -10,9 +9,9 @@
 
 class Soil:
     def __init__(self, soil_data: Optional[SoilData] = None):
-        data = soil_data or SoilData()
-        self.evapotranspiration = Evapotranspiration(data)
-        self.infiltration = Infiltration(data)
+        self.data = soil_data or SoilData()
+        self.evapotranspiration = Evapotranspiration(self.data)
+        self.infiltration = Infiltration(self.data)
```
ltp23, line 16: If possible, it is better to only pass in just enough data to make each class functional.

ltp23 (approved):

SC_redesign/Crop_and_Soil/soil/soil.py
morrowcj: @ltp23, this follows the composite pattern that we used for the `Crop` class. All the component classes utilize the same data component class reference, since they utilize and update the same attributes within the data component. Is there a reason this doesn't fit in the `Field` class when it does fit the `Crop` class?

SC_redesign/Crop_and_Soil/field/field.py
morrowcj: Really? I like the flexibility that is allowed by giving the option to input a pre-configured component object while also setting up a default otherwise.

SC_redesign/Crop_and_Soil/field/crop_planting_config.py
Isgotkowitz: In that case, I don't think it would be necessary then. I was only thinking it would be necessary when a crop is divided into sections and each section has a different crop.

SC_redesign/Crop_and_Soil/field/crop_planting_config.py
Isgotkowitz: Ok, I think as long as the mix of crops is uniformly distributed across the field then it shouldn't be an issue to just keep a single parameter.

morrowcj: Your point about storing the date as a tuple is good. And also your point about my docs is valid. `day` should be "the sequential day within the current year of the simulation."

Isgotkowitz: Also multi-cropping seems like a doable feature to include in v1, as long as it is only multi-cropping in the way you describe above.

morrowcj: I agree. We should include it in this form and we can tweak it later if needed.

Isgotkowitz:
> As mentioned above, I think the more common situation would be a uniform mixture of plants. However, I have though about partitioning a field (left half corn, right half alfalfa for example). A way to solve this would simply be to, again, divide the resource pools into fractions based on the partition sizes. However, in this scenario, those pools would be independent from each other. But at that point, just treat each partition as its own field, because there really isn't a difference.

Cool, this would only be a problem for a field divided up into monocrop sections with different species in each section.

Isgotkowitz:
> Great question, I should have explained this. ...

Now that I understand what multi-cropping means in this context, my main worry about that has been alleviated. However you raised a really good concern about higher priority crops using up all available resources before lower priority crops get a chance to access them. Your idea about dividing the resource pool into fractions sounds like a really solution, though it also sounds like it would introduce significant complexity into managing the resources (so maybe it should wait until v2).

Isgotkowitz:
> > 2. Is it possible/helpful to define a heuristic for deciding the order in which processes occur? For example, we try to adhere as closely as possible to the following order:
> >    
> >    1. weather processes
> >    2. farmer-controlled processes (tilling, fertilizing, cutting, killing, ...)
> >    3. animal processes (grazing, manuring, ...)
> >    4. plant processes (growing, uptaking water/nutrients, ...)
> >    5. soil processes (erosion, water infiltration/percolation, ...)
> >    
> >    when we are writing the `Field` method that will actually execute all these processes (because they have to happen sequentially).
> 
> I'm not entirely sure what you mean by this but if I'm understanding correctly, yes, it is important that we generalize the order in which processes occur. I tried to do that with my heading blocks within `manage_field()`. Was that not clear/sufficient, or do you think there's a better way?

My bad, I did not look closely enough at the heading blocks, they are a very good start on the ordering. Though since we have the ability to break up module processes and mix them around, like how we are planning to break off water processes from different modules and recombine them, we could review all the other modules and see if there are any more processes like that.

Isgotkowitz (approved):

morrowcj:
I'm going to merge this branch and leave any additions to another PR.

PR #307, opened 2023-01-26
Evapotranspiration Refactor

### LINKED ISSUE #303: Refactor/Implement transpiration
Implement transpiration and water uptake into the crop `water_dynamics.py` file. This will need to work in concert with Soil, so it may wait until the Soil class is further along.

### LINKED ISSUE #306: refactor and test evapotranspiration.py (soil) into SC_redesign directory

### LINKED ISSUE #313: Potential evapotranspiration adjusted for water use needs to be implemented inside of Evapotranspiration module
In the `Evapotranspiration.py`, the function `_determine_soil_evaporation()` takes the adjusted potential evapotranspiration as a parameter even though it can/should be calculated within the Evapotranspiration module.

The equations for calculating the adjusted potential evapotranspiration can be found in SWAT 2:2.3.1. These equations use the amount of free water held in the canopy as a parameter, and the calculations for this parameter can be found in SWAT 2:2.1.

Isgotkowitz:
# Summary
This PR refactors the `Evapotranspiration` module within the `Soil` module based on SWAT 2:2.3. All but three methods are `@staticmethods`, which eliminates the need to pass objects into methods as was done in the old code. `evapotranspirate()` is the main method within `Evapotranspiration` that handles all evapotranspiration operations from the soil. All methods are tested in the `test_evapotranspiration.py` file.

It also introduces the `SoilData` and `LayerData` classes. `SoilData` is the soil equivalent to `CropData`; it will store data that is generated and tracked by modules within `Soil`. Each instance of the `LayerData` class represents a single layer of soil within a soil profile, and a `Soil` object maintains a list of `LayerData` objects to manage the soil profile.

# Notes
* In the old code, `Evapotranspiration` included an operation to determine the maximum transpiration based on attributes of `CropData`. This method has been moved to the `WaterDynamics` module within the `Crop` module, and is now tested within the `test_water_dynamics.py` file.
* A separate folder has been added for `Soil` module tests, which includes an empty `init` file. At some future point all tests for `Crop` module should be moved into their own module.


SC_redesign/Crop_and_Soil/soil/evapotranspiration.py
```diff
@@ -0,0 +1,264 @@
...
+        Details: calculates and stores the potential evapotranspiration, soil evaporation in the SoilData object
+        """
+        self.data.potential_evapotranspiration = self._determine_potential_evapotranspiration(
+            extraterrestrial_radiation,
+            max_air_temp,
+            min_air_temp,
+            avg_air_temp)
+        self.data.potential_evapotranspiration_adjusted = self._determine_potential_evapotranspiration_adjusted(
+            initial_canopy_free_water)
+        # TODO: add field (in CropData?) to track amount of free water in canopy as it gets adjusted - issue #316
```
morrowcj:
```suggestion
        # TODO: add attribute (in CropData?) to track amount of free water in canopy as it gets adjusted - issue #316
```

Let's avoid using the term "field" to refer to anything other than an agricultural field, if possible.

SC_redesign/Crop_and_Soil/soil/evapotranspiration.py
```diff
@@ -0,0 +1,264 @@
...
+        """calculates the evaporative demand for a given layer of soil
+
+        Args:
+            max_soil_water_evaporation: maximum water evaporation from soil on given day in mm
+            layer_data: LayerData object of the soil layer to be analyzed
+
+        Returns:
+            evaporative demand for given layer of soil in mm
+
+        SWAT Reference: 2:2.3.17
```
morrowcj: I'm having a hard time linking this SWAT equation to this code.

SC_redesign/Crop_and_Soil/soil/evapotranspiration.py
```diff
@@ -0,0 +1,264 @@
...
+        if layer_data.top_depth is None or \
+                layer_data.top_depth < 0 or \
+                layer_data.bottom_depth is None or \
+                layer_data.bottom_depth < 0 or \
+                (layer_data.bottom_depth <= layer_data.top_depth):
+            raise ValueError("Missing or illegal values for top or bottom depths")
+
+        # Calculate top evaporative demand
+        top_depth = layer_data.top_depth
+        top_quotient = top_depth / (top_depth + exp(2.374 - (0.00713 * top_depth)))
```
morrowcj: Where does this term (`top_quotient`) come from??

SC_redesign/Crop_and_Soil/soil/evapotranspiration.py
```diff
@@ -0,0 +1,264 @@
...
+        SWAT Reference: 2:2.3.3.1
+        """
+        if soil_evaporation_adj < snow_water_content:
+            return 0  # 2:2.3.10
+        else:
+            return soil_evaporation_adj - snow_water_content  # 2:2.3.15
+
+    @staticmethod
+    def _determine_evaporative_demand(max_soil_water_evaporation: float, layer_data: LayerData) -> float:
+        """calculates the evaporative demand for a given layer of soil
```
morrowcj: Let's indicate in the name that this is a layer function. Also, how do you feel about removing the dependency on `LayerData`? My preference is to always do so (in `@staticmethod`s) unless it would require an unreasonable number of parameters

```suggestion
    def _determine_layer_evaporative_demand(max_soil_water_evaporation: float, top_depth: float, bottom_depth: float, compensation: float) -> float:
        """calculates the evaporative demand for a given layer of soil
```

note that the acronym "esco" should be replaced with a better name. As a parameter, I'm good with simply `compensation` but the corresponding `LayerData` attribute should be something like `evapotranspiration_compensation_coefficient`

SC_redesign/Crop_and_Soil/soil/evapotranspiration.py
```diff
@@ -0,0 +1,264 @@
...
+
+        # Calculate top evaporative demand
+        top_depth = layer_data.top_depth
+        top_quotient = top_depth / (top_depth + exp(2.374 - (0.00713 * top_depth)))
+        top_evaporative_demand = max_soil_water_evaporation * top_quotient
+
+        # Calculate bottom evaporative demand
+        bottom_depth = layer_data.bottom_depth
+        bottom_quotient = bottom_depth / (bottom_depth + exp(2.374 - (0.00713 * bottom_depth)))
+        bottom_evaporative_demand = max_soil_water_evaporation * bottom_quotient
```
morrowcj: This violates "don't repeat yourself" because it is doing exactly the same thing as lines 214-217. It seems to me that a cleaner solution would be to change this function to something like:

```python
def _determine_depth_evaporative_demand(depth: float, max_evaporation: float) -> float:
    """calculate the evaporative demand at a particular soil depth"""
    pass

def _determine_layer_evaporative_demands(depths: List[float], compensations: List[float], max_evaporation: float) -> List[float]:
    """calculate the evaporative demand for each soil layer
    
    Args: 
        depths: the deepest depths for each soil layer (mm)
        compensations:  the compensation factors for each layer (mm)
        max_evaporation: the maximum water evaporation for the soil (layers?, if so should be a List) during the current day.
    """

    # add in the depth of the soil surface (depth of 0)
    new_depths = [0] + depths
   
    # get depth demands
    depth_demands = [Evapotranspiration._determine_depth_evaporative_demand(d) for d in new_depths]

    # calculate and return the layer demands for each layer
    return [lower - upper*coef for lower, upper, coef in zip(depth_demands[1:], depth_demands, compensations)]
```

It doesn't necessarily need to be split into two functions, but its how I think of this type of problem. What are your thoughts on this? It is how I've handled layers previously (i.e., in `nitrogen_incorporation.py`). But, as always, if you think there's a good reason not to do this, feel free to say so.

morrowcj (changes requested):
# Summary
Ed, I think that, in general, this PR is quite good: great work. My main comments involve `_determine_evaporative_demand()`, which I'd like you to address before moving forward. The comments on this method also apply generally to `determine_evaporative_demand_reduced()` and `_determine_amount_water_removed()`. Feel free to push back, if desired. I am hopeful that once we go through a few back-and-forths on this file, it will make the other soil refactors a bit easier. 

# Other Comments
* Try to be much more descriptive in the main comment of the PR. It is OK to open a draft with a placeholder, 
but once you're ready for review, it is very important to give reviewers information about what they're looking
at. What was the goal of the PR? What are the primary files affected? Are there any special considerations or 
context that should be available? etc. 
* Add a top-level docstring describing where in SWAT the functions in `evapotranspiraiton.py` and `water_dynamics.py` come
come from (see lines 5-8 in `crop/biomass_allocation.py`).

SC_redesign/Crop_and_Soil/soil/evapotranspiration.py
Isgotkowitz: This is function is actually a combination of 2:2.3.16 (the first one, there is a duplicate) and 2:2.3.17. The 16 is to calculate the evaporative demand at a certain depth, and 17 is to calculate evaporative demand for a layer based on the its top and bottom depths

SC_redesign/Crop_and_Soil/soil/evapotranspiration.py
Isgotkowitz: Equation 2:2.3.16, I'll add that into the comments

morrowcj: My bad, I don't know how I missed it.

SC_redesign/Crop_and_Soil/soil/evapotranspiration.py
Isgotkowitz: Removing the dependency is no problem here, I'll take care of it

SC_redesign/Crop_and_Soil/soil/evapotranspiration.py
Isgotkowitz: First, I think it would be a good idea to split it into two function cause I am definitely repeating myself.

Isgotkowitz: Second, I think it is better to do one layer at a time, and have a call for each layer in `evaporate_from_soil`. In this case either method works alright, but I find it more straightforward to work on one object a time. Also, as soon as multiple objects get separated into corresponding lists of attributes, the danger arises that somehow those lists will get/be used out of sync.

morrowcj (approved):
Looks great! Thanks for your patience.

SC_redesign/Crop_and_Soil/soil/evapotranspiration.py
```diff
@@ -0,0 +1,293 @@
...
+    def __init__(self, soil_data: Optional[SoilData] = None):
+        self.data = soil_data or SoilData()  # initialize with defaults, if not given
+
+    # --- main routine ---
+    def evapotranspirate(self, extraterrestrial_radiation: float, max_air_temp: float, min_air_temp: float,
+                         avg_air_temp: float, above_ground_biomass: float, residue: float, snow_water_content: float,
+                         initial_canopy_free_water: float) -> None:
+        """does the evapotranspiration of the soil on a given day
+
+        Details: calculates and stores the potential evapotranspiration, soil evaporation in the SoilData object
```
morrowcj: I think we need a better verb than "does" - maybe "executes evapotranspiration processes"?

Also you're missing docstrings for the arguments.

SC_redesign/Crop_and_Soil/soil/evapotranspiration.py
```diff
@@ -0,0 +1,293 @@
...
+            """
+            return 0  # 2:2.3.1
+        else:
+            """
+            Evaporation from free water in canopy on given day is set equal to initial free water in the canopy 
+            (2:2.3.3), and the potential evapotranspiration adjusted for evaporation of free water in copy is equal to 
+            difference between potential evapotranspiration and evaporation from free water in canopy 
+            (E'_0 = E_0 - E_CAN)
+            """
+            return self.data.potential_evapotranspiration - initial_canopy_free_water
```
morrowcj, line 82: I like these contextual additions a lot.

morrowcj (commented):
I noticed a couple things in `evapotranspiration.py` while I was reviewing #319 that I missed previously.

SC_redesign/Crop_and_Soil/soil/evapotranspiration.py
Isgotkowitz: Thanks

SC_redesign/Crop_and_Soil/soil/evapotranspiration.py
Isgotkowitz: I have added a docstring and changed the description to use your suggested verb, I think it should be good now

SC_redesign/Crop_and_Soil/soil/evapotranspiration.py
```diff
@@ -125,7 +138,7 @@ def _determine_latent_heat_vaporization(avg_air_temp: float) -> float:
             avg_air_temp: average air temperature in degrees C
 
         Returns:
-            latent heat of vaporization in MJ kg^(-1)
+            latent heat of vaporization in MJ per kg
```
morrowcj, line 142: Good thinking to change this. Superscript notation in plain text is a pain. Hooray for parsimony!

SC_redesign/Crop_and_Soil/crop/water_dynamics.py
```diff
@@ -2,20 +2,47 @@
...
+
 # TODO: This module needs to be updated to include water uptake by plants and evapotranspiration (implemented in Field?)
 
 class WaterDynamics:
     def __init__(self, crop_data: Optional[CropData] = None):
         self.data = crop_data or CropData()  # initialize with defaults, if not given
 
-    def cycle_water(self, evaporation: float, transpiration: float, max_evapotranspiration: float) -> None:
+    def cycle_water(self, evaporation: float, transpiration: float, max_evapotranspiration: float,
+                    potential_evapotranspiration_adjusted: float) -> None:
```
ew3361zh, line 13: Add doc strings for this method

ew3361zh (approved):
Really nice work. I'll leave the science part to Clay's review but the coding practices look sound. There was one place I didn't see docstrings (`def cycle_water` in the `WaterDynamics` class) but otherwise really consistent, well-documented doc strings, generally good typing practices, code is well broken up using single-responsibility principle guidelines. Testing from "old" testing modules were failing (see attached photo) but discussed with Ed and he assured me this was not a concern at this juncture as those will be obsolete.
Screenshot 2023-02-02 at 3 20 02 PM

morrowcj:
> Testing from "old" testing modules were failing (see attached photo) but discussed with Ed and he assured me this was not a concern at this juncture as those will be obsolete. 

To clarify, @ew3361zh, as Ed mentioned these tests are obsolete `pytest SC_redesign` is what we're most interested in. That being said, I do plan to fix those old tests, since I broke them by merging Reid's code into the `SC_Redesign` branch but did not update tests.

SC_redesign/Crop_and_Soil/soil/soil.py
```diff
@@ -14,9 +22,10 @@ def __init__(self):
...
-        # profile layer attributes
-        self.depths = [5, 8, 20]  # arbitrary
-        self.nitrates = [0.5, 1, 5]  # arbitrary
+        self.data.soil_layers = [LayerData(top_depth=0, bottom_depth=5, nitrate=0.5),
+                                 LayerData(top_depth=5, bottom_depth=8, nitrate=1),
+                                 LayerData(top_depth=8, bottom_depth=20, nitrate=5)]
+
 
     @classmethod
     def make_from_config(cls, soil_config) -> Soil:
```
matthew7838, line 30: missing docstring

SC_redesign/Crop_and_Soil/crop/water_dynamics.py
Isgotkowitz: Docstring is now present, but Clay and I will need to do some `TODO`'s before it can be finished

matthew7838 (approved):
Most styling and comments look good, all I can find are some missing docstrings. One thing I noticed is that there are no docstrings in the tests. I was told to add docstrings while writing my tests so I'm not sure if the same principle would apply to this PR.

SC_redesign/Crop_and_Soil/soil/soil.py
Isgotkowitz: This function is not actually implemented yet, I think it will be ok to wait until it is implemented to write the docstring. Nice catch though

PR #308, opened 2023-01-26
Info map objects fix pt2

ew3361zh:
**What**
This PR seeks to address an issue that arose in [PR299](https://github.com/RuminantFarmSystems/MASM/pull/299) around converting objects that were added to Output Manager (OM) interfering to be serialized by the json.dumps() method (i.e. converting python object instances into json readable data). 

**Why**
Object instances were removed to ensure the proper function of the json.dumps() method on data collected by OM but it was later confirmed that using the built in vars() method successfully converted objects to json.dumps() serializable formats so this PR readds those removed objects back to the OM.

**How**
PR 299 was used for comparison and each individual object was tested to ensure function with the json.dumps() method usage in OM. There were certain data storage objects that contained objects within them (e.g. the "herd_data" dictionary in the _print_animal_num_warnings method of the animal_management.py method) which made them unable to be added back.

Note:
- There are some additional formating changes fixed that are small and hopefully not too distracting along with some additional data added to OM in the feed.py module.
- I also added some typing to one of the methods whose docstrings incorrectly labeled an arg type which led to some of the decision-making in PR299.


JoeWaddell (approved):
This looks good to me: with the caveat that we'll have to decide on which daily updates are added to the variable pool to avoid inflating output file sizes in longer simulations.

ancheypeng (approved):
Looks good! Is there anything that can be done later on about variables like the herd_data dictionary to allow them to be added to the logs?

ew3361zh:
> Looks good! Is there anything that can be done later on about variables like the herd_data dictionary to allow them to be added to the logs?

Do you know if there are specific pieces of data that are needed in the logs/variables pools? 

I imagine we could either individually take out the pieces of the herd_data dictionary we needed for reporting/sharing or if there are multiple pieces of data needed we could look at remaking the herd_data dictionary so that it's not passing around object instances (or if we need the object instances, removing them and using vars() on them to be added to the appropriate OM pool)

PR #314, opened 2023-01-27
Crop Species Data

### LINKED ISSUE #180: Integrate BaseCrop and Crop
Recently, I restructured the `BaseCrop` class in order to cut down on duplicate code and redundancy. However, I left some original features in tact to preserve the program's current usage. Some of these aspects need to be changed in future versions: 

* [ ] Currently, data for each crop only contains a name parameter (`crop_name`), which takes the form "cropXX" where "XX" is the last two digits of the year in which the first instance of the crop was planted. `BaseCrop` (and `Crop`) determine the species of the crop using its name (see `_set_species()`). Instead, the species should be given explicitly. 
* [ ] The `Crop` class currently assigns an internal object to a 'BaseCrop' (child) instance using a switch. `Crop` and `BaseCrop` should be better integrated.

### LINKED ISSUE #231: Validate starting crop-specific parameters with SWAT
SWAT distributes with a database of species-specific plant growth parameters. Since we primarily use SWAT methods and almost all the same inputs, it is important that we validate that our default starting parameters are the same as those distributed with SWAT. Reference: https://swat.tamu.edu/media/69341/ch14_input_plantdb.pdf

morrowcj:
## Summary
This PR Incorporates the species data from the SWAT database and allows for different crops to be used interchangeably in RUFAS. The PR is misleadingly large and should be very straightforward to review (see details)

Thank you for your review!

## Changes 
* `CropSpecies` is an `Enum` of all the different crop species handled by RUFAS (`species_data_factory.py`)
* a new crop species class exists for each of the supported `CropSpecies` (`crop_data.py`)
* all crop species have attribute values taken from the [SWAT database](https://swat.tamu.edu/media/69419/Appendix-A.pdf), and have been fully validated (with the exception of `Triticale`).
* `CropSpeciesDataFactory` builds the crop data for the desired species. (`species_data_factory.py`)

## Details
* All changes to the `CropData` class are simply re-arranges or updating default variable values
* All other changes to `crop_data.py` setup the species default attribute values
* `species_data_factory.py` contains the main updates in need of review
* The tests should also serve as a demonstration of the functionality. In particular:
  - `test_species_factory_defaults()` is very long and simply tests that all attributes are initialized to their appropriate values for each data class.
  - `test_manual_custom_crop_data()` demonstrates how to manually create custom crops by directly altering the values of the data class.
  - `test_factory_alterations()` demonstrates the factory's utility at extending this customization


SC_redesign/Crop_and_Soil/crop/crop_data.py
```diff
@@ -2,15 +2,103 @@
...
+    Details:
+      The kw_only=True argument of the @dataclass decorator specifies that this class' attributes can only be
+      initialized with values other than defaults by explicitly including the keyword (i.e, positional arguments are
+      disabled).
+      For example CropData() will initialize with the default light_extinction value (0.65),
+      CropData(light_extinction=0.7) will initialize with the vallue set to 0.7, but CropData(0.65) will not work.
+      The upside is that this facilitates dataclass inheritance. For example, the CornData class inherits from CropData
+      but will have its values set to different defaults.
+
+      The slots=True argument sets up the data class so that memory is pre-allocated for the instance. This means that
```
Isgotkowitz: Where is the `slots=True` argument?

SC_redesign/Crop_and_Soil/crop/crop_data.py
```diff
@@ -73,6 +157,7 @@ class CropData:
     """heat units accumulated on the current day; degrees C above minimum growth temperature (Celsius*)"""
     heat_fraction: Optional[float] = None
     """fraction of potential heat units accumulated to date (unitless)"""
+    """fraction of potential heat units accumulated to date (unitless)"""
```
Isgotkowitz:
```suggestion

```

SC_redesign/Crop_and_Soil/crop/crop_data.py
```diff
@@ -204,20 +266,11 @@ class CropData:
...
-    """expected species-specific optimal harvest index for the plant at maturity under ideal
-     growth conditions (unitless)"""
-    min_harvest_index: float = 0.2
-    """expected species-specific harvest index for the plant under drought conditions; represents minimum harvest index
-    allowed for the plant (unitless)"""
+
     # is_residue_added: bool = False ## not needed?
     harvest_efficiency: float = 1.0
     """efficiency of the harvest operation: the proportion of yield that will be extracted from the field 
     (unitless; [0, 1])"""
```
Isgotkowitz, line 268:
```suggestion
    (unitless; must be greater than or equal to 0 and less than or equal to 1)"""
```
Not really a necessary change, but it would eliminate the 30 seconds I needed to think about what `[0, 1]` meant

SC_redesign/Crop_and_Soil/crop/crop_data.py
```diff
@@ -265,3 +318,408 @@ def has_given_harvest_index(self) -> bool:
     def is_in_senescence(self) -> bool:
         """check if the plant is in senescence"""
         return self.heat_fraction > self.senescent_heat_fraction
+
+
+"""
```
Isgotkowitz, line 318: Would it be possible to put each child class into its own file and put each file into a dedicated folder? If so that would make it possible for someone to see what is and isn't there without having to actually open up the file and read through it.

Isgotkowitz (changes requested):
Looks really good, especially `species_data_factory.py`. There are just a few (hopefully minor) things to discuss, and after that it should be good to merge

SC_redesign/Crop_and_Soil/crop/crop_data.py
morrowcj: It accidentally got removed. I'll put it back in (though I'm not 100% sure it will remain).

SC_redesign/Crop_and_Soil/crop/crop_data.py
morrowcj: This is a good point - I have been using set notation occasionally. I'll need to think about this more.

Isgotkowitz (approved):
Nice

SC_redesign/Crop_and_Soil/crop/crop_data.py
morrowcj: I've decided to leave this as-is for now. I may decide to do a reorganization in a separate PR.

SC_redesign/Crop_and_Soil/tests_SC/test_crop_data_and_species_factory.py
```diff
@@ -0,0 +1,473 @@
...
+    assert generic.half_mature_nitrogen_fraction == 0.02
+    assert generic.mature_nitrogen_fraction == 0.01
+    assert generic.emergence_phosphorus_fraction == 0.005
+    assert generic.half_mature_phosphorus_fraction == 0.003
+    assert generic.mature_phosphorus_fraction == 0.002
+    assert generic.optimal_harvest_index == 0.5
+    assert generic.min_harvest_index == 0.2
+    assert generic.yield_nitrogen_fraction == 0.2
+    assert generic.yield_phosphorus_fraction == 0.003
+    del generic  # cleanup
```
JoeWaddell: This is a question for @ltp23 and others - would using a pytest fixture get around the need to delete a temporary object like this? It could be over-engineering things, but it's what I've tried doing in the [animal_tests branch](https://github.com/RuminantFarmSystems/MASM/pull/284/files), but I'm not sure if I've done it the best way (it works, but it's still in review so I can't vouch that it's 100% correct).

SC_redesign/Crop_and_Soil/tests_SC/test_crop_data_and_species_factory.py
```diff
@@ -0,0 +1,473 @@
...
+
+    # check that each attribute is set appropriately
+    assert aspen.name == "custom crop: aspen"
+    assert aspen.species == "aspen"
+    assert aspen.scientific_name == "Populus tremuloides"
+    assert aspen.plant_code == "PTREM"
+    assert aspen.is_perennial is True
+    assert aspen.is_nitrogen_fixer is False
+    assert aspen.max_leaf_area_index == 5.0
+
```
JoeWaddell, line 411: If the fixture method isn't the way to go, would this require a `del aspen` line here, for cleanup?

JoeWaddell (approved):
This looks like good work! Straightforward code, and it runs without problems on my setup. 

Only a couple minor comments where I try to summon @ltp23 for some insight in how to handle unit tests. 

Also: I appreciate the Summary/Changes/Details you wrote: that helps a lot in reviewing a PR, especially for folks outside your subject area.

SC_redesign/Crop_and_Soil/tests_SC/test_crop_data_and_species_factory.py
morrowcj: I'm not sure it was necessary to delete anything at all. As I understand it, all these objects only exist in the scope of the function and are destroyed once it is done running. But, because this particular function creates many different class instances during it's execution, I thought it might be good to prevent clutter. I'm curious to know what @ltp23 has to say.

JoeWaddell: Oh of course! It would be a local, since it's just inside the test function. Makes me feel better, because in the branch I linked there, I wasn't deleting those objects either (in the cases I didn't use a fixture).

ltp23: I agree it is not necessary to delete those local objects. It is only tricky when we have class variables, then we would have to save the initial states and then reassign afterward.

SC_redesign/Crop_and_Soil/tests_SC/test_crop_data_and_species_factory.py
ltp23: Same as above. We don't need to delete the object. 

Just an aside, when we assert a boolean variable, we don't need to write `is True` or `is False`. A simpler `assert aspen.is_perennial` or `assert not aspen.is_perennial` would work too.

ltp23:
I really like how you documented those dataclasses. I need to follow that in the code I wrote. 

There is another way to handle default values for each crop data object but it takes about the same number of lines of code. It may help streamline the unit testing a bit if using parameterization. 

```
@dataclass(kw_only=True)
class Corn(CropData):
    """crop data class with default values for corn"""
    pass
```

```
class DefaultCropDataFactory:
    DEFAULT_CORN = Corn(
            species="corn",
            name="default corn",
            plant_code="CORN",
            # ....
    )

    @classmethod
    def get_default_crop_data(cls, crop_species: CropSpecies) -> CropData:
        default_crop_data_by_species = {
            CropSpecies.CORN: cls.DEFAULT_CORN,
            # ...
        }
        return default_crop_data_by_species[crop_species]

```

ltp23 (approved):

PR #319, opened 2023-01-31
Infiltration Refactor

### LINKED ISSUE #318: refactor and test infiltration.py (soil) into SC_redesign directory

Isgotkowitz:
# Summary
This PR refactors the `Infiltration` module (`infiltration.py`) of the `Soil` module from SWAT 2:1.1. All methods except the main routine are static methods.

Every new method within `Infiltration` is tested, though almost all of them with entirely arbitrary values. 

# Other Changes
* `SoilData`  has been updated to track attributes used and modified in `Infiltration`
* `SoilData` also now includes 3 `@property` methods to render needed data for `Infiltration` methods

# Notes
This PR should not be merged into `SC_Redesign` until #307 (Evapotranspiration Refactor) has been merged.
All reviewers should ignore changes in the files 
* `crop_data.py`
* `water_dynamics.py`
* `test_water_dynamics.py`
* `evapotranspiration.py`
* `test_evapotranspiration.py`
* `soil.py`
* `soil_tests/__init__.py`

while reviewing this PR


SC_redesign/Crop_and_Soil/crop/water_dynamics.py
```diff
@@ -2,20 +2,47 @@
...
         self.data.cumulative_evaporation = evaporation
         self.data.cumulative_transpiration = transpiration
         self.data.max_cumulative_evapotranspiration = max_evapotranspiration
         self.data.cumulative_evapotranspiration = self._determine_evapotranspiration(self.data.cumulative_evaporation,
                                                                                      self.data.cumulative_transpiration)
         self.data.water_deficiency = self._determine_water_deficiency(self.data.cumulative_evapotranspiration,
                                                                       self.data.max_cumulative_evapotranspiration)
+        # @CHECKME: cumumlative evaporation, transpiration, evapotranspiration, and maximum cumulative
+        # evapotranspiration are all listed as yearly totals, but maximum transpiration is a daily value.
+        # Do they need to calculated in separate methods?
```
morrowcj: I don't know much about `@CHECKME`, but I think this should be changed to a `TODO`, since I don't think the former is used by our team.

SC_redesign/Crop_and_Soil/crop/water_dynamics.py
```diff
@@ -2,20 +2,47 @@
...
+
+        Returns:
+            maximum transpiration in mm
+
+        SWAT Reference: 2:2.3.5, 6
+        """
+        if leaf_area_index <= 3:  # 2:2.3.5
+            return (potential_evapotranspiration_adjusted * leaf_area_index) / 3
+        else:  # 2:2.3.6
+            return potential_evapotranspiration_adjusted
```
morrowcj: What if $LAI \leq 0$ (2:2.3.5)? Is there already something in place to prevent that?

SC_redesign/Crop_and_Soil/crop/water_dynamics.py
```diff
@@ -32,7 +59,8 @@ def _determine_evapotranspiration(evaporation: float, transpiration: float) -> f
         return evaporation + transpiration
 
     @staticmethod
-    def _determine_water_deficiency(evapotranspiration: float, max_evapotranspiration: float) -> float:  # pseudocode: C.9.C.1
+    def _determine_water_deficiency(evapotranspiration: float,
+                                    max_evapotranspiration: float) -> float:  # pseudocode: C.9.C.1
```
morrowcj: No need for the pseudo-code reference any more.

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -0,0 +1,27 @@
...
+    top_depth: Optional[float] = None
+    """top depth of the layer (mm)"""
+    bottom_depth: Optional[float] = None
+    """bottom depth of the layer (mm)"""
+    nitrate: float = 1.5
+    """nitrate level of the layer ()"""  # TODO: units?
+    soil_water_content: float = 0.25  # arbitrary
+    """soil water content of the layer (mm)"""
+    field_capacity_water_content: float = 0.3  # arbitrary
+    """water content of soil layer at field capacity (mm)"""
```
morrowcj: As you know, the old code had a `field_capacity` variable and a `fc_water` variable. The former was actually the *concentration* of water in each layer (i.e., $\frac{\text{mm } H_{2}O }{\text{mm soil}}$) while the latter was a *volume* of water for the layer (i.e., $\text{concentration} \times \text{layer thickness}$).

I believe that, even though we need the volume, the more realistic input value is actually the concentration (since it is a function of soil composition). This means both variables should probably be retained by the `LayerData` class, with a method for converting concentration into volume. Does that make sense and do you agree? This will likely apply to other variables like "saturation capacity", etc.

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+@dataclass
+class LayerData:
+    top_depth: Optional[float] = None
+    """top depth of the layer (mm)"""
+    bottom_depth: Optional[float] = None
+    """bottom depth of the layer (mm)"""
+    nitrate: float = 1.5
+    """nitrate level of the layer ()"""  # TODO: units?
```
morrowcj: The units for nitrates are *kilograms per hectare*.

```suggestion
    """nitrate level of the layer (kg/ha)"""
```

but, in addition to the volume (OK, technically this is a density at the large scale of a field) , I suspect we will also need concentration (i.e., small-scale density; see SWAT 3:1.1.1), but we can cross that bridge when we get to it.

SC_redesign/Crop_and_Soil/soil/soil.py
```diff
@@ -1,9 +1,17 @@
...
+from SC_redesign.Crop_and_Soil.soil.evapotranspiration import Evapotranspiration
+from SC_redesign.Crop_and_Soil.soil.soil_data import SoilData
+from SC_redesign.Crop_and_Soil.soil.layer_data import LayerData
 
 
 class Soil:
-    def __init__(self):
+    def __init__(self, soil_data: Optional[SoilData] = None):
+        data = soil_data or SoilData()
+        self.evapotranspiration = Evapotranspiration(data)
```
morrowcj: This class looks good as a start, but note that the "whole-profile attributes" below will need to be moved to the `SoilData` (but not `LayerData`) class, or removed entirely if they already exist there.

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -0,0 +1,81 @@
+from dataclasses import dataclass
+from typing import List, Optional
+from SC_redesign.Crop_and_Soil.soil.layer_data import LayerData
+
+
+@dataclass
+class SoilData:
```
morrowcj, line 12: Let's add class-level docstrings with a concise overall description of the classes

morrowcj (approved):
This looks great. I've made a handful of mostly minor comments. Address them as you see fit, but I'm approving.

SC_redesign/Crop_and_Soil/crop/water_dynamics.py
Isgotkowitz: Done

SC_redesign/Crop_and_Soil/soil/infiltration.py
```diff
@@ -0,0 +1,291 @@
...
+        self.data = soil_data or SoilData()
+
+    def infiltrate(self, second_moisture_condition_parameter: float, rainfall: float, is_top_layer_frozen: bool,
+                   weighting_coefficient: float) -> None:
+        """main routine for determining runoff and infiltration of soil for a given day"""
+        third_moisture_condition_parameter = self._determine_third_moisture_condition_parameter(
+                                                                                    second_moisture_condition_parameter)
+
+        # --- adjust moisture condition parameters for slope of soil, if necessary -------------------------------------
+        if self.data.average_slope_fraction != 0.05:
```
ltp23: This is potentially buggy when you compare two floats directly for non-equality like this. A safer option is to check the absolute difference.

SC_redesign/Crop_and_Soil/crop/water_dynamics.py
Isgotkowitz: I don't know.  The default `leaf_area_index` in `CropData` is greater than 0, none of the crop species in your PR for `CropData` use a value other than the default, and in `LeafAreaIndex` the function that determines the change to `leaf_area_index` is lower bounded at 0, so it shouldn't be possible for `leaf_area_index` to be less than 0. 

However, it might be a really good idea to raise an error here is `leaf_area_index` is less than 0, cause it would indicate a bug somewhere. Then again, if we add input validation here it might mean we need validate input everywhere, which would make our code more complex. This is probably something we should talk to Pooya about.

SC_redesign/Crop_and_Soil/crop/water_dynamics.py
Isgotkowitz: Done

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: Yes, this makes sense, it should be easy to add `@property` methods into `LayerData` to calculate volume for a given layer.

Isgotkowitz: I had no idea about this distinction though, so I will need to go back through the code with a pretty fine-toothed comb.

ltp23 (approved):
Great job, Ed! 
Overall, everything looks very good and consistent. 

I have one question about the naming of some parameters that are suffixed with the word 'parameter'. That seems somewhat redundant unless it is part of the literature.

SC_redesign/Crop_and_Soil/soil/soil_data.py
Isgotkowitz: Ok, I will get on that

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: Ok. I would like to walk through some of the SWAT sections and figure out how you tell whether an equation is using volume, density, or concentration

SC_redesign/Crop_and_Soil/soil/infiltration.py
Isgotkowitz: Ok, I will change that.

Isgotkowitz:
It is definitely a bit redundant to have `parameter` at the end of every variable, but in SWAT they're called "curve numbers", which is much more abstract as they are used as parameters for a lot of the equations.

SC_redesign/Crop_and_Soil/soil/soil.py
Isgotkowitz: Noted

SC_redesign/Crop_and_Soil/crop/water_dynamics.py
morrowcj: `leaf_area_index` gets updated every day. But it is bounded at 0 (line 67 `leaf_area_index.py`). I don't think you need to do anything here. I don't think an error is necessary in this case. We will need to do input-validation but it should be at a higher level (right after receiving input).

No need to do anything further here.

SC_redesign/Crop_and_Soil/soil/layer_data.py
morrowcj: Yeah, `SoilLayer.field_capacity` (line 394 old `soil.py`) is always initialized between 0 and 1 (see the "field_capacity" key in any of the `input/soil/*.json` files). Then `SoilLayer.fc_water` is calculated as mentioned above (line 569 old `soil.py`).

SC_redesign/Crop_and_Soil/soil/layer_data.py
morrowcj: Yeah, no worries. Note that in this specific case the answer is found at the end of the first paragraph on SWAT page 333 (5:2.3.8) 

> ... and $NO3_{ly}$ is the nitrate content of soil layer $ly$ (kg $\text{NO}_{3}\text{-N}/\text{ha}$).

And the units translate to "kilograms of 'nitrate-nitrogen' (i.e., nitrogen from nitrate molecules) per hectare" and including the name of the compound in the units is redundant, so it can be reduced to (kg/ha). 

"Content" and "amount" (words that SWAT uses interchangeably) implies a quantity. In reality, though, it isn't a quantity, it as an area density (mass / area). SWAT assumes that the base unit for everything is a hectare so it uses that fact to simplify how it discusses amounts. In short: units are complicated.

That's probably more info than you needed but, there it is.

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: I believe `SoilData` and `LayerData` now properly use concentration and volume

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: No, that is good to know

SC_redesign/Crop_and_Soil/soil/soil.py
Isgotkowitz: I have updated it so that almost all of the attributes in `Soil` are now in `SoilData`, with the exception of the list of `LayerData` objects because it won't let me set the defaults in `SoilData`

SC_redesign/Crop_and_Soil/soil/soil.py
```diff
@@ -11,17 +12,9 @@ class Soil:
...
-        self.evaporation = 20  # arbitrary
-        self.transpiration = 30  # arbitrary
-        self.evapotranspiration_max = 100  # arbitrary
-        self.available_water = 80  # arbitrary
-        self.water_capacity = 100  # arbitrary
-
-        self.evapotranspiration = None
-        self.water_factor = None
-
+        # TODO: Find a way to set defaults for soil layers in SoilData
```
morrowcj, line 17: Just a thought: another factory class might be a good answer to this, with a few default configurations along with more customizable options.

Isgotkowitz: That is a good idea, similar to how you did the `CropSpecies?

PR #321, opened 2023-01-31
updated herd_data output manager logging for _print_animal_num_warnings 

ancheypeng:
This PR modifies the `info_map` passed to the output managers logging for the `_print_animal_num_warnings` function in the `AnimalManagement` class. It adds elements from the `herd_data` dictionary to `info_map`, specifically the values relating to the number of animals in the simulation.


ew3361zh (approved):
I think this looks good, Anchey. Do we care about herd_num and replace_num at all?

The root of the issue seems to come from the AnimalManagement `init_animals` method which extraneously adds the instance of the Config object to the herd_data dictionary and then sends all that data, unaltered in any way other than structurally, to the initialize_herd method.

I think your solution gets us by that but I wonder what you think about undoing that unnecessary line `herd_data['config'] = config`?

I'll attach some photos of what I did and you can think it through whether it's a good fit for this PR. I think this PR is fine as stands so have approved the changes as is.

**In animal_management.py**
Screenshot 2023-02-01 at 10 43 58 AM

**In life_cycle.py**
Screenshot 2023-02-01 at 10 45 16 AM

**and then a little further down in life_cycle.py**
Screenshot 2023-02-01 at 10 45 03 AM

ancheypeng:
> I think this looks good, Anchey. Do we care about herd_num and replace_num at all?

Thanks for the feedback! I'm not really sure what `replace_num` does, I don't think it relates to this particular function. For `herd_num`, I just pushed a commit that updates the warning text. We changed the check to the `simulate_animals` flag rather than checking if `herd_num` is 0, and I think I just forgot to change the output text when switch to OM for logging. 

> I think your solution gets us by that but I wonder what you think about undoing that unnecessary line herd_data['config'] = config?

I think your suggested changes make sense and are an improvement.  The reason I added the `herd_data['config'] = config` line was to enable dictionary unpacking to be used to pass the arguments to `initialize_herd`. Originally, it just had each of the 10 arguments passed in individually. The `herd_data['config'] = config` and dictionary packing is a bit confusing, so I think your suggestion is better. It would be great if you could commit the change, but if you just had the changes temporarily, I can redo the changes myself.

ew3361zh:
> > I think this looks good, Anchey. Do we care about herd_num and replace_num at all?
> 
> Thanks for the feedback! I'm not really sure what `replace_num` does, I don't think it relates to this particular function. For `herd_num`, I just pushed a commit that updates the warning text. We changed the check to the `simulate_animals` flag rather than checking if `herd_num` is 0, and I think I just forgot to change the output text when switch to OM for logging.
> 
> > I think your solution gets us by that but I wonder what you think about undoing that unnecessary line herd_data['config'] = config?
> 
> I think your suggested changes make sense and are an improvement. The reason I added the `herd_data['config'] = config` line was to enable dictionary unpacking to be used to pass the arguments to `initialize_herd`. Originally, it just had each of the 10 arguments passed in individually. The `herd_data['config'] = config` and dictionary packing is a bit confusing, so I think your suggestion is better. It would be great if you could commit the change, but if you just had the changes temporarily, I can redo the changes myself.

Thanks, Anchey, I understand why you made that change better now - I did commit the changes locally. I will add some typing and change the doc strings to reflect the new arguments and then push to this branch.

Fjordrunner (approved):
LGTM! Following the convo between Niko and yourself, your design choices make sense, this should be good to go.

PR #322, opened 2023-02-01
Animal life cycle refactoring

ltp23:
Some notes:
- Currently, all the instance variables (those prefixed with `self`) are public (i.e., can be seen when accessed from outside the file). I guess not all of them need to be exposed but only those that are used by other classes. We can have a discussion on what should be public or semi-private (we would prefix it with an underscore `_`).
- Feel free to make suggestions on how to make names of variables and functions more descriptive and specific. I also think it is a good idea to spell out names instead of abbreviating them. 
- Please review the functionality of each function to see if it makes sense. 
- Please make suggestions on how to make docstrings more informative.


RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -235,408 +278,606 @@ def daily_update(self, date, calves, heiferIs, heiferIIs,
...
+    def _handle_new_born(self, sim_day: int, cow: Cow, calves_born: List[Calf]) -> None:
+        args = {
+            'id': self.animal_initializer.next_id(),
+            'breed': 'HO',
+            'birth_date': sim_day,
+            'days_born': 0,
+            'p_init': cow.p_gest_for_calf,
+            'birth_weight': cow.calf_birth_weight
+        }
+        # at parturition, the sum of P absorbed for gestation rqmts is
```
JoeWaddell, line 876: I think these comments could be modified and made part of a docstring, with cited references and/or the existing references to pseudocode

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -25,137 +40,148 @@ class LifeCycleManager:
     """
     Manages the life cycles of the animals.
```
JoeWaddell, line 43: This and other docstrings could be expanded on: see a later comment with more detail.

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -25,137 +40,148 @@ class LifeCycleManager:
...
+
+    # TODO: Annotate config after removing all the imports in all the __init__.py files
+    def initialize_herd(self, herd_num: int, calf_num: int, heiferI_num: int, heiferII_num: int,
+                        heiferIII_num: int, cow_num: int, replace_num: int, herd_init: bool, breed: str,
+                        config) -> Tuple[List[Calf], List[HeiferI], List[HeiferII], List[HeiferIII], List[Cow]]:
         """
         Generates a replacement herd to simulate the market, for the herd to get
          replacements. Initializes the herd.
 
         Args:
```
JoeWaddell: Some/most of the docstrings are from the previous version, but we'll need to update these to keep consistent across modules (following the guidelines I've linked below). Note the typing specifically, but updating Args to an underlined "Parameters", with type annotation in the docstring will make things look nice online in Sphinx format.

https://numpydoc.readthedocs.io/en/latest/example.html#example

@YijingGong would be a big help in writing/updating these, given her experience with this part of the model and the underlying animal science.

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -175,44 +201,61 @@ def initialize_herd(self, herd_num, calf_num, heiferI_num, heiferII_num,
...
+        heiferIIIs = self._get_animals(HeiferIII, heiferIII_num, breed)
+        cows = self._get_animals(Cow, cow_num, breed)
+        self.replacement_market = self.animal_initializer.get_replacement_cows(replace_num, breed)
+        return calves, heiferIs, heiferIIs, heiferIIIs, cows
 
-        if self.config['use_input_calving_interval']:
-            self.avg_CI = self.config['calving_interval']
+    # TODO: In the animal_management_animal.json,
+    #  the use_input_calving_interval attribute is set to false while the
+    #  calving_interval attribute is present, maybe set that to true instead?
```
JoeWaddell: This is a good question, and I think a separate issue to review of the default JSON values for a handful of scenarios would be useful.

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -235,408 +278,606 @@ def daily_update(self, date, calves, heiferIs, heiferIIs,
...
-            '3': 0,
-            'greater_than_3': 0
-        }
-        calving_to_preg_time_available_num = {
-            '1': 0,
-            '2': 0,
-            '3': 0,
-            'greater_than_3': 0
-        }
+        preg_heifer_num = 0  # TODO: Seems unused after calculation
```
JoeWaddell, line 276: This could be a useful metric to collect in the output manager.

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -235,408 +278,606 @@ def daily_update(self, date, calves, heiferIs, heiferIIs,
...
+                                             "culled_cow_num", "GnRH_injection_num_h", "GnRH_injection_num",
+                                             "PGF_injection_num", "PGF_injection_num_h", "ai_num", "preg_check_num",
+                                             "preg_check_num_h", "sold_calf_num", "daily_milk_production",
+                                             "open_cow_num", "vwp_cow_num", "preg_cow_num", "milking_cow_num",
+                                             "dry_cow_num", "avg_days_in_milk", "avg_days_in_preg",
+                                             "avg_cow_body_weight", "avg_parity_num", "avg_calving_interval",
+                                             "avg_breeding_to_preg_time", "avg_heifer_culling_age",
+                                             "avg_cow_culling_age", "avg_mature_body_weight"]
+
+        life_cycle_daily_herd_update = {key: vars(self)[key] for key in life_cycle_daily_herd_update_keys}
```
JoeWaddell, line 317: This is wonderfully concise! These improvements to the code are great

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -25,137 +40,148 @@ class LifeCycleManager:
...
-
-    def initialize_herd(self, herd_num, calf_num, heiferI_num, heiferII_num,
-                        heiferIII_num, cow_num, replace_num, herd_init, breed,
-                        config):
+        self.animal_config = data  # animal_config in animal_management
+        self.avg_daily_cow_milking = 0.0
+        self.initialize_db_summary: Optional[InitializationDBSummaryTypedDict] = None
+        self.avg_CI = 0.0
+
+        self.sold_heifers: List[HeiferIII] = []
```
JoeWaddell, line 83: Question for my own edification: is this essentially a clean way of showing that the empty list should be filled with a given type? e.g. this list is instantiated empty, but will later be filled with HeiferIII (or similar) objects?

tests/test_animal_life_cycle.py
```diff
@@ -0,0 +1,1350 @@
...
+    mock_calf = mocker.MagicMock(autospec=Calf)
+    mock_calf.body_weight_history = calf_body_weight_history
+    mock_calf.pen_history = calf_pen_history
+
+    mock_calf_vals = mocker.MagicMock(autospec=Dict)
+    mock_calf.get_calf_values.return_value = mock_calf_vals
+
+    heiferIs: List[HeiferI] = []
+    mock_new_heiferI = mocker.MagicMock(autospec=HeiferI)
+    mocker.patch('RUFAS.routines.animal.life_cycle.life_cycle.HeiferI',
```
JoeWaddell, line 221: To clear up some of my confusion on the patch process here:

Does it need to be patched from the `RUFAS.routines.animal.life_cycle.life_cycle` (instead of the location of the class definition itself, in: `RUFAS.routines.animal.life_cycle.HeiferI`) because the method we're testing "sees" the class inside the life_cycle.life_cycle location since it's been imported?

tests/test_animal_life_cycle.py
```diff
@@ -0,0 +1,1350 @@
...
+        'body_weight_history': heiferII_body_weight_history,
+        'pen_history': heiferII_pen_history,
+        'conceptus_weight': heiferII_conceptus_weight,
+        'calf_birth_weight': heiferII_calf_birth_weight
+    })
+    assert len(heiferIIIs) == 1
+    assert heiferIIIs[0] == mock_new_heiferIII  # check identity
+
+
+def test_extract_repro_stats_from_heiferII(mocker: MockerFixture,
```
JoeWaddell, line 375: This is a nice example of a test for a subtly complex function: just wanted to note my appreciation for your elegant unit tests!

JoeWaddell (changes requested):
Various docstrings will need to be updated, but overall this cleans things up nicely. Great work!

tests/test_animal_life_cycle.py
ltp23: Yes, this is often a tricky point. We patch at the location the class/object is being used, instead of where it is defined.

RUFAS/routines/animal/life_cycle/life_cycle.py
ltp23: Yes, the type annotation is needed here to remind us that we would only add HeiferIII objects to this list later on. Hopefully, the editor will flash a warning when we add an object of a different type. The type checker cannot infer what the desired type is from an empty list.

RUFAS/routines/animal/animal_typed_dicts.py
```diff
@@ -0,0 +1,247 @@
...
+    lactation_curve: str
+    repro_cull_time: int
+    do_not_breed_time: int
+    cull_milk_production: int
+    cow_times_milked_per_day: int
+
+    # farm level -> calf
+    male_calf_rate_sexed_semen: float
+    male_calf_rate_conventional_semen: float
+    keep_female_calf_rate: int
```
YijingGong: This should be a float. It's a ratio (0-1)

RUFAS/routines/animal/animal_typed_dicts.py
```diff
@@ -0,0 +1,247 @@
...
+    heifer_user_defined_tai_cr: float
+    cow_presynch_protocol: str
+    cow_TAI_protocol: str
+    ovsynch56_conception_rate: float
+    ovsynch48_conception_rate: float
+    cosynch72_conception_rate: float
+    cosynch5d_conception_rate: float
+    cow_user_defined_tai_cr: float
+    cow_resynch_protocol: str
+    user_define_tai_length: int
```
YijingGong, line 163: The current structure is: Heifers have 3 repro methods (ED, TAI, Synch-ED). TAI has 2 sub-methods (5dCG2P and 5dCGP), and Synch-ED has 2 sub-methods (2P and CP). Similarly, cows have 3 repro methods (ED, TAI, ED-TAI). TAI has 4 sub-methods (OvSynch 56, 48, 72, and 5d) as you wrote here. Currently, if a heifer/cow uses the TAI method, no matter which sub-method she uses, she has the same TAI_conception_rate. 

So 1) I am curious why you determine to separate the 5dCG2P_conception_rate and 5dCGP_conception_rate from TAI_conception_rate and separate cows' ovsynch56_conception_rate, etc. as well? Did Kristan suggest you to do so?
2) In addition, in Manfei's branch, the user-defined variables are deleted. I don't think we have the user-defined repro method now. 
3) I think you forget the TAI_conception_rate_1 (conception rate for cows' first service) and TAI_conception_rate_sub (conception rate for cows' second+ service)

RUFAS/routines/animal/animal_typed_dicts.py
```diff
@@ -0,0 +1,247 @@
...
+
+    # TODO: For these two variables, either leave them like this or
+    # add another typed dict for each
+    ED_related: Dict[str, Union[float, str]]
+    TAI_related: Dict[str, Union[float, str]]
```
YijingGong, line 167: Not sure what is the different between those 2 structures. About which way is more efficient. probably talk to other software engineers? or could you explain what's the difference and advantages of both? Maybe we can discuss during our Wednesday meeting.

RUFAS/routines/animal/animal_typed_dicts.py
```diff
@@ -0,0 +1,247 @@
...
+
+    # remaining attributes in farm level -> repro
+    voluntary_waiting_period: int
+    tai_program_start_day: int
```
YijingGong: tai_program_start_day should be under farm level -> repro -> TAI_related

RUFAS/routines/animal/animal_typed_dicts.py
```diff
@@ -0,0 +1,247 @@
...
+    conception_rate_decrease: float
+    avg_gestation_len: int
+    std_gestation_len: int
+    prefresh_day: int
+    num_21_days_repro: int
```
YijingGong: This one seems not used in the code. Can be deleted

RUFAS/routines/animal/animal_typed_dicts.py
```diff
@@ -0,0 +1,247 @@
...
+    # add another typed dict for each
+    ED_related: Dict[str, Union[float, str]]
+    TAI_related: Dict[str, Union[float, str]]
+
+    # remaining attributes in farm level -> repro
+    voluntary_waiting_period: int
+    tai_program_start_day: int
+    conception_rate_decrease: float
+    avg_gestation_len: int
+    std_gestation_len: int
```
YijingGong: I think it might be good to move this avg_gestation_len and std_gestation_len to the section of # from_literature -> repro. Also, the type might be good to be consistent with avg_estrus_cycle_heifer and std_estrus_cycle_heifer. Int might be ok, but maybe float would be more general for like 1.5.

RUFAS/routines/animal/animal_typed_dicts.py
```diff
@@ -0,0 +1,247 @@
...
+    preg_loss_rate_1: float
+    preg_check_day_2: int
+    preg_loss_rate_2: float
+    preg_check_day_3: int
+    preg_loss_rate_3: float
+    avg_estrus_cycle_return: float
+    std_estrus_cycle_return: float
+    avg_estrus_cycle_cow: float
+    std_estrus_cycle_cow: float
+    avg_estrus_cycle_p: float
```
YijingGong: variable name of avg_estrus_cycle_after_pgf might be more descriptive.

RUFAS/routines/animal/animal_typed_dicts.py
```diff
@@ -0,0 +1,247 @@
...
+    parity_death_prob: List[float]
+    death_cull_prob: List[float]
+    parity_cull_prob: List[float]
+    mastitis_cull_prob: List[float]
+    feet_leg_cull_prob: List[float]
+    injury_cull_prob: List[float]
+    disease_cull_prob: List[float]
+    udder_cull_prob: List[float]
+    unkown_cull_prob: List[float]  # TODO: Fix this typo in `animal_management_animal.json` and here
+    cull_day_count: List[float]
```
YijingGong: this should be int. It is day.

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -8,15 +8,30 @@
...
 from RUFAS.routines.animal.life_cycle.heiferI import HeiferI
 from RUFAS.routines.animal.life_cycle.heiferII import HeiferII
 from RUFAS.routines.animal.life_cycle.heiferIII import HeiferIII
-from RUFAS.routines.animal.life_cycle.cow import Cow
-from RUFAS.routines.animal.life_cycle.animal_base import AnimalBase
-from RUFAS.routines.animal.life_cycle.animal_initialization import AnimalInitialization
-from RUFAS.routines.animal.life_cycle import animal_constants as const
+from RUFAS.util import Utility
+
+GenericAnimal = TypeVar("GenericAnimal", bound=Union[Calf, HeiferI, HeiferII, HeiferIII, Cow])
```
YijingGong, line 36: To be honest, I am not very familiar with the typing class. Do you think it might be good to ask for a review from one more software developer? Or maybe we can use the Wednesday meeting to talk about this, or schedule a separate meeting.

YijingGong: I guess I get the idea that it is checking the type of objects, but not sure about the details like TypeVar and Union.

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -25,137 +40,148 @@ class LifeCycleManager:
     """
     Manages the life cycles of the animals.
     """
-    # statistics
```
YijingGong, line 31: Just trying to understand, could you explain why you delete all those variables without .self? Are they originally used in the life_cycle_report.py to get the values?

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -25,137 +40,148 @@ class LifeCycleManager:
...
-    culled_heifer_num = 0
-    culled_cow_num = 0
-
+    # The following class variables are used in HerdReport.
     num_cow_for_parity = {
         '1': 0,
         '2': 0,
         '3': 0,
         'greater_than_3': 0
     }
```
YijingGong, line 52: why this is outside (and the below avg_calving_to_preg_time and cull_reason_stats) while everything is inside of init function too?

RUFAS/routines/animal/life_cycle/life_cycle.py
YijingGong: In terms of your question that 'use_input_calving_interval' is false but 'calving_interval' attribute is present, they are not related. The attribute of 'calving_interval' will always be present in the json no matter 'use_input_calving_interval' is true or false. 

I believe the logic of the original code is correct. If the 'use_input_calving_interval' is true, use the user-input calving interval, else, use the calving interval from the database initialization.

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -175,44 +201,61 @@ def initialize_herd(self, herd_num, calf_num, heiferI_num, heiferII_num,
...
+            Calf: self.animal_initializer.get_calves,
+            HeiferI: self.animal_initializer.get_heiferIs,
+            HeiferII: self.animal_initializer.get_heiferIIs,
+            HeiferIII: self.animal_initializer.get_heiferIIIs,
+            Cow: self.animal_initializer.get_cows
+        }
+        animals = animal_getter_by_animal_type[animal_type](num, breed)
+        for animal in animals:
+            animal.events.add_event(animal.days_born, 0, animal_constants.INIT_HERD)
+        return animals
```
YijingGong, line 240: Nice! Very concise!

RUFAS/routines/animal/life_cycle/life_cycle.py
YijingGong, line 276: It is used to calculate 'avg_breeding_to_preg_time' (line 374-378 in left hand side old code).

RUFAS/routines/animal/life_cycle/life_cycle.py
YijingGong: I think I get it. They are used in later _reset_parity() and _reset_cull_reason_stats() functions, right?

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -235,408 +278,606 @@ def daily_update(self, date, calves, heiferIs, heiferIIs,
...
-        calving_to_preg_time_available_num = {
-            '1': 0,
-            '2': 0,
-            '3': 0,
-            'greater_than_3': 0
-        }
+        preg_heifer_num = 0  # TODO: Seems unused after calculation
+
+        self._reset_parity()
+        self._reset_cull_reason_stats()
```
YijingGong, line 280: I am thinking do we need these two functions? Since life_cycle_manager object will be initialized every day in the animal_management.py, which means every day, the parity and cull_stats will be reset to zero already. I might be wrong. Maybe further testing is needed.

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -235,408 +278,606 @@ def daily_update(self, date, calves, heiferIs, heiferIIs,
...
+
+        total_animal_num = self._evaluate_calves_for_weaning(sim_day, calves, heiferIs, total_animal_num)
+        total_animal_num = self._evaluate_heiferIs_for_transitioning_to_heiferIIs(sim_day, heiferIs, heiferIIs,
+                                                                                  total_animal_num)
+        total_animal_num, preg_heifer_num = self._evaluate_heiferIIs_for_transitioning_to_heiferIIIs(sim_day, heiferIIs,
+                                                                                                     heiferIIIs,
+                                                                                                     preg_heifer_num,
+                                                                                                     total_animal_num)
+        total_animal_num = self._evaluate_heiferIIIs_for_transitioning_to_cows(sim_day, heiferIIIs, cows,
+                                                                               total_animal_num)
```
YijingGong, line 290: Love those 4 functions. Beautiful.

RUFAS/routines/animal/life_cycle/life_cycle.py
YijingGong, line 317: Very cool. To enhance my understanding, Is this supposed to get the values from the class and later can be accessed in the life_cycle_report? But it seems to be locol (not public)? Or it is made public using the info_map? Can you explain a little bit the logic and syntax here?

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -235,408 +278,606 @@ def daily_update(self, date, calves, heiferIs, heiferIIs,
...
+
+        self._check_if_heifers_need_to_be_sold(heiferIIIs, cows, ids_removed)
+        self._check_if_replacement_heifers_needed(sim_day, heiferIIIs, cows, animals_added)
+
```
YijingGong, line 296: I believe those two functions should be after the _cull_cows_and_record_stats() because the _cull_cows_and_record_stats() function can change the cow number in the herd.

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -235,408 +278,606 @@ def daily_update(self, date, calves, heiferIs, heiferIIs,
...
+        return total_animal_num, preg_heifer_num
+
+    def _extract_repro_stats_from_heiferII(self, heiferII: HeiferII) -> None:
+        """Extracts the reproductive stats from a heiferII.
+
+        Args:
+            heiferII: The heiferII to extract the stats from.
+
+        """
+        # self.CIDR_count += heiferII.CIDR_count  # TODO: add CIDR count to heiferII
```
YijingGong, line 594: Why this is commented out?

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -235,408 +278,606 @@ def daily_update(self, date, calves, heiferIs, heiferIIs,
...
+        If the number of heifers exceeds what is needed for the herd,
+        sell those as replacement
+
+        Args:
+            heiferIIIs: The list of heiferIIIs.
+            cows: The list of cows.
+            ids_removed: The list of ids of animals removed from the herd.
+
+        """
+        sell_threshold = 1.03
```
YijingGong, line 668: I think it might be good to make the buy_threshold and sell_threshold user-input variables. I can bring this up for discussion.

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -235,408 +278,606 @@ def daily_update(self, date, calves, heiferIs, heiferIIs,
...
-                self.culled_cow_num, self.avg_cow_culling_age = \
-                    self.calc_average(self.culled_cow_num,
-                                      self.avg_cow_culling_age, cow.days_born)
-
-                # print(len(culled_cows))
+                self._cull_cow(cow)
                 ids_removed.append(cow.id)
-                del cows[index]
-
+                removed_cows_idx.append(index)
```
YijingGong, line 736: If I have this removed_cows_idx.append(index), and later Utility.remove_items_from_list_by_indices(cows, removed_cows_idx), why do I still need to keep the ids_removed?

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -235,408 +278,606 @@ def daily_update(self, date, calves, heiferIs, heiferIIs,
...
+        """
+        denominator = total_animal_num if total_animal_num > 0 else 1
+        pc = Utility.percent_calculator(denominator)
+        self.calf_percent = pc(self.calf_num)
+        self.heiferI_percent = pc(self.heiferI_num)
+        self.heiferII_percent = pc(self.heiferII_num)
+        self.heiferIII_percent = pc(self.heiferIII_num)
+        self.cow_percent = pc(self.cow_num)
+
+    def _calculate_cow_percentages(self) -> None:
```
YijingGong, line 909: Like this function. Very concise.

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -235,408 +278,606 @@ def daily_update(self, date, calves, heiferIs, heiferIIs,
...
+        sell those as replacement
+
+        Args:
+            heiferIIIs: The list of heiferIIIs.
+            cows: The list of cows.
+            ids_removed: The list of ids of animals removed from the herd.
+
+        """
+        sell_threshold = 1.03
+        while len(heiferIIIs) + len(cows) > self.herd_num * sell_threshold and len(heiferIIIs) > 0:
```
YijingGong, line 669: I previously found there was an inconsistency between len(cows) and cow_num, which I think was because the old code used the `del` in the loop to delete the animals. Since you modified that already, I guess now it should be fine. But just want to point this out. 

So I am thinking maybe there's no need to use both (cow_num and len(cows)) in the life_cycle.py, which causes confusion and potential problems. What do you think?

YijingGong (changes requested):
Great job, Loi. That's a lot of work. Thank you for your efforts. I really like how you make a lot of functions much more concise and readable. I also made some minor comments on the code.

RUFAS/routines/animal/life_cycle/life_cycle.py
ltp23: I understand this point. Since the daily update function is being used somewhere else  (in animal_management.py, I think) and ids_removed is one of the return variables, we would need to remove it here and there at the same time.

YijingGong: I see. Sounds good

RUFAS/routines/animal/animal_typed_dicts.py
ltp23: Thanks Yijing for the details. Here, I was just listing out the attributes I saw in the `animal_management_animal.json`. Would you like to update that JSON file in lines 37-57, just because I don't understand the subject matter enough? Then, I would update the typed dictionaries accordingly.  

The purpose of all the typed dictionaries is to guard against using undefined keys, so there is a need for an exact match between what is in the JSON file and what is in the type dicts definition.

RUFAS/routines/animal/animal_typed_dicts.py
ltp23: So, in the `animal_management_animal.json`, we have the following:

```
{
    "herd_information": {
        ...
    },
    "animal_config": {
       "management_decisions": {
           ...
       },
       "farm_level": {
           "calf": {
                ...
            },
            "repro": {
                "ED_related": {
                    ...
                },
                "TAI_related": {
                    ...
                 }
            }
       },
       "from_literature": {
           ...
        }
   },
   ...
}
```

The two keys `ED_related` and `TAI_related` are nested two levels deep, but when we define and use the config object, we want to flatten it all out. So we go from `animal_config -> farm_level -> repro -> ED_related & TAI_related` in the JSON object to just `animal_config -> ED_related & TAI_related` in the config object.  It may seem complicated, but you don't need to worry about it because these two keys are only there for completeness' sake.

RUFAS/routines/animal/animal_typed_dicts.py
ltp23: Alternatively, I think it's probably better to make a separate pull request to address the contents of the JSON file along with your suggestions for buy/sell thresholds.

RUFAS/routines/animal/animal_typed_dicts.py
ltp23: It's fixed now.

RUFAS/routines/animal/animal_typed_dicts.py
ltp23: I could only delete the `num_21_days_repro` key. The other three are used in `heiferII.py`.

RUFAS/routines/animal/animal_typed_dicts.py
ltp23: I changed the type to float for these two attributes and moved them to be under the `from_literature -> repro` section.

RUFAS/routines/animal/animal_typed_dicts.py
ltp23: OK. I changed `avg_estrus_cycle_p` to `avg_estrus_cycle_after_pgf`, and `std_estrus_cycle_p` to `std_estrus_cycle_after_pgf`.

RUFAS/routines/animal/animal_typed_dicts.py
ltp23: It's fixed now, and I fixed the type above too.

RUFAS/routines/animal/life_cycle/life_cycle.py
ltp23: I added some descriptions. Hopefully, it helps clarify a bit more.

RUFAS/routines/animal/life_cycle/life_cycle.py
ltp23: Before those variables were class variables, i.e., they were defined right under the class name. So I moved them into the `__init__` function to make them instance variables. Class variables are shared among all the objects of the class, whereas instance variables are associated with a specific object.

In the case of the LifeCycleManager class, as we have here, I think it's ok to go with the previous design of using class variables too. It is likely that we only have one LifeCycleManager object, so it doesn't matter. 

Also, both class variables and instance variables can be used with `self.`.

RUFAS/routines/animal/animal_typed_dicts.py
YijingGong: ohhhh, I see. We are looking about the different json files... I will slack you about it.

RUFAS/routines/animal/life_cycle/life_cycle.py
ltp23: Same as above.

RUFAS/routines/animal/life_cycle/life_cycle.py
ltp23: That makes sense and I agree. It is easier to always have a calving interval in the JSON file and we can decide to use it or not, instead of adding or removing it depending on the true/false value of use_input_calving_interval.

RUFAS/routines/animal/life_cycle/life_cycle.py
ltp23: Generally, manager classes tend to be instantiated once, i.e., only have only one object. Of course, we could create many objects out of a class, but the expected pattern of usage is to have only one manager object.

RUFAS/routines/animal/life_cycle/life_cycle.py
ltp23: Yes, the `vars()` function will return a dictionary that contains object attributes, and you can look things up using keys. What you see here is the results being recorded to the output manager (the variable `om`).

RUFAS/routines/animal/life_cycle/life_cycle.py
ltp23: I moved those two functions down as you suggested. 

This is a good reminder for me to keep functions side-effect free (no changes to `self` variables) and make them return values instead. This way, any changes to the `self` variables only happen in one or a few places like in the `daily_update()`.

RUFAS/routines/animal/life_cycle/life_cycle.py
ltp23: The current version of HeiferII class in the master branch doesn't have the `CIDR_count` variable.

RUFAS/routines/animal/life_cycle/life_cycle.py
ltp23: We can add these two parameters to the json files later.

RUFAS/routines/animal/life_cycle/life_cycle.py
ltp23: I think this depends on what `cow_num` means. Does it always reflect the current size of the `cows` list? I don't fully grasp how these two variables correlate yet.

RUFAS/routines/animal/life_cycle/life_cycle.py
ltp23: Yijing clarifies this point below so I have this TODO removed.

RUFAS/routines/animal/life_cycle/life_cycle.py
ltp23: Yes, I will update the docstring style from the Google style to the Numpy style. I will add some more text to the docstrings. We can always update them as we go along.

input/animal/animal_feed_management.json
```diff
@@ -110,7 +109,8 @@
 				"injury_cull_prob": [0, 0.08, 0.18, 0.28, 0.38, 0.47, 0.56, 0.64, 0.71, 0.78, 0.85, 0.90, 0.95, 1],
 				"disease_cull_prob": [0, 0.04, 0.12, 0.24, 0.34, 0.42, 0.50, 0.57, 0.64, 0.72, 0.81, 0.89, 0.95, 1],
 				"udder_cull_prob": [0, 0.12, 0.24, 0.33, 0.41, 0.48, 0.55, 0.62, 0.68, 0.76, 0.82, 0.89, 0.95, 1],
-				"unkown_cull_prob": [0, 0.05, 0.11, 0.18, 0.27, 0.37, 0.45, 0.54, 0.62, 0.70, 0.77, 0.84, 0.92, 1],
+				"unknown_cull_prob": [0, 0.05, 0.11, 0.18, 0.27, 0.37, 0.45,
+					0.54, 0.62, 0.70, 0.77, 0.84, 0.92, 1],
```
YijingGong, line 114: Just curious, what's the difference between the new version and the old version?

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -898,7 +898,7 @@ def resynch_ed_tai(self, sim_day):
                 AnimalBase.config['conception_rate_decrease']
             self.estrus_day = self.determine_estrus_day(
                 self.abortion_day, const.ESTRUS_AFTER_PGF_NOTE,
-                AnimalBase.config['avg_estrus_cycle_p'],
+                AnimalBase.config['avg_estrus_cycle_after_pgf'],
                 AnimalBase.config['std_estrus_cycle_p'], sim_day)
```
YijingGong: I think you forgot to change the 'std_estrus_cycle_p' to the same format as avg

RUFAS/routines/animal/life_cycle/life_cycle.py
YijingGong: I see... It will come later in the new repro branch. We can worry about it later.

RUFAS/routines/animal/life_cycle/life_cycle.py
YijingGong: Issue submitted

RUFAS/routines/animal/life_cycle/life_cycle.py
YijingGong: `cow_num` should be the number of lactating cows in the herd on this simulation day. At the same time, the `cows` list keeps the ids of the cows in the herd on this simulation day. So I believe they should be the same. You deal with it in `_cull_cows_and_record_stats` function, if the cow is culled, you delete her id from the list. If she is not culled, you use `_handle_cow_body_weight_and_parity` to count how many cows are there.

YijingGong:
Good job! I think it looks good now. Just a few follow-up discussions and one thing you need to change.

YijingGong (changes requested):

input/animal/animal_feed_management.json
ltp23: It's the same, but with the typo fixed (unkown -> unknown)

RUFAS/routines/animal/life_cycle/cow.py
ltp23: You are right. I'll fix it.

RUFAS/routines/animal/life_cycle/life_cycle.py
ltp23: This can potentially become a difficult bug to track. The count of each animal type can get out of sync with the size of their list. 

At the start of the `daily_update()`, should we reset all the counts (e.g., `self.calf_num`, `self.heiferI_num`, ...) to 0?

YijingGong: Yes, at the start of the `daily_update()`, those counts (e.g., `self.calf_num`, `self.heiferI_num,` ...) should be reset to zero. Because they are supposed to count the animal number of today. 

As I commented, I found this bug was due to the old code using the `del` in the loop to delete the animals. Since it was modified, maybe now we don't have this bug. But it would be good to keep it in mind. And if we can test this in the unit test, or talk about the possibility to use`cow_num` instead of `len(cows)`, it might be better.

input/animal/animal_feed_management.json
YijingGong: I see the typo now... Really hard to find lol

YijingGong:
I think this PR looks good to me now. It looks like you need to re-request the review and I can approve it? Anyway, I am ok to merge it.

YijingGong (approved):
Looks good to me!

KFosterReed:
Hi team! I had a bit of extra time today and wanted to take a look at this PR. Looking at the code, the clarity and documentation is leagues above where it was before. Kudos for that and thank you for all the work to get to this stage! When I ran the model however, it looks like there is something wrong with the outputs because the animal numbers returned in the outputs are continuously increasing instead of tracking the current herd demographics. See the image below.  
image (1)

I am assuming that this is just in the outputs because it should cause other problems in the model (i.e. create more and more pens!) if it was actually simulating 120,000 heiferI animals after 1 year. So hopefully it will be a relatively easy fix :)

YijingGong:
As reminded by Kristan, I checked other outputs of the life cycle part. Besides the increase in the count of all animal classes,  the same pattern is also found in culled animal number (culled_cow_num, culled_heifer_num), sold animal number, animal number for each parity, number of hormone injections, number of pregnancy checks, number of ai. 

However, I found that the culled cow number for the 8 different reasons is correct. So it's likely related to how you initialize the variables. For the cull_reason, you set them to zero at 2 places (under the `class LifeCycleManager` and the `def __init__(self, data: AnimalConfigTypedDict)` function).

For most other variables, you only set them to be zero in the `def __init__(self, data: AnimalConfigTypedDict)` function.

Another thing I found weird is the animal initialization:
Screenshot 2023-03-02 at 4 00 25 PM

Screenshot 2023-03-02 at 4 02 00 PM

Screenshot 2023-03-02 at 4 02 09 PM

It looks like the program initializes with cows only with fresh first lactation cows and cows in parity 3 or higher. During simulation, some cows in parity 1 grow to parity 2.

JoeWaddell:
To build on the comments here: looking through the graphics output, there's many places related to the bug in the total number of animals. I'm still troubleshooting, but wanted to add a couple example figures that should be instructive. 

Below is from this branch: not only does the number of animals increase, but so does the percentage! 40,000% is certainly not correct. 
image

Related, is that the cull percentage is shrinking. Below is from this branch: 
image
It seems like the value being referenced here for the denominator/total in the percent calculation is just constantly increasing, like the figures in Kristan's comment. 


Below is from the master branch. The reported ranges are percentages from 0-100% of the total culled animals, which is correct. 
image

Cow percent also reports reasonable values in the master branch: 
image

JoeWaddell (approved):
This looks great: with the new function that resets the animal numbers, fixing the bug reported in the comments, it looks ready to merge.

PR #323, opened 2023-02-01
Adding a python-requires in setup.py

ltp23:
In this very small pull request, I specified the minimum Python version in the setup file. 

I think this enforcement only takes effect for a fresh download and setup of the project. 

For the error that Kristan encountered with the type annotation of the Union type, a simple fix is to add the following - `from __future__ import annotations` - to the top of that file. This should help with backward compatibility regarding type annotation.


JoeWaddell (approved):
Great! My first review of a one line PR.  It follows the guidelines here to the letter: https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires , so this is an easy approval.

PooyaHekmati (approved):

PR #324, opened 2023-02-02
Laplacette cm patch 1

LaplacetteCM:
Files updated with no need to review (I've deleted the math blocks and changed the way we explain the equations according to the last PR comments):
- growth_constraints.md
- biomass_allocation.md
- leaf_area_index.md

Files updated that need to be reviewed:
- root_development.md


morrowcj (approved):
Celina, this looks great to me. 

One small thing: would you move the docs in `Crop_and_Soil/docs/` into `docs_SC/model_docs/`? If that is a pain with github, let me know and I can do it.

LaplacetteCM:
Done! If I did it correctly, all the files should be now in the model_docs folder. I think I need to add one more reviewer to be allowed to merge.

LaplacetteCM:
Clay, I have included the update on water_dynamics.md in this PR as well. It'd need a review because it's the same as for root_development.md. 

Especially, if you could check in the first paragraph that I am correctly applying the concept of method vs function that Niko shared. I understand we only have methods here. (self.data.method)

morrowcj:
These changes look good, Celina. Yes, I think the usage of method is now in line with Niko's recommendations. 

You could request another reviewer or, if you're satisfied, I can bypass that requirement and merge the branch when ready. I think doing so would be fine for these small changes.

LaplacetteCM:
Great Clay, go ahead and merge it whenever you want. I'll try to finish with the remaining wikis next week but I'll start a new one with them. Or we could leave this one open. You'll know better than me what works best!

morrowcj:
OK. I'm merging this now. Thanks again for all the great work!

PR #328, opened 2023-02-03
Refactor and test percolation.py in the Soil module

### LINKED ISSUE #327: refactor and test percolation.py (soil) into SC_redesign directory

Isgotkowitz:
# Summary
This PR refactors the `Percolation` module (`percolation.py`) of the `Soil` module from SWAT 2:3.2. It is entirely comprised of `@staticmethods` except for the main routine (`percolate()`) that does the calculations and updates for an object instance. 

Every method in `Percolation` is tested, however `test_percolate_between_layers()` takes on a sequential format instead of the usual format: using multiple sets of parameters. `test_percolate_between_layers()` aims for full code coverage by testing 3 different cases, all of which are spelled out in its docstring.

# Other Changes
- `Infiltration` has been slightly refactored to use the new `temperature` attribute of `LayerData` objects instead of accepting a parameter in the main routine to determine if the top layer of soil is frozen or not. `test_infiltration.py` has also been updated to reflect these changes
-  `Infiltration` has had a missing docstring added to a static method
- the `LayerData` module has had several attributes and `@property` methods added that are used in `Percolation` (and possibly other modules), some attribute names have been changed, method signatures now have return types, and it now uses a `__post_init__` function to initialize mutable variables that are dependent upon other variables in the class
- `soil_water_content` inside the `LayerData` module has been modified so that it is now mutable. This has been achieved by changing it from being calculated by an `@property` method to initializing it as none, and adding a `__post_init__()` method to calculate its initial value as soon as a `LayerData` object is created.
- the `SoilData` module has two attributes added for use in `Percolation`: a `LayerData` object representing the vadose zone, and a time_step over which percolation occurs. There is a correction made to `profile_soil_water_content()` so it is now correctly lower bounded at 0. It now includes a `__post_init__` method that will initialize its `soil_layers` and `vadose_zone_layer` 
- `Soil` has had the initialization of its soil layers removed, because that is now done inside the `SoilData` class
- All files have had their units changed so they are displayed as `(...)` instead of `in ...`


SC_redesign/Crop_and_Soil/soil/percolation.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def percolate(self, has_seasonal_high_water_table: bool) -> None:
+        """executes percolation of excess water in each layer of soil profile to the layer directly beneath it
+
+        Args:
+            has_seasonal_high_water_table: if the HRU has a seasonal high water table (true/false)
+
+        SWAT Reference: sections 2:3.1 and 2
+        """
+        for i in len(self.data.soil_layers):
```
morrowcj: I think you're missing a call to `range()`.

Isgotkowitz: Yes

SC_redesign/Crop_and_Soil/soil/infiltration.py
```diff
@@ -56,7 +56,7 @@ def infiltrate(self, second_moisture_condition_parameter: float, rainfall: float
                                                                   second_shape_coefficient)
 
         # --- Adjust retention parameter only if top layer of soil is frozen -------------------------------------------
-        if is_top_layer_frozen:
+        if self.data.soil_layers[0].temperature <= 0:
```
morrowcj, line 59: I wonder if we'll need to account for frozen soil elsewhere. If so, it might be nice to have a similar `SoilData.is_frozen` variable.

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -17,36 +18,59 @@ class LayerData:
...
     field_capacity_water_concentration: float = 0.3  # arbitrary
     """water concentration of soil layer at field capacity (mm)"""
     wilting_point_water_concentration: float = 0.2  # arbitrary
     """water concentration of soil layer at wilting point (mm)"""
     saturation_point_water_concentration: float = 0.5
     """water concentration of soil layer at saturation point (mm)"""
     soil_evaporation_compensation_coefficient: float = 1
-    """coefficient that allows user to modify depth distribution used to meet the soil evaporative demand (2:2.3.17)"""
+    """coefficient that allows user to modify depth distribution used to meet the soil evaporative demand (unitless) 
+        (2:2.3.17)"""
```
morrowcj:
```suggestion
        (SWAT 2:2.3.17)"""
```

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -17,36 +18,59 @@ class LayerData:
...
-    def field_capacity_content(self):
+    def field_capacity_content(self) -> float:
         """volume of water in layer when at field capacity in mm"""
-        return self.field_capacity_water_concentration / self.layer_thickness
+        return self.field_capacity_water_concentration * self.layer_thickness
 
     @property
-    def wilting_point_content(self):
+    def wilting_point_content(self) -> float:
         """volume of water in layer when at wilting point"""
```
morrowcj:
```suggestion
    def wilting_point_content(self) -> float:
        """amount of water in layer when at wilting point (mm)"""
```

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -17,36 +18,59 @@ class LayerData:
...
     soil_water_concentration: float = 0.25  # arbitrary
     """soil water concentration of the layer (mm)"""
+    soil_water_content:  Optional[float] = None  
+    """volume of soil water in the layer (mm)"""
     field_capacity_water_concentration: float = 0.3  # arbitrary
     """water concentration of soil layer at field capacity (mm)"""
     wilting_point_water_concentration: float = 0.2  # arbitrary
     """water concentration of soil layer at wilting point (mm)"""
     saturation_point_water_concentration: float = 0.5
     """water concentration of soil layer at saturation point (mm)"""
```
morrowcj: Just to make this a bit clearer, since water is measured in a confusing unit:

```suggestion
    field_capacity_water_concentration: float = 0.3  # arbitrary
    """water concentration of soil layer at field capacity (mm water / mm soil)"""
    wilting_point_water_concentration: float = 0.2  # arbitrary
    """water concentration of soil layer at wilting point (mm water / mm soil)"""
    saturation_point_water_concentration: float = 0.5
    """water concentration of soil layer at saturation point (mm water / mm soil)"""
```

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -17,36 +18,59 @@ class LayerData:
...
+    def excess_water_available(self):
+        """volume of water available for percolation in the soil layer in mm
+
+        SWAT Reference: 2:3.2.1, 2
+        """
+        return max(0, self.soil_water_content - self.field_capacity_content)
+
+    @property
+    def saturation_content(self) -> float:
         """volume of water in layer when saturated in mm"""
```
morrowcj: Let's stick to the unit style used elsewhere.

```suggestion
        """volume of water in layer when saturated (mm)"""
```

SC_redesign/Crop_and_Soil/soil/percolation.py
```diff
@@ -0,0 +1,145 @@
...
+        Args:
+            has_seasonal_high_water_table: if the HRU has a seasonal high water table (true/false)
+
+        SWAT Reference: sections 2:3.1 and 2
+        """
+        for i in range(len(self.data.soil_layers)):
+            """iterate through each layer of soil in the soil profile"""
+            upper_layer = self.data.soil_layers[i]
+            if i != (len(self.data.soil_layers) - 1):
+                """upper layer is not the bottom layer of the soil profile"""
```
morrowcj: Can we reserve quote comments for docstrings or multi-line standalone comment/note blocks?

```suggestion
        for i in range(len(self.data.soil_layers)):  # iterate through each layer of soil in the soil profile
            upper_layer = self.data.soil_layers[i]
            if i != (len(self.data.soil_layers) - 1):  # upper layer is not the very bottom layer of the soil profile
```

SC_redesign/Crop_and_Soil/soil/percolation.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def percolate(self, has_seasonal_high_water_table: bool) -> None:
+        """executes percolation of excess water in each layer of soil profile to the layer directly beneath it
+
+        Args:
+            has_seasonal_high_water_table: if the HRU has a seasonal high water table (true/false)
+
+        SWAT Reference: sections 2:3.1 and 2
+        """
+        for i in range(len(self.data.soil_layers)):
```
morrowcj: I also like to use `i` by default for iterators, but the team wants to move away from this. Use a short word to indicate what the iterator represents. `layer` or `layer_num` would be appropriate here.

SC_redesign/Crop_and_Soil/soil/percolation.py
```diff
@@ -0,0 +1,145 @@
...
+
+from SC_redesign.Crop_and_Soil.soil.layer_data import LayerData
+from SC_redesign.Crop_and_Soil.soil.soil_data import SoilData
+
+"""
+This module is based on the section 'Percolation' (2:3.2) in SWAT
+"""
+
+
+class Percolation:
```
morrowcj, line 12: I think the logic in the `percolate()` method can be improved. See the file I added (`TMP_alternative_percolates.py`) and let's discuss externally.

SC_redesign/Crop_and_Soil/soil/percolation.py
```diff
@@ -0,0 +1,145 @@
...
+            """iterate through each layer of soil in the soil profile"""
+            upper_layer = self.data.soil_layers[i]
+            if i != (len(self.data.soil_layers) - 1):
+                """upper layer is not the bottom layer of the soil profile"""
+                lower_layer = self.data.soil_layers[i + 1]
+                if upper_layer.temperature > 0 and self._determine_if_percolation_allowed(
+                        lower_layer.soil_water_content,
+                        lower_layer.field_capacity_content,
+                        lower_layer.saturation_content,
+                        has_seasonal_high_water_table):
```
morrowcj:
```suggestion
                can_percolate = self._determine_if_percolation_allowed(
                        lower_layer.soil_water_content,
                        lower_layer.field_capacity_content,
                        lower_layer.saturation_content,
                        has_seasonal_high_water_table)
                if upper_layer.temperature > 0 and can_percolate:

```

SC_redesign/Crop_and_Soil/soil/percolation.py
```diff
@@ -0,0 +1,145 @@
...
+                    """percolation not allowed from upper to lower layer"""
+                    continue
+            else:
+                """upper layer is the bottom layer of the soil profile, percolation goes into vadose zone which can 
+                    accept infinite amount of water
+                """
+                if upper_layer.temperature > 0:
+                    """percolation is allowed from upper to lower layer"""
+                    lower_layer = self.data.vadose_zone_layer
+                    if upper_layer.temperature > 0:
```
morrowcj: checking the same condition twice.

SC_redesign/Crop_and_Soil/soil/percolation.py
```diff
@@ -0,0 +1,145 @@
...
+                                                                              upper_layer.saturated_hydraulic_conductivity)
+            amount_to_percolate = Percolation._determine_percolation_to_next_layer(upper_layer.excess_water_available,
+                                                                                   time_step, percolation_time)
+
+            #  Limit the maximum amount of water allowed to percolate so that lower layer cannot become overly saturated
+            if amount_to_percolate > lower_layer.acceptable_percolation_amount:
+                amount_to_percolate = lower_layer.acceptable_percolation_amount
+
+            # move water from upper layer to lower layer
+            return amount_to_percolate
```
morrowcj: bound to zero
```suggestion
            return max(0, amount_to_percolate)
```

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -17,36 +18,59 @@ class LayerData:
     """nitrate level of the layer (kg/ha)"""
     soil_water_concentration: float = 0.25  # arbitrary
     """soil water concentration of the layer (mm)"""
+    soil_water_content:  Optional[float] = None
```
morrowcj:
```suggestion
    water_content:  Optional[float] = None  
```

SC_redesign/Crop_and_Soil/soil/percolation.py
morrowcj: Furthermore, I'd recommend including the temperature check in `_determine_if_percolation_allowed()`

SC_redesign/Crop_and_Soil/tests_SC/TMP_alternative_percolates.py
```diff
@@ -0,0 +1,81 @@
...
+                layer_below.soil_water_content += percolated_water
+
+
+
+
+
+
+
+
+
```
morrowcj:
```suggestion
```

morrowcj (changes requested):
This is a great start. I think one more round of revision will greatly improve it.
* All of the static methods look good to me
* Most of my comments involve `percolate`. Look over them and let me know what you think.
* I've added a file to the PR as a supplement to the comments, which should be deleted once you make a decision regarding its contents.

SC_redesign/Crop_and_Soil/soil/infiltration.py
Isgotkowitz: I think `infiltration.py` is only dealing with how much of the water from the environment makes it into the soil, and `percolation.py` deals with how the water moves down once it is in the soil. So we only have to worry about frozen soil beneath the surface layer in `percolation.py`

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: Nice catch

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: Whatever we can do to make the water units less confusing, I am all for it

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: Ok

Isgotkowitz: All units in all files inside `Crop_and_Soil/soil/` have now been made consistent

SC_redesign/Crop_and_Soil/soil/percolation.py
Isgotkowitz: Yes. Are quote comments the ones that are going to be automatically put in the documentation?

Isgotkowitz: I believe all instances of this have now been corrected

SC_redesign/Crop_and_Soil/soil/percolation.py
Isgotkowitz: Ok, will do

SC_redesign/Crop_and_Soil/soil/percolation.py
Isgotkowitz: Whoops

SC_redesign/Crop_and_Soil/soil/percolation.py
Isgotkowitz: I think it would be better to keep the temperature check out of `_determine_if_percolation_allowed()` because that method directly corresponds to logic in SWAT

SC_redesign/Crop_and_Soil/soil/percolation.py
Isgotkowitz: Ok

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: Done

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_percolation.py
```diff
@@ -0,0 +1,221 @@
...
+
+def test_percolate_between_layers():
+    """this function tests _percolate_between_layers() in Percolation.py
+
+    It tests 3 different cases:
+        1: when the amount of water that can be percolated from the upper layer to the lower layer is greater than the
+            amount of water that the lower layer can absorb
+        2: when the amount of water that can be percolated from the upper layer to the lower layer is less than the
+            amount of water that the lower layer can absorb
+        3: when the upper layer does not not have any water that can be percolated to the lower layer
```
morrowcj, line 78: I like this.

morrowcj (approved):

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -17,36 +18,59 @@ class LayerData:
     """nitrate level of the layer (kg/ha)"""
     soil_water_concentration: float = 0.25  # arbitrary
     """soil water concentration of the layer (mm)"""
+    water_content:  Optional[float] = None
+    """volume of soil water in the layer (mm)"""
```
morrowcj:
```suggestion
    """water present in the layer (mm)"""
```

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -17,36 +18,59 @@ class LayerData:
...
-    """water concentration of soil layer at saturation point (mm)"""
+    """water concentration of soil layer at saturation point (mm water / mm soil)"""
     soil_evaporation_compensation_coefficient: float = 1
-    """coefficient that allows user to modify depth distribution used to meet the soil evaporative demand (2:2.3.17)"""
+    """coefficient that allows user to modify depth distribution used to meet the soil evaporative demand (unitless) 
+        (SWAT 2:2.3.17)"""
+
+    # --- Percolation
+    temperature: float = 15.05
+    """temperature of soil layer (degrees Celsius)"""
```
morrowcj:
```suggestion
    """current temperature of this soil layer (degrees Celsius)"""
```

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -17,36 +18,59 @@ class LayerData:
...
+    saturated_hydraulic_conductivity: float = 9.5
+    """saturated hydraulic conductivity for this layer of soil (mm per hour)"""
+    available_water_capacity: float = 0.2
+    """available water capacity expressed as fraction of total soil volume (unitless)"""
```
morrowcj:
```suggestion
    """available water capacity expressed as fraction of total soil volume [for this layer?] (unitless)"""
```

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -17,36 +18,59 @@ class LayerData:
...
+
+    def __post_init__(self):
+        """This function initializes all attributes in the dataclass that depend on other attributes in the class"""
```
morrowcj:
```suggestion
        """Initialize all attributes in the dataclass that depend on other attributes"""
```

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -30,16 +31,35 @@ class SoilData:
...
     """retention parameter for the previous day (mm) (used in SWAT 2:1.1.9)"""
     average_slope_fraction: float = 0.05
-    """average slope fraction of the subbasin"""
+    """average slope fraction of the subbasin (unitless)"""
     moisture_condition_parameter: Optional[float] = None
     """curve number value adjusted for moisture content (unitless) (SWAT 2:1.1.11)"""
 
+    # ---- percolation
+    vadose_zone_layer: Optional[LayerData] = None
+    """Datalayer object that represents the vadose zone, arbitrary top and bottom depths, starts with no water"""
```
```suggestion
    """LaterData object that represents the vadose zone layer, arbitrary top and bottom depths [still true?], starts with no water"""
```

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -30,16 +31,35 @@ class SoilData:
...
+    time_step: float = 24
+    """length of time step over which percolatigit adon occurs (hours) """
```
morrowcj: Is this what you meant?
```suggestion
    """length of time step over which percolation occurs (hours) """
```

morrowcj (commented):
I found a few documentation spots that could be improved while looking at #336.

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: Yeah I don't really understand this attribute either, it gets used for equation 2:3.1.6 but that is only in SWAT, and not the codebase. SWAT says it gets input by the user, but it never gets used in the `Percolation` module so I don't think we even need it.

SC_redesign/Crop_and_Soil/soil/soil_data.py
Isgotkowitz: No, I will update it.

SC_redesign/Crop_and_Soil/soil/soil_data.py
Isgotkowitz: No, I will also update this

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: I deleted it and everything still works, so I guess we did not need it.

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -30,16 +31,36 @@ class SoilData:
...
+    def __post_init__(self):
+        if self.soil_layers is None:
+            # sets the soil layers to a default set if user does not provide any
+            self.soil_layers = [LayerData(top_depth=0, bottom_depth=50, nitrate=0.5),
+                                LayerData(top_depth=50, bottom_depth=80, nitrate=1),
+                                LayerData(top_depth=80, bottom_depth=200, nitrate=5)]
+
+        # configures the vadose zone LayerData object based on where the soil profile ends
+        self.vadose_zone_layer = LayerData(top_depth=self.soil_layers[-1].bottom_depth,
+                                           bottom_depth=10000000,    # bottom depth is 10,000 meters by default
```
matthew7838, line 54: Not sure if this is a mistake, but the number here does not match the 10,000 default value in the commetnts.

matthew7838 (commented):
Except for one part that I figure it would be better for me to mention and double check, everything else looks good. Missing documentations in the test files earlier were fixed.

SC_redesign/Crop_and_Soil/soil/soil_data.py
morrowcj: Soil depth is in mm. 10k meters is 10M mm

matthew7838 (approved):

PR #329, opened 2023-02-03
Refactor decomposition.py

### LINKED ISSUE #334: refactor and test decomp_factors.py (soil/carbon_cycling) into SC_redesign directory
The soil composition method needs refactoring, in line with the rest of SC_redesign.

matthew7838:
### **Summary**
This PR refactors the `Decomposition` sub-module (`decomposition.py`) of the `Soil` module from "pseudocode_soil" S.6.A.1 and "pseudocode_soil" S.6.A.2. All methods except the main routine are static methods.

Every method is tested covering possibly most edge cases.

### **Changes**

-  Except for the main routine `decompose`, all other methods are now static
-  Except for the main routine `decompose`, all methods do not rely on any external classes
-  Descriptive names modified and added
-  Tests added for all methods
-  Now only the main routine `decompose` updates and tracks attributes added in the `SoilData` class

### **Other Changes**

- Added property function in the `LayerData` class to calculate the correct water factor

### DIFF
--- a/SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
+++ b/SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
@@ -1,7 +1,8 @@
 import math
-from typing import Optional
+from typing import Optional, List
 
 from SC_redesign.Crop_and_Soil.soil import soil_data
+from SC_redesign.Crop_and_Soil.soil.layer_data import LayerData
 from SC_redesign.Crop_and_Soil.soil.soil_data import SoilData
 
 
@@ -16,16 +17,14 @@ class Decomposition:
             "pseudocode_soil" S.6.A
         Args:
             temp_average:
-            soil: an instance of the Soil class defined in soil.py
-            weather: an instance of the Weather class defined in classes.py
-            time: an instance of the Time class defined in classes.py
-        """
-        self.data.decomposition_temperature_effect = self.calc_temp_factor(temp_average)
 
-        #self.data.decomposition_moisture_effect = self.moisture_factor(soil)
+        """
+        self.data.decomposition_temperature_effect = self._calc_temp_factor(temp_average)
+
+        self.data.decomposition_moisture_effect = self.moisture_factor(self.data.soil_layers)
 
     @staticmethod
-    def calc_temp_factor(temp_average: float) -> float:
+    def _calc_temp_factor(temp_average: float) -> float:
         """
         Description: calculates the Temperature factor for carbon decomposition
             "pseudocode_soil" S.6.A.1
@@ -44,14 +43,14 @@ class Decomposition:
                    (decomposition_inflection_y + (max_min_distance / math.pi) * math.atan(math.pi * inflection_slope * (
                            temp_average - decomposition_inflection_x))) / normalizer)
 
-    #@staticmethod
-    def calc_moisture_factor(self) -> float:
+    @staticmethod
+    def _calc_moisture_factor(soil_layers: List[LayerData]) -> float:
         """
         Description: calculates the moisture factor for carbon decomposition
             "pseudocode_soil" S.6.A.2
             defaults drawn from defac: course soil
         Args:
-            soil
+            soil_layers:
         """
         dimensionless_empirical_factor_a = 0.55
         dimensionless_empirical_factor_b = 1.7
@@ -59,7 +58,7 @@ class Decomposition:
         dimensionless_empirical_factor_e1 = 6.648115
         dimensionless_empirical_factor_e2 = 3.22
 
-        for layer in self.data.soil_layers:
+        for layer in soil_layers:
             # S.6.A.5
             base_1 = (layer.water_fac - dimensionless_empirical_factor_b) / (dimensionless_empirical_factor_a -
                                                                              dimensionless_empirical_factor_b)
--- a/SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
+++ b/SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
@@ -5,7 +5,10 @@ from SC_redesign.Crop_and_Soil.soil import soil_data
 from SC_redesign.Crop_and_Soil.soil.layer_data import LayerData
 from SC_redesign.Crop_and_Soil.soil.soil_data import SoilData
 
-
+# TODO: the equations for this model, referenced in the soil psuedocode, are derived from the excel file
+#   located on basecamp: https://3.basecamp.com/3486446/buckets/5296287/vaults/2740532358
+#   but the meaning (and validity) of the terms is extremely unclear from either source. The 
+#   documentation cannot be adequately completed without a better understanding of these methods.
 class Decomposition:
     def __init__(self, soil_data: Optional[SoilData] = None):
         self.data = soil_data or SoilData()  # initialize with defaults, if not given
--- a/SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
+++ b/SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
@@ -10,7 +10,7 @@ class Decomposition:
     def __init__(self, soil_data: Optional[SoilData] = None):
         self.data = soil_data or SoilData()  # initialize with defaults, if not given
 
-    def decomposite(self, temp_average: float) -> None:
+    def decompose(self, temp_average: float) -> None:
         """
         Description: calculates temperature and moisture decomposition factors for
             Carbon based on weather and soil profile.
@@ -21,7 +21,7 @@ class Decomposition:
         """
         self.data.decomposition_temperature_effect = self._calc_temp_factor(temp_average)
 
-        self.data.decomposition_moisture_effect = self.moisture_factor(self.data.soil_layers)
+        self.data.decomposition_moisture_effect = self._calc_moisture_factor(self.data.soil_layers)
 
     @staticmethod
     def _calc_temp_factor(temp_average: float) -> float:
--- a/SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
+++ b/SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
@@ -9,17 +9,18 @@ class Decomposition:
     def __init__(self, soil_data: Optional[SoilData] = None):
         self.data = soil_data or SoilData()  # initialize with defaults, if not given
 
-    def decomposite(self, soil, weather, time):
+    def decomposite(self, temp_average: float) -> None:
         """
         Description: calculates temperature and moisture decomposition factors for
             Carbon based on weather and soil profile.
             "pseudocode_soil" S.6.A
         Args:
+            temp_average:
             soil: an instance of the Soil class defined in soil.py
             weather: an instance of the Weather class defined in classes.py
             time: an instance of the Time class defined in classes.py
         """
-        self.data.decomposition_temperature_effect = self.temp_factor(soil, weather, time)
+        self.data.decomposition_temperature_effect = self.temp_factor(temp_average)
 
         self.data.decomposition_moisture_effect = self.moisture_factor(soil)
 
@@ -44,7 +45,7 @@ class Decomposition:
                            temp_average - decomposition_inflection_x))) / normalizer)
 
     @staticmethod
-    def moisture_factor(soil) -> float:
+    def calc_moisture_factor(soil) -> float:
         """
         Description: calculates the moisture factor for carbon decomposition
             "pseudocode_soil" S.6.A.2
--- a/SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
+++ b/SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
@@ -60,10 +60,10 @@ class Decomposition:
 
         for layer in soil_layers:
             # S.6.A.5
-            base_1 = (layer.water_fac - dimensionless_empirical_factor_b) / (dimensionless_empirical_factor_a -
-                                                                             dimensionless_empirical_factor_b)
-            base_2 = (layer.water_fac - dimensionless_empirical_factor_c) / (dimensionless_empirical_factor_a -
-                                                                             dimensionless_empirical_factor_c)
+            base_1 = (layer.water_factor - dimensionless_empirical_factor_b) / (dimensionless_empirical_factor_a -
+                                                                                dimensionless_empirical_factor_b)
+            base_2 = (layer.water_factor - dimensionless_empirical_factor_c) / (dimensionless_empirical_factor_a -
+                                                                                dimensionless_empirical_factor_c)
             hold = (base_1 ** dimensionless_empirical_factor_e1) * (base_2 ** dimensionless_empirical_factor_e2)
 
         return hold
--- a/SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
+++ b/SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
@@ -38,9 +38,9 @@ class Decomposition:
             "pseudocode_soil" S.6.A.1
             defaults drawn from defac: course soil
         Args:
-            temp_average: Average temperature (unitless)
+            temp_average: Average temperature (celsius)
 
-        Returns: temperature effect
+        Returns: temperature effect (unitless)
         """
         # S.6.A.4
         return (y_inflection + (point_distance / math.pi) * math.atan(math.pi * inflection_slope * (
@@ -55,7 +55,7 @@ class Decomposition:
             "pseudocode_soil" S.6.A.2
             defaults drawn from defac: course soil
         Args:
-            soil_layers: Layer data
+            water_factor: relative water saturation (%)
         Returns: moisture effect (unitless)
         """
         # S.6.A.5
--- a/SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
+++ b/SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
@@ -20,9 +20,9 @@ class Decomposition:
             weather: an instance of the Weather class defined in classes.py
             time: an instance of the Time class defined in classes.py
         """
-        self.data.decomposition_temperature_effect = self.temp_factor(temp_average)
+        self.data.decomposition_temperature_effect = self.calc_temp_factor(temp_average)
 
-        self.data.decomposition_moisture_effect = self.moisture_factor(soil)
+        #self.data.decomposition_moisture_effect = self.moisture_factor(soil)
 
     @staticmethod
     def calc_temp_factor(temp_average: float) -> float:
@@ -31,7 +31,7 @@ class Decomposition:
             "pseudocode_soil" S.6.A.1
             defaults drawn from defac: course soil
         Args:
-            temp_average:
+            temp_average: Average temperature
         """
         decomposition_inflection_x = 15.400
         decomposition_inflection_y = 11.750
@@ -44,8 +44,8 @@ class Decomposition:
                    (decomposition_inflection_y + (max_min_distance / math.pi) * math.atan(math.pi * inflection_slope * (
                            temp_average - decomposition_inflection_x))) / normalizer)
 
-    @staticmethod
-    def calc_moisture_factor(soil) -> float:
+    #@staticmethod
+    def calc_moisture_factor(self) -> float:
         """
         Description: calculates the moisture factor for carbon decomposition
             "pseudocode_soil" S.6.A.2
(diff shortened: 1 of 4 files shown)


SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
```diff
@@ -0,0 +1,73 @@
...
+        Args:
+            temp_average: Average temperature
+
+        Returns: Unitless temperature effect
+        """
+        decomposition_inflection_x = 15.400
+        decomposition_inflection_y = 11.750
+        max_min_distance = 29.700
+        inflection_slope = 0.03
+        normalizer = 20.80546
```
morrowcj: Can we move these variables into the signature as parameters? I think you can also simply the names a bit. You did a great job picking descriptive names, but they can be a bit cleaner. For example, "decomposition" is redundant, since this is the decomposition module and these variables don't get used outside it.

New signature example:

```python
def _calc_temp_factor(temperature: float, x_inflection: float = 15.4, y_inflection: float = 11.75, 
                      point_distance: float = 29.7, inflection_slope = 0.03, 
                      normalizer = 20.80546) -> float:
```

Do something similar for the other static method as well and don't forget update the docstrings.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
```diff
@@ -0,0 +1,73 @@
...
+
+        # S.6.A.4
+        return max(0.0,
+                   (decomposition_inflection_y + (max_min_distance / math.pi) * math.atan(math.pi * inflection_slope * (
+                           temp_average - decomposition_inflection_x))) / normalizer)
```
morrowcj: I know it is written this way in the old code, but I don't think this should be limited to be strictly positive.
I can't find anything in the pseudocode or the referenced excel models to suggest that it should be 
positive. Did you? The name "temperature effect on decomposition" implies that it can have a positive or negative effect. 

Otherwise, the math looks right.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
```diff
@@ -0,0 +1,73 @@
...
+
+        for layer in soil_layers:
+            # S.6.A.5
+            base_1 = (layer.water_factor - dimensionless_empirical_factor_b) / (dimensionless_empirical_factor_a -
+                                                                                dimensionless_empirical_factor_b)
+            base_2 = (layer.water_factor - dimensionless_empirical_factor_c) / (dimensionless_empirical_factor_a -
+                                                                                dimensionless_empirical_factor_c)
+            hold = (base_1 ** dimensionless_empirical_factor_e1) * (base_2 ** dimensionless_empirical_factor_e2)
+
+        return hold
```
morrowcj: What is "hold"? Also, there's no need to save this, just directly return:

```suggestion
            return (base_1 ** dimensionless_empirical_factor_e1) * (base_2 ** dimensionless_empirical_factor_e2)
```

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
```diff
@@ -0,0 +1,73 @@
...
+            defaults drawn from defac: course soil
+        Args:
+            soil_layers: Layer data
+        Returns: unitless moisture effect
+        """
+        dimensionless_empirical_factor_a = 0.55
+        dimensionless_empirical_factor_b = 1.7
+        dimensionless_empirical_factor_c = -0.007
+        dimensionless_empirical_factor_e1 = 6.648115
+        dimensionless_empirical_factor_e2 = 3.22
```
morrowcj: just call these `a_term`, `b_term`, `c_term`, `first_exponent`, `second_exponent` for now.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
```diff
@@ -0,0 +1,73 @@
+import math
+from typing import Optional, List
+
+from SC_redesign.Crop_and_Soil.soil import soil_data
+from SC_redesign.Crop_and_Soil.soil.layer_data import LayerData
+from SC_redesign.Crop_and_Soil.soil.soil_data import SoilData
+
+
```
morrowcj, line 5: I'm going to make a commit adding a TODO for future reference.

```suggestion
# TODO: the equations for this model, referenced in the soil psuedocode, are derived from the excel file
#   located on basecamp: https://3.basecamp.com/3486446/buckets/5296287/vaults/2740532358
#   but the meaning (and validity) of the terms is extremely unclear from either source. The 
#   documentation cannot be adequately completed without a better understanding of these methods.
```

morrowcj (approved):
# Summary
This looks good. I've requested a few minor changes, referenced in the review comments below. 
Excellent job, Matthew. Keep up the great work!

# Additional Notes
* Excellent job with the main comment on the pull request. It is very descriptive and easy to follow.
* Signatures look good, type hints are present
* The equations match the source documents
* Docstrings look good, everything is as well documented as it can be given the source material
* I appreciate how thoughtful you were with renaming variables.
* The tests all pass, look like the do what they are supposed to, and consider an adequate range of cases.
  - Note that, after updating the signatures of those methods, you'll need to update your tests too, but you
  only will need to update the lines that actually call the method.

morrowcj (changes requested):
Sorry, I accidentally selected "approve" when I meant to request changes.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
matthew7838: I think it is reasonable to say that it will always be between 0 a 1.2 for two reasons. The first is provided by the graph in the excel file, and the second is that the function in S.6.A.1 involves the inverse of a tangent, therefore I think as temp get's lower it will get infinitely close to 0 and infinitely close to 1.2 as the temperature gets higher.

morrowcj: @matthew7838, the graph for this function ("tfunc" tab in the excel sheet) is for the temperature ranges of -10 to 60 degrees C. Note that values $\leq$ -15 degrees C (5 degrees F) give negative results. These temperatures values occur in much of the midwest, so we can't discount them. Second, those values are based heavily on the parameter values other than temperature. The values that are hard-coded are only one parameterization of the model but we need to plan for other parameterizations as well.

The limits of arctan(x) are $(-\frac{\pi} { 2},~\frac{\pi}{2})$.

So, the only reason this function should have a lower-limit of 0 is if the equation returns the "actual rate of decomposition" and not the "effect of temperature on decomposition." I'm not positive which is true, since there's no documentation in the original model. But I'm inclined to believe it is the latter, since the title of the graph and the pseudocode reflect that.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
```diff
@@ -0,0 +1,76 @@
...
+        inflection_slope = 0.03
+        normalizer = 20.80546
+
+        # S.6.A.4
+        return max(0.0,
+                   (decomposition_inflection_y + (max_min_distance / math.pi) * math.atan(math.pi * inflection_slope * (
+                           temp_average - decomposition_inflection_x))) / normalizer)
+
+    @staticmethod
+    def _calc_moisture_factor(soil_layers: List[LayerData]) -> float:
```
Isgotkowitz: First, I think it would be better to have `_calc_moisture_factor()` only take the `water_factor` attribute instead of the entire `LayerData` object - in general it is good to only pass what is needed to a method.

Second (and this is less important), I prefer to pass one value at a time to methods and make the calling method handle the multiple return values. This would be helpful for someone in the future who wants to model something this method, but doesn't need to use it on a whole list of attributes.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
```diff
@@ -0,0 +1,76 @@
...
+        dimensionless_empirical_factor_e1 = 6.648115
+        dimensionless_empirical_factor_e2 = 3.22
+
+        for layer in soil_layers:
+            # S.6.A.5
+            base_1 = (layer.water_factor - dimensionless_empirical_factor_b) / (dimensionless_empirical_factor_a -
+                                                                                dimensionless_empirical_factor_b)
+            base_2 = (layer.water_factor - dimensionless_empirical_factor_c) / (dimensionless_empirical_factor_a -
+                                                                                dimensionless_empirical_factor_c)
+            hold = (base_1 ** dimensionless_empirical_factor_e1) * (base_2 ** dimensionless_empirical_factor_e2)
```
Isgotkowitz: What is this method supposed to be doing? Because currently, it iterates through `soil_layers`, calculates stuff based on the current layer, then moves on to the next without doing anything with it. So the only value that gets returned is the `hold` of the bottom layer

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
```diff
@@ -0,0 +1,76 @@
...
+
+    @staticmethod
+    def _calc_moisture_factor(soil_layers: List[LayerData]) -> float:
+        """
+        Description: calculates the moisture factor for carbon decomposition
+            "pseudocode_soil" S.6.A.2
+            defaults drawn from defac: course soil
+        Args:
+            soil_layers: Layer data
+        Returns: unitless moisture effect
```
Isgotkowitz:
```suggestion
        Returns: moisture effect (unitless)
```
I know it is a pain but it is good to keep the style of documentation as consistent as possible across the codebase

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
```diff
@@ -0,0 +1,76 @@
...
+        self.data.decomposition_moisture_effect = self._calc_moisture_factor(self.data.soil_layers)
+
+    @staticmethod
+    def _calc_temp_factor(temp_average: float) -> float:
+        """
+        Description: calculates the Temperature factor for carbon decomposition
+            "pseudocode_soil" S.6.A.1
+            defaults drawn from defac: course soil
+        Args:
+            temp_average: Average temperature
```
Isgotkowitz: Needs units here

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_decomposition.py
```diff
@@ -0,0 +1,92 @@
...
+               wilting_point_water_concentration=1.8),
+     LayerData(top_depth=8, bottom_depth=20, soil_water_concentration=1.4, field_capacity_water_concentration=1.8,
+               wilting_point_water_concentration=0.8),
+     LayerData(top_depth=20, bottom_depth=22, soil_water_concentration=0.8, field_capacity_water_concentration=1,
+               wilting_point_water_concentration=0.6)],
+])
+def test_calc_temp_factor(layers):
+    """ensures that moisture effect was calculated according to the formula in "pseudocode_soil" S.6.A.2"""
+    data = SoilData(soil_layers=layers)
+    decomp = Decomposition(data)
```
Isgotkowitz: These objects are not necessary for an `@staticmethod` test

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_decomposition.py
```diff
@@ -0,0 +1,92 @@
...
+
+    for layer in layers:
+        # S.6.A.5
+        base_1 = (layer.water_factor - dimensionless_empirical_factor_b) / (dimensionless_empirical_factor_a -
+                                                                            dimensionless_empirical_factor_b)
+        base_2 = (layer.water_factor - dimensionless_empirical_factor_c) / (dimensionless_empirical_factor_a -
+                                                                            dimensionless_empirical_factor_c)
+        hold = (base_1 ** dimensionless_empirical_factor_e1) * (base_2 ** dimensionless_empirical_factor_e2)
+    expect = hold
+    assert decomp._calc_moisture_factor(layers) == hold
```
Isgotkowitz:
```suggestion
    assert Decomposition._calc_moisture_factor(layers) == hold
```

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -34,6 +34,11 @@ class SoilData:
     moisture_condition_parameter: Optional[float] = None
     """curve number value adjusted for moisture content (unitless) (SWAT 2:1.1.11)"""
 
+    # ---- decomposition
+    decomposition_temperature_effect: Optional[float] = None
+    decomposition_moisture_effect: Optional[float] = None
```
Isgotkowitz: These should both have docstrings beneath them

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -50,3 +50,14 @@ def wilting_point_content(self):
     def saturation_content(self):
         """volume of water in layer when saturated in mm"""
         return self.saturation_point_water_concentration / self.layer_thickness
+
+    @property
+    def water_factor(self):
+        """volume of water in layer when saturated in mm"""
```
Isgotkowitz: I don't think that is what's being calculated here

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -50,3 +50,14 @@ def wilting_point_content(self):
...
+        if self.soil_water_content <= self.field_capacity_content:
+            return (self.soil_water_content - self.wilting_point_content)/(
+                    self.field_capacity_content - self.wilting_point_content)
+        else:
+            return (self.saturation_content - self.soil_water_content) / (
+                        self.saturation_content - self.field_capacity_content)
```
Isgotkowitz: If this comes from an equation in the documentation, there should be a reference to it

Isgotkowitz (changes requested):
Good first PR. There are a couple of formatting issues in the docstrings that need to corrected, and some changes need to be made in `_calc_moisture_factor()`, but other than that looks good.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
matthew7838: @morrowcj Not sure what this method is supposed to be doing.

morrowcj: @matthew7838, the old code calculates the moisture for each layer and saves the output as an attribute **of that layer's object**. 

You'll ultimately want to do the same thing, but break the functionality down:
* first make the `_calc_moisture_factor()` method act on a single layer.
* then, from the main method, apply it to *each layer*, saving the results within the layer. 

Here's a hint (simplified):

```python
class Decomposition:
    def decompose(self, temp_average): -> None:
        for layer in self.data.soil_layers:
            layer.moisture_decomposition_factor = self._calculate_moisture_factor(layer.water_factor)
    
    def _calculate_moisture_factor(water_factor, a=0.55, b=1.7, c=-0.007, e1=6.65, e2=3.22) -> float:
        return ((water_factor - b) / (a - b))**e1 * ((water_factor - c) / (a - c))**e2
```

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
```diff
@@ -0,0 +1,65 @@
+import math
+from typing import Optional, List
+
+from SC_redesign.Crop_and_Soil.soil import soil_data
```
Isgotkowitz: Is this import statement necessary?

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
```diff
@@ -0,0 +1,65 @@
...
+class Decomposition:
+    def __init__(self, soil_data: Optional[SoilData] = None):
+        self.data = soil_data or SoilData()  # initialize with defaults, if not given
+
+    def decompose(self, temp_average: float) -> None:
+        """
+        Description: Updates and tracks all the related attributes in SoilData class
+
+        Args:
+            temp_average: Average temperature to pass in the _calc_temp_factor(temp_average) function
```
Isgotkowitz: In docstrings for main routines I don't think it is necessary to say what it is used for, just what it is. Also it should have units

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
```diff
@@ -0,0 +1,65 @@
...
+        Returns: None
+
+        """
+        self.data.decomposition_temperature_effect = self._calc_temp_factor(temp_average)
+
+        for layer in self.data.soil_layers:
+            layer.decomposition_moisture_effect = self._calc_moisture_factor(layer.water_factor)
+
+    @staticmethod
+    def _calc_temp_factor(temp_average, x_inflection: float = 15.4, y_inflection: float = 11.75,
```
Isgotkowitz, line 30: Why are these being declared as parameters? If they are the same for every layer of soil and work for all soil profiles then they should be hardcoded into the method. If they are features of a soil profile/layer and a user may want to modify them based on the data they are working with, they should be attributes in either `SoilData` or `LayerData` (set to the values used here as defaults) and passed to the main routine as parameters

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
```diff
@@ -0,0 +1,65 @@
...
+            temp_average: Average temperature (celsius)
+
+        Returns: temperature effect (unitless)
+        """
+        # S.6.A.4
+        return (y_inflection + (point_distance / math.pi) * math.atan(math.pi * inflection_slope * (
+                temp_average - x_inflection))) / normalizer
+
+    @staticmethod
+    def _calc_moisture_factor(water_factor, a_term: float = 0.55, b_term: float = 1.7,
```
Isgotkowitz, line 47: Same as above comment, these should not just be declared as parameters in the method signature with no way for a user to modify them

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -25,6 +25,8 @@ class LayerData:
     """water concentration of soil layer at saturation point (mm)"""
     soil_evaporation_compensation_coefficient: float = 1
     """coefficient that allows user to modify depth distribution used to meet the soil evaporative demand (2:2.3.17)"""
+    decomposition_moisture_effect: Optional[float] = None
+    """moisture effect on decomposition factor (unitless) (pseudocode_soil S.6.A.2)"""
```
Isgotkowitz, line 63: It is good style to group attributes used by the same module together and label them
```suggestion
    
    # --- Decomposition
    decomposition_moisture_effect: Optional[float] = None
    """moisture effect on decomposition factor (unitless) (pseudocode_soil S.6.A.2)"""
```
This is a pretty old version of `LayerData` so if you want to see more examples you should check the `SC_redesign` branch

Isgotkowitz (changes requested):
Good stuff, getting very close to being finished. My comments are on how default values should be handled by methods, and some minor things in docstrings

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
matthew7838: These were changes requested by Clay in previous requested changes, they were orginially hardcoded since I don't think they're suppose to be modified.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
matthew7838: same as above

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
Isgotkowitz: Ok, I just asked Clay about it too. I think they are good for right now, maybe make an issue in the Soil and Crop Redesign project and mark it as a Future Issue so we don't forget about it?

Isgotkowitz (approved):
Nice

SC_redesign/Crop_and_Soil/soil/carbon_cycling/decomposition.py
morrowcj: To follow up (@Isgotkowitz, @matthew7838), this is because this is just a single parameterize of the model (these values were selected because they best fit a specific dataset). Rather than being hard-coded, these methods should provide an option for alternative parameterizations of the model.

As an example, the moisture function `calc_moisture_factor` has three parameterizations in the excel file for different grades of soil:
image

morrowcj (approved):
Looks good to me. Thank you for your hard work, Matthew! You are free to merge.

PR #330, opened 2023-02-03
reworked life_cycle reporting variables to OM

ew3361zh:
**What**
Reworking the `life_cycle_manager` daily update reporting to the OM to be more complete and hopefully easier to track because they are all added at once, once per day. Readding some previous variables removed from OM pool.

**Why**
Looking through the current herd_report from the output_handler, there are a number of variables that hadn't been added to the output_manager in part because their names change multiple times throughout the model but I believe this captures everything in the herd_report now.

**How**
Looking through the list of variables in the output_handler herd_report and trying to match them up with attributes in the LifeCycleManager class.

**Notes**
- One addition to general .gitignore file to prevent output generated in manure module from being added to github.


.gitignore
```diff
@@ -24,6 +24,7 @@ others
 # only stores non-empty directories
 output/*
 !output/.keep
+RUFAS/routines/manure/output
```
JoeWaddell, line 27: Great addition: been meaning to bring this up.

JoeWaddell (approved):
This works on my end, and searching in the variables output for given keywords looks to correctly collect those data. Nice work!

ew3361zh:
At Loi's suggestion, I am closing out this PR which branched off master and instead migrated the changes to PR #337 which is essentially the same but branches off the animal_life_cycle_refactor_pr branch, hopefully preventing merge conflicts down the road.

ltp23 (approved):
Thanks, Niko. This looks good to me.

PR #300, opened 2023-01-24
General mills morrowcj

morrowcj:
## Summary
This PR is meant to incorporate some fixes that @rhensen made in the `general_mills_crop` branch. It appears that this branch fixes some of the major errors that were present in the old crop and soil module. 
* It appears that yields are much more consistent and inlign with expectation for the `SWAT.json` input file
* It also appears the bug involving nitrogen uptake in the soil has been resolved (or at least improved).

I wanted these changes to be present in the `SC_redesign` branch so that these fixes can be referenced during the course of the Soil and Crop refactoring effort.

PR #331, opened 2023-02-03
fixed failing tests in old SC module

morrowcj:
Updating the current SC module with Reid's code (PR #300) broke a few tests that I forgot to fix. This PR implements those fixes.

All of the fixes were very simple changes to reflect the updated functions they were designed to test (e.g., mocking the `BaseCrop` class instead of `Crop`, adding bounds to test expectation, etc.)

### DIFF
--- a/RUFAS/routines/field/crop/leaf_area_index.py
+++ b/RUFAS/routines/field/crop/leaf_area_index.py
@@ -147,4 +147,4 @@ def calculate_d_LAI_actual(crop_type, d_LAI_max):
         float: change in LAI actual
     """
     #print('d_LAI_max',d_LAI_max,'crop_type.gamma_reg',crop_type.gamma_reg)
-    return min(d_LAI_max * sqrt(crop_type.gamma_reg),0.01)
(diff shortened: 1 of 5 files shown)


morrowcj:
@ew3361zh I just wanted to let you know that this has been resolved. If you're cool with it, I'll just go ahead and merge this once you've had a chance to look at it. I don't think it warrants another review, since it does not concern `master` nor will these files be permanent.

ew3361zh (approved):
This looks good, Clay. Every test now passing for me locally within this branch.

One longer-term thing I'd recommend (i.e. not necessary to do on this PR) is to change the star imports in all these test files to the actual names of the modules/classes needed.

[Flake8 doesn't like the star import because it's harder to tell where something is being imported from, if it's actually imported, and whether you've unintentionally imported something else that might have a naming conflict with another import.](https://www.flake8rules.com/rules/F405.html)

morrowcj:
> One longer-term thing I'd recommend (i.e. not necessary to do on this PR) is to change the star imports in all these test files to the actual names of the modules/classes needed.

I completely agree. I believe we've gotten rid of all star imports in the new module.

PR #333, opened 2023-02-07
Manure field connector design

morrowcj:
This PR is meant to track the connection between the manure module and the soil and crop module. 

@ltp23 and @varrma, rather than send you a list, I decided to make a markdown file (`field-manure-connection.md`) to track the variables. Please review this PR to let me know if the proposed variable list makes sense and is doable.


RUFAS/routines/manure/field-manure-connection.md
```diff
@@ -0,0 +1,83 @@
...
+Below are the variables that we anticipate the modules will need access to from the other module.
+
+#### Manure variables used by SC module
+
+The manure will need the following variables. The units don't matter for now (but should be metric), 
+just be clear on what units you are reporting in, and we can handle conversion later, if necessary.
+* Manure amount:
+  - mass
+  - volume (necessary for liquid) and/or density (nice to have for solid)
+  - percent solid (for slurry; >= 20% is considered solid manure)
```
morrowcj:
```suggestion
  - percent solid (for slurry; >= 80% is considered solid manure)
```

varrma (approved):
Looks good to me, includes all our Initial discussion.

RUFAS/routines/manure/field_manure_manager/field_manure_connector.py
```diff
@@ -0,0 +1,351 @@
...
+    """Amount of nitrogen in the sludge portion of the manure, kg."""
+
+    sludge_manure_phosphorus: float = 0.0
+    """Amount of phosphorus in the sludge portion of the manure, kg."""
+
+    sludge_manure_potassium: float = 0.0
+    """Amount of potassium in the sludge portion of the manure, kg."""
+
+    sludge_manure_volume: float = 0.0
+    """Total volume of sludge manure, m^3."""
```
Isgotkowitz, line 56: Instead of separating `sludge` and `liquid` in these, I think we should just have one attribute for each characteristic, and add a dry matter fraction attribute which indicates how much of the manure application is dry matter (by mass)

RUFAS/routines/manure/field_manure_manager/field_manure_connector.py
```diff
@@ -0,0 +1,351 @@
...
+
+@dataclass(frozen=True)
+class FieldManureData(FieldManureDataLiquidPortion, FieldManureDataSludgePortion):
+    """Class to hold data for field manure."""
+
+    liquid_manure_nitrogen: float = 0.0
+    """Amount of nitrogen in the liquid portion of the manure, kg."""
+
+    liquid_manure_phosphorus: float = 0.0
+    """Amount of phosphorus in the liquid portion of the manure, kg."""
```
Isgotkowitz, line 38: It is possible to add manure to a field only knowing how much phosphorus is in it in total, but sometime in the future we may want to add characteristics like what fraction of that manure is water-extractable inorganic phosphorus and what the source animal is.

RUFAS/routines/manure/field_manure_manager/field_manure_connector.py
```diff
@@ -0,0 +1,351 @@
...
+
+
+class FieldManureDataSludgePortion(FieldManureDataSludgeNutrients):
+    sludge_manure_volume: float
+
+
+@dataclass(frozen=True)
+class FieldManureData(FieldManureDataLiquidPortion, FieldManureDataSludgePortion):
+    """Class to hold data for field manure."""
+
```
Isgotkowitz, line 33: There needs to be some sort of `field_coverage` attribute that indicates the percentage of the field that is covered by this manure application.

RUFAS/routines/manure/field_manure_manager/field_manure_connector.py
```diff
@@ -0,0 +1,351 @@
...
+    sludge_manure_potassium: float
+
+
+class FieldManureDataSludgePortion(FieldManureDataSludgeNutrients):
+    sludge_manure_volume: float
+
+
+@dataclass(frozen=True)
+class FieldManureData(FieldManureDataLiquidPortion, FieldManureDataSludgePortion):
+    """Class to hold data for field manure."""
```
Isgotkowitz, line 32: This may not be the appropriate class/file/module, but somewhere we will need to set up a way to take manure from animals grazing directly on a field and put it into the grazing manure field pools.

Isgotkowitz (commented):
I think we should have a meeting to discuss the API of `ManureApplication` in the Soil and Crop module, and how to best connect that with the `Manure` module.

RUFAS/routines/manure/field_manure_manager/field_manure_connector.py
morrowcj: This is a post v1 feature.

RUFAS/routines/manure/field_manure_manager/field_manure_connector.py
morrowcj: That is provided by the user, not the manure module.

RUFAS/routines/manure/field_manure_manager/field_manure_connector.py
Isgotkowitz: Cool.

morrowcj: But I think it is a good idea to consider that we **will** need it.

RUFAS/routines/manure/field_manure_manager/field_manure_connector.py
KFosterReed: This would be a good question for Peter.

RUFAS/routines/manure/field_manure_manager/field_manure_connector.py
KFosterReed: I think the important thing is how these get incorporated into the different soil N, P, and K fractions. Would the liquid manure P be put in a different soil P pool?

KFosterReed:
Hi Team, I think this is a great start! In particular, I like the idea of this connector being stateless and the field model not being aware of the amount of manure in the manure module. I also like the clarity in all of the class functions. 

What is missing for me is how an exchange of manure information from the manure module to the crop module is initiated and executed. In particular, what information is passed from the crop module to the manure module. The information required to pass from the manure module to the crop module is very clear. 

I think the best representation of reality would be for the crop and soil module to request a certain amount of N, P, and K from the manure module. Based on the composition of the manure in storage the manure module will pass the most amount of manure without exceeding the P request. 

Example: The field manager requests: 300 kg of N, 500 kg of P and 100 kg of K. The current manure N:P:K ratio is 1:1:1 at 1 % of the manure (never the case but for ease of the example). The manure module can pass 50,000 kg of manure to the field connector delivering 500 kg of N, 500 kg of P, and 500 kg of K. The field manager now has to apply an additional 200 kg of N fertilizer to meet the demand.

Why not limit for N & K too? As far as I am aware (and I could be wrong!) nutrient management plans do not restrict the amount of manure that can be applied based on N or K. But this is only based on my very limited knowledge. Also because dairy soils tend to have a lot of P build up. 

A few other comments: 
- In my opinion, we don't need to distinguish between cow and non-cow manure. as far as the soil is concerned, if the manure has the same composition, it will not make a difference if it is cow or horse manure. 
- In the manure composition it requests only total N. Is that all the soil N model needs?

PR #336, opened 2023-02-08
Refactor and test `soil_temp.py`

### LINKED ISSUE #335: refactor and test soil_temp.py (soil) into SC_redesign directory

Isgotkowitz:
# Summary
This PR refactors the `SoilTemp` module (`soil_temp.py`) in the `Soil` module based on section 1:1.3.3 of SWAT. It contains only `@staticmethods` except for `daily_soil_temperature_update()`, which is the main routine of this module. 

Every method is tested, and in the test for `daily_soil_temperature_update()` the method is run twice, to test that the method works initially (no previous day temperatures present) and for all calls after the first (when there is a temperature for the previous day recorded)
# Note
This PR should only be merged *AFTER* #328. All changes can be ignored in 
- `evapotranspiration.py`
- `infiltration.py` and `test_infiltration.py`
- `percolation.py` and `test_percolation.py`
# Changes
All of the below files have changes from both this PR and #328. For all changes that should be ignored, it should be assumed that they will be reviewed for #328
- `soil_data.py`: the only changes that need to be reviewed for this PR are the addition of the `@property` `profile_bulk_density()` method. All other changes should be ignored.
- `soil.py` has had an import statement and attribute added for `SoilTemp`. All other changes should be ignored.
- `layer_data.py` has had two attributes added, both under the comment `# ---- Temperature`. All other changes should be ignored.

### DIFF
--- a/SC_redesign/Crop_and_Soil/soil/layer_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/layer_data.py
@@ -32,7 +32,7 @@ class LayerData:
 
     # --- Percolation
     temperature: float = 15.05
-    """temperature of soil layer (degrees Celsius)"""
+    """temperature of soil layer (degrees C)"""
     saturated_hydraulic_conductivity: float = 9.5
     """saturated hydraulic conductivity for this layer of soil (mm per hour)"""
     available_water_capacity: float = 0.2
@@ -41,6 +41,8 @@ class LayerData:
     # --- Temperature
     bulk_density: float = 1.4
     """bulk density of the soil layer (Mg per cubic meter)"""
+    previous_day_temperature: Optional[float] = None
+    """temperature of soil layer on the previous day (degrees C)"""
 
     def __post_init__(self):
         """This function initializes all attributes in the dataclass that depend on"""
@@ -50,6 +52,11 @@ class LayerData:
     def layer_thickness(self) -> float:
         """thickness of soil layer in mm"""
         return self.bottom_depth - self.top_depth
+
+    @property
+    def depth_of_layer_center(self) -> float:
+        """depth beneath the surface of the center this layer (mm)"""
+        return self.top_depth + (self.layer_thickness / 2)
 
     @property
     def field_capacity_content(self) -> float:
@@ -62,7 +69,7 @@ class LayerData:
         return self.wilting_point_water_concentration * self.layer_thickness
 
     @property
-    def excess_water_available(self):
+    def excess_water_available(self) -> float:
         """volume of water available for percolation in the soil layer in mm
 
         SWAT Reference: 2:3.2.1, 2
@@ -76,5 +83,5 @@ class LayerData:
 
     @property
     def acceptable_percolation_amount(self) -> float:
-        """volume of water that can be accepted by layer before reaching saturation (in mm)"""
+        """volume of water that can be accepted by layer before reaching saturation (mm)"""
         return max(0, self.saturation_content - self.soil_water_content)
--- a/SC_redesign/Crop_and_Soil/soil/layer_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/layer_data.py
@@ -40,7 +40,8 @@ class LayerData:
 
     # --- Temperature
     bulk_density: float = 1.4
-    """bulk density of the soil layer (Mg per cubic meter)"""
+    """bulk density of the soil layer (Mg per cubic meter) (provided by user, but SWAT 2:3.1.1 has an equation for 
+        calculating this field as well)"""
     previous_day_temperature: Optional[float] = None
     """temperature of soil layer on the previous day (degrees C)"""
 
--- a/SC_redesign/Crop_and_Soil/soil/layer_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/layer_data.py
@@ -38,6 +38,10 @@ class LayerData:
     available_water_capacity: float = 0.2
     """available water capacity expressed as fraction of total soil volume"""
 
+    # --- Temperature
+    bulk_density: float = 1.4
+    """bulk density of the soil layer (Mg per cubic meter)"""
+
     def __post_init__(self):
         """This function initializes all attributes in the dataclass that depend on"""
         self.soil_water_content = self.soil_water_concentration * self.layer_thickness
--- a/SC_redesign/Crop_and_Soil/soil/soil_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/soil_data.py
@@ -41,6 +41,15 @@ class SoilData:
     time_step: float = 24
     """length of time step over which percolation occurs (hours) """
 
+    # ---- temperature
+    albedo: float = 0.16
+    """ratio of reflected to solar radiation (unitless)"""
+    # TODO: check above definition with a SME because all I did to get it was google 'soil albedo'
+    previous_temperature_effect: float = 0.8
+    """variable that controls the influence of previous day's temperature on current day's temperature, range is from 0
+    to 1, inclusive. SWAT sets the lag coefficient to 0.8 (paragraph between equations 1:1.3.3, 4) (unitless)
+    """
+
     def __post_init__(self):
         if self.soil_layers is None:
             # sets the soil layers to a default set if user does not provide any
@@ -53,15 +62,6 @@ class SoilData:
                                            bottom_depth=10000000,    # bottom depth is 10,000 meters by default
                                            soil_water_concentration=0,
                                            saturation_point_water_concentration=inf)
-
-    # ---- temperature
-    albedo: float = 0.16
-    """ratio of reflected to solar radiation (unitless)"""
-    # TODO: check above definition with a SME because all I did to get it was google 'soil albedo'
-    lag_coefficient: float = 0.8
-    """variable that controls the influence of previous day's temperature on current day's temperature, range is from 0
-    to 1, inclusive. SWAT sets the lag coefficient to 0.8 (paragraph between equations 1:1.3.3, 4) (unitless)
-    """
 
     @property
     def profile_soil_water_content(self) -> float:
--- a/SC_redesign/Crop_and_Soil/soil/soil_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/soil_data.py
@@ -41,6 +41,15 @@ class SoilData:
     time_step: float = 24
     """length of time step over which percolation occurs (hours) """
 
+    # ---- temperature
+    albedo: float = 0.16
+    """ratio of reflected to solar radiation (unitless)"""
+    # TODO: check above definition with a SME because all I did to get it was google 'soil albedo'
+    lag_coefficient: float = 0.8
+    """variable that controls the influence of previous day's temperature on current day's temperature, range is from 0
+    to 1, inclusive. SWAT sets the lag coefficient to 0.8 (paragraph between equations 1:1.3.3, 4) (unitless)
+    """
+
     @property
     def profile_soil_water_content(self) -> float:
         """
--- a/SC_redesign/Crop_and_Soil/soil/soil_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/soil_data.py
@@ -108,3 +108,18 @@ class SoilData:
     #   - soil evaporation adjusted is always less than or equal to maximum soil evaporation
     #   - potential evapotranspiration adjusted is always less than or equal to maximum evapotranspiration
     #   - average slope fraction is always in the range 0 to 1 inclusive (I think)
+
+    @property
+    def profile_bulk_density(self):
+        """
+
+        Returns: average bulk density of the soil profile based on the bulk density of each soil layer and weighted by
+            the thickness of that layer
+
+        """
+        weighted_densities_sum = 0
+        weights_sum = 0
+        for layer in self.soil_layers:
+            weighted_densities_sum += (layer.layer_thickness * layer.bulk_density)
+            weights_sum += layer.layer_thickness
+        return weighted_densities_sum / weights_sum
--- a/SC_redesign/Crop_and_Soil/soil/soil_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/soil_data.py
@@ -39,7 +39,7 @@ class SoilData:
     vadose_zone_layer: Optional[LayerData] = None
     """Datalayer object that represents the vadose zone, arbitrary top and bottom depths, starts with no water"""
     time_step: float = 24
-    """length of time step over which percolatigit adon occurs (hours) """
+    """length of time step over which percolation occurs (hours) """
 
     def __post_init__(self):
         if self.soil_layers is None:
--- a/SC_redesign/Crop_and_Soil/soil/soil_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/soil_data.py
@@ -56,8 +56,7 @@ class SoilData:
 
     # ---- temperature
     albedo: float = 0.16
-    """ratio of reflected to solar radiation (unitless)"""
-    # TODO: check above definition with a SME because all I did to get it was google 'soil albedo'
+    """proportion of solar radiation that is reflected by the soil surface (unitless)"""
     lag_coefficient: float = 0.8
     """variable that controls the influence of previous day's temperature on current day's temperature, range is from 0
     to 1, inclusive. SWAT sets the lag coefficient to 0.8 (paragraph between equations 1:1.3.3, 4) (unitless)
@@ -103,7 +102,7 @@ class SoilData:
     def profile_field_capacity(self) -> float:
         """
 
-        Returns: total amount of water contained in the soil profile at field capacity (but not saturated) (mm)
+        Returns: total amount of water contained in the entire soil profile at field capacity (but not saturated) (mm)
 
         """
         if self.soil_layers is None:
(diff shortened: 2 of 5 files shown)


matthew7838 (approved):
Looks pretty good to me, caught another missing comment in the tests, but other than that seems good. The reference is clear enough for me to double-check the calculations.

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -17,36 +18,70 @@ class LayerData:
...
+    def layer_thickness(self) -> float:
+        """thickness of soil layer (mm)"""
         return self.bottom_depth - self.top_depth
 
     @property
-    def soil_water_content(self):
-        """volume of soil water in the layer in mm"""
-        return self.soil_water_concentration / self.layer_thickness
+    def depth_of_layer_center(self) -> float:
+        """depth beneath the surface of the center this layer (mm)"""
```
morrowcj, line 57:
```suggestion
        """depth beneath the surface to the center of this layer (mm)"""
```

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -30,16 +31,44 @@ class SoilData:
...
+        # configures the vadose zone LayerData object based on where the soil profile ends
+        self.vadose_zone_layer = LayerData(top_depth=self.soil_layers[-1].bottom_depth,
+                                           bottom_depth=10000000,    # bottom depth is 10,000 meters by default
+                                           soil_water_concentration=0,
+                                           saturation_point_water_concentration=inf)
+
+    # ---- temperature
+    albedo: float = 0.16
+    """ratio of reflected to solar radiation (unitless)"""
+    # TODO: check above definition with a SME because all I did to get it was google 'soil albedo'
```
morrowcj: This is correct, but I'd reword:

```suggestion
    albedo: float = 0.16
    """proportion of solar radiation that is reflected by the soil surface (unitless)"""
```

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -30,16 +31,44 @@ class SoilData:
...
+    lag_coefficient: float = 0.8
+    """variable that controls the influence of previous day's temperature on current day's temperature, range is from 0
+    to 1, inclusive. SWAT sets the lag coefficient to 0.8 (paragraph between equations 1:1.3.3, 4) (unitless)
```
morrowcj: I know this is what SWAT calls this, but from a temporal modelling point-of-view, this is **not** a lag term. It is more closely related to an auto-regression parameter. Specifically, it is a proportional scalar that "determines the **_realtive_** magnitude of the effect of the previous day's temperature on the current day's temperature". `previous_temperature_effect` would be a better name.

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -74,7 +103,7 @@ def profile_saturation(self) -> float:
     def profile_field_capacity(self) -> float:
         """
 
-        Returns: total amount of water contained in the soil profile at field capacity (but not saturated) in mm
+        Returns: total amount of water contained in the soil profile at field capacity (but not saturated) (mm)
```
morrowcj: for emphasis:
```suggestion
        Returns: total amount of water contained in the entire soil profile at field capacity (but not saturated) (mm)
```

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -101,3 +130,18 @@ def soil_water_factor(self):
...
     #   - average slope fraction is always in the range 0 to 1 inclusive (I think)
+
+    @property
+    def profile_bulk_density(self):
+        """
+
+        Returns: average bulk density of the soil profile based on the bulk density of each soil layer and weighted by
+            the thickness of that layer
+
+        """
```
morrowcj: You can document `@properties` as though they are regular attributes if it makes sense to do so.

```suggestion
        """average bulk density of the soil profile based on the bulk density of each soil layer, weighted by
            the thickness
        """
```

SC_redesign/Crop_and_Soil/soil/soil_temp.py
```diff
@@ -0,0 +1,544 @@
...
+#
+# Soil attribute definitions
+#
+#     T_soil = soil temperature (ºC) at depth z (mm)
+#
+#     L = lag coefficient, set to 0.8
+#
+#     T_soil_prev_day = soil temperature (ºC) at depth z (mm) on the previous day
+#
+#     df = depth factor
```
morrowcj: Delete all of this commented stuff (through line 322), it is stale.

SC_redesign/Crop_and_Soil/soil/soil_temp.py
```diff
@@ -0,0 +1,544 @@
...
+
+from SC_redesign.Crop_and_Soil.soil.soil_data import SoilData
+
+"""
+This module is based on the "Soil Temperature" section of SWAT (1:1.3.3)
+"""
+class SoilTemp:
+
+    def __init__(self, soil_data: Optional[SoilData] = None):
+        self.data = soil_data or SoilData()
```
morrowcj: docstring for `soil_data`

SC_redesign/Crop_and_Soil/soil/soil_temp.py
```diff
@@ -0,0 +1,544 @@
...
+        """this is the main routine that updates the soil temperature
+
+        Args:
+            solar_radiation: solar radiation reaching the ground on a given day (MJ per square meter per day)
+            avg_temp: average temperature of a given day (degrees C)
+            min_temp: minimum temperature of a given day (degrees C)
+            max_temp: maximum temperature of a given day (degrees C)
+            plant_cover: total aboveground plant biomass and residue on a given day (kg per hectare)
+            snow_cover: water content of the snow cover on a given day (mm)
+            avg_annual_air_temp: average annual air temperature (degrees C)
```
morrowcj: from the perspective of the simulation:

```suggestion
            solar_radiation: solar radiation reaching the ground on the current day (MJ per square meter per day)
            avg_temp: average temperature of the current day (degrees C)
            min_temp: minimum temperature of the current day (degrees C)
            max_temp: maximum temperature of the current day (degrees C)
            plant_cover: total aboveground plant biomass and residue on the current day (kg per hectare)
            snow_cover: water content of the snow cover on the current day (mm)
            avg_annual_air_temp: average annual air temperature (degrees C)
```

morrowcj: Related: is `avg_annual_air_temp` the average for the entire calendar year, the current calendar year up to the current day (I hope not), or for the last 365 days?

This variable worries me for when we want the model to forecast into the future (not for v1).

SC_redesign/Crop_and_Soil/soil/soil_temp.py
```diff
@@ -0,0 +1,544 @@
...
+
+        Important: SWAT does not specify how to start the simulation i.e. it does not specify what to do on day 0, when
+            there is no previous day's temperature. Currently, the implementation just uses the temperature that the
+            soil starts (it sets the previous days temperature equal to the current day's temperature)
```
morrowcj: Further elaborating:

```suggestion
        Important: SWAT does not specify how to start the simulation i.e. it does not specify what to do on day 0, when
            there is no previous day's temperature. Currently, the implementation just uses the temperature that the
            soil starts (it sets the previous days temperature equal to the current day's temperature). This assumption
            is fairly reasonable due to temporal auto-correlation, but does not account for the random fluctuations
            that can occur throughout the year.
```

SC_redesign/Crop_and_Soil/soil/soil_temp.py
```diff
@@ -0,0 +1,544 @@
...
+                                                                             layer.previous_day_temperature,
+                                                                             layer_depth_factor,
+                                                                             avg_annual_air_temp,
+                                                                             actual_soil_surface_temp)
+            else:
+                layer.temperature = self._determine_average_soil_temperature(self.data.lag_coefficient,
+                                                                             layer.temperature,
+                                                                             layer_depth_factor,
+                                                                             avg_annual_air_temp,
+                                                                             actual_soil_surface_temp)
```
morrowcj: Simplify?

```python
            if layer.previous_day_temperature is None:
                layer.previous_day_temperature = layer.temperature
            # Alternatively, store in variable and use conditional properties of None:
            # prev_temp = layer.previous_day_temperature or layer.temperature

            layer.temperature = self._determine_average_soil_temperature(self.data.lag_coefficient,
                                                                         layer.previous_day_temperature,  # or prev_temp
                                                                         layer_depth_factor,
                                                                         avg_annual_air_temp,
                                                                         actual_soil_surface_temp)
```

SC_redesign/Crop_and_Soil/soil/soil_temp.py
```diff
@@ -0,0 +1,544 @@
...
+    def _determine_maximum_damping_depth(bulk_density: float) -> float:
+        """calculates maximum damping depth of a soil profile based on bulk density
+
+        Args:
+            bulk_density: the soil profile bulk density (Mg per cubic meter)
+
+        Returns:
+            the maximum damping depth (mm)
+
+        SWAT Reference: 1:1.3.7
```
morrowcj: Right?
```suggestion
        SWAT Reference: 1:1.3.6
```

SC_redesign/Crop_and_Soil/soil/soil_temp.py
```diff
@@ -0,0 +1,544 @@
...
+
+        return ratio / (ratio + exp(-0.867 - (2.078 * ratio)))
+
+    @staticmethod
+    def _determine_radiation_factor(solar_radiation: float, albedo: float) -> float:
+        """calculates the radiation term for use in calculating the bare soil surface temp
+
+        Args:
+            solar_radiation: solar radiation reaching the ground on a given day (MJ per square meter per day)
+            albedo: the albedo of the soil (Google says "ratio of reflected to incident solar radiation") (unitless?)
```
morrowcj: Reflect my comments on albedo doc above.

SC_redesign/Crop_and_Soil/soil/soil_temp.py
```diff
@@ -0,0 +1,544 @@
...
+            the radiation factor for the day (unitless)
+
+        SWAT Reference: 1:1.3.10
+        """
+        return ((solar_radiation * (1 - albedo)) - 14) / 20
+
+    @staticmethod
+    def _determine_bare_soil_surface_temp(radiation_factor: float, avg_temp: float, min_temp: float,
+                                          max_temp: float) -> float:
+        """calculates the temperature of the soil surface if it were bare
```
morrowcj: How about this?

```suggestion
        """calculates the temperature at the surface of bare soil.
```

SC_redesign/Crop_and_Soil/soil/soil_temp.py
```diff
@@ -0,0 +1,544 @@
...
+
+        Args:
+            radiation_factor: radiation factor for a given day (unitless)
+            avg_temp: average temperature of a given day (degrees C)
+            min_temp: minimum temperature of a given day (degrees C)
+            max_temp: maximum temperature of a given day (degrees C)
```
morrowcj: Change these to "the current day" as before, here and elsewhere that makes sense.

SC_redesign/Crop_and_Soil/soil/soil_temp.py
```diff
@@ -0,0 +1,544 @@
...
+
+        Returns:
+            the temperature of the surface soil if it were bare (degrees C)
```
morrowcj:
```suggestion
            bare soil surface temperature (degrees C)
```

SC_redesign/Crop_and_Soil/soil/soil_temp.py
```diff
@@ -0,0 +1,544 @@
...
+        return cover_weighting_factor * previous_top_soil_layer_temp + \
+            (1 - cover_weighting_factor) * bare_soil_surface_temp
+
+    @staticmethod
+    def _determine_average_soil_temperature(lag_coefficient: float, previous_day_soil_temp: float, depth_factor: float,
+                                            avg_annual_air_temp: float, soil_surface_temp: float) -> float:
+        """calculates daily average soil temperature at center of a given soil layer
+
+        Args:
+            lag_coefficient: coefficient that controls influence of previous day's temp on current day's temp (degrees C)
```
morrowcj: see previous comments about lag.

SC_redesign/Crop_and_Soil/soil/soil_temp.py
```diff
@@ -0,0 +1,544 @@
...
+            previous_day_soil_temp: soil temperature in the layer from the previous day (degrees C)
+            depth_factor: factor that quantifies the influence of depth below surface on soil temp (unitless)
```
morrowcj:
```suggestion
            depth_factor: factor that quantifies the influence of depth below surface on soil temperature (unitless)
```

morrowcj (changes requested):
# Summary
This looks great. I'm requesting some minor changes, mostly in reference to documentation. Hopefully we can merge these today.

# Other Comments
* I just noticed that the comment block on lines 5-8 of `layer_data.py` could be used as the main docstring for the class 
(i.e., move it right below `class LayerData:` [line 12]). What do you think?

SC_redesign/Crop_and_Soil/soil/soil_temp.py
Isgotkowitz: Whoops yes sorry about that

SC_redesign/Crop_and_Soil/soil/soil_temp.py
Isgotkowitz: I don't know. The most detail I have found about `avg_annual_air_temp` is "average annual air temperature is calculated from  the long-term monthly maximum and minimum temperatures reported in the weather generator input (.wgn) file" on page 42 of SWAT

SC_redesign/Crop_and_Soil/soil/soil_temp.py
Isgotkowitz: Yes, good thinking

SC_redesign/Crop_and_Soil/soil/soil_temp.py
Isgotkowitz: Yup, whoops

SC_redesign/Crop_and_Soil/soil/soil_temp.py
Isgotkowitz: Done

SC_redesign/Crop_and_Soil/soil/soil_data.py
Isgotkowitz: Agreed. I will change it to that. Do you think the docstring is ok as it is?

SC_redesign/Crop_and_Soil/soil/soil_data.py
Isgotkowitz: Yes that would be better. I was just going along with the docstrings that were being autogenerated for me

SC_redesign/Crop_and_Soil/soil/soil_temp.py
Isgotkowitz: Done, I think I got all of them but may not have

SC_redesign/Crop_and_Soil/soil/soil_temp.py
Isgotkowitz: I think I have taken care of this

SC_redesign/Crop_and_Soil/soil/soil_data.py
morrowcj: Yes

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -17,36 +18,70 @@ class LayerData:
...
+    # --- Percolation
+    temperature: float = 15.05
+    """temperature of soil layer (degrees C)"""
+    saturated_hydraulic_conductivity: float = 9.5
+    """saturated hydraulic conductivity for this layer of soil (mm per hour)"""
+    available_water_capacity: float = 0.2
+    """available water capacity expressed as fraction of total soil volume (unitless)"""
+
+    # --- Temperature
+    bulk_density: float = 1.4
```
matthew7838, line 40: In some of the above docstrings, there are references to where they are in the SWAT. I wonder if the same principle would apply here since unlike `previous_day_temperature`, this is a less intuitive attribute from my point of view. I was able to find reference on SWAT 2:3.1.1.

matthew7838 (commented):
Looks good, just one reference I would recommend for improving readability.

morrowcj (approved):

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: Usually I don't think it is necessary to include a docstring for an attribute that is assumed by SWAT to be given in the input and is actually included in some of our inputs. If you check out some of the example soil input files (there are a bunch in `MASM\input\soil\`) you'll see that each layer has a bulk density field. This is in contrast to attributes in `LayerData` like `soil_evaporation_compensation_coefficient`, where SWAT assumes that the user provides it as input but we do not have it in any of our soil data input files.

That said it's a good idea to give users a reference for how SWAT thinks this attribute should be calculated, in the case they want to check it against their own method for calculating bulk density.

PR #337, opened 2023-02-08
added life_cycle variables to OM pool

ew3361zh:
## What
Reworking the life_cycle_manager daily update reporting to the OM to be more complete and hopefully easier to track because they are all added at once, once per day. Readding some previous variables removed from OM pool.

## Why
Looking through the current herd_report from the output_handler, there are a number of variables that hadn't been added to the output_manager in part because their names change multiple times throughout the model but I believe this captures everything in the herd_report now.

## How
Looking through the list of variables in the output_handler herd_report and trying to match them up with attributes in the LifeCycleManager class. Added a few more that seemed to go hand in hand with the ones already added. Easy enough to remove any unneeded/unwanted at any point.

### Notes

- One addition to general .gitignore file to prevent output generated in manure module from being added to github.
- At Loi's suggestion, I am closing out the previous version of this PR (#330) which branched off master to this version which is essentially the same but branches off the animal_life_cycle_refactor_pr branch, hopefully preventing merge conflicts down the road.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,11 +19,13 @@
 build
 others
 
-# Ignore contents of output directory, except for .keep file
-# This is because we want to keep the output direcectory, and git
+# Ignore contents of output directories, except for .keep files
+# This is because we want to keep the output directories, and git
 # only stores non-empty directories
 output/*
 !output/.keep
+RUFAS/routines/manure/output/*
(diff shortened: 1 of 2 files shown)


ltp23 (approved):
Thanks Niko for reworking your other PR.

PR #338, opened 2023-02-09
Crop_yields wiki

LaplacetteCM:
This is the proposed wiki for crop_yields.py that would replace the current crop_yields.md.

I uploaded a new file since this one is with the formatting we are currently using and you might want to keep the other explaining the whole routine for later. If not, we should delete it to avoid confussions or I can start this PR from there. 

One comment before you take a look: when updating above-ground biomass because of dry-down, the new above-ground biomass is still the same variable. However, when writing the equation I had to point a difference in the notation to be able to explain what we are doing. Do you have a better idea for that?

Also, note that in lines 149 and 165 of yields.py it is explained as a commet that we use "potential harvest index of a given day" when we are actually using harvest_index (after the water def adjustment). Just a quick note but I didn't want to forget.

### DIFF
--- a/SC_redesign/docs_SC/model_docs/crop.md
+++ b/SC_redesign/docs_SC/model_docs/crop.md
@@ -1,0 +1,32 @@
+## __Crop__
+*Under construction*
+
+### __Crop data__
+
+The growth cycle of a crop is controlled by attributes contained in the `CropData` data class. `CropData` 
+stores and updates all the crop variables needed to configure and initialize the processes involved when
+a crop grows in a field. 
+
+Differences in growth between crops are also defined by the parameters contained in `CropData`. The class
+holds default values for a generic crop as well as specific crop data classes (*e.g.*, `CornData` class) with 
+its values set to different defaults.
+
+The **crops supported by RUFAS** at present with available crop data classes within `CropData` are: 
+
+- Corn (*Zea mays*)
+- Spring wheat (*Triticum aestivum*)
+- Winter wheat (*Triticum aestivum*)
+- Cereal rye (*Secale cereale*)
+- Spring barley (*Hordeum vulgare*)
+- Fall oats (*Avena sativa*)
+- Tall fescue (*Festuca arundinacea*)
+- Alfalfa (*Medicago sativa*)
+- Soybean (*Glycine max*)
+- Sugar beet (*Beta vulgaris var. saccharifera*)
+- Potato (*Solanum tuberosum*)
+- Triticale (*x Triticosecale*)
+
+---
+**References**: for more information visit [Sphinx][1] documentation (TBA). 
+
+[1]:https://www.sphinx-doc.org/en/master/
--- a/SC_redesign/docs_SC/model_docs/crop.md
+++ b/SC_redesign/docs_SC/model_docs/crop.md
@@ -25,7 +25,8 @@ The **crops supported by RUFAS** at present with available crop data classes wit
 - Soybean (*Glycine max*)
 - Sugar beet (*Beta vulgaris var. saccharifera*)
 - Potato (*Solanum tuberosum*)
-- Triticale (*x Triticosecale*)
+<!-- - Triticale (*x Triticosecale*) -->
+
 
 ---
 **References**: for more information visit [Sphinx][1] documentation (TBA). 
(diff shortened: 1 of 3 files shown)


SC_redesign/docs_SC/model_docs/yields.md
```diff
@@ -0,0 +1,114 @@
...
+$$HI_{\text{act}} = (HI_{\text{pot}} - HI_{\text{min}})\times \frac {W_{\text{def}}}{W_{\text{def}} + exp(6.13 - 0.883\times W_{\text{def}})} + HI_{\text{min}}
+$$
+
+where $HI_{\text{act}}$ is the actual harvest index for a given day (`harvest_index`), $HI_{\text{pot}}$ is the potential harvest index for a given day  
+(`potential_harvest_index`), $HI_{\text{min}}$ is the minimum harvest index allowed for the plant under drought conditions (`min_harvest_index`), and 
+$W_{\text{def}}$ is the water deficiency factor (`water_deficiency`).
+
+Alternatively, the user can provide a **target harvest index** for the crop (`user_harvest_index`). When that occurs, the provided value is used in place of the **harvest index** (`harvest_index`) calculated with the previous method.  
+
+Once crops reach maturity (`heat_fraction`>= 1.0), the **above-ground biomass** is adjusted by the water losses due to the start of the dry-down process. The new above-ground biomass value is calculated using the pre-dry-down above-ground biomass and the dry-down fraction:
```
morrowcj: LaTeX mode also works inline to make expression like this inequality a little better.
```suggestion
Once crops reach maturity (`heat_fraction` $\geq 1.0$), the **above-ground biomass** is adjusted by the water losses due to the start of the dry-down process. The new above-ground biomass value is calculated using the pre-dry-down above-ground biomass and the dry-down fraction:

```

morrowcj (approved):
Celina, I think this looks great. I had one very small suggestion about formatting an expression below. 

This file should replace the old yields file (you can move that one to a new `docs_SC/model_docs/archive/` folder)

morrowcj:
> Also, note that in lines 149 and 165 of yields.py it is explained as a commet that we use "potential harvest index of a given day" when we are actually using harvest_index (after the water def adjustment). Just a quick note but I didn't want to forget.

I've just pushed a change addressing thins. Thank you for pointing it out.

morrowcj:
> One comment before you take a look: when updating above-ground biomass because of dry-down, the new above-ground biomass is still the same variable. However, when writing the equation I had to point a difference in the notation to be able to explain what we are doing. Do you have a better idea for that?

No, I think how you did this is best.

SC_redesign/docs_SC/model_docs/yields.md
LaplacetteCM: Yes, that's better! In fact I did it that way in one of the equations but totally forgot to make the change in the paragraph. Thanks

SC_redesign/docs_SC/model_docs/crop.md
```diff
@@ -0,0 +1,32 @@
...
+
+### __Crop data__
+
+The growth cycle of a crop is controlled by attributes contained in the `CropData` data class. `CropData` 
+stores and updates all the crop variables needed to configure and initialize the processes involved when
+a crop grows in a field. 
+
+Differences in growth between crops are also defined by the parameters contained in `CropData`. The class
+holds default values for a generic crop as well as specific crop data classes (*e.g.*, `CornData` class) with 
+its values set to different defaults.
```
morrowcj:
```suggestion
their values set to different defaults.

```

SC_redesign/docs_SC/model_docs/crop.md
```diff
@@ -0,0 +1,32 @@
...
+- Winter wheat (*Triticum aestivum*)
+- Cereal rye (*Secale cereale*)
+- Spring barley (*Hordeum vulgare*)
+- Fall oats (*Avena sativa*)
+- Tall fescue (*Festuca arundinacea*)
+- Alfalfa (*Medicago sativa*)
+- Soybean (*Glycine max*)
+- Sugar beet (*Beta vulgaris var. saccharifera*)
+- Potato (*Solanum tuberosum*)
+- Triticale (*x Triticosecale*)
```
morrowcj: I'm not comfortable supporting triticale right now, since we don't have any defaults. Let's comment this out until we can validate its attributes.

```suggestion
<!-- - Triticale (*x Triticosecale*) -->

```

morrowcj (approved):
This looks great. I have two minor requests about `crop.md`. You can merge whenever you're comfortable.

SC_redesign/docs_SC/model_docs/crop.md
LaplacetteCM: Thought about that too, I agree.

PR #339, opened 2023-02-09
Pen unit tests and refactoring

ancheypeng:
This PR focuses on increasing unit test coverage for the Pen class along with some minor refactoring to improve the testability of the code.

Unit tests were added for the following functions:
- `set_avg_nutrient_rqmts`
- `set_milk_avgs`
- `update_animals `
    - `add_new_animals`
    - `update_pen_populated`
    - `update_stocking_density`
    - `set_animal_combination`

`update_animals` was refactored to include more helper functions to improve testability and SRP.


RUFAS/routines/animal/pen.py
```diff
@@ -265,60 +271,68 @@ def __init__(self, pen_id, vertical_dist_to_milking_parlor, horizontal_dist_to_m
...
         """
-        Sets the pen's id to id.
+        self.stocking_density = len(self.animals_in_pen) / self.num_stalls * 100
 
+    def set_animal_combination(self, animal_combination: AnimalCombination) -> None:
+        """
+        Sets the pen's animal combination to animal_combination
         Args:
-            pen_id: The pen's unique pen ID, obtained from the input file.
+            animal_combination: the new AnimalCombination
```
Fjordrunner: A minor change to read "the new AnimalCombination for the/this/the given/etc. pen" might be helpful here for readability purposes?

RUFAS/routines/animal/pen.py
```diff
@@ -265,60 +271,68 @@ def __init__(self, pen_id, vertical_dist_to_milking_parlor, horizontal_dist_to_m
...
-        self.id = pen_id
-
-    def set_avg_nutrient_rqmts(self, avg_nutrient_rqmts):
-        self.avg_nutrient_rqmts = {key: value for (key, value) in avg_nutrient_rqmts.items()}
+        self.animal_combination = animal_combination
 
-    def set_milk_avgs(self, avg_milk, avg_CP_milk):
-        self.avg_milk = avg_milk
-        self.avg_CP_milk = avg_CP_milk
+    # TODO: Remove this functionality once pen has been fully switched to AnimalCombination enum
```
Fjordrunner, line 346: Has this TODO been met yet, or are we still in the process of switching over to using the AnimalCombination enum?

RUFAS/routines/animal/pen.py
```diff
@@ -265,60 +271,68 @@ def __init__(self, pen_id, vertical_dist_to_milking_parlor, horizontal_dist_to_m
...
+    def update_classes_in_pen(self) -> None:
+        """
+        Updates the classes contained within the pen
+        """
+        self.classes_in_pen = set()
+        for animal in self.animals_in_pen:
+            stage = type(animal).__name__
```
Fjordrunner: `stage` by itself is a bit misleading here, maybe a change to "life_cycle_stage" or "animal_stage" could be appropriate?

Fjordrunner (approved):

RUFAS/routines/animal/pen.py
ancheypeng: No, not yet, `calc_ration` still uses the `classes_in_pen` variable. It will be refactored soon

RUFAS/routines/animal/pen.py
```diff
@@ -265,60 +271,68 @@ def __init__(self, pen_id, vertical_dist_to_milking_parlor, horizontal_dist_to_m
...
 
-    def set_id(self, pen_id):
+    def update_stocking_density(self) -> None:
+        """
+        Updates the stocking density of the pen
         """
-        Sets the pen's id to id.
+        self.stocking_density = len(self.animals_in_pen) / self.num_stalls * 100
 
+    def set_animal_combination(self, animal_combination: AnimalCombination) -> None:
```
JoeWaddell: Should this be renamed as `update_animal_combination` instead of `set...`, since the `animal_combination` is first defined/set in `__init__`?

JoeWaddell (approved):
This looks great: nice work on the refactor!

The only general comment to make: this could be a good time to update the docstrings for the functions so that they follow the Sphinx formatting (it's already done here for the Pen class).  That's an easy fix addition though, since only a few functions here require parameters other than self.

RUFAS/output_handler/reports/pen_report.py
```diff
@@ -135,7 +135,7 @@ def __init__(self, data: Dict[Any, Any], feed, pen_id: int) -> None:
 
                 self.daily_variables[str(feed_id) + "(" + feed_name + ")"] = \
                     [
-                        'pen.ration[\'%s\'] if pen.pen_populated and \'%s\' in pen.ration.keys() else 0' % (
+                        'pen.ration[\'%s\'] if pen.populated and \'%s\' in pen.ration.keys() else 0' % (
```
PooyaHekmati, line 138: can we use f-string here?

ancheypeng: I don't think so, the code isn't actually meant to be a string. It's evaluated later on as python code using the `eval()` function (line 62 of pen_report). (this seems like quite bad practice, but the whole reporting system is built on this as of now)

PR #342, opened 2023-02-10
Refactor and test `SoilErosion` module

### LINKED ISSUE #341: refactor and test soil_erosion.py (soil) into SC_redesign directory

### LINKED ISSUE #343: refactor and test erosion.py (soil/phosphorus_cycling) into SC_redesign_directory

Isgotkowitz:
# Summary
This PR refactors the `SoilErosion` module housed in the `soil_erosion.py` file based on section 4:1.1 of SWAT. The module is entirely comprised of static methods with the exception of the main routine that handles updating , `erode()`. All methods are tested in `test_soil_erosion.py`

# Changes
- `crop_data.py`: added attribute that is used to calculate erosion
- `layer_data.py`: added two attributes for use in calculating erosion, and one that is calculated in `Infiltration` but is also used by `SoilErosion`
- Also in `layer_data.py`: added the `@property` to calculate `percent_organic_matter_content` based on `percent_organic_carbon_content` in `LayerData` module. `percent_organic_matter_content` is provided in the example soil input files, but it appears to have been calculated with a formula different from the one SWAT specifies (4:1.1.4).
- `soil.py`: added import statement and attribute for `SoilErosion`, removed initialization of `SoilData` attributes which were already being initialized in `SoilData`
- `soil_data.py`:  added a number of attributes for `SoilErosion` to use. Fixed small mistakes in some `@property` methods. Added one `@property` method for `SoilErosion` to use
- `infiltration.py`: now records runoff in a `SoilData` attribute
- `test_infiltration.py`: tests that runoff is stored properly in `SoilData` object

### DIFF
--- a/SC_redesign/Crop_and_Soil/soil/soil_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/soil_data.py
@@ -67,6 +67,7 @@ class SoilData:
     """the peak runoff rate (meters cubed per second)"""
     field_size: float = 1
     """area of the field which contains this soil (hectares)
+    
         NOTE: the field is treated as its own HRU, so this attribute may be used as the area of the HRU
     """
     snow_cover_water_content: float = 0
--- a/SC_redesign/Crop_and_Soil/soil/soil_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/soil_data.py
@@ -34,6 +34,8 @@ class SoilData:
     """average slope fraction of the subbasin (unitless)"""
     moisture_condition_parameter: Optional[float] = None
     """curve number value adjusted for moisture content (unitless) (SWAT 2:1.1.11)"""
+    accumulated_runoff: Optional[float] = None
+    """accumulated runoff or rainfall excess (mm)"""
 
     # ---- percolation
     vadose_zone_layer: Optional[LayerData] = None
@@ -49,6 +51,26 @@ class SoilData:
     """variable that controls the influence of previous day's temperature on current day's temperature, range is from 0
     to 1, inclusive. SWAT sets the lag coefficient to 0.8 (paragraph between equations 1:1.3.3, 4) (unitless)
     """
+
+    # ---- Erosion
+    slope_length: float = 3
+    """length of the slope (meters)"""
+    percent_sand_content: float = 15
+    """percent sand content in the soil profile (unitless)"""
+    percent_silt_content: float = 65
+    """percent silt content in the soil profile (unitless)"""
+    percent_rock_content: float = 0.1
+    """percent rock content in the first layer of soil (unitless)"""
+    manning: float = 0.4
+    """the Manning roughness coefficient for this subbasin (unitless)"""
+    peak_runoff_rate: Optional[float] = None
+    """the peak runoff rate (meters cubed per second)"""
+    field_size: float = 1
+    """area of the field which contains this soil (hectares)
+        NOTE: the field is treated as its own HRU, so this attribute may be used as the area of the HRU
+    """
+    eroded_sediment: float = 0
+    """cumulative amount of sediment that has been eroded off of the field (metric tons)"""
 
     def __post_init__(self):
         if self.soil_layers is None:
@@ -82,14 +104,12 @@ class SoilData:
             water_sum = 0
             for layer in self.soil_layers:
                 water_sum += max(0, (layer.water_content - layer.wilting_point_content))
-            return
+            return water_sum
 
     @property
     def profile_saturation(self) -> float:
         """
-
         Returns: amount of water in the soil profile when completely saturated (mm)
-
         """
         if self.soil_layers is None:
             return 0
@@ -102,9 +122,7 @@ class SoilData:
     @property
     def profile_field_capacity(self) -> float:
         """
-
-        Returns: total amount of water contained in the entire soil profile at field capacity (but not saturated) (mm)
-
+         Returns: total amount of water contained in the entire soil profile at field capacity (but not saturated) (mm)
         """
         if self.soil_layers is None:
             return 0
@@ -115,10 +133,8 @@ class SoilData:
             return field_capacity_sum
 
     @property
-    def soil_water_factor(self):
-        """
-
-        Returns: the soil water factor (unitless)
+    def soil_water_factor(self) -> float:
+        """Returns: the soil water factor (unitless)
 
         SWAT Reference: 5:2.3.18
         """
@@ -132,9 +148,9 @@ class SoilData:
     #   - average slope fraction is always in the range 0 to 1 inclusive (I think)
 
     @property
-    def profile_bulk_density(self):
-        """average bulk density of the soil profile based on the bulk density of each soil layer, weighted by
-            the thickness
+    def profile_bulk_density(self) -> float:
+        """average bulk density of the soil profile based on the bulk density of each soil layer, weighted by the
+            thickness
         """
         weighted_densities_sum = 0
         weights_sum = 0
@@ -142,3 +158,13 @@ class SoilData:
             weighted_densities_sum += (layer.layer_thickness * layer.bulk_density)
             weights_sum += layer.layer_thickness
         return weighted_densities_sum / weights_sum
+
+    @property
+    def surface_runoff_volume(self) -> float:
+        """volume of surface runoff (mm per hectare)
+
+        Used in SWAT equation 4:1.1.1
+        """
+        if self.accumulated_runoff is None or self.field_size is None:
+            return None
+        return self.accumulated_runoff / self.field_size
--- a/SC_redesign/Crop_and_Soil/soil/soil_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/soil_data.py
@@ -30,8 +30,8 @@ class SoilData:
     # ---- infiltration
     previous_retention_parameter: Optional[float] = None
     """retention parameter for the previous day (mm) (used in SWAT 2:1.1.9)"""
-    average_slope_fraction: float = 0.05
-    """average slope fraction of the subbasin (unitless)"""
+    average_subbasin_slope: float = 0.05
+    """average slope of the subbasin expressed as rise over run (meters / meters)"""
     moisture_condition_parameter: Optional[float] = None
     """curve number value adjusted for moisture content (unitless) (SWAT 2:1.1.11)"""
     accumulated_runoff: Optional[float] = None
@@ -55,12 +55,6 @@ class SoilData:
     # ---- Erosion
     slope_length: float = 3
     """length of the slope (meters)"""
-    percent_sand_content: float = 15
-    """percent sand content in the soil profile (unitless)"""
-    percent_silt_content: float = 65
-    """percent silt content in the soil profile (unitless)"""
-    first_rock_percent: float = 0.1
-    """percent rock content in the first layer of soil (unitless)"""
     manning: float = 0.4
     """the Manning roughness coefficient for this subbasin (unitless)"""
     peak_runoff_rate: Optional[float] = None
--- a/SC_redesign/Crop_and_Soil/soil/soil_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/soil_data.py
@@ -59,7 +59,7 @@ class SoilData:
     """percent sand content in the soil profile (unitless)"""
     percent_silt_content: float = 65
     """percent silt content in the soil profile (unitless)"""
-    percent_rock_content: float = 0.1
+    first_rock_percent: float = 0.1
     """percent rock content in the first layer of soil (unitless)"""
     manning: float = 0.4
     """the Manning roughness coefficient for this subbasin (unitless)"""
--- a/SC_redesign/Crop_and_Soil/soil/soil_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/soil_data.py
@@ -65,14 +65,12 @@ class SoilData:
     """the Manning roughness coefficient for this subbasin (unitless)"""
     peak_runoff_rate: Optional[float] = None
     """the peak runoff rate (meters cubed per second)"""
-    field_size: float = 1
-    """area of the field which contains this soil (hectares)
-        NOTE: the field is treated as its own HRU, so this attribute may be used as the area of the HRU
-    """
     snow_cover_water_content: float = 0
     """water content of the snow cover (mm)"""
     eroded_sediment: float = 0
     """cumulative amount of sediment that has been eroded off of the field (metric tons)"""
+    surface_volume_runoff: Optional[float] = None
+    """volume of surface runoff (mm per hectare), used in SWAT equation 4:1.1.1."""
 
     def __post_init__(self):
         if self.soil_layers is None:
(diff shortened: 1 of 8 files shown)


SC_redesign/Crop_and_Soil/soil/soil_erosion.py
```diff
@@ -0,0 +1,251 @@
...
+        topographic_factor = self._determine_topographic_factor(self.data.slope_length,
+                                                                self.data.average_slope_fraction)
+        fragment_factor = self._determine_coarse_fragment_factor(self.data.percent_rock_content)
+
+        if self.data.peak_runoff_rate is None:
+            raise TypeError("SoilData peak_runoff_rate cannot be NoneType")
+            return
+        elif self.data.surface_runoff_volume is None:
+            raise TypeError("SoilData surface_runoff_rate cannot be NoneType")
+            return
```
ltp23: I think you don't need to include the return statements here because after an error has been raised, the program will stop.

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_soil_erosion.py
```diff
@@ -0,0 +1,263 @@
...
+# --- Static method tests ---
+@pytest.mark.parametrize("sand,silt", [
+    (15, 65),
+    (0, 0),
+    (17, 60),
+    (9, 80),
+    (12.339485, 61.1938549),
+    (23.4958769, 58.1093485),
+])
+def test_determine_coarse_sand_factor(sand, silt):
```
ltp23: Add type annotations in tests too if you can. 
```
def test_determine_coarse_sand_factor(sand: float, silt: float) -> None:
```

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -50,6 +52,28 @@ class SoilData:
...
+    """percent silt content in the soil profile (unitless)"""
+    percent_rock_content: float = 0.1
+    """percent rock content in the first layer of soil (unitless)"""
+    manning: float = 0.4
+    """the Manning roughness coefficient for this subbasin (unitless)"""
+    peak_runoff_rate: Optional[float] = None
+    """the peak runoff rate (meters cubed per second)"""
+    field_size: float = 1
+    """area of the field which contains this soil (hectares)
+        NOTE: the field is treated as its own HRU, so this attribute may be used as the area of the HRU
```
```suggestion
    """area of the field which contains this soil (hectares)
    
        NOTE: the field is treated as its own HRU, so this attribute may be used as the area of the HRU
```

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -50,6 +52,28 @@ class SoilData:
...
+    """percent sand content in the soil profile (unitless)"""
+    percent_silt_content: float = 65
+    """percent silt content in the soil profile (unitless)"""
+    percent_rock_content: float = 0.1
+    """percent rock content in the first layer of soil (unitless)"""
+    manning: float = 0.4
+    """the Manning roughness coefficient for this subbasin (unitless)"""
+    peak_runoff_rate: Optional[float] = None
+    """the peak runoff rate (meters cubed per second)"""
+    field_size: float = 1
```
morrowcj: I don't think this should be a Soil attribute. It is a field attribute. Take it as an input parameter for the methods that need it.

SC_redesign/Crop_and_Soil/soil/soil_erosion.py
```diff
@@ -0,0 +1,251 @@
...
+
+    def erode(self, minimum_cover_management_factor: float, surface_residue: float) -> None:
+        """this is the main routine for SoilErosion. It is responsible for running all the necessary soil erosion
+            methods and updating attributes as necessary
+
+        Args:
+            minimum_cover_management_factor: minimum value for cover and management factor for water erosion applicable
+                to land cover/plant (unitless)
+            surface_residue: amount of residue on the soil surface (kg per hectare)
+
```
morrowcj, line 24: Can we add a "Details" field to the main method documentation (may need to go back and do so for the entire SC module at some point)? I'd like the document to include a detailed, but concise (plain English) summary of the logic.

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_soil_erosion.py
```diff
@@ -0,0 +1,263 @@
...
+    incorp = SoilErosion(data)
+
+    # Mock helper function
+    incorp._determine_soil_erodibility_factor = MagicMock(return_value=0.95)
+    incorp._determine_cover_management_factor = MagicMock(return_value=0.95)
+    incorp._determine_support_practice_factor = MagicMock(return_value=0.95)
+    incorp._determine_topographic_factor = MagicMock(return_value=0.95)
+    incorp._determine_coarse_fragment_factor = MagicMock(return_value=0.95)
+    incorp._determine_sediment_yield = MagicMock(return_value=0.05)
+    incorp._determine_adjusted_sediment_yield = MagicMock(return_value=0.0498)
```
ltp23: This is interesting. I haven't thought of patching functions directly like this. Normally, I'm just too used to using the more verbose way of `mocker.patch()`, but this looks more concise.

ltp23 (approved):
Thanks, Ed. This PR looks great to me.

SC_redesign/Crop_and_Soil/soil/soil_erosion.py
```diff
@@ -0,0 +1,251 @@
...
+    @staticmethod
+    def _determine_coarse_sand_factor(percent_sand_content: float, percent_silt_content: float) -> float:
+        """calculates factor based on sand content for use in determining soil erodibility factor
+
+        Args:
+            percent_sand_content: percent of soil content that is sand
+            percent_silt_content: percent of soil content that is silt
+
+        Returns:
+            factor based on sand content (unitless)
```
morrowcj: Can we describe this as something other than just "factor". Perhaps something like the following

```suggestion
        """calculates the coarseness factor of soil erodbility

        Args:
            percent_sand_content: percent of soil content that is sand
            percent_silt_content: percent of soil content that is silt
            
        Details: 
          The coarseness of a soil effects the overall erodibility of the soil. Specifically, soils with high-levels of 
          coarse-sand content will have relatively low erodibility compared to soils with less sand. 

        Returns:
            coarseness factor of erodibility, based on sand content (unitless)
```

Fjordrunner (approved):
Overall, this PR looks good, but a singular (albeit nitpicky) critique I add here instead of adding a bunch of comments for you to scroll through and review is that for many of your function docstrings, you alternate between capitalizing and not capitalizing the first word in sentences. It's really not a big deal, but in the case that you revise the functions before submitting I believe that capitalizing the first word of sentences in documentation is standard.

SC_redesign/Crop_and_Soil/soil/soil_erosion.py
```diff
@@ -0,0 +1,251 @@
...
+
+        SWAT Reference: 4:1.1.6
+        """
+        return 0.2 + 0.3 * exp((-0.256) * percent_sand_content * (1 - (percent_silt_content / 100)))
+
+    @staticmethod
+    def _determine_clay_silt_ratio_factor(percent_silt_content: float, percent_clay_content: float) -> float:
+        """calculates factor based on the clay-silt ratio for use in calculating soil erodibility factor
```
morrowcj:
```suggestion
        """calculates the component factor of erodibility that is based on the clay-silt ratio.
```

SC_redesign/Crop_and_Soil/soil/soil_erosion.py
Isgotkowitz: Good suggestion

SC_redesign/Crop_and_Soil/soil/soil_erosion.py
```diff
@@ -0,0 +1,251 @@
...
+        Args:
+            percent_sand_content: percent of sand in the given layer of soil
+
+        Returns:
+            factor for adjusting soil erodibility factor based on high sand contents
+
+        SWAT Reference: 4:1.1.9
+        """
+        inverse_sand_percentage = 1 - (percent_sand_content / 100)
+        # TODO: better name for this variable
```
morrowcj: This isn't important at al, but how about

```suggestion
        not_sand_fraction = 1 - (percent_sand_content / 100)
```

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_soil_erosion.py
Isgotkowitz: Ok, I will add them in this file, though all of the previous test classes I have written do not have them. Do you have a certain rationale for why you like type annotations in test classes? To me, they have always seemed a little unnecessary but I would appreciate hearing your thoughts on this

SC_redesign/Crop_and_Soil/soil/soil_erosion.py
```diff
@@ -0,0 +1,251 @@
...
+        first_multiplicative_term = log(0.8) - log(minimum_cover_management_factor)
+        second_multiplicative_term = exp(-0.00115 * surface_residue)
+        second_additive_term = log(minimum_cover_management_factor)
+        return exp(first_multiplicative_term * second_multiplicative_term + second_additive_term)
+
+    @staticmethod
+    def _determine_support_practice_factor() -> float:
+        """TODO: implement this for version 2 (only applies to fields that are doing contour tillage/planting,
+            stripcropping, and/or terracing) SWAT Reference: section 4:1.1.3"""
+        return 1
```
morrowcj:
```suggestion
    def _determine_support_practice_factor() -> float:  # TODO: implement this for version 2
        """SWAT Reference: section 4:1.1.3 (only applies to fields that are doing contour tillage/planting,
            stripcropping, and/or terracing)"""
        return 1
```

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_soil_erosion.py
```diff
@@ -0,0 +1,263 @@
...
+    # Initialize objects
+    data = SoilData(accumulated_runoff=13, peak_runoff_rate=0.11)
+    incorp = SoilErosion(data)
+
+    # Mock helper function
+    incorp._determine_soil_erodibility_factor = MagicMock(return_value=0.95)
+    incorp._determine_cover_management_factor = MagicMock(return_value=0.95)
+    incorp._determine_support_practice_factor = MagicMock(return_value=0.95)
+    incorp._determine_topographic_factor = MagicMock(return_value=0.95)
+    incorp._determine_coarse_fragment_factor = MagicMock(return_value=0.95)
```
morrowcj: Can we set these to have different (i.e., unique) return values? That way, it will be a little easier to detect which factor is the problem if there ever is a bug added.

morrowcj (approved):
This looks good to me. I didn't notice anything except the few minor comments below. Great work, as usual!

All clear to merge once you address the comments.

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_soil_erosion.py
Isgotkowitz: It works well in this case, I believe because the method of a specific object is being directly patched. When the method is directly patched on the class itself (i.e. `Erosion._determine_something = MagicMock(...)` that patch doesn't necessarily go away after the test is finished running, which can make other tests fail. If you're curious to see an example of this, check out TODOs I left in `test_evapotranspiration.py` because I could not get `mocker.patch()` to work correctly, and the current workaround just rearranges the order of the tests so they don't mess each other up.

Isgotkowitz:
@Fjordrunner thanks for the review, all nitpicky critiques are welcomed. I will pay more attention to the capitalizations going forward

SC_redesign/Crop_and_Soil/soil/soil_data.py
Isgotkowitz: Done

SC_redesign/Crop_and_Soil/soil/soil_erosion.py
Isgotkowitz: I like it.

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_soil_erosion.py
Isgotkowitz: Yes

SC_redesign/Crop_and_Soil/soil/soil_erosion.py
Isgotkowitz: I have added a details section, it could use a second pair of eyes though

morrowcj: It looks quite good. I'd add a bit saying that the total eroded sediment is tracked (and accumulated daily/after each erosion event)

Which makes me wonder: Should we keep track of the day's erosion amount? I'm inclined to think so [lines 49-50]:

```python
        self.data.eroded_sediment = self._determine_adjusted_sediment_yield(sediment_yield,
                                                                             self.data.snow_cover_water_content)
        self.data.accumulated_sediment_erosion += self.data.eroded_sediment
```

What do you think?

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -50,6 +52,26 @@ class SoilData:
...
 
+    # ---- Erosion
+    slope_length: float = 3
+    """length of the slope (meters)"""
+    percent_sand_content: float = 15
+    """percent sand content in the soil profile (unitless)"""
+    percent_silt_content: float = 65
+    """percent silt content in the soil profile (unitless)"""
+    percent_rock_content: float = 0.1
+    """percent rock content in the first layer of soil (unitless)"""
```
morrowcj: Are you sure these are whole-profile attributes and not layer attributes? I'm inclined to think it is the latter. I expect that erosion processes only use the top layer properties. Is that correct?

If you need a percent for the whole profile, it might be best to have properties that calculate them from the layers.

Also, `first_rock_percent` might be better for the rock, if it is truly only for the first layer.

Isgotkowitz: The reason I have sand and silt as profile attributes is because that is how they are listed in the example soil inputs, even though it does not make sense.

Isgotkowitz: They should really be layer attributes but if we changed them  in `SoilData`/`LayerData` without also changing them in the soil input files it could cause problems when we try to write the code that takes in JSON inputs and outputs the data objects

morrowcj: We don't have soil input files for `SC_redesign` yet - the world is our oyster! 

Let's make things how we want them to be, that's why we're doing this silly redesign after all.

SC_redesign/Crop_and_Soil/soil/soil_erosion.py
morrowcj:
```suggestion
            raise TypeError("SoilData peak_runoff_rate cannot be NoneType")
        elif self.data.surface_runoff_volume is None:
            raise TypeError("SoilData surface_runoff_rate cannot be NoneType")
```

Issue #351, opened 2023-02-14
`SoilErosion` does not adjust soil profile based on erosion


Isgotkowitz:
Currently, the `SoilErosion` module only calculates the amount sediment that is eroded from the soil (in metric tons) and accumulates it over the course of the simulation, but does not remove that eroded sediment from the soil profile. The main issue here is units: a suitable method (assumption, input from user, etc.) will have to be found so that metric tons can be converted to mm / ha.

morrowcj:
Note that this will affect the thickness of the top soil layer and may even result in complete loss of a soil layer (if erosion > top layer depth), which has implications for the "top 20 cm" layer required by SurPhos.

Isgotkowitz:
SMEs need to weigh in as to whether this actually needs to be accounted for or not.

morrowcj:
@KFosterReed, thoughts on this?

morrowcj:
SME grooming: closing as not planned, since this is not causing a known problem, nor is there an obvious solution.

PR #355, opened 2023-02-15
Implement features of old `soil.py` into new `soil.py` (in `SC_redesign`)and test

### LINKED ISSUE #354: implement features of old soil.py (soil) into new soil.py (SC_redesign/) and test

Isgotkowitz:
# Summary
This PR implements many of the functionalities from the old `Soil` module into the various modules within the `Crop_and_Soil/soil/` module. It also introduces new functionality for creating `SoilData` objects with the `SoilConfigFactory` module, based on the way `CropData` objects for specific species are created by the `CropSpeciesFactory` module.

# Main Changes
- `SoilData`:
    - several attributes have been added to track annual totals of or changes in profile attributes
    - there is now an optional `name` attribute that can be used to identify different `SoilData` objects
    - `second_moisture_condition_parameter` (used in the `Infiltration` module) is now an attribute of `SoilData`
    - `surface_runoff_volume` is now named consistently across `SoilData` and `SoilTemp` modules
    - `__post_init__()` has had a docstring added, checks that there is not a vadose zone `LayerData` object already initialized before initializing one, and sets some of the yearly totals
    - an `@property` method (`profile_nitrates_total()`) has also been added to calculate the total amount of nitrates in the soil profile
    - `SoilData` now takes keyword arguments only, and an import statement has been added to make this possible
    - there is an `annual_reset` method which resets yearly totals when called
    - the docstring for `profile_bulk_density` has been reformatted a little
- `LayerData`:
  - `layer_thickness()` now checks the validity of the data it is using, and throws a `ValueError` if it finds invalid attributes
  - `saturation_content()` has been moved up so it is now directly below the other `@property` methods that calculate water-related contents
- `Soil`:
    - `__init__()` now has a docstring, as well as docstrings for all the attributes in it
    - there are now two methods that are meant to be run daily, one to run water-related processes and one to run all other processes
- `SoilConfigFactory`:
  - this module lays the groundwork for more default soil configurations to be created and easily  used by the simulation/users. It is meant to follow `CropSpeciesFactory` as closely as is appropriate
- `test_soil_data_and_config_factory.py`:
  - this module functions as a test for all modules not associated with a specific section of SWAT
  - `SoilData` is now tested within this module, including configuring `SoilData` objects, `annual_reset()`, and all the `@property` methods within `SoilData`
  - `LayerData` is now tested within this module, including that `__post_init__()` works correctly, and all `@property` methods work correctly
- `Evapotranspiration` and test module:
    - `evapotranspirate()` now adjusts the annual totals for attributes that it calculates
    - the test for `evapotranspirate()` checks that totals have been updated correctly
    - the name of an attribute used has been updated to reflect how it is named in `LayerData`
    - `test_potential_evapotranspiration()` and `test_determine_soil_evaporation()` have been rewritten to use `patch` from `pytest`, which allows tests to be run independent of order within the module. `test_determine_latent_heat_vaporization()` and `test_determine_soil_cover_index()` have been moved down in the module.
- `Infiltration` and test module:
   - `infiltrate()` has had argument descriptions added to its docstring, gets `second_moisture_condition_parameter` from `SoilData` object, and updates annual totals for attributes it calculates
   - the test for `infiltrate()` has been updated to reflect the new method signature and checks that the annual totals that should be updated are getting updated correctly
- `SoilErosion` and test module:
   - the main routine, `erode()` now calculates the daily amount of sediment that gets eroded, stores it, and uses that value to update the annual total. It also updates a second annual total based on the surface runoff volume of a given day
   - the test for `erode()` now checks that updates to annual totals have been applied correctly

# Other Changes
- all tests for `Crop` modules are now in their own directory, `crop_tests`. Only the names of these files have been changed. The directory `crop_tests` also has an empty `__init__.py` file.
- `test_crop_data_and_species_factory.py` has had a small typo corrected in one of its comments
- `species_data_factory.py` has had a comment added that explains the purpose of an if condition
- `test_percolation.py` has been updated to work with `SoilData` now only taking keyword arguments


SC_redesign/Crop_and_Soil/soil/evapotranspiration.py
```diff
@@ -54,6 +54,13 @@ def evapotranspirate(self, extraterrestrial_radiation: float, max_air_temp: floa
         #  monitor?) - issue #317
         self._evaporate_from_soil()
 
+        # Update annual totals
```
morrowcj:
```suggestion
        # Update accumulated (annual) totals
```

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -64,6 +64,12 @@ def __post_init__(self):
     @property
     def layer_thickness(self) -> float:
         """thickness of soil layer (mm)"""
+        if self.top_depth < 0:
+            raise ValueError("Top depth of a soil layer cannot be less that 0")
+        elif self.bottom_depth <= 0:
+            raise ValueError("Bottom depth of a soil layer cannot be above or at the top of the soil surface")
+        elif self.top_depth >= self.bottom_depth:
+            raise ValueError("Top depth of a soil layer cannot be at or below the bottom depth of the layer")
```
morrowcj: I don't know how I feel about validation taking place within a `@property`.

@ltp23, can you provide any thoughts on this?

SC_redesign/Crop_and_Soil/soil/soil.py
```diff
@@ -12,15 +12,80 @@
...
         self.soil_erosion = SoilErosion(data)
+        """Process component that track erosion from the soil profile"""
 
     @classmethod
     def make_from_config(cls, soil_config) -> Soil:
         """"""
         Warning("create from config file not yet implement, returning default Soil()")
         return Soil()
+
+    def daily_soil_routine(self, solar_radiation: float, avg_temp: float, min_temp: float, max_temp: float,
```
morrowcj, line 45: I'd like to see these main methods utilized. 

Who is going to call them? The `Field` class? If so, add them where they should go.

SC_redesign/Crop_and_Soil/soil/soil.py
```diff
@@ -12,15 +12,80 @@
...
+            solar_radiation: solar radiation reaching the ground on the current day (MJ per square meter per day)
+            avg_temp: average temperature of the current day (degrees C)
+            min_temp: minimum temperature of the current day (degrees C)
+            max_temp: maximum temperature of the current day (degrees C)
+            plant_cover: total aboveground plant biomass and residue on the current day (kg per hectare)
+            snow_cover: water content of the snow cover on the current day (mm)
+            avg_annual_air_temp: average annual air temperature (degrees C)
+        """
+        self.soil_temp.daily_soil_temperature_update(solar_radiation, avg_temp, min_temp, max_temp, plant_cover,
+                                                     snow_cover, avg_annual_air_temp)
```
morrowcj, line 61: The only thing this method does is call another method? and pass the same arguments? That means this method is entirely unnecessary. Right? Just call `soil.soil_temp.daily_soil_temperature_update()` wherever you would call this method.

OR, are other methods going to be appended to this list? If so, add any method calls that exist and add placeholder (comments) otherwise.

SC_redesign/Crop_and_Soil/soil/soil.py
```diff
@@ -12,15 +12,80 @@
...
+    def daily_soil_routine(self, solar_radiation: float, avg_temp: float, min_temp: float, max_temp: float,
+                           plant_cover: float, snow_cover: float, avg_annual_air_temp: float) -> None:
+        """this method calls all now-water related daily update routines
+
+        Args:
+            solar_radiation: solar radiation reaching the ground on the current day (MJ per square meter per day)
+            avg_temp: average temperature of the current day (degrees C)
+            min_temp: minimum temperature of the current day (degrees C)
+            max_temp: maximum temperature of the current day (degrees C)
+            plant_cover: total aboveground plant biomass and residue on the current day (kg per hectare)
```
morrowcj:
```suggestion
            plant_cover: total above-ground plant biomass and residue on the current day (kg per hectare)
```

SC_redesign/Crop_and_Soil/soil/soil.py
```diff
@@ -12,15 +12,80 @@
...
+            snow_cover: water content of the snow cover on the current day (mm)
+            avg_annual_air_temp: average annual air temperature (degrees C)
+        """
+        self.soil_temp.daily_soil_temperature_update(solar_radiation, avg_temp, min_temp, max_temp, plant_cover,
+                                                     snow_cover, avg_annual_air_temp)
+
+    def daily_soil_water_routine(self, rainfall: float, weighting_coefficient: float,
```
morrowcj, line 63: Same question: where is this used?

SC_redesign/Crop_and_Soil/soil/soil.py
```diff
@@ -12,15 +12,80 @@
...
+                                 minimum_cover_management_factor: float) -> None:
+        """this method calls all water related daily update routines
+
+        Args:
+            rainfall: rainfall depth of current day (mm)
+            weighting_coefficient: weighting coefficient used to calculate retention coefficient for daily curve number
+                calculations dependent on plant evapotranspiration (unitless)
+            has_seasonal_high_water_table: if the HRU has a seasonal high water table (true/false)
+            extraterrestrial_radiation: radiation from the aliens, in MJ per square meter per day
+                TODO: better description
```
morrowcj:
```suggestion
            solar_radiation: incoming solar, in MJ per square meter per day
```

SC_redesign/Crop_and_Soil/soil/soil_config_factory.py
```diff
@@ -0,0 +1,39 @@
...
+
+"""
+This module serves as a way to easily create and modify SoilData objects based on a set of given configurations. It is 
+based on the CropSpeciesDataFactory model, and should contain all equivalent functionalities.
+"""
+
+
+class SoilConfigurations(Enum):
+    """Enum of all currently support soil configurations"""
+    GENERIC = "generic"
```
morrowcj:
```suggestion
class SoilConfigurations(Enum):
    """Enum of all currently support soil configurations"""
    GENERIC = "generic"
    # TODO: implement additional soil types
```

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -143,12 +196,22 @@ def soil_water_factor(self) -> float:
 
     @property
     def profile_bulk_density(self) -> float:
-        """average bulk density of the soil profile based on the bulk density of each soil layer, weighted by the
-            thickness
-        """
+        """Average bulk density of the soil profile based on the bulk density of each soil layer, weighted by the
```
matthew7838, line 203: Although it's straightforward, I think it is still less straightforward than calculating the total. Perhaps adding the reference to the definition of bulk density if possible?

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -64,6 +64,12 @@ def __post_init__(self):
     @property
     def layer_thickness(self) -> float:
         """thickness of soil layer (mm)"""
```
matthew7838, line 70: Just a suggestion, not sure if it's that helpful, but I wonder if there's a way to also check if more than one variable is invalid. Right now, the error message will only alert the first invalidity it encounters. 

My suggestion is perhaps setting a boolean for all three conditions and throwing a value error for all combinations of boolean values. There aren't too many variables to check so I think there will only be 6 combinations in total if implemented.

SC_redesign/Crop_and_Soil/soil/soil.py
```diff
@@ -12,15 +12,80 @@
...
+        """Process component that controls water infiltration from the soil surface into the profile"""
         self.percolation = Percolation(data)
+        """Process component that controls percolation of water from upper layers to lower layers"""
         self.soil_temp = SoilTemp(data)
+        """Process component that tracks and updates the temperatures within the soil profile"""
         self.soil_erosion = SoilErosion(data)
+        """Process component that track erosion from the soil profile"""
 
     @classmethod
     def make_from_config(cls, soil_config) -> Soil:
```
matthew7838, line 40: Might need docstring?

SC_redesign/Crop_and_Soil/soil/soil_erosion.py
```diff
@@ -46,8 +46,12 @@ def erode(self, field_size: float, minimum_cover_management_factor: float, surfa
         sediment_yield = self._determine_sediment_yield(self.data.surface_runoff_volume, self.data.peak_runoff_rate,
                                                         field_size, erodibility_factor, cover_factor,
                                                         support_practice_factor, topographic_factor, fragment_factor)
-        self.data.eroded_sediment += self._determine_adjusted_sediment_yield(sediment_yield,
-                                                                             self.data.snow_cover_water_content)
+        self.data.eroded_sediment = self._determine_adjusted_sediment_yield(sediment_yield,
+                                                                            self.data.snow_cover_water_content)
+
+        # Update annual totals
+        self.data.annual_eroded_sediment_total += self.data.eroded_sediment
```
matthew7838, line 55: Is there a reason why not just `self.data.annual_eroded_sediment_total += self._determine_adjusted_sediment_yield(sediment_yield, self.data.snow_cover_water_content)` ?

matthew7838 (changes requested):
### Summary
Great PR, lots of things done. All the docstrings look right to me. I don't think I caught anything missing or many things things that are needed for changes. Comments are mostly just suggestions.

SC_redesign/Crop_and_Soil/soil/soil.py
Isgotkowitz: You mean `make_from_config()` might need a docstring?

SC_redesign/Crop_and_Soil/soil/soil_erosion.py
Isgotkowitz: At this point, no. At a future point though we will need to take the value from `self._determine_adjusted_sediment_yield()` and use it to actually remove mass from the top of layer of the soil profile (issue #351). I will add a `TODO` so that is clearer.

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: That is a good idea. I think it would be good to talk to @morrowcj  first though because he might know more about how errors are going to be pushed back up to end users, and that will inform how complex the error handling should be

SC_redesign/Crop_and_Soil/soil/soil_data.py
Isgotkowitz: What do you mean by "calculating the total"?

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: talked to Pooya and he made a good suggestion for the best error message format, I think it should be up to snuff now

SC_redesign/Crop_and_Soil/soil/soil.py
matthew7838: Yes, sorry for the confusion,

SC_redesign/Crop_and_Soil/soil/soil_data.py
matthew7838: I just mean that compare to other functions that are just looping and calculating the total, this one is less straightforward. Basically just suggesting a reference for bulk density. Sorry for the confusion.

Isgotkowitz: The reason it isn't calculating a simple total is because it is doing a weighted average, not a regular average. Also there is a reference to SWAT for the definition of bulk density in `layer_data.py` which I think is a better place to have it, because that is where users will be directly inputting their data.

SC_redesign/Crop_and_Soil/soil/soil.py
Isgotkowitz: Added

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -28,6 +55,8 @@ class SoilData:
     """maximum amount of evaporation from soil on a given day (mm)"""
 
     # ---- infiltration
+    second_moisture_condition_parameter: float = 85
+    """curve number for average moisture conditions (unitless)"""
```
```suggestion
    """'curve number' parameter for average moisture conditions equation (unitless)"""
```

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -62,22 +91,46 @@ class SoilData:
...
     eroded_sediment: float = 0
-    """cumulative amount of sediment that has been eroded off of the field (metric tons)"""
-    surface_volume_runoff: Optional[float] = None
+    """cumulative amount of sediment that has been eroded off of the field in a day (metric tons)"""
+    surface_runoff_volume: Optional[float] = None
     """volume of surface runoff (mm per hectare), used in SWAT equation 4:1.1.1."""
 
     def __post_init__(self):
+        """This method initializes attributes that either cannot be set to a default above or depend on need other
+            attributes in the object to be set before they can be set"""
```
```suggestion
        """This method initializes attributes that either cannot be set to a default above or depend on other
            attributes in the object to be set before they can be set"""
```

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -62,22 +91,46 @@ class SoilData:
...
+            self.vadose_zone_layer = LayerData(top_depth=self.soil_layers[-1].bottom_depth,
+                                               bottom_depth=10000000,  # bottom depth is 10,000 meters by default
+                                               soil_water_concentration=0,
+                                               saturation_point_water_concentration=inf)
+
+        # Set the initial water content for the first year of the simulation
+        self.initial_water_content = self.profile_soil_water_content
+        self.initial_nitrates_total = self.profile_nitrates_total
+
+    def annual_reset(self) -> None:
```
morrowcj: method names should start with verbs: `reset_annually()`, `do_annual_reset()`, `perform_annual_reset()`?

Also, where/how will this be called?

SC_redesign/Crop_and_Soil/soil/soil_erosion.py
```diff
@@ -46,8 +46,14 @@ def erode(self, field_size: float, minimum_cover_management_factor: float, surfa
         sediment_yield = self._determine_sediment_yield(self.data.surface_runoff_volume, self.data.peak_runoff_rate,
                                                         field_size, erodibility_factor, cover_factor,
                                                         support_practice_factor, topographic_factor, fragment_factor)
-        self.data.eroded_sediment += self._determine_adjusted_sediment_yield(sediment_yield,
-                                                                             self.data.snow_cover_water_content)
+        self.data.eroded_sediment = self._determine_adjusted_sediment_yield(sediment_yield,
+                                                                            self.data.snow_cover_water_content)
+
+        # TODO: the soil profile needs to have mass removed from it as it gets eroded - Issue #351
```
morrowcj:
```suggestion
        # TODO: does the soil profile needs to have mass removed from it as it gets eroded? - Issue #351
```

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_crop_data_and_species_factory.py
```diff
@@ -35,7 +35,7 @@ def test_invalid_crop_species_enum(invalid_species):
 
 def test_species_factory_defaults():
     """Ensure that each species come shipped with their factory-default settings. This is a check of the
-    CropSpeciesDataFactory.create_secies_data() method.
+    CropSpeciesDataFactory.create_species_data() method.
```
morrowcj, line 38: Thank you!

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_soil_data_and_config_factory.py
```diff
@@ -0,0 +1,402 @@
...
+])
+def test_soil_config_enum(config: str, expected: SoilConfigurations) -> None:
+    """Tests that SoilConfigurations properly enumerates accepted configuration names"""
+    soil_config = SoilConfigurations(config)
+    assert soil_config == expected
+
+
+@pytest.mark.parametrize("invalid_config", [
+    "prairie seafloor",
+    "indoor floor",
```
morrowcj, line 27: Nice!

SC_redesign/Crop_and_Soil/soil/soil_config_factory.py
```diff
@@ -0,0 +1,39 @@
...
+import dataclasses
+from SC_redesign.Crop_and_Soil.soil.soil_data import SoilData
+
+"""
+This module serves as a way to easily create and modify SoilData objects based on a set of given configurations. It is 
+based on the CropSpeciesDataFactory model, and should contain all equivalent functionalities.
+"""
+
+
+class SoilConfigurations(Enum):
```
morrowcj: I think this should be singular. The instantiated object is a single configuration. 

```suggestion
class SoilConfiguration(Enum):
```

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_soil_data_and_config_factory.py
```diff
@@ -0,0 +1,402 @@
...
+    assert generic.time_step == 24
+    assert generic.previous_temperature_effect == 0.8
+    assert generic.slope_length == 3
+    assert generic.manning == 0.4
+    assert generic.peak_runoff_rate is None
+    assert generic.snow_cover_water_content == 0
+    assert generic.eroded_sediment == 0
+
+    # TODO: would this maybe be a cleaner, easier way to test the validity of the SoilData object that gets returned?
+    assert generic == SoilData()
```
morrowcj: Is this `TODO` referring to the `assert generic == SoilData()` line? If so, I think it is good to have all the individual tests (above) because if one of them fails, we know where to look. It is probably a good idea to **also** include the test of overall equality, as you have done.

morrowcj (approved):
This looks really good. I have a few comments that I'd like you to consider addressing, but you are free to merge once you are happy with it. Great work!

matthew7838 (approved):
Looks good.

SC_redesign/Crop_and_Soil/soil/soil.py
Isgotkowitz: I'm gonna miss this comment, it was my first comedic homerun as a software dev

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_soil_data_and_config_factory.py
Isgotkowitz: Thanks, I try:)

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_soil_data_and_config_factory.py
Isgotkowitz: Yes, it is referring to the `assert generic == SoilData()`, I'll update the code comment based on your comment

SC_redesign/Crop_and_Soil/soil/soil.py
Isgotkowitz: I think other methods are going to be appended, to this list, but I am not sure. It depends on how we end up structuring the phosphorus, nitrogen, and carbon modules. I am going to leave it in here for now, but I will add a comment noting this.

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_crop_data_and_species_factory.py
Isgotkowitz: You're welcome:)

SC_redesign/Crop_and_Soil/soil/soil_config_factory.py
Isgotkowitz: Done

SC_redesign/Crop_and_Soil/soil/soil.py
Isgotkowitz: They have now been added into `Field`

SC_redesign/Crop_and_Soil/soil/soil.py
Isgotkowitz: Now it has been added as the only method within `cycle_water()` inside `Field`. Though I think it is likely that this method will go away at some point in the future and we will wind up calling all water methods directly.

SC_redesign/Crop_and_Soil/soil/soil_data.py
Isgotkowitz: I added a method in `Field` that will handle all annual data collection and resetting, but I think we will need to change the design of our `Soil` objects (and `Crop` objects, it has all the same issues). This is because the `SoilData` object inside of the `Soil` object is not actually a field, so there is no way for the `Field` that contains the `Soil` object to read data from or perform operations on the `SoilData` object.

Isgotkowitz: An extremely simple solution that passes all the tests would be to change `data` in `Soil` to `self.data`. This is something that I think will be pretty necessary to analyze before going with it though

Isgotkowitz: And the names for annual reset methods have been changed

SC_redesign/Crop_and_Soil/soil/soil.py
morrowcj: OK, that makes sense. I just wasn't sure.

SC_redesign/Crop_and_Soil/soil/soil_data.py
morrowcj: I agree that `self.data` should be used instead of `data` (not doing so was just an oversight, I think). I don't quite follow your first reply in this thread, though. Why would we need to redesign?

The field should be able to access the soil layer data with `self.soil.data.soil_layers`. Right?

Isgotkowitz: Because `SoilData` wasn't actually an attribute of `Soil`, there was no way to access it through `Soil`. So you're right, the field should be able to access the soil layer data with `self.soil.data.soil_layers` but it currently cannot

Isgotkowitz: Now resolved by #367

SC_redesign/Crop_and_Soil/soil/layer_data.py
ltp23: Style-wise, I think it's better to raise the errors the moment they occur, so in this case, when the variables become negative somewhere else. We can move those error checks inside the post_init() function, or write a class/static method for checking and call it from post_init(). Ideally, we want the code in property methods to be simple return statements.

Isgotkowitz: I will add an issue to address this

PR #362, opened 2023-02-21
Output Manager 2.5 / 3 Changing structure of pools

PooyaHekmati:
This PR modifies a few things, all of which happen under the hood. It modifies the data structure of the pools in the output manager. This is to make it more robust. It is also needed for future PRs. The PR also fixes the issue of the saved JSON files not being valid. Also, it modifies the name of the files into a more readable format.

For the review, I recommend running the code multiple times and seeing the output. Each time, change variables 'exclude_info_maps' (line 313 in output_manager.py), 'max_depth' (line 257 in output_manager.py), and 'indent' (line 259 in output_manager.py).


main.py
```diff
@@ -41,24 +44,45 @@ def execute_simulations_from_files(files: List[Path]) -> None:
...
+        "-v", "--verbose", help="print progress messages", action="store_true"
+    )
     # parser.add_argument("-i", "--interactive", help="run in interactive mode", action="store_true")
 
     return parser.parse_args()
 
 
-if __name__ == '__main__':
+if __name__ == "__main__":
+    # Clear the output directory before running the model
```
JoeWaddell: Can this be (easily) made into an optional argument without conflicting with anything?

RUFAS/output_manager.py
```diff
@@ -232,54 +226,81 @@ def _get_prefix(self, caller_class: str, caller_function: str) -> str:
...
             with open(path, 'w') as json_file:
-                json_file.write(json.dumps(dict))
+                json.dump(Utility.make_serializable(data_dict, max_depth=3),
+                          json_file,
+                          indent=0)
         except Exception as e:
             raise e
 
-    def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
+    def _generate_file_name(self, base_name: str, extension: str = "json"
```
JoeWaddell: Would it be straightforward to add an option for the user to define the prefix on each run?

This might be something better opened in a separate issue for an additional feature (which may not yet be a high priority). It could be potentially useful for testing different inputs, but if the work is more than the current system (whereby a user simply needs to carefully organize/file results between runs) it might not be worth it at this stage.

JoeWaddell (approved):
This is looking great! 

I've left just a couple questions about future development, which may not be relevant at this stage. I've been parsing/exploring the outputs with & without info_maps, so thinking about how we can best ensure stored data is easy to archive and interpret later.

RUFAS/output_manager.py
KFosterReed: I can definitely see how a pre-fix to the output file's name would be a useful addition. Joe's context for how to gauge it's priority is spot on.

KFosterReed (approved):
This looks like a big improvement! Looking forward to the next steps :)

ew3361zh:
> For the review, I recommend running the code multiple times and seeing the output. Each time, change variables 'exclude_info_maps' (line 313 in output_manager.py), 'max_depth' (line 257 in output_manager.py), and 'indent' (line 259 in output_manager.py).

Really nice work overall! Appreciate the variability theses options add to the levels/depths of results but also to the adaptability to handle different kinds of data added to the OM. I think it will allow for really precise and detailed data collection to meet a wide variety of reporting and stakeholder requests/needs.

Output files are much more readable and the indent option adds a great deal as well.

Love all of this - looks good to me.

RUFAS/output_manager.py
```diff
@@ -37,17 +40,37 @@ def __new__(cls):
...
+                     key: str, value: Any, info_map: Dict[str, Any]) -> None:
+        """Adds value and info map at key in the given pool."""
+        key_not_exists_in_pool = pool.get(key) is None
+        if key_not_exists_in_pool:
+            pool[key] = self._pool_element_factory()
+        # reduced_info_map is identical to info_map without the class key and
+        # the function key; as they are already stored in element key and
+        # having them increases the final file size.
+        reduced_info_map = {k: info_map[k]
+                            for k in info_map.keys() - {'class', 'function'}}
```
ew3361zh: great addition here on the reduced_info_map - makes results much clearer and reduces redundancy (and file size, as stated)

ew3361zh:
I did get 5 failed tests in the save_xx section when I changed 'exclude_info_maps' to False:

Screenshot 2023-04-14 at 11 20 28 AM

The same set of tests all pass when I change it back to True.

RUFAS/output_manager.py
PooyaHekmati: I relate to this value as well. This PR is already large with a broad scope. Let's do this in the next PR.

PooyaHekmati:
I added two new gnu args which allow clearing the output path and excluding info maps.

ew3361zh (approved):
New gnu arguments working great. 
All tests were passing with multiple different args combinations.
Looking really sharp - excited about these changes for OM!

JoeWaddell (approved):
Wonderful job. The ability to decide whether or not to delete old results from the command line is great.

PR #364, opened 2023-02-22
Refactor pool_gas_partition.py

matthew7838:
### Summary
This PR refactors the Decomposition sub-module `pool_gas_partition.py` of the Soil module from "pseudocode_soil" S.6.C.1 to S.6.C.13. All methods except the main routine are static methods.

Every method is tested covering possibly most edge cases.

### Changes 

- Except for the main routine `partition`, all other methods are now static
- Except for the main routine `partition`, all methods do not rely on any external classes
- Descriptive names modified and added 
- Now only the main routine `partition `updates and tracks attributes added in the `SoilData` and `LayerData` class

### Other changes
- Added attributes in both the `SoilData` and `LayerData` class 
- Wordings of **above ground, AG, and Below ground, BG**, have been changed to plant and soil

### **NOTE**
Please also review if things match the `pseudo_soil.doc` documents, also see if there could be improvements in naming


SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -103,6 +103,14 @@ class SoilData:
     decomposition_temperature_effect: Optional[float] = None
     """temperature effect on decomposition factor (unitless) (pseudocode_soil S.6.A.1)"""
 
+    # --- pool gas partition
+    silt_clay_content: Optional[float] = None
+    """silt and clay content in the soil (%)"""
```
Isgotkowitz: `layer_data.py` already contains attributes that track the soil and clay contents of each soil layer, so I think this should be replaced with an `@property` method that iterates through the soil layers and calculates a weighted average. I'm not sure what the best factor would be to weigh it by, @morrowcj what do you think?

SC_redesign/Crop_and_Soil/soil/carbon_cycling/pool_gas_partition.py
```diff
@@ -0,0 +1,629 @@
...
+        self.data = soil_data or SoilData()  # initialize with defaults, if not given
+
+    def partition(self):
+        """
+        main routine to update variable in all layers
+
+        Returns: None
+
+        """
+        # S.6.C.2
```
Isgotkowitz:
```suggestion
```
I think it is redundant to put the pseudocode reference above each method call, because they're already referenced in the method description.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/pool_gas_partition.py
```diff
@@ -0,0 +1,629 @@
...
+
+    @staticmethod
+    def _passive_to_active_carbon_amount(passive_carbon_decomposition_amount: float,
+                                         passive_carbon_co2_lost_rate=0.55) -> float:
+        """
+        Calculates passive carbon decomposed into active carbon (kg/ha)
+
+        Args:
+            passive_carbon_decomposition_amount: passive carbon decomposed into active or passive carbon and CO2 (kg/ha)
+            passive_carbon_co2_lost_rate: passive carbon lost as CO2 during decomposition (%)
```
Isgotkowitz: In SWAT they are unfortunately not great at clearly spelling out how they use fractions vs percentages, but in general when something is a percentage 1 = 1% and 100 = 100%, and when something is a fraction 0.01 = 1% and 1.0 = 100%. On top of that each new scientific model I've had to work with seems to do it a different way, but when we write code for RuFaS I think we should try to keep it consistent with SWAT. @morrowcj is this what you have been doing/should we be making this the SC team standard?

SC_redesign/Crop_and_Soil/soil/carbon_cycling/pool_gas_partition.py
```diff
@@ -0,0 +1,629 @@
...
+
+    @staticmethod
+    def _soil_structural_slow_carbon_remaining(soil_structural_slow_carbon_usage: float,
+                                               structural_slow_carbon_loss_rate: float = 0.3) -> float:
+        """
+        Calculates soil structural carbon decomposed to slow carbon after accounting for carbon dioxide loss (kg/ha)
+        Args:
+            soil_structural_slow_carbon_usage: soil structural carbon decomposed into slow carbon (kg/ha)
+            structural_slow_carbon_loss_rate: rate of carbon dioxide loss during transformation of structural to
+                                                                                                            slow carbon
```
Isgotkowitz:
```suggestion
                                                                                                            slow carbon (unitless)
```

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_pool_gas_partition.py
```diff
@@ -0,0 +1,540 @@
...
+from SC_redesign.Crop_and_Soil.soil.layer_data import LayerData
+from SC_redesign.Crop_and_Soil.soil.soil_data import SoilData
+
+
+@pytest.mark.parametrize("soil_structural_slow_carbon_usage", [
+    70,  # lower values
+    150,  # higher values
+    88.8,  # arbitrary
+])
+def test_soil_structural_slow_carbon_remaining(soil_structural_slow_carbon_usage):
```
Isgotkowitz:
```suggestion
def test_soil_structural_slow_carbon_remaining(soil_structural_slow_carbon_usage: float) -> None:
"""Tests that the amount of structural soil carbon decomposed is calculated correctly."""
```
Test method signatures should also be type annotated and have a docstring, but the docstrings can be really simple (No args description or pseudocode reference, just a one sentence test description.)

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_pool_gas_partition.py
```diff
@@ -0,0 +1,540 @@
...
+                field_capacity_water_concentration=1.8,
+                wilting_point_water_concentration=0.8),
+      LayerData(top_depth=20, bottom_depth=22, soil_water_concentration=0.8, field_capacity_water_concentration=1,
+                wilting_point_water_concentration=0.6)])
+])
+def test_partition(layers):
+    data = SoilData(soil_layers=layers)
+    partition = PoolGasPartition(data)
+
+    # S.6.C.1 mocking
```
Isgotkowitz:
```suggestion
```
All of the "# \<pseudocode reference\> mocking" inline comments can be removed.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/pool_gas_partition.py
```diff
@@ -0,0 +1,629 @@
+import math
+from typing import Optional, List
+from SC_redesign.Crop_and_Soil.soil.soil_data import SoilData
+
```
Isgotkowitz, line 4: This module should have a top-level docstring description.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/pool_gas_partition.py
```diff
@@ -0,0 +1,629 @@
...
+                layer.soil_structural_slow_carbon_remaining, layer.active_carbon_to_slow_amount,
+                layer.slow_carbon_decomposition_amount)
+            # aggregate passive carbon pool flux
+            # S.6.C.13
+            layer.passive_carbon_amount = self._soil_passive_carbon_amount(
+                layer.passive_carbon_amount, layer.slow_to_passive_carbon_amount,
+                layer.active_carbon_to_passive_amount, layer.passive_carbon_decomposition_amount)
+
+    @staticmethod
+    def _soil_passive_carbon_amount(passive_carbon_amount: float, slow_to_passive_carbon_amount: float,
```
Isgotkowitz: All method names should start with a verb.

Isgotkowitz (changes requested):
There are a couple of things that need to be addressed across this PR, including unnecessary inline comments, use of percentages vs. fractions, test method signature type annotations and docstrings, and method signature names (all of these are described in more detail in comments). A big problem with this PR though is size, it is way too large. The Code Review section of the wiki says that the maximum size of a PR should be 200 lines, but because we are refactoring entire modules this number is extremely constraining. I have been trying to keep my PRs below 500 lines (although I should be aiming for lower), and once they get bigger split them into stacked PRs (Clay sent me [this ](https://blog.logrocket.com/using-stacked-pull-requests-in-github/) article, its quite helpful).

morrowcj:
Hey, @matthew7838, I agree with @Isgotkowitz about the size of this PR. From now on, try to make them smaller (by breaking the work into pieces). For example, the first few methods (and their tests) could be a PR. Then the next few methods could be a separate PR, and so on. Make sense?

SC_redesign/Crop_and_Soil/soil/carbon_cycling/pool_gas_partition.py
```diff
@@ -0,0 +1,629 @@
+import math
+from typing import Optional, List
+from SC_redesign.Crop_and_Soil.soil.soil_data import SoilData
+
+
+class PoolGasPartition:
+    def __init__(self, soil_data: Optional[SoilData] = None):
+        self.data = soil_data or SoilData()  # initialize with defaults, if not given
+
+    def partition(self):
```
morrowcj: What is this method partitioning?

SC_redesign/Crop_and_Soil/soil/carbon_cycling/pool_gas_partition.py
morrowcj: I think we should keep everything as a fraction (i.e., `passive_carbon_loss_rate: fraction of passive carbon lost as CO2 during decomposition (unitless)`

morrowcj: I've suggested a change to this variable. Carbon and CO2 are redundant and this is the "loss rate" not "lost rate":

```suggestion
            passive_carbon_loss_rate: passive carbon lost as CO2 during decomposition (%)
```

SC_redesign/Crop_and_Soil/soil/carbon_cycling/pool_gas_partition.py
morrowcj: What he said.

SC_redesign/Crop_and_Soil/soil/soil_data.py
morrowcj: Ooh, yes. let's not duplicate attributes.

morrowcj (approved):
I don't have much to add beyond @Isgotkowitz's comments. I agree with all of what he said. So address those items and you're good to go!

SC_redesign/Crop_and_Soil/soil/soil_data.py
matthew7838: @morrowcj Any suggestion on the weight factor?

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -135,3 +233,8 @@ def water_factor(self):
         else:
             return (self.saturation_content - self.soil_water_content) / (
                     self.saturation_content - self.field_capacity_content)
+
+    @property
+    def silt_clay_content(self):
```
Isgotkowitz, line 238:
```suggestion
    def silt_clay_ratio(self):
```
I think this is a better name for this quantity actually represents.

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_pool_gas_partition.py
```diff
@@ -0,0 +1,574 @@
...
+
+    assert PoolGasPartition._determine_plant_active_decompose_carbon.call_count == 3
+    assert PoolGasPartition._determine_soil_active_decompose_carbon.call_count == 3
+    assert PoolGasPartition._determine_soil_active_carbon_amount.call_count == 3
+
+    assert PoolGasPartition._determine_soil_slow_carbon_amount.call_count == 3
+
+    assert PoolGasPartition._determine_soil_passive_carbon_amount.call_count == 3
+
+    # Checking values were set correctly by the main routine
```
Isgotkowitz, line 533:
```suggestion
```

Isgotkowitz (approved):
Looks good. Only comments are for a name change if you feel its necessary and removal of a duplicate comment.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/pool_gas_partition.py
```diff
@@ -0,0 +1,626 @@
+import math
+from typing import Optional, List
+from SC_redesign.Crop_and_Soil.soil.soil_data import SoilData
+
+"""
+This module contains all necessary methods for updating all kinds of carbon in the field, based on the
+pseudocode soil documentation.
```
morrowcj, line 7: Add a reference to the relevant section(s) of the pseudo code

SC_redesign/Crop_and_Soil/soil/carbon_cycling/pool_gas_partition.py
```diff
@@ -0,0 +1,626 @@
...
+"""
+
+
+class PoolGasPartition:
+    def __init__(self, soil_data: Optional[SoilData] = None):
```
morrowcj: Add class docstrings here and elsewhere:

```suggestion
class PoolGasPartition:
    """[Add description of what the class is for]
    
    Attributes
    ----------
    data : SoilData, optional
        a reference to the soil data object, which will be used and updated. If not given, a default SoilData
        object is created.
    """
    def __init__(self, soil_data: Optional[SoilData] = None):
```

SC_redesign/Crop_and_Soil/soil/carbon_cycling/pool_gas_partition.py
```diff
@@ -0,0 +1,626 @@
...
+        self.data = soil_data or SoilData()  # initialize with defaults, if not given
+
+    def partition_pool_gas(self):
+        """
+        main routine to update variable in all layers
+
+        Returns: None
+
+        """
```
morrowcj: Formatting, and be more specific:
```suggestion
        """main routine to update variables [associated with gas partitioning?] in all layers

        Notes
        ------
        This method applies all the gas-partitioning sub routines to each layer of soil that is present in the soil `data` object.
        """
```

Try to add context like this, for any code you write, wherever it might not be clear.

morrowcj (approved):
LGTM, just a few small comments about docstrings.

PR #366, opened 2023-02-24
Refactor animal allocation to pens

ltp23:
This PR focuses on rewriting the method `allocate_all_pens()` in the animal management class. Also, it will be renamed to `allocate_animals_to_pens()`. At first, I wanted to divide the process into several PRs, but then I thought it would probably be harder to check if the code worked as intended, so I did it all in one go. 
It helps that we have decided to no longer use the animal combination None, so a lot of the complexity of the previous implementation can be simplified. For now, the combination Growing-and-Closeup is not being considered yet, though I think that should be fairly easy.


KFosterReed:
Thanks for the explanation Loi! I will take a look at it soon :)

tests/test_animal_management.py
```diff
@@ -476,3 +479,341 @@ def test_get_life_cycle_output():
...
+        cow.milking = i % 2 == 0
+        cows.append(cow)
+
+    # Act
+    dry_cows = AnimalManagement._get_dry_cows(cows)
+
+    # Assert
+    assert len(dry_cows) == num_dry_cows
+    for i in range(num_dry_cows):
+        assert dry_cows[i] == cows[i * 2 + 1]
```
JoeWaddell: This is a wonderfully clever way to set up a test!

RUFAS/routines/animal/animal_management.py
```diff
@@ -491,6 +496,460 @@ def daily_update_id_pen(self, animals_added, ids_removed, calves_born, feed, tem
...
+            pen_type='tiestall',
+            manure_handling='manual_scraping',
+            manure_separator='sedimentation',
+            manure_storage='storage_pit',
+            animal_combination=animal_combination,
+            max_stocking_density=1.2
+        )
+
+    @classmethod
+    def _remove_items_from_dicts_by_id(cls,
```
JoeWaddell: Minor thing, but should this be called `_remove_item_from_dicts_by_id` instead, to indicate that it only works on a single value? Or is the idea that the same ID could correspond to different items inside a given dict?

RUFAS/routines/animal/animal_management.py
```diff
@@ -491,6 +496,460 @@ def daily_update_id_pen(self, animals_added, ids_removed, calves_born, feed, tem
...
+        pens : List[Pen]
+            List of pens to calculate the total maximum stocking density for.
+
+        Returns
+        -------
+        int
+            Total maximum stocking density for the given list of pens.
+
+        """
+        return int(sum(pen.num_stalls * pen.max_stocking_density for pen in pens))
```
JoeWaddell: This method I'm unclear on. It's calculating the total number of animal "spaces" (since the stocking density term adjusts for fewer/more animals than stalls), correct? From the name I assumed it was calculating the max density across pens (to mimic the stocking density values in animal_management_animal.json), but it's used in the other methods as a helper to calculate the number of available animal "spaces" available.

Since this is an part 1 PR, I wanted to check that I'm interpreting the eventual logic correctly.

RUFAS/routines/animal/animal_management.py
```diff
@@ -491,6 +496,460 @@ def daily_update_id_pen(self, animals_added, ids_removed, calves_born, feed, tem
...
+
+        Returns
+        -------
+        Optional[Pen]
+            The pen with the maximum number of stalls that can house the given animal combination.
+            If no pen can house the given animal combination, returns None.
+
+        """
+        max_stalls_found = 0
+        pen: Optional[Pen] = None
```
JoeWaddell: I appreciate how succint this is!

RUFAS/routines/animal/animal_management.py
```diff
@@ -491,6 +496,460 @@ def daily_update_id_pen(self, animals_added, ids_removed, calves_born, feed, tem
...
+            horizontal_dist_to_milking_parlor=1.6,
+            number_of_stalls=number_of_stalls,
+            housing_type='open air barn',
+            bedding_type='straw',
+            pen_type='tiestall',
+            manure_handling='manual_scraping',
+            manure_separator='sedimentation',
+            manure_storage='storage_pit',
+            animal_combination=animal_combination,
+            max_stocking_density=1.2
```
JoeWaddell: In the next steps of this PR, will this (and other values) be populated from the `animal_management_animal.json`?

JoeWaddell (approved):
This is looking nice! I left various questions and comments, most of which I imagine pertain to the next stages of this PR. Great job

RUFAS/routines/animal/animal_management.py
```diff
@@ -491,6 +496,460 @@ def daily_update_id_pen(self, animals_added, ids_removed, calves_born, feed, tem
...
+
+            stall_shortages = cls._update_stall_shortage_for_animal_combination(stall_shortages,
+                                                                                animal_combination_with_max_shortage,
+                                                                                pen)
+            current_max_shortage = max(stall_shortages.values())
+
+        return default_pens_created
+
+    # Note: This function will replace allocate_all_pens eventually.
+    def allocate_all_pens_2(self):
```
KFosterReed: What do you think about renaming this function. The name 'allocate all pens' implies that we are distributing the pens to the animals. But really we are putting animals in pens. My other ideas so far are: assign_all_animals_to_pens,  put_all_animals_in_pens, sort_animals_into_pens

RUFAS/routines/animal/animal_management.py
KFosterReed: I agree with Joe's suggestion. Perhaps a better name is 'calculate_animal_spaces'?

KFosterReed (changes requested):
This PR is definitely on the right track as far as separating the different functions that allocate_all_pens is doing and improving the logic and code clarity. After finally giving it enough time to figure out what was going on with the mixed-type pens, I don't think this is actually achieving what I was hoping to achieve with the mixed type pens. Here is a link to a diagram of how I was hoping the animal classes would be sorted into animal combinations for putting into pens [https://whimsical.com/pen-grouping-2TDjBuTYGVDEzaDGyoJRFo](url)

In scenario 1, gorwing and close up animals are housed and fed in different pens. In scenario 2, growing and close up animals are housed and fed together. However, after reflection on this, I think we should get rid of the 'mixed type pen' option all together for now and just work towards scenario 1. I will put together some more diagrams/documentation related to pen grouping for future grouping scenarios that we would like to add in the future but in the spirit of getting an MVP out there, I think the mixed pens adds a bunch of complexity without giving us much functional benefit.

RUFAS/routines/animal/animal_management.py
ltp23: The plural `items` means the same pen id can be present in more than one dictionary. Two in this context.

RUFAS/routines/animal/animal_management.py
ltp23: The name change should be easy.

Just to double-check the logic of this part, I'm trying to calculate the animal space shortage.

Let's use calves as an example. There are 100 calves, and currently, 4 pens are designated for holding those calves. Each of these pens has 10 stalls and a max stocking density of 1.2.

So we are, 100 - 4 x 10 x 1.2 = 100 - 48 = 52 animal spaces short. 

Because I followed the code from the previous implementation, the math should still be correct, but I will double-check. The flaw in my understanding has been that I thought there is a one-to-one correspondence between an animal and a stall, without taking into account of the maximum stocking density. That adds some extra wiggle room.

RUFAS/routines/animal/animal_management.py
ltp23: Good suggestion. I'll look to see if there is config data inside the animal management class. I can add it to the next PR.

RUFAS/routines/animal/animal_management.py
ltp23: Yes, that makes sense. I'll pick the first one. Actually, I've been working on the second PR and using a similar name for another function. I'll make adjustments so all the names don't mean the same thing.

ltp23:
Thanks Kristan. 

Perhaps the name `mixed type` is somewhat confusing and/or misleading. It is something like `unclassified` or `unspecialized`.

There are two kinds of mixed-type pens - one with an animal combination of NONE and the other with GROWING_AND_CLOSE_UP. 

Then during the allocation process, these act as backup pens in case we need more animal spaces, so we can switch them to a more specific combination (calf, growing, close-up, lactating).

Also, the use of mixed-type pens is only limited to the allocation process, so I think if we decide to make changes to the design, it shouldn't affect other places much.

KFosterReed:
> 

@ltp23 the intended use of a Growing_and_Closeup pen is not that it be backup but that it be used on farms that would intentionally house them all together as in the scenario 2 in the whimsical diagram. Thus, we would have animals of classes Heifer1, HeiferII, HeiferIII and dry cows all in the same pen - intentionally - and the AnimalCombination associated with that pen would remain as Growing_and_closeup

KFosterReed (commented):
I made one small change to the is_far_off_dry function. We don't need a day when the far off dry period starts because that automatically starts when they stop lactating. Thus, when they are dry and their pregnancy is less than the start of the close-up period then they are already in the far-off dry period :)

Other than that, I think this looks good. I am looking forward to seeing the next step so that we can actually use the new function

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -135,6 +137,78 @@ def __init__(self, args):
...
+            True if the cow is in the far-off dry state, False otherwise.
+
+        """
+
+        return (self.is_pregnant and self.is_dry and
+                (self.days_in_preg < AnimalModuleConstants.DRY_CLOSE_UP_START_DATE))
+
+    # TODO: Not used yet & check for correctness
+    @property
+    def is_close_up_dry(self):
```
Fjordrunner, line 198: Maybe is_close_up_or_dry might be a better function name to distinguish that there are two animal types?

RUFAS/routines/animal/animal_management.py
```diff
@@ -491,6 +504,467 @@ def daily_update_id_pen(self, animals_added, ids_removed, calves_born, feed, tem
...
+        Returns
+        -------
+        int
+            The shortage of animal spaces. If there is no shortage, the result will be 0.
+
+        """
+        max_animal_spaces = 0
+        for pen in pens:
+            max_animal_spaces += cls._calc_max_animal_spaces_per_pen(pen.num_stalls, pen.max_stocking_density)
+        return num_animals - max_animal_spaces
```
Fjordrunner, line 625: Wouldn't this return negative numbers at times? Is that the desired functionality here, or am I misunderstanding something? It's totally fine if I am but. Also, doesn't "shortage" imply that we don't have enough of something? Is the name still appropriate, from an animal science perspective @JoeWaddell @KFosterReed ?

RUFAS/routines/animal/animal_management.py
```diff
@@ -491,6 +504,467 @@ def daily_update_id_pen(self, animals_added, ids_removed, calves_born, feed, tem
...
+            raise ValueError("The number of animals cannot exceed the total number of spaces.")
+
+        num_animals_in_pens = [0] * num_pens
+        allocation_limits = [math.ceil(overall_density * max_spaces) for max_spaces in max_spaces_in_pens]
+        # Sort pens by allocation limit, then by index
+        sorted_pen_indices = sorted(range(num_pens), key=lambda pen_idx: (allocation_limits[pen_idx], pen_idx))
+
+        for i in sorted_pen_indices[:num_pens - 1]:
+            num_animals_to_allocate = min(num_animals, allocation_limits[i])
+            num_animals_in_pens[i] += num_animals_to_allocate
```
Fjordrunner: Is this line correct if there are already animals in the pen, since the variable num_animals_in_pens starts out as a list of all 0s, or is num_animals_in_pen a variable that is to be used exclusively in reference to animals that are being allocated, not animals as a whole? In that case, it might be worth renaming that variable.

tests/test_animal_allocation.py
Fjordrunner, line 1: Is there a reason that we are splitting the animal allocation logic's tests to their own file? I don't have an issue with it at all, was just curious.

Fjordrunner (approved):
Overall this seems like a massive improvement over the former version of the animal allocation logic. I think that the addition of several helper functions is definitely a change in the right direction, but depending on whether or not the functions that we relied on before are correct or not, I can see us still having some pen allocation issues in the future. LGTM for now!

RUFAS/routines/animal/animal_management.py
ltp23: Yes, if there is no shortage, the return value will be either 0 or negative. Then, in the main allocation function, we would check if the shortage is positive or not. I'll update the docstring for this.

RUFAS/routines/animal/animal_management.py
ltp23: The variable `num_animals_in_pens` is a local variable here. Perhaps the `+=` is confusing and implies adding to some existing number, so I'll change it to a simple assignment with `=` instead. That variable is simply a list of numbers that correspond to the numbers of animals we want to put in the pens. This function is about 'planning' only, and the actual moving of the animals is reserved for the other 'executing' function.

tests/test_animal_allocation.py
ltp23: I moved the tests to a separate file mostly because they were quite long as a whole. I'm ok with moving them back to the test_animal_management.py too.

ltp23:
One extra thing of note that I haven't mentioned in the description that is related to what animals would go to what pens when there are multiple animal types and pens of the same type:

It would be easier for me to explain with an example using some concrete numbers. 
Let say I have two GROWING pens, and I want to allocate a list of heiferI's and heiferII's to those pens. 

number of animals to be allocated in the 2 pens = [40, 60]
list of 50 heiferI's (id: 0-49)
list of 50 heiferII's (id: 50-99)

Current approach: Each animal list will be considered in order, one after another. 
Pen 1 will only get heiferI's (id: 0-39), and pen 2 will get the rest, i.e., heiferI's (id: 40-49) and heiferII's (id: 50-99)

An alternative approach (not implemented yet): 
Since the ratio of heiferI's to heiferII's is 1:1, we can preserve this ratio in each pen 
Pen 1 will get 20 heiferI's (id: 0-19) and 20 heiferII's (id: 50-69), and Pen 2 will get the rest of the animals. If there are more pens, the same principle still applies. 

Regardless of which approach we choose, we can also shuffle each animal's list before any allocation, so the animals don't look like they always stay together. Please let me know if this is a desired feature.

JoeWaddell:
> Regardless of which approach we choose, we can also shuffle each animal's list before any allocation, so the animals don't look like they always stay together. Please let me know if this is a desired feature.

@ltp23  My instinct is that the first option is better, in part because it is what's already implemented, but also because it potentially makes ration formulation more accurate, since more similar animals are grouped together (and have more similar requirements). However, @KFosterReed and others can comment as to whether shuffling the animals would be more a realistic scenario (or if they are indeed usually grouped by size/age...though I suspect each farm has its own quirks).

RUFAS/routines/animal/animal_management.py
JoeWaddell: I think changing from += to = is good, since this part also threw me. I see how it's 0 prior to assignment, but it might be more immediately clear what it's doing with your suggested change.

RUFAS/routines/animal/animal_management.py
```diff
@@ -491,6 +504,467 @@ def daily_update_id_pen(self, animals_added, ids_removed, calves_born, feed, tem
...
+        [45, 27, 18]
+
+        >>> AnimalManagement.plan_animal_allocation(num_animals=70, max_spaces_in_pens=[50, 30, 20])
+        [35, 21, 14]
+
+        >>> AnimalManagement.plan_animal_allocation(num_animals=47, max_spaces_in_pens=[50, 30, 20])
+        [22, 15, 10]
+
+        """
+        num_pens = len(max_spaces_in_pens)
```
JoeWaddell: This might be overkill, but I think the variable name `num_pens` could instead be something like `pens_available_for_combination`. That's perhaps too long, but drawing the distinction between the total number of pens and the number of pens available for the specific animal combination might help a more general audience when later reading the code.

RUFAS/routines/animal/animal_management.py
```diff
@@ -491,6 +504,467 @@ def daily_update_id_pen(self, animals_added, ids_removed, calves_born, feed, tem
...
+
+        for i, count in enumerate(allocation_plan):
+            animal_combination = animal_pens[i].animal_combination
+            animal_pens[i].update_animals(animals[:count], animal_combination)
+            animals = animals[count:]
+
+    @classmethod
+    def plan_animal_allocation(cls, num_animals: int, max_spaces_in_pens: List[int]) -> List[int]:
+        """
+        Make an allocation plan to move animals to pens and match pen density as closely as possible
```
JoeWaddell, line 851: This whole method is wonderful! You've done a nice job explaining the process in the docstring, and the examples make it crystal clear what's going on.

RUFAS/routines/animal/animal_management.py
```diff
@@ -491,6 +504,467 @@ def daily_update_id_pen(self, animals_added, ids_removed, calves_born, feed, tem
...
+        Returns
+        -------
+        None
+
+        """
+
+        self.pens_by_animal_combination = self._group_pens_by_animal_combination(self.all_pens)
+
+        # For now, we are only considering the following animal combinations:
+        animals_by_combination = {
```
JoeWaddell, line 948: This is great, and if I follow the logic, I see how to add new groups in the future. Also: this means that animal classes must be exclusive to a single AnimalCombination group, correct?

KFosterReed:
> > Regardless of which approach we choose, we can also shuffle each animal's list before any allocation, so the animals don't look like they always stay together. Please let me know if this is a desired feature.
> 
> @ltp23 My instinct is that the first option is better, in part because it is what's already implemented, but also because it potentially makes ration formulation more accurate, since more similar animals are grouped together (and have more similar requirements). However, @KFosterReed and others can comment as to whether shuffling the animals would be more a realistic scenario (or if they are indeed usually grouped by size/age...though I suspect each farm has its own quirks).

I agree that the first option is preferred since farms will most often group animals based on their age so lets leave it as is. Great work @ltp23 :)

KFosterReed:
Hi Team, I have read through the code updates and it looks great and makes a lot of sense! The only thing stopping me from approving is that I haven't found time to run it yet. If others have run the simulation and it seems like it is working then let's go ahead and merge.

KFosterReed (approved):

RUFAS/routines/animal/animal_management.py
ltp23: Yes Joe, that's the key thing: we want the animal combinations we specify in the input json to be both exhaustive (i.e., covering all the animal types), and non-overlapping (i.e., each animal type is associated with only one animal combination). Since we don't have a mechanism to validate the input yet, we would need to check the combinations ourselves.

Issue #371, opened 2023-03-01
`field/` module is full of files with misformatted docstrings


Isgotkowitz:
# Problem
Most (if not all) files that have been created do not adhere to the style guides as linked in the README (https://realpython.com/documenting-python-code/ and https://numpydoc.readthedocs.io/en/latest/format.html). This is a pretty significant problem that could really bog the process of merging `SC_redesign` into `master` when the soil and crop redesign has been completed.
# Solution
This issue should be addressed with a series of PR's that reformat the comments of groups of files at a time. The size of these sets and the best way to group them should be discussed. I think reformatting 1 - 2 files at a time would make this process drag on for too long, and reviewing entire sub-modules at a time (`soil`, `crop`, `field`, `tests_SC`) would result in PRs that are too large and broad.
# Strategy
- Start executing these PRs as soon as possible, in tandem with starting new refactors of old code. The reason for this is that someone who has just corrected the comments in an older file will know exactly what the comments and docstrings should look like and will be much less likely to make the same mistakes while working on a new file.
- Regularly have people from outside the Soil and Crop team participate in the review process, so that we know the reformatted files are meeting the style standards of the entire RuFaS team.

morrowcj:
@Isgotkowitz, I don't agree that this is a priority at the moment. We are trying to finish the redesign of this module soon and our docstring format fully complies with the instructions we were given when we started this redesign. 

We were originally told to follow the [google style guide](https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html), which allows for exactly the kind of dosctrings we are using *consistently* throughout the module. We recently switched to using [numpy style](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard). When I asked, I was told this was the same as google style - I now know that not to be correct. 

Changes to the numpy format can occur **after** the redesign is complete.

PooyaHekmati:
>When I asked, I was told this was the same as google style - I now know that not to be correct.

Sphinx engine is capable of handling both styles, so they are the same from that perspective. If by being "the same" you mean they are equal char for char; then it means the same thing has two different names which in turn means Google and NumPy are both claiming to own it. I don't think that makes much sense.

[...]

PR #374, opened 2023-03-06
Implement and test `Fertilizer` module within `soil.PhosphorusCycling` module

Isgotkowitz:
# Summary
This PR implements the sub-sections "Rain-Induced Phosphorus Release from Fertilizer" and "Distribution of Released Fertilizer P between Infiltration and Runoff" from Pete Vadas' SurPhos model (the sub-sections listed above are from the theoretical documentation). The bulk of the implementation is in the new `Fertilizer` module, housed in `fertilizer.py'. 

# Main Changes
- `fertilizer.py`
  - This file contains all necessary methods for adding phosphorus via surface-applied fertilizer to a field, and removing that phosphorus by either adding it to the soil profile or removing it from the field with runoff.
- `test_fertilizer.py`
  - This file contains tests for each method in the `Fertilizer` module.
- `soil_data.py`
  - New attributes have been added (under the `# ---- Fertilizer` comment) to track fertilizer phosphorus-related pools and counters.
  -  Two new `@property` methods have been added to at the bottom of the file to specify attributes of `SoilData` that based on other fertilizer phosphorus-related attributes.

# Other Changes
- `crop_and_soil_constants.py` has been added as a place to store global constants, mainly for unit conversion. The list of best practices within it should be checked.
- The `phosphorus_cycling/` and `phosphorus_cycling_tests/` directories have had `__init__.py` files added to both to make them modules.

# Notes
- This file uses Google-style docstrings, not NumPy. Please ignore their style for now, all docstrings will be updated to the correct format at a later point in time. 
- All non-empty files add/changed in this PR have been manually linted with `flake8`.
  - To lint them yourself from the console, download `flake8`, and lint with the command 
  `flake8 --count --select=E501,E9,F63,F7,F82 --show-source -- statistics --max-line-length=120 path/to/file`
  - Alternatively, in your local copy of the branch, you can create a `.flake8` file that contains
    ```
    [flake8]
    count = True
    select =
        E501,
        E9,
        F63,
        F7,
        F82
    show-source = True
    statistics = True
    max-line-length = 120
    ```
    and then run the flake command with no flags.


SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/fertilizer.py
```diff
@@ -0,0 +1,240 @@
+from typing import Optional
+from math import log, exp
+
+from SC_redesign.Crop_and_Soil.soil.soil_data import SoilData
+
+"""
+This module incorporates equations from the SurPhos model to simulate the Phosphorus from fertilizer applied to the soil
```
morrowcj: Small nitpick: we're not simulating phosphorus, but some process related to phosphorus.

```suggestion
This module incorporates equations from the SurPhos model to simulate the [cycling of?] Phosphorus from fertilizer applied to the soil
```

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/fertilizer.py
```diff
@@ -0,0 +1,240 @@
...
+        first_rainfall_occurred = self.data.rain_events_counter == 1 and self.data.available_phosphorus_pool > 0
+
+        if first_rainfall_occurred and runoff == 0:
+            self._add_to_labile_phosphorus(self.data.available_phosphorus_pool, field_size)
+            self.data.available_phosphorus_pool = 0
+            self.data.days_since_application += 1
+            return
+        elif first_rainfall_occurred and runoff > 0:
+            self._leach_available_phosphorus(rainfall, runoff, field_size)
+            self.data.days_since_application += 1
```
morrowcj: @Isgotkowitz, I have a few thoughts about this method that we should maybe think about together:
1. I wonder if it should be kept as two separate methods. The first would be something like `add_phosphorus` that simply adds phosphorus to the main pool - and update counters - and is only called on the fertilization day (currently `FieldData.is_amendment_day`, but should probably be split into something like `add_fertilizer_today` and `add_manure_today`). The second would be called daily and cycle the phosphorus based on the counters.
2. I think that weather events (`rain_events_counter`) should always be tracked by the field manager (perhaps through a weather tracking component class).
3. I think the logic may not be quite exhaustive. What happens when it rains on application day, for instance?
4. The if statements can be simplified a bit, I think. Here's an example that maintains your logic, starting after the rainfall counter:
```python
        # increase temporal counter (always)
        self.data.days_since_application += 1

        # do nothing if there's no P
        if self.data.available_phosphorus_pool <= 0:
            return

        # do nothing if P fert was added today AND no rain occurred today. (What happens when it rains on app. day?)
        if add_fertilizer_today and self.data.rain_events_counter <= 0:
            return

        # End early if *more than* 1 rain event has occurred
        if self.data.rain_events_counter > 1:
            return

        # Absorb P, and do nothing else, if no rain has occurred
        if self.data.rain_events_counter <= 0:
            self._absorb_phosphorus(field_size)
            return  # end after absorption

        # The rest of the code pertains to exactly one rainfall event
        if runoff > 0:
            self._leach_available_phosphorus(rainfall, runoff, field_size)
        else:
            self._add_to_labile_phosphorus(self.data.available_phosphorus_pool, field_size)
            self.data.available_phosphorus_pool = 0
```

And the first few lines could be even further simplified:
```python
        ## early return conditions
        no_phosphorus = self.data.available_phosphorus_pool <= 0
        clear_app_day = add_fertilizer_today and self.data.rain_events_counter <= 0
        more_than_one_rain = self.data.rain_events_counter > 1

        if no_phosphorus or clear_app_day or more_than_one_rain:
            return
```

Isgotkowitz: 1. Two separate methods is a very good idea, I will change a few things right now and it will be taken care of

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/fertilizer.py
Isgotkowitz: I have updated the top-level docstring, is much more descriptive

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/fertilizer.py
Isgotkowitz: 2. I disagree, I think it would be better to keep `rain_event_counter` grouped with the other `SoilData` attributes that are used only by `Fertilizer`. Putting it somewhere else would possibly mean that we have to update it somewhere else, which would be much more complex and error-prone. We should discuss this further, and also I will change the name to make it clear that it is associated with `Fertilizer` and is not some generic rain counter

Isgotkowitz: 3. I agree, and making sure the logic is exhaustive in this module is going to be more challenging than previous ones I think. Maybe writing two helper functions, one that manages all operations before and on the day of the first rain event, and one that handles all operations after the first rain event, and letting the top-level method decide which one to run would make the logic easier to check?

    To respond to the example you brought up, I think it is covered if the phosphorus addition and management functions are called in the correct order. The field manager would add fertilizer phosphorus via the addition method (setting `days_since_application` to 0), do some stuff, then call the daily fertilizer phosphorus routine. The first thing the routine does is update the rainfall event counter (in this case to 1), so `no_phosphorus_absorbed` and `phosphorus_absorbed_only` are false and `first_rainfall_occurred` is true. The program will then empty the available phosphorus pool.

Isgotkowitz: 4. I agree that the `if` statements are not as clean as they could be, but simplifying them is going to be a bit of a complex process.

morrowcj: > Maybe writing two helper functions, one that manages all operations before and on the day of the first rain event, and one that handles all operations after the first rain event, and letting the top-level method decide which one to run would make the logic easier to check?

I think that's a good idea. Thanks.

SC_redesign/Crop_and_Soil/crop_and_soil_constants.py
```diff
@@ -0,0 +1,20 @@
+"""
+Description:
+    These constants are globals that should be used for unit conversion.
+
+Details:
+    Please follow the proceeding best practices:
+        1. Write names of constants in all caps, with words separated by underscores.
+        2. Use the full name of units, not abbreviations. This avoids confusion in cases where case matters (for
+            example, megagrams (Mg) and milligrams (mg).)
+        3. Do not include the constants for reverse conversion directions.
```
morrowcj:
```suggestion
        3. For conversion constants, do not include the value for reverse conversion directions. 
            To convert in the opposite direction, simply divide instead of multiply.
```

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/fertilizer.py
```diff
@@ -0,0 +1,258 @@
...
+
+    def do_fertilizer_phosphorus_operations(self, rainfall: float, runoff: float, field_size: float) -> None:
+        """The main routine that updates phosphorus in surface-applied fertilizer on a daily basis.
+
+        Args:
+            rainfall: amount of rainfall on this day (mm)
+            runoff: amount of runoff on this day (mm)
+            field_size: size of the field (ha)
+
+        """
```
morrowcj, line 27: Should we start using Numpy style here? I think this is the way:

```suggestion
        """The main routine that updates phosphorus in surface-applied fertilizer on a daily basis.

        Parameters
        -------------
        rainfall : float
            amount of rainfall on this day (mm) 
        runoff : float 
            amount of runoff on this day (mm)
        field_size : float
            size of the field (ha)
        """
```

I'll just go ahead and say it: I don't like this. It is not efficient to have to write the type of the argument when that information already exists in the signature. Other sphinx types do this automatically but it seems Numpy style needs it explicitly (although type hints are not included in Numpy method signatures according to [this example doc](https://numpydoc.readthedocs.io/en/latest/example.html#example)).

morrowcj (approved):
This looks good to me. My only comment involves use of Numpy docstrings, but I think that is a large enough change that it would warrant its own PR (i.e., when we go back to rewrite the rest of the docstrings for already completed modules). But I guess, try to use the correct style for all future modules.

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/fertilizer.py
Isgotkowitz: Yeah I'm pretty sure NumPy style also likes optional or defaults specified after the type, which is yet another thing to repeat from the signature. All new modules that I start will use NumPy style; this one was already started with Google style and I didn't want to make it a hodgepodge of different styles.

morrowcj: Agreed

Isgotkowitz:
> This looks good to me. My only comment involves use of Numpy docstrings, but I think that is a large enough change that it would warrant its own PR (i.e., when we go back to rewrite the rest of the docstrings for already completed modules). But I guess, try to use the correct style for all future modules.

I think it is gonna warrant a lot of PRs, but we shouldn't be able to forget about it cause issue #371 is on the books. Also I will be using NumPy style in all new files that I start now.

SC_redesign/Crop_and_Soil/crop_and_soil_constants.py
```diff
@@ -0,0 +1,21 @@
...
+    Please follow the proceeding best practices:
+        1. Write names of constants in all caps, with words separated by underscores.
+        2. Use the full name of units, not abbreviations. This avoids confusion in cases where case matters (for
+            example, megagrams (Mg) and milligrams (mg).)
+        3. For conversion constants, do not include the value for reverse conversion directions. 
+            To convert in the opposite direction, simply divide instead of multiply.
+        4. Prefer constants that are greater than one.
+"""
+# Area
+HECTARES_TO_SQUARE_MILLIMETERS = 10000000000
```
ltp23, line 15: As a visual aid for writing big numbers like this, the underscore separator can be used. 

```
HECTARES_TO_SQUARE_MILLIMETERS = 10_000_000_000
```

SC_redesign/Crop_and_Soil/crop_and_soil_constants.py
```diff
@@ -0,0 +1,21 @@
...
+Description:
+    These constants are globals that should be used for unit conversion.
+
+Details:
+    Please follow the proceeding best practices:
+        1. Write names of constants in all caps, with words separated by underscores.
+        2. Use the full name of units, not abbreviations. This avoids confusion in cases where case matters (for
+            example, megagrams (Mg) and milligrams (mg).)
+        3. For conversion constants, do not include the value for reverse conversion directions. 
+            To convert in the opposite direction, simply divide instead of multiply.
```
ltp23: This point may go either way. For me, I think it's more natural to use multiplication most of the time, so including the reverse constants is not a bad idea. You can express one in terms of the other so there is not much duplication really. 

```
KM_TO_M = 1000
M_TO_KM = 1 / KM_TO_M

distance_in_m = distance_in_km * KM_TO_M
distance_in_km = distance_in_m * M_TO_KM  
```

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/phosphorus_cycling_tests/test_fertilizer.py
```diff
@@ -0,0 +1,289 @@
...
+        determined correctly."""
+    observe = Fertilizer._determine_phosphorus_distribution_factor(rainfall, runoff)
+    expect = 0.034 * exp(3.4 * (runoff / rainfall))
+    assert observe == expect
+
+
+@pytest.mark.parametrize("phosphorus,frac_released,distribution_factor,total_rainfall", [
+    (10000000, 1, 0.28347283947, 12300),
+    (1295043, 0.4, 0.238947265, 12320),
+    (100421, 0.075, 0.3485781, 9183),
```
ltp23, line 42: I would say in general it is not necessary to use logical numbers. We are just testing a math expression. Using nicer numbers that cover enough edge cases is usually sufficient.

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/phosphorus_cycling_tests/test_fertilizer.py
```diff
@@ -0,0 +1,289 @@
...
+    if rainfall_events == 1:
+        data.available_phosphorus_pool = pool_amount
+    else:
+        data.recalcitrant_phosphorus_pool = pool_amount
+    fert = Fertilizer(data)
+
+    fert._determine_phosphorus_distribution_factor = MagicMock(return_value=0.05)
+    fert._determine_dissolved_phosphorus_concentration = MagicMock(return_value=0.05)
+
+    pool_amount_mg = pool_amount * 1000000
```
ltp23, line 127: It's ok to just use a named constant here instead of a hard-coded value, assuming you have tested it somewhere else.

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/phosphorus_cycling_tests/test_fertilizer.py
```diff
@@ -0,0 +1,289 @@
...
+                      "absorbed_phosphorus": (0.5 * (recalcitrant_pool * fert.data.solubilizing_factor))})
+
+    fert._update_after_first_rain(rainfall, runoff, field_size)
+
+    if not rainfall:
+        assert fert._add_to_labile_phosphorus.call_count == 0
+        assert fert._determine_leached_phosphorus.call_count == 0
+        assert fert.data.recalcitrant_phosphorus_pool == recalcitrant_pool
+    elif rainfall and not runoff:
+        fert._add_to_labile_phosphorus.assert_called_with(recalcitrant_pool * fert.data.solubilizing_factor, field_size)
```
ltp23, line 214: Just an FYI, there is a more exact version - `assert_called_once_with()`.

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/phosphorus_cycling_tests/test_fertilizer.py
```diff
@@ -0,0 +1,289 @@
...
+    (0, 5, 13, 0, 4),  # First rainfall, after day of application
+    (2, 3, 17, 4, 1.8),  # Not first rainfall, after day of application
+    (3, 8, 0, 0, 2.1),    # No rainfall, some rain events, after day of application
+])
+def test_do_fertilizer_phosphorus_operations(rain_events: int, days_since_application: int, rainfall: float,
+                                             runoff: float, field_size: float) -> None:
+    """Tests that correct action is taken on fertilizer phosphorus in the system, and that temporal counters are
+        incremented correctly.
+    """
+    data = SoilData(rain_events_after_fertilizer_application=rain_events, days_since_application=days_since_application)
```
ltp23, line 272: I think in general you should mock objects like SoilData here and make direct assignments. Even though you pass in those arguments, what if in the init function, there are some behind-the-scenes calculations going on, you may get unexpected numbers. This puts an extra burden on you for having to know more than what is needed to create a SoilData object just for testing purposes.

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/phosphorus_cycling_tests/test_fertilizer.py
```diff
@@ -0,0 +1,289 @@
...
+
+    fert.do_fertilizer_phosphorus_operations(rainfall, runoff, field_size)
+
+    assert fert.data.days_since_application == days_since_application + 1
+    if rainfall:
+        assert fert.data.rain_events_after_fertilizer_application == rain_events + 1
+        rain_events += 1
+    if rain_events == 0 or (rainfall and rain_events == 1):
+        fert._update_before_and_at_first_rain.assert_called_with(rainfall, runoff, field_size)
+        assert fert._update_after_first_rain.call_count == 0
```
ltp23, line 286: An alternative is 
```
fert._update_after_first_rain.assert_not_called() 
```

ltp23 (approved):
Overall, everything looks good! Great job, Ed!

SC_redesign/Crop_and_Soil/crop_and_soil_constants.py
Isgotkowitz: That is very useful to know

SC_redesign/Crop_and_Soil/crop_and_soil_constants.py
Isgotkowitz: Ok

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/phosphorus_cycling_tests/test_fertilizer.py
Isgotkowitz: Ok, I will keep that in mind

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/phosphorus_cycling_tests/test_fertilizer.py
Isgotkowitz: Ok, I had not though to mock objects, but going forward I will do so because you make a very good point.

PR #377, opened 2023-03-08
Crop planting methods

morrowcj:
Implement methods for specifying, configuring, and planting crops in a field.

## Notes
* some crop management methods were removed from `Crop` and moved to `Field`.


SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -255,13 +311,13 @@ def assess_dormancy(self, daylength: float) -> None:
             daylength: length of time from sunup to sundown on the current day (hours)
 
         Details:
-            If the length of the current day is at or below the dormancy threshold length, all crops that can go dormant 
-            should be put into dormancy. If the length is greater than the greater than the threshold length, all crops should
-            be brought out of dormancy.
-            
+            If the length of the current day is at or below the dormancy threshold length, all crops that can go dormant
+            should be put into dormancy. If the length is greater than the greater than the threshold length, all crops
```
```suggestion
            should be put into dormancy. If the length is greater than the threshold length, all crops
```

Isgotkowitz (changes requested):
Looks really good. Particularly liked the way you organized `make_crop_from_config_dict()`, `make_supported_crop()`, and `make_custom_crop()`. Only found one typo in a docstring that needs to be changed

SC_redesign/Crop_and_Soil/tests_SC/field_tests/test_field.py
```diff
@@ -39,3 +29,104 @@ def test_start_dormancy(daylength: float, threshold_daylength: float) -> None:
...
+    new_field.add_crop(Crop(), 0.33)
+    assert new_field.crops[0].data.field_proportion == 0.10
+    assert new_field.crops[1].data.field_proportion == 0.20
+    assert new_field.crops[2].data.field_proportion == 0.33
+
+    # --- failing cases ---
+    newer_field = Field()
+    with pytest.raises(Exception):
+        newer_field.add_crop(Crop(), 1.1)  # over 1
+        new_field.add_crop(Crop(), 0.5)  # total is over 1
```
Isgotkowitz, line 96: These test cases are not all being properly run. pytest does not make it to 
```
new_field.add_crop(Crop(), 0.5)
```
I suggest having only the "over 1" case in the first `with` statement, and adding a second `with` statement to test the "total is over 1" case.

Isgotkowitz (changes requested):
I found that one of the cases in `test_add_crop()` that is designed to fail is not being reached. Also there are about 8 issues that `flake8` raises when run on the files that have been changed for this PR.

SC_redesign/Crop_and_Soil/tests_SC/field_tests/test_field.py
morrowcj: Thanks for pointing this out! I had no idea that only one error would be recognized by `pytest.raises` I think I've done this in a few places. I'll change that.

morrowcj:
> Also there are about 8 issues that `flake8` raises when run on the files that have been changed for this PR.

@Isgotkowitz, this is super annoying since I must've spent like an hour and a half on linting. Haha

Isgotkowitz:
> > Also there are about 8 issues that `flake8` raises when run on the files that have been changed for this PR.
> 
> @Isgotkowitz, this is super annoying since I must've spent like an hour and a half on linting. Haha

Yeah I think it highly variable because of how configurable it is (plus there isn't a clear teamwide standard yet to make things), I will share my configuration so it is reproducible

Isgotkowitz:
# Configuring and running `flake8`
This is how I have been configuring and running `flake8` on new code to make sure it not adding to the list of pre-existing linter errors that the GitHub Action linting test finds.

# Configuration
Add a file to the top level directory of your project folder called `.flake8` containing the following lines:
```
[flake8]
count = True
show-source = True
statistics = True
max-line-length = 120
max-complexity = 10
```
This is a mash up of different flags found in the two different `flake8` calls that GitHub runs automatically on new code before it can be merged (See `.github/workflows/lint_flake8.yml`)

# Script
Running `flake8` on the entirety of RuFaS makes it so much harder to sort out which errors are actually in the files that you modified. The following script will run `flake8` only on the files that are different between the branch you are working on and the one you are going to merge your changes into.
Copy the following into a script called `runflake8onbranch.sh` in the top-level directory of your project (same place as `.flake8`):
```
#!/bin/bash
if [ $# -eq 0 ]; then
    base_branch="master"
else
    base_branch=$1
fi

flake8 $(git diff --name-only $(git merge-base ${base_branch} HEAD))
```
Then run the command `chmod +x runflake8onbranch.sh`
You should now be able to run the script with the command `./runflake8onbranch.sh`
Note: by default, it assumes the branch you will merge into is `master`, and currently something in `master` makes `flake8` crash. To run it on a different base branch, add the name as an argument to the script.
For example, for this PR, run `./runflake8onbranch.sh origin/field-manager_SCred`

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -158,28 +154,105 @@ def check_schedule(self, year: int, day: int) -> None:
...
     # <editor-fold desc="--- Crop Management Methods ---">
-    def plant_crops(self, current_crop_config: CropPlantingConfig) -> None:
-        """plant the current crop(s) into the filed"""
-        for sp in current_crop_config.species:
-            self.add_crop(sp)
-        pass
+    def plant_crops(self, crops_config: List[Dict], coverage: Optional[List[float]] = None) -> None:
+        """adds all crop(s) into the field from the current configuration specs
+
+        Args:
```
ew3361zh, line 160: I believe with the numpy documentation style we're using "Parameters" in place of "Args" for docstrings with this style:

Parameters
----------
x : type
    Description of parameter `x`.
y
    Description of parameter `y` (with type not specified).

https://numpydoc.readthedocs.io/en/latest/format.html#parameters

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -158,28 +154,105 @@ def check_schedule(self, year: int, day: int) -> None:
...
-    def add_crop(self, species: str, field_cover: float = 1.0,
-                 priority: int = 1) -> None:
+        Args:
+            config: a dictionary containing specifications for the crop to be initialized.
+
+        Details: if the "species" key is present in the dictionary, that value is checked against the supported
+            crop species. If it is supported, that supported crop is initialized. Otherwise, a custom crop is
+            created (with 'custom' prepended to species name, if given).
+
+        Returns: a Crop object, initialized with the desired attribute values.
```
ew3361zh, line 186: And same for returns:

Returns
-------
int
    Description of anonymous integer return value.

https://numpydoc.readthedocs.io/en/latest/format.html#returns

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -255,13 +326,13 @@ def assess_dormancy(self, daylength: float) -> None:
             daylength: length of time from sunup to sundown on the current day (hours)
 
         Details:
```
ew3361zh, line 328: Not sure if this would be "Notes" now?

https://numpydoc.readthedocs.io/en/latest/format.html#notes

ew3361zh (approved):
A few formatting notes from the Numpy docstrings section but really nice overall.

Isgotkowitz (approved):

SC_redesign/Crop_and_Soil/field/field.py
morrowcj: Yes, we've got an issue to track this (#371). We're not updating older modules yet, but files will contain Numpy style from now on. We'll go back and fix docstrings before merging into master.

SC_redesign/Crop_and_Soil/field/field.py
morrowcj: I think Numpy style has a section for extended explanation, right below the first line. We'll split this sort of thing up into either "Extended Summary" or `Notes` depending on the content.

PR #383, opened 2023-03-11
User defined ration - now for review!

### LINKED ISSUE #311: Add user-defined ration functionality to animal.ration
Currently using `user-defined-ration` branch to develop ideas.
This is currently acting as a placeholder to link to other items in the Animal Module 'Project': Joe Waddell will update this task with his notes soon.

PR 1: will add the strict override version: fixed ration as % of DMI or kg for each animal class.
    - simply prints warnings if it can't meet requirements
PR 2: will implement the reduction in milk if it can't meet requirement, and print which requirements are failing and why (over, under, etc.)

JoeWaddell:
This is the initial implementation for user-defined ration functionality.

User must fill out an input JSON such that calves, heifers, lactating cows, and dry cows have their feeds listed with percentages for each. It must be a file named user_input.json in the feed input folder

From there, user must change the animal_config `"user_input": false` assignment to True. 

When flagged as true, ration_driver.py will check that flag, and automatically generate user_defined_purchased_feed.JSON in the input/feed folder. Prices of feeds not in a reference feed JSON will need to be modified by the user.


JoeWaddell:
So I've made a user manual for testing out this branch, and as noted in the description, this google doc can serve as the base for the design doc/pseudocode. Stored on basecamp here: https://3.basecamp.com/3486446/buckets/5296287/google_documents/6009429761

Methods will be improved upon once we're settled on the overall workflow, prior to moving this from Draft to full PR.

Note the various TODOs littered in the new code. Extracting nutrient composition of a formulated diet and a user-defined diet is a high priority. 

Further, there's thought we'll need to give re: how to structure the input files, how to streamline the user setup process (if necessary), and for a second PR I'd like to focus on how to better guide the user on reformulating their ration.  Perhaps simply providing the optimal diet (e.g. one not bound by the user defined percentages) alongside the warnings would be sufficient, but showing where the nutrient composition of the user-defined ration fails would be ideal.

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -56,19 +58,26 @@ def optimization(requirements, available_feeds, animal_type, cow_type):
...
         # this case should not be called, but is in place to not crash the
         # simulation if bounds error is not resolved
+        #print(count)
         if count > 30:
             solution = None
-            break
+            mock_solution = user_defined_solution(pen, requirements.DMIest)
+            ration_vals = NLP.get_ration_vals_null(mock_solution)
+            #print('nullvals')
+            return solution, ration_vals
```
KFosterReed: I don't understand why this return value needs to be here in addition to below. If I am following the logic correctly, this return will include a 'None' solution and the ration values from the mock_solution. Below, if the solution is None, it will also return ration_vals that are None as well. Is there a reason we need to be able to return a None solution with both types of ration values?

JoeWaddell: I'll need to change the name from `mock_solution` to avoid confusion here, since it's not actually returning a None solution, but an object that mimics a returned solution from the optimization step. 

The idea is that after 30 attempts at optimization (a number we can/should change, noting that it was an arbitrary value) it will accept the user-defined solution as the ideal, and generate that mimicked solution. The ration_vals returning as null are a different problem: I have a TODO to return a real set of `ration_vals` at that stage, so the current null function is a placeholder, to not confuse the logic any more than it already is.

The whole process is overly complex, hence doing an overall refactor of ration methods alongside this.

RUFAS/routines/animal/ration/ration_NLP.py
```diff
@@ -537,24 +592,70 @@ def optimize():
...
+            if not failed_constraints:
+                print('No constraints violated')
+        
+            for constr in failed_constraints:
+                # add warnings to output manager
+                # or print out to console or both
+                #print(f'Constraint {constr} violated')
+                #print(f'Constraint value: {constr["fun"](usermod.x)}')
+                #print(f'Constraint type: {constr["type"]}')
+                print(f'Supplied ration could not meet the following: {constr["fun"].__name__}')
```
KFosterReed: Is there a way to add the animal type or pen that this is failing for?

RUFAS/routines/animal/ration/ration_driver.py
KFosterReed: Ok, that makes more sense to me. I'll keep the ration refactor in mind when reviewing this as well

RUFAS/routines/animal/ration/ration_NLP.py
JoeWaddell: Absolutely! We'll need to discuss exactly what we want to report here, but the relevant information can easily be supplied to output manager as a warning or log (since it's been so nicely written by Pooya).

My thought was we'd want to report how many times it failed, the history of the checks that failed, and the condition in which it was finally resolved (e.g. lactation reduction by X%).

KFosterReed: That sounds like a great plan. Summarizing the failures with the relevant contextual information will be much more useful than a log of all the warnings - though the warnings are helpful as well! With just a little bit of guess work, I was able to modify the user_defined_ration_input_percentages inputs so that there are no failures

JoeWaddell:
Pulling this PR out of draft now that the lactation reduction branch was merged in.

## Design concerns ##
This PR adds yet another input JSON: I've kept it separate for now to not interfere with other methods that read existing JSONs, but the "percentages" file could be added to the existing JSONs, or kept as-is. 

Future changes to this methodology might require more than a single set of percentages (e.g. seasonal changes in ration composition), so thought to it's placement should take that into account.

## Known tweaks needed ##
1: How it adds the summaries of optimization failures to the output manager (in both "normal" and user-defined ration formulation methods) should be reevaluated: especially in light of point 2.

2: Likewise, this is a good PR to add pen level summary of requirements to the output manager, but it requires some careful thought on when to add them, since they're (potentially) recalculated many times. I'm meeting with @ew3361zh tomorrow to get his review on my proposal (not committed here yet).

## Basecamp/Google Doc 'manual' ## 
For anyone reviewing/using this branch, I wrote a short manual for this branch, to explain the options in the new input JSON. Find it on basecamp here: https://3.basecamp.com/3486446/buckets/5296287/google_documents/6009429761

RUFAS/routines/animal/ration/user_defined_ration.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def __new__(cls):
+        if not hasattr(cls, 'instance'):
+            cls.instance = super(UserDefinedRationManager, cls).__new__(cls)
+        return cls.instance
+
+    def __init__(self) -> None:
+        if UserDefinedRationManager.__instance is None:
+            UserDefinedRationManager.__instance = self
+            with open('input/userdefinedration/user_defined_ration_input_percentages.json', 'r') as f:
```
KFosterReed: So my understanding is that most of the json files are read in the simulation engine and classes.py files. Will it make life easer for the input manager implementation if this file is accessed and read in the Stats.__init__ function with most of the other jsons?

KFosterReed: May be a question for @PooyaHekmati

JoeWaddell: I think so: once we have a decision on where to put the contents of the JSON (e.g. standalone as it is now, or part of the feed or animal JSONs) I'll update the code so that it's read in there, to make the transition to Input Manager easier to sort out.

KFosterReed:
Hi Joe! So far I have been able to successfully change the diets and get successful rations that are pretty close to what was inputted! This is awesome!

After a few successful tests, I created this google sheet:

https://docs.google.com/spreadsheets/d/1F8sJ3P7nR8SWmnzAIJi8jplOddoPQ2Ug9LpPikOqWss/edit?usp=sharing

to start tracking which diets I had tested. My thought was that @tomhuhh and @Ed731551 could also run a few diets to help expand the number of diets that get tested. However, doing this, I started to get the following error message: 

image

I haven't dug into this too much but it looks like you tried to handle the case when the solution is not successful in the get_user_defined_ration() but it seems like this ration is failing that case.

JoeWaddell:
Great! Glad it's working for a number of tests you've tried so far. 

As for the error you got: it seems to be not even able to return a failed solution, which could be an optimization problem issue, or it could be a problem with the list of available feeds (e.g. mismatch in the various input JSONs).

I'll add in one more check at that stage, to cover the possibility that the optimization problem doesn't converge. Did you find that error trying the first diet in the linked google doc?

Edit: Following up, we could also try to return the pen.ration at this point, like in the usual workflow: however that doesn't resolve the problem in the case that it's the first formulation. I'll open another Issue for V2 to consider how we handle heifer rations that can't be optimized (since lactation can't be reduced), which should be used in both user-defined and the 'automatic' formulation methods.

KFosterReed:
> Great! Glad it's working for a number of tests you've tried so far.
> 
> As for the error you got: it seems to be not even able to return a failed solution, which could be an optimization problem issue, or it could be a problem with the list of available feeds (e.g. mismatch in the various input JSONs).
> 
> I'll add in one more check at that stage, to cover the possibility that the optimization problem doesn't converge. Did you find that error trying the first diet in the linked google doc?
> 
> Edit: Following up, we could also try to return the pen.ration at this point, like in the usual workflow: however that doesn't resolve the problem in the case that it's the first formulation. I'll open another Issue for V2 to consider how we handle heifer rations that can't be optimized (since lactation can't be reduced), which should be used in both user-defined and the 'automatic' formulation methods.

It was a version of that first diet in the google doc. I also pushed a commit with the ration percent and feed files I was using to try to make it easy to replicate

tomhuhh:
I was not able to completely replicate the issue but I got a similar error with the 'NoneType'. Please see below. 

Screenshot 2023-06-20 at 9 53 11 AM

Hello @KFosterReed, the google sheet you shared is view only.  Would you like us to add diets to that sheet when testing? Thanks!

KFosterReed:
> I was not able to completely replicate the issue but I got a similar error with the 'NoneType'. Please see below.
> 
> Screenshot 2023-06-20 at 9 53 11 AM
> Hello @KFosterReed, the google sheet you shared is view only. Would you like us to add diets to that sheet when testing? Thanks!

@tomhuhh yes please! Permissions should be updated

JoeWaddell:
The first error I was able to replicate by having a mismatch in the (currently hardcoded, see below) `input\userdefinedration\user_defined_ration_input_percentages.json` file and the feed JSON in the main udr.json input. The second bug I haven't been able to replicate.

I've updated some logic here to avoid potential causes of this bug. Any mismatch in purchased feeds, their prices, or the lists in the ration_percents file (which is still currently hardcoded to `input\userdefinedration\user_defined_ration_input_percentages.json`) will return None in the optimization process, which leads to errors later in the process. 

Also: I'll update the hardcoded location to be part of the feed JSON to further simplify this process (one less JSON to worry about!). 

I'll also look into how we can best add in checks that the user input (both in the user-defined ration scope and the usual automatic ration formulation process) matches the available and purchased feeds, to avoid failures to optimize.

tomhuhh:
I was able to successfully test a few diets, including the two that I generated for my abstract for the modeler's meeting. One thing that I noticed is that the estimated DMI is much lower than expected, at 18 kg for the NRC diet and  20 kg for the NASEM diet. Details can be found on that google sheet.

JoeWaddell:
> I was able to successfully test a few diets, including the two that I generated for my abstract for the modeler's meeting. One thing that I noticed is that the estimated DMI is much lower than expected, at 18 kg for the NRC diet and 20 kg for the NASEM diet. Details can be found on that google sheet.

This is expected when rations fail, since lactation will be reduced, and lead to a lower estimated DMI. However, I've seen rations fail upwards of 80 times (reaching that lower lactation threshold set in the percentages.json), so we should give some consideration to what constraints are OK to fail (if any) under user-defined conditions, or otherwise greatly raising the lactation reduction threshold, allowing the user to "force" a ration that the optimization method doesn't qualify as a success.

When we have expanded the scope of the `ration_report` method #531 , it will be easier to see how close a given ration is to passing, and set another threshold allowing for "close enough" fits to work. As it is, the user must decide how dramatically lactation should be reduced, which is a different way of achieving the same result.

RUFAS/routines/animal/ration/ration_NLP.py
```diff
@@ -502,6 +506,48 @@ def get_ration_vals(x):
     ration_vals = {'ME_tot': ME_tot}
     return ration_vals
 
+
+def make_user_bounds(ration_percents: Dict, DMIest: float) -> List:
```
ltp23: For a more specific return type, we could use `List[Tuple[float, float]]`.

RUFAS/routines/animal/ration/ration_NLP.py
```diff
@@ -502,6 +506,48 @@ def get_ration_vals(x):
...
+    DMIest: float
+        average estimated DMI for pen
+    Returns
+    -------
+    List
+        List of each bound, divided by three and reported in triplicate for scipy.minimize function
+    """
+    tribounds = []
+    # udr = user defined ration
+    udr_tolerance = udrv.tolerance
```
ltp23: I think it would be easier to unit-test if we could pass in `udrv` as a parameter.

RUFAS/routines/animal/ration/ration_NLP.py
```diff
@@ -502,6 +506,48 @@ def get_ration_vals(x):
...
+    DMIest_lower = DMIest-DMIest*AnimalModuleConstants.DMI_CONSTRAINT_PERCENT
+    DMIest_upper = DMIest+DMIest*AnimalModuleConstants.DMI_CONSTRAINT_PERCENT
+    ration_key_list = sorted([int(key) for key in ration_percents.keys()])
+    for key in ration_key_list:
+        # target = ration_percents[str(key)]/100*(DMIest_upper+0.0001) # change from percent to decimal percent, adding a little bit in case of 0 return
+        # target = ration_percents[key]
+        targetlower = ration_percents[str(key)]/100*(DMIest_lower+0.0001)
+        targetupper = ration_percents[str(key)]/100*(DMIest_upper+0.0001)
+        targetbounds = ((targetlower-targetlower*udr_tolerance)/3,
+                        (targetupper+targetupper*udr_tolerance)/3)
```
ltp23: Maybe using underscores to separate words in names (snake_case) and adding whitespaces in the calculations? And for the constants, we can declare them at the start of the function to avoid duplicating hard-coded values.

RUFAS/routines/animal/ration/ration_NLP.py
```diff
@@ -541,14 +595,28 @@ def optimize(animal_combination):
...
-        bnds.append((0, (limit[i] / 3) + 0.0001))
-    bnds = tuple(bnds)
-
+    if udrv.udr_or_not:
+        bnds = make_user_bounds(UserDefinedRationManager.ration_to_use(animal_combination, available_feeds), DMIest)
+    else:    
+        bnds = []
+        for i in range(len(limit)):
+            bnds.append((0, (limit[i] / 3) + 0.0001))
+        bnds = tuple(bnds)
```
ltp23:
```
bnds = [(0, (lim / 3) + 0.0001) for lim in limit]
```

RUFAS/routines/animal/ration/ration_NLP.py
```diff
@@ -522,7 +568,8 @@ def get_ration_vals(x):
 
 heifer_cons = [cons for cons in cow_cons if cons['fun'] not in [NEl_constraint, DMI_constraint_lower]]
 
-def optimize(animal_combination):
+def optimize(animal_combination, available_feeds: Dict) -> None:
+
```
ltp23: Here is a small helper function for initializing `x0`.

```
def initialize_decision_variables(n: int, scale: float = 10.0) -> List[float]:
    x0 = [1] + [random.random() * scale for _ in range(n-1)]
    return x0
```

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -86,6 +90,214 @@ def optimization(requirements, available_feeds, animal_combination):
...
+    -------
+    float
+        Average running milk
+    """
+    total_milk_in_pen = 0.0
+    for animal in pen.animals_in_pen:
+        total_milk_in_pen += animal.estimated_daily_milk_produced
+    num_animals = len(pen.animals_in_pen)
+    starting_milk_average = total_milk_in_pen/num_animals
+    return starting_milk_average
```
ltp23:
```
total_milk_in_pen = sum(animal.estimated_daily_milk_produced for animal in pen.animals_in_pen)
starting_milk_average = total_milk_in_pen / len(pen.animals_in_pen)
return starting_milk_average
```

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -86,6 +90,214 @@ def optimization(requirements, available_feeds, animal_combination):
...
+    ration: Dict
+
+    Returns
+    -------
+    List
+
+    """
+    solution_from_ration = []
+    for key in ration.keys():
+        if key != 'status' and key != 'objective':
```
ltp23: Maybe add a set variable for excluded keys

```
excluded_keys = {'status', 'objective'}
```
and then use it as
```
key not in excluded_keys
```

ltp23 (approved):
Thanks Joe for the extensive work.

KFosterReed (approved):

PooyaHekmati:
Thanks, @KFosterReed , I see this PR was started before IM became a thing; however, now that we have IM as the Single Source of Truth, all inputs have to come from IM.
CC @ltp23

KFosterReed:
My understanding is that the inputs in this PR are in alignment with the the IM meta data (more so that the current feed jsons). Am I missing something?

PooyaHekmati:
> My understanding is that the inputs in this PR are in alignment with the the IM meta data (more so that the current feed jsons). Am I missing something?

There are two steps to use IM: 1) make sure the data and metadata capture what you need, and 2) use IM.get_data() to get that input. The whole idea of IM is around the fact that we don't the model to get its inputs directly, instead, Input Manager manages the input for it.
So, it is good that we already have the data and metadata, now we need to use them (aka, integrating the IM with the rest of the code). @allisterakun is currently leading the effort of going through the entire codebase and making sure IM is the sole channel for getting inputs delivered to the model. Let's not introduce new cases for him.

Issue #379, opened 2023-03-09
`soil.phosphorus_cycling.Fertilizer` needs to be integrated into `Field` and `Soil` modules


Isgotkowitz:
The `Fertilizer` module has been implemented and is in the process of being reviewed (PR #374) but has not been integrated with the  `Field` and `Soil` modules.

morrowcj:
Going to close this, since it is covered by #381

PR #385, opened 2023-03-13
Integrate `Fertilizer` module into `Soil` and `Field` modules

Isgotkowitz:
# Main Changes
- `soil.py`
  -  Import statement for `Fertilizer` has been added, an unnecessary one has been removed. 
  - There is now a `Fertilizer` attribute to in `Soil`
  - The `make_from_config()` class method now takes field size as an argument (it is needed by methods within the `Fertilizer` module
  - `Soil` now calls the daily fertilizer phosphorus update routine as part of the `daily_soil_water_routine()`
- `soil_data.py`
  - The annual total of phosphorus leached from the field by runoff is now tracked, and is reset along with all other annual totals tracked by the `SoilData` module
  - The incorporation of `Fertilizer` into `SoilData` is tested in `test_soil_data_and_config_factory.py`
- `layer_data.py`
  - a labile phosphorus content attribute has been added to track the amount of labile phosphorus in each soil layer 
- `field.py`
  - `amend_soil()` now adds fertilizer phosphorus to the field
  - the daily update routine for soil is now called with the new method sigature 
# Other Changes
- `fertilizer.py` had a small docstring formatting issue corrected 
- Most files in this PR have had small formatting issues addressed that were raised when linting with `flake8`
- `layer_data.py` has had duplicate `@property` method removed
# Notes
- Many of the integrations of `Fertilizer` in the `Field` module are not tested, and will wait to be tested until they are merged with a branch that includes the `test_field.py` file
- Closes https://github.com/RuminantFarmSystems/MASM/issues/379


morrowcj (approved):
LGTM!

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -56,11 +56,11 @@ def manage_field(self, day: int, year: int, current_weather: CurrentWeather) ->
         # determine total amount of residue and above-ground biomass present on the given day
         total_plant_cover = self.field_data.current_residue + self._determine_total_above_ground_biomass()
 
+        # TODO: track snow cover on soil surface somewhere - Issue #317
         self.soil.daily_soil_routine(current_weather.incoming_light, current_weather.mean_air_temperature,
                                      current_weather.min_air_temperature, current_weather.max_air_temperature,
```
ltp23, line 69: This list of parameters is getting long, so maybe we should start to include parameter names as well.

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -134,7 +134,11 @@ def till_soil(self) -> None:
 
     def amend_soil(self) -> None:
         """amend the soil with nutrients"""
-        pass
+        # TODO: test this method when it is merged with a branch containing field_test.py
+        fertilizer_phosphorus_applied = 90
+        if fertilizer_phosphorus_applied:
+            self.soil.fertilizer_phosphorus.add_fertilizer_phosphorus(fertilizer_phosphorus_applied)
```
ltp23: Probably only this line is needed here. I think the `add_fertilizer_phosphorus` should be able to handle the case of input 0 by doing nothing.

ltp23 (approved):
Looks good to me! Thanks, Ed.

SC_redesign/Crop_and_Soil/field/field.py
morrowcj: This is a good point. I've wondered when/if we need to employ keywords explicitly when calling methods.

morrowcj (approved):

SC_redesign/Crop_and_Soil/field/field.py
Isgotkowitz: I will change this up

SC_redesign/Crop_and_Soil/field/field.py
Isgotkowitz: Good point, that would be a more sensible placeholder

PR #392, opened 2023-03-15
Implemented IPCC equations for heifers and dry cows

tomhuhh:
IPCC Tier 2 methane emission equations for heifers and dry cows from the pseudocode has been implemented in this commit.


KFosterReed (approved):
I made one change to remove the Boadi equation for growing heifers. I don't think it adds anything to our prediction abilities and we don't want users to input 'Boadi' as a methane model option so heifers will be the same as calves - IPCC method only. Other than that, it looks good to me and works as expected!

tomhuhh:
> 

Hello Kristan, 

As we discussed yesterday at our meeting, the methane equation for calves is from Pattanaik et al. (2003) not IPCC. I wanted to make a note here for future reference. 

Thanks!
Haowen

RUFAS/routines/animal/manure/calf_manure_excretion.py
```diff
@@ -17,7 +17,8 @@ def manure_calculations(ration_formulation,
                         feed,
                         body_weight: float,
                         fecal_phosphorus: float,
-                        urine_phosphorus_required: float) \
+                        urine_phosphorus_required: float,
+                        methane_model: str) \
```
ltp23, line 22: You could set a default methane model like this `methane_model='IPCC'`. Then, you wouldn't have to pass in an extra argument in the call to function `manure_calculations()` in the animal classes, unless it is using some other model.

ltp23 (approved):
Great job, Haowen! Thank you.

RUFAS/routines/animal/manure/calf_manure_excretion.py
tomhuhh: Hello Loi, 

That is correct. There is not an IPCC equation for calves according to the pseudocode. That is why it is not set to "IPCC" for calves. 

Thanks! 
Haowen

PR #397, opened 2023-03-17
NASEM logic updates

JoeWaddell:
This is adding more nuanced logic to the NASEM nutrient calculations for P and Ca, following NASEM guidelines. Those were only slightly changed: if parity is over 2, then certain parts of the equations aren't used. Likewise, the NASEM equations sometimes return negative values, so we set 0.0 as a minimum return.

I've also added a minimum for the DMI estimates, as during data exploration for model validation we found that some edge cases returned negative values. The lower bound should be a discussion point for this PR, since we can set it as a flat kg value, or as some basic percentage of body weight, for the cases in which we find nonlinear behavior (since the equations in the literature were developed for 'average' animals). 
Note that DMI can't be set to 0, as it will raise an error in ration formulation when trying to divide by 0.


JoeWaddell:
I've created a ration constants file to make it easier to manipulate minimums and other magic numbers in the animal_requirements file. For now, I've started with just a handful, and notably the ones that prompted this fix. We can/should think carefully about whether we want one big animal constants file in the base animal directory, given there's already a life cycle-specific one in a subdirectory. 

Likewise, these minimums are just placeholders, and upping those minimums necessitates changing the test functions, as well. I've also reverted the default calculation method to NRC for now, since that is largely unchanged (though the minimum DMI for heifers here will come into play, since it's set to 0 in animal_requirements.py for heifers, by default).

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -883,6 +887,8 @@ def calculate_NASEM_calcium_requirements(body_weight: float, mature_body_weight:
                      (day_of_pregnancy - 1)) * (body_weight/715)
     Ca_Lact = (0.295 + 0.239 * milk_true_protein) * milk_production
     calcium_requirement = Ca_Maint + Ca_Growth + Ca_Preg + Ca_Lact
+    if calcium_requirement < ration_constants.minimum_calcium:
+        calcium_requirement = ration_constants.minimum_calcium
     return calcium_requirement
```
YijingGong: Those 3 lines can be made concise to 'return max(calcium_requirement, ration_constants.minimum_calcium)'

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -990,19 +996,24 @@ def calculate_NASEM_phosphorus_requirements(body_weight: float, mature_body_weig
...
             math.exp((0.05527-0.000075*(day_of_pregnancy-1)) *
                      (day_of_pregnancy-1)*(body_weight / 715))
     if milk_true_protein == None or milk_production == None:
         P_Lact = 0.0
     else:
         P_Lact = milk_production * (0.49 + 0.13*milk_true_protein)
     phosphorus_requirement = P_Maint + P_Growth + P_Preg + P_Lact
+    if phosphorus_requirement < ration_constants.minimum_phosophorus: 
+        phosphorus_requirement = ration_constants.minimum_phosophorus
     return phosphorus_requirement
```
YijingGong: Same as the calcium requirement.

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -990,19 +996,24 @@ def calculate_NASEM_phosphorus_requirements(body_weight: float, mature_body_weig
         P_Maint = 0.8 * dry_matter_intake_estimate + 0.0006 * body_weight
     else:
         P_Maint = 0.0
-    P_Growth = (1.2+4.635*mature_body_weight**0.22 *
+    if parity <= 2:
+        P_Growth = (1.2+4.635*mature_body_weight**0.22 *
```
YijingGong, line 998: I was thinking do we want to keep track of what has been changed from the NDC (2001)? For the purpose of distinguishing what changes are due to the code improvement practices and what changes are because of the NASEM 2021. Maybe it is an unnecessary workload. Just thinking aloud. 
One easy way might be just adding a line in the docstring notes part saying 'revised in NASEM 2021 from NRC 2001'

RUFAS/routines/animal/ration/ration_constants.py
```diff
@@ -0,0 +1,12 @@
+"""
+RUFAS: Ruminant Farm Systems Model
+File name: animal_requirements.py
+Description: This file contains constants used in calculations of animal requirements
+    May also serve as repository for values associated with ration formulation in the future
+Author(s): Joe Waddell jw2574@cornell.edu
+"""
+
+# animal_requirements.py minimum values for returns
+minimum_DMI = 1.0 # kg/day minimum instituted for all animals
```
YijingGong, line 11: Is that possible that a calf eats less than this?

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -1058,6 +1069,10 @@ def calculate_NRC_DMI(animal_type: str, body_weight: float, day_of_pregnancy: in
         # TODO: Actual calculation for dry_matter_intake_estimate
         dry_matter_intake_estimate = 0.0
         # this comment is a holdover from the previous version
+    # TODO: below (and in the NASEM calculation) we use a flat minimum DMI value, but...
```
KFosterReed, line 1075: I think a minimum DMI based on the percentage of BW makes a lot of sense. I'll look into this more but 1% of BW seems like a reasonable minimum to me.

KFosterReed (approved):
This seems like a pretty straight forward/simple start to updating some of the logic to prevent the requirement estimates from going out of bounds. 

Also, the addition of a constants file to keep the magic values is great. I have a slight preference for all the animal constants to be in one place so combining with the lifecycle constants seems like a good idea but open to discussion

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -1066,14 +1062,22 @@ def calculate_NRC_DMI(animal_type: str, body_weight: float, day_of_pregnancy: in
...
     else:
-        # TODO: Actual calculation for dry_matter_intake_estimate
-        dry_matter_intake_estimate = 0.0
+        net_energy_maintenance_diet = 1 # TODO update this method to retrieve values from nutrient composition of 
+                                        # ration from previous formulation.
+                                        # Currently using magic value set by Edward and Haowen
+        dry_matter_intake_estimate = body_weight**0.75 * (0.2435*net_energy_maintenance_diet \
+                                                          - 0.0466*net_energy_maintenance_diet**2 \
+                                                          - 0.1128) / net_energy_maintenance_diet
+        if day_of_pregnancy and day_of_pregnancy >= 210:
```
KFosterReed, line 1071: is this if statement correct? Why are there two day_of_pregnancy variables?

JoeWaddell: The first half of the statement is to check that `day_of_pregnancy` is not `None`, and the second is to apply it only in cases where it is over 210. Since heifers start with a `day_of_pregnancy` set to `None` (but others do have an integer value), it seemed like a succinct way to make both checks.

On a related note, in the ration refactor I'll need update all of these methods to make it clear which block of code applies to which animal class, to improve overall readability.

RUFAS/routines/animal/ration/ration_constants.py
JoeWaddell: To follow up from this: calves are handled entirely separately for the time being. In forthcoming refactors we'll be bringing them into these methods, when I'll have to add logic to handle them properly.

RUFAS/routines/animal/ration/animal_requirements.py
JoeWaddell: I will talk to @Ed731551 about this, since he will be able to comment on any modifications from the original equations. We'll add any notes like this to the docstring.

RUFAS/routines/animal/ration/animal_requirements.py
JoeWaddell: Handled here and elsewhere, thanks!

JoeWaddell (commented):
This should be ready to go for now. 

Related to this: since we are now calculating a DMI estimate for both NRC and NASEM methods, the next small PR related to the ration refactor actually use these DMI constraints for heifers.

RUFAS/routines/animal/ration/animal_requirements.py
Ed731551: Yes, we can have a talk about this. Some equations are the same in both feeding systems.

RUFAS/routines/animal/ration/animal_requirements.py
KFosterReed: Ok! That makes sense. And better organization/logic to separate the animal classes is a great goal

Ed731551: Just to add on the "magic value". That refers to an assumption of 1.0 Mcal of NEm/kg of diet. Values make sense for growing animals by using this as an initial input.

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -1055,10 +1062,22 @@ def calculate_NRC_DMI(animal_type: str, body_weight: float, day_of_pregnancy: in
...
+                                        # Currently using magic value set by Edward and Haowen
+        dry_matter_intake_estimate = body_weight**0.75 * (0.2435*net_energy_maintenance_diet 
+                                                          - 0.0466*net_energy_maintenance_diet**2 
+                                                          - 0.1128) / net_energy_maintenance_diet
+        if day_of_pregnancy and day_of_pregnancy >= 210:
+            adjustment_factor = 1+((210-day_of_pregnancy) * 0.0025)
+            dry_matter_intake_estimate -= adjustment_factor
         # this comment is a holdover from the previous version
-    return dry_matter_intake_estimate
+    # TODO: below (and in the NASEM calculation) we use a flat minimum DMI value, but...
```
tomhuhh, line 1075: With this update in DMI for heifers, it would make sense to add con11 (DMI_constraint) to heifers in the ration_NLP.py file. I tested with different combinations of feed ingredients. When I used soybean meal as a major protein source, the intake with the updated code was 11.11 kg, which is reasonable.

tomhuhh (approved):
Looks good to me! Thanks Joe!

Ed731551 (approved):
Changes in logic makes sense to me. Thanks

RUFAS/routines/animal/ration/animal_requirements.py
Ed731551: Agree

JoeWaddell: Excellent. Adding the constraint at the optimization step will be the first change in my next small ration refactor PR, so it is in a separate PR.

PR #399, opened 2023-03-18
CWolber draft pr. 

carsonwolber:


JoeWaddell (approved):
Looks good to me! The new tests pass on my system, and the overall cleanup of tests and their names is wonderful.

tests/test_animal.py
```diff
@@ -1218,12 +1242,12 @@ def test__calc_average():
     pass
 
 
-def test_manure_calculations():
+def test_calf_manure_calculations():
```
tomhuhh, line 1255: Thanks Carson. There are already unit tests for manure calculations in tests/test_manure_module/test_animal_manure_excretions. So I am uncertain if it is still necessary to write separate unit tests. Thanks!

tomhuhh (approved):
Good work Carson! This looks good to me. Thanks!

PR #394, opened 2023-03-17
Bookshelf: Generalized sensitivity analysis

morrowcj:
# NOTE 

This PR will be split into multiple PRs. It is too large ATM. AND contains changes beyond those present in #402 

----

This PR will setup a generalized sensitivity analysis module to be used with RUFAS. I will adapt the tools that @JoeWaddell has created for the more general case. 

# Changes

Most of the changes are additions of reference files and can be ignored for the Review. The 3 files that should be reviewed are

* The `README.md` file, which contains information about the module generally
* The `generalized_sensitivity.py` file, which contains the relevant class.
* The `SC_sensitivity.md` file, which will serve as a design/plan for executing SA on the crop and soil modules. 

# Notes
This is just a draft for now. I wanted to make it available early so that feedback can be added at any stage. 
I'll convert to a full pull request once it is ready for review.

PR #402, opened 2023-03-22
Generalized Sensitivity Analysis Module

morrowcj:
# Summary
* Adds a module with methods for conducting global sensitivity analysis with any objective function.
* Contains a detailed README and a module with the new class.

# Next steps
* provide examples of how to wrap the RuFaS modules to work with the new class
* extend the class to perform an additional SA step, with a reduced parameter set based on the initial SA result (maybe, see #394).
* Improve and extend the README file.

### DIFF
--- a/model_evaluation/sensitivity_analysis/README.md
+++ b/model_evaluation/sensitivity_analysis/README.md
@@ -1,13 +1,13 @@
 # Sensitivity Analyses for RuFaS
 
-Authors: Clay J. Morrow <!-- TODO: Joe, Varma, and others, please add your name here when you make changes/additions --> 
-Date Created: 17 Mar 2023  
+Authors: Clay J. Morrow <!-- TODO: Joe, Varma, and others, please add your name here when you make changes/additions -->   
+Date Created: 17 Mar 2023   
 Last Updated: 22 Mar 2023 <!-- NOTE: please remember to change this date when editing this file -->> 
 
 __Contents:__
 1. [Introduction](#1-introduction)
 2. [Summary of RuFaS Sensitivity Analyses](#2-summary-of-rufas-sensitivity-analyses)  
-   a. [Parallelization](#parallelization) 
+   a. [Parallelization](#parallelization)   
    b. [Types of SA](#types-of-sa)
 3. [References and Links](#3-references-and-links)  
    a. [Primary Sources](#primary-sources)  
--- a/model_evaluation/sensitivity_analysis/README.md
+++ b/model_evaluation/sensitivity_analysis/README.md
@@ -1,8 +1,8 @@
 # Sensitivity Analyses for RuFaS
 
-Authors: Clay J. Morrow <!-- TODO: Joe, Varma, and others, please add your name here when you make changes/additions -->   
+Authors: Clay J. Morrow, Joseph Waddell <!-- TODO: Joe, Varma, and others, please add your name here when you make changes/additions -->   
 Date Created: 17 Mar 2023   
-Last Updated: 22 Mar 2023 <!-- NOTE: please remember to change this date when editing this file --> 
+Last Updated: 23 Mar 2023 <!-- NOTE: please remember to change this date when editing this file --> 
 
 __Contents:__
 1. [Introduction](#1-introduction)
--- a/model_evaluation/sensitivity_analysis/README.md
+++ b/model_evaluation/sensitivity_analysis/README.md
@@ -29,7 +29,7 @@ to create these methods.
 
 ## 2. Summary of RUFAS Sensitivity Analyses
 
-Broadly speaking, a global sensitivity analysis (SA) simply aims to answer the questions, "which parameters have the 
+Broadly speaking, a global sensitivity analysis (SA) simply aims to answer the question, "which parameters have the 
 largest impact on the results of a model".
 
 Within the [generalized_sensitivity.py](generalized_sensitivity.py) file, is the class `SensitivityAnalysis`. This 
@@ -50,9 +50,13 @@ parallelize its execution. Luckily the package that we are using to perform SA
 ### Types of SA
 
 In this document, we'll consider 3 methods of global sensitivity analysis: 1) Morris' method, 2) the extended FAST
-method, and 2) the Sobol method. They each have their advantages and disadvantages. Generally speaking, Sobol is
+method, and 3) the Sobol method. They each have their advantages and disadvantages. Generally speaking, Sobol is
 the most accurate but is the slowest by far. The other methods are fast and accurate but can't quantify
 interactive effects (Sobol can quantify 2nd order).
+
+A solid review/summary/critique that covers different approaches in more detail:[Paleari et al, 2021](https://www.sciencedirect.com/science/article/pii/S0304380021002088?casa_token=xTPqc6Gp5w4AAAAA:HUJUaa__fELazAdkJrfCtK5DyILPrRqUdvHBKN_8t6ikjLYsb7UpdpmR_KYgiqpRf9tcgdmGcZI) 
+
+They outline some of the drawbacks of different methods, and may serve as a good citation when publishing results. Particularly, when we are able to demonstrate concordance between the different metrics. Part of the overall workflow can/should include various techniques, with the fastest methods being used as a first pass, allowing us to have both preliminary results and detailed second-order effects calculated - when appropriate - using the computationally intensive Sobol method. 
 
 ###### Morris' method
 
--- a/model_evaluation/sensitivity_analysis/README.md
+++ b/model_evaluation/sensitivity_analysis/README.md
@@ -1,0 +1,122 @@
+# Sensitivity Analyses for RuFaS
+
+Authors: Clay J. Morrow <!-- TODO: Joe, Varma, and others, please add your name here when you make changes/additions --> 
+Date Created: 17 Mar 2023  
+Last Updated: 22 Mar 2023 <!-- NOTE: please remember to change this date when editing this file -->> 
+
+__Contents:__
+1. [Introduction](#1-introduction)
+2. [Summary of RuFaS Sensitivity Analyses](#2-summary-of-rufas-sensitivity-analyses)  
+   a. [Parallelization](#parallelization) 
+   b. [Types of SA](#types-of-sa)
+3. [References and Links](#3-references-and-links)  
+   a. [Primary Sources](#primary-sources)  
+   b. [SALib](#salib)
+
+---
+
+## 1. Introduction
+
+The purpose of this document is to guide the developers of RuFaS in conducting sensitivity analyses (SA). These methods
+can be applied to either 1) The entire RuFaS model or 2) any submodule, with minimal work on the user's part. 
+Joe Waddell began this process, by finding the relevant packages and writing some scripts to implement SA for the full 
+RuFaS module. My goal is to build on his work to write a module that generalises the SA methods and allows them to be 
+applied to any python function/module. 
+
+The [References and Links](#references-and-links) section provides both the primary and secondary sources that we used
+to create these methods. 
+----
+
+## 2. Summary of RUFAS Sensitivity Analyses
+
+Broadly speaking, a global sensitivity analysis (SA) simply aims to answer the questions, "which parameters have the 
+largest impact on the results of a model".
+
+Within the [generalized_sensitivity.py](generalized_sensitivity.py) file, is the class `SensitivityAnalysis`. This 
+class executes a standardized procedure for conducting SA. The user needs to provide an objective function, names of
+parameters of that function to be analyzed, upper and lower bounds for each parameter, and some additional 
+specifications about how the SA should be conducted. 
+
+Currently, this class supports 3 methods for conducting SA. These methods (Sobol, FAST, and Morris' method) are
+further discussed below in [Types of SA](#types-of-sa). 
+
+### Parallelization
+
+Because RuFaS takes a long time to run and will need to be executed many times (thousands), it will be important to
+parallelize its execution. Luckily the package that we are using to perform SA 
+[SALib](https://salib.readthedocs.io/en/latest/) (additional links and description provided in the 
+[SALib Section](#salib) below) already utilizes parallel processing. We leverage that functionality in our SA class.
+
+### Types of SA
+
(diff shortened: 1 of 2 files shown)


model_evaluation/sensitivity_analysis/README.md
```diff
@@ -0,0 +1,122 @@
...
+  https://waterprogramming.wordpress.com/2014/02/11/extensions-of-salib-for-more-complex-sensitivity-analyses/
+  )
+* [Applied example 1](http://keyboardscientist.weebly.com/blog/sensitivity-analysis-with-salib)
+* [Applied example 2](https://notebook.community/locie/locie_notebook/misc/Sensitivity_analysis)
+* [Package example using a variable and parameters](https://salib.readthedocs.io/en/latest/basics.html#another-example)
+* [Categorical data GitHub Issue](https://github.com/SALib/SALib/issues/195)
+* [Negative sobol index GitHub Issues](https://github.com/SALib/SALib/issues/102)
+* [How to wrap existing models](https://salib.readthedocs.io/en/latest/user_guide/wrappers.html)
+
+The documentation for the package pretty good, and the two blogs are helpful at seeing how others have used it.
```
Isgotkowitz, line 119:
```suggestion
The documentation for the package is pretty good, and the two blogs are helpful at seeing how others have used it. 
```

model_evaluation/sensitivity_analysis/generalized_sensitivity.py
```diff
@@ -0,0 +1,295 @@
...
+        self._check_inputs(pars, bounds, groups)
+
+        self.objective_function: Callable = fun
+        self.parameter_names: List[str] = pars
+        self.parameter_bounds: List[Tuple[float, float]] = bounds
+        self.sensitivity_method: SupportedSensitivityMethods = SupportedSensitivityMethods(method)
+        self.parameter_groups: Optional[List[str]] = groups
+        self.output_names: Optional[List[str]] = outputs
+        self.sample_n: int = sample_n
+        self.parallel_processors: int = n_cores
```
Isgotkowitz:
```suggestion
        if n_cores is None:
            self.parallel_processors: int = os.cpu_count()
        else:
            self.parallel_processors: int = n_cores
```
I think it would be better to set the default number of cores to the number of system (logical) cores, instead of one. This makes the class as powerful as possible by default, which will probably help a lot of the SME's who may not know what this parameter does and what it should be set too.
Also if the above suggestion is committed, the `os` library will need to be imported, and the default in the `init()` signature changed to `None`.

model_evaluation/sensitivity_analysis/generalized_sensitivity.py
```diff
@@ -0,0 +1,295 @@
...
+        if self.sensitivity_method == SupportedSensitivityMethods.SOBOL:
+            self.problem.sample(func=sobol_sampler.sample, N=self.sample_n, *args, **kwargs)
+        if self.sensitivity_method == SupportedSensitivityMethods.MORRIS:
+            self.problem.sample(func=morris_sampler.sample, N=self.sample_n, *args, **kwargs)
+
+    def evaluate_model(self) -> None:
+        """Evaluates the objective function over the entire parameter sample space
+
+        Notes
+        ----
```
Isgotkowitz:
```suggestion
        -----
```

Isgotkowitz (approved):
Really cool stuff here Clay, and kudos on making it possible to run on any subsection of RuFaS. I have one suggestion for making `SensitivityAnalysis` parallelized by default, see the comment below. I also noticed in the richdiff view of the README that some things looked wonky. 

I am concerned about some of the parallelized SALib methods being experimental, and thus potentially not working correctly. If they usually run correctly but sometimes crash, that is ok because whoever is performing the sensitivity analysis will know something went wrong and can try running the tests again. But if a method runs incorrectly and does not crash the results will be invalid with no warning or indication, which seems (to me) like a very dangerous outcome.

model_evaluation/sensitivity_analysis/generalized_sensitivity.py
morrowcj: @Isgotkowitz, I thought about doing exactly this and decided not to for one very annoying reason: Pycharm absolutely will not run any code that utilizes the `multiprocessing` package (which the parallel versions of `SALib` code does). So, if it were setup to run in parallel by default, users wouldn't be able to test it in Pycharm. Instead, they would need to either need to use a terminal or alternate IDE. 

I'll be sure to add notes about this exact issue to the docstrings - I meant to do so and forgot.

morrowcj:
@Isgotkowitz, I want to reply to some of your main points:

> I also noticed in the richdiff view of the README that some things looked wonky.

Thanks for pointing this out - I'll fix that.

> I am concerned about some of the parallelized SALib methods being experimental, and thus potentially not working correctly. If they usually run correctly but sometimes crash, that is ok because whoever is performing the sensitivity analysis will know something went wrong and can try running the tests again. 

I had a conversation with the developers of `SALib` about this exact thing and they assured me that it is stable from all the testing they have done. It just is not officially supported yet and may have unexpected behavior occasionally (see my reply about multiprocessing by default). In my testing (albeit somewhat limited), I have had perfect success with the parallelization. Looking at the source code of the methods in question, I would have implemented parallelization in almost the exact same way that the authors of  `SALib` did.

[This thread](https://github.com/SALib/SALib/issues/555) has the exchange with `SALib` devs and [this open issue](https://github.com/SALib/SALib/issues/559) is a follow up regarding Pycharm's trouble with parallelization.

model_evaluation/sensitivity_analysis/generalized_sensitivity.py
Isgotkowitz: Shoot that is disappointing

model_evaluation/sensitivity_analysis/README.md
```diff
@@ -0,0 +1,122 @@
...
+RuFaS module. My goal is to build on his work to write a module that generalises the SA methods and allows them to be 
+applied to any python function/module. 
+
+The [References and Links](#references-and-links) section provides both the primary and secondary sources that we used
+to create these methods. 
+----
+
+## 2. Summary of RUFAS Sensitivity Analyses
+
+Broadly speaking, a global sensitivity analysis (SA) simply aims to answer the questions, "which parameters have the
```
JoeWaddell:
```suggestion
Broadly speaking, a global sensitivity analysis (SA) simply aims to answer the question, "which parameters have the 
```

model_evaluation/sensitivity_analysis/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+Because RuFaS takes a long time to run and will need to be executed many times (thousands), it will be important to
+parallelize its execution. Luckily the package that we are using to perform SA 
+[SALib](https://salib.readthedocs.io/en/latest/) (additional links and description provided in the 
+[SALib Section](#salib) below) already utilizes parallel processing. We leverage that functionality in our SA class.
+
+### Types of SA
+
+In this document, we'll consider 3 methods of global sensitivity analysis: 1) Morris' method, 2) the extended FAST
+method, and 2) the Sobol method. They each have their advantages and disadvantages. Generally speaking, Sobol is
```
JoeWaddell:
```suggestion
method, and 3) the Sobol method. They each have their advantages and disadvantages. Generally speaking, Sobol is
```

model_evaluation/sensitivity_analysis/README.md
```diff
@@ -0,0 +1,122 @@
...
+the most accurate but is the slowest by far. The other methods are fast and accurate but can't quantify
+interactive effects (Sobol can quantify 2nd order).
```
JoeWaddell, line 55:
```suggestion
interactive effects (Sobol can quantify 2nd order).

A solid review/summary/critique that covers different approaches in more detail:[Paleari et al, 2021](https://www.sciencedirect.com/science/article/pii/S0304380021002088?casa_token=xTPqc6Gp5w4AAAAA:HUJUaa__fELazAdkJrfCtK5DyILPrRqUdvHBKN_8t6ikjLYsb7UpdpmR_KYgiqpRf9tcgdmGcZI) 

They outline some of the drawbacks of different methods, and may serve as a good citation when publishing results. Particularly, when we are able to demonstrate concordance between the different metrics. Part of the overall workflow can/should include various techniques, with the fastest methods being used as a first pass, allowing us to have both preliminary results and detailed second-order effects calculated - when appropriate - using the computationally intensive Sobol method. 
```

model_evaluation/sensitivity_analysis/generalized_sensitivity.py
```diff
@@ -0,0 +1,295 @@
...
+        `Sobol method <https://salib.readthedocs.io/en/latest/api/SALib.analyze.html#module-SALib.analyze.sobol>`_
+    'fast'
+       `FAST method <https://salib.readthedocs.io/en/latest/api/SALib.analyze.html#module-SALib.analyze.fast>`_
+    `morris`
+        `Morris' method <https://salib.readthedocs.io/en/latest/api/SALib.analyze.html#module-SALib.analyze.morris>`_
+    """
+    SOBOL = "sobol"
+    FAST = "fast"
+    MORRIS = "morris"
+
```
JoeWaddell, line 34: I think we could discuss how best to implement something like the fractional factorial approach as well, since it is supported by SA lib. It has far more drawbacks than even the FAST method, but it does have the advantage of being even faster, and nicely handles binary inputs (with some conversion, can manage string input, e.g. if we sample between two manure treatment types, two feed types, etc.).

JoeWaddell (approved):
This is a great first step, nice job!  

The trickiest part might be ensuring that RuFaS runs properly inside the parallelized wrapper, as I had some problems with it early on (Clay uploaded the code, which has various iterations of my attempts) I think mostly related to different cores trying to access the same files at the same time.

The helper functions I wrote can likely act as a base for modifying the JSON inputs according to the parameter space that the selected sampling functions generate, and since they're sparsely documented, I can write the documentation prior to implementation here. We can all meet to coordinate the next steps, and/or open a sensitivity analysis project where we can hash out details and take ownership of various steps along the way. 

@YijingGong is also a superb resource for reviewing sensitivity analysis methods, as much of the code I was running was adapted from her analyses.

model_evaluation/sensitivity_analysis/README.md
morrowcj: Great addition, @JoeWaddell

model_evaluation/sensitivity_analysis/generalized_sensitivity.py
morrowcj: That's a great Idea - I think it can easily be added in this framework. I'll look into that. in the next PR.

morrowcj:
@JoeWaddell, I think you raise some good points. I wanted to reply to a few of your points:

> The trickiest part might be ensuring that RuFaS runs properly inside the parallelized wrapper, as I had some problems with it early on (Clay uploaded the code, which has various iterations of my attempts) I think mostly related to different cores trying to access the same files at the same time.

Yeah, I'm a bit worried about this as well, particularly with the memory leak issue. I think I know how to get around the reading from the same file, component. I'm thinking that what we talked about earlier - all inputs being contained within a class - will help with that. Specifically, which files were being read from multiple times? Just the JSONs?

> and/or open a sensitivity analysis project where we can hash out details and take ownership of various steps along the way.

I think this is a good idea.

> @YijingGong is also a superb resource for reviewing sensitivity analysis methods, as much of the code I was running was adapted from her analyses.

Great to know!, I'll make sure to request her review in the next iterations.

varrma (approved):
@morrowcj  This is very interesting having a generalized SA method applicable to all modules.  About setting our own input and output parameters, does the SA method randomize the number of iterations. For example, one input parameter against the "n" output parameters at a time or all input parameters iterates against all the outputs. I believe it could be sub-module specific and type of method selection.

morrowcj:
@varrma, in response to your question

> About setting our own input and output parameters, does the SA method randomize the number of iterations. For example, one input parameter against the "n" output parameters at a time or all input parameters iterates against all the outputs.

The number of iterations is not randomized, it is fixed based on the `N` value that you supply. This causes the sampler to randomly sample parameter space and provide some number of cases $K$ (which depends on the method and is not equal to `N`). The model is then run $K$ times - one for each set of parameters in the sample. 

Another part of your question appears to ask if given a fixed set of input values, let's say $x=7$ and $y=4$, the model runs multiple times. The answer to that is no. Each of the K samples has a unique parameter combinations (meaning if $x$ and $y$ are your only parameters and one sample has $x=7$ and $y=7$, then the model will only run once with these parameters). That is because these types of SA assume that the model is deterministic (i.e., the same input always gives the same output, no randomness). 

Does that answer your question?

varrma:
> @varrma, in response to your question
> 
> > About setting our own input and output parameters, does the SA method randomize the number of iterations. For example, one input parameter against the "n" output parameters at a time or all input parameters iterates against all the outputs.
> 
> The number of iterations is not randomized, it is fixed based on the `N` value that you supply. This causes the sampler to randomly sample parameter space and provide some number of cases K (which depends on the method and is not equal to `N`). The model is then run K times - one for each set of parameters in the sample.
> 
> Another part of your question appears to ask if given a fixed set of input values, let's say x=7 and y=4, the model runs multiple times. The answer to that is no. Each of the K samples has a unique parameter combinations (meaning if x and y are your only parameters and one sample has x=7 and y=7, then the model will only run once with these parameters). That is because these types of SA assume that the model is deterministic (i.e., the same input always gives the same output, no randomness).
> 
> Does that answer your question?

Thanks Clay for the explanation, yes that helps. I understand that the sampler and the selected method would define the iterations.

JoeWaddell:
> Yeah, I'm a bit worried about this as well, particularly with the memory leak issue. I think I know how to get around the reading from the same file, component. I'm thinking that what we talked about earlier - all inputs being contained within a class - will help with that. Specifically, which files were being read from multiple times? Just the JSONs?

Those certainly, and I was also worried about how the csv and sqlite databases are being accessed.

The temporary solution I landed on was duplicating all those files, so that there is a unique version for reach run: a lot of setup, and it requires cleanup following a successful run, but it seems like a dead-simple way around potential problems. 

That said, as you mention: reading those once, assigning their contents to a singleton class could be the way forward. It's similar to what I'm implementing in the user_defined_ration branch, to avoid accessing the same JSON multiple times.

PR #406, opened 2023-03-23
Refactor methods to add to phosphorus pools - closes #403

### LINKED ISSUE #403: Refactor methods to add to phosphorus contents of a soil layer
Currently the `_add_to_labile_phosphorus()` method is implemented in two modules: `phosphorus_cycling/ManureApplication` and `phosphorus_cycling/Fertilizer`. It should be moved to the `LayerData` module so that any module may call it.
The method `_add_to_active_phosphorus()` (in `phosphorus_cycling/ManureApplication`) should also be moved to `LayerData`.

Isgotkowitz:
# Summary
This PR eliminates a method for adding phosphorus to a soil pool that was duplicated across modules (`fertilizer.py` and `manure_application.py`), by moving it to `layer_data.py` where it can now be accessed by any module. It also changes the design from having a method to add phosphorus to each individual pool to having a single base method that can calculate the new value of any pool, and different wrapper methods for calling the base method and resetting the pool to the result.

# Main Changes
- `layer_data.py`
  - `add_to_labile_phosphorus()`, `add_to_stable_phosphorus()` (the wrapper methods) and `_add_phosphorus_to_pool()` (the base method) have been added.
  - All are tested in `test_soil_data_and_config_factory.py`
- `fertilizer.py` and `manure_application.py`
  - All uses of the old methods that add to soil phosphorus pools have been changed to use the new ones, and all tests have been updated to mock the new functions.
  - The old methods and their tests have been removed from both of these modules.

# Other Changes
- `layer_data.py` has had a stale TODO removed

# Notes
- Please scrutinize the pattern for the base and wrapper methods, there may be a cleaner or more efficient way to implement them.
- Closes #403

### DIFF
--- a/SC_redesign/Crop_and_Soil/soil/layer_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/layer_data.py
@@ -1,4 +1,4 @@
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Optional
 
 """
@@ -19,7 +19,7 @@ class LayerData:
     """phosphate content of the layer (kg/ha)"""
     soil_water_concentration: float = 0.25  # arbitrary
     """soil water concentration of the layer (mm)"""
-    water_content:  Optional[float] = None
+    water_content:  float = field(init=False)
     """water present in the layer (mm)"""
     field_capacity_water_concentration: float = 0.3  # arbitrary
     """water concentration of soil layer at field capacity (mm water / mm soil)"""
--- a/SC_redesign/Crop_and_Soil/soil/layer_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/layer_data.py
@@ -76,9 +76,9 @@ class LayerData:
 
         Parameters
         ----------
-            phosphorus_to_add: float
+            phosphorus_to_add : float
                 Amount of phosphorus to add (kg)
-            field_size: float
+            field_size : float
                 Size of the field (ha)
 
         """
@@ -91,9 +91,9 @@ class LayerData:
 
         Parameters
         ----------
-            phosphorus_to_add: float
+            phosphorus_to_add : float
                 Amount of phosphorus to add (kg)
-            field_size: float
+            field_size : float
                 Size of the field (ha)
 
         """
@@ -101,7 +101,7 @@ class LayerData:
                                                                       field_size)
 
     @staticmethod
-    def _add_phosphorus_to_pool(self, pool_to_add_to: float, phosphorus_to_add: float, field_size: float) -> float:
+    def _add_phosphorus_to_pool(pool_to_add_to: float, phosphorus_to_add: float, field_size: float) -> float:
         """This is a generic method to be used by wrapper functions to add phosphorus to any of the phosphorus pools.
 
         Parameters
--- a/SC_redesign/Crop_and_Soil/soil/layer_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/layer_data.py
@@ -70,6 +70,65 @@ class LayerData:
         """Initialize all attributes in the dataclass that depend on other attributes"""
         self.water_content = self.soil_water_concentration * self.layer_thickness
 
+    def add_to_labile_phosphorus(self, phosphorus_to_add: float, field_size: float) -> None:
+        """This method is a wrapper for adding a specified mass of phosphorus to the labile phosphorus content of this
+            soil layer.
+
+        Parameters
+        ----------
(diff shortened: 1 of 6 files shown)


SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -70,6 +70,65 @@ def __post_init__(self):
         """Initialize all attributes in the dataclass that depend on other attributes"""
         self.water_content = self.soil_water_concentration * self.layer_thickness
 
+    def add_to_labile_phosphorus(self, phosphorus_to_add: float, field_size: float) -> None:
```
morrowcj, line 171: I have a question about this method. I fully support moving this to a higher level so that it can be called from multiple lower-level modules. However, I wonder if `LayerData` is the best home for it? Will there be an overall `phosphorus_cycling.py` file (as there is in the old code)? If the answer to that question is "yes", then I think methods like this belong there instead. 

What do you think?

morrowcj (approved):
I have only one potentially substantial comment regarding the structure of this PR. I'll leave the final decision about its resolution up to you.

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: I think that `phosphorus_cycling.py` could be an equally good place for the wrapper methods, but I think it would be better to keep the base method in `layer_data.py` because it could easily be extended to be the base method of any method that adds to (or removes from) a soil component pool in a layer of soil, as long it tracked in \<units\> / ha.

Moving the wrapper functions from `layer_data.py` to `phosphorus_cycling.py` *may* also make it significantly more complicated for other modules at the level of `phosphorus_cycling.py` to access said wrapper methods. An instance of `phosphorus_cycling.py` will have instances of `manure_application.py`, `fertilizer.py`, etc. as attributes, but not the other way around. And I can't think of a simple, clean way that these attributes would call a method from the instance that contains them (hopefully that makes sense). Again I emphasize that this may not be an issue because I might not be conceptualizing the problem right or am missing a clear solution.

Also, I am still unsure of whether or not a `phosphorus_cycling.py` module will be necessary (currently leaning towards a yes). My loose plan when I started was to implement all the sub-modules, then evaluate the most efficient, concise way for the `Soil` and `Field` modules to group and interact with them. Definitely the wrong way to go about this refactor though, should have written a design document and come up with a clear plan then (perhaps I should still write one?)

morrowcj: Yeah, I think your first paragraph may be the right path.

Your second point is a good one. Could a `@staticmethod` solve the problem?

This is all moot if `phosphorus_cycling.py` won't exist, so I'm OK with putting this on the backburner for now.

As for a design doc - it's up to you. I'm OK with you taking what ever path will be most efficient for you. I could see either way being valuable.

Isgotkowitz: Making the wrappers `@staticmethod`'s is one solution, probably the best. I consider this a less cleanly solution though because it would require passing the `LayerData` instance that is being modified as a parameter, which is less ideal than just passing the necessary values.

ltp23 (approved):
Good job, Ed!

I only have some comments about the file `layer_data.py`.

- The field `self.water_content` should probably be set to `field(init=False)` in the class definition to disallow any accidental initialization, since it is meant to be calculated in `__post_init__()`. 
- The `layer_thickness` property preferably shouldn't perform a value-error check. We should do the check when the condition is first violated like in the initialization phase, and not wait until we want to do some calculations. 
- Overall, I think this class may be getting a little too big. It would be great if you could find a way to break it down further when you know more.

Isgotkowitz:
> Good job, Ed!
> 
> I only have some comments about the file `layer_data.py`.
> 
> * The field `self.water_content` should probably be set to `field(init=False)` in the class definition to disallow any accidental initialization, since it is meant to be calculated in `__post_init__()`.
> * The `layer_thickness` property preferably shouldn't perform a value-error check. We should do the check when the condition is first violated like in the initialization phase, and not wait until we want to do some calculations.
> * Overall, I think this class may be getting a little too big. It would be great if you could find a way to break it down further when you know more.

Thanks Loi, I did not know about `field(init=False)` but that is handy for working with `@dataclasses`.

Refactoring error-raising for `layer_thickness` is an issue so it will eventually be taken care of.

Is there a certain size or rule of thumb you use when you're deciding if a module (specifically a `@dataclass`) is too big?

PR #412, opened 2023-03-29
Mini Design Doc: `phosphorus_cycling/`

Isgotkowitz:
This is a design doc for the phosphorus cycling submodule, though not a full one.

Note: 
- 2 todo's were removed, they were related to future issues so I created a GitHub issue and took them out
- accidentally wrote this on the wrong git branch, so it's now stacked on `add_P_to_soil_SCred` branch (that PR should be closed soon though)

### DIFF
--- a/SC_redesign/docs_SC/Admin/phosphorus_design_doc.md
+++ b/SC_redesign/docs_SC/Admin/phosphorus_design_doc.md
@@ -49,7 +49,6 @@ and in the soil, including
   * Manure applied with a machine.
   * Manure applied with by animals grazing in said field.
 
-### Other Requirement Details
 * This section does not flush out the requirements that may be addressed in future versions of this module. See 
 [Beyond Version 1](#beyond-version-1)
 
@@ -63,7 +62,7 @@ based on the user-defined input.
 
 ## Proposed Solution
 Automatically divide the user-defined top layer of soil into two layers: a top layer that is 20 mm thick, so that the 
-simulation can follow critical parts of the SurPhos model, and then a second layer. This second layer is the same as the
+simulation can follow critical parts of the SurPhos model and then a second layer. This second layer is the same as the
 user-defined top layer of soil except for the fact that it's top depth is 20 mm below where the user specified it was.  
 
 Most of the Soil and Crop modules outside of PC will not have their operations disrupted by this new top layer, however
--- a/SC_redesign/docs_SC/Admin/phosphorus_design_doc.md
+++ b/SC_redesign/docs_SC/Admin/phosphorus_design_doc.md
@@ -38,7 +38,7 @@ module, described in the above document describing the Soil and Crop redesign.
 ---
 
 ## Requirements
-This section will list the requirements of PC in order to be integrated with and used by other modules. 
+This section will list the features of PC required in order to be integrated with and used by other modules. 
 * Provide a composite class that can be used by higher level modules to efficiently and effectively manage phosphorus on
 and in the soil, including
   * Phosphorus content on the soil surface.
@@ -73,7 +73,7 @@ cases the model will treat the soil profile as if it had the original top layer 
 ---
 
 ## Alternative Solution
-Some solutions considered when this design was being flushed out:
+Some solutions considered when this design was being fleshed out:
 * Maintain a special set of pools in the soil profile that are used to track what is in the top 20 mm of the soil 
     profile.
   * Pros:
--- a/SC_redesign/docs_SC/Admin/phosphorus_design_doc.md
+++ b/SC_redesign/docs_SC/Admin/phosphorus_design_doc.md
@@ -38,63 +38,57 @@ module, described in the above document describing the Soil and Crop redesign.
 ---
 
 ## Requirements
-This section will outline how the module is/will be designed and organized in order to fulfill its prescribed 
-functionality. It will also make note of any important details that determine how the module will function.
-
-### Phosphorus Cycling
-
-
-### Fertilizer
-
-
-### Manure Application
-
-
-### Manure
-
-
-### Mineralization
-
-
-### Soluble Phosphorus
-
+This section will list the requirements of PC in order to be integrated with and used by other modules. 
+* Provide a composite class that can be used by higher level modules to efficiently and effectively manage phosphorus on
+and in the soil, including
+  * Phosphorus content on the soil surface.
+  * Phosphorus content in each layer of the soil profile.
+  * Amount of phosphorus that have been lost from the field due to runoff.
+* Provide ability to add phosphorus via fertilizer to a field.
+* Provide ability to add phosphorus via manure to a field, with options for
+  * Manure applied with a machine.
+  * Manure applied with by animals grazing in said field.
 
 ### Other Requirement Details
-* In order to accurately simulate phosphorus lost via runoff, the soil profile will need to maintain a set of phosphorus
-pools that represent phosphorus in the top 20 mm of the soil profile. This is a critical part of SurPhos, and is 
-discussed more in the [Open Questions](#open-questions) section.
 * This section does not flush out the requirements that may be addressed in future versions of this module. See 
 [Beyond Version 1](#beyond-version-1)
 
 ---
 
-## Open Questions
-* Currently (Mar 29, 2023), the module does not maintain a top layer soil with a depth of 20 mm (as required by 
-SurPhos), and there are many potential ways that this requirement could be realized:
-  * Override user input and always add a 20 mm top layer to the soil profile.
-    * Pros: 
-      * Does not require change in the implementation details of modules that have already been implemented.
-      * Implementing this override would likely be relatively simple (consisting of just cutting off the top 20 mm of 
-        the given top soil layer).
-    * Cons:
-      * Mutates input given directly by user with no option to override.
-      * May result in inaccurate modeling of other aspects of the soil profile that are not critically dependent on 
-        tracking the top 20 mm of the soil profile (i.e., water, other nutrients, etc.)
-  * Maintain a special set of pools in the soil profile that are used to track what is in the top 20 mm of the soil 
+## Existing Solution
+In the previous implementation of the Phosphorus Cycling module, the model simply maintained a soil profile that was 
+based on the user-defined input.
+
+---
+
+## Proposed Solution
+Automatically divide the user-defined top layer of soil into two layers: a top layer that is 20 mm thick, so that the 
+simulation can follow critical parts of the SurPhos model, and then a second layer. This second layer is the same as the
+user-defined top layer of soil except for the fact that it's top depth is 20 mm below where the user specified it was.  
+
+Most of the Soil and Crop modules outside of PC will not have their operations disrupted by this new top layer, however
+in a few specific cases processes that work with the whole soil profile are dependent on the top layer of soil. In these
+cases the model will treat the soil profile as if it had the original top layer of soil, as defined by the user.
+
+---
+
+## Alternative Solution
+Some solutions considered when this design was being flushed out:
+* Maintain a special set of pools in the soil profile that are used to track what is in the top 20 mm of the soil 
     profile.
-    * Pros:
-      * Enables most accurate picture of soil profile to be maintained.
-      * Does not mutate the soil profile specified by the user.
-    * Cons:
-      * Would likely result in an extremely complex implementation that touches many modules outside of Phosphorus 
-      Cycling.
-  * Do not simulate the top 20 mm of soil in the profile (this one is kind of a non-starter, but listed to fully flush 
-    out options).
-    * Pros:
-      * Does not require any additional implementation at all.
-      * Does not change user-input attributes at all.
-    * Cons:
-      * Severely hinders the accuracy of the SurPhos model that this module is based on.
+  * Pros:
+    * Enables most accurate picture of soil profile to be maintained.
+    * Does not mutate the soil profile specified by the user.
+  * Cons:
+    * Would likely result in an extremely complex implementation that touches many modules outside of Phosphorus 
+    Cycling.
+* Do not simulate the top 20 mm of soil in the profile (this one is kind of a non-starter, but listed to fully flush 
+out options).
+  * Pros:
+    * Does not require any additional implementation at all.
+    * Does not change user-input attributes at all.
+  * Cons:
+    * Severely hinders the accuracy of the SurPhos model that this module is based on.
 
 ---
 
@@ -174,3 +168,8 @@ This submodule will be responsible for maintaining an accurate state of the soil
 content. There are two main processes it will need to simulate in order to accomplish this
 * Erosion from the top layer of soil
 * Phosphorus movement between different layers in the soil profile
+
+### Other Details
+* The SurPhos model specifically tracks the amount of phosphorus in the top 20 mm of the soil profile, as this 
+phosphorus was determined to be critical to accurately tracking how much phosphorus was removed from the field in 
+runoff. To see how we integrate this into PC, see the [solution](#proposed-solution) that we have elected to implement.
--- a/SC_redesign/docs_SC/Admin/phosphorus_design_doc.md
+++ b/SC_redesign/docs_SC/Admin/phosphorus_design_doc.md
@@ -42,6 +42,77 @@ This section will outline how the module is/will be designed and organized in or
 functionality. It will also make note of any important details that determine how the module will function.
 
 ### Phosphorus Cycling
+
+
+### Fertilizer
+
+
+### Manure Application
+
+
+### Manure
+
+
+### Mineralization
+
+
+### Soluble Phosphorus
+
+
+### Other Requirement Details
+* In order to accurately simulate phosphorus lost via runoff, the soil profile will need to maintain a set of phosphorus
+pools that represent phosphorus in the top 20 mm of the soil profile. This is a critical part of SurPhos, and is 
+discussed more in the [Open Questions](#open-questions) section.
+* This section does not flush out the requirements that may be addressed in future versions of this module. See 
+[Beyond Version 1](#beyond-version-1)
+
+---
+
+## Open Questions
+* Currently (Mar 29, 2023), the module does not maintain a top layer soil with a depth of 20 mm (as required by 
+SurPhos), and there are many potential ways that this requirement could be realized:
+  * Override user input and always add a 20 mm top layer to the soil profile.
+    * Pros: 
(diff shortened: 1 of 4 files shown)


SC_redesign/docs_SC/phosphorus_design_doc.md
```diff
@@ -0,0 +1,152 @@
+# Design Document: Phosphorus Cycling Module Redesign
+
+Authors: Ed Hansen  
+Date Created: 27 Mar 2023  
+Last Updated: 27 Mar 2023  
+Reviewers: [Add your names here!]
```
morrowcj:
```suggestion
Reviewers: Clay Morrow
```

SC_redesign/docs_SC/phosphorus_design_doc.md
```diff
@@ -0,0 +1,152 @@
...
+   f. [Mineralization](#mineralization)  
+   g. [Soluble Phosphorus](#soluble-phosphorus)
+   i. [Other Requirements](#other-requirement-details)
+4. [Open Questions](#open-questions)
+5. [Beyond Version 1](#beyond-version-1)
+
+--- 
+
+## Overview
+This document will detail how the redesigned Phosphorus Cycling (PC) submodule within the Soil module of RuFaS. The redesign
```
morrowcj: I think you're missing a verb in this sentence.

```suggestion
This document will detail how the redesigned Phosphorus Cycling (PC) submodule [works? fits?] within the Soil module of RuFaS. The redesign
```

SC_redesign/docs_SC/phosphorus_design_doc.md
```diff
@@ -0,0 +1,152 @@
...
+that is described in this document is a part of the larger Soil and Crop redesign project, on which details can be found
+here: [design-document_SC-redesign.md](design-document_SC-redesign.md).
+
+---
+
+## Context
+The Phosphorus Cycling submodule is based largely on the SurPhos model for predicting Phosphorus lost from the soil 
+surface in runoff. It was designed specifically to improve on the shortcomings of the SWAT model in this aspect.
```
morrowcj:
```suggestion
The Phosphorus Cycling submodule is based largely on the SurPhos model for predicting the portions of phosphorus that are lost from the soil 
surface in runoff. It was designed specifically to improve on the shortcomings of the SWAT model in this aspect. 
```

SC_redesign/docs_SC/phosphorus_design_doc.md
```diff
@@ -0,0 +1,152 @@
...
+   i. [Other Requirements](#other-requirement-details)
+4. [Open Questions](#open-questions)
+5. [Beyond Version 1](#beyond-version-1)
+
+--- 
+
+## Overview
+This document will detail how the redesigned Phosphorus Cycling (PC) submodule within the Soil module of RuFaS. The redesign
+that is described in this document is a part of the larger Soil and Crop redesign project, on which details can be found
+here: [design-document_SC-redesign.md](design-document_SC-redesign.md).
```
morrowcj: double check this link. I think that it might be in the `Admin` sub-folder. This document should probably be in the same directory as the main design doc. Rename/reorganize stuff if you need to.

SC_redesign/docs_SC/phosphorus_design_doc.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Requirements
+This section will outline how the module is/will be designed and organized in order to fulfill its prescribed 
+functionality. It will also make note of any important details that determine how the module will function.
+
+### Phosphorus Cycling
+This submodule will serve as the composite class for PC. It will include
+- a main routine that can be called by higher-level modules and will execute all daily updates on phosphorus already in
+the field.
+- methods used by multiple PC sub-modules
```
morrowcj: Is this revision accurate?

```suggestion
- calls to methods from multiple PC sub-modules
```

SC_redesign/docs_SC/phosphorus_design_doc.md
```diff
@@ -0,0 +1,152 @@
...
+
+It will be structured the same as its counterpart composite classes in higher level modules with one notable difference:
+all states related to phosphorus will be maintained in `SoilData` and `LayerData` instances, not in a 
+`@dataclass` specific to PC.
```
morrowcj: A few suggestions (check the link, please - I'm doing this in a browser):

```suggestion
It will be structured similar to other [Soil and Crop classes](design-document_SC-redesign.md#module-design-and-structure), with one notable difference:
all states related to phosphorus will be maintained in `SoilData` and `LayerData` instances, not in a 
`@dataclass` specific to PC.
```

SC_redesign/docs_SC/phosphorus_design_doc.md
```diff
@@ -0,0 +1,152 @@
...
+soil surface. There are two methods that can be called by higher-level modules:
+1. add_fertilizer_phosphorus(), which adds phosphorus to fertilizer phosphorus surface pools
+2. do_fertilizer_phosphorus_operations(), which conducts daily leaching and runoff activities based on only 3 
+parameters:
+   * the amount of rainfall on the current day
+   * the amount of runoff on the current day
+   * the size of the field on which these operations are occurring
+
+The `Fertilizer` submodule is designed with the assumption that there will not be frequent applications of phosphorus 
+via fertilizer (~1 time per year). This reflects real-world agricultural practices.
```
morrowcj:
```suggestion
via fertilizer (~1 time per year). This reflects the most common real-world agricultural practices.
```

SC_redesign/docs_SC/phosphorus_design_doc.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Manure Application
+This submodule handles **applying** phosphorus from manure to a field (which is, arguably, a shittier way to apply 
+phosphorus than applying via fertilizer). When implementing this submodule, the original intent was to have everything
```
morrowcj: I like the pun, but let's church it up a little bit.

```suggestion
This submodule handles **applying** phosphorus from manure to a field (which is, arguably, a crappier way to apply 
phosphorus than applying via fertilizer). When implementing this submodule, the original intent was to have everything 
```

SC_redesign/docs_SC/phosphorus_design_doc.md
```diff
@@ -0,0 +1,152 @@
...
+   * the amount of runoff on the current day
+   * the size of the field on which these operations are occurring
+
+The `Fertilizer` submodule is designed with the assumption that there will not be frequent applications of phosphorus 
+via fertilizer (~1 time per year). This reflects real-world agricultural practices.
+
+### Manure Application
+This submodule handles **applying** phosphorus from manure to a field (which is, arguably, a shittier way to apply 
+phosphorus than applying via fertilizer). When implementing this submodule, the original intent was to have everything 
+in the `Manure` module but the application operations became too complex, and necessitated having their own module.
```
morrowcj:
```suggestion
in the `Manure` module but the application operations became too complex, and necessitated separate modules. 
```

SC_redesign/docs_SC/phosphorus_design_doc.md
```diff
@@ -0,0 +1,152 @@
...
+
+The `Fertilizer` submodule is designed with the assumption that there will not be frequent applications of phosphorus 
+via fertilizer (~1 time per year). This reflects real-world agricultural practices.
+
+### Manure Application
+This submodule handles **applying** phosphorus from manure to a field (which is, arguably, a shittier way to apply 
+phosphorus than applying via fertilizer). When implementing this submodule, the original intent was to have everything 
+in the `Manure` module but the application operations became too complex, and necessitated having their own module. 
+There are two methods to be used by higher level modules. One is for manure applied by animals grazing in a field, and 
+the other is for manure applied by machine. Each type of manure has its own set of pools and factors that includes
```
morrowcj:
```suggestion
There are two methods to be used by higher level modules. One is for manure applied by animals grazing in a field [grazing not yet implemented in RuFaS], and 
the other is for manure applied by machine. Each type of manure has its own set of pools and factors that includes
```

SC_redesign/docs_SC/phosphorus_design_doc.md
```diff
@@ -0,0 +1,152 @@
...
+Unlike the `Fertilizer` submodule, `ManureApplication` is designed to handle more frequent applications of manure 
+phosphorus. It does this by resetting the various factors it maintains to an average weighted by masses of the new 
+phosphorus application and the phosphorus already on the field.
+
+### Manure
+The `Manure` submodule handles all daily operations on the phosphorus from manure that is currently on the field, and 
+will only have one method that gets called by higher-level modules. These operations consist of
+* Transferring phosphorus from the surface pools to soil phosphorus pools
+* Transferring phosphorus between different surface pools to reflect the chemical composition of the manure changing
+* Removing phosphorus from the surface pools as runoff occurs.
```
morrowcj: By my reading, almost all of the subsections under `##Requirements` aren't actually requirements *per se*. Rather, they are design details, which should go in the last section according to Pooya's template. 

For actual requirements, think higher level: what does the phosphorus module *need to do* within RuFaS and what are the tangible ways that we will know that these requirements have been met? This section does not need to be long. A single paragraph or short bulleted list is probably sufficient.

For example, the phosphorus module should:
* simulate the amount of phosphorus lost to runoff and the amount that remains in the field for other processes
* determine this process based on environmental factors (weather, topography/runoff, literature [for volatilization])
* discriminate between phosphorus added via manure or via fertilizer
* etc. 

Does that make sense?

SC_redesign/docs_SC/phosphorus_design_doc.md
```diff
@@ -0,0 +1,152 @@
...
+pools that represent phosphorus in the top 20 mm of the soil profile. This is a critical part of SurPhos, and is 
+discussed more in the [Open Questions](#open-questions) section.
+* This section does not flush out the requirements that may be addressed in future versions of this module. See 
+[Beyond Version 1](#beyond-version-1)
+
+---
+
+## Open Questions
+* Currently (Mar 29, 2023), the module does not maintain a top layer soil with a depth of 20 mm, and there are many 
+potential ways that this requirement could be realized:
```
morrowcj:
```suggestion
* Currently (Mar 29, 2023), the module does not maintain a top layer soil with a depth of 20 mm (as required by SurPhos), and there are many 
potential ways that this requirement could be realized:
```

SC_redesign/docs_SC/phosphorus_design_doc.md
```diff
@@ -0,0 +1,152 @@
...
+### Other Requirement Details
+* In order to accurately simulate phosphorus lost via runoff, the soil profile will need to maintain a set of phosphorus
+pools that represent phosphorus in the top 20 mm of the soil profile. This is a critical part of SurPhos, and is 
+discussed more in the [Open Questions](#open-questions) section.
+* This section does not flush out the requirements that may be addressed in future versions of this module. See 
+[Beyond Version 1](#beyond-version-1)
+
+---
+
+## Open Questions
```
morrowcj: Let's make a decision and then move the things in this section to "Proposed Solution" and "Alternative Solutions".

SC_redesign/docs_SC/phosphorus_design_doc.md
```diff
@@ -0,0 +1,152 @@
...
+      * Would likely result in an extremely complex implementation that touches many modules outside of Phosphorus 
+      Cycling.
+  * Do not simulate the top 20 mm of soil in the profile (this one is kind of a non-starter, but listed to fully flush 
+    out options).
+    * Pros:
+      * Does not require any additional implementation at all.
+      * Does not change user-input attributes at all.
+    * Cons:
+      * Severely hinders the accuracy of the SurPhos model that this module is based on.
+
```
morrowcj: My suggestion (and by no means the last word): Why don't we simply subdivide the top layer into two pools (e.g., `top_twenty_phosphorus`, and `top_remainder_phosphorus`) within the phosphorus module, but continue treating the layer as a whole for everything else? This would simply involve summing the two pools before passing data outside of P cycling. Perhaps this is defeats the entire purpose, though. 

In that case, we could permanently subdivide the layer and treat the top 20mm as its own layer (with identical composition the the rest of the upper layer but with depths changed). Most soil methods already handle a variable number of layers (at least `Crop` does). The only place that I can see this needing additional attention is where the "top layer" attributes are explicitly used, in which case we need to aggregate the top 20 layer with the rest of the top layer. It shouldn't be too difficult to track those instances down though, right?

SC_redesign/docs_SC/phosphorus_design_doc.md
```diff
@@ -0,0 +1,152 @@
...
+---
+
+## Beyond Version 1
```
morrowcj: excellent.

morrowcj (changes requested):
Ed, Kudos on this great work. This is a **totally** different skill set from programming and you've done a really fine job - your communication skills are good and improving constantly. In addition to wording/grammar suggestions, there are a few substantial points that I think we should address together. Hopefully my comments speak for themselves but please reach out for clarification or further discussion if needed.

SC_redesign/docs_SC/phosphorus_design_doc.md
Isgotkowitz: I don't think so, because all the method calls will be made in the other PC sub-modules, and the composite `PhosphorusCycling` class is the place where are those methods are defined. Is that clearer?

morrowcj: OK, then disregard! Thanks.

SC_redesign/docs_SC/phosphorus_design_doc.md
Isgotkowitz: Lol yes, sorry about that. I just couldn't dismiss the little voice in the back of my head saying "Pooya said to make it funny..."

SC_redesign/docs_SC/phosphorus_design_doc.md
Isgotkowitz: Yes, it does

SC_redesign/docs_SC/phosphorus_design_doc.md
Isgotkowitz: Sounds good

SC_redesign/docs_SC/phosphorus_design_doc.md
Isgotkowitz: >My suggestion (and by no means the last word): Why don't we simply subdivide the top layer into two pools (e.g., top_twenty_phosphorus, and top_remainder_phosphorus) within the phosphorus module, but continue treating the layer as a whole for everything else? This would simply involve summing the two pools before passing data outside of P cycling. Perhaps this is defeats the entire purpose, though.

The problem with summing the pools (in my mind) is that there is a large loss of information as soon as we do so. If we were to sum them and then the data gets passed to the part of `Crop` that uptakes phosphorus (just one example), it will see only the one pool and uniformly draw phosphorus from it, then the next day the simulation gets back to P cycling and finds the top pool has significantly less phosphorus, or something like that. We should have a meeting about this to flush out details though

SC_redesign/docs_SC/phosphorus_design_doc.md
Isgotkowitz: Will do.

SC_redesign/docs_SC/phosphorus_design_doc.md
Isgotkowitz: Nice catch

SC_redesign/docs_SC/phosphorus_design_doc.md
morrowcj: Yeah, that's a very good point and sort of defeats the purpose. What do you think of my second point, then?

Isgotkowitz: I think it is a good one, and off the top of my head I can't think of any instances where the top layer attributes are used specifically

SC_redesign/docs_SC/Admin/phosphorus_design_doc.md
```diff
@@ -0,0 +1,175 @@
...
+  * Phosphorus content in each layer of the soil profile.
+  * Amount of phosphorus that have been lost from the field due to runoff.
+* Provide ability to add phosphorus via fertilizer to a field.
+* Provide ability to add phosphorus via manure to a field, with options for
+  * Manure applied with a machine.
+  * Manure applied with by animals grazing in said field.
+
+### Other Requirement Details
+* This section does not flush out the requirements that may be addressed in future versions of this module. See 
+[Beyond Version 1](#beyond-version-1)
```
morrowcj: I don't think this needs its own heading.

SC_redesign/docs_SC/Admin/phosphorus_design_doc.md
```diff
@@ -0,0 +1,175 @@
...
+The Phosphorus Cycling submodule is based largely on the SurPhos model for predicting the portions of phosphorus that 
+are lost from the soil surface in runoff. It was designed specifically to improve on the shortcomings of the SWAT model 
+in this aspect. Currently (March 2023), the state of the Phosphorus Cycling submodule 
+(`RUFAS/routines/field/soil/phosphorus_cycling/`)is in the same state of disarray as the rest of the Soil and Crop 
+module, described in the above document describing the Soil and Crop redesign.
+
+---
+
+## Requirements
+This section will list the requirements of PC in order to be integrated with and used by other modules.
```
morrowcj:
```suggestion
This section will list the features of PC required in order to be integrated with and used by other modules. 
```

SC_redesign/docs_SC/Admin/phosphorus_design_doc.md
```diff
@@ -0,0 +1,175 @@
...
+user-defined top layer of soil except for the fact that it's top depth is 20 mm below where the user specified it was.  
+
+Most of the Soil and Crop modules outside of PC will not have their operations disrupted by this new top layer, however
+in a few specific cases processes that work with the whole soil profile are dependent on the top layer of soil. In these
+cases the model will treat the soil profile as if it had the original top layer of soil, as defined by the user.
+
+---
+
+## Alternative Solution
+Some solutions considered when this design was being flushed out:
```
morrowcj: I'm not sure if this is a pun or a typo.
```suggestion
Some solutions considered when this design was being fleshed out:
```

morrowcj (approved):
Looks good to me.

matthew7838 (approved):
Looks good to me, again, this is really helpful documentation to help me understand other modules. Thanks.

Issue #408, opened 2023-03-24
Make database files threadsafe


morrowcj:
# Problem
As @JoeWaddell pointed out in #402, running RuFaS in parallel might be problematic. This is because some data files (input JSONs, weather CSVs, etc.) may be accessed by multiple threads simultaneously. To prevent data corruption, steps should be taken to make our databases thread safe (whatever that means).

# Potential solutions
* Perhaps we can just explicitly lock files (read only) before reading from them to ensure that they can't be written to: [StackOverflow link](https://stackoverflow.com/questions/489861/locking-a-file-in-python)
* Perhaps using a database (SQL) format would solve the problem?

Actual computer people likely know much more about this than I do, so that is the extent to which I'm am comfortable making suggestions. 

# Context
I feel strongly having this implemented would greatly benefit sensitivity analyses, validation, or mathematical optimization. So, if it were done before those tasks, it would be very useful.

Isgotkowitz:
I think both solutions are good ones, both have pros and cons.

1. Making files read-only:
  - Pros: this should be easier to implement, would hopefully only require specifying what mode you are opening the file in. In an ideal world you should be able to have multiple processes open the same file in read-only mode at the same time and the OS will be smart enough to allow this because it is inherently thread-safe.
  - Cons: OS' are not always this smart, so it might only allow this to happen after importing a bunch of libraries to ensure thread-safety. Also, it may vary by OS which would increase the time needed to implement a complete solution.

2. Using a database:
  - Pros: this solution would scale better and (ideally) make storage and management of large datasets easier and more standardized. It also puts the burden of thread-safety on the database system that is chosen.
  - Cons: setting up a dedicated database that the entire RuFaS team can use may require a significant amount of dev-hours, and we'll need to take time to get everyone up to speed on how to run the database.

@ all the other computer people, please read this comment carefully and correct anything I said that is wrong, or that you disagree with.

Isgotkowitz:
@JoeWaddell can you share the code you ran that caused issues for you?

JoeWaddell:
> @JoeWaddell can you share the code you ran that caused issues for you?

@Isgotkowitz Here's the crudest version: the first comments explain what works on my system. The latter attempts steps were useful in seeing what crashes, but didn't work. At the very least, you can see my thought process as I worked my way down the script. I'll refrain from posting this code as a proper file for now, since I don't want to confuse folks.

[...]

PR #413, opened 2023-03-29
Examples for conducting SA on RuFaS modules.

morrowcj:
## Summary
Provides an example demonstrating how to use the `SensitivityAnalysis` class to perform SA on RuFaS sub-modules. 

## Notes
* most of this PR (in terms of lines) is docstrings. I wanted to be as clear as possible, please let me know if I need to clarify anything.
* @JoeWaddell, I left a couple of `#TODO`s directed at you. 
* I really wanted `Example-animal_management_SA-wrapper.py` to be contained within `model_evaluation/sensitivity_analysis/`, but I **could not** figure out how to get the script working from anywhere except the root `MASM` folder. The problem seems to be relative paths of imports. @ltp23, do you have any ideas on how to do what I want?


model_evaluation/sensitivity_analysis/RuFaS-sensitivity-analysis-guide.md
morrowcj, line 1: @JoeWaddell and @varrma, would you review this file and make sure I didn't miss anything from our discussion?
Also, take a look at the figure that I created for it `SA_conceptual_model.jpg`. 

 I'm still working on `Example-animal_management_SA_wrapper.py`, which I'll get back to on Monday and will probably solicit some help from @JoeWaddell. 

-----
This file is an extension of the README.md file from the last PR in this branch. The "Summary of SA Module" section was largely unchanged and can be skipped.

Isgotkowitz (approved):
Nice job Clay. The documentation on this is really good

JoeWaddell:
> * @ltp23, do you have any ideas on how to do what I want?

image
Not ltp23, but this is the way I've imported from MASM without working in the base directory. Importing works fine, but as you can see VSCode gets a little goofy trying to understand where modules coming from, so it gets a little heavy-handed with the underlining. There may be a way to resolve that in my editor, but I only use this method in cases where I want to be certain that a method/function works as expected outside of the RUFAS ecosystem. 

Loi likely has a more elegant solution, but this may serve your needs in the interim.

model_evaluation/sensitivity_analysis/generalized_sensitivity.py
```diff
@@ -32,6 +32,9 @@ class SupportedSensitivityMethods(Enum):
     MORRIS = "morris"
     # TODO: add fractional factorial method
 
+# TODO: for some reason, the docstring formatting in this and Example-animal_management_SA-wrapper.py isn't rendering
+#   properly (e.g., equations) in the Pycharm preview. They may need to be tweaked.
+
 
 class SensitivityAnalysis:
```
JoeWaddell, line 39: I think here we'll want to partition this either into two different classes, or two main options within the main Sensitivity Analysis class, reflecting the conceptual model. Full-model versus Sub-model

You've done a wonderful job of writing support functions for the wrapper method that we'll use in sub-model analyses. I'll put together the pieces of what I wrote for the Full-model analysis. I'll outline that process in the analysis guide, and put it together from the helper functions I wrote previously (since it was the method I was using prior). It'll be contingent on resolution of issue #408

model_evaluation/sensitivity_analysis/RuFaS-sensitivity-analysis-guide.md
```diff
@@ -0,0 +1,262 @@
...
+
+An alternative solution would be to simply conduct the global SA of the entire RuFaS module, but this would lead to a 
+relatively poor understanding of the whole system 
+(see [Additional Discussion of RuFaS SA](#additional-discussion-of-rufas-sa) below). 
+
+---
+
+## Testability, Monitoring, and Alerting
+
+I am unsure how to test the `SensitivityAnalysis` module, and I am unsure if it even needs to be tested. Perhaps
```
JoeWaddell: Once we're settled on the approach, we can add a test file for unit and integration tests, to ensure that individual methods are working as expected. 

As for a sort of integration test, I think it would be useful to set up a dummy function that performs some rudimentary calculation based on a set of inputs, and we test the process from start to finish on this. We could even use the example you've already made for the SA-wrapper on animal_management.

JoeWaddell (commented):
This is looking great! I've added some comments and a fair bit of text in a separate commit.

varrma (approved):

model_evaluation/sensitivity_analysis/generalized_sensitivity.py
morrowcj: @JoeWaddell, could you elaborate on this a bit more? As I see it, this class could be used for the full model analysis too, but I might be missing something.

[edit] never mind, I see that it is in your commit. I'll look over that now.

model_evaluation/sensitivity_analysis/RuFaS-sensitivity-analysis-guide.md
```diff
@@ -101,12 +101,87 @@ An alternative solution would be to simply conduct the global SA of the entire R
 relatively poor understanding of the whole system 
 (see [Additional Discussion of RuFaS SA](#additional-discussion-of-rufas-sa) below). 
 
+Comment from @JoeWaddell: I am going to outline my process for this alternative solution below. My proposal is that we develop - either as a separate class entirely, or as an option within the existing SensitivityAnalysis class - a global method, despite the drawbacks of that approach.
+
+## Full-model analysis, e.g. global sensitivity analysis
+
+@JoeWaddell is formalizing global/full-model Sensitivity Analysis for RuFaS.
+As outlined above, RuFaS involves many stochastic processes, and any interpretation of outputs must be made
+ with careful consideration to initial context.
```
morrowcj, line 110: > RuFaS involves many stochastic processes

@JoeWaddell, is this actually true? Does the model itself contain stochasticity, or is it simply that it depends on stochastic input (e.g., weather)? I was under the impression that this was a mostly deterministic model (Soil and Crop is).

model_evaluation/sensitivity_analysis/RuFaS-sensitivity-analysis-guide.md
```diff
@@ -101,12 +101,87 @@ An alternative solution would be to simply conduct the global SA of the entire R
...
 
-I am unsure how to test the `SensitivityAnalysis` module, and I am unsure if it even needs to be tested. Perhaps
-some aspects should have formal tests, but because it is a stochastic process, it will be difficult.
+Because RuFaS involves many stochastic processes, testing full-model sensitivity analysis will be difficult.
+
+Nevertheless, once we've settled on the approach, we can start to build  unit tests to ensure that individual methods are working as expected.
+
+Likewise, we will develop integration tests, so we can ensure that a given dummy function performs some rudimentary calculation correctly.
+With a set of inputs, and we test the process from start to finish.
+The example SA-wrapper on animal_management is an alternative candidate.
```
morrowcj, line 183: I like this. Thank you, @JoeWaddell.

model_evaluation/sensitivity_analysis/generalized_sensitivity.py
JoeWaddell: @morrowcj I think it's certainly possible!  So I outlined my process for running full-model analyses in the .md file: it's what we've discussed previously as a group, I just wanted to include it in the ongoing document. The idea being that the process outlined can serve as the basis for a different method within this class - essentially, it requires an `objective_function` that's already "ready" for a general whole model analysis.

The reason I speculated we could make a separate class is since the workflow is so different - since it requires modifying files themselves, instead of attributes in an instance of some portion of RuFaS (like in the `daily_update` example). The process I've been using involves pregenerating modified jsons (to make it robust to parallelization), and tidying the workspace following an analysis. What I've added in the markdown file is a bit too wordy, so please edit/comment as you (or anyone reading this) see fit. 

Edit: Also, it goes without saying (but I'll say it anyway) that it's entirely possible I'm missing something, too! There may be a way to make RuFaS able to handle modification of those inputs in parallel, once issue #408 is solved

model_evaluation/sensitivity_analysis/RuFaS-sensitivity-analysis-guide.md
varrma: Great work on this Clay. As Joe mentioned, we can start with having two options in performing these tests.

model_evaluation/sensitivity_analysis/generalized_sensitivity.py
morrowcj: I just read over what you'd written, @JoeWaddell and I think it is great. I think we can simplify the full model (easily) to more better facilitate these types of analyses. `run_rufas()` is the main method that runs the full simulation. 

I propose that, rather than using the JSON files directly, we retool that method to run from a config class or dictionary (that is itself created from the JSON files in the normal executions). It would allow us to create/tweak the config object directly when running the objective function rather than constantly overwriting JSON files. This would also solve the parallelization problem, since a different object would be initialized with different values on each thread.

What do you think about that? If it seems useful, I can get started on that - I don't think it will be too difficult.

model_evaluation/sensitivity_analysis/RuFaS-sensitivity-analysis-guide.md
JoeWaddell: It is mostly deterministic, but there are elements that are stochastic. Ones that come to mind are the culling methods and pregnancy losses: however what I understand is that since we're currently fixing the seed, it won't vary right now.

So when its in the hand of a user, it will vary under identical starting conditions, since there are some metrics that will be random (with rates defined in various jsons). If you look at `['culling']` inside the animal_management_animal.json, you'll see ranges used when randomly determining cull reasons/probabilities.

morrowcj: This is good to know! I had no idea.

model_evaluation/sensitivity_analysis/generalized_sensitivity.py
JoeWaddell: That sounds like a great plan! That may solve the questions I had about changing values in "constants" files, as well (or at the least it can provide guidance on how we should restructure those files to accommodate SA methods).

PR #414, opened 2023-03-30
Reconfigure `flake8` to fail build if it finds any errors

Isgotkowitz:
This PR removes the second `flake8` run that happens when checking code before approving it to merge into `SC_redesign`. 

It also fixes the handful of linter errors that were present in `SC_redesign` when this reconfiguration was done.

### DIFF
--- a/.github/workflows/lint_flake8.yml
+++ b/.github/workflows/lint_flake8.yml
@@ -27,7 +27,5 @@ jobs:
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
     - name: Lint with flake8
       run: |
-        # stop the build if there are Python syntax errors or undefined names
-        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
-        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
-        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+        # stop the build if there are any errors or warnings
+        flake8 SC_redesign/ --count --show-source --statistics --max-line-length=120 --max-complexity=10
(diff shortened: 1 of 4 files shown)


.github/workflows/lint_flake8.yml
```diff
@@ -27,7 +27,5 @@ jobs:
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
     - name: Lint with flake8
       run: |
-        # stop the build if there are Python syntax errors or undefined names
-        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
```
morrowcj, line 31: What are we losing by removing the `--select` flag? I don't understand what these values indicate.

morrowcj (approved):

.github/workflows/lint_flake8.yml
Isgotkowitz: Those are different error codes. It tells flake8 to stop the build **only** if it finds errors listed in the `select` flag. Check out [error codes](https://flake8.pycqa.org/en/latest/user/error-codes.html) for more info

matthew7838 (approved):
Looks like a good change to linting. One question, are the changes to other files addressing the warning issues brought up by the updated lint?

Isgotkowitz:
> Looks like a good change to linting. One question, are the changes to other files addressing the warning issues brought up by the updated lint?

Yes and no. The issues were raised by both versions of the linter, but in the old version it was set so that they were not fatal and you wouldn't know they had been found unless you looked through the giant list of warnings raised by the linter after running over the entirety of the codebase. Which is I am very gung-ho on changing the linting config, the old workflow IMO has very bad usability.

With the new linting workflow when there are formatting errors, it will (fingers crossed) fail the linting check, list where all the errors were found, and won't let the PR be merged until they get fixed.

PR #416, opened 2023-03-31
Override user input to create new top layer of soil (1 / 2)

Isgotkowitz:
# Summary

This is the first PR in a series that overrides user-defined the soil profile by adding a 20 mm thick top layer, and refactors the other modules in `soil/` (outside of `phosphorus_cycling/`) to account for this new top layer. The need for this top layer comes from SurPhos, which tracks phosphorus content specifically in the top 20 mm of soil because that is where most phosphorus that comes from manure and fertilizer stays, and tracking the concentrated part near the soil surface is crucial to accurately predicting phosphorus runoff.

# Outline for this PR series

1. (this PR) implements the overriding of user-given soil specification so that there is always a top layer of soil that is 20 mm thick. Also refactors all old tests that deal with the soil profile so they don't fail. 
2. Refactor modules (hopefully just 1-2) so that they treat the top two layers in the soil profile as one to ensure the integrity of non-SurPhos modules.
3. Refactor additional modules that were not covered in second PR of this series. This one may not happen, depends on how many modules end up being refactored.

# Main Changes:

- `soil_data.py`:
  - Change defaults so default soil profile has the top 20 mm soil layer.
  - Added `_subdivide_top_layer()`, is called in `post_init()` when user defines a soil profile.
  - **PLEASE CRITIQUE** specifically how `_subdivide_top_layer()` is implemented and the bounds used by `post_init()` to decide whether or not to call it .
- Test files for `Evapotranspiration`, `Infiltration`, `PoolGasPartition`, `Decomposition`, and `SoilTemperature`
  - These test files now have much thicker, more realistic (I think) soil layers.
  - The number of times subroutines are assumed to have been called is now a function of how many layers of soil are in the profile being tested, instead of being hardcoded. 
- `test_soil_data_and_config_factory.py`
  - Tests that new default soil profile is properly created.
  - Tests that user-defined soil profile is properly overridden, and error is raised when inappropriate profile is passed.
  - Same changes as in test files listed above.
- `test_percolation.py` 
  - Last test case in `test_percolate_between_layers()` now has a `LayerData` method patched so it correctly follows that control branch is supposed to be tested in this case
  - Same changes as in the changes to the long list of test files above.

### DIFF
--- a/SC_redesign/Crop_and_Soil/soil/soil_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/soil_data.py
@@ -161,13 +161,23 @@ class SoilData:
 
     def __post_init__(self):
         """This method initializes attributes that either cannot be set to a default above or depend on other
-            attributes in the object to be set before they can be set"""
+            attributes in the object to be set before they can be set
+
+        Raises
+        ------
+        ValueError
+            If the bottom depth of the top layer of soil is < 20
+
+        """
         if self.soil_layers is None:
             self.soil_layers = [LayerData(top_depth=0, bottom_depth=20, nitrate=0.3),
                                 LayerData(top_depth=20, bottom_depth=50, nitrate=0.5),
                                 LayerData(top_depth=50, bottom_depth=80, nitrate=1),
                                 LayerData(top_depth=80, bottom_depth=200, nitrate=5)]
-        else:
+        elif self.soil_layers[0].bottom_depth < 20:
+            raise ValueError(f"Expected bottom depth of top soil layer must be 20 mm or greater, received "
+                             f"'{self.soil_layers[0].bottom_depth}'.")
+        elif self.soil_layers[0].bottom_depth > 20:
             self._subdivide_top_layer()
 
         if self.vadose_zone_layer is None:
--- a/SC_redesign/Crop_and_Soil/soil/soil_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/soil_data.py
@@ -1,6 +1,7 @@
 from dataclasses import dataclass
 from typing import List, Optional
 from math import inf
+from copy import deepcopy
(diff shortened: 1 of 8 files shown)


SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -160,12 +161,24 @@ class SoilData:
...
         if self.soil_layers is None:
-            # sets the soil layers to a default set if user does not provide any
-            self.soil_layers = [LayerData(top_depth=0, bottom_depth=50, nitrate=0.5),
+            self.soil_layers = [LayerData(top_depth=0, bottom_depth=20, nitrate=0.5),
+                                LayerData(top_depth=20, bottom_depth=50, nitrate=0.5),
                                 LayerData(top_depth=50, bottom_depth=80, nitrate=1),
                                 LayerData(top_depth=80, bottom_depth=200, nitrate=5)]
+        elif self.soil_layers[0].bottom_depth < 20:
+            raise ValueError(f"Expected bottom depth of top soil layer must be 20 mm or greater, received "
+                             f"'{self.soil_layers[0].bottom_depth}'.")
```
morrowcj, line 179: This is fine for now, but it may become a problem later. There are soil types where the layer might be very shallow. It may be a non-issue, though, so I'd cross that bridge when you come to it.

morrowcj (approved):
Looks great! Sorry for the delay.

SC_redesign/Crop_and_Soil/soil/soil_data.py
Isgotkowitz: Ok, noted. I will add an issue labeled future so we don't forget.

matthew7838 (approved):
Also looks good to me.

PR #418, opened 2023-04-03
Override user input to create new top layer of soil (2 / 2, refactor `Soil` modules)

Isgotkowitz:
# Summary

This is the second PR in the series to implement overriding user input so that the soil profile maintains a 20 mm thick top layer. This PR refactors the `Infiltration` and `SoilTemp` modules to account for this new top layer of soil (and tests their new functionality).

# Main Changes
- `SoilTemp`
  - Instead of passing just the temperature of the top layer of soil to the static helper methods, the top two layers have all necessary attributes either combined or averaged so that they are treated as one layer.
  - Note that because the top two layers of soil are set to the same temperature as each other every time the main routine is called, it is only possible for them to be different temperatures on the first day of the simulation, and have different previous temperatures on the first and second days of the simulation.
  - A `@staticmethod` helper has been introduced to calculate weighted average temperatures. It is tested with the rest of the helper methods.
- `Infiltration`
  - This file has had functionality added so that the soil will actually absorb the rainfall that does not runoff the field (a functionality that I did not originally implement for some reason), and this absorbed rainfall is proportionally divided between the top two layers of soil in the profile.
  - This new functionality is tested, and a helper method has been added in the test file to make this a bit cleaner.
  - A potential issue with the way water is absorbed in this module is that it is effectively skipping the percolation process for the top 20 mm of soil. This may have significant effects on the accuracy of the model, because water percolating through the top 20 mm of soil may have a significant effect on the phosphorus that is stored in that layer. It may be better to have all absorbed water stay in the top 20 mm, as this is more realistic and the amount of water absorbed is the same with either method. EDIT: this module's behavior has been changed. It now only absorbs water into the top 20 mm of soil, where it stays until the `Percolation` module causes it to move downward through the soil profile.

In both modules, please check the sections added to the docstrings for clarity and sensibility.

# Going Forward

I believe only modifying these two modules within `Soil` is sufficient. The other potential modules to refactor are `Evapotranspiration` and `SoilErosion`, which I think will retain their integrity even if they treat the top 20 mm of soil as a separate layer.

### DIFF
--- a/SC_redesign/Crop_and_Soil/soil/infiltration.py
+++ b/SC_redesign/Crop_and_Soil/soil/infiltration.py
@@ -18,6 +18,15 @@ class Infiltration:
         weighting_coefficient: weighting coefficient used to calculate retention coefficient for daily curve number
             calculations dependent on plant evapotranspiration (unitless)
 
+        Notes
+        -----
+        This module, in part, relies on the temperature of the top soil layer to determine infiltration. Because RuFaS
+        overrides the user-defined soil profile to maintain a top soil layer that is 20 mm thick for the purpose of
+        tracking phosphorus more accurately, this module distributes the water infiltrated between the two top layers of
+        proportionately by thickness. It also assumes that the top two layers of soil will have the same temperature
+        every day, which will be true for every day of the simulation as long as daily routine in `soil_temp.py` is run
+        prior to this routine.
+
         """
         third_moisture_condition_parameter = self._determine_third_moisture_condition_parameter(
                                                                         self.data.second_moisture_condition_parameter)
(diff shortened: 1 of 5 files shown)


morrowcj (approved):
This looks great! Nice work!

Isgotkowitz:
> This looks great! Nice work!

RUFAS/routines/field/soil/carbon_cycling/residue_partitioning.py
Isgotkowitz, line 1: The `== True` here was causing a Flake8 error, but was not caught by the GitHub action linting process because any branch that is being merged into `SC_redesign` only has flake8 run on `SC_redesign/`

ltp23 (approved):
Everything looks good, Ed!

One comment I have is sometimes I see we use 'temperature' and other times we use 'temp'. Similar patterns with min, max, avg maybe. I do this myself too. Do you have a preference?

Isgotkowitz:
> Everything looks good, Ed!
> 
> One comment I have is sometimes I see we use 'temperature' and other times we use 'temp'. Similar patterns with min, max, avg maybe. I do this myself too. Do you have a preference?

I don't have a preference, sometimes which I choose is a function of whether or not it is part of a much longer variable name.

PR #424, opened 2023-04-05
Removal of pyomo logic and related comments

JoeWaddell:
# This PR
Ration methods are due for a refactor, and this is the first in what will initially be a series of short PRs. 

I've started with removal of the pyomo logic, since it's no longer in use in the model. This code could be archived on Basecamp for posterity, in case it is useful for future application. For anyone reviewing, please a comment if you think this is needed or not. 

Likewise, I've removed comments associated with the pyomo logic, and in a few other places where it was clear the comments were no longer needed. 

# Next PR:
For the next PR I will focus on renaming functions used in the principal ration files, and I will bring the docstrings up to spec. For example: optimize, optimization, and objective are all too similar and not descriptive. 

# Subsequent PRs
Following these initial changes, the refactor will be a little more intensive. 

### `ration_driver.py`
Most functions in this file could use one or two helper functions for clarity.
The function in most need of refactoring: set_requirements needs to be broken up into multiple smaller functions. 
Finally, there is some conflict between the use of animal types and animal combinations when formulating ration - most notably coming into play when formulating a ration for the close up pen. 

### `ration_NLP.py`
There's a technical discussion to be had about whether we should be relying so heavily on globals in `ration_NLP.py`

### `calf_ration.py`
This one is possibly low priority since calves operate so differently with respect to other animals, but renaming the functions in this file such that they're clearly calf methods may be useful. 

### `animal_requirements.py`
We could consider a refactor so that individual methods are called on an animal object instead of a list, but this is also likely low priority.


KFosterReed (approved):
This PR is pretty straight forward to remove some old code from a previous method that was used for the NLP. The ration formulation works the same before and after the code was removed and I don't think there is any reason we need to store the old code in Basecamp. Simple but good improvement.

morrowcj (approved):
Looks good to me: a clean removal of the unnecessary code.

PR #429, opened 2023-04-07
Flake8 Config file and checking script

ew3361zh:
Working on a script to help check branch code for flake8 violations for users to be able to address these violations before pushing code to github

For testing, follow instructions in the flake8_instructions.md file which will eventually be what's put in the wiki. Flake8 should be able to be easily uninstalled after testing within the tester's virtual environment if desired.


cleanup.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+# git status -s collects all modified files both committed and uncommitted. -s gives the shortened version just listing the files
```
Isgotkowitz: I think it actually doesn't collect the modified committed files

ew3361zh: Good catch, I think I meant staged and unstaged. But yes, once committed, it wouldn't catch that. So maybe not a comprehensive enough scan.

Isgotkowitz: I think so, I think a more comprehensive scan will be durable to different workflows. Like often I will forget to run the local linter before I commit changes, and the more comprehensive scan would make order irrelevant (people could lint anytime - before or after adding, committing or pushing, and the same set of files will always be linted).

ew3361zh: Great call. I will review your updated script today. 

I wonder if you could take a look at the flake8_instructions.md file as well - essentially what I would add to the wiki for how RuFaS is using flake8 and how we're asking people adding to the code to use flake8.

I know there are options like [pre-commit](https://pre-commit.com/) to kind of force linting and not allow merging with linting issues but I think that's too severe given there are some flake8 violations that are signficantly more than a quick fix like deleting a space or breaking up a line for length and I wouldn't want to not allow people to merge their code for a lack of fixing a violation.

So I want the instructions to be accessible so people have a process down and can easily understand how to fix their own flake8 violations as well as hopefully legacy violations.

cleanup.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
Isgotkowitz:
```suggestion
#!/bin/sh
```
On POSIX-compliant systems, `/bin/sh` links to whatever the default shell is on that system, so I think making it `sh` instead of `bash` will keep it working in case someone is working on a non-BASH shell (likely someone on a Mac, whose default shell is ZSH)

cleanup.sh
```diff
@@ -0,0 +1,19 @@
...
+
+display_usage() {
+    echo "Usage: ./runflake8onbranch.sh [BASEBRANCH]"
```
Isgotkowitz:
```suggestion
    echo "Usage: ./cleanup.sh [BASEBRANCH]"
```

flake8_instructions.md
```diff
@@ -0,0 +1,54 @@
...
+
+This guide will help show how to install flake8 locally and to show you how to invoke it and to correct the styleguide violations present **before** pushing code to GitHub.
+
+Whenever code is pushed to the RuFaS GitHub repository, it should first be checked locally using flake8 to ensure no additional styleguide violations are occurring.
+
+Additionally, any fixes to legacy styleguide violations you can do within the files you're working on are greatly appreciated. Often these are quick fixes but may need to be done manually.
+
+## Installing Flake8 Locally
+
+To get started, install flake8 locally by opening an interactive shell or terminal and run this command:
```
Isgotkowitz:
```suggestion
To get started, first make sure that your virtual environment is activated.
Then install flake8 locally by opening an interactive shell or terminal and run this command:
```
Maybe instead of instructions for adding flake8 specifically we just add instructions for how to update the virtual environment?

ew3361zh: I think this is a good idea and definitely one where I want to take into account the different IDEs - I'm pretty sure PyCharm automatically creates a new venv for every project but I'd like to help make sure users are opening a shell to update/install flake8 only for the RuFaS venv.

Isgotkowitz:
The desired outcomes from this PR have been achieved with #806

PR #430, opened 2023-04-07
Implement, integrate, and test manure in soil.phosphorus_cycling module (5/5, Manure decomposition & assimilation)

### LINKED ISSUE #389: refactor and test manure_leach.py (soil/phosphorus_cycling) into SC_redesign directory

Isgotkowitz:
# Summary
This PR implements assimilation and decomposition functionality in the `Manure` module of `phosphorus_cycling/`, and completes the functionality of this sub-module. 

# Main Changes
- `manure.py`
  - `daily_manure_update()` now adjusts the moisture factor of the two manure pools.
  - `daily_manure_update()` now decomposes and assimilates manure and the phosphorus in it (these processes comprise most of the method). First, the amounts of decomposed and assimilated manure and phosphorus are calculated, then they are removed from the pools at the same time (this presents a serious issue, see note 3 in the [Notes](#Notes) section for more information).
  - non-static subroutines `_adjust_manure_moisture_factor()`, `_determine_decomposed_surface_manure()`, and `_determine_assimilated_surface_manure()`  have been added. For both manure pools, the first method calculates and updates the moisture factor. The last two determine the changes in manure mass and field coverage due to decomposition and assimilation, then return the changes in dictionaries.
  - two `@staticmethods` have been added, `_determine_assimilated_phosphorus_amount()` and `_determine_mineralized_surface_phosphorus()`. They calculate how much phosphorus is assimilated and decomposed from the given pool on the current day, respectively.
  - `_determine_mineralization_rate()` has been replaced with `_determine_mineralized_surface_phosphorus()`. The former only calculated the rate that phosphorus mineralized at, the latter does the same but also calculates the amount of phosphorus actually lost from the given pool. The latter is called repeatedly in the daily update routine, so introducing more functionality makes the method overall cleaner and easier to read.
  - all methods of `Manure` have a dedicated test method in `test_manure.py`

# Other Changes
- `manure_applied_mass`
  - The method `_adjust_manure_moisture_factor()` requires knowing the original mass of manure in the application, and adding it required a small refactor of the following below modules.
  - `soil_data.py`: now contains two applied mass attributes, one for each manure pool.
  - `manure_application.py`:
    - The application methods now reset the applied mass attribute when manure is applied to the field.
    - The test methods for said application methods also check that the attribute is set correctly.

# Notes
1. Originally, I planned to have the final PR in this series implement the `PhosphorusCycling` composite class. However, I now think that it would be better to implement it in the PR that finishes the `phosphorus_cycling/` submodule, because it will likely require only one simple and relatively small PR to implement it and integrate with both `Soil` and the classes that comprise it. (At most, it will require two PR's)
2. The current way that the two manure pools are managed is very bad. They have the same set of attributes, and are only differentiated by prefixing `machine_` or `grazing_` in front of the attribute name. This will be addressed by issue #433 
3. As it is currently designed, `Manure` has a big issue in that it potentially creates phosphorus out of thin air, see `manure.py` lines 114 - 118 for more details and a possible solution. Any feedback on the listed solution or suggestions for others would be appreciated. Also, if this a solution is decided upon and it is decided that it should be implemented immediately, I think would be good to split that into it's own PR (instead of incorporating it into this one) 
4. For phosphorus (but not manure!), the terms "decomposing" and "mineralizing" mean the same thing in the context of this module.


morrowcj (approved):
This looks fine to  me. We already discussed the refactor/design changes that would improve this module in the future, but for now it looks good to go.

Be sure to periodically check in on `field-manure-connection.md` in #333 and make sure that it contains an up-to-date list of variables that we'll need from the Manure module (separately from the user's input regarding management practices).

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/manure.py
```diff
@@ -126,6 +255,129 @@ def _add_infiltrated_phosphorus_to_soil(self, infiltrated_phosphorus_amount: flo
...
+            and how much the field coverage changes as a result.
+
+        Parameters
+        ----------
+        temperature_factor : float
+            The temperature factor on the current day (unitless)
+
+        Returns
+        -------
+        Dict
```
matthew7838: I think there are a few ways to document dictionaries better. I googled some and I'm not sure which one works on our current format, you might want to take a look. For me, I think it's more intuitive to indicate that the below arguments are keys.

matthew7838 (approved):
Looks pretty good to me. For note 3, I'm not sure if I can make a good comment on that, but it seems like you and Clay had a discussion already. I only have a few suggestions for docstrings. Great PR.

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/manure.py
Isgotkowitz: Ok, thanks for the suggestion. Do you have any of the links to those sites?

PR #435, opened 2023-04-12
Input system design document

morrowcj:
Creates a design document for a new input system for RuFaS.

A new and improved input system is critically needed to best facilitate sensitivity analyses, but it also has many other benefits for the RuFaS module - as outlined by the markdown document. 

---

**Note** that the draw.io file is the source file used by diagrams.net to produce the class diagram shown in the jpeg file.


morrowcj:
As a note: I've begun work on similarly structured input systems for the Crop and Soil module: #466

ltp23:
Hi everyone, I added a new diagram based on Clay's diagram. It is fairly complicated but I tried to add some flexibility for the reading, validating, and providing default values for input sources. I made an attempt at incorporating some structure to the sensitivity analysis component into the design. 

The diagram is in markdown format, so you can make edits to it too. 
To render it as an SVG image, I installed an npm package called `mermaid` (I think) and the command I used was
`mmdc -i input-system-class-diagram.mmd -o input-system-class-diagram.svg`.
You can view the SVG file in a browser and would need to zoom in quite a bit. I have also included an SVG image in the latest commit.

PooyaHekmati:
Incorporated into Input Manager's design doc, closing the PR.

PR #440, opened 2023-04-13
Refactor, test, and implement `soluble_P.py`

### LINKED ISSUE #439: refactor and test soluble_P.py (soil/phosphorus_cycling) into SC_redesign directory

Isgotkowitz:
# Summary
This PR refactors the file `soluble_P.py` from the old code. The two processes this submodule executes are 1) removing phosphorus from the top layer of soil when there is runoff, and 2) transferring phosphorus from each soil layer to the layer below it as water percolates through the layers.

# Main Changes
- `soluble_phosphorus.py` (replaces `soluble_P.py`)
  - This module contains 1 public-facing routine, `daily_update_routine()`, which conducts all runoff and phosphorus percolation operations.
  - The module contains 1 non-static helper method, `_remove_runoff_phosphorus_from_top_soil()` which handles all phosphorus removal due to runoff.
  - All other methods in the module are `@staticmethods` that calculate various parameters and conduct unit conversion.
  - All methods are tested in `test_soluble_phosphorus.py`. *NOTE*: please pay extra attention to the test method for the main routine (`test_daily_update_routine()`). This was a tricky function to test, because the changes in the labile phosphorus cascade down the soil profile and thus would be difficult to compute. Currently, there is only soil layer simulated, but is tested to account for different cases and is verified by checking the amount of phosphorus in it after percolation.
- `percolation.py`
  - Because the amount of phosphorus that moves from a layer to the one below it depends on how much water percolates from the upper layer to the lower layer, this amount needed to be recorded. So the main routine of `Percolation` now records the amount of water that percolates out of each layer of soil everyday.
  - This new functionality is tested in `test_percolation.py`
- `soil_data.py`
  - The attribute `annual_soil_phosphorus_runoff` has been added that now tracks how much runoff phosphorus there is from the soil surface.
  - This attribute is also reset by the `do_annual_reset()` method.
- `layer_data.py`
  - The attribute `percolated_water` now tracks how much water is percolated out of the layer on the current day. 
- `crop_and_soil_constants.py`
  -  The conversion factors for cubic meters to cubic millimeters and kilograms to megagrams (and the factors for going in the reverse direction) have been added.


SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/soluble_phosphorus.py
```diff
@@ -0,0 +1,227 @@
...
+        soil_data : SoilData, optional
+            The SoilData object used by this module to track manure phosphorus activity, creates new one if one is not
+            provided.
+
+        """
+        self.data = soil_data or SoilData()
+
+    def daily_update_routine(self, runoff: float, field_size: float) -> None:
+        """Removes phosphorus from the top layer of soil due to runoff, and moves phosphorus downward through the soil
+            profile as water percolates through it.
```
morrowcj, line 29: This module doesn't handle percolating of P, right?

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/soluble_phosphorus.py
```diff
@@ -0,0 +1,227 @@
...
+
+        Parameters
+        ----------
+        runoff : float
+            Amount of rainfall that runs off the field on the current day (mm)
+        field_size : float
+            Size of the field (ha)
+
+        """
+        if runoff:
```
morrowcj: Should the condition be > 0?

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/soluble_phosphorus.py
```diff
@@ -0,0 +1,227 @@
...
+        -----
+        TODO: this equation is in the code, both old RuFaS and SurPhos, but is not in the documentation. Also not clear
+            what the units are, amend this notes section after talking with Pete.
+
+        """
+        dissolved_reactive_phosphorus_leachate = exp((soil_phosphorus * 1.5 - isotherm_intercept) / isotherm_slope)
+        return min(20.0, dissolved_reactive_phosphorus_leachate)
+
+    @staticmethod
+    def _determine_percolated_water_volume(percolated_water: float, field_size: float) -> float:
```
morrowcj, line 214: Ignore my previous comment. My bad. Should this be in the percolate module though?

morrowcj (approved):
Looks good to me, with few minor comments.

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/soluble_phosphorus.py
```diff
@@ -0,0 +1,227 @@
...
+        runoff_in_liters_per_hectare = runoff_in_liters / field_size
+
+        top_layer_soil_phosphorus_concentration = self._determine_soil_phosphorus_concentration(
+            self.data.soil_layers[0].labile_phosphorus_content, self.data.soil_layers[0].bulk_density,
+            self.data.soil_layers[0].layer_thickness, field_size)
+        extraction_coefficient = 0.005
+        top_layer_dissolved_reactive_phosphorus_runoff = top_layer_soil_phosphorus_concentration * \
+            extraction_coefficient * runoff_in_liters_per_hectare * (10 ** (-6))
+
+        adjusted_phosphorus_runoff = min(self.data.soil_layers[0].labile_phosphorus_content,
```
matthew7838: Wouldn't it be better if this method just returns the runoff and remove the amount in the main routine?

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/soluble_phosphorus.py
```diff
@@ -0,0 +1,227 @@
...
+            Bulk density of the soil layer (Megagram per cubic meter)
+        layer_thickness : float
+            Thickness of the soil layer (mm)
+        field_size : float
+            Area of the field (ha)
+
+        Returns
+        -------
+        float
+            The concentration of phosphorus in the soil layer (mg phosphorous per kg soil)
```
matthew7838:
```suggestion
            The concentration of phosphorus in the soil layer (mg phosphorous/kg soil)
```
Maybe this would be more in line with how we document units in other files? e.g. (kg/ha)

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/soluble_phosphorus.py
```diff
@@ -0,0 +1,227 @@
...
+            Concentration of phosphorus in the soil layer (mg phosphorous per kg soil)
+        isotherm_slope : float
+            Slope of the phosphorus sorption isotherm (unitless)
+        isotherm_intercept
+             Intercept of the phosphorus sorption isotherm (unitless)
+
+        Returns
+        -------
+        float
+            Concentration of dissolved phosphorus in the soil water that can be leached into the next layer (mg per L)
```
matthew7838: Same suggestion as above.

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/soluble_phosphorus.py
```diff
@@ -0,0 +1,227 @@
...
+
+        References
+        ----------
+        APLE Theoretical Documentation eqn. [14]
+
+        Notes
+        -----
+        TODO: this equation is in the code, both old RuFaS and SurPhos, but is not in the documentation. Also not clear
```
matthew7838, line 206: Add an issue?

matthew7838 (approved):
Looks like a good PR. Again, some suggestions that you can decide to adapt or not. Great work.

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/soluble_phosphorus.py
Isgotkowitz: Same difference. Python treats non-zero numbers as `True`, and numbers set to 0 as `False`. I'm gonna add the explicit conditional though.

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/soluble_phosphorus.py
Isgotkowitz: No worries. I don't think this method should be in `Percolation` (for now) and my reasoning is that this is the only place where the percolated volume of water is needed in the Soil and Crop module.

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/soluble_phosphorus.py
Isgotkowitz: Done

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/soluble_phosphorus.py
Isgotkowitz: Done

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/soluble_phosphorus.py
Isgotkowitz: Good call.

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/soluble_phosphorus.py
Isgotkowitz: Good question, I'm not 100% sure. I like the current design because it keeps the main routine cleaner. However if the current design for `_remove_runoff_phosphorus_from_top_soil()` is kept, do you think it would be good to create a similar function that handles the percolation of phosphorus (i.e. a method that does all the stuff contained in the `for` loop) and call that function every iteration?

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/soluble_phosphorus.py
morrowcj: I figured that was the case, but I think the explicit condition reads clearer. 

Interestingly, it looks like negative numbers are also evaluate to True: 

> Integers and floating point numbers can be converted to the boolean data type using Python's bool() function. An int, float or complex number set to zero returns False. An integer, float or complex number set to any other number, positive *or negative*, returns True .

SC_redesign/Crop_and_Soil/soil/phosphorus_cycling/soluble_phosphorus.py
matthew7838: Yes. I do think that if the ultimate goal is to still keep a cleaner main routine, then it would be good to create a similar function.

PR #446, opened 2023-04-14
Refactor Residue Partition (4/4)!

matthew7838:
### Summary
This PR is the last part of refactoring `residue_partition.py` of the Soil module from "pseudocode_soil" S.6.B.II. As for this part, the PR cover all the main routine. Every method is tested covering most edge cases.

### Changes
- Now the main routines mostly only update the attributes
- Deleted the else condition of incorporate biomass because it will already be updated correctly in the corp module

### Notes
There's a lot of modification compared to the original code in this PR, it's better to have a second opinion that compares with the original code to see if there's anything that doesn't make sense.


SC_redesign/Crop_and_Soil/soil/carbon_cycling/residue_partition.py
```diff
@@ -17,6 +18,124 @@ class ResiduePartition:
     def __init__(self, soil_data: Optional[SoilData] = None):
         self.data = soil_data or SoilData()  # initialize with defaults, if not
 
+    def partition_residue(self, rainfall: float, crop: CropData):
+        self.data.plant_residue_lignin_composition = self._determine_plant_residue_lignin_composition(
```
Isgotkowitz, line 33: Docstring

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -177,6 +177,16 @@ class SoilData:
...
 
+    # ---- Residue partition (Carbon Cycling)
+    plant_residue_lignin_composition: float = 0
+    """lignin fraction of plant residue (unitless)"""
+    plant_lignin_nitrogen_ratio: float = 0
+    """plant lignin to nitrogen ratio (unitless)"""
+    plant_residue_metabolic_fraction: float = 0
+    """plant residue fraction that is metabolic (unitless)"""
+    total_residue = 0
+    """"amount of total residue (kg/ha)"""
```
Isgotkowitz: Is this residue on the soil surface or total residue in the soil profile?

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_residue_partition.py
```diff
@@ -196,23 +199,23 @@ def test_determine_plant_structural_carbon_amount(plant_dry_matter_residue_amoun
...
+#     (0, 0.4),  # no dry matter residue amount
+#     (2, 0),  # no tillage fraction
+#     (0, 0),  # neither
+# ])
+# def test_determine_soil_dry_matter_residue_amount(plant_dry_matter_residue_amount: float,
+#                                                   tillage_fraction: float) -> None:
+#     """Tests that the soil dry matter residue amount was updated correctly"""
+#     expected = plant_dry_matter_residue_amount * tillage_fraction
+#     assert expected == ResiduePartition._determine_soil_dry_matter_residue_amount(plant_dry_matter_residue_amount,
+#                                                                                   tillage_fraction)
```
Isgotkowitz:
```suggestion
```

Isgotkowitz (approved):
Nice job. I only found a missing docstring and a commented out test that can be deleted.

RUFAS/routines/field/soil/carbon_cycling/residue_partitioning.py
morrowcj, line 1: let's not make any changes to the original code. If you could revert this, that'd be great. If you need help doing so, I think the following git command should work (checkout a single file):

```shell
git checkout origin/SC_redesign -- RUFAS/routines/field/soil/carbon_cycling/residue_partitioning.py
```

SC_redesign/Crop_and_Soil/soil/carbon_cycling/residue_partition.py
morrowcj: In the docstring, make sure to note something like "this method should *only* be called (by the field/field manager) on the day that a cut, harvest, or kill operation occurs and should be called *after* that operation."

SC_redesign/Crop_and_Soil/soil/carbon_cycling/residue_partition.py
```diff
@@ -17,6 +18,124 @@ class ResiduePartition:
...
+                layer.decomposition_moisture_effect,
+                self.data.decomposition_temperature_effect,
+                layer.plant_metabolic_carbon_amount)
+
+            layer.plant_metabolic_to_soil_carbon_amount = self._determine_plant_metabolic_to_soil_carbon_amount(
+                layer.plant_metabolic_carbon_amount,
+                layer.tillage_fraction
+            )
+
+            if self.data.soil_layers.index(layer) == 0:
```
morrowcj, line 51: > `self.data.soil_layers.index(layer)`

I'm not familiar with this syntax: I think you're checking if this is the top layer, but can you explain to me how it does that?

morrowcj: @Isgotkowitz, If I'm correct about this being for the "top layer", it may be affected by the 'top 20cm' layer. Would you look over this chunk again with that in mind?

SC_redesign/Crop_and_Soil/soil/carbon_cycling/residue_partition.py
```diff
@@ -17,6 +18,124 @@ class ResiduePartition:
...
+                self.data.decomposition_temperature_effect,
+                layer.soil_metabolic_carbon_amount,
+            )
+
+            layer.soil_metabolic_carbon_amount = self._determine_soil_metabolic_carbon_amount(
+                layer.soil_metabolic_carbon_amount,
+                layer.plant_metabolic_to_soil_carbon_amount,
+                crop.root_biomass,
+                layer.soil_residue_metabolic_fraction,
+                layer.soil_metabolic_active_carbon_usage
```
morrowcj, line 129: Pycharm is telling me that this argument has the "incorrect type". It should be a `float`, but I'm wondering if it the error is because the `SoilData` class is initialized with this attribute set to `None` by default. 

The attribute gets set, as a `float` above this on line 104, so I'm not sure if it is a real problem.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/residue_partition.py
```diff
@@ -17,6 +18,124 @@ class ResiduePartition:
...
+                layer.structural_carbon_transfer_amount,
+                layer.soil_structural_active_carbon_usage,
+                layer.soil_structural_slow_carbon_usage,
+                crop.root_biomass,
+                layer.soil_structural_carbon_amount
+            )
+
     @staticmethod
     def _determine_plant_residue_lignin_composition(plant_residue_lignin_composition: float,
                                                     rainfall: float) -> None:
```
morrowcj: This method should return a `float`, right?

```suggestion
    def _determine_plant_residue_lignin_composition(plant_residue_lignin_composition: float,
                                                    rainfall: float) -> float:
```

morrowcj (approved):
This looks pretty good. I left a few comments. This module is a mess and you did your best with what you were given, so you should be proud of that. 

Hopefully we'll be able to improve it in the future. 

I also pushed an empty `__init__.py` file to make `carbon_cycling/` a proper module.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/residue_partition.py
morrowcj, line 51:
```suggestion
            if self.data.soil_layers.index(layer) == 0:
            # TODO: rather than simply mixing the crop residue into the top layer, this module should consider tillage
            #     depth and mix the residue into all soil down to that depth. This is not a priority, though.
```

matthew7838: Not sure how this goes with the 'top 20cm' layer, but yes it is taking the top layer following the original code.

Isgotkowitz: @morrowcj yes, it definitely would be affected by that separate top 20 mm of soil. If I understand what this control block is doing correctly, it is transferring carbon from the soil surface to the top layer of soil, the exact amount being determined by how much carbon is on the surface and how much that carbon is affected by tilling. Which makes this code very unrealistic, because it seems kinda impossible to only plow the top 20mm of soil.

That said, implementing a solution that distributes carbon proportionally to every layer affected by tillage sounds like a lot right now, so maybe it would be better to wait till after V1?

Isgotkowitz: @morrowcj also `my_list.index(x)` returns the index of `x` in `my_list` (if it's there, which it always should be in this case. Raises a ValueError if it doesn't find it)

Isgotkowitz: @matthew7838 The top 20mm (not cm) comes from SurPhos. Phosphorus gets really concentrated in the top 20mm of the soil profile, so it is important to simulate it separately to get accurate results. The problem is that the actual top soil layer might be a lot larger than 20 mm, so we override the user input soil profile in `SoilData` and always keep a 20mm-thick soil layer at the top of the profile. Sometimes this conflicts with what other soil modules do, so it's important to consider whether the top two layers in our code should be treated as one for the purposes of a specific module.

morrowcj: > That said, implementing a solution that distributes carbon proportionally to every layer affected by tillage sounds like a lot right now, so maybe it would be better to wait till after V1?

I agree.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/residue_partition.py
matthew7838: I'm not quite sure why there's a warning tho, but I've tested with a few scenarios and don't think it actually causes problems.

SC_redesign/Crop_and_Soil/soil/soil_data.py
matthew7838: I think it's total residue in the soil profile, I can't recall clearly and it's not referenced. @morrowcj I think we might have talked about this attribute do you have a better insight on this?

morrowcj: I think this is simply for mass balance purposes - a tally of all the residue ever added to the field. I don't think it gets used anywhere though.

We'll probably need to rework this a bit so that multiple crops can add their residue to the field instead of just one, though.

PR #252, opened 2022-12-23
Crop transpiration

morrowcj:
This sub-PR of #223 refactors water uptake by crops. 

I was originally going to include tests separately, but I decided to make them small, since they are identical to N and P tests and need to be combined in the future.

PR #451, opened 2023-04-19
Tests of the crop transpiration refactor

morrowcj:
This is a stacked PR on top of #252, which implements unit tests.


morrowcj:
These tests are so small that I'm just going to add them to the main PR.

PR #452, opened 2023-04-19
Pen Feed Data Added to OM

ew3361zh:
This PR adds some requested data on rations and feed by individual pen to the Output Manager.

It adds: 
- the daily ration data (animalmanagement.daily_update_id_pen.pen_ration_data_pen_num) by pen

- the ration formula at each ration formulation (animalmanagement.daily_update_id_pen.pen_specific_feed_data_pen_num) by pen

- several pieces of ration nutrition data at the point of ration formulation:
1. ration_nutrient_amount
2. ration_nutrient_conc
3. MEdiet (metabolizable energy concentration of the diet fed to the pen)
4. dry_matter_intake

To make this work, I had to slightly tweak the logic in a part of the daily_update_id_pen function to ensure pens with a ration set are being updated with ration amounts daily.

### DIFF
--- a/RUFAS/routines/animal/animal_management.py
+++ b/RUFAS/routines/animal/animal_management.py
@@ -23,7 +23,6 @@ from typing import Any, Dict, Tuple, List
 from RUFAS.general_constants import GeneralConstants
 from RUFAS.output_manager import OutputManager
 from RUFAS.routines.animal.animal_module_constants import AnimalModuleConstants
-from RUFAS.routines.animal.clustering_pen_grouping import grouping
 from RUFAS.routines.animal.life_cycle.animal_base import AnimalBase
 from RUFAS.routines.animal.life_cycle.cow import Cow
 from RUFAS.routines.animal.life_cycle.life_cycle import LifeCycleManager
@@ -384,12 +383,12 @@ class AnimalManagement:
             feed: an instance of the Feed class defined in feed.py
             temp: the temperature on the current day
         """
-
+        calf_id_list = [calf.id for calf in calves_born]
         info_map = {"class": self.__class__.__name__,
                     "function": self.daily_update_id_pen.__name__,
                     "animals_added": animals_added,
                     "ids_removed": ids_removed,
-                    "calves_born": calves_born,
+                    "calves_born_id_list": calf_id_list,
                     "temp": temp,
                     }
 
@@ -468,7 +467,7 @@ class AnimalManagement:
 
         for i in range(len(self.all_pens)):
             if self.all_pens[i].ration == {}:
-                if len(self.all_pens[i].animals_in_pen) > 0:            
+                if len(self.all_pens[i].animals_in_pen) > 0:
                     available_feeds = ration_driver.AvailableFeeds()
                     available_feeds.feed_nutrients(feed)
                     self.all_pens[i].allocated_feeds = feed.input_feed_combinations[self.all_pens[i].animal_combination]
--- a/RUFAS/routines/animal/animal_management.py
+++ b/RUFAS/routines/animal/animal_management.py
@@ -373,6 +373,14 @@ class AnimalManagement:
             temp: the temperature on the current day
         """
 
+        info_map = {"class": self.__class__.__name__,
+                    "function": self.daily_update_id_pen.__name__,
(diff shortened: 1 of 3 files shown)


RUFAS/routines/animal/animal_management.py
```diff
@@ -385,6 +385,14 @@ def daily_update_id_pen(self, animals_added, ids_removed, calves_born, feed, tem
             temp: the temperature on the current day
         """
 
+        info_map = {"class": self.__class__.__name__,
+                    "function": self.daily_update_id_pen.__name__,
+                    "animals_added": animals_added,
+                    "ids_removed": ids_removed,
+                    "calves_born": calves_born,
```
JoeWaddell: If these are calf objects, I think it would be better to add just their IDs, to avoid the problems we discussed about `max_depth` and blowing up output file size.

Something like `[calf.id for calf in calves_born]` would probably do the trick

RUFAS/routines/animal/pen.py
```diff
@@ -438,6 +438,15 @@ def calc_ration(self, feed, available_feeds: DefaultDict[Any, Any]):
         self.MEdiet = ration_vals['ME_tot']
         self.dry_matter_intake = nutrient_amount['dm']
 
+        info_map = {"class": self.__class__.__name__,
+                    "function": self.calc_ration.__name__,
+                    "feed": vars(feed),
```
JoeWaddell: Here I'd get rid of `vars(feed)`, since the subsequent `available_feeds` is the relevant subset of those data

RUFAS/routines/animal/pen.py
```diff
@@ -438,6 +438,15 @@ def calc_ration(self, feed, available_feeds: DefaultDict[Any, Any]):
         self.MEdiet = ration_vals['ME_tot']
         self.dry_matter_intake = nutrient_amount['dm']
 
+        info_map = {"class": self.__class__.__name__,
```
JoeWaddell, line 435: I think we'd also want the pen ID, perhaps the pen's animal combination, and the simulation day in the info map here, so that we can later reconstruct the nutrient amount/conc over time, by pen.

JoeWaddell (changes requested):
A few initial comments on tweaks to the info maps. 

Also as we chatted about, this would be a good place to add the fix to the other case of the info_map increasing output file size to 20+GB

RUFAS/routines/animal/animal_management.py
ew3361zh: sounds good. This is making me think a good review of all the objects being added to info_maps is probably in order.

RUFAS/routines/animal/pen.py
ew3361zh: I'm going to get rid of it in the second info_map as well (roughly around line 525) - let me know if you think it needs to be there and I can add it back in.

RUFAS/routines/animal/pen.py
ew3361zh: pen ID and animal combination are both easy enough as they're elements of the pen class instance but simulation day isn't tracked in the pen class so we'd have to add it specially or perhaps there's another reference to it we can use - I will think about it.

JoeWaddell: OK yeah that's something that will be necessary in many cases, I think. `sim_day` might be reconstructable, so to speak, just by applying some assumption about the number of pens and total simulation days, but that will be messier if pens are added during the simulation. 

Just thinking that when we're analyzing nutrient concentration in a ration, we'd want to see how it changed over the course of the simulation. So we'll just need some way to calibrate these variable measures in time, essentially.

JoeWaddell (approved):
Nice work on the changes! I think this looks good, and if we need to make additions to the info_maps or tweaks to the variable name, we know where to look. 

The only other additions would be: 1) the retrieval of sim day, possibly from `PenHistory` in an `Animal` object inside the `Pen` object, and 2) the fix for the output manager file size we talked about on Slack.

tomhuhh (approved):
Looks good to me! Joe has spoke to me about the changes in the info maps, which I think makes sense as long as information needs to be tracked is being tracked. Thank you Niko!

PR #456, opened 2023-04-20
Refactor carbon_cycle.py

matthew7838:
### Summary
This PR is the last part of the carbon cycle module, it also aggregates some value calculated throughout other files as well as runs all the main routines. The aggregation was done by following the pseudocode_soil S.6.D.1 to S.6.D7.

### Changes
- All other methods to calculate aggregation are now static
- There is now a separate method aside from the main routine just to update the aggregated attributes


SC_redesign/Crop_and_Soil/soil/carbon_cycling/carbon_cycle.py
```diff
@@ -0,0 +1,347 @@
...
+            Size of the field (ha)
+
+        Returns
+        -------
+        float
+            soil volume (cubic meters)
+
+        References
+        -------
+        pseudoode_soil S.6.D.1
```
Isgotkowitz:
```suggestion
```
Generally unit conversion doesn't need references. I think it would only be appropriate in cases where really uncommon/novel/proprietary units are used, @morrowcj what do you think?

SC_redesign/Crop_and_Soil/soil/carbon_cycling/carbon_cycle.py
```diff
@@ -0,0 +1,347 @@
...
+            soil volume (cubic meters)
+
+        Returns
+        -------
+        float
+            mass of soil (kg)
+
+        References
+        -------
+        pseudoode_soil S.6.D.1
```
Isgotkowitz:
```suggestion
```

SC_redesign/Crop_and_Soil/soil/carbon_cycling/carbon_cycle.py
```diff
@@ -0,0 +1,347 @@
...
+        Returns
+        -------
+        float
+            fraction of active carbon in the soil (unitless)
+
+        References
+        -------
+        pseudoode_soil S.6.D.2
+        """
+        return active_carbon_amount/soil_mass
```
Isgotkowitz: Active carbon needs to be multiplied by field size before being divided by the soil mass here (I think) because if that doesn't happen the units will be mismatched (kilograms per hectare divided by kilograms results in kilograms per hectare-kilogram)

SC_redesign/Crop_and_Soil/soil/carbon_cycling/carbon_cycle.py
```diff
@@ -0,0 +1,347 @@
...
+        Returns
+        -------
+        float
+            soil volume (cubic meters)
+
+        References
+        -------
+        pseudoode_soil S.6.D.1
+        """
+        return (layer_thickness * field_size * HECTARES_TO_SQUARE_MILLIMETERS) * CUBIC_MILLIMETERS_TO_CUBIC_METERS
```
morrowcj: Great job. This looks good.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/carbon_cycle.py
```diff
@@ -0,0 +1,347 @@
...
+        Returns
+        -------
+        float
+            fraction of slow carbon in the soil (unitless)
+
+        References
+        -------
+        pseudoode_soil S.6.D.2
+        """
+        return slow_carbon_amount/soil_mass
```
Isgotkowitz: Same here as above method

SC_redesign/Crop_and_Soil/soil/carbon_cycling/carbon_cycle.py
```diff
@@ -0,0 +1,347 @@
...
+    @staticmethod
+    def _determine_soil_mass(bulk_density: float, soil_volume: float) -> None:
+        """This method calculates the mass of soil
+
+        Parameters
+        ----------
+        bulk_density: float
+            bulk density of the soil layer (Mg per cubic meter)
+        soil_volume: float
+            soil volume (cubic meters)
```
morrowcj: Spell out the units here to avoid confusion (double-check that this is correct).

```suggestion
        bulk_density: float
            bulk density of the soil layer (Megagrams per cubic meter)
        soil_volume: float
            soil volume (cubic meters)
```

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -198,6 +198,16 @@ class LayerData:
     soil_structural_carbon_amount: Optional[float] = None
     """amount of soil structural carbon decomposed into slow or active carbon (kg/ha)"""
 
+    # --- Carbon cycling
+    soil_overall_carbon_fraction: Optional[float] = None
+    """the total fraction of carbon in the soil (unitless)"""
```
Isgotkowitz, line 234:
```suggestion
    soil_overall_carbon_fraction: Optional[float] = None
    """the total fraction of carbon in the soil by mass (unitless)"""
```

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -198,6 +198,16 @@ class LayerData:
...
+    total_soil_carbon_amount: Optional[float] = None
+    """the total amount of soil carbon (kg/ha)"""
+    total_decomposition_carbon_CO2_lost: Optional[float] = None
+    """amount of total carbon lost as CO2 during decomposition(kg/ha)"""
+    total_carbon_CO2_lost: Optional[float] = None
+    """total amount of carbon lost as CO2 (kg/ha)"""
```
Isgotkowitz: I think this should be an annual total, and then be reset once a year. An example of this process is `soil_data.py`, variables in lines 33 to 66 and the method `do_annual_reset()` at line 294. In fact if there winds up being an annual reset method in `LayerData`, a good place for it to be called on every layer would be `do_annual_reset()`

SC_redesign/Crop_and_Soil/soil/carbon_cycling/carbon_cycle.py
```diff
@@ -0,0 +1,347 @@
...
+
+        Returns
+        -------
+        float
+            mass of soil (kg)
```
morrowcj: This can't be correct if the units given for inputs are correct: For example, if our bulk density is 5 $Mg / m^3$ and we have a soil volume of 10 $m^3$ then we have a total of 50 $Mg$ of soil:

$$ \frac{5~Mg}{m^3} \times \frac{10~m^3}{1} = 50~Mg$$ 

Note: the $m^3$ gets canceled out by the denominator of the first term and the numerator of the second term.

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_carbon_cycle.py
```diff
@@ -0,0 +1,234 @@
...
+    cycle = CarbonCycle(data)
+    cycle.decomposition.decompose = MagicMock()
+    cycle.pool_gas_partition.partition_pool_gas = MagicMock()
+    cycle.residue_partition.partition_residue = MagicMock()
+    cycle._soil_carbon_aggregation = MagicMock()
+    cycle.cycle_carbon(rainfall, crop, temp_average, field_size)
+
+    assert cycle.decomposition.decompose.call_count == 1
+    assert cycle.pool_gas_partition.partition_pool_gas.call_count == 1
+    assert cycle.residue_partition.partition_residue.call_count == 1
```
Isgotkowitz, line 238:
```suggestion
    assert cycle.residue_partition.partition_residue.call_count == 1
    assert cylce._soil_carbon_aggregation.call_count == 1
```

SC_redesign/Crop_and_Soil/soil/carbon_cycling/carbon_cycle.py
morrowcj: Something is wrong here as well. If we're trying to get the fraction of carbon in the soil, we need to divide the mass of carbon by the mass of soil. This equation has the *density* of carbon (kg/ha) divided by the mass of soil. 

On a more important level: are the any of these computed fractions (slow, active, passive, overall) actually used anywhere? If not, I don't think it is important to include them.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/carbon_cycle.py
```diff
@@ -0,0 +1,347 @@
...
+        Returns
+        -------
+        float
+            the total amount of soil carbon (kg/ha)
+
+        References
+        -------
+        pseudoode_soil S.6.D.4
+        """
+        return active_carbon_amount + slow_carbon_amount + passive_carbon_amount
```
morrowcj, line 249: This is important to track, and it looks right to me.

morrowcj (changes requested):
The calculation of carbon mass needs some clarification. My other comments refer to the different carbon fractions and whether or not we actually need them.

Isgotkowitz (changes requested):
Great job on carbon cycling! Some of the calculations need to have their units revisited, and I think some of the carbon attributes could be further integrated with `LayerData`/`SoilData`, but these will hopefully all be easy fixes

SC_redesign/Crop_and_Soil/soil/carbon_cycling/carbon_cycle.py
morrowcj: Yeah, I agree.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/carbon_cycle.py
morrowcj: Yeah, this is exactly what I was referring to in my comment here as well.

SC_redesign/Crop_and_Soil/soil/layer_data.py
morrowcj: I'm indifferent to this. I think either way is fine, since this variable is purely for mass-balance checking (I'm pretty sure this, like many variables only exist so that they can be written to an output file). Whatever we decide to go with, let's change the name to clarify: something like `annual_accumulated_CO2_lost` for the annual reset case or `simulation_accumulated_CO2_lost` if it is a running tally.

Isgotkowitz: Wouldn't CO2 lost be one of the numbers that the economics and emissions module would be really interested in? That was what I was thinking when I suggested tracking the annual amount lost, but if it's gonna be an insignificant amount than we can ignore it.

morrowcj (approved):
LGTM! Nice work!

SC_redesign/Crop_and_Soil/soil/carbon_cycling/carbon_cycle.py
```diff
@@ -0,0 +1,353 @@
...
+
+        """
+        for layer in self.data.soil_layers:
+            soil_volume = self._determine_soil_volume(layer.layer_thickness, field_size)
+            soil_mass = self._determine_soil_mass(layer.bulk_density, soil_volume)
+            soil_active_carbon_fraction = self._determine_soil_active_carbon_fraction(layer.active_carbon_amount,
+                                                                                      soil_mass)
+            soil_slow_carbon_fraction = self._determine_soil_slow_carbon_fraction(layer.slow_carbon_amount,
+                                                                                  soil_mass)
+            soil_passive_carbon_fraction = self._determine_soil_passive_carbon_fraction(layer.passive_carbon_amount,
```
Isgotkowitz: `soil_mass` here is in megagrams, but the three functions beneath it take the soil mass in kilograms.
I suggest adding a conversion factor in `_determine_soil_mass()` so that it returns kg instead of megagrams.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/carbon_cycle.py
```diff
@@ -0,0 +1,353 @@
...
+        Returns
+        -------
+        float
+            fraction of passive carbon in the soil (unitless)
+
+        References
+        -------
+        pseudoode_soil S.6.D.2
+        """
+        return passive_carbon_amount*field_size/soil_mass
```
Isgotkowitz, line 197: `_determine_soil_active_carbon_fraction()`, `_determine_soil_slow_carbon_fraction()`, and `_determine_soil_passive_carbon_fraction()` do the same thing so they could be replaced by a single function, up to you though

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -502,3 +512,7 @@ def water_factor(self):
     def silt_clay_content(self):
         """silt and clay fraction in the soil (unitless)"""
         return self.percent_silt_content / self.percent_clay_content
+
+    def do_annual_rest(self):
```
Isgotkowitz:
```suggestion
    def do_annual_reset(self):
```

SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -322,6 +322,10 @@ def do_annual_reset(self) -> None:
 
         self.annual_soil_phosphorus_runoff = 0
 
+        # Reset carbon cycle
+        for layer in self.soil_layers:
+            layer.do_annual_rest()
```
Isgotkowitz:
```suggestion
            layer.do_annual_reset()
```

Isgotkowitz (changes requested):
Good changes. There is still one units error that can be quickly remedied, a refactor that could be done to turn 3 methods into one, and a typo in the annual reset method name.

SC_redesign/Crop_and_Soil/soil/carbon_cycling/carbon_cycle.py
matthew7838: I agree that it's redundant, but I think that it does improve clarity this way. I think a more generalized function does bring a bit more ambiguity when others are reading this code.

Isgotkowitz (approved):
Nice

PR #457, opened 2023-04-24
Restructured Methane Model

tomhuhh:
The code for methane emissions for all animal categories has been restructured so we can output methane estimates from different methods at the same time.

### DIFF
--- a/RUFAS/routines/animal/life_cycle/cow.py
+++ b/RUFAS/routines/animal/life_cycle/cow.py
@@ -331,22 +331,22 @@ class Cow(HeiferIII):
             methane_model: methane model used for methane emission calculations
             ME_intake: metabolizable energy intake, Mcal/kg DM
         """
-        info_map = {"class": self._class_._name_,
-                "function": self.calc_manure_excretion._name_}
-
         p_urine, p_feces_excrt = self.calc_base_manure()
 
         if self.milking:
-            self.p_excrt, self.manure_excretion, self.methane_emission_dict = lactating_manure_calculations(
+            info_map = {"class": self.__class__.__name__,
+                "function": self.calc_manure_excretion.__name__,}
+
+            self.p_excrt, self.manure_excretion, methane_emission_dict = lactating_manure_calculations(
                 self.ration_formulation, feed, self.body_weight,
                 self.days_in_milk, self.mPrt, self.estimated_daily_milk_produced,
                 p_feces_excrt, p_urine, methane_model, self.fat_percent, ME_intake)
+            
+            om.add_variable("methane emissions", methane_emission_dict, info_map)
         else:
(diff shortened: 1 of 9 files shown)


Ed731551 (commented):
@tomhuhh for dry the matter intake (DMI) values within methane equations, I couldn't see the connection with feeding system (NRC vs NASEM). It is expected that methane predictions may differ because of different DMI predictions. Likely I have missed something that I can't find yet?... or are you considering the NRC predictions for DMI as default? Thanks.

tomhuhh:
> @tomhuhh for dry the matter intake (DMI) values within methane equations, I couldn't see the connection with feeding system (NRC vs NASEM). It is expected that methane predictions may differ because of different DMI predictions. Likely I have missed something that I can't find yet?... or are you considering the NRC predictions for DMI as default? Thanks.

Hello Edward,

Thanks for getting back to me so quickly. My understanding is that the user will need to decide which ration formulation method they want to use in the input files, specifically in the "animal_management.animal.json" file. Currently, we do not output methane emissions for both methods at the same time. 

Thanks!
Haowen

Ed731551:
> > @tomhuhh for dry the matter intake (DMI) values within methane equations, I couldn't see the connection with feeding system (NRC vs NASEM). It is expected that methane predictions may differ because of different DMI predictions. Likely I have missed something that I can't find yet?... or are you considering the NRC predictions for DMI as default? Thanks.
> 
> Hello Edward,
> 
> Thanks for getting back to me so quickly. My understanding is that the user will need to decide which ration formulation method they want to use in the input files, specifically in the "animal_management.animal.json" file. Currently, we do not output methane emissions for both methods at the same time.
> 
> Thanks! Haowen

Ed731551:
Thanks for clarifying this Haowen. My question arose because I have seen in the code that nutritional requirements functions are very specific on this regard. For example: def calculate_NASEM_protein_requirements( )   I though that for DMI it is a similar thing.

KFosterReed:
Hi Tom, 
I think this would go in the category of an acceptable 'Quick Fix' to add a desired functionality for testing before we do a larger refactor on the manure excretion methods. However, it doesn't work because the dictionary can't be summed by the pen calc_manure function so we need another solution if we want to move forward with this: 

image

tomhuhh:
Hello Kristan @KFosterReed! Thanks for letting me know! I will check with Loi and Niko to see how to proceed. 

Thanks!
Haowen

tomhuhh:
An issue (#656) has been created to address the difficulty with multiple methane outputs.

tomhuhh:
Hello @ltp23! I would like to revisit this pull request and draw your attention to it. Have we reached an agreement that this pull request is oriented more towards version 2 and thus falls into a lower-priority category? Thank you.

PR #459, opened 2023-04-24
OM Variable Tracker

ew3361zh:
This PR adds 2 methods to the Output Manager that look through the variables_pool after the simulation runs and pull out the names of all the variables tracked during a simulation and add them, alphabetized, to a .txt file for easy viewing.

This feature was requested as a way to continually be able to easily track what variables are collected by the Output Manager.

To test, run a simulation (try with and without info_maps to ensure both options work) and look at the newly-generated variable names tracking txt file:
Screenshot 2023-04-25 at 8 13 36 AM

It should generate a list of all the variables added to the OM variables pool including the class and function from which they were added:
Screenshot 2023-04-25 at 8 14 11 AM

Unit testing was added and was running successfully for me.

### DIFF
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -324,15 +324,20 @@ class OutputManager(object):
         file_path = os.path.join(path, self._generate_file_name("variable_names", "txt"))
         var_dict = {}
         with open(file_path, 'w') as var_names_file:
-            for k1, v1 in vars_pool.items():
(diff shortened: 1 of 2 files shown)


KFosterReed (approved):
This is a great addition! It will do a lot to facilitate communication with stakeholders and future users.

ltp23 (approved):
It looks good to me. Thanks, Niko!

PR #461, opened 2023-04-24
Heat scheduling

### LINKED ISSUE #368: Implement heat scheduling
Currently the crop module does not have heat scheduling supported (SWAT 5:1.1.1). 

There should be a management option to harvest a crop when it has achieved its optimum growth (as determined via heat units). This practice is called heat scheduling.

morrowcj:
Implements SWAT's heat unit scheduling for crop harvesting.

closes #368


SC_redesign/Crop_and_Soil/crop/crop_data.py
```diff
@@ -51,6 +53,24 @@ class CropData:
...
+    planting_year: int = 0
+    """the year that this crop was planted"""
+    planting_day: int = 100
+    """the (julian) day that this crop was planted"""
+    next_harvest_year: int = 0
+    """the year in which this crop should be harvested next"""
+    next_harvest_day: int = 250
+    """the (julian) day on which this crop should be harvested next"""
+    use_heat_scheduling: bool = False
+    """should heat unit scheduling be used to determine harvesting instead of user-specified harvest dates?"""
```
matthew7838: Not sure about the formality of the documentation. I think that putting a question as attribute documentation is kinda odd. I'm not that fluent in English but perhaps something like this?
```suggestion
    """Status indicating if heat unit scheduling will be used to determine harvesting instead of user-specified harvest dates"""
```

SC_redesign/Crop_and_Soil/crop/crop_data.py
```diff
@@ -51,6 +53,24 @@ class CropData:
...
+    harvest_heat_fraction: float = 1.10
+    """the fraction of potential heat units that denotes the optimal growth stage for harvest"""
+    is_harvest_day: bool = False
+    """is today the day on which harvest operations should occur for this plant?"""
```
matthew7838: Same as above.

matthew7838 (approved):
LGTM, nice to review such a short PR. Seems like just some simple tweak and I don't see many problems. Just two small suggestions.

SC_redesign/Crop_and_Soil/crop/crop_data.py
```diff
@@ -51,6 +53,24 @@ class CropData:
...
 
+    # Management variables
+    planting_year: int = 0
+    """the year that this crop was planted"""
+    planting_day: int = 100
+    """the (julian) day that this crop was planted"""
+    next_harvest_year: int = 0
+    """the year in which this crop should be harvested next"""
+    next_harvest_day: int = 250
+    """the (julian) day on which this crop should be harvested next"""
```
Isgotkowitz, line 64: Do these values start at 1 or 0? It may be helpful indicate that in the docstring, in case someone where to use this module in their own simulation and used a way of tracking time different than the one we use.

SC_redesign/Crop_and_Soil/crop/crop_data.py
Isgotkowitz: I like this pattern for boolean attributes.

Isgotkowitz (approved):
Nice PR. My only comment is about clarifying how the harvest operation works with dates.

SC_redesign/Crop_and_Soil/crop/crop_data.py
morrowcj: That is a really good question. I think from a farmer's perspective they need to start at 1. I think this is something that should be addressed in either the user-input step or in the `Time` class or the overall `SimulationEngine`. For this class, it actually doesn't matter since it is just checking equality.

SC_redesign/Crop_and_Soil/crop/crop_data.py
morrowcj: @Isgotkowitz, to clarify, do you like the original wording or Matthew's alterations?

Isgotkowitz: Matthew's alterations

SC_redesign/Crop_and_Soil/crop/crop_data.py
Isgotkowitz: Gotcha

SC_redesign/Crop_and_Soil/crop/crop_data.py
morrowcj: OK good. I agree, just checking. Thanks guys.

SC_redesign/Crop_and_Soil/crop/crop_data.py
morrowcj:
```suggestion
    """Status indicating if today is the day on which harvest operations should occur for this plant"""
```

Issue #460, opened 2023-04-24
Milk production reduction at ration_driver step is "overwritten" at milking_update step


JoeWaddell:
Milk production should be reduced if a ration cannot be formulated such that it meets animal requirements in a pen. While milk production is reduced at the `ration_formulation` step, it is quickly "overwritten" by a new estimate that doesn't take into account the reduction.

## Current Behavior
This graph is from the user-defined ration branch, so note that it is not representative of a truly realistic simulation (but is informative).
Note the sharp drops in lactation (red line) every ration interval. Those are reductions because the supplied ration cannot meet the animal's requirements. 

However, note that in the supplied graph it immediately "recovers". 

![image](https://user-images.githubusercontent.com/109082924/234092349-948a0322-c8b5-43e2-ac21-ef37de2065db.png)

## Expected Behavior
The milk production line should not rebound immediately to the predicted values, but should be penalized based on the reduction at the ration formulation step.

## Possible Solution
What I propose to implement is to add a `milk_production_reduction` attribute to an animal, with an initialized value of 0.
At each ration formulation that is reset to 0.
If a ration cannot be formulated given the pen's average requirements: milk production is reduced, and that reduction value is set as the `milk_production_reduction` attribute.
That attribute value is then subtracted from the `estimated_daily_milk_produced` value. 
The estimate of daily milk attribute is calculated and set as an attribute in `cow.py`, in the `milking_update` method (I'm also checking that it doesn't occur elsewhere).

### More notes:
The production reduction happens in the `ration_formulation` method in `ration_driver.py`, in the following section. 

```
if animal_type == 'cow':
        while not solution.success:
        # This values for reduction are not from pseudocode, but the vales below
        # are based on fastest case runtime testing
        # TODO: continue testing for more efficient reductions
        NEl_con = NLP.NEl_constraint(solution.x)
        if NEl_con < -0.5:
            reduction = 3 * (-NEl_con)
        else:
            reduction = 1.5
        
        for animal in pen.animals_in_pen:
            animal.estimated_daily_milk_produced -= reduction
        # recalculating requirements after reduction
        req.set_requirements(pen, animal_type, True)
        solution, ration_vals = optimization(pen, req, available_feeds, animal_type, cow_type)

```
Outside the while loop I propose to reset the `milk_production_reduction` attribute to 0. It may be better to have this reset be a separate method entirely outside of `ration_driver.py`
[...]

Issue #363, opened 2023-02-21
X-axis for ration_report graphics reporting incorrect values


JoeWaddell:
Ration report graphics inside `pen_report` are using x-axis labels for years outside the simulation time. 

![image](https://user-images.githubusercontent.com/109082924/220436447-db758b8d-8557-4f22-8e9e-6c7a0d90e888.png)

Note that the last x-axis tick is well outside the end date. It seems that the intent was to add a tick at each ration interval.

## Expected Behavior
Ration report graphics should use the values found in the ration_report.csv for each pen. 

## Possible Solution
Changes to `ration_graphics` inside `graphics.py` to correctly assign x-axis values, with a tick for each ration interval date.

KFosterReed:
So weird!

JoeWaddell:
output handler was deprecated, new graphics functionality is being implemented

PR #462, opened 2023-04-24
Milk_production_reduction

JoeWaddell:
Initial attempt at a solution for issue #460

Note that the addition of the `milk_production_reduction` attribute in cow.py is because that is generated as a negative value in ration_driver.py


RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -298,6 +299,8 @@ def milking_update(self, sim_day, calving_interval):
             self.estimated_daily_milk_produced = estimated_daily_milk_produced
         else:
             self.estimated_daily_milk_produced = 0
+        estimated_daily_milk_produced += self.milk_production_reduction
+        self.estimated_daily_milk_produced += self.milk_production_reduction
         self.single_acc_milk_prod += estimated_daily_milk_produced
```
KFosterReed, line 305: I think lines 302-303 should be within the if self.milking statement because otherwise it will be subtracting the milk reduction after the animals are dried off, right?

Also, I think that we need to update L304 to: 
`self.single_acc_milk_prod += self.estimated_daily_milk_produced`
so the accumulated milk production also incorporates the reduction.

JoeWaddell: Good catch, thanks! I've moved those up, and changed the argument for single_acc_milk_prod (and the others below).

I've made these same changes on the user_defined_ration branch, but I'm still seeing the spikes and instant recovery in most cases...so I'm still missing some part of this fix.

YijingGong:
Hi Joe, regarding your question about milk production estimation elsewhere in the code, to the best of my knowledge, I don't believe it is estimated elsewhere.

 I looked at this PR and the branch of "user-defined-ration", I guess my biggest question is that I did not see how the `self.milk_production_reduction` in the cow.py is connected to the ration. 

When I run the "user-defined-ration" and print the value of `self.milk_production_reduction`, the value is always 0. Maybe this is because I did not (and do not know how to) input a user-defined ration that has less nutrition than needed. But at least from the code in cow.py, I did not see how `self.milk_production_reduction` is connected to the ration. 

We can talk more at tomorrow's meeting.

JoeWaddell:
Thanks @YijingGong! I can explain how this works in detail in the meeting tomorrow, but in short: the user_defined_ration branch needs to use the udr.json input, as that is set up to use a non-optimized ration (so I can test how it handles a bad fit). 

In that scenario the self.milk_production_reduction attribute will drop, however I think it's either still being reset somewhere unexpected, or I'm somehow doubly penalizing the milk production attribute in a single day. I'll be able to clarify in today's meeting. Hopefully I'm just missing something obvious.

YijingGong:
After talking with Joe, I now have a better understanding of the relationship between ration and cow milk production. Thanks Joe @JoeWaddell. I looked again at this weird behavior of milk production (seem to drop at ration interval). So I printed out the cow id, cow age, and her milk production, and try to match that to the graphs. Then, interestingly, I found that there is a discrepancy between the values printed in the terminal and the values shown in the graphs. 

For example, according to the below figure, around day 2025, there is a day that this cow (id = 252196) has milk production around 30kg. However, according to the printed values, there is not a day between 2000 to 2050 that the cow has milk production under 35kg. 

So I started to suspect the correctness of the function which plots the figures. But that seems ok. But on the way tracking back to the plotting function, I saw that in the function of update() in the cow.py, the current code runs the `self.update_body_weight_history(sim_day)
        self.update_milk_production_history(sim_day)` at the beginning of this function, which seems not very reasonable. So I changed those to lines of code to after `estimated_daily_milk_produced, fat_percent, \
        daily_fat_correct_milk_production = self.milking_update(sim_day, calving_interval)`, then I had the milk production figure of the following:

I am not confident that the bug is completely solved by moving those 2 lines of code, but I think this at least provides some new insights and directions that are worth double-checking.

JoeWaddell:
Fantastic @YijingGong ! I just had a discussion with @Fjordrunner , and he had some good thoughts related to that update function, as well. Changed the milk_production_reduction reset to prior to the daily update, and we're seeing similar behavior to what you've reported above. I'll continue investigating, but it looks like either or both fixes will resolve this issue/PR.

image

KFosterReed:
I've been playing with this a bit today and keep getting this odd behavior in the milk production that changes drastically between ration formulation intervals. I am putting it down for now and will come back to it tomorrow ... one of the recommended debugging strategies ;)

image
image

KFosterReed:
This seems to be working well for normal milk production ranges with 1 or multiple pens of lactating cows, but when the ration formulation is challenged with high milk production, it still seems to be producing the oscillating pattern: 
1 pen, normal Milk production: 
image
image

1 pen, High Milk production: 
image
2 pens, high milk production: 
image


Also, I tried to fix a few tests but the test for the milk_reduction_reset function in animal_management is still broken because I added a filter for lactating cow pens.

JoeWaddell:
Great work documenting this behavior: very odd. I have seen oscillations like that on my user_defined_ration branch, and thought it was an error isolated to my modified ration formulation code, so hopefully the root cause is the same. 

However, I haven't been able to replicate this so far on this branch, and instead get occasional KeyErrors with `manure_handlers[mm_pen.id].daily_update` (I think it has to do with pens being created during the simulation, but I'm still tracking down the cause).

Can you add the input jsons you're using to this branch?

JoeWaddell:
I've tried a few methods, but with increased lactation the issue of sharply oscillating reductions still occurs.

Not all the logic I've added in the latest commit will be retained, since the reduction attribute for CLOSE_UP pen animals shouldn't need to be reset once I'm sure the reset is happening when they first go dry. 

The next steps I'm taking to suss out what's causing this issue: 
- Add requirements to output manager (for initial diagnostic purposes, save to csv) to see if they oscillate in the same way, and if so: if it is due to animal additions/removals
- Merge in methods from user-defined ration branch to check which constraints aren't being met. (with logic to save summary of ration formulation failures)
-  ???

JoeWaddell:
image

Getting closer to a solution: I reduced the reduction amount, so it has more chances to solve at the optimization step, but one can still see a sharp rise a few years into this simulation. 

I did another for a much longer time, and what seems to be happening is that ration formulation is flipping between different solutions: 
image

One can see the way that changes here are reflected in the milk production (note that the x-axis for ration figures are still incorrect #363).
image

So the big swings in milk production appear to be a combination of factors at play: sharp rises in animal populations and a large steps in lactation reduction. Reducing the lactation reduction step helped, and the fix to tracking the number of animals should fix the issue.

JoeWaddell:
All new functions/methods have tests and documentation, so it is ready for re-review.

## Comments
The simulation is longer than the `input/animal_management.py` on master, done here for testing purposes. Will be reverted to default values prior to merge.

Likewise, the lactation curve parameters were modified to force failures when attempting to optimize a ration. Those will also be reverted prior to merge.

I'm not entirely sold on the percentiles functionality: the option to use it, and the percentile value itself, will be moved to the animal_constants file prior to merge, but I've left its current place because I wanted to see what everyone thought about the functionality and results.

Likewise some of the new variables reported in om are duplicated for testing purposes, and I'll remove extra information prior to merge.

## Questions 
I moved the constraints and lists of constraints out of the `optimize` function in `ration_NLP.py` to use them in the new constraint checking functions (thanks @ltp23 for writing those!), but would it be better to make a RationCosntraints class? 
- If so, should the rest of the functions in `ration_NLP.py` be part of a class? (e.g. RationOptimization)

Related to these constraints: I wrote tests for the new functions, and the tests both pass and function as expected. However, I couldn't figure out how to mock them properly, to use pytest fixtures. I left in one of my attempts commented out, but would love some insight as to whether it's worth trying to mock those, or if they're OK as-is. 

Re: the Animal Module discussion today, tagging you two, since you might be able to provide some insight here. Likewise we have a when2meet poll in the Animal_Module slack channel. @ew3361zh @ltp23

KFosterReed:
Unfortunately, I can't help with the testing questions or the constraints class question - maybe a discussion for next week's animal module meeting?

I did run the model a few times and we are certainly getting closer! I am really happy with the average milk production at a pen level outputs: 
image

The individual animal milk production outputs are also MUCH better but we are kind of having the opposite problem to what initiated this PR. When the milk production reduction is in effect it seems like there is one day that is left as 'not reduced'. Here is an example: 

image

JoeWaddell:
Curious! I don't see the same behavior on my end, but I'll keep trying to replicate this. I tried longer and shorter simulations and I'm not seeing those spikes. 

Can you provide the diet info you're using, if that might is how you're able to push the production down so much lower? I see that production is far lower than what I'm reporting using the `ànimal_management_animal` values on this branch.

image
image
image

And yes, great idea to discuss the test and class/constraint question on Wednesday. There should be at least one new class for the ration methods, so getting group input (with perspective from both SMEs and Developers) on those will be helpful.

KFosterReed:
One of these days I'll remember to give the information needed to replicate the errors I'm finding...

I was using the same animal info that was in the original PR but with 2 lactating cow pens: 
image

And then for the diet, I was giving the lactating cows an unreasonably bad diet by removing feeds and forcing them to exist on only grass silage and oats: 

image

JoeWaddell:
The problem comes from the order of operations in `AnimalManagement.daily_updates`, since `calc_nutrient_rqmts` is what sets the lactation reduction and updates the daily milk produced - however, life_cycle_manager.daily_update` has already been called at that point, and inside that function is the `cow.update_milk_production_history` method that updates and store the milk production history. 

The quick fix I implemented was to call the `cow.update_milk_production_history` again after ration formulation, but there might be other methods in daily_update that should be called again. 

A remaining oddity I see has to do with the cases where an animal first begins lactating after a dry period: though I think this will be fixed when https://github.com/RuminantFarmSystems/MASM/pull/496 is merged, because I *think* it's a problem with cows not being moved into the LAC_COW pen until the ration interval, but there could be that since they're starting lactation between ration intervals, they don't have an applied lactation reduction attribute. In the latter case, we could apply the pen average reduction value in the cases where that is lower than their attribute, but I think it needs some careful thought (and it may not be a major issue, in any case).
image

YijingGong:
Great work, Joe. I tested 'animal_management_bad.json' (thanks for providing it, otherwise, I am not sure how to alter the feed to a limited ration), it seems the milk production reduction is very reasonable both on the pen level and an individual level. The only two things I noticed are 1) pen allocation is pretty off (which should be solved by PR#497 soon); 2) the high-level milk production at the beginning of the lactation as you mentioned above. We should definitely double-check this after #496 is merged into the master to make sure this kind of behavior does not happen. 

I am thinking probably I should 'officially' review this PR after the merge of #496.

JoeWaddell:
Barring the issue I just opened, this should be ready for re-review by @YijingGong and others. I've run it on a few scenarios, and while in the `animal_management_bad.json` we're still seeing _some_ sharp dips/spikes, I think that's mainly due to culling of cows (note the spikes in culling for low production, which makes sense given the poor feed supply). A less-terrible feed supply (    `"lac_cow_feeds": [81, 103, 136, 139],`) still challenges production, but doesn't fail as completely as the current `bad` json, and so I think this is ready for full review and can be merged.

### note ###
PR #496 has a new version of `daily_updates` called `daily_updates2` (both are still in the code, but the former is not being used, and kept only for reference/easier rollback). In this PR my changes have now been duplicated, so behavior seen in the previous comments can be compared to the workflow as implemented in `daily_updates`.

tomhuhh:
Thanks Joe! I observed the same behavior as you described in issue #542. As Kristan commented below that issue, it is an issue worth looking into for version 2. That said, I noticed with animal 213266, its milk production experienced two drastic drop, one right after calving, and another one on ~ DIM 70. Please see below. I think I should expect a milder change in milk production on DIM 70, but I could be wrong. Another thing that I noticed is that when running a year of data, using the bad json file takes about 145 seconds while using the default json file takes about 48 seconds, with the former being 3 times of the latter. I wanted to bring it up but it might just has to do with the time it takes to formulate diets with restricted feed ingredients. Thanks! 



Screenshot 2023-06-01 at 4 27 13 PM

JoeWaddell:
Thanks @tomhuhh : you're exactly right about runtime. I think one reason it's happening is because the reductions require a re-assessment of animal requirements and then another nonlinear optimization attempt each time optimization fails. So what this means is that with the small reduction in lactation I've been using here, it can require many attempts, and this really slows things down. We can play around with the previous code (commented out here) that tries to base the reduction off the net energy for lactation requirement, but that was over-compensating and leading to much larger drops than needed.

JoeWaddell:
Update to `calc_pen_requirements` because percentile functionality causes erratic ration formulation behavior month to month. I'll open an issue to expand that functionality to use median and other methods in a separate PR.

YijingGong:
Hi Joe, I think this PR looks great to me! The milk reduction functionality is working well from my point of view. 

I think one thing that I noticed is interesting and might be worth discussing, but it propabaly goes beyond the scope of this PR. I noticed that a cow's milk production changed greatly once during her lactation. Here is an example, but not only this cow, but also other cows. 
Screenshot 2023-06-07 at 12 40 03 PM
So I took a look into it, and found that it seems that the big change (here it is a big increase) happens when the cow switches from pen 3 to pen 4 (I have two lactating cow pens, as shown in the below screenshot)
Screenshot 2023-06-07 at 12 42 42 PM
Then I took a look at the ration report for pen 3 and pen 4. Even though both pens have similar cow numbers, and they use the same feed (id = [51, 103, 136, 139], I used the purchased_feed_bad.json), but the amount of each feed is different for pen 3 and pen 4.
Pen3:
Screenshot 2023-06-07 at 12 47 21 PM
Pen4:
Screenshot 2023-06-07 at 12 48 10 PM
This is probably why there is a change in milk production when the cow switches to another pen. 

I believe this further verified your function of milk reduction is working correctly. We just need to think about whether this type of big milk production change after moving cows is what we want/expect.

YijingGong (approved):
Great work! One comment above.

tests/test_animal.py
```diff
@@ -1384,10 +1384,100 @@ def test_ration_report():
...
+
+    test_constraints = [{'type': 'ineq', 'fun': dummy_constraint_1},
+                        {'type': 'ineq', 'fun': dummy_constraint_2},
+                        {'type': 'eq', 'fun': dummy_constraint_3},
+                        {'type': 'eq', 'fun': dummy_constraint_1},]
+    result = RUFAS.routines.animal.ration.ration_driver.find_failed_constraints(dummy_solution(), test_constraints)
+    assert len(result)==2
+    assert result[0]['fun'].__name__ == 'dummy_constraint_2'
+    assert result[1]['fun'].__name__ == 'dummy_constraint_1'
+
```
ltp23: This one is somewhat trickier. My approach would be to pretend that `is_contraint_violated()` already works so I want to 'patch' it with my own fake values. So, we are only interested in the filtered results. 

```
@mark.parametrize('mock_results, expected_result', [
    ([False, True, False], [1]),  
    ([False, False, False], []),  
    ([True, True, True], [0, 1, 2]),  
])
def test_find_failed_constraints(mocker, mock_results, expected_result):
    # Arrange
    solution_x = MagicMock()  
    constraints = [mocker.MagicMock(), mocker.MagicMock(), mocker.MagicMock()]
    
    # The exact path depends on where is_constraint_violated() is 'used' (which may differ from where it is 'defined') - a common bug
    # The path below may need to be adjusted
    mocker.patch('RUFAS.routines.animal.ration.ration_driver.is_constraint_violated', side_effect=mock_results)
    
    # Act
    failed_constraints = find_failed_constraints(solution_x, constraints)
    failed_constraints_indices = []

    for i, constraint in enumerate(constraints): 
        if constraint in failed_constraints:
            failed_constraints_indices.append(i)
                    
    # Assert
    assert failed_constraints_indices == expected_result
```

tests/test_animal.py
```diff
@@ -1384,10 +1384,100 @@ def test_ration_report():
...
+    def dummy_constraint_3(dummy_solution):
+        return sum(dummy_solution)-10
+    test_constraints = [{'type': 'ineq', 'fun': dummy_constraint_1},
+                        {'type': 'ineq', 'fun': dummy_constraint_2},
+                        {'type': 'eq', 'fun': dummy_constraint_3},
+                        {'type': 'eq', 'fun': dummy_constraint_1},]
+    for i, constraint in enumerate(test_constraints):
+        result = RUFAS.routines.animal.ration.ration_driver.is_constraint_violated(dummy_solution(), constraint)
+        assert result == [False, True, False, True][i]
+
```
ltp23, line 1394: Here is an example using parametrization. 

```
def eq_constraint(x):
    return np.sum(x) - 10  # This 'eq' constraint checks if the sum of x is equal to 10

def ineq_constraint(x):
    return np.sum(x) - 10  # This 'ineq' constraint checks if the sum of x is greater than 10

@pytest.mark.parametrize("solution_x,constraint,expected", [
    (np.array([2, 3, 5]), {'type': 'eq', 'fun': eq_constraint}, False),  # Constraint not violated, hence expecting False
    (np.array([2, 3, 4]), {'type': 'eq', 'fun': eq_constraint}, True),
    (np.array([1, 3, 5]), {'type': 'ineq', 'fun': ineq_constraint}, True),
    (np.array([3, 4, 5]), {'type': 'ineq', 'fun': ineq_constraint}, False),
])
def test_is_constraint_violated(solution_x, constraint, expected):
    assert is_constraint_violated(solution_x, constraint) == expected
```

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -228,6 +290,82 @@ def __init__(self):
...
+            self.NEl = np.percentile(NEl, requirement_percentile)
+            self.MP_req = np.percentile(MP_req, requirement_percentile)
+            self.Ca_req = np.percentile(Ca_req, requirement_percentile)
+            self.P_req = np.percentile(P_req, requirement_percentile)
+            self.DMIest = np.percentile(DMIest, requirement_percentile)
+            self.avg_BW = np.percentile(BW, requirement_percentile)
+            self.avg_milk = np.percentile(milk, requirement_percentile)
+            self.avg_CP_milk = np.percentile(CP_milk, requirement_percentile)
+            self.avg_milk_production_reduction = np.percentile(milk_production_reduction, requirement_percentile)
+
```
ltp23: I could simplify this like the following but the trade-off is that when using `setattr()`, the IDE can't really help with type-checking or checking if an attribute exists on an object until runtime. 

```
attributes = {
    "NEmaint": NEmaint,
    "NEa": NEa,
    "NEg": NEg,
    "NEpreg": NEpreg,
    "NEl": NEl,
    "MP_req": MP_req,
    "Ca_req": Ca_req,
    "P_req": P_req,
    "DMIest": DMIest,
    "avg_BW": BW,
    "avg_milk": milk,
    "avg_CP_milk": CP_milk,
    "avg_milk_production_reduction": milk_production_reduction,
}

calc_method = 'mean'
for attr_name, data in attributes.items():
    if calc_method == 'mean':
        # populating the class variables as an average across cows for each requirement
        setattr(self, attr_name, stat.mean(data))
    else:
        # here we'd implement another method, e.g. percentile, median, etc.
        requirement_percentile = 90
        setattr(self, attr_name, np.percentile(data, requirement_percentile))
```

RUFAS/routines/animal/ration/ration_NLP.py
```diff
@@ -508,6 +508,23 @@ def get_ration_vals(x):
...
+con5 = {'type': 'ineq', 'fun': phosphorus_constraint}
+con6 = {'type': 'ineq', 'fun': protien_constraint}
+con7 = {'type': 'ineq', 'fun': NDF_constraint_1}
+con8 = {'type': 'ineq', 'fun': NDF_constraint_2}
+con9 = {'type': 'ineq', 'fun': forage_NDF_constraint}
+con10 = {'type': 'ineq', 'fun': fat_constraint}
+con11 = {'type': 'ineq', 'fun': DMI_constraint_upper}
+con12 = {'type': 'ineq', 'fun': DMI_constraint_lower}
+cow_cons = [con1, con2, con3, con4, con5, con6, con7, con8, con9, con10, con11, con12]
+heifer_cons = [con1, con3, con4, con5, con6, con7, con8, con9, con10, con11, con12]
```
ltp23: It looks like these numbered constraints are only set up here so this may help simplify it a bit

```
constraint_functions = [
    NEmact_constraint,
    NEl_constraint,
    NEgact_constraint,
    calcium_constraint,
    phosphorus_constraint,
    protien_constraint,  # Mispelled
    NDF_constraint_1,
    NDF_constraint_2,
    forage_NDF_constraint,
    fat_constraint,
    DMI_constraint_upper,
    DMI_constraint_lower
]

cow_cons = [{'type': 'ineq', 'fun': func} for func in constraints]

heifer_cons = [cons for cons in cow_cons if cons['fun'] not in [NEl_constraint, DMI_constraint_lower]]

```

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -309,8 +312,8 @@ def milking_update(self, sim_day, calving_interval):
                 0.0926 * (math.log(self.days_in_milk)) ** 2) * \
                           (math.log(self.days_in_milk) ** 1.107)
             daily_fat_correct_milk_production = \
-                0.4 * estimated_daily_milk_produced + \
-                0.15 * fat_percent * estimated_daily_milk_produced
+                0.4 * self.estimated_daily_milk_produced + \
+                0.15 * fat_percent * self.estimated_daily_milk_produced
```
ltp23: Just some math factoring

```
daily_fat_correct_milk_production = self.estimated_daily_milk_produced * (0.4 + 0.15 * fat_percent)
```

RUFAS/routines/animal/animal_management.py
```diff
@@ -1071,6 +1086,24 @@ def calc_avg_growth(self):
...
+
+        Returns
+        -------
+        float: The total milk produced in the herd (kg milk/day)
+        """
+        total_milk = 0.0
+        for cow in cows:
+            total_milk += cow.estimated_daily_milk_produced
+        return(total_milk)
+
```
ltp23, line 1103: If you like a one-liner,
```
return sum(cow.estimated_daily_milk_produced for cow in cows)
```

ltp23 (approved):
Great job, Joe!

RUFAS/routines/animal/ration/ration_driver.py
JoeWaddell: Nice option: I'll add your comment here to the issue I've opened about this, so that we can decide whether we use the method when we add alternative calculation methods.

PR #471, opened 2023-04-26
refactor leaching runoff erosion (1/3)

matthew7838:
### Summary
This is the first PR of refactoring `leaching_runoff_erosion`, the second PR will wrap up all functions and the third will be the main routine refactor works. If the second PR is too short, it will be combined with the main routine and just create a total of 2 PR. This PR focus on the pseudocode_soil documentation pseudocode_soil S.4.C.1 to  S.4.C.1. The purpose of this relatively short PR is to make sure current work is on right track with other works in the same module e.g. naming conversion etc.

### Changes
- All methods in this part are static

- Attribute and method names were changed to be more descriptive

- As for the current PR, there will be no attributes being modified directly

### Notes
@Isgotkowitz Please give suggestions if any of the variable naming does not align with your work in layer data or other files.


SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/leaching_runoff_erosion.py
```diff
@@ -0,0 +1,173 @@
...
+
+class LeachingRunoffErosion:
+
+    def __init__(self, soil_data: Optional[SoilData], field_size: Optional[float] = None):
+        """This method initializes the SoilData object that this module will work with, or create one if none provided.
+
+        Parameters
+        ----------
+        soil_data : SoilData, optional
+            The SoilData object used by this module to track carbon in the soil profile, creates new one if one is not
```
Isgotkowitz:
```suggestion
            The SoilData object used by this module to track nitrogen leaching and runoff in the soil profile, creates new one if one is not
```

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/leaching_runoff_erosion.py
```diff
@@ -0,0 +1,173 @@
...
+            volume of water in layer when saturated (mm)
+
+        Returns
+        -------
+        float
+            the concentration of the inorganic pools NO3/NH4 in the top soil layer (kg /mm H20)
+
+        References
+        -------
+        pseudocode_soil S.4.C.1
```
Isgotkowitz:
```suggestion
        SWAT Theoretical documentation eqn. 4:2.1.2
        
        Notes
        -----
        This equation has been modified for use in RuFaS, it no longer accounts for the fraction of porosity from which anions are excluded.
```
We don't want any references to pseudocode_soil, unless there are absolutely no other sources for where equations or processes come from.

Also, we should add this question to a list for Pete: "Should ammonium be leached and runoff the field in the same way as nitrates?" and reference this equation. pseudocode_soil says that this equation should be used for both nitrates and ammonium, but SWAT only specifies this for use with nitrates.

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/leaching_runoff_erosion.py
```diff
@@ -0,0 +1,173 @@
...
+        Returns
+        -------
+        float:
+            the amount of ammonium runoff from the first layer (kg/ha)
+
+        References
+        -------
+        pseudocode_soil S.4.C.2
+        """
+        return ammonium_concentration * runoff * runoff_extraction_coef
```
Isgotkowitz:
```suggestion
    @staticmethod
    def _determine_nitrate_runoff_amount(nitrogen_concentration: float,
                                         runoff: float,
                                         runoff_extraction_coef: float) -> float:
        """This method determines the amount of nitrate runoff for the first layer

        Parameters
        ----------
        nitrogen_concentration: float
            the content of inorganic (nitrate or ammonium) nitrogen in the top soil layer (kg N/mm H20)
        runoff_extraction_coef: float
            coefficient of extraction for runoff (unitless)
        runoff: float
            daily runoff of H2O (mm)

        Returns
        -------
        float:
            the amount of nitrate runoff from the first layer (kg/ha)

        References
        -------
        SWAT Theoretical documentation eqn. 4:2.1.5
        
        Notes
        -----
        The SWAT equation uses beta_NO3 as the runoff extraction coefficient (see SWAT Input file .BSN "NPERCO", page 104). RuFaS instead simplifies the runoff extraction coefficient to be 0.1 for determining nitrate runoff and 1.0 for ammonium runoff.
        """
        return nitrogen_concentration * runoff * runoff_extraction_coef

```
There should not be two methods that do the exact same thing.

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/leaching_runoff_erosion.py
```diff
@@ -0,0 +1,173 @@
...
+            amount of Fresh, Active, and Stable nitrogen (kg/ha)
+        bulk_density: float
+    `       bulk density of the soil layer (Mg per cubic meter)
+        Returns
+        -------
+        float
+            the soil nitrogen concentrations for the Fresh, Active, and Stable pools in soil(mg/kg)
+        References
+        -------
+        pseudocode_soil S.4.C.3
```
Isgotkowitz:
```suggestion
        SWAT Theoretical documentation eqn. 4:2.2.2
```

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/leaching_runoff_erosion.py
```diff
@@ -0,0 +1,173 @@
...
+        Parameters
+        ----------
+        nitrogen_amount: float
+            amount of Fresh, Active, and Stable nitrogen (kg/ha)
+        bulk_density: float
+    `       bulk density of the soil layer (Mg per cubic meter)
+        Returns
+        -------
+        float
+            the soil nitrogen concentrations for the Fresh, Active, and Stable pools in soil(mg/kg)
```
Isgotkowitz, line 98:
```suggestion
            the soil nitrogen concentrations for the Fresh, Active, and Stable pools in soil (g / metric ton soil)
```
I believe the pseudocode is wrong here, because it S.4.C.3 is exactly the same as 4:2.2.2, but SWAT specifies the units of conc_orgN as "(g N / metric ton soil)" in the paragraph beneath 4:2.2.1

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/leaching_runoff_erosion.py
```diff
@@ -0,0 +1,173 @@
...
+        enrichment_ratio: float
+            Enrichment ratio (unitless)
+        Returns
+        -------
+        float
+            nitrogen mass loss in erosion (kg/ha)
+
+        References
+        -------
+        pseudocode_soil S.4.C.4
```
Isgotkowitz:
```suggestion
        SWAT Theoretical documentation eqn. 4:2.2.1
```

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/leaching_runoff_erosion.py
```diff
@@ -0,0 +1,173 @@
...
+        float
+            enrichment ratio (unitless)
+
+        References
+        -------
+        pseudocode_soil S.4.C.5
+
+        Notes
+        -------
+        These numbers are modified ans suspected of retrieved from other references instead of SWAT, it will be kept
```
Isgotkowitz:
```suggestion
        @TODO These numbers are modified ans suspected of retrieved from other references instead of SWAT, it will be kept
```

Isgotkowitz (changes requested):
Good start on the equations. I think I miscommunicated that it was ok to just put references to the pseudocode instead of SWAT, I apologize for that. We want only SWAT references (or references to the model we're working off of), and to only make temporary exceptions in cases like S.4.C.5 where it's clear that the structure of the equation in RuFaS is different from SWAT.

morrowcj (approved):
I agree with Ed's comments and didn't notice anything else.

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/leaching_runoff_erosion.py
matthew7838: Any suggestions on identifying between nitrate and ammonium when deciding to pass in the right coef?  Right now, since NH4 coef is 1. Maybe I still keep 2 methods, and the one calculating ammonium will have no coef parameter. Do you have other suggestions?

Isgotkowitz: No, I think it should be one method. With situations like this I think that as long as the method is clearly documented (in this case, explicitly stating that the extraction coefficient should be 1.0 for ammonium and 0.1 for nitrates) it is ok put the responsibility of using the right coefficient value on the user/method that is calling the function.

That said, there will definitely be ways to use this equation in the main method that are better and clearer than others but we can figure that out later.

Isgotkowitz (approved):
Noice looks good

PR #472, opened 2023-04-26
Add residue to soil

### LINKED ISSUE #245: Add residue to refactored Soil class (SC_redesign)
The refactoring of `SC_redesign/Crop_and_Soil/crop/yields.py` calculates the residue to be added to soil at harvest. 

The updated `Soil` class will need to accumulate this residue, whenever it gets refactored.

----

Note that, it may be important to change `partition_residue` to NOT depend on individual crops. Rather, the method just needs
to know about the total 1) the total residue remaining on the surface and 2) the total below-ground (root) biomass.

morrowcj:
Implement residue transfer: When crops are harvested in a field, their residue is added to the soil. 

closes #245

### DIFF
--- a/SC_redesign/Crop_and_Soil/field/field.py
+++ b/SC_redesign/Crop_and_Soil/field/field.py
@@ -303,7 +303,10 @@ class Field:
 
     def harvest_scheduled_crops(self) -> None:
         """perform the harvest operation on all crops in the field, depending on the harvest operation, if today is
-        the crop's harvest day
+        the crop's harvest day.
+
+        After the harvest, this method adds any residue to the soil. Root residue is only added if the
+        harvest operation killed the crop.
         """
         for crop in self.crops:
             if not crop.data.is_harvest_day:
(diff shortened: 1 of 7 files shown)


SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -315,6 +318,10 @@ def harvest_scheduled_crops(self) -> None:
             if crop.data.next_harvest_operation == HarvestOperation.HARVEST_NOKILL:
                 crop.crop_management.manage_harvest(cut=True, collect_yield=True, kill=False)
 
+            self.soil.data.plant_surface_residue += (crop.data.yield_residue or 0)
```
Isgotkowitz, line 321: How does `or` work in this context?

Isgotkowitz (approved):
Nice PR, very clean. 
An operation that is still not implemented though is updating the organic matter content of a `LayerData` object when a plant is killed and the roots suddenly become part of the soil, but I think this will be pretty complicated to implement and merits its own future issue/PR.

SC_redesign/Crop_and_Soil/field/field.py
morrowcj: if `crop.data.yield_residue = None`, then 0 is used. Otherwise, the value of `yield_residue` is used:

```python
x = (None or 0)
print(x)  # 0
y = (5 or 0)
print(y)  # 5
z = (-1 or 0)
print(z)  # -1
```

morrowcj:
@Isgotkowitz, I've created issue #475 to address your point.

matthew7838 (approved):
Looks good to me, is this change from `yield_residue` to `plant_surface_residue` caused by a misunderstanding during refactoring? or is it just a new decision?

Issue #404, opened 2023-03-22
Rename attributes in `soil` module to better reflect meaning


Isgotkowitz:
In the dataclasses `soil_data.py` and `layer_data.py` there are some attributes with names that are not correct. They should be updated according to the following guide:

density/concentration/ratio: an amount per another amount
content: an amount on its own

morrowcj:
We should probably have a whole-team discussion about such things when the time comes to address this and similar issues.

PR #473, opened 2023-04-26
Add nitrogen pools to `LayerData` and initialize them

### LINKED ISSUE #370: Move validation/error raising from `layer_thickness` to `post_init` in `LayerData`
# Problem
The validity of the boundaries of a soil layer are checked in the `@property` method `layer_thickness` in `LayerData`. This is a bad way of checking for errors because it complicates an `@property` method and does not do it immediately after the `LayerData` object is initialized.
# Solution
Move the checking into the `post_init` method of `LayerData`.

Isgotkowitz:
# Summary
This PR add the nitrogen pools in each layer of the soil profile and initializes them as defined by SWAT.

Note: the substantive changes in this PR have not been tested, this is because it is important to get these changes into the codebase ASAP so that I (Ed) and Matthew can start working on different nitrogen cycling submodules while using the same nitrogen pools. Testing should and will be done, and should not interfere with the ongoing work in the nitrogen cycling module (because the testing will mostly be concerned with `SoilData` and `LayerData` being init'ed correctly)

# Main Changes
- `layer_data.py`:
  - the `residue` attribute has been introduced as a temporary variable for initialization (similar to `field_size`)
  - `nitrate` has been replaced with `nitrate_content`, and 4 other nitrogen pools have been added to the code.
  - `__post_init__()` now validates the top and bottom depths of the layer, and the validation of `field_size` has been moved upward but its mechanics have not changed.
  - Nitrogen pools are initialized in `_initialize_nitrogen_pools()`, which is called by `__post_init__()`
  - `determine_soil_phosphorus_area_density()` has been renamed and refactored so it works with any nutrient passed to it.
- `soil_data.py`:
  - There is a temporary `residue` attribute to make these changes work, can be removed after #472 is closed
  - The default soil profile does not set nitrate levels.
  - `_subdivide_top_layer()` now passes residue amounts when `LayerData.__post_init__()` is called.
  - The nitrate attribute name has been updated for `profile_nitrates_total()`

# Other Changes
- fixed a small type I made in `decomposition.py` docstring
- due to the changes in attribute and method names, the following test files are trivial changes that make the nitrate levels properly initialize or remove them entirely
  - `test_phosphorus_mineralization.py`
  - `test_soluble_phosphorus.py`
  - `test_evapotranspiration.py`
  - `test_infiltration.py`
  - `test_soil_config_factory.py`

### DIFF
--- a/SC_redesign/Crop_and_Soil/soil/layer_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/layer_data.py
@@ -268,7 +268,6 @@ class LayerData:
         References
         ----------
         SWAT Theoretical documentation eqn. 3:2.1.1, 2 and last paragraph on page 208 (for phosphorus initialization)
-                                       eqn. 3:1.1.1 - 5 and paragraph beneath eqn. 3:1.1.4
 
         """
         if self.top_depth < 0 or self.bottom_depth <= 0 or self.top_depth >= self.bottom_depth:
--- a/SC_redesign/Crop_and_Soil/soil/layer_data.py
+++ b/SC_redesign/Crop_and_Soil/soil/layer_data.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass, field, InitVar
 from typing import Optional
-from math import log
+from math import log, exp
 
 from SC_redesign.Crop_and_Soil.crop_and_soil_constants import MEGAGRAMS_TO_KILOGRAMS, HECTARES_TO_SQUARE_MILLIMETERS, \
     CUBIC_MILLIMETERS_TO_CUBIC_METERS, KILOGRAMS_TO_MILLIGRAMS, MILLIGRAMS_TO_KILOGRAMS
@@ -15,6 +15,10 @@ to represent its soil
 class LayerData:
     field_size: InitVar[float] = None
     """Size of the field (ha)
+        Note: this attribute is only used for initialization. After that it cannot be used.
+    """
+    residue: InitVar[float] = 0
+    """Amount of residue on the soil surface when this soil layer is initialized (kg / ha)
         Note: this attribute is only used for initialization. After that it cannot be used.
     """
     top_depth: Optional[float] = None
@@ -229,13 +233,28 @@ class LayerData:
     soil_structural_carbon_amount: Optional[float] = None
     """amount of soil structural carbon decomposed into slow or active carbon (kg/ha)"""
 
-    def __post_init__(self, field_size: float):
+    # ---- Nitrogen
+    nitrate_content: float = field(init=False)
+    """Nitrate (NO3) content of this soil layer (kg / ha)"""
+    ammonium_content: float = 0
+    """Ammonium (NH4+) content of this soil layer (kg / ha)"""
+    active_organic_nitrogen_content: float = field(init=False)
+    """Active organic nitrogen content of this soil layer (kg / ha)"""
+    stable_organic_nitrogen_content: float = field(init=False)
+    """Stable organic nitrogen content of this soil layer (kg / ha)"""
+    fresh_organic_nitrogen_content: float = field(init=False)
+    """Fresh organic nitrogen content of this soil layer (kg / ha)
+        Note: all layers except the top layer are initialized with 0 fresh organic nitrogen."""
+
+    def __post_init__(self, field_size: float, residue: float):
         """Initialize all attributes in the dataclass that depend on other attributes
 
         Parameters
         ----------
         field_size: float
             Size of the field (ha)
+        residue: float
+            Amount of residue on the soil surface when this soil layer is initialized (kg / ha)
 
         Raises
         ------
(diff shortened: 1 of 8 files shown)


SC_redesign/Crop_and_Soil/soil/soil_data.py
```diff
@@ -192,6 +192,9 @@ class SoilData:
     total_residue = 0
     """"amount of total residue ever added to the field(kg/ha)"""
 
+    residue: float = 0
+    """TEMPORARY: the amount of residue currently on the soil surface (kg / ha)"""
+
```
morrowcj: I added an attribute called `plant_surface_residue` in PR #472. I think that variable should be used in place of this one.

morrowcj (approved):
Looks good to me - a small comment about residue.

SC_redesign/Crop_and_Soil/soil/soil_data.py
Isgotkowitz: I agree. I think it is likely that PR gets merged first, so I plan on merging SC_redesign into this one, deleting the above `residue` and replacing it with `plant_surface_residue`

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -229,6 +231,21 @@ class LayerData:
...
+    """Concentration of nitrates in this soil layer at beginning of the simulation (mg / kg soil)"""
+    nitrate_content: Optional[float] = None
+    """Nitrate (NO3) content of this soil layer (kg / ha)"""
+    ammonium_content: float = 0
+    """Ammonium (NH4+) content of this soil layer (kg / ha)"""
+    active_organic_nitrogen_content: float = field(init=False)
+    """Active organic nitrogen content of this soil layer (kg / ha)"""
+    stable_organic_nitrogen_content: float = field(init=False)
+    """Stable organic nitrogen content of this soil layer (kg / ha)"""
+    fresh_organic_nitrogen_content: float = 0
```
matthew7838, line 245: I used the word amount instead of content in other similar attributes in the carbon cycle module, we might have to unifiy this sometime in the future. I do think your naming is better than mine.

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -409,14 +484,14 @@ def determine_soil_phosphorus_concentration(labile_phosphorus: float, bulk_densi
...
+    def determine_soil_nutrient_area_density(nutrient_concentration: float, bulk_density: float,
+                                             layer_thickness: float, field_size: float) -> float:
+        """Converts a concentration of nutrients in the soil from (mg / kg soil) to (kg / ha)
 
         Parameters
         ----------
-        labile_phosphorus : float
-            Labile phosphorus content of this soil layer (mg phosphorous / kg soil)
+        nutrient_concentration : float
+            Nutrient concentration of this soil layer (mg / kg soil)
```
matthew7838, line 494: since the description already mentioned in the soil, maybe we can skip the soil in the unit?
```suggestion
            Nutrient concentration of this soil layer (mg / kg)
```

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -409,14 +484,14 @@ def determine_soil_phosphorus_concentration(labile_phosphorus: float, bulk_densi
...
-                                               layer_thickness: float, field_size: float) -> float:
-        """Converts a concentration of soil from (mg / kg soil) to (kg / ha)
+    def determine_soil_nutrient_area_density(nutrient_concentration: float, bulk_density: float,
+                                             layer_thickness: float, field_size: float) -> float:
+        """Converts a concentration of nutrients in the soil from (mg / kg soil) to (kg / ha)
 
         Parameters
         ----------
-        labile_phosphorus : float
-            Labile phosphorus content of this soil layer (mg phosphorous / kg soil)
```
matthew7838, line 419: Same as below.

matthew7838 (approved):
LGTM, one thing about the naming that I think I should bring it up for discussion, another is just the unit labeling simplification. Good PR.

SC_redesign/Crop_and_Soil/soil/layer_data.py
morrowcj: We were just talking about something similar: #475

Isgotkowitz: I think both 'amount' and 'content' are good and definitely clear enough, though I agree it would be good to unify them at some point. Also I think renaming these attributes to be consistent would fall under the scope of #404

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: You raise a good point, this was a little redundant. I did sort of the opposite though, I removed the specifics from the sentence description and kept the specific units in the parameter descriptions.

PR #476, opened 2023-04-27
Add animal nutrient requirement data by Pen to OM

ew3361zh:
This PR adds 2 main groups of data to the Output Manager:
- average animal nutrient requirement data by pen for all animal types except Calves
- milk data for cows


RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -316,6 +319,23 @@ def milking_update(self, sim_day, calving_interval):
 
         self.body_weight += self.daily_growth
 
+        info_map = {"class": self.__class__.__name__,
+                    "function": self.milking_update.__name__,
+                    "simulation_day": sim_day,
+                    "calving_interval": calving_interval,
```
KFosterReed: I am not sure if calving interval is really needed here? I think the number of calves (self.calves)  would be better to include here

KFosterReed: @YijingGong what do you think?

KFosterReed: Actually, after looking at this, I don't think either the calving interval or parity should go in the info map. Perhaps we could add parity to the values if desired. I pushed a commit that removes the calving interval (and adjusts one other info map in the pen)

YijingGong: I agree. To my understanding, info_map is to keep metadata, not the actual variable data (which is in the variables.json). 

To me, it looks like the milking-related data has been added to variables.json by this part: 
milk_data_update = {}
        milk_data_update["days_in_milk"] = self.days_in_milk
        milk_data_update["estimated_daily_milk_produced"] = self.estimated_daily_milk_produced
        milk_data_update["milk_protein"] = self.mPrt
        milk_data_update["milk_fat"] = self.fat_percent
        milk_data_update["milk_lactose"] = self.lactose_milk
        milk_data_update["lactating"] = self.milking
        milk_data_update["cow_id"] = self.id
        om.add_variable("milk_data_at_milk_update", milk_data_update, info_map)


Maybe I am wrong or missing some context information. Let me know.

KFosterReed: That seems right to me. @ew3361zh , could we add the parity to the milk_data_update?

input/animal_management.json
tomhuhh, line 1: Thanks Niko! I would change the end date back to "2010:264". Thanks!

ew3361zh: It looks like this change came from Kristan. I think maybe I had accidentally changed the end date to be 1 year out when the standard animal_management.json simulation should just be 20 days so I think we should probably leave this change in.

tomhuhh (approved):
Thanks Niko! This looks good to me!

KFosterReed (approved):
Looks good to me too!

PR #477, opened 2023-04-27
Refactor and test `humus_mineralization.py`

### LINKED ISSUE #474: refactor and test humus_mineralization.py (soil/nitrogen_cycling) into SC_redesign directory

Isgotkowitz:
# Summary
This PR refactors `humus_mineralization.py`, which handles the mineralization of active and stable organic phosphorus, both between the two and from active to nitrate, within each layer of the soil profile.

Note: in the old `humus_mineralization.py` file, the mineralization of active organic nitrogen to the nitrate pool was not conducted (it was in `mineralization_decomp.py`). I chose to implement it here because it more closely follows the SWAT sections. I (Ed) am tasked with refactoring `mineralization_decomp.py` so I will make sure to not repeat the process there.

# Main Changes
- `humus_mineralization.py`
  - Contains the `HumusMineralization` class, which contains one main update routine (`mineralize_organic_phosphorus()`) and two static helper methods (`_determine_intra_organic_mineralization()` and `_determine_organic_to_nitrate_mineralization()`)
  - All methods are tested in `test_humus_mineralization.py`
- `__init__.py`'s have been added inside the new directories `nitrogen_cycling/` and `test_nitrogen_cycling/` to make them proper modules.
- `crop_and_soil_constants.py` has had `FRACTION_OF_HUMIC_NITROGEN_IN_ACTIVE_POOL` added. It is used in both `HumusMineralization` and the `post_init` in `LayerData` so this seemed like a logical place to put it, although it may be better to have it as an attribute in either `LayerData` or `SoilData` so that the user may modify it.
- `layer_data.py`
  - The `humus_mineralization_rate_factor` attribute has been added, as well as a reference to its description in the SWAT docs. Its default is the same as in SWAT (0.0003).
  - `active_humic_nitrogen_fraction` has been replaced with `FRACTION_OF_HUMIC_NITROGEN_IN_ACTIVE_POOL` in `post_init`
  - The `@property` methods `nutrient_cycling_temp_factor` and `nutrient_cycling_water_factor` have been added and tested. @matthew7838 do you think this is a good place/way to implement them for use by multiple different modules?


morrowcj (approved):
Looks good, I didn't notice any issue.

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/humus_mineralization.py
```diff
@@ -0,0 +1,117 @@
...
+        ----------
+        soil_data : SoilData, optional
+            The SoilData object used by this module to track phosphorus cycling, creates new one if one is not provided.
+        field_size : float, optional
+            Size of the field (ha)
+
+        Notes
+        -----
+        Used to initialize a SoilData object for this module to work with, if a pre-configured SoilData object is not
+        provided (ha)
```
matthew7838:
```suggestion
        provided
```

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -511,6 +514,39 @@ def determine_soil_nutrient_area_density(nutrient_concentration: float, bulk_den
         total_nutrient_mass_in_kg = nutrient_concentration * soil_mass_in_kg * MILLIGRAMS_TO_KILOGRAMS
         return total_nutrient_mass_in_kg / field_size
 
+    # TODO: coordinate with Matthew to find best place for this so that it can be accessed by multiple modules
+    @property
+    def nutrient_cycling_temp_factor(self) -> float:
+        """The nutrient cycling temperature factor (unitless)
+
+        References
```
matthew7838, line 522: missing return doc string

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -511,6 +514,39 @@ def determine_soil_nutrient_area_density(nutrient_concentration: float, bulk_den
...
+        second_term = self.temperature / (self.temperature + exp(9.93 - 0.312 * self.temperature))
+        factor = 0.9 * second_term + 0.1
+        return max(0.1, factor)
+
+    @property
+    def nutrient_cycling_water_factor(self) -> float:
+        """The nutrient cycling water factor (unitless)
+
+        References
+        ----------
```
matthew7838, line 540: same as above

matthew7838 (approved):
Nice clean PR. Just some typos and missing docstrings.

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: I think that it is ok to skip the return section for `@property` methods, because they aren't exactly full methods. Plus we don't have any returns sections for any of the other `@property` methods. @morrowcj what do you think?

Isgotkowitz: Also Matthew will these `@property` methods be reasonably easy to work into the refactor you're doing right now? One alternative would be to make them `@staticmethods`

morrowcj: I agree. Document a `@property` it as though it is an attribute.

PR #479, opened 2023-04-28
Refactor and test `mineralization_decomp.py`

### LINKED ISSUE #478: refactor and test mineralization_decomp.py (soil/nitrogen_cycling) into SC_redesign directory

Isgotkowitz:
# Summary
This PR refactors and tests `mineralization_decomp.py`, which handles the decomposition and transfer of fresh nitrogen into the nitrate and active inorganic nitrogen soil pools.

# Main Changes
-  `mineralization_decomp.py`:
   - This module contains one non-static method, `mineralize_and_decompose_nitrogen()`, and 4 `@staticmethod`s that serve as helpers.
   - All methods are tested in `test_mineralization_decomp.py`
- `layer_data.py`:
  -  `fresh_organic_phosphorus_content` has been to make `mineralization_decomp.py` work, but has not actually been implemented yet. It does have a TODO and associated issue.
  - `residue_fresh_organic_mineralization_rate` has also been added, with references to it's definition in the SWAT input files. Can be modified by user, but not the functionality to modify it by crop type (which would be pretty complicated, not sure if it would be worth implementing this issue.


SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/mineralization_decomp.py
```diff
@@ -0,0 +1,202 @@
...
+        so they have been implemented in the same method, hence why this method takes in a generic nutrient.
+
+        TODO: In SWAT, this method takes the amount of residue in the soil (instead of carbon) and multiplies it by 0.58
+            to get the amount of carbon in the soil. This method should be refactored to do that when we get a tracker
+            for residue in LayerData - issue #481
+
+        """
+        nutrient_total = organic_nutrient + inorganic_nutrient
+        if nutrient_total == 0:
+            nutrient_total = 0.000_001
```
JoeWaddell: This is neat! I didn't realize until trying out this line that we can do this to break up long (potentially confusing) decimals

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/mineralization_decomp.py
```diff
@@ -0,0 +1,202 @@
...
+        The equations for determining the carbon-nitrogen ratio and carbon-phosphorus ratio are identical in structure
+        so they have been implemented in the same method, hence why this method takes in a generic nutrient.
+
+        TODO: In SWAT, this method takes the amount of residue in the soil (instead of carbon) and multiplies it by 0.58
+            to get the amount of carbon in the soil. This method should be refactored to do that when we get a tracker
+            for residue in LayerData - issue #481
+
+        """
+        nutrient_total = organic_nutrient + inorganic_nutrient
+        if nutrient_total == 0:
```
JoeWaddell: Is there ever a chance this could be negative, or even closer to 0.0 than the value below?

Also on another branch we used syntax that would be (in this case) something like: return max(nutrient_total, 0.000_001)

Also, is a minimum like this coming from the lit, or is it a "magic value" that would be worth putting in a constants file (so that when evaluating moderl performance it's easier to tinker with) - or is this close enough to 0 that it won't matter in any case?

JoeWaddell (approved):
The methods look good, the branch runs on my system, but note that the following block in the ReceptionPit class is adding vars(pen) to the info_map on a daily update, which will cause the JSON output to explode in file size! It's fixed on master, so once those changes are merged in, it'll be fixed (you can also delete that bit out of info_map directly).

```
        info_map = {"class": cls.__name__,
                    "function": cls.daily_update.__name__,
                    "manure_handler_daily_output": vars(manure_handler_daily_output),
                    _"pen": vars(pen),_
                    "bedding": vars(bedding),
                    }

```

Otherwise it looks good! Since I'm not a SME for this portion of the model I can't say much to the methods re: the science, but it looks like clean code, and the documentation looks to explain things well.  I just left a couple small comments. The tests look good, and they passed on my system. Looks like it's coming along nicely!

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/mineralization_decomp.py
Isgotkowitz: I know, I didn't know about it either until Loi told me about them, but they're a real life saver especially when writing conversion constants.

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/mineralization_decomp.py
Isgotkowitz: It should definitely never be negative, neither the organic or inorganic nutrient totals should ever be negative, that would be indicative of a major bug in the code. It is theoretically possible that the nutrient total here could be less than 0.000_001, though I am pretty sure it is very unlikely.

I like your suggestions about using `max(nutrient_total, <arbitrary lower bound>)` and placing the lower bound in a constants file instead of having it hardcoded into an `if` block.

As for the value of `0.000_001`, I got it by asking @morrowcj about what should happen in this situation, unfortunately SWAT does not specify. I will change it to be more easily modified, but I think a SME is needed to decide what value is appropriate to return here when there are no nutrients in the soil

SC_redesign/Crop_and_Soil/crop_and_soil_constants.py
```diff
@@ -38,3 +38,7 @@
 # Nitrogen
 FRACTION_OF_HUMIC_NITROGEN_IN_ACTIVE_POOL = 0.02
 """Defined in SWAT Theoretical documentation, page 186 in paragraph beneath eqn. 3:1.1.4."""
+
+MINIMUM_NUTRIENT_TOTAL = 0.000_000_001
```
ltp23: This is interesting. Wouldn't this result in a very large number when dividing by it?

ltp23 (approved):
Overall, the code looks good to me!

I know it's already too late to raise this point, but I'm wondering if there is another way to access soil layers besides indexing (`self.data.soil_layers[0]`). It seems error-prone to type an integer index every time and probably harder to refactor if you need to make many small changes later. I don't have any good suggestions yet, but since each soil layer is complex in itself, I would have a variable for each soil layer, but that may be impractical if there are many layers. At least, we have the unit tests in place, so that's great.

SC_redesign/Crop_and_Soil/crop_and_soil_constants.py
Isgotkowitz: Yes, which I believe is ok in this context. I'm going to leave this PR open for a while though so that Clay and I can take a look at the behavior of this module and make sure it logical and scientifically sound, and we'll probably add some more thorough documentation as we go to.

Isgotkowitz:
> I know it's already too late to raise this point, but I'm wondering if there is another way to access soil layers besides indexing (self.data.soil_layers[0]). It seems error-prone to type an integer index every time and probably harder to refactor if you need to make many small changes later. I don't have any good suggestions yet, but since each soil layer is complex in itself, I would have a variable for each soil layer, but that may be impractical if there are many layers.

@ltp23 I agree with all of this. Maybe it would be good to add `top_soil_layer` and `bottom_soil_layer` attributes in `SoilData` that point to `self.soil_layers[0]` and `self.soil_layers[-1]` respectively? I don't think soil profiles can really be generalized more than that, especially because different soil profiles can have different numbers of layers.

PR #480, opened 2023-04-28
DMI constraint update

JoeWaddell:
Draftiest of a draft: this will be a short PR to add the DMI constraint to heifers, but also to change the DMI constraint from one upper constraint into an upper DMI constraint and a lower DMI constraint. 

They are the DMI estimate plus or minus the `DMI_CONSTRAINT_PERCENT` value in the `animal_module_constants.py` file.

Example: if the DMI estimate is 10kg, and the `DMI_CONSTRAINT_PERCENT` is 0.2, the upper DMI constraint will be 12kg, and the lower DMI constraint will be 8kg.

I've also added a crude script that allows one to more rapidly assess what's going on in the ration reports.

### DIFF
--- a/RUFAS/routines/animal/animal_management.py
+++ b/RUFAS/routines/animal/animal_management.py
@@ -436,10 +436,10 @@ class AnimalManagement:
         """
 
         all_animals_added = animals_added + calves_born
+        
+        original_pen_populations = self.track_former_pen_population()
 
         self.remove_animals_from_herd(animals_removed)
-
-        original_pen_populations = self.track_former_pen_population()
 
         animal_type_mapping_dict = {
             'Calf': {'p_conc': self.p_conc['calf'], 'animal_list': self.calves,
--- a/RUFAS/routines/animal/animal_module_constants.py
+++ b/RUFAS/routines/animal/animal_module_constants.py
@@ -21,3 +21,5 @@ class AnimalModuleConstants:
     DEFAULT_MANURE_HANDLER = 'manual scraping'
     DEFAULT_MANURE_SEPARATOR = 'screw press'
     DEFAULT_MANURE_STORAGE = 'slurry storage outdoor'
+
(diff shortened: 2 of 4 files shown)


JoeWaddell:
Here I've added upper and lower DMI constraints, as well as a potential fix to the ration per animal, as also instituted in #483 (separate branches to test behavior independently). Having the fix in this branch helps assess any potential changes in DMI constraints. 

I've also added the DMI constraint for heifers, now that we have DMI calculation methods for both NASEM and NRC.

Finally: if you run `python model_evaluation\\quick_ration_analysis.py` you'll see the first few ration intervals for the first four pens, summarized as the ration fed to each animal, the total kg, and the percentages of each feed in the diet. This can be removed before merging to master, or kept in as a crude development tool.

KFosterReed (approved):
Nice work Joe! I think this works exactly as expected. If I change the DMI constraint percentage, I can see a difference in the ration. The lactating cow DMI tends to increase with increasing DMI leeway but the heifer ration actually goes both up and down. 

I do like the idea of having output assessment files in the master branch in the model evaluation folder. My vote would be to push it to the master branch so everyone can use it easily. I think @YijingGong might have some similar life cycle evaluation methods in a separate branch?

Either way, I think this is good to go!

KFosterReed (approved):
This is awesome! We are definitely removing animals on a daily basis now! Thanks Joe!

Ed731551 (approved):
Hi Joe. Proposed changes makes sense to me. Great job! Thanks.

tomhuhh:
Thanks Joe! This looks good to me. 

As we discussed this morning, the feature to add the animals to the pen on a daily basis will be implemented later. 

Also, with the file quick_ration_analysis.py, I had to change the pathname on my Mac to make it work, but other than that, everything works well. 

Thanks again!

PR #482, opened 2023-05-01
removed objects from OM info_maps to reduce size

ew3361zh:
This PR seeks to reduce the size of the output variables pool json file when a simulation is run with info_maps included.

When doing longer simulations, the size of the info_maps can balloon to unmanageable size if there are large objects present in info_maps or variables added - especially when they're added on a daily basis.

In the cases here, I looked to replace the data potentially lost from the object being removed with the relevant information. I will tag the relevant SME to ensure we're still capturing the necessary data as far as we can tell.

This PR I believe removes or replaces all the remaing serialized objects at the recursion max_depth level of 3 (the default standard recursion depth).


ltp23 (approved):
It looks good to me. I like that you are being more specific with what to be included in the output manager. Thanks Niko.

ew3361zh:
@varrma I think we could fold in #490 to what I've started here. This PR was originally to reduce info_map size associated with the manure module but I think we could easily expand it to handle some of the repetition and other excess data Kristan mentioned in issue 490. Let me know what you think.

KFosterReed:
After looking at this more closely, I think this actually covers some of what I shared in issue #490.. that's great! I would vote for pulling this in, keeping the clean_up_info_maps branch open and continuing to work towards some of the issues in 490 in the future.

KFosterReed (approved):
Nice improvement!

varrma:
@ew3361zh agree with that Niko, we can use this to be active branch as Kirstan mentioned. I can review the variables that are repetitive/ seeming more information to be omitted.

ew3361zh:
> @ew3361zh agree with that Niko, we can use this to be active branch as Kirstan mentioned. I can review the variables that are repetitive/ seeming more information to be omitted.

Sounds good, Varma. Let me know if you need any help reviewing the output json variables file. Happy to talk it over or go through it with you.

Issue #464, opened 2023-04-25
Ration supplied does not seem to scale with number of animals in a pen.


JoeWaddell:
Ration supplied does not scale with number of animals in a pen.

Note: pen.py calc_ration multiplies ration for pen by the length of animals_in_pen

Note: we can see from figures below that *sometimes* the feed scales with ration, so it may be a bug related to animal type (heifer i, ii or iii, dry and lactating cows). 

This may also be purely a graphical bug: I haven't fully investigated this problem, but wanted to be sure to bring attention to it, and show figures explaining the current behavior. 

## Expected Behavior
When animal populations change between ration intervals (note issue #189), amount fed (e.g. as reported in ration report CSVs and graphics) should increase or decrease relative to the number of animals in a pen.

## Current Behavior
Some examples below

From current master, in pen 2
![image](https://user-images.githubusercontent.com/109082924/234342167-bcff7bc1-5667-4b71-961a-5e27744c415e.png)

From PR #283 , which appears to have improved (yet not fully solved) problem #189 
pen 2
![image](https://user-images.githubusercontent.com/109082924/234341466-4d801c94-677f-4fc4-932a-9254f0da4d16.png)
pen 3
![image](https://user-images.githubusercontent.com/109082924/234341389-7f776510-3367-488a-bc53-40122a67fcd0.png)

JoeWaddell:
In animal_management.py:

```
        original_pen_populations = self.track_former_pen_population()
        self.remove_animals_from_herd(animals_removed)
```

Since currently (related to issue #189) pen populations only seem to decrease, reversing the order of these two lines (the reversed state, relative to master, is what I've quoted above) seems to improve the ration behavior. The reason is that the `original_pen_populations` value is calculated using `track_former_pen_population`, and that value is being fed into the `calculate_pen_rations` method in animal_management. 

![image](https://user-images.githubusercontent.com/109082924/235482517-e2e323bb-6de8-42cc-a18b-04e8e22eaa60.png)

The method used a difference in the previous and current pen populations - but since animals were removed *following* the original calculation, these values never varied, and as such the ration fed was never actually changing. This will require confirmation from an independent reviewer, but I think this change will close this issue.

JoeWaddell:
Closed following merge of PR #480 which contained a fix.

PR #483, opened 2023-05-01
Ration per animal: reordered logic to account for animal removal prior to calculate_pen_rations

JoeWaddell:
Refer to issue #464 for full details: this is the potential fix outlined in the last comment.


JoeWaddell:
Simple change in order of logic seems to fix allocation of rations to pen. Review still required to ensure that the order of operations hasn't changed any other functionality.

JoeWaddell:
this logic was added in PR #480 , which merged first

Issue #475, opened 2023-04-27
[Crop&Soil] Reconcile soil organic carbon between carbon cycling and phosphorus mineralization


morrowcj:
# Problem
Within Crop and Soil (redesign), it appears that there are 2 pairs of `LayerData` attributes that track organic carbon: `soil_overall_carbon_fraction` and `total_soil_carbon_amount` are both used by the `CarbonCycling` module while `percent_organic_carbon_content` is used by the `Erosion` and `PhosphorusMineralization` modules. Based on their descriptions, it appears that `soil_overall_carbon_fraction` and `percent_organic_carbon_content` are synonymous (though one a percent and one a proportion).

# Solution
These variables should be reconciled so that they are consistent across the entire Soil and Crop module. It appears that the value is calculated by the `CarbonCycling` module and then used by the other modules. This will require fixing equations that require % to work with fraction (or *vice versa*).

Better names should also be chosen for the selected variable. I'd suggest `LayerData.organic_carbon_fraction` and `LayerData.organic_carbon` for the actual amount.

#### Notes
This issue was identified in review comments by @Isgotkowitz on PR #472.

morrowcj:
I've added this to both SC projects, because it seems very important (and simple) and may be finished prior to v.1

Isgotkowitz:
Another aspect of this that I think is worth mentioning explicitly: `percent_organic_carbon_content` is used in relation to several other `LayerData` attributes (`percent_<name>_content` for clay, sand, silt, and rock, all under the Erosion header comment in `LayerData`), and they were all easy to manage before because they were all static. But when we replace `percent_organic_carbon_content` with some attributes from `CarbonCycling` those attributes will be dynamic, and we will need some way of keeping all the attributes balanced as the carbon attributes change. If we don't, we may find ourselves in the situation where the sum of carbon, rock, sand, silt, and clay percents are > 100.

ew3361zh:
Moving this to Blocked. @KFosterReed notifying you we discussed this in our dev-team meeting and it was determined to be too risky to work on right now (affects too many different outputs). The modeling of Carbon and Phosphorus was received with positive feedback from SMEs so we determined this is okay to delay.

Isgotkowitz:
Grooming session: @Isgotkowitz will look in `LayerData` to see if this duplication of soil carbon levels is still present.

allisterakun:
Grooming session: @Isgotkowitz will double-check the status of this issue.

[...]

PR #489, opened 2023-05-03
Refactor and test `denitrification.py`

### LINKED ISSUE #488: refactor and test denitrification.py (soil/nitrogen_cycling) into SC_redesign directory

Isgotkowitz:
# Summary
This PR refactors and tests `denitrification.py`, which handles the denitrification of nitrates in the soil profile (denitrification is the process of bacteria turning NO3 (solid nitrates) in gases (N2 or N2O)), and is based on section 3:1.4 of SWAT.

# Main Changes
- `denitrification.py`
  - This module contains only two methods, one non-static main routine and one `@staticmethod`.
  - There is a TODO referencing issue #475 to revisit the method call associated with it, once the Carbon module has been fully sorted out.
  - Both methods are tested in `test_denitrification.py`
- `layer_data.py` 
  - 3 attributes have been added, one of them tracks the annual total of denitrified nitrates, and the others are used to determine how much denitrification occurs.
  - The annual total is reset by the `do_annual_reset()` method.


SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
```diff
@@ -0,0 +1,80 @@
...
+        References
+        ----------
+        SWAT Theoretical documentation section 3:1.4, eqn. 3:1.4.1, 2
+
+        """
+        for layer in self.data.soil_layers:
+            if layer.nutrient_cycling_water_factor < layer.denitrification_threshold_water_content:
+                continue
+
+            # TODO: check that soil_overall_carbon_fraction is being used correctly in this context - issue #475
```
morrowcj, line 52: Has this been looked into yet?

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
```diff
@@ -0,0 +1,80 @@
...
+            nitrified_nitrates = self._calculate_nitrification_amount(layer.nitrate_content,
+                                                                      layer.denitrification_rate_coefficient,
+                                                                      layer.nutrient_cycling_temp_factor,
+                                                                      layer.soil_overall_carbon_fraction * 100)
+            layer.nitrate_content -= nitrified_nitrates
```
morrowcj, line 57: Does this need to be bounded to 0? I'm genuinely asking.

```suggestion
            layer.nitrate_content -= max(nitrified_nitrates, 0)
```

If so, perhaps the bounding should be part of the return statement for `_calculate_nitrification_amount()` instead of here.

morrowcj (approved):
Looks good to me. Just a small comment about a `#TODO` and another about potentially preventing negative nitrification.

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
Isgotkowitz: I agree, I will make that change to `_calculate_nitrification_amount()`

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
Isgotkowitz: No, #475 is still unresolved. In this case it just depends on whether `soil_overall_carbon_fraction` represents only organic carbon or not.

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -250,9 +250,19 @@ class LayerData:
...
 
+    annual_denitrified_nitrogen_total: float = 0
+    """Cumulative total amount of nitrates that have denitrified in a year (kg / ha)"""
+
     humus_mineralization_rate_factor: float = 0.0003
     """Rate factor for humus mineralization of active organic nutrients (nitrogen and phosphorus) (unitless)
         Reference: SWAT Input .BSN file, see "CMN" on page 101."""
+    denitrification_rate_coefficient: float = 1.4
+    """Controls the rate of denitrification in this layer of soil (unitless)
+        Note: acceptable values for this attribute are in the range [0.0, 3.0].
```
PooyaHekmati, line 261: what happens if the value is not in this range? do have test coverage for such scenarios?

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
```diff
@@ -0,0 +1,86 @@
...
+
+"""
+This module handles the denitrification of nitrogen in the nitrates pool, based on SWAT section 3:1.4.
+"""
+
+
+class Denitrification:
+
+    def __init__(self, soil_data: Optional[SoilData] = None, field_size: Optional[float] = None):
+        """This method initializes the SoilData object that this module will work with, or create one if none provided.
```
PooyaHekmati:
```suggestion
        """Initializes the SoilData object that this module will work with, or create one if none is provided.
```

PooyaHekmati: Since soil_data can be passed, I assume it exists somewhere else as well. So, when we create it or otherwise manipulate it, how do we handle side effects?

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
```diff
@@ -0,0 +1,86 @@
...
+        ----------
+        soil_data : SoilData, optional
+            The SoilData object used by this module to track the denitrification of nitrates in the soil profile,
+            creates new one if one is not provided.
+        field_size : float, optional
+            Size of the field (ha)
+
+        Notes
+        -----
+        The field size is used to initialize a SoilData object for this module to work with, if a pre-configured
```
PooyaHekmati, line 26: What if both are none?

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
```diff
@@ -0,0 +1,86 @@
...
+        """Conducts the daily denitrification operations.
+
+        References
+        ----------
+        SWAT Theoretical documentation section 3:1.4, eqn. 3:1.4.1, 2
+
+        """
+        for layer in self.data.soil_layers:
+            if layer.nutrient_cycling_water_factor < layer.denitrification_threshold_water_content:
+                continue
```
PooyaHekmati: maybe something like this would help? Just an idea, feel free to use a better name or discard this comment 
```suggestion
            nutrient_is_below_threshold= layer.nutrient_cycling_water_factor < layer.denitrification_threshold_water_content:
            if nutrient_is_below_threshold:
                continue
```

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
```diff
@@ -0,0 +1,86 @@
...
+
+            # TODO: check that soil_overall_carbon_fraction is being used correctly in this context - issue #475
+            nitrified_nitrates = self._calculate_nitrification_amount(layer.nitrate_content,
+                                                                      layer.denitrification_rate_coefficient,
+                                                                      layer.nutrient_cycling_temp_factor,
+                                                                      layer.soil_overall_carbon_fraction * 100)
+            layer.nitrate_content -= nitrified_nitrates
```
PooyaHekmati, line 57: why?

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
```diff
@@ -0,0 +1,86 @@
...
+
+        Notes
+        -----
+        The field size is used to initialize a SoilData object for this module to work with, if a pre-configured
+        SoilData object is not provided.
+
+        """
+        self.data = soil_data or SoilData(field_size=field_size)
+
+    def do_denitrification(self) -> None:
```
PooyaHekmati: I don't have enough subject matter expertise to judge this name, maybe denitrification is a keyword. But if it is not, maybe 'denitrify' can be the name?

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
```diff
@@ -0,0 +1,86 @@
...
+            if layer.nutrient_cycling_water_factor < layer.denitrification_threshold_water_content:
+                continue
+
+            # TODO: check that soil_overall_carbon_fraction is being used correctly in this context - issue #475
+            nitrified_nitrates = self._calculate_nitrification_amount(layer.nitrate_content,
+                                                                      layer.denitrification_rate_coefficient,
+                                                                      layer.nutrient_cycling_temp_factor,
+                                                                      layer.soil_overall_carbon_fraction * 100)
+            layer.nitrate_content -= nitrified_nitrates
+            layer.annual_denitrified_nitrogen_total += nitrified_nitrates
```
PooyaHekmati, line 58: same question: why? 
maybe the answer is obvious for an SME, but shedding more light on these two operations in the docstring might be a good idea.

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # --- Static methods ---
```
PooyaHekmati: I don't think this comment is needed.

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
```diff
@@ -0,0 +1,86 @@
...
+    @staticmethod
+    def _calculate_nitrification_amount(nitrate_content: float, denitrification_rate_coefficient: float,
+                                        temp_factor: float, percent_organic_carbon_content: float) -> float:
+        """This method calculates the amount of nitrate lost to denitrification.
```
PooyaHekmati: would it make sense to bring this here?
```suggestion
        """Calculates the fraction of nitrates lost to nitrification as nitrification factor, and bounds it to be in the range [0.0, 1.0].
```
maybe explain how bounding works? do we clamp or normalize?

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
```diff
@@ -0,0 +1,86 @@
...
+                                                                      layer.denitrification_rate_coefficient,
+                                                                      layer.nutrient_cycling_temp_factor,
+                                                                      layer.soil_overall_carbon_fraction * 100)
+            layer.nitrate_content -= nitrified_nitrates
+            layer.annual_denitrified_nitrogen_total += nitrified_nitrates
+
+    # --- Static methods ---
+    @staticmethod
+    def _calculate_nitrification_amount(nitrate_content: float, denitrification_rate_coefficient: float,
+                                        temp_factor: float, percent_organic_carbon_content: float) -> float:
```
PooyaHekmati: maybe?
```suggestion
                                        temp_factor: float, organic_carbon_content_percent: float) -> float:
```

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
```diff
@@ -0,0 +1,86 @@
...
+        References
+        ----------
+        SWAT Theoretical documentation eqn. 3:1.4.1
+
+        Notes
+        -----
+        This calculates the fraction of nitrates lost to nitrification as nitrification factor, and bounds it to be in
+        the range [0.0, 1.0].
+
+        """
```
PooyaHekmati:
```suggestion
        References
        ----------
        SWAT Theoretical documentation eqn. 3:1.4.1

        """
```

PooyaHekmati:
```suggestion
        References
        ----------
        SWAT Theoretical documentation eqn. 3:1.4.1

        """
```

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
```diff
@@ -0,0 +1,86 @@
...
+        exponential_term = exp(-1 * denitrification_rate_coefficient * temp_factor * percent_organic_carbon_content)
+        nitrification_factor = max(min(1.0, (1 - exponential_term)), 0.0)
+        return nitrate_content * nitrification_factor
```
PooyaHekmati:
```suggestion
        exponential_term = exp(-1 * denitrification_rate_coefficient * temp_factor * percent_organic_carbon_content)
        nitrification_factor = 1 - exponential_term
        clamped_nitrification_factor = max(min(1.0, (nitrification_factor )), 0.0)
        return nitrate_content * clamped_nitrification_factor 
```

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/nitrogen_cycling_tests/test_denitrification.py
```diff
@@ -0,0 +1,51 @@
+import pytest
+from math import exp
+from unittest.mock import patch, MagicMock, PropertyMock, call
+
+from SC_redesign.Crop_and_Soil.soil.nitrogen_cycling.denitrification import Denitrification
+from SC_redesign.Crop_and_Soil.soil.soil_data import SoilData
+
+
+# --- Static method test ---
```
PooyaHekmati: unnecessary

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/nitrogen_cycling_tests/test_denitrification.py
```diff
@@ -0,0 +1,51 @@
...
+    (78, 1.4, 0.88, 60),
+    (104.55, 0.33, 0.233, 44.55),
+    (22.12, 0.0, 0.5561, 71.223),
+    (204.445, 3.0, 0.781, 22.334),
+    (0.0, 2.55, 0.1554, 39.112)
+])
+def test_calculate_nitrification_amount(nitrates: float, denitrification_rate: float, temp_factor: float,
+                                        percent_carbon: float) -> None:
+    """Tests that the amount of nitrified nitrates is calculated correctly."""
+    observed = Denitrification._calculate_nitrification_amount(nitrates, denitrification_rate, temp_factor,
```
PooyaHekmati: we usually use 'actual' as the name of this variable.

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/nitrogen_cycling_tests/test_denitrification.py
```diff
@@ -0,0 +1,51 @@
...
+                                                               percent_carbon)
+    expected_nitrification_factor = max(min(1 - exp(-1 * denitrification_rate * temp_factor * percent_carbon), 1.0),
```
PooyaHekmati: it would be better if you hardcode the expected values in the data provider and remove this calculation. 
Generally speaking, it is a good practice to avoid duplicating code logic in the test. Also, hardcoding the expected value is easier to read and judge if things are making sense.

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/nitrogen_cycling_tests/test_denitrification.py
```diff
@@ -0,0 +1,51 @@
...
+                                        0.0)
+    expected = nitrates * expected_nitrification_factor
+    assert observed == expected
+
+
+# --- Test main routine ---
```
PooyaHekmati: unnecessary

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/nitrogen_cycling_tests/test_denitrification.py
```diff
@@ -0,0 +1,51 @@
...
+def test_do_denitrification() -> None:
```
PooyaHekmati: there are lots of magic values in this test case, is it possible to explain them a bit? So, if I changed the code and this test failed, how do I know which value is doing what? How do I know if it is ok to change these values to make the test pass or not? What is the significance of indices 0, 2, 3? Why index 1 is handled differently?

PooyaHekmati (approved):
Approving to unblock the PR.
The PR is good enough to be merged, but I think there is room for improvement before getting merged into the master.

PooyaHekmati:
For record, it took me an hour to review this.

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: As of right now, I think it is ok to leave the `denitrification_rate_coefficient` invalidated because it is such an  advanced attribute that anyone who messes with it will probably be looking at what it is and what acceptable values for it are.

More generally, input validation is light in the current state of the Soil and Crop module, and I think it would be good to decide on a strategy for input validation / error raising before diving into it and/or adding it for each piece of data that comes up. Some considerations for said strategy:

1. It needs to be consistent with how errors are handled throughout the rest of RuFaS, and as of this writing I do not know that team standard.
2. If errors are going to be raised, where will they be propagated up to, if propagated up at all? If propagated up, there should be appropriate handling in the methods/modules etc.
3. What are the metrics for deciding whether are error is unrecoverable or not?
  - For example, with the `denitrification_rate_coefficient`, if a user tries to set it to 3.5, should an error be raised that they tried to set it to an invalid value and the simulation stopped, or set it to 3.0 (the closest valid value), send a warning to the user, and run the simulation?
4. For attributes/elements that potentially change everyday, should these values be checked in every method they get passed to, or would it be better to have a method that runs once a day to validate every value?

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
Isgotkowitz: I am not sure exactly what you mean by side effects. If a pre-existing `SoilData` object is not passed to this module and it creates its own, the module will run in isolation unless that newly-created `SoilData` object is extracted from this object and used in all other modules being run.

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
Isgotkowitz: It will first see that `soil_data` is None, then move to trying to create a new `SoilData` instance. When it does this, the `post_init` method of `SoilData` will see that it is none and throw an error.

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
Isgotkowitz: Good suggestion

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
Isgotkowitz: I am in the same boat in terms of subject matter expertise, but I think denitrify is accurate for the operations being conducted here and is a much better name that the current one.

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
Isgotkowitz: Good call

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
Isgotkowitz: It is a good suggestion, but I am going to keep it as `percent_organic_carbon_content` so that it stays in sync with the attribute names in `LayerData`. This can be addressed by #404 though.

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
Isgotkowitz: I like keeping comments about specific implementation details in the `Notes` section, IMO it keeps the docstring a bit cleaner. Also this method calculates the amount of nitrate lost, calculating the nitrification factor is one of the components of this process.

I am not sure what the difference between normalizing and clamping is (but if I had to guess, I'd say clamp). As for an explanation of why the bounds are what they are, I think it would be pretty obvious to a SME but adding a note wouldn't hurt.

SC_redesign/Crop_and_Soil/soil/nitrogen_cycling/denitrification.py
Isgotkowitz: Good suggestion

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/nitrogen_cycling_tests/test_denitrification.py
Isgotkowitz: Good to know

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/nitrogen_cycling_tests/test_denitrification.py
Isgotkowitz: Interesting, I will use this testing strategy moving forward.

I am curious if there is a best practice for calculating those expected values, especially for a method that performs complex calculations? If I were going to determine expected values for this function outside of this test, I think I would write a method in interactive Python mode and use that to calculate the expected values. However this would still be problematic, because I still might make the same mistakes in both method implementations. Worse yet, if the method giving me the expected values was wrong and I don't have it saved then there wouldn't be a way for me to go back and check it as a possible reason for why my tests are failing.

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/nitrogen_cycling_tests/test_denitrification.py
Isgotkowitz: Good q's, I will add to the docstring about how this test achieves full coverage, which will explain those values.

PR #494, opened 2023-05-08
CropRotation input

morrowcj:
This PR updates the *design* of the `CropRotation` class. The next PR will try to implement/address the new design.

# Notes to reviewers
The code does not yet coincide with the design. Therefore, it is acceptable to just review the design document. The PR will try to implement the designs.

### DIFF
--- a/SC_redesign/Crop_and_Soil/manager/example_rotations.md
+++ b/SC_redesign/Crop_and_Soil/manager/example_rotations.md
@@ -1,6 +1,6 @@
 # Introduction
 
-This document is meant to outline how the CropInput system should work. 
+This document is meant to outline how the `CropInput` system should work. 
 
 ## Scheduling an individual crop practice
 
@@ -8,7 +8,10 @@ The `CropScheduleSpec` class is a dataclass that contains the specifications of 
 concerning the planting and harvesting of individual crops. The specification of this class should be flexible and allow
 for different types of patterns. 
 
-#### Example 1: A user wants to plant an instance of "corn" every year for 4 years starting in 1990. The planting and harvest
+Below are examples for how crop management might be specified.
+
+#### Example 1: 
+A user wants to plant an instance of "corn" every year for 4 years starting in 1990. The planting and harvest
 events should always take place on the same of the year. The following call to `CropScheduleSpec` should work:
 
 ```python
@@ -19,7 +22,8 @@ print(all_corn.planting_years)  # (1990, 1991, 1992, 1993)
 print(all_corn.harvest_years)  # (1990, 1991, 1992, 1993)
 ```
 
-#### Example 2: This time, the user wants to plant an instance of "corn" every other year, for 6 years, starting in 1990
+#### Example 2: 
+This time, the user wants to plant an instance of "corn" every other year, for 6 years, starting in 1990
 
 ```python
 from SC_redesign.Crop_and_Soil.manager.crop_input import CropScheduleSpec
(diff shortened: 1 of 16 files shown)


morrowcj:
@Isgotkowitz, I was hoping that, for an initial review, you would just look over `example_rotations.md`. The actual implementation is not ready for review yet, and I want to have some conversation about design. 

In particular, I'd like your critique of the `User Input` section ("Approach #1"). Does this seem like a useful structure for users to give input?

SC_redesign/Crop_and_Soil/manager/example_rotations.md
Isgotkowitz, line 1: This seems like a solid design to me.

Do you have any ideas about how `CropScheduleSpec`, `CropRotation`, and custom crop types will be handed to individual `Field` instances?

Isgotkowitz (approved):

SC_redesign/Crop_and_Soil/manager/example_rotations.md
morrowcj: As I see it, the `FieldManager` would get the data from the input system and the `CropRotation` for each field would be initialized and contained within `Field`s at the time of `Field` initialization. Does that make sense?

matthew7838 (approved):
Looks like a reasonable implementation, can't come up with other suggestions for now.

morrowcj:
@Isgotkowitz, I was unable to finish this. I'm struggling to get the implementation to work. I think you may have to take it over. I'll provide some instructions for what I've envisioned below. Please feel free to reach out if you have any questions at all.

## Goal

The goal here is to create a class that will be used by the `FieldManager` that tells it when and how crops should be planted and harvested in a particular field. In the last section of `example_rotations.md`, I've outlined how I envisioned data being structured when received from the user. This module needs to convert that raw input data into a more useful structure so that `FieldManager` can better utilize it.

## First attempt
My first attempt at handling Crop input can be found in `crop_rotation.py`. The goal for the classes in this file is explained in `example_rotations.md`. You reviewed both of these files last week. 

The more I thought about it, the more I realized that this implementation was pretty messy and I foresee some issues with it. So, I decided to make a second pass by breaking the problem down into smaller chunks.

## Second attempt
With the intention of replacing `crop_rotation.py`, I wrote `events.py` and `alternative_rotation_class.py` (which will need renaming).

I decided to create simple classes called `Events` (`events.py`). A `CropSchedule` object, which indicates how an individual crop instance will be handled, is composed of two different `Events`, one for planting, and another for harvesting. 

A collection of `CropSchedules` make up a `CropSequence`. This object's purpose is to house all the timing instructions (planting and harvesting) for a single crop species in a field. All the `CropSequence`s for a field's simulation will then be stored in a `CropRotationSchedule`. This is the object that `FieldManager` will use to execute management practices. 

#### Remaining issues
The main issue I was working on that I have not finished is converting user-input into `CropRotationSchedule` objects. As I imagined it:
* users will specify crop events by specifying a pattern and projecting that pattern into the future. The updated sequence projection function is `project_sequence()` (in `events.py`). I've tested and thought about it extensively and I'm pleased with it.
* My thought was that we use `project_sequence()` to expand the user input into long-lists and use those lists to create the `CropSequence`. My attempt at implementing the second part of that process can be found in the `CropSequence.create_from_sequences()` static method. I was not able to figure out how to make it work properly and I ran out of time (sorry)
* I then envision the `create_from_pattern()` method to create the sequences needed (via `project_sequence()`) and then call `create_from_sequences()` on them to create the object.
* Once the above is worked out, a `CropRotationSchedule` method should take the raw user input and convert with the steps above.

The user input outlined in `example_rotations.md` was based off of the older sequence projection function (`repeat_pattern()` from `crop_rotation.py`), so it will need to be updated to conform to the new pattern. 

## Next steps
After this PR is worked out, there are a few additional things needed for the Crop Input system:
* `FieldManager` (once created) needs to use `CropRotationSchedule` (one per field). It can loop through each `CropSequence` (and the nested `CropSchedule`s) to check if anything should be planted/harvested on a given day, through use of `Event.occurs_today()`.
* It needs to search for `crop_reference`s in the list of supported crops. If a match exists, the `FieldManager` needs to initialize that crop when appropriate. If a match doesn't exist, we need to check if the user defined a crop with that reference name. In that case, `FieldManager` will need to initialize a custom crop with the user's specifications (via `CropSpeciesDataFactory`). Similarly, if a user provides a custom crop with the name of a one of the default crops, the user-specifications should be used.

---

This is all I can think of for now. As I said, please let me know if you have questions or want to meet to talk about it more.

morrowcj:
I just added a few examples of `create_from_sequence()` in attempts to aid in its development/implementation.

PR #496, opened 2023-05-09
Track movement of animals between pens after life cycle daily update

### LINKED ISSUE #188: Timing of animal milking status update is not synched to pen changes
The animal's physiological status (or life cycle update) is changed before the ration, manure, and pen updates are made. This means that an animal will change from, for example, non-milking to milking and then be included in the pen level estimates for feed, manure, etc. in the non-milking pen. 

## Expected Behavior

All animals in each pen on every day are part of the animal combination that is assigned to that pen

## Current Behavior

An animal's status is updated before she moves pens and so is fed the wrong feed. 

## Possible Solution

change the order of the life cycle and pen calculations

## Steps to Reproduce

1.  If you run a simulation with a large herd, on any given day there will be one or two animals in the pen that are not in the same animal class as those specified by the animal combination
2.
3.
4.

### LINKED ISSUE #189: Animals in Pen only updated with ration formulation interval

### LINKED ISSUE #493: Sort lactating cows by days in milk before allocating to pens
Currently in the allocate_animals_to_pens function, if there is more than one animal pen for an animal group, they are split into the two pens based on their animal ID. This results in some animals with a very high milk production being in the same pen with animals with a very low milk production. It would be more representative if the animals were sorted based on their day in lactation (i.e. day in milk)

I would like the lactating cows to first be ordered based on their days in lactation and then split between pens.

An alternative that we used to have was to group them by their nutrient requirements (see animal/clustering_pen_grouping.py). In the future, I would like there to be multiple options for how the animals can be grouped (day in milk, parity, nutrient requirement) but day in milk is the most common and would make the ration formulation process easier when there are multiple pens of lactating cows. 

**Additional context**

ltp23:
Diagrams:
[Example flowchart for life cycle daily update](https://github.com/RuminantFarmSystems/MASM/blob/95ea9cc7adccd26cbcc322b0041cb73442dd74c8/RUFAS/routines/animal/diagrams_and_guides/life_cycle_daily_update_example.mmd)

[Flowchart for new version of animal management daily updates()](https://github.com/RuminantFarmSystems/MASM/blob/95ea9cc7adccd26cbcc322b0041cb73442dd74c8/RUFAS/routines/animal/diagrams_and_guides/animal_mgmt_daily_updates_flowchart.mmd)

When animals change pens, first they will be removed from their current pens and then get assigned to one of the pens of the new combination they belong to. It would be harder to observe these effects if we only have one pen for each combination, and the new, to-be-graduated-to class and the current class of an animal both belong to the same combination (e.g., heifer I becomes heifer II, but both types belong to the same combination). 

It is important that we have a consistent way of finding out what combination an animal should be assigned to. The tricky part is that depending on the grouping scenario an animal can be assigned to different combinations. I created a separate class for this called AnimalGroupingScenario. 

This class has two important methods: `get_animal_type(animal)` and `find_animal_combination(animal)`. Every time we want to find out the type of animal or the combination that an animal is associated with, we should use those two methods and minimize the usage of hard-coded names anywhere outside of that class.

I also incorporated this new setup into the ration driver and the optimization parts, and the changes I made here were only related to checking the types of animals. 

Regarding Git issue #493: Cows are now sorted by days in milk before allocation at ration intervals. This is easier to see if we have multiple pens for lactating cows. 

For the next PR, I'd want to move the AnimalCombination enum class to a separate file and remove the NONE option. The primary purpose of this is to reduce the chances of running into circular imports error. I may also re-address the addition of default pens to smooth out some of the inconsistencies in the code.


YijingGong:
Great update, Loi! A lot of work. I think the main functionalities are all there. Just some small points:

1. In the diagram (life_cycle_daily_update_example.mmd), why there are 33 heiferIII graduated, but there are only 13 calves being born? 
Screenshot 2023-05-10 at 11 17 39 AM

2. I spend some time figuring out what is the definition of 'combination', I think maybe you can provide a description of it to help the reviewers understand the problem better. 

3. I am looking at the herd_report.csv and the pen_reports (growth_report, manure_report, and ration_report) to check the functionality though I am not sure if it is the best way. I found inconsistency in the calf number in herd_report.csv (which is 48 for the first day) and the animal number of pen_0, which should be the pen for calves (the number is 61 for the first day). Maybe this is related to point 1?
I also checked other pens, pen_1 animal number matches heiferI + heiferII number; pen_2 animal number does not match heiferIII + dry cow number; pen_3 animal number matches lactating cow number.  

4. I also looked at the individual_animal_report.json, the "pen_history" matches exactly with the "event_history" and makes perfect sense to me!

5. For issue #493, it is hard to tell now because we only have one pen for lactating cows, but it seems the function is there. We can keep an eye on that when we have multiple pens of milking cows.

I guess only point 3 will need a bit of coding work. Other than that, it looks great to me.

ltp23:
Thanks Yijing for your review. Much appreciated. 

To answer some of your points:

1. Those were the numbers I saw at the end of day 1. I think we would need to review those two steps in the life cycle to figure out the reasons.

3. The number 61 you saw was the combined total of current calves and newborn calves after day 1. 
For pen 2, I reviewed the part in the life cycle code that was related to the checking of heifer III's for sale. I noticed I only had the number of sold heifers incremented, but didn't decrement the number of heiferIIIs by the same quantity. After making that update, I think the numbers look better now.

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -9,14 +9,17 @@
...
 from RUFAS.routines.animal.life_cycle.animal_base import AnimalBase
 from RUFAS.output_manager import OutputManager
 om = OutputManager()
 from typing import Optional
 from typing import Dict
 from RUFAS.routines.animal.ration import ration_constants
 
-def calc_rqmts(body_weight: float, mature_body_weight: float, day_of_pregnancy: int, animal_type: str, parity: Optional[int] = 0,
+def calc_rqmts(body_weight: float, mature_body_weight: float, day_of_pregnancy: int,
+               animal_type: AnimalType, parity: Optional[int] = 0,
```
JoeWaddell, line 22: The methods calling calc_rqmts will have to be refactored to use `AnimalType`, since currently they are still using string arguments like `cow` or `heifer`. This currently fails when using the NRC calculations as a result.

All of the calls to `calc_rqmts` will need to be updated accordingly. 

Likewise `ration_driver.set_requirements` and all calls (`animal_type` as a string is used in `set_requirements`, and in calls to it)

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -568,15 +570,15 @@ def calculate_NASEM_energy_lactation_requirements(animal_type: str, milk_fat: fl
...
+    Same calculations as done in the NRC (2001). Requirements are based on milk yield and composition.
 
     References
     ----------
     .. [1] The National Academies of Sciences, Engineering, and Medicine "Nutrient Requirements of Dairy Cattle, 8th edition."
         National Academic Press, Chapter 3 "Energy", pp. 30, 2021.
 
     """
-    if animal_type == 'cow':
+    if animal_type in [AnimalType.LAC_COW]:
```
JoeWaddell, line 581: Since animal_type is a string, it's never meeting the conditional here, and instead lactating cows are erroneously given the `else` value of 0.0, causing a huge drop in estimated DMI. This is likely an issue in all functions in this file

ltp23:
Thanks Joe! 

I have updated those nutrient requirements functions to incorporate the new AnimalType system. For the most part, I modified the following two checks:

`if animal_type == 'cow':` -> `if animal_type in [AnimalType.LAC_COW, AnimalType.DRY_COW]:`

`if animal_type == 'heifer':` -> `if animal_type in [AnimalTYpe.HeiferI, AnimalType.HeiferII,  AnimalType.HeiferIII]:`

RUFAS/routines/animal/animal_management.py
```diff
@@ -942,6 +984,42 @@ def allocate_animals_to_pens(self) -> None:
...
+            )
+            self.all_pens.extend(new_default_pens)
+            self.pens_by_animal_combination[animal_combination].extend(new_default_pens)
+            self._allocate_animals_to_pens_helper(animals, self.pens_by_animal_combination[animal_combination])
+
+        self.fully_update_animal_to_pen_id_map()
+
+    def _sort_animals_before_allocation(self):
+        # Sort lactating cows by days in milk in increasing order
+        self.cows = self._get_dry_cows(self.cows) + \
```
YijingGong, line 1020: I believe it's not necessary to include dry cows here since all dry cow's days in milk are zero. And they will likely never be in the same pen as lactating cows in reality.

YijingGong (approved):
I tested multiple pens for lactating cows this time. It seems it's working well. 

I noticed that fresh cows (who recently gave birth) will be first allocated to the last pen (in my case, it is pen5). Then at the ration interval, it will be moved based on her days in milk. Do you think the behavior could be improved by always allocating a fresh cow to the first lactating pen (in my case, it should be pen3) after she calves, because we know that her days in milk must be very small?
It's probably also a design question. @KFosterReed Do you think it is worth doing or it does not matter too much? 

Just put below individual output as an example, I have here pen0 for calves, pen1 for growing heifers, pen2 for springers and dry cows, and pen3,4,5 for lactating cows.
`"pen_history": [
                {
                    "classes_in_pen": [
                        "HeiferIII",
                        "DryCow"
                    ],
                    "end_date": 44,
                    "pen": 2,
                    "start_date": 1
                },
                {
                    "classes_in_pen": [
                        "LacCow"
                    ],
                    "end_date": 61,
                    "pen": 5,
                    "start_date": 45
                },
                {
                    "classes_in_pen": [
                        "LacCow"
                    ],
                    "end_date": 151,
                    "pen": 3,
                    "start_date": 62
                },
                {
                    "classes_in_pen": [
                        "LacCow"
                    ],
                    "end_date": 271,
                    "pen": 4,
                    "start_date": 152
                },
                {
                    "classes_in_pen": [
                        "LacCow"
                    ],
                    "end_date": 342,
                    "pen": 5,
                    "start_date": 272
                },
                {
                    "classes_in_pen": [
                        "HeiferIII",
                        "DryCow"
                    ],
                    "end_date": 386,
                    "pen": 2,
                    "start_date": 343
                }`
You can see this cow after calving, she's first allocated to pen5, then at ration interval, she is moved to pen3, then gradually move to pen4 and pen5.

RUFAS/routines/animal/animal_management.py
ltp23: Hi Yijing, this part is ok. It simply combines the two sublists of dry cows and sorted lactating cows into one big list. During the sorting at ration interval, the animals in those sublists will go their own way.

ltp23:
Thanks Yijing for making the git issue [535](https://github.com/RuminantFarmSystems/MASM/issues/535). I assigned myself to it.

JoeWaddell (approved):
This looks great: I haven't been able to find any futher issues, and I look forward to having this merged to master since it fixes the animal tracking issue! Should close out that issue, and is a wonderful contribution to code cleanliness and expanding the test coverage. Nice work!

PR #508, opened 2023-05-11
Fixed issues with MP calculations using both NRC and NASEM

tomhuhh:
Issue with NASEM: A conversion in the metabolizable protein calculation was incorrectly applied. Specifically, we should divide the milk_true_protein value by 100 instead of the net proteins because that value is a percentage. That issue resulted in underestimation of MP using NASEM in RuFaS.

Issue with NRC: The equation for MPm (MP for maintenance) is incomplete. It is missing the dry matter intake and metabolizable bacterial protein part.

### DIFF
--- a/RUFAS/routines/animal/ration/animal_requirements.py
+++ b/RUFAS/routines/animal/ration/animal_requirements.py
@@ -588,7 +588,7 @@ def calculate_NASEM_energy_lactation_requirements(animal_type: str, milk_fat: fl
 def calculate_NRC_protein_requirements(body_weight: float, conceptus_weight: float, day_of_pregnancy: Optional[int],
                                        animal_type: str, milk_production: float, milk_true_protein: float,
                                        calf_birth_weight: float, net_energy_growth: float, average_daily_gain: float,
-                                       equivalent_shrunk_body_weight: float) -> float:
+                                       equivalent_shrunk_body_weight: float, dry_matter_intake_estimate: float) -> float:
     """ Protein requirement for maintenance according to NRC (2001).
 
     Calculates the estimated total metabolizable protein requirement (MP) in kilograms per day
@@ -639,8 +639,13 @@ def calculate_NRC_protein_requirements(body_weight: float, conceptus_weight: flo
     # Metabolizable protein requirement for maintenance (g)
     # (note this is not the full calculation, which will be completed within the
     # non-linear program)
+    TDN_estimate = 0.7 #communication with Dr. Edward Garcia
+    MP_bactria_estimate = dry_matter_intake_estimate * 1000 * TDN_estimate * 0.13 #communication with Dr. Edward Garcia
+
     MPm = 0.3 * (body_weight - conceptus_weight) ** 0.6 + \
-        4.1 * (body_weight - conceptus_weight) ** 0.5
+        4.1 * (body_weight - conceptus_weight) ** 0.5 + \
+        (dry_matter_intake_estimate * 1000 * 0.03 - 0.5 * (MP_bactria_estimate / 0.68 - MP_bactria_estimate)) + \
+        0.4 * 11.8 * dry_matter_intake_estimate / 0.67
     # Growth Requirement
     # ---------------------
     # [A.Cow.B.2]-[A.Heifer.B.2]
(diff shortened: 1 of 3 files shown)


RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -639,8 +639,13 @@ def calculate_NRC_protein_requirements(body_weight: float, conceptus_weight: flo
     # Metabolizable protein requirement for maintenance (g)
     # (note this is not the full calculation, which will be completed within the
     # non-linear program)
+    TDN_estimate = 0.7 #communication with Dr. Edward Garcia
```
Ed731551: This value is a general assumption indicating a feed with 70% total digestible nutrients (TDN) just to give an initial input for allowing subsequent NRC calculations for metabolizable protein with the actual diet formulated based on nutritional requirements.

Ed731551 (approved):

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -743,11 +748,11 @@ def calculate_NASEM_protein_requirements(lactating: bool, body_weight: float, fr
     NPMFP = CPMFP*0.73
     NPGrowth = frame_weight_gain*0.11*0.86
     NPGest = gravid_uterine_weight_gain * 125
-    NPMilk = milk_true_protein*milk_production*1000
+    NPMilk = (milk_true_protein / 100) * milk_production * 1000
```
JoeWaddell: Here and elsewhere, let's use the `GeneralConstants` class from `general_constants.py`, since it includes things like `GRAMS_TO_KG`

The docstring in the file specifies that it's for the manure treatment module, but I think it's best to use these functions where possible for code clarity

RUFAS/routines/animal/ration/animal_requirements.py
JoeWaddell: Is this value worth putting in the `animal_module_constants.py` file (and associated `AnimalModuleConstants` class? Or should this be a TODO to calculate TDN from the previous rations, when formulated (using this as a placeholder value for the first formulation?)

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -743,11 +748,11 @@ def calculate_NASEM_protein_requirements(lactating: bool, body_weight: float, fr
...
     TargetEffMP = 0.69
     if lactating:
-        metabolizable_protein_requirement = (((NPscurf + NPMFP + NPMilk + NPGrowth) /
-                                              TargetEffMP) + (NPGest/0.33) + NPEndUrin)/100  # final div/100 is correcting the units g/kg
+        metabolizable_protein_requirement = ((NPscurf + NPMFP + NPMilk + NPGrowth) /
```
JoeWaddell, line 779: I think it would be helpful to have these abbreviations in the notes section (or elsewhere) in the docstring: at least to note that NP = Net Protein, but the others like `TargetEffMP` are a little arcane still. It's legacy code, but since we're updating this method already, it shouldn't be much extra work.

RUFAS/routines/animal/ration/animal_requirements.py
tomhuhh: Sounds good Joe! I will update these constants all together as soon as I finish the review.

RUFAS/routines/animal/ration/animal_requirements.py
tomhuhh: Thanks Joe and Edward! You are corret. This constant value of 0.7 is used as a placeholder value for the first formulation because this information comes from the diet.

RUFAS/routines/animal/ration/animal_requirements.py
tomhuhh: Thanks Joe! I have the docstring updated.

tomhuhh:
TODO: Update the unit test for NASEM DMI calculations.

KFosterReed (approved):
This looks good to me! The updates definitely change the composition of the rations formulated which is fantastic. Well done Haowen

JoeWaddell (approved):
Looks great! Nice job adding in the extra details about the abbreviations, that should be a big help moving forward and keeping the code succinct but still legible to non-SMEs.

PR #509, opened 2023-05-12
Implement irrigation system in `Field` module

### LINKED ISSUE #326: Combine all water operations into a single Field-level module
# Problem
Currently there are many different submodules within `Soil` and `Crop` (`Evapotranspiration` in `Soil`, `WaterDynamics` in `Crop`) that handle water operations for the `Field`. This is problematic because the processes in these submodules are interdependent, and having them in separate modules makes them difficult to coordinate and maintain. 

# Solution
Create a single, unified submodule within `Field` that coordinates and executes all water operation.

# Steps
- [x] Create a simplified way to irrigate a field. It will have the following input characteristics
    - `irrigation_amount` - an amount to irrigate the field with when irrigation happens.
    - `irrigation_interval` - the number of days to wait to irrigate again.
    - `rainfall_irrigation_threshold` - an amount of rainfall that will reset the days since irrigation counter.
  There will also be the following attributes to support irrigation
   - `days_since_irrigation` - counter for the number of days since irrigation occurred. When this counter reaches `irrigation_interval` or the rainfall on a day is >= `rainfall_irrigation_threshold`, this counter is reset. Also when the `irrigation_interval` is reached, the field will be irrigated.
   - `annual_irrigation_water_use_total` - total amount of water used for field irrigation in a year. Will be reset on a yearly basis.
- [ ] Finalize the daily update operations for `Crop` and `Soil` objects.
- [ ] Put irrigation methods, daily `Crop` update calls, and daily `Soil` water update calls into single `Field` level method (`cycle_water()`) 

Related Issues: #316, #450

Isgotkowitz:
# Summary
This PR implements a rudimentary method for irrigating fields. Each field needs two parameters to execute irrigation: an amount of water that the field should be watered with over a certain number of days. By not providing at least one of these variables, the user specifies that they wish for no irrigation to occur. When rainfall occurs, it counts as irrigation and reduces the amount of water that must be pulled from farm resources for that irrigation interval. Irrigation is only applied to a field when there is at least one crop planted in it that is not in dormancy.

# Main Changes
- `field_data.py`:
  - Has a number of irrigation variables.
  - `post_init` will initialize all irrigation variables based on the two user-specified ones (is tested in `test_field.py`)
  - `convert_liters_to_millimeters()` has been added as a `@staticmethod` (is tested in `test_field.py`)
- `field.py`
  - Method `_determine_watering_amount()` has been added and is the driver of irrigation in this module (is tested in `test_field.py`)
  - `cycle_water()` in `field.py` now checks if there is a non-dormant crop planted in the field, and if so executes the irrigation routine.

# Beyond V1
There are a number of different ways that irrigation in RuFaS could be enhanced. 
1) Adding an option to execute irrigation based on crop's water deficit.
    - This seems like a very unrealistic scenario to me, but SWAT has this option so it may be useful to have.
2) Adding crop-specific irrigation configuration options.
    - Some crops are thirstier than others, so adding a way to define irrigation routines based on the crop would like be more representative of how a farmer operates.
3) Adding time-based irrigation configuration options.
    - At some point in a crop's life cycle, it may be more vulnerable to water deficits than at other times. So it would be helpful to specify that when a crop has been in the ground for a certain number of days, or has reached a certain level of growth, the irrigation routine shoul change.
4) Adding alternate irrigation methods to RuFaS.
    - I am not 100%, but I believe the current irrigation implementation on simulates water being sprayed down on the field (which is why it gets treated the same as rain). But there are many other options for irrigation, including things like drip tape.


SC_redesign/Crop_and_Soil/tests_SC/field_tests/test_field.py
```diff
@@ -77,7 +81,7 @@ def test_add_crop():
         field.add_crop(crop)
         assert type(field.crops[i]) is Crop
     for crop in field.crops:
-        assert crop.data.field_proportion == 1/5
+        assert crop.data.field_proportion == 1 / 5
```
Isgotkowitz: The extra spaces were added here automatically by PyCharm `Reformat file` feature

morrowcj (approved):
Looks great. nice work.

ltp23 (approved):
The PR looks great to me! Thanks Ed. 

I like the use of the CurrentWeather class. It is short, simple, and to the point. 

The name Dormancy, though, is a little unusual for a class name.

PR #511, opened 2023-05-16
Refactor and unify water processes in Crop and Soil module

### LINKED ISSUE #316: Track free water available in the canopy
In the Crop module, there is no place where amount of free water held in the canopy is being tracked. The equations for how the amount of free water changes over a given day are in SWAT 2:2.3.1

### LINKED ISSUE #326: Combine all water operations into a single Field-level module

Isgotkowitz:
# Summary
This PR begins the changes that will unify the crop and soil module's water operations, and likely both of their daily operations. 

# Changes
-  `crop_data.py`
   - The attribute `initial_canopy_free_water` has been renamed to `canopy_water`, and an arbitrary maximum canopy capacity has been added.
   - The `@property` method `water_canopy_storage_capacity()` which returns the actual canopy storage capacity on the current day.
- `water_dynamics.py`
  - The main routine `cycle_water()` has had processes removed and placed into their own methods, in order to give the `Field` module more fine-grained control over what happens in what order.
  - `evaporate_from_canopy()` now evaporates water from the canopy based on what the evapotranspirative demand is on the current day.
  - `set_maximum_transpiration()` sets the maximum transpiration based on the evapotranspirative demand that has been reduced by evaporating from **all** the canopies of crops in the field.
  - Tests for the new methods have been added, and the test for `cycle_water()` has been paired down.
- `transpiration.py` has been renamed to `water_uptake.py`
  - This more clearly reflects what the module does.
  - The import statement in `crop.py` reflects this change.
  - The associated test file has also been updated, with some changes to patch a failing test.
- `field.py`
  - The `cycle_water()` method now takes a more fine-grained approach to handling water in the crop and soil modules. It starts by applying precipitation to the field, then calculating the day's evapotranspirative demand, then removing water from elements in the specified order (Note this method is not complete yet).
  - `_handle_water_in_crop_canopies()` was added to manage how much precipitation gets stored in the canopies of crops before reaching the soil surface.
  - `_evaporate_from_crop_canopies()` was added to handle evaporation from the crop canopies of all the crops in the field, and update the evapotranspirative demand as it did so.
  - `_determine_potential_evapotranspiration()` and `_determine_latent_heat_vaporization()` were copied from `soil/evapotranspiration.py`, as evapotranspiration is a field-level process.
  - Tests for all the new methods mentioned above have been added, except `cycle_water()` (will be tested when completed).
- `soil/evapotranspiration.py`:
  - A TODO has been removed for #316


SC_redesign/Crop_and_Soil/crop/crop_data.py
```diff
@@ -305,8 +305,19 @@ class CropData:
...
+    """Maximum amount of water that can be trapped in canopy on a given day when fully developed (mm).
+        References: SWAT Theoretical documentation eqn. 2:2.1.1 (see also SWAT Input file .HRU ("CANMX" page 233).
+        Note: this default is super arbitrary. It comes from the paper:
+            'Holder AJ, Rowe R, McNamara NP, Donnison IS, McCalmont JP. Soil & Water Assessment Tool (SWAT) simulated
+            hydrological impacts of land use change from temperate grassland to energy crops: A case study in western
+            UK. GCB Bioenergy. 2019;11:1298–1317.  https ://doi.org/10.1111/gcbb.12628'
+        which cites the following paper that I could not find:
+           'Wang, D., Li, J. S., & Rao, M. J. (2006). Winter wheat canopy interception under sprinkler irrigation.
+            Scientia Agricultura Sinica, 39(9), 1859–1864.'
+    """
```
morrowcj, line 320: glad you found something to put here. Hopefully our SEs can provide greater insight for a larger variety of crop species beyond wheat.

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -351,15 +351,37 @@ def assess_dormancy(self, daylength: float) -> None:
...
+        self.soil.percolation.percolate(self.field_data.seasonal_high_water_table)
+
+        full_evapotranspirative_demand = self._determine_potential_evapotranspiration(
+            current_weather.incoming_light, current_weather.max_air_temperature, current_weather.min_air_temperature,
+            current_weather.mean_air_temperature)
+
+        remaining_evapotranspirative_demand = self._evaporate_from_crop_canopies(full_evapotranspirative_demand)
+
+        for crop in self.crops:
+            crop.water_dynamics.set_maximum_transpiration(remaining_evapotranspirative_demand)
```
morrowcj, line 383: So, all the crops share the same evapotranspirative demand, because it is at the field level, correct? This (looping through crops sequentially) will run into problems if there are multiple crops in a field. The last crop in the list will (almost) never have to transpire since there will be little-to-no demand remaining. That will, in turn, lead to that crop not taking up water. 

I think this is OK for now but in the future we'll need to come up with a way to allow the crops to transpire simultaneously so that neither is getting consistently shafted. The first thing that comes to mind is using proportional allocation of the demand based on the relative crop densities, allow all crops to draw from their fraction of the pool, combine all of the fractional remainders and pass them to plants who's water demands weren't met by the initial round. I'm sure there's a better solution, though.

morrowcj: On a similar note, once the "max_canopy_water" thing gets sorted out, it might be best to calculate the total canopy water by averaging all the crops in a field together, weighted by their relative densities.

morrowcj (approved):
I think this is a solid V1 implementation, but see my comment. It has implications for the future.

SC_redesign/Crop_and_Soil/field/field.py
Isgotkowitz: > So, all the crops share the same evapotranspirative demand, because it is at the field level, correct? This (looping 
> through crops sequentially) will run into problems if there are multiple crops in a field. The last crop in the list will 
> (almost) never have to transpire since there will be little-to-no demand remaining. That will, in turn, lead to that crop not taking up water.

Actually, I think this is not an issue in this loop. In the above section, each crop in the field gets the same maximum transpiration value, so they will all uptake water at whatever rate is determined for them. 

However in the methods `_handle_water_in_crop_canopies()` and `_evaporate_from_crop_canopies()`, this is absolutely an issue because it does not fairly distribute the resource/demand shared by all crops. It iterates through the list of crops sequentially and uses that crop as much as possible before moving on to the next one.

I'll write an issue and add TODO's to these methods so this is not forgotten. Edit: issue #513

SC_redesign/Crop_and_Soil/crop/crop_data.py
Isgotkowitz: Fingers crossed. I wound up looking through some SWAT Google group and it seems like a lot of people have trouble finding values for this.

SC_redesign/Crop_and_Soil/crop/crop_data.py
```diff
@@ -400,6 +411,17 @@ def is_in_senescence(self) -> bool:
         """check if the plant is in senescence"""
         return self.heat_fraction > self.senescent_heat_fraction
 
+    @property
+    def water_canopy_storage_capacity(self) -> float:
+        """Maximum amount of water that can be held in the canopy (mm).
```
ew3361zh, line 416: Does this need a unit test?

ew3361zh (approved):
I think this looks good overall - there's a lot going on at once and some of it is a little dense in terms of seeing all the different moving pieces. Glad that Clay was able to review it as well.

I had one note about adding a unit test for the water_canopy_storage_capacity property.

Testing suite looked solid though. I saw the note about finishing up the cycle_water() testing later.

I was a little curious about that note alongside the note for the test for cycle_water() being pared down. It's labeled as an integration test - is the idea to add unit tests once it's further along? I guess I'm also unfamiliar with what is left to do with the function so maybe the test as written now covers enough to ensure it's doing what you need it to do right for now.

SC_redesign/Crop_and_Soil/crop/crop_data.py
Isgotkowitz: Yes good catch

Isgotkowitz:
> I was a little curious about that note alongside the note for the test for cycle_water() being pared down. It's labeled as an integration test - is the idea to add unit tests once it's further along? I guess I'm also unfamiliar with what is left to do with the function so maybe the test as written now covers enough to ensure it's doing what you need it to do right for now.

It has been the convention on the soil and crop team to label the tests of main routines of modules as "integration tests". The idea was that the methods tested are reliant on other methods in the class and also set values in the associated  `@dataclass`, so in that sense they are 'integrated' with larger pieces of the program. This is probably not the correct usage of the term 'integration test', however I do think the test suite is sufficient as is because tests that calculations are done correctly and that values are stored correctly.

PR #516, opened 2023-05-17
Updates to OM variable name tracker

ew3361zh:
This PR makes some updates to improve and expand performance of the OM variable-tracking method:

1. Renamed method to `collect_variable_names_and_contexts` to better describe what it does.
2. Adjusted docstrings of method to also better describe it's function.
3. Removed unneeded copying of variables pool within method.
4. Passed this method the exclude_info_maps flag to then send to its partner method, `_list_to_file_text` which allows that method to write an appropriate header message in the saved file regarding whether or not the list of variables includes info_map variable names.
5. Renamed variables within nested looping structure that pulls out variable names from variables pool to hopefully better show its logic.
6. Reworked testing to reflect these changes both for the `collect_variable_names_and_contexts` and `_list_to_file_text` methods.

The resulting variable names list with the info_maps included is nearly twice as long as the previous version of this method which did not collect info_map variable data.

### DIFF
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -351,16 +351,19 @@ class OutputManager(object):
         Collects names of all variables added to variables_pool along with the caller class and function contextual
         information to eventually be saved into a txt file in the given path to a directory.
         """
-        # vars_pool = self.variables_pool.copy()
-        # for key, value in vars_pool.items():
-        #     if isinstance(value, dict) and "info_maps" in value:
-        #         value.pop("info_maps")
+
         file_path = os.path.join(path, self._generate_file_name("variable_names", "txt"))
         var_set = set()
-        for key, value in self.variables_pool.items():
-            var_set.add(key)
-            var_set.update(f"{key}: {variable_name}" for values_list in value.values() for variable_dict in values_list
-                           if isinstance(variable_dict, dict) for variable_name in variable_dict.keys())
+        for class_and_function, variables_data in self.variables_pool.items():
+            for variable_data_value in variables_data.values():
+                for variable_data_item in variable_data_value:
+                    if isinstance(variable_data_item, dict):
+                        for variable_data_item_key, variable_data_item_value in variable_data_item.items():
+                            if isinstance(variable_data_item_value, dict):
+                                for variable_data_item_value_key in variable_data_item_value.keys():
+                                    var_set.add(f"{class_and_function}: {variable_data_item_value_key}")
+                            else:
+                                var_set.add(f"{class_and_function}: {variable_data_item_key}")
         var_list = sorted(var_set)  # sorted(set) sorts and then converts set into a list
 
         self._list_to_file_txt(var_list, file_path, exclude_info_maps=exclude_info_maps)
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -272,7 +272,7 @@ class OutputManager(object):
         except Exception as e:
             raise e
 
-    def _list_to_file_txt(self, data_list: List[str], path: str, exclude_info_maps: bool = False) -> None:
+    def _list_to_file_txt(self, data_list: List[str], path: str) -> None:
         """Saves a list into a text file
 
         Parameters
@@ -290,18 +290,7 @@ class OutputManager(object):
         """
         try:
             with open(path, "w") as var_names_file:
-                if exclude_info_maps:
-                    info_maps_absent_notice = "This simulation ran excluding info maps from the variables pool so no"\
-                     " data from info maps are included.\nIf you would like to see that data, run the simulation with"\
-                     " info maps included."
-                    var_names_file.write(info_maps_absent_notice + "\n\n")
-                else:
-                    info_maps_present_notice = "This simulation ran including info maps in the variables pool so"\
-                     " data from info maps are included.\nIf you would not like to see that data, run the simulation"\
-                     " with info maps excluded."
-                    var_names_file.write(info_maps_present_notice + "\n\n")
-                for variable_name in data_list:
-                    var_names_file.write(variable_name + "\n")
+                var_names_file.writelines(data_list)
         except Exception as e:
             raise e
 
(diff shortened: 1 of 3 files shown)


ew3361zh:
@PooyaHekmati would you be able to take a look and see if this addresses the issues we discussed for this method?

PooyaHekmati:
> @PooyaHekmati would you be able to take a look and see if this addresses the issues we discussed for this method?

It certainly is better than the last version. The changes I wanted were larger than the code suggestion, so I made a commit. I fixed some design issues, changed names, reduces complexity, and added format_options. In addition to the TODO that I left in the code, test cases need to be updated as well.

PooyaHekmati:
I'd also like to hear @JoeWaddell 's opinion on this.

ew3361zh:
> > @PooyaHekmati would you be able to take a look and see if this addresses the issues we discussed for this method?
> 
> It certainly is better than the last version. The changes I wanted were larger than the code suggestion, so I made a commit. I fixed some design issues, changed names, reduces complexity, and added format_options. In addition to the TODO that I left in the code, test cases need to be updated as well.

Thanks Pooya! 

One quick question after running the simulation and comparing the output files:

This code doesn't appear capture the same depth of variables - some of the values/info_map variables recorded are dictionaries and the variable names within those dictionaries aren't captured. This results in ~100 fewer variables in the results. Can you expand a little on the reasoning of that decision?

I will take a closer look though and give more feedback/thoughts.

PooyaHekmati:
> > > @PooyaHekmati would you be able to take a look and see if this addresses the issues we discussed for this method?
> > 
> > 
> > It certainly is better than the last version. The changes I wanted were larger than the code suggestion, so I made a commit. I fixed some design issues, changed names, reduces complexity, and added format_options. In addition to the TODO that I left in the code, test cases need to be updated as well.
> 
> Thanks Pooya!
> 
> One quick question after running the simulation and comparing the output files:
> 
> This code doesn't appear capture the same depth of variables - some of the values/info_map variables recorded are dictionaries and the variable names within those dictionaries aren't captured. This results in ~100 fewer variables in the results. Can you expand a little on the reasoning of that decision?
> 
> I will take a closer look though and give more feedback/thoughts.

That makes sense because I removed multiple nested loops. SMEs can give you better feedback if all those details are needed. In case they need those detail, you can create a new function whose job is to get a dict (values or info map) and return the flattened structure. This can be done by having a recursive function, or by serialization. We already have a serialization flow in the OM, it would be helpful to take a look.

RUFAS/output_manager.py
```diff
@@ -336,30 +335,54 @@ def save_errors(self, path: str) -> None:
...
 
-    def save_variable_names(self, path: str) -> None:
+    def save_variable_names_and_contexts(
+        self, path: str, exclude_info_maps: bool, format_option: str = "block"
+    ) -> None:
         """
-        Saves names of all variables added to variables_pool into a json file in the given path to a directory.
+        saves names of all variables added to variables_pool along with the caller class
+        and function contextual information into a txt file in the given path to a directory.
+        TODO: expand the docstring and explain format_options: "block", "inline", and "verbose"
```
Isgotkowitz: Is there an issue for this or otherwise some record/timeframe for it?

RUFAS/output_manager.py
```diff
@@ -336,30 +335,54 @@ def save_errors(self, path: str) -> None:
...
-        var_list = sorted(var_set)  # sorted(set) sorts and then converts set into a list
 
+        var_list = [f"_{exclude_info_maps=}, expect info_maps accordingly.\n"]
+        for name, variable_data in self.variables_pool.items():
+            if not variable_data["values"]:
+                var_list.append(f"{name}: **NO VARIABLES**\n")
+                continue
+
+            is_varible_nested = isinstance(variable_data["values"][0], Dict)
+            if is_varible_nested:
```
```suggestion
            is_variable_nested = isinstance(variable_data["values"][0], Dict)
            if is_variable_nested:
```

Isgotkowitz (approved):
Overall looks good to me. Spotted what may be a dangling TODO and a small typo.

RUFAS/output_manager.py
PooyaHekmati: We create issues when we want to postpone some todos to be done in the future. The TODO that I made here is expected to be done in this PR, so no issues are needed. Basically, I was asking Niko to do some stuff before merging the PR.

JoeWaddell:
> I'd also like to hear @JoeWaddell 's opinion on this.

This looks great: it's a lot tidier than my local code for looping through key value pairs, and the formatting makes it easy to quickly see what's available. This is nice work. I'll chime in more on the other comments here, where I have something to contribute.

JoeWaddell:
> That makes sense because I removed multiple nested loops. SMEs can give you better feedback if all those details are needed. In case they need those detail, you can create a new function whose job is to get a dict (values or info map) and return the flattened structure. This can be done by having a recursive function, or by serialization. We already have a serialization flow in the OM, it would be helpful to take a look.

The answer will vary from person to person, but my instinct is that it's showing an appropriate amount of information.

Re: Niko's concern about information being hidden, it could be useful (though certainly not imperative) to report the typing for each of the variables. If there's not a way to automate that process it's likely not worth the trouble, but it's something that might be able to split the difference between showing the full depth of nested variables/dicts/lists like this.

ew3361zh:
> > That makes sense because I removed multiple nested loops. SMEs can give you better feedback if all those details are needed. In case they need those detail, you can create a new function whose job is to get a dict (values or info map) and return the flattened structure. This can be done by having a recursive function, or by serialization. We already have a serialization flow in the OM, it would be helpful to take a look.
> 
> The answer will vary from person to person, but my instinct is that it's showing an appropriate amount of information.
> 
> Re: Niko's concern about information being hidden, it could be useful (though certainly not imperative) to report the typing for each of the variables. If there's not a way to automate that process it's likely not worth the trouble, but it's something that might be able to split the difference between showing the full depth of nested variables/dicts/lists like this.

I think it brings up a good question which is what this function will be used for and by whom - initially it came out of a request from Kristan to have a record of what the variables were that we're collecting (I think as a continually updating list on what we have available in the Output Manager). Pooya suggested adding in the info_map variable names as well. 

What were you imagining using this for, Joe? Or is there anything in your current local looping script that would be helpful to add in here as you imagine how you might use this information?

morrowcj (approved):
This looks good to me. I don't really have any specific comments to add to the conversation. I do have a question related to the OM that is not related to this PR, so I'll make an issue and tag @ew3361zh in it. 

Edit: The issue is #518

JoeWaddell:
> What were you imagining using this for, Joe? Or is there anything in your current local looping script that would be helpful to add in here as you imagine how you might use this information?

Just to follow up from our in-person meeting: I think it's good as-is, especially considering the deadlines we have. 

The use case I imagine here is this functionality being a utility for an SME or developer trying to quickly suss out if a specific variable/attribute is being saved, and where to find it. To that point, it could be useful to show the entire nested hierarchy of saved variables (so that a user could find a specific value inside a saved dictionary), but that might be an uncommon use. For now the variable names are certainly enough for a user to know where to look in the output JSONs, and in the code itself (in the case that they need to add another variable).

PR #526, opened 2023-05-22
Crop and Soil finish line document

Isgotkowitz:
This PR contains a document that outlines what work still needs to be done to finish the v1 Crop and Soil module, as well a timeline for tasks.

Note: this is by no means a finalized document. It will need to be updated as we discuss how requirements should be addressed and what a reasonable timeline is.


JoeWaddell (approved):
This is a big help! Nice work

matthew7838 (approved):
Looks good, is well thought out, and is pretty clear.

PR #527, opened 2023-05-22
Connect Soil and Crop module to OutputManager

Isgotkowitz:
### Summary
This PR tries to collect essential **daily** results simulated from the whole SC module by utilizing the output manager. This PR will try to collect variables as concisely as possible, but it will still require communication between SMEs from different modules to make sure this output gatherer meets their needs.

### Changes
- `output_gatherer` gathers data from soil, layer, field, and crops. Data that are needed to check if the simulation is modified daily will be sent via the output manager.
- `test_output_gatherer` were done on `send_daily_variables` and checks that `info_map` and `values` were added as expected.

### Notes
Current tests are done by asserting the variable dictionary pool value with stdout output that was inspected manually and making sure that it makes sense. There still exists some sort of formatting and spacing issues, that could be on the OM side. Should be some small problem.


SC_redesign/Crop_and_Soil/manager/output_gatherer.py
```diff
@@ -0,0 +1,191 @@
...
+
+            # ----------------------------adding soil data
+            self.om.add_variable("potential_evapotranspiration", field.soil.data.potential_evapotranspiration,
+                                 info_map)
+            self.om.add_variable("potential_evapotranspiration_adjusted",
+                                 field.soil.data.potential_evapotranspiration_adjusted,
+                                 info_map)
+            self.om.add_variable("transpiration", field.soil.data.transpiration, info_map)
+            self.om.add_variable("soil_evaporation_adjusted", field.soil.data.soil_evaporation_adjusted, info_map)
+            self.om.add_variable("maximum_soil_evaporation", field.soil.data.maximum_soil_evaporation, info_map)
```
Isgotkowitz: These are no longer present in `SoilData`. They should be replaced by just `evaporation`.

SC_redesign/Crop_and_Soil/manager/output_gatherer.py
```diff
@@ -0,0 +1,191 @@
...
+            self.om.add_variable("eroded_sediment", field.soil.data.eroded_sediment, info_map)
+            self.om.add_variable("accumulated_runoff", field.soil.data.accumulated_runoff, info_map)
+            # not sure-------------------
+            self.om.add_variable("surface_runoff_volume", field.soil.data.surface_runoff_volume, info_map)
```
Isgotkowitz:
```suggestion
```

SC_redesign/Crop_and_Soil/manager/output_gatherer.py
```diff
@@ -0,0 +1,191 @@
...
+            self.om.add_variable("available_phosphorus_pool", field.soil.data.available_phosphorus_pool, info_map)
+            self.om.add_variable("runoff_phosphorus_pool", field.soil.data.runoff_phosphorus_pool, info_map)
+            self.om.add_variable("days_since_application", field.soil.data.days_since_application, info_map)
+            # ----------------------------
```
Isgotkowitz: Attributes from the `vadose_zone_layer` should be recorded as well.

SC_redesign/Crop_and_Soil/manager/output_gatherer.py
```diff
@@ -0,0 +1,191 @@
...
+            # not sure
+            self.om.add_variable("rain_events_after_fertilizer_application",
```
Isgotkowitz: All the attributes underneath the headers `Fertilizer (Phosphorus Cycling)` and `Manure (Phosphorus Cycling)` should be put in here too.

SC_redesign/Crop_and_Soil/manager/output_gatherer.py
```diff
@@ -0,0 +1,191 @@
...
+                                 field.soil.data.rain_events_after_fertilizer_application,
+                                 info_map)
+
+            # ----------------------------adding layer data
+            for layer in field.soil.data.soil_layers:
+                info_map["prefix"] = "field" + field.field_data.name + " layer top depth " + str(layer.top_depth) + \
+                                     " layer bottom depth " + str(layer.bottom_depth)
```
Isgotkowitz: Instead of recording the depths, the index of the soil layer in `field.soil.data.soil_layers` should be used.

SC_redesign/Crop_and_Soil/manager/output_gatherer.py
```diff
@@ -0,0 +1,191 @@
...
+                # not sure----------------------------------------
+                # does the temp change?
```
Isgotkowitz: Yes, the temp changes and it should be recorded.

SC_redesign/Crop_and_Soil/manager/output_gatherer.py
```diff
@@ -0,0 +1,191 @@
...
+                self.om.add_variable("temperature", layer.temperature, info_map)
+                # --------------------------------------------
+                self.om.add_variable("percolated_water", layer.percolated_water, info_map)
+                # not sure ---------------------------------------
```
Isgotkowitz: `water_content` should be recorded.

SC_redesign/Crop_and_Soil/manager/output_gatherer.py
```diff
@@ -0,0 +1,191 @@
...
+                                     , info_map)
+                self.om.add_variable("passive_to_active_carbon_amount", layer.passive_to_active_carbon_amount
+                                     , info_map)
+                self.om.add_variable("passive_carbon_co2_lost_amount", layer.passive_carbon_co2_lost_amount
+                                     , info_map)
+                self.om.add_variable("plant_active_decompose_carbon", layer.plant_active_decompose_carbon
+                                     , info_map)
+                self.om.add_variable("soil_active_decompose_carbon", layer.soil_active_decompose_carbon
+                                     , info_map)
+                # --------------------------------------------------------------------
```
Isgotkowitz: Non-organic phosphorus contents and `mean_phosphorus_sorption_parameter` should be recorded

SC_redesign/Crop_and_Soil/manager/output_gatherer.py
```diff
@@ -0,0 +1,191 @@
...
+                self.om.add_variable("active_inorganic_unbalanced_counter", layer.active_inorganic_unbalanced_counter
+                                     , info_map)
+                self.om.add_variable("labile_inorganic_unbalanced_counter", layer.labile_inorganic_unbalanced_counter
+                                     , info_map)
+                self.om.add_variable("previous_phosphorus_balance", layer.previous_phosphorus_balance
```
Isgotkowitz:
```suggestion

```

SC_redesign/Crop_and_Soil/manager/output_gatherer.py
```diff
@@ -0,0 +1,191 @@
...
+                self.om.add_variable("slow_carbon_co2_lost_amount", layer.slow_carbon_co2_lost_amount
+                                     , info_map)
+                self.om.add_variable("passive_to_active_carbon_amount", layer.passive_to_active_carbon_amount
+                                     , info_map)
+                self.om.add_variable("passive_carbon_co2_lost_amount", layer.passive_carbon_co2_lost_amount
+                                     , info_map)
+                self.om.add_variable("plant_active_decompose_carbon", layer.plant_active_decompose_carbon
+                                     , info_map)
+                self.om.add_variable("soil_active_decompose_carbon", layer.soil_active_decompose_carbon
+                                     , info_map)
```
Isgotkowitz: In addition to all the carbon attributes above, I think it would be good to record the `soil_overall_carbon_fraction` and the `total_soil_carbon_amount` since they are both a bit easier to understand.

SC_redesign/Crop_and_Soil/manager/output_gatherer.py
```diff
@@ -0,0 +1,191 @@
...
+                self.om.add_variable("water_uptake", crop.data.water_uptake, info_map)
+                self.om.add_variable("water_stress", crop.data.water_stress, info_map)
+                self.om.add_variable("temp_stress", crop.data.temp_stress, info_map)
+                self.om.add_variable("nitrogen_stress", crop.data.nitrogen_stress, info_map)
+                self.om.add_variable("phosphorus_stress", crop.data.phosphorus_stress, info_map)
+                self.om.add_variable("accumulated_heat_units", crop.data.accumulated_heat_units, info_map)
+                self.om.add_variable("is_growing", crop.data.is_growing, info_map)
+                self.om.add_variable("is_dormant", crop.data.is_dormant, info_map)
+                self.om.add_variable("canopy_height", crop.data.canopy_height, info_map)
+                self.om.add_variable("leaf_area_added", crop.data.leaf_area_added, info_map)
```
Isgotkowitz: `leaf_area_index` as well

SC_redesign/Crop_and_Soil/manager/output_gatherer.py
```diff
@@ -0,0 +1,191 @@
...
+                                     , info_map)
+
+            for crop in field.crops:
+                info_map["prefix"] = "field" + field.field_data.name + " crop" + crop.data.name
+                self.om.add_variable("root_depth", crop.data.root_depth, info_map)
+                self.om.add_variable("biomass", crop.data.biomass, info_map)
+                self.om.add_variable("usable_light", crop.data.usable_light, info_map)
+                self.om.add_variable("biomass_growth_max", crop.data.biomass_growth_max, info_map)
+                self.om.add_variable("biomass_growth", crop.data.biomass_growth, info_map)
+                self.om.add_variable("previous_biomass", crop.data.previous_biomass, info_map)
```
Isgotkowitz:
```suggestion
```

SC_redesign/Crop_and_Soil/manager/output_gatherer.py
matthew7838: I thought that we would focus on identifying the layers with the depth. or is that just the case for identifying the top layer?

matthew7838 (approved):

SC_redesign/Crop_and_Soil/manager/output_gatherer.py
```diff
@@ -0,0 +1,230 @@
...
+                                 field.soil.data.grazing_stable_organic_phosphorus,
+                                 info_map)
+
+            # ----------------------------adding layer data
+            for index, layer in enumerate(field.soil.data.soil_layers):
+                info_map["prefix"] = "field" + field.field_data.name + " layer index " + str(index)
+
+                self.om.add_variable("temperature", layer.temperature, info_map)
+                self.om.add_variable("percolated_water", layer.percolated_water, info_map)
+                # This carbon part is confusing from the start, maybe look back later
```
Isgotkowitz:
```suggestion
```

Isgotkowitz (commented):
Looks good Matthew, just found one inline comment that I think is unnecessary.

SC_redesign/Crop_and_Soil/manager/output_gatherer.py
matthew7838: good catch, thanks.

SC_redesign/Crop_and_Soil/manager/output_gatherer.py
```diff
@@ -0,0 +1,229 @@
+from typing import List, Optional
+
+from RUFAS.output_manager import OutputManager
+from SC_redesign.Crop_and_Soil.field.field import Field
+
+
+class OutputGatherer:
+    def __init__(self, fields: List[Field]):
+        self.fields = fields
+        self.om = OutputManager()
```
ew3361zh: For all the other instantiations of OM, we're putting them outside the class def. I'd probably recommend doing the same unless there's a particularly strong reason you have it where it is in the class init method.

ew3361zh (approved):
I think this looks good. I had one note about where you're instantiating the OM which I'd recommend moving to align with how we're doing it in other modules.

One big request to you both though: 
I did not look through each of the daily variables you're adding so I don't know what types they all are. But we've had some issues with large objects being added on a daily basis (big dicts, big class instances). If you could review the individual variables again and ensure there's nothing particularly large that is being added that might make the output variables file particularly unmanageable, that would be greatly appreciated. I can expand on specific examples of this and what we've done to address it if that's helpful.

ew3361zh (approved):
I think this looks good. I had one note about where you're instantiating the OM which I'd recommend moving to align with how we're doing it in other modules.

One big request to you both though: 
I did not look through each of the daily variables you're adding so I don't know what types they all are. But we've had some issues with large objects being added on a daily basis (big dicts, big class instances). If you could review the individual variables again and ensure there's nothing particularly large that is being added that might make the output variables file particularly unmanageable, that would be greatly appreciated. I can expand on specific examples of this and what we've done to address it if that's helpful.

PR #529, opened 2023-05-23
`Field`-level method to execute daily process components.

Isgotkowitz:
# Summary
Adds a method that handles calling all daily update routines for crops and soil in the field.

# Changes
- `_execute_daily_processes()` added. This is the method that actually calls all the growth and non-water related update routines. It is tested in `test_field.py`.
- `cycle_water()` has been changed to `_cycle_water()`, because it should only be called internally by `Field`.
- `_execute_daily_processes()` is now called by `manage_field()` which is the main public-facing method that will be called daily during the simulation. Note `manage_field()` is not tested because it is not finished, and some important features like scheduling mechanism have not been implemented yet.

### DIFF
--- a/SC_redesign/Crop_and_Soil/field/field.py
+++ b/SC_redesign/Crop_and_Soil/field/field.py
@@ -350,7 +350,7 @@ class Field:
         it will allow subject-matter experts to more easily experiment with different orders.
 
         """
-        # TODO - issue #317
+        # TODO: implement snow addition, melting, and sublimation - issue #317
         snow_cover = 0
         total_plant_cover = self.field_data.current_residue + self._determine_total_above_ground_biomass()
         self.soil.soil_temp.daily_soil_temperature_update(current_weather.incoming_light,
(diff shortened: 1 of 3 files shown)


SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -349,7 +336,47 @@ def assess_dormancy(self, daylength: float) -> None:
...
+        current_weather : CurrentWeather
+            Object containing the environment conditions on this day.
+
+        Notes
+        -----
+        This method is designed to make it easier to change the order of process execution, which is desirable because
+        it will allow subject-matter experts to more easily experiment with different orders.
+
+        """
+        # TODO - issue #317
```
matthew7838: Maybe add a simple description to this TODO?

matthew7838 (approved):
Good PR, one small suggestion on the TODO comments. This is pretty straightforward and clear.

matthew7838 (approved):
Good PR, one small suggestion on the TODO comments. This is pretty straightforward and clear.

JoeWaddell (approved):
Looks good to me!

Issue #463, opened 2023-04-24
Milking Parameters are Not Assigned Correctly


KFosterReed:
## Expected Behavior

The parameters that define an animals lactation curve (l, m, and n) are randomly selected from a distribution. These parameters should be assigned to the animal once at the beginning of the lactation to establish their pre-determined expected milk production that varies between cows. Then, each day, a new random value should be added (or subtracted from their expected milk production) to represent daily variability 
## Current Behavior

Currently the random parameters are re-selected each day and there is no additional daily variation added to the milk production
## Possible Solution

Update the parameters only on the first day of lactation

## Steps to Reproduce

1. Review the milking_update() function in cow.py
2.
3.
4.

## Context (Environment)

The current implementation of the Monte Carlo methods for the lactation curve does not reflect our understanding of biology or how the lactation curve parameters were estimated. Each cow has a set of lactation curve parameters that defines her expected lactation. If those change every day, then each cow has too much randomness in their individual lactation curve. This was a big discussion point in Manfei's PhD dissertation and defense and it was my understanding that it was updated but we must have either lost the update or it has not been merged in yet.

YijingGong:
I think what you said makes perfect sense to me. My attention was never drawn to this issue previously and I also did not discover it by myself when reviewing the code. I also had no idea whether we had an update related to this. But I am glad that you bring this issue up. I think we should fix it.

A fix might be in life_cycle.py, every time a cow has a newborn, set the parameters (i.e. cow.l, cow.m, cow.n). 

I can draft a design doc for this. This should be a relatively easy fix. I think Loi would be a perfect person to code it since he refactored the life cycle. But I am not sure if he has the bandwidth at this moment. Let me know what you think.

KFosterReed:
Ok, great! Thanks Yijing. 

Setting the parameters at calving might be a good option - but how would that work for multiple parities? Also, I wonder if Niko might have more bandwidth to help with this and if you and he can work together to get it accomplished?

YijingGong:
Works for me! Could you please let Niko know about that? Or maybe we can talk as a team in next week's Animal meeting. I will write a design doc for this task (you already covered most of it in the issue, I will just copy and paste and add some details). 

For different parities, I think it should work the same as in cow.py. We can fetch the cow's parity in life_cycle.py, and assign the parameters based on the parity number. Of course, the details need further discussion, the above solution is based on my quick look at the code.

YijingGong:
[...]

PR #534, opened 2023-05-24
Assign lactation curve parameters at parturition

ew3361zh:
This PR seeks to address the lactation parameter setting design doc from @YijingGong and @KFosterReed which addresses issue #463:

# Overview

- The current implementation of the Monte Carlo methods for the lactation curve does not reflect our understanding of biology or how the lactation curve parameters were estimated. 

# Expected behavior

- The parameters that define an animal's lactation curve (l, m, and n in Wood’s model) are randomly selected from a distribution. These parameters should be assigned to the animal once at the beginning of the lactation to establish their pre-determined expected milk production that varies between cows. 
- There are 3 parameters of Wood’s lactation curve (l, m, n)
- Each parameter should be sampled from a normal distribution with mean and standard deviation defined in the input JSON file. 
- The mean and standard deviation are related to animal breed (Holstein or Jersey), and parity (1, 2, or >=3)
- Then, each day, a new random value should be added (or subtracted from their expected milk production) to represent daily variability

# Current behavior

- Currently, the random parameters are re-selected each day and there is no additional daily variation added to the milk production

# How this PR addresses these issues:

- Creates a function in cow.py for setting lactation curve parameters, removing it from the daily milking_update function which now gets called by life_cycle.py's _cull_cows_and_record_stats function after parturition (so no longer on a daily basis). 
- Creates a function in cow.py for calculating daily milk produced which moves logic from the daily milking_update function though still gets called from daily milking_update function.
- Adds logic in the milking_update function to provide a small daily variation in production based on the guidelines provided in the design doc (0 mean, 1kg variance).
- Adds wood params as cow class attributes set to 0 at initialization.
- Adds lactation_curve config, breed_index, and parity_index as cow class attributes.
- Adds functions in cow.py to set breed and parity indexes.
- Adds unit testing for all new functions

# Questions:
- How can this be successfully tested from the Animal Science side?
- Do we need logic to reset the wood params to 0 when the cow stops producing milk for their current new born calf?


YijingGong:
Hi Niko, I took a look at both the codes and the outputs. I think this PR is definitely in the right direction and already in good shape. Great work!

In terms of your first question, I don't have a complete answer. For me, I just look at the codes, to see if the logic makes sense. And then I run the model, to see if the milk production of cows looks reasonable (when dry, it should be 0; when lactating, the shape should roughly look like below with a small variation that we added). 
Screenshot 2023-05-25 at 3 47 25 PM
Here is one of the milk production outputs of a cow:
Screenshot 2023-05-25 at 5 03 14 PM
You can see that milk production is as expected in terms of shape, parity, dry period, etc. Also, because now we assign the parameters once at the beginning of lactation, the curve does not have large variations as before. Below is an example from the previous code, you can see sometimes the variation is very large (lactation 2):
Screenshot 2023-05-25 at 5 12 15 PM

For your second question, I think we don't need to set the wood params to 0, because when the cow stop lactating (and become dry cow), the days_in_milk is 0, and thus the estimated milk production is 0 based on the equation. But looking at the codes, I believe we can simplify this (and potentially also save some computation) by adding a line (`return 0, 0, 0`) in the else section. 
```
if self.milking:
    self.days_in_milk += 1
else:
    self.days_in_milk = 0
```  
and this part can be changed accordingly 
```
if self.milking:
    self.estimated_daily_milk_produced = estimated_daily_milk_produced
else:
```
But I think this does not affect the big picture.  

Another point I want to suggest is that we can `put the daily_milk_variation_mean = 0` and `daily_milk_variation_std_dev = 1` to a constant file (I am thinking either animal_constants.py or animal_module_constants.py) in order to avoid magic value problem. In addition, I can test whether the daily variation works or not more easily as well.

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -126,7 +127,15 @@ def __init__(self, args):
...
 
+        self.wood_l = 0
+        self.wood_m = 0
+        self.wood_n = 0
+
+        self.lactation_curve = AnimalBase.config['lactation_curve']
         self.milk_production_history = []
+        self.breed_index = 0
+        self.parity_index = 0
+        self.set_breed_index()
```
YijingGong, line 140: Should we also call self.set_parity_index() here too?

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -212,6 +221,43 @@ def is_close_up_dry(self):
...
+    def set_parity_index(self):
+        """Sets the cow's parity index for use in the lactation curve parameter calculation"""
+        self.parity_index = 2 if self.calves - 1 > 2 else self.calves - 1
+
+    def set_lactation_curve_params(self):
+        """
+        Sets cow's lactation curve parameters based on cow's lactation curve attribute.
+        Currently only set up for wood model.
+        """
+        if AnimalBase.config['lactation_curve'] == 'wood':
```
YijingGong: Just to be consistent with the below calculate_daily_milk_produced() function, it might be good to have 'if self.lactation_curve == 'wood':'

RUFAS/routines/animal/life_cycle/cow.py
YijingGong: I realized that you called cow.set_parity_index() in the life_cycle.py. I think you are right that set_breed_index() only needs to be called once when the cow is initiated, but cow.set_parity_index() needs to be called every time they give birth to calves.

YijingGong (approved):
Great work, Niko. I think it is working as we expected. Also, thanks for correcting the format of other places as well. One small comment, but not a big deal. 

A side note about merging into the master branch is that I saw Joe merged his 'milk_production_reduction' (PR #462) yesterday, and I saw you pulled the master branch yesterday too. But it seems that you pulled before he merged that branch. So It might cause some conflicts when merging? I am not sure though, just a heads-up for you.

RUFAS/routines/animal/life_cycle/cow.py
ew3361zh: My thoughts exactly

RUFAS/routines/animal/life_cycle/cow.py
ew3361zh: Great catch, thank you, I will make that update.

ew3361zh:
> Great work, Niko. I think it is working as we expected. Also, thanks for correcting the format of other places as well. One small comment, but not a big deal.
> 
> A side note about merging into the master branch is that I saw Joe merged his 'milk_production_reduction' (PR #462) yesterday, and I saw you pulled the master branch yesterday too. But it seems that you pulled before he merged that branch. So It might cause some conflicts when merging? I am not sure though, just a heads-up for you.

Thanks, Yijing. I will resolve those conflicts right now.

JoeWaddell:
Something a little funny is happening in this branch, relative to master, with respect to how total milk is calculated on a monthly basis:

Master on the left, current branch on right. Data apparently come from `life_cycle_manager.daily_milk_production`
image


Likewise if you look at the monthly means: Data apparently come from `pen.avg_milk`
image

Documenting the behavior for now, I'll see if I can work out what's going on. What's interesting is that it does seem to reach some plateau reflecting the predicted totals.

ew3361zh:
> Something a little funny is happening in this branch, relative to master, with respect to how total milk is calculated on a monthly basis:
> 
> Master on the left, current branch on right. Data apparently come from `life_cycle_manager.daily_milk_production` image
> 
> Likewise if you look at the monthly means: Data apparently come from `pen.avg_milk` image
> 
> Documenting the behavior for now, I'll see if I can work out what's going on. What's interesting is that it does seem to reach some plateau reflecting the predicted totals.

It appears that this has to do with where we set the lactation curve parameters interestingly. 

I ran through each of the changes in this PR and making them one by one, they each continued to produce similar cumulated milk totals and pen 3 average milk produced...until I changed where the call to newly established set_lactation_curve_parameter() was.

As soon as it's moved outside of cow.milking_update(), both the cumulated milk production and the pen 3 average milk produced switch from the graphs you have on the left to the ones on the right.

This PR currently calls cow.set_lactation_curve_params() from lifecycle.py in `_cull_cows_and_record_stats` at the `if new_born` check.


Screenshot 2023-06-15 at 5 44 15 PM

I tried moving it to be called it from cow.update() if a cow has a calf (i.e. this line in cow.update() comes back true:

`if self.days_in_preg > 0 and self.days_in_preg == self.gestation_length:`

Screenshot 2023-06-15 at 5 44 39 PM

This change met with the same result of having the graphs on the right. But it does show that the lactation curve parameters are not being reset every day which I know was a major goal of this PR and design document.

I'm going to look further into this but thought I would update with what I found.

ew3361zh:
Okay, I think I figured out what was happening - when I moved the wood lactation curve parameters to become attributes of a cow object, I set them to 0 to initialize them. But I didn't account for the cows created that already have calves. There is a check at the end of the cow object initialization that looks at `if 'days_in_milk' in args`. I just added calls to `set_parity_index()` and `set_lactation_curve_params()` if that check is true.

I will push my changes so you can see the effects.

My question is whether we want this lactation curve reset for cows that previously had calves at the point where they have another calf? 

@YijingGong  @JoeWaddell thoughts?

update note: I spoke with Kristan and she thought resetting the lactation curve parameters at every calving was a good idea so the changes currently in this PR are up to date with that logic in place.

JoeWaddell:
Looks good to me: I tried it with the small/bad feed lists, and behavior looks as expected. Only issues are the existing #569 and #542 (the latter of which I might have a fix for soon, after explaining the issue to @ew3361zh)
Nice work!

JoeWaddell (approved):

PR #536, opened 2023-05-24
Implement `TillageApplication`

Isgotkowitz:
# Summary
This PR implements and tests a tillage module, which is an important functionality on any farm for maintaining the fertility of the soil.

- Contains a public-facing method that both mixes stuff from the soil surface into the top soil layer, and mixes tilled soil layers in the profile.
- Contains two helper methods, one removes stuff from the top and the other actually mixes the soil profile.
- All methods are tested in `test_tillage_application.py`
- `Field` has a `TillageApplication` instance as an attribute.
- `finish_line_doc.md` is updated to show progress made by this PR.
- ***PLEASE CRITIQUE*** the docstrings especially. The helper methods in this module are written to handle every attribute that can be effected by tillage, so they use a lot of generic terms and descriptions that may not be super clear.


SC_redesign/Crop_and_Soil/field/tillage_application.py
```diff
@@ -0,0 +1,199 @@
...
+from SC_redesign.Crop_and_Soil.soil.soil_data import SoilData
+
+"""
+This module contains all necessary methods for executing tillage operations on a field, based on SWAT Theoretical
+documentation section 6:1.6 and the SurPhos plow.f file.
+
+Notes
+-----
+This module was written to be as flexible as possible, because every attribute on the soil surface and in the soil
+profile gets incorporated and/or mixed with the same logic. That is why the term "stuff" is used in the docstrings to
```
matthew7838: Do you think it would be a good idea to just use soil attributes instead of stuff?

SC_redesign/Crop_and_Soil/field/tillage_application.py
```diff
@@ -0,0 +1,199 @@
...
+        mixing_fraction : float
+            Fraction of stuff mixed and redistributed from each layer in the soil profile (unitless)
+
+        References
+        ----------
+        SWAT Theoretical documentation section 6:1.6, SurPhos Fortran code plow.f
+
+        Notes
+        -----
+        The tillage process starts by removing stuff from the soil surface and putting it into the top soil layer, then
```
matthew7838:
```suggestion
        The tillage process starts by removing surface soil attributes from the soil surface and putting it into the top soil layer, then
```
like this maybe?

SC_redesign/Crop_and_Soil/field/tillage_application.py
```diff
@@ -0,0 +1,199 @@
...
+        """
+        Calculates amount of stuff incorporated from the soil surface into the soil profile.
+
+        Parameters
+        ----------
+        data_container : object
+            Class instance containing the soil surface pool to be removed from (unitless)
+        attribute_name : str
+            Name of the pool to be removed from (unitless)
+        incorporation_fraction : float
```
matthew7838:
```suggestion
        data_container : object
            Class instance containing the soil surface pool to be removed from 
        attribute_name : str
            Name of the pool to be removed from
        incorporation_fraction : float
```
I don't think they're kind of redundant in this use case.

matthew7838 (approved):
Looks good, minor suggestions.

Isgotkowitz:
@matthew7838 good suggestions, I think it is going to make the docstrings read a lot better.

SC_redesign/Crop_and_Soil/field/tillage_application.py
```diff
@@ -0,0 +1,198 @@
...
+        SurPhos fortran code, plow.f lines 20 - 32.
+
+        Notes
+        -----
+        This method both calculates the amount that is removed from the soil surface and actually removes it from the
+        soil surface, returning the removed amount. The units of the value returned are the same as the units of the
+        pool being removed from.
+
+        """
+        amount_in_pool = getattr(data_container, attribute_name)
```
ltp23, line 204: Maybe do some error handling here in case the attribute doesn't exist. If possible, it would be better to more specific about the type of the data container.

ltp23 (approved):
It looks good to me. Lots of good naming. Thanks Ed!

I was thinking about how we access the soil layers by direct indexing. We could mitigate this a bit by creating a class whose sole job is to manage the soil layers. We can do the normal CRUD operations and some shortcut methods like get_top_layer(), get_second_layer(), etc.

Isgotkowitz:
> I was thinking about how we access the soil layers by direct indexing. We could mitigate this a bit by creating a class whose sole job is to manage the soil layers. We can do the normal CRUD operations and some shortcut methods like get_top_layer(), get_second_layer(), etc.

@ltp23 I will create an issue for this

Issue #521, opened 2023-05-22
Rename save functions to dump functions


PooyaHekmati:
### What
Rename all functions with names matching the pattern save_xyz to dump_xyz. Find references to these functions and fix them.

### How
It depends on the IDE. In VScode, you can right-click on the RUFAS folder in Explorer and click on find in folder. Alternatively, you can temporarily rename the functions back to save_xyz and right click on their names, and click on find all references. 

### Why
What these functions really do is dump and not save. We need to free up the save so we can use it more appropriately. In computer science, "dumping" typically refers to the process of generating a comprehensive report or output that contains detailed information about the state of a system or a specific component. It is commonly used for debugging purposes or when troubleshooting issues. The generated dump file may include data such as memory contents, register values, stack traces, or log information, which can be analyzed to diagnose and fix problems.

ew3361zh:
Addressed by PR #537

PR #537, opened 2023-05-25
updated OM save_xyz func names to dump_xyz across model

ew3361zh:
This PR seeks to address issue #521 to update the names of all the OM save_xyz functions to dump_xyz which more accurately reflects what they do.

Changes were made in OutputManager.py, test_misc.py, test_user_prompt.py, and main.py.

Doc strings were also updated from save -> dump.

A search across files for "save_" after these changes were made yielded no further references to the save_xyz OM functions.

Tested by running a simulation as well as running the test suite and did not find any issues.

### DIFF
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -237,26 +237,26 @@ class OutputManager(object):
         return f"{caller_class}.{caller_function}"
 
     def _dict_to_file_json(self, data_dict: Dict[str, Any], path: str) -> None:
-        """Dumps a dictionary into a JSON file
+        """Saves a dictionary into a JSON file
 
         Parameters
         ----------
         data_dict : Dict[str, Any]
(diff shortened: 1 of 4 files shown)


PooyaHekmati (changes requested):
Nice work.
I think the issue and the title of this PR are specifically asking to change the functions that have save_ prefix in their names. This doesn't apply to _dict_to_file_json and _list_to_file_txt.

ew3361zh:
> Nice work. I think the issue and the title of this PR are specifically asking to change the functions that have save_ prefix in their names. This doesn't apply to _dict_to_file_json and _list_to_file_txt.

Makes sense, I returned the docstrings on those functions back to their original state.

PooyaHekmati (approved):
LGTM

JoeWaddell (approved):
Looks good!

Issue #492, opened 2023-05-05
[Crop Soil] Implement `FertilizerApplication` class


Isgotkowitz:
# Problem
The current way that fertilizer may be added to a field is through `add_fertilizer_phosphorus()` method in `phosphorus_cycling/fertilizer.py`. This method only adds phosphorus to the field via fertilizer, which is a problem, because there are many other nutrients that can and are added to fields with via fertilizer.

# Possible Solutions
- Create a `FertilizerApplication` class within the `soil/` or `field/` modules that can take a fertilizer application with all its attributes and modify the soil appropriately

# Note
This is closely related to issue #433

Isgotkowitz:
## Update
A `FertilizerApplication` module has been added, #539. But it does not contain enough to close this issue.
Things to do/consider before closing this issue:
- move `add_fertilizer_phosphorus()` from `soil/phosphorus_cycling/fertilizer.py` into `FertilizerApplication`
- Find (more) appropriate parameters for fertilizer applications, and increase the number of ways that fertilizer can be applied. For example, providing a method for users to specify just masses of nitrogen applied instead of fractions.
- Allow users to specify what fraction of fertilizer should go below the first soil layer - this is a feature in SWAT (section 6:1.7) and should probably be implemented in RuFaS as well

Isgotkowitz:
This issue should only be addressed after the pilot testing of the Crop and Soil module wraps up and SMEs give the go-ahead to start changing things in the Crop and Soil module.

KFosterReed:
Groomed March 27, 2026: There is currently ability to apply N and P below the surface and assign proprortion of nutrient application to the suface and below surface. Also, currently the total mass of nutrient application is the only way for the user to provide fertilizer application rate input. I believe the relevent objectives from this issue are:

1. move add_fertilizer_phosphorus() from soil/phosphorus_cycling/fertilizer.py into FertilizerApplication
2.  Allow users to specify what fraction of fertilizer should go below the first soil layer - this is a feature in SWAT (section 6:1.7) and should probably be implemented in RuFaS as well

However both are low priority at this time.

PR #539, opened 2023-05-25
Implement `FertilizerApplication` module

Isgotkowitz:
This PR implements a `FertilizerApplication` module that sits at the level of `Field` and provides an interface for adding fertilizer to a field.

- `FertilizerApplication` contains a single method, `apply_fertilizer()` that applies fertilizer to the soil profile.
  - `apply_fertilizer()` uses a method for adding phosphorus via fertilizer, which is defined in the `soil/phosphorus_cycling/fertilizer` module. This method will be moved to `FertilizerApplication` eventually - issue #492 
  - `apply_fertilizer()` is tested in `test_fertilizer_application.py`
- `Soil` has had its `init` method changed - it now properly creates a `SoilData` object with the provided field size when a `SoilData` instance is not provided.
- `Field` has an instance of `FertilizerApplication` that it can use to apply fertilizer to the field.
- The finish line doc has been updated to reflect progress made by this PR.

Note: this implementation is not feature-rich (as it should be for the MVP), but will have more added to it as part of addressing #492


SC_redesign/Crop_and_Soil/field/fertilizer_application.py
```diff
@@ -0,0 +1,65 @@
...
+
+class FertilizerApplication:
+
+    def __init__(self, soil: Optional[Soil] = None, field_size: Optional[float] = None):
+        """This method initializes the Soil object that this module will work with, or create one if none provided.
+
+        Parameters
+        ----------
+        field_size : float, optional
+            Used to initialize a Soil object for this module to work with, if a pre-configured SoilData object is not
```
matthew7838:
```suggestion
           Size of the field. Used to initialize a Soil object for this module to work with, if a pre-configured SoilData object is not
```

matthew7838 (approved):
LGTM, I have just one small suggestion to make the docstrings more clear.

SC_redesign/Crop_and_Soil/field/fertilizer_application.py
```diff
@@ -0,0 +1,65 @@
...
+from SC_redesign.Crop_and_Soil.soil.soil import Soil
+
+"""
+This module provides a way for Field to apply fertilizer, based on SWAT Theoretical documentation section 6:1.7.
+"""
+
+
+class FertilizerApplication:
+
+    def __init__(self, soil: Optional[Soil] = None, field_size: Optional[float] = None):
```
ltp23, line 12: This part seems a bit tricky. Ideally, we want exactly one argument. When no arguments are provided, should we throw an error? But when both are provided, the field size just becomes redundant. I think the confusion comes from having two optional parameters but at least one is actually required. 

I think the better design is to require a soil object since the client code should have either a soil object or know the field size. For the latter, it may as well create a soil object anyway.

SC_redesign/Crop_and_Soil/field/fertilizer_application.py
```diff
@@ -0,0 +1,65 @@
...
+        """This method initializes the Soil object that this module will work with, or create one if none provided.
+
+        Parameters
+        ----------
```
ltp23, line 16: Missing the soil parameter.

ltp23 (approved):
Overall, it looks good. Just a minor comment. Thanks Ed!

SC_redesign/Crop_and_Soil/field/fertilizer_application.py
Isgotkowitz: Good catch

SC_redesign/Crop_and_Soil/field/fertilizer_application.py
Isgotkowitz: You make a good point here - one argument would be a lot easier to understand.

Some time ago, there was only one optional input - `Soil`, and if not provided would create a new `Soil` object to work with. Then some features got added, and lead to the necessity for the field size to be provided when creating new `SoilData` objects, which is why this module (and all others in the `soil/` submodule) have two optional initialization parameters.

If there is an attempt to instantiate a `FertilizerApplication` object with `soil` and `field_size` both being `None`, an error will be raised. But it is raised in the `post_init()` of `SoilData` which seems not ideal.

The idea behind making the `Soil` (or in other modules, `SoilData`) optional was that if a user or researcher was interested in using or studying the behavior of only one module, this design would require them to do as little overhead setup as possible.

We should talk more about what a better input/initialization system would be, but for now I think it should be left like this because it is consistent with the rest of the Crop and Soil module.

PR #540, opened 2023-05-26
Implement `ManureApplication` module

### LINKED ISSUE #433: Implement `ManurePool` class to make operations in `soil/phosphorus_cycling/` better
# Problem
The behavior of manure, as defined in SurPhos, is dependent on two attributes: method of application (applied by machine or grazing animals) and type (source of manure can be cows, or non-cows). In order to fully implement SurPhos, we need to track these different manure types separately (issue #420).

Also, at the time of this writing the interface for applying manure in a field is located in `Crop_and_Soil/soil/phosphorus_cycling/manure_application.py` which does not reflect how this module will actually be used (at the `Soil` or `Field` levels of the Crop and Soil module).

# Current solution
Two separate manure pools are tracked, one for machine-applied cow manure, and one for directly applied cow manure. Each pool is composed of an identical set of attributes (save that they are prefixed with either `machine_` or `grazing_`), and are entirely contained in the `SoilData` module. This has resulted in a lot of duplicate code building up throughout the `phosphorus_cycling` submodule.

# Possible Solutions
1. Implement a base `ManurePool` class that defines all the common attributes and behavior of the manure pools, and then implement 4 different pool classes that inherit from `ManurePool` with the correct constants and behaviors for that pool's type. 
  - Also, move the `ManureApplication` module up from the `phosphorus_cycling` submodule, likely into the `field` or `manager` sub-modules

# Other Concerns
- Manure is also a major source of nitrogen, so any solution to this issue should consider compatibility with `nitrogen_cycling/`. `nitrogen_cycling/` has not been refactored yet as part of the Soil and Crop module redesign; how easily/simply it can be integrated with `phosphorus_cycling` is currently a big x-factor.
- It is possible that this feature will become fairly complex/involved and require a design doc & design review process before implementing. In this case a better course of action may be to make this a "Future issue" and complete V1 without it.

Isgotkowitz:
This PR expands the capabilities for adding manure to a field by moving the location of the `ManureApplication` module and allowing it to add more nutrients to the field.

- `manure_application.py` has been moved from `soil/phosphorus_cycling/` to `field/` and now takes parameters to calculate the amount of nitrogen in manure applications.
  - A missing method (`_determine_water_extractable_inorganic_phosphorus_fraction_by_animal()`) and test have been added.
  - The method `_add_nitrogen_to_soil_layer()` makes it easy to add a specified amount of nitrogen to a specified soil layer.
  - All methods that handle adding manure to the field now add nitrogen as well as phosphorus.
  - Many of the TODOs have been removed, and docstrings updated with the origin/reason why equations are there.
  - All new and changed methods are tested in `test_manure_application.py`.
- `ManureApplication` is no longer a process component of `Soil`.
-  `Field` now has a `ManureApplication` process component.


matthew7838 (approved):
Good PR, great notes and insights were given about your communication with Pete.

ltp23 (approved):
I would say the two methods `_apply_solid_machine_manure` and `_apply_liquid_machine_manure` could use some refactoring, but that may take some time. They seem to have some overlap in functionality. The long parameter lists could be replaced with normal dictionaries or TypedDicts. 

Overall, the code looks really good and well-documented. Good job, Ed!

Isgotkowitz:
> I would say the two methods `_apply_solid_machine_manure` and `_apply_liquid_machine_manure` could use some refactoring, but that may take some time. They seem to have some overlap in functionality. The long parameter lists could be replaced with normal dictionaries or TypedDicts.
> 
> Overall, the code looks really good and well-documented. Good job, Ed!

Thanks Loi. 

#433 calls for `ManurePool` objects to be implemented - those classes should take this behavior of adding manure to fields, so I will add a note in there that the behavior should be added, and the design refactored to eliminate redundancy.

Issue #522, opened 2023-05-22
Implement save_variables function


PooyaHekmati:
Implement save_variables function. This function shares code with dump_variables function. So, create a function with _ prefix and make both save and dump variables call this function. dump_variables only call this function; however, save_variables first filters the pool and then passes the filtered pool to this function. For filtering, first, take a look at https://learnpython.com/blog/filter-dictionary-in-python/ . You will need a function to pass to filter(). Let's start simple: we want save_variables to do this:
1. reads a txt file (the path to this txt is given as an arg to the function). The txt file contains the keys that we want to save. Each line in the txt file contains the exact key (no pattern matching, just use == operator) (one key per line). 
2. Uses filter () to filter the keys to figure out which entries in variables_pool need to be saved.
3. Calls the newly developed private function to handle exclude_info_map and the saves the results.

ew3361zh:
A couple questions:
1. Do we want the _exclude_info_maps() function to be dependent on the setting of the exclude_info_maps flag? So if the user has set exclude_info_maps to False, should this function effectively do nothing?
2. Into what format should the new save_variables function save the results from step 3 above?

PooyaHekmati:
I'd say the caller should make decision to either call _exclude_info_maps or not based on that flag.
The rest if the same as the existing code, so no change in formatting.

ew3361zh:
Handled in PR #541

PR #541, opened 2023-05-26
Create save_variable function in OM

ew3361zh:
This PR seeks to address issue #522 to create a save_variable function in OM that reads a filtered list of requested variables and saves this filtered pool of variables accordingly.

# To accomplish this, the PR:

1. Creates a new private function `_load_txt_file_to_list` in OM to read txt files and add them to a list.
2. Creates a new private function `_filter_variables_pool` to filter the variables_pool using the list of keys provided from the txt file.
3. Creates a new private function `_exclude_info_maps` which makes a copy of the given pool and removes info_maps from it.
5. Creates a new function `save_variables` which reads a text file containing the keys by which the user wants to filter the variables pool and saves the resulting filtered pool to a json file.
6. Calls `save_variables` from main's `execute_simulations_from_files` function.
7. Moved logic to exclude info_maps out of the dump_variables function and replaced it with the call to the new `_exclude_info_maps` function. It still then saves the all variables pool.
8. Adds unit testing for new functions and updates existing unit testing appropriately to cover new functionality.
9. Adds a dummy list of variables file to input folder which is not present in this PR but which was a list of variable names.
Screenshot 2023-05-27 at 2 58 13 PM

Screenshot 2023-05-31 at 9 51 07 AM


ew3361zh:
@PooyaHekmati would you please take a look?

RUFAS/output_manager.py
```diff
@@ -53,6 +53,7 @@ def __init__(self) -> None:
                     "function": self.__init__.__name__,
                 },
             )
+            self.keys_list = []
```
PooyaHekmati: unclear and vague name, this is the list of what keys? what purpose does it serve? why is it needed? what happens if it gets changed as a side effect of another process?

RUFAS/output_manager.py
```diff
@@ -301,36 +302,159 @@ def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
         timestamp = time.strftime(r"%d-%b-%Y_%a_%H-%M-%S", time.localtime())
         return f"{base_name}_{timestamp}.{extension}"
 
+    def _exclude_info_maps(self, pair: dict[str, Any]) -> bool:
```
PooyaHekmati: The function name starts with an action verb, suggesting that it does something. Yet, the return value is bool, suggesting the function makes a decision of some sort. There is an inconsistency here.

RUFAS/output_manager.py
```diff
@@ -301,36 +302,159 @@ def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
...
+            If "info_maps" is found as a value in the pair
+
+        """
+
+        unwanted_value = "info_maps"
+        key, value = pair
+        if key == unwanted_value:
+            return False
+        else:
+            return True
```
```suggestion
    def _exclude_info_maps(self, pool:Dict[str, OutputManager.pool_element_type]) -> Dict[str, OutputManager.pool_element_type]:

        """ Makes a copy of the given pool and removes info_maps from it.

        Returns
        -------
        Dict[str, OutputManager.pool_element_type]
            A copy of the given pool with info_maps removed from it.

        """
  
        pool_copy = pool.copy()
            for key, value in pool_copy.items():
                if isinstance(value, dict) and "info_maps" in value:
                    value.pop("info_maps")
        return pool_copy 
```

RUFAS/output_manager.py
```diff
@@ -301,36 +302,159 @@ def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
...
+        ----------
+            path : str
+                Path of the input file to be read.
+
+        """
+        try:
+            with open(path) as keys_doc:
+                self.keys_list = keys_doc.read().splitlines()
+        except Exception as e:
+            raise e
```
```suggestion
    def _load_txt_file_to_list(self, path: str) -> List[Str]:
        """ Reads a text file into a list.

        Parameters
        ----------
            path : str
                Path of the input file to be read.
                
         Returns
         -------
            TODO
            
         Raises
         -------
         TODO            

        """
        try:
            with open(path) as keys_doc:
                return keys_doc.read().splitlines()
        except Exception as e:
            raise e
```

RUFAS/output_manager.py
```diff
@@ -301,36 +302,159 @@ def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
...
+            bool
+                If the key from pair is found in the list of keys from the input file.
+
+        """
+
+        key, value = pair
+        if key in self.keys_list:
+            return True
+        else:
+            return False
```
```suggestion
    def _variables_pool_inclusion_filter(self, inclusion_key: str) -> bool:
        """returns True is inclusion_key has a match in variables_pool"""

        return True if inclusion_key in self.variables_pool.keys() else False
```

RUFAS/output_manager.py
```diff
@@ -301,36 +302,159 @@ def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
...
+                        for data_pool_key in var_data_list[index].keys():
+                            self.variables_pool[name][var_data_key][index][data_pool_key] = dict(filter(
+                                self._filter_variables_pool, var_data_list[index].items()))
+
+        if exclude_info_maps:
+            for name, variable_data in self.variables_pool.items():
+                self.variables_pool[name] = dict(filter(self._exclude_info_maps, variable_data.items()))
+
+        file_path = os.path.join(path, self._generate_file_name("saved_variables", "json"))
+        self._dict_to_file_json(self.variables_pool, file_path)
```
```suggestion
        inclusion_keys = self._load_txt_file_to_list(keys_file_path)
        filtered_pool = dict(filter(self._variables_pool_inclusion_filter,self.variables_pool))
        final_pool = filtered_pool
        if exclude_info_maps:
            final_pool = self._exlude_info_maps(filtered_pool)
        file_path = os.path.join(path, self._generate_file_name("saved_variables", "json"))
        self._dict_to_file_json(final_pool, file_path)
```

RUFAS/output_manager.py
```diff
@@ -301,36 +302,159 @@ def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
...
         if exclude_info_maps:
-            for key, value in vars_pool.items():
-                if isinstance(value, dict) and "info_maps" in value:
-                    value.pop("info_maps")
+            for name, variable_data in self.variables_pool.items():
+                self.variables_pool[name] = dict(filter(self._exclude_info_maps, variable_data.items()))
 
-        file_path = os.path.join(path, self._generate_file_name("variables", "json"))
+        file_path = os.path.join(path, self._generate_file_name("dumped_variables", "json"))
         self._dict_to_file_json(self.variables_pool, file_path)
```
```suggestion
        file_path = os.path.join(path, self._generate_file_name("all_variables", "json"))
        self._dict_to_file_json(pool, file_path)
```

RUFAS/output_manager.py
```diff
@@ -301,36 +302,159 @@ def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
...
+            path : str
+                Path to the directory where the file will be saved.
+
+            exclude_info_maps : bool
+                Flag for whether or not the user wants to inlcude info_maps data in their results files.
+
         """
-        vars_pool = self.variables_pool.copy()
+
         if exclude_info_maps:
```
```suggestion
        pool=self.variables_pool
        if exclude_info_maps:
```

RUFAS/output_manager.py
```diff
@@ -301,36 +302,159 @@ def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
...
-            for key, value in vars_pool.items():
-                if isinstance(value, dict) and "info_maps" in value:
-                    value.pop("info_maps")
+            for name, variable_data in self.variables_pool.items():
+                self.variables_pool[name] = dict(filter(self._exclude_info_maps, variable_data.items()))
```
```suggestion
            pool =self._exclude_info_maps(self.variables_pool)
```

RUFAS/output_manager.py
```diff
@@ -408,15 +534,29 @@ def dump_variable_names_and_contexts(
...
+            exclude_info_maps : bool
+                Flag for whether or not the user wants to inlcude info_maps data in their results files.
+
         """
         self.dump_variables(path, exclude_info_maps)
         self.dump_variable_names_and_contexts(path, exclude_info_maps)
         self.dump_errors(path)
         self.dump_logs(path)
         self.dump_warnings(path)
+        self.save_variables(path, input_path, exclude_info_maps)
```
PooyaHekmati: Let's not call this here. Instead, move it to execute_simulations_from_files in main.py

PooyaHekmati (changes requested):
Please take a look at the code suggestion.

When updating a significant amount of things that are not the main purpose of the PR (docstrings for example), it is a good practice to do them in a separate PR to make the review easier.

RUFAS/output_manager.py
ew3361zh: What do you think about this renaming and description for the docstring? (would also change every reference to match)

user_input_keys_list : List[str]
        Contains a list of user-input keys used to filter the
        variables_pool to keep only their requested data.

RUFAS/output_manager.py
ew3361zh: The issue I'm seeing here is with self._variables_pool_inclusion_filter being able to access the inclusion_keys returned from self._load_txt_file_to_list. The python filter function doesn't allow sending any additional arguments which was the original reason I added that keys_list attribute to the OM.

The two options I thought of when I was originally working on this were:
1. Call self._load_txt_file_to_list within self._variables_pool_inclusion_filter to check against but that seemed like poor design to have it load and read that file every time self._variables_pool_inclusion_filter was called.
2. Add that keys_list attribute to OM (would need to rename and add docstrings as noted above). We could then still call self._load_txt_file_to_list in self.save_variables where it is now to ensure it's available but then could have that list available in self._variables_pool_inclusion_filter as an attribute of OM.

PooyaHekmati: both of these approaches have design problems. Let's change _variables_pool_inclusion_filter to:
`def _filter_variables_pool(self, inclusion_keys: List[str]):
        """TODO"""
        return {
            key: self.variables_pool[key]
            for key in inclusion_keys
            if key in self.variables_pool.keys()
        }`
and not use filter().

RUFAS/output_manager.py
```diff
@@ -301,18 +301,108 @@ def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
...
+        """ Reads a text file into a list.
+
+        Parameters
+        ----------
+        path : str
+            Path of the input file to be read.
+
+        Returns
+        -------
+        List
```
```suggestion
        List[str]
```

RUFAS/output_manager.py
```diff
@@ -301,18 +301,108 @@ def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
...
+            A list of keys the user has selected to filter the variables pool.
```
PooyaHekmati: This is just a specific use case of this function. The functionality is broader and more general.

ew3361zh:
@PooyaHekmati can you please review again to see if your requested changes have been addressed?

RUFAS/output_manager.py
```diff
@@ -301,18 +301,108 @@ def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
...
+
+        Raises
+        -------
+        Exception
+            If an error occurs while opening or reading the file.
+
+        """
+        try:
+            with open(path) as keys_doc:
```
PooyaHekmati: same as above, this function can be used for different purposes, keys_doc is too specific.

RUFAS/output_manager.py
PooyaHekmati: Let's also make it clear that each line in the files ends up being an element in the list.

RUFAS/output_manager.py
```diff
@@ -301,18 +301,108 @@ def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
...
+                return keys_doc.read().splitlines()
+        except Exception as e:
+            raise e
+
+    def _filter_variables_pool(self, inclusion_keys: List[str]):
```
PooyaHekmati: missing type annotation for the return value

RUFAS/output_manager.py
```diff
@@ -301,18 +301,108 @@ def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
...
+
+        Parameters
+        ----------
+        inclusion_keys : List[str]
+            A list of keys the user has selected to filter the variables pool.
+
+        Returns
+        -------
+        Dict[str, OutputManager.pool_element_type]
+            The variables_pool with only the values paired with the keys
```
PooyaHekmati: I'm not sure if this is correct. It does not return the variables_pool, it makes a new dictionary; i.e., it has no side effect on the vars pool.

RUFAS/output_manager.py
```diff
@@ -301,18 +301,108 @@ def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
...
+
+    def save_variables(self, path: str, keys_file_path: str,
+                       exclude_info_maps: bool = False) -> None:
+        """
+        Reads a text file containing a list of keys and filters the variables pool by those keys.
+        Saves resulting data pool to a json file in the given path to a directory.
+
+        Parameters
+        ----------
+        path : str
```
PooyaHekmati: can we rename this save_path?

RUFAS/output_manager.py
```diff
@@ -301,18 +301,108 @@ def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
...
+            Path of the input file containing the list of keys.
+
+        """
+        inclusion_keys = self._load_txt_file_to_list(keys_file_path)
+        filtered_pool = self._filter_variables_pool(inclusion_keys)
+        final_pool = filtered_pool
+        if exclude_info_maps:
+            final_pool = self._exclude_info_maps(filtered_pool)
+        file_path = os.path.join(path, self._generate_file_name("saved_variables", "json"))
+        self._dict_to_file_json(final_pool, file_path)
```
PooyaHekmati: I know I wrote this code, but now that I'm looking at it, the extra variable is not needed.
```suggestion
        if exclude_info_maps:
            filtered_pool = self._exclude_info_maps(filtered_pool)
        file_path = os.path.join(path, self._generate_file_name("saved_variables", "json"))
        self._dict_to_file_json(filtered_pool, file_path)
```

tests/test_misc.py
```diff
@@ -749,6 +755,162 @@ def test_list_to_file_txt(
...
+    expected_result = {}
+    assert mock_output_manager._exclude_info_maps(pool) == expected_result
+
+    # Test case 2: Pool with info_maps
+    pool = {
+        "key1": {
+            "info_maps": "value1",
+            "other_key": "other_value"
+        },
+        "key2": {
```
PooyaHekmati, line 774: how is it possible to not have info map?

tests/test_misc.py
```diff
@@ -749,6 +755,162 @@ def test_list_to_file_txt(
...
+        "key1": {
+            "other_key": "other_value"
+        },
+        "key2": {
+            "other_key": "other_value"
+        }
+    }
+    assert mock_output_manager._exclude_info_maps(pool) == expected_result
+
+    # Test case 3: Pool without info_maps
```
PooyaHekmati: Same as above

tests/test_misc.py
```diff
@@ -749,6 +755,162 @@ def test_list_to_file_txt(
...
+
+def test_load_txt_file_to_list(
+    mock_output_manager: OutputManager,
+    output_manager_original_method_states: Dict[str, Callable],
+    tmpdir
+) -> None:
+    """Test case for function _load_txt_file_to_list in output_manager.py"""
+    content = "apple\nbanana\ncherry"
+
+    file_path = tmpdir.join("test_file.txt")
```
PooyaHekmati: is my understanding correct that this test case is actually saving a file in tmp path?
If yes, then this is not the right approach because failure to save/open the file can make this test fail while the actual code can work; therefore, getting misleading signals. Instead, the open() function needs to be mocked. Also, since the function can raise an exception, that behavior needs to be tested as well (mock open() such that it raises an exception, then call _load_txt_file_to_list() and ensure it raises the exception.)

tests/test_misc.py
```diff
@@ -749,6 +755,162 @@ def test_list_to_file_txt(
...
+        "_load_txt_file_to_list"
+    ]
+
+
+def test_filter_variables_pool(
+    mock_output_manager: OutputManager,
+    output_manager_original_method_states: Dict[str, Callable]
+) -> None:
+    """Test case for function _filter_variables_pool in output_manager.py"""
+    mock_output_manager.dummy_keys_list = ['apple', 'orange', 'banana']
```
PooyaHekmati: why is this needed?

tests/test_misc.py
```diff
@@ -749,6 +755,162 @@ def test_list_to_file_txt(
...
+    }
+    assert mock_output_manager._filter_variables_pool(inclusion_keys) == expected_result
+
+    # Test case 3: inclusion_keys with non-existing keys
+    inclusion_keys = ["key1", "key4"]
+    mock_output_manager.variables_pool = {
+        "key1": "value1",
+        "key2": "value2",
+        "key3": "value3"
+    }
```
PooyaHekmati: don't repeat yourself

tests/test_misc.py
```diff
@@ -749,6 +755,162 @@ def test_list_to_file_txt(
...
+    mock_output_manager._generate_file_name = MagicMock(return_value="dummy_name")
+    mock_output_manager._dict_to_file_json = MagicMock()
+    mock_output_manager._load_txt_file_to_list = MagicMock()
+
+    mock_output_manager.save_variables("dummy_path", "dummy_input_path", False)
+
+    mock_output_manager._generate_file_name.assert_called_once_with("saved_variables", "json")
+    mock_output_manager._dict_to_file_json.assert_called_once_with(
+        mock_output_manager.variables_pool, os.path.join("dummy_path", "dummy_name")
+    )
```
PooyaHekmati, line 891: save_variables also makes calls to _load_txt_file_to_list and _filter_variables_pool and makes decision to either call _exclude_info_maps or not. These are not being tested.

tests/test_misc.py
ew3361zh: I was just trying to supply different possibilities of potential pools (even those outside the expected). I can delete this section if it seems extraneous.

RUFAS/output_manager.py
JoeWaddell: For my own clarity: should this be explained in the short summary, or better in the Returns section so that the user has a better idea of the list structure in relation to the input? Or both?

main.py
```diff
@@ -64,6 +64,7 @@ def execute_simulations_from_files(
         simulator = SimulationEngine(input_file_path)
         simulator.simulate()
         output_manager.dump_all_pools(r"output", exclude_info_maps)
+        output_manager.save_variables(r"output", r"input/list_of_keys.txt", exclude_info_maps)
```
JoeWaddell: Running this branch fails because `input/list_of_keys.txt` doesn't exist (you noted that it won't be in this PR in the description, but should this method simply save a single JSON with everything if no keys are specified?).

Also, the idea here is that we'll add a new line for each output, specifying a different list of keys each time? Will this be handled differently in future OM PRs, or in the Input Manager PR? Alternatively, could/should this be all done in one JSON, to reduce the times that a user needs to modify the code itself?

Also to check that I'm understanding  the workflow as it is: if we want some general analysis with all keys, but also a pair of smaller analyses, we would have three calls to `output_manager.save_variables` using e.g. `general_analysis_keys.txt`, `methane_analysis_keys.txt`, and a `ration_analysis_keys.txt`
It then saves the three different output JSONs with the name as defined in the `_generate_file_name` method, correct?

JoeWaddell (changes requested):

RUFAS/output_manager.py
ew3361zh: My thought - based on the description of what the short summary should be ("A one-line summary that does not use variable names or the function name") - is that this type of info should be in the returns section.

main.py
ew3361zh: > Running this branch fails because `input/list_of_keys.txt` doesn't exist (you noted that it won't be in this PR in the description, but should this method simply save a single JSON with everything if no keys are specified?).

Maybe I should have better explained - this is the 2nd of 5 parts to the filtering variables functionality of save_variables. I didn't add this because the next couple PRs expand the capabilities of the save_variables function.

I provided the example images so the tester could run the simulation successfully by adding a similar txt file (but with the same name) to the inputs folder.

Not sure if this was a good practice but I didn't want to add files that we would be deleting in the next PR. Sorry for the confusion.

> Also, the idea here is that we'll add a new line for each output, specifying a different list of keys each time? Will this be handled differently in future OM PRs, or in the Input Manager PR? Alternatively, could/should this be all done in one JSON, to reduce the times that a user needs to modify the code itself?

> Also to check that I'm understanding the workflow as it is: if we want some general analysis with all keys, but also a pair of smaller analyses, we would have three calls to `output_manager.save_variables` using e.g. `general_analysis_keys.txt`, `methane_analysis_keys.txt`, and a `ration_analysis_keys.txt` It then saves the three different output JSONs with the name as defined in the `_generate_file_name` method, correct?

Yes, the next few PRs will expand the capabilities of the save_variables function to: 
1. Handle multiple input variable filter files, use them to filter the variables_pool accordingly, and then save each resulting filtered_pool.
2. Differentiate between whether each file wants save_variables to be inclusive or exclusive of the list of variables provided. Inclusive is what we have currently and the exclusive version will save everything in the variables_pool EXCEPT for the variables listed in the file.

If you're interested, you can take a peek at draft PR #544 which is the next step (handling multiple files) and I have a draft of the step after that I should get to Github tonight (both of which were branched off of this branch before today's changes requested from Pooya).

Let me know if you have any other questions and thanks for the great feedback and questions!

tests/test_misc.py
PooyaHekmati: It is a good idea to have more comprehensive test cases. My concern is that someone may read this in the future and assume it is ok to not have info maps. Maybe we can fix it by adding a comment to emphasize that info map is expected to exist, we are just being extra cautious here.

main.py
PooyaHekmati: I think Joe has a point: what is the expected behavior if the user doesn't provide the file containing the list of keys? I would say in that case the save_variables shouldn't do anything because it is possible that the user is only interested in the dumps. However, I wouldn't add an if condition here to check if the file exists or not because in the next PR, there will be a for loop which runs for each file in the directory. If there are no files, then that for loop will not do anything. In other words, the non-existing file is temp issue.

ew3361zh: I think we can do a couple things:

1. Add a temporary file so to prevent crashes for the period where this is merged into master before the next PR.
2. Merge the next PR into this PR before merging this PR into master.

It is a reminder to me though that I need to add a .keep file to each of the directories containing these files.

JoeWaddell: Thanks for the feedback/answers, and I think your two suggestions here sound good. I'll look ahead at your other PRs, as well, to have a better sense of where you're headed with this. Nice work!

ew3361zh: I'll go ahead and add the temp txt file which, even if we decide to merge everything into this PR before merging into master, we can delete later. It seems like a small enough thing not to need to set up an issue for but I know that can get us in trouble so I'll make a quick issue to delete it later as well.

RUFAS/output_manager.py
ew3361zh: Implemented

RUFAS/output_manager.py
ew3361zh: Moved to main.py

RUFAS/output_manager.py
ew3361zh: Updated to "with open(path) as text_file:"

ew3361zh:
@PooyaHekmati I've addressed the changes requested, please take another look and let me know if you have any other thoughts/feedback

JoeWaddell:
I've tested it after the updates and example keys.txt file, and it works well! Nice job.

JoeWaddell (approved):

PooyaHekmati (approved):

PR #543, opened 2023-05-30
Implement Crop Scheduling mechanism #1 - `Event`s

Isgotkowitz:
# Summary
This PR implements pieces of code that will be used for scheduling events that occur in a field, principally crop plantings and harvestings.

# Changes
- `crop_input.py` (and `test_crop_input.py`) have been removed, they will be replaced by the `CropSchedule` module in subsequent PRs.
- `events.py`
  - Contains a base class `Event` that serves as a way to schedule events to happen on a certain year and day during a simulation.
  - Two child classes are present in this file, `PlantingEvent` and `HarvestEvent` which contain additional fields for planting and harvesting crops, respectively.   
  - At some point in the future, this class will be extended for tillage, fertilizer, and manure application events.
- `harvest_operations.py` has had a list of acceptable final harvest operations added, which will be used when ensuring that the final harvest operation on a crop will be a correct one. 
- `crop_schedules_overview.md` provides an overview and examples of how crop scheduling will work. It is not a full blown design doc, but will be useful in helping people understand the mechanisms for crop scheduling.


matthew7838 (approved):
Ed, I think this looks really good. I don't have anything to comment for this one.

SC_redesign/Crop_and_Soil/manager/events.py
```diff
@@ -0,0 +1,95 @@
...
+
+        Returns
+        -------
+        event : Event
+            an Event instance
+        """
+        self.year = year
+        self.day = day
+
+    def __eq__(self, other):
```
ltp23, line 36: Since we implement `__eq__` here, to be safe we would probably need to do the same in all the subclasses and add hashing methods `__hash__` too.

ltp23 (approved):
It looks great to me. Good job, Ed!

SC_redesign/Crop_and_Soil/manager/events.py
Isgotkowitz: Sounds good on `__eq__`. But could you explain a bit more about `__hash__` and why it is necessary here - and by extension what a good way of going about implementing it would be for `Event` and its descendants?

Isgotkowitz: @ltp I came up with the following:
```python
    def __eq__(self, other):
        """Overrides the equality operator for Event objects."""
        correct_type = isinstance(other, Event)
        hash_value = self.__hash__()
        other_hash_value = other.__hash__()
        equal_hash_values = hash_value == other_hash_value
        return correct_type and equal_hash_values

    def __hash__(self):
        """Overrides the hash method for Event objects."""
        str_representation = str(self.year) + str(self.day)
        return hash(str_representation)
```
I think it is alright so I am going to merge to move things along, but if you see any glaring issues with this LMK and I will fix them.

Issue #523, opened 2023-05-22
Expand save_variables function to encompass multiple files


PooyaHekmati:
In issue 552 we made save_variables function to read from a single file. Expand this function such that it can read multiple files and generate an output for each of them.

ew3361zh:
Addressed in PR #544

PR #544, opened 2023-05-31
Expand save variables fun

ew3361zh:
This PR seeks to address issue #523 to expand the save_variable function in OM to read multiple filtered lists of requested variables and saves each filtered pool of variables accordingly.

# To accomplish this, the PR:

1. Creates a new directory in the inputs folder which contains the variable_pool filters example txt files:
Screenshot 2023-06-02 at 10 28 07 AM

2. Creates a new function `_load_input_txt_file_names_to_list` which looks in the aforementioned new directory for all txt files and collects and returns all their names in a list.
3. Updates `save_variables` function to first call `_load_input_txt_file_names_to_list` and then loops through this list and saves each filtered list of variables with its own unique name based on the input file name in the output folder.
4. Updates the save_variables call in main's `execute_simulations_from_files` function to look in the appropriate directory for the variable pool filters.
5. Adds unit testing for new function and updates old unit testing accordingly.


ew3361zh:
@PooyaHekmati I have this as a draft PR until we have #541 merged to master but I think this is ready to be looked at (currently have it comparing to #541 branch because I branched off of there).

RUFAS/output_manager.py
```diff
@@ -317,6 +318,29 @@ def _exclude_info_maps(self, pool: Dict[str, pool_element_type]) -> Dict[str, po
                 value.pop("info_maps")
         return pool_copy
 
+    def _load_txt_file_names_to_list(self, dir_path: str) -> List[str]:
```
JoeWaddell: This is nice: would it make sense to generalize it to use any specified file type, so that it could be used in other modules?

JoeWaddell (approved):
This looks great, runs well, and I played around with some different additional filter files and it works nicely.

If I put in bad names to the variable list, it doesn't throw an error, it just doesn't report anything for that. Should it report "not_a_key: not found in variables pool" at the top of the custom output, or should we just consider this a "user beware" feature?

ew3361zh:
> This looks great, runs well, and I played around with some different additional filter files and it works nicely.
> 
> If I put in bad names to the variable list, it doesn't throw an error, it just doesn't report anything for that. Should it report "not_a_key: not found in variables pool" at the top of the custom output, or should we just consider this a "user beware" feature?

That's an interesting question. A couple PRs ahead of this one I implement a pattern matching on the list from the txt file both for the inclusion and exclusion filters. In that way it will more easily capture (or exclude) a wider array of variables but the more specific they get the fewer (or more if it's an exclusion filter) matches they'll get back (e.g. for inclusion filters, "milk" will get a lot of hits whereas "MilkingParlor" would get fewer - it's not case sensitive btw). 

I wonder if one solution might be to have the list of filters printed at the top of the output file to reiterate what their filters were (which might highlight any spelling errors or omissions they had and weren't thinking about). 

Or if we want to leave it more up to the user to go back and look at the filters they've input if they don't think they've gotten back what they expected because we don't want to clutter the output file with the list that's available to them as an input.

RUFAS/output_manager.py
ew3361zh: I think we could certainly expand its capabilities by passing it a list of file extensions we want to look for in the given directory and then have it retrieve all the matching documents. 

Did you have any example use cases in mind? That might be helpful.

Right now this is set as a private function so not really intended to be used by other modules but I'd be curious to hear @PooyaHekmati thoughts on whether that's an appropriate reworking to do for where this is and what it's doing for the OM.

JoeWaddell: The only other modules that come to mind: A) in support of the input manager , and possibly B) dealing with Feed libraries. However, I think in both cases it might be overkill, so I think in answering your question I've answered my initial question, that no - it probably doesn't need to be generalized, since it's a narrow use case in what we're handling in this project.

PooyaHekmati: Hi Joe, I like your thinking here. For this particular case, the answer is no because other parts of the codebase are not supposed to do file handling. The input manager will need a similar function, for which we will figure it out.

PooyaHekmati: what about
```suggestion
    def _list_txt_file_names_in_dir(self, dir_path: str) -> List[str]:
```
?

RUFAS/output_manager.py
```diff
@@ -317,6 +318,29 @@ def _exclude_info_maps(self, pool: Dict[str, pool_element_type]) -> Dict[str, po
...
+        ----------
+        path : str
+            Path of the directory to be searched.
+
+        Returns
+        -------
+        List
+            A list of txt file names.
+
+        """
```
PooyaHekmati: I think a one-liner is enough for this case.
```suggestion
        """ Returns the list of files in the given path"""
```
It's just a suggestion though. If you don't like the one-liner, please add exception to your docstring.

RUFAS/output_manager.py
```diff
@@ -317,6 +318,29 @@ def _exclude_info_maps(self, pool: Dict[str, pool_element_type]) -> Dict[str, po
...
+        dir_path = Path(dir_path)
```
PooyaHekmati: I have a vague memory of Path() being problematic in the past. If I recall correctly, Loi had this issue. Can you please confirm with him?

RUFAS/output_manager.py
```diff
@@ -317,6 +318,29 @@ def _exclude_info_maps(self, pool: Dict[str, pool_element_type]) -> Dict[str, po
...
+        if not dir_path.is_dir():
+            raise IsADirectoryError("specified path is not a directory")
```
PooyaHekmati: It is a very good practice to raise specific exceptions and avoid raising the generic Exception. However, the Python doc suggests that NotADirectoryError should be raised here.
https://docs.python.org/3/library/exceptions.html#NotADirectoryError

RUFAS/output_manager.py
```diff
@@ -317,6 +318,29 @@ def _exclude_info_maps(self, pool: Dict[str, pool_element_type]) -> Dict[str, po
...
+        txt_files = []
+        for filename in os.listdir(dir_path):
```
```suggestion
        all_files = os.listdir(dir_path)
        for filename in all_files:
```
I'm not sure if python is intelligent enough to do caching here. If it doesn't, having os.listdir() in the for loop can lead to multiple requests to the disk, which is not good. Better safe than sorry, let's make sure it doesn't happen.

RUFAS/output_manager.py
```diff
@@ -372,16 +396,22 @@ def save_variables(self, save_path: str, keys_file_path: str,
         save_path : str
             Path to the directory where the file will be saved.
 
-        keys_file_path : str
-            Path of the input file containing the list of keys.
+        dir_path : str
+            Path of the input file containing the list of keys for inclusion filter.
```
PooyaHekmati: input file?!

RUFAS/output_manager.py
```diff
@@ -317,6 +318,29 @@ def _exclude_info_maps(self, pool: Dict[str, pool_element_type]) -> Dict[str, po
...
+            A list of txt file names.
+
+        """
+        dir_path = Path(dir_path)
+        if not dir_path.is_dir():
+            raise IsADirectoryError("specified path is not a directory")
+        txt_files = []
+        for filename in os.listdir(dir_path):
+            if filename.endswith(".txt"):
+                txt_files.append(filename)
```
```suggestion
                txt_files.append(os.join(dir_path,filename))
```

RUFAS/output_manager.py
```diff
@@ -372,16 +396,22 @@ def save_variables(self, save_path: str, keys_file_path: str,
...
+            Flag for whether or not the user wants to inlcude info_maps data in their results files.
 
         """
-        inclusion_keys = self._load_txt_file_to_list(keys_file_path)
-        filtered_pool = self._filter_variables_pool(inclusion_keys)
-        if exclude_info_maps:
-            filtered_pool = self._exclude_info_maps(filtered_pool)
-        file_path = os.path.join(save_path, self._generate_file_name("saved_variables", "json"))
-        self._dict_to_file_json(filtered_pool, file_path)
+        list_of_input_files = self._load_txt_file_names_to_list(dir_path)
```
PooyaHekmati: can we find a better name for this variable? list_of_input_files is generic, it doesn't convey what kind of input is this.

RUFAS/output_manager.py
```diff
@@ -372,16 +396,22 @@ def save_variables(self, save_path: str, keys_file_path: str,
...
+        for input_file in list_of_input_files:
+            input_path = dir_path + input_file
```
PooyaHekmati: Let's have _load_txt_file_names_to_list handle this.

PooyaHekmati (changes requested):
LGTM overall, some minor comments.

PooyaHekmati (changes requested):
LGTM overall, some minor comments.

RUFAS/output_manager.py
ew3361zh: I like it

RUFAS/output_manager.py
ew3361zh: I'm fine with a one-liner here if you think it's acceptable.

RUFAS/output_manager.py
ew3361zh: I will follow up with Loi.

RUFAS/output_manager.py
ew3361zh: The reason I kept the name and dir_path separate and joined them here was to have access to the file name without the dir_path attached. 

When _generate_file_name() is called, it crashes if the dir_path is part of the file name. 

I could do the reverse of what I have here and remove the dir_path from the filename prior to calling _generate_file_name() but I'm not sure if that's better or worse than how it's set up.

Or if you have thoughts on another way to name the saved files for the call to _generate_file_name(), please let me know.

PooyaHekmati: Makes sense

RUFAS/output_manager.py
```diff
@@ -317,6 +318,19 @@ def _exclude_info_maps(self, pool: Dict[str, pool_element_type]) -> Dict[str, po
...
+        dir_path_check = Path(dir_path)
+        if dir_path_check.is_dir():
+            txt_files = []
+            all_files = os.listdir(dir_path)
+            for filename in all_files:
+                if filename.endswith(".txt"):
+                    txt_files.append(filename)
+            return txt_files
+        else:
+            raise NotADirectoryError("specified path is not a directory")
```
PooyaHekmati: 'NotADirectoryError' already says what's wrong, the message string can be used to guide the user on what is needed.

```suggestion
            raise NotADirectoryError("The specified path must be a directory")
```

RUFAS/output_manager.py
```diff
@@ -372,16 +386,22 @@ def save_variables(self, save_path: str, keys_file_path: str,
         save_path : str
             Path to the directory where the file will be saved.
 
-        keys_file_path : str
-            Path of the input file containing the list of keys.
+        dir_path : str
+            Path of the directory containing the list of files containg the keys for inclusion filter.
```
PooyaHekmati: The directory doesn't contain the 'list' of the files, it contains the files which contain the list of keys

PooyaHekmati (approved):
Please address the comments before merging.

PR #545, opened 2023-05-31
Implement Crop Scheduling mechanism #2 - `CropSchedule`

Isgotkowitz:
This method implements the `CropSchedule` module, which takes in data provided by the user and generates a series of events for planting and harvesting a crop over the run of the simulation.

- `CropSchedule` takes in a name, a reference to the crop that is to be planted, planting years and days, harvesting years, days and operations, a flag for whether or not to harvest optimally, and parameters to dictate the expanded patterns of planting and harvesting.
- All input parameters are validated through the methods `_validate_planting_parameters()`, `_validate_harvest_parameters()`, and `_validate_pattern_parameters()` (implemented in `Schedule`)
- `CropSchedule` has two methods, one that generates a list of all `PlantingEvent`'s and one that generates a list of all `HarvestEvent`'s that occur over the expanded crop schedule.
- Both methods and the initialization of `CropSchedule` are tested in `test_crop_schedule.py`.
- `finish_line_doc.md` is updated to reflect the progress made by this PR as well.

Note: the way this class uses inheritance may not be the best/most efficient way to do so, but will be addressed by #561


SC_redesign/Crop_and_Soil/manager/crop_schedule.py
```diff
@@ -0,0 +1,178 @@
...
+        ValueError
+            If not number of planting years and days are not equal.
+
+        """
+        valid_years = self._validate_years(self.planting_years)
+        if not valid_years:
+            raise ValueError(f"'{self.name}': expected all years to be > 0 and in non-descending order, received "
+                             f"'{self.planting_years}'.")
+
+        valid_days = self._validate_days(self.planting_days)
```
matthew7838: Maybe separate out the error message in the case of <=0 or non-descending order so it will be easier for users to handle the error if it's more specific.

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
```diff
@@ -0,0 +1,178 @@
...
+            If not all harvest days are valid.
+        ValueError
+            If the number of harvest years, days, and operations are not equal.
+        ValueError
+            If the last harvest operation is not a final one, or if any operations before the last are final ones.
+
+        """
+        harvest_years_valid = self._validate_years(self.harvest_years)
+        if not harvest_years_valid:
+            raise ValueError(f"'{self.name}': expected all harvest years to be > 0 and in non-descending order, "
```
matthew7838, line 109: This is pretty minor, just a suggestion not sure if it's needed to be addressed or not. Maybe we can narrow down the years? I'm pretty sure there won't be things planted in year 1 or so.

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
```diff
@@ -0,0 +1,178 @@
...
+        if not harvest_days_valid:
+            raise ValueError(f"'{self.name}': expected all harvest days to be in range [1, 366], received "
+                             f"'{self.harvest_days}'.")
+
+        equal_harvest_parameters = len(self.harvest_years) == len(self.harvest_days) == len(self.harvest_operations)
+        if not equal_harvest_parameters:
+            raise ValueError(f"'{self.name}': expected number of values for harvest years, days, and operations to be "
+                             f"equal, received '{self.harvest_years}' years, '{self.harvest_days}' days, and "
+                             f"'{self.harvest_operations}' operations.")
+
```
matthew7838, line 122: Just indicate the length of each list, it might be a big list and the message could be a mess.

matthew7838 (approved):
Ed seems like a good PR, notes really helped me understand some functions. Just a few suggestions for this PR.

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
```diff
@@ -0,0 +1,178 @@
...
+        Parameters
+        ----------
+        name : str
+            Reference to the name of this crop schedule that will be used to distinguish this schedule from others.
+        crop_reference : str
+            Reference to name of the crop that will be used to identify the correct crop specifications.
+        planting_years : List[int]
+            Year(s) in which crop is planted.
+        planting_days : List[int]
+            Day(s) on which crop is planted.
```
ew3361zh: I'm assuming these are Julian Days, is that correct? Worth specifying? Same for harvest_days.

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
```diff
@@ -0,0 +1,178 @@
...
+        ValueError
+            If not all planting days are valid.
+        ValueError
+            If not number of planting years and days are not equal.
+
+        """
+        valid_years = self._validate_years(self.planting_years)
+        if not valid_years:
+            raise ValueError(f"'{self.name}': expected all years to be > 0 and in non-descending order, received "
+                             f"'{self.planting_years}'.")
```
ew3361zh, line 80: Do you need to do a check for order if this message is saying the order might be one of the reasons a ValueError was raised?

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
```diff
@@ -0,0 +1,178 @@
...
+
+        valid_days = self._validate_days(self.planting_days)
+        if not valid_days:
+            raise ValueError(f"'{self.name}': expected all planting days to be in range [1, 366], received "
+                             f"'{self.planting_days}'.")
```
ew3361zh, line 85: Need a check on whether planting days were within the specified range?

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
```diff
@@ -0,0 +1,178 @@
...
+
+        if len(self.planting_years) != len(self.planting_days):
+            raise ValueError(f"'{self.name}': expected number of planting years and days to be the same, received "
+                             f"'{self.planting_years}' years and '{self.planting_days}' days.")
```
ew3361zh, line 89: Why do the planting_years and planting_days lists need to be the same length?

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
ew3361zh, line 109: Same question as for validating planting years above

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
```diff
@@ -0,0 +1,178 @@
...
+
+        """
+        harvest_years_valid = self._validate_years(self.harvest_years)
+        if not harvest_years_valid:
+            raise ValueError(f"'{self.name}': expected all harvest years to be > 0 and in non-descending order, "
+                             f"received '{self.harvest_years}'.")
+
+        harvest_days_valid = self._validate_days(self.harvest_days)
+        if not harvest_days_valid:
+            raise ValueError(f"'{self.name}': expected all harvest days to be in range [1, 366], received "
```
ew3361zh, line 114: Same question as for valitdating planting days as above

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
```diff
@@ -0,0 +1,178 @@
...
+                             f"'{self.harvest_days}'.")
+
+        equal_harvest_parameters = len(self.harvest_years) == len(self.harvest_days) == len(self.harvest_operations)
+        if not equal_harvest_parameters:
```
ew3361zh, line 118: Same question on all of these lists needing to be equal in length as for planting_days and planting_years above

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
```diff
@@ -0,0 +1,178 @@
...
+            List of all planting events that will happen for this crop schedule.
+
+        """
+        all_planting_years = self._repeat_pattern(self.planting_years, self.pattern_skip, self.pattern_repeat)
+        all_planting_days = self.planting_days * (self.pattern_repeat + 1)
+        all_planting_dates = list(zip(all_planting_years, all_planting_days))
+
+        planting_events = []
+        for date in all_planting_dates:
+            new_planting_event = PlantingEvent(self.crop_reference, date[0], date[1], self.heat_scheduled)
```
ew3361zh: What do you think about creating variables for the 0 and 1 for date[0], date[1] to understand what those dates mean in the planting event?

ew3361zh (commented):
Really nice work. This looks good overall! Test suite is solid. 

I have a few questions but nothing that feels too structural so I'm happy to approve once the comments are addressed. Maybe once they are just let me know and I can take another look.

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
Isgotkowitz: I added in keyword arguments in this initialization, do you think that makes the code clear enough?

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
Isgotkowitz: Yes, will add that

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
Isgotkowitz: `_validate_years()` checks for non-descending order and that all years are > 0.

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
Isgotkowitz: Do you mean it still needs a check or is the checking necessary?

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
Isgotkowitz: Because if they were not equal, that would indicate an attempt to plant (or any other operation) without knowing all the parameters of that planting.

Example: planting years = [1991, 1992, 1993], planting days = [120, 120].
This tells us that the user wants plantings to occur in 3 different years. And we know that for 1991 and 1992, the plantings should occur on day 120 of that year. But there is not a corresponding day for 1993, so we have no idea what day the user would want us to plant on in that year.

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
Isgotkowitz: True it is definitely messy, but all these lists have potentially been expanded since they were passed to `init`. So by printing the full list, the person checking the error message can see how each list they passed was expanded, which would be more helpful to them than just printing the length of each list, in case they incorrectly understood the expansion mechanics.

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
ew3361zh: nm, don't know why I didn't look at the _validate_days method. This looks great as is

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
ew3361zh: Ah, gotcha. I thought that the planting_days list was the Julian days within a single planting year - for example if there are multiple plantings of the same crop in the same field within the same year. Maybe this isn't a common occurrence the model needs to account for but it was something we built into the Data Collection App so maybe will need to come back to this at some point to reconcile the two.

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
Isgotkowitz: I think we should accept all cases because how the time works is fairly arbitrary. By having the acceptable range be so large, it is possible to work with the inputs we already have (with years that range from ~1990 to present) and any new input sets that get created but may just have all years start at 1 (which is the way I personally would create my input sets)

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
Isgotkowitz: I think that is a good idea, but should be implemented in #561 because changing `_validate_years()` would mean changing the base class, and subsequently all the other child classes.

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
ew3361zh: I think I might do it here as well for clarity but it seems pretty minor and probably just a personal preference.

ew3361zh (approved):

SC_redesign/Crop_and_Soil/manager/crop_schedule.py
Isgotkowitz: I see. It is still possible to specify multiple plantings of the same crop within the same year, it just has to be specified a little more verbosely.

Issue #524, opened 2023-05-22
Expand save_variables function to handle key exclusion


PooyaHekmati:
So far, the save_variables function saves the keys which have a match in the txt files. Here, we want to add new set of files. Everything in the pool ends up in the output except for the keys that are written in these files.

ew3361zh:
Addressed in #546

PR #546, opened 2023-06-01
Save_variables fun key exclusion

ew3361zh:
This PR seeks to address issue #524 to expand the save_variable function in OM to handle exclusionary filter lists of requested variables in addition to the previous inclusionary filter lists.

# To accomplish this, the PR:
1. Updates `_filter_variables_pool` to filter by either exclusion and inclusion by checking the value of the first filter key in a given filter document and acting appropriately.
2. Adds an example exclusion filter file to the output_filters directory.
3. Updates unit testing across the board.


PooyaHekmati (changes requested):
Hi Niko, I requested some changes in Slack but didn't update them here. I'm requesting changes here so you can re-request review when you are done adding those.

RUFAS/output_manager.py
```diff
@@ -373,7 +373,12 @@ def _filter_variables_pool(self, inclusion_keys: List[str]) -> Dict[str, pool_el
             from the inclusion_keys list remaining from the variables_pool.
 
         """
-        return {key: self.variables_pool[key] for key in inclusion_keys if key in self.variables_pool.keys()}
+        exclude_keyword_location = 0
+        exclude_keyword = "exclude"
```
PooyaHekmati, line 383: nice!

RUFAS/output_manager.py
```diff
@@ -356,14 +356,14 @@ def _load_txt_file_to_list(self, path: str) -> List[str]:
         except Exception as e:
             raise e
 
-    def _filter_variables_pool(self, inclusion_keys: List[str]) -> Dict[str, pool_element_type]:
+    def _filter_variables_pool(self, filter_keys: List[str]) -> Dict[str, pool_element_type]:
         """
-        Takes the list of keys the user wants in their final data pool,
+        Takes the list of keys the user wants to their final data pool either by inclusion or exclusion,
```
PooyaHekmati: This sentence doesn't sound right. 
Can we rephrase it to something like: returns a filtered variables pool based on either inclusion or exclusion?

Also, the fact that the first element of the list is important is not mentioned in the docstring.

PooyaHekmati (approved):
[LGTM](https://lgtm.lol/i/758)

JoeWaddell (approved):
This looks great! Some clever work in here, and I've tested it a number of ways and it works nicely. Well done

Issue #525, opened 2023-05-22
Implement pattern matching for save_variables


PooyaHekmati:
save_variables reads the files which contain the exact matches with keys in the pools and uses == operator. Here, we want to change the content of these file with regular expressions and use regex instead of ==.

ew3361zh:
Addressed with PR #549

PR #549, opened 2023-06-01
Filter_variables_pool Pattern Matching

ew3361zh:
This PR seeks to address issue #525 to implement pattern matching by variables read from the filter keys text files in the `_filter_variables_pool` function in OM. 

# To accomplish this, the PR:
1. Adapts the existing dictionary comprehension filters in the `_filter_variables_pool` to filter by partial matches and ignores case.
2. Updates testing for `_filter_variables_pool` to test new functionality.


tests/test_misc.py
```diff
@@ -876,26 +886,48 @@ def test_filter_variables_pool(
...
     }
+    expected_result_exclude = {
+        "key3": "value3"
+    }
+
     assert mock_output_manager._filter_variables_pool(filter_keys) == expected_result
     assert mock_output_manager._filter_variables_pool(exclude_filter_keys) == expected_result_exclude
 
-    # Restore original method and variables_pool
+    # Test case 6: filter_keys with partial matches
```
JoeWaddell: Nice test, and brings to mind that we'll need to be careful about writing variable names that are partial matches for other variable names. We can of course simply *exclude* the unwanted names, as long as those exclusions aren't partial matches for _another_ name.

JoeWaddell (approved):
Looks good to me

tests/test_misc.py
ew3361zh: Agreed. I think this may influence how we name variables added to the variables pools as well. I will be adding notes on all these changes to the OM wiki page that will provide an overview with some examples to highlight how it works and hopefully show people how the naming can influence the filtering.

PooyaHekmati (changes requested):
The implemented logic works by finding substrings in keys. Issue #525 explicitly asks for pattern matching using regex, which is different from checking if a substring exists in a key. For instance, `LifeCycleManager.daily_update.life_cycle_daily_herd_update` is a match for pattern `^LifeCycleManager.*update$` while `Cow.milking_update.milk_data_at_milk_update` is not a match for this pattern. This also addresses @JoeWaddell 's concern/question.

- Python's Regex doc can be found [here](https://docs.python.org/3/library/re.html)
- A sample Regex tutorial can be found [here](https://www.geeksforgeeks.org/regular-expression-python-examples-set-1)
- [Regex Cheatsheet](https://www.debuggex.com/cheatsheet/regex/python)

RUFAS/output_manager.py
```diff
@@ -372,19 +373,22 @@ def _filter_variables_pool(self, filter_keys: List[str]) -> Dict[str, pool_eleme
...
         variable defined in this function, it will treat the rest of the filter list as exclusionary
         and filter the variables_pool accordingly. Otherwise, it will treat the list of filters
         as inclusionary.
 
         """
         exclude_keyword_location = 0
         exclude_keyword = "exclude"
-        if filter_keys and filter_keys[exclude_keyword_location] == exclude_keyword:
-            return {key: self.variables_pool[key] for key in self.variables_pool.keys() if key not in filter_keys}
+        filter_by_exclusion = filter_patterns and filter_patterns[exclude_keyword_location] == exclude_keyword
```
PooyaHekmati, line 385: In the next PR, if filter_patterns is empty, we should log it. Also, if it only contains 'exclude' we should log it. Generally speaking, we should log how many filters are provided.

RUFAS/output_manager.py
```diff
@@ -372,19 +373,22 @@ def _filter_variables_pool(self, filter_keys: List[str]) -> Dict[str, pool_eleme
...
+        if filter_by_exclusion:
+            return {key: self.variables_pool[key] for key in self.variables_pool.keys() if not
+                    any(re.match(pattern, key) for pattern in filter_patterns)}
         else:
-            return {key: self.variables_pool[key] for key in filter_keys if key in self.variables_pool.keys()}
+            return {key: self.variables_pool[key] for key in self.variables_pool.keys() if
+                    any(re.match(pattern, key) for pattern in filter_patterns)}
```
PooyaHekmati, line 391: In the next PR, where you are adding logs, you can create a variable and assign these dics to it (instead of directly returning them). This way you log the number of keys that are matched.

tests/test_misc.py
```diff
@@ -843,59 +843,325 @@ def test_list_txt_file_names_in_dir(
...
+    """Test case for pattern pool using regex patterns with
+    function _filter_variables_pool in output_manager.py"""
+    mock_output_manager.variables_pool = {
+        "DummyClass1.dummy_fun1.dummy_var1": "value1",
+        "DummyClass1.dummy_fun1.dummy_var2": "value2",  # same class as prev, same fun, different var
+        "DummyClass2.dummy_fun2.dummy_var3": "value3",  # new class, new fun, new var
+        "DummyClass2.dummy_fun3.dummy_var4": "value4",  # same class as prev, new fun, new var
+        "DummyClass2.dummy_fun4.dummy_var4": "value5",  # same class as prev, new fun, same var
+        "DummyClass3.dummy_fun4.dummy_var2": "value6",  # new class, new fun, same var name as 2nd entry
+        "DummyClass4.dummy_fun2.dummy_var5": "value7"   # new class, same fun name as 3rd entry, new var
```
PooyaHekmati, line 1054: I like this

tests/test_misc.py
```diff
@@ -843,59 +843,325 @@ def test_list_txt_file_names_in_dir(
...
+    }
```
PooyaHekmati, line 1116: We don't need to hold this PR for this issue, but this is duplicated code. Also, I see similar structures. This can be greatly improved using fixtures and data providers.

PooyaHekmati (approved):

PR #555, opened 2023-06-02
Implement Methane Mitigation Methods 

tomhuhh:
This PR is to implement methane mitigation methods such as 3-NOP according to the design docs done by Dr. Edward Garcia.


Ed731551 (commented):
Implementation of methane yield equations for both feed additives (3-NOP and monensin) appear to be correct from my side. However, I am not sure if unit tests for dry cows should be kept as CH4 mitigation effect from feed additives is not longer considered for this animal category (see lines 49 to 54). Thanks.

tomhuhh:
@Ed731551 I double checked the test file for dry cows and don't think we currently have any unit tests for methane mitigation in dry cows. Thanks!

JoeWaddell:
I think you'll want a check here to circumvent the calculations if the amount is set to 0.0, because currently if I select mitigation methods but set the amount to 0.0, this is the result: decreased or *increased* methane. It makes sense, since the mitigation equations don't zero out in the case that the additive is 0.

image

I'd have the first `if` block in `methane_mitigation` be for the case that the amount is 0.0 or the method is `None`, to avoid this particular issue. Something like `if methane_mitigation_method == "None" or methane_mitigation_additive_amount == 0.0:` might do the trick.

RUFAS/routines/animal/manure/lactating_cow_manure_excretion.py
```diff
@@ -15,6 +15,50 @@
...
+    EE_concentration : float
+        Concentration of ether extract (EE) in the ration.
+    starch_concentration : float
+        Concentration of starch in the ration.
+    CP_concentration : float
+        Concentration of crude protein (CP) in the ration.
+    methane_mitigation_method: str 
+        Methane mitigation method used to reduce enteric methane emissions, including 3-NOP, monensin, essential oils, and seaweed. 
+    methane_mitigation_additive_amount: float 
+        The amount of methane mitigation feed additive that is added, g/kg dry matter intake.
```
JoeWaddell: Should this be mg or g?

RUFAS/routines/animal/manure/lactating_cow_manure_excretion.py
```diff
@@ -15,6 +15,50 @@
...
+    NDF_concentration : float
+        Concentration of neutral detergent fiber (NDF) in the ration.
+    EE_concentration : float
+        Concentration of ether extract (EE) in the ration.
+    starch_concentration : float
+        Concentration of starch in the ration.
+    CP_concentration : float
+        Concentration of crude protein (CP) in the ration.
+    methane_mitigation_method: str 
+        Methane mitigation method used to reduce enteric methane emissions, including 3-NOP, monensin, essential oils, and seaweed.
```
JoeWaddell: Capitalization and quotes to make sure the user is clear these are the actual options.
```suggestion
        Methane mitigation method used to reduce enteric methane emissions, including '3-NOP', 'Monensin', 'Essential Oils', and 'Seaweed'. 
```

tomhuhh:
@JoeWaddell Thanks Joe! Do you think instead we should set bounds for the amount to be supplemented for each feed additive? Thanks!

For example, in the plot below shows the effect of 3-NOP with an assumption of 36% NDF, 4% EE and 19% starch. X axis is the amount of 3-NOP in mg/g and y-axis shows the mitigation effect in % CH4 yield.  

Screenshot 2023-06-15 at 4 03 51 PM

tomhuhh:
Hello @JoeWaddell ! I have addressed the issues of upper and lower limits for feed additives you brought up based on the values @Ed731551 provided. Thanks for your feedback! Please let me know if you have any other feedback or comments.

tomhuhh:
Please see some preliminary tests with this function below. 

CONTROL - No feed additive
Screenshot 2023-07-01 at 10 53 35 AM

MONENSIN 14 mg/kg DMI - expected no change in enteric methane 
Screenshot 2023-07-01 at 10 53 35 AM

MONENSIN 24 mg/kg DMI - expected decrease in enteric methane emissions 
Screenshot 2023-07-01 at 10 54 32 AM

Observation and explanation: increased methane emission due to low CP% (~10%) 
TODO: Implement a reasonable default diet 

3-NOP - 110 mg/kg DMI - expected no change in enteric methane 
Screenshot 2023-07-01 at 10 53 35 AM

3-NOP - 80 mg/kg DMI - expected decrease in enteric methane emissions
Screenshot 2023-07-01 at 10 54 43 AM

Conclusion: The methane mitigation function is working. 
Comment: the current run time is 47 to 49 seconds, which is not different from the run time before this function is implemented. 

Thanks!

tomhuhh:
A new issue #619 is created to address the issue mentioned above.

tomhuhh:
Monensin works now after a condition in CP is added. A reduction in methane emission when 24 mg/kg DMI is applied is shown below. 

image

RUFAS/routines/animal/manure/lactating_cow_manure_excretion.py
```diff
@@ -15,6 +16,60 @@
...
+    methane_mitigation_additive_amount: float 
+        The amount of methane mitigation feed additive that is added, mg/kg dry matter intake (DMI). The recommended dose for 3-NOP is between 40 and 100 mg/kg DMI, while that for monensin is between 16 and 36 mg/kg DMI. 
+
+    Returns
+    -------
+    float 
+        Reduction in methane yield (methane production/DMI), %.   
+    """
+
+    methane_yield_reduction = 0.0
```
JoeWaddell, line 48: My only concern here is how to handle these lower and upper bounds. 

My inclination is to have them report warnings if the user selects an option beyond those bounds, instead of returning the equivalent of no treatment. Then, the equations can perform as predicted - even in cases of low treatment application amounts _increasing_ methane output, because that's what's expected from Edward's work. Even though the very low dose is not recommended, I think it should be allowed since as Edward presented, efficacy of low Monensin doses are related to protein intake, so I think it's reasonable that we may find users attempting peculiar combinations of input variables.

This is a big design decision though, so it needs some careful thought. I tend to think it wise to allow the user to 'break' assumptions, as long as it's well documented in the code/wiki/design docs that there are settings that there are realistic bounds on values they _should_ enter.

RUFAS/routines/animal/manure/lactating_cow_manure_excretion.py
```diff
@@ -15,6 +16,60 @@
...
+    NOP_upper_bound = 100
+    Monensin_lower_bound = 16
+    Monensin_upper_bound = 36
+    Monensin_CP_lower_bound = 15
+    Monensin_CP_upper_bound = 19
+
+    if methane_mitigation_method == "3-NOP" and NOP_lower_bound <= methane_mitigation_additive_amount <= NOP_upper_bound:
+        methane_yield_reduction = -30.8 - 0.226 * (methane_mitigation_additive_amount - 70.5) + 0.906 * (
+            NDF_concentration - 32.9) + 3.871 * (EE_concentration - 4.2) - 0.337 * (starch_concentration - 21.1)
+    elif methane_mitigation_method == "Monensin" and Monensin_lower_bound < methane_mitigation_additive_amount <= Monensin_upper_bound:
```
JoeWaddell: should be `<=` to follow others, correct?

tomhuhh: According to the design doc, "addition of monensin to the diet around <= 16 mg/kg DMI may seriously compromise abatement effect and even increase CH4 yield emissions". So I did not include the "=" condition for acceptable levels of monensin. Thanks!

RUFAS/routines/animal/pen.py
```diff
@@ -762,7 +773,7 @@ def subset_class_feeds(self, feed):
     # =========================================================================
     # Manure-related methods
     # ----------------------
-    def calc_total_manure(self, feed, methane_model: str) -> None:
+    def calc_total_manure(self, feed, methane_model: str, methane_mitigation_method: str, methane_mitigation_additive_amount: float) -> None:
```
JoeWaddell, line 693: These new parameters need to be added to the parameter list

tests/test_manure_module/test_animal_manure_excretions/test_lactating_cow_manure_excretion.py
```diff
@@ -7,85 +7,91 @@
...
+    assert actual_reduction == pytest.approx(expected_reduction, abs=1e-2)
+
+
+@pytest.mark.parametrize(
+    "methane_model, methane_mitigation_method",
+    [
+        ("Mutian", "3-NOP"),
+        ("Mills", "Monensin"),
+        ("IPCC", "Essential Oils"),
+        ("IPCC", "Seaweed")
```
JoeWaddell: I'd add in an test case for a `methane_mitigation_method` of "None"

RUFAS/routines/animal/animal_module_constants.py
```diff
@@ -24,4 +24,15 @@ class AnimalModuleConstants:
     DEFAULT_MANURE_STORAGE = 'slurry storage outdoor'
 
     DMI_CONSTRAINT_PERCENT = 0.20
+
+    MONENSIN_CP_LOWER_BOUND = 15
+    '''The lower bound of dietary CP(Crude Protein) level (%DM) that should be considered when using the CP equation to caculate the mitigation effects of monensin '''
```
JoeWaddell: These inline docstrings are great! Makes it much clearer to folks elsewhere in the code what each is doing.

JoeWaddell (approved):
Looks good to me: just left comments that it needs some addition to a docstring and maybe expanded tests.

RUFAS/routines/animal/pen.py
tomhuhh: Thanks Joe! What do you mean by the parameter list? If you meant the entries in the input manager/data collection app, these have been added by @allisterakun.

tests/test_manure_module/test_animal_manure_excretions/test_lactating_cow_manure_excretion.py
tomhuhh: Thanks for catching that Joe! I have it updated now.

RUFAS/routines/animal/animal_module_constants.py
Ed731551: Yes, I do agree with Joe's comment.

Ed731551 (approved):

ltp23 (approved):

PR #556, opened 2023-06-05
Base schedule implementation for all other schedules

Isgotkowitz:
This PR introduces `Schedule`, a base class from which the more specific schedule classes will inherit from.

- `Schedule` takes 5 parameters which every child class will need.
- Contains `_validate_pattern_parameters()` for validating the skip repeat pattern parameters, because these are two values which are consistent enough across different schedules that I deemed it ok to use the same error message for different classes.
- Contains `_validate_days()` and `_validate_years()`, which can be used by child classes because the requirements for years and days do not change for different types of schedules.
  - Thanks to @ltp23 for the excellent one-liners in these methods.
- Contains `_elongate_list()`, which provides a common way to elongate lists to a specified length, and will save lots of repeated code in child classes. 
- Contains `_repeat_pattern()` for repeating patterns based on user input, which will be used to create full schedules of events for simulation runs.
- All methods are tested in `test_schedule.py`.


SC_redesign/Crop_and_Soil/manager/schedule.py
```diff
@@ -0,0 +1,171 @@
...
+        >>> repeat_pattern([2, 3, 7], 3, 2)
+        [2, 3, 7, 11, 12, 16, 20, 21, 24]
+
+        """
+        differences = [skip + 1]
+        in_pattern_differences = range(1, len(pattern[1:]) + 1)
+        for difference in in_pattern_differences:
+            differences.append(pattern[difference] - pattern[difference - 1])
+
+        full_pattern = deepcopy(pattern)
```
ew3361zh: I think copy() would also work here (I tried and it was still passing all tests) but I'm not sure of the full implications of what types of schedules might get fed to this function and maybe this accounts for future flexibility.

ew3361zh (approved):
Looks great! Testing looks solid, really nice work!

SC_redesign/Crop_and_Soil/manager/schedule.py
```diff
@@ -0,0 +1,171 @@
...
+        -------
+        bool
+            True if all days are valid.
+
+        Notes
+        -----
+        A day is 'valid' if it in the range [1, 366].
+
+        """
+        return all(0 < day <= 366 for day in days)
```
matthew7838: I suggest having a validation on leap years also if you check for 366 days.

SC_redesign/Crop_and_Soil/manager/schedule.py
```diff
@@ -0,0 +1,171 @@
...
+        A list of years is valid if every year is > 0, and the list of years does not descend at all.
+
+        """
+        return all(0 < years[index] <= years[index + 1] for index in range(0, len(years) - 1))
+
+    @staticmethod
+    def _repeat_pattern(pattern: List[int], skip: int = 0, repeat: int = 0) -> List[int]:
+        """
+        Takes a pattern of numbers and repeats it a specified number of times, skipping over specified gaps between
+        repetitions.
```
matthew7838:
```suggestion
        """
        Takes a pattern of differences between the list numbers and repeats it a specified number of times, skipping over specified gaps between
        repetitions.
```
or something similar to give more details about the pattern. Also, I'm not sure if this is redundant, but maybe maybe add a note to indicated if the skip is inclusive or not.

matthew7838 (approved):
Looks good, one suggestion is to improve the range of validation, and another about docstrings. Testing coverage also looks good. Great work.

SC_redesign/Crop_and_Soil/manager/schedule.py
Isgotkowitz: Good suggestion

SC_redesign/Crop_and_Soil/manager/schedule.py
Isgotkowitz: Good call.

PR #558, opened 2023-06-06
Soil amendment schedules 2: `ManureSchedule` and `ManureEvent`

Isgotkowitz:
This PR contains a scheduler for manure application events, and an event class specifically for manure applications.

- `events.py` now contains a `ManureEvent` subclass which defines when and how much manure gets applied in a single application.
- `ManureSchedule` contains a method for initializing schedules, and `_validate_manure_parameters()` and `_determine_if_all_non_negative_values()` are called within it to validate the fields of the schedule.
- `generate_manure_events()` takes the fields of the schedule and turns it into a list of all `ManureEvent`s that will happen as defined by the `ManureSchedule`

Note: the way this class uses inheritance may not be the best/most efficient way to do so, but will be addressed by https://github.com/RuminantFarmSystems/MASM/issues/561


SC_redesign/Crop_and_Soil/manager/manure_schedule.py
```diff
@@ -0,0 +1,181 @@
...
+    def __init__(self, name: str, years: [List], days: [List], nitrogen_masses: List[float],
+                 phosphorus_masses: List[float], field_coverages: List[float], application_depths: List[float] = None,
+                 surface_remainder_fractions: List[float] = None, pattern_skip: int = 0, pattern_repeat: int = 0):
+        """
+        Creates and validates a manure application schedule.
+
+        Parameters
+        ----------
+        name : str
+            Name of this tillage schedule.
```
Isgotkowitz: Fix

SC_redesign/Crop_and_Soil/manager/manure_schedule.py
```diff
@@ -0,0 +1,181 @@
...
+        if not valid_coverage_fractions:
+            raise ValueError(error_header + f"expected all field coverage fractions to be in the range [0.0, 1.0], "
+                                            f"received '{self.field_coverages}'.")
+
+        valid_depths = all(depth >= 0.0 for depth in self.application_depths)
+        if not valid_depths:
+            raise ValueError(error_header + f"expected all manure application depths to be >= 0, received "
+                                            f"'{self.application_depths}'.")
+
+        valid_surface_fractions = all(0.0 <= fraction <= 1.0 for fraction in self.surface_remainder_fractions)
```
Isgotkowitz, line 124: Change to use non-negative method

Isgotkowitz: Actually nope

SC_redesign/Crop_and_Soil/manager/events.py
```diff
@@ -93,3 +93,36 @@ def __init__(self, crop_reference: str, year: int = 1, day: int = 240, operation
...
+            Day in which this manure application occurs.
+        nitrogen_mass : float
+            Minimum mass of nitrogen that should be contained in this manure application (kg)
+        phosphorus_mass : float
+            Minimum mass of phosphorus that should be contained in this manure application (kg)
+        field_coverage : float
+            Fraction of the field covered by this manure application (unitless)
+        application_depth : float
+            Depth that manure is injected into the soil at (mm)
+        surface_remainder_fraction
```
JoeWaddell: missing typing here (though it's in the arg list)

SC_redesign/Crop_and_Soil/manager/manure_schedule.py
```diff
@@ -0,0 +1,183 @@
...
+This module contains `ManureSchedule`, a `Schedule` child class that defines when and how much manure will be applied to
+a field.
+"""
+
+
+class ManureSchedule(Schedule):
+
+    def __init__(self, name: str, years: [List], days: [List], nitrogen_masses: List[float],
+                 phosphorus_masses: List[float], field_coverages: List[float], application_depths: List[float] = None,
+                 surface_remainder_fractions: List[float] = None, pattern_skip: int = 0, pattern_repeat: int = 0):
```
JoeWaddell: So the way I did this typing in previous branches was: `calving_interval: Optional[int] = None`

There was some discussion about how to handle Optional and setting defaults in the hints, and this is what we landed on a few months back. It was in the NASEM updates branch, and is now in master, so if you search for what I quoted above you'll find it. Word of warning, that whole file needs to be refactored, since it's a series of classless functions.

SC_redesign/Crop_and_Soil/manager/manure_schedule.py
```diff
@@ -0,0 +1,183 @@
...
+        manure_application_events = []
+        for event in all_manure_application_events:
+            new_event = ManureEvent(year=event[0], day=event[1], nitrogen_mass=event[2], phosphorus_mass=event[3],
+                                    field_coverage=event[4], application_depth=event[5],
+                                    surface_remainder_fraction=event[6])
+            manure_application_events.append(new_event)
+        return manure_application_events
+
+    @staticmethod
+    def _determine_if_all_non_negative_values(values: List[Any]) -> bool:
```
JoeWaddell: Should this be something like: `List[Union[int, float]]` ? If you pass a list that includes str or other objects it will result in a TypeError. Probably should be obvious to anyone trying to use the method -  which is private anyway - but might give the impression that it tries to convert objects in a list to ints/floats, or otherwise excludes a list that includes multiple types.

SC_redesign/Crop_and_Soil/tests_SC/manager_tests/test_manure_schedule.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+@pytest.mark.parametrize("values,expected", [
+    ([1, 3, 4], True),
+    ([0.0, 1.2, 3.8], True),
+    ([], True),
+    ([-0.1, 0.1], False),
+    ([-2, -4], False)
+])
+def test_determine_if_all_non_negative_values(values: List[Any], expected: bool) -> None:
```
JoeWaddell: same as previous re: `List[Union[int, float]]`

JoeWaddell (approved):
Looks good! Just a few small comments/questions about typing.

SC_redesign/Crop_and_Soil/manager/manure_schedule.py
Isgotkowitz: Good suggestion. I like `Any` cause its so convenient, but you're right - it should be limited to only accept valid types.

SC_redesign/Crop_and_Soil/manager/manure_schedule.py
Isgotkowitz: Good catch. We've been doing that in our modules, slipped through the cracks here.

SC_redesign/Crop_and_Soil/manager/manure_schedule.py
```diff
@@ -0,0 +1,185 @@
...
+                                            f"'{self.nitrogen_masses}'.")
+
+        valid_phosphorus_masses = self._determine_if_all_non_negative_values(self.phosphorus_masses)
+        if not valid_phosphorus_masses:
+            raise ValueError(error_header + f"expected all phosphorus masses to be >= 0, received "
+                                            f"'{self.phosphorus_masses}'.")
+
+        valid_coverage_fractions = all(0.0 <= fraction <= 1.0 for fraction in self.field_coverages)
+        if not valid_coverage_fractions:
+            raise ValueError(error_header + f"expected all field coverage fractions to be in the range [0.0, 1.0], "
```
matthew7838: Just a suggestion, not sure if it will be too much work, but I think it would be better to keep executing and return everything that throws an error. Maybe like adding errors to a list and returning or ExceptionGroup, here's what I found: https://stackoverflow.com/questions/12826291/raise-two-errors-at-the-same-time
Again, I think it's currently fine and this is just a suggestion that might require major redo on testing, so it could be something to improve in the future.

matthew7838 (approved):
I think this looks good, and one suggestion on returning errors that could be improved in the future. The testing coverage also seems detailed enough.

SC_redesign/Crop_and_Soil/manager/manure_schedule.py
Isgotkowitz: This is a very good suggestion. I think it would be better to wait on implementing that though because I would want to do that in all the error raising methods in the child classes of `Schedule` which would be a lot of work right now as some are already merged, but I'll add it to #561

PR #559, opened 2023-06-07
Model documentation

tomhuhh:
A folder is created to document all the design docs and pseudocode. Started with Sphinx.


PooyaHekmati (changes requested):
We already have sphinx setup, and the docs folder already exists. Why we need to duplicate those?

tomhuhh:
Thanks @PooyaHekmati for the feedback! This is good to know. I was not aware that Sphinx is applied to the entire MASM folder. So if I understand it correctly, we can start adding rst files into the folder? Do you have recommendations for tools to convert word document to rst files? I apologize if you have mentioned it in our meeting yesterday but I did not quite catch it.

PooyaHekmati:
> Thanks @PooyaHekmati for the feedback! This is good to know. I was not aware that Sphinx is applied to the entire MASM folder. So if I understand it correctly, we can start adding rst files into the folder? Do you have recommendations for tools to convert word document to rst files? I apologize if you have mentioned it in our meeting yesterday but I did not quite catch it.

No worries, there was a lot for you to digest in the meeting. Here is a summary of our conclusions from that meeting:
1- Wiki is not the appropriate place for this type of documentation. Wikis are useful when the content needs to be easy to access and are subject to frequent changes by users and maintainers. Whereas detailed documentation should go to GitHub pages.
2- We have a folder called docs, which hosts a bunch of rst files which are the source files for our GitHub pages. An rst file can be compiled by the Sphinx engine by either commanding the Sphinx engine to auto-generate results from docstrings or by explicitly providing the content.
3- As described [this wiki](https://github.com/RuminantFarmSystems/MASM/wiki/Using-Sphinx), Sphinx can generate HTML and Latex files. Right now, we have an action that runs the Sphinx engine and compiles HTML files every time a merge to master happens. Down the road, we can compile latex as well which in turn can be used to generate PDF and manuscripts as well.
4- I suggested that you take a look at what we already have, both rst files and the current HTML files under docs/_build/html. And come up with a plan for how to best add docs to what we already have. This can be done either by adding new rst files or adding to the existing ones. Then share this plan with Subject Matter Experts, especially with @KFosterReed and move forward after they approve.
5- There are many online resources for learning about rst as well as converting different formats such as markdown (md), and word files (doc, docx) into rst. My quick Google top results are: https://www.vertopal.com/en/convert/doc-to-rst , https://pypi.org/project/rstdoc/ , https://cloudconvert.com/rst-converter . I have not used them myself, that is just the result of my search. After conversion, make sure to double-check the results and fix issues, a perfect result is not guaranteed. 


Please let me know if you have more questions.

KFosterReed:
Hi @tomhuhh and @PooyaHekmati, I like how this plan is emerging. A couple of comments/thoughts: 
1. I think we should still have some high level description of each module on the wiki but this can be short and pretty general like we have for the Feed module now. 
2. The existing documentation that is compiled by Sphinx is the in-code documentation which we should keep separated from the external/scientific documentation that Haowen is working on. There will be some overlap in content between the two but they are distinct. To me, it makes sense to have a separate folder within the docs folder to house all of the scientific documentation one place. What should we call it? ... Scientific Manual? User Guide? Model Documentation?

PooyaHekmati:
> Hi @tomhuhh and @PooyaHekmati, I like how this plan is emerging. A couple of comments/thoughts:
> 
> 1. I think we should still have some high level description of each module on the wiki but this can be short and pretty general like we have for the Feed module now.
> 2. The existing documentation that is compiled by Sphinx is the in-code documentation which we should keep separated from the external/scientific documentation that Haowen is working on. There will be some overlap in content between the two but they are distinct. To me, it makes sense to have a separate folder within the docs folder to house all of the scientific documentation one place. What should we call it? ... Scientific Manual? User Guide? Model Documentation?

I agree that auto-generated content and manually written content should be separated. I also agree that a brief overview can go to the wiki. We can put a link to the extensive doc in the wiki and vice versa.

For the folder, we first need to check if Sphinx allows that. If it does, a name that implies these are not auto-generated would be preferred.

tomhuhh:
Hello @PooyaHekmati and @KFosterReed! Thanks for the feedback. I was able to create a folder named "Model Documentation" in the "docs" folder. With that folder in place and a test rst file, I was able to generate a html page. Of course, details need to be fine-tuned. My next step is to look into how to nest different documents under one topic, for example, animal module. Additionally, I understand there are online tools to convert docx/pdf files to rst files. However, I was uncertain if there is any security concerns given that RuFaS has not been published yet. I am currently using a tool named Vertopal which is secure according to [this](https://www.saasworthy.com/product/vertopal#:~:text=The%20files%20are%20absolutely%20secured,safely%20secured%20in%20other%20sources.) source. Please let me know your thoughts on using these online sources to process RuFaS documentations (i.e. design docs, pseudocode). Thanks!

tomhuhh:
Closing this PR & the branch for now. Please refer to the branch 'Manual-Pseudocode' that @EmmanuelDodoo and @jackie-wells  have been working on for more up-to-date information on this project. Thanks!

PR #560, opened 2023-06-07
Soil amendments 3: `FertilizerSchedule` and `FertilizerEvent`

Isgotkowitz:
This PR contains `FertilizerEvent` and `FertilizerSchedule`, the building blocks that will be used to define a schedule of fertilizer applications for a field.

- `FertilizerEvent` is an `Event` child class that contains the necessary information for a single fertilizer application.
- `FertilizerSchedule` is a `Schedule` subclass that takes in user input and creates and validates a schedule which can produce a list of all `FertilizerEvent`s that would happen as defined by that schedule.
- Input validation is done partly with `_validate_fertilizer_parameters()`, which is supplemented by `_determine_if_all_non_negative_values()`.
- `generate_fertilizer_events()` is the method that handles the actual creation of `FertilizerEvent`s
- all methods are tested in `test_fertilizer_schedule.py`
- progress made by this PR is reflected in `finish_line_doc.md`

Note: the way this class uses inheritance may not be the best/most efficient way to do so, but will be addressed by https://github.com/RuminantFarmSystems/MASM/issues/561


SC_redesign/Crop_and_Soil/manager/fertilizer_schedule.py
```diff
@@ -0,0 +1,169 @@
...
+        valid_fractions = all(0.0 <= fraction <= 1.0 for fraction in self.surface_remainder_fractions)
+        if not valid_fractions:
+            raise ValueError(error_header + f"expected all surface remainder fractions to be in range [0.0, 1.0], "
+                                            f"received '{self.surface_remainder_fractions}'.")
+
+        equal_fertilizer_parameters = len(self.years) == len(self.days) == len(self.mix_names) == \
+            len(self.nitrogen_masses) == len(self.phosphorus_masses) == len(self.application_depths) == \
+            len(self.surface_remainder_fractions)
+        if not equal_fertilizer_parameters:
+            raise ValueError(error_header + f"expected equal numbers of fertilizer application parameters, received "
```
JoeWaddell, line 123: How long are these lists typically going to be? 

Just wondering because this will be a long error message even with a small number, so I'm curious if it would be better to show the length of each bin of data here so the user can spot what's missing, e.g. 2 years, 1 day, 2 mix_names etc...

Then you could log the full message here in the output manager in the error pool. Might be moot if it's only ever a small number.

SC_redesign/Crop_and_Soil/tests_SC/manager_tests/test_fertilizer_schedule.py
```diff
@@ -0,0 +1,82 @@
...
+    ("test_5", ["chex_mix"], [1990, 1992], [100], [10], [10], [-30.0, 30.0], [0.8],
+     "'test_5': expected all application depths to be >= 0, received '[-30.0, 30.0]'."),
+    ("test_6", ["mix_4"], [1991, 1991], [100, 200], [10], [10], [0.0], [1.0, 1.02],
+     "'test_6': expected all surface remainder fractions to be in range [0.0, 1.0], received '[1.0, 1.02]'."),
+    ("test_7", ["mix_5"], [1999, 2000, 2001], [100], [15, 15], [10], [0.0], [1.0],
+     "'test_7': expected equal numbers of fertilizer application parameters, received '[1999, 2000, 2001]' years, "
+     "'[100, 100, 100]' days, '['mix_5', 'mix_5', 'mix_5']' mix names, '[15, 15]' nitrogen masses, '[10, 10, 10]' "
+     "phosphorus masses, '[0.0, 0.0, 0.0]' application depths, and '[1.0, 1.0, 1.0]' surface remainder fractions.")
+])
+def test_validate_fertilizer_parameters(name: str, mix_names: List[str], years: List[int], days: List[int],
```
JoeWaddell, line 33: Should one final test be done to ensure no errors are raised with a valid input?

JoeWaddell (approved):
Looks good to me, just had a couple test questions/thoughts.

SC_redesign/Crop_and_Soil/manager/events.py
```diff
@@ -209,3 +196,49 @@ def __hash__(self):
...
+        hash_value = self.__hash__()
+        other_hash_value = other.__hash__()
+        equal_hash_values = hash_value == other_hash_value
+        return correct_type and equal_hash_values
+
+    def __hash__(self):
+        """Overrides the hash method for FertilizerEvent objects."""
+        str_representation = str(self.year) + str(self.day) + self.mix_name + str(self.nitrogen_mass) + \
+            str(self.phosphorus_mass) + str(self.depth) + str(self.surface_remainder_fraction)
+        return hash(str_representation)
```
ltp23: I would suggest something like the following

In the base `Event` class
```
def __eq__(self, other):
    if isinstance(other, Event):
        return self.year == other.year and self.day == other.day
    return False

def __hash__(self):
    return hash((self.year, self.day))
```

For an event subclass

```
def __eq__(self, other):
    if isinstance(other, FertilizerEvent):
        return super().__eq__(other) and self.mix_name == other.mix_name \
            and self.nitrogen_mass == other.nitrogen_mass \
            # check the remaining attributes
    return False

def __hash__(self):
    return hash((self.year, self.day, self.mix_name, ...))
```

Hashing by concatenating strings is probably not ideal since a string like 123 can mean year1 and day23 or year12 and day3. Despite that, as you concatenate more strings, it is likely to be unique enough anyways. Prioritize attributes that you think are unique and you don't have to include all attributes in a hash. There is a trade-off between uniqueness and speed of hash value computation, which is just as important.

SC_redesign/Crop_and_Soil/manager/fertilizer_schedule.py
```diff
@@ -0,0 +1,169 @@
...
+        all_nitrogen_masses = self.nitrogen_masses * (self.pattern_repeat + 1)
+        all_phosphorus_masses = self.phosphorus_masses * (self.pattern_repeat + 1)
+        all_depths = self.application_depths * (self.pattern_repeat + 1)
+        all_surface_fractions = self.surface_remainder_fractions * (self.pattern_repeat + 1)
+        all_events = list(zip(all_mix_names, all_years, all_days, all_nitrogen_masses, all_phosphorus_masses,
+                              all_depths, all_surface_fractions))
+
+        fertilizer_events = []
+        for event in all_events:
+            new_event = FertilizerEvent(mix_name=event[0], year=event[1], day=event[2], nitrogen_mass=event[3],
```
ltp23, line 151: We can destructure the event to avoid list indexing. 

```
mix_name, year, day, nitrogen_mass, phosphorus_mass, depth, surface_remainder_fraction = event
```

I think it's probably simpler to just use normal for-loop indexing

```
for i in range(self.pattern_repeat + 1):
    fertilizer_events.append(FertilizerEvent(
      mix_name=all_mix_names[i],
      # ... 
    ))

ltp23 (approved):
Some comments, but overall it looks good to me. Good job, Ed!

SC_redesign/Crop_and_Soil/manager/fertilizer_schedule.py
Isgotkowitz: Matthew raised this exact concern in a previous PR too. Length of lists depends on the origin of the input. The ones in `input/` can get pretty long, ~20 values. But if it were only ever going to be novel inputs for testing different scenarios, I wouldn't expect them to get any longer than 4 or 5 because they would utilize the pattern skip and repeat parameters.

The reason I prefer printing the full lists here is because at this stage they have been elongated and the full list would give the user insight into what the expanded inputs look like, which may help them fix their input faster.

20 does seem to be unwieldy though, I'll log the long error message to OutputManager and raise a shorter one. 

Edit: I am actually going to keep it as is. Logging the full error message and raising a shorter one would be better, but I don't want to change it here without changing all the other `Schedule` child classes. And that would be too much work considering our timeline and that this input validation will be replaced by the `InputManager`.

SC_redesign/Crop_and_Soil/manager/events.py
Isgotkowitz: > Hashing by concatenating strings is probably not ideal since a string like 123 can mean year1 and day23 or year12 and day3.

That is a dangerous thing to not consider, thank you for catching that.

But if the hash isn't going to be used for determining equality, what is the benefit of having it defined for `Event` and its subclasses?

ltp23: I think these are separate questions. 

It's good practice that if you decide to implement a semantic equality check, then you should also implement a hash function that is consistent with that check. Normally, the default behavior of the equality check (`==`) is just an identity comparison (ie, same memory location or not, equivalent to `is`). 

You can do away with all of this if you don't need to compare one whole object to another semantically or store them in a dictionary or set. 

Defining methods in a base class helps with code reuse in the subclasses. When you invoke a method in a subclass object, the version in the subclass is looked up first, so it is ok to not do anything in the base class.

SC_redesign/Crop_and_Soil/tests_SC/manager_tests/test_fertilizer_schedule.py
Isgotkowitz: It is implicitly tested through `test_generate_fertilizer_events()`, though you raise a good point. I am going to put it off for now, but will add it to #561 so that equivalent tests can be written for all the child classes of `Schedule` at the same time.

SC_redesign/Crop_and_Soil/manager/fertilizer_schedule.py
Isgotkowitz: I think this would be a better model for creating events, but since all the `Schedule` subclasses generate events with the same structure I will wait for a better time when all the event generation methods can modified at once. In the meantime a note will get added to #561 for this.

SC_redesign/Crop_and_Soil/tests_SC/manager_tests/test_fertilizer_schedule.py
JoeWaddell: Understood! Sounds like a plan, nice work keeping Issues organized.

Isgotkowitz:
@ltp23 and @JoeWaddell thank you for the reviews. You both raised a lot of good points about how the code should be improved, but because of the current time constraints and maintaining consistency between child classes I've decided to record those points in #560 instead of fixing them in this PR. After this push when we have time to start addressing the post-V1 crop and soil issues, we will be able to go back and change all the child classes at once.

PR #562, opened 2023-06-07
Integrate `PlantingEvent`s into `Field`

Isgotkowitz:
This PR integrates `PlantingEvent`s into the `Field` module, making it capable for a field to execute all plantings as dictated by a user through multiple `CropSchedule`s.

- Integration and execution of `PlantingEvent` is handled by 4 new methods
  - `check_crop_planting_schedule()` is called by the daily update routine and checks if any `PlantingEvent`s occur on the current day. If so, it removes them from the list of all planting events and executes them.
  - `_create_and_update_events()` takes a list of `Event`s and returns two lists of `Event`s. The first list is all events from the original list that do not occur on the current day, and the second list is all the events that do occur on the current day. It is written to work with the base `Event`s class so that it may be easily applied when working with other types of events.
  - `plant_crop()` takes a reference to the crop to be planted and a flag indicating how it will be harvested, creates a new `Crop` object using the reference, then adds it to the field's list of currently active crops.
  - `_reset_crop_field_coverage_fractions()` resets all actively growing crops in a field to occupy the same proportion of the field. This is a severely simplified way of handling field coverage in multicropping situations, but is passable for now.
  - all methods are tested in `test_field.py`
- `Field` now accepts a list of `PlantingEvent`s and a dictionary of crop specifications as initialization parameters. 
- Methods `check_schedule()`, `plant_crops()`, and `add_crop()` have been replaced and removed by the above methods.
- The `id` field in `CropData` has been changed to type `Any`, to add greater flexibility in the when uniquely differentiating different `Crop` instances.
- `test_output_gatherer.py` has been tweaked so that it no longer uses one of the methods that was deleted.
- `finish_line_doc.md` has been updated with the progress made by this PR.


SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -141,39 +148,111 @@ def amend_soil(self) -> None:
...
+
+        Notes
+        -----
+        This method is written to work with generic Events so that it may be used on all the different child classes of
+        Event: PlantingEvent, HarvestEvent, ManureEvent, FertilizerEvent, and TillageEvent.
+
+        """
+        todays_events = [event for event in all_events if event.occurs_today(time)]
+        remaining_events = [event for event in all_events if event not in todays_events]
+        return remaining_events, todays_events
```
ltp23: Assuming the list of events is small, the current code is good. 

Alternatively, we want to loop over the list just once. 

```
todays_events = []
remaining_events = []
for event in all_events:
    if event.occurs_today(time):
        todays_events.append(event)
    else:
        remaining_events.append(event)
return remaining_events, todays_events
```

ltp23 (approved):
The code looks great! Thanks, Ed!

SC_redesign/Crop_and_Soil/field/field.py
Isgotkowitz: Good call Loi, we probably won't ever have a list of events that is crazy long but definitely still shouldn't be looping multiple times when we don't have to.

JoeWaddell (approved):
Looks good to me!

PR #565, opened 2023-06-08
Handling Tillage Events

matthew7838:
### Summary
This PR handles how fields will be tilled each day, it goes through every tillage event in the field and checks if events match the date. When the date matches the events, it will be tilled and removed.

### Changes

- Renames the `till_soil` function in `field.py`  to make avoid naming redundancy with `till_soil` in `TillageApplication`.
- Added helper method to filter out today's event, the method will be further tested in PR#564.
- Test coverage now includes the new `till_soil_event()` function.
- Now tillage events that happened on the day will be recorded via the output manager.
- `Record_tillage() ` added in `TillageApplication` to record if `till_soil()` happened.
- Updated finish line document to reflect current progress


SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -24,7 +26,13 @@
 class Field:
     """object representing an agricultural field"""
 
-    def __init__(self, field_data: Optional[FieldData] = None, soil: Optional[Soil] = None):
+    def __init__(self, tillage_events: Optional[TillageEvent] = None, field_data: Optional[FieldData] = None,
+                 soil: Optional[Soil] = None, ):
```
```suggestion
    def __init__(self, field_data: Optional[FieldData] = None, soil: Optional[Soil] = None, 
                 tillage_events: Optional[TillageEvent] = None):
```
Adding to the initialization parameters this way ensures compatibility with any old code that creates field instances with calls like
```python
test_field = Field(field_data, soil)
```

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -24,7 +26,13 @@
...
+
+        # Tillage events
+        self.tillage_events: List[TillageEvent] = tillage_events
+        """List of all tillage events that will occur over the run of the simulation in this field."""
```
Isgotkowitz: This should be grouped with similar attributes, i.e. `self.tillage`

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -49,28 +57,28 @@ def __init__(self, field_data: Optional[FieldData] = None, soil: Optional[Soil]
         self.manure_applicator = ManureApplication(self.soil.data)
         """Manure application interface."""
 
-    def manage_field(self, day: int, year: int, current_weather: CurrentWeather) -> None:
+    def manage_field(self, time: Time, current_weather: CurrentWeather) -> None:
         """main Field function, runs all field routines based on current attribute configuration
 
         Args:
             day: the current (sequential) day of the simulation  - TODO: not yet implemented
             year: the current (sequential) year of the simulation - TODO: not yet implemented
```
Isgotkowitz: Should be removed as they are longer present

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -49,28 +57,28 @@ def __init__(self, field_data: Optional[FieldData] = None, soil: Optional[Soil]
...
             current_weather: a CurrentWeather object, containing a collection of today's weather variables needed
                 for field processes.
+            time: a time object that wil help indicate the year and day
```
Isgotkowitz: The order of parameters' descriptions should match the order in the method signature.

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -129,9 +162,19 @@ def setup_crop_schedule(self, crop_config):
         # </editor-fold>
 
     # <editor-fold desc="--- Soil Management Methods ---">
-    def till_soil(self) -> None:
-        """till the soil"""
-        pass
+    def till_soil_event(self, time: Time) -> None:
```
Isgotkowitz: I think a better name for this method would be `check_tillage_schedule()`

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -129,9 +162,19 @@ def setup_crop_schedule(self, crop_config):
...
+        executed.
+
+        Parameters
+        ----------
+        time : Time
+            Time object containing the current day and year of the simulation.
+        """
+        self.tillage_events, todays_events = self._create_and_update_events(self.tillage_events, time)
+        for event in todays_events:
+            self.tiller.till_soil(event.tillage_depth, event.incorporation_fraction, event.mixing_fraction)
```
Isgotkowitz: Can you add record this event occuring in the `OutputManager`? This will help the econ and emissions module estimate how much energy/fossil fuel is used on the farm.

SC_redesign/Crop_and_Soil/tests_SC/field_tests/test_field.py
```diff
@@ -515,6 +517,25 @@ def test_annual_reset() -> None:
...
+    setattr(mocked_time, "calendar_year", year)
+    setattr(mocked_time, "day", day)
+
+    field = Field(events)
+    tillage_events, todays_events = field._create_and_update_events(events, mocked_time)
+    todays_count = len(todays_events)
+    field.tiller.till_soil = MagicMock()
+    field.till_soil_event(mocked_time)
+
+    assert field.tiller.till_soil.call_count == todays_count
```
Isgotkowitz: There are a few things wrong with this test. `TillageEvent`s takes 5 parameters, `__init__(self, tillage_depth: float, incorporation_fraction: float, mixing_fraction: float, year: int = 1, day: int = 160):`, so the tillage events as specified here have a mixing fraction of 7, and they all have the default day of 160. 

Also, I will pass on a lesson that Pooya taught me in a PR review. When writing tests for a method, you should try to define all the expected results for each test case as much as possible and avoid duplicating the logic of the method being tested as much as possible. The behavior of the function tested here is to remove all the tillage events that happen on the current day from the list of all remaining tillage events, and call `till_soil()` with all the events that occur on the current day. So, the expected results should include two lists of tillage events, one of events that do not occur on the current day and one that does. And the test should check that the `tillage_events` attribute in the field instance is set to the list of the events that do not occur on the current day, and that the `till_soil()` calls have a 1:1 correspondence with the events that do occur on the current day. 

Structuring tests like this will help to catch mistakes in both the thing being tested and test itself.

Isgotkowitz (changes requested):

SC_redesign/Crop_and_Soil/field/tillage_application.py
```diff
@@ -206,3 +216,25 @@ def _remove_amount_incorporated(data_container: object, attribute_name: str,
...
+        Records the mass and nutrients collected in an individual harvest and sends them to the OutputManager.
+
+        Parameters
+        ----------
+        field_name : str
+            Name of the field that contains this crop.
+        year : int
+            Year in which this harvest occurred.
+        day : int
+            Julian day on which this harvest occurred.
```
Isgotkowitz: Need tillage depth, incorporation fraction, and mixing fraction descriptions here.

Isgotkowitz (approved):
Nice PR. A few parameter descriptions missing from one docstring, everything else looks perfect.

ltp23 (approved):
Overall, the code looks good to me. Thanks Matthew!

I think I understand the `Event` class and its subclasses a little better. 

One suggestion I have is to use the heap data structure to order the events from earliest to latest. 

To do that, first I would implement the special method __lt__() (which stands for less than comparison) in the `Event` class. 

```
class Event:
    ...
    
    def __lt__(self, other):
        return (self.year, self.day) < (other.year, other.day)

    ...
```

During initialization,
```
heapq.heapify(tillage_events)
self.tillage_events = tillage_events
```

When checking for today's events
```
def check_tillage_schedule(self, time):
    while self.tillage_events and self.tillage_events[0].occurs_today(time):
        event = heapq.heappop(self.tillage_events)
        self.tiller.till_soil(...)
```
This would eliminate the need for the helper method `_create_and_update_events()` and probably the `__eq__()` and `__hash__()` methods in `Event`s classes as well.

If there is a time when we need to add an event ad hoc,
```
heapq.heappush(self.tillage_events, TillageEvent(...))
```

Isgotkowitz:
> Overall, the code looks good to me. Thanks Matthew!
> 
> I think I understand the `Event` class and its subclasses a little better.
> 
> One suggestion I have is to use the heap data structure to order the events from earliest to latest.
> 
> To do that, first I would implement the special method **lt**() (which stands for less than comparison) in the `Event` class.
> 
> ```
> class Event:
>     ...
>     
>     def __lt__(self, other):
>         return (self.year, self.day) < (other.year, other.day)
> 
>     ...
> ```
> 
> During initialization,
> 
> ```
> heapq.heapify(tillage_events)
> self.tillage_events = tillage_events
> ```
> 
> When checking for today's events
> 
> ```
> def check_tillage_schedule(self, time):
>     while self.tillage_events and self.tillage_events[0].occurs_today(time):
>         event = heapq.heappop(self.tillage_events)
>         self.tiller.till_soil(...)
> ```
> 
> This would eliminate the need for the helper method `_create_and_update_events()` and probably the `__eq__()` and `__hash__()` methods in `Event`s classes as well.
> 
> If there is a time when we need to add an event ad hoc,
> 
> ```
> heapq.heappush(self.tillage_events, TillageEvent(...))
> ```

@ltp23 I thinking about this same issue while I was writing `Event`s. As you suggested, it would save a lot of time to simply check the front element only of a list and continually popping off elements until one is reached that does not occur on the current day. Would a data structure for this be a heap or a queue though?

ltp23:
A heap is more general and self-organizes, but a queue would work the same if you add the events in the right order - monotonically increasing.

PR #566, opened 2023-06-12
NRC Maintenance energy typo fix

KFosterReed:
There was a small typo in the calculate_NRC_energy_maintenance_requiremets function in animal_requirements.py that resulted in over estimation of energy requirements. 

See the current pseudocode on Basecamp or here: https://docs.google.com/document/d/1jLxqY2rJC4BFfHlOpllHfbUItNUMRyyulKSS9X4hPOM/edit?usp=sharing  equation A.Heifer.A.4 for reference.


Ed731551 (changes requested):
Looking at page 318 of NRC (2001) book, this value should be 0.05 and not 0.5. I agree, this may lead to overestimate the energy requirement for this animal category

tomhuhh:
@KFosterReed Thanks for catching this! This might also explain the odd behavior with heifers' diets we noticed before. Also thanks @Ed731551 for double checking the literature. I updated the unit tests based on the fix. Thus, it should be good to be merged to master. Thanks!

Ed731551 (approved):

JoeWaddell (approved):
Nice work all round!

PR #568, opened 2023-06-12
Integrate `HarvestEvent`s into the `Field` module

Isgotkowitz:
This PR `HarvestEvent`s into the `Field` module, allowing crops to now be harvested as scheduled by users.

- `Field` now maintains a list of all remaining harvest events for a simulation.
- `check_crop_harvest_schedule()` determines which scheduled harvest operations occur on the current day and executes them, if any, as well as calling some other methods to handle other aspects of crop harvesting.
- `_harvest_heat_scheduled_crops()` harvests all crops that have heat scheduled harvests and are ready to be harvested.
- `harvest_crop()` takes the parameters for a single harvest event and executes them on a single crop in the field, otherwise raises warnings to the `OutputManager`
- `_remove_dead_crops()` removes all dead crops from the field's list of active crops.
- all methods are tested in `test_field.py`


SC_redesign/Crop_and_Soil/tests_SC/field_tests/test_field.py
```diff
@@ -122,13 +200,105 @@ def test_plant_crop_error(field_name: str, crop_reference: str, custom_crop_spec
...
+    mocked_time = MagicMock(Time)
+    setattr(mocked_time, "day", 200)
+    setattr(mocked_time, "calendar_year", 2000)
+
+    field.harvest_crop("test", "default", mocked_time)
+
+    actual = om.warnings_pool["Field:'test'.harvest_warning"]
+    print(actual)
+    print(actual['values'])
+    print(expected_message)
```
Isgotkowitz: Ed remove these

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -161,6 +167,48 @@ def check_crop_planting_schedule(self, time: Time) -> None:
...
+        """
+        self.harvest_events, todays_harvest_events = self._create_and_update_events(self.harvest_events, time)
+        for event in todays_harvest_events:
+            self._harvest_crop(event.crop_reference, event.operation, time)
+
+        self._harvest_heat_scheduled_crops()
+
+        self._remove_dead_crops()
+
+        self._reset_crop_field_coverage_fractions()
```
matthew7838: One question, if this method is meant to deal with all harvests not just the ones that match the time, why not move `_harvest_heat_scheduled_crops()` in here and just be one method? I don't think it adds too much complexity to the method.

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -253,6 +301,56 @@ def _plant_crop(self, crop_reference: str, use_heat_scheduled_harvesting: bool,
...
+        """
+        crops_to_be_harvested = [crop for crop in self.crops if crop.data.id == crop_reference]
+
+        info_map = {"class": self.__class__.__name__, "function": self._harvest_crop.__name__,
+                    "prefix": f"field_name:'{self.field_data.name}'",
+                    "date": {"day": time.day, "year": time.calendar_year}}
+        if len(crops_to_be_harvested) > 1:
+            om.add_warning("harvest_warning", "Multiple crops to be harvested by single HarvestEvent.", info_map)
+        elif len(crops_to_be_harvested) < 1:
+            om.add_warning("harvest_warning", "No crop found to be harvested by a HarvestEvent.", info_map)
```
matthew7838, line 354: Just like the tillage event, maybe move anything OM involved to be a separate method?

matthew7838 (approved):
Looks good, just some questions and suggestions. I like how you separate the OM-related test in a different method, which would relate to one of the points I brought up.

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -161,6 +167,48 @@ def check_crop_planting_schedule(self, time: Time) -> None:
         for event in todays_planting_events:
             self._plant_crop(event.crop_reference, event.use_heat_scheduled_harvest, time)
 
+    def check_crop_harvest_schedule(self, time: Time) -> None:
```
ew3361zh, line 173: It seems like this method does a few more things than just checking the crop harvest schedule. Do you think there's a more encompassing verb that would cover all the activities? I see in the docstring notes that these other activities are listed - I don't necessarily think it needs to be broken up but it feels like the notes section is covering central and key activities this method does that are vital for crop harvesting.

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -161,6 +167,48 @@ def check_crop_planting_schedule(self, time: Time) -> None:
...
+            Time object containing the current day and year of the simulation.
+
+        Notes
+        -----
+        This method checks for scheduled harvests, i.e. checks all the remaining HarvestEvents. It also checks if any of
+        the active crops are to be harvested using heat scheduling, and if so checks if they have met the harvesting
+        threshold.
+
+        """
+        self.harvest_events, todays_harvest_events = self._create_and_update_events(self.harvest_events, time)
```
ew3361zh: I know this isn't part of this PR but seeing 'and' in a function name usually means a function is doing 1 too many things. I looked at it though and I think it's good as is but I'd think about renaming it as well (sorry, not immediately relevant to this PR but thought I would note it).

SC_redesign/Crop_and_Soil/field/field.py
ew3361zh: I like the idea of moving all variables being added into one method but I think warnings are useful to be added at the point in the code where the warning is arising. I would worry about losing too much context if it were moved out of this context.

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -253,6 +301,56 @@ def _plant_crop(self, crop_reference: str, use_heat_scheduled_harvesting: bool,
...
+        ----------
+        crop_reference : str
+            Name used to get the specifications for the crop to be harvested.
+        harvest_operation : str
+            Name of the harvest operation to be performed on the referenced crop.
+        time : Time
+            Object containing the current day and year of the simulation.
+
+        Raises
+        ------
```
ew3361zh: I don't think adding warnings to the OM warnings pool should be considered the same as raising an official python warning and I don't think necessarily needs to be added in the docstrings. We're not noting these this way anywhere else in the model.

ew3361zh (approved):
Some comments about docstrings and function names but overall looks really good!

SC_redesign/Crop_and_Soil/field/field.py
Isgotkowitz: Good to know

SC_redesign/Crop_and_Soil/field/field.py
Isgotkowitz: I agree, `_harvest_heat_scheduled_crops()` is not very complex but I like having it separated out here because 1) it more distinctly separates the two types of harvesting and 2) it makes testing a lot easier/simpler.

SC_redesign/Crop_and_Soil/field/field.py
Isgotkowitz: This is a good point. I think instead of changing the name though I will move `_remove_dead_crops()` and `_reset_crop_field_coverage()` to a higher level method. Part of why I wish to make `check_crop_harvest_schedule()` simpler is so that name can stay and be more appropriate, which keeps the names of all the different methods that check schedules consistent (`check_crop_planting_schedule()`, `check_tillage_schedule()`, etc)

Also now that I think about it, there are edge cases that would require the above methods to be called in a higher-level method.

SC_redesign/Crop_and_Soil/field/field.py
Isgotkowitz: You raise an important point though, I'll change it to `_filter_events()`

SC_redesign/Crop_and_Soil/field/field.py
Isgotkowitz: I will leave the warnings where they are for now, but I did notice that a variable is being added in `_plant_crop()` when it should be in a separate method, will change that in this PR. I think for now this is a good standard to follow: warnings can be raised where they are found, but variables should be added in their own methods.

ew3361zh: I think that sounds good unless you have specific context data that wouldn't be able to be added with the variable if it were added in a separate method. I should have said that I like how you're adding the daily and annual update variables all in one method because they're all kind of associated with the update. But there isn't any reason a variable can't be added at a separate point if it's generated by a method/action that's not associated with either a daily or annual update.

PR #570, opened 2023-06-12
Recording crop yields and transferring yield residue in `CropManagement`

Isgotkowitz:
This PR introduces some mechanisms into the `CropManagement` module to better handle the recording of harvested crops and the transfer of unharvested biomass to the appropriate trackers.

- `_record_yield()` records the mass of the crop harvested, as well as the nutrients contained, to the `OutputManager`. It also saves the date of the harvest, harvest name, and species name as metadata in the `info_map`.
- `_transfer_residue()` collects the residue that was cut during the harvest but was not collected and adds it to the `SoilData` instance that is associated with the field the crop is kept in.
- `manage_harvest()` uses the above two methods, and takes a couple new parameters as a result. 
- All changes and new methods are tested in `test_crop_management.py`.


matthew7838 (approved):
LGTM, a straightforward PR. Nicely done.

ew3361zh (approved):
Looks great, nicely done! No questions from me.

Issue #567, opened 2023-06-12
Double Counting of Maintenance Protein Requirements


KFosterReed:
I believe that the new updates to the requirements calculations are resulting in a double counting of some of the protein requirements for maintenance. 
## Expected Behavior

The metabolizable protein requirement for maintenance should only be counted once. 

## Current Behavior

In the protein constraint function within the ration_NLP.py file, L393 estimates the metabolizable protein requirements for maintenance and adds it to the MP_req before finding the difference between the MP_supply and requirement.  However, I believe the same requirement is also estimated in L661-664 in the animal_requirements.py file and included in the MP_req that is passed to the ration_NLP.py file

## Possible Solution

Delete the MPm calculation in the protein_constraint function in ration_NLP

## Steps to Reproduce

1.
2.
3.
4.

PR #571, opened 2023-06-13
Fixed the double counted MP in NLP

tomhuhh:
Also added a TODO: Separate the DMI components from MPm requirements calculated by both NRC (2001) and NASEM (2021) and put them in this section because DMI gets reestimated during each iteration.

### DIFF
--- a/RUFAS/routines/animal/ration/ration_NLP.py
+++ b/RUFAS/routines/animal/ration/ration_NLP.py
@@ -385,6 +385,8 @@ def protein_constraint(x):
     # Total metabolizable protein supply
     MP_supply = MPbact + RUP_diet + 0.4 * 11.8 * DMI
 
+    return (MP_supply - (MP_req / 1000))
+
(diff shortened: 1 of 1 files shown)


Ed731551 (approved):
Implemented changes make sense to me according to previous discussions.

tomhuhh:
Hello @KFosterReed! The issue described in Issue #567 is addressed here. There currently seems to be no unit test for the modified function `def protein_constraint(x):`. Other than that, the duplicate section has been deleted from this function in the NLP.py file and a TODO and a new issue #572  are created. Please let us know what your thoughts are. Thanks!

Screenshot 2023-06-13 at 4 02 25 PM

Ed731551 (approved):

JoeWaddell (changes requested):
Looks good, I just think we should remove the TODO and make it an issue, instead.

tomhuhh:
Add issue #622.

JoeWaddell (approved):

PR #578, opened 2023-06-16
Create feed db from csv

PooyaHekmati:
This PR created the feed database before running the simulation engine. While this adds a little bit of extra time to runtime (which can be avoided by removing the call to setup_feed_db_from_csv() in the main function), it allows us to alter the feeds by simply updating these csv files instead of needing to update the sqlite db. This is particularly important for our integration with Vyal. This PR makes them agnostic about how we handle the feed, all they need to know and do is about csv files.

Note 1: it is also possible to use json files instead of csv file; but since we are making tables, csv is more efficient.
Note 2: The feed db is expected to be decommissioned, this PR implements a temp code until we do that.


KFosterReed:
This doesn't seem to work for me. I get this error when I try to run it before the I have the opportunity to put the input file path in: 
image

I tried moving the call to set_up_feed_db to within the run_rufas function after the obtain_file_list function was called like this: 

image

but then I got this error: 

image

Not sure where to go next :/

PooyaHekmati:
> This doesn't seem to work for me. I get this error when I try to run it before the I have the opportunity to put the input file path in: image
> 
> I tried moving the call to set_up_feed_db to within the run_rufas function after the obtain_file_list function was called like this:
> 
> image
> 
> but then I got this error:
> 
> image
> 
> Not sure where to go next :/

Thanks for taking the time and reviewing this. Different reasons can be behind this error message, but my first guess is that it is happening because you are running it in drive C: and your drive C: is protected. So when the code attempts to create the DB, Windows rejects the request. Is it possible for you to run it in a different drive?

JoeWaddell:
I've tested this on my home desktop over the weekend, and just now in the lab on my work laptop, and both run fine. I'll see if I can replicate the bug Kristan found, but it looks good on my end.

allisterakun:
Hi @KFosterReed, @PooyaHekmati has asked me to help with this issue. One of Pooya's suspicions for the reason is file access. 
Would you please run two lines of code to help me narrow the problem?

- Judging from the screenshots you shared earlier in the conversation, I believe you are using PyCharm as the IDE. Please click the **Terminal** tab at the bottom to open up a new terminal. 
image

- Please run `icacls input/databases/feeds.sqlite` and post a screenshot of the result.

- Finally, please try running `icacls input\databases\feeds.sqlite /grant Users:F` and see if it will fix the problem.

KFosterReed:
Hi @allisterakun , Thanks for working on this!
Here is the screenshot when I ran the lines of code you shared:
image

But, unfortunately, I still get the same error: 
image

PR #581, opened 2023-06-20
Integrate `ManureEvent`s into `Field` module 

Isgotkowitz:
This PR adds the functionality for `Field` to apply manure to the field based on a series of `ManureEvent`s.

- A list of `ManureEvent`s has been added which will dictate exactly what the `Field` module requests from the manure manager when a manure application is scheduled to happen.
- `available_fertilizer_mixes` now has some default mixes that will always be available if no other mixes are provided.
- `check_manure_application_schedule()` now checks for all manure applications that should happen on a single day and passes them to be executed, and is called every day from the main routine in `Field`
- `_execute_manure_application()` takes the fields from a single `ManureEvent`, request and apply manure based on those fields, and meet any unmet nutrient demands with fertilizer.
- `_determine_optimal_fertilizer_mix()` takes nutrient demands and a list of available fertilizers, and determines which mix can be used to meet the demands with the least amount of fertilizer.
- `_record_manure_application()` records the actual amounts of manure and nutrients that are applied to the field to the `OutputManager`. 

Note: once the method for requesting manure in the `ManureManagement` module is finished, it will need to be integrated into `_execute_manure_application()` - there are placeholder values hardcoded there for now.

### DIFF
--- a/SC_redesign/Crop_and_Soil/field/field.py
+++ b/SC_redesign/Crop_and_Soil/field/field.py
@@ -289,7 +289,7 @@ class Field:
             Julian day on which this fertilizer application is occurring.
 
         """
-        info_map = {"class": self.__class__.__name__, "function": self._execute_fertilizer_application.__name__,
+        info_map = {"class": self.__class__.__name__, "function": self._record_fertilizer_application.__name__,
                     "prefix": f"field_name:'{self.field_data.name}'", "date": {"year": year, "day": day},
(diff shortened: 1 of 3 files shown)


ltp23 (approved):
The code looks great to me. I think it could use some defensive programming here and there to verify your expectation of arguments and return values.

matthew7838 (approved):
Both code and test coverage look good. Nice PR.

PR #583, opened 2023-06-20
Adding manure nutrients and nutrient request classes

ltp23:
This PR is the first of several that deals with connecting the manure module to the crop and soil module. 

The two new code files are:

`manure_nutrients.py`: introduces the ManureNutrients class that stores relevant manure nutrient content to the field. This class also handles validation and arithmetic operations for nutrient amounts, allowing them to behave like primitive data types.

`nutrient_request.py`: introduces the NutrientRequest class that represents a request from the field. It verifies that the values in a request are non-negative and they have at least a positive value.

### DIFF
--- a/RUFAS/routines/manure/manure_nutrients/manure_nutrients.py
+++ b/RUFAS/routines/manure/manure_nutrients/manure_nutrients.py
@@ -1,0 +1,179 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, fields
+
+
+@dataclass(kw_only=True, frozen=True)
+class ManureNutrients:
+    """A class to store the relevant manure nutrient information to be passed to the crop and soil module"""
+
+    nitrogen: float = 0.0
+    """Amount of accumulated manure nitrogen derived from the manure module, kg."""
+
+    phosphorus: float = 0.0
+    """Amount of accumulated manure phosphorus derived from the manure module, kg."""
+
+    potassium: float = 0.0
+    """Amount of accumulated manure potassium derived from the manure module, kg."""
+
+    dry_matter: float = 0.0
+    """Amount of accumulated dry matter derived from the manure module, kg."""
+
+    total_manure_mass: float = 0.0
+    """Amount of accumulated manure mass derived from the manure module, kg."""
+
+    def __post_init__(self):
+        """
+        Validate the dataclass fields.
+
+        Raises
+        ------
+        ValueError
+            If any field is negative.
+
+        """
+        for field in fields(self):
+            if getattr(self, field.name) < 0:
+                raise ValueError(f'Field {field.name} must be non-negative.')
+
+    @property
+    def dry_matter_fraction(self) -> float:
+        """
+        Calculate the dry matter fraction of the manure.
+
+        Returns
+        -------
+        float
+            The dry matter fraction of the manure, unitless, between 0 and 1.
+
+        """
+        if self.total_manure_mass == 0.0:
+            return 0.0
+        return self.dry_matter / self.total_manure_mass
+
+    @property
+    def nitrogen_composition(self) -> float:
+        """
+        Calculate the nitrogen composition of the manure.
+
+        Returns
+        -------
+        float
+            The nitrogen composition of the manure, unitless, between 0 and 1.
+
+        """
+        if self.total_manure_mass == 0.0:
+            return 0.0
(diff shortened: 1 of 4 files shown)


RUFAS/routines/manure/manure_nutrients/manure_nutrients.py
```diff
@@ -0,0 +1,179 @@
...
+        scalar : int | float
+            The scalar to multiply by.
+
+        Returns
+        -------
+        ManureNutrients
+            The product of the ManureNutrients object and the scalar.
+
+        """
+        return self.__mul__(scalar)
```
Isgotkowitz, line 179: Why is it necessary to include `__rmul__` if it does the same thing as `__mul__`, I assume just a best practice to always have a left multiply and a right multiply?

RUFAS/routines/manure/manure_nutrients/manure_nutrients.py
```diff
@@ -0,0 +1,179 @@
...
+
+from dataclasses import dataclass, fields
+
+
+@dataclass(kw_only=True, frozen=True)
+class ManureNutrients:
+    """A class to store the relevant manure nutrient information to be passed to the crop and soil module"""
+
+    nitrogen: float = 0.0
+    """Amount of accumulated manure nitrogen derived from the manure module, kg."""
```
Isgotkowitz, line 11:
```suggestion
    """Amount of accumulated manure nitrogen derived from the manure module (kg)"""
```
Not an important issue at all, but in the Crop and Soil module we always format the units like this. Doesn't need to changed here  (especially if this is the style the Manure module already uses), but it's another thing we should decide a team standard for at some point.

tests/test_manure_module/manure_nutrients/test_manure_nutrients.py
```diff
@@ -0,0 +1,151 @@
...
+def test_manure_nutrients_invalid_init(nutrient_values: dict[str, float]):
+    """
+    Unit test for the __post_init__ method in the ManureNutrients class in manure_nutrients.py.
+
+    Test the validation in the initialization of the ManureNutrients class,
+    expecting it to raise an error when any of the attributes are negative.
+
+    """
+    with pytest.raises(ValueError):
+        ManureNutrients(**nutrient_values)
```
Isgotkowitz: It may be good to test that the error message is correct as well.

Isgotkowitz (approved):
Nice PR Loi, very clean. I had some minor comments about docstring formatting, the two multiply methods, and a suggestion for testing error cases.

varrma (approved):
Looks good Loi !!

RUFAS/routines/manure/manure_nutrients/manure_nutrients.py
ltp23: I agree. I think if you ever need to implement one of the two, you should implement the other as well. Normally, the `__mul__()` method will be called first.  If it doesn't exist, then the `__rmul__()` method will be called. In this case, `__rmul__()` looks more natural with the object on the right (a * obj), but it is ok to switch the order too.

RUFAS/routines/manure/manure_nutrients/manure_nutrients.py
ltp23: Great that you pointed this out. I could never be consistent about this. We have several choices:
- `description, <unit>.`
- `description, in <unit>.`
- `description (<unit>).`
- `description, measured in <unit>.`

I tend to flip-flop between the first two. But I can adopt your convention too. It looks good. 

Another thing is when there are no units, I think `unitless` sounds better, but technically `dimensionless` is more correct.

tests/test_manure_module/manure_nutrients/test_manure_nutrients.py
ltp23: Ok. I will add that to the test.

RUFAS/routines/manure/manure_nutrients/manure_nutrients.py
Isgotkowitz: I think the third option, `description, (<unit>).`,  would be best because it makes the description a proper sentence which will benefit the sphinx documentation greatly (IMO).

Also I agree that `unitless` sounds better, but maybe it would be best to let the SME's decide if it is important enough to use `dimensionless` in the docstrings.

PR #584, opened 2023-06-20
Logging Output Manager Filtering Activity

ew3361zh:
This PR adds OM logging within OM's own filtering functions.

This tracks vital OM filtering activity including:
- The status of the exclude_info_maps flag for the simulation
- The number of filter files found in the filter files directory
- Whether the exclude_keyword was found in the first position of each particular filter file
- The number of keys resulting from each application of the filter file's patterns to the variables pool

This will be helpful to users and developers diagnose unexpected results found in filtered variables output files.

Also updates unit testing accordingly.

Note:
- Fixes issue of OM only logging one of the new logs by switching the order of OM function calls in main.py

### DIFF
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -416,7 +416,6 @@ class OutputManager(object):
                                       any(re.match(pattern, key) for pattern in filter_patterns)}
         filter_log_count_msg = f"There were {len(filter_pattern_matches)} matches for {input_file} filter patterns"
         self.add_log("num_filter_pattern_matches", filter_log_count_msg, info_map)
-        print(f"There were {len(filter_pattern_matches)} matches for {input_file} filter patterns")
         return filter_pattern_matches
 
     def save_variables(self, save_path: str, dir_path: str,
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -264,6 +264,10 @@ class OutputManager(object):
         serialized dictionary using the max_depth parameter.
 
         """
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._dict_to_file_json.__name__,
+                    }
+        self.add_log("save_dict_file_try", f"Attempting to save to {path}.", info_map)
         try:
             with open(path, "w") as json_file:
                 json.dump(
@@ -271,6 +275,7 @@ class OutputManager(object):
                     json_file,
                     indent=0,
                 )
+                self.add_log("save_dict_file_success", f"Successfully saved to {path}.", info_map)
         except Exception as e:
             raise e
 
@@ -290,9 +295,14 @@ class OutputManager(object):
             If an error occurs while saving to the file
 
         """
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._list_to_file_txt.__name__,
+                    }
+        self.add_log("save_txt_file_try", f"Attempting to save to {path}.", info_map)
         try:
             with open(path, "w") as var_names_file:
                 var_names_file.writelines(data_list)
+                self.add_log("save_txt_file_success", f"Successfully saved to {path}.", info_map)
         except Exception as e:
             raise e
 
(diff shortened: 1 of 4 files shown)


RUFAS/output_manager.py
```diff
@@ -351,13 +366,20 @@ def _load_txt_file_to_list(self, path: str) -> List[str]:
...
         """
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._load_txt_file_to_list.__name__,
+                    }
+        self.add_log("open_text_file", f"Attempting to open {path}.", info_map)
         try:
             with open(path) as text_file:
-                return text_file.read().splitlines()
+                list_of_elements = text_file.read().splitlines()
+                load_message = f"Successfully opened {path}, it contained {len(list_of_elements)} filter patterns."
```
```suggestion
                load_message = f"Successfully opened {path} and read {len(list_of_elements)} lines."
```
This function is "_load_txt_file_to_list" which means here we don't what are these lines yet. We had the same thing in a previous PR that led to renaming this function to its current name.
Also, at this point, we have successfully read the content of the file, we should log that too. Hence changing 'contained' to 'read'.

RUFAS/output_manager.py
```diff
@@ -366,6 +388,9 @@ def _filter_variables_pool(self, filter_patterns: List[str]) -> Dict[str, pool_e
         filter_patterns : List[str]
             A list of patterns the user has selected to filter the variables pool.
 
+        input_file : str
```
```suggestion
        input_file_name : str, optional
```

RUFAS/output_manager.py
```diff
@@ -351,13 +366,20 @@ def _load_txt_file_to_list(self, path: str) -> List[str]:
...
-                return text_file.read().splitlines()
+                list_of_elements = text_file.read().splitlines()
+                load_message = f"Successfully opened {path}, it contained {len(list_of_elements)} filter patterns."
+                self.add_log("filter_pattern_file_load_log", load_message, info_map)
+                return list_of_elements
         except Exception as e:
             raise e
 
-    def _filter_variables_pool(self, filter_patterns: List[str]) -> Dict[str, pool_element_type]:
+    def _filter_variables_pool(self, filter_patterns: List[str], input_file: str) -> Dict[str, pool_element_type]:
```
```suggestion
    def _filter_variables_pool(self, filter_patterns: List[str], input_file_name: Optional[str]) -> Dict[str, pool_element_type]:
```
1. It is the file name, not the file object itself.
2. The function can work without it, hence optional.

RUFAS/output_manager.py
```diff
@@ -380,15 +405,27 @@ def _filter_variables_pool(self, filter_patterns: List[str]) -> Dict[str, pool_e
...
+                    "function": self._filter_variables_pool.__name__,
+                    }
         exclude_keyword_location = 0
         exclude_keyword = "exclude"
         filter_by_exclusion = filter_patterns and filter_patterns[exclude_keyword_location] == exclude_keyword
         if filter_by_exclusion:
-            return {key: self.variables_pool[key] for key in self.variables_pool.keys() if not
-                    any(re.match(pattern, key) for pattern in filter_patterns)}
+            filter_vars_msg = f"{input_file} has exclude-keyword '{exclude_keyword}' at"\
+                f" position {exclude_keyword_location}."
```
```suggestion
                f" position {exclude_keyword_location}. Performing filtering by exclusion."
```

RUFAS/output_manager.py
```diff
@@ -380,15 +405,27 @@ def _filter_variables_pool(self, filter_patterns: List[str]) -> Dict[str, pool_e
...
+            filter_pattern_matches = {key: self.variables_pool[key] for key in self.variables_pool.keys() if not
+                                      any(re.match(pattern, key) for pattern in filter_patterns)}
         else:
-            return {key: self.variables_pool[key] for key in self.variables_pool.keys() if
-                    any(re.match(pattern, key) for pattern in filter_patterns)}
+            filter_vars_msg = f"{input_file} does NOT contain exclude-keyword '{exclude_keyword}'"\
+                f" at position {exclude_keyword_location}."
```
```suggestion
                f" at position {exclude_keyword_location}. Performing filtering by inclusion."
```

RUFAS/output_manager.py
```diff
@@ -408,15 +445,18 @@ def save_variables(self, save_path: str, dir_path: str,
             Flag for whether or not the user wants to include info_maps data in their results files.
 
         """
+        info_map = {"class": self.__class__.__name__,
```
PooyaHekmati, line 449: I feel bad for this info_map, we are about to exclude it :(

PooyaHekmati (approved):

RUFAS/output_manager.py
ew3361zh: Haha, it did make me giggle thinking about that when writing it in there.

JoeWaddell (approved):
Looks good to me! Nice work, and I always appreciate the detailed comment to open the review.

PR #585, opened 2023-06-20
User defined ration input change

JoeWaddell:
Changed logic so the `input\userdefinedration\user_defined_ration_input_percentages.json` is no longer used, and the date inside was moved to another field in the feed input JSON. 

Other changes to the code were necessary to accommodate and avoid circular imports, but it appears to work as intended.


tomhuhh:
I tested again with the Northeastern diet I used for my case study and it turns out that the DMI is still underestimated. The predicted DMI using NRC as the nutrient requirement model is 18.5 kg on average, while the expected DMI is ~ 28 kg. Looking at the milk production, a penalty seems to have been correctly applied.  

image

JoeWaddell:
> I tested again with the Northeastern diet I used for my case study and it turns out that the DMI is still underestimated. The predicted DMI using NRC as the nutrient requirement model is 18.5 kg on average, while the expected DMI is ~ 28 kg. Looking at the milk production, a penalty seems to have been correctly applied.
> 


This is the anticipated behavior: low DMI is expected given that lactation has been reduced. 

However, this indicates that the ration could not meet the requirements. If this is unexpected, the follow-up questions are: is lactation reduced in all months, or is this intermittent behavior? 

Also note there are reports in the outputs that will show which constraints are failing for each attempt prior to reducing lactation. We can pinpoint why apparently OK rations are failing.

Finally, note that you can force some override here: in the lactation reduction input, if you lower the amount it is reduced, you won't see such a drop in DMI (since it will "OK" the user defined ration before too many reformulation attempts).

tomhuhh:
> 

Thank you Joe. Very clear logic. I was actually looking into the variables that failed to be met and found the following variables with a high occurrence. 

"NEmact_constraint",
"NEl_constraint",
"NEgact_constraint",
"DMI_constraint_lower", 
"NDF_constraint_2", 
"phosphorus_constraint"

RUFAS/routines/animal/ration/user_defined_ration.py
```diff
@@ -6,8 +6,7 @@
 Author: Joseph C. Waddell, jw2574@cornell.edu
 """
 
-from typing import Any, Dict
-import json
+from typing import Dict
 
 class UserDefinedRationManager(object):
     """
```
ltp23, line 12: Hi Joe, for this class, I have a couple of suggestions

```
def __new__(cls):
    if not hasattr(cls, '__instance'):  # currently missing the two underscores
        cls.__instance = super(UserDefinedRationManager, cls).__new__(cls)
    return cls.__instance
```

then in the `__init__()`, we could include a dictionary for different ration types

```
self.ration_dict = {
    'LAC_COW': [],
    'GROWING': [],
    'CLOSE_UP': [],
    'DEFAULT': []
}
```

In the `ration_to_use()`, we could simplify the if/elif branches with this

```
def ration_to_use(self, pen_animal_combo, available_feeds: Dict) -> Dict:
    group = pen_animal_combo.name 
    ration_percents = self.ration_dict.get(group, self.ration_dict['DEFAULT'])
    return self.feed_quality_fix(ration_percents, available_feeds)
```

And the three methods `feed_quality_fix`, `ration_to_use`, and `make_ration_from_user_values` look like they are missing the  `self` parameters.

We may need to adjust line 601 in `ration_NLP.py` if the `self` parameter is used in `ration_to_use()`:

```
udr_manager = UserDefinedRationManager()
bnds = make_user_bounds(udr_manager.ration_to_use(animal_combination, available_feeds), DMIest)
```

input/feed/purchased_feed.json
```diff
@@ -27,5 +27,13 @@
             "initial_dry_matter": 0
         }
 
-    }
+    },
+    "user_defined_ration_percentages": {
+        "calf": {},
+        "growing_heifers": {},
+        "close_up": {},
+        "cow_lactating": {},
```
ltp23: Would it be more consistent to name this one as `lactating_cow(s)`?

ltp23 (approved):
Other than the suggestions I had, the code looks good to me.

input/feed/purchased_feed.json
JoeWaddell: Absolutely: I'll make sure the entries here match the entries for the other inputs in this JSON.

RUFAS/routines/animal/ration/user_defined_ration.py
JoeWaddell: Wonderful: I'll make these changes ASAP, and thanks for the detailed improvements!

tomhuhh:
Hello Joe, just a follow up. I tested with a milk reduction percent of both 0.05 and 0.01 with different diets, the situation improved but I still observed failed constraints. I narrowed down to the following constraints based on the tests that I have run so far. 

'DMI_constraint_lower', 'NEl_constraint', 'NEmact_constraint' 

I am uncertain what causes this but I will keep looking. 

Thank you!

KFosterReed:
Hi Joe and Haowen! I am also getting a lot of violated constraints and I'm not really sure why. I am also getting a lot of lower DMI constraint violations as well as the energy constraint violations which is confusing to me because it seems like a simple way to increase the energy delivery would be to increase the intake. I haven't reviewed the constraint function in detail yet but I will do that tomorrow to see if any bright ideas come my way

JoeWaddell:
> > 
> 
> Thank you Joe. Very clear logic. I was actually looking into the variables that failed to be met and found the following variables with a high occurrence.
> 
> "NEmact_constraint", "NEl_constraint", "NEgact_constraint", "DMI_constraint_lower", "NDF_constraint_2", "phosphorus_constraint"

So the `NDF_constraint_2` (which we can rename for clarity) is an upper NDF bound. So I think that's why the lower DMI and all three net energy constraints are failing: it's trying to simultaneously reduce the NDF, and raise the others. Looking into it now!

tomhuhh:
Hello Joe, a quick update on the diet tests. With the NE diet that is already in the branch, there are still some failed constraints: "NEmact_constraint", "NEl_constraint", "calcium_constraint", "phosphorus_constraint". Good news is that "DMI_constraint_lower" is no longer on the list. Good work!

tomhuhh:
First of all, great work Joe! I chatted with you about the modifications you made to the ration formulation functionality. What you explained to me makes sense. Energy and protein supply should not be divided into three portions to be compared with the requirements. With that, I have tested with a few tested diets and some new diets. Diets and notes are added to the google sheet: https://docs.google.com/spreadsheets/d/1F8sJ3P7nR8SWmnzAIJi8jplOddoPQ2Ug9LpPikOqWss/edit#gid=0. Here is a summary of what I found. 

1. There are cases (NE: Diet 7 NRC, Diet 7 NASEM, Diet 15 NASEM) where a ration can be formulated with similar compositions as the input and the estimated intake is reasonable. 
2. There are some other cases (NE: Diet 15 NRC, Diet 35 NRC, Diet 35 NASEM) where forage NDF fails (the constraint should be > 19%) or the estimated DMI is low or the ingredient inclusion rates do not match with the inputs. 

Some further troubleshooting is recommended. Thank you!

tomhuhh:
I have now uploaded the diets that I worked with. Also, in the case of Diet 35 NRC, it fails 84 attempts for a single ration interval. Thanks!

JoeWaddell:
> I have now uploaded the diets that I worked with. Also, in the case of Diet 35 NRC, it fails 84 attempts for a single ration interval. Thanks!

Thanks! For everyone else: Haowen and I met and discussed the reason it's failing, and it's because of the lower limit required for forage NDF: 19% of the diet or more. What's happening is that valid diets that use a high percentage of NDF from concentrate (per a previous discussion with Edward) means that there is less than that specific constraint. 

We should therefore consider modifying this constraint.

Alternatively, this is where the user can decide to set the `milk_reduction_maximum` to a low value, or 0.0, so that there is no reduction in lactation (and consequently no reduction in DMI) due to that single constraint being violated.

JoeWaddell:
From conversations with @KFosterReed and @tomhuhh, some custom diets use a lower percentage of forage NDF than we want in the automated formulation. 

I tried setting the minimum forage NDF% in the constraint file to 15, but ration NRC_35 would still fail. Therefore acceptable outputs required my setting the `milk_reduction_maximum` value to 0.0 (or another small value). Note that this solution would be somewhat difficult to use if there are additional failing constraints.

So instead of making the value in that constraint much lower, in the ongoing Ration Refactor we will move the values in the constraints to the `animal_module_constants` file so that the option is there to change those for specific cases, for advanced users.

I think in the end this branch should be ready for approval and merging, if all others agree! 

Note: I will remove comments and the additional rations used for testing from the code prior to merging.

KFosterReed (approved):

KFosterReed:
LGTM :)

Issue #580, opened 2023-06-19
Move output filters to output dir


PooyaHekmati:
## Expected Behavior
Output filter files should be in the output directory.

## Current Behavior
These files are in the input directory.

## Possible Solution
Move the files and make the needed adjustments in the code.

## Steps to Reproduce
N/A

## Context (Environment)
There are 2 reasons for this:

1. The input directory is responsible to contain the inputs that are needed to run the simulation and the required data for the EEE module. The output description is not an input to the model.
2. Down the road, when users want to share their results, it would be beneficial to share the filters as well. If they are in the same dir, all the user needs to do is share just the output folder and which contains everything.

PooyaHekmati:
This is a p1.

ew3361zh:
Addressed with PR #586

PR #586, opened 2023-06-20
Move output filters to output dir

ew3361zh:
This PR seeks to address issue #580 which wants the output filters directory moved to the output directory (from its current location in the input directory).

# To accomplish this, the PR:
1. Moves the output_filters dir and its example filters to the outputs directory: MASM/output.
2. Adds new location of output_filters to main .gitignore file to ensure it's updated as people add new filters.
3. Updates path argument for OM save_variables method in main.
4. Updates run_rufas method to include output_filters in the list of things to keep if the clear_input flag is true.


PooyaHekmati (approved):

allisterakun (approved):
Looks good.
Passed all test scenarios.
Straight forward and simple.

PR #587, opened 2023-06-20
Clean and polish `Field` and `CropManagement`

Isgotkowitz:
This PR cleans and polishes the `Field` and `CropManagement` module. This includes removing all unused and placeholder methods, as well as removing some todo's and rewriting old docstrings.

- `CropManagement` has had the placeholder methods for grazing and collecting cut yields removed and replaced with TODO's. The place holder tests for these two methods have also been removed.
- `Field` has had the module and `manage_field()` docstrings rewritten to reflect their (more) finalized states. 
- Several old methods that were unused or had been replaced by new methods have been deleted, as well as all their usages and tests.
- All methods besides `manage_field()` and `perform_annual_reset()` have been made private (i.e. prefixed with an underscore)
- `_execute_fertilizer_application()` now accesses mixes in a way that will throw an error if the specified mix is not found. A test has been added to verify correct error handling.
- `manage_field()` has also had a test written, now that it is very close to being finalized.
- With the new tests, the code coverage has reached 100%
- `finish_line_doc.md` has been updated to with progress made by this PR.

### DIFF
--- a/SC_redesign/Crop_and_Soil/crop/crop_management.py
+++ b/SC_redesign/Crop_and_Soil/crop/crop_management.py
@@ -61,13 +61,10 @@ class CropManagement:
         self._record_yield(field_name, field_size, year, day)
         self._transfer_residue(soil_data)
 
-    def graze(self):  # TODO: implement grazing method (SWAT 6:1.3)
-        pass
+    # TODO: implement grazing feature - issue #590
 
     # ---- Sub Methods ----
-    def collect_cut_yield(self, collected_fraction):  # TODO: implement for in-field drying - needed? GitHub Issue #162
-        """collect yield that has been previously cut and left in the field"""
-        pass
+    # TODO: implement management practice for dry-down and collecting cut yields that have been left in the field - #353
 
     def kill(self) -> None:
         """kills the plant, preventing it from growing, and converts all biomass to residue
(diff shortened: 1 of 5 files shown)


matthew7838 (approved):
Nice touch, really helpful for my current PR. Also no comment for this PR.

SC_redesign/Crop_and_Soil/crop/crop_management.py
```diff
@@ -61,13 +61,10 @@ def manage_harvest(self, harvest_op: HarvestOperation, field_name: str, field_si
         self._record_yield(field_name, field_size, year, day)
         self._transfer_residue(soil_data)
 
-    def graze(self):  # TODO: implement grazing method (SWAT 6:1.3)
-        pass
+    # TODO: implement grazing feature - issue #590
```
JoeWaddell, line 64: This is great: making issues out of TODOs is good practice IMHO, since they can linger.

JoeWaddell (approved):
This looks great: the initial comments continue to be a great resource for reviewing, so thanks for keeping those well-documented.

SC_redesign/Crop_and_Soil/crop/crop_management.py
Isgotkowitz: Absolutely. One of the things I have learned since starting is that if a TODO is not connected to an issue and it doesn't say anything about when it will get addressed, then it is a dangling TODO and those are big no-no

PR #593, opened 2023-06-21
Setup field management practices

Isgotkowitz:
This PR implements the method that will take a dictionary of farm management practices (fertilizer application, manure application, and tillage specifications) and turn them into object instances usable by the simulation.

- `_setup_field()` handles the specification for an entire field, and is responsible for grabbing the correct files and calling subroutines to actually setup object instances usable for the simulation. It is not completed in this PR, so a test is not included for it here.
- `_setup_management()` takes a dictionary containing the management practices for a single field, and returns a tuple including a list of fertilizer mixes, a `FertilizerSchedule`, a `ManureSchedule`, and a `TillageSchedule`.
- `_setup_management()` is tested in `test_field_manager.py`
- Progress is reflected in `finish_line_doc.md`


SC_redesign/Crop_and_Soil/manager/field_manager.py
```diff
@@ -11,8 +15,11 @@
 
 
 class FieldManager:
-    def __init__(self, _fields_config: Optional[List[Dict[str, str]]] = None):
+    def __init__(self, fields_config: List[Dict[str, str]]):
         self.fields: List[Field] = []
+        for field in fields_config:
+            field_name, field_config = field.items()
+            self.fields.append(self._setup_field(field_name, field_config))
```
ltp23, line 22: I think the structure of the json input seems off a little bit. Here, it looks like we expect a list of dictionaries, each of which only has a key-value pair. 

Maybe we could change from 

```
fields_config = [
{ "field1": {"config_field1": "...", "config_field2": "..."},
{ "field2": {"config_field1": "...", "config_field2": "..."}
```

to

```
fields_config = [
{
"name": "field1",
"config": {"config_field1": "...", "config_field2": "..."}
},
{
"name": "field2",
"config": {"config_field1": "...", "config_field2": "..."}
}
]
```
And we could break down the config portion even further.

SC_redesign/Crop_and_Soil/manager/field_manager.py
```diff
@@ -102,3 +109,90 @@ def _date_conversion_day(time: Time) -> int:
...
+        Returns
+        -------
+        Field
+            A `Field` instance configured with the specified input data
+
+        """
+        input_directory = Utility.get_base_dir() / 'input'
+
+        management_config = \
+            Utility.read_json_file(input_directory / 'field_management' / field_config['field_management'])
```
ltp23, line 132: This part may need some tweaks. I don't know if we can find a way to move this part to the State class and then receive a config object from there. It is more of a design choice. We should avoid doing file handling here if we can.

SC_redesign/Crop_and_Soil/manager/field_manager.py
```diff
@@ -102,3 +109,90 @@ def _date_conversion_day(time: Time) -> int:
...
+        fertilizer_events = fertilizer_schedule.generate_fertilizer_events()
+        manure_events = manure_schedule.generate_manure_events()
+        tillage_events = tillage_schedule.generate_tillage_events()
+
+        return Field(tillage_events=tillage_events, fertilizer_events=fertilizer_events,
+                     fertilizer_mixes=available_fertilizer_mixes, manure_events=manure_events)
+
+    @staticmethod
+    def _setup_management(field_name: str,
+                          management_config: Dict) -> Tuple[Dict, FertilizerSchedule, ManureSchedule, TillageSchedule]:
```
ltp23, line 145: When there are missing values in the config object, the use of `get()` will return None. I assume you wanted to handle that possibility in other places. You could also provide the defaults here or simply raise errors.

ltp23 (approved):
Thanks Ed. I have some comments, but I'm not sure if they are applicable.

SC_redesign/Crop_and_Soil/manager/field_manager.py
Isgotkowitz: I think that would be a better way to structure the JSONs also. But the way it is currently done is based on the structure of the top-level input JSONs, and I would like to change those as little as possible because we are likely to reevaluate them anyway as part of the development of the `InputManager`.

SC_redesign/Crop_and_Soil/manager/field_manager.py
Isgotkowitz: Again I think that would be better too, but this is just a temporary implementation to mimic the old code so that we can begin testing while making as little change to `State` and `SimulationEngine` as possible. All I/O will eventually be handled by the `InputManager`, so that will eliminate all of these reads.

SC_redesign/Crop_and_Soil/manager/field_manager.py
Isgotkowitz: Good point, I will double check all the uses of `get()` here to make sure `None` is acceptable.

Isgotkowitz:
> Thanks Ed. I have some comments, but I'm not sure if they are applicable.

@ltp23 you made a lot of good points about how the input structure could be better designed here, but because a lot of it will be removed or changed when the `InputManager` is added I am going to keep it mostly the same for now and wait until the `InputManager` is integrated to make it really well designed

matthew7838 (approved):
LGTM, a great PR, nothing to comment on for this one.

PR #594, opened 2023-06-22
Adding a class for Storing Results of Nutrient Requests

ltp23:
This PR introduces the NutrientRequestResults class, which stores the information about the nutrient amounts fulfillable based on the dynamics between what is requested and what is available. The actual calculations will be part of the next PR. 

The class also includes some relevant data validation.

### DIFF
--- a/RUFAS/routines/manure/manure_nutrients/nutrient_request_results.py
+++ b/RUFAS/routines/manure/manure_nutrients/nutrient_request_results.py
@@ -1,0 +1,70 @@
+from __future__ import annotations
+
+import math
+from dataclasses import dataclass, fields
+
+
+@dataclass(frozen=True)
+class NutrientRequestResults:
+    nitrogen: float = 0.0
+    """Amount of manure nitrogen requested that can be fulfilled (kg). Default to 0."""
+
+    phosphorus: float = 0.0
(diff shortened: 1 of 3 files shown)


RUFAS/routines/manure/manure_nutrients/nutrient_request_results.py
```diff
@@ -0,0 +1,72 @@
...
+    dry_matter_fraction: float = 0.0
+    """Fraction of manure that is dry matter, between 0 and 1 (unitless). Default to 0."""
+
+    def __post_init__(self) -> None:
+        """
+        Validate the dataclass fields.
+
+        Returns
+        -------
+        None
```
Isgotkowitz, line 47: Does `None` need to be specified as the return type in the docstring? I ask because if so there are a lot of docstrings in the Crop and Soil module that need to be updated.

Isgotkowitz (approved):
Excellent PR Loi. And kudos on how you test for the correct error message being raised, it is extremely clean and concise

RUFAS/routines/manure/manure_nutrients/nutrient_request_results.py
ltp23: Thanks Ed. I don't know definitively. I've seen both ways. We can update as we go along.

varrma (approved):
Looks good Loi !!

PR #595, opened 2023-06-22
Removing old daily_updates() in animal management class

ltp23:
In this PR, I removed the previous `daily_updates()` in the AnimalManagement class and then renamed `daily_updates2()` to `daily_updates()`. I also included a unit test for it and more docstrings to some of the helper methods.


RUFAS/routines/animal/animal_management.py
```diff
@@ -1699,24 +1607,89 @@ class names as keys and sets of animals as values. There should also be a specia
...
+        """
         graduated_animals = set()
         for animal_class_name in ['heiferIs', 'heiferIIs', 'heiferIIIs', 'cows']:
             graduated_animals.update(animals_snapshot_after_update[animal_class_name]
                                      - animals_snapshot_before_update[animal_class_name])
         for animal in graduated_animals:
             self._add_animal_to_pen_and_id_map(animal, feed, temp)
 
-    def _handle_newly_added_animals(self, new_animals, feed, temp):
+    def _handle_newly_added_animals(self, new_animals, feed, temp) -> None:
```
JoeWaddell, line 1582: This and a few other private functions are missing typing hints (some have the types in the docstrings, though).

JoeWaddell (approved):
Looks good, runs well! 

Just left a comment about typing, and deleted a few other methods that were removed in the same PR. Would be worth double-checking that I didn't remove too much, but I followed your comments in the code and checked that things still run as normal.

RUFAS/routines/animal/animal_management.py
ew3361zh: Agree with Joe on this. I think typing hints would be really helpful where they can be added. I see that the frequently used `animals` variable has a TODO on it for typing but the others I think would be really good to do before merging.

RUFAS/routines/animal/animal_management.py
```diff
@@ -1725,9 +1646,9 @@ def _add_animal_to_pen_and_id_map(self, animal, feed, temp):
                                                  self.phosphorus_concentration_by_animal_class[type(animal)])
         self.animal_to_pen_id_map[animal.id] = pen_with_min_stocking_density.id
 
-    def daily_updates2(self, feed, weather, time):
+    def daily_updates(self, feed, weather, time):
```
ew3361zh, line 1649: Add typing hints here

ew3361zh (approved):
Very nice, Loi. Test suite was working well. Did a brief check to see that it covered everything added and I believe it does but it would be helpful to incorporate that tool Ed was talking about for automating test coverage analysis.

Only comments are around adding typing for some of the parameters in the function definitions.

RUFAS/routines/animal/animal_management.py
ew3361zh: Another thought is you could create an issue for someone to go through and do the type hints. Maybe another good first issue for one of the new dev team members?

JoeWaddell: I think adding type hints would be great task for new members, because it will give them some experience wading into the code to figure out what's what. Likewise for turning `Args` into `Parameters` in many functions that have yet to be refactored.

ltp23: Thanks Joe and Niko, the type hints for those classes cause import errors at least for this animal_management.py file. So, I have been skipping on them when adding type hints.

ew3361zh: I forgot about that issue with certain animal module files. Thanks, Loi.

PR #596, opened 2023-06-22
Crop schedule input processing in `FieldManager`

Isgotkowitz:
This PR adds the functionality for creating a series of `CropSchedule`s based on input dictionary of crop management practices. It also modifies the current crop management input files to be compatible with the new crop and soil module.

- `_setup_field()` now gets the crop management specifications from the input file and passes them to the subroutine that creates the `CropSchedule`s.
- `_setup_crop_schedules()` takes a dictionary of crop scheduling specifications and creates `CropSchedule`s based on them.
- There are 2 tests for `_setup_crop_schedules()` - 1 tests it with all crop management input files currently in RuFaS and makes sure none raise any errors, and the other checks specifically for correctness.
- Progress made by this PR is reflected in `finish_line_doc.md`.
-  All 8 crop input JSONs have been modified to add missing parameters needed and so that their contents can be put directly into `CropSchedule`s. This includes adding the crop species to be planted/harvested, changing dates to be contained in lists, and adding harvest years and operations.


SC_redesign/Crop_and_Soil/tests_SC/manager_tests/test_field_manager.py
```diff
@@ -162,3 +164,72 @@ def test_setup_management(field_name: str, config: Dict) -> None:
...
+    "testing_rotation.json"
+])
+def test_setup_crop_schedules(crop_input_file_name: str) -> None:
+    """Tests that the crop schedule setup method is able to correctly parse all the currently available crop
+        datasets."""
+    input_directory = Utility.get_base_dir() / 'input'
+    crops_config = Utility.read_json_file(input_directory / 'crop' / crop_input_file_name)
+    crop_specifications = crops_config.get("crops")
+    FieldManager._setup_crop_schedules(crop_specifications)
+    assert True
```
matthew7838, line 186: Should there be a certain condition to assert?

matthew7838 (approved):
Don't see major problems, just one question to double check about the testing.

SC_redesign/Crop_and_Soil/tests_SC/manager_tests/test_field_manager.py
Isgotkowitz: Ideally yes - I would have a list of `CropSchedule`s as the expected output for each input file. But because the input files can be very large and involved, I decided it would be more efficient to just check that all the modified input files can be parsed without crashing, and test for correctness with `test_correct_crop_schedule_setup()`. Between these two tests, I am confident that FieldManager will be able to handle the real inputs and produce the correct list of `CropSchedule`s from them.

ltp23: This seems like it should be part of a dedicated module for validating user input files. I don't think we have given much thought to the design of that kind of thing yet.

ltp23 (approved):
The PR looks good to me! Thanks Ed.

SC_redesign/Crop_and_Soil/tests_SC/manager_tests/test_field_manager.py
Isgotkowitz: Yes, I assume that each module will be tapped to write a set of validation rules for their inputs, but I am not sure what stage that will happen at in the development of `InputManager`

PR #599, opened 2023-06-23
Final setup methods in `FieldManager`

Isgotkowitz:
This PR implements the last pieces needed for `FieldManager` to setup `Field` instances by creating `Soil` and `FieldData` objects.

- `_setup_field()` now calls the subroutine for creating `Soil` objects, and creates `FieldData` objects with values taken from the various configuration files.
- `_setup_soil()` takes a dictionary of different soil profile parameters, including a list of different soil layer configurations. The layer configurations are passed on to a subroutine that creates `LayerData` objects from them, and passes the `LayerData` objects along with the other parameters to a method that creates a `SoilData` instance from them. That `SoilData` object is used to create a `Soil` instance which is then returned.
- `_setup_soil_layer()` takes a dictionary with specifications for a `LayerData` object and some other parameters that are necessary for the construction of a `LayerData` object, and creates and returns an instance based on those parameters.
- `test_field_manager.py` includes tests for `_setup_field()`, `_setup_soil()`, and `_setup_soil_layer()`
- `LayerData` is now capable of taking an initial amount of ammonium in a soil layer (the mechanism is exactly the same for taking an initial nitrate level). This change helps accommodate some (but not all) of the input documents, and some updates have been made because of this in `layer_data.py`, `test_leaching_runoff_erosion.py`, and `test_soil_data_and_config_factory.py`.
- Progress made by this PR is reflected in `finish_line_doc.md`.


matthew7838 (approved):
LGTM, great tests.

SC_redesign/Crop_and_Soil/manager/field_manager.py
```diff
@@ -239,3 +250,115 @@ def _setup_crop_schedules(crop_config: Dict) -> List[CropSchedule]:
...
+        soil_layers_config.sort(key=lambda x: x.get("bottom_depth"))
+        soil_layers = []
+        for index, layer_config in enumerate(soil_layers_config):
+            if index == 0:
+                new_layer = FieldManager._setup_soil_layer(field_size, 0.0, sand_content, silt_content, residue,
+                                                           nitrogen_mineralization_rate, layer_config)
+            else:
+                new_layer = FieldManager._setup_soil_layer(field_size, soil_layers[-1].bottom_depth, sand_content,
+                                                           silt_content, residue, nitrogen_mineralization_rate,
+                                                           layer_config)
```
ltp23: Maybe combine these two branches into one, since they only differ by one argument.

SC_redesign/Crop_and_Soil/manager/field_manager.py
```diff
@@ -239,3 +250,115 @@ def _setup_crop_schedules(crop_config: Dict) -> List[CropSchedule]:
...
+
+        config_dictionary = {}
+
+        config_dictionary["second_moisture_condition_parameter"] = soil_config.get("CN2")
+        config_dictionary["average_subbasin_slope"] = soil_config.get("field_slope")
+        config_dictionary["slope_length"] = soil_config.get("slope_length")
+        config_dictionary["manning"] = soil_config.get("manning")
+        config_dictionary["albedo"] = soil_config.get("soil_albedo")
+        config_dictionary["cover_type"] = soil_config.get("soil_cover_type")
+        config_dictionary["soil_layers"] = soil_layers
```
ltp23: Maybe to avoid some repetition, you could write it as 

```
soild_data_config = {
  'second_moisture_condition_paramenter': soil_config.get('CN2'),
  'average_subbasin_slope': soil_config.get('field_slope'),
  ...
}
```

SC_redesign/Crop_and_Soil/manager/field_manager.py
```diff
@@ -239,3 +250,115 @@ def _setup_crop_schedules(crop_config: Dict) -> List[CropSchedule]:
...
+        -----
+        Whoever wrote the JSON's for soil profile inputs wrote "N03" (the digit zero) instead of "NO3" (the letter 'O'),
+        and that is why it is written with a zero and not the letter here.
+
+        """
+        config_dictionary = {}
+
+        config_dictionary["bottom_depth"] = layer_config["bottom_depth"]
+        config_dictionary["wilting_point_water_concentration"] = layer_config.get("wilting_point")
+        config_dictionary["field_capacity_water_concentration"] = layer_config.get("field_capacity")
```
ltp23: It looks like when you use square brackets around a field, you assume that that field should be there. When you use `get()`, it means optional and can return None. I think you should raise an error every time a required field is absent.

SC_redesign/Crop_and_Soil/tests_SC/manager_tests/test_field_manager.py
```diff
@@ -233,3 +238,103 @@ def test_correct_crop_schedule_setup(crop_schedule_config: Dict, expected: List[
...
+@pytest.mark.parametrize("field_name,field_config", [
+    ("field_1", {"soil": "ARL_soil.json", "crop": "ARL_rotation.json",
+                 "field_management": "ARL_field_management.json"}),
+    ("field_2", {"soil": "barnyard_soil.json", "crop": "corn_scheduled_rotation.json",
+                 "field_management": "barnyard_scheduled_field_management.json"})
+])
+def test_setup_field(field_name: str, field_config: Dict[str, str]) -> None:
+    """Tests that a Field instance is correctly initialized with a given input configuration."""
+    Utility.get_base_dir = MagicMock()
+    Utility.read_json_file = MagicMock(return_value={"field_size": 1.0, "initial_residue": 0.0, "latitude": 40.0})
```
ltp23: I would try to minimize doing direct assignments like these and use patching instead because they are not automatically reverted to their original states unless you do it manually (ie., save the original state, reassign with the new state, reset with the original state). Though I assume the tests still run fine, they are not entirely self-contained.

ltp23 (approved):
Good job, Ed! Thank you.

SC_redesign/Crop_and_Soil/manager/field_manager.py
Isgotkowitz: Good call

SC_redesign/Crop_and_Soil/manager/field_manager.py
Isgotkowitz: Good call

SC_redesign/Crop_and_Soil/tests_SC/manager_tests/test_field_manager.py
Isgotkowitz: You're right, especially with functions like those, spillover into other tests when `SC_redesign` would be really bad.

PR #600, opened 2023-06-26
Updating housing ammonia emission calculations in manure handlers

ltp23:
This PR updates the calculations of housing ammonia emissions in the manure handlers.

The main changes happen in the two files manure_handler_classes.py and gas_emission.py.

### DIFF
--- a/RUFAS/routines/manure/constants/gas_emission_constants.py
+++ b/RUFAS/routines/manure/constants/gas_emission_constants.py
@@ -47,3 +47,4 @@ class GasEmissionConstants:
     SPECIFIC_GROWTH_RATE = 0.637  # micrometers
 
     DEFAULT_HOUSING_SPECIFIC_CONSTANT = 260.0  # s/m
+    DEFAULT_PH_FOR_HOUSING_AMMONIA = 7.5
--- a/RUFAS/routines/manure/constants/gas_emission_constants.py
+++ b/RUFAS/routines/manure/constants/gas_emission_constants.py
@@ -27,7 +27,7 @@ class GasEmissionConstants:
     MCF = 0.79
     """Methane conversion factor (unitless)."""
 
-    MS = 1.0
+    MS = 0.9
     """#TODO"""
 
     METHANE_FACTOR = 0.67
--- a/RUFAS/routines/manure/constants/gas_emission_constants.py
+++ b/RUFAS/routines/manure/constants/gas_emission_constants.py
@@ -1,50 +1,86 @@
 class GasEmissionConstants:
-    """Constants for gas emission calculations.
+    """Constants used in gas emission calculations."""
 
-    Attributes:
-        b1: rate correcting factor, dimensionless.
-        b2: rate correcting factor, dimensionless.
-        lnA: natural log of Arrhenius constant, g CH4/kg VS-h.
-        E: activation energy, J/mol.
-        R: gas constant, J/mol-K.
+    b1 = 1.0
+    """Rate correcting factor (unitless)."""
 
-        Bo: achievable emission of CH4 during anaerobic digestion, kg CH4/kg VS.
-        MCF: TODO
-        MS: TODO
-        METHANE_FACTOR: TODO.
-        METHANE_ENERGY_DENSITY: methane energy density, MJ/kg.
-        METHANE_DENSITY: methane density, kg/m^3.
-        METHANE_POTENTIAL_Go: methane potential, mL/g VS.
-        POTENTIAL_METHANE_YIELD_OF_MANURE: potential methane yield of manure, kg CH4/kg VS.
+    b2 = 0.01
+    """Rate correcting factor (unitless)."""
 
-        DEFAULT_VOLATILE_SOLIDS_FRACTION: default volatile solids fraction, dimensionless.
-        CHEN_HASHIMOTO_KINETIC_CONSTANT_KCH: Chen-Hashimoto kinetic constant, dimensionless.
-        SPECIFIC_GROWTH_RATE: specific growth rate, micrometers.
+    lnA = 43.33
+    """Natural log of Arrhenius constant (unitless)."""
 
-        DEFAULT_HOUSING_SPECIFIC_CONSTANT: default housing specific constant, s/m.
-
+    E = 112700.0
     """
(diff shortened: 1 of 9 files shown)


RUFAS/routines/manure/constants/gas_emission_constants.py
```diff
@@ -1,49 +1,86 @@
...
+    minimum energy that must be available to molecules for a reaction to occur.
+    """
+
+    R = 8.314
+    """
+    The ideal gas constant (J/mol :math:`\cdot` K).
+    """
+
+    Bo = 0.24
+    """Achievable emission of :math:`CH_4` during anaerobic digestion (#TODO: units?)."""
```
varrma: Bo = achievable emission of CH4 during anaerobic digestion, kg CH4/kg VS

RUFAS/routines/manure/constants/gas_emission_constants.py
```diff
@@ -1,49 +1,86 @@
...
+
     MCF = 0.79
+    """Methane conversion factor (unitless)."""
+
     MS = 0.9
-    METHANE_FACTOR = 0.67  # TODO: use a better name
+    """#TODO"""
+
+    METHANE_FACTOR = 0.67
+    """#TODO"""
```
varrma: 0.67 = conversion factor of m3 CH4 to kg CH4

varrma (approved):
Looks good Loi, couple of comments about the variable description and units.

RUFAS/routines/manure/constants/gas_emission_constants.py
```diff
@@ -1,49 +1,86 @@
...
+
+    R = 8.314
+    """
+    The ideal gas constant (J/mol :math:`\cdot` K).
+    """
+
+    Bo = 0.24
+    """Achievable emission of :math:`CH_4` during anaerobic digestion (#TODO: units?)."""
+
     MCF = 0.79
```
JoeWaddell: Should these names be extended for clarity, e.g. for this case `methane_conversion_factor`, or does that lead to excessive line length in some areas?

RUFAS/routines/manure/gas_emissions/gas_emissions.py
```diff
@@ -194,6 +194,79 @@ def calc_housing_carbon_dioxide_emission(cls, num_animals: int, barn_area: float
...
+            Number of animals in the barn (unitless).
+        barn_area : float
+            Barn area per animal based on housing type (:math:`m^2`).
+        urine_total_ammoniacal_nitrogen : float
+            Total ammoniacal nitrogen in urine (kg).
+        urine : float
+            Amount of urine produced by animals in the barn (kg).
+        temp : float
+            Current barn temperature (:math:`^{\circ}C`).
+        pH : float, optional
```
JoeWaddell, line 298: Since these are labeled optional here, should the typing hint in the args be something like `ph: Optional[float] = GasEmissionConstants.DEFAULT_PH_FOR_HOUSING_AMMONIA`? Or is it simply a style choice?

RUFAS/routines/manure/gas_emissions/gas_emissions.py
```diff
@@ -194,6 +194,79 @@ def calc_housing_carbon_dioxide_emission(cls, num_animals: int, barn_area: float
...
+
+        total_barn_area = num_animals * barn_area
+        TAN = urine_total_ammoniacal_nitrogen / total_barn_area
+        p = ManureConstants.MANURE_DENSITY  # kg/m^3
+        c = GeneralConstants.SECONDS_PER_DAY  # s/day
+        tempK = cls._convert_temperature_celsius_to_kelvin(temp)  # K
+        r = cls._calc_barn_resistance(temp, hsc)
+        M = urine / total_barn_area  # manure per area of exposed surface, kg/m^2
+        Q = cls._calc_Q(tempK, pH)
+        loss = (TAN * c * p) / (r * M * Q)
```
JoeWaddell: Should these variables be given longer names, or listed with their definitions in a Notes section in the docstring? Just thinking ahead when folks might later tweak or add to these equations. If they come directly from the literature, maybe just adding that to a References section would be a good thing for documentation

RUFAS/routines/manure/pen_manure/pen_manure.py
```diff
@@ -77,6 +77,13 @@ def __post_init__(self):
         """Performs any necessary unit conversion after initialization."""
         self.manure_volume = self.manure_mass / ManureConstants.MANURE_DENSITY
 
+        # Zero out any negative field
+        # TODO: This is a temporary fix. Need to find out why negative values are being generated
```
JoeWaddell, line 81: Could you make an issue with what inputs/conditions generate a negative value?

JoeWaddell (changes requested):
Looks good to me: my only comments overlap a little with those of Varma, as they're mostly about documentation.

RUFAS/routines/manure/constants/gas_emission_constants.py
ltp23: Thanks Joe.  I can change MCF to METHANE_CONVERSION_FACTOR.

RUFAS/routines/manure/gas_emissions/gas_emissions.py
ltp23: That's a good point, Joe. But conveniently, when a parameter is written like that`param=default_value`, the parameter will assume the type of `default_value_type | None` (alternative notation for Optional). In this case, it would be `float | None`.

RUFAS/routines/manure/gas_emissions/gas_emissions.py
ltp23: I think a good trade-off would be this: In the Notes section of the docstring, I would state the original formula as presented in the literature, but then in the code, I would replace the symbols in the formula with descriptive names.

RUFAS/routines/manure/pen_manure/pen_manure.py
ltp23: Ok Joe. I made a git issue for this ([#609](https://github.com/RuminantFarmSystems/MASM/issues/609)).

JoeWaddell (approved):
Thanks for the informative replies and relevant modifications: looks great!

PR #602, opened 2023-06-26
Calculate and use peak runoff rate in `SoilErosion`

Isgotkowitz:
This PR implements the calculation of the peak runoff rate value, used to determine how much soil erosion occurs during a rainfall event. Before this, the peak runoff rate was kept as a value to be input by users.

- `SoilErosion` has 6 new `@staticmethod`s that handle the calculation of the peak runoff rate, all have tests in `test_soil_erosion.py`.
- The main routine in `SoilErosion`, `erode()` now calculates the peak runoff rate instead of pulling it from `SoilData` and takes the rainfall amount for the current day. These changes are tested as well. Please note that the docstring was added to, but is still not formatted correctly - this will be addressed at a later date. The test for `erode()` has been updated as well.
- `peak_runoff_rate` has been removed from `SoilData`, and tests have been adjusted accordingly.
- Calls to `erode` have been updated in `Field` and `Soil`.
- A new constant has been added to the Crop and Soil constants file.


SC_redesign/Crop_and_Soil/soil/soil_erosion.py
```diff
@@ -32,6 +34,7 @@ def erode(self, field_size: float, minimum_cover_management_factor: float, surfa
             minimum_cover_management_factor: minimum value for cover and management factor for water erosion applicable
                 to land cover/plant (unitless)
             surface_residue: amount of residue on the soil surface (kg per hectare)
+            rainfall: amount of rain that fell on the field on the current day (mm).
```
JoeWaddell, line 37: Could be a good time to bring this (and other) method(s) up to spec re: Sphinx documentation, like you did in `_determine_peak_runoff_rate` and others - wonderful job on those!

JoeWaddell (approved):
This all looks great! Nice job on the docstrings, and my only comment was that you could update the older ones: but that could also be an entirely separate PR.

Isgotkowitz:
> This all looks great! Nice job on the docstrings, and my only comment was that you could update the older ones: but that could also be an entirely separate PR.

@JoeWaddell thank you. I decided against updating older ones in this PR partly because it would violate the single-responsibility principle, but also because if I updated one I would have the itch to update all of them in one file, and I didn't want this PR growing any larger or more complex than it already is.

SC_redesign/Crop_and_Soil/soil/soil_erosion.py
```diff
@@ -265,6 +270,205 @@ def _determine_coarse_fragment_factor(percent_rock_content: float) -> float:
...
+        surface_runoff : float
+            Amount of rainfall that did not infiltrate into the soil on the current day (mm).
+        rainfall : float
+            Amount of rainfall on the current day (mm).
+        slope_length : float
+            Length of the subbasin slope (m).
+        manning : float
+            Manning roughness coefficient for the subbasin (unitless).
+        average_subbasin_slope : float
+            Average slope length of the subbasin expressed as rise over run (m / m).
```
matthew7838: m/m is a bit confusing, not sure if there's a better way to specify this unit.

SC_redesign/Crop_and_Soil/soil/soil_erosion.py
```diff
@@ -265,6 +270,205 @@ def _determine_coarse_fragment_factor(percent_rock_content: float) -> float:
...
+        Calculates the time of concentration for the subbasin.
+
+        Parameters
+        ----------
+        slope_length : float
+            Length of the subbasin slope (m).
+        manning : float
+            Manning roughness coefficient for the subbasin (unitless).
+        average_subbasin_slope : float
+            Average slope length of the subbasin expressed as rise over run (m / m).
```
matthew7838, line 390: same

matthew7838 (approved):
Some concerns on the unit, otherwise all LGTM. Noice

SC_redesign/Crop_and_Soil/soil/soil_erosion.py
Isgotkowitz: There is, thanks Matthew

Issue #603, opened 2023-06-27
Create IM file and class


PooyaHekmati:
create input_manager.py and declare class InputManager in it. Make sure it is a singleton and it has test coverage for this. Take OM as an example.

ew3361zh:
Resolved in PR #606

PR #606, opened 2023-06-28
Create Input Manager file and class

ew3361zh:
This PR seeks to address issue #603 to create an input_manager.py file and declare the InputManager class in it.

Unit testing has been added to ensure it is a singleton.

This is the starting point for the Input Manager.

### DIFF
--- a/RUFAS/input_manager.py
+++ b/RUFAS/input_manager.py
@@ -7,7 +7,7 @@ class InputManager:
     """
     Input Manager class responsible for loading, validating, and providing access to input data.
     """
-    __instance = Optional['InputManager'] = None
+    __instance = None
 
     def __new__(cls):
         if not hasattr(cls, "instance"):
@@ -17,5 +17,3 @@ class InputManager:
     def __init__(self) -> None:
         if InputManager.__instance is None:
             InputManager.__instance = self
-        self.__pool: Dict[str, Any] = {}
-        self.__metadata: Dict[str, Any] = {}
--- a/RUFAS/input_manager.py
+++ b/RUFAS/input_manager.py
@@ -1,0 +1,21 @@
+# !/usr/bin/env python3
+
+from typing import Any, Dict, Optional
+
+
+class InputManager:
+    """
+    Input Manager class responsible for loading, validating, and providing access to input data.
+    """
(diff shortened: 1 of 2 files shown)


PooyaHekmati (approved):
LGTM
I think the test cases for IM will be large enough to deserve a dedicated file. Let's create a test file for it in the next PR.

RUFAS/input_manager.py
```diff
@@ -0,0 +1,19 @@
...
+
+class InputManager:
+    """
+    Input Manager class responsible for loading, validating, and providing access to input data.
+    """
+    __instance = None
+
+    def __new__(cls):
+        if not hasattr(cls, "instance"):
+            cls.instance = super(InputManager, cls).__new__(cls)
```
Isgotkowitz, line 14: Asking because I am curious, what class is `super()` referring to here? `InputManager` doesn't appear to be a child class

RUFAS/input_manager.py
```diff
@@ -0,0 +1,19 @@
+# !/usr/bin/env python3
+
+from typing import Any, Dict, Optional
```
Isgotkowitz, line 3: Any, Dict, and Optional are unused, throws a flake8 error for me when I run the linting manually. That said, I am sure they will be used in subsequent PRs so leaving them in now seems inconsequential.

Isgotkowitz (approved):
Nice job Niko, a very clean start to the `InputManager`. Two comments, one about import statements and one about `super` that is just because I am curious. Both are take it or leave it.

RUFAS/input_manager.py
ew3361zh: Good catch, yes I had those added for what's coming in the next PR but then opted to move that stuff out of this PR. I may leave it in as it's going to be part of the next PR.

RUFAS/input_manager.py
ew3361zh: This is part of the Singleton design that we used with OM and that I found in pretty much every article I looked at for the Singleton pattern in Python. This was the best explanation I found of why it is used:

https://stackoverflow.com/questions/15522162/usage-of-super-in-singleton

PR #610, opened 2023-06-28
Add timestamps to the OutputManager's info_map data when adding logs, errors, and warnings.

### LINKED ISSUE #598: Add timestamp to logs, warnings, and errors in the Output Manager
We don't have a way to follow and track the sequence of events.

In functions `add_log`, `add_warning`, and `add_error`, add a timestamp entry in the `info_map' before calling the `_add_to_pool`.

1. We could move the timestamp adding logic to `_add_to_pool` and avoid repetition, but that will add timestamps for variables as well. I think that would be unnecessary and TMI.
2. We could let the callers of  `add_log`, `add_warning`, and `add_error` to add their timestamps if they want. But that would lead to too many changes across the codebase. 

**Additional context**

michaelRichards99:
Added timestamp info to the OutputManager when adding logs, errors, and warnings. Consolidated file timestamp creation to use the same method. Added checks to the existing unit tests to check that the timestamp information is present.

Fixes #598.


RUFAS/output_manager.py
```diff
@@ -187,9 +189,21 @@ def add_error(self, name: str, msg: str, info_map: Dict[str, Any]) -> None:
         info_map["suffix"] : str, optional
             If present, gets appended to the key
         """
+        self._add_timestamp_to_info_map(info_map)
         key = self._generate_key(name, info_map)
         self._add_to_pool(self.errors_pool, key, msg, info_map)
 
+    def _get_timestamp(self, include_millis: bool = False) -> str:
+        """Produce the current system time as a timestamp string."""
+        base_timestamp_str: str = "%d-%b-%Y_%a_%H-%M-%S"
```
PooyaHekmati, line 217: Can you add an example in the docstring to show what this looks like?

RUFAS/output_manager.py
```diff
@@ -187,9 +189,21 @@ def add_error(self, name: str, msg: str, info_map: Dict[str, Any]) -> None:
         info_map["suffix"] : str, optional
             If present, gets appended to the key
         """
+        self._add_timestamp_to_info_map(info_map)
         key = self._generate_key(name, info_map)
         self._add_to_pool(self.errors_pool, key, msg, info_map)
 
+    def _get_timestamp(self, include_millis: bool = False) -> str:
+        """Produce the current system time as a timestamp string."""
```
```suggestion
        """Produces the current system time as a timestamp string."""
```

RUFAS/output_manager.py
```diff
@@ -187,9 +189,21 @@ def add_error(self, name: str, msg: str, info_map: Dict[str, Any]) -> None:
...
+        base_timestamp_str: str = "%d-%b-%Y_%a_%H-%M-%S"
+        timestamp_format_string: str = f"{base_timestamp_str}.%f" if include_millis else base_timestamp_str
+        return datetime.datetime.now().strftime(timestamp_format_string)
+
+    def _add_timestamp_to_info_map(self, info_map: Dict[str, Union[str, bool]]) -> None:
+        """Add the current system time to the info_map."""
```
```suggestion
        """Adds the current system time to the info_map."""
```

RUFAS/output_manager.py
```diff
@@ -187,9 +189,21 @@ def add_error(self, name: str, msg: str, info_map: Dict[str, Any]) -> None:
...
         key = self._generate_key(name, info_map)
         self._add_to_pool(self.errors_pool, key, msg, info_map)
 
+    def _get_timestamp(self, include_millis: bool = False) -> str:
+        """Produce the current system time as a timestamp string."""
+        base_timestamp_str: str = "%d-%b-%Y_%a_%H-%M-%S"
+        timestamp_format_string: str = f"{base_timestamp_str}.%f" if include_millis else base_timestamp_str
+        return datetime.datetime.now().strftime(timestamp_format_string)
+
+    def _add_timestamp_to_info_map(self, info_map: Dict[str, Union[str, bool]]) -> None:
```
```suggestion
    def _add_timestamp_to_info_map(self, info_map: Dict[str, Any]) -> None:
```
Why Union[str, bool]?

RUFAS/output_manager.py
```diff
@@ -310,7 +324,7 @@ def _generate_file_name(self, base_name: str, extension: str = "json") -> str:
         """
         Returns a file name using the given base_name and timestamp.
         """
-        timestamp = time.strftime(r"%d-%b-%Y_%a_%H-%M-%S", time.localtime())
+        timestamp: str = self._get_timestamp()
```
```suggestion
        timestamp: str = self._get_timestamp(include_millis=False)
```

tests/test_misc.py
```diff
@@ -368,6 +368,8 @@ def test_add_error(
     mock_output_manager.add_error(name, value, info_map)
 
     mock_output_manager._generate_key.assert_called_once_with(name, info_map)
+
+    assert info_map["timestamp"] is not None
```
PooyaHekmati: Let's also check for the correctness of the value. You'd need to mock datetime.datetime.now() and manually set its return value.

tests/test_misc.py
```diff
@@ -395,6 +396,8 @@ def test_add_warning(
     mock_output_manager.add_warning(name, value, info_map)
 
     mock_output_manager._generate_key.assert_called_once_with(name, info_map)
+
+    assert info_map["timestamp"] is not None
```
PooyaHekmati: Same

tests/test_misc.py
```diff
@@ -422,6 +424,8 @@ def test_add_log(
     mock_output_manager.add_log(name, value, info_map)
 
     mock_output_manager._generate_key.assert_called_once_with(name, info_map)
+
+    assert info_map["timestamp"] is not None
```
PooyaHekmati: Same

RUFAS/output_manager.py
PooyaHekmati: Maybe this function is overkill and not needed?

RUFAS/output_manager.py
```diff
@@ -160,6 +161,7 @@ def add_warning(self, name: str, msg: str, info_map: Dict[str, Any]) -> None:
         info_map["suffix"] : str, optional
             If present, gets appended to the key
         """
+        self._add_timestamp_to_info_map(info_map)
```
```suggestion
        info_map["timestamp"] = self._get_timestamp(include_millis=True)
```

PooyaHekmati (changes requested):
Nice work! It's good to have functions with clear responsibilities. The newly developed functions need to have test cases as well.

RUFAS/output_manager.py
michaelRichards99: done

RUFAS/output_manager.py
michaelRichards99: yeah it's not really necessary. removed.

tests/test_misc.py
michaelRichards99: done.

tests/test_misc.py
michaelRichards99: done.

tests/test_misc.py
michaelRichards99: It's tricky to directly mock python built-ins such as as DateTime without adding a new 3rd party library or manually mocking with some sort of reflection. I've mocked _get_timestamp from the OutputManager class instead.

RUFAS/output_manager.py

RUFAS/output_manager.py
michaelRichards99: I copied the type from another method in the OutputManager. I removed the method completely so no longer applies.

RUFAS/output_manager.py
michaelRichards99: method removed.

RUFAS/output_manager.py

RUFAS/output_manager.py
michaelRichards99: done.

PooyaHekmati (approved):
[LGTM](https://lgtm.lol/i/718)

RUFAS/output_manager.py
```diff
@@ -187,9 +189,25 @@ def add_error(self, name: str, msg: str, info_map: Dict[str, Any]) -> None:
...
         """
+        info_map["timestamp"] = self._get_timestamp(include_millis=True)
         key = self._generate_key(name, info_map)
         self._add_to_pool(self.errors_pool, key, msg, info_map)
 
+    def _get_timestamp(self, include_millis: bool = False) -> str:
+        """
+        Produces the current system time as a timestamp string.
+
+        Example
```
Isgotkowitz: There should be `Parameters` and `Returns` sections here

RUFAS/output_manager.py
```diff
@@ -187,9 +189,25 @@ def add_error(self, name: str, msg: str, info_map: Dict[str, Any]) -> None:
...
+        ------
+            >>> self._get_timestamp(include_millis=True)
+            28-Jun-2023_Wed_15-48-21.406585
+            >>> self._get_timestamp(include_millis=False)
+            28-Jun-2023_Wed_15-48-21
```
```suggestion
        Examples
        --------
        >>> self._get_timestamp(include_millis=True)
        28-Jun-2023_Wed_15-48-21.406585
        >>> self._get_timestamp(include_millis=False)
        28-Jun-2023_Wed_15-48-21
```

Isgotkowitz (changes requested):
Nice PR Michael, looks very good overall. First comment is on some missing sections in the `_get_timestamp()` docstring, and the second has some minor formatting corrections to the `Examples` section of the same docstring.

RUFAS/output_manager.py
michaelRichards99: Done.

RUFAS/output_manager.py
michaelRichards99: Done.

Isgotkowitz (approved):
Nice

PR #611, opened 2023-06-29
End-to-End testing on the redesigned Crop and Soil module

### LINKED ISSUE #361: Update Soil and Crop Input
## Problem Summary
The redesign of the soil and crop module has/will have led to a substantially different set of expected input variables from the user. That means that the input system (i.e., the input json files, code to convert them into objects, and the resulting data/config objects themselves) will need to be re-written. 

This should be one of the final steps in the soil redesign process.

## Details: 
Each of the inputs required by the crop and soil module should be contained within model configuration classes. The attributes of these classes would specify the starting conditions of the model and the way the simulations progress. 
For each field (or management unit) simulated, we'll need the following information at the least:
* Field specification:
    - Field size: how large is the field (or management unit) in terms of area?
    - Topography: how much of a slope exists (on average)?
    - Location: approximately where is the field (or management unit) located on the globe? This will allow us to acquire 
    weather (predictions?) and soil type/composition data.
    - subdivisions: an optional number of fields or other subdivisions that a single management unit is divided into.
* Crop management:
    - planting schedules: which crop species are planted on which years? are there repeated rotation patterns?
    - harvest schedules: should crops be cut on a schedule or should harvest dates be based on optimal growth conditions? 
    If a schedule is used, harvest dates are needed for each unique component of the rotation (i.e., one per year)
    - harvest index override: an optional fraction that manually specifies what proportion of plant is targeted for harvest, if 
    provided this value is used in place of the internally calculated harvest index.
    - harvest efficiency: an optional fraction that manually specifies how efficient the harvest operation actually is at removing 
    the targeted amount.
* Soil management:
    - Soil type: the general type of soil to be used (from a list of supported options) this specification will pull a 
    pre-made soil configuration (with relevant composition attributes) from our database to use in the field. 
    - Soil composition: as an alternative to soil type, advanced options should allow user-defined soil configuration. 
    - Tillage: what type of tillage method will be used on the field? How much mixing will occur? On what date(s) should 
    tillage occur?
    - Amendment: what type of amendments (manure, fertilizer) will be added to the soil? by what method will they be 
    applied? How much? What is the nutrient composition?

### LINKED ISSUE #442: Finish FieldManager to connect and execute full SC simulation/routines
This should be one of the last tasks completed for v1.0 of SC_redesign before merging into master. 

For details, see [the SC redesign design doc](SC_redesign/docs_SC/Admin/design-document_SC-redesign.md), the 
[roadmap](SC_redesign/docs_SC/Admin/road-map_soil-and-crop.md), and possibly [functionality requirements](SC_redesign/docs_SC/Admin/functionality-requirements.md).

Objectives: 
* [ ] Finish/finalize the `Field` class (started in #346), which should run all the daily routines for a given agricultural field. 
* [ ] Create a `FieldManager` class that handles the entire SC simulation through its main method. This class/method should accept a `SCInput` class containing all configuration specifications provided by the 'user'. 
* [ ] Check that all SC methods are being called at the right times and producing reasonable output.
* [ ] Be sure that the *general* functionality of the old code in routines/field/field_management/ is in `Field`.
* [ ] Be sure that the *general*  functionality of the old in routines/field/field.py is in `FieldManager`  

## Associated issues
It may be useful to complete these similar issues in concert with this one:
* #361 
* #159

Isgotkowitz:
This is an extensive PR that addresses bugs which prevent full RuFaS simulations from running. Subsequent PRs will fix issues causing outputs to be whacky.

- `crop_data.py` has been modified so that initial values are valid and not set to `None`. `test_heat_units.py` reflects this change.
- `nitrogen_incorporation.py` 
  - Has had a parameter renamed to be correct. 
  - Has had the near mature nutrient fraction removed as an argument, and simply uses a SWAT assumption instead. The test for this method reflects this.
  - Now takes up no nutrients when the root depth is zero. The test for this method reflects this also.
- `phosphorus_incorporation.py`, like nitrogen incorporation, has had a parameter renamed to be correct and now uses the same SWAT assumption as nitrogen incorporation. These changes are also reflected in the method tests.
- `field.py`
  - The fertilizer application method checks that nitrogen and phosphorus are actually being applied, and that the optimal fertilizer mix will actually include fertilizer with it.
  - The manure application method stand-in has been corrected so it is a valid call.
  - `manage_harvest()` is called correctly now.
  - `is_mature` and `is_dormant` are used instead of `in_growing_season`
  - a division by zero error is avoid by setting average weighted transpiration to 0 when there is no transpiration.
  - `SoilData` is passed instead of `Soil` (as is correct).
  - soil evaporation and sublimation are automatically set to 0 when there is no evaporative demand.
  - these changes are reflected in their individual tests.
- `manure_application.py` has had a call changed to a subroutine to included a missing parameter, and the test now reflects that parameter being included.
- `carbon_cycle.py` has had method calls corrected.
- `decomposition.py` has had a weird bug corrected, caused by syntax intermittently producing complex numbers. Updated method is tested as well.
- `layer_data.py` has had defaults changed from `None` to `0.0`
- `soluble_phosphorus.py` catches OverflowErrors and handles them gracefully now.
- `test_field_manager.py` has a test for `create_current_weather()` now and mocks use in other tests - the method implementation was stacked under this PR.
- inputs
  - `barnyard.json` now uses `no_crop.json` - as it should because that simulation should have no crops.
  - extra values in `multi_crop_field_management.json` have been removed.
  - `testing_field_management.json` has correct fertilizer mix now.
  - `multi_crop.json`, `swat.json`, and `testing.json` now use existing feed storage and manure management configurations.


SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_decomposition.py
```diff
@@ -25,17 +25,26 @@ def test_calc_temp_factor(temp_average, x_inflection: float = 15.4, y_inflection
...
-    assert Decomposition._calc_moisture_factor(water_factor) == expect
+    if expected_base_1 < 0.0:
+        expected_term_1 = (-1) * ((-1 * expected_base_1) ** first_exponent)
+    else:
+        expected_term_1 = expected_base_1 ** first_exponent
+    if expected_base_2 < 0.0:
+        expected_term_2 = (-1) * ((-1 * expected_base_2) ** second_exponent)
+    else:
+        expected_term_2 = expected_base_2 ** second_exponent
+    expected = expected_term_1 * expected_term_2
```
Isgotkowitz, line 48: Ed rewrite this test

Isgotkowitz: Fixed

SC_redesign/Crop_and_Soil/crop/nitrogen_incorporation.py
```diff
@@ -207,8 +208,10 @@ def determine_nutrient_shape_parameters(half_mature_heat_fraction: float, mature
             heat_fraction=half_mature_heat_fraction, nitrogen_fraction=half_mature_nutrient_fraction,
             mature_nitrogen_fraction=mature_nutrient_fraction, emergence_nitrogen_fraction=emergence_nutrient_fraction
         )
+
+        adjusted_near_mature_nutrient_fraction = mature_nutrient_fraction + 0.00001
```
matthew7838: I think that declaring this 0.00001 with a descriptive variable should be simple and significantly improves readability.

SC_redesign/Crop_and_Soil/tests_SC/manager_tests/test_field_manager.py
```diff
@@ -42,6 +43,35 @@ def test_date_conversion_day(year: int, day: int, expected_day: int):
...
+    time = MagicMock()
+    setattr(time, "year", year)
+    setattr(time, "day", day)
+
+    actual = FieldManager._create_current_weather(weather, time, 4)
+
+    assert actual == expected
+    CurrentWeather.determine_daylength.assert_called_once()
+
+
```
matthew7838, line 74: Good that this part is covered, it's part of the missing coverage.

matthew7838 (approved):
Good debugging and great notes. One comment on improving readability, that's all.

SC_redesign/Crop_and_Soil/crop/nitrogen_incorporation.py
Isgotkowitz: Good suggestion

ew3361zh (approved):
Honestly pretty challenging to review with the number of different things going on in different parts of the module but looking through the changes and running the tests suite everything looked good (i.e. nothing jumped out to me as potentially harmful). Looks like this is cleaning up a lot of smaller pieces though so I understand why it was done this way.

Isgotkowitz:
> Honestly pretty challenging to review with the number of different things going on in different parts of the module but looking through the changes and running the tests suite everything looked good (i.e. nothing jumped out to me as potentially harmful). Looks like this is cleaning up a lot of smaller pieces though so I understand why it was done this way.

@ew3361zh I understand, my apologies for letting it get away from me like this. It was very easy to keep making little fixes and quickly snowballed into this juggernaut, I will be more careful to keep track of size and scope as I work on future PRs.

PR #614, opened 2023-06-29
Adding a class to manage manure nutrient values from the manure module

ltp23:
This PR introduces the `ManureNutrientManager` class, which manages manure nutrient supply from the manure module output and the requests from the crop and soil module. 

The class validates nutrient requests based on nitrogen and phosphorus needs and the available manure mass. It allows both full and partial fulfillment of nutrient requests, considering the nutrient compositions.


RUFAS/routines/manure/manure_nutrients/manure_nutrient_manager.py
```diff
@@ -0,0 +1,232 @@
...
+        Request nutrients from the crop and soil module.
+
+        Parameters
+        ----------
+        request : NutrientRequest
+            The request for nutrients.
+
+        Returns
+        -------
+        NutrientRequestResults
```
Isgotkowitz:
```suggestion
        NutrientRequestResults | None
```
Also I think the case when None is returned should be described

RUFAS/routines/manure/manure_nutrients/manure_nutrient_manager.py
```diff
@@ -0,0 +1,232 @@
...
+
+        """
+        min_positive = math.inf
+        for mass in projected_manure_masses:
+            if mass < 0:
+                raise ValueError(f'Projected manure mass cannot be negative: {mass}')
+            elif 0 < mass < min_positive:
+                min_positive = mass
+
+        return min_positive if min_positive != math.inf else 0
```
Isgotkowitz:
```suggestion
        return min_positive if min_positive != math.inf else 0.0
```

RUFAS/routines/manure/manure_nutrients/manure_nutrient_manager.py
```diff
@@ -0,0 +1,232 @@
...
+        -------
+        float
+            The projected manure mass.
+
+        Raises
+        ------
+        ValueError
+            If the request for a nutrient or the nutrient composition is negative.
+
+        """
```
Isgotkowitz: Is the nutrient composition fractional? If so, it may be a good idea to raise an error if it is > 1.0

Isgotkowitz (approved):
Looks good Loi. Left comments about a few small things

ltp23:
Thank you, Ed! I added some updates based on your feedback.

varrma (approved):

Issue #608, opened 2023-06-28
Create load_data() function in Input Manager


ew3361zh:
Create a new function `load_data()` in the Input Manager class that loads data from a JSON or CSV file and parses the it as instructed by the metadata.

PR #615, opened 2023-06-29
Create Input Manager Load Data Function

ew3361zh:
This PR tries to address issue #608 to create a new function `load_data()` in the Input Manager that will load data from a JSON or CSV file and parse it as instructed by the metadata.

Notes:
1. I'm unsure if there is additional parsing that should be done on the data read from the files.
2. I don't know about the best format for the CSV data. It seemed like a list of rows as dictionaries was one of the standard options but I'm not sure if there's something else that would work better for RuFaS's needs.


RUFAS/input_manager.py
```diff
@@ -19,19 +25,58 @@ def __init__(self) -> None:
...
+        """Loads data from JSON or CSV fileRaises
+
+        Raises
+        ------
+        Exception
+            If an error occurs while opening or reading a data file.
+
+        """
+        metadata_files_key = "files"
+        data_files = self.__metadata[metadata_files_key]
```
```suggestion
        files_details = self.__metadata["files"]
```
I think this is an overkill.
Also changed the name to a more descriptive one.

RUFAS/input_manager.py
```diff
@@ -19,19 +25,58 @@ def __init__(self) -> None:
...
+                    "function": self._load_data.__name__,
+                    }
+        for key, value in data_files.items():
+            file_path = value[path_key]
+            try:
+                if value["type"] == "json":
+                    with open(file_path) as json_file:
+                        data = json.load(json_file)
+                        self.__pool[key] = data
+                if value["type"] == "csv":
```
```suggestion
                elif value["type"] == "csv":
```

RUFAS/input_manager.py
```diff
@@ -19,19 +25,58 @@ def __init__(self) -> None:
...
+
+        """
+        metadata_files_key = "files"
+        data_files = self.__metadata[metadata_files_key]
+        path_key = "path"
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._load_data.__name__,
+                    }
+        for key, value in data_files.items():
+            file_path = value[path_key]
```
```suggestion
        for key, details in files_details.items():
            file_path = details[path_key]
```

RUFAS/input_manager.py
```diff
@@ -19,19 +25,58 @@ def __init__(self) -> None:
...
+                    with open(file_path) as json_file:
+                        data = json.load(json_file)
+                        self.__pool[key] = data
+                if value["type"] == "csv":
+                    with open(file_path, "r") as csv_file:
+                        data_reader = csv.DictReader(csv_file)
+                        self.__pool[key] = list(data_reader)
+                else:
+                    om.add_warning("InputManager load data file not csv/json", f"{file_path} not csv nor json and not"
+                                   f" added to data pool", info_map)
```
PooyaHekmati: I think a "be" verb is missing in this sentence. Also, it is a good idea to mention the key. The if statement works on the type that is provided in the metadata, not the file_path.

PooyaHekmati (commented):
And don't forget about logging :)

RUFAS/input_manager.py
ew3361zh: Addressed! Good call, thank you.

RUFAS/input_manager.py
```diff
@@ -49,3 +51,39 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
                 om.add_log("load_metadata_success", f"Successfully loaded metadata from {metadata_path}", info_map)
         except Exception as e:
             raise e
+
+    def _load_data(self) -> None:
+        """Loads data from JSON or CSV fileRaises
```
PooyaHekmati: typo?

RUFAS/input_manager.py
```diff
@@ -49,3 +51,39 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+        for key, details in files_details.items():
+            file_path = details[path_key]
+            om.add_log("load_data_attempt", f"Attempting to load data for {key} from {file_path}.", info_map)
+            try:
+                if details["type"] == "json":
+                    with open(file_path) as json_file:
+                        data = json.load(json_file)
+                        om.add_log("load_data_successful", f"Successfully loaded data for {key} from {file_path}.",
+                                   info_map)
+                        self.__pool[key] = data
```
```suggestion
                        self.__pool[key] = data
                        om.add_log("load_data_successful", f"Successfully loaded data for {key} from {file_path}.",
                                   info_map)
```
Not a big deal but better safe than sorry.

RUFAS/input_manager.py
```diff
@@ -49,3 +51,39 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+                elif details["type"] == "csv":
+                    with open(file_path, "r") as csv_file:
+                        data_reader = csv.DictReader(csv_file)
+                        om.add_log("load_data_successful", f"Successfully loaded data for {key} from {file_path}.",
+                                   info_map)
+                        self.__pool[key] = list(data_reader)
```
```suggestion
                        self.__pool[key] = list(data_reader)
                        om.add_log("load_data_successful", f"Successfully loaded data for {key} from {file_path}.",
                                   info_map)
```

RUFAS/input_manager.py
```diff
@@ -49,3 +51,39 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+                else:
+                    om.add_warning("InputManager load data file is not csv/json", f"File for {key} data in path"
+                                   f" {file_path} was not a csv nor json file and was not added to data pool", info_map)
```
PooyaHekmati: is this correct?

PooyaHekmati (changes requested):
I was going to go for LGTM, but lines 86 and 87 don't look correct.

RUFAS/input_manager.py
ew3361zh: I was wondering if we'd wanted to keep that separate in case there was some issue with adding the data to the pool separate from just loading the data but can certainly go along with this suggestion.

RUFAS/input_manager.py
ew3361zh: Do you mean the part about it being a warning vs. error? 
Or about taking the action of adding a warning as the main action if this else statement is reached?
Or the message itself?

RUFAS/input_manager.py
ew3361zh: Yep! Addressed.

tests/test_input_manager.py
```diff
@@ -30,11 +30,74 @@ def test_load_metadata(mock_input_manager: InputManager) -> None:
...
+            "dummy_data_file": {
+                "path": "dummy_data.txt",
+                "type": "txt"
+            }
+        }
+    }
+
+    with patch("builtins.open", mock_open(read_data="key_and_value")):
+        mock_input_manager._load_data()
+    assert mock_input_manager._InputManager__pool == {}
```
Isgotkowitz: Should the `OutputManager` be checked to make sure the correct warning was raised?

Isgotkowitz (approved):
Looks good Niko, only had one comment about how one of the tests could be expanded.

tests/test_input_manager.py
ew3361zh: It's an interesting idea - we haven't really been testing individual calls to OM methods. I'd be curious what @PooyaHekmati thinks.

PooyaHekmati: In an ideal world, yes! Since that is concerning a requirement, it needs to be checked.

ew3361zh: Sounds good, I updated the test!

PooyaHekmati (approved):

Issue #616, opened 2023-06-30
Create validate_data function for IM


ew3361zh:
Create a new function in the IM class called `validate_data` that performs validation on the data in the IM pool as instructed by the metadata.

From IM design doc:
-IM must provide data validation functionality to ensure that the structure of the data is usable and the values of the data are within the pre-defined ranges and follow specified patterns. For instance, if user wants to simulate 10^20 cows, this validation should fail as this number exceeds the maximum allowed number of cows.
-IM must terminate execution when critical inputs make validations fail. To achieve this, the function receives a flag to determine whether a validation is critical. This flag will be called `eager_termination` and will have the default value of True which means the default behavior is to terminate the process once invalid data is found.
-Minimum, maximum, minimum length, maximum length, and pattern will be used to perform validation. Since all of them are optional, the validation is contingent on their availability.

Nice to have (MVP?):
Severity value added to validation rules so some validations can be blocking and some can be non-blocking (not sure if this is planned for MVP - Sanity Check is not MVP though).

PR #620, opened 2023-07-03
Create Input Manager Validate Data Function

ew3361zh:
This PR tries to address issue #616 to create a new function `validate_data()` in the IM that will coordinate the validation and fixing of data in the IM data pool.

This PR adds placeholder functions for the actions `validate_data()` coordinates: `_validate_element()` and `_fix_data()`.


RUFAS/input_manager.py
```diff
@@ -91,3 +91,78 @@ def _load_data(self) -> None:
                                    info_map)
             except Exception as e:
                 raise e
+
+    def _validate_data(self, eager_termination: bool = True) -> bool:
+        """
+        Validates input data.
+
+        Args
+        ----
```
```suggestion
        Parameters
        ----------
```

RUFAS/input_manager.py
```diff
@@ -91,3 +91,78 @@ def _load_data(self) -> None:
...
+                                       f"Invalid data found: {key} - {value} but was fixable.",
+                                       info_map)
+        return True
+
+    def _validate(self, key: str, value: Any) -> bool:
+        """
+        Perform data validation checks.
+
+        Args
+        ----
```
Isgotkowitz: Parameters

RUFAS/input_manager.py
```diff
@@ -91,3 +91,78 @@ def _load_data(self) -> None:
...
+        # Perform data validation checks
+        # Return True if the data is valid, False otherwise
+        pass
+
+    def _fix_data(self, key: str, value: Any) -> bool:
+        """
+        Attempt to fix the invalid data.
+
+        Args
+        ----
```
Isgotkowitz: Parameters

RUFAS/input_manager.py
```diff
@@ -91,3 +91,78 @@ def _load_data(self) -> None:
...
             except Exception as e:
                 raise e
+
+    def _validate_data(self, eager_termination: bool = True) -> bool:
+        """
+        Validates input data.
+
+        Args
+        ----
+        eager_termination : bool, optional
```
```suggestion
        eager_termination : bool, default=True
```

RUFAS/input_manager.py
```diff
@@ -91,3 +91,78 @@ def _load_data(self) -> None:
...
+                    om.add_error("Invalid data.", f"Invalid data found: {key} - {value}", info_map)
+                    if not self._fix_data(key, value):
+                        om.add_error("Data not fixable.",
+                                     f"Unable to fix the invalid data: {key} - {value}. Terminating the process.",
+                                     info_map)
+                        return False
+                    else:
+                        om.add_warning("Data was fixable.",
+                                       f"Invalid data found: {key} - {value} but was fixable.",
+                                       info_map)
```
Isgotkowitz: Even if `eager_termination` is set to False, I think it would still be good to raise a warning and/or attempt to fix any invalid data found. The difference should be whether or not the the function exits when invalid/unfixable data is found. 

However if this is not done, then it would be better to return True at the very beginning of the function when `eager_termination` is False, because it won't matter if any of the data is invalid so iterating through the pool could be entirely skipped.

RUFAS/input_manager.py
```diff
@@ -91,3 +91,78 @@ def _load_data(self) -> None:
...
                 raise e
+
+    def _validate_data(self, eager_termination: bool = True) -> bool:
+        """
+        Validates input data.
+
+        Args
+        ----
+        eager_termination : bool, optional
+            Flag to determine if the process should terminate upon finding invalid data.
```
Isgotkowitz: I think either this description should be changed or the method behavior should be changed, because this process does not necessarily terminate upon finding invalid data. It has to find invalid data *and* be unable to fix the data to terminate.

Isgotkowitz (changes requested):
Nice PR Niko, this is a really good start on input validation. Changes requested on the docstrings - some on section names, others on better aligning the docstring with the method properties/behaviors. Also I think it would be good to expand when warnings are raised and when data is attempted to be fixed, but that would be a bigger deal so up to you.

RUFAS/input_manager.py
```diff
@@ -91,3 +91,78 @@ def _load_data(self) -> None:
...
+
+        Returns
+        -------
+        bool
+            True if all data is valid; False otherwise.
+        """
+        info_map = {"class": self.__class__.__name__,
```
michaelRichards99, line 109: `info_map` is a vague term and it's not clear how it's being used. Though not specific to this PR, I think a name such as `logging_context` would be more descriptive.

RUFAS/input_manager.py
```diff
@@ -91,3 +91,78 @@ def _load_data(self) -> None:
...
+                    }
+        for key, value in self.__pool.items():
+            if not self._validate(key, value):
+                if eager_termination:
+                    om.add_error("Invalid data.", f"Invalid data found: {key} - {value}", info_map)
+                    if not self._fix_data(key, value):
+                        om.add_error("Data not fixable.",
+                                     f"Unable to fix the invalid data: {key} - {value}. Terminating the process.",
+                                     info_map)
+                        return False
```
michaelRichards99: We should probably give a specific reason as to why the invalid data was not able to be fixed.
If this reason will be reported in the _fix_data method when it's implemented, then this error log might be redundant.

RUFAS/input_manager.py
```diff
@@ -91,3 +91,78 @@ def _load_data(self) -> None:
...
+                    else:
+                        om.add_warning("Data was fixable.",
+                                       f"Invalid data found: {key} - {value} but was fixable.",
+                                       info_map)
+        return True
```
michaelRichards99, line 143: Before returning, should we verify that all critical data for the simulation is present?

RUFAS/input_manager.py
```diff
@@ -91,3 +91,78 @@ def _load_data(self) -> None:
...
+            True if all data is valid; False otherwise.
+        """
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._validate_data.__name__,
+                    }
+        for key, value in self.__pool.items():
+            if not self._validate(key, value):
+                if eager_termination:
+                    om.add_error("Invalid data.", f"Invalid data found: {key} - {value}", info_map)
+                    if not self._fix_data(key, value):
```
michaelRichards99: Shouldn't we only be doing data fixing for non-critical data?

RUFAS/input_manager.py
michaelRichards99: This can be ignored in this PR. I do think we should consider the name change, though.

RUFAS/input_manager.py
ew3361zh: What about something like this?
```suggestion
       for key, value in self.__pool.items():
            if not self._validate(key, value):
                om.add_error("Invalid data.", f"Invalid data found: {key} - {value}", info_map)
                if not self._fix_data(key, value) and eager_termination:
                    om.add_error("Data not fixable.",
                                 f"Unable to fix the invalid data: {key} - {value}. Terminating the process.",
                                 info_map)
                    return False
                elif not self._fix_data(key, value):
                    om.add_error("Data not fixable.",
                                 f"Unable to fix invalid data: {key} - {value}. Continuing the process.",
                                 info_map)
                else:
                    om.add_warning("Data was fixable.",
                                   f"Invalid data found: {key} - {value} but was fixable.",
                                   info_map)
        return True
```

RUFAS/input_manager.py
ew3361zh: Good catch. That was in a version of the docstrings I thought I'd pushed. My thought is that the latter is the correct change to make.

RUFAS/input_manager.py
ew3361zh: Info_map has a wide array of uses and since it is a part of all 4 main data pools in OM, I'd probably opt to keep it as named. And I think depending on the IDE you're using, if you hover over the function that uses info_map shortly after it's created, it should show the docstrings with the explanation of what its purpose is.

RUFAS/input_manager.py
ew3361zh: Yea the question of where the error message is going is a good one to consider. I'm still working on the validate and fix functions in separate PRs but maybe it makes more sense to have the error message at the point of validation since we'll know what the violation is.

RUFAS/input_manager.py
Isgotkowitz: That looks good, I suggest running `_fix_data()` once and saving the result in variable though

RUFAS/input_manager.py
ew3361zh: Correct. The requirements state that IM must terminate execution when critical inputs fail validation but one of the use-cases is being able to run through all the inputs before terminating so that the user can get a full rundown of all the failures. I'll have to think through how to fit that logic in. Great point!

RUFAS/input_manager.py
ew3361zh: @michaelRichards99 I've added an issue for this: #635 
It's a good idea but is pushed for V2 to allow us to keep focused on priorities we need for V1.

michaelRichards99: Sounds good to me. If the MVP will be run on our predefined input data that is known to be correct then this wouldn't bee an urgent issue but should be addressed in V2.

RUFAS/input_manager.py
michaelRichards99: I agree that finding all data that is invalid and being able to report that would be a nice feature. On the other hand many compilers terminate on the first error encountered. If designing it the first way is a lot more challenging, terminating on the first failure could be a valid option. I don't think the IM spec prefers one way or the other IIRC.

RUFAS/input_manager.py
```diff
@@ -91,3 +91,78 @@ def _load_data(self) -> None:
...
                 raise e
+
+    def _validate_data(self, eager_termination: bool = True) -> bool:
+        """
+        Validates input data.
+
+        Parameters
+        ----------
+        eager_termination : bool, default=True
+            Flag to determine if the process should terminate upon finding invalid data.
```
PooyaHekmati: Specify the expected behavior.
```suggestion
            If true, the process will be terminated upon finding invalid data.
```

RUFAS/input_manager.py
PooyaHekmati: Relying on external actors (compiler, OS, platform, etc...) to achieve a desired behavior is a risk that needs justification. In this case, seems to be an unnecessary risk.

RUFAS/input_manager.py
```diff
@@ -91,3 +91,78 @@ def _load_data(self) -> None:
...
+        bool
+            True if all data is valid; False otherwise.
+        """
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._validate_data.__name__,
+                    }
+        for key, value in self.__pool.items():
+            if not self._validate(key, value):
+                if eager_termination:
+                    om.add_error("Invalid data.", f"Invalid data found: {key} - {value}", info_map)
```
PooyaHekmati: 1- This message blongs to self._validate.
2- At this stage, it is a warning, it becomes an error when we try to fix it and it fails.
```suggestion
                    om.add_warning("Invalid data", f"Invalid data found: {key=}; {value=}", info_map)
```

RUFAS/input_manager.py
```diff
@@ -91,3 +91,78 @@ def _load_data(self) -> None:
...
+                                     f"Unable to fix the invalid data: {key} - {value}. Terminating the process.",
+                                     info_map)
+                        return False
+                    else:
+                        om.add_warning("Data was fixable.",
+                                       f"Invalid data found: {key} - {value} but was fixable.",
+                                       info_map)
+        return True
+
+    def _validate(self, key: str, value: Any) -> bool:
```
PooyaHekmati: not a good name, maybe rename it to _validate_item or _validate_element

RUFAS/input_manager.py
```diff
@@ -91,3 +91,78 @@ def _load_data(self) -> None:
                                    info_map)
             except Exception as e:
                 raise e
+
+    def _validate_data(self, eager_termination: bool = True) -> bool:
+        """
+        Validates input data.
```
PooyaHekmati: It also tries to fix the data

RUFAS/input_manager.py
```diff
@@ -91,3 +91,78 @@ def _load_data(self) -> None:
...
+                    "function": self._validate_data.__name__,
+                    }
+        for key, value in self.__pool.items():
+            if not self._validate(key, value):
+                if eager_termination:
+                    om.add_error("Invalid data.", f"Invalid data found: {key} - {value}", info_map)
+                    if not self._fix_data(key, value):
+                        om.add_error("Data not fixable.",
+                                     f"Unable to fix the invalid data: {key} - {value}. Terminating the process.",
+                                     info_map)
```
PooyaHekmati: This message belongs to self._fix_data.
```suggestion
                        om.add_error("Data not fixable",
                                     f"Unable to fix the invalid data: {key=}, {value=}. Terminating the process.",
                                     info_map)
```

RUFAS/input_manager.py
```diff
@@ -91,3 +91,78 @@ def _load_data(self) -> None:
...
+                        return False
+                    else:
+                        om.add_warning("Data was fixable.",
+                                       f"Invalid data found: {key} - {value} but was fixable.",
```
PooyaHekmati: Move it to _fix_data.
```suggestion
                        om.add_warning("Data fixed",
                                       f"Invalid data fixed: {key=}; {value=}",
```

RUFAS/input_manager.py
PooyaHekmati: Yes, we should also log the number of items that were tried, number of items that were valid, number of invalid items, and number of items that we fixed.

PooyaHekmati (changes requested):
There are many comments on the PR which makes it difficult to read, follow, and review the code. Please resolve.

RUFAS/input_manager.py
ew3361zh: Addressed

RUFAS/input_manager.py
ew3361zh: I've changed it to `_validate_element()`

RUFAS/input_manager.py
ew3361zh: I've added counts for:
- Valid elements
- Invalid elements
- Fixed elements
- Unfixable elements
- Total checked elements
- Invalid critical elements

RUFAS/input_manager.py
```diff
@@ -91,3 +91,113 @@ def _load_data(self) -> None:
...
+                        invalid_critical_elements_counter += 1
+                    is_data_fixable = self._fix_data(variable, value)
+                    if is_data_fixable:
+                        fixed_elements_counter += 1
+                    elif not is_data_fixable and eager_termination:
+                        return False
+                    else:
+                        unfixed_elements_counter += 1
+
+        om.add_log("Total Valid Elements", f"{valid_elements_counter} valid elements found.", info_map)
```
PooyaHekmati: Simplify:
```suggestion
        om.add_log("Total Valid Elements", f"{valid_elements_counter=}", info_map)
```
The same applies to the rest.

RUFAS/input_manager.py
```diff
@@ -91,3 +91,113 @@ def _load_data(self) -> None:
...
+                    "function": self._validate_data.__name__,
+                    }
+        valid_elements_counter = 0
+        invalid_elements_counter = 0
+        fixed_elements_counter = 0
+        unfixed_elements_counter = 0
+        invalid_critical_elements_counter = 0
+        total_elements_checked_counter = 0
+
+        for key in self.__pool.keys():
```
PooyaHekmati, line 118: Looping through the pool ensures all data is being validated. However, it is possible to have something in the metadata which doesn't have a match in the data. Looping through the data doesn't catch things like that. If we loop through the metadata though, it will be possible to leave some data unchecked. So I think it is good to keep this data as is, but a solution is needed to ensure nothing is being left unused in the metadata.

RUFAS/input_manager.py
```diff
@@ -91,3 +91,113 @@ def _load_data(self) -> None:
...
+            for variable, value in self.__pool[key].items():
+                total_elements_checked_counter += 1
+                property_map_key = self.__metadata["files"][key]["properties"]
+                if self._validate_element(variable, value):
+                    valid_elements_counter += 1
+                else:
+                    invalid_elements_counter += 1
+                    is_data_critical = "default" not in self.__metadata["properties"][property_map_key][variable].keys()
```
PooyaHekmati: I think this is not what we agreed on. We said this condition should be handled inside `_fix_data`. If validation fails, we call `_fix_data` unconditionally. It is `_fix_data`'s job to tell us if the data was fixable. If the data could not be fixed, it means (by definition) it was critical data.

RUFAS/input_manager.py
```diff
@@ -91,3 +91,113 @@ def _load_data(self) -> None:
...
+                    if is_data_critical:
+                        invalid_critical_elements_counter += 1
+                    is_data_fixable = self._fix_data(variable, value)
```
PooyaHekmati: Exactly! Line 127 is the same as this one. (127 needs to be removed)

We also said that the `able` suffix denotes the potential, here we have already taken the action so we need to have a name that refers to the past.

PooyaHekmati (changes requested):

RUFAS/input_manager.py
ew3361zh: Makes sense, I will coordinate this with @allisterakun and his work on the `_fix_data` function.

RUFAS/input_manager.py
ew3361zh: My initial thought would be a separate function called after `validate_data()` which would compare the metadata properties sections with the pool data and add in the default values for anything missing (logging/adding a warning for anything added to the pool, of course).
`_reconcile_data()` or something to that effect.
I will create an issue for it.

ew3361zh: Also makes me consider the case where there's a piece of data in the pool loaded from one of the input_data files that doesn't have a properties counterpart in the metadata to do a check.
Possibly this comes from someone entering a new piece of data into an input file (that is added into the code as well) but forgetting to add it to the metadata file. I would think in this case we'd want to add an error at the `_validate_element()` check when it can't find the counterpart in the metadata but continue to proceed with whichever validation process the user selected (eager_termination true or false).
We'd also probably want to have an error logged in `_fix_data()`.
I can create another issue for this but would be curious your thoughts @PooyaHekmati.

PooyaHekmati: Good catch Niko!
I think having data that doesn't have an explanation (aka metadata) is severe enough to be brought to the user's attention, but it is not severe enough to stop the execution. It falls somewhere between warning and error. So we should log this (error or warning) and remove this element from the pool. It doesn't need to get to `_fix_data()` because if the metadata is missing then we don't know how to fix it.

RUFAS/input_manager.py
```diff
@@ -91,3 +91,106 @@ def _load_data(self) -> None:
...
+                if self._validate_element(variable, value):
+                    valid_elements_counter += 1
+                else:
+                    invalid_elements_counter += 1
+                    is_data_fixed = self._fix_data(variable, value)
+                    if is_data_fixed:
+                        fixed_elements_counter += 1
+                    elif not is_data_fixed and eager_termination:
+                        invalid_critical_elements_counter += 1
+                        return False
```
Isgotkowitz, line 130:
```suggestion
```
Should the program be returning right here, without checking all the values? Also, if it should be returning here, should it log some kind of message to indicate to the user why it stopped here, like the name of the variable that it stopped on?

RUFAS/input_manager.py
```diff
@@ -91,3 +91,106 @@ def _load_data(self) -> None:
...
+        bool
+            True if all data is valid; False otherwise.
+        """
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._validate_data.__name__,
+                    }
+        valid_elements_counter = 0
+        invalid_elements_counter = 0
+        fixed_elements_counter = 0
+        invalid_critical_elements_counter = 0
```
Isgotkowitz, line 115: When recording these, would it also be a good idea which to record exactly which elements were invalid, fixed, and critically invalid in addition to the counts of each? It would potentially add a lot more information to the logs, but would also save a lot of time for someone who is developing input that has mistakes in it.

Isgotkowitz (approved):
Looks really good Niko. 2 comments, both related to potentially logging more information, which would help someone looking at the logs debug their input faster.

RUFAS/input_manager.py
ew3361zh: Good question, Ed. Yes, it should be returning False here. This is the use case where the user has the eager_termination flag set to True which means as soon as they encounter the first invalid critical element (i.e. an element that cannot be fixed), they want to stop the simulation. The second use case of wanting to run the data digestion pipeline all the way through but still stopping before the simulation starts requires the user to set eager_termination to False.

One piece of the IM that is still to come is the order of operations function start() which will start IM's data digestion pipeline and look something like this:

```
def start(self) -> None:
        """
        Start the data digestion pipeline.
        """
        self.load_metadata()
        self.load_data()
        self.validate_data()
```
I'll add a boolean check on self.validate_data() and if it comes back false, this will raise an exception and the simulation will stop.

This does bring up an interesting point though that we'll want to make sure OM has been called to dump its data before we exit the system when self.validate_data() returns False to make sure the user has access to the logs/errors/warnings. That will be where the information about what was the variable it encountered before exiting.

RUFAS/input_manager.py
ew3361zh: Definitely. This will be captured in the `_validate_element()` and `_fix_data()` functions as they receive each piece of data.

tests/test_input_manager.py
```diff
@@ -102,3 +103,110 @@ def test_load_data_raises_exception(mock_input_manager: InputManager) -> None:
     with patch("builtins.open", mock_open_func):
         with pytest.raises(Exception):
             mock_input_manager._load_data("bad/path.csv")
+
+
+@pytest.fixture
+def mock_metadata(mocker) -> Dict[str, Dict[str, Any]]:
```
michaelRichards99:
```suggestion
def mock_metadata(mocker: MockerFixture) -> Dict[str, Dict[str, Any]]:
```
Add type annotations for the mocker object here and for the others unit tests.

michaelRichards99 (approved):
LGTM

PooyaHekmati (approved):
Please fix/add type annotations in the test files before merging.

PR #623, opened 2023-07-03
Integrate `FieldManager` into `State` and `SimulationEngine`

Isgotkowitz:
This PR integrates the new Soil and Crop module into the main section of the codebase, in the `RUFAS/` directory.

- `RUFAS/__init__.py` has been cleared, because the import statement that was there before was causing circular import errors when the test suite for the new Crop and Soil module was being run.
- `classes.py` replaces `Fields` with a `FieldManager` instance, and swaps out the old annual update routine for the new.
- `output_handler.py` has had the `FieldsReport` removed.
- `base_report.py` has had the old `fields` removed.
- `feed.py` now immediately returns before it can try to interact with the new `FieldManager` instance that is passed to it. This was the way I could think of eliminating this issue with the fewest lines possible, but I'm sure there are better, more future-proof ways of eliminating this issue and I am very open to suggestions.
- `simulation_engine.py` now runs the new daily routine for the Crop and Soil module, and also removes the call to the old annual update routine.
- `field.py` has had the `Time` type annotation removed - `Time` being imported causes circular import errors.
- `current_weather.py` has had `check_current_weather()` replaced by a `@staticmethod` in `FieldManager`. The reason was because it was accessing `Weather` objects incorrectly, and I couldn't get it work correctly where it was (something to do with classmethods), so I just removed it and moved the functionality to `FieldManager`. Also, `determine_daylength()` is no longer a protected method, as it is now needed by `FieldManager`.
- `events.py` has had the `Time` import and type annotation removed.
- `field_manager.py`
  - Problematic import statements and type annotations have been removed.
  - The type annotations of `init` have been corrected.
  - The `is_leap_year()` from `classes.py` has been replaced with one from `Schedule` (because the one from `classes.py` can't be imported).
  - The new method for creating `CurrentWeather` objects is implemented.
- `schedule.py` now contains an implementation of `is_leap_year()`. It is the same as it is in `classes.py`, and this seemed like the most logical place to duplicate it.
- The tests for `CurrentWeather`, `Events`, `Schedule`, and `FieldManager` all reflect the new methods and lack of type annotations.
- `test_user_prompt.py` has had the import statement for `SimulationEngine` changed so that it now works with the modified `init` file.
- `test_misc.py` has had the mocked old daily field function removed and checks that the new `FieldManager` daily update routine is called once.
- `finish_line_doc.md` has been updated to reflect progress made.


RUFAS/routines/feed/feed.py
```diff
@@ -684,6 +684,8 @@ def daily_feed_storage(self, fields, feed_report):
                 can be added to output.
         """
         # aggregate crop yield across fields
+        return
```
matthew7838, line 687: what is this return for?

SC_redesign/Crop_and_Soil/manager/current_weather.py
```diff
@@ -33,22 +31,8 @@ class CurrentWeather:
...
-        cls.max_air_temperature = weather.T_max
-        cls.annual_mean_air_temperature = weather.T_avg_annual
-        cls.rainfall = weather.rainfall
-        cls.irrigation = weather.irrigation
-        cls.daylength = CurrentWeather._determine_daylength(month=month)
-        return CurrentWeather()  # TODO: placeholder for typing, needs implementation
-
     @staticmethod
-    def _determine_daylength(month: int) -> int:
+    def determine_daylength(month: int) -> int:
```
matthew7838, line 35: When I started I was told by Clay to create the static method with starting `_`, is there a particular reason to not use the underscore here? or have we shifted to a different naming convention?

matthew7838 (approved):
LGTM. Just a few questions. I agree the class method thing is complicated and I think you handled it really well here.

RUFAS/routines/feed/feed.py
Isgotkowitz: Prevents all of the code in that method from running. Before the entire old `Fields` object was passed to the feed module, and now the `FieldManager` is passed, but if the code were to run it would crash because the `Fields` and `FieldManager` have very different structures.

SC_redesign/Crop_and_Soil/manager/current_weather.py
Isgotkowitz: No, the way you did it was correct. I shifted the logic for creating `CurrentWeather` objects from the `CurrentWeather` class to `FieldManager` - but it still needed to use `determine_daylength()` so I made it public.

RUFAS/routines/feed/feed.py
JoeWaddell: So does that mean this method should be removed entirely? Or is it going to be re-implemented in another PR?

Isgotkowitz: I do not know to the first question. I am unsure about what is going to happen with the code in the Feed module. I figured this was the simplest way to eliminate the issue without having to remove large parts of the Feed codebase.

I am fairly sure that it is not going to be reimplemented anytime soon. I asked @KFosterReed and she said the Feed module was not going to be included with V1, so I am not sure what the timeline would be for re-implementing it.

JoeWaddell (approved):
Looks good to me: the explanation of the Feed responsibility makes sense.

PR #624, opened 2023-07-04
Adding a nutrient request method to the manure manager class

ltp23:
This PR mainly introduces four new methods to the ManureManager class, but only the public `request_nutrients()` method should be relevant to the crop&soil module. The other three methods are intended to be private.  

`_get_manure_type`: This method determines the type of manure produced based on the manure treatment system. This is essential for mapping between treatment types and manure types.

`_get_manure_density`: This method looks up the density of manure produced by a given pen based on the pen's manure treatment system. The density corresponds to predefined densities for different types of manure.

`_add_manure_nutrients`: This method adds the nutrients in the manure produced by a specific pen to the manure nutrient manager. This is done using the daily output data of the manure treatment system.

`request_nutrients`: This method handles nutrient requests from the crop and soil module. The method evaluates the nutrient request, calculates the projected manure mass that would satisfy the request, and checks against the available nutrients.

Relevant unit tests are also included. 

In case anyone misses it, a recent PR has changed the names of AnimalManagement and ManureManagement to AnimalManager and ManureManager.


RUFAS/routines/manure/constants/manure_constants.py
```diff
@@ -1,18 +1,29 @@
...
-
-    Attributes:
-        MANURE_DENSITY: The density of manure, kg/m^3.
-        UREA_MOLAR_MASS: The molar mass of urea, g/mol.
-        UREA_DENSITY: The density of urea, g/cm^3.
-        TAN_MOLAR_MASS: The molar mass of TAN, g/mol.
-        URINE_TAN_FACTOR: The factor to convert urine nitrogen to urine total ammoniacal nitrogen, unitless.
-        MANURE_SOLIDS_BEDDING_DENSITY: The density of manure solids bedding, kg/m^3.
-
-    """
```
Isgotkowitz: I think this is actually the way that the NumPy style guide specifies how class docstrings should be written.

RUFAS/routines/manure/manure_manager.py
```diff
@@ -178,6 +186,109 @@ def daily_update(self, animal_manager: AnimalManager) -> None:
...
+
+    def _add_manure_nutrients(self, pen: ManureManagerPen,
+                              manure_treatment_daily_output: ManureTreatmentDailyOutput) -> None:
+        """
+        Add the nutrients in the manure produced by a given pen to the manure nutrient manager.
+        Parameters
+        ----------
+        pen : ManureManagerPen
+            A pen object to look up the type of its manure treatment system.
+        manure_treatment_daily_output
```
Isgotkowitz:
```suggestion
        manure_treatment_daily_output : ManureTreatmentDailyOutput
```

RUFAS/routines/manure/manure_manager.py
```diff
@@ -178,6 +186,109 @@ def daily_update(self, animal_manager: AnimalManager) -> None:
...
+            The daily output data of the manure treatment system.
+        Returns
+        -------
+        None
+        """
+        # TODO: With the introduction of different manure types, we should rename attributes in ManureTreatmentDailyOutput
+        # to make it more generic and not specific to any manure type.
```
Isgotkowitz, line 266: Is this is a TODO for something that should be addressed in this PR or in a future one? If a future one, I think it would be good to make an issue for it and add the reference number to it.

RUFAS/routines/manure/manure_manager.py
```diff
@@ -178,6 +186,109 @@ def daily_update(self, animal_manager: AnimalManager) -> None:
...
+        request : NutrientRequest
+            The specific nutrient request, including quantities of nitrogen and phosphorus.
+        Returns
+        -------
+        NutrientRequestResults | None
+            The results of the nutrient request, detailed in a `NutrientRequestResults` object, which includes
+            the amount of nitrogen, phosphorus, total manure mass, dry matter, and others that can be provided to fulfill the request.
+            Returns None if the request cannot be fulfilled.
+        """
+        # This is a wrapper method that calls the request_nutrients method of the manure nutrient manager.
```
Isgotkowitz:
```suggestion
        
        Notes
        ------
        This is a wrapper method that calls the request_nutrients method of the manure nutrient manager.

        """
```

Isgotkowitz (approved):
Excellent PR, very clean code Loi. Four comments, only related to docstrings and comments

RUFAS/routines/manure/manure_manager.py
varrma: At Present the manure request is based on N and P mass, while the manure output has three different types based on liquid, slurry and solid manure from the different manure scenario IDs. This is should be revisited, how the manure accumulation could be done.

varrma (approved):
Looks good Loi !!

RUFAS/routines/manure/manure_manager.py
Isgotkowitz: Then there should be a GitHub issue made of it, otherwise there is a good chance that it will be forgotten and that TODO will be left here dangling

PR #605, opened 2023-06-28
test coverage for SC_redesign branch  (1/3)

matthew7838:
### Summary
This PR covers the first half of any files that were uncovered while running the coverage tool, some tests were added and some were patched with conditions that did not meet and led to insufficient coverage. The next related PR will be solely focusing on rewriting tests for `water_uptakes.py` since it's missing a significant amount of tests and might requires a major rewrite on testing.

### Changes
- Tests for the crop.py is added, it's missing completely
- Fixed `crop_data.py` and `dormancy.py` bugs that are related to the same syntax error and were never tested.
- In `test_crop_data.py`, fixed testing for `test_in_growing_season_property`, now the `growing` variable is used
- In `cropdata.py`, deleted post init part that checks for type `TREE` since it wasn't part of any crop and was never used
- Added tests to test extend_phosphate_uptakes_to_full_profile in `phosphorus_incorporation.py`
- In `test_crop_management.py`, `cut_crop` is now properly tested
- In `test_dormacy.py` and `test_nitrogen_incorporation.py` new parameters were passed in to make sure errors are caught

### **Notes**
!!**Ignore**  anything to `test_water_uptake.py` since it is merged from a preapproved PR.!!

PR #625, opened 2023-07-05
Water uptake test rework

matthew7838:
### Summary
This PR's purpose is to patch up missing tests in `water_uptake.py`, it is stacked on #605  and is part of the work to reach 100% test coverage progress. It's separated as an independent PR due to the fact that most of the functions were not tested.

### Changes
- Added another set of values that tests conditions that were not tested in `test_correct_layer_for_efficiency`
- Added `test_extract_water_from_soil` to test `extract_water_from_soil`
- Added `test_reduce_efficiency_of_uptake` to test `reduce_efficiency_of_uptake`
- Added `test_adjust_water_uptakes` to test `adjust_water_uptakes`
- Added `test_find_stratified_max_water_uptakes` to test `find_stratified_max_water_uptakes`
- Added `test_determine_max_water_uptake_to_depth` to test `determine_max_water_uptake_to_depth`
- Added `test_take_up_water` to test `_take_up_water`


SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -54,17 +56,130 @@ def test_uptake_water(max_trans):
...
     """This method only tests for edge cases, other parts of the method already have coverage"""
     crop_data = CropData(actual_water_uptakes=uptakes)
     soil_data = SoilData(soil_layers=layers, field_size=3)
     uptake = WaterUptake(crop_data=crop_data)
-    try:
+    if should_fail:
+        try:
+            uptake.extract_water_from_soil(soil_data)
+        except Exception as e:
+            assert str(e) == "actual_water_uptakes should be the same length as the number of soil layers"
```
Isgotkowitz: I should have mentioned this before but when testing for errors the `pytest.raises()` context manager should be used. In this case, when using that method, this test would be 
```python
with pytest.raises(Exception) as e:
    uptake.extract_water_from_soil(soil_data)
assert str(e.value) == "expected error message"
```

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -54,17 +56,130 @@ def test_uptake_water(max_trans):
...
+                          ([0.1, 0.2, 0.4], [9.24, 7.7], [2.0, 3.5, 4.2], True),
+                          ([0.1, 0.2, 0.4], [9.24, 7.7, 1.31], [2.0, 3.5], True)])
+def test_reduce_efficiency_of_uptake(potential_uptakes: List[float], water_availabilities: List[float],
+                                     available_capacities: List[float], should_fail: bool) -> None:
+    """Tests that the reduced efficiency of uptake is calculated correctly and correct exceptions are thrown"""
+    if should_fail:
+        try:
+            WaterUptake._reduce_efficiency_of_uptake(potential_uptakes, water_availabilities, available_capacities)
+        except Exception as e:
+            assert str(e) == "potential_uptakes, water_availabilities, and available_capacities must be of equal length"
```
Isgotkowitz: pytest.raises

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -54,17 +56,130 @@ def test_uptake_water(max_trans):
...
+    else:
+        zipped = zip(potential_uptakes, water_availabilities, available_capacities)
+        expected = [WaterUptake._correct_layer_for_efficiency(pot, avail, cap) for pot, avail, cap in zipped]
+        assert expected == WaterUptake._reduce_efficiency_of_uptake(potential_uptakes, water_availabilities,
```
Isgotkowitz: It would be better to have `_correct_layer_for_efficiency()` mocked and test that it is being called with the expected combinations from the input. Because as this test is now, it is duplicating the exact same logic as what is supposed to be tested.

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -54,17 +56,130 @@ def test_uptake_water(max_trans):
...
+                          ([0.1, 0.2], [0.4, 0.5, 0.6], [0.7, 0.8, 0.9], 2.5, True)])
+def test_adjust_water_uptakes(potential_uptakes: List[float], water_availabilities: List[float],
+                              unmet_demands: List[float], uptake_compensation: float, should_fail: bool) -> None:
+    """Tests that the adjusted water uptakes are calculated correctly and correct exceptions are thrown"""
+    if should_fail:
+        try:
+            WaterUptake._adjust_water_uptakes(potential_uptakes, water_availabilities, unmet_demands,
+                                              uptake_compensation)
+        except Exception as e:
+            assert str(e) == "potential_uptakes and unmet_demands must be the same length."
```
Isgotkowitz: pytest.raises

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -54,17 +56,130 @@ def test_uptake_water(max_trans):
...
+        expected = [WaterUptake._correct_layer_for_efficiency(pot, avail, cap) for pot, avail, cap in zipped]
+        assert expected == WaterUptake._reduce_efficiency_of_uptake(potential_uptakes, water_availabilities,
+                                                                    available_capacities)
+
+
+@pytest.mark.parametrize("potential_uptakes,water_availabilities,unmet_demands,uptake_compensation,should_fail",
+                         [([0.1, 0.2, 0.3], [0.4, 0.5, 0.6], [0.7, 0.8, 0.9], 2.3, False),
+                          ([0.1, 0.2], [0.4, 0.5, 0.6], [0.7, 0.8, 0.9], 2.5, True)])
+def test_adjust_water_uptakes(potential_uptakes: List[float], water_availabilities: List[float],
+                              unmet_demands: List[float], uptake_compensation: float, should_fail: bool) -> None:
```
Isgotkowitz: Because it isn't used at all in the actual method, it may be appropriate to remove the `water_availabilities` parameter from the method. However, before removing it, it should be carefully checked how this method is being used, because it also may be possible that `water_availabilities` should be used here and there was an oversight that lead to it not being used.

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -54,17 +56,130 @@ def test_uptake_water(max_trans):
...
+    """Tests that the adjusted water uptakes are calculated correctly and correct exceptions are thrown"""
+    if should_fail:
+        try:
+            WaterUptake._adjust_water_uptakes(potential_uptakes, water_availabilities, unmet_demands,
+                                              uptake_compensation)
+        except Exception as e:
+            assert str(e) == "potential_uptakes and unmet_demands must be the same length."
+    else:
+        expected = [uptake + (demand * uptake_compensation) for uptake, demand in zip(potential_uptakes, unmet_demands)]
```
Isgotkowitz: `expected` should be pre-computed and included with the test case.

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -54,17 +56,130 @@ def test_uptake_water(max_trans):
...
+def test_find_stratified_max_water_uptakes(root_depth: float, max_transpiration: float, water_distro_parameter: float,
+                                           upper_depths: List[float],
+                                           lower_depths: List[float], should_fail: bool) -> None:
+    """Tests that the stratified max water uptakes are calculated correctly and correct exceptions are thrown"""
+    if should_fail:
+        try:
+            WaterUptake._find_stratified_max_water_uptakes(root_depth, max_transpiration, water_distro_parameter,
+                                                           upper_depths, lower_depths)
+        except Exception as e:
+            assert str(e) == "upper_depths and lower_depths must be the same length"
```
Isgotkowitz: pytest.raises

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -54,17 +56,130 @@ def test_uptake_water(max_trans):
...
+    else:
+        expected = []
+        for upper, lower in zip(upper_depths, lower_depths):
+            top_potential = WaterUptake._determine_max_water_uptake_to_depth(root_depth, upper, max_transpiration,
+                                                                             water_distro_parameter)
+            bottom_potential = WaterUptake._determine_max_water_uptake_to_depth(root_depth, lower, max_transpiration,
+                                                                                water_distro_parameter)
+            expected.append(bottom_potential - top_potential)
+
```
Isgotkowitz: `expected` should be pre-computed and included with each test case

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -54,17 +56,130 @@ def test_uptake_water(max_trans):
...
+                          (0, 25.7, 33.4, 42.3)])
+def test_determine_max_water_uptake_to_depth(root_depth: float, depth: float, max_transpiration: float,
+                                             water_distro_parameter: float) -> None:
+    """Tests that the stratified max water uptake to depth are calculated correctly and correct exceptions are thrown"""
+    if root_depth == 0:
+        expected = 0
+    else:
+        term1 = max_transpiration / (1 - exp(-water_distro_parameter))
+        term2 = 1 - exp(-water_distro_parameter * depth / root_depth)
+        expected = term1 * term2
```
Isgotkowitz: This is a case where pre-computing the expected value is (IMO) not the best strategy for effective testing, because the equation that is at the heart of what is being tested has multiple layers and the result has a lot of decimals. However, word-for-word duplication still renders this test case inconsequential.

So what I to do in cases like these is to move the actual implementation off of my screen so I can't look at it, look at the original equation (in this case, SWAT eqn. 5:2.2.1) and reimplement the equation in the test in a way that differs from the original. This could mean implementing the whole equation in one-line, adding in parentheses that are not strictly necessary, changing the order of terms when possible, ideally all of the above and more. Usually the test logic and the actual method logic produce the same results, but sometimes they will be different and this forces you to recheck both implementations against the equation in the documentation, which helps to catch a lot of bugs.

Also an important thing to know, the `pytest.approx()` feature is really useful when doing this because if two different implementations produce results that are just slightly different then can use `approx` to check that the results are approximately equal.

.coverage
Isgotkowitz, line 1: This file should not be included in the PR.

Isgotkowitz (changes requested):
This is a good start on filling out the test suite, but I noticed some underlying issues with a lot of the tests that need to be addressed. The 2 main things that need changing is the use of `pytest.raises` instead of try-except blocks, and not reproducing logic inside the code.

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
matthew7838: good catch, looking at the reference I think it's not needed.

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -1,7 +1,10 @@
 from math import exp
+from typing import List
 
 import pytest
 from unittest.mock import MagicMock, patch
+from mock import call
```
Isgotkowitz:
```suggestion
from unittest.mock import MagicMock, patch, call
```

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -54,17 +58,126 @@ def test_uptake_water(max_trans):
...
+                               " length"
+    else:
+        WaterUptake._correct_layer_for_efficiency = MagicMock()
+        result = WaterUptake._reduce_efficiency_of_uptake(potential_uptakes, water_availabilities, available_capacities)
+        expected_calls = []
+        for i in range(len(potential_uptakes)):
+            expected_calls.append(call(potential_uptakes[i], water_availabilities[i], available_capacities[i]))
+        WaterUptake._correct_layer_for_efficiency.assert_has_calls(expected_calls)
+        zipped = zip(potential_uptakes, water_availabilities, available_capacities)
+        expected = [WaterUptake._correct_layer_for_efficiency(pot, avail, cap) for pot, avail, cap in zipped]
```
Isgotkowitz:
```suggestion
        WaterUptake._correct_layer_for_efficiency = MagicMock(return_value=0.5)
        result = WaterUptake._reduce_efficiency_of_uptake(potential_uptakes, water_availabilities, available_capacities)
        expected_calls = []
        for i in range(len(potential_uptakes)):
            expected_calls.append(call(potential_uptakes[i], water_availabilities[i], available_capacities[i]))
        WaterUptake._correct_layer_for_efficiency.assert_has_calls(expected_calls)
        expected = [0.5] * len(potential_uptakes)
```

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -54,17 +58,126 @@ def test_uptake_water(max_trans):
...
+                                           upper_depths: List[float],
+                                           lower_depths: List[float], should_fail: bool, expected: List[float]) -> None:
+    """Tests that the stratified max water uptakes are calculated correctly and correct exceptions are thrown"""
+    if should_fail:
+        with pytest.raises(Exception) as e:
+            WaterUptake._find_stratified_max_water_uptakes(root_depth, max_transpiration, water_distro_parameter,
+                                                           upper_depths, lower_depths)
+        assert str(e.value) == "upper_depths and lower_depths must be the same length"
+    else:
+        assert pytest.approx(expected, abs=0.0001) == WaterUptake._find_stratified_max_water_uptakes(
```
Isgotkowitz:
```suggestion
        assert pytest.approx(expected) == WaterUptake._find_stratified_max_water_uptakes(
```

Isgotkowitz: We should stick with the default absolute tolerance for approx

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -54,17 +58,126 @@ def test_uptake_water(max_trans):
...
+        WaterUptake._determine_actual_layer_uptake = MagicMock()
+        result = WaterUptake._take_up_water(potential_uptakes, water_availabilities,
+                                            wilting_points)
+        expected_calls = []
+        for i in range(len(potential_uptakes)):
+            expected_calls.append(call(potential_uptakes[i], water_availabilities[i], wilting_points[i]))
+        WaterUptake._determine_actual_layer_uptake.assert_has_calls(expected_calls)
+        zipped = zip(potential_uptakes, water_availabilities, wilting_points)
+        expected = [WaterUptake._determine_actual_layer_uptake(pot, avail, wilt) for pot, avail, wilt in zipped]
+        assert expected == result
```
Isgotkowitz: Same comment here as for `test_reduce_efficiency_of_uptake()`

SC_redesign/Crop_and_Soil/crop/water_uptake.py
```diff
@@ -305,7 +306,7 @@ def _determine_max_water_uptake_to_depth(root_depth: float, depth: float, max_tr
             return 0
 
         term1 = max_transpiration / (1 - exp(-water_distro_parameter))
-        term2 = 1 - exp(-water_distro_parameter * depth / root_depth)
+        term2 = math.floor(1 - exp(-water_distro_parameter * depth / root_depth))
```
Isgotkowitz: I thought the same thing when I was working on some SWAT equations, I had to ask Clay about it. The floor function isn't actually called on `term2`, the formatting of the SWAT theoretical documentation pdf is just really bad so the tops of square brackets have been cutoff.

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -54,17 +58,126 @@ def test_uptake_water(max_trans):
...
+
+
+@pytest.mark.parametrize("potential_uptakes,water_availabilities,available_capacities,should_fail",
+                         [([0.1, 0.2, 0.3], [9.24, 7.7, 1.31], [2.0, 3.5, 4.2], False),
+                          ([0.1, 0.2], [9.24, 7.7, 1.31], [2.0, 3.5, 4.2], True),
+                          ([0.1, 0.2, 0.4], [9.24, 7.7], [2.0, 3.5, 4.2], True),
+                          ([0.1, 0.2, 0.4], [9.24, 7.7, 1.31], [2.0, 3.5], True)])
+def test_reduce_efficiency_of_uptake(potential_uptakes: List[float], water_availabilities: List[float],
+                                     available_capacities: List[float], should_fail: bool) -> None:
+    """Tests that the reduced efficiency of uptake is calculated correctly and correct exceptions are thrown"""
```
Isgotkowitz: In the future, it would be good to write the descriptions of these tests without using 'and'. These descriptions are accurate and easy to read (and so are the tests), but by including 'and' the docstrings make it sound like the tests are dual-purpose.

Isgotkowitz (changes requested):
This is a much more robust test suite, thank you Matthew. I still found some duplicated logic and added suggestions for how they could be addressed. Also, the floor function isn't actually used in equation 5:2.2.1 - the pdf is just really poorly formatted.

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
matthew7838: That looks like a way better way to get expected, thanks!

Isgotkowitz (approved):
Nice, thanks Matthew

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -1,7 +1,9 @@
 from math import exp
+from typing import List
 
 import pytest
-from unittest.mock import MagicMock, patch
+from unittest.mock import MagicMock, patch, call
+import math
```
ew3361zh: vs_code is showing `math` as imported but never accessed.

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -54,17 +57,123 @@ def test_uptake_water(max_trans):
...
+        assert str(e.value) == "upper_depths and lower_depths must be the same length"
+    else:
+        assert pytest.approx(expected) == WaterUptake._find_stratified_max_water_uptakes(
+            root_depth, max_transpiration, water_distro_parameter, upper_depths, lower_depths)
+
+
+@pytest.mark.parametrize("root_depth,depth,max_plant_transpiration,water_distribution_parameter",
+                         [(69.4, 25.7, 33.4, 69.4),
+                          (0, 25.7, 33.4, 42.3)])
+def test_determine_max_water_uptake_to_depth(root_depth: float, depth: float, max_plant_transpiration: float,
```
ew3361zh, line 140: Getting a duplicated test name flake8 violation with this one and what looks like a placeholder test function on line 42.

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -54,17 +57,123 @@ def test_uptake_water(max_trans):
...
+
 ])
-def test_extract_water_from_soil(layers, uptakes) -> None:
+def test_extract_water_from_soil(layers, uptakes, should_fail: bool) -> None:
     """This method only tests for edge cases, other parts of the method already have coverage"""
     crop_data = CropData(actual_water_uptakes=uptakes)
     soil_data = SoilData(soil_layers=layers, field_size=3)
     uptake = WaterUptake(crop_data=crop_data)
-    try:
+    if should_fail:
```
ew3361zh, line 68: I think it's usually a good practice to split these types of if/else scenarios into separate tests. This still supplies good coverage but I think it's helpful to have the individual test (e.g. test_extract_water_from_soil fails_when_xyz, and test_extract_water_from_soil_passes_when_xyz) so if/when it fails, you know immediately what's failing based on the name of the test.

SC_redesign/Crop_and_Soil/tests_SC/crop_tests/test_water_uptake.py
```diff
@@ -54,17 +57,123 @@ def test_uptake_water(max_trans):
...
+                          ([0.1, 0.2, 0.4], [9.24, 7.7, 1.31], [2.0, 3.5], True)])
+def test_reduce_efficiency_of_uptake(potential_uptakes: List[float], water_availabilities: List[float],
+                                     available_capacities: List[float], should_fail: bool) -> None:
+    """Tests that the reduced efficiency of uptake is calculated correctly when there's no error in input"""
+    if should_fail:
+        with pytest.raises(Exception) as e:
+            WaterUptake._reduce_efficiency_of_uptake(potential_uptakes, water_availabilities, available_capacities)
+        assert str(e.value) == "potential_uptakes, water_availabilities, and available_capacities must be of equal" \
+                               " length"
+    else:
```
ew3361zh, line 93: similar comment on splitting these up. Not a requirement necessarily but I think a better practice.

ew3361zh (approved):
Looks good! Nice work, Matthew.

A couple small flake8 checks/changes and then a suggestion about splitting the tests up into more specifically-named scenario tests to get a clearer understanding to future devs for if something fails.

PR #627, opened 2023-07-05
Expand ration report - ration refactor m3

### LINKED ISSUE #531: Expand functionality of ration_report method
Ration report lists nutrient composition and concentration for a given ration, but it needs to be expanded to cover some needs in `animal_requirements.py`, and also to improve user-defined ration functionality.

The solution is fairly straightforward: we can use the existing constraint functions in `ration_NLP.py` as a reference and implement the same or similar logic into the ration_report method. Since the `ration_NLP` constraints assume a tri-decision format, it's unlikely we can use exactly the same function, however.

The following methods in `ration_NLP.py` need implementation in `ration_report.py`: 
`NEmact_constraint`, `NEl_constraint`, `NEgact_constraint`, `calcium_constraint`, `protein_constraint`, `forage_NDF_constraint`, `fat_constraint`.

Finally, these values will need to be retrievable when calling `calc_rqmts` in `animal_requirements.py`. `calc_rqmts` is called in different locations, but the nutrient composition is saved at the pen level, so it's likely that we should be able to add that report attribute to input args.

Highest priority are TDN (total digestible nutrients) retrieval for `calculate_NRC_protein_requirements`, `NDF_conc` for `calculate_NASEM_protein_requirements` and `calculate_NASEM_DMI`, and `net_energy_diet_concentration` for `calculate_NRC_DMI`

Notes
-----
net_energy_diet_concentration = NE of ration / total fed DMI in kg
To get NE: multiply ME by 0.64

JoeWaddell:
This PR has been superseded by branch `ration_refactor_m4`, though I have yet to open a PR for that branch. Changes in base branch was too complex to battle GitHub to properly cherrypick.


tests/test_animal.py
```diff
@@ -54,7 +57,7 @@ def cow_b() -> dict:
         'body_weight': 680,
         'mature_body_weight': 700,
         'day_of_pregnancy': 150,
-        'animal_type': AnimalType.LAC_COW,
+        'animal_type': AnimalType.DRY_COW,
```
Ed731551, line 68: Do you also want to add a case study for a lactating cow?

tests/test_animal.py
```diff
@@ -43,7 +43,10 @@ def cow_a() -> dict:
         'ADG_heifer': None,
         'daily_growth': None,
         'age': 1000,
-        'distance': None
+        'distance': None,
+        'NDF_conc': 0.3,
+        'TDN_conc': 0.7,
+        'net_energy_diet_concentration': 1.0
```
Ed731551, line 57: My understanding is that you want to reproduce two test cases: 1. cow(a) and 2. cow(b), lactating and dry cow respectively???. If this is a test case for a lactating cow, the net energy concentration of the diet should be higher. For this test case, I would suggest you use a value something like 1.60 Mcal/kg DM

Ed731551 (commented):
Overall, equations look. I added specific a comments on test cases (line 49 and 60).

What's the difference between ADG_heifer and daily_growth? same thing? (for example see lines 118 and 119). Thanks.

JoeWaddell:
This PR has been moved to branch `ration_refactor_m4`, though yet to open a PR. Changes in base branch was too complex to battle GitHub to properly cherrypick.

PR #628, opened 2023-07-05
Removing the None option in AnimalCombination enum class

ltp23:
This PR removes the None option in the AnimalCombination enum class located inside the Pen class. The `init_default_pens()` is also removed because the method `allocate_animals_to_pen()` is called initially anyway and it can handle the case when the input json file doesn't specify enough pens of the right type.


JoeWaddell (approved):
Looks great, runs well, and nicely tidies animal_manager.

ew3361zh (approved):
Looks great, Loi! Ran well with different input files and test suite as well.

PR #629, opened 2023-07-06
Manure-Field module connector integration

Isgotkowitz:
This PR integrates the Manure and Field modules by passing the Manure module into the `FieldManager` and having the contained `Field` objects request manure.

- When a manure application is scheduled to occur, `_execute_manure_application()` requests manure by passing a `NutrientRequest` to the `request_nutrients()` method in the `ManureManager` instance, and processes how much manure is supplied through a `NutrientRequestResults` instance.
- Note: the `ManureApplication` class that takes the supplied manure and applies it the soil uses different organic and inorganic nitrogen fractions from the ones provided in `NutrientRequestResults`. Hence the translation of the provided fractions into the ones that are passed to `ManureApplications`. 
- `_record_manure_application()` has been modified so that it records the amount of potassium applied as None if no amount is provided. This provides more accurate output, because it will more clearly indicate to whomever is handling the output that potassium is not supported in this version of RuFaS.
- `ManureManager` has been integrated into `FieldManager`, so that the `ManureManager` instance must be passed when `FieldManager` is initialized, and then `ManureManager` is passed to every `Field` instance when it is initialized.
- `Field` checks that the `ManureManager` passed to it is not None - if it is, it will throw an error.
- The tests for `_execute_manure_application()`, `_record_manure_application()`, as well as all tests that instantiate `Field` objects have been modified to meet the new requirements of the `Field` module.
- The order of initializations in `State` has changed - it now instantiates the `ManureManager`, then instantiates the `FieldManager` so that the `ManureManager` may be passed to it. The code in this PR has run the 6 complete simulations that are possible with the current input correctly.


varrma (approved):
Looks good Ed !!

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -318,14 +326,52 @@ def _execute_manure_application(self, requested_nitrogen: float, requested_phosp
...
+            supplied_nitrogen = 0.0
+            supplied_phosphorus = 0.0
 
-        unmet_nitrogen_demand = max(0.0, requested_nitrogen - manure_filled_by_request["nitrogen"])
-        unmet_phosphorus_demand = max(0.0, requested_phosphorus - manure_filled_by_request["phosphorus"])
+        unmet_nitrogen_demand = max(0.0, requested_nitrogen - supplied_nitrogen)
+        unmet_phosphorus_demand = max(0.0, requested_phosphorus - supplied_phosphorus)
         if unmet_nitrogen_demand == 0.0 and unmet_phosphorus_demand == 0.0:
             return
         elif not unmet_nitrogen_demand == 0.0 and unmet_phosphorus_demand == 0.0:
```
ltp23: I think to improve readability we can spell out the two cases

```
elif unmet_nitrogen_demand > 0.0 and unmet_phosphorus_demand == 0.0:
    optimal_mix = self.OPTIMAL_MIX  # maybe saving the string '100_0_0' as a constant somewhere
elif unmet_nitrogen_demand > 0.0 or unmet_phosphorus_demand > 0.0:
    optimal_mix = self._determine_optimal_fertilizer_mix(unmet_nitrogen_demand, unmet_phosphorus_demand, self.available_fertilizer_mixes)
```

ltp23 (approved):
The code looks good to me. Thanks Ed!

SC_redesign/Crop_and_Soil/field/field.py
Isgotkowitz: Good suggestion Loi

Issue #637, opened 2023-07-09
Build example metadata


ew3361zh:
Build a metadata json structure that fulfills the requirements for the IM:

-is a single file
-stored directly under input directory
-describes which inputs are critical and which are non-critical
-describes what validation rules are
-provides default values for non-critical data
-provides links/instructions to IM on loading and parsing input data
-can validate both json and csv from properties section

PR #638, opened 2023-07-09
Establish Metadata Structure

ew3361zh:
This PR seeks to address issue #637 to build a metadata json structure that fulfills the requirements from the IM design doc.

The starting point was the IM design doc Appendix 1 metadata structure.

Notes:
- The properties section has a single layer of complexity (i.e. no nested data validation/fixing capabilities) currently. Will expand once a structure is agreed upon and approved.
- Data defaults and validation guidelines are a mixture from guidelines in the Data Collection App and the current RuFaS input json files.
- Far from comprehensive because it's just used to establish a pattern and create a place to build the eventual final metadata structure from.
- Not sure what `cross-validation` section at the end will be used for.

### DIFF
--- a/input/example_metadata.json
+++ b/input/example_metadata.json
@@ -1,81 +1,280 @@
 {
-    "config": {
-      "start_date": "2009:1",
-      "end_date": "2009:100",
-      "graphic_dir": "output/graphics/",
-      "output_dir": "output/",
-      "random_seed": 42,
-      "simulate_animals": true
+  "config": {
+    "start_date": "2009:1",
+    "end_date": "2009:100",
+    "graphic_dir": "output/graphics/",
+    "output_dir": "output/",
+    "random_seed": 42,
+    "simulate_animals": true
+  },
+  "files": {
+    "animal": {
+      "title": "Animal data",
+      "description": "Moo Moo!",
+      "path": "animal/example_animal.json",
+      "type": "json",
+      "properties": "a"
     },
-    "files": {
-      "animal": {
-        "title": "Animal data",
-        "description": "Moo Moo!",
-        "path": "animal/barnyard_animal.json",
-        "type": "json",
-        "properties": "a"
+    "crop": {
+      "title": "Crop data",
+      "description": "Alfafa is fun!",
+      "path": "crop/example_rotation.json",
+      "type": "json",
+      "properties": "b"
+    },
+    "economy": {
+      "title": "Economy data",
+      "description": "Time is money!",
+      "path": "EEE/example_costs.csv",
+      "type": "csv",
+      "properties": "c"
+    },
+    "emission": {
+      "title": "Emissions data",
+      "description": "Let's think about global warming",
+      "path": "EEE/example_emissions.json",
+      "type": "json",
+      "properties": "d"
+    },
+    "feed": {
+      "title": "Feed data",
+      "description": "Those cows need to eat!",
+      "path": "feed/example_user_feeds.csv",
+      "type": "csv",
+      "properties": "e"
+    },
+    "manure": {
+      "title": "Manure data",
+      "description": "You thought it's a waste, but those crops need to eat as well",
+      "path": "manure/example_manure_management.json",
+      "type": "json",
+      "properties": "f"
+    },
+    "soil": {
+      "title": "Soil data",
+      "description": "Stop treating soil like dirt!",
+      "path": "soil/example_soil.json",
+      "type": "json",
+      "properties": "g"
+    },
+    "weather": {
+      "title": "Weather data",
+      "description": "Let's turn on the AC",
+      "path": "weather/example_weather.csv",
+      "type": "csv",
+      "properties": "h"
+    }
+  },
+  "properties": {
+    "a": {
+      "calf_num": {
+        "type": "number",
+        "description": "The average number of pre-weaned calves (head)",
+        "minimum": 0,
+        "default": 8
       },
-      "crop": {
-        "title": "Crop data",
-        "description": "Alfafa is fun!",
-        "path": "crop/alfalfa_rotation.json",
-        "type": "json",
-        "properties": "b"
+      "heiferI_num": {
+        "type": "number",
+        "description": "HeiferI is the number of youngstock post-weaning and before breeding (head)",
+        "minimum": 0,
+        "default": 44
       },
-      "economy": {
-        "title": "Economy data",
-        "description": "Time is money!",
-        "path": "EEE/costs.csv",
-        "type": "csv",
-        "properties": "c"
+      "heiferII_num": {
+        "type": "number",
+        "description": "HeiferII is the number of heifers eligible for breeding and in early pregnancy (head)",
+        "minimum": 0,
+        "default": 38
       },
-      "emission": {
-        "title": "Emissions data",
-        "description": "Let's think about global warming",
-        "path": "EEE/emissions.json",
-        "type": "json",
-        "properties": "d"
+      "heiferIII_num_springers": {
+        "type": "number",
+        "description": "HeiferIII is the number of close-up heifers (head)",
+        "minimum": 0,
+        "default": 5
       },
-      "feed": {
-        "title": "Feed data",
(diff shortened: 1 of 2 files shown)


input/example_metadata.json
```diff
@@ -1,81 +1,280 @@
...
+        "type": "boolean",
+        "description": "Whether or not to initialize a herd if the milk prod and body weight numbers are exceedingly abnormal",
+        "default": false
+      },
+      "breed": {
+        "type": "string",
+        "description": "The predominant breed of the herd (Holstein - HO or Jersey - JE)",
+        "default": "HO",
+        "minimum_length": 2,
+        "maximum_length": 2
```
PooyaHekmati:
```suggestion
        "pattern": "[a-zA-Z]{2}"
```
a more relaxed version would be:
```suggestion
        "pattern": ".{2}"
```

input/example_metadata.json
```diff
@@ -1,81 +1,280 @@
...
+        "type": "array",
+        "description": "Years in which the harvesting occurs",
+        "minimum_length": 1
+      },
+      "harvest_days": {
+        "type": "array",
+        "description": "Julian day(s) of year to harvest",
+        "minimum_length": 1
+      }
+
```
PooyaHekmati: A to Z, the same applies to other blobs.
```suggestion
    "b": {
       "crop_species": {
        "type": "string",
        "description": "the name of the crop being grown",
        "minimum_length": 1,
        "maximum_length": 50
      },
      "harvest_days": {
        "type": "array",
        "description": "Julian day(s) of year to harvest",
        "minimum_length": 1
      },
      "harvest_years": {
        "type": "array",
        "description": "Years in which the harvesting occurs",
        "minimum_length": 1
      },
      "planting_days": {
        "type": "array",
        "description": "Julian day(s) of year to plant",
        "minimum_length": 1
      },
      "planting_years": {
        "type": "array",
        "description": "Years in which the planting occurs",
        "minimum_length": 1
      }
```

input/example_metadata.json
```diff
@@ -1,81 +1,280 @@
...
+        "minimum": 1980,
+        "default": 2000
+      },
+      "jday": {
+        "description": "Julian day of simulation",
+        "type": "number",
+        "minimum": 1,
+        "maximum": 366,
+        "default": 1
+      },
```
PooyaHekmati: We already have these two in the config blob.

PooyaHekmati (approved):
I left some comments, but they are not big enough to block the PR. Don't worry about the correctness of the data presented in this file, we will fix it. The `cross-validation` section is a placeholder for the metadata that we need for requirement no. 3. We will get to it later.

input/example_metadata.json
ew3361zh: This was specifically for the weather data - I can be clearer in my descriptions that this is weather csv data specific julian day. I believe we do need this data from the weather csv in order to align weather data with simulation day.

input/example_metadata.json
```diff
@@ -1,81 +1,273 @@
...
-      "random_seed": 42,
-      "simulate_animals": true
+  "config": {
+    "start_date": "2009:1",
+    "end_date": "2009:100",
+    "graphic_dir": "output/graphics/",
+    "output_dir": "output/",
+    "random_seed": 42,
+    "simulate_animals": true
+  },
```
Isgotkowitz: Is the config blob designed to always have these keys? If so, should it also have a `properties`?

input/example_metadata.json
```diff
@@ -1,81 +1,273 @@
...
+      "type": "json",
+      "properties": "g"
+    },
+    "weather": {
+      "title": "Weather data",
+      "description": "Let's turn on the AC",
+      "path": "weather/example_weather.csv",
+      "type": "csv",
+      "properties": "h"
+    }
```
Isgotkowitz: I think this metadata structure looks good, but I want to confirm that we will have some blobs within the `files` section that will be strictly necessary (i.e. need to have an exact name, like `weather` and `fields`) and some that will be variable (like having two `soil` blobs, `soil1` and `soil2`). @PooyaHekmati can you also confirm this?

Isgotkowitz (approved):
Looks good Niko, two comments. Config properties is less important, required blob names is very important - have asked for Pooya's input as well.

input/example_metadata.json
PooyaHekmati: I met with Niko today and we decided to save the config in json file and treat it like other files.

input/example_metadata.json
PooyaHekmati: The metadata is not something that the end user interacts with. Any deficiencies in metadata lead to unexpected runtime behaviors. If I understand correctly, your concern/question is that we are going to access this data in the code, so if we want to access something that does not exist, that will be a failure. That is true but is the same as asking what happens if the user changes the content of a Python file. We don't worry about this, for the same reason we don't worry about the metadata. We ensure the correctness of the code by unit testing and integration testing, and the correctness of the metadata by cross-validation (yet to be designed) and end-to-end testing.
Another way of stating what I wrote is that the metadata is an internal component of RuFaS. It can only be modified by those who know about its existence and details.

input/example_metadata.json
PooyaHekmati: I don't think so. This is duplication and duplication is bad.
It also violates requirement No. 7. Simulation starts from day `x` and continues to day `y`. `x` and `y` are stored in the config blob. There is no reason to store them somewhere else as well.

ew3361zh: Right but the weather data is just a big CSV of historical weather data with the column headers:
year,jday,rainfall,tmax,tmin,tavg,hday,irrigation

So for example the LT_weather CSV has daily weather data starting from Julian day 1 of 1980 through Julian Day 365 of 2019. If the simulation started on Julian Day 150 of 1995, we'd need some reference in the LT_weather CSV side to be able to find the weather for that specific day and continue from that point in the weather data. I'm not sure how we'd know how to align them without this.

input/example_metadata.json
```diff
@@ -1,81 +1,309 @@
...
+      "type": "csv",
+      "properties": "i"
+    }
+  },
+  "properties": {
+    "a": {
+      "start_date": {
+        "type": "string",
+        "description": "The year and Julian day on which the simulation will start.",
+        "pattern": "[12][019][0-9]{2}:[1-9][0-9]{0,2}",
```
ltp23: I don't know if this is relevant yet since this is an example file. This pattern also allows some invalid combinations like any day number more than 366 and any mismatch between a leap year and day number.

input/example_metadata.json
```diff
@@ -1,81 +1,309 @@
...
-      "f": {},
-      "g": {},
-      "h": {}
+    "b": {
+      "calf_num": {
+        "type": "number",
+        "description": "The average number of pre-weaned calves (head)",
+        "minimum": 0,
+        "default": 8
+      },
```
ltp23, line 111: In general, is it also possible to specify a 'maximum' field?

ltp23 (approved):
The structure looks good to me. Thanks Niko.

input/example_metadata.json
ew3361zh: Good catch, Loi. I realized this pattern is much more complicated than I had initially thought and have updated it to prevent day numbers greater than 366. 

Can you expand on what you mean in the second part about allowing mismatches between a leap year and day number? I want to be sure I account for that but not sure I understand. Thanks!

input/example_metadata.json
ew3361zh: Yes! It is possible to specify a "maximum" field.
The guidelines for validation of each of the 4 allowed data types are laid out well in the Input Manager design doc pages 12-13.

input/example_metadata.json
Isgotkowitz: Ok, that puts my mind more at ease. Thank you Pooya

input/example_metadata.json
PooyaHekmati: We would know that by looking into the config blob. There is no need to duplicate the date here.

PR #640, opened 2023-07-10
test coverage for SC_redesign branch (2/3)

matthew7838:
### Summary
This is a second part of coverage to ensure 100% test coverage on the Soil and Crop branch. The next PR regarding this will be the last in the series if no other major tests are missing in the rest of the files.

### Changes
- Coverage for `perform_annual_field_reset` in `FieldData`.
- Error handling for `manure_application`
- Complete tests for equality and hash function for all the event's children classes.
- Added test case that checks ` FertilizerSchedule` with None values for depth and fraction.
- Added a test to fully `cover is_leap_year` in the `schedule.py`
- Added test cases and condition statements to check that the main routine of `Infiltration` accounts for the specific cases of input
- Added a test to test the `water_factor` property method in layer_data.py


SC_redesign/Crop_and_Soil/tests_SC/field_tests/test_manure_application.py
```diff
@@ -286,40 +286,55 @@ def test_apply_grazing_manure(dry_mass: float, dry_fraction: float, phosphorus_m
...
-                                                                    coverage, area, expected_weiP_frac, inorganic_frac,
-                                                                    ammonium_frac, organic_frac)
+    if should_fail:
+        with pytest.raises(ValueError) as e:
+            data = SoilData(field_size=1.1)
+            incorp = ManureApplication(data)
+
+            incorp._determine_water_extractable_inorganic_phosphorus_fraction_by_animal = MagicMock(return_value=0.25)
+            incorp._apply_liquid_machine_manure = MagicMock()
+            incorp._apply_solid_machine_manure = MagicMock()
```
Isgotkowitz: Would it be possible to move this setup outside of the `if should_fail:` control block? If so, I think that would be good so that it doesn't have to be duplicated in both branches of the `if else` statement.

SC_redesign/Crop_and_Soil/tests_SC/field_tests/test_manure_application.py
```diff
@@ -286,40 +286,55 @@ def test_apply_grazing_manure(dry_mass: float, dry_fraction: float, phosphorus_m
...
+
+            incorp.apply_machine_manure(dry_mass, dry_fraction, total_phosphorus_mass, coverage, area, inorganic_frac,
+                                        ammonium_frac, organic_frac, weiP_frac, source_animal)
+        assert str(e.value) == f"Water extractable inorganic phosphorus fraction must be in the range [0.0, 0.95]," \
+                               f" received '{weiP_frac}'."
```
Isgotkowitz, line 309: Since there are some subroutines that get mocked they should be checked that they were called the appropriate number of times, or if you think it is not really necessary that they be checked at all when checking error handling then they don't have to be mocked at all.

SC_redesign/Crop_and_Soil/tests_SC/manager_tests/test_events.py
```diff
@@ -22,15 +23,184 @@ def test_occurs_today(year: int, day: int, current_year: int, current_day: int,
...
+def test_event_equality(event1, event2, expected: bool) -> None:
     """Tests that equality is tested correctly between Event objects."""
-    event1 = Event(year1, day1)
-    event2 = Event(year2, day2)
     actual = event1 == event2
     assert actual == expected
+
+
+@pytest.mark.parametrize("event,expected", [
+    (Event(1, 120), hash((1, 120)))
```
Isgotkowitz: I think a few more test cases would be helpful here, to give a greater degree of confidence that all edge cases are being covered. Also, I think it would be better to pass the expected hash value, instead of the hash call, like
```python
(Event(1, 120), -7023525560569449493)
```
(if that is possible - I assume it is. But if the hash function gets reseeded or something every time the program is run, then this strategy would not work and the current way is fine)

Edit: an even more thorough testing strategy for hashing may be to pass the Event, a hash value, and the expected equality between the hash of the Event and the hash value (this is just a suggestion and is kind of overkill in this case, so take it or leave it). This would make it possible to explicitly verify that edge cases do not break the hashing method. For example, it would be possible to explicitly check that `hash((1, 120))` does not equal `hash((11, 20))`. Fortunately Loi caught an error in my original implementation of the hashing method that would have allowed similar mistakes and suggested the current hashing method, so it is pretty robust.

SC_redesign/Crop_and_Soil/tests_SC/manager_tests/test_events.py
Isgotkowitz, line 1: Nice job with this test suite Matthew, I really appreciate how thorough it is. Also, I apologize for not having written a test suite when I implemented these modules, that is my bad.

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_infiltration.py
```diff
@@ -181,32 +181,40 @@ def test_determine_moisture_condition_parameter(retention_param):
...
                                                field_size=1.33),
                                      LayerData(top_depth=20, bottom_depth=50, soil_water_concentration=0,
                                                field_size=1.33)])
     incorp = Infiltration(data)
-    assert incorp.data.average_subbasin_slope == 0.07
-    assert incorp.data.previous_retention_parameter == 27
+    if incorp.data.previous_retention_parameter is None:
+        status = True
+    else:
+        status = False
```
Isgotkowitz: Is there a more specific variable name that could be used here?

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_layer_data.py
```diff
@@ -0,0 +1,27 @@
...
+                      saturation_content: float, expected: float) -> None:
+    """Tests that water factor was calculated correctly"""
+    with patch('SC_redesign.Crop_and_Soil.soil.layer_data.LayerData.soil_water_content',
+               new_callable=PropertyMock, return_value=soil_water_content):
+        with patch('SC_redesign.Crop_and_Soil.soil.layer_data.LayerData.field_capacity_content',
+                   new_callable=PropertyMock, return_value=field_capacity_content):
+            with patch('SC_redesign.Crop_and_Soil.soil.layer_data.LayerData.wilting_point_content',
+                       new_callable=PropertyMock, return_value=wilting_point_content):
+                with patch('SC_redesign.Crop_and_Soil.soil.layer_data.LayerData.saturation_content',
+                           new_callable=PropertyMock, return_value=saturation_content):
```
Isgotkowitz: This test looks really good Matthew and no changes are necessary, but I wanted to make you aware of some handy syntax for using context managers to mock things.
An equivalent way to mock these functions would be 
```python
    with patch('SC_redesign.Crop_and_Soil.soil.layer_data.LayerData.soil_water_content',
               new_callable=PropertyMock, return_value=soil_water_content) as mocked_soil_water, \
        patch('SC_redesign.Crop_and_Soil.soil.layer_data.LayerData.field_capacity_content',
              new_callable=PropertyMock, return_value=field_capacity_content) as mocked_field_cap, \
        patch('SC_redesign.Crop_and_Soil.soil.layer_data.LayerData.wilting_point_content',
              new_callable=PropertyMock, return_value=wilting_point_content) as mocked_wilt_point, \
        patch('SC_redesign.Crop_and_Soil.soil.layer_data.LayerData.saturation_content',
              new_callable=PropertyMock, return_value=saturation_content) as mocked_sat_content:
```
Keeping all the patches in the same with statement is nice because it helps avoid the nesting/makes the code more readable. Another way to accomplish this is to use `patch.multiple` (an example is in `test_excess_water_available()` in `test_soil_data_and_config_factory.py`), though I personally have had found it to more finicky to use.

Also, adding the `as mocked_<method_name>` allows all the assert statements normally used on mocked methods to be used inside the context manager as well (`assert_called_once()`, `call_count`, etc)

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_layer_data.py
Isgotkowitz, line 1: Kudos on creating this file. There are a number of other `LayerData` tests, but they are in `test_soil_data_and_soil_config.py`. That file has become quite large and definitely violates the single-responsibility principle, so the `LayerData` tests should be moved to this file at some point - I will create an issue for this.

Isgotkowitz (approved):
Excellent PR Matthew. Multiple comments with suggestions about minor ways that readability could be improved and how testing could be expanded, but are all take it or leave it.

SC_redesign/Crop_and_Soil/tests_SC/field_tests/test_manure_application.py
matthew7838: They are still needed when there's no value error, so I'll just move them down when it doesn't fail.

SC_redesign/Crop_and_Soil/tests_SC/manager_tests/test_events.py
matthew7838: Yes, they do get reseeded. For now, I'll just add more cases.

PR #644, opened 2023-07-11
Updating storage methane emission calculations

ltp23:
This PR updates the equation for calculating methane emission for slurry storage systems. This is done via the `calc_methane_emission_from_slurry_storage` method. 

The `_calc_arrhenius_exponent` and `_calc_volatile_solid_components` methods are helper methods that compute the Arrhenius exponent and the degradable and non-degradable volatile solid components respectively, which are used in the methane emission calculation. 

I also added unit tests for the changes I made.


varrma (approved):
Looks good Loi !!

RUFAS/routines/manure/constants/gas_emission_constants.py
```diff
@@ -1,27 +1,30 @@
 class GasEmissionConstants:
     """Constants used in gas emission calculations."""
 
-    b1 = 1.0
-    """Rate correcting factor (unitless)."""
+    METHANE_EMISSION_COEFFICIENT = 24
+    """Methane emission coefficient, used in calculation of slurry storage methane emission (unitless)."""
```
ew3361zh, line 5: Is there another way to help explain what this is? I think the repeating of the name of variable in the definition could be changed to explain better what the coefficient is. Like for example the other two say "rate correcting factor for...". Is there something similar for this one? Or maybe it's just changing it to "The coefficient used in calculation of slurry storage methane emission"

ew3361zh (approved):
Nice work, Loi! I had one comment about the definition of a variable not being as clear as I think it could be but otherwise, this looks great and testing was working well.

JoeWaddell (approved):
This all looks great! Nice work.

PR #645, opened 2023-07-11
Carbon Initialization Bug(s)

Isgotkowitz:
This PR refines the way that the carbon cycling submodule initializes and handles some of the carbon pools within the soil.
Before this PR, all carbon pools were initialized as being empty, which resulted in 0 activity in the carbon module. With the pools being set to reasonable defaults, there is now carbon decomposition and activity that can be observed to verify if it is behaving in a somewhat reasonable way.

- `decomposition.py`
  - The decomposition factors calculated in this module (one based on temperature, the other moisture) are now both lower-bounded at 0.0. The tests for these methods have been updated as well.
  - This prevents situations where negative amounts of decomposition are computed.
  - The tests for these methods have been updated to reflect the changes made.
- `layer_data.py`
  - The active, passive, and slow carbon pools are initialed as `None` and initialized by `_initialize_carbon_pools()`, which is called in `post_init()`.
  - This new method and its integration into `post_init()` is tested in `test_soil_data_and_config_factory.py`.
  - The `soil_water_content` property of `LayerData` has been removed - it is a field which represents the same thing as `water_content`, and it's uses have been replaced with `water_content`.
  - `@property` `silt_clay_content()` has been changed so that it computes the combined fraction of silt and clay content in a soil layer instead of the silt to clay content ratio in the soil. ***NOTE*** this is not necessarily correct, docstring for details.

***Important*** The carbon cycling module is still not fully functional. The main thing it currently lacks is the ability to have new carbon enter the system via residue left and tilled into the soil. A major obstacle to implementing this is that I am pretty sure the old carbon cycling implementation, upon which ours is based, has a very significant design flaw. I can go into more detail if necessary, but it boils to down to the module not properly distinguishing between residue carbon on the soil surface and carbon residue in the soil profile. Fixing this would be complicated and time-intensive (time we may not have) and before implementing a fix it should be discussed how executing that fix will fit in with implementing other V1 features and the evaluation.


SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -396,6 +398,31 @@ def _initialize_nitrogen_pools(self, field_size: float, residue: float) -> None:
...
+
+        Parameters
+        ----------
+        field_size: float
+            Size of the field (ha).
+
+        Notes
+        -----
+        This is an extremely simply and arbitrary way of initializing carbon pools in the soil, and is intended to be a
+        temporary solution until SWAT-C is implemented. TODO - #512
```
JoeWaddell, line 413: I like that you already have issues open!

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -396,6 +398,31 @@ def _initialize_nitrogen_pools(self, field_size: float, residue: float) -> None:
...
+
+        """
+        soil_volume_in_cubic_meters = self.layer_thickness * (field_size * HECTARES_TO_SQUARE_MILLIMETERS) * \
+            CUBIC_MILLIMETERS_TO_CUBIC_METERS
+        soil_mass_in_kg = self.bulk_density * MEGAGRAMS_TO_KILOGRAMS * soil_volume_in_cubic_meters
+
+        one_third_soil_carbon = (1 / 3) * (soil_mass_in_kg * (self.percent_organic_carbon_content / 100) / field_size)
+
+        self.active_carbon_amount = one_third_soil_carbon
```
JoeWaddell, line 422: So is the idea in this simple implementation to partition carbon into these three places, and that the full implementation will better allocate carbon to each of the three `LayerData` attributes?

JoeWaddell (approved):
Looks good to me! Just a couple minor comments

SC_redesign/Crop_and_Soil/soil/layer_data.py
```diff
@@ -396,6 +398,31 @@ def _initialize_nitrogen_pools(self, field_size: float, residue: float) -> None:
...
+        -----
+        This is an extremely simply and arbitrary way of initializing carbon pools in the soil, and is intended to be a
+        temporary solution until SWAT-C is implemented. TODO - #512
+
+        """
+        soil_volume_in_cubic_meters = self.layer_thickness * (field_size * HECTARES_TO_SQUARE_MILLIMETERS) * \
+            CUBIC_MILLIMETERS_TO_CUBIC_METERS
+        soil_mass_in_kg = self.bulk_density * MEGAGRAMS_TO_KILOGRAMS * soil_volume_in_cubic_meters
+
+        one_third_soil_carbon = (1 / 3) * (soil_mass_in_kg * (self.percent_organic_carbon_content / 100) / field_size)
```
matthew7838, line 420: Is there a reference to this temporary one-third assumption? If so, add them into docStrings.

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_soil_data_and_config_factory.py
```diff
@@ -352,24 +352,44 @@ def test_layer_thickness_error(top: float, bottom: float) -> None:
...
+        # Check everything
+        assert layer.water_content == expected_water_content
+        calc_psp.assert_called_once_with(layer.percent_clay_content, 25, layer.percent_organic_carbon_content)
+        assert determine_nutrient_density.call_count == 3
+        assert layer.mean_phosphorus_sorption_parameter == 0.5
+        assert layer.labile_inorganic_phosphorus_content == 22
+        assert layer.active_inorganic_phosphorus_content == 22
+        assert layer.stable_inorganic_phosphorus_content == 22
+        init_nitrogen_pools.assert_called_once()
+        init_carbon_pools.assert_called_once()
```
matthew7838, line 378: Nice addition to these tests.

matthew7838 (approved):
Great PR, Nice tests added. However, even with this PR merged there still seems to be a major problem regarding carbon suit that might be unsolvable under current circumstances. Is there any temporary solution before SWAT-C implementation?

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: Yes, the idea in this simple implementation is to have carbon partitioned between the three pools in each layer. But the full implementation should have the capacity to have carbon enter the system via residue, which it currently does not.

SC_redesign/Crop_and_Soil/soil/layer_data.py
Isgotkowitz: There is not a reference, but I will make it clear in the docstrings that the one-third of carbon in each pool is an arbitrary assumption that I have made.

SC_redesign/Crop_and_Soil/tests_SC/soil_tests/test_soil_data_and_config_factory.py
Isgotkowitz: Thanks Matthew:)

Isgotkowitz:
> Great PR, Nice tests added. However, even with this PR merged there still seems to be a major problem regarding carbon suit that might be unsolvable under current circumstances. Is there any temporary solution before SWAT-C implementation?

Right now the biggest issue I see in the carbon module is that carbon is not able to enter the system through residue. Unfortunately the mechanisms for this seem much more complex than simply having a couple pools that get initialized at the beginning of the simulation and having them move exchanged carbon after that.

PR #646, opened 2023-07-11
IM fix_data()

### LINKED ISSUE #621: Create fix_data function in IM
Closely associated with issue #616 and #617, this issue is to create a new function in the IM class called `fix_data()` which will attempt to fix any non-critical invalid data in the IM data pool and return a boolean based on the result of that attempt. Should include unit tests for this new function.

allisterakun:
# #621  `_fix_data()` in `RUFAS.input_manager`:

> Note: this PR implements the `fix_data()` function in the Input Manager and the related unit tests. All changes to other parts of the code are due to resolving conflicts of merging to the base branch `im_refactor`. Please ignore all unrelated changes and only review the relevant ones.

## `_fix_data() -> bool:`

This function attempts to fix the invalid data with the provided default value. 

It takes in the path to the data, and checks if the variable of interest is critical data by checking if a default value is provided for this variable in the metadata. If the data is fixable (not critical data), this function will replace the value of the variable of interest in the IM variable pool with the provided default value; else, it will return False, indicating critical data (data unfixable).

#### `variable_properties: Dict[str, Any]`
The properties for the variable of interest.

#### `element_hierarchy : List[str]`
A list of strings indicating the path to reach the variable of interest in `self.__metadata` and `self.__pool`.

#### `input_data: Dict[str, Any]`
A buffer dictionary that holds the input data for validation and fixing.

#### bool

- **False** If no default value is found (aka critical data), the data is unfixable. 
- **True** If the data is successfully fixed with the provided default value.


allisterakun:
@ew3361zh As I was implementing the fix_data() function, I found that `self.__pool[module_key][variable_name]` (for example, `self.__pool["animal"]["herd_information"]`) is a nested structure. Will speak to you about this later Wednesday at the work-together session.

allisterakun:
- [x] Feature Implementation
- [ ] Docstring
- [ ] Logging
- [ ] Unit tests

allisterakun:
- [x] Feature Implementation
- [x] Docstring
- [x] Logging
- [ ] Unit tests

allisterakun:
- [x] Feature Implementation
- [x] Docstring
- [x] Logging
- [x] Unit tests

RUFAS/input_manager.py
```diff
@@ -320,29 +320,70 @@ def _validate_string_type_element(self, variable_to_check: Dict[str, Any], var_n
...
+        Return True if the data is fixed, False for critical data
 
         Parameters
         ----------
-        key : str
-            The key of the data to fix.
-
-        value : Any
-            The value of the data to fix.
+        variable_path : List[str]
```
PooyaHekmati: This doesn't exist in the function signature.
Please add the args to docstring.

RUFAS/input_manager.py
```diff
@@ -320,29 +320,70 @@ def _validate_string_type_element(self, variable_to_check: Dict[str, Any], var_n
...
+                return is_valid_string
+        if is_maximum_length:
+            is_valid_string = len(input_data_value) <= variable_to_check["maximum_length"]
+            if not is_valid_string:
+                om.add_warning("String out length range.", f"{var_name=}", info_map)
+                return is_valid_string
+
+        return True
+
+    def _fix_data(self, module_key: str, property_map_key: str, element_hierarchy: List[str]) -> bool:
```
PooyaHekmati: Can you come up with better names? These names are not clear, even to me.

RUFAS/input_manager.py
```diff
@@ -320,29 +320,70 @@ def _validate_string_type_element(self, variable_to_check: Dict[str, Any], var_n
 
         return is_valid_string
 
-    def _fix_data(self, key: str, value: Any) -> bool:
+    def _validate_string_type_element(self, variable_to_check: Dict[str, Any], var_name: str,
```
PooyaHekmati: Isn't this PR about fix_data? What is this function doing here?

RUFAS/input_manager.py
```diff
@@ -320,29 +320,70 @@ def _validate_string_type_element(self, variable_to_check: Dict[str, Any], var_n
...
-        value : Any
-            The value of the data to fix.
+        variable_path : List[str]
+            The path to reach the variable of interest.
 
         Returns
         -------
         bool
-            True if the data is fixed, False otherwise.
+            True if the data is fixed, False for critical data.
```
PooyaHekmati: Let's not use critical data because then we have to define what it means. This terminology was used in the design phase to differentiate the cases, but I don't think it is necessary here. We know what it means today, but those who join later have to figure it out. Let's be kind to them.

PooyaHekmati (changes requested):
The logic seems to be correct; however, the docstring is not correct. Also, the variable names can be improved both in the function signature and inside the function body. You have also _validate_string_type_element but there is no mention of it in the PR title or description.

RUFAS/input_manager.py
```diff
@@ -355,29 +357,76 @@ def _validate_string_type_element(self, variable_to_check: Dict[str, Any], var_n
...
-        pass
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._fix_data.__name__,
+                    }
+
+        variable_metadata: Dict = reduce(lambda d, key: d[key], element_hierarchy,
+                                         self.__metadata['properties'][property_map_key])
+        if 'default' in variable_metadata.keys():
+            variable_parent = reduce(lambda d, key: d[key], element_hierarchy[:-1],
+                                     self.__pool[module_key])
```
ew3361zh: The variable parent will need to come from the temporary variable holding the data loaded from the file, not the `self.__pool`. With the way I was asked to refactor `_validate_data()`, a variable won't get added to the pool until after it's shown to be valid (or is fixed). 
So `_validate_element()` now takes the full data loaded from each file which you can use in `_fix_data()`.

RUFAS/input_manager.py
```diff
@@ -355,29 +357,77 @@ def _validate_string_type_element(self, variable_to_check: Dict[str, Any], var_n
...
+                om.add_warning("String out length range.", f"{var_name=}", info_map)
+                return is_valid_string
+
+        return True
+
+    def _fix_data(self, module_key: str, property_map_key: str, element_hierarchy: List[str],
+                  input_data: dict[str, Any]) -> bool:
         """
         Attempt to fix the invalid data.
+        Return True if the data is fixed, False for critical data
```
ew3361zh: Is there a reason this isn't in the Returns section of the docstrings? Seems like it could be confusing to have to separate statements about what this method returns that aren't the same. I think this return statement is a better of the two.

RUFAS/input_manager.py
```diff
@@ -355,29 +357,77 @@ def _validate_string_type_element(self, variable_to_check: Dict[str, Any], var_n
...
+        module_key : str
+            The module where data of interest belongs to.
+
+        property_map_key: str
+            The metadata properties section keyword for the data of interest.
 
-        value : Any
-            The value of the data to fix.
+        element_hierarchy: List[str]
+            A list of strings indicating the path to reach the variable of interest in self.__metadata and self.__pool.
```
ew3361zh, line 363: add docstring for input_data

ew3361zh (changes requested):
Looks really good, Allister! I have a couple notes about docstrings to update and then it looks like there's some conflicts to resolve based on changes to the im_refactor branch from this morning.

Also I think in one of the previous resolved conflicts it looks like the conflicting `_validate_string_type_element` didn't get removed so it's in the code twice which is why it's showing up as new code here.

Screenshot 2023-07-20 at 12 15 00 PM

RUFAS/input_manager.py
```diff
@@ -357,29 +357,43 @@ def _validate_string_type_element(self, variable_properties: Dict[str, Any], var
...
+
+        input_data: dict[str, Any]
+            A buffer dictionary that holds the input data for validation and fixing.
```
ew3361zh, line 365: Nice! I like this, I may update im_refactor `_validate_element` docstrings with this definition as well.

ew3361zh (approved):
Nice work, Allister. LGTM.

RUFAS/input_manager.py
```diff
@@ -356,29 +356,43 @@ def _validate_string_type_element(self, variable_properties: Dict[str, Any], var
...
 
         Returns
         -------
         bool
-            True if the data is fixed, False otherwise.
+            True if the data is fixed, False if there is no default value found.
```
PooyaHekmati: I think this is too specific, we will change this function down the road and this description won't be true anymore. Can we do something like this to satisfy that as well?
```suggestion
            True if the data is fixed, False otherwise.
```

RUFAS/input_manager.py
```diff
@@ -356,29 +356,43 @@ def _validate_string_type_element(self, variable_properties: Dict[str, Any], var
...
+                  input_data: dict[str, Any]) -> bool:
         """
         Attempt to fix the invalid data.
 
         Parameters
         ----------
-        key : str
-            The key of the data to fix.
+        module_key : str
+            The module where data of interest belongs to.
```
PooyaHekmati: What does module mean here?

PooyaHekmati (approved):

RUFAS/input_manager.py
```diff
@@ -356,29 +356,43 @@ def _validate_string_type_element(self, variable_properties: Dict[str, Any], var
...
-        # where element is fixed, place this warning:
-        # om.add_warning("Data fixed", f"Invalid data fixed: {key=}; {value=}", info_map)
-        # where data is not fixable:
-        # om.add_error("Data not fixable.", f"Unable to fix the invalid data: {key=}, {value=}.", info_map)
-        pass
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._fix_data.__name__,
+                    }
+
+        variable_metadata: Dict = reduce(lambda d, key: d[key], element_hierarchy,
```
michaelRichards99:
```suggestion
        variable_metadata: Dict[str, Any] = reduce(lambda d, key: d[key], element_hierarchy,
```
Adds more clarity.

RUFAS/input_manager.py
```diff
@@ -356,29 +356,43 @@ def _validate_string_type_element(self, variable_properties: Dict[str, Any], var
...
+                                         self.__metadata['properties'][property_map_key])
+        if 'default' in variable_metadata.keys():
```
michaelRichards99:
```suggestion
        if 'default' not in variable_metadata.keys():
            return False
        variable_parent = reduce(lambda d, key: d[key], element_hierarchy[:-1],
                                     input_data[module_key])
        variable_parent[element_hierarchy[-1]] = variable_metadata['default']
        om.add_warning("Data fixed",
                           f"Invalid data fixed: {element_hierarchy[-1]} => {variable_metadata['default']}",
                           info_map)
        return True
```
Do the error check first to remove indentation.

RUFAS/input_manager.py
```diff
@@ -356,29 +356,43 @@ def _validate_string_type_element(self, variable_properties: Dict[str, Any], var
 
         return is_valid_string
 
-    def _fix_data(self, key: str, value: Any) -> bool:
+    def _fix_data(self, module_key: str, property_map_key: str, element_hierarchy: List[str],
+                  input_data: dict[str, Any]) -> bool:
         """
         Attempt to fix the invalid data.
```
michaelRichards99, line 354:
```suggestion
        Attempt to fix the invalid non-critical data.
```
Should we mention that `_fix_data` only fixes non-critical data and that it doesn't fix critical data?

michaelRichards99 (changes requested):

michaelRichards99 (approved):

Issue #641, opened 2023-07-10
Save keys as column headers and 'values' fields as column values (OM CSV #1)


ltp23:
In this step, we want to process each key-value pair in the aggregate json file generated by the output manager. For each pair, we should save the key as a column header and the 'values' field of its value dictionary as the column values in a separate CSV file. 

The 'values' field can be a list of simple types (int, float, str, bool) or a list of more complex types. For the latter, let's use `str()` method to handle them for now.

We can keep the keys as-is without any string processing. We'll save each key to a separate CSV file. Each of these CSV files will be stored in the `output/CSVs/om/variables/` directory. 

This CSV output handling process should occur at the end of the simulation to make sure all data is captured.

For example:

```json
{
    "simple_key": {
        "values": ["Hi", "How", "Are", "You"]
    },
    "list_key": {
        "values": [
            [1, 2, 3],
            [4, 5, 6]
        ]
    }
}
```

The CSV output for the simple_key.csv would be:

```json
simple_key
"Hi"
"How"
"Are"
"You"
```

And the CSV output for the `list_key.csv` would be:

```json
list_key
"[1, 2, 3]"
"[4, 5, 6]"
```

Checklist:

- [x] For each key-value pair in the aggregate json, create a new CSV file using the key as the file name.
- [x] Save the key as a column header in its respective CSV file.
- [x] For each key-value pair, save the 'values' field of its value dictionary as the column values.
- [x] If the 'values' field consists of simple types (int, float, str, bool), save them as they are.
- [x] If the 'values' field consists of more complex types, convert them into string using `str()` method.
- [x] Save each CSV output file into the `output/CSVs/om/variables/` directory.
- [x] Check the correctness of the output, i.e., each CSV file is correctly formatted (e.g., headers and values align correctly), and the files contain the correct data.
- [x] Write unit tests:
     - [x] Check if the JSON to CSV conversion is performed correctly.
     - [x] Check that complex data types are correctly stringified.
     - [x] Check that the CSV file is saved at the correct location.

michaelRichards99:
Merged.

PR #650, opened 2023-07-12
Save simulation variables and the values into a CSV file.

michaelRichards99:
This address part 1 of the CSV generation for the OutputManager: issue #641.

It pads the values with None, which are inserted into the csv file as nothing instead of the string "None".

### DIFF
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -1,11 +1,13 @@
 # !/usr/bin/env python3
 
+from enum import Enum
 from pathlib import Path
 import re
 from typing import Any, Dict, List, Optional, Union
 import json
 import os
 import datetime
+import csv
 
 from RUFAS.util import Utility
 
@@ -34,6 +36,14 @@ class OutputManager(object):
 
     __instance = None
     pool_element_type = Dict[str, List[Dict[str, Any]]]
+
+    class OutputFormat(Enum):
+        """
+        Enumeration that represents valid output encodings for OutputManger pools
+        """
+        JSON = 0
+        CSV = 1
+
 
     def __new__(cls):
         if not hasattr(cls, "instance"):
@@ -307,6 +317,48 @@ class OutputManager(object):
         except Exception as e:
             raise e
 
+    def _save_variables_to_csv(self, data_dict: Dict[str, Any], path: str) -> None:
+        """Saves variables and their values to a csv file.
+
+        Parameters
+        ----------
+        data_dict : Dict[str, Any]
+            The variable dictionary to be saved
+        path : str
+            The path to the file to be saved
+
+        Raises
+        ------
+        Exception
+            If an error occurs while writting to the file
+        """
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._save_variables_to_csv.__name__,
+                    }
+        self.add_log("_save_variables_to_csv", f"Attempting to save to {path}.", info_map)
+        if len(data_dict) == 0:
+            self.add_warning("_save_variables_to_csv", f"No variables in the pool to save to {path}", info_map)
+            with open(path, "w") as file:
+                file.truncate(0)
+            return
+
+        keys = list(data_dict.keys())
+        values = list(data_dict.values())
+        # Determine the maximum length of values to account for uneven lengths
+        max_length = max(len(d['values']) for d in values)
+
+        rows = []
+        for i in range(max_length):
+            row = [d['values'][i] if i < len(d['values']) else None for d in values]
+            rows.append(row)
+
+        with open(path, 'w', newline='') as file:
+            writer = csv.writer(file)
+            writer.writerow(keys)
+            writer.writerows(rows)
+
+        self.add_log("_save_variables_to_csv", f"Successfully saved to {path}.", info_map)
+
     def _list_to_file_txt(self, data_list: List[str], path: str) -> None:
         """Saves a list into a text file
 
@@ -487,7 +539,7 @@ class OutputManager(object):
             file_path = os.path.join(save_path, self._generate_file_name(f"saved_variables_{filter_file}", "json"))
             self._dict_to_file_json(filtered_pool, file_path)
 
-    def dump_variables(self, path: str, exclude_info_maps: bool = False) -> None:
+    def dump_variables(self, path: str, format: OutputFormat, exclude_info_maps: bool = False) -> None:
         """
         Dumps variables_pool into a json file in the given path to a directory.
 
@@ -504,8 +556,12 @@ class OutputManager(object):
         if exclude_info_maps:
             pool = self._exclude_info_maps(self.variables_pool)
 
-        file_path = os.path.join(path, self._generate_file_name("all_variables", "json"))
-        self._dict_to_file_json(pool, file_path)
+        file_path = os.path.join(path, self._generate_file_name("all_variables", format.name.lower()))
+        match format:
+            case OutputManager.OutputFormat.JSON:
+                self._dict_to_file_json(pool, file_path)
+            case OutputManager.OutputFormat.CSV:
+                self._save_variables_to_csv(pool, file_path)
 
     def dump_logs(self, path: str) -> None:
         """
@@ -607,7 +663,8 @@ class OutputManager(object):
         """
         dumps all pool into the given path to a directory.
         """
-        self.dump_variables(path, exclude_info_maps)
+        self.dump_variables(path, OutputManager.OutputFormat.JSON, exclude_info_maps)
+        self.dump_variables(path, OutputManager.OutputFormat.CSV, exclude_info_maps)
         self.dump_variable_names_and_contexts(path, exclude_info_maps)
         self.dump_errors(path)
         self.dump_logs(path)
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -57,14 +57,6 @@ class OutputManager(object):
                 },
             )
 
-    def _csv_output_directory(self, base_output_dir: str) -> Path:
-        """Directory where csv files are written to."""
-        return Path(base_output_dir) / "CSVs" / "om"
-
-    def _csv_variable_output_directory(self, base_output_dir: str) -> Path:
-        """Directory where variable pool's csv files are written to."""
-        return self._csv_output_directory(base_output_dir) / "variables"
-
     def _pool_element_factory(self) -> pool_element_type:
         """Factory for elements added to pools"""
         info_maps: List[Dict[str, Any]] = []
@@ -670,7 +662,7 @@ class OutputManager(object):
         """
         self.dump_variables(path, exclude_info_maps)
         self.dump_variable_names_and_contexts(path, exclude_info_maps)
-        self.save_variables_to_csv_files(str(self._csv_variable_output_directory(path)))
+        self.save_variables_to_csv_files(str(Path(path) / "CSVs" / "om" / "variables"))
         self.dump_errors(path)
         self.dump_logs(path)
         self.dump_warnings(path)
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -1,13 +1,12 @@
 # !/usr/bin/env python3
 
-from enum import Enum
 from pathlib import Path
-import re
 from typing import Any, Dict, List, Optional, Union
+import datetime
 import json
 import os
-import datetime
-import csv
+import pandas as pd
+import re
 
 from RUFAS.util import Utility
 
@@ -36,14 +35,6 @@ class OutputManager(object):
 
     __instance = None
     pool_element_type = Dict[str, List[Dict[str, Any]]]
-
-    class OutputFormat(Enum):
-        """
-        Enumeration that represents valid output encodings for OutputManger pools
-        """
-        JSON = 0
-        CSV = 1
-
 
     def __new__(cls):
         if not hasattr(cls, "instance"):
@@ -317,7 +308,7 @@ class OutputManager(object):
         except Exception as e:
             raise e
 
-    def _save_variables_to_csv(self, data_dict: Dict[str, Any], path: str) -> None:
+    def _dict_to_file_csv(self, data_dict: Dict[str, Any], path: str) -> None:
         """Saves variables and their values to a csv file.
 
         Parameters
@@ -333,31 +324,21 @@ class OutputManager(object):
             If an error occurs while writting to the file
         """
         info_map = {"class": self.__class__.__name__,
-                    "function": self._save_variables_to_csv.__name__,
+                    "function": self._dict_to_file_csv.__name__,
                     }
-        self.add_log("_save_variables_to_csv", f"Attempting to save to {path}.", info_map)
-        if len(data_dict) == 0:
-            self.add_warning("_save_variables_to_csv", f"No variables in the pool to save to {path}", info_map)
-            with open(path, "w") as file:
-                file.truncate(0)
+        self.add_log("save_dict_file_try", f"Attempting to save to {path}.", info_map)
+
+        if not all("values" in v for v in data_dict.values()):
+            self.add_error("save_dict_file_try", f"Unable to save {path} due to missing values.", info_map)
             return
 
-        keys = list(data_dict.keys())
-        values = list(data_dict.values())
-        # Determine the maximum length of values to account for uneven lengths
-        max_length = max(len(d['values']) for d in values)
+        df = pd.DataFrame({k: pd.Series(v['values'], dtype=object) for k, v in data_dict.items()})
+        try:
+            df.to_csv(path, index=False)
+        except Exception as e:
+            raise e
 
-        rows = []
-        for i in range(max_length):
-            row = [d['values'][i] if i < len(d['values']) else None for d in values]
-            rows.append(row)
-
-        with open(path, 'w', newline='') as file:
-            writer = csv.writer(file)
-            writer.writerow(keys)
-            writer.writerows(rows)
-
-        self.add_log("_save_variables_to_csv", f"Successfully saved to {path}.", info_map)
+        self.add_log("save_dict_file_try", f"Successfully saved to {path}.", info_map)
 
     def _list_to_file_txt(self, data_list: List[str], path: str) -> None:
         """Saves a list into a text file
@@ -539,9 +520,9 @@ class OutputManager(object):
             file_path = os.path.join(save_path, self._generate_file_name(f"saved_variables_{filter_file}", "json"))
             self._dict_to_file_json(filtered_pool, file_path)
 
-    def dump_variables(self, path: str, format: OutputFormat, exclude_info_maps: bool = False) -> None:
+    def dump_variables(self, path: str, exclude_info_maps: bool = False) -> None:
         """
-        Dumps variables_pool into a json file in the given path to a directory.
+        Dumps variables_pool into a json file and a csv file in the given path to a directory.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


RUFAS/output_manager.py
```diff
@@ -307,6 +317,48 @@ def _dict_to_file_json(self, data_dict: Dict[str, Any], path: str) -> None:
...
+        """
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._save_variables_to_csv.__name__,
+                    }
+        self.add_log("_save_variables_to_csv", f"Attempting to save to {path}.", info_map)
+        if len(data_dict) == 0:
+            self.add_warning("_save_variables_to_csv", f"No variables in the pool to save to {path}", info_map)
+            with open(path, "w") as file:
+                file.truncate(0)
+            return
```
ltp23: I think we should add a check to make sure we have a `values` field for each key-value pair. It is ok to not create a file at all if we have no data.

RUFAS/output_manager.py
```diff
@@ -307,6 +317,48 @@ def _dict_to_file_json(self, data_dict: Dict[str, Any], path: str) -> None:
...
+
+        rows = []
+        for i in range(max_length):
+            row = [d['values'][i] if i < len(d['values']) else None for d in values]
+            rows.append(row)
+
+        with open(path, 'w', newline='') as file:
+            writer = csv.writer(file)
+            writer.writerow(keys)
+            writer.writerows(rows)
```
ltp23: Although the current implementation works correctly, I would suggest handling the data in a column-oriented fashion instead of a row-oriented one. Most data analysis libraries like `pandas` are optimized for columnar operations through vectorization. For example, here's an alternative way to implement the function:

```python
df = pd.DataFrame({k: pd.Series(v['values']) for k, v in data_dict.items()})
df.to_csv(path, index=False)
```

With this code, pandas automatically handles the variations in the lengths of the value lists by filling in NaN for any missing values, and it also manages the writing of the CSV file.

RUFAS/output_manager.py
```diff
@@ -487,7 +539,7 @@ def save_variables(self, save_path: str, dir_path: str,
             file_path = os.path.join(save_path, self._generate_file_name(f"saved_variables_{filter_file}", "json"))
             self._dict_to_file_json(filtered_pool, file_path)
 
-    def dump_variables(self, path: str, exclude_info_maps: bool = False) -> None:
+    def dump_variables(self, path: str, format: OutputFormat, exclude_info_maps: bool = False) -> None:
```
ltp23: Can you update the docstring this extra parameter?

RUFAS/output_manager.py
```diff
@@ -504,8 +556,12 @@ def dump_variables(self, path: str, exclude_info_maps: bool = False) -> None:
...
             pool = self._exclude_info_maps(self.variables_pool)
 
-        file_path = os.path.join(path, self._generate_file_name("all_variables", "json"))
-        self._dict_to_file_json(pool, file_path)
+        file_path = os.path.join(path, self._generate_file_name("all_variables", format.name.lower()))
+        match format:
+            case OutputManager.OutputFormat.JSON:
+                self._dict_to_file_json(pool, file_path)
+            case OutputManager.OutputFormat.CSV:
+                self._save_variables_to_csv(pool, file_path)
```
ltp23: This looks good. I would also add a default case (`case _:`) and handle it (e.g., add a warning, raise an error, do nothing). You could also set a default option and put it at the back of the list. It looks like you want it to be an optional parameter.

RUFAS/output_manager.py
```diff
@@ -35,6 +37,14 @@ class OutputManager(object):
     __instance = None
     pool_element_type = Dict[str, List[Dict[str, Any]]]
 
+    class OutputFormat(Enum):
+        """
+        Enumeration that represents valid output encodings for OutputManger pools
```
PooyaHekmati: 1. These are data formats not encoding.
2. Right now it is only for the vars pool (not pools)

RUFAS/output_manager.py
```diff
@@ -307,6 +317,48 @@ def _dict_to_file_json(self, data_dict: Dict[str, Any], path: str) -> None:
         except Exception as e:
             raise e
 
+    def _save_variables_to_csv(self, data_dict: Dict[str, Any], path: str) -> None:
```
PooyaHekmati: So it should be save dict to csv, not save vars to dict

RUFAS/output_manager.py
```diff
@@ -307,6 +317,48 @@ def _dict_to_file_json(self, data_dict: Dict[str, Any], path: str) -> None:
...
+            If an error occurs while writting to the file
+        """
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._save_variables_to_csv.__name__,
+                    }
+        self.add_log("_save_variables_to_csv", f"Attempting to save to {path}.", info_map)
+        if len(data_dict) == 0:
+            self.add_warning("_save_variables_to_csv", f"No variables in the pool to save to {path}", info_map)
+            with open(path, "w") as file:
+                file.truncate(0)
```
PooyaHekmati: why do you need to truncate the file here?

RUFAS/output_manager.py
```diff
@@ -307,6 +317,48 @@ def _dict_to_file_json(self, data_dict: Dict[str, Any], path: str) -> None:
...
+            return
+
+        keys = list(data_dict.keys())
+        values = list(data_dict.values())
+        # Determine the maximum length of values to account for uneven lengths
```
PooyaHekmati: comments are bad, don't use them.

RUFAS/output_manager.py
PooyaHekmati: isn't this function's responsibility to generate names? why is it not doing it and instead calling save functions?

RUFAS/output_manager.py
```diff
@@ -307,6 +317,48 @@ def _dict_to_file_json(self, data_dict: Dict[str, Any], path: str) -> None:
...
+        ----------
+        data_dict : Dict[str, Any]
+            The variable dictionary to be saved
+        path : str
+            The path to the file to be saved
+
+        Raises
+        ------
+        Exception
+            If an error occurs while writting to the file
```
PooyaHekmati, line 324: i don't see any exceptions being caught in this function.

tests/test_misc.py
```diff
@@ -320,6 +320,24 @@ def mock_output_manager(mocker) -> OutputManager:
...
+        file_path = os.path.join(temp_dir, "test.csv")
+        om._save_variables_to_csv(variable_data, file_path)
+        with open(file_path, 'r') as file:
+            assert file.read() == "var1,var2\n1.0,{'test': 1}\nTrue,\ntest,\n"
+        om._save_variables_to_csv(empty_data, file_path)
+        with open(file_path, 'r') as file:
+            assert file.read() == ""
+
+    assert os.path.exists(file_path) == False
+
```
ltp23: I see that you're using `tempfile` to handle throwaway file data, which I think is a nice approach, but in the context of unit testing, we might want to avoid actual I/O operations where possible.

I would suggest using pytest's parametrization feature to test various scenarios and edge cases more concisely (The two provided test cases are good here, but I think you can come up with more.) and also use more mocking to mock out external functions, which helps with focusing on the logic of the function being tested.

RUFAS/output_manager.py
PooyaHekmati: don't modify this function

ltp23 (approved):
Thanks Michael. The code's functionality works as expected. I have a few comments for improvement along the way. Thank you for teaching me a few things too.

RUFAS/output_manager.py
```diff
@@ -307,6 +317,48 @@ def _dict_to_file_json(self, data_dict: Dict[str, Any], path: str) -> None:
...
+            return
+
+        keys = list(data_dict.keys())
+        values = list(data_dict.values())
+        # Determine the maximum length of values to account for uneven lengths
+        max_length = max(len(d['values']) for d in values)
+
+        rows = []
+        for i in range(max_length):
+            row = [d['values'][i] if i < len(d['values']) else None for d in values]
```
PooyaHekmati: unreadable, add clarity

RUFAS/output_manager.py
PooyaHekmati: besides, how is it possible to have "uneven" lengths?

RUFAS/output_manager.py
```diff
@@ -35,6 +37,14 @@ class OutputManager(object):
     __instance = None
     pool_element_type = Dict[str, List[Dict[str, Any]]]
 
+    class OutputFormat(Enum):
```
PooyaHekmati: I don't think this is necessary.

PooyaHekmati (changes requested):
For each entry in the vars pool save a CSV file. To deal with nested structures, use str() for now.

RUFAS/output_manager.py
ltp23: I think uneven lengths in values lists happen when somewhere in the codebase the add_variables() only runs when certain local conditions are met.

RUFAS/output_manager.py
michaelRichards99: They weren't being caught because in the _dict_to_file_json they were being caught just to be thrown again without any additional computation done on the exception. Therefore, I did not wrap the IO operations in a try/catch block. I modified `_dict_to_file_csv` resemble the existing `_dict_to_file_json` method to add clarity as to where the exception may occur.

RUFAS/output_manager.py
michaelRichards99: The `dump_variables` method needs to create a unique file name for the json and csv files and then write the variable pool to those files. the `_generate_file_name` method's responsibility is to generate the names.

RUFAS/output_manager.py
michaelRichards99: removed with most recent commit.

RUFAS/output_manager.py
michaelRichards99: Removed with most recent commit. I had this originally for unit testing purposes, which is admittedly a poor reason to add this. It's been removed in the most recent commit and the unit tests now mock I/O operations.

RUFAS/output_manager.py
michaelRichards99: renamed to `_dict_to_file_csv`

RUFAS/output_manager.py
michaelRichards99: Removed.

tests/test_misc.py
michaelRichards99: That's a really good idea. I've removed all filesystem trips by mocking `open()` in this unit test.

I've used pytests `parametrization` feature to reuse the function and add more data concisely.

RUFAS/output_manager.py
michaelRichards99: Using pandas is much more concise and much more readable. I've removed the csv library and am using your suggestion to generate the csv file from the dictionary.

RUFAS/output_manager.py

RUFAS/output_manager.py
```diff
@@ -489,7 +522,7 @@ def save_variables(self, save_path: str, dir_path: str,
 
     def dump_variables(self, path: str, exclude_info_maps: bool = False) -> None:
         """
-        Dumps variables_pool into a json file in the given path to a directory.
+        Dumps variables_pool into a json file and a csv file in the given path to a directory.
```
PooyaHekmati: As I mentioned in my comment in the last review, you shouldn't be saving just a single CSV file. I think that is why you were concerned about varying lengths. Instead, for each entry in the vars pool, create a CSV file. Since we are assuming data types are primitive (or flattened using `str()`) you will end up with multiple CSV files each having a single column (it would be nice to add an id column if possible). Also, `dump_variables` should not be modified. Instead, create a function and call it, we will later expand it.

PooyaHekmati (changes requested):

RUFAS/output_manager.py
michaelRichards99: I forgot the comment containing all code change requests can itself contain code change requests. I'll create a csv file for each variable in the vars pool and will keep `dump_variables` unmodified, extracting csv generation to its own method.

PooyaHekmati: Thanks :)

michaelRichards99: I've added the functionality to create 1 csv per variable in the variables pool. 

I wasn't sure where to place the creation of the directory where these csv files go. Looks like currently both the `SimulationEngine` and `OutputHandler` [create directories](https://github.com/RuminantFarmSystems/MASM/blob/2af2ab10da382dbffb7434f38060740cd8fef121/RUFAS/simulation_engine.py#L181C13-L181C13)in the `output` directory, but the `OutputManager` doesn't create anything on its own. Do you think this logic could be consolidated into the `OutputManager`'s constructor?

ltp23:
Thank you, Pooya, for the clarifications. I misunderstood our objectives initially. I now understand that we want to save each variable in a separate CSV file, and this should happen right after the simulation ends, not later. I've updated the git issue description to reflect these points. Sorry Michael for the rework.

RUFAS/output_manager.py
```diff
@@ -307,6 +308,38 @@ def _dict_to_file_json(self, data_dict: Dict[str, Any], path: str) -> None:
         except Exception as e:
             raise e
 
+    def _dict_to_file_csv(self, data_dict: Dict[str, Any], path: str) -> None:
+        """Saves a variable and its values to a csv file.
```
PooyaHekmati: what does `variable and its values` mean? How can a variable have values? Why not explicitly stating `saves a dict to a csv file`?

RUFAS/output_manager.py
```diff
@@ -307,6 +308,38 @@ def _dict_to_file_json(self, data_dict: Dict[str, Any], path: str) -> None:
...
+
+        Parameters
+        ----------
+        data_dict : Dict[str, Any]
+            The variable dictionary to be saved
```
```suggestion
            The dictionary to be saved
```

RUFAS/output_manager.py
```diff
@@ -307,6 +308,38 @@ def _dict_to_file_json(self, data_dict: Dict[str, Any], path: str) -> None:
...
+        ------
+        Exception
+            If an error occurs while writting to the file
+        """
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._dict_to_file_csv.__name__,
+                    }
+        self.add_log("save_dict_file_try", f"Attempting to save to {path}.", info_map)
+
+        if len(data_dict) != 1 or all("values" not in v for v in data_dict.values()):
```
PooyaHekmati: why `len(data_dict) != 1`? What is the reason and significance of 1?

RUFAS/output_manager.py
```diff
@@ -487,6 +520,27 @@ def save_variables(self, save_path: str, dir_path: str,
...
+        Dumps variables_pool into one csv file per variable in the given path to a directory.
+
+        Parameters
+        ----------
+        path : str
+            Path to the output directory for the OutputManager.
+
+        """
+        pool = self.variables_pool
+        variables_csv_path = Path(path) / "CSVs" / "om" / "variables"
```
PooyaHekmati: avoid hardcoding values. Assume the whole path (except for the file name) is being passed to the function.

RUFAS/output_manager.py
```diff
@@ -487,6 +520,27 @@ def save_variables(self, save_path: str, dir_path: str,
...
+        try:
+            variables_csv_path.mkdir(parents=True, exist_ok=True)
+        except Exception as e:
+            raise e
+
+        for key, value in pool.items():
+            csv_file_path = os.path.join(variables_csv_path, self._generate_file_name(key, "csv"))
+            self._dict_to_file_csv({key: value}, csv_file_path)
```
PooyaHekmati: I think this is more efficient.
```suggestion
        for key in pool.keys():
            csv_file_path = os.path.join(variables_csv_path, self._generate_file_name(key, "csv"))
            self._dict_to_file_csv(pool[key], csv_file_path)
```

RUFAS/output_manager.py
```diff
@@ -487,6 +520,27 @@ def save_variables(self, save_path: str, dir_path: str,
             file_path = os.path.join(save_path, self._generate_file_name(f"saved_variables_{filter_file}", "json"))
             self._dict_to_file_json(filtered_pool, file_path)
 
+    def dump_variables_to_csv_files(self, path: str) -> None:
```
PooyaHekmati: I agree that this function is dumping right now, but in future PRs it will change to saving. I leave it to you to decide if you want to want to keep its current name and rename it later or if you want to call it save in this PR.

PooyaHekmati (changes requested):

RUFAS/output_manager.py
michaelRichards99: This goes back to [this comment](https://github.com/RuminantFarmSystems/MASM/pull/650#discussion_r1263150030) I left on your previous review. The OM does not create the output directories, the output handler does. I'll create a helper method for this for now.

RUFAS/output_manager.py
michaelRichards99: Since we were only saving the variables pool and are making sure each dictionary has the `values`, I used this name to signify that it was saving any python dicitonary but a dictionary from the variable pool.

I've renamed it to _dict_to_file_csv

RUFAS/output_manager.py
michaelRichards99: This was to ensure only 1 variable was being saved. Since this is renamed to `_dict_to_file_csv`, it no longer makes sense for this check to be here.

RUFAS/output_manager.py

RUFAS/output_manager.py
michaelRichards99: done. I've updated the name to use save.

tests/test_misc.py
```diff
@@ -313,13 +313,36 @@ def test_get_prefix() -> None:
...
+     False),
+    ({"var1": {"values": [1,2,3]}},
+     "var1\n1\n2\n3\n",
+     True),
+    ({"var1": {"not a value": [1]}}, "",
+     False),
+])
+def test_dict_to_csv(data: Dict[str, Any], expected_result: str, should_write: bool) -> None:
+    """Unit test for the function _dict_to_file_csv in the file output_manager.py"""
+    om = OutputManager()
```
ew3361zh: I think it would fix some of the singleton testing issues you were running into if you used the mock_output_manager fixture similar to other OM test functions rather than instantiating OM anew here.

tests/test_misc.py
```diff
@@ -658,6 +688,34 @@ def test_dump_variables(
         "_dict_to_file_json"
     ]
 
+def test_save_variables_to_csv_files(
```
ew3361zh, line 704: Add extra blank line before and after test funs (gets a flake8 violation without two spaces between test funs)

tests/test_misc.py
```diff
@@ -5,14 +5,14 @@
...
 """
 
-import re
 import os
-from typing import Callable
-from typing import Dict
+import re
+from typing import Any, Callable, Dict
 from mock import Mock, mock_open, patch
+from pathlib import Path
```
ew3361zh: showing up as "imported but unused."

tests/test_misc.py
```diff
@@ -5,14 +5,14 @@
...
 
 import pytest
-from mock.mock import MagicMock
+from mock.mock import MagicMock, call
```
ew3361zh: Also showing up as imported but unused. The only reference is showing up in the test_dict_to_csv but as flake8 violation of redefining call from this import.

ew3361zh (changes requested):
I think this looks good. Nice work! 

I had some questions about the function names that look like they were already addressed in previous reviews/conversations.

I do a have a few requested changes on flake8 violations:
- 2 blank lines between test functions on both sides
- whitespace after commas in lists [1,2,3] -> [1, 2, 3]
- remove unused import statements

Screenshot 2023-07-18 at 12 59 53 PM

michaelRichards99:
Thanks for bringing the Flake8 errors to my attention. I'll add this check to my commits to ensure that I am not adding more formatting errors to the code base from now on.

ew3361zh (approved):
Excellent, LGTM

PooyaHekmati (approved):

Issue #617, opened 2023-06-30
Create validate_element function in IM


ew3361zh:
Closely associated with issue #616, this issue is to create a new function in the IM class called `_validate_element()` which does the actual validation on the data.

It will be called from `validate_data()`.

It should return true if the data is valid and false otherwise.

ew3361zh:
Addressed in #652.

PR #652, opened 2023-07-13
Create IM validate_element function

ew3361zh:
This PR tries to address issue #617 to create a new function `validate_element()` in the IM that will validate each input data element based on the metadata guidelines.

Note:
Testing is not yet added. Want to make sure this is headed in the right direction.


ew3361zh:
moving to another branch

PR #654, opened 2023-07-13
Fix management event recording

Isgotkowitz:
This PR fixes some small issues in various value collection methods in the Crop and Soil module.

- When recording tillage, fertilizer, and manure applications, the info maps used `"field_name:'<field name>'"` instead of `"field:'<field name>'"`, which is how prefixes are constructed for info maps in `OutputGatherer`. All tests of these methods have been updated.
- In `OutputGatherer`, three variables were being collected twice and this PR removes the extra calls.


ew3361zh (commented):
I noticed this is still in draft form so I'll wait to approve until it's labeled ready just in case there's more changes coming. But as is right now, this looks good and I'd approve.

Isgotkowitz:
> I noticed this is still in draft form so I'll wait to approve until it's labeled ready just in case there's more changes coming. But as is right now, this looks good and I'd approve.

@ew3361zh Thank you for looking at this, my apologies - I did not realize I had left it in draft. It is ready to be approved, and there have been no changes since you last looked through it.

ew3361zh (approved):

matthew7838 (approved):
Looks good to me, good small touches. Sorry for the extra calls for you to fix, nice catch.

PR #655, opened 2023-07-14
Rework how crop variables and events are recorded

Isgotkowitz:
This PR modifies how Crop-related events are recorded, so that they are recorded consistently with other Crop and Soil variables, are recorded with essential information even if info maps are excluded, and are better grouped together.

- The prefix for info maps uses `field:'...',...` instead of `field_name:'...',...`, so that they can be filtered with daily and annual Crop and and Soil variables.
- Planting and harvesting dates are both included in the value recorded to the `OutputManager` instead of as an element of the info_map or within the prefix, when Crop-related events occur.
- Daily variables collected from crops (in `OutputGatherer`) include the planting date in the prefix, so that the daily values from different plantings of the same crop may be distinguished (more easily).
- Two small typos have been fixed in `OutputGatherer`.
- All tests that have been effected by these changes in recording variables have been updated.


matthew7838 (approved):
Looks good, seems like some good important minor changes.

ltp23 (approved):
The code looks good to me. Thank you, Ed!

PR #658, opened 2023-07-14
Add example object structure for example metadata

ew3361zh:
This PR is a short patch to the example metadata to include some nested data.

Adding nested data with the "type" of "object" and a nested structure within that object can serve as a guidline for other people adding to the metadata.

The existing `animal_management_animal.json` `herd_information` object was incorporated as the example. Data can be further nested than one layer and I can certainly add that if it's necessary/helpful.


PooyaHekmati (approved):

allisterakun (approved):
LGTM

PR #660, opened 2023-07-15
Im refactor

ew3361zh:
This PR tries to address issue #617 to create a new function validate_element() in the IM that will validate each input data element based on the metadata guidelines.

To achieve this, the PR:

1. Creates a `start_data_processing()` method to organize metadata and input data processing. This method will eventually be called from and return a boolean to `main()`. I thought it was helpful here to see the IM processing pipeline laid out and have laid out a proposal for how IM will be incorporated into `main.py` in issue #679.
2. Switches to using the metadata as the guideline to looping through what data to validate.  
3. To accomodate input data only being added to the IM pool after it has been validated, `validate_data()` is the second function to be called after `_load_metadata()` in the `start_data_processing()` pipeline. `validate_data()` loops through the file paths outlined by the metadata and opens and loads those files one at a time to a temporary data buffer dictionary which is used to validate and fix the data from each input file. Once the data has been validated and/or fixed, it will be added to the IM pool. 
4. Creates `validate_element()` method to recursively check each element expected to be in the input data file based on metadata guidelines which calls `fix_data()` to attempt to fix any invalid data found. 
5. Creates 3 separate functions (`_validate_string_type_element()`, `_validate_array_type_element()`, and `_validate_number_type_element()`) to validate by data type based on metadata guidelines.
6. Updates unit testing.

Notes:
-`get_data()` and `_fix_data()` are being added in PRs #675 and #646 respectively.
- `main()` was not updated with any expected behavior based on results of the new `start_data_processing()` function. That felt like a separate PR and something to do once we've got a validatable metadata in place.


RUFAS/input_manager.py
```diff
@@ -27,7 +28,29 @@ def __init__(self) -> None:
         self.__metadata: Dict[str, Any] = {}
         self.__pool: Dict[str, Any] = {}
 
-    def _load_metadata(self, metadata_path: str = "input/example_metadata.json") -> None:
+    def start_data_processing(self, metadata_path,
```
```suggestion
    def start_data_processing(self, metadata_path: str,
```

RUFAS/input_manager.py
```diff
@@ -27,7 +28,29 @@ def __init__(self) -> None:
...
+                              eager_termination: bool = True) -> bool:
+        """
+        Organize metadata and input data processing pipeline.
+
+        Parameters
+        ----------
+        metadata_path : str
+            File path to the metadata.
+        eager_termination : bool, optional, default=True
```
```suggestion
        eager_termination : bool, default=True
```

RUFAS/input_manager.py
```diff
@@ -27,7 +28,29 @@ def __init__(self) -> None:
...
+        -------
+        bool
+            Flag indicating whether input data is valid.
+        """
+        self._load_metadata(metadata_path)
+        self._load_data()
+        is_input_data_valid = self._validate_data(eager_termination)
+        return is_input_data_valid
+
+    def _load_metadata(self, metadata_path) -> None:
```
```suggestion
    def _load_metadata(self, metadata_path: str) -> None:
```

RUFAS/input_manager.py
```diff
@@ -142,31 +159,171 @@ def _validate_data(self, eager_termination: bool = True) -> bool:
...
+            is_in_range = variable_to_check["minimum"] <= input_data_value
+            if not is_in_range:
+                om.add_warning("Value out of range.", f"{var_name=}", info_map)
+                return is_in_range
+        if is_maximum_check:
+            is_in_range = input_data_value <= variable_to_check["maximum"]
+            if not is_in_range:
+                om.add_warning("Value out of range.", f"{var_name=}", info_map)
+                return is_in_range
+        return True
```
```suggestion
        minimum_value = variable_to_check.get("minimum")
        maximum_value = variable_to_check.get("maximum")
        is_in_range = True
        if minimum_value is not None and maximum_value is not None:
            is_in_range = minimum_value <= input_data_value <= maximum_value
            warning_string = f"Value not in range [{minimum_value}, {maximum_value}]."
        elif minimum_value is not None:
            is_in_range = minimum_value <= input_data_value
            warning_string = f"Value less than {minimum_value}."
        elif maximum_value is not None:
            is_in_range = input_data_value <= maximum_value
            warning_string = f"Value greater than {maximum_value}."
            
        if not is_in_range:
            om.add_warning(warning_string, f"{var_name=}", info_map)
            
        return is_in_range
```
I think adding the values checked against to the warning would make them a little bit more helpful. Also there is a lot of repeated repeated calls to the OutputManager and `variable_to_check`, by reducing their numbers I think the code becomes more readable.

RUFAS/input_manager.py
```diff
@@ -142,31 +159,171 @@ def _validate_data(self, eager_termination: bool = True) -> bool:
...
+
+    def _validate_num_type_element(self, variable_to_check: Dict[str, Any], var_name: str,
+                                   input_data_value: int) -> bool:
+        """Validates a __pool number element."""
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._validate_num_type_element.__name__,
+                    }
+        is_minimum_check = variable_to_check.get("minimum")
+        is_maximum_check = variable_to_check.get("maximum")
+        if is_minimum_check and is_maximum_check:
```
Isgotkowitz: This breaks if either "minimum" or "maximum" is 0, because 0 and None both evaluate to False.

RUFAS/input_manager.py
```diff
@@ -142,31 +159,171 @@ def _validate_data(self, eager_termination: bool = True) -> bool:
...
+        if is_minimum_length:
+            is_valid_string = variable_to_check["minimum_length"] <= len(input_data_value)
+            if not is_valid_string:
+                om.add_warning("String out length range.", f"{var_name=}", info_map)
+                return is_valid_string
+        if is_maximum_length:
+            is_valid_string = len(input_data_value) <= variable_to_check["maximum_length"]
+            if not is_valid_string:
+                om.add_warning("String out length range.", f"{var_name=}", info_map)
+                return is_valid_string
```
Isgotkowitz: Same comments here as `_validate_num_type_element()`

RUFAS/input_manager.py
```diff
@@ -142,31 +159,171 @@ def _validate_data(self, eager_termination: bool = True) -> bool:
...
+        if is_minimum_length:
+            is_in_range = variable_to_check["minimum_length"] <= len(input_data_value)
+            if not is_in_range:
+                om.add_warning("Array out of length range.", f"{var_name=}", info_map)
+                return is_in_range
+        if is_maximum_length:
+            is_in_range = len(input_data_value) <= variable_to_check["maximum_length"]
+            if not is_in_range:
+                om.add_warning("Array out of length range.", f"{var_name=}", info_map)
+                return is_in_range
```
Isgotkowitz: Same comments here as `_validate_num_type_element()`

RUFAS/input_manager.py
```diff
@@ -110,30 +133,24 @@ def _validate_data(self, eager_termination: bool = True) -> bool:
...
-        invalid_elements_counter = 0
-        fixed_elements_counter = 0
         invalid_critical_elements_counter = 0
         total_elements_checked_counter = 0
 
-        for key in self.__pool.keys():
-            for variable, value in self.__pool[key].items():
+        for module_key in self.__pool.keys():
+            property_map_key = self.__metadata["files"][module_key]["properties"]
+            module_properties = self.__metadata["properties"][property_map_key]
```
Isgotkowitz: Should either of these get wrapped to in `try except` blocks for `KeyErrors`?

Isgotkowitz (changes requested):
Good start on the data validation Niko. Requested changes on the three validation methods for arrays, numbers, and strings - there is an edge case that breaks them. Also some suggestions on small stuff mostly type annotations

RUFAS/input_manager.py
ew3361zh: It's a good question - I don't think so... 

The module_keys in the `__pool` will be set from `_load_data` method where it loops through the keys of metadata['files'] (config, animal, crop, etc) and sets those as the key for the `__pool` data loaded from the file. 

Only keys present in the `metadata['files'].keys()` will be set as `module_keys` for the `__pool` so it should be expected we'd find them again in the metadata as we loop through to get the nested properties keyword.

Let me know if you agree or if I'm missing something.

RUFAS/input_manager.py
ew3361zh: Good call.

RUFAS/input_manager.py
ew3361zh: I remember that crossing my mind at some point and forgot to come back to it. Nice catch!

RUFAS/input_manager.py
ew3361zh: This was interestingly generated by my autoDocstring vs code plug-in which I finally updated to numpy style. I checked out the numpy docstring documentation and it says: "If the default value would not be used as a value, optional is preferred." So I think getting rid of it is correct here.

RUFAS/input_manager.py
ew3361zh: updated

RUFAS/input_manager.py
```diff
@@ -27,7 +28,29 @@ def __init__(self) -> None:
         self.__metadata: Dict[str, Any] = {}
         self.__pool: Dict[str, Any] = {}
 
-    def _load_metadata(self, metadata_path: str = "input/example_metadata.json") -> None:
+    def start_data_processing(self, metadata_path: str,
+                              eager_termination: bool = True) -> bool:
+        """
+        Organize metadata and input data processing pipeline.
```
```suggestion
        Starts the pipeline for organizing metadata and input data processing.
```

RUFAS/input_manager.py
```diff
@@ -27,7 +28,29 @@ def __init__(self) -> None:
...
+
+        Parameters
+        ----------
+        metadata_path : str
+            File path to the metadata.
+        eager_termination : bool, default=True
+            If true, the process will be terminated upon finding invalid data.
```
```suggestion
        eager_termination : bool, default=True
            If True, the process will be terminated as soon as finding invalid data and failing to fix it.
            If False, the process will be terminated after going through and validating the entire data.
```

RUFAS/input_manager.py
```diff
@@ -27,7 +28,29 @@ def __init__(self) -> None:
...
+
+        Returns
+        -------
+        bool
+            Flag indicating whether input data is valid.
```
```suggestion
            True if data is valid, otherwise False.
```

PooyaHekmati (changes requested):
The implemented logic is wrong. As we discussed on July 13th, we add an element to the pool after checking its validity and trying to fix it if it was invalid. Please use the code that I wrote and shared with you that day on Slack (new 12.py) as the guideline for the logic.
This implementation has gone well beyond its expected delivery date. Please make sure it is implemented, reviewed, and merged by EOW.

RUFAS/input_manager.py
Isgotkowitz: Ok, that sounds good to me. And even if there were `try except` blocks added, it seems like there wouldn't be anything better to do than create a slightly more detailed error message and raise an error anyways

RUFAS/input_manager.py
```diff
@@ -53,120 +76,284 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+            else:
+                om.add_warning("InputManager load data file is not csv/json",
+                               f"{module_key} data must be available in either csv or json file type.",
+                               info_map)
+                continue
+
+            property_map_key = details["properties"]
+            module_properties = self.__metadata["properties"][property_map_key]
+            for element in module_properties.keys():
+                total_elements_counter += 1
```
allisterakun: I think this counter is inaccurate: it only keeps track of the top-level elements; it does not account for all the nested variables.

I am unsure if this is the intention, but instead of counting the top module elements, would it make more sense to keep track of the status (total/valid/invalid) of all the variables?

allisterakun: The same goes for the other counters: `valid_elements_counter` and `invalid_critical_elements_counter`:
    One invalid variable would make the entire module invalid, and there may be more than one invalid variables in the same module.

RUFAS/input_manager.py
```diff
@@ -53,120 +76,284 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+        element_hierarchy = element.split(".")
+        variable_to_check = reduce(lambda d, key: d[key], element_hierarchy,
+                                   self.__metadata["properties"][property_map_key])
+        var_type = variable_to_check["type"]
+        is_nested = var_type == "object"
+        if is_nested:
+            children_status: Dict[str, bool] = {}
+            false_counter = 0
+            for nested_key in variable_to_check.keys():
+                whole_key = f"{element}.{nested_key}"
```
allisterakun: Why do we need to pass the variable path as a string? Can we simply pass the `element_hierarchy` list?

RUFAS/input_manager.py
```diff
@@ -53,120 +76,284 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+                child_status = self._validate_element(self, whole_key, property_map_key, eager_termination)
```
allisterakun: The inputs here are incorrect. You are missing the `module_key` and `input_data`
Your input                             |    Expected
`self: InputManager`            => `module_key: str`
`whole_key: str`                    => `element: str` (which is the variable path)
`property_map_key: str`       => `property_map_key: str`
`eager_termination: bool`    => `input_data: Dict[str, Any]`
None                                    => `eager_termination: bool = True`

RUFAS/input_manager.py
```diff
@@ -53,120 +76,284 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
-
-        # TODO in validate_element fun branch
-        # where element is found to be invalid, place this warning:
-        # om.add_warning("Invalid data", f"Invalid data found: {key=}; {value=}", info_map)
-        pass
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._validate_element.__name__,
+                    }
+        element_hierarchy = element.split(".")
+        variable_to_check = reduce(lambda d, key: d[key], element_hierarchy,
```
allisterakun: Maybe rename it to `"variable_properties"`?

RUFAS/input_manager.py
```diff
@@ -53,120 +76,284 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+            is_in_range = len(input_data_value) <= variable_to_check["maximum_length"]
+            warning_string = f"Array length more than {maximum_length}."
+
+        if not is_in_range:
+            om.add_warning(warning_string, f"{var_name=}", info_map)
+
+        return is_in_range
+
+    def _validate_num_type_element(self, variable_to_check: Dict[str, Any], var_name: str,
+                                   input_data_value: int) -> bool:
```
allisterakun: Maybe change the `input_data_value` type to float to handle decimal data?

RUFAS/input_manager.py
```diff
@@ -53,120 +76,284 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+                                    "bool":
+                                    True}
+            is_valid = type_validation_dict.get(var_type)
+
+            if is_valid:
+                return True
+            elif is_valid is None:
+                raise Exception("Element must be type number, array, string, or bool")
+            else:
+                is_fixed = self._fix_data(module_key, element_hierarchy)
```
allisterakun: `(module_key, property_map_key, element_hierarchy, input_data)`

RUFAS/input_manager.py
```diff
@@ -53,120 +76,284 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+                return is_fixed
+
+    def _validate_array_type_element(self, variable_to_check: Dict[str, Any], var_name: str,
+                                     input_data_value: list) -> bool:
+        """Validates a __pool element of type array."""
```
allisterakun: Technically, the element/variable is not yet in self.__pool

RUFAS/input_manager.py
```diff
@@ -53,120 +76,284 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+        """Validates a __pool string element."""
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._validate_string_type_element.__name__,
+                    }
+        pattern_check = variable_to_check.get("pattern")
+        is_valid_string = True
+        if pattern_check is not None:
+            is_valid_string = bool(re.match(pattern_check, input_data_value))
+            warning_string = f"String variable must match pattern {variable_to_check['pattern']}."
+            if not is_valid_string:
```
allisterakun, line 339: I think you can do the warning altogether at the end and use `&=` for the length checks to update `is_valid_string`. So the code would be something like this:
```
                    "function": self._validate_string_type_element.__name__,
                    }
        pattern_check = variable_to_check.get("pattern")
        is_valid_string = True
        warning_string = ""
        if pattern_check is not None:
            is_valid_string = bool(re.match(pattern_check, input_data_value))
            warning_string += f"String variable must match pattern {variable_to_check['pattern']}. "

        minimum_length = variable_to_check.get("minimum_length")
        maximum_length = variable_to_check.get("maximum_length")
        if minimum_length is not None and maximum_length is not None:
            is_valid_string &= variable_to_check["minimum_length"] <= len(input_data_value) <= \
                variable_to_check["maximum_length"]
            warning_string += f"String out length range [{minimum_length}, {maximum_length}]."
        elif minimum_length is not None:
            is_valid_string &= variable_to_check["minimum_length"] <= len(input_data_value)
            warning_string += f"String length less than {minimum_length}."
        elif maximum_length is not None:
            is_valid_string &= len(input_data_value) <= variable_to_check["maximum_length"]
            warning_string += f"String length more than {maximum_length}."

        if not is_valid_string:
            om.add_warning(warning_string, f"{var_name=}", info_map)

        return is_valid_string
```

allisterakun (changes requested):

RUFAS/input_manager.py
ew3361zh: I think we need the new nested key at each level down in the hierarchy.

Say the metadata is {'animal_properties': {'herd_data': {'type': 'object', 'calf_num': {'type': 'number', 'minimum': 2}, 'cow_num': {'type': etc...

So the first time the element may just be 'herd_data' so the element_hierarchy will just be that element. This is an object type so it will trigger the recursive loop.
The first time through that loop the whole_key will now be {'herd_data'}.{'calf_num'} and it will be sent as the new element to the recursive call to 'validate_element'.
The second time through that loop the whole_key will now be {'herd_data'}.{'cow_num'} and so each time the element_hierarchy changes.

Maybe I'm missing your point here but I think it is set up correctly as is. I'm not sure we'd know which nested element we're on otherwise.

RUFAS/input_manager.py
ew3361zh: Good catch, thank you. Updating that and will push with other changes.

RUFAS/input_manager.py
ew3361zh: Good catch! Updating.

RUFAS/input_manager.py
ew3361zh: I like this. I think the initial idea was to break out of this validation as soon as the first violation is found but maybe because there are multiple possible combinations of violations, it would be helpful to get all the warnings at once so the user isn't rerunning the code after fixing the pattern violation just to get back a length violation.

However, I think the value returned by each individual check would need to inform whether we're adding the warning string which I think would create different booleans we'd need to check and then do a more cumbersome check at the end of whether any of the checks is False.

The other possibility Pooya and I talked about was writing the length min and max into the regex pattern but I think he didn't like that idea in the end so I'm going to keep things as is for now. There's a good idea in here though.

RUFAS/input_manager.py
ew3361zh: Definitely. Do you have any thoughts on how to get the actual nested numbers? I tried within the validate_element function but it became complicated to bring that back to validate_data and I don't know that we want to create global variables for the counters. I can rename these for now and maybe create an issue to try to find a way to accurately get the statuses of all the nested elements. I think it's still useful to get the top level object validity statuses and report those to OM.

RUFAS/input_manager.py
allisterakun: Right, I understand the current logic. My question is: if we need to parse the `variable_path: str` into a list of strings `element_hierarchy: List[str]` eventually, why don't we just pass the List[str] instead?

So it would take in `_validate_element(...args, element_hierarchy: List[str], ...args)`. In that case, the first time `["herd_data"]`, and it triggers the recursive loop: it calls itself with  `["herd_data", "calf_num"]` and `["herd_data", "cow_num"]`, respectively.

By doing that, we can skip unnecessary string parsing and string manipulation for each iteration of the recursive loop.

RUFAS/input_manager.py
allisterakun: Right, I have some ideas for the actual nested counter beside the global counter. Since we are using a buffer dictionary `input_data`, we can set up a `["metrics"/"counters"]` attribute (needs a better name) in the parent function `_validate_data`, and have all the children functions update this field correspondingly.

So in `_validate_data()`, you would do something like `input_data["metrics"] = {"total_element": 0, "valid_element": 0, "invalid_element": 0}`. 

Not sure if this would be something we can do.

ew3361zh: I have added this possible solution to the issue #677.

RUFAS/input_manager.py
ew3361zh: Implemented.

RUFAS/input_manager.py
```diff
@@ -231,14 +230,15 @@ def _validate_element(self, module_key: str, element: str,
             children_status: Dict[str, bool] = {}
             false_counter = 0
             for nested_key in variable_properties.keys():
-                whole_key = f"{element}.{nested_key}"
-                child_status = self._validate_element(self, module_key, whole_key, property_map_key, input_data,
-                                                      eager_termination)
+                element_hierarchy.append(nested_key)
```
allisterakun, line 240: After appending to the list `element_hierarchy`, you should pop the last element at the end of the for loop.

allisterakun (changes requested):

allisterakun (approved):

RUFAS/input_manager.py
```diff
@@ -53,120 +76,286 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
-        Validates input data and attempts to fix any invalid input data.
+        Validates data from input files and adds data to pool if valid.
 
         Parameters
         ----------
-        eager_termination : bool, default=True
-            If true, the process will be terminated upon finding invalid data.
+        eager_termination : bool
+            If True, the process will be terminated as soon as finding invalid data and failing to fix it.
+            If False, the process will be terminated after going through and validating the entire data.
```
michaelRichards99:
```suggestion
            If False, the process will be terminated after going through and validating the entire data, if invalid data is found.
```

RUFAS/input_manager.py
```diff
@@ -53,120 +76,286 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
-
-    def _validate_element(self, key: str, value: Any) -> bool:
+                    invalid_critical_items_counter += 1
+
+            om.add_log("Total Valid Items", f"{valid_items_counter=}", info_map)
+            om.add_log("Total Checked Items", f"{total_items_counter=}", info_map)
+            om.add_log("Total Invalid Critical Items", f"{invalid_critical_items_counter=}", info_map)
+            if invalid_critical_items_counter > 0:
+                return False
+            return True
```
michaelRichards99:
```suggestion
            return invalid_critical_items_counter == 0
```

RUFAS/input_manager.py
```diff
@@ -53,120 +76,286 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
         """
         Perform data validation checks.
 
         Parameters
         ----------
-        key : str
+        module_key : str
+            The module whose data is being validated.
+
+        element : str
```
michaelRichards99:
```suggestion
        element_hierarchy : List[str]
```

RUFAS/input_manager.py
```diff
@@ -53,120 +76,286 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
             The key of the data to validate.
 
-        value : Any
-            The value of the data to validate.
+        property_map_key : str
+            The metadata properties section keyword for the data input file being checked.
 
+        eager_termination : bool, default=True
```
michaelRichards99:
```suggestion
        eager_termination : bool
```

RUFAS/input_manager.py
```diff
@@ -53,120 +76,286 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+                                    "array":
+                                    self._validate_array_type_element(variable_properties, var_name, input_data_value),
+                                    "bool":
+                                    True}
+            is_valid = type_validation_dict.get(var_type)
+
+            if is_valid:
+                return True
+            elif is_valid is None:
+                raise Exception("Element must be type number, array, string, or bool")
```
michaelRichards99:
```suggestion
                raise Exception(f"Invalid type {var_type}: Element must be type number, array, string, or bool")
```
I think more information about the exception would be useful.

ew3361zh: Agreed.

RUFAS/input_manager.py
```diff
@@ -53,120 +76,285 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+
+            type_validation_dict = {"string":
+                                    self._validate_string_type_element(variable_properties, var_name, input_data_value),
+                                    "number":
+                                    self._validate_num_type_element(variable_properties, var_name, input_data_value),
+                                    "array":
+                                    self._validate_array_type_element(variable_properties, var_name, input_data_value),
+                                    "bool":
+                                    True}
+            is_valid = type_validation_dict.get(var_type)
```
michaelRichards99: Instantiating this dictionary will result in all 3 validation methods being called on the variable, which is only 1 of the 3 types. 
```suggestion
match var_type:
    case "string": 
        is_valid = self._validate_string_type_element(variable_properties, var_name, input_data_value)
      case "number":
          is_valid = self._validate_num_type_element(variable_properties, var_name, input_data_value)
      case "array":
           is_valid = self._validate_array_type_element(variable_properties, var_name, input_data_value)
    case "bool":
        is_valid = True
```

michaelRichards99 (changes requested):
Hi Niko. This looks great to me, good job!

I have 1 more requested change.

RUFAS/input_manager.py
michaelRichards99: A catch-all case `case _:` can also be added after all other cases and can raise an `error` for even more validation.

ew3361zh: Thanks, Michael, nice catch! I added your suggested change as well as a catch-all for the final case if the var_type doesn't match any of the given cases.

michaelRichards99 (approved):

RUFAS/input_manager.py
```diff
@@ -53,120 +76,285 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+                if not data_frame.empty:
+                    om.add_log("load_data_successful",
+                               f"Successfully loaded data from {file_path}.",
+                               info_map)
+                return data_dict
+        except Exception as e:
+            raise e
 
-    def _validate_data(self, eager_termination: bool = True) -> bool:
+    def _validate_data(self, eager_termination: bool) -> bool:
```
PooyaHekmati: As discussed on July 13th, this function is now doing more than validation and needs an appropriate name.

RUFAS/input_manager.py
```diff
@@ -53,120 +76,285 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
-        total_elements_checked_counter = 0
-
-        for key in self.__pool.keys():
-            for variable, value in self.__pool[key].items():
-                total_elements_checked_counter += 1
-                if self._validate_element(variable, value):
-                    valid_elements_counter += 1
+        valid_items_counter = 0
+        invalid_critical_items_counter = 0
+        total_items_counter = 0
```
PooyaHekmati: why is `fixed_elements_counter` removed? It is helpful to have it.

RUFAS/input_manager.py
```diff
@@ -53,120 +76,285 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+
+        for module_key, details in self.__metadata["files"].items():
```
PooyaHekmati: `module_key` is not the correct name. The file blobs don't necessarily map to modules. `file_blob_key` is probably a better name.
Also `details` is not specific enough. Details of what? use `file_details` or `file_description`.

RUFAS/input_manager.py
```diff
@@ -53,120 +76,285 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+            file_path = details["path"]
+            if details["type"] == "json":
+                data = self._load_data_from_json(file_path)
+            elif details["type"] == "csv":
+                data = self._load_data_from_csv(file_path)
+            else:
+                om.add_warning("InputManager load data file is not csv/json",
+                               f"{module_key} data must be available in either csv or json file type.",
+                               info_map)
+                continue
```
PooyaHekmati: As discussed on July 17th, avoid using `elif` structure for routing the flow to functions. 
```suggestion
            data_type_to_loadder_map = {
                "json": self._load_data_from_json(file_path),
                 "csv": self._load_data_from_csv(file_path),
            }
            try:
                data_type_to_loadder_map.get(details["type"])
            except:
               raise KeyException(f"Faulty data type in {module_key}, only CSV and JSON are supported."}
```

RUFAS/input_manager.py
```diff
@@ -53,120 +76,285 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+
+            property_map_key = details["properties"]
```
PooyaHekmati: `property_map_key` is misleading. We don't have a mapping here. Use `associated_properties` or `corresponding_properties` or `properties_blob_key`.

RUFAS/input_manager.py
```diff
@@ -53,120 +76,285 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+            module_properties = self.__metadata["properties"][property_map_key]
```
PooyaHekmati: Same as above, `module_properties` is not correct. Use `properties` or `blueprint`.

RUFAS/input_manager.py
```diff
@@ -53,120 +76,285 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+            for element in module_properties.keys():
```
PooyaHekmati: nit. when iterating through an iterable with plural name, individual items in it are singular:
```suggestion
            for property in module_properties.keys():
```

RUFAS/input_manager.py
```diff
@@ -53,120 +76,285 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+                total_items_counter += 1
+                is_valid_element = self._validate_element(module_key, [element], property_map_key, data,
+                                                          eager_termination)
+                if is_valid_element:
+                    valid_items_counter += 1
+                elif not is_valid_element and eager_termination:
+                    invalid_critical_items_counter += 1
+                    return False
```
```suggestion
                if is_valid_element:
                    valid_items_counter += 1
                    self.__pool[module_key] = data
                else:
                    if not eager_termination:                        
                        invalid_items_counter +=1
                     else:
                        return False
                        
```

RUFAS/input_manager.py
```diff
@@ -53,120 +76,285 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
                 else:
```
PooyaHekmati, line 185:
```suggestion
```

RUFAS/input_manager.py
```diff
@@ -53,120 +76,285 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
-
-        if invalid_critical_elements_counter > 0:
-            return False
-
-        return True
-
-    def _validate_element(self, key: str, value: Any) -> bool:
+                    invalid_critical_items_counter += 1
+            if invalid_critical_items_counter == 0:
+                self.__pool[module_key] = data
```
```suggestion
```

RUFAS/input_manager.py
```diff
@@ -53,120 +76,285 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+
+        om.add_log("Total Valid Items", f"{valid_items_counter=}", info_map)
+        om.add_log("Total Checked Items", f"{total_items_counter=}", info_map)
+        om.add_log("Total Invalid Critical Items", f"{invalid_critical_items_counter=}", info_map)
+        return invalid_critical_items_counter == 0
+
+    def _validate_element(self, module_key: str, element_hierarchy: List[str],
```
PooyaHekmati: why this function needs module_key? At this point, we have read the file and we are done with it. All we need is the property's entry.

RUFAS/input_manager.py
```diff
@@ -53,120 +76,285 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+                               info_map)
+                return False
+        else:
+            var_name = element_hierarchy[-1]
+            try:
+                input_data_value = reduce(lambda d, key: d[key], element_hierarchy, input_data)
+            except KeyError as e:
+                raise KeyError(f"Key {var_name} not found in pool: {e}")
+
+            match var_type:
```
PooyaHekmati: use dict instead. (discussed on July 17th)

RUFAS/input_manager.py
```diff
@@ -53,120 +76,285 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+    def _validate_array_type_element(self, variable_properties: Dict[str, Any], var_name: str,
+                                     input_data_value: list) -> bool:
+        """Validates an input data element of type array."""
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._validate_array_type_element.__name__,
+                    }
+        maximum_length = variable_properties.get("maximum_length")
+        minimum_length = variable_properties.get("minimum_length")
+        is_in_range = True
+        if maximum_length is not None and minimum_length is not None:
```
PooyaHekmati: don't mix different constraints in the same if statement (discussed on July 17th)

PooyaHekmati (changes requested):

RUFAS/input_manager.py
ew3361zh: How about `_validate_data_and_add_to_pool()`. Please let me know if you have any other suggestions.

RUFAS/input_manager.py
ew3361zh: The elements `_validate_data()` sends to `_validate_element()` are potentially nested dictionaries and with the fixing happening in `_validate_element()`, it became too complicated to try and track and report back the individual data fixing numbers to `_validate_data()`. I've documented this in issue #677.

RUFAS/input_manager.py
michaelRichards99: The expressions in python dictionaries are not lazily evaluated, so all the values in the dictionary are created when the dictionary is instantiated.

In this case, it will result in all 3 of the _validate... methods being called on the data. This could pollute the OM's logs if one of these causes an error.

We could write a [custom dictionary implementation](https://stackoverflow.com/questions/16669367/setup-dictionary-lazily) that decouples the method from the data, and executes the method when `__getitem__` is called or use a control flow construct.

RUFAS/input_manager.py
ew3361zh: Updated.

RUFAS/input_manager.py
ew3361zh: In conversations and testing with @michaelRichards99, instantiating a dictionary with multiple function calls as values will cause all functions to be called at instantiation (in this case with the `file_path` argument). This will lead to a number of unintended consequences so this isn't a good approach.
image (1)
image (2)

RUFAS/input_manager.py
ew3361zh: Updated to properties_blob_key.

RUFAS/input_manager.py
ew3361zh: Updated to properties.

RUFAS/input_manager.py
ew3361zh: Updated.

RUFAS/input_manager.py
ew3361zh: This would add the full input data loaded from the file to the pool each time through the loop which could be a number of times. I think the it's better to wait until we've completed the full inner `for property in properties.keys():` loop to then add the data (if we haven't gotten back any invalid elements).

RUFAS/input_manager.py
PooyaHekmati: simply move the arg from lines 162 and 163 to line 166.
Running 
```
print("Start")
class foo():
    def __init__(self):
        print("init, creating func mapping dict")
        d={
            "bar1": self.bar1,
            "bar2": self.bar2,
            "bar3": self.bar3,
            "bar4": self.bar4,
        }
        print("dict is created")
        x=1
        for k,v in d.items():
            if k=="bar2":
                continue
            print(f"calling {k}")
            v(x)
            x+=1
    def bar1(self, x):
        print(f"bar1 {x=}")
    def bar2(self, x):
        print(f"bar2 {x=}")
    def bar3(self, x):
        print(f"bar3 {x=}")
    def bar4(self, x):
        print(f"bar4 {x=}")
print("creating foo object")
f=foo()
```
Generates
```
Start
creating foo object
init, creating func mapping dict
dict is created
calling bar1
bar1 x=1
calling bar3
bar3 x=2
calling bar4
bar4 x=3
```
As you see, bar2 is not called.
CC @michaelRichards99

michaelRichards99: That works cleanly in this instance because the function's arity is the same.

It does become more difficult for variadic functions as is the case for the 2nd instance of this in this PR. In that case, the types `string`, `array`, and `number` all have functions with 3 parameters and the `bool` type needs a function that always returns true. @ew3361zh An additional `self._validate_bool_type_element(variable_properties, var_name, input_data_value)` that always returns `True` would remove this complexity and then the dictionary that stores function pointers works as planned.

PooyaHekmati: Can you write a piece of code that shows your point?

michaelRichards99: We had the following dictionary. (I've taken out the parameters so it doesn't call the functions and replaced True with bool().
```
func_map = {
    "string": self._validate_string_type_element,
    "number": self._validate_num_type_element,
    "array": self._validate_array_type_element,
    "bool": bool
}
```
bool() takes 1 arg while the others have 3 (variable_properties, var_name, input_data_value). when you get the function with `func_map(var_type)`, you'd need to check if the type was a `bool` so you can pass the proper arguments to avoid a TypeError.

Instead of special-casing the `bool` type, we could introduce `self._validate_bool_type_element(variable_properties, var_name, input_data_value)` which removes the complexity.

PooyaHekmati: Yes, bool can have a function. It will do nothing right now but allows adding validation for bool down the road.

RUFAS/input_manager.py
ew3361zh: Good catch. I think we had it here because we mistakenly thought fix_data() needed it but @allisterakun have resolved this and will make the adjustments here and in the fix_data() branch.

RUFAS/input_manager.py
ew3361zh: Updated.

RUFAS/input_manager.py
ew3361zh: Addressed.

RUFAS/input_manager.py
ew3361zh: Addressed as well. Dicionary method implemented.

RUFAS/input_manager.py
ew3361zh: Implemented.

RUFAS/input_manager.py
ew3361zh: I have implemented this as requested.

RUFAS/input_manager.py
ew3361zh: Removed from code as requested.

RUFAS/input_manager.py
ew3361zh: I have addressed this now and it will report the total items checked, total invalid items, total valid items, and total fixed items.

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+                               info_map)
+                return data_dict
+        except Exception as e:
+            raise e
 
-    def _validate_data(self, eager_termination: bool = True) -> bool:
+    def _validate_data_and_add_to_pool(self, eager_termination: bool) -> bool:
         """
-        Validates input data and attempts to fix any invalid input data.
+        Validates data from input files and adds data to pool if valid.
```
PooyaHekmati: I don't think this statement is accurate. It first loads the files, then runs the validations, for invalid ones it attempts to fix them, then adds them to the pool.
Also `_populate_pool` is probably a better name for this function.

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
-        for key in self.__pool.keys():
-            for variable, value in self.__pool[key].items():
-                total_elements_checked_counter += 1
-                if self._validate_element(variable, value):
-                    valid_elements_counter += 1
+
+        for file_blob_key, file_details in self.__metadata["files"].items():
+            file_path = file_details["path"]
+            data_type_to_loader_map = {"json": self._load_data_from_json,
+                                       "csv": self._load_data_from_csv}
```
PooyaHekmati: Doesn't need to be in the for-loop.
```suggestion
        data_type_to_loader_map = {"json": self._load_data_from_json,
                                   "csv": self._load_data_from_csv}
        for file_blob_key, file_details in self.__metadata["files"].items():
            file_path = file_details["path"]
```

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+            try:
+                data_loader = data_type_to_loader_map[file_details["type"]]
+                data = data_loader(file_path)
+            except KeyError:
+                raise KeyError(f"Faulty data type in {file_blob_key}, only CSV and JSON are supported.")
```
PooyaHekmati: I know I wrote this error message in an earlier comment, but now with the additional context I think this is better. So later on when we want to add new types we just need to implement its function and add its handle to the dict.
```suggestion
                raise KeyError(f"Faulty data type in {file_blob_key}, supported types are: {data_type_to_loader_map.keys()}.")
```

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+
+            counter_dict = {"fixed_elements": 0, "total_elements": 0, "valid_elements": 0, "invalid_elements": 0}
+            properties_blob_key = file_details["properties"]
+            properties = self.__metadata["properties"][properties_blob_key]
+            for property in properties.keys():
+                is_valid_element = self._validate_element([property], properties_blob_key, data,
```
PooyaHekmati: not a biggie, but `is_element_valid` sounds better.

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+                                                          eager_termination, counter_dict)
```
PooyaHekmati: If I'm understanding correctly, `counter_dict` is being passed so it can be used to retrieve the values. This is a bad practice. It is so bad that a whole language (Rust) is developed around the ownership and borrowing concepts to prevent something like this. Instead, instantiate a dictionary or named tuple (both are fine for this use case, choose whichever you think fits better), add all the counters and the bool flag that is currently used in line 174 to it. So instead of returning a bool, `self._validate_element()` now returns a package (a complex data structure either dict or named tuple).
There are two reasons why this is bad practice: 
1. Memory management, two actors are sharing the same resource when they don't have to. 
2. Readability and clean code: When someone sees the signature of a function, the args are what the function needs to be able to do its job. In here, element validation is not dependent on having the counters.

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+        om.add_log("Total Fixed Items", f"{fixed_elements_counter=}", info_map)
+        om.add_log("Total Invalid Items", f"{invalid_elements_counter=}", info_map)
+        return invalid_elements_counter == 0
 
-    def _validate_element(self, key: str, value: Any) -> bool:
+    def _validate_element(self, element_hierarchy: List[str], properties_blob_key: str,
+                          input_data: Dict[str, Any], eager_termination: bool,
+                          counter_dict: Dict[str, int], ) -> bool:
         """
         Perform data validation checks.
```
PooyaHekmati: can we add more details here?

RUFAS/input_manager.py
PooyaHekmati: I'm happy to see how you guys are working together :)

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+        variable_properties = reduce(lambda d, key: d[key], element_hierarchy,
+                                     self.__metadata["properties"][properties_blob_key])
+        var_type = variable_properties["type"]
+        is_nested = var_type == "object"
+        if is_nested:
+            children_status: Dict[str, bool] = {}
+            false_counter = 0
+            for nested_key in variable_properties.keys():
+                element_hierarchy.append(nested_key)
+                child_status = self._validate_element(element_hierarchy, properties_blob_key,
```
PooyaHekmati: [Recursion](https://i0.wp.com/miro.medium.com/max/640/1*__jhoWXEa_INM44V5jx30w.jpeg?resize=831%2C592&ssl=1)

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+                                                      input_data, eager_termination)
```
PooyaHekmati: Don't we need to pass `counter_dict`?

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+                if eager_termination and not child_status:
```
PooyaHekmati: rename it in line 234 as well
```suggestion
                if eager_termination and not is_child_valid:
```

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+                children_status[element_path] = child_status
+                if not child_status:
+                    om.add_warning("Invalid nested element found", f"{element_path}", info_map)
+                    false_counter += 1
+                element_hierarchy.remove(nested_key)
+            is_valid = false_counter == 0
+            if is_valid:
+                return True
+            else:
+                invalid_child_vars = [key for key, value in children_status.items() if value is False]
```
PooyaHekmati: If we keep track of these right after line 239 then we don't need to scan the children_status again to find them. i.e. we can trade time complexity for space complexity, which I think is worth it.

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+            type_validation_dict = {"string": self._validate_string_type_element,
+                                    "number": self._validate_num_type_element,
+                                    "array": self._validate_array_type_element,
+                                    "bool": self._validate_bool_type_element, }
+
+            try:
+                validity_checker = type_validation_dict[var_type]
+                is_valid = validity_checker(variable_properties, var_name, input_data_value)
+            except KeyError:
+                raise KeyError(f"Invalid type {var_type}: Element must be type number, array, string, or bool")
```
PooyaHekmati: 1. Adding dict at the end of a name doesn't provide more info, we can see that we have dict here. That's why we generally choose names based on functionality and behavior. In this case, we have a map because we are mapping one domain to another.
2. `validator` is shorter than `validity_checker`. It is also used in other codebases as well. [See this](https://en.wikipedia.org/wiki/Validator)
3. Let's be programmatic in the error message to reduce the needed maintenance down the road.
4. The function call doesn't need to be in the try block.

```suggestion
            data_type_to_validator_map = {"string": self._validate_string_type_element,
                                    "number": self._validate_num_type_element,
                                    "array": self._validate_array_type_element,
                                    "bool": self._validate_bool_type_element, }
            try:
                validator = data_type_to_validator_map [var_type]
            except KeyError:
                raise KeyError(f"Invalid type {var_type}: Element must be type {data_type_to_validator_map.keys()}")
            is_valid = validator (variable_properties, var_name, input_data_value)
```

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+                warning_string = f"Array length less than {minimum_length}."
+                om.add_warning(warning_string, f"{var_name=}", info_map)
+                return is_in_range
+        if maximum_length is not None:
+            is_in_range = len(input_data_value) <= variable_properties["maximum_length"]
+            if not is_in_range:
+                warning_string = f"Array length more than {maximum_length}."
+                om.add_warning(warning_string, f"{var_name=}", info_map)
+
+        return is_in_range
```
PooyaHekmati: The same applies to other validation functions as well. There is no reason to obfuscate the behavior by `is_in_range`. Explanatory variables should not be used in the logic.
```suggestion
        if minimum_length is not None:
            is_in_range = variable_properties["minimum_length"] <= len(input_data_value)
            if not is_in_range:
                warning_string = f"Array length less than {minimum_length}."
                om.add_warning(warning_string, f"{var_name=}", info_map)
                return False
        if maximum_length is not None:
            is_in_range = len(input_data_value) <= variable_properties["maximum_length"]
            if not is_in_range:
                warning_string = f"Array length more than {maximum_length}."
                om.add_warning(warning_string, f"{var_name=}", info_map)
                return False

        return True
```

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+            else:
+                is_fixed = self._fix_data(variable_properties, element_hierarchy, input_data)
+                if is_fixed:
+                    counter_dict["fixed_elements"] += 1
+                else:
+                    counter_dict["invalid_elements"] += 1
+                return is_fixed
 
-    def _fix_data(self, key: str, value: Any) -> bool:
+    def _validate_array_type_element(self, variable_properties: Dict[str, Any], var_name: str,
```
PooyaHekmati: what about?
```suggestion
    def _array_type_validator(self, variable_properties: Dict[str, Any], var_name: str,
```
and for other similar functions.

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+        if maximum_length is not None:
+            is_in_range = len(input_data_value) <= variable_properties["maximum_length"]
+            if not is_in_range:
+                warning_string = f"Array length more than {maximum_length}."
+                om.add_warning(warning_string, f"{var_name=}", info_map)
+
+        return is_in_range
+
+    def _validate_num_type_element(self, variable_properties: Dict[str, Any], var_name: str,
+                                   input_data_value: float) -> bool:
```
```suggestion
                                   input_data_value: union[int, float]) -> bool:
```

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+            return False
+        variable_parent = reduce(lambda d, key: d[key], element_hierarchy[:-1],
+                                 input_data)
+        variable_parent[element_hierarchy[-1]] = variable_properties['default']
+        om.add_warning("Data fixed",
+                       f"Invalid data fixed: {element_hierarchy[-1]} => {variable_properties['default']}",
+                       info_map)
+        return True
+
+    def get_data(self, data_address: str) -> Optional[Any]:
```
PooyaHekmati: I think type `Any` includes type `None` so `Optional` is unnecessary.

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+            The requested data if found.
+
+        Raises
+        -------
+        KeyError
+            If the requested data is not found.
+
+        Examples
+        -------
+        >>> InputManager.get_data('animal.herd.calf_num')
```
PooyaHekmati: I don't think this example is correct. `get_data` is not a static function, so it can't be called without calling the constructor first.
```suggestion
        >>> IM=InputManager()
        >>> IM.get_data('animal.herd.calf_num')
```
or
```suggestion
        >>> InputManager().get_data('animal.herd.calf_num')
```

RUFAS/input_manager.py
```diff
@@ -53,144 +77,368 @@ def _load_metadata(self, metadata_path: str = "input/example_metadata.json") ->
...
+        try:
+            data_value = reduce(lambda d, key: d[key], element_hierarchy,
+                                self.__pool)
+            return data_value
+
+        except KeyError as key_error:
+            invalid_key = str(key_error).strip("\'")
+            parent_address = str(data_address.split("." + invalid_key)[0])
+
+            om.add_warning("Data not found:", f"Cannot find \"{data_address}\", "
```
PooyaHekmati: we are literally talking about an error here, so an error should be added to the pool, not a warning.

PooyaHekmati (approved):

RUFAS/input_manager.py
ew3361zh: Updated docstrings and I like _populate_pool, thank you!

RUFAS/input_manager.py
ew3361zh: moved outside for-loop.

RUFAS/input_manager.py
ew3361zh: I like it.

RUFAS/input_manager.py
ew3361zh: updated

RUFAS/input_manager.py
ew3361zh: That makes sense. I realize we're essentially adding these to OM twice with what's already happening after 239 and here.

RUFAS/input_manager.py
ew3361zh: updated, thank you.

RUFAS/input_manager.py
ew3361zh: Updated

RUFAS/input_manager.py
ew3361zh: I like this!

RUFAS/input_manager.py
ew3361zh: Updated

RUFAS/input_manager.py
ew3361zh: Updated.

RUFAS/input_manager.py
ew3361zh: I think the first way will be more helpful.

RUFAS/input_manager.py
ew3361zh: Updated

RUFAS/input_manager.py
ew3361zh: Right! Will make sure the arguments sent and returned will be correct with the updates to where counter_dict is instantiated per your comments above.

RUFAS/input_manager.py
ew3361zh: Thank you for the explanation and background! I've updated to instantiate the dictionary in `_validate_element()` (with the 'is_valid' bool added) and return it to `_populate_pool()`

PooyaHekmati:
Bypassing branch protection as a change was requested by @Isgotkowitz , however, he is not available to re-review. His comments are addressed and the PR has got 3 approvals out of the needed 2.

PR #665, opened 2023-07-17
Soil and crop metadata

### LINKED ISSUE #667: Write metadata properties for Field, Soil, Crop management, and Farm management

Isgotkowitz:
This PR fills in the metadata for the crop, soil, field management, and field configurations. 

There is also an example of what a field configuration file would look like following the properties specified in the metadata, I thought it would be helpful to have it here so that it makes clear one of the notes I added at the bottom of issue #668.

Also, there are comments on the files pointing out specific ways the metadata might be improved.


input/example_metadata.json
```diff
@@ -257,33 +288,245 @@
...
+        "default": 0.95
+      },
+      "soil_layers": {
+        "type": "array",
+        "description": "The soil layers that this profile contains. Each element of the array should be a `soil_layer` object.",
+        "minimum_length": 1
+      },
+      "soil_layer_properties": {
+        "type": "object",
+        "description": "This represents a single soil layer.",
```
Isgotkowitz, line 641: @ew3361zh @PooyaHekmati I am tagging you both to alert you to this situation. One of the inputs for soil is an array of objects, each object representing a soil layer. As you can see from the soil_layer_properties description, it is an involved and complex object, so some level of input validation would be ideal.

input/example_metadata.json
```diff
@@ -257,33 +288,245 @@
...
+          "minimum": 0.0,
+          "maximum": 100.0,
+          "default": 1.3
+        },
+        "initial_labile_inorganic_phosphorus_concentration": {
+          "type": "number",
+          "description": "Concentration of labile inorganic phosphorus at the beginning of the simulation.",
+          "units": "milligrams / kilogram soil",
+          "minimum": 0.0,
+          "default": null
```
Isgotkowitz: @ew3361zh @PooyaHekmati I am also curious if this would be an acceptable specification for within a property?

input/example_metadata.json
PooyaHekmati: Hi @Isgotkowitz , let's meet and talk about this.

input/example_ARL_field.json
```diff
@@ -0,0 +1,12 @@
+{
+  "soil_specification": "soil/ARL_soil.json",
+  "crop_specification": "crop/ARL_rotation.json",
+  "field_management_specification": "field_management/ARL_field_management_1.json",
```
Isgotkowitz: An alternative here would be to specify the name of the blob that contains the file with specification data, instead of specifying the file path.

input/example_metadata.json
```diff
@@ -168,18 +189,211 @@
...
+          "description": "Number of times that this tillage schedule should be repeated.",
+          "minimum": 0,
+          "default": 0
+        },
+        "pattern_skip": {
+          "type": "number",
+          "description": "Number of years to be skipped between schedule repetitions.",
+          "minimum": 0,
+          "default": 0
+       }
```
Isgotkowitz: Currently, the field management properties contains 3 objects, one each for fertilizer, manure, and tillage applications. This could easily be flattened would just result in 3 blobs instead of one, which would not be a big deal to handle. Just let me know if you would like me to make this change.

Edit: I flattened these into 3 different properties.

input/example_metadata.json
```diff
@@ -49,12 +49,33 @@
...
+      "properties": "soil_profile_properties"
+    },
+    "soil_2": {
+      "title": "Second soil data",
+      "description": "Soil is still being treated like dirt.",
+      "path": "soil/example_soil_2.json",
+      "type": "json",
+      "properties": "soil_profile_properties"
+    },
+    "field_management": {
```
allisterakun: Just a question I had, you mentioned in #668 that we should "allow users to specify multiple fields to be simulated for a single simulation." I am wondering if each field needs its own `field_management` or if this applies to all fields.

input/example_metadata.json
```diff
@@ -168,18 +189,211 @@
...
+        "minimum": 0,
+        "default": 0
+      }
+    },
+    "field_management_properties": {
+      "fertilizer_schedule_properties": {
+        "type": "object",
+        "description": "Defines the schedule of fertilizer applications for a field.",
+        "available_fertilizer_mixes": {
+          "type": "object",
```
allisterakun: Should this be an array if you're listing all the available fertilizer mixes?

allisterakun (approved):

input/example_metadata.json
Isgotkowitz: Good question. Both of the scenarios you have described are possible - two different fields may use the same field management, and they may use different field management specifications (same with crop and soil specifications). So when we get to using metadata files to actually run the simulation, there will sometimes be multiple blobs with the same properties that specify different files.

input/example_metadata.json
Isgotkowitz: No, because each available fertilizer mix is specified as a key value pair, where the key is the name of the mix as a string and the value is another dictionary containing the keys `N`, `P`, and `K`, so the more appropriate structure for this is a dictionary.

input/example_metadata.json
ew3361zh: I forgot to ask if this was addressed already - I do think because we're using json.load it should be able to work. Is it important to have null/None instead of 0.0 as a default?

input/example_ARL_field.json
```diff
@@ -0,0 +1,14 @@
...
+  "manure_management_specification": "manure_schedules/ARL_manure_schedule_1.json",
+  "tillage_management_specification": "tillage_schedules/ARL_tillage_schedule_1.json",
+  "field_size": 1.0,
+  "absolute_latitude": 43.304346,
+  "longitude": null,
+  "minimum_daylength": 7.95,
+  "seasonal_high_water_table": false,
+  "watering_amount_in_liters": 0.0,
+  "watering_interval": 0.0
+}
```
ew3361zh: Can you explain again what this is and how it relates to the metadata?

input/example_metadata.json
```diff
@@ -49,12 +49,47 @@
...
+    "soil_2": {
+      "title": "Second soil data",
+      "description": "Soil is still being treated like dirt.",
+      "path": "soil/example_soil_2.json",
+      "type": "json",
+      "properties": "soil_profile_properties"
+    },
+    "fertilizer_schedule": {
+      "title": "Fertilizer schedule.",
+      "description": "Fertilizer is like butter, you can never have too much.",
```
ew3361zh, line 68: these are all very funny.

input/example_metadata.json
```diff
@@ -168,20 +203,205 @@
...
+        "description": "Number of years to be skipped between schedule repetitions.",
+        "minimum": 0,
+        "default": 0
       }
     },
+    "fertilizer_schedule_properties": {
+      "available_fertilizer_mixes": {
+        "type": "object",
+        "description": "All the available fertilizer mixes that are available besides the default ones.",
+        "default": {}
```
ew3361zh: Right now we're not doing defaults for object types because we're only validating the data within the object. So this can probably be removed.

input/example_metadata.json
```diff
@@ -257,33 +477,323 @@
...
+        "description": "Amount of residue present on the soil surface at the start of the simulation.",
+        "units": "kg / ha",
+        "minimum": 0.0,
+        "default": 0.0
+      },
+      "soil_cover_type": {
+        "type": "string",
+        "description": "Type of soil cover at the beginning of the simulation.",
+        "reference": "SurPhos Theoretical documentation eqn. [14].",
+        "pattern": "^(BARE|RESIDUE_COVER|GRASSED)$",
```
ew3361zh: I wondered what you thought about the capitalization here - right now this would turn back a false match on "residue_cover". We could try to make the pattern checker a little more flexible, or I guess in the DCA we could ensure that it's a drop-down menu/radio button that has this exact capitalization. Probably the latter is the preferred since we don't want to make the pattern checker too lenient.

input/example_metadata.json
```diff
@@ -257,33 +477,323 @@
...
+          "minimum": 0.0,
+          "default": 0.05
+        }
+      }
+    },
+    "field_properties": {
+      "soil_specification": {
+        "type": "string",
+        "description": "Path to the soil profile specification for this field.",
+        "pattern": "*.json$"
```
ew3361zh: Sorry if I misunderstood in our meeting with Allister but can you explain this section again and why these aren't treated as other metadata files (i.e. what differentiates them from other files in the files section and what goal is putting them here trying to achieve)? Is the thought these paths would be validated by the Input Manager and then followed after they're validated and loaded into the input data?

ew3361zh (changes requested):
I think this looks really good overall. I have a few concerns and questions I've added and want to make sure we're all on the same page as far as expectations for how this data will be processed by IM. In particular the paths to other files within the "field_properties" section.

input/example_metadata.json
Isgotkowitz: I do not think it has been addressed. Is it important - yes, because there are some `post_init` functions which use internal methods to initialize variables, but only in cases when the user does not provide any value (as indicated by `None`).

input/example_ARL_field.json
Isgotkowitz: So this `example_ARL_field.json` file is not related to the metadata - I wrote it up because the input system for the Crop and Soil module is changing very significantly with the incorporation of the metadata and `InputManager` and I wanted to give a tangible example of what the new field input files will look like; the field configuration files are totally novel so I thought it important to have their design looked at too. This example does not necessarily need to be merged into `master` though.

input/example_metadata.json
Isgotkowitz: I matched it to what is used internally in the Crop and Soil module (i.e., it will only accept those three values). However it would be easy to make these all lower case - if that is the best practice then I can do that.

input/example_metadata.json
Isgotkowitz: > why these aren't treated as other metadata files (i.e. what differentiates them from other files in the files section

The soil specification files are treated the same as other metadata files - but only when they are specified in the `files` section of the metadata. 

> what goal is putting them here trying to achieve)?

This `"soil_specification"` property is verifying and validating that a field configuration file has a reference to a soil profile. 

>  Is the thought these paths would be validated by the Input Manager and then followed after they're validated and loaded into the input data?

No. Only paths specified in the `path` of a blob will be loaded and validated. The pattern here is verifying that the path contained by a field configuration file is formatted correctly, but it does not load or open any files based on that. 

So the order of execution will be:
- The `InputManager` loads and validates every file specified in blobs in the `files` section of the metadata.
- The Crop and Soil module pulls the data for one field (at a time) from `InputManager`
- After getting the path to the configuration files associated with the field it is currently setting up, Crop and Soil module will pull the data from the `InputManager` that is associated with the blob that contains the same `path` as the one specified in data for the field that is being set up.

Note that if a field configuration file contains the path to a file that is not specified in any of the other blobs, the simulation should throw an error and terminate.

Another open question is how the Crop and Soil module will pull data from the `InputManager` based on the file that the data originated from - this is noted at the bottom of issue #668.

This was a pretty long-winded explanation and may not be very clear, so please let me know if you want/need to talk about it more.

input/example_metadata.json
Isgotkowitz: ACK'd

input/example_metadata.json
Isgotkowitz: Thanks:) I am trying to work up to Pooya's level of engineering humor.

input/example_metadata.json
ew3361zh: I think it's okay to keep as is. We can certainly account for it in the DCA. If someone wrote their own soil input file though and didn't capitalize it would return a warning and would sub in the default. I'm not sure how great the difference would be and what effects the user would see between these results and they could check the OM warnings log and see that the pattern for this element wasn't properly matched.

input/example_metadata.json
ew3361zh: I think this makes sense now and if I'm understanding it correctly, would work. Thank you for the longer explanation, it's helpful.

input/example_metadata.json
ew3361zh: I think the issue I see here is our num type validator in IM validates the value as if it's a num. So if someone didn't enter a value and we sent this through the num_type_validator, it would crash. Same for if someone entered "six" for this field labeled as "type": number. I'm not sure the best way to address these types of differences - right now we're relying on the type being representative of the value being sent to the type validator. We could do a check on type or just a check for None but curious what you think.

ew3361zh (approved):
I think we may need to come back to this structure once we've settled on the best way to validate array element data (object and primitive types) and will need to figure out in IM how to address non-matching type data when the data type is labeled in a specific way. But that feels like an IM decision and not necessarily a metadata decision.

input/example_metadata.json
Isgotkowitz: Maybe just add a rule that if a number is `None` it is acceptable? Or add a flag that indicates `None` is acceptable along with the min, max, and default, but that would necessitate a lot more change in `InputManager`

input/example_metadata.json
Isgotkowitz: It probably would not make a big difference, and since the cover type is dynamic the value the user provides is just an initial input. I will leave it as is for now.

input/example_metadata.json
ew3361zh: I think we'd want to catch the cases where someone just left it blank and they want the default to be subbed in (and then the default could be None or any number as outlined in the metadata).

Isgotkowitz: Catching blank values and subbing the default in sounds good - what about applying the same solution to the case when the user provides an incorrect type for a value? It would be a little clunky in this case for `InputManager` to see `null` in a number input, say that it's the wrong type, then replace it with the default `None`. But it would hopefully be pretty simple and could be applied pretty universally

ew3361zh: The json.load() method we're using to read the data from the input json file will convert the json null to None so I believe we could handle a blank entry and the null in a json file in the same fashion.

input/example_metadata.json
```diff
@@ -155,33 +190,326 @@
...
         "type": "array",
         "description": "Julian day(s) of year to harvest",
-        "minimum_length": 1
+        "minimum_length": 0,
+        "properties": {
+          "type": "number",
+          "minimum": 1,
+          "maximum": 366,
+          "default": []
+        }
```
ew3361zh, line 208: If I'm reading these correctly, the `default` in the `properties` section is for the array itself while the max/min/type are for the data in the array, is that right?

I'm curious what your thinking was here because it does raise an interesting question:

If the array's `maximum_length` or `minimum_length` guidelines were violated, we'd want the `default` empty array to be substituted in at that point to fix the data. But if the guidelines of the array's elements were violated (in this case if a Julian day number were supplied that were outside the 1-366 range), would we want to sub in the `default` empty array then as well?

ew3361zh: Are there default harvest Julian days you'd want subbed in? Or is that handled in the code if the crop receives a blank array?

Isgotkowitz: You raise a good point - this section is not well written. If the days are outside the range 1 - 366, there should be no default so the program should stop running. Will make that change now

Isgotkowitz:
@ew3361zh @allisterakun I wanted to let you know after talking with Pooya I changed the properties of field input files - instead of containing the path to the file with the soil, crop, etc specification, it contains the name of the blob that contains the associated specification data instead. This will not require any significant or structural changes in the metadata, and the implications for the `InputManager` are largely unchanged.

PR #675, opened 2023-07-19
IM get_data()

### LINKED ISSUE #664: Create IM get_data() function
This issue is to create a new funciton in the IM class called `get_data()` which will allow other parts of the RuFaS system to call and request data from IM.

allisterakun:
# #621  `_get_data()` in `RUFAS.input_manager`:

## `_get_data() -> Optional[Any]:`

This function gets the requested data from `self.__pool` and returns the value. 

#### `data_address: str`
The address of the requested data.

#### Optional[Any]

- The value of the requested data if found.
- Raise a `KeyError` if not found.

### DIFF
--- a/RUFAS/input_manager.py
+++ b/RUFAS/input_manager.py
@@ -414,8 +414,8 @@ class InputManager:
             parent_address = str(data_address.split("." + invalid_key)[0])
 
             om.add_warning("Data not found:", f"Cannot find \"{data_address}\", "
-                                              f"\"{parent_address}\" does not have attribute \"{invalid_key}\"",
+                                              f"\"{parent_address}\" does not have attribute \"{invalid_key}\".",
                            info_map)
 
             raise KeyError(f"Data not found: Cannot find \"{data_address}\", "
-                           f"\"{parent_address}\" does not have attribute \"{invalid_key}\"")
+                           f"\"{parent_address}\" does not have attribute \"{invalid_key}\".")
--- a/RUFAS/input_manager.py
+++ b/RUFAS/input_manager.py
@@ -6,7 +6,7 @@ import re
 
 import pandas as pd
 from RUFAS.output_manager import OutputManager
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 om = OutputManager()
 
@@ -345,4 +345,7 @@ class InputManager:
         # om.add_warning("Data fixed", f"Invalid data fixed: {key=}; {value=}", info_map)
         # where data is not fixable:
         # om.add_error("Data not fixable.", f"Unable to fix the invalid data: {key=}, {value=}.", info_map)
+        pass
+
+    def get_data(self, data_address: str) -> Optional[Any]:
         pass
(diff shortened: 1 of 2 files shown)


allisterakun:
- [x] Feature Implementation
- [x] Docstring
- [x] Logging
- [ ] Unit tests

RUFAS/input_manager.py
```diff
@@ -346,3 +346,31 @@ def _fix_data(self, key: str, value: Any) -> bool:
...
+        -------
+        Any
+            The requested data if found, else None.
+        """
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._fix_data.__name__,
+                    }
+        element_hierarchy = data_address.split('.')
+        variable_parent = reduce(lambda d, key: d[key], element_hierarchy[:-1],
+                                 self.__pool)
```
PooyaHekmati: This one is being used repeatedly in different parts of the IM code. Can we wrap it inside a function?

RUFAS/input_manager.py
```diff
@@ -346,3 +346,31 @@ def _fix_data(self, key: str, value: Any) -> bool:
...
+
+        if element_hierarchy[-1] in variable_parent.keys():
+            return variable_parent[element_hierarchy[-1]]
+
+        else:
+            om.add_warning("Data not found:", f"Cannot find {data_address}", info_map)
+            return None
```
PooyaHekmati: I think we should raise an exception here.

allisterakun: Done!

RUFAS/input_manager.py
```diff
@@ -381,3 +382,39 @@ def _fix_data(self, key: str, value: Any) -> bool:
...
+                    }
+
+        element_hierarchy = data_address.split('.')
+
+        try:
+            data_value = reduce(lambda d, key: d[key], element_hierarchy,
+                                self.__pool)
+            return data_value
+
+        except KeyError as key_error:
```
allisterakun, line 422: @PooyaHekmati I added some error handling so that the message would be more helpful in locating the error.
It will be something like this: 

> KeyError: 'Data not found: Cannot find variable "var1.var2.var5", "var1.var2" does not have attribute "var5"'

RUFAS/input_manager.py
```diff
@@ -382,3 +383,39 @@ def _fix_data(self, key: str, value: Any) -> bool:
...
         # om.add_error("Data not fixable.", f"Unable to fix the invalid data: {key=}, {value=}.", info_map)
         pass
+
+    def get_data(self, data_address: str) -> Optional[Any]:
+        """
+        Get the requested data from the pool
+
+        Parameters
+        ----------
+        data_address : str
```
ew3361zh, line 393: What do you think about adding some information about the format for the `data_address` for people calling it to know what the structure should look like? If it doesn't seem appropriate to go here, maybe it could go in the Examples section: https://numpydoc.readthedocs.io/en/latest/format.html#examples

PooyaHekmati (approved):

allisterakun:
- [x] Feature Implementation
- [x] Docstring
- [x] Logging
- [x] Unit tests

RUFAS/input_manager.py
allisterakun: Great Point! Just added one example.

ew3361zh (approved):
LGTM!

Issue #642, opened 2023-07-10
Handle `info_maps` field in aggregate output JSON file (OM CSV #2)


ltp23:
This second task is very similar to the first one. Here, we want to process the `info_maps` field in each key of the aggregate output JSON file. The `info_maps` field consists a list of dictionaries, all sharing a consistent set of keys. 

Our goal is to create a separate CSV file for each top-level key in the JSON. Each of these CSV files contains both the `values` and the `info_maps` field of the associated key in the JSON. Within the CSV file, every row represents a single dictionary, converted into a string using the str() method.

We will store these CSV files in the `output/CSVs/om/variables` directory. 

We also want to make this info_maps process optional by introducing an enable/disable flag into the design, with the 'disable' option set as default. Come up with the design as you see fits. 

Here is an example for this task:

```json
{
    "simple_key": {
        "values": ["Hi", "How", "Are", "You"],
        "info_maps": [
            {"map1": "value1", "map2": "value2", "map3": "value3"},
            {"map1": "value4", "map2": "value5", "map3": "value6"}
        ]
    }
}
```

The CSV output for the `simple_key_info_maps.csv` would be:

```csv
values,info_maps
hi,"{'map1': 'value1', 'map2': 'value2', 'map3': 'value3'}"
how,"{'map1': 'value4', 'map2': 'value5', 'map3': 'value6'}"
are,
you,
```

**Checklist**:

- [x] Iterate over the list of dictionaries within the `info_maps` field, and convert each dictionary into a string using the `str()` method.
- [x] For each key-value pair, create a new CSV file using the key as the file name.
- [x] Save each stringified dictionary in the `info_maps` list as a row in the respective CSV file.
- [x] Save each CSV output file into the `output/CSVs/om/variables/` directory.
- [x] Introduce an enable/disable flag into the design to make the conversion optional.
    - [x] Set 'disable' as the default option.
- [x] Check that each CSV file is correctly formatted, and the files contain the correct data.
- [x] Write unit tests:
     - [x] Check if the JSON to CSV conversion is performed correctly for `info_maps`.
     - [x] Check that each CSV file is saved at the correct location. 
     - [x] Test the behaviors of the enabled/disabled flag.

PR #676, opened 2023-07-19
OM CSV Task #2: info_map csv files.

michaelRichards99:
Adds the info_map field from each variale in the `OutputManager`'s variable_pool to the csv file. Saves these csv files in the `/output/CSVs/om/variables` directory.

This PR addresses issue #642, which has a more detailed explanation of the requested changes.


RUFAS/output_manager.py
```diff
@@ -520,25 +524,35 @@ def save_variables(self, save_path: str, dir_path: str,
...
 
         Parameters
         ----------
-        path : str
-            Path to the output directory for the OutputManager.
+        base_csv_dir : str
+            Path to the output directory for the OutputManager's csv files.
+
+        exclude_info_maps : bool
+            Flag for whether or not the user wants to inlcude info_maps data in their results files.
```
```suggestion
            Flag for whether or not the user wants to include info_maps data in their results files.
```

RUFAS/output_manager.py
```diff
@@ -520,25 +524,35 @@ def save_variables(self, save_path: str, dir_path: str,
...
         except Exception as e:
             raise e
 
         for key in pool.keys():
-            csv_file_path = os.path.join(path, self._generate_file_name(key, "csv"))
-            self._dict_to_file_csv({key: pool[key]}, csv_file_path)
+            variable_file_path = os.path.join(str(variables_path), self._generate_file_name(key, "csv"))
+            self._dict_to_file_csv({key: pool[key]}, variable_file_path, "values")
+            if not exclude_info_maps:
+                info_maps_file_path = os.path.join(str(info_map_path), self._generate_file_name(f"{key}_info_maps", "csv"))
```
ew3361zh: getting a line too long error here. I think even though we've got the flake8 line length violation at 127 in github we're still trying to be below 120 when possible.

RUFAS/output_manager.py
```diff
@@ -328,11 +332,11 @@ def _dict_to_file_csv(self, data_dict: Dict[str, Any], path: str) -> None:
...
-        if all("values" not in v for v in data_dict.values()):
+        if all(field not in v for v in data_dict.values()):
             self.add_error("save_dict_file_try", f"Unable to save {path} due to missing values.", info_map)
             return
 
-        df = pd.DataFrame({k: pd.Series(v['values'], dtype=object) for k, v in data_dict.items()})
+        df = pd.DataFrame({k: pd.Series(v[field], dtype=object) for k, v in data_dict.items()})
         try:
             df.to_csv(path, index=False)
         except Exception as e:
```
ew3361zh, line 369: write unit test to cover this exception correctly being raised.

RUFAS/output_manager.py
```diff
@@ -520,25 +524,35 @@ def save_variables(self, save_path: str, dir_path: str,
...
 
         """
         pool = self.variables_pool
+        variables_path = Path(base_csv_dir) / "variables"
+        info_map_path = Path(base_csv_dir) / "info_maps"
         try:
-            Path(path).mkdir(parents=True, exist_ok=True)
+            variables_path.mkdir(parents=True, exist_ok=True)
+            if not exclude_info_maps:
+                info_map_path.mkdir(parents=True, exist_ok=True)
```
ew3361zh: similarly I'd add a test to cover this exception being raised properly.

ew3361zh (changes requested):
I think this looks good, Michael. I have a couple small comments about spelling and a flake8 violation.

I think it would be good to expand test coverage for the exceptions being raised in `save_variables_to_csv_files` and `dict_to_file_csv`. If those are coming in a later PR, please just let me know as I think you mentioned this is part of a series of 3-4 PRs all related to this functionality.

RUFAS/output_manager.py
michaelRichards99: Thanks. I'll use 120 from now on.

RUFAS/output_manager.py
michaelRichards99: Done.

ew3361zh (approved):
Looks good to me!

RUFAS/output_manager.py
```diff
@@ -308,31 +308,72 @@ def _dict_to_file_json(self, data_dict: Dict[str, Any], path: str) -> None:
...
 
-        if all("values" not in v for v in data_dict.values()):
-            self.add_error("save_dict_file_try", f"Unable to save {path} due to missing values.", info_map)
+        if len(data_dict) == 0:
             return
 
-        df = pd.DataFrame({k: pd.Series(v['values'], dtype=object) for k, v in data_dict.items()})
+        variable_name = next(iter(data_dict))
+        mandatory_fields = ["values"] if exclude_info_maps else ["values", "info_maps"]
+        if not all(key in data_dict[variable_name].keys() for key in mandatory_fields):
```
ew3361zh: Nice, this is an improvement over the previous version. More clear to understand.

ew3361zh (approved):
Looks good, Michael. I think this addresses the requested changes as I understand them. Testing looks improved as well.

RUFAS/output_manager.py
michaelRichards99: Thanks Niko! I agree, much cleaner this time.

PooyaHekmati: Why is this needed? How is it possible to have missing values or info_maps?

RUFAS/output_manager.py
```diff
@@ -308,31 +308,72 @@ def _dict_to_file_json(self, data_dict: Dict[str, Any], path: str) -> None:
         except Exception as e:
             raise e
 
-    def _dict_to_file_csv(self, data_dict: Dict[str, Any], path: str) -> None:
+    def _dict_to_csv_column_list(self, data_dict: Dict[str, List[Any]], mandatory_fields: List[str]
```
PooyaHekmati, line 311: I don't think this function is needed, this is an aggressive application of SRP. Preparing the content fits well in the responsibility domain of `_dict_to_file_csv()`.

RUFAS/output_manager.py
```diff
@@ -308,31 +308,72 @@ def _dict_to_file_json(self, data_dict: Dict[str, Any], path: str) -> None:
...
                     }
         self.add_log("save_dict_file_try", f"Attempting to save to {path}.", info_map)
 
-        if all("values" not in v for v in data_dict.values()):
-            self.add_error("save_dict_file_try", f"Unable to save {path} due to missing values.", info_map)
+        if len(data_dict) == 0:
             return
 
-        df = pd.DataFrame({k: pd.Series(v['values'], dtype=object) for k, v in data_dict.items()})
+        variable_name = next(iter(data_dict))
```

RUFAS/output_manager.py
```diff
@@ -308,31 +308,72 @@ def _dict_to_file_json(self, data_dict: Dict[str, Any], path: str) -> None:
...
+        mandatory_fields = ["values"] if exclude_info_maps else ["values", "info_maps"]
+        if not all(key in data_dict[variable_name].keys() for key in mandatory_fields):
+            self.add_error(
+                "save_dict_file_try", f"Unable to save {path} due to missing values and/or info_maps.", info_map
+            )
+            return
+
+        data = data_dict[variable_name]
+        # preserve variable name in csv columns
+        data[variable_name] = data.pop("values")
```
PooyaHekmati: I think this leads to side effects.

PooyaHekmati (changes requested):
I think a little bit of over-engineering is happening here. Are all these complications needed? Are they justified?

RUFAS/output_manager.py
michaelRichards99: assuming the variable pool is valid then this check is not. I have this check here because the method name `_dict_to_file_csv` mentions nothing about the structure of the dictionary.

That being said, the rest of the method does assume some predefined structure. This can be removed if we assume all dictionaries have values and info_maps.

RUFAS/output_manager.py
michaelRichards99: Good catch, I'll remove this.

RUFAS/output_manager.py
michaelRichards99: `data_dict` is dictionary with a single key value pair. 
```json
{
    "simple_key": {
        "values": ["Hi", "How", "Are", "You"],
        "info_maps": [
            {"map1": "value1", "map2": "value2", "map3": "value3"},
            {"map1": "value4", "map2": "value5", "map3": "value6"}
        ]
    }
}
```
`next(iter(data_dict))` was the cleanest way to get the key. I implemented this to produce this csv:

```csv
simple_key,info_maps
hi,"{'map1': 'value1', 'map2': 'value2', 'map3': 'value3'}"
how,"{'map1': 'value4', 'map2': 'value5', 'map3': 'value6'}"
are,
you,
```

Looking back at the design, the name of the variable is not actually needed here, I should just use the string "values".

```csv
values,info_maps
hi,"{'map1': 'value1', 'map2': 'value2', 'map3': 'value3'}"
how,"{'map1': 'value4', 'map2': 'value5', 'map3': 'value6'}"
are,
you,
```

I'll remove this.

RUFAS/output_manager.py
michaelRichards99: For this specific PR, I agree. 

The next task for output manager csv file generation includes flattening complex data structures. This will result in 1 object being split into multiple csv columns. In that case, the logic in this method will be more complex, which is why I'm adding this helper method.

I can remove this from it's own function for this specific PR but I think I'll add it back when the logic becomes more complex in the following PR.

PooyaHekmati: Well in that case I think it can stay.

PooyaHekmati (approved):

PR #680, opened 2023-07-21
Irrigation backward compatibility

### LINKED ISSUE #674: Make irrigation functionality backwards compatible
This issue is for making the `Field` module compatible with 2 methods of determining irrigation for a field.
1) The old method - the amount of water (in mm) that is used to irrigate a field on a given day is hardcoded into the weather input file, and the amount of water specified in it is exactly the amount of water that will be applied to the field on any given day.
2) The new method - users specify *x* amount of water to be applied every *y* days, and this amount of water minus the total rainfall that occurred over the interval is the amount of water that will be applied at the end of every interval. 

**Note** - method 2 is already implemented in the code, this issue is for method 1 to be implemented in the code too.

**Solution**
The implementation for this can be pretty simply, possibly just modifying the function in `Field` to be

```python
    def _determine_watering_amount(self, rainfall: float, irrigation: float) -> float:
        """
        Snip
        irrigation: float
            The amount of irrigation specified in the weather input (mm).
        Snip
        """
        if not self.field_data.watering_occurs and irrigation > 0.0:
            self.field_data.annual_irrigation_water_use_total += irrigation
            return irrigation

       # Snip
```

Of course, the call to this function in `Field`'s `_cycle_water()` method will need to be modified to
```python
    def _cycle_water(self, current_weather: CurrentWeather):
        """
        Snip
        """
        watering_amount = self._determine_watering_amount(current_weather.rainfall, current_weather.irrigation)
        # Snip
```

A related feature that would be a nice-to-have, but is not a must-have, would be a function called something like `_record_field_watering()` that would record the date and amount of irrigation applied to a field. This function would be called in two places in `_determine_watering_amount()`, one spot where the old irrigation method is used, and the other when the new one is used.

matthew7838:
### Summary
This draft is to make the method `_determine_watering_amount()` backward compatible with the hard-coded irrigation in weather data. With this PR, it will add a condition to simply return the irrigation data straight from the provided weather data. Also, there is a method added to record the amount and time. All changes are tested to ensure full coverage. 

### Changes
- In `field.py`, `_determine_watering_amount` has added a field to take in irrigation, if the new irrigation field is provided it will simply use that as the watering amount
-  In `field.py`, `_determine_watering_amount` will now record the time and amount of watering when the old or new method is used.
- In `field.py`,  `_execute_daily_processes` and `_cycle_water` will now take in extra time parameters to help record the watering amount.
-  In `field.py`,  `_record_field_watering` is created to record the time and amount of watering occurrences..
- All changes are covered and tested

### DIFF
--- a/SC_redesign/Crop_and_Soil/field/field.py
+++ b/SC_redesign/Crop_and_Soil/field/field.py
@@ -888,7 +888,7 @@ class Field:
                 crop.data.cumulative_transpiration = 0.0
                 crop.data.cumulative_potential_evapotranspiration = 0.0
 
-    def _determine_watering_amount(self, rainfall: float) -> float:
+    def _determine_watering_amount(self, rainfall: float, irrigation: float) -> float:
         """Manages watering of the field.
 
         Parameters
--- a/SC_redesign/Crop_and_Soil/field/field.py
+++ b/SC_redesign/Crop_and_Soil/field/field.py
@@ -1192,6 +1192,7 @@ class Field:
         return
 
     def _record_field_watering(self, year: int, day: int, watering_amount: float):
+        """Records a watering event to the OutputManager."""
         info_map = {"class": self.__class__.__name__, "function": self._record_field_watering.__name__,
                     "prefix": f"field_name:'{self.field_data.name}'", "date": {"year": year, "day": day},
                     "field_size": self.field_data.field_size}
--- a/SC_redesign/Crop_and_Soil/field/field.py
+++ b/SC_redesign/Crop_and_Soil/field/field.py
@@ -925,7 +925,6 @@ class Field:
         user-specified data provided, the hard-coded irrigation will be ignored and only uses the new method.
 
         """
-        # Old method that uses hard coded irrigation amount from the weather data
         if self.field_data.watering_occurs:
             self.field_data.current_water_deficit -= rainfall
             self.field_data.current_water_deficit = max(0.0, self.field_data.current_water_deficit)
--- a/SC_redesign/Crop_and_Soil/field/field.py
+++ b/SC_redesign/Crop_and_Soil/field/field.py
@@ -122,7 +122,7 @@ class Field:
 
         # --- Whole-Field Methods ---
         # Allow non-management field processes (water/nutrient cycling) to occur
-        self._execute_daily_processes(current_weather)
+        self._execute_daily_processes(current_weather, time)
         # ... Other ...
 
         # --- Crop Management ---
@@ -766,13 +766,15 @@ class Field:
     # </editor-fold>
 
     # <editor-fold desc="--- Field-level Methods ---">
-    def _execute_daily_processes(self, current_weather: CurrentWeather) -> None:
+    def _execute_daily_processes(self, current_weather: CurrentWeather, time) -> None:
         """Executes all daily updates on this field's soil and crop objects.
 
         Parameters
         ----------
         current_weather : CurrentWeather
             Object containing the environment conditions on this day.
+        time : Time
+            Object containing the current year and day of the simulation.
 
         Notes
         -----
@@ -791,7 +793,7 @@ class Field:
                                                           snow_cover,
                                                           current_weather.annual_mean_air_temperature)
 
-        self._cycle_water(current_weather)
+        self._cycle_water(current_weather, time)
 
         for crop in self.crops:
             if crop.data.is_mature or crop.data.is_dormant:
@@ -806,12 +808,14 @@ class Field:
             crop.leaf_area_index.grow_canopy()
             crop.biomass_allocation.allocate_biomass(current_weather.incoming_light)
 
-    def _cycle_water(self, current_weather: CurrentWeather):
+    def _cycle_water(self, current_weather: CurrentWeather, time):
         """allow the water to cycle through the field.
 
         Args:
             current_weather: a CurrentWeather object, containing a collection of today's weather variables needed
                 for field processes.
+        time : Time
+            Object containing the current year and day of the simulation.
 
         Details: This method executes all water-related processes that occur within Crop and Soil objects. Having a
             separate method to handle water processes altogether is necessary because processes that effect water in the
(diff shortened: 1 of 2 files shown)


SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -911,21 +921,28 @@ def _determine_watering_amount(self, rainfall: float) -> float:
         resources is tracked on an annual basis, so that water budgeting may be accurately predicted.
 
         """
-        if not self.field_data.watering_occurs:
+        # Old method that uses hard coded irrigation amount from the weather data
+        if self.field_data.watering_occurs and irrigation > 0:
+            raise ValueError("Expected to use hardcoded irrigation data or specified amount, but tried to use both")
```
Isgotkowitz: This should behavior should not throw an error, and instead default to the new method of watering over the hard-coded irrigation values.

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -911,21 +921,28 @@ def _determine_watering_amount(self, rainfall: float) -> float:
         resources is tracked on an annual basis, so that water budgeting may be accurately predicted.
 
         """
-        if not self.field_data.watering_occurs:
+        # Old method that uses hard coded irrigation amount from the weather data
```
Isgotkowitz: Instead of an in-line comment, it would be more appropriate to append to the `Notes` section a description of how this method checks for the multiple ways of watering, how it decides between them, etc.

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -888,13 +893,18 @@ def _cycle_water(self, current_weather: CurrentWeather):
...
 
         Parameters
         ----------
         rainfall : float
             Amount of rainfall that occurs on this day (mm)
+        year : int
+            Year in which this watering occurs.
+        day : int
+            Julian day on which this watering occurs.
+        irrigation
```
Isgotkowitz: Irrigation needs description.

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -888,13 +893,18 @@ def _cycle_water(self, current_weather: CurrentWeather):
                 crop.data.cumulative_transpiration = 0.0
                 crop.data.cumulative_potential_evapotranspiration = 0.0
 
-    def _determine_watering_amount(self, rainfall: float) -> float:
+    def _determine_watering_amount(self, rainfall: float, year: int, day: int, irrigation=0.0) -> float:
```
Isgotkowitz: irrigation should  be type annotated. Also I'm curious, why default to 0.0?

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -1173,3 +1190,11 @@ def perform_annual_reset(self) -> None:
         self.soil.data.do_annual_reset()
         self.field_data.perform_annual_field_reset()
         return
+
+    def _record_field_watering(self, year: int, day: int, watering_amount: float):
+        """Records a watering event to the OutputManager."""
+        info_map = {"class": self.__class__.__name__, "function": self._record_field_watering.__name__,
+                    "prefix": f"field_name:'{self.field_data.name}'", "date": {"year": year, "day": day},
+                    "field_size": self.field_data.field_size}
```
```suggestion
                    "prefix": f"field:'{self.field_data.name}'", "date": {"year": year, "day": day},
                    "field_size": self.field_data.field_size, "units": "mm"}
```

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -1173,3 +1190,11 @@ def perform_annual_reset(self) -> None:
         self.soil.data.do_annual_reset()
         self.field_data.perform_annual_field_reset()
         return
+
+    def _record_field_watering(self, year: int, day: int, watering_amount: float):
+        """Records a watering event to the OutputManager."""
```
Isgotkowitz: Proper docstring

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -1173,3 +1190,11 @@ def perform_annual_reset(self) -> None:
         self.soil.data.do_annual_reset()
         self.field_data.perform_annual_field_reset()
         return
+
+    def _record_field_watering(self, year: int, day: int, watering_amount: float):
```
Isgotkowitz: Return type annotation

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -1173,3 +1190,11 @@ def perform_annual_reset(self) -> None:
...
+        """Records a watering event to the OutputManager."""
+        info_map = {"class": self.__class__.__name__, "function": self._record_field_watering.__name__,
+                    "prefix": f"field_name:'{self.field_data.name}'", "date": {"year": year, "day": day},
+                    "field_size": self.field_data.field_size}
+        value = {"watering_amount": watering_amount}
```
Isgotkowitz: This doesn't need to be in a dictionary; you can just pass the value in directly.

Isgotkowitz (changes requested):
Excellent start Matthew, this PR is most of the way there. Only one significant piece of method behavior should be changed, everything else is minor tweaks and clarity related.

SC_redesign/Crop_and_Soil/field/field.py
matthew7838: There's no need for that after checking all conditions, I forgot to remove it.

SC_redesign/Crop_and_Soil/field/field.py
matthew7838: I don't really understand this, might need more clarifications or examples.

SC_redesign/Crop_and_Soil/field/field.py
matthew7838: Yea, I was trying to ask for clarification on this case while you're away. Thanks for the clarification, Ed.

SC_redesign/Crop_and_Soil/field/field.py
Isgotkowitz:
```suggestion
        value = watering_amount
```

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -910,22 +921,30 @@ def _determine_watering_amount(self, rainfall: float) -> float:
         needs to be applied are reset at the end of every interval. The water that is added to the field from the farm's
         resources is tracked on an annual basis, so that water budgeting may be accurately predicted.
 
+        Old method of using hard-coded irrigation data will be used if there's no user specified data. If there's any
+        user-specified data provided, the hard-coded irrigation will be ignored and only uses the new method.
+
         """
-        if not self.field_data.watering_occurs:
+        # Old method that uses hard coded irrigation amount from the weather data
```
```suggestion
```

Isgotkowitz (approved):
Looks great Matthew, nice job

JoeWaddell (approved):
Looks great, and nice work on the tests - they're very straightforward.

PR #681, opened 2023-07-21
OM csv task #3: parsing complex datastructures.

michaelRichards99:
In this task, we are improving on the approach we used in Task 1. When the field of the aggregate JSON output file contains a list of dictionaries, we need a way to flatten it. 

For each dictionary in the list, we should save each key as a separate column header and the corresponding values as column values. If the value is a simple type (int, float, str, bool), it can be saved directly. For more complex types, they should be converted to a string using the `str()` method.

Here's an example to contrast this task with task 1

```json
{
    "simple_key": {
        "values": [{"key1": 1, "key2": [1, 1]}, {"key1": 2, "key2": [2,2]}, {"key1": [3,3], "key2": 3}],
         "info_maps": [
            {
                "subkey1": 1,
                "subkey2": "Hello",
                "subkey3": [1, 2, 3],
                "subkey4": {
                    "nestedkey1": "World",
                    "nestedkey2": [4, 5, 6]
                }
            },
            {
                "subkey1": 2,
                "subkey2": "Hi",
                "subkey3": [4, 5, 6],
                "subkey4": {
                    "nestedkey1": "There",
                    "nestedkey2": [7, 8, 9]
                }
            }
        ]
    }
}
```

(Task 1) The CSV output for the `simple_key.csv` would be:

```csv
simple_key
{"key1": 1, "key2": 1}
{"key1": 2, "key2": 2}
{"key1": 3, "key2": 3}
```

(This task) And the CSV output for the `complex_key.csv` would be:

```csv
key1,key2,info_map_subkey1,info_map_subkey2,info_map_subkey3,info_map_subkey4
1,"[1,1]",1,"Hello","[1, 2, 3]","{'nestedkey1': 'World', 'nestedkey2': [4, 5, 6]}"
2,"[2,2]",2,"Hi","[4, 5, 6]","{'nestedkey1': 'There', 'nestedkey2': [7, 8, 9]}"
3,"[3,3]",,,,
```


RUFAS/output_manager.py
```diff
@@ -328,7 +328,21 @@ def _dict_to_csv_column_list(self, data_dict: Dict[str, List[Any]], mandatory_fi
         """
         column_list = []
         for field in mandatory_fields:
-            column_list.append((field, pd.Series(data_dict[field], dtype=object)))
+            data_list = data_dict[field]
+            if len(data_list) != 0 and isinstance(data_list[0], dict):
```
ew3361zh: I think this could be shortened slightly:
```suggestion
            if data_list and isinstance(data_list[0], dict):
```

RUFAS/output_manager.py
```diff
@@ -328,7 +328,21 @@ def _dict_to_csv_column_list(self, data_dict: Dict[str, List[Any]], mandatory_fi
...
+                csv_column_lists: Dict[str, List[Any]] = {}
+                for data_dictionary in data_list:
```
ew3361zh: What do you think about using a different name for `data_dictionary`? With `data_dict` already being used, I think using a different name might help keep things clear.

tests/test_misc.py
```diff
@@ -330,29 +330,69 @@ def test_dict_to_csv_column_list(mock_output_manager: OutputManager) -> None:
...
     result = mock_output_manager._dict_to_csv_column_list(data, ["values", "info_maps"])
+    assert len(result) == 3
     (_, data_series) = result[0]
-    (_, info_maps_series) = result[1]
-    assert data_series.to_list() == data['values'] and info_maps_series.to_list() == data['info_maps']
+    (_, map1_series) = result[1]
+    (_, map2_series) = result[2]
+    assert data_series.to_list() == data['values'] \
+        and map1_series.to_list() == ['value1', 'value2'] \
+        and map2_series.to_list() == [1, 2]
```
ew3361zh: I think having separate assert statements would help with potential debugging vs. one longer one.
```suggestion
    assert data_series.to_list() == data['values'] 
    assert map1_series.to_list() == ['value1', 'value2']
    assert map2_series.to_list() == [1, 2]
```

RUFAS/output_manager.py
michaelRichards99: I renamed to `nested_dictionary`. Do you think this is better?

ew3361zh: I like that and think it helps with clarity, thanks.

ew3361zh (approved):
Nice work, Michael!

PooyaHekmati (approved):

PR #684, opened 2023-07-24
Fix newlines to be platform independent

michaelRichards99:
## Problem
Windows encodes newlines with 2 bytes: `\r\n`, whereas most other OSs encode it with 1 byte: `\n`

The `test_dict_to_csv` unit test in `test_misc.py` hard-coded newline as `\n`, which caused the unit test to fail when running on a windows machine.

## The Fix
Uses `os.linesep`, which encapsulates this difference.

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -320,12 +320,12 @@ def mock_output_manager(mocker) -> OutputManager:
 
 
 @pytest.mark.parametrize("data, expected_result, should_write", [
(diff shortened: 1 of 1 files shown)


JoeWaddell (approved):
Tested on my desktop PC, all tests now pass. Nice job getting this fixed!

ew3361zh (approved):
Everything working for me locally with this fix. Looks good to me, Michael, nice work!

PR #686, opened 2023-07-24
Fix Coverage Action

allisterakun:
Currently, the Coverage Action will post a comment to the PR with the current coverage status. However, if the user merges the branch into the master or if the user is pushing to the master branch directly, there will be nowhere to post such a comment. An error will then occur due to no URL set.

This PR fixes the above issue by checking if the URL is set before posing the comment.


ew3361zh (approved):
That looks like a good fix to me.

michaelRichards99 (approved):
Thanks for the quick fix.

Issue #683, opened 2023-07-24
Expand array input data validation capabilities in IM


ew3361zh:
When an array is detected in the data loaded from an input file, the first thing to do is to run the validation on itself (min len, max len, etc). 

Once the array is validated, loop through its elements and run validation for them as well. 

We probably need to call validate_element for them. 

This was requested by SMEs

ltp23:
- [X] Validate object type
- [X] Validate array type
- [X] Validate the default metadata json files
- [x] Keep track of the counters

Isgotkowitz:
This has been attempted once but was not completed. Needs to be groomed again.

Isgotkowitz:
Closed by #1345

PR #687, opened 2023-07-24
IM validate array elements

ew3361zh:
PR to validate array elements of any type.

### DIFF
--- a/RUFAS/input_manager.py
+++ b/RUFAS/input_manager.py
@@ -269,7 +269,8 @@ class InputManager:
             except KeyError:
                 raise KeyError(f"Invalid type {var_type}: Element must be type {data_type_to_validator_map.keys()}")
 
-            is_valid = validator(variable_properties, var_name, input_data_value)
+            is_valid = validator(variable_properties, var_name, input_data_value,
+                                 properties_blob_key, element_hierarchy)
 
             element_counter_and_validity["total_elements"] += 1
             if is_valid:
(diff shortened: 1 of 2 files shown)


ew3361zh:
Two main issues with solution as currently written:

1. The potential solution for validating an array of objects works but it essentially creates two parallel tracks within the validate_element() function to accommodate the structure of an array of objects which is contributing to _validate_data() now violating the complexity Codefactor limit.

I am working on trying to reduce the complexity but any suggestions are welcome.

2. The current solution does not work to validate a simple array of primitive types.

I see several possible ways forward for the latter issue of not being able to validate primitive types:
-Making all arrays be arrays of objects.
-Continuing as we had it before where we’re not validating the elements of arrays of primitive types but are validating elements of arrays of objects.
-Continuing to look for solution that will work with both types of arrays. If this is the preferred solution, I would want to have a meeting to discuss possible avenues forward.

ew3361zh:
Being taken care of elsewhere.

PooyaHekmati:
> Being taken care of elsewhere.

Where?

ew3361zh:
> > Being taken care of elsewhere.
> 
> Where?

It sounded like from the ATM that @ltp23 was working on it. We have an issue for this #683 which I thought would be a more appropriate place to note progress and next steps on addressing it.

PooyaHekmati:
> > > Being taken care of elsewhere.
> > 
> > 
> > Where?
> 
> It sounded like from the ATM that @ltp23 was working on it. We have an issue for this #683 which I thought would be a more appropriate place to note progress and next steps on addressing it.

I think it is better to keep the branch and let @ltp23 decide if he wants to use the work that is here or not.

Isgotkowitz:
@ew3361zh can this PR be closed?

ew3361zh:
> @ew3361zh can this PR be closed?

I think so. We've surpassed the need for it as a reference. This is being handled by #1345.

PR #688, opened 2023-07-25
Updating storage ammonia emission

ltp23:
This PR improves on the current implementation of ammonia calculations for slurry storage classes by adding more thorough documentation, error handling, and reorganizing helper functions and constants. Unit tests were also updated accordingly.


RUFAS/routines/manure/constants/gas_emission_constants.py
```diff
@@ -91,3 +91,18 @@ class GasEmissionConstants:
...
+    """Housing specific constant for solid and semi-solid manure (s/m)."""
+
+    SLURRY_MANURE_HSC = 19.0
+    """Housing specific constant for slurry manure (s/m)."""
+
+    LIQUID_MANURE_HSC = 4.1
+    """Housing specific constant for liquid manure (s/m)."""
+
+    SOLID_MANURE_THRESHOLD = 8.0
+    """Dry matter threshold for classifying solid and semi-solid manure."""
```
ew3361zh, line 105: Is there a unit on this? Maybe add (unitless) if not.

RUFAS/routines/manure/constants/gas_emission_constants.py
```diff
@@ -91,3 +91,18 @@ class GasEmissionConstants:
...
+
+    SLURRY_MANURE_THRESHOLD = 5.0
+    """Dry matter threshold for classifying slurry manure."""
```
ew3361zh, line 108: Same as for `SOLID_MANURE_THRESHOLD` - add unit or add (unitless) if no unit.

ew3361zh (approved):
Looks good, Loi!

RUFAS/routines/manure/gas_emissions/gas_emissions.py
```diff
@@ -590,6 +590,200 @@ def calc_housing_ammonia_emission(cls, num_animals: int, barn_area_per_animal: f
...
+        """
+        Calculate storage ammonia emissions for manure treatments.
+
+        Notes
+        -----
+        The equation used to calculate storage ammonia emission is:
+
+        .. math::
+
+            E_{NH_3} = total\\_storage\\_area \\cdot \\frac{TAN \\cdot c \\cdot \gamma}{r \\cdot M \\cdot Q}
```
JoeWaddell, line 608: What LaTeX viewer do you use? The VSCode popup isn't recognizing this equation, and using some of the online tools (like latexbase.com) it's failing to render properly. 

You can see in this screenshot how it shows up in the VSCode tooltip, where it skips the equation entirely:
image

And here for the wonky formatting in latexbase (which is likely a config error on my end):
image

JoeWaddell (approved):
Looks great, just a comment on how to quickly view LaTeX formatted equations.

RUFAS/routines/manure/constants/gas_emission_constants.py
ltp23: Those thresholds are unitless so I will update the docstrings on that.

RUFAS/routines/manure/gas_emissions/gas_emissions.py
ltp23: Screenshot 2023-07-27 at 9 23 08 AM

Thanks Joe. The latex renderers can be a hit or miss. I think it works ok in pycharm when I hover over it. I still need to check how the sphinx docs look.

PR #690, opened 2023-07-26
Moving outputs of manure module to the same location as animal module's

ltp23:
This PR adds the graphics-producing feature for the manure module and moves all the outputs (csv and graphics files) to the `output` directory, where the animal module outputs are also located. 

Design-wise, before, the manure module output handler was part of the manure manager, but now I decouple them and make the manure manager unaware of the output handler. I added two lines in the simulation engine's `simulate` to run the handler when the simulation loop was finished. Eventually, when new modules for making generic csv and graphics files are more robust, this whole manure module output handler will be phased out.


RUFAS/routines/manure/manure_manager.py
```diff
@@ -88,33 +87,34 @@ def __init__(self,
...
-                value = a list of tuples where each tuple represents a group of daily output objects.
-                    More precisely, each such tuple is of length 5, consisting of a
-                    pen object, a ManureHandlerDailyOutput object, a ReceptionPitDailyOutput object,
-                    a ManureSeparatorDailyOutput object, a ManureTreatmentDailyOutput object,
-                    and another ManureTreatmentDailyOutput object for the anaerobic digestion treatment if it is used.
+        list[dict]
+            A list of dictionaries containing the daily output data for each animal pen.
+            The keys and their associated data types of each dictionary are as follows:
+            {
+                'simulation_day': int,
```
Isgotkowitz, line 105: Will this list of dictionaries span multiple years? If so, it would be good the include the simulation year as well

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
```diff
@@ -114,27 +124,27 @@ def _initialize_daily_output_during_update(self, manure_treatment_daily_input: L
...
+                    (1 - self.config.total_solids_removal_efficiency_for_treatment)),
+            liquid_manure_total_volatile_solids=(
+                    manure_treatment_daily_input.liquid_manure_total_volatile_solids *
+                    (1 - self.config.volatile_solids_removal_efficiency_for_treatment)),
+            liquid_manure_phosphorus=(
+                    manure_treatment_daily_input.liquid_manure_phosphorus *
+                    (1 - self.config.phosphorus_removal_efficiency_for_treatment)),
+            liquid_manure_potassium=(
+                    manure_treatment_daily_input.liquid_manure_potassium *
+                    (1 - self.config.potassium_removal_efficiency_for_treatment)),
```
Isgotkowitz: I think moving each of these calculations to their own line before the `return` statement would make this code a lot more readable.

RUFAS/routines/manure/output_handler/manure_manager_output_handler.py
```diff
@@ -23,447 +22,698 @@
...
         ----------
         pen : ManureManagerPen
-            A ManureManagementPen object to be processed.
+            The ManureManagerPen object to process.
 
         Returns
         -------
-        Dict[str, List[Any]]
-            A dictionary of important pen attributes that is properly formatted to be converted to dataframe.
+        dict[str, list[Any]]
```
Isgotkowitz, line 110: What is the difference between these type annotations, and also is one of them the team standard? We have been using the capitalized ones, but if the lower case ones should be preferred we can switch over to them.

Isgotkowitz (approved):
LGTM, a few small comments about minor things but I was not able to spot any serious issues due to size and my unfamiliarity with the manure module.

RUFAS/routines/manure/output_handler/manure_manager_output_handler.py
ltp23: Since we are using Python version 3.10+, those type annotations are now supported natively without the need to be imported from somewhere else, hence the lowercase forms. I also like the use of pipe `|` for the Union type because it looks cleaner compared to the `Union` annotation. But overall, it is not necessary to make changes to the existing annotations we have.

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
ltp23: I will update this.

RUFAS/routines/manure/manure_manager.py
ltp23: I haven't used simulation year anywhere in the manure module, so I don't know how relevant it is. But by the way, how do you get quick access to the simulation year?

varrma (approved):
Looks good Loi !!

PR #693, opened 2023-07-27
CBPB class definition and potassium calculation.

michaelRichards99:
## Overview
This PR creates a manure treatment class for the Compost Bedded Pack Barn (CBPB) manure treatment. 

Additionally, It defines the method to update the potassium content of the bedding-manure mixture. 

### Potassium equation
K<sub>t+1</sub> = K<sub>t</sub> + K<sub>bedding</sub> + K<sub>manure</sub> - K<sub>loss</sub> 

We assume that both K<sub>bedding</sub> and K<sub>loss</sub> are 0 every day of the simulation.


RUFAS/routines/manure/manure_treatments/manure_treatment_types.py
```diff
@@ -9,4 +9,5 @@ class ManureTreatmentType(DefaultEnum):
     ANAEROBIC_DIGESTION = 'anaerobic digestion'
     ANAEROBIC_DIGESTION_AND_LAGOON = 'anaerobic digestion and lagoon'
     ANAEROBIC_DIGESTION_AND_LAGOON_WITH_SPLIT = 'anaerobic digestion and lagoon with split'
+    COMPOST_BEDDED_PACK_BARN = 'cbpb'
```
ltp23: Can you change the value to 'compost bedded pack barn'? Just to keep it consistent with what I will have in the manure_mangement.json.

RUFAS/routines/manure/manure_treatments/manure_treatment_cbpb.py
```diff
@@ -0,0 +1,79 @@
...
+
+        super().__init__(weather, time, manure_treatment_config)
+      
+    def calc_bedding_potassium_content(
+        self,
+        current_manure_bedding_mix_potassium: float,
+        additional_potassium_in_manure: float,
+        additional_potassium_in_bedding: float = 0,
+        potassium_loss: float = 0
+    ) -> float:
```
ltp23: I think this is probably the right place to put it here and later use it in the daily helper function. Can you prefix the name with underscore _ to make it private?

ltp23 (approved):

varrma (approved):

PR #696, opened 2023-07-28
Add corn silage and Triticale

### LINKED ISSUE #692: Add silage corn and triticale/durum wheat to available crop species

matthew7838:
### Summary
This PR aims to complete the V1 requirements to support corn silage and Triticale.

### Changes
- In `crop_data.py`, two children classes of `CropData`, `Triticale`, and `CornSilage` was created and will assign the correct values according to SWAT.
- In `species_data_factory.py`, both classes were added `CropSpecies` to enumerate
- The test was updated to reflect this change.

### Notes
- We use the closest analog data available for triticale which is the Durum Wheat data
- Not sure where to determine this `is_nitrogen_fixer` field, the current value is according to **chatGPT**.

### DIFF
--- a/SC_redesign/Crop_and_Soil/crop/crop_data.py
+++ b/SC_redesign/Crop_and_Soil/crop/crop_data.py
@@ -838,9 +838,41 @@ class Potato(CropData):
 
 @dataclass(kw_only=True)
 class Triticale(CropData):
-    """crop data class with default values for triticale"""
-    # TODO: triticale has unknown parameters, since it is not present in SWAT database.
-    #     Durum wheat is likely the closest analog.
+    """crop data class with default values for triticale
+    Notes
+    -------
+    We use the closest analog data available which is the Durum Wheat data
+    """
+    species: str = "durum_wheat"
+    name: str = "default durum_wheat"
+    plant_code: str = "DWHT"
+    scientific_name: str = "Triticum durum"
+    plant_category: PlantCategory = PlantCategory("cool_annual")
+    is_nitrogen_fixer: bool = False
+
+    minimum_temperature: float = 0.0
+    optimal_temperature: float = 15.0
+
+    max_leaf_area_index: float = 4.0
+    first_heat_fraction_point: float = 0.15
+    first_leaf_fraction_point: float = 0.01
+    second_heat_fraction_point: float = 0.50
+    second_leaf_fraction_point: float = 0.95
+    senescent_heat_fraction: float = 0.90
+
+    light_use_efficiency: float = 30.0
+
+    emergence_nitrogen_fraction: float = 0.0600
+    half_mature_nitrogen_fraction: float = 0.0231
+    mature_nitrogen_fraction: float = 0.0130
+    emergence_phosphorus_fraction: float = 0.0084
+    half_mature_phosphorus_fraction: float = 0.0032
+    mature_phosphorus_fraction: float = 0.0019
+
+    optimal_harvest_index: float = 0.40
+    min_harvest_index: float = 0.20
+    yield_nitrogen_fraction: float = 0.0263
+    yield_phosphorus_fraction: float = 0.0057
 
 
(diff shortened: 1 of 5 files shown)


SC_redesign/Crop_and_Soil/crop/crop_data.py
```diff
@@ -838,6 +838,73 @@ class Potato(CropData):
...
-    """crop data class with default values for triticale"""
-    # TODO: triticale has unknown parameters, since it is not present in SWAT database.
-    #     Durum wheat is likely the closest analog.
+    """crop data class with default values for triticale
+    Notes
+    -------
+    We use the closest analog data available which is the Durum Wheat data
+    """
+    species: str = "durum_wheat"
+    name: str = "default durum_wheat"
```
Isgotkowitz: I think these should be changed to triticale, the current is the more accurate way but it may be confusing if someone specifies triticale to be planted then sees durum wheat recorded in the output.

SC_redesign/Crop_and_Soil/crop/crop_data.py
```diff
@@ -838,6 +838,73 @@ class Potato(CropData):
...
+    half_mature_nitrogen_fraction: float = 0.0177
+    mature_nitrogen_fraction: float = 0.0138
+    emergence_phosphorus_fraction: float = 0.0048
+    half_mature_phosphorus_fraction: float = 0.0018
+    mature_phosphorus_fraction: float = 0.0014
+
+    optimal_harvest_index: float = 0.90
+    min_harvest_index: float = 0.90
+    yield_nitrogen_fraction: float = 0.0140
+    yield_phosphorus_fraction: float = 0.0016
```
Isgotkowitz: This looks perfect, could you move it up to be just beneath the other Corn data class?

SC_redesign/Crop_and_Soil/crop/species_data_factory.py
```diff
@@ -19,7 +19,8 @@ class CropSpecies(Enum):
     SOYBEAN = "soybean"
     SUGAR_BEET = "sugar_beet"
     POTATO = "potato"
-    TRITICALE = "triticale"
+    TRITICALE = "durum_wheat"
+    CornSilage = "corn_silage"
```
Isgotkowitz:
```suggestion
    CORN_SILAGE = "corn_silage"
```

Isgotkowitz (approved):
Nice job Matthew, only a few comments on naming and ordering.

SC_redesign/Crop_and_Soil/crop/species_data_factory.py
```diff
@@ -19,7 +19,8 @@ class CropSpecies(Enum):
     SOYBEAN = "soybean"
     SUGAR_BEET = "sugar_beet"
     POTATO = "potato"
-    TRITICALE = "triticale"
+    TRITICALE = "durum_wheat"
```
ew3361zh: I think similar to what Ed said in the crop_data file, it's a little confusing to have triticale and durum_wheat be conflated like this. I understand the data you have is for the durum wheat but it seems like acknowledging that somewhere and then keeping triticale as triticale throughout would be clearer.

ew3361zh: But obviously do what makes sense for the C&S team as I haven't been privy to most of those conversations.

ew3361zh (approved):
Looks good. Just a note of confusion from my end on the conflation of durum_wheat and triticale. I think the note in the docstrings is enough to explain the data being used is durum wheat data as a close analog but I don't know if I'd change the name of species from triticale in the code.

SC_redesign/Crop_and_Soil/crop/species_data_factory.py
matthew7838: Thanks, Niko, I have followed Ed's suggestions which should help with the confusion.

PR #701, opened 2023-07-28
Om CSV filtering with OM's regex filters

michaelRichards99:
In this task, we are integrating the output manager's filtering capabilities to allow the users to select which variables they want in the CSV output.

Regex filtering already exists in the output manager. We should investigate if that infrastructure can be reused for this task. If not, we can re-use as much of it as we can to add this functionality.

Filter patterns for `json` and `csv` output are distinct. The users will be able to specify which patters apply to which format. This is achieved by having 1 distinct files for json and csv filters. Any file starting with `json_` will be interpreted as having regex patterns for json output. Any file starting with `csv_` will interpreted as having regex patterns for csv output.

This builds off of the previous output manager csv pr #681 .


RUFAS/output_manager.py
```diff
@@ -546,33 +561,44 @@ def save_variables(self, save_path: str, dir_path: str,
...
-            file_path = os.path.join(save_path, self._generate_file_name(f"saved_variables_{filter_file}", "json"))
-            self._dict_to_file_json(filtered_pool, file_path)
+            if filter_file.startswith("json_"):
+                if exclude_info_maps:
+                    filtered_pool = self._exclude_info_maps(filtered_pool)
+                file_path = os.path.join(save_path, self._generate_file_name(f"saved_variables_{filter_file}", "json"))
+                self._dict_to_file_json(filtered_pool, file_path)
+            elif filter_file.startswith("csv_"):
+                self._save_variables_to_csv_files(
+                    filtered_pool, str(Path(save_path) / "CSVs" / "om" / "variables"), exclude_info_maps
```
PooyaHekmati: This is not consistent. Info map exclusion is handled differently for csv and json. For JSON exclusion happens here, where as for csv it is handled in the downstream function. Can you make it consistent by either handling both here or handling both in the downstream function?

RUFAS/output_manager.py
```diff
@@ -328,7 +329,21 @@ def _dict_to_csv_column_list(self, data_dict: Dict[str, List[Any]], mandatory_fi
...
-            column_list.append((field, pd.Series(data_dict[field], dtype=object)))
+            data_list = data_dict[field]
+            if data_list and isinstance(data_list[0], dict):
+                csv_column_lists: Dict[str, List[Any]] = {}
+                for nested_dictionary in data_list:
+                    for subkey, value in nested_dictionary.items():
+                        if subkey in csv_column_lists:
+                            csv_column_lists[subkey].append(value)
+                        else:
+                            csv_column_lists[subkey] = [value]
```
PooyaHekmati: can you explain what is happening here?

PooyaHekmati (changes requested):

RUFAS/output_manager.py

RUFAS/output_manager.py
michaelRichards99: This is how we go from a list of dictionaries to a single dictionary.

```json
[
  {
     "subkey1": 1,
     "subkey2": "Hello",
     "subkey3": [1, 2, 3],
      },
      {
      "subkey1": 2,
      "subkey2": "Hi",
      "subkey3": [4, 5, 6],
       }
]
```
becomes

```json
{
   "subkey1": [1,2],
   "subkey2": ["hello", "hi"],
   "subkey3": [1, 2, 3, 4, 5, 6],
}
```

Looping over an array of dictionaries, on the first iteration, we haven't seen the key yet so we need to initialize a list.

In the newest commit, rather than having this branch inside the loop, I've created the dictionary with all the keys and empty lists before the loop.

RUFAS/output_manager.py
```diff
@@ -359,10 +366,8 @@ def _dict_to_file_csv(self, data_dict: Dict[str, Any], path: str, exclude_info_m
         if len(data_dict) == 0:
             return
```
ew3361zh, line 368: Apologies that this is not the focus of the PR but what do you think about adding a log/warning if the data_dict is empty? I think it could be helpful since there will be a log about trying to save to the path but if the length is zero it just returns with no further info to the user on what happened and why the csv wasn't saved.

ew3361zh (approved):
LGTM. I previously reviewed and approved #681 which looks like it's incorporated here. Not sure if it's important to merge them separately to have the distinct file histories/comments conversations.

ew3361zh:
Also curious if you could explain why it's important to have distinct csv and json filter files? Is it just for saved variables file naming purposes?

RUFAS/output_manager.py
michaelRichards99: Is an empty `data_dict` possible? If it's not possible, then I think it'll be good to add a log here. If it's possible that the simulation results in an empty variable in the variable pool, then I think it'll be better to move the log about trying to save to the path after this check.

michaelRichards99:
> Also curious if you could explain why it's important to have distinct csv and json filter files? Is it just for saved variables file naming purposes?

In discussions with @PooyaHekmati , we decided to offer the ability for users to be able to distinguish their filter patterns based on the output type. This gives the user more fine-grained control of the outputs.

RUFAS/output_manager.py
ew3361zh: It's definitely possible to have an empty info_map. I think it's also possible to have an empty values section. I think regardless of the first log statement, if we have a check on the len of the data_dict being zero, it seems like valuable information to log/warn the user about. Maybe this isn't the most appropriate place for that check and warning to be done but it triggered this question because you had the return statement there.

ew3361zh:
> > Also curious if you could explain why it's important to have distinct csv and json filter files? Is it just for saved variables file naming purposes?
> 
> In discussions with @PooyaHekmati , we decided to offer the ability for users to be able to distinguish their filter patterns based on the output type. This gives the user more fine-grained control of the outputs.

That makes sense. I guess it's easier and saves space to have more (and more precise) filter flies than it is to have OM go through every filter file to create both a csv and json file each time. Thank you!

RUFAS/output_manager.py
michaelRichards99: Thank you for the feedback, Niko. I agree with you here, more logging information is never a bad thing. I'll add a log here.

PooyaHekmati (approved):

PR #702, opened 2023-07-29
init EEE module

PooyaHekmati:
implement Crop Management Practice Fuel Consumption Estimate, IDs 1001 to 6037
Input data and accompanying meta data yet to be developed.

### DIFF
--- a/RUFAS/economy_emission_energy.py
+++ b/RUFAS/economy_emission_energy.py
@@ -1,0 +1,150 @@
+from math import cos, sin
+from typing import Dict
+from RUFAS.input_manager import InputManager
+from RUFAS.output_manager import OutputManager
+
+IM = InputManager()
+OM = OutputManager()
+
+
+class EconomyEneregyEmission:
+    def __init__(self) -> None:
+        pass
+
+    def _calculate_diesel_consumption(
+        self,
+        acre_size: float,
+        drawbar_speed: float,
+        field_capacity: float,
+        functional_draft: float,
+        harvest_yield: float,
+        IFSM_operation_data_E: float,
+        IFMS_operation_data_F: float,
+        IFSM_operation_data_G: float,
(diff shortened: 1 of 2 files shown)


RM2789 (approved):
Dear Professor and Pooya ,

Yes, this code is fine to go ahead with. Thank you for your effort. 
Roohi

PooyaHekmati:
Thanks @RM2789 
I have a question though: why do we need to calculate sin and cos of 0 every time? They don't changed, why not just using their known results (0 and 1)?

RUFAS/economy_emission_energy.py
```diff
@@ -0,0 +1,150 @@
...
+        -------
+        None
+
+        Notes
+        -----
+        This method assumes that the instance of the class has access to a data source with fuel consumption scenarios,
+        and the scenarios are stored as a dictionary where the keys are scenario names and the values are dictionaries
+        containing the details required for calculating diesel consumption. The format of the scenarios dictionary
+        should be:
+
```
KFosterReed, line 117: I think this is a great, simple structure for these calculations. I want to highlight here a couple of things about how the scenarios will integrate with the information coming out of the biophysical model.

1. Not all of the information for the scenarios will come from the Input manager. 
2. I think the scenarios in this case will correspond to an operation per field per year so would a good scenario name be something like 'operation_field_year'?

Otherwise, it looks great!

RM2789 (commented):
Dear Pooya,

I developed the equation by referring to the method shown in the attached Excel sheet (shared by Peter Vadas). For all variation, it is cos(0) and sin(0), so going by that, yes, we can change it to constants. 

[4 pass shred rake bale.xls](https://github.com/RuminantFarmSystems/MASM/files/12231094/4.pass.shred.rake.bale.xls)
Thank you
Roohi

PR #704, opened 2023-08-01
IM object filtering patch

ew3361zh:
This PR fixes an overlooked issue with validating nested objects in the IM's `_validate_element()` method by creating a list of keys to ignore in the nested validation loop.

Per the IM design doc, the metadata will have this structure for a nested json blob:

`"properties": {
   "<var name>": {
    "type": "object" 
    "description": "<optional description for this variable>",
     "<nested var name>": {
     "type": "<an item from list:[number, array, bool, string, –what else?-]>" 
     "description": "<optional description for this variable>",
     <data description template>
     }
   }
}`

In `_validate_element()`, if the property type is _object_, it will loop through each of the keys in the object and attempt to validate the corresponding data from the input file against them.

We don't want the IM to attempt to validate the values for the keys "type" and (optional) "description".

This PR also updates unit testing and fixes another error in the return value of the nested validation loop.

### DIFF
--- a/tests/test_input_manager.py
+++ b/tests/test_input_manager.py
@@ -350,6 +350,11 @@ def test_validate_element_string_type(mock_input_manager: InputManager,
 
     assert result["is_valid"] is True
 
+    input_data = {"element1": "invalid_value"}
+    result = mock_input_manager._validate_element(["element1"], "property_map_key1", input_data, True)
+
+    assert result["is_valid"] is False
+
     mock_input_manager._validate_element = input_manager_original_method_states["_validate_element"]
 
 
@@ -363,6 +368,11 @@ def test_validate_element_number_type(mock_input_manager: InputManager,
     result = mock_input_manager._validate_element(["element2"], "property_map_key1", input_data, True)
 
     assert result["is_valid"] is True
+
+    input_data = {"element2": 500}
+    result = mock_input_manager._validate_element(["element2"], "property_map_key1", input_data, True)
+
+    assert result["is_valid"] is False
 
     mock_input_manager._validate_element = input_manager_original_method_states["_validate_element"]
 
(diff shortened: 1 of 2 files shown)


tests/test_input_manager.py
```diff
@@ -341,6 +354,20 @@ def test_validate_element_array_type(mock_input_manager: InputManager,
     mock_input_manager._validate_element = input_manager_original_method_states["_validate_element"]
 
 
+def test_validate_element_object_type(mock_input_manager: InputManager,
```
Isgotkowitz: Is there a test that checks that invalid nested data causes `"is_valid"` to be false?

Isgotkowitz (approved):
One question about test suite coverage, otherwise LGTM. Nice job Niko

Isgotkowitz:
> `"type": "<an item from list:[number, array, bool, string, –what else?-]>"`

What about a `map` type (i.e. dictionary)? There is one variable in particular in the Crop and Soil module that would benefit from having this type supported

tests/test_input_manager.py
ew3361zh: I will add one for that!

tests/test_input_manager.py
```diff
@@ -341,6 +381,77 @@ def test_validate_element_array_type(mock_input_manager: InputManager,
     mock_input_manager._validate_element = input_manager_original_method_states["_validate_element"]
 
 
+def test_validate_element_valid_object_type(mock_input_manager: InputManager,
```
michaelRichards99, line 399: For these unit tests would it be possible to add a comment explaining why the test should pass or fail? I think a simple comment would go a long way in maintainability. Do you think it is worth adding? I know we try to avoid comments when possible.

ew3361zh: What do you think about more descriptive docstrings vs. inline comments?

michaelRichards99: For a more high-level description of why the unit test should pass/fail, docstrings work better, IMO. 

It'll be much easier to read the comment rather than having to trace through all the steps to find out why the test passes/fails.

ew3361zh: Currently have some info in the docstrings about each test but I can expand that info to be a little clearer.

ew3361zh: I also did spread these out into individual tests rather than parametrizing the inputs in hopes to provide clearer results.

michaelRichards99: Thanks.

michaelRichards99 (approved):
Properly fixes the problem.

Issue #695, opened 2023-07-28
Unrealistic initial herd structure from the current animal database 


tomhuhh:
## Expected Behavior
Cow percent for parity would follow a similar pattern as shown below. 

1st lactation: 40% 
2nd lactation:  36% 
3rd lactation:  17.5% 
3rd lactation+: 6.5%   

We would like to randomly select 1000 cows from the animal database but with a seed. 

## Current Behavior

1st lactation: 2% 
2nd lactation:  0% 
3rd lactation:  18% 
3rd lactation+: 80%   

![image](https://github.com/RuminantFarmSystems/MASM/assets/52800053/335096e0-e029-4618-b723-c19f11cf0dd5)
![image](https://github.com/RuminantFarmSystems/MASM/assets/52800053/f7a8fac9-2920-4634-a3bf-c6520c5473f7)
![image](https://github.com/RuminantFarmSystems/MASM/assets/52800053/643f1be7-e5bf-4ae0-ad23-e8832a99befa)
![image](https://github.com/RuminantFarmSystems/MASM/assets/52800053/2e540628-9ed1-40cd-85bd-7e4d47a89bb5)

This initial herd structure makes it more difficult (takes longer time) to achieve a stable herd structure in the simulations. 

The first 1000 cows were selected from the animal database in order because 'set_seed' is set to TRUE. 

## Possible Solution

Reset the animal database by changing the "herd_init" in the "animal_management_animal.json" file from "false" to "true" with a reasonable herd inputs in the same file. 

## Steps to Reproduce

1. Simulate 10 years' data using the current version of RuFaS with the "animal_management_animal.json" file as an input. 
2.
3.
4.

tomhuhh:
Updating this issue to a future issue because we have since discovered that the issue is not related to the database itself but rather with how 'set_seed' functions in the code. In the development version of RuFaS, 'set_seed' is set to True. Within the AnimalInitialization class, the 'set_seed' value is used to determine whether animals are selected in a random order from the animal database, ensuring representative sampling from each parity. However, when 'set_seed' is set to True, the animals are not chosen randomly, resulting in an unrepresentative herd structure. In the user version of RuFaS, the code functions correctly because 'set_seed' is set to False. 

A refactoring is likely needed in the future but this is at a lower priority.

JoeWaddell:
When we aren't using a seed, we can randomly sample from the animal database correctly. 

[...]

PR #708, opened 2023-08-01
Fix animal database

tomhuhh:
This PR is to regenerate a new animal database to replace the old database where the percentages for parities one and two are low (< 5%) after the selection.


tomhuhh:
Closing this pull request because we have since discovered that the issue is not related to the database itself but rather with how 'set_seed' functions in the code. In the development version of RuFaS, 'set_seed' is set to True. Within the AnimalInitialization class, the 'set_seed' value is used to determine whether animals are selected in a random order from the animal database, ensuring representative sampling from each parity. However, when 'set_seed' is set to True, the animals are not chosen randomly, resulting in an unrepresentative herd structure.

In the user version of RuFaS, the code functions correctly because 'set_seed' is set to False. This issue has also been discussed in issue #695. Therefore, there is no need to regenerate a database for development purposes.

Closing the branch as well...

PR #710, opened 2023-08-02
Negative water evaporation bug draft

### LINKED ISSUE #643: Negative `water_evaporated` values
## Expected Behavior
The amount of evaporation from a field on a single day should be lower bounded at 0.

## Current Behavior

The evaporation from the field is sometimes a negative amount, which is physically impossible.

## Possible Solution

Find where evaporation from the field is calculated, and lower bound it at 0.

## Steps to Reproduce

1. Run `python main.py input/ARL.json`
2. Look for the variable with the name `"field:'field_1'.water_evaporated"`, and look through the values present. There are a fair number of negative ones.
3.
4.

matthew7838:
### Summary
This small PR is to find out what causing the negative water_evporation value in the output and fix it. This will solve issue #643.

### Changes

- In `field.py`, `_determine_potential_evapotranspiration` will now set a minimum return value to 0.0 if `avg_air_temp` was too small.
- A few changes to condition statements in `evaporation.py`.
### Notes
These values might need to be re-evaluated again by science experts.

### DIFF
--- a/SC_redesign/Crop_and_Soil/crop/water_uptake.py
+++ b/SC_redesign/Crop_and_Soil/crop/water_uptake.py
@@ -196,7 +196,7 @@ class WaterUptake:
     @staticmethod
     def _adjust_water_uptakes(potential_uptakes: List[float], unmet_demands: List[float],
                               uptake_compensation: float) -> List[float]:
-        """adjusts the potential water uptakes from each layer based by drawing from deeper layers when possible.
+        """Adjusts the potential water uptakes for each layer based on drawing from deeper layers when possible.
 
         References
         ----------
--- a/SC_redesign/Crop_and_Soil/crop/water_uptake.py
+++ b/SC_redesign/Crop_and_Soil/crop/water_uptake.py
@@ -196,7 +196,7 @@ class WaterUptake:
     @staticmethod
     def _adjust_water_uptakes(potential_uptakes: List[float], unmet_demands: List[float],
                               uptake_compensation: float) -> List[float]:
-        """adjusts the potential water uptakes from each layer based by drawing from deeper layeres when possible.
+        """adjusts the potential water uptakes from each layer based by drawing from deeper layers when possible.
 
         References
         ----------
--- a/SC_redesign/Crop_and_Soil/field/field.py
+++ b/SC_redesign/Crop_and_Soil/field/field.py
@@ -1113,8 +1113,10 @@ class Field:
         for crop in self.crops:
             amount_evaporated = crop.water_dynamics.evaporate_from_canopy(remaining_evapotranspirative_demand)
             remaining_evapotranspirative_demand -= amount_evaporated
-            if remaining_evapotranspirative_demand == 0.0:
+            if remaining_evapotranspirative_demand - amount_evaporated <= 0.0:
                 break
+            else:
+                remaining_evapotranspirative_demand -= amount_evaporated
         return remaining_evapotranspirative_demand
 
     def _determine_total_above_ground_biomass(self) -> float:
--- a/SC_redesign/Crop_and_Soil/field/field.py
+++ b/SC_redesign/Crop_and_Soil/field/field.py
@@ -1153,20 +1153,22 @@ class Field:
         Notes
         -----
         This method calculates the evapotranspirative demand for the entire field on any given day using the Hargreaves
-        method.This method will also cap the minimum to zero if negative average values were inputted, which causes
-        nonsense negative evapotranspirative demand.
+        method.This method lower-bounds the potential evapotranspiration at 0.0 mm.
         """
-        if max_air_temp - min_air_temp < 0:
-            raise Exception
         if avg_air_temp is None:
             calculated_avg_air_temp = (max_air_temp + min_air_temp) / 2
             latent_heat_vaporization = Field._determine_latent_heat_vaporization(calculated_avg_air_temp)
-            return max(0, (0.0023 * extra_terrestrial_radiation * ((max_air_temp - min_air_temp) ** (-0.5))
-                           * (calculated_avg_air_temp + 17.8)) / latent_heat_vaporization)
+            potential_evapotranspiration = (0.0023 *
+                                            extra_terrestrial_radiation * ((max_air_temp - min_air_temp) ** (-0.5))
+                                            * (calculated_avg_air_temp + 17.8)) / latent_heat_vaporization
+            return max(0, potential_evapotranspiration)
+
         else:
             latent_heat_vaporization = Field._determine_latent_heat_vaporization(avg_air_temp)
-            return max(0, (0.0023 * extra_terrestrial_radiation * ((max_air_temp - min_air_temp) ** (-0.5))
-                           * (avg_air_temp + 17.8)) / latent_heat_vaporization)
+            potential_evapotranspiration = (0.0023 *
+                                            extra_terrestrial_radiation * ((max_air_temp - min_air_temp) ** (-0.5))
+                                            * (avg_air_temp + 17.8)) / latent_heat_vaporization
+            return max(0, potential_evapotranspiration)
 
     @staticmethod
     def _determine_latent_heat_vaporization(avg_air_temp: float) -> float:
--- a/SC_redesign/Crop_and_Soil/field/field.py
+++ b/SC_redesign/Crop_and_Soil/field/field.py
@@ -1153,18 +1153,20 @@ class Field:
         Notes
         -----
         This method calculates the evapotranspirative demand for the entire field on any given day using the Hargreaves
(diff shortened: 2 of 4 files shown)


SC_redesign/Crop_and_Soil/crop/water_uptake.py
```diff
@@ -196,7 +196,7 @@ def _correct_layer_for_efficiency(potential_uptake: float, available_water: floa
     @staticmethod
     def _adjust_water_uptakes(potential_uptakes: List[float], unmet_demands: List[float],
                               uptake_compensation: float) -> List[float]:
-        """adjusts the potential water uptakes from each layer based by drawing from deeper layeres when possible.
+        """adjusts the potential water uptakes from each layer based by drawing from deeper layers when possible.
```
Isgotkowitz:
```suggestion
        """Adjusts the potential water uptakes from each layer based by drawing from deeper layers when possible.
```

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -1153,18 +1153,20 @@ def _determine_potential_evapotranspiration(extra_terrestrial_radiation: float,
         Notes
         -----
         This method calculates the evapotranspirative demand for the entire field on any given day using the Hargreaves
-        method.
-
+        method.This method will also cap the minimum to zero if negative average values were inputted, which causes
+        nonsense negative evapotranspirative demand.
```
```suggestion
        method. This method lower-bounds the potential evapotranspiration at 0.0 mm.
```

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -1153,18 +1153,20 @@ def _determine_potential_evapotranspiration(extra_terrestrial_radiation: float,
...
         """
+        if max_air_temp - min_air_temp < 0:
+            raise Exception
```
Isgotkowitz: This should either have a much more descriptive error raised, or it should be removed. If there is a problem with the weather data it should be caught and dealt with at a much higher level than the `Field` module, but on the other hand, if having invalid min and max air temps will really wreck the result of this function then its a good idea to check them.

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -1153,18 +1153,20 @@ def _determine_potential_evapotranspiration(extra_terrestrial_radiation: float,
...
         if avg_air_temp is None:
             calculated_avg_air_temp = (max_air_temp + min_air_temp) / 2
             latent_heat_vaporization = Field._determine_latent_heat_vaporization(calculated_avg_air_temp)
-            return (0.0023 * extra_terrestrial_radiation * ((max_air_temp - min_air_temp) ** (-0.5))
-                    * (calculated_avg_air_temp + 17.8)) / latent_heat_vaporization
+            return max(0, (0.0023 * extra_terrestrial_radiation * ((max_air_temp - min_air_temp) ** (-0.5))
+                           * (calculated_avg_air_temp + 17.8)) / latent_heat_vaporization)
```
```suggestion
            potential_evapotranspiration = (0.0023 * extra_terrestrial_radiation * ((max_air_temp - min_air_temp) ** (-0.5))
                           * (calculated_avg_air_temp + 17.8)) / latent_heat_vaporization
            return max(0, potential_evapotranspiration)
```
I think this would improve readability.

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -1153,18 +1153,20 @@ def _determine_potential_evapotranspiration(extra_terrestrial_radiation: float,
...
         else:
             latent_heat_vaporization = Field._determine_latent_heat_vaporization(avg_air_temp)
-            return (0.0023 * extra_terrestrial_radiation * ((max_air_temp - min_air_temp) ** (-0.5))
-                    * (avg_air_temp + 17.8)) / latent_heat_vaporization
+            return max(0, (0.0023 * extra_terrestrial_radiation * ((max_air_temp - min_air_temp) ** (-0.5))
+                           * (avg_air_temp + 17.8)) / latent_heat_vaporization)
```
Isgotkowitz: Same as above

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -1153,18 +1153,20 @@ def _determine_potential_evapotranspiration(extra_terrestrial_radiation: float,
...
         -----
         This method calculates the evapotranspirative demand for the entire field on any given day using the Hargreaves
-        method.
-
+        method.This method will also cap the minimum to zero if negative average values were inputted, which causes
+        nonsense negative evapotranspirative demand.
         """
+        if max_air_temp - min_air_temp < 0:
+            raise Exception
         if avg_air_temp is None:
```
Isgotkowitz, line 1158: Is the lower bounding in this function tested?

Isgotkowitz (changes requested):
Nice fix Matthew. Requested changes on some docstrings, and also want to make sure that changes are being properly tested.

SC_redesign/Crop_and_Soil/crop/water_uptake.py
ew3361zh: This sentence is still a little confusingly worded. I think it's something in the "...from each layer based by drawing from..." construction. 

```suggestion
        """Adjusts the potential water uptakes from each layer by drawing from deeper layers when possible.
```

Are there multiple potential water uptakes for each layer?

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -1153,18 +1153,20 @@ def _determine_potential_evapotranspiration(extra_terrestrial_radiation: float,
         Notes
         -----
         This method calculates the evapotranspirative demand for the entire field on any given day using the Hargreaves
-        method.
-
+        method.This method will also cap the minimum to zero if negative average values were inputted, which causes
+        nonsense negative evapotranspirative demand.
         """
+        if max_air_temp - min_air_temp < 0:
```
ew3361zh: With the check on avg_air_temp below being None, is it worth doing a similar check on these two? I guess at least you can do something to address avg_air_temp  is None whereas there's not as easy a solution for either of these being None.

SC_redesign/Crop_and_Soil/field/field.py
ew3361zh: I think testing should be appropriately expanded to show any potentially negative water evaporation is caught now.

ew3361zh (commented):
Looks good! Added a few comments, I agree with the changes requested from Ed and will approve once everything's addressed.

SC_redesign/Crop_and_Soil/crop/water_uptake.py
matthew7838: Honestly, I'm not sure either. I did not write that, let me look into it.

SC_redesign/Crop_and_Soil/field/field.py
matthew7838: Thanks.

SC_redesign/Crop_and_Soil/field/field.py
matthew7838: Sorry, this is just used to find the bugs, there's no actual use, so I'll delete it. Good catch.

SC_redesign/Crop_and_Soil/field/field.py
matthew7838: Good suggestion.

SC_redesign/Crop_and_Soil/field/field.py
matthew7838: Good suggestion.

SC_redesign/Crop_and_Soil/field/field.py
matthew7838: I think that the validation of min and max temperature will be something checked on a higher level when dealing with input files. There's not much to do with these value's sanity checks in this scope.

SC_redesign/Crop_and_Soil/crop/water_uptake.py
matthew7838: Now I looked into it. No, there's only 1 uptake for each layer.

SC_redesign/Crop_and_Soil/field/field.py
```diff
@@ -1153,18 +1153,22 @@ def _determine_potential_evapotranspiration(extra_terrestrial_radiation: float,
         Notes
         -----
         This method calculates the evapotranspirative demand for the entire field on any given day using the Hargreaves
-        method.
-
+        method.This method lower-bounds the potential evapotranspiration at 0.0 mm.
```
```suggestion
        method. This method lower-bounds the potential evapotranspiration at 0.0 mm.
```

Isgotkowitz (approved):
Looks good, thanks for making those changes Matthew

ew3361zh (approved):
LGTM!

Issue #668, opened 2023-07-18
Grab data from `InputManager` by property type


Isgotkowitz:
The way the Crop and Soil module is designed allows users to specify multiple fields to be simulated for a single simulation. This necessitates users being able to specify configurations for multiple different fields, and by extension the `InputManager` will need to be able to provide multiple field specifications to the Crop and Soil module.

**Proposed Solution**
When requesting data from the `InputManager`, there would be a method for doing so where the calling module only specifies the property of data that it wants, and the `InputManager` will return a list of data with all the data sets that meet the given property (it might be 0, 1, or many).

**Alternatives**
An alternative for storing the configurations of multiple fields would be to store all of them in only one file. The benefits are that there would only ever be one field blob in the metadata (making it simpler to retrieve the data from the `InputManager`), the downside is the the property section for this blob would be much more complicated (an array of objects) and thorough validation of it may not be feasible for V1.

**Additional context**
- If someone takes this issue but is not totally sure exactly how the feature should work, please reach out to Ed.
- There may need to be a similar feature implemented but instead of getting data by blob property, it is retrieved by specifying a file path. But that bridge should be crossed after the design for the field specification files has been more flushed out. (And again, if there is any clarification needed on this point do not hesitate to reach out to Ed!)

Isgotkowitz:
Update: after talking with Pooya, it seems that a feature will be implemented that allows modules to grab information from the *metadata* itself, and if implemented then the feature described in this issue will not be needed.

Edit: this feature is contained in #711

Issue #711, opened 2023-08-02
Create get_metadata function


ew3361zh:
This issue is to create a new funciton in the IM class called get_metadata() which will allow other parts of the RuFaS system to call and request metadata from IM. It will be very similar in structure to the IM get_data() function.

ew3361zh:
Addressed in #712.

PR #712, opened 2023-08-02
Input Manager Get Metadata Function

ew3361zh:
This PR addresses issue #711 by creating a new IM function, `get_metadata()` which allows the user to request data from the IM metadata dictionary.

This function looks in the metadata dictionary and returns the requested metadata to the user or raises a KeyError if the key is not found.

Unit testing and docstrings have been updated.


RUFAS/input_manager.py
```diff
@@ -446,3 +465,68 @@ def get_data(self, data_address: str) -> Any:
...
+        >>> input_manager = InputManager()
+
+        This will return the 'type' for albedo in the soil_profile_properties section of the metadata's properties:
+        >>> input_manager.get_metadata('properties.soil_profile_properties.albedo.type')
+        "number"
+
+        If a broader spectrum of the metadata is needed, the user can expand the query to get_metadata
+        by shortening the metadata_address. This will return the full 'albedo' object containing its type,
+        description, minimum, maximum, and default:
+        >>> input_manager.get_metadata('properties.soil_profile_properties.albedo')
```
Isgotkowitz: Will it be possible to call `input_manager.get_metadata('files')` and get access to all the data that is contained in all blobs? Based on this description I think it will be, but I want to double check because if so #668 can be closed with this PR

Isgotkowitz (approved):
Looks good Niko, only comment is a question specifying `get_metadata()`s capabilities which will determine if an extra issue can be closed by this PR

RUFAS/input_manager.py
ew3361zh: The way it's set up, entire sections of the IM's metadata would indeed be able to be retrieved at that level (so sending just 'files' or 'properties' as the `metadata_address` argument for this function would return those entire sections to the requester).

ew3361zh: Just added a test case where the metadata_address request is just 'properties' and it successfully returns the whole object.

Isgotkowitz: Awesome, I will add a comment noting this closes #668 then

Isgotkowitz:
Closes #668

RUFAS/input_manager.py
```diff
@@ -421,9 +421,28 @@ def get_data(self, data_address: str) -> Any:
...
         Examples
         -------
+        The user can request as broad or narrow a selection of the input data pool as is needed.
+
+        Input Manager must first be instantiated:
         >>> input_manager = InputManager()
-        >>> input_manager.get_data('animal.herd.calf_num')
-        This will return the value of `calf_num` of the `herd` section in the `animal` module.
+
+        This will return the value of `calf_num` of the `herd_information` section in the `animal` module:
```
PooyaHekmati: the keyword `module` is misleading. Let's use blob or section or something like that.

RUFAS/input_manager.py
```diff
@@ -421,9 +421,28 @@ def get_data(self, data_address: str) -> Any:
...
+        >>> input_manager.get_data('animal.herd_information.calf_num')
+        8
```
PooyaHekmati, line 432: 8?

RUFAS/input_manager.py
```diff
@@ -421,9 +421,28 @@ def get_data(self, data_address: str) -> Any:
...
+
+        If a broader spectrum of data is needed, the user can expand the query to get_data
```
PooyaHekmati: I don't think `spectrum` is a good word here. Range or domain are probably better.

RUFAS/input_manager.py
```diff
@@ -446,3 +465,68 @@ def get_data(self, data_address: str) -> Any:
...
+        The user can request as broad or narrow a selection of the metadata as is needed.
+
+        Input Manager must first be instantiated:
+        >>> input_manager = InputManager()
+
+        This will return the 'type' for albedo in the soil_profile_properties section of the metadata's properties:
+        >>> input_manager.get_metadata('properties.soil_profile_properties.albedo.type')
+        "number"
+
+        If a broader spectrum of the metadata is needed, the user can expand the query to get_metadata
```
PooyaHekmati: same as above

PooyaHekmati (approved):
Please address comments before mering.

RUFAS/input_manager.py
ew3361zh: Changed to blob.

RUFAS/input_manager.py
ew3361zh: 8 is the default number for calf_num so running `input_manager.get_data('animal.herd_information.calf_num')` would return 8 in this example case. I can make that clearer in the notes.

PR #716, opened 2023-08-04
Example Input Files with Defaults

ew3361zh:
This PR seeks to create new example input files that match the format of the metadata.


input/crop/example_rotation.json
```diff
@@ -0,0 +1,11 @@
...
+    "crop_species": "generic",
+    "harvest_days": [],
+    "harvest_years": [],
+    "harvest_operations": ["default"],
+    "harvest_type": ["scheduled"],
+    "planting_days": [],
+    "planting_years": [],
+    "pattern_repeat": 0,
+    "pattern_skip": 0
+}
```
Isgotkowitz: I think this is good as a default, it would be acceptable to pass
```json
{ }
```
as the crop input file though? It was acceptable with the old input system, but as I think about it, it seems like this would not be found to be valid by the InputManager.

input/fertilizer_schedule/example_fertilizer_schedule.json
```diff
@@ -0,0 +1,14 @@
...
+        "years": [],
+        "days": [],
+        "nitrogen_masses": [],
+        "phosphorus_masses": [],
+        "potassium_masses": [],
+        "application_depths": [],
+        "surface_remainder_fractions": [],
+        "pattern_repeat": 0,
+        "pattern_skip": 0
+    }
```
Isgotkowitz: I think I may have wrote the properties section of this one incorrectly. `"available_fertilizer_mixes"` should be at the same level as all the other variables, and in this example contain an empty dictionary.

Isgotkowitz (changes requested):
I think these files look good Niko, but I am still a little unsure about the benefit of having these "example" files. If they are to demonstrate what the newly formatted files should look like, that could be taken care of by just having the rewritten input files available. And likewise if their purpose is to provide default files in case a user does not specify a file for an input, couldn't one of the other prewritten input files provide that instead?

Also one of the example files needed for the Crop and Soil module is `example_field.json`.

ew3361zh:
> I think these files look good Niko, but I am still a little unsure about the benefit of having these "example" files. If they are to demonstrate what the newly formatted files should look like, that could be taken care of by just having the rewritten input files available. And likewise if their purpose is to provide default files in case a user does not specify a file for an input, couldn't one of the other prewritten input files provide that instead?

Good questions I'm not sure I know the complete answers to.

1. I do think they're going to be used as both defaults and as examples of the format these files should take should someone want to create their own to use and experiment with. Maybe the latter would be better served in the Wiki? Or maybe it's helpful to have the example files here so people can reference them locally and easily copy/paste the format to their new custom input file.

2. As far as rewriting the current input files - I wasn't sure how up-to-date those were with the C&S redesign. I could certainly take values available from those and use them here if it's the case that they represent the baseline/standard for a simulation. I can reach out to Kristan as we discussed in the worktogether meeting today.

3. My last thought is that we may have very similar sets of input files for a transition period because we still need the IM get_data functions throughout the model to get the data loaded and validated by the IM. And until that happens, I think we still need the existing input files in their current formats and locations to be able to run simulations.

> Also one of the example files needed for the Crop and Soil module is `example_field.json`.

Still working on that one.

input/fertilizer_schedule/example_fertilizer_schedule.json
ew3361zh: Thank you, will update that.

input/crop/example_rotation.json
ew3361zh: I don't think it would work to have an empty object here. IM will be looking for each of these fields to validate and will raise an error if they're missing.

Isgotkowitz:
> 1. I do think they're going to be used as both defaults and as examples of the format these files should take should someone want to create their own to use and experiment with. Maybe the latter would be better served in the Wiki? Or maybe it's helpful to have the example files here so people can reference them locally and easily copy/paste the format to their new custom input file.

I think having them locally would be good for those reasons. Would it be possible to name them as "default" files though, instead of "example"? It is not a big deal, but when I think about the two terms I imagine example being something that is just a demonstration and would not actually work, while default makes me think of something that is a standard, usable baseline. Also, I agree - having a section of the wiki dedicated to explaining and having examples of the input files would be really good to have, though they may not be complete inputs.

On the subject of names too, if a file is being used as the default I think it may still be good to give it a descriptive name if possible - for example the crop input file. Having an input file that specifies no crops to be planted in a simulation seems like the best candidate for the default, but I think it would still be better to name it as `no_crop_schedule.json` or something similar.

> 2. As far as rewriting the current input files - I wasn't sure how up-to-date those were with the C&S redesign. I could certainly take values available from those and use them here if it's the case that they represent the baseline/standard for a simulation. I can reach out to Kristan as we discussed in the worktogether meeting today.

The current input files have been rewritten to work with the redesigned Soil and Crop module (at least, the ones that are used in full simulations have been), but those changes have not been merged to master yet. For creating the baselines, there are some which are going to be a lot easier than others. For crops, tillage, fertilizer, and manure applications I think it is reasonable to make the baselines have no activity (no crops planted, no tilling occurs, etc.). For soil it is more difficult, because in the real world each field has its own unique soil profile. For a default though, it will probably be ok to just choose one of the soil profiles as the default because if someone is using the default soil profile then they probably don't care too much about what is happening in the soil profile. Definitely something that would be good to check in with Kristan about though.

> 3. My last thought is that we may have very similar sets of input files for a transition period because we still need the IM get_data functions throughout the model to get the data loaded and validated by the IM. And until that happens, I think we still need the existing input files in their current formats and locations to be able to run simulations.

This makes sense. I am not 100% sure what the schedule is for the InputManager integration but I think the full integration is far enough out that we will need the old-style inputs for a while longer.

input/field_management/Barnyard_field_1.json
```diff
@@ -0,0 +1,14 @@
+{
+    "soil_specification": "example_soil.json",
+    "crop_specification": "example_rotation.json",
+    "fertilizer_management_specification": "example_fertilizer_schedule.json",
+    "manure_management_specification": "example_manure_schedule.json",
+    "tillage_management_specification": "example_tillage_schedule.json",
+    "field_size": 0.0,
```
Isgotkowitz:
```suggestion
    "field_size": 1.0,
```

input/field_management/Barnyard_field_1.json
```diff
@@ -0,0 +1,14 @@
+{
+    "soil_specification": "example_soil.json",
+    "crop_specification": "example_rotation.json",
+    "fertilizer_management_specification": "example_fertilizer_schedule.json",
+    "manure_management_specification": "example_manure_schedule.json",
+    "tillage_management_specification": "example_tillage_schedule.json",
```
Isgotkowitz:
```suggestion
    "soil_specification": "soil_1",
    "crop_specification": "crop",
    "fertilizer_management_specification": "manure_schedule",
    "manure_management_specification": "fertilizer_schedule",
    "tillage_management_specification": "tillage_schedule",
```
These should be the name of the metadata blob that contains the corresponding specification. I had a conversation with Pooya before I merged the Crop and Soil metadata about this and he recommended using the blob names instead of file paths. This is reflected in the current example metadata on the `master` branch.

Isgotkowitz (approved):
Looks good. Only comments are on the example field file

allisterakun (approved):

PR #721, opened 2023-08-07
replace python multi-line string with Windows compatible string concatenation.

michaelRichards99:
Python's multi-line string appears to not properly encode window's `\r\n` for newlines, resulting in a unit test failing on windows platforms.

Tested on windows and ubuntu machines and passes on both.


Isgotkowitz (approved):
LGTM

ew3361zh (approved):
Works well!

Issue #699, opened 2023-07-28
[Animal] Update the methane model input


tomhuhh:
**Is your feature request related to a problem? Please describe.**
The methane estimate methods are grouped by animal category in the current code. For calves, there is the "Pattanaik" method. For heifers, there are the IPCC Tier method and the "Boadi and Wittenberg" method. For lactating cows, there are IPCC Tier 2, Mutian, and Mills methods. Lastly, for dry cows, there are IPCC Tier 2 and Mills methods. However, in the input json file, there is only one option for the methane estimate methods across all animal categories.

**Describe the solution you'd like**
In the data collection app, include an input for methane estimate method for each animal category. Code structure will need further adjustment as well. If we end up outputing methane estimates for all possible methane estimate methods across all animal categories, there may be no need to include methane estimate method as an input. So in that case, this input needs to be removed. 

**Describe alternatives you've considered**
NA

**Additional context**
This is a request beyond version 1.

tomhuhh:
@ltp23 Hello Loi, similarly, I assigned this issue to you only because I know you have been working on some of the stuff. Please feel free to assign to others if you think they are more suitable for the task. Thank you!

PooyaHekmati:
TODO: Follow up with author to check if this is still the case.

tomhuhh:
I can confirm that this issue has not been resolved yet. However, as far as the priority of the issue, I'll have to check with Kristan.

tomhuhh:
This issue can also be potentially addressed by the animal module refactoring led by @JoeWaddell in combination with #656.

ew3361zh:
From backlog grooming session:
@JoeWaddell what are your thoughts on the progress being made on this issue and whether it can be addressed in the animal module refactoring?

Isgotkowitz:
Blocked by #1680

Isgotkowitz:
Grooming session: @KFosterReed and @krb78, can you please give your opinions on whether this issue still needs to be implemented, and if so what the best approach would be?

krb78:
[...]

PR #722, opened 2023-08-08
Updates to metadata 

### LINKED ISSUE #1054: Enteric methane for lactating cows is reported as "0" when using 3-Mutian for methane_model
## Expected Behavior

When using either 3-Mutian or IPCC methane models, we expect there to be values for `AnimalManager.daily_updates.daily_aggregate_(calf|heifer|dry_cow|lactating_cow)_enteric_methane_g` and for `pen.calc_total_manure.pen_0_daily_enteric_methane_g` for all pens. 

## Current Behavior

When using methane_model = "3-Mutian", there are values reported for `AnimalManager.daily_updates.daily_aggregate_(calf|heifer|dry_cow)_enteric_methane_g` but 0's for `AnimalManager.daily_updates.daily_aggregate_lactating_cow_enteric_methane_g`

## Steps to Reproduce

1. Run defaults on main (default_animal.json has methane_model = IPCC). Look at values for "`AnimalManager.daily_updates.daily_aggregate_lactating_cow_enteric_methane_g`" and "`pen.calc_total_manure.pen_3_daily_enteric_methane_g`". They should be >0. 
2. Change methane_model in the input animal.json to "3-Mutian". Look at values for the same outputs. (They are 0's). 
3.
4.
5.

### LINKED ISSUE #1084: Animal Module pseudocode references
Peer-reviewed references for some values in the Animal Module are missing from the Animal LifeCycle pseudocode. A consolidated list of inputs or references should be developed, using [Manfei's paper](https://doi.org/10.3168/jds.2022-22396) and [Taylor's paper](https://doi.org/10.3390/ani11051373) as starting points.

This issue came up in discussions of adjusting the Jersey calf birth weight mean and SD, and identifying where these values came from. A [Google doc has been created](https://docs.google.com/document/d/1eUCWQCmCs7MnHdLoyOejf94JB69SWls7AJWEQH-WoOg/edit?usp=sharing) to begin filling in missing references or other information, and contains the reference information for the Jersey calf body weight values.

allisterakun:


JoeWaddell:
Just updated to main: most of the previous changes were already fixed. So this still remains a good place to add any necessary changes.

Also note: Before merging to main, we should copy the properties info in the `default_metadata.json` to all other metadata files, barring reasons not to for certain metadata files.

tomhuhh:
Missing metadata for `bunk_type` under 'storage_options'. I am on the `pilot_testing_Nov2023` branch. 

Screenshot 2023-11-14 at 2 08 43 PM

Thanks!

tomhuhh:
Also, in the same section, I am uncertain if the default values for `initial_dry_matter` makes sense to me. 

Screenshot 2023-11-14 at 2 15 35 PM

The data that I received from the farm are in % somewhere between 20 to 40%. However, I am not an expert in this area, so I would rely on a SME for clarification.  

Thanks!

input/metadata/default_metadata.json
```diff
@@ -187,55 +187,56 @@
         "description": "Herd Demographics",
         "calf_num": {
           "type": "number",
-          "description": "Number of Calves (head) -- The average number of pre-weaned calves (default is ~8% of herd number)",
+          "description": "Calves (head) -- The initial number of pre-weaned calves",
```
elle-andreen: All animal category values should be **initial** numbers?

input/metadata/default_metadata.json
```diff
@@ -187,55 +187,56 @@
...
         },
         "herd_init": {
           "type": "bool",
           "default": false,
-          "description": "Initialize Herd? -- Whether to re-initialize the animal database, where the herd animals are selected from."
+          "description": "Initialize Herd? -- Whether to re-initialize the animal database, where the herd animals are selected from"
         },
         "replace_num": {
           "type": "number",
+          "description": "Replacements (head) -- Number of replacement animals available in the replacement market",
```
elle-andreen, line 230: Correct definition for replace_num? And, this value does **not** need to be user-inputtable?

input/metadata/default_metadata.json
```diff
@@ -246,55 +247,55 @@
...
           },
           "cull_milk_production": {
             "type": "number",
-            "description": "Cull Milk Production (kg/d) -- Minimum daily milk production for do-not breed cow to be culled (default is 30 kg/d)",
+            "description": "Cull Milk Production (kg/d) -- Milk production threshold at which 'do not breed' cows are culled if they fall below",
             "default": 30,
             "minimum": 0
           },
           "cow_times_milked_per_day": {
             "type": "number",
```
elle-andreen, line 313: Should cow_times_milked_per_day be a pattern with discrete numbers (i.e. 1|2|3) or left as an integer to accomodate non-standard milking routines that may be added in the future? (e.g. seasonal, robotic)

input/metadata/default_metadata.json
```diff
@@ -526,43 +527,43 @@
...
-              "description": "Average Holstein Birth Weight (kg/head) -- The average birthweight for Holstein cows (default is 43.9)",
+              "description": "Average Holstein Birth Weight (kg/head) -- The average birthweight for Holstein cows",
               "default": 43.9,
               "minimum": 0
             },
             "birth_weight_std_ho": {
               "type": "number",
-              "description": "Holstein Birth Weight Standard Deviation (kg/head) -- The standard deviation birthweight for Holstein cows (default is 1)",
+              "description": "Holstein Birth Weight Standard Deviation (kg/head) -- The standard deviation of birthweight for Holstein cows",
               "default": 1,
```
elle-andreen: Is SD of 1 kg for birth weight is too low? Applies to Jersey breed as well

input/metadata/default_metadata.json
```diff
@@ -866,47 +867,47 @@
         },
         "3-NOP_additive_amount": {
           "type": "number",
-          "description": "3-NOP Additive Amount (mg/kg DMI) -- The dosage of the 3-NOP mitigation additive (mg/kg DMI)",
+          "description": "3-NOP Additive Amount (mg/kg DMI) -- The dosage of the 3-NOP mitigation additive",
           "minimum": 40,
```
elle-andreen: @Ed731551 @KFosterReed The methane mitigation design document states the usable range for 3-NOP is 37-137 mg/kg DMI, would you like the mix/maxes here expanded to those values rather than the recommended range? ([Similar to the discussion in issue 905](https://github.com/RuminantFarmSystems/MASM/issues/905)) If so, the values for monensin should be updated as well.

input/metadata/default_metadata.json
```diff
@@ -866,47 +867,47 @@
...
       },
       "housing": {
         "type": "string",
-        "description": "Housing -- The type of housing management for the herd.",
+        "description": "Housing Type (select one) -- The type of housing management for the herd",
         "default": "barn",
         "pattern": "^(barn|3-pasture|drylot)$"
       },
       "pasture_concentrate": {
-        "description": "Pasture Concentration -- Amount of concentrate supplementation provided if animals are housed on pasture. (kg)",
```
elle-andreen, line 948: @Ed731551 (I tagged you Ed because I think this value is utilized by the methane mitigation equations, but let me know if I'm incorrect) Are the units of pasture_concentrate kg/**head**/day or just kg/day offered to a group of animals?

input/metadata/default_metadata.json
```diff
@@ -866,47 +867,47 @@
...
+        "description": "Pasture Concentrate Fed (kg/head/day) -- Amount of concentrate provided to lactating cows each day if animals are housed on pasture",
         "type": "number",
         "minimum": 0,
         "default": 0
       },
       "methane_model": {
-        "description": "Methane Model -- The method used to estimate enteric methane emissions.",
+        "description": "Methane Model (select one) -- The method used to estimate enteric methane emissions for lactating cows",
```
elle-andreen: For methane_model, see Haowen’s issue [699](https://github.com/RuminantFarmSystems/MASM/issues/699) on model selection for estimating enteric methane emissions. It seems the model selected here would be applied for lactating cows, and potentially any classes where the same model is available, but I may be wrong there. E.g., if you select “Mills”, lactating and dry cow emissions will be computed according to Mills, but calves will be according to Pattanaik and growing heifers according to IPCC (their respective defaults)? Either way, assuming the intention is still to make model selection specifiable for each animal class _after_ V1, I think for now we can specify that the model here is for estimating enteric methane from lactating cows and give further specifics in documentation.

input/metadata/default_metadata.json
```diff
@@ -920,69 +921,69 @@
...
             "type": "number",
             "minimum": 0,
             "default": 0.1
           },
           "horizontal_dist_to_milking_parlor": {
-            "description": "Horizontal Distance to Milking Parlor (m) -- The distance (in meters) between the pen and the milking parlor.",
+            "description": "Horizontal Distance to Milking Parlor (m) -- The distance (in meters) between the pen and the milking parlor",
             "type": "number",
             "minimum": 0,
             "default": 1.6
```
elle-andreen: Default (1 meter/5.25 feet) for horizontal_dist_to_milking_parlor seems like too small of a default distance unless I'm missing something?

input/metadata/default_metadata.json
```diff
@@ -920,69 +921,69 @@
...
+            "description": "Number of Stalls -- The number of stalls in the pen",
             "type": "number",
             "minimum": 0,
             "default": 1000
           },
           "housing_type": {
-            "description": "Housing Type -- The type of barn the pen is in.",
+            "description": "Housing Type (select one) -- The type of barn the pen is in",
             "type": "string",
             "pattern": "^(open air barn)$",
```
elle-andreen: Housing_type: I know this may be a place-holder type variable for future implementations of other housing types (e.g. bedded-pack barn) but just want to highlight in case there is another housing type that should be present here.

input/metadata/default_metadata.json
Ed731551: Hi Elle, for typical dairy cattle rations, the recommended range for adding 3-NOP in the diet should be: 40 to 100 mg/kg DMI (please see Page 5) - https://docs.google.com/document/d/1Ot6sQGinCJiFazsqtsXXWebjzP4PdK3l/edit 

I would favor the safe use of 3-NOP as CH4 mitigation option within this recommended range. On the range you referred to, which is in the same page (Table 1), that comes from the descriptive statistics shown in the Kebreab et al. JDS_2023 paper. Evidence from literature discourage use of 3-NOP beyond 100 mg/kg DMI. Hope that helps. Thanks.

elle-andreen:
Sorry for the plethora of comments! In this last commit (animal_properties section), almost all variable definitions were edited, but the majority of changes were just standardizing the formatting, adding/moving units, removing (default is ___) instances, etc. Is seems like more than it really is. For any major changes or clarifications, I have added comments and/or tagged folks. I know it seems like there are a lot of non-critical (capitalization, etc.) changes but, because the metadata will be directly translated into the updated DCA, I though it was more efficient for everything to be as close to ideal as possible on the first pass rather than returning to correct minor things.

input/metadata/default_metadata.json
KFosterReed: This is correct. This is the hypothetical number of replacement animals available for purchase if the farm needs it.

input/metadata/default_metadata.json
KFosterReed: ohh, great question! I hadn't thought about unusual milking routines. I think for now leaving it as an integer is ok. I know that at some point in the not too distant future there will be interest in representing robotic herds and we might even have to think about a distribution for this value but for now, we can keep it simple

input/metadata/default_metadata.json
KFosterReed: Hmm, I didn't remember off the top of my head where this value came from but looking at Manfei's reference table, it was from this paper: [https://www.journalofdairyscience.org/article/S0022-0302(09)70573-9/fulltext](url). It looks like she may have taken the SD from the modeled value for an average live heifer calf instead of the actual birthweight distribution. I think we should use the value reported for all of the calves which would be an average BW of 42.9 and an SD of 6 (that is more variable than I would have expected! Have to go with the data though)

input/metadata/default_metadata.json
KFosterReed: This value was added in anticipation of adding a grazing model (which we haven't done yet) so it is not actually used... yet!

input/metadata/default_metadata.json
KFosterReed: Great plan and addition!

input/metadata/default_metadata.json
KFosterReed: Yes, it is definitely too small for pretty much anything but the smallest of small farms! I am not sure where this number came from but we will need to do some investigating to find a defensible default value. For now, we could just change it to some even and plausible number like ...  how does10 sound?

input/metadata/default_metadata.json
KFosterReed: yes, I think the alternative input should be pasture, which we don't have represented right now. Maybe it was removed so people don't think they have the option of grazing?

tomhuhh:
Noticed another issue with the metadata file, addressing it in this PR: one of the options for `'cow_presynch_protocol' `should be `'PreSynch' `instead of` 'Presynch'`. 

Screenshot 2023-12-28 at 5 35 24 PM

input/metadata/default_metadata.json
```diff
@@ -191,50 +191,51 @@
         "description": "Herd Demographics",
         "calf_num": {
           "type": "number",
-          "description": "Number of Calves (head) -- The average number of pre-weaned calves (default is ~8% of herd number)",
+          "description": "Calves (head) -- The initial number of pre-weaned calves",
```
tomhuhh: I would go with the original phrasing 'Number of Calves (head)' instead of 'Calves (head)' to improve clarity. Similarly for the number of animals in other categories.

input/metadata/default_metadata.json
```diff
@@ -191,50 +191,51 @@
...
           "default": 5000,
           "minimum": 0
         },
         "breed": {
           "type": "string",
           "default": "HO",
           "pattern": "(HO|JE)",
-          "description": "Breed (select one Holstein/Jersey) -- The predominant breed of the herd (Holstein or Jersey)"
+          "description": "Breed (HO/JE) -- The predominant breed of the herd (Holstein or Jersey)"
         }
```
tomhuhh, line 239: My suggestion is to use "HO" or "JE" or simply just remove the parentheses and the content in the parentheses instead of using just 'HO/JE' to improve consistency and clarity.

input/metadata/default_metadata.json
```diff
@@ -261,55 +262,55 @@
           "description": "General Management",
           "breeding_start_day_h": {
             "type": "number",
-            "description": "Breeding Start Day - Heifers (days) -- Days old when a heifer would be eligible for breeding (days)",
+            "description": "Breeding Start Day - Heifers (days) -- Days old when a heifer would be eligible for breeding",
             "default": 380,
```
tomhuhh, line 266: I would use 'Heifer Breeding Start Day (days)' instead of 'Breeding Start Day - Heifers (days)' to improve clarity.

input/metadata/default_metadata.json
```diff
@@ -261,55 +262,55 @@
...
             "default": "TAI",
             "pattern": "^(TAI|ED|ED-TAI)$"
           },
           "semen_type": {
             "type": "string",
             "default": "conventional",
             "pattern": "^(sexed|conventional)$",
-            "description": "The semen type that makes up a majority of the farm's breedings"
+            "description": "Semen type (sexed or conventional) -- The semen type that makes up a majority of the farm's breedings"
           },
```
tomhuhh, line 287: For consistency, should we keep the keys in the same order? By this I mean that the 'description' should be before 'default' and after 'type'. That said, I don't think the order will impact how the information is retrieved by the code.

tomhuhh: Also, I would write "Semen type (select one) -- The semen type ("sexed" or "conventional") that makes up a majority of the farm's breedings" instead.

input/metadata/default_metadata.json
```diff
@@ -365,54 +366,55 @@
...
                 "minimum": 0,
                 "maximum": 1
               },
               "estrus_conception_rate": {
                 "type": "number",
-                "description": "Estrus Conception Rate -- The % of inseminated animals that would be concepted in the ED (Estrus Detection) programs (default is 0.6)",
-                "default": 0.4,
+                "description": "Estrus Service Rate (percent) -- The percentage of animals detected in estrus that then would be serviced in the ED (Estrus Detection) and SynchED protocols",
+                "default": 1,
                 "minimum": 0,
```
tomhuhh, line 419: If this is the heifer estrus conception rate, I believe the default value should be '0.9' instead of '1'. 

Also, I do not think the description here is correct. I believe it should be the conception rate instead of the service rate. And this number should be for heifers only. We should specify that in the description as well. 

Lastly, the structure of the reproduction protocol metadata has significantly changed at least in the repro refactoring branches so it may require some rework in the future when those updates are merged.

input/metadata/default_metadata.json
```diff
@@ -365,54 +366,55 @@
             "description": "Repro Management",
             "voluntary_waiting_period": {
               "type": "number",
-              "description": "Voluntary Waiting Period (days) -- If 'TAI' (Timed Artificial Insemination) is chosen for cow, it is days after parturition when the 1st hormonal injection is made in TAI programs. If 'ED' or 'ED-TAI' is chosen for cow, it is days after parturition when estrus is started to be monitored.",
+              "description": "Voluntary Waiting Period (days) -- If TAI (Timed Artificial Insemination) protocol is chosen for cows, voluntary waiting period (VPW) is days after parturition at which the 1st hormonal injection is made. If ED (Estrus Detection) or ED-TAI (combination of Estrus Detection and TAI) is chosen for cows, VWP is days after parturition when estrus begins to be monitored.",
               "default": 50,
```
tomhuhh, line 371: This definition may change after the cow repro refactoring branch is merged.

input/metadata/default_metadata.json
```diff
@@ -490,77 +492,77 @@
...
                 "maximum": 1,
                 "default": 0.6
               },
               "cow_resynch_protocol": {
                 "type": "string",
-                "description": "Cow Resynch Protocol -- Resynch protocol for cow before/at/after pregnancy check day in the TAI (Timed Artificial Insemination) program",
+                "description": "Cow Resynch Protocol (select one) -- Resynch protocol for cows before/at/after pregnancy check day in the TAI (Timed Artificial Insemination) protocol",
                 "default": "TAIafterPD",
                 "pattern": "^(TAIafterPD|TAIbeforePD|PGFatPD)$"
               },
```
tomhuhh, line 563: We could also include what each option means in the 'description'.  Here is a description from Kaitlyn for pilot testing. 

'The resynch protocol for cow that is diagnosed open at pregnancy in the TAI program options: Start Ovsynch at Open Check (TAIafterPD), Start Ovsynch at Before Check (give GnRH one week prior = TAIbeforePD), Give lut only at Open Check (lut and heat detect, not TAI = PFGatPD)'

input/metadata/default_metadata.json
```diff
@@ -490,77 +492,77 @@
...
               "tai_program_start_day": {
                 "type": "number",
-                "description": "TAI program start day for ED-TAI protocol -- The TAI program start day for ED-TAI protocol.",
+                "description": "TAI protocol start day for ED-TAI protocol (day) --",
                 "minimum": 0,
```
tomhuhh, line 567: I don't think we need the '--' here.

input/metadata/default_metadata.json
```diff
@@ -571,43 +573,43 @@
             "description": "Bodyweight",
             "birth_weight_avg_ho": {
               "type": "number",
-              "description": "Average Holstein Birth Weight (kg/head) -- The average birthweight for Holstein cows (default is 43.9)",
-              "default": 43.9,
+              "description": "Average Holstein Birth Weight (kg/head) -- The average birthweight for Holstein cows",
+              "default": 42.9,
```
tomhuhh, line 578: I would just write 'The average birthweight for Holstein calves, kg/head'. 

Similarly for the terms below.

input/metadata/default_metadata.json
```diff
@@ -621,88 +623,88 @@
             "description": "Literature Repro Values",
             "preg_check_day_1": {
               "type": "number",
-              "description": "Pregnancy Check Day 1 (days) -- Days since last insemination to check whether the animal is pregnant or not for the first time (default is 32).",
+              "description": "Pregnancy Check Days, Check 1 (days) -- Days in pregnancy at the 1st pregnancy check",
               "default": 32,
```
tomhuhh, line 628: I think it should be the days since the last insemination based on what I understand from the code. The animal is not necessarily pregnant.

input/metadata/default_metadata.json
```diff
@@ -621,88 +623,88 @@
...
+              "description": "Pregnancy Loss Rate, Check 3 (percent) -- The percentage of cows that abort their pregnancy between the 2nd and 3rd pregnancy checks",
               "default": 0.017,
               "minimum": 0,
               "maximum": 0.1
             },
             "avg_estrus_cycle_return": {
               "type": "number",
-              "description": "Average days between calving and first estrus after calving (default is 23).",
+              "description": "Average Estruos Cycle Return (days) -- Average days between calving and first estruos cycle after calving",
               "minimum": 1,
```
tomhuhh, line 667: This might be a typo and should be 'estrus' or 'estrous' instead of 'estruos'. Similar typos can be found below.

input/metadata/default_metadata.json
```diff
@@ -794,27 +796,27 @@
             "description": "",
             "parity_death_prob": {
               "type": "array",
-              "description": "Death rate for first, second, third, and later lactations",
+              "description": "Death Probability, by Parity",
               "properties": {
                 "type": "number",
-                "description": "Death rate for first, second, third, and later lactations",
+                "description": "Death Probability, by Parity -- The probability of death for cows of a single parity (first lactation, second lactation, etc.); a separate entry should be included for 1st, 2nd, 3rd, and 4th+ parities (4 entries total)",
```
tomhuhh: I would use 'for a parity group' instead of 'for a single parity' because we grouped 3+ lactation animals into one group. 

Similar for the culling probability.

input/metadata/default_metadata.json
```diff
@@ -889,7 +891,7 @@
             "description": "",
             "still_birth_rate": {
               "type": "number",
-              "description": "Stillbirth rate",
+              "description": "Rate of stillbirths (percent)",
               "minimum": 0,
               "maximum": 1
```
tomhuhh: The default value is '0.065'?

input/metadata/default_metadata.json
```diff
@@ -965,69 +967,69 @@
...
+            "description": "Pen Type select one) -- The type of stalls the pen has",
             "type": "string",
-            "pattern": "^(freestall|tilestall|drylot)$",
+            "pattern": "^{freestall|tiestall|drylot}$",
             "default": "freestall"
           },
           "max_stocking_density": {
-            "description": "Max Stocking Density -- The maximum ration between the number of animals in the pen and the number of stalls in the pen.",
+            "description": "Max Stocking Density -- The maximum stocking density of the pen (ratio of the number of animals in the pen to the number of stalls in the pen)",
             "type": "number",
```
tomhuhh, line 1028: I would write 'The maximum stocking density of the pen  -- ratio of the number of animals in the pen to the number of stalls in the pen' instead.

input/metadata/default_metadata.json
```diff
@@ -2408,7 +2410,7 @@
         },
         "milk_reduction_maximum": {
           "type": "number",
-          "description": "Milk Reduction Percent -- Allowable amount of milk reduction (kg)",
+          "description": "Milk Reduction Maximum -- Allowable amount of milk reduction (kg)",
           "minimum": 0,
```
tomhuhh, line 2416: I would write 'Allowable amount of milk reduction (kg) when dietary nutrient supply cannot meet the requirements.'

tomhuhh (commented):
Great work! Thank you Elle! I left some comments for discussion.

input/metadata/default_metadata.json
elle-andreen: I think you’re right that it wouldn’t affect the way the code functions, but since the MD will be used for other things like updating the DCA, it’s probably better to be consistent just in case it matters for something else. Thanks for catching that!

input/metadata/default_metadata.json
elle-andreen: I think you’re correct that this is the wrong definition, this might have been from a conflict resolution error after the repro sections of the metadata were significantly updated. I have updated this definition accordingly.

input/metadata/default_metadata.json
elle-andreen: Just to confirm, do you know if for culling reasons are there are still four parity groups (1, 2, 3, 4+), or only 3 (1, 2, 3+)?

input/metadata/default_metadata.json
elle-andreen: I think the definitions of each option may be a bit too complex to fully include in the MD definition. They should definitely be present in the DCA and documentation though!

input/metadata/default_metadata.json
elle-andreen: You’re correct, technically they are not pregnant! Although we’d like to hope they are 😊

Isgotkowitz:
@elle-andreen should issue #905 get resolved with this PR?

tomhuhh (approved):
LGTM! Thanks for addressing those comments in a timely manner, Elle!

elle-andreen:
> @elle-andreen should issue #905 get resolved with this PR?

This PR will not resolve #905! Thank you for catching that - the intention originally was that this PR _would_ resolve that issue as part of the larger metadata review, but we elected to have this PR only cover the animal schema/module portions of the metadata for now. The feed module is the next one on my list to review; when that PR is opened, I can link it to issue #905 if that works for you. Or we can make a separate small PR to just fix the feed min/max ranges from #905.

input/metadata/default_metadata.json
```diff
@@ -621,88 +624,88 @@
...
               "type": "number",
-              "description": "Average length of estrus cycle for heifers (default is 21).",
+              "description": "Heifer Estrous Cycle Length (days) -- Average length of estrous cycle for heifers",
               "minimum": 1,
               "default": 21
             },
             "std_estrus_cycle_heifer": {
               "type": "number",
-              "description": "Standard deviation of the length of estrus cycle for heifers (default is 2.5).",
+              "description": "Heifer Estous Cycle Length Standard Deviation (days)",
```
JoeWaddell:
```suggestion
              "description": "Heifer Estrous Cycle Length Standard Deviation (days)",
```

JoeWaddell (approved):
Looks great! Just one mispelling suggestion.

input/metadata/default_metadata.json
```diff
@@ -191,50 +191,51 @@
...
           "type": "number",
-          "description": "Number of HeiferIs (head) -- HeiferI is the number of youngstock post-weaning and before breeding",
+          "description": "Number of HefierI's (head) -- The initial number of heifers that are weaned but not yet bred",
           "default": 44,
           "minimum": 0
         },
         "heiferII_num": {
           "type": "number",
-          "description": "Number of HeiferIIs (head) -- HeiferII is the number of heifers eligible for breeding and in early pregnancy",
+          "description": "Number of HefierII's (head) -- The initial number of heifers that are either eligible for breeding (based on user-inputted Breeding Start Day for heifers), or in early pregnancy",
```
jadamchick: typo: should be "Number of HeiferII's" :)

input/metadata/default_metadata.json
```diff
@@ -191,50 +191,51 @@
...
           "default": 38,
           "minimum": 0
         },
         "heiferIII_num_springers": {
           "type": "number",
-          "description": "Number of HeiferIIIs (head) -- HeiferIII is the number of close-up heifers",
+          "description": "Number of HefierIII's (head) -- The initial number of close-up heifers (heifers within the user-defined close-up period, i.e. Prefresh Day)",
```
jadamchick: That typo again! "Number of HeiferIII's"

input/metadata/default_metadata.json
```diff
@@ -261,55 +262,56 @@
           "description": "General Management",
           "breeding_start_day_h": {
             "type": "number",
-            "description": "Breeding Start Day - Heifers (days) -- Days old when a heifer would be eligible for breeding (days)",
+            "description": "Heifer Breeding Start Day (days) -- Days old when a heifer would be eligible for breeding",
```
jadamchick: While intuitive, that isn't what this variable means to RuFaS. I think we might be updating that at some point, but for now maybe we can update the metadata description instead to something like: "Days old when RuFaS initiates estrus protocol (first cycle to occur ~21 days later)"?

input/metadata/default_metadata.json
```diff
@@ -261,55 +262,56 @@
...
             "type": "number",
-            "description": "Cull Milk Production (kg/d) -- Minimum daily milk production for do-not breed cow to be culled (default is 30 kg/d)",
+            "description": "Cull Milk Production (kg/d) -- Milk production threshold at which 'do not breed' cows are culled if they fall below",
             "default": 30,
             "minimum": 0
           },
           "cow_times_milked_per_day": {
             "type": "number",
-            "description": "Number of Times Cows Milked (per day) -- The average or most common number of times the cows are milked, whole numbers 1-3",
+            "description": "Number of Milkings (per day) -- The average or most common number of times cows are milked per day (1, 2, or 3 times daily)",
```
jadamchick, line 314: What happens if you put a decimal here (eg 2.5)? Will it change to the default value?  That might be worth clarifying, whether this can be any value between 1 and 3, or if it needs to be the pattern of "1", "2", or "3"

input/metadata/default_metadata.json
```diff
@@ -365,54 +367,55 @@
             "description": "Repro Management",
             "voluntary_waiting_period": {
               "type": "number",
-              "description": "Voluntary Waiting Period (days) -- If 'TAI' (Timed Artificial Insemination) is chosen for cow, it is days after parturition when the 1st hormonal injection is made in TAI programs. If 'ED' or 'ED-TAI' is chosen for cow, it is days after parturition when estrus is started to be monitored.",
+              "description": "Voluntary Waiting Period (days) -- If TAI (Timed Artificial Insemination) protocol is chosen for cows, voluntary waiting period (VPW) is days after parturition at which the 1st hormonal injection is made. If ED (Estrus Detection) or ED-TAI (combination of Estrus Detection and TAI) is chosen for cows, VWP is days after parturition when estrus begins to be monitored.",
```
jadamchick: I would suggest: replace "when estrus begins to be monitored" with "when a cow is eligible to be bred during estrus"

input/metadata/default_metadata.json
jadamchick: A note on the defaults -- I think (for heifers) default Detection rate should be 0.9, and default Conception rate should be 0.6.

input/metadata/default_metadata.json
```diff
@@ -490,77 +493,77 @@
...
                 "type": "string",
-                "description": "Cow Resynch Protocol -- Resynch protocol for cow before/at/after pregnancy check day in the TAI (Timed Artificial Insemination) program",
+                "description": "Cow Resynch Protocol (select one) -- Resynch protocol for cows before/at/after pregnancy check day in the TAI (Timed Artificial Insemination) protocol",
                 "default": "TAIafterPD",
                 "pattern": "^(TAIafterPD|TAIbeforePD|PGFatPD)$"
               },
               "tai_program_start_day": {
                 "type": "number",
-                "description": "TAI program start day for ED-TAI protocol -- The TAI program start day for ED-TAI protocol.",
+                "description": "TAI protocol start day for ED-TAI protocol (day)",
```
jadamchick: It would be helpful to emphasize/clarify that this is start day for the Ovsynch portion of the TAI protocol (as long as that doesn't change with the cow repro refactor, and I don't think it will...).

jadamchick (approved):
Great work, Elle!  Just a heads up that the cow repro section will change when Loi merges the cow repro refactor (hopefully soon). Though of course you have to work with what you have in the present :)

input/metadata/default_metadata.json
elle-andreen: Good question! I had asked a similar question above and we want to leave things as integers for now, but you're correct that until we decide to accommodate non-standard milkings, this value should have a different pattern and/or type to avoid any decimals being used. I will ping someone from the dev team to help with that (I don't trust my grasp of types/patterns haha)
image

elle-andreen:
> Great work, Elle! Just a heads up that the cow repro section will change when Loi merges the cow repro refactor (hopefully soon). Though of course you have to work with what you have in the present :)

Thanks for looking it over Julie! I made a few further corrections based on your comments. Yes, we will have to rework things some once the repro refactor is complete, I started the MD review before any of those modifications hit the MD and before I knew the extent of the repro refactor! Hopefully the updates here will be usable or translatable into any new or modified sections after the repro refactor merges.

PR #726, opened 2023-08-09
100% test coverage for input and output managers.

michaelRichards99:
This PR brings the test coverage for `input_manager.py` and `output_manager.py` as well as their respective unit test files to `100%`.

Very slight modifications were made to the source files themselves in order to remove unreachable code / conditions and to remove occurrences of a single `\n` byte.

### DIFF
--- a/RUFAS/input_manager.py
+++ b/RUFAS/input_manager.py
@@ -233,7 +233,7 @@ class InputManager:
             variable_properties = reduce(lambda d, key: d[key], element_hierarchy,
                                          self.__metadata["properties"][properties_blob_key])
         except KeyError as e:
-            raise e
+            raise KeyError(f"{str(e)} not found in input data")
 
         var_type = variable_properties["type"]
         is_nested = var_type == "object"
--- a/RUFAS/input_manager.py
+++ b/RUFAS/input_manager.py
@@ -229,8 +229,12 @@ class InputManager:
                     }
         element_counter_and_validity = {"fixed_elements": 0, "total_elements": 0, "valid_elements": 0,
                                         "invalid_elements": 0, "is_valid": True}
-        variable_properties = reduce(lambda d, key: d[key], element_hierarchy,
-                                     self.__metadata["properties"][properties_blob_key])
+        try:
+            variable_properties = reduce(lambda d, key: d[key], element_hierarchy,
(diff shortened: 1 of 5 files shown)


RUFAS/input_manager.py
```diff
@@ -258,11 +262,7 @@ def _validate_element(self, element_hierarchy: List[str], properties_blob_key: s
             return element_counter_and_validity
         else:
             var_name = element_hierarchy[-1]
-            try:
-                input_data_value = reduce(lambda d, key: d[key], element_hierarchy, input_data)
-            except KeyError:
-                raise KeyError(f"Key {var_name} not found in input data")
```
michaelRichards99, line 264: This exception case was unreachable due to the same `reduce()` being called above.

RUFAS/input_manager.py
```diff
@@ -229,8 +229,12 @@ def _validate_element(self, element_hierarchy: List[str], properties_blob_key: s
                     }
         element_counter_and_validity = {"fixed_elements": 0, "total_elements": 0, "valid_elements": 0,
                                         "invalid_elements": 0, "is_valid": True}
-        variable_properties = reduce(lambda d, key: d[key], element_hierarchy,
-                                     self.__metadata["properties"][properties_blob_key])
+        try:
+            variable_properties = reduce(lambda d, key: d[key], element_hierarchy,
+                                         self.__metadata["properties"][properties_blob_key])
+        except KeyError as e:
+            raise e
```
Isgotkowitz: Would it be better/is it possible to display a more detailed error message here?

Isgotkowitz (approved):
Nice PR Michael, only comment is about possibly raising a more detailed error message but is up to you. Also have pulled this code to local and can confirm that everything passes on Windows

allisterakun (approved):
All the changes look good to me! 
I reviewed the IM-related changes, great changes and additions. I am not as familiar with the OM, but I read through the changes and they seem sensible to me.
Great PR!

RUFAS/input_manager.py
michaelRichards99: Done.

michaelRichards99:
> Nice PR Michael, only comment is about possibly raising a more detailed error message but is up to you. Also have pulled this code to local and can confirm that everything passes on Windows

Thanks for checking on windows.

PR #727, opened 2023-08-09
Fix formatting of arrays of objects in the example metadata.

Isgotkowitz:
This is a small PR that corrects some formatting mistakes of arrays of objects in the properties section of `example_metadata.json`.

I had written the crop schedule to be a single collection of values, when it should be multiple collections of the same set of values. In other words, I specified it to be a single object when it should have been an array of objects. Also, the way that the array of soil layer objects was being specified was inconsistent with the current standard (see Niko's comment below), so that has been corrected as well.


ew3361zh (changes requested):
Thanks, Ed, good catch on this. While I'm not sure what the end format will be, we're currently adding arrays of objects in this format*:

*apologies it's not keeping a readable format so I'm going to take a screenshot of the manure storage options array of objects setup which I think will be more helpful as a guideline:
Screenshot 2023-08-09 at 3 16 42 PM

This format may end up changing in the future when we figure out how to properly validate and fix array data but in the meantime I think having all the arrays of objects in the same format will be helpful.

ew3361zh (approved):
Looks great, thank you Ed!

allisterakun (approved):
LGTM

PR #730, opened 2023-08-10
IM get_data function expanded capabilities

ew3361zh:
This PR seeks to expand the IM's `get_data()` function to be able to return partial array data if requested.

For example:
```
__pool = {"cows": ["a", "b", "c", "d"]}
im = InputManager()
data_requested = im.get_data('cows.:3')
data_requested = ["a", "b", "c"]
```

To accomplish this, the PR:

1. Adds a new helper function `_get_array_data` which helps handle requests to `get_data()` for partial array data or single indexes from an array.
2. Updates unit testing.


Isgotkowitz (approved):
Very clean, looks good Niko.

PooyaHekmati:
Why do we need this? What is the use case? Why this can't be handled downstream?

ew3361zh:
> Why do we need this? What is the use case? Why this can't be handled downstream?

This was created in response to a request from @Isgotkowitz that you'd asked me to pursue, I believe. Ed, can you expand at all? I know it's been a while but do you recall?

Isgotkowitz:
> This was created in response to a request from @Isgotkowitz that you'd asked me to pursue, I believe. Ed, can you expand at all? I know it's been a while but do you recall?

@ew3361zh I'm sorry, I do not recall. There are some ways that I think the functionality of `get_data()` could be expanded, but off the top of my head I do not know any use cases in the code base where getting subarrays would be useful.

ew3361zh:
> > This was created in response to a request from @Isgotkowitz that you'd asked me to pursue, I believe. Ed, can you expand at all? I know it's been a while but do you recall?
> 
> @ew3361zh I'm sorry, I do not recall. There are some ways that I think the functionality of `get_data()` could be expanded, but off the top of my head I do not know any use cases in the code base where getting subarrays would be useful.

Well then...maybe I will just delete this branch since it seems like an unnecessary feature :). Thanks!

PR #731, opened 2023-08-13
Removing previous manure module code and connections to output handler

ltp23:
This PR helps remove the old manure module code. I can remove it now since the old `field` module has been replaced by the newer crop and soil module. Before, those two modules were connected.


allisterakun (approved):
Looks good to me!

varrma (approved):

PR #739, opened 2023-08-28
Deprecate OutputHandler

ew3361zh:
PR to deprecate and remove OH from model and ensure OM is suitably handling all output needs.
- [x] remove OH files
- [x] remove OH references
- [ ] check with SMEs on reporting they need from OH
- [ ] thorough file by file removal of all OH references 
- [x] ensure unit testing up to date

Tracking of reports needed both from Output Handler and new reports requested that don't currently exist in Output Handler: [https://docs.google.com/spreadsheets/d/1LIqEVpIl8zrL1zZEtAv0NCiwlHBbBIHw0ZQspXF_2LU/edit#gid=0](https://github.com/RuminantFarmSystems/MASM/pull/url)


PooyaHekmati (approved):
[LGTM](https://lgtm.lol/i/126)

KFosterReed:
I started to try to make some reports and although I haven't been successful with that yet (admittedly only had about 20 min to work on this) I did notice that the old output handler inputs are still in the json files and the input jsons are still there. Can/should we delete those now?

PooyaHekmati:
> I started to try to make some reports and although I haven't been successful with that yet (admittedly only had about 20 min to work on this) I did notice that the old output handler inputs are still in the JSON files and the input jsons are still there. Can/should we delete those now?
Thanks for catching this. I think we can and we should. 
[LGTM](https://i.imgflip.com/7xmk8k.jpg)

allisterakun (approved):
No issues for me!

Issue #679, opened 2023-07-20
Incorporate Input Manager into main.py


ew3361zh:
This issue seeks to find the best implementation strategy for how and where Input Manager should be called in main.py and what should happen depending on the result of the data processing pipeline in Input Manager.

# Solution 1:
There is a function in Input Manager `start_data_processing()` which returns a boolean of whether all the data from the input files is valid (meaning it was valid or has been fixed by the Input Manager).

This function receives 2 arguments from main.py 
- a string path to the input folder which contains the metadata file to be used to guide loading and validating input data.
- a boolean `eager_termination` which determines if the user wants to terminate data validation as soon as the first invalid critical data element is found (i.e. a data element that does not have a default value given by the metadata). Having eager_termination set as False would allow for a user to fully load all the data and have all invalid data logged to the OM pools.

<img width="809" alt="Screenshot 2023-07-20 at 4 12 11 PM" src="https://github.com/RuminantFarmSystems/MASM/assets/70217952/73a7f37f-3573-48d4-9c51-367bf09447ef">

If `start_data_processing()` returns False, the simulation will not run and Output Manager's `save_variables` and `dump_all_pools` methods will be called and the next input file will be sent to the Input Manager for the same validation process. 

`main.py`would manage this process from the method like this:

<img width="752" alt="Screenshot 2023-07-20 at 4 38 09 PM" src="https://github.com/RuminantFarmSystems/MASM/assets/70217952/e10d501b-fdf1-48fa-92c4-680ca70297ee">

**Describe alternatives you've considered**
It has been discussed on the dev-team where someone may want the simulation to continue depsite the Input Manager not being able to successfully validate and fix all the input data. This could be triggered using another flag the user could set as a gnu argument that could be checked in the `execute_simulations_from_files()` method in main.

ew3361zh:
Addressed in #810.

PR #740, opened 2023-08-30
Input files fix

ew3361zh:
This PR fixes some small formatting errors found when adding and initializing Input Manager in main.py to validate the input files.

- [x] fix input files
- [x] fix metadata structure
- [x] update unit testing
- [x] add IM to main.py
- [ ] remove print statement from main.py

Notes to reviewers:
1. I added a print message in main.py after IM is initialized that shows the result of the data being validated. Currently the validation works but you can adjust input file values for critical data to see validation fail. This will be removed before merging.
2. Currently the success of data validation is not linked to proceeding with the simulation but that is addressed in issue #679.


ew3361zh:
Changes to `example_animal.json`:
- Added parity_cull_prob in culling section
- Changed methane_mitigation_method default from null to "None"

Changes to `example_rotation.json`:
- Added key "crop_schedules" as the value to the array of a crop rotation

Changes to `example_feed.json`
- Removed excess commas

Changes to `example_metadata.json`:
- Updated all filepaths to start with `input/...`
- Updated example_emissions file to be a csv and for the type to be csv (json was incorrect).
- Updated field blob filepath to look in input/field_management/....
- Updated weather blob filepath to look for barnyard_weather.csv which is the correct filename.
- Updated all instances of type "boolean" to be type "bool" as IM is looking for.
- Removed initial type and description properties in animal_properties section as they were unnecessary.
- Updated all RegEx string patterns used for validation that were in the format of a choice (i.e. (A|B|C)) to have parantheses instead of curly brackets which RegEx couldn't parse.
- Move location of life_cycle object out a level to match the pattern in the animal input file.

allisterakun (approved):

Isgotkowitz (approved):

PR #742, opened 2023-08-31
Im integration other

allisterakun:
This PR changes the input source of `classes.py`, `main.py`, and `simulation_engine.py` from separate input files to using the Input Manager.

### `main.py`
Moved `input_manager.start_data_processing()` into a for loop, which iterates through all the files under the input folder and passes them in as input metadata. 
Also, before `input_manager.start_data_processing()`, ``input_manager.flush_pool()` is called to ensure no confusion with the previous simulation.
#### `State` class
Replaced all `Utility.read_json_file()` with `im.get_data()`, so it is getting all the data from InputManager rather than directly reading JSON files
#### `Config` class
Removed the process of reading weather data CSV file, receiving the weather data as input instead.
Removed the validation of the weather data dates, checking that all the simulation dates have corresponding weather data. This validation process should not be happening here but rather in the InputManager (a future TODO functionality; this would be a cross-validation case).
#### `Weather` class
Removed weather data validation and simplified the weather data pre-processing (formatting)
### `simulation_engine.py`
Replaced all `Utility.read_json_file()` with im.get_data()

### `example_metadata.json`
Rolled back the metadata for manure management to an older version due to the complexity of adapting to the new metadata format. @ltp23 also suggested that the new metadata format is too complex for the manure module to implement.


RUFAS/classes.py
```diff
@@ -48,6 +47,7 @@ def __init__(self, data, config, weather, time):
             time: instance of the Time class containing information necessary to
                 initialize the state
         """
+        data = {}
```
ew3361zh: We can probably get rid of this once we merge in #744

RUFAS/classes.py
```diff
@@ -97,167 +97,22 @@ def __init__(self, data, weather_file):
         self.end_day = int(self.end_full_date[1])
 
         # set seed attributes
-        self.set_seed = data['set_seed']
-        self.seed = data['seed']
+        self.random_seed = data['random_seed']
```
ew3361zh, line 103: So I think if I follow correctly the new example_config.json file contains the `random_seed` setting and this would get loaded into `data_config` in simulation_engine.py which then gets added to create self.config. So I think this would be available from `config` argument and should be:
`self.random_seed = config[random_seed]`

allisterakun: Will update!

RUFAS/classes.py
allisterakun: Yes!

RUFAS/classes.py
allisterakun: This is the Config class __init__ function. So it is setting the random_seed for the config

RUFAS/classes.py
```diff
@@ -98,166 +98,18 @@ def __init__(self, data, weather_file):
...
-        # years and mid-year start/end dates
-        # each year in years starts with the calendar date at index zero and then
-        # is filled with the days of the year that the program will run
-        # the start and end dates will be specified by the user
-        # years is used as a correctly size template for each of the weather arrays
-        # in Weather()
+        self.w_start_year = self.start_year
+        self.w_start_day = self.start_day
+        self.w_end_year = self.end_year
+        self.w_end_day = self.end_day
```
Isgotkowitz, line 113: I think renaming these vars to `weather_start_year` etc would be good for clarifying the code. However, if that would require changing references to these variables in a lot of other places, then maybe a better solution would be to just add a comment specifying that weather start and end dates are being set here.

Isgotkowitz (approved):
Looks good, one comment on code clarity but may be outside scope of this PR. Also, is the changed functionality of `Weather` and `SimulationEngine` being tested?

RUFAS/input_manager.py
```diff
@@ -648,3 +648,13 @@ def get_metadata(self, metadata_address: str) -> Any:
...
+
+    def flush_pools(self) -> None:
+        """
+        Clear the variable pool.
+        """
+        info_map = {"class": self.__class__.__name__,
+                    "function": self.flush_pools.__name__,
+                    }
+        self.__pool = {}
+        om.add_log("Clear variable pool", f"The pool is emptied.", info_map)
```
```suggestion
        om.add_log("Clear variable pool", "The pool is emptied.", info_map)
```

PR #744, opened 2023-09-01
IM integration - Feed, Animal Manager, Manure Manager

ew3361zh:
This PR changes the source of the data used by the Feed class, the Animal Manager class, and the Manure Manager class from separate input files to the Input Manager Pool.


ew3361zh:
Changes to `example_animal.json` input file
Added because model was looking for them:
- replace_num
- injury_cull_prob
- death_cull_prob
- cull_day_count
- udder_cull_prob
- unknown_cull_prob
- feet_leg_cull_prob
- disease_cull_prob
- mastitis_cull_prob
- methane_mitigation_additive_amount

`pen_information` changed to object of pen objects with `penX` as key for each pen object within the overall pen_information object.

Changes to `example_feed.json` input file:
purchased_feeds object changed to match previous purchased feeds object in `purchased_feeds.json`.
Added because model was looking for them:
- feed_database
- feeds_table
- feed_quality_table
- nutrient_table
Added in storage options section:
- storage_type
- moisture
- bunk_type

Changes to `example_manure_management.json`:
- Left as is but model is not ready to use this updated document and not sure where all the new variables should be applied/used.
- Updated metadata to use previous `manure_management.json` file that model currently relies on.

ew3361zh:
Changes to `example_metadata.json`:
- changed manure_management blob to look for current manure_management.json file as source of data instead of updated example_manure_management.json document which the model is not prepared to use as intended.
- Updated animal_properties section to validate all above-listed culling probabilities, feed database table information.
- Updated manure_management_properties section to validate current manure_management.json file data.

RUFAS/routines/feed/feed.py
```diff
@@ -67,10 +67,13 @@ def __init__(self, data: PurchasedFeedTypedDict):
         self.feed_quality_table = data['feed_quality_table']
         self.nutrient_table = data['nutrient_table']
         self.db_reader = DatabaseReader(self.feed_database)
+        purchased_feeds_list = [feed_item["purchased_feed"] for feed_item in data["purchased_feeds"]]
+        purchased_feed_costs = {str(feed_item["purchased_feed"]): feed_item["purchased_feed_cost"]
```
JoeWaddell, line 71: This is great: will be necessary in the user-defined-ration branches when we merge those to master, so seeing this technique is helpful.

JoeWaddell (commented):
This is looking nice! After a good discussion with @ew3361zh , I'll explain a little bit about how I (and a few others) use RuFaS, to help guide IM development.

From the code (and discussion), the current point of entry to RuFaS is a json, which soon(tm) will be in the format of `example_metadata.json`. For the user, this will still function like `animal_management.json` as a point of entry, since it both has config information, but also acts as a pointer to other files with more detailed config information.

The way I typically use RuFaS is to have a number of input jsons, both the "main entry" json and animal and/or feed configs (in most of my simulations). The "main" input json helps me switch between scenarios/settings that I want to test particular functionality on, e.g. very different feed compositions to evaluate how ration formulation is working (unique feed JSONs).  My understanding is that the way IM works will not change that workflow: it just means I'll instead have multiple metadata files as the point of entry, and the various animal/feed/manure metadata files as the respective JSONs. 


What this might mean in practice (and how I've seen a couple others use it), is that some combination of files make up a given "farm", which would be the name of the entry point. 

- So for example: "Farm A" might use all the same input metadata JSONs as "Farm B", but with the exception of "Feed_northeast.json" instead of "Feed_southeast.json". 

- If this is correct, the main entry JSON would be identical except for `Farm_A_metadata.json.files.feed` pointing to different files. It just means duplicating a metadata json, then changing some fields (the same process we use now). 

- Then, Farm C may have a whole series of changes, including (perhaps) different minx/maxes for some of the inputs in the `properties` section of the metadata file, because they are running some extreme experiments (outside of the normal bounds of simulations).

Assuming the above assumptions are correct, my question becomes: should the properties  section of the metadata be a separate file, if those are not frequently modified? This means the workflow for a user would be very similar to the current implementation, but with more descriptions/titles that can be filled out, and splitting the general config into an entirely separate file. 
- Or is the idea that the typical user will never deal much with the metadata, because it will be the backend to some other user interface?

RUFAS/classes.py
```diff
@@ -245,8 +245,8 @@ def __init__(self, data, weather_file):
         # checks that start date is not after end date
         if self.start_year > self.end_year \
                 or (self.start_year == self.end_year and self.start_day > self.end_day):
-            # TODO determine if this below raised error should be rewritten to reassign 
-            # either the the start or end year to then allow the simulation to continue 
+            # TODO determine if this below raised error should be rewritten to reassign
+            # either the the start or end year to then allow the simulation to continue
```
Isgotkowitz: Is this an old TODO?

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -132,7 +132,7 @@ def __init__(self, args):
         self.wood_m = 0
         self.wood_n = 0
 
-        self.lactation_curve = AnimalBase.config['lactation_curve']
+        self.lactation_curve = 'wood'
```
Isgotkowitz, line 135: Is there any need to document somewhere that users are not currently able to set the lactation curve?

Isgotkowitz (approved):
Looks good to me. Had two questions about minor details, nothing important

allisterakun (approved):

input/example_metadata.json
```diff
@@ -1898,435 +1976,161 @@
       }
     },
     "manure_management_properties": {
-      "manure_management_scenario": {
+      "manure_management_scenarios": {
```
allisterakun, line 1973: @ltp23 Here is where we reverted back to the old input data

PR #746, opened 2023-09-01
Update README.md

PooyaHekmati:
Add priority level definitions to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,15 @@ When decision-making is difficult and doubtful, we use these as our North Star t
 ## Past
 N/A
 
+# Definitions for Priority Levels
+- **P0 (Priority 0):** These are the 🔥 "fire-breathing dragon" tasks! They are critical and need immediate attention. Team members should don their hero capes 🦸‍♂️🦸‍♀️ and charge into action on P0 tasks. However, if someone finds themselves in a knightly standstill while battling these dragons 🐉, they can embark on a noble quest to tackle lower-priority tasks while working to rescue the damsel in distress (unblocking the higher priority task).
+
+- **P1 (Priority 1):** Think of these as the trusty steed tasks 🐎. They're important and require attention once the dragons (P0) are vanquished. Knights of the project realm should prioritize P0 tasks but, if they encounter a moat or a drawbridge on their way, they can hop on their trusty steed 🏇 and joust with P1 tasks, all while working on strategies to lower the drawbridge (resolve blockers).
+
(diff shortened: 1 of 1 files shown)


JoeWaddell (approved):
My immediate enthusiastic approval of this PR is in no way related to the fact that I'm currently re-reading the Once and Future King for about the tenth time. Also not related to "Priority One" messages being a thing in Star Trek.

On a more serious note: are we considering these levels as roughly exponential? As in the jump from P0 from P1 is a bigger step than P2 to P1? 

Also: should we have a name like this for for non-priority tasks, so that people get in the habit of labeling things every time they open an issue/PR? Just thinking about Issues or Project tasks that aren't _yet_ a priority, but may be a Future Priority, or even a Non-Priority task, though I can see the argument that those are still priorities, just low-ranked.

KFosterReed (approved):
I regret taking so long to read these!

KFosterReed:
I also like Joe's suggestion for a future priority category as well :)

PooyaHekmati:
> On a more serious note: are we considering these levels as roughly exponential? As in the jump from P0 from P1 is a bigger step than P2 to P1?

It should be that way, but it usually isn't :D 

> Also: should we have a name like this for for non-priority tasks, so that people get in the habit of labeling things every time they open an issue/PR? Just thinking about Issues or Project tasks that aren't _yet_ a priority, but may be a Future Priority, or even a Non-Priority task, though I can see the argument that those are still priorities, just low-ranked.

Let's chat about this so I can have a better understanding of what you are thinking about. If my current understanding of your comment is correct, a typical labeling system would be: 

1. wish
2. low
3. mid
4. high
5. UBN (unbreak now!)

JoeWaddell:
> It should be that way, but it usually isn't :D

I'll be sure to remind folks in the Animal Module that when we get the real fire-breathing dragon tasks, that we need to drop what we're doing (unless it's slaying a different dragon) to help out. I've worked in a lab before where *everything* was labeled as a critical task, which means that *nothing* was treated a critical task - it became a challenge to sort out what's *actually* important. I think keeping that P0 in reserve for when it's really needed is a good choice.

> Let's chat about this so I can have a better understanding of what you are thinking about. If my current understanding of your comment is correct, a typical labeling system would be:
> 
>     1. wish
> 
>     2. low
> 
>     3. mid
> 
>     4. high
> 
>     5. UBN (unbreak now!)

We can definitely chat about this, but to put my thoughts into words: I think what you've merged to master is great. Instead of adding two more levels, I'd just think about adding a wish/future label. That way we retain the minimalist nuance of low/high priority for P1 and P2, and we reserve P0 for the critically time-sensitive tasks.

PR #748, opened 2023-09-05
Updating nitrogen equations for compost bedded pack

ltp23:
This PR improves on the existing code for calculating nitrogen losses for the compost bedded pack system by reorganizing the hard-coded constants, adding more test cases and error-checking for negative nitrogen values.

### DIFF
--- a/RUFAS/routines/manure/gas_emissions/gas_emissions.py
+++ b/RUFAS/routines/manure/gas_emissions/gas_emissions.py
@@ -1307,7 +1307,7 @@ class GasEmissions:
         return total_carbon_decomposition
 
     @staticmethod
-    def _calc_nitrogen_loss_from_ammonia_emission(
+    def _calc_nitrogen_loss_in_compost_bedded_pack_barn_due_to_ammonia_emission(
             daily_nitrogen_input: float,
             is_bedding_tilled: bool
     ) -> float:
@@ -1345,7 +1345,7 @@ class GasEmissions:
         return nitrogen_loss_tilled + nitrogen_loss_untilled
 
     @staticmethod
-    def _calc_nitrogen_loss_to_leaching(daily_nitrogen_input: float) -> float:
+    def _calc_nitrogen_loss_in_compost_bedded_pack_barn_due_to_leaching(daily_nitrogen_input: float) -> float:
         """
         Calculate the mass of nitrogen that leaches out of the compost bedded pack barn's manure-bedding mixture.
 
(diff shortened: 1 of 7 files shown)


RUFAS/routines/manure/gas_emissions/gas_emissions.py
Isgotkowitz, line 1: It seems that most of the equations modified in this file are for calculating the gas emissions when compost bedded pack is being used. Are there other equations for computing emissions when non-CBPB bedding is used? If so, it may be good to indicate these things in the names of the methods, so it is extra clear which equations should be used for each bedding type.

RUFAS/routines/manure/manure_treatments/manure_treatment_cbpb.py
```diff
@@ -66,14 +67,26 @@ def _calc_bedding_potassium_content(
...
 
         """
-        daily_output = self._initialize_daily_output_during_update(self._current_manure_treatment_daily_input)
+        daily_input = self._current_manure_treatment_daily_input
+        total_nitrogen_loss = GasEmissions.calc_total_nitrogen_loss_from_compost_bedded_pack_barn(
+            daily_nitrogen_input=daily_input.liquid_manure_nitrogen,
+            is_bedding_tilled=True
+        )
+        manure_nitrogen = daily_input.liquid_manure_nitrogen - total_nitrogen_loss
+        # TODO: To be implemented further later.
```
Isgotkowitz, line 86: Dangling TODO

tests/test_manure_module/test_gas_emissions.py
Isgotkowitz, line 1: These test cases look great, but I think that "Act and Assert" and "Arrange" in-line comments are unnecessary.

Isgotkowitz (approved):
Code looks great, 1 comment about the naming of methods specific to CBPB treatment method, 2 comments on other comments in the code.

RUFAS/routines/manure/gas_emissions/gas_emissions.py
ltp23: That sounds good. I will update the names of those functions to be more specific.

RUFAS/routines/manure/manure_treatments/manure_treatment_cbpb.py
ltp23: Yes, I need to update this method incrementally.

tests/test_manure_module/test_gas_emissions.py
ltp23: Those are entirely optional. I keep those to remind myself of the structure of each test.

varrma (approved):

PR #756, opened 2023-09-07
Refactor Field module to use `InputManager`

Isgotkowitz:
This PR integrates the `InputManager` into the `FieldManager` module, completely deprecating the old input system in the `Field` module. It also refactors some of the inputs, to make them compatible with the metadata properties.

Code to look at closely:
- `field_manager.py`:
  - `init` now calls the appropriate methods to setup fields with data pulled from the IM
  - All methods responsible for setting up Field or components of Field pull data from IM.
  - The method to pull data for field management (fertilizer, manure, and tillage applications) has been split up into three methods, to reflect the new structure of data as specified in the metadata properties section.
- `test_field_manager.py`: tests all the changes made to `FieldManager`, maintains 100% code coverage. 

Code to look at not as closely:
- `classes.py`: updates the initialization of the `FieldManager`.
- `field_data.py`: adds default longitude.

***Testing***: RuFaS should run fully when used with `example_metadata.json`, `ARL_metadata.json`, and `barnyard_metadata.json`. Please try them on your machine.

Inputs:
- The soil properties section of `example_metadata.json` has been rewritten - it is compatible with the existing soil profiles. This is because converting the soil input files to the new metadata would have been a very intensive task, but the inputs and properties will eventually be updated to include all values that we want to incorporate in the soil module.
- All of the field management files have been written into new files that are consistent with the corresponding properties in metadata.
- `no_animal.json` has been added so that it is easier to run simulations with no animals.
- `purchased_feed.json` has been rewritten to follow the corresponding properties.
- The example soil files have been rewritten to follow the soil properties.
- `ARL_metadata.json` and `barnyard_metadata.json` have been written, and should run full simulations successfully.
- `example_metadata.json` has had the file references for NRC and NASEM comp changed so that they are correct.


allisterakun (approved):
Runs without error!

RUFAS/routines/field/manager/field_manager.py
```diff
@@ -145,15 +149,42 @@ def _create_current_weather(weather, time, month: int) -> CurrentWeather:
...
+        field_blob_names = []
+
+        try:
+            blobs = im.get_metadata("files")
+        except KeyError:
+            raise KeyError("Could not find 'files' section of metadata.")
+
+        for blob_name, blob_values in blobs.items():
+            try:
+                property = blob_values["properties"]
```
ltp23: Maybe using a different name here because property is a built-in name.

RUFAS/routines/field/manager/field_manager.py
```diff
@@ -145,15 +149,42 @@ def _create_current_weather(weather, time, month: int) -> CurrentWeather:
...
+        Returns
+        -------
+        List[str]
+            List of blob names that contain field configurations.
+
+        """
+        field_blob_names = []
+
+        try:
+            blobs = im.get_metadata("files")
```
ltp23, line 165: I'm wondering if there is a need to store this data for later use.

ltp23 (approved):
The code looks great to me. Very thorough. Great job, Ed!

One potentially helpful note on simplifying the instanstiation of the schedule objects: Because the parameter names and the key names from data dictionaries match for the most parts, you can take advantage the hidden attribute `__annotations__` on any function, which will give you info about function signature, and then you can work from there.

RUFAS/routines/field/manager/field_manager.py
Isgotkowitz: Good call

RUFAS/routines/field/manager/field_manager.py
Isgotkowitz: Do you think it would be stored in the `InputManager` somewhere?

Issue #755, opened 2023-09-07
Provide default user-defined ration for testing purposes


JoeWaddell:
We've merged the user-defined ration branch to master! However: only *after* the merge (and removing the modified feed sqlite database) did I realize that the user-defined ration example in the Feed json provided includes a feed item (158) that was a custom user input. 

@tomhuhh is going to provide a breakdown of that item (it was composed of multiple already-existing ingredients), but any other formulated ration would also be appropriate, as a 'typical' Northeast diet/ratio for testing.

Equally valid is removing the example feed JSON, and putting those percentages in `purchased_feed.json`, so users can more rapidly test the automated v. user-defined ration formulation methods.

Issue #577, opened 2023-06-15
Heifer ration formulation 'reduction' method


JoeWaddell:
Currently, if a heifer ration cannot be formulated, it takes the previous month's ration instead. Of course, this cannot be done on the first formulation.

As such, in cases of poor feed ingredients, a formulation might *never* be possible, in which case it fails entirely.

Furthermore, there is currently no way to reduce requirements (as there is for lactating cows, see the milk production reduction penalty, like in #462). 

Consideration should be given to penalizing growth rate or other parameters to allow for poor outcomes.

One implementation for the user-defined ration method is implemented in a commit in #383, though there is not yet an implementation for the non-user-defined method. 

Likewise, we will need to add temperature to the input arguments when recalculating requirements for heifers in animal_requirements.py (see method `recalculate_requirements`).

PR #758, opened 2023-09-09
Added two new diets

tomhuhh:
The NE diet is based on one of the diets from Edward; the W diet is based on the UDR test google sheet.


tomhuhh:
I tested with these two diets and they work fine. However, in the output, it shows that the requirements below are not met. 
- "NEmact_constraint",
- "NDF_constraint_upper"

JoeWaddell:
> I tested with these two diets and they work fine. However, in the output, it shows that the requirements below are not met.
> 
>     * "NEmact_constraint",
> 
>     * "NDF_constraint_upper"

After testing I see this behavior is the same for both diets.

For the failed constraints for pen2, reviewing the user-defined values, they don't total 100%! That's why it's giving the NEmact and NDF failures.
Pen 2 fails but uses the default ration because there are no methods to reduce requirements for animals that aren't lactating. See Issue #577 

Pen 3 fails for `total_energy` just once: so after one reduction in lactation, it passes. This single reduction (which doesn't greatly affect overall production) may be alleviated when we have the NASEM nutrient values in RuFaS, as these may be assuming slightly different files.

tomhuhh:
Thanks for catching that Joe! If you are okay with these two diets, we can merge them into Masters. Thanks!

JoeWaddell:
I've updated the dry cow ration so that hay forage makes up the rest of the ration (since it previously totaled less than 100%), and I've deleted the Western Diet (I have a local copy), so that we keep the number of diets on master relatively low (following discussion with @KFosterReed ).

JoeWaddell:
Since output handler has now been deprecated, I've also removed the `quick_ration_analysis.py` script in this same PR, since it was dependent on the old structure of ration CSVs. I have a local copy to modify/reinstate once CSVs are being printed again.

JoeWaddell (approved):
Harder to evaluate after deprecation of output handler and lack of CSVs, but relevant info is in the output manager, and this seems to be a good "default" for testing/development purposes. Should close https://github.com/RuminantFarmSystems/MASM/issues/755

KFosterReed (approved):

PR #765, opened 2023-09-13
Adding ration per animal to om

JoeWaddell:
Following deprecation of output handler, feed information is no longer being stored

This quick PR simply adds the number of animals in each pen to the info_map (necessary to calculate total feed ingredients fed per pen), and adds the ration per animal itself to om.

This also removes extraneous info from om (concentrations, when those are a product of DMI, and DMI, which is a sum of ration_per_animal)


allisterakun (approved):

ltp23 (approved):
This looks simple enough. Good job, Joe! Thanks.

Just want to double-check with you that the output data flows one way from the simulation to the output manager, so we won't have a way to access the data later within the simulation.

PR #768, opened 2023-09-14
Adding more output data from manure module to output manager

ltp23:
In this PR I want to add the following 7 groups of output variables to the output manager:

- `AnimalModuleInputToManureModule_Pen_<pen_id>.<variable_name>`: List the variables that have been passed to the manure module from the animal module but have been processed somewhat. 
- `ManureHandlerDailyOutput_Pen_<pen_id>.<variable_name>`: List the variables associated with manure handling.
- `ReceptionPitDailyOutput_Pen_<pen_id>.<variable_name>`: List the variables associated with reception pits.
- `AnaerobicDigestion_ManureTreatmentDailyOutput_Pen_<pen_id>.<variable_name>` (optional): List the variables associated with the use of an anaerobic digester.
- `ManureSeparatorDailyOutput_Pen_<pen_id>.<variable_name>` (optional): List the variables associated with the use of manure separators.
- `ManureTreatmentDailyOutput_Pen_<pen_id>.<variable_name>`: List the variables associated with the manure treatment process.
- `Accumulated_ManureTreatmentDailyOutput_Pen_<pen_id>.<variable_name>`: List the variables associated with manure treatment but they have been added up over time.


RUFAS/routines/manure/manure_handlers/milking_parlor.py
```diff
@@ -32,25 +29,14 @@ def __init__(self,
...
-                    "wash_water_use_rate": wash_water_use_rate,
-                    }
 
         self.num_milkings = num_milkings
         self.minutes_spent_in_holding_area = minutes_spent_in_holding_area
         self.minutes_spent_per_milking = minutes_spent_per_milking
         self.wash_water_use_rate = wash_water_use_rate
         self.fresh_water_use_rate = fresh_water_use_rate
 
-        om.add_variable("fresh_water_use_rate",
```
JoeWaddell, line 51: It looks like this - and many other variables in this file - were removed from the output manager but not replaced: was this intentional?

ltp23: Thanks Joe. I'm not sure which variables to report from this file so I will revisit it later. Though, I think water usage is relevant. Also, they should probably be reported from some place else.

varrma (approved):
Looks good Loi with the new names

KFosterReed:
Questions about the overall goal for the manure module outputs: 
1. Is the manure module still using the old manure output handler? If the goal is to transition the old manure output handler to add variables to the OM then would it be a good idea to change the name to manure output manager instead of output handler?
2. When I run the `animal_managment.json` in this branch, I am still getting a separate director within the output -> CSVs directory called `manure_module_outputs` even though there is not a csv filter for these output files. Is the goal to keep an option for producing these separate directories?

Any help to clarify this for me is much appreciated!

ltp23:
Thanks Kristan for the questions.

> 1. Is the manure module still using the old manure output handler? If the goal is to transition the old manure output handler to add variables to the OM then would it be a good idea to change the name to manure output manager instead of output handler?

I think I should come up with a better name instead of manure module output handler (maybe output manager helper) because what I wanted to do was to have a way to add some extra functionalities to the output manager that are only specific to the manure module. 

> 2. When I run the `animal_managment.json` in this branch, I am still getting a separate directory within the output -> CSVs directory called `manure_module_outputs` even though there is not a csv filter for these output files. Is the goal to keep an option for producing these separate directories?

Right, so currently this option of producing csv files and graphics files still exists, but I will remove that option in this PR so we only have one outputing system that is the output manager.

JoeWaddell (approved):

PR #771, opened 2023-09-14
Correct max root depth and add values for crop different configurations

Isgotkowitz:
Previously, the default for all crops was that the maximum rooting depth was 20 mm. This was not correct. According to SWAT Table A-6 in [Appendix A - Model Databases](https://swat.tamu.edu/media/69419/Appendix-A.pdf), the crops available to simulate in RuFaS have maximum rooting depths that range from 0.6 to 3 meters. This PR corrects the default and adds different values for each preconfigured crop.

Note that SWAT Table A-6 does not specify the units of the maximum rooting depths, but it does cite Kiniry et al, 1995 as the source of the maximum rooting depth values listed. That [article](https://cdnsciencepub.com/doi/pdf/10.4141/cjps95-114) specifies maximum rooting depths in meters, and since the values listed there match the ones in the SWAT table, I assume the SWAT table is also using meters.

### DIFF
--- a/RUFAS/routines/field/crop/crop_data.py
+++ b/RUFAS/routines/field/crop/crop_data.py
@@ -28,6 +28,8 @@ class CropData:
       CropData(light_extinction=0.7) will initialize with the vallue set to 0.7, but CropData(0.65) will not work.
       The upside is that this facilitates dataclass inheritance. For example, the CornData class inherits from CropData
       but will have its values set to different defaults.
+
+      This "Generic" crop uses the values for Grain Sorghum.
     """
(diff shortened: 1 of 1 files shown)


JoeWaddell (approved):
Looks good to me: I always like seeing the underscore to denote a comma for clarity in large numbers.

matthew7838 (approved):
LGTM, I agree with Joe, the numbers were a lot clearer.

Isgotkowitz:
@matthew7838 and @JoeWaddell thank you guys

PR #772, opened 2023-09-14
Bug fixes in `Field` and `Infiltration` modules

Isgotkowitz:
This PR adds simple solutions for two bugs that can occur in the Crop and Soil module:

1. `Denitrification` crashes when `soil_overall_carbon_fraction` is `None` in a soil layer. `soil_overall_carbon_fraction` is initialized as `None`, but is set based on soil layer contents but is set by `CarbonCycle._soil_carbon_aggregation()`. 
The simple solution to this is to switch the order of `CarbonCycling` and `NitrogenCycling` so that carbon is always run first.

2. In the infiltration module, when the second moisture curve number gets below a certain threshold, the first moisture curve number goes negative. This is an issue because it causes the program to try to take the log of a negative number, crashing the program.
The simple solution is to lower bound the first moisture curve number to one
***Pending SME approval***

**Edit:** After discussing with the SMEs, a different approach will be taken to address the bug in the `Infiltration` module.
Originally, the `Infiltration` module automatically adjusted `second_moisture_condition_parameter` that the user input based on the slope of the field. This was (sometimes) causing valid user input to be lowered to unacceptable values, which would in turn crash the program.

After discussing with SMEs, it was decided that this automatic curve number adjustment should not occur because it was causing bugs, but also because it should the responsibility of the user to adjust the curve number for slope as they deem necessary.

Note: the `second_moisture_condition_parameter` will have lower and upper bounds of 20 and 99, respectively, specified in the metadata - issue #782


JoeWaddell (approved):
LGTM: runs properly and justification in description is clear.

RUFAS/routines/field/soil/infiltration.py
```diff
@@ -49,25 +49,13 @@ def infiltrate(self, rainfall: float, weighting_coefficient: float, potential_ev
...
-            adjusted_third_moisture_condition_parameter = self._determine_third_moisture_condition_parameter(
-                                                                        adjusted_second_moisture_condition_parameter)
-        else:
-            adjusted_second_moisture_condition_parameter = self.data.second_moisture_condition_parameter
-            adjusted_third_moisture_condition_parameter = third_moisture_condition_parameter
-        # --------------------------------------------------------------------------------------------------------------
-        adjusted_first_moisture_condition_parameter = self._determine_first_moisture_condition_parameter(
-                                                                        adjusted_second_moisture_condition_parameter)
+        first_moisture_condition_parameter = self._determine_first_moisture_condition_parameter(
+                                                                        self.data.second_moisture_condition_parameter)
```
matthew7838, line 53:
```suggestion
                                                                        self.data.second_moisture_condition_parameter)
```
Where is the `second_moisture_condition_parameter` bounded?

matthew7838 (commented):
Looks like a good change, but there's one small question.

RUFAS/routines/field/soil/infiltration.py
Isgotkowitz: It will be bounded in the soil profile properties section of the metadata

matthew7838: So that will not be addressed in this PR?

Isgotkowitz: #782

matthew7838: Got it. Thanks for the clarification.

Isgotkowitz: > So that will not be addressed in this PR?

Yes, it will be addressed in #774

matthew7838 (approved):
LGTM

PR #776, opened 2023-09-15
Fix nitrate factor equation

### LINKED ISSUE #775: Fix equation in `_determine_nitrate_factor()` in `NitrogenIncorporation`
In the `NitrogenIncorporation` submodule of `Crop`, the method `_determine_nitrate_factor()` contains the equation
```python
        elif total_accessible_nitrates <= 300:
            return 1.5 - (0.0005 * total_accessible_nitrates)
```
which matches the referenced equation (5:2.3.16) in the SWAT Theoretical documentation. 

However, the value returned by this statement is wildly outside the bounds of acceptable ranges for the nitrate factor. If the equation is changed to 
```python
1.5 - (0.005 * total_accessible_nitrates)
```
then the values produced are perfectly distributed within the acceptable bounds of the nitrate factor.

For this reason it seems like the SWAT theoretical documentation is wrong, but this needs to be confirmed with SMEs before proceeding to rewrite the equation and reaching out to the SWAT team.

Isgotkowitz:
Closes #775, issue is described well there.

All lines changed here are fully covered by test suite, no new linting errors introduced.

This solution is pending SME approval.

**Edit:** SME approved, rewrote docstring in NumPy style and noted error in SWAT docs.


allisterakun (approved):
Straight-forward changes, LGTM!

matthew7838 (approved):
Looks good, simple changes.

PR #784, opened 2023-09-19
ration_refactor_m3

JoeWaddell:
This PR is intended to reflect goals of Milestone 3 in the Ration Refactor Design Doc: https://docs.google.com/document/d/1AvB1Lee54gf_6vsOo03BolMJ4qabZWoqJP9QkDD-yig/edit

This is partially implementing the changes in (the now closed) #627 which resulted in so many merge conflicts that it was only logical to split it into two PRs. The addition of the `report_ration_supply` method is now milestone 4, with this simply expanding the `report_ration` method. See Issue #531

Note that the new methodology - using values from the ration from the previous formulation into the requirement calculation process - is not relevant to the initial ration formulation, because there is no "previous" pen. This means that we use the previously defined defaults.


RUFAS/routines/animal/animal_manager.py
```diff
@@ -367,16 +367,30 @@ def calc_nutrient_rqmts(self, feed, temp):
             calf.calc_nutrient_rqmts(feed, temp)
 
         for heiferI in self.heiferIs:
-            heiferI.set_nutrient_rqmts(temp, self.ANIMAL_GROUPING_SCENARIO)
+            pen = heiferI.pen_history[-1].pen
```
Fjordrunner: Maybe `latest_pen` would be a more apt name here, depending on if I'm understanding the code correctly. I assumed that it's not the current pen given that we're pulling from pen_history, but if that's what we're referring to then we could just call it `current_pen` or something around those lines.

RUFAS/routines/animal/animal_manager.py
```diff
@@ -367,16 +367,30 @@ def calc_nutrient_rqmts(self, feed, temp):
...
+            heiferI.set_nutrient_rqmts(temp, self.ANIMAL_GROUPING_SCENARIO,
+                                       nutrient_conc = self.all_pens[pen].ration_nutrient_conc,
+                                       metabolizable_energy = self.all_pens[pen].MEdiet,
+                                       previous_DMI = self.all_pens[pen].dry_matter_intake)
 
         for heiferII in self.heiferIIs:
-            heiferII.set_nutrient_rqmts(temp, self.ANIMAL_GROUPING_SCENARIO)
-
+            pen = heiferII.pen_history[-1].pen
```
Fjordrunner: See first comment.

RUFAS/routines/animal/animal_manager.py
```diff
@@ -367,16 +367,30 @@ def calc_nutrient_rqmts(self, feed, temp):
...
+            heiferII.set_nutrient_rqmts(temp, self.ANIMAL_GROUPING_SCENARIO,
+                                       nutrient_conc = self.all_pens[pen].ration_nutrient_conc,
+                                       metabolizable_energy = self.all_pens[pen].MEdiet,
+                                       previous_DMI = self.all_pens[pen].dry_matter_intake)
+            
         for heiferIII in self.heiferIIIs:
-            heiferIII.set_nutrient_rqmts(temp, self.ANIMAL_GROUPING_SCENARIO)
-
+            pen = heiferIII.pen_history[-1].pen
```
Fjordrunner: See first comment.

RUFAS/routines/animal/animal_manager.py
```diff
@@ -367,16 +367,30 @@ def calc_nutrient_rqmts(self, feed, temp):
...
+            heiferIII.set_nutrient_rqmts(temp, self.ANIMAL_GROUPING_SCENARIO,
+                                       nutrient_conc = self.all_pens[pen].ration_nutrient_conc,
+                                       metabolizable_energy = self.all_pens[pen].MEdiet,
+                                       previous_DMI = self.all_pens[pen].dry_matter_intake)
+            
         for cow in self.cows:
-            cow.set_nutrient_rqmts(self.ANIMAL_GROUPING_SCENARIO)
+            pen = cow.pen_history[-1].pen
```
Fjordrunner: See first comment.

RUFAS/routines/animal/life_cycle/heiferI.py
```diff
@@ -45,18 +45,32 @@ def get_heiferI_values(self):
         """
         return self.get_calf_values()
 
-    def set_nutrient_rqmts(self, temp, animal_grouping_scenario):
+    def set_nutrient_rqmts(self, temp, animal_grouping_scenario, nutrient_conc: dict = {}, 
+                           metabolizable_energy: float = 15.625, previous_DMI: float = 10.0):
         """
         Calculates this heiferI's nutrient requirements.
         """
+        if metabolizable_energy == 0.0: metabolizable_energy = 15.625
```
Fjordrunner: This block of code is essentially being copied in each of the non-calf cow classes. Is this potentially an opportunity to bring this into animal_base and inherit from there?

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -1355,36 +1353,28 @@ def calculate_NRC_DMI(self, animal_type: AnimalType, body_weight: float, day_of_
...
-            # TODO: Refactor this so lactating is not needed
-            if lactating:
-                fat_corrected_milk_kg = (
-                    0.4 * milk_production) + (15 * milk_fat * (milk_production / 100))
-                dry_matter_intake_estimate = (0.372 * fat_corrected_milk_kg + 0.0968 * body_weight ** 0.75) \
-                    * (1 - math.exp(-0.192 * ((days_in_milk / 7) + 3.67)))
-            else:
-                dry_matter_intake_estimate = (
-                    (1.97 - 0.75 * math.exp(0.16 * (day_of_pregnancy - 280))) / 100) * body_weight
+        if animal_type in [AnimalType.LAC_COW]:
```
Fjordrunner, line 1598: Couldn't we just do a direct equality check here to see if the animal type is cow?

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -1355,36 +1353,28 @@ def calculate_NRC_DMI(self, animal_type: AnimalType, body_weight: float, day_of_
...
+            fat_corrected_milk_kg = (
+                0.4 * milk_production) + (15 * milk_fat * (milk_production / 100))
+            dry_matter_intake_estimate = (0.372 * fat_corrected_milk_kg + 0.0968 * body_weight ** 0.75) \
+                * (1 - math.exp(-0.192 * ((days_in_milk / 7) + 3.67)))
+        elif animal_type in [AnimalType.DRY_COW]:
```
Fjordrunner, line 1603: Same as above, can't we just do an equality check here?

Fjordrunner (dismissed):
Overall this looks good. I think there are three things I was able to identify, feel free to ignore them/start a discussion over better alternatives in the comments below.

RUFAS/routines/animal/animal_manager.py
JoeWaddell: Agreed! This is to get the last formulated ration that an animal ate, so it has to find where it last ate, first.

RUFAS/routines/animal/life_cycle/heiferI.py
JoeWaddell: I don't think so, since metabolizable_energy is something calculated from their previous ration, so it's dynamic. Just checking here that there was a formulated ration, in case it's a recently initialized animal, or purchased, etc.

RUFAS/routines/animal/ration/animal_requirements.py
JoeWaddell: I think this (and the similar comment) will be refactored in the upcoming Animal Manager refactor

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -1572,36 +1585,26 @@ def calculate_NRC_DMI(
...
-            net_energy_maintenance_diet = 1  # TODO update this method to retrieve values from nutrient composition of
-            # ration from previous formulation.
-            # Currently using magic value set by Edward and Haowen
-            # see Issue #531
             dry_matter_intake_estimate = (
                 body_weight**0.75
-                * (0.2435 * net_energy_maintenance_diet - 0.0466 * net_energy_maintenance_diet**2 - 0.1128)
-                / net_energy_maintenance_diet
+                * (0.2435 * net_energy_diet_concentration - 0.0466 * net_energy_diet_concentration**2 - 0.1128)
+                / net_energy_diet_concentration
```
tomhuhh: Hello Joe, where can I find this equation? I looked up in the 'RuFaS Heifer requirements PseudoCode' file in the 'Ration Formulation (NRC 2001)' folder on Basecamp but did not find it.

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -1665,14 +1669,22 @@ def calculate_NASEM_DMI(
...
+                0.082
+                * (
+                    NDF_conc
+                    - (
+                        23.1
+                        + 56 * (body_weight / mature_body_weight)
+                        - 30.6 * (body_weight / mature_body_weight) ** 2.0
+                    )
+                )
             )
```
tomhuhh, line 1697: Similarly, where can I find this equation? The equation in the file 'RuFaS Cow Requirements Pseudocode NASEM' for estimating DMI for dry cows is a different one.

tomhuhh (commented):
Nice work Joe! Please see my comments below. I was able to run the code with no difficulty. I checked the failed constraints, and they look fine. However, I cannot review any other outputs easily. Do you have the code to extract the ration information from the output csv file? I basically want to check the average daily intake for each animal.

RUFAS/routines/animal/ration/animal_requirements.py
JoeWaddell: Excellent note: this will need to be added to the Pseudocode, as it came directly from @Ed731551, who used the source material as reference (we cited it in the docstring, but the Pseudocode was apparently not updated). I'll check to see if he has an updated version.

RUFAS/routines/animal/ration/animal_requirements.py
JoeWaddell: Excellent note (again): this will need to be added to the Pseudocode, as they came directly from @Ed731551, who used the source material as reference (we cited it in the docstring).

JoeWaddell:
> Do you have the code to extract the ration information from the output csv file? I basically want to check the average daily intake for each animal.

What I used is the script below (no judgement, I wrote it quickly!): note that this is hardcoded to a specific output file (the last in the output file list, so it depends on how it's sorted), so use with caution, and/or add more args to the command line input to change while file you want to examine.

```
import os
import json
import sys

if len(sys.argv)>1:
    pen_select = sys.argv[1]
else:
    pen_select = 3
    

filelist = os.listdir('output')
file = [file for file in filelist if file.startswith('saved_variables_json_all_variables')][-1]

keyselect = f'ration_per_animal_for_pen_{pen_select}'

with open(str('output/' + file)) as f:
    data = json.load(f)
actual_key = [key for key in list(data.keys()) if key.endswith(keyselect)][0]
ration_raw = data[actual_key]
ration_list = list(ration_raw.values())[-1]

try:
    for line in ration_list:
        del line['status']
        del line['objective']
except:
    pass

percent_dicts = []
DMI_list = []
for line in ration_list:
    percent_dict = {}
    DMI = 0
    for key in line.keys():
        DMI += line[key]
        line[key] = round(line[key],4)
    DMI_list.append(DMI)
    for key in line.keys():
        percent_dict[key] = round(line[key]/DMI, 4)
    percent_dicts.append(percent_dict)

for pos in range(len(percent_dict) - 3, len(percent_dict)):
    DMItotal = round(DMI_list[pos],2)
    print('\n')
    print(f'ration for pen {pen_select} interval {pos}')
    print(f'raw kgs, total = {DMItotal}kg')
    print(ration_list[pos])
    print('percent of DMI')
    print(percent_dicts[pos])
```

RUFAS/routines/animal/ration/animal_requirements.py
Ed731551: Thanks for noticing this. Now updates have been implemented in the pseudocode: 'RuFaS Heifer requirements PseudoCode' file.

RUFAS/routines/animal/ration/animal_requirements.py
Ed731551: The equation for both dry cows and growing heifers is the same (Eq. 2.3 in the NASEM book). Now it is included in the current version of the NASEM pseudocode. Again, thanks for noticing this!

tomhuhh:
Thanks for sharing the code @JoeWaddell! The intakes look ok to me too.

RUFAS/routines/animal/ration/animal_requirements.py
Ed731551: For your information, NRC formulas for dry matter intake are now included in NRC pseudode for heifers: Section D in document: https://docs.google.com/document/d/16z9y9tmYQ3ymRKm-7S_FpDkRDxHIm_L8/edit

I do appreciate comments on this. Thanks!

RUFAS/routines/animal/life_cycle/heiferI.py
ancheypeng: unless I'm missing something, it seems like the `set_nutrient_rqmts` function is the exact same for heiferI, heiferII, and heiferIII, so it would only need to be defined once?

JoeWaddell: Great comments, I see now that @Fjordrunner was likely talking about the whole method, too. I'll note this to be improved in the upcoming animal_manager refactor.

We could make a base method that applies to heifers that later gets updated with one that applies to cows, but we could also consider it the responsibility of Animal Manager itself. I'm not sure which path makes more sense, so if anyone has some input please let me know! I'm collecting notes from all other comments that are slightly outside the scope of the Ration Refactor, to make sure we get to them when refactoring Animal Manager.

ancheypeng: Sounds good! Because the heifer classes already inherit from one another (HeiferI -> HeiferII -> HeiferIII), would it be possible to remove the method from HeiferII and HeiferIII, so that they would just use the method from HeiferI?

ancheypeng (approved):
LGTM!

tomhuhh (approved):
This looks good to me! I ran a simulation with the user defined ration metadata and the nutrient requirements are met and the intakes look within the expected range. Good work Joe~

Ed731551:
To me this look good. Thanks.

PR #785, opened 2023-09-20
Add deep copy to IM

PooyaHekmati:
The entire codebase uses data stored in the IM pool. Since Python passes data by reference only, this means that the whole code base has write access to the IM pool. Consequently, if somewhere in the code we retrieve data from IM and change it, other parts of the code will be affected, which is not desirable.
To mitigate this issue, this PR adds deep copy to the get_data() and get_metadata() functions. Deep copy copies all elements and values within a nested data structure. It adds some computation overhead, but it is needed to prevent side effects. 
For ref, deep copy is implemented here: https://github.com/python/cpython/blob/3.11/Lib/copy.py#L128

### DIFF
--- a/RUFAS/input_manager.py
+++ b/RUFAS/input_manager.py
@@ -1,5 +1,6 @@
 # !/usr/bin/env python3
 
+from copy import deepcopy
 from functools import reduce
 import json
 import re
@@ -562,7 +563,7 @@ class InputManager:
         try:
             data_value = reduce(lambda d, key: d[key], element_hierarchy,
                                 self.__pool)
-            return data_value
+            return deepcopy(data_value)
 
         except KeyError as key_error:
             invalid_key = str(key_error).strip("\'")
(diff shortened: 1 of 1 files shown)


RUFAS/input_manager.py
Isgotkowitz, line 1: LGTM, only question is testing-related: should the addresses of the values returned be checked to make sure they are not the same as the data contained in the `InputManager`?

If not necessary, I'll approve.

allisterakun (approved):
Logic is straightforward, LGTM. Just the same concern about testing that Ed menetioned.

RUFAS/input_manager.py
PooyaHekmati: I'm not sure if I'm understanding your question/comment correctly. Can you please clarify what you mean?

JoeWaddell (approved):

RUFAS/input_manager.py
PooyaHekmati: @Isgotkowitz I'm going to merge this PR as it doesn't change anything about the validation. We can revisit your comment when you are available.

PR #790, opened 2023-09-20
Om improvement

PooyaHekmati:
When we developed save_variables() with the ability to filter the pool, we wanted to deprecate dump_variables() as the save function can have the same functionality (just put .* in a filter file so it matches with everything). However, it seems that has not happened. This PR, first fixes that. Note that the dump_vriables() is still in the code base but it is not used. We want to keep it for debugging purposes when save_variables() is not working. Also, we don't want to force a full dump after every run. Since OM looks for filter files whose names start with either csv or json, prefixing a filter file name with _ (or anything) makes OM ignore that file. I added the filter files to git ignore to make sure local changes won't be synced. I also renamed the filter files to capture their true purpose.


RUFAS/output_manager.py
```diff
@@ -699,11 +699,10 @@ def dump_variable_names_and_contexts(
         )
         self._list_to_file_txt(var_list, file_path)
 
-    def dump_all_pools(self, path: str, exclude_info_maps: bool = False) -> None:
+    def dump_all_nondata_pools(self, path: str, exclude_info_maps: bool = False) -> None:
         """
-        dumps all pool into the given path to a directory.
+        Dumps all non-data pool into the given path to a directory.
```
```suggestion
        Dumps all non-data pools into the given path to a directory.
```

ew3361zh (approved):
Looks good.

A couple questions: 

1. Before merging this PR, will the filter files be renamed with the underscore in front so they're not running each time? e.g. `csv_all_variables.txt` -> `_csv_all_variables.txt`?
2. Is there any docstring notes we should add about the deprecation of `dump_variables()` since it's still in the code? Suggested wording could either be something along the lines of what is written in this PR "Note that the dump_vriables() is still in the code base but it is not used. We want to keep it for debugging purposes when save_variables() is not working." or Numpy has some suggested deprecation warning syntax: https://numpydoc.readthedocs.io/en/latest/format.html#deprecation-warning

JoeWaddell (approved):
Changes look good to me, and I appreciate the thorough explanation. Seconding the questions @ew3361zh asked.

PooyaHekmati:
> Looks good.
> 
> A couple questions:
> 
> 1. Before merging this PR, will the filter files be renamed with the underscore in front so they're not running each time? e.g. `csv_all_variables.txt` -> `_csv_all_variables.txt`?

Due to the conflict with .gitignore, that needs to be done after the merge.

> 2. Is there any docstring notes we should add about the deprecation of `dump_variables()` since it's still in the code? Suggested wording could either be something along the lines of what is written in this PR "Note that the dump_vriables() is still in the code base but it is not used. We want to keep it for debugging purposes when save_variables() is not working." or Numpy has some suggested deprecation warning syntax: https://numpydoc.readthedocs.io/en/latest/format.html#deprecation-warning

Let's use the deprecated library instead which allows us to use versionadded() and versionchanged() as well.

PR #796, opened 2023-09-21
Average annual air temperature calculation change

Isgotkowitz:
This PR changes the way the average annual temperature is calculated and set in the `Weather` module, and incorporates this new value into the `Field` module (which is the only module in RuFaS that uses it).

`classes.py` has been modified so that it has 0 linting issues, and a test file has been added to test the new method in `Weather`.


RUFAS/classes.py
```diff
@@ -240,20 +242,36 @@ def __init__(self, weather_file, config):
             self.radiation[current_year_index][current_day_index] = weather_file['Hday'][i]
             self.irrigation[current_year_index][current_day_index] = weather_file['irrigation'][i]
 
-        # calculates T_avg_annual for each year
-        for i in range(len(years)):
-            avg = sum(self.T_avg[i]) / (len(years[i]))
-            self.T_avg_annual.append(avg)
+        self.T_avg_annual = self._calculate_average_annual_temperature(weather_file['avg'])
```
ew3361zh, line 245: It looks like the previous way this was calculated did a calculation for each year in a simulation but this looks like it's just doing one calculation for the T_avg_annual and applies it to each year? The Weather objects are so complicated so maybe I'm not interpreting what's happening here properly.

RUFAS/classes.py
```diff
@@ -240,20 +242,36 @@ def __init__(self, weather_file, config):
...
+        Notes
+        -----
+        This method calculates the average annual air temperature by taking the average of all daily average air
+        temperatures provided in the weather input file.
+
+        This method is intended to approximate SWAT's method for calculating the average annual temperature. SWAT
+        calculates average high and low temperatures for each month over every simulated year, then averages those
+        values to get a single annual average air temperature for the entire simulation. The exact implementation for
+        this can be found at in the SWAT source code file `readwgn.f
+        <https://bitbucket.org/blacklandgrasslandmodels/swat_development/src/master/readwgn.f>`_
```
ew3361zh, line 272: I think (briefly) outlining the issue of how avg air temp was calculated previously in the initial comment about the PR would be helpful for reference (i.e. why did this change need to happen?).

RUFAS/classes.py
ew3361zh: Also, this line (245) was showing up as not covered by the testing suite. I'm not sure how much you want to get into the weeds with expanding coverage on the whole Weather class but maybe there's a simple init of a mocked Weather object that shows `_calculate_average_annual_temperature()` has been called.

Isgotkowitz: > this looks like it's just doing one calculation for the T_avg_annual and applies it to each year?

Yes, that is apparently the expected behavior. I don't totally understand why (and Pete was a little puzzled by it too) but having a single value for the entire run of the simulation is the desired outcome.

> Also, this line (245) was showing up as not covered by the testing suite.

Yeah, unfortunately the entire `Weather` init function is totally uncovered. I started writing a test for it and then realized how complex it would have to be in order to even be decent. I think there is a fair argument for not writing any kind of decent testing for weather as it is right now because its behavior is so problematic - but that is the lazy programmer in me talking. If you think it would be worth it to have at least some testing then I can write a simple one.

RUFAS/classes.py
Isgotkowitz: Good suggestion

RUFAS/classes.py
ew3361zh: Okay, I just gave it a quick attempt and I agree it's not worth the set-up to get this one line covered especially if the Weather class is basically uncovered.

ew3361zh (approved):
LGTM

allisterakun (approved):

PR #802, opened 2023-09-25
Clean `input` directory

### LINKED ISSUE #769: Clean up Crop and Soil inputs in `input/` directory
During the integration of `InputManager` into the Crop and Soil module, only the ARL and Barnyard simulation files were completely rewritten to be compatible with the metadata structure. There are currently a number of different Field input files that need to be rewritten and put together into new metadata files so that they can be run with the new input system.

The configurations that need to be rewritten are:
- [x] Multi crop - completed in #774 
- [ ] Swat - completed in #801
- [x] Testing - completed in #799
- [ ] Remove unused inputs - completed in #802

Isgotkowitz:
Removes unused and old files from `input` directory.

There are 7 metadata files that should all run full simulations of RuFaS: 
- ARL
- Barnyard
- Default
- Multi-crop
- SWAT
- Testing
- User defined ration

All Field module related input files that are not used in these metadata files are removed by this PR.

Assuming all PRs stacked beneath this one get merged before this one, closes #769.

_Note_: this PR breaks a dependency in the linting workflow. This happens because the name of one of the deleted files contains a space. Given that this file is deleted and there are no Python files touched in this PR, I think it would be safe to override the Linting workflow for this PR.


input/data/crop/corn_rotation.json
```diff
@@ -1,16 +1,15 @@
...
+            "crop_species": "corn",
+            "planting_years": [2009],
+            "planting_days": [121],
             "harvest_years": [2009],
-            "harvest_day": [319],
+            "harvest_days": [319],
             "harvest_operations": ["default"],
-            "harvest_type": "scheduled"
+            "harvest_type": "scheduled",
+            "pattern_repeat": 1,
```
Fjordrunner: My only concern here is if the values for "pattern_repeat" and "pattern_skip" are binary values, I'd prefer for them to be True or False instead, but if I'm misunderstanding the values then please ignore this.

Fjordrunner (commented):
Ready to approve after confirming the nature of those "pattern_repeat" and "pattern_skip" values in corn_rotation.json

input/data/crop/corn_rotation.json
Isgotkowitz: They are not binary values, good note though. Pattern repeat is the number of times the pattern is repeated, and pattern skip is the number of steps to skip between repetitions of the pattern (the patterns for crop inputs are planting years and harvesting years).

Fjordrunner: In that case, do you think we could rename those variables to something around the lines of "num_pattern_repetitions" and "num_pattern_skips", or those without the "num" prefix, if you think that's too verbose? Thanks for the clarification!

Fjordrunner (approved):

matthew7838 (approved):
LGTM. I also agree with @Fjordrunner 's suggestion. I think it's more clear.

Issue #533, opened 2023-05-23
Insufficient test coverage for OutputManager.save_variable_names_and_contexts()


PooyaHekmati:
OutputManager.save_variable_names_and_contexts() has some logic in it but the unit case is not checking any of them. The test coverage for this function needs to be increased.

ew3361zh:
Test coverage for OM is 100% as of 9/25/2023.

PR #804, opened 2023-09-25
small patch to address OM.py lines missing coverage

ew3361zh:
When looking through open issues I saw #533 and when double checking that we had 100% coverage for OM, saw that two lines were reading as uncovered (660-661 - an exception being raised by a bad path for saving variables to a CSV file).

This PR addresses that coverage discrepancy by more correcting the types of potential exceptions/errors that would be raised in a bad path exception - this should get OM test coverage to 100%.


Isgotkowitz (approved):
Nice

allisterakun (approved):
Straight-forward change, LGTM

PR #807, opened 2023-09-25
Add more extensive data validation in `InputManager`

### LINKED ISSUE #783: `InputManager` does not correctly replace invalid inputs with default values
When an invalid value is passed as an input to the `InputManager` and the property associated with that input has a specified default, the invalid value passed should be replaced with the default value.

## Expected Behavior

An invalid input value should be replaced with the default value, if one is provided.

## Current Behavior

The invalid value is not corrected, and whichever module that uses it grabs it from the `InputManager` and uses it.

## Steps to Reproduce

1. Pick any input a file which has a default specified in the `properties` section of the metadata, and change it to `null` or a value outside of the acceptable bounds.
2. Run RuFaS
3.
4.

Isgotkowitz:
Adds better type checking during the input validation stage, so now if a data of the wrong type is provided to the validation methods, instead of crashing they will mark it as an invalid value and let it be corrected later on.

This PR contains a test for one method not covered at all in `input_manager.py`, brings coverage to 100%. That method is `flush_pool()` in `InputManager`, which was formerly `flush_pools()` - its usage is changed in `main.py`.

It also changes 3 input files that converts dictionaries to lists.


Isgotkowitz:
@PooyaHekmati in the functional requirements section of the InputManager design doc, it says that the InputManager must be able to differentiate between critical and non-critical inputs, and be able to fix non-critical inputs when they are invalid.

I am wondering how this extends to the case when a variable is specified in the properties section, but is not present in the input file. Currently the InputManager raises an error specifying the missing key, then stops running when something like this happens. But I think better behavior would be to replace that value with the default if there is one. Thoughts?

allisterakun (approved):
LGTM

matthew7838 (approved):
LGTM, good changes.

PR #809, opened 2023-09-27
Change pen information input from dictionary to list

Isgotkowitz:
This PR changes the way that pen information is specified in the animal input file, now matching what is specified in the properties for animal inputs.

It also adjusts the way that pens are initialized in `init_pens()` in `AnimalManager`, though these changes are quite simple. The existing test for this method was also adjusts to follow the new input format.

By fixing pen information, strict type checking can be implemented in `_array_type_validator()` in `InputManager` and the TODO removed, completing the work started in #783.

Also, a number of linting issues have been fixed in `AnimalManager`.


Isgotkowitz:
@KFosterReed @JoeWaddell hoping you two can take a look at this PR. It makes minor changes to the way pen information is passed to the animal module, but want to make sure you both know about it and have the chance to consider any downstream effects.

KFosterReed (approved):
Looks like a simple change and runs well with a different number of pens!

JoeWaddell (approved):
LGTM: works properly, and removing the repetition makes sense from a design perspective, too.

PR #801, opened 2023-09-25
Swat input rewrite

Isgotkowitz:
This PR rewrites the input files needed to run the SWAT simulation with the `InputManager`.

PR #799, opened 2023-09-22
`testing` input rewrite

Isgotkowitz:
This PR rewrites the `testing.json` file to follow the metadata format and be compatible with the `InputManager`.

Note this PR is deceivingly large - most of the lines added come from the brand new metadata file.

PR #779, opened 2023-09-15
Output Graph Generation

allisterakun:
This PR implements issue #777. We do so by introducing a new class GraphGenerator. This class serves the purpose of generating plots and graphs from the simulation results.

To use this class to generate a graph, the user should create a `JSON` file (please use `png_default_filter.json` as an example) and place it under the `output/output_filter/` folder. The file name start should start with the `"png_"` prefix, which is used in `output_handler.py` to identify the desired output file type.

The `JSON` file should contain the following information:
1. Graph Type (line graph, bar graph)
2. RegEx filter(s) to filter the output to get the desired variables for the graph
3. (Optional) Other information (title, x-axis and y-axis label): all other information is optional. If there are any other customizations that you need, please contact me via Slack

Currently, the GraphGenerator class supports two types of graphs: line graphs and bar graphs, as they are the most commonly used types. If you ever need to generate a graph of any other type, please contact me via Slack.

--------------------------------
Edit and updates from Pooya
--------------------------------
This PR implements issue #777. We do so by introducing a new class GraphGenerator. This class serves the purpose of generating plots and graphs from the simulation results.

This PR introduces new filter files for the output manager which are needed for graph generation. Therefore, the names of these filters start with `graph_`. Since they need to contain more information, they are JSON rather than text. These JSON filter files contain:
1. `type`: the type of the graph. While many types are being introduced, only plot and stackplot are tested in this PR. It is likely that other graph types fail. In that case, please reach out to the dev team and we will address the issue in future PRs.
1. `filters`: the good old regex-based filters. A list of regex filters is expected here. Any matching variable in the OM pool will be plotted on the same graph.
1. `variables`: When the data is stored in dictionaries (i.e., when reporting the variable to the output manager using `OutputManager.add_variable()`, a dict was passed), the `variables` entry allows selecting which variables in the dictionary are being plotted.
1. other details: including legend, title, etc... The following list is currently implemented:
```
align_labels
aspect
canvas
constrained_layout
dpi
edgecolor
facecolor
figheight
figsize
figwidth
frameon
grid
legend
snap
subplot_adjust
tight_layout
title
transform
xlabel
xticklabels
xticks
xlim
ylabel
yticklabels
yticks
ylim
yscale
xscale
zorder
```

An example JSON filter is provided in this PR:
```
{
  "type": "plot",
  "filters": [
    "Pen\\.calc_total_manure\\.daily_aggregate_calf_(degradable|non_degradable)_volatile_solids"
  ],
[...]

PR #810, opened 2023-09-27
Sim eng IM integration

ew3361zh:
This PR takes the next step in IM integration to the model by:

1. Updating Simulation Engine methods to reflect that they no longer rely on loading input_file_paths as that data has been shifted to be loaded and validated by Input Manager.
2. Updates main.py functions to reflect that they now take in and process paths to metadata files, not input files.
3. Updates main.py `execute_simulations_from_files` function to `execute_simulations` and fully incorporates Input Manager data validation into the decision to run a simulation.
4. Updates unit testing and docstrings for main.py functions. 
5. Reorganizes file structure in input folder to have a metadata folder and a data folder which contains subfolders for each blob input file. Old-style input files have been temporarily moved to a folder in input called "deprecated_input_files" for reference purposes.
6. Deprecates user_prompt.py - simulation runs with a hardcoded list of metafiles in the main.py global METADATA_PATHS variable which is fed to `execute_simulations`.

Notes:
1. Changes to main.py and to the use of IM will be announced prior to merging in the dev team and general channels to ensure everyone knows it's coming.
2. gitignore file is not behaving as expected for me locally. The two recent changes (adding output/output_filters/ and this PR adding input/metadata/) are not working. If I change files in those folders locally, they are appearing in my git changes and will be added to my staged changes if I run `git add .`. Not sure why.
3. I got a large file warning from Git but am not sure to which file it is referring (seems like maybe it's the entire commit): 
remote: warning: File dc171fbcfee093c031fe67f47f550bce6a7ecebb is 79.43 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB
4. This PR is deceptively large because of moving the files around.


PooyaHekmati (changes requested):
LGTM overall, but there are few high level things:

> 5. Updates to user prompt and main.py gnu -h arg to indicate that the model is looking for a metadata file(or files), not an input file (or files).

We extracted the metadata from the code and saved it as a JSON file for a variety of reasons, one of which is to make it possible to run the model with different versions of metadata. However, the metadata is meant to be under the hood. It is not something that a user would normally care about. An average user won't even know that such a thing exists. Therefore, I think the signature of `execute_simulations()` should be the same as what you have in this PR to allow running the sim engine with multiple metadata files in one go, if that is desired. But, we don't want the user prompt here. Instead, let's make a variable named `METADATA_PATHS` in main.py and hardcode the list of metadata there. And pass that to `execute_simulations`. Later on, the scenario manager will handle this list more elegantly, once we implemented it. 
> 
> Notes:
> 
> 1. @allisterakun and I briefly discussed a roll-out on the switch to using metadata files (and which metadata files are currently viable). I think messaging in the general Slack channel would be a good place to start but are there other avenues that would be helpful?

Implementing different metadata files is not the goal here (at least for now). Right now, we want to have one working metadata that we can rely on (please don't call it example_xyz). We can add more matadata files down the road if needs be. But for now, the emphasis is on fully integrating the input manager and making sure it is serving as SSOT; which doesn't seem to be done yet, contrary to what IM integration spreadsheet says.

> 2. I didn't clear out the input folder of non-metadata files since there may be some reference files of scenarios people ran with the previous input files and there may be some settings in there they want to reference when making their own custom metadata input files. Feedback on any action that would be helpful to take here would be appreciated.

I noticed that too. I think for now, the right course of action is to first make sure we have a reliable alternative (IM) then communicate the change and remove the deprecated files. Again, I want to emphasize that while we want to accommodate versioning in metadata and allow (in your words) customization, that shouldn't be seen as a normal day-to-day operation. Two things that come to my mind (please evaluate how much they make sense) are: 1. Can we organize the data and metadata files in folders with similar names under the input folder? So the metadata would be under inputs/metadata/ and the data would be under inputs/data/. 2. can we add the metadata to .gitignore?

PooyaHekmati:
I forgot to mention that this PR will have a merge conflict with PR #779 . The graph generation gnu arg and its time calculation (both of which need to be added to #779) are also modified here. Let's get this one merged first.

Isgotkowitz:
@PooyaHekmati @ew3361zh A couple of PRs I want to draw your attention to:
1. There are two PRs for rewriting old top-level inputs as metadata - #799 and #801
2. There is a PR for cleaning out unused or deprecated files from the `input/` directory - #802. It does not completely clean the directory, but removes all unused or deprecated Crop and Soil related files.

Once the above PRs have been merged, there will be a bunch of different metadata files specifying different simulations that can be run:
- Example
- ARL
- Barnyard
- Multi Crop
- SWAT
- Testing
- User defined ration

After 802 is merged, there will be one top level input file that uses the deprecated format, animal_management.json.

Isgotkowitz:
> However, the metadata is meant to be under the hood. It is not something that a user would normally care about. An average user won't even know that such a thing exists.

@PooyaHekmati I am not understanding how this will be reasonably achieved or even why this is the desired outcome. If users want to specify a simulation of RuFaS for their purposes, they will need to know not just how to write individual input files but how to bundle those input files together in a way that RuFaS can correctly understand and parse. There is a case where a correct input file can be written only if it references names in the metadata correctly, and it seems impossible for users to be able to do this and not be aware of the existence of the metadata.

Isgotkowitz (approved):
LGTM, nice clean code. Maybe it would be good to move the tests for `SimulationEngine` to a dedicated file at some point (i.e. `test_simulation_engine.py`) but that may be outside of the scope of this PR.

PooyaHekmati:
> > However, the metadata is meant to be under the hood. It is not something that a user would normally care about. An average user won't even know that such a thing exists.
> 
> @PooyaHekmati I am not understanding how this will be reasonably achieved or even why this is the desired outcome. If users want to specify a simulation of RuFaS for their purposes, they will need to know not just how to write individual input files but how to bundle those input files together in a way that RuFaS can correctly understand and parse. There is a case where a correct input file can be written only if it references names in the metadata correctly, and it seems impossible for users to be able to do this and not be aware of the existence of the metadata.

This was the idea behind separating code, metadata, and data from the very beginning. This is the case because the metadata and code are expected to be tightly coupled and metadata is expected to change only when the code changes. The metadata describes what the model expects to receive. In other words, metadata is the API agreement. It can only change when the backend changes. If the user wants to modify the inputs, that change is data, not metadata; and by definition, data can not be treated as metadata. If some part of the code is using the metadata to accommodate variations in the inputs, that doesn't seem to be in the right direction.

As we have iterated in meetings, metadata should be seen and treated as part of the model and the codebase, not part of the input. It closely mimics the role of schemas in DCA. It will be invisible to the user thanks to the scenario manager and the interfaces that it will provide such as APIs, DB, Caches, etc... Remember, we are using files as the only IO stream right now, but that will change. Any development in data and metadata should be done with that in mind.

Another note is that the alternative solution was to bake all the metadata information into the model itself. We extracted that to make it more visible, accessible, and flexible. We shouldn't push the flexibility and accessibility of a component to such an extreme that it makes the other components inaccessible and rigid. All the conversations we had about iPhone vs Android originated from here. We also had a conversation about having two sets of input: a and b, where there is n to n mapping between them. We discussed how the mapping can be stored and retrieved and we concluded that a new data file, c, can be added to do this.

ew3361zh:
>Therefore, I think the signature of `execute_simulations()` should be the same as what you have in this PR to allow running the sim engine with multiple metadata files in one go, if that is desired. But, we don't want the user prompt here. Instead, let's make a variable named `METADATA_PATHS` in main.py and hardcode the list of metadata there. And pass that to `execute_simulations`. Later on, the scenario manager will handle this list more elegantly, once we implemented it.

What do you think about deprecating user_prompt.py? Removing the call from main.py to `user_prompt.obtain_file_list()` means nothing in the user_prompt.py file is used any further in any part of the model.


> Implementing different metadata files is not the goal here (at least for now). Right now, we want to have one working metadata that we can rely on (please don't call it example_xyz). We can add more matadata files down the road if needs be. But for now, the emphasis is on fully integrating the input manager and making sure it is serving as SSOT; which doesn't seem to be done yet, contrary to what IM integration spreadsheet says.

There are no other input files being used and all input data is coming from the IM pool at this point.
 

> I noticed that too. I think for now, the right course of action is to first make sure we have a reliable alternative (IM) then communicate the change and remove the deprecated files. Again, I want to emphasize that while we want to accommodate versioning in metadata and allow (in your words) customization, that shouldn't be seen as a normal day-to-day operation. 

I think the point here is how to make this work best in a transition period and how best to roll that out. For example if Joe or Kristan wants to run different simulations as they do now, how should they best work with a single metadata file that points to one set of input files? I don't know how many different input files they use between different simulations but in the old way of doing things, I believe there were 4 main input files that Kristan ran the simulation with to test how things were working - animal_management.json, ARL.json, barnyard.json, and LT.json. Each one of those had a different group of settings.

I don't think we'd want people to alter the metadata "properties" sections but the "files" section being flexible to define different "metadata" files might be helpful to SMEs between now and implementing the Scenario Manager.

> Two things that come to my mind (please evaluate how much they make sense) are: 1. Can we organize the data and metadata files in folders with similar names under the input folder? So the metadata would be under inputs/metadata/ and the data would be under inputs/data/. 2. can we add the metadata to .gitignore?

I think this could work - I like the metadata being in a metadata folder so people can identify it as such. I think having all the input data files in one folder could prove a little crowded depending on how many different sets of input files the SMEs (and others) want as kind of a standard set of input files to test different scenarios in the model.

PooyaHekmati:
> > Therefore, I think the signature of `execute_simulations()` should be the same as what you have in this PR to allow running the sim engine with multiple metadata files in one go, if that is desired. But, we don't want the user prompt here. Instead, let's make a variable named `METADATA_PATHS` in main.py and hardcode the list of metadata there. And pass that to `execute_simulations`. Later on, the scenario manager will handle this list more elegantly, once we implemented it.
> 
> What do you think about deprecating user_prompt.py? Removing the call from main.py to `user_prompt.obtain_file_list()` means nothing in the user_prompt.py file is used any further in any part of the model.
> 

I agree, I just didn't want to overwhelm you. 
> > Implementing different metadata files is not the goal here (at least for now). Right now, we want to have one working metadata that we can rely on (please don't call it example_xyz). We can add more matadata files down the road if needs be. But for now, the emphasis is on fully integrating the input manager and making sure it is serving as SSOT; which doesn't seem to be done yet, contrary to what IM integration spreadsheet says.
> 
> There are no other input files being used and all input data is coming from the IM pool at this point.
> 

In that case, we should do some housekeeping and clean up. Maybe we announce it in both dev team and general channels to make sure everyone has seen it?

> > I noticed that too. I think for now, the right course of action is to first make sure we have a reliable alternative (IM) then communicate the change and remove the deprecated files. Again, I want to emphasize that while we want to accommodate versioning in metadata and allow (in your words) customization, that shouldn't be seen as a normal day-to-day operation.
> 
> I think the point here is how to make this work best in a transition period and how best to roll that out. For example if Joe or Kristan wants to run different simulations as they do now, how should they best work with a single metadata file that points to one set of input files? I don't know how many different input files they use between different simulations but in the old way of doing things, I believe there were 4 main input files that Kristan ran the simulation with to test how things were working - animal_management.json, ARL.json, barnyard.json, and LT.json. Each one of those had a different group of settings.
> 
> I don't think we'd want people to alter the metadata "properties" sections but the "files" section being flexible to define different "metadata" files might be helpful to SMEs between now and implementing the Scenario Manager.
> 

I agree and that's why I suggested adding this to gitignore. So they can have their own input files (data and metadata) without worrying. Another thing is that maybe having different metadata files and asking them change the hardcoded variable in main.py?

> > Two things that come to my mind (please evaluate how much they make sense) are: 1. Can we organize the data and metadata files in folders with similar names under the input folder? So the metadata would be under inputs/metadata/ and the data would be under inputs/data/. 2. can we add the metadata to .gitignore?
> 
> I think this could work - I like the metadata being in a metadata folder so people can identify it as such. I think having all the input data files in one folder could prove a little crowded depending on how many different sets of input files the SMEs (and others) want as kind of a standard set of input files to test different scenarios in the model.

We can have sub-directories under inputs/data/ similar to what we currently have under inputs/ right now.

Isgotkowitz:
> If the user wants to modify the inputs, that change is data, not metadata; and by definition, data can not be treated as metadata. If some part of the code is using the metadata to accommodate variations in the inputs, that doesn't seem to be in the right direction.

Ok, this helps things make more sense. If I am understanding correctly, the `files` section of the metadata is actually more like data and not metadata (and should be easily modified by users) while the `properties` section is purely metadata and should not be touched at all by users?

Also, we should review the field input structure at our next one-on-one or sooner, because if the data relying on "metadata" is wrong that should be addressed.

main.py
```diff
@@ -12,24 +12,23 @@
...
 import config.global_variables
 from RUFAS.simulation_engine import SimulationEngine
-from RUFAS.user_prompt import obtain_file_list
 from RUFAS.input_manager import InputManager
 from RUFAS.output_manager import OutputManager
 from RUFAS.util import Utility
 
 
+METADATA_PATHS = ['input/metadata/default_metadata.json']
+
```
Isgotkowitz: This seems like a good temporary solution. Is there a certain design in mind for what the longer term solution will be though?

Isgotkowitz: > Another thing is that maybe having different metadata files and asking them change the hardcoded variable in main.py?

What if instead of having the path to the metadata hardcoded in main.py, a path to a file in `input/` was hardcoded into main.py and this file contained the path(s) to all the metadata files to be run. This file could be ignored by git, which would make it really easy to change the input files without having to actually touch the hardcoded value in main.py.

Of course this would take more work to implement, and if it would just be replacing one temp solution with another then it may not be worth it.

PooyaHekmati: Well, this can be a good starting point for the Scenario Manager. We can have it the bare minimum right now; later on, when we have more bandwidth, we can design and implement it the right way. @ew3361zh Do you think this is a good path forward?

PooyaHekmati:
> > If the user wants to modify the inputs, that change is data, not metadata; and by definition, data can not be treated as metadata. If some part of the code is using the metadata to accommodate variations in the inputs, that doesn't seem to be in the right direction.
> 
> Ok, this helps things make more sense. If I am understanding correctly, the `files` section of the metadata is actually more like data and not metadata (and should be easily modified by users) while the `properties` section is purely metadata and should not be touched at all by users?
> 

There are no rules that say anything can't be touched by users. Users, if they are comfortable and confident, can change everything. The point is that we shouldn't develop the system/product with the assumption that users are willing to change things here and there (again, think about Apple products). That being said, the metadata in its entirety is metadata. A good way of thinking about it is to think of metadata as a piece of code. Can users change it? Yes! Do we expect them to do so? No! To put things in simple terms, metadata is a part of the system, not part of the input to the system. If the confusion is coming from metadata being located in the input folder, we can move it to somewhere else (we probably should).

JoeWaddell:
Apologies that this is a really long comment, but I'm trying to summarize/comment on the whole conversation, because there's likely things folks on this PR can clarify. Likewise it might be useful to see how I've been using RuFaS (though from reading the previous comments, I think everyone already knows all of this). 

So the way that many of us use RuFaS right now is that we'll run comparisons between different scenarios (hence the idea of Scenario Manager) - the outputs being data we'll use for presentations/publications. 

As an example, in refactoring Ration methodology, I often switch between multiple Feed JSONs, and one or two Animal JSONs. I have a few different "entry point jsons" (not sure what else to call the initial input file that RuFaS uses - currently the metadata file) that have pointers to the appropriate Animal/Feed JSONs, which themselves have the different combinations of animal management/feed values that I want to test during development (and to collect data for publication!). 

What it looks like in practice:  `python main.py input/'Combination1.json`, , `python main.py input/'Combination2.json`....`python main.py input/'Combination10.json`, where the combinations refer to the lower-level input JSONs, e.g., AnimalA.json+FeedA.json, then AnimalA.json+FeedB.json....and so forth, until it's something like AnimalB.json + FeedE.json.

My understanding is that the eventual Scenario Manager will handle the above more elegantly than having to create so many high- and low-level input files, especially since at present it means modifying the metadata file (which I see should not be thought of as an input). We certainly don't want to waste time developing architecture that we'll throw away once Scenario Manager is rolled out, but we also have to think about how the interim system will work so that we don't lose time renaming/modifying files in the meantime.

So as I understood the idea of hardcoding a path to a single file: I'd worry that having to modify/rename those JSONs each run would make it significantly clunkier to use. Or that it's essentially the same behavior, but now modifying main.py instead of changing the command line arguments.

PooyaHekmati:
> Apologies that this is a really long comment, but I'm trying to summarize/comment on the whole conversation, because there's likely things folks on this PR can clarify. Likewise it might be useful to see how I've been using RuFaS (though from reading the previous comments, I think everyone already knows all of this).
> 
> So the way that many of us use RuFaS right now is that we'll run comparisons between different scenarios (hence the idea of Scenario Manager) - the outputs being data we'll use for presentations/publications.
> 
> As an example, in refactoring Ration methodology, I often switch between multiple Feed JSONs, and one or two Animal JSONs. I have a few different "entry point jsons" (not sure what else to call the initial input file that RuFaS uses - currently the metadata file) that have pointers to the appropriate Animal/Feed JSONs, which themselves have the different combinations of animal management/feed values that I want to test during development (and to collect data for publication!).
> 
> What it looks like in practice: `python main.py input/'Combination1.json`, , `python main.py input/'Combination2.json`....`python main.py input/'Combination10.json`, where the combinations refer to the lower-level input JSONs, e.g., AnimalA.json+FeedA.json, then AnimalA.json+FeedB.json....and so forth, until it's something like AnimalB.json + FeedE.json.
> 
> My understanding is that the eventual Scenario Manager will handle the above more elegantly than having to create so many high- and low-level input files, especially since at present it means modifying the metadata file (which I see should not be thought of as an input). We certainly don't want to waste time developing architecture that we'll throw away once Scenario Manager is rolled out, but we also have to think about how the interim system will work so that we don't lose time renaming/modifying files in the meantime.
> 
> So as I understood the idea of hardcoding a path to a single file: I'd worry that having to modify/rename those JSONs each run would make it significantly clunkier to use. Or that it's essentially the same behavior, but now modifying main.py instead of changing the command line arguments.

I think the proposed solution allows us to do the same thing in an easier way. This is because you can still have as many metadatas as you would like. The only difference is that currently, you list them as GNU args (main.py a main.py b etc...); with the changes in this PR you will save the list in a Python file, which means your experiments are easily repeatable and you can always go back and see what you did yesterday.

ew3361zh:
@PooyaHekmati if you could please review again now.

Updated changes:
1. Created `scenario_manager.py` which holds the METADATA_PATHS variable used to run a/multiple simulation(s).
2. Removed the hard-coded METADATA_PATHS variable in main.py since that same info is now coming from `scenario_manager.py`.
3. `user_prompt.py` completely deprecated.
4. Updated file/directory structure of input directory - now separate metadata and data dirs with the data dir having all the previous blob subdirs.
5. Updated and tested each of the metadata files to ensure their paths are updated with the new file/dir structure.
6. Updated gitignore to include metadata subdir as well as scenario_manager.py file.
7. Updated location of "Simulating..." message to appear earlier to help users know simulation is active.
8. Resolved conflicts with master.

I spoke with @JoeWaddell about the idea of changing the METADATA_PATHS variable in the `scenario_manager.py` to be a list of tuples that contains the path to the metadata used along with a name that could be used to ID the output files associated with each particular run and I think that better fits in a separate PR/issue. So I wrote it up in #817. Can address as urgency dictates.

Once approved, I will hold off merging until I've sent out a summary on dev-team and general Slack channels.

RUFAS/simulation_engine.py
```diff
@@ -40,19 +32,17 @@ def simulate(self) -> None:
...
         if config.global_variables.PRODUCE_GRAPHICS:
             sys.stdout.write('Producing Graphics\n')
-            t_start_graphics = timer.time()
-            t_end_graphics = timer.time()
-            graphics_prod_time = t_end_graphics - t_start_graphics
-        else:
-            graphics_prod_time = 0
+        t_start_graphics = timer.time()
+        t_end_graphics = timer.time()
+        graphics_prod_time = t_end_graphics - t_start_graphics
```
PooyaHekmati: Let's remove this altogether.

RUFAS/simulation_engine.py
```diff
@@ -139,37 +128,19 @@ def _annual_simulation(self) -> None:
...
         self._run_post_annual_routines()
 
-    def _initialize_simulation(self, file_path: Path) -> None:
-        """Reads the json file, writes information to the simulation variables.
-
-        Reads and interprets the (json) file at the given path. Compiles the
-        information into dictionaries and instantiates the simulation objects with
+    def _initialize_simulation(self) -> None:
+        """Requests data from the Input Manager pool to compile the
+        information into dictionaries and instantiate the simulation objects with
```
PooyaHekmati:
```suggestion
        """Instantiates the simulation object by requesting data from the Input Manager."""
```

RUFAS/simulation_engine.py
```diff
@@ -139,37 +128,19 @@ def _annual_simulation(self) -> None:
...
         them. Assigns the objects to the global simulation variables.
```
PooyaHekmati:
```suggestion
```

RUFAS/simulation_engine.py
```diff
@@ -139,37 +128,19 @@ def _annual_simulation(self) -> None:
...
+        """
```
PooyaHekmati, line 115:
```suggestion
```

input/scenario_manager.py
```diff
@@ -0,0 +1,22 @@
+# !/usr/bin/env python3
```
PooyaHekmati: This file is misplaced. The path to this file should be RUFAS/scenario_manager.py

main.py
```diff
@@ -8,28 +8,26 @@
 """
 import argparse
 from pathlib import Path
+import sys
 from typing import List
+from input.scenario_manager import METADATA_PATHS
```
PooyaHekmati:
```suggestion
from RUFAS.scenario_manager import METADATA_PATHS
```

RUFAS/simulation_engine.py
```diff
@@ -139,37 +128,19 @@ def _annual_simulation(self) -> None:
...
-            self.state = State(self.config,
-                               self.weather, self.time)
-
-        except errors.JSONfileData as e:
-            print(
-                f"JSON FILE ERROR: {file_path.name}\n\t{e.section} Section\n{e.msg}\n")
-            raise errors.InvalidJSONfile(file_path.name)
+        self.weather = Weather(data_weather, self.config)
+        self.time = Time(self.config)
+        self.state = State(self.config, self.weather, self.time)
```
PooyaHekmati: missing test coverage

PooyaHekmati (approved):
This PR seemed easy at first but got many twists and turns. Nicely done.

RUFAS/simulation_engine.py
ew3361zh: Updated

RUFAS/simulation_engine.py
ew3361zh: Attempted to add test coverage but issues with complexity of Weather class made it too difficult. Made a new issue to address SimulationEngine test coverage when we revisit classes.py for Milestone 3: #821

PR #811, opened 2023-09-28
Ration optimizer unit tests

ancheypeng:
Adds some unit tests for the RationOptimizer functions


JoeWaddell (changes requested):
These look great! Just a few more tests to add in: the `__init__` for of the Ration Optimizer, `set_constraints`, `total_energy`, and `get_ration_vals`. Most of those should be easy, especially since `total_energy` is just the three NE methods stacked together.

Two remain to do after those: `optimize` and `attempt_optimization`, but those are a lot trickier to write tests for, because they rely heavily on helper functions. I'm at the same crossroads in ration_driver, because formulate_ration is equally complex and reliant on other methods.  Patching in results for each of the helper functions should work, but we can chat about what strategy we ought to take.

Issue #805, opened 2023-09-25
Make percolation out of soil layer occur before percolation into soil layer


Isgotkowitz:
Water percolation downwards through the soil profile is not occurring as desired.

## Current
In the method `percolate()` of the `Percolation` module, the water that percolates out of a soil layer is *immediately* added to the next layer below it.

## Problem
The result of the above order of operations is that there is no opportunity for the water that percolates into a soil layer to stay there before it is available to percolate out of the layer.

## Solution
Adjust the percolation method so that water percolating into a layer on a certain day is not available to percolate out of it on the same day.

ew3361zh:
Resolved in #818

PR #818, opened 2023-10-02
Percolation patch

ew3361zh:
This PR seeks to address issue #805 by:

1. Rearranging the logic in the `Percolation.percolate()` method so that the the percolated water is added to the layer below the day after it's percolated from the layer above rather than the same day (comments added for clarity in comment but removed in actual code):
Screenshot 2023-10-02 at 11 10 19 AM
2. Updating unit testing for the `percolate()` method.

Notes:
There was previously only integration testing for the `percolate()` method and while @Isgotkowitz and I tried to dissect the logic, we couldn't quite figure it out. I wrote a new unit test more specifically for the `percolate()` method and tested whether it achieved the same level of coverage for percolation.py if the integration test was removed and coverage held steady at 100%. I don't know what other .py files the `test_percolate()` integration test may affect though so I am leaving it in place for the time being. If reviewers would prefer it be removed because it's not certain what it's testing, please comment that.


RUFAS/routines/field/soil/percolation.py
```diff
@@ -58,7 +58,9 @@ def percolate(self, has_seasonal_high_water_table: bool) -> None:
                 percolated_water = self._percolate_between_layers(self.data.time_step, current_layer, layer_below)
                 current_layer.water_content -= percolated_water
                 current_layer.percolated_water = percolated_water
-                layer_below.water_content += percolated_water
+                if layer_number > 0:
+                    layer_above = self.data.soil_layers[layer_number - 1]
+                    current_layer.water_content += layer_above.percolated_water
```
Isgotkowitz, line 63: This works for all the soil layers contained in `self.data.soil_layers`, but water doesn't percolate into the vadose zone layer.

Isgotkowitz (changes requested):
This is a good start but the vadose zone layer needs to be percolated into.

Isgotkowitz:
> Notes:
There was previously only integration testing for the percolate() method and while @Isgotkowitz and I tried to dissect the logic, we couldn't quite figure it out. I wrote a new unit test more specifically for the percolate() method and tested whether it achieved the same level of coverage for percolation.py if the integration test was removed and coverage held steady at 100%. I don't know what other .py files the test_percolate() integration test may affect though so I am leaving it in place for the time being. If reviewers would prefer it be removed because it's not certain what it's testing, please comment that.

@ew3361zh thank you for writing this test. I am working on another bug in percolation right now, and when I finalize the PR for it I will update `test_percolate()` to work in a correct and understandable manner.

RUFAS/routines/field/soil/percolation.py
Isgotkowitz: I think just adding `self.data.vadose_zone_layer.water_content += self.data.soil_layers[-1].percolated_water` after the for loop has finished executing would work.

ew3361zh: I tried it with the line after the loop but the testing wasn't behaving as expected so added the line within the loop based on some conditional logic which I think does very close to what it does outside the loop but behaves more as expected with water_content and percolated_water amounts by layer.

Isgotkowitz (approved):
LGTM. Thank you Niko

matthew7838 (approved):
Niko, nice change and good coverage on the tests.

PR #819, opened 2023-10-03
Refactoring formulate-ration method in ration-manager

ltp23:
This PR refactors the formulate_ration() method in the RationManager class by breaking it up into a handful of smaller static methods.


JoeWaddell (approved):
Nice work breaking this method up: I've tested it locally and it produces results identical to master (which is the idea for this PR!).

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -33,68 +36,363 @@ class RationManager:
         collect the necessary information, which is then sent to RationOptimization
         Finally, RationReporter is a suite of methods used to send information to output manager 
         (and other submodules of RuFaS)
-    
+
     """
+    REDUCTION_FACTOR = 0.25
```
JoeWaddell, line 41: This is now `MILK_REDUCTION_KG` in animal_module_constants.py

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -33,68 +36,363 @@ class RationManager:
...
+    MAX_REATTEMPTS = 5
```
JoeWaddell: We don't want to hardcode a number of max_reattempts here: maybe another constant, but I think instead of flat reattempts, it should be a total reduction in milk kg produced (so it would scale to the `MILK_REDUCTION_KG` value). Definitely difficult to set a default value here too.

ltp23:
Closed because it is too out of date.

PR #806, opened 2023-09-25
Better integrate `flake8` into RuFaS

Isgotkowitz:
This PR contains a configuration file for `flake8`, a script for running `flake8` locally in a convenient and precise way, and a modified GitHub action workflow for linting PRs in a more precise way.

- The `flake8` config file has been added so that linting is done with consistent settings.
- The GitHub action now only lints Python files that have been added or modified in the PR.
  - This was achieved using the [get-changed-files](https://github.com/marketplace/actions/get-all-changed-files) action developed by jitterbit.
- `cleanup.sh` has been added here from #429 and renamed as `lint_branch_changes.sh` (there is also a Batch file version), it does the same thing as the GitHub action - just locally.
  - Also an additional check has been added so if no Python files have been modified in the branch yet then no files are checked.

A wiki page has also been written describing how to use `flake8` in RuFaS - it should be consistent with all the tools/configurations added in this PR.

PR #820, opened 2023-10-03
GitHub Linting Action Quick Fix

Isgotkowitz:
After merging #806 this morning, the newly written GitHub Action failed for linting failed on #796 and #807, like this: 
linting_github_action_failure
I don't totally understand this issue, but I'm pretty sure it has something to do with writing a multi-line string into the GITHUB_ENV variable (there is actually a dedicated way to do [this](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings), but I didn't use this in the fix).

My solution is to reformat the output of the grep command to be on a single line, using a second call to `tr`.
I have tested it on 796 and 807 and it has worked as expected:

796:
796_linting_fix_demo

807:
807_linting_fix_demo

### DIFF
--- a/.github/workflows/lint_flake8.yml
+++ b/.github/workflows/lint_flake8.yml
@@ -30,7 +30,10 @@ jobs:
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
     - name: Get all added or modified files
       run: |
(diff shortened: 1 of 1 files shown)


ew3361zh (approved):
Fix worked for me on #818. Nice work, Ed!

Isgotkowitz:
> Fix worked for me on #818. Nice work, Ed!

@ew3361zh Thank you for testing it out:)

PooyaHekmati (approved):

PR #822, opened 2023-10-04
Update `.gitignore`

Isgotkowitz:
Changes the gitignore to `RUFAS/scenario_manager.py` instead of `input/scenario_manager.py`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,7 @@
 build
 others
 
-input/scenario_manager.py
+RUFAS/scenario_manager.py
 
 # Ignore contents of output directories, except for .keep files
 # This is because we want to keep the output directories, and git


ew3361zh (approved):
Thank you, Ed, nice catch! Good to note that the auto-refactoring that VS Code does when you move files around doesn't update the .gitignore file!

Just a note here that in my check-in with @PooyaHekmati earlier this week, we both confirmed that .gitignore wasn't behaving as expected so files that were already tracked by github were able to be changed and those changes were being tracked by git. The expected behavior is that with a specific file named in the .gitignore, any local changes to it would not be tracked by git. So we may see some changes to this file tracked by git and accidentally pushed to github despite it explicitly being in the .gitignore file.

Isgotkowitz:
> The expected behavior is that with a specific file named in the .gitignore, any local changes to it would not be tracked by git. So we may see some changes to this file tracked by git and accidentally pushed to github despite it explicitly being in the .gitignore file.

Well that is not ideal. Hopefully it will not be too bad to deal with.

PooyaHekmati (approved):

PR #824, opened 2023-10-04
Clamp `soil_water_factor` in `SoilData`

Isgotkowitz:
This PR clamps the value returned for the soil water factor by `SoilData` between 0.0 and 1.0, which is the range of acceptable values for it as defined in the SWAT theoretical documentation pages 333 - 334.

It also refactors the test to pass the inputs _as well as_ the expected output.


ew3361zh (approved):
Nice fix, Ed. Was working for all 7 metadata paths for me. Testing looks good as well.

Do you think it's important at all to note if/when the unclamped_water_factor is > 1.0? As in, is it helpful information that's telling about something else going on in the model? If so, maybe adding a warning/log to OM would be worthwhile.

Isgotkowitz:
> Nice fix, Ed. Was working for all 7 metadata paths for me. Testing looks good as well.
> 
> Do you think it's important at all to note if/when the unclamped_water_factor is > 1.0? As in, is it helpful information that's telling about something else going on in the model? If so, maybe adding a warning/log to OM would be worthwhile.

@ew3361zh Good question, my thought is no. I was looking at some of the input values, and the soil water factor will get above 1 when soil water is at pretty reasonable/normal levels so I don't think it is important to warn users when it gets clamped.

That said, the final answer on this should come from a SME.

JoeWaddell (approved):
Elegant fix! This works for all files I tested, and while I'm not a S&C SME, I agree with Ed's conclusion about not reporting values that fall outside of the 0-1 range. My intuition is that abnormal moisture values should be reported to OM at an earlier step.

Isgotkowitz:
Thanks @JoeWaddell

Issue #737, opened 2023-08-25
Create user option for data validation without simulation running


ew3361zh:
In the Input Manager design document there is discussion of a feature to allow a user to use the IM's data validation feature without running a simulation after. Currently the IM allows this feature via the early-termination flag but only if invalid-data is found during the process. There should be an option to not run the simulation even if all data is found to be valid/fixable.

**Describe the solution you'd like**
The first thought I had was creating a separate termination flag for the user to select whether they wanted to not run the simulation after data validation is complete. This will allow the early-termination flag to function as intended with a separate mechanism to control whether the simulation should run after data validation has run.

**Describe alternatives you've considered**
A separate issue #738 was created for the option for the user to turn validation off and go directly into the simulation which could be paired with the solution to this issue. Via Pooya, "there is conversation going on about adding new features to IM/adding a new module upstream to it. If we decide to do that, both of these validation things can be done there."

ew3361zh:
Accomplished with PR #825.

PR #825, opened 2023-10-04
Validate data no simulation

ew3361zh:
This PR seeks to address #737 in which a user can utilize the InputManager's data validation capacity without running a simulation afterward.

1. Adds flag `terminate_after_validation` (tav) along with gnu arg for said flag. Default is set to False.
2. Updates logic in main.py `execute_simulations()` to include this flag's value.
3. Dumps nondata pools but doesn't save variables in case of tav. This was tested and variables pool was empty which makes sense given all the variables are added during a simulation.
4. Updates unit testing accordingly.

Notes:
- The default for exclude_info_maps was set to True in master but since the gnu arg flag is to set it to True, I changed the default back to False. This may have been accidentally changed on a local branch and merged into master.
- I also added an option for the user to control the early_termination flag from the gnu args.
- I'm differentiating early_termination and terminate_after_validation. Early_termination ends the validation process if the Input Manager encounters unfixable invalid critical data and sends back a False bool to main.py which will prevent a simulation from being run. Terminate_after_validation is the option to not run a simulation even after a full validation/fixing of the data by Input Manager has been achieved, regardless of the value of the early_termination flag.


main.py
```diff
@@ -23,7 +23,9 @@ def run_rufas(
     make_graphs: bool = True,
     verbose: bool = True,
     clear_output: bool = False,
-    exclude_info_maps: bool = True,
+    exclude_info_maps: bool = False,
+    early_termination: bool = True,
```
Fjordrunner: This is definitely nitpicky but I prefer including a verb in most Boolean names...the first suggestions that come to mind are `is_early_termination` or `terminate_early`. I think that `terminate_early` is better here since it matches the following variable, `terminate_after_validation`. This applies to everywhere else where this variable is used, including the test file in case we do go through with this.

main.py
```diff
@@ -113,6 +122,18 @@ def parse_gnu_args():
         help="Exclude info_maps from the output",
         action="store_true",
     )
+    parser.add_argument(
+        "-et",
+        "--early_termination",
+        help="Terminate data validation if invalid critical data found",
```
Fjordrunner: Maybe `early_termination` isn't actually the best variable name here since it doesn't shine light on termination occurring early because of invalid critical data?

Fjordrunner (changes requested):
Overall, this looks really good: thanks for the manageable PR size! My comments are somewhat nitpicky but the more I read into the new code, the less I liked that variable name, happy to have a discussion on it though.

Isgotkowitz (approved):
LGTM.

Fjordrunner:
@ew3361zh @PooyaHekmati I'm not the biggest fan of the variable name "early_termination," as I highlight above. The most striking issue here is that early termination doesn't have anything to do with users themselves terminating runs early. I feel as if the name suggests that a user is manually exiting or aborting a run via an early termination, as opposed to there being missing critical data. If this Boolean variable is related to there being invalid critical data, I suggest a different name entirely with the focus being on the absent critical data. I propose the name "invalid_data_found" or "invalid_critical_data_found," because it can be reasonably implied that if such data is found, the program would terminate early.

ew3361zh:
> @ew3361zh @PooyaHekmati I'm not the biggest fan of the variable name "early_termination," as I highlight above. The most striking issue here is that early termination doesn't have anything to do with users themselves terminating runs early. I feel as if the name suggests that a user is manually exiting or aborting a run via an early termination, as opposed to there being missing critical data. If this Boolean variable is related to there being invalid critical data, I suggest a different name entirely with the focus being on the absent critical data. I propose the name "invalid_data_found" or "invalid_critical_data_found," because it can be reasonably implied that if such data is found, the program would terminate early.

What about something like `terminate_early_on_invalid_data`?

PooyaHekmati (changes requested):
I agree that `early_termination` is not a good choice as it is 1) ambiguous, and 2) similar to `eager_termination` that is used in input manager. You can rename the GNU arg to something like `only_run_validation` or something of that nature. Users care about what the code does, not what it doesn't. 
Also, if this arg is true, then we won't run the simulation. The question is, why do we need to call `execute_simulations()` when we are not going to run the simulation? This doesn't seem right.

ew3361zh:
Updates:

- [x] `early_termination` renamed to `eager_termination` to match its use elsewhere and for clarity.
- [x] `eager_termination `removed from gnu arguments - not something users should be changing in gnu args.
- [x] `terminate_after_validation` -> `only_run_validation`
- [x] `only_run_validation` being True triggers call to new function in main.py `run_validation()` which just runs data validation.
- [x] `execute_simulations()` only called if `only_run_validation` is False.
- [x] Unit testing updated.

PooyaHekmati (approved):
[LGTM](https://lgtm.lol/i/714)

PooyaHekmati:
The code looks good, all comments have been resolved; and the PR has the needed approvals. Overriding the requested changes to get this PR merged and unblock other PRs that are waiting for this.

PR #827, opened 2023-10-05
Correct Barnyard simulation inputs

Isgotkowitz:
Changes the inputs for the barnyard simulation to match what is in Pete's Excel Spreadsheet model.

The inputs should match those on the "PARAMETERS" sheets of "__model_PlantGrowth_3.5 sand barnyards.xlsx" and "__model_PlantGrowth_3.5 soil barnyards 2.xlsx" (in [this](https://3.basecamp.com/3486446/buckets/5296287/vaults/2788737073) folder on Basecamp).

Note that the inputs in here will probably need to be evolve as the Crop and Soil module evaluation progresses, but these are correct enough for now that they can be compared to the output of the Excel model.


allisterakun (approved):
Looks good

ew3361zh (approved):
Looks GTM, Ed. 

Curious about the intention and result behind changing the blob names in the metadata though - can you speak to that?

Isgotkowitz:
> Looks GTM, Ed.
> 
> Curious about the intention and result behind changing the blob names in the metadata though - can you speak to that?

@ew3361zh sure. Short answer: soil blobs were changed to more clearly describe what differentiates them, same with field blobs, and now only one input each is being used for tillage, fertilizer, and manure application so they don't need to be numbered.

Long answer: Pete's model has two versions because there are two soil profiles, one with "regular" soil and one with soil that is really sandy. The output of the two model runs can be compared to show the effect of different soil profiles under the exact same management/environmental conditions. We want to do the same thing in RuFaS, so we need two fields that _only_ differ in soil profiles. Renaming the blobs makes this clearer, and the Crop and Soil module uses blob names to construct output prefixes so it makes the output clearer as well.

ew3361zh:
> > Looks GTM, Ed.
> > Curious about the intention and result behind changing the blob names in the metadata though - can you speak to that?
> 
> @ew3361zh sure. Short answer: soil blobs were changed to more clearly describe what differentiates them, same with field blobs, and now only one input each is being used for tillage, fertilizer, and manure application so they don't need to be numbered.
> 
> Long answer: Pete's model has two versions because there are two soil profiles, one with "regular" soil and one with soil that is really sandy. The output of the two model runs can be compared to show the effect of different soil profiles under the exact same management/environmental conditions. We want to do the same thing in RuFaS, so we need two fields that _only_ differ in soil profiles. Renaming the blobs makes this clearer, and the Crop and Soil module uses blob names to construct output prefixes so it makes the output clearer as well.

Thanks, Ed, that makes sense. I didn't know the blob names informed output prefixes for the C&S module. Was mostly thinking changing the blob names would confuse calls to the IM's `get_data()`. Something we may want to keep in mind down the line as IM gets more specifically incorporated into the model.

Isgotkowitz:
> Thanks, Ed, that makes sense. I didn't know the blob names informed output prefixes for the C&S module. Was mostly thinking changing the blob names would confuse calls to the IM's get_data(). Something we may want to keep in mind down the line as IM gets more specifically incorporated into the model.

Definitely something to keep in mind. For better or for worse, the Field module currently operates totally independently of what the blob names are for field inputs (sort-of), and this may be a desirable outcome to try and achieve with other modules.

Issue #817, opened 2023-09-29
Output files named by user in Scenario Manager.


ew3361zh:
**Is your feature request related to a problem? Please describe.**
In the forthcoming/new method of running simulations, it's possible for users to run multiple simulations sequentially and we need a way to more-easily differentiate the output files from each sequential simulation.

**Describe the solution you'd like**
Converting the METADATA_PATHS variable in `scenario_manager.py` from a list of string paths to metadata files to a list of tuples that would hold the string path to the metadata file and the name the user would like to use and have incorporated at the front of the file names generated from the simulation by that particular metadata file.

e.g: METADATA_PATHS = [('large_farm_metadata_path.json', 'large_farm'), ('small_farm_metadata_path.json', 'small_farm')]

Which would generate variables output files (for example), that looked like:
"large_farm_saved_variables_json..."
"small_farm_saved_variables_json..."

**Describe alternatives you've considered**
@JoeWaddell and I briefly discussed using the metadata file names instead of custom names.

PooyaHekmati:
Generally speaking, it is better to use named tuples or typed dictionaries rather than tuples or dicts. They provide cleaner and more informative code; containing more information withing the code itself which reduces the need for documentation.

ew3361zh:
Addressed by PR #833.

Issue #211, opened 2022-11-08
Ability to suppress messages (global verbose=False)


morrowcj:
RuFaS should have an option to not print any basic status messages during a run (e.g., "running simulation", "creating graphics", etc). 

This feature would greatly help when running RuFaS in batch mode. 

There currently exists a `PRINT_STATUS_MESSAGES` global variable (`config/global_variables`) but it needs to be implemented. Perhaps there could also be levels of verbosity: 0 for no messages, 1 for only warnings or important abnormal behavior, 2 for all messages.

morrowcj:
Similarly, we should be able to specify where output files are saved (`OUT_DIR`)

ew3361zh:
Addressed with PR #828

PR #828, opened 2023-10-05
Trial CLI messaging

ew3361zh:
With recent changes to how IM data validation occurs via #807 in which more of the validation process is logged to OM vs raising errors, the user can potentially run multiple sequential scenario-simulations and not be sure if one of them failed unless they go to the nondata logs. This also ties in with issue #817 in which the user has somewhat limited abilities to discern between output files in a multi-scenario situation.

This PR seeks to address these issues by adding a small number of sys.stdout.write statements that update the user as to which metadata file is being run in what order and whether the result of that validation process is valid/invalid.

This PR will also implement the verbose gnu arg referenced in issue #211.

To test:
1. Run a simulation with the verbose option: `python main.py -v` or
2. Run just a validation of the data with the verbose option: `python main.py -v -ov`


PooyaHekmati (changes requested):
This should consider the -v or --verbose gnu arg. Line 98 in main.py on master.

JoeWaddell (changes requested):
This is a nice improvement!  

Only note: 'Data is valid' will show even if IM had to correct input data. I think it could be useful (though maybe outside the scope of this PR) if instead of 'Data is valid' it printed something like 'Invalid data was fixed successfully.', so the user is reminded to check the warnings file.

This would of course have to avoid the 'OutputManager.save_variables.invalid filter file' warning for the defaults (I suppose that check can be updated to not flag those defaults in the warnings pool).

ew3361zh:
> This is a nice improvement!
> 
> Only note: 'Data is valid' will show even if IM had to correct input data. I think it could be useful (though maybe outside the scope of this PR) if instead of 'Data is valid' it printed something like 'Invalid data was fixed successfully.', so the user is reminded to check the warnings file.

Thanks, Joe, I've made some changes towards this end. Maybe try rerunning the model now both with only_run_validation `python main.py -ov` run and a standard run. But maybe intentionally break some fields both those with defaults and those without to see the different messaging. 

One field without a default I found is in default_animal.json>animal_config>from_literature>life_cycle>still_birth_rate

> This would of course have to avoid the 'OutputManager.save_variables.invalid filter file' warning for the defaults (I suppose that check can be updated to not flag those defaults in the warnings pool).

I'm not sure I understand this part - could you expand a little?

JoeWaddell:
>>    This would of course have to avoid the 'OutputManager.save_variables.invalid filter file' warning for the defaults (I suppose that check can be updated to not flag those defaults in the warnings pool).

> I'm not sure I understand this part - could you expand a little?

We discussed this in person, but for clarification for others: this was just so the command line messages won't report that default output filters (the `example` or those with underscores, e.g. ignored filters) are erroneous...which appears to be fixed already. 

A quick look at the way this works seems great: however it shows 0 elements fixed, even after showing that it did fix something. I might be missing something though. I like the general "Check logs for more detailed data validation info." message, since that covers all cases at once. Otherwise it's exactly what I had in mind, so nice job!

image

ew3361zh:
> > > This would of course have to avoid the 'OutputManager.save_variables.invalid filter file' warning for the defaults (I suppose that check can be updated to not flag those defaults in the warnings pool).
> 
> > I'm not sure I understand this part - could you expand a little?
> 
> We discussed this in person, but for clarification for others: this was just so the command line messages won't report that default output filters (the `example` or those with underscores, e.g. ignored filters) are erroneous...which appears to be fixed already.
> 
> A quick look at the way this works seems great: however it shows 0 elements fixed, even after showing that it did fix something. I might be missing something though. I like the general "Check logs for more detailed data validation info." message, since that covers all cases at once. Otherwise it's exactly what I had in mind, so nice job!

Thanks, Joe. After doing some debugging, it looks like the recursive nature of the `validate_json_element()` method is causing the tracking system of elements to be reset mid-validation process. I will create an issue about this and look into it further on Monday.

JoeWaddell (approved):
Tested again: LGTM! Nice work

YijingGong (approved):
I tested several invalid reproduction protocol inputs and invalid detection/conception rates, this PR is able to print the warning and fix that to default values, which is great! 
Perfectly solved the concern that I have in PR #858. Great work, Niko.

RUFAS/input_manager.py
```diff
@@ -199,6 +205,9 @@ def _populate_pool(self, eager_termination: bool) -> bool:
         om.add_log("Total Checked Items", f"{total_elements_counter=}", info_map)
         om.add_log("Total Fixed Items", f"{fixed_elements_counter=}", info_map)
         om.add_log("Total Invalid Items", f"{invalid_elements_counter=}", info_map)
+        if config.global_variables.PRINT_STATUS_MESSAGES:
+            sys.stdout.write(f"{fixed_elements_counter} element(s) fixed during the validation process.\n")
+            sys.stdout.write(f"{invalid_elements_counter} element(s) found invalid and unfixable.\n")
```
```suggestion
            sys.stdout.write(f"{fixed_elements_counter} element(s) fixed during the validation process.\n")
            sys.stdout.write(f"{invalid_elements_counter} invalid and unfixable element(s) were found.\n")
```

RUFAS/input_manager.py
```diff
@@ -289,7 +300,7 @@ def _validate_input_type_dynamic(self, variable_properties: Dict[str, Any], var_
         return validator(variable_properties, var_name, input_data_value)
 
     def _validate_csv_element(self, var_name: str, properties_blob_key: str, input_data: Dict[str, Any],
-                              eager_termination: bool) -> dict:
+                              eager_termination: bool, element_counter_and_validity: Dict[str, Any]) -> dict:
```
PooyaHekmati: Why is `element_counter_and_validity` Any? It's either bool or int. It seems the same problem exists in other functions as well.
```suggestion
                              eager_termination: bool, element_counter_and_validity: Dict[str, int|bool]) -> Dict[str, int|bool]:
```

PooyaHekmati (changes requested):
Looks good overall. 
But `if config.global_variables.PRINT_STATUS_MESSAGES:` is repeated 10 times, and `sys.stdout.write` is repeated 21 times, which is not good. We also have issue #838 . All of these can be resolved by: instead of using `sys.stdout.write`, send a log to OM. In `OM._add_log()` first add the log to the pool, then check if printing is needed, if yes, use `sys.stdout.write` there. By doing this:

1. You can directly pass the gnu arg value to OM and eliminate the global config
1. All logs are in one single place.
1. You are not repeating yourself.
1. Later, we can develop a logger if needs be.
1. All the messages that are being printed here are logs in nature so they have to be in the logs pool as well. It seems some of them are already being reported to the logs pool, you'd just need to report the new one.

Note that it is a good idea to:

1. handle formatting in OM as much as possible. (Instead of sending formatted messages to OM, try to find a way that it can handle the formatting by itself, as much as possible)
1. warnings and errors can benefit from this as well. So maybe convert the verbose arg from bool to something that can capture errors, warnings, and logs. Each level contains the previous levels as well. You may want to create an enum for this. I'd set the default to either error or warning.

ew3361zh:
> Looks good overall. But `if config.global_variables.PRINT_STATUS_MESSAGES:` is repeated 10 times, and `sys.stdout.write` is repeated 21 times, which is not good. We also have issue #838 . All of these can be resolved by: instead of using `sys.stdout.write`, send a log to OM. In `OM._add_log()` first add the log to the pool, then check if printing is needed, if yes, use `sys.stdout.write` there. By doing this:
> 
> 1. You can directly pass the gnu arg value to OM and eliminate the global config
> 2. All logs are in one single place.
> 3. You are not repeating yourself.
> 4. Later, we can develop a logger if needs be.
> 5. All the messages that are being printed here are logs in nature so they have to be in the logs pool as well. It seems some of them are already being reported to the logs pool, you'd just need to report the new one.

I like this idea overall and I think it makes sense. However, part of the request/goal for this PR was to have the specific messages related to data validation printed out for the user as they're running multiple simulations. It sounds like the approach here would effectively be an on/off switch for printing out one of the following: 
1. All errors 
2. All errors and warnings
3. All errors, warnings, and logs

I would worry that even if there were an option to print out just all the logs, we'd have a flood of printed messages that would be hard to sort through for a user looking for something specific like which fields from their input data files were "fixed" or were invalid-critical data.

Do you have any thoughts on how we might be able to align it with the original goals of this PR centered around solely data validation and simulation messages being printed in the CL? 

Or is a verbose gnu arg not the proper avenue to take for something specific as printing out messages related to just the data validation process and we should look for another avenue?

PooyaHekmati:
> > Looks good overall. But `if config.global_variables.PRINT_STATUS_MESSAGES:` is repeated 10 times, and `sys.stdout.write` is repeated 21 times, which is not good. We also have issue #838 . All of these can be resolved by: instead of using `sys.stdout.write`, send a log to OM. In `OM._add_log()` first add the log to the pool, then check if printing is needed, if yes, use `sys.stdout.write` there. By doing this:
> > 
> > 1. You can directly pass the gnu arg value to OM and eliminate the global config
> > 2. All logs are in one single place.
> > 3. You are not repeating yourself.
> > 4. Later, we can develop a logger if needs be.
> > 5. All the messages that are being printed here are logs in nature so they have to be in the logs pool as well. It seems some of them are already being reported to the logs pool, you'd just need to report the new one.
> 
> I like this idea overall and I think it makes sense. However, part of the request/goal for this PR was to have the specific messages related to data validation printed out for the user as they're running multiple simulations. It sounds like the approach here would effectively be an on/off switch for printing out one of the following:
> 
> 1. All errors
> 2. All errors and warnings
> 3. All errors, warnings, and logs

Correct
The rationale is that if there are errors or warnings, the user needs to be aware of them.

> I would worry that even if there were an option to print out just all the logs, we'd have a flood of printed messages that would be hard to sort through for a user looking for something specific like which fields from their input data files were "fixed" or were invalid-critical data.
>
> Do you have any thoughts on how we might be able to align it with the original goals of this PR centered around solely data validation and simulation messages being printed in the CL?

I understand your point. If you format the output like:
```
[2023-10-26 13:45:32] [ERROR]: <name>: <msg>
[2023-10-26 13:45:32] [WARNING]: <name>: <msg>
[2023-10-26 13:45:32] [Log]: <name>: <msg>
```
it will be clean and clear, and the name can be used to find the relevant piece of information. For example, if all validation logs start with `validation_` then the user can simply use ctrl+f (cmd+f) to search for validation logs.

> Or is a verbose gnu arg not the proper avenue to take for something specific as printing out messages related to just the data validation process and we should look for another avenue?

The right answer is to have a dedicated logging system that can handle such cases. However, given other priorities, I think that is a nice thing to do in the future.

ew3361zh:
- [x] LogType Enum class created in OM.py - used to help set which logs are printed during a simulation (logs, errors, warnings, or none)
- [x] `v=errors` only prints errors, `v=warnings` prints warnings and errors, `v=logs` prints logs, warnings, and errors, `v=none` is the default and doesn't print out any warnings, errors, or logs.
- [x] verbose gnu arg now has options based on the LogType class
- [x] data validation-related messages include "validation" in their name for easier searching
- [x] printed logs/errors/warnings have the metadata prefix at their start to help for easier searching
- [x] updated unit testing

RUFAS/output_manager.py
```diff
@@ -13,6 +15,24 @@
 from RUFAS.util import Utility
 
 
+class LogType(Enum):
```
PooyaHekmati: can we rename this class to LogVerbosity?

main.py
```diff
@@ -54,26 +54,23 @@ def run_rufas(
         keep_list = [".keep", "output_filters"]
         Utility.empty_dir(output_dir, keep=keep_list)
 
-    set_global_variables(make_graphs, verbose)
-    if verbose:
-        print("RuFaS: Ruminant Farm Systems Model 2023")
+    set_global_variables(make_graphs)
+    if verbose != LogType.NONE:
+        sys.stdout.write("RuFaS: Ruminant Farm Systems Model 2023\n")
```
PooyaHekmati: This leads to a behavior like this:
image
which implies that `RuFaS: Ruminant Farm Systems Model 2023` is an error. It is also inconsistent with `Simulating` and `Simulation Successful` messages. I think we want to print out the `RuFaS: Ruminant Farm Systems Model 2023` anyways.

main.py
```diff
@@ -90,18 +87,23 @@ def run_validation(metadata_files: List[Path], exclude_info_maps: bool = False,
                 }
     output_manager = OutputManager()
     input_manager = InputManager()
+    output_manager.add_log("Validation only", "***Only validating data, no simulation will follow.***", info_map)
     for metadata_file in metadata_files:
         input_manager.flush_pool()
         output_manager.flush_pools()
+        output_manager.add_log("Validation start.", f"Validating data for {str(metadata_file['path'])}...\n", info_map)
```
PooyaHekmati:
```suggestion
        output_manager.add_log("Validation start", f"Validating data for {str(metadata_file['path'])}...\n", info_map)
```

main.py
```diff
@@ -124,13 +126,16 @@ def execute_simulations(
...
```
PooyaHekmati:
```suggestion
        output_manager.add_log("Validation start", f"Validating data for {str(metadata_file['path'])}...\n", info_map)
```

RUFAS/output_manager.py
```diff
@@ -144,6 +165,8 @@ def add_log(self, name: str, msg: str, info_map: Dict[str, Any]) -> None:
         info_map["timestamp"] = self._get_timestamp(include_millis=True)
         key = self._generate_key(name, info_map)
         self._add_to_pool(self.logs_pool, key, msg, info_map)
+        if self.__log_type == LogType.LOGS:
+            sys.stdout.write(f"{self.__metadata_prefix}[LOG]: {name}: {msg}\n")
```
PooyaHekmati: Applies to other instances as well.
```suggestion
            sys.stdout.write(f"[Time Stamp][LOG][{self.__metadata_prefix}] {name}: {msg}\n")
```

RUFAS/output_manager.py
ew3361zh: Definitely.

main.py
ew3361zh: Moved this to print every time.

RUFAS/output_manager.py
ew3361zh: Addressed with logs, errors, and warnings.

RUFAS/output_manager.py
```diff
@@ -14,6 +16,43 @@
...
+    Selecting WARNINGS will tell OutputManager to print out all warnings and errors added during a simulation.
+    Selecting LOGS will tell OutputManager to print out all logs, warnings, and errors added during a simulation.
+    """
+
+    NONE = "none"
+    ERRORS = "errors"
+    WARNINGS = "warnings"
+    LOGS = "logs"
+
+    def __le__(self, other) -> bool:
```
ew3361zh, line 37: I like this - really helpful comparison method.

RUFAS/output_manager.py
```diff
@@ -211,11 +253,38 @@ def add_error(self, name: str, msg: str, info_map: Dict[str, Any]) -> None:
         info_map["timestamp"] = self._get_timestamp(include_millis=True)
         key = self._generate_key(name, info_map)
         self._add_to_pool(self.errors_pool, key, msg, info_map)
+        self._handle_log_output(name, msg, info_map, LogVerbosity.ERRORS)
+
+    def _handle_log_output(
+        self, name: str, msg: str, info_map: Dict[str, Any], log_level: LogVerbosity
+    ) -> None:
+        colors: Dict[LogVerbosity, str] = {
```
ew3361zh, line 274: I like the addition of the colors - especially at the LOGS level with everything being printed out.

ew3361zh:
@PooyaHekmati I really like your additions here - I think it produces clearer output for the different types of log messages which will be helpful especially when users are running multiple scenarios.

I added docstrings for the _handle_log_level method - please feel free to amend as needed.

And thank you from the bottom of my heart for handling the merge conflicts :).

RUFAS/output_manager.py
PooyaHekmati: I forgot to add unit tests for it though

RUFAS/output_manager.py
PooyaHekmati: Maybe in the future, we allow different color pallets or different formatting styles. But for now, I think this is enough.

PooyaHekmati:
> @PooyaHekmati I really like your additions here - I think it produces clearer output for the different types of log messages which will be helpful especially when users are running multiple scenarios.
> 
> I added docstrings for the _handle_log_level method - please feel free to amend as needed.

I think the only thing left here to do is the unit test for `__le__`, which shouldn't be hard to do. 
Also, now that `LogVerbosity` is being used for `log_level` as well, I think a more general name like `LogLevel` is probably more appropriate. It can be done in this PR or in a future one.

> And thank you from the bottom of my heart for handling the merge conflicts :).

You're welcome :)

RUFAS/output_manager.py
ew3361zh: It's interesting we're still at 100% coverage for output_manager.py
Screenshot 2023-11-06 at 4 20 27 PM

PooyaHekmati (approved):

PR #829, opened 2023-10-05
Fix water buildup in top soil layer

### LINKED ISSUE #468: `soil.Infiltration` submodule does not check bounds for runoff or store it clearly
# Context
In the `Infiltration` submodule of `soil`, the amount of rainfall that runs off the field that day is calculated and stored as `accumulated_runoff` in its associated `SoilData` instance so that it can be easily accessed by other modules.

# Problem
1. `Infiltration` does not check that the amount of runoff it calculates is <= the amount of rainfall on that day.
2. The description of `accumulated_runoff` in `SoilData` does not indicate that it is the daily measure of runoff from the field.

# Solution
Store `accumulated_runoff` as follows (in `infiltration.py`):
```
self.data.accumulated_runoff = min(rainfall, self._determine_accumulated_runoff(rainfall, retention_parameter))
```
Update the docstring for `accumulated_runoff` in `SoilData` to clearly indicate it is a daily measure, and more explicitly what it *is*.

Isgotkowitz:
This PR addresses the large buildup of water that was occurring in the surface soil layer. The most substantive changes are in the Percolation module, which makes two changes:
1. Allows percolation even when soil temperature is frozen.
2. Reverses the order in which percolation occurs in the soil profile (now works from bottom to top).

The new percolation method has had it's unit/integration test rewritten following a much better testing structure.

Also, the infiltration module was closely checked during the process of debugging this issue and while it's overall behavior was correct it actually had to different methods of calculating infiltration implemented. So the extra method was removed and the interaction with `Infiltration` changed in `Field` as well. Tests have been changed accordingly. Some important variables are also now tracked in `SoilData` and recorded to the `OutputManager`.

Closes #468 

Before:
download

After:
download


JoeWaddell (approved):
The reversal in the percolation process is so that the lowest layers percolate down first, to ensure there's not an unexpected "backup" of water on a layer that could otherwise percolate down, right?

Otherwise: LGTM!

Isgotkowitz:
> The reversal in the percolation process is so that the lowest layers percolate down first, to ensure there's not an unexpected "backup" of water on a layer that could otherwise percolate down, right?

@JoeWaddell yes, exactly. I have not tested infiltration/percolation with extreme weather conditions yet (hurricane levels of precip), but it seems to do a good job of not getting backed up over multiple days of high infiltration so far.

RUFAS/routines/field/soil/percolation.py
```diff
@@ -33,14 +33,18 @@ def percolate(self, has_seasonal_high_water_table: bool) -> None:
...
+
         References
         ----------
         SWAT sections 2:3.1 and 2
         """
         layer_count = len(self.data.soil_layers)
         deepest_layer = layer_count - 1
 
-        for layer_number in range(layer_count):  # loop through each layer
+        for layer_number in reversed(range(layer_count)):  # loop through each layer
```
matthew7838: Either remove this comment or state that it's looped in a reversed way.

matthew7838 (approved):
LGTM, good changes. I like the testing improvements.

RUFAS/routines/field/soil/percolation.py
Isgotkowitz: Good catch

PR #830, opened 2023-10-08
Carbon Cycling Integration

### LINKED ISSUE #512: Revisit Carbon Cycling submodule of Soil and Crop
# Problem
The v1 version of Crop and Soil uses a carbon cycling sub-module that is based upon the old code (prior to SC_redesign). The module is documented in the old soil pseudocode, but the primary sources are unclear. I believe that it is a novel model developed by Hector M. based upon DAYCENT. It is unclear to me if the current carbon module is the best approach for RuFaS. Along with the other subject-matter experts (Xuesong, Pete Vadas), I believe that a revision is warranted. 

# Solution
The Soil and Crop team believes that the carbon functionality of SWAT-C (developed by Xuesong) should replace the current carbon submodule. The implementation should conform to the standards set by the other SC modules.

matthew7838:
--- Edited by Ed ---
This PR implements most of the functionality of the carbon cycle, and the outputs produced by the Carbon Cycling module on this branch have been reviewed and approved by SMEs. There are two remaining pieces of functionality that are needed to fully integrate carbon with the rest of RuFaS, documented in issues #944 and #945.

This PR addresses two main issues that were preventing the carbon cycling module from running as expected:
1) Carbon (or more accurately, plant residue) was not being transferred appropriately from the Crop Management module to the Carbon Cycling module.
2) The carbon decomposition in the surface soil layer was not being appropriately distinguished from the carbon decomposition in the subsurface soil layers.

The first issue has been addressed through enhancing the `_transfer_residue()` method in the Crop Management module, and adding the `add_residue_to_pools()` methods and subroutines in the Residue Partition module of Carbon Cycling. The main routines of both the Residue Partition modules and Pool Gas Partition modules have been refactored to handle the decomposition activities of the surface soil layer differently from the activities of the subsurface soil layers.

All modules touched in this PR maintain their levels of code coverage.

Note that the code clarity and cleanliness of the Carbon Cycling module and related components are not up to the standards we aim for in RuFaS. However, given the current deadlines and the fact that we would like to replace the current carbon cycling module with one based on SWAT-C, it seems inappropriate to try to achieve a super clean Carbon Cycling module in this PR.

Below are some outputs from a 100 year simulation that has one large addition of plant residue, then lets it decompose.

Carbon dioxide emissions (kg / ha) on a daily time step:
image

Metabolic (orange line) and structural (blue line) litter in the surface soil layer (kg / ha) on a daily time step:
image

Slow (blue line), active (orange line), and passive (grey line) carbon amounts in the 4th soil layer (kg / ha) on a daily time step:
image


RUFAS/routines/field/crop/crop_management.py
```diff
@@ -49,19 +49,20 @@ def manage_harvest(self, harvest_op: HarvestOperation, field_name: str, field_si
             The object tracking the attributes of the soil profile.
 
         """
+        killed = False
```
ew3361zh:
```suggestion
        is_killed = False
```
or just `is_killed_by_harvest`
Or `was_killed`/`was_killed_by_harvest`?

RUFAS/routines/field/crop/crop_management.py
```diff
@@ -49,19 +49,20 @@ def manage_harvest(self, harvest_op: HarvestOperation, field_name: str, field_si
...
         self.determine_harvest_index()
 
         if harvest_op == HarvestOperation.HARVEST:
             self.cut_crop(collected_fraction=self.data.harvest_efficiency)
             self.kill()
+            killed = True
```
ew3361zh: same as above

RUFAS/routines/field/crop/crop_management.py
```diff
@@ -49,19 +49,20 @@ def manage_harvest(self, harvest_op: HarvestOperation, field_name: str, field_si
...
 
         if harvest_op == HarvestOperation.HARVEST_NOKILL:
             self.cut_crop(collected_fraction=self.data.harvest_efficiency)
+            # non of the roots gets killed, soil litter not used
```
ew3361zh:
```suggestion
            # none of the roots gets killed, soil litter not used
```

RUFAS/routines/field/crop/crop_management.py
```diff
@@ -253,10 +254,27 @@ def _transfer_residue(self, soil_data: SoilData) -> None:
...
+        Notes
+        -----
+        If a crop is harvested but not killed, then there is only residue added to the surface. If it is harvested and
+        killed, then both surface and root residue is added to the soil profile.
 
         """
-        soil_data.plant_surface_residue += self.data.yield_residue
-        soil_data.soil_layers[0].fresh_organic_nitrogen_content += self.data.yield_nitrogen
+        soil_data.crop_yield_nitrogen = self.data.residue_nitrogen
+        soil_data.plant_residue_lignin_composition = 0.17
```
ew3361zh, line 266: I think it would be helpful to know why this variable is set to 0.17 here. Is this a constant/standard?

ew3361zh (approved):
Tried to keep my comments to code-based since the science is beyond me. I think this looks good overall. A few minor suggestions.

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: Good suggestion

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: Nice catch

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: Excellent question. This was a constant set in the previous version of the carbon cycling module. Clay tried to find where this value came from and how we should really be setting it based on the type of crop that residue is coming from, but he did not turn up any proper answers. However, knowing how much lignin is in a crop is required for the Feed storage module, so we will be verifying what this value should be based on the crop type. Will make an issue outlining this.

RUFAS/routines/field/crop/crop_management.py
```diff
@@ -253,10 +253,27 @@ def _transfer_residue(self, soil_data: SoilData) -> None:
...
+        if killed:
+            soil_data.plant_surface_residue = self.data.yield_residue - self.data.root_biomass
```
JoeWaddell, line 268: Is it possible to have more than one crop in the same soil profile/soil_data object? e.g. multicropping?
If so (or there are future plans), I think it would make sense to use '+=' actions (also assuming there is a reset at some stage.)

RUFAS/routines/field/soil/carbon_cycling/pool_gas_partition.py
```diff
@@ -29,37 +29,83 @@ def __init__(self, soil_data: Optional[SoilData], field_size: Optional[float] =
...
+
         Notes
         ------
         This method applies all the gas-partitioning sub routines to each layer of soil that is present in the soil
         `data` object.
-        """
 
-        for layer in self.data.soil_layers:
+        """
+        layer = self.data.soil_layers[0]
```
JoeWaddell, line 41: You noted in the description that this bringing carbon_cycling up to spec isn't a priority right now, so disregard this suggestion if it would take longer than a couple minutes.

Simply splitting this into a `partition_pool_gas_surface_layer` and `partition_pool_gas_subsurface_layers` (those may not be the correct terms) methods may improve clarity, and presumably wouldn't necessitate writing new tests (just splitting existing tests). Again though...disregard if this is more complex than it appears on the _surface_.

JoeWaddell (approved):

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: Yes, multicropping is currently possible and you raise a good point. The scenario where multiple crops are harvested on the same day has been accounted for, but definitely in a way that is not ideal.

When crops A and B are harvested on the same day (in that order) the following happens:
- Crop A is harvested and some of it is stored as residue in SoilData (what is happening in the code above).
- The `add_residue_to_pools()` method is called, which takes all the residue that was stored in SoilData, adds it to the litter pools, then clears the residue storage pools.
- Crop B is then harvested and some of it is stored as residue in SoilData.
- `add_residue_to_pools()` is called again, adding residue to litter pools and clearing the residue pools.

RUFAS/routines/field/soil/carbon_cycling/pool_gas_partition.py
Isgotkowitz: This is really good idea, but I will hold off on it in the interest of keeping this PR from ballooning even more than it already has. I will coordinate with @KFosterReed and Pete to define exactly what the appropriate levels of cleanliness are for the V1 Carbon Cycling module, and if appropriate/time allows this suggestion will be implemented.

PR #833, opened 2023-10-09
Output file naming

ew3361zh:
This PR seeks to address issue #817 which requests an adjustment to how output file names are generated; to be associated with a particular metadata scenario being run.

Changes:
1. Expands scenario_manager.py to now have a custom TypedDict class for MetadataFiles objects which includes both a prefix the user can assign to associate with a metadata file scenario and a path to a metadata file.
2. Updates main.py to take these new metadata prefix arguments in and sends them to the appropriate OutputManager methods for filesaving.
3. Updates unit testing. Coverage for affected files (scenario_manager.py, main.py, and output_manager.py all at 100%).

For testing:
1. Suggested to try single and multiple metadata files with custom prefixes (in scenario_manager.py).
2. Be sure to remove the underscore in the output_filters _csv_all_variables.txt if you don't have any other csv filter files available.
3. Look at generated output files.


JoeWaddell (approved):
LGTM! Nice feature to have, ran properly when I tested it. Will make it far easier to evaluate RuFaS under a variety of conditions at once.

Isgotkowitz (approved):
Looks good, nice job Niko.

My only suggestion is that the names of the output files could be pared down a bit, cause they are starting to get a bit long. For example if the filter is `csv_dummy_filter_one.txt`, I think it would be better to save the output as `{prefix}_dummy_filter_one_{timestamp}.csv` instead of `{prefix}_saved_variables_csv_dummy_filter_one.txt_{timestamp}.csv`. It may not be feasible/practical to make that change in this PR though, and there may be some good reasons for keeping the output file names more verbose.

ew3361zh:
Changes after approval from @Isgotkowitz and @JoeWaddell requested by @PooyaHekmati:
- [x] `__metadata_prefix` attribute added to OutputManager class
- [x] `set_metadata_prefix()` method added to OM
- [x] updates unit testing
- [x] updates main.py to reflect difference in above set-up with changes in OM

RUFAS/output_manager.py
```diff
@@ -624,14 +629,12 @@ def save_variables(
             if filter_file.startswith("json_"):
                 file_path = os.path.join(
                     save_path,
-                    self._generate_file_name(f"saved_variables_{filter_file}", "json"),
+                    self._generate_file_name(f"{self.__metadata_prefix}_saved_variables_{filter_file}", "json"),
```
PooyaHekmati: Don't change calls to `_generate_file_name`; just use the metadata_prefix in the function itself (line 442 on master)

PooyaHekmati (changes requested):

RUFAS/output_manager.py
ew3361zh: addressed

PooyaHekmati (approved):

PR #844, opened 2023-10-11
update master 2 main

PooyaHekmati:
After renaming the default branch to the main, it is time to update our actions in the workflow.

### DIFF
--- a/.github/workflows/coverage_pytest-cov.yml
+++ b/.github/workflows/coverage_pytest-cov.yml
@@ -5,9 +5,9 @@ name: Coverage
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
   pull_request:
-    branches: [ master ]
+    branches: [ main ]
 
 jobs:
   build:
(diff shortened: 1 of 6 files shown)


Isgotkowitz (changes requested):
`lint_branch_changes.sh` and `lint_branch_changes.bat` need to have master changed to main. Also looks like the CodeFactor workflow is not running.

Isgotkowitz (changes requested):
`lint_branch_changes.bat` still needs to be updated.

PooyaHekmati:
> `lint_branch_changes.bat` still needs to be updated.

I had changed but it was not saved. It should be good now.

PooyaHekmati:
Forcing the merge as it is an urgent case.

Isgotkowitz (commented):
Retroactive approval. Seems like everything is in good working order except the CodeFactor test

PooyaHekmati:
> Retroactive approval. Seems like everything is in good working order except the CodeFactor test

That needs to be fixed in codefactore. I think we should depreciate it anyways.

Issue #813, opened 2023-09-29
Variable Name Not Matching 


tomhuhh:
I noticed that the variable names in the variable_name.txt file do not match with the column names in the output cvs file. 

## Expected Behavior

I would expect the names are matching. 

## Current Behavior

For example, in the variable_name.txt, the variable name is 'LifeCycleManager.daily_update.life_cycle_daily_herd_update.values: num_cow_for_parity', but in the csv file, the corresponding column name is 'LifeCycleManager.daily_update.life_cycle_daily_herd_update.num_cow_for_parity' instead.

## Possible Solution

I was able to generate a list of column names from the output csv file using the code below 

`column_names = df.columns`

I was also able to filter the column names in the csv file by searching for a key word using the code below.  
```
# Define the keyword you want to search for
keyword = 'methane'

# Use a list comprehension to find column names containing the keyword
matching_columns = [col for col in df.columns if keyword in col]
```

## Steps to Reproduce

1.
2.
3.
4.

ew3361zh:
@tomhuhh can you describe why it's important to you to have the names in the variable_name.txt file matching the column names in the csv output columns?

tomhuhh:
@ew3361zh Thank you for your question. I certainly don't want to add more work to your plate, but I also aim to streamline the workflow, avoid redundancy, and simplify the process of evaluating RuFaS outputs, enabling us to receive quicker feedback during the development phase and stay on track to meet our milestones. Regarding the 'variable_names.txt' file, could you please remind me of its initial purpose? My challenge has been navigating through the extensive output CSV file, which contains over 1000 columns. The inconsistency in variable names (i.e., column names) makes it more challenging to filter the necessary variables. For instance, I mentioned the 'num_cow_for_parity' variable as an example. I believe that the situation could be improved with the implementation of graphs. However, my understanding is that graph implementation is currently of lower priority. In the meantime, we are expected to do all the post-processing analyses on our own. If I'm mistaken or overlooking any details, please feel free to clarify. Thank you for your hard work!

ew3361zh:
>Regarding the 'variable_names.txt' file, could you please remind me of its initial purpose?

[...]

PR #846, opened 2023-10-11
Gnu save var names formats

ew3361zh:
This PR seeks to address issue #813 in which SMEs are not able to use the variable_names.txt in the capacity they desire to.

The Output Manager had an existing array of formatting options which were previously not easily accessed. At Pooya's suggestion, these formatting options are being tested here as possible solutions to address #813. To help with this process, this PR:

1. Adds gnu args for the 3 formatting options: "verbose", "inline", and "block." The previous default setting for formatting was "verbose" and was kept as the default here as well.
~~2. Adds a function to determine which of the 3 options the user selected and passes that value to the main.py run_rufas() function which eventually gets it to OutputManager for formatting the variable_names.txt file.~~
3. Updates unit testing.

To test out these new format_option gnu args:
1. Run a simulation specifying a formatting option. E.g. for the "inline" formatting option, use the command-line `python main.py -f=inline`. For "block" use `python main.py -f=block`. The default is the "verbose" option so to see this version, the user can either use `python main.py -f=verbose` or `python main.py`.
2. After each simulation, look at the resulting variable_names.txt file in the output folder and see if any of them will fulfill their needs.

If none of these formats fulfills the needs of the SMEs, please specify what you would like and we can work on a different format that does meet those needs.


PooyaHekmati (changes requested):
This is a good starting point, however, the GNU arg shouldn't be boolean. Letting user type in the option makes more sense. Something like 'python main.py --format-option=block'. See https://docs.python.org/3/library/argparse.html#module-argparse 
Also, the convention is that after - only one character follows. For example --debug and -d are the same, but -dbg is not allowed.

PooyaHekmati:
The code looks good and works; but it seems you have merge conflicts. Also, test_user_prompt doesn't sound right; we deprecated user_prompt. It seems this is more like test_main.py.

ew3361zh:
> The code looks good and works; but it seems you have merge conflicts. Also, test_user_prompt doesn't sound right; we deprecated user_prompt. It seems this is more like test_main.py.

Can definitely make that change. And yes, there will be a few merge conflicts with this as well as with #825 once that gets merged in. Thought I would resolve those all at once rather than piecemeal.

One other question. Right now test coverage for main.py is at 96% because of the lines under `if __name__ == "__main__":`. If I move those into a main() function and just have it be:

Screenshot 2023-10-12 at 11 12 38 AM

Screenshot 2023-10-12 at 11 12 49 AM
     
 It could get coverage up to 98%. Not sure if it's worth it but wanted to run that by you. The line main() would be the only remaining uncovered line.

PooyaHekmati:
> > The code looks good and works; but it seems you have merge conflicts. Also, test_user_prompt doesn't sound right; we deprecated user_prompt. It seems this is more like test_main.py.
> 
> Can definitely make that change. And yes, there will be a few merge conflicts with this as well as with #825 once that gets merged in. Thought I would resolve those all at once rather than piecemeal.
> 
> One other question. Right now test coverage for main.py is at 96% because of the lines under `if __name__ == "__main__":`. If I move those into a main() function and just have it be:
> 
> Screenshot 2023-10-12 at 11 12 38 AM Screenshot 2023-10-12 at 11 12 49 AM
> It could get coverage up to 98%. Not sure if it's worth it but wanted to run that by you. The line main() would be the only remaining uncovered line.

#825 has the approvals, what is it waiting for?

Having
```
if __name__ == "__main__":
     main()
```
is a very common practice. If I remember correctly, we had it here as well at some point but apparently, it is not the case anymore. Thanks for checking, I think that is a good thing to do. One other thing that I was thinking is that since you have a fresh memory on GNU args, maybe revisiting those and making sure they are setup correctly is a good idea.

Isgotkowitz (changes requested):
Code LGTM.

Found some stuff in the output from the block option that seems buggy to me:
```
AnimalManager._calc_ration_at_interval.MEdiet_pen_0
                                                   .info_maps: number_animals_in_pen_0
                                                   .info_maps: available_feeds
                                                      .info_maps: number_animals_in_pen_0
                                                      .info_maps: available_feeds
                                                                  .info_maps: number_animals_in_pen_0
                                                                  .info_maps: available_feeds
                                                                  .values: 155
                                                                  .values: 157
                                                                  .values: status
                                                                  .values: objective
                                                                   .info_maps: available_feeds
                                                                   .info_maps: number_animals_in_pen_1
                                                                   .values: dm
                                                                   .values: as_fed
                                                                   .values: CP
                                                                   .values: ADF
                                                                   .values: NDF
                                                                   .values: lignin
                                                                   .values: ash
                                                                   .values: phosphorus
                                                                   .values: potassium
                                                                   .values: N
                                                                   .values: EE
                                                                   .values: starch
```
Looks like some stuff is being duplicated here, and also like some information from pen_1 is mixed in with information from pen_0? I don't know this part of RuFaS well though, so this may be the correct behavior.

Also this:
```
field='field',crop='default alfalfa',planted=89,2010.residue_phosphorus
                                                                       .info_maps: prefix
                           .info_maps: prefix
                           .info_maps: field_size
                           .info_maps: species
                           .values: crop
                           .values: yield
                           .values: nitrogen
                           .values: phosphorus
                           .values: planting_date
                           .values: harvest_date
```
The `field='field',crop='default alfalfa',planted=89,2010.residue_phosphorus` having the info_maps: prefix field below it is correct, but all the info_maps and values that are not indented as far back are missing the associated variable name.
I believe it should look like:
```
field='field',crop='default alfalfa',planted=89,2010.residue_phosphorus
                                                                       .info_maps: prefix
field='field'.harvest_yield
                                                                       .info_maps: prefix
                                                                       .info_maps: field_size
                                                                       .info_maps: species
                                                                       .values: crop
                                                                       .values: yield
                                                                       .values: nitrogen
                                                                       .values: phosphorus
                                                                       .values: planting_date
                                                                       .values: harvest_date

```

Isgotkowitz:
I have a suggestion about how the variable names could be formatted.

In the verbose formatting option, variables that are added to the OutputManager as dictionaries are listed like
```
field='field'.harvest_yield.info_maps: prefix
field='field'.harvest_yield.info_maps: species
field='field'.harvest_yield.info_maps: field_size
field='field'.harvest_yield.values: crop
field='field'.harvest_yield.values: yield
field='field'.harvest_yield.values: nitrogen
field='field'.harvest_yield.values: phosphorus
field='field'.harvest_yield.values: planting_date
field='field'.harvest_yield.values: harvest_date
```
Right now, it is possible to filter out specific values from dictionaries with filters formatted like
```
field='field'.harvest_yield
field='field'.harvest_yield.values: crop
field='field'.harvest_yield.values: yield
field='field'.harvest_yield.values: nitrogen
```
If the above filter was used to write a CSV the names of the columns would be

| field='field'.harvest_yield.crop | field='field'.harvest_yield.yield | field='field'.harvest_yield.nitrogen |
|--|--|--|
"alfalfa" | 5000 | 1100
"corn" | 2500 | 400
... | ... | ...

My suggestion is to have a formatting option for variable names where the format is _exactly_ the same as what is reported in output files and how variables are filtered. For example:
 - the variable is written to the names file as "field='field'.harvest_yield.yield"
 - it can be filtered into an output file using a pattern like "field='.*'.harvest_yield.yield"
 - it is written as a CSV column header/JSON key in an output file as "field='field'.harvest_yield.yield"

Or
 - the variable is written to the names file as "field='field'.harvest_yield: yield"
 - it can be filtered into an output file using a pattern like "field='.*'.harvest_yield: yield"
 - it is written as a CSV column header/JSON key in an output file as "field='field'.harvest_yield: yield"

@tomhuhh and @JoeWaddell please expand on this if there is any information you see missing.

ew3361zh:
> I have a suggestion about how the variable names could be formatted.
> 
> In the verbose formatting option, variables that are added to the OutputManager as dictionaries are listed like
> 
> ```
> field='field'.harvest_yield.info_maps: prefix
> field='field'.harvest_yield.info_maps: species
> field='field'.harvest_yield.info_maps: field_size
> field='field'.harvest_yield.values: crop
> field='field'.harvest_yield.values: yield
> field='field'.harvest_yield.values: nitrogen
> field='field'.harvest_yield.values: phosphorus
> field='field'.harvest_yield.values: planting_date
> field='field'.harvest_yield.values: harvest_date
> ```
> 
> Right now, it is possible to filter out specific values from dictionaries with filters formatted like
> 
> ```
> field='field'.harvest_yield
> field='field'.harvest_yield.values: crop
> field='field'.harvest_yield.values: yield
> field='field'.harvest_yield.values: nitrogen
> ```
> 
> If the above filter was used to write a CSV the names of the columns would be
> 
> field='field'.harvest_yield.crop	field='field'.harvest_yield.yield	field='field'.harvest_yield.nitrogen
> "alfalfa"	5000	1100
> "corn"	2500	400
> ...	...	...
> My suggestion is to have a formatting option for variable names where the format is _exactly_ the same as what is reported in output files and how variables are filtered. For example:
> 
> * the variable is written to the names file as "field='field'.harvest_yield.yield"
> * it can be filtered into an output file using a pattern like "field='.*'.harvest_yield.yield"
> * it is written as a CSV column header/JSON key in an output file as "field='field'.harvest_yield.yield"
> 
> Or
> 
> * the variable is written to the names file as "field='field'.harvest_yield: yield"
> * it can be filtered into an output file using a pattern like "field='.*'.harvest_yield: yield"
> * it is written as a CSV column header/JSON key in an output file as "field='field'.harvest_yield: yield"
> 
> @tomhuhh and @JoeWaddell please expand on this if there is any information you see missing.

Thank you, Ed. I'm working on something in this vein and will have it pushed here Monday for review.

ew3361zh:
@Isgotkowitz @tomhuhh @JoeWaddell 
This update adds the format option called "match" which makes the format of the variable_names.txt file match the column headers for the csv file and doesn't include the info_maps/values data. This version adds too much complexity to the dump_variable_names_and_contexts() method but I wanted to at least run it by everybody before continuing to work towards having this as a viable output format option.

To test: run `python main.py -f=match`

Check the variable_names.txt file and then one of the generated csv files.

Isgotkowitz:
> Check the variable_names.txt file and then one of the generated csv files.

@ew3361zh I like the match format, especially the automatic flattening of the info_maps. The column names seem intuitive enough to me without adding in `.values` and `.info_maps` at the end, so I will approve.

Isgotkowitz (approved):

ew3361zh:
Updates:

- [x] "match" format option renamed to "basic" - to test this format option, run `python main.py -f=basic`
- [x] formatting option removed from `_dict_to_csv_column_list` method
- [x] standard format of csv output file column headers looks like "basic" formatting option (`Class.function.variable`)
- [x] unit testing updated

ew3361zh:
- [x] "block" formatting fixed
- [x] removed "prefix" and "suffix" from adding to variables pools after discussion with @Isgotkowitz (c&s module using OM prefixes most heavily)
- [x] updated unit testing to reflect these changes

RUFAS/output_manager.py
```diff
@@ -81,7 +81,7 @@ def _add_to_pool(
         # the function key; as they are already stored in element key and
         # having them increases the final file size.
         reduced_info_map = {
-            k: info_map[k] for k in info_map.keys() - {"class", "function"}
+            k: info_map[k] for k in info_map.keys() - {"class", "function", "prefix", "suffix"}
```

RUFAS/output_manager.py
```diff
@@ -721,9 +719,9 @@ def dump_errors(self, path: str) -> None:
         file_path = os.path.join(path, self._generate_file_name("errors", "json"))
         self._dict_to_file_json(self.errors_pool, file_path)
 
-    def dump_variable_names_and_contexts(
-        self, path: str, exclude_info_maps: bool, format_option: str = "verbose"
-    ) -> None:
+    def dump_variable_names_and_contexts(self, path: str, exclude_info_maps: bool, # noqa
```
PooyaHekmati: why noqa?

PooyaHekmati (commented):
The code works but why "prefix" are "suffix" added to the reduced info map?
Also, why noqa?

RUFAS/output_manager.py
ew3361zh: In discussion with Ed, it was seen as extraneous information - at this point the prefix and suffix have already been added to the variable name. There wasn't a good reason to add it to the info_maps pool.

RUFAS/output_manager.py
ew3361zh: Will add the mccabe code.

Isgotkowitz:
@tomhuhh @JoeWaddell I want to raise a point that is related to the conversation we had about output variable names and filter files the other day.

With the new `basic` formatting option for output variable names, it is a lot easier to use copy paste to create filter files. But there are still some cases where debugging the Regex filter files may be needed, because of the nature of Regex.

For example, my preferred way of recording variable units is by making changes like the following: `om.add_variable("temperature", ...)` -> `om.add_variable("temperature(C)", ...)`.

The downside to this is that it prevents creating filter files by only copying and pasting from the variable names file. Continuing the above example, the variable names file would have
```
...
Field.soil.temperature(C)
...
```
and the filter file to collect this variable will not work if it has `Field.soil.temperature(C)` (because "(" and ")" are special characters in Regex). Instead it would need to be specified in the filter file as `Field.soil.temperature` or `Field.soil.temperature.*` or `Field.soil.temperature\(C\)`, etc.

I will create an issue related to the above example.

RUFAS/output_manager.py
ew3361zh: After further discussion with Pooya, prefix and suffix were decidedly helpful in providing context for where in the model the variables were added to the pool and so these have been added back to the reduced info map.

PooyaHekmati (approved):

tomhuhh:
> With the new `basic` formatting option for output variable names, it is a lot easier to use copy paste to create filter files. 

I agree with this. Thanks @ew3361zh for your hard work.  

> I will create an issue related to the above example.

@Isgotkowitz  Has this issue been created? If so, would it be okay if we merge this PR and open a new PR for the additional feature?  

Thanks!

Isgotkowitz:
@tomhuhh it is #869. Though there is not a clear path forward at this time for what the feature should be to address this.

tomhuhh (approved):
LGTM! Good work Niko!

PR #847, opened 2023-10-12
Deprecate `classes.py`

### LINKED ISSUE #839: Revisit classes.py
we have a file  RUFAS/classes.py which contains multiple classes. We need to break this into multiple files such that each file contains only one class. Then we have to evaluate each class and decide on further steps (refactoring, renaming, adding test coverage, etc...)

allisterakun:
- Deleted `classes.py` 
- Moved all four classes that were in `classes.py` into a folder `classes` as `state.py` `config.py` `weather.py` and `time.py` 
- Rerouted all references to the four classes
- Deleted unused `weather_path` parameter in Config class init


ew3361zh (approved):
Nice work, Allister!

One suggestion - I realize almost all this code will be rewritten but with these separate class files the `is_leap_year()` helper function shows up twice. I think indicating somehow thar two separate classes are dependent on this output of this function would be helpful for the eventual redesign team.

Maybe put it in its own file and import it into the two class files that use it? Or maybe making a note of it in the PR would be sufficient.

Isgotkowitz (approved):
Looks great, love seeing classes.py finally being split up. One suggestion on naming conventions though.

I think `classes` is not a very good name, and this PR is a really good opportunity to change it to something more descriptive. Something like `conditions` or `simulation_environment` would be better, though these names are not necessarily the best either.

PooyaHekmati:
> Looks great, love seeing classes.py finally being split up. One suggestion on naming conventions though.
> 
> I think `classes` is not a very good name, and this PR is a really good opportunity to change it to something more descriptive. Something like `conditions` or `simulation_environment` would be better, though these names are not necessarily the best either.

Do we even need to put these files in a separate folder?

Isgotkowitz:
> > Looks great, love seeing classes.py finally being split up. One suggestion on naming conventions though.
> > I think `classes` is not a very good name, and this PR is a really good opportunity to change it to something more descriptive. Something like `conditions` or `simulation_environment` would be better, though these names are not necessarily the best either.
> 
> Do we even need to put these files in a separate folder?

I like the idea of having them in a separate folder, though maybe it would be better to have that folder be all inclusive for everything that isn't part of `routines`.

allisterakun:
> > Looks great, love seeing classes.py finally being split up. One suggestion on naming conventions though.
> > I think `classes` is not a very good name, and this PR is a really good opportunity to change it to something more descriptive. Something like `conditions` or `simulation_environment` would be better, though these names are not necessarily the best either.
> 
> Do we even need to put these files in a separate folder?

I do agree with @Isgotkowitz on keeping them in a separate folder because there are already too many files under the `RUFAS/` folder. And I do believe that it could use a better name rather than `classes/`. Any suggestions, @PooyaHekmati?

allisterakun:
> > > Looks great, love seeing classes.py finally being split up. One suggestion on naming conventions though.
> > > I think `classes` is not a very good name, and this PR is a really good opportunity to change it to something more descriptive. Something like `conditions` or `simulation_environment` would be better, though these names are not necessarily the best either.
> > 
> > 
> > Do we even need to put these files in a separate folder?
> 
> I do agree with @Isgotkowitz on keeping them in a separate folder because there are already too many files under the `RUFAS/` folder. And I do believe that it could use a better name rather than `classes/`. Any suggestions, @PooyaHekmati?

Maybe we can have another folder for `I/O management`, which would contain the IM, OM, and GraphGenerator.

PooyaHekmati:
> > > > Looks great, love seeing classes.py finally being split up. One suggestion on naming conventions though.
> > > > I think `classes` is not a very good name, and this PR is a really good opportunity to change it to something more descriptive. Something like `conditions` or `simulation_environment` would be better, though these names are not necessarily the best either.
> > > 
> > > 
> > > Do we even need to put these files in a separate folder?
> > 
> > 
> > I do agree with @Isgotkowitz on keeping them in a separate folder because there are already too many files under the `RUFAS/` folder. And I do believe that it could use a better name rather than `classes/`. Any suggestions, @PooyaHekmati?
> 
> Maybe we can have another folder for `I/O management`, which would contain the IM, OM, and GraphGenerator.

Let's bring them under RUFAS for now, after refactoring them and giving them descriptive names we can/should re-org the file structure of the code base.

PR #852, opened 2023-10-12
Ration optimizer unit tests

ancheypeng:
Adds some unit tests for the RationOptimizer functions.

Redo of https://github.com/RuminantFarmSystems/MASM/pull/811 because too many merge conflicts


JoeWaddell (approved):
These look great, and bring the ration_optimizer coverage up substantially! Nice work

tests/test_animal.py
```diff
@@ -173,8 +173,143 @@ def heifer_c() -> dict:
...
+    available_feeds['CP'] = [1.0, 2.0, 3.0, 4.0, 5.0, 6.0]
+    available_feeds['dRUP'] = [1.0, 2.0, 3.0, 4.0, 5.0, 6.0]
+    available_feeds['lactating_cow_limit'] = [1.0, 2.0, 3.0, 4.0, 5.0, 6.0]
+
+    return available_feeds
+
+
 def test_calculate_NRC_energy_maintenance_requirements(
-    cow_a: dict, cow_b: dict, heifer_a: dict, heifer_b: dict
+        cow_a: dict, cow_b: dict, heifer_a: dict, heifer_b: dict
```
Fjordrunner, line 312: A nitpick here which pertains to places later in code: `cow_a` is introduced as a dict, but we don't know what types of information is held within it. Maybe making these variable names slightly more descriptive might make future work on these tests easier.

tests/test_animal.py
```diff
@@ -1756,74 +1891,403 @@ def test_triple_values_in_list(input, expected) -> None:
...
+        ration_optimizer.forage_NDF_constraint,
+        ration_optimizer.fat_constraint,
+        ration_optimizer.DMI_constraint_upper,
+        ration_optimizer.DMI_constraint_lower
+    ]
+
+    assert ration_optimizer.cow_cons == [{'type': 'ineq', 'fun': func, 'args': arguments} for func in
+                                         ration_optimizer.constraint_functions]
+
+    assert ration_optimizer.heifer_cons == [cons for cons in ration_optimizer.cow_cons if
```
Fjordrunner, line 1928: This is also nitpicky but if we're working with numpy arrays using np.where might be preferred to the somewhat long list comprehension.

tests/test_animal.py
```diff
@@ -1756,74 +1891,403 @@ def test_triple_values_in_list(input, expected) -> None:
...
+    def objective(x, _): sum(x)
+
+    ration_optimizer.objective = objective
+    ration_optimizer.cow_cons = mock_cow_cons
+    ration_optimizer.heifer_cons = mock_heifer_cons
+
+    return ration_optimizer
+
+
+@pytest.mark.parametrize("udr_or_not, animal_combination, expected_x0, expected_bounds, expected_constraints", [
```
Fjordrunner, line 2216: `is_udr` makes more sense than `udr_or_not` since Boolean values answer that question.

Fjordrunner (changes requested):
Looks good overall, added some nitpicky comments that could be implemented or discussed.

tests/test_animal.py
ancheypeng: `udr_or_not` is the variable name used in the function, in a different PR both can be changed to clean it up

tests/test_animal.py
ancheypeng: They are not numpy arrays, `cow_cons` and `heifer_cons` are just lists. I don't think it would make sense to make them numpy arrays because they are holding a list of dictionaries that contain functions.

tests/test_animal.py
ancheypeng: I think they are just meant to be mock cows to use for unit testing

tests/test_animal.py
Fjordrunner: I would make a note of this and then change them both in the future if possible, but nothing major.

tests/test_animal.py
Fjordrunner: I think it's a bit unintuitive to have `cow_a` be a dictionary given that we have Cow objects in the model, but this is something that can and I think should be fixed in a future PR.

Fjordrunner (approved):

PR #858, opened 2023-10-13
Fixing repro protocols usage in heifer2 and cow classes

### LINKED ISSUE #853: Manure Excretion Output Values are 0
Looking at the values reported to the OM with the variable 'Pen.calc_total_manure', they are all 0

## Expected Behavior

These values should be the sum of the manure values excreted in each pen. 
For example, Pen.calc_total_manure.pen_0_daily_methane should be the daily sum of the methane excreted by all the animals in Pen 0. 

## Current Behavior

All values that start with `Pen.calc_total_manure.pen_X_daily...` are zero. Values that start with `Pen.calc_total_manure.daily_aggregate_ANIMAL_TYPE...` are non-zero and reasonable values

## Steps to Reproduce

1. Run any scenario on the main branch with the following output filter: Pen.calc_total_manure
2.
3.
4.

## Context (Environment)

I am trying to evaluate and summarize the manure excretion outputs

It is important for us to be able to collect the results at different levels of aggregation. It is un

### LINKED ISSUE #854: Not Breeding Heifers
The HeiferII repro programs never succeed in breeding this group so eventually the herd will reduce to zero animals

## Expected Behavior

Heifers should be able to get pregnant (we need baby cows to make milk and produce more cows!)

## Current Behavior

Somewhere along the line from the inputs to the heifer II attributes the variable names of these attributes became disconnected from the functions in the heifer repro functions. So far I have found the following: 

-  on L488 in heiferII.py: `if self.tai_method_h == "d5CG2P":` should be `if self.tai_method_h == "md5CG2P":` because that is what the self.tai_method_h will be assigned as. If it took you a minute to figure out what the difference between these two values are, that is part of the point. The repro protocol methods currently rely on complicated names that are very easy to introduce typos. The other problem this highlights is that the logic allows the animals to just pass through without any assigned breeding protocol. If it doesn't find 'md5CG2p' or any of the other options, it just moves on without providing any errors. 
- Another thing to note, is that previously this variable was called '5dCG2P'. When I run the main branch, for a little over a year the heifers have '5dCG2P' as their TAI method instead of 'md5CG2P'. Thus, I am inferring that the  animals that are selected from the Initialization database will likely have the old value instead of 'md5CG2P'. I am not sure why animals are not established with the breeding protocol used in the current simulation, but that would be the ideal outcome. 
- When I changed the if statement on L488 to  `if self.tai_method_h == "md5CG2P":` , animals will start the `d5CG2p_update()`, but when they get to the point of assigning the conception rate on L453, it does not find the conception rate: 
-
image

## Possible Solution

We have designed the refactor of this part of the model so that is the best solution. I am mostly sharing this information with the hopes that it is helpful so that we can squash this bug for good. 

## Steps to Reproduce

1. One way is to run the default json for a longer period, you will notice that the animals in all groups slowly decline to 0. 
2. Another, quicker way is to run the default json with a print statement on L637 nested under the if statement on L635, you will see that it never inseminates heifers 
3.
4.
5.

ltp23:
In this PR, I made temporary fixes to make things more consistent in the HeiferII and Cow classes when they are integrated with the recent changes in the input json structure. 

I also fixed the inconsistency in the names of the two confusing heifer tai protocols starting from the animal initialization step. It is an acceptable workaround, though. 

I added some error-checking to make sure that when an unexpected repro protocol occurs for some reason, the program will stop so we know what is going on. I have tried a lot of different breeding protocol combinations for both heifers and cows and the program seems to be running without errors but the numbers still need to be checked for reasonableness.

To have some way to do a quick visual inspection, I also coded up some throwaway code to autogenerate a graph for the herd composition. So, anything new in the `output` folder will be removed before merging to the main branch.

I also included the small fix for the zero values currently observed in the animal manure excretion values collected on a per-pen basis. I saw some negative values that only happened on day 1, probably due to the small values obtained from ration calculations. For example, on day 1 the manure mass for heifer1 and heifer2 was negative (because dry matter intake was 0). Also on day 1, cows had a handful of negative-valued manure properties.


KFosterReed:
This does fix the bug Loi! Thank you! I tested it for all the different Repro protocols and it actually gets the animals pregnant! Hooray! 

Before merging though, do we want to have the graph generated in the Main branch? I know Allister and Pooya are working on some graphics generation across the whole model so maybe we should check with them?

RUFAS/graph_herd_composition.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # Labels and Legend
+        plt.xlabel('Simulation Day')
+        plt.ylabel('Number of Animals')
+        plt.title('Herd Composition Over Time')
+        plt.legend(loc='upper left', bbox_to_anchor=(1, 1))
+
+        # Save the Graph
+        plt.savefig(outdir / outfile_name, dpi=300, bbox_inches='tight')
+        plt.close()
```
YijingGong: Love this function. This will also be very helpful when we do the refactoring.

RUFAS/routines/animal/life_cycle/animal_initialization.py
```diff
@@ -466,6 +466,13 @@ def get_heiferIIs(self, num, breed):
                 'p_gest_for_calf': int(row[AnimalValues.p_gest_for_calf]),
                 'calf_birth_weight': float(row[AnimalValues.calf_birth_weight])
             }
+            tai_protocol_name_conversion_map = {
+                '5dCG2P': 'md5CG2P',
+                '5dCGP': 'md5CGP',
+            }
+            if args['tai_method_h'] in tai_protocol_name_conversion_map:
+                old_tai_method_h = args['tai_method_h']
+                args['tai_method_h'] = tai_protocol_name_conversion_map[old_tai_method_h]
```
YijingGong, line 475: Are there any rules in programming that prevent people from using '5dCG2P' as a function name or a variable name? Because it is weird to me why people started to use 'md5CG2P'. 
But anyway I think we should avoid multiple names in the code when refactoring.

RUFAS/routines/animal/life_cycle/heiferII.py
```diff
@@ -330,8 +330,11 @@ def update(self, sim_day):
                 self.ed_update(sim_day)
             elif self.repro_program == "TAI":
                 self.tai_update(sim_day)
-            elif self.repro_program == "synch-ED":
```
YijingGong, line 333: Good catch! We are using Synch-ED all the time.

RUFAS/routines/animal/life_cycle/heiferII.py
```diff
@@ -362,8 +365,8 @@ def determine_estrus_day(self, start_date, estrus_note, sim_day):
         estrus_cycle = truncnorm.rvs(
             -const.STDI,
             const.STDI,
-            AnimalBase.config["avg_estrus_cycle_heifer"],
-            AnimalBase.config["std_estrus_cycle_heifer"],
+            AnimalBase.config["avg_estrus_cycle_cow"],
+            AnimalBase.config["std_estrus_cycle_cow"],
```
YijingGong: Maybe it's lost in the input structure changes, but the heifer estrus distribution is slightly different than cows. Cows have a normal distribution of N(21, 4), and heifers have a normal distribution of N(21, 2.5). I think it's ok to have the temporary fix as this. But I would like to have them different eventually, and since they are 'from literature' section, we won't increase the number of user-input variables. Do you mind changing it here or in later refactoring PRs?

YijingGong (commented):
Good job, Loi. I tested some valid and invalid protocol names, and all of them seem to work. I put some minor comments when I reviewed the code line by line. Besides, there is one thing that caught my attention. When I tested the invalid breeding protocol names, the Rufas program seemed to fix that to the default breeding protocol, which is hard to tell. It seems users can only open the warning log to see that. For example, if I accidentally put "ed" (lower case or some typo, wanted to put 'ED') for heifer_repro_method and cow_repro_method, the program directly fixes them to default TAI:
"values": [
"Invalid data fixed: heifer_repro_method => TAI",
"Invalid data fixed: cow_repro_method => TAI",
"Invalid data fixed: heifer_TAI_protocol => md5CG2P",
"Invalid data fixed: cow_presynch_protocol => Double OvSynch",
"Invalid data fixed: cow_TAI_protocol => OvSynch 56"
]
In this case, the user likely thinks the used method is "ED", but Rufas actually uses "TAI". Do we have options to have more explicit warning or error messages?
I used this input file: [default_animal.json](https://github.com/RuminantFarmSystems/MASM/files/12935341/default_animal.json)

I am not sure if you want to solve the above and my comments in this PR, or later PRs. Let me know. I can make issues if you want me to.

RUFAS/routines/animal/life_cycle/heiferII.py
ltp23: Yes, I will add two more fields to the `from_literature` section for `avg_estrus_cycle_heifer` (21) and `std_estrus_cycle_heifer` (2.5). I used the cow's numbers because I couldn't find them while working on it.

ew3361zh:
> Good job, Loi. I tested some valid and invalid protocol names, and all of them seem to work. I put some minor comments when I reviewed the code line by line. Besides, there is one thing that caught my attention. When I tested the invalid breeding protocol names, the Rufas program seemed to fix that to the default breeding protocol, which is hard to tell. It seems users can only open the warning log to see that. For example, if I accidentally put "ed" (lower case or some typo, wanted to put 'ED') for heifer_repro_method and cow_repro_method, the program directly fixes them to default TAI: "values": [ "Invalid data fixed: heifer_repro_method => TAI", "Invalid data fixed: cow_repro_method => TAI", "Invalid data fixed: heifer_TAI_protocol => md5CG2P", "Invalid data fixed: cow_presynch_protocol => Double OvSynch", "Invalid data fixed: cow_TAI_protocol => OvSynch 56" ] In this case, the user likely thinks the used method is "ED", but Rufas actually uses "TAI". Do we have options to have more explicit warning or error messages? I used this input file: [default_animal.json](https://github.com/RuminantFarmSystems/MASM/files/12935341/default_animal.json)
> 
> I am not sure if you want to solve the above and my comments in this PR, or later PRs. Let me know. I can make issues if you want me to.

I've got a PR(#828) that is in the early stages but it will address this issue by having an option for a more verbose simulation run - e.g. updating the user if input data is valid for metadata file used, giving progress in validation and simulation. I could add something to potentially print out a list of data that has been fixed by the Input Manager during the process. I will tag you in that PR when it is further along and ready for review @YijingGong. Thank you.

RUFAS/routines/animal/life_cycle/animal_initialization.py
ltp23: To keep the naming consistent, I have changed all the usage of '5dCG2P' and '5dCGP' to the corresponding 'md5CG2P' and 'md5CGP'. The only place you may see the '5dCGP/2P' names is in the database, but with the current fix this will get converted during the animal initialization anyway.

At first, I wanted to allow both name styles but then I thought that would make things even more error-prone with the typos.

Programming-wise, yes variable names and probably function names too need to start with an alphabetic character or an underscore. Though I don't know if the motivation for the name change was because of this quirk.

input/metadata/default_metadata.json
```diff
@@ -596,6 +596,18 @@
...
             },
+            "avg_estrus_cycle_heifer": {
+              "type": "number",
+              "description": "Average length of estrus cycle for heifers (default is 21).",
+              "minimum": 1,
+              "default": 21
+            },
+            "std_estrus_cycle_heifer": {
+              "type": "number",
+              "description": "Standard deviation of the length of estrus cycle for cows (default is 2.5).",
```
YijingGong: A typo here: "Standard deviation of the length of estrus cycle for **heifers** (default is 2.5)."

YijingGong (approved):
Solved all my comments. Good job Loi. Just a small typo. 
I approved it.

JoeWaddell (approved):
This looks good, and performs as expected. Nice work Loi!

PR #859, opened 2023-10-13
Record simulation information in OutputManager

### LINKED ISSUE #849: Report Time to the Output Manager
Reporting the values that the Time class maintains (julian day, calendar year, and simulated year) is very helpful when analyzing outputs from RuFaS, and it is cumbersome to have to reimplement the collecting of these values every time a new evaluation branch is created.

The Julian day, calendar year, and simulated year get reported to the OutputManager every day.

- Record nothing about the time and let users create lists representing the time when analyzing the outputs they got.
- Record only the julian day.

**Additional context**

### LINKED ISSUE #850: Report Weather to the OuputManager
Reporting the values that the Weather class contains (minimum daily temperature, maximum daily temperature, precip, etc) is very helpful when analyzing outputs from RuFaS, and it is cumbersome to have to reimplement the collecting of these values every time a new evaluation branch is created.

The weather conditions that changed everyday (min, max, average daily temperatures, precip, etc.) get reported to the OutputManager every day.

- Record nothing about the weather and let users parse input files to get the Weather values they need.

**Additional context**

Isgotkowitz:
This PR adds recording time and weather variables. Closes #849, closes #850. My motivation for adding this to `main` is that I find it very convenient to be able to filter them into the CSVs with variables I need to evaluate/debug/graph, and it was getting annoying to have to reimplement collecting them every time I created a new evaluation branch.

Time variables recorded are julian day, simulated year, and calendar (all on daily timesteps). Weather variables recorded daily are precipitation, maximum temperature, minimum temperature, average temperature, radiation, and irrigation. The average annual temperature is also recorded, but it is only recorded once because it remains static throughout a simulation run. The dedicated functions for recording these variables are called by `SimulationEngine`. All code added in this PR is tested appropriately.

Note these variables are recorded with the prefix "Time" and "Weather" as appropriate. This choice was made because, to me, it seems cleaner to have:
```
Time.day
Weather.precipitation
```
than
```
Time.record_time.day
Weather.record_weather.precipitation
```

`Time` has been removed from the `Utility` module, it was causing some circular import errors.

Edit: I know there are some concerns about program memory when recording these extra values. I tried putting all 7 metadata files we currently have into ScenarioManager and running them together, seemed to work fine on my machine (8 GB ram, which would be the limiting factor right?)


allisterakun (approved):
PR looks good. I left some comments on Slack about `Time`. I will leave a copy here as well:
1. The name `years` and `year` seems a bit confusing to me, as `years` is a list of “absolute” years in the simulation like [2000, 2001, 2002] while the `year` is a “relative” year starting at 1. I know this is “inherited” from the `Config` class, but I think a better name (and some descriptions) can be added for the local variables in the `Time` class.
2. For the `end_simulation()`  function in `time.py`, in the `elif self.year == len(self.years)`: statement, on line 85, I think you can just call the `is_last_day_of_simulation()` function.

Isgotkowitz:
> 1. The name years and year seems a bit confusing to me, as years is a list of “absolute” years in the simulation like [2000, 2001, 2002] while the year is a “relative” year starting at 1. I know this is “inherited” from the Config class, but I think a better name (and some descriptions) can be added for the local variables in the Time class.
> 2. For the end_simulation() function in time.py, in the elif self.year == len(self.years): statement, on line 85, I think you can just call the is_last_day_of_simulation() function.

Strongly agreed. These should be addressed when `Time` gets refactored.

RUFAS/util.py
```diff
@@ -294,14 +293,16 @@ def empty_dir(cls, dir_path: Path, keep: Optional[List[str]] = None) -> None:
                     shutil.rmtree(file)
 
     @staticmethod
-    def day_to_month_conversion(time: Time) -> int:
+    def day_to_month_conversion(day: int, calendar_year: int) -> int:
         """
         Converts the day number into the corresponding month of the year.
```
matthew7838: I know I wrote this, but now I think adding a brief description of how and why calender_year is used would make it a lot easier to understand without looking into the code.

RUFAS/weather.py
```diff
@@ -165,7 +171,7 @@ def get_current_day_conditions(self, time: Time) -> CurrentDayConditions:
         """
         year = time.year
         day = time.day
-        month = Utility.day_to_month_conversion(time)
+        month = Utility.day_to_month_conversion(day, time.calendar_year)
```
matthew7838, line 174: Minor question, is there a difference between these two-year variables?

tests/test_time.py
```diff
@@ -0,0 +1,22 @@
...
+    setattr(config, "leap_year_length", 366)
+    setattr(config, "year_length", 365)
+    return config
+
+
+def test_record_time(mock_config: Config) -> None:
+    time = Time(mock_config)
+    with patch("RUFAS.output_manager.OutputManager.add_variable") as add_var:
+        time.record_time()
+        assert add_var.call_count == 3
```
matthew7838: `is_last_day_of_simulation` is not tested, I think it's worth adding if it's not too time consuming.

matthew7838 (approved):
LGTM, minor comments and questions. Nothing's stopping me from approving.

RUFAS/weather.py
Isgotkowitz: Yes. `year` is the simulated year, `calendar_year` based on when the simulation is scheduled to start and finish. 

So if a simulation is specified to start in 1999 and run until 2001, then the years will be 1 -> 2 -> 3 while the calendar years will be 1999 -> 2000 -> 2001.

tests/test_time.py
Isgotkowitz: Testing `is_last_day_of_simulation` wouldn't be too time consuming, but writing tests for all the untested methods in `Time` would be so I will instead create a scaffold for Time tests and write an issue for this.

Issue #649, opened 2023-07-12
Break up `test_soil_data_and_config_factory.py` file into module specific files


Isgotkowitz:
`test_soil_data_and_config_factory.py` contains tests for several different modules, and each module is in a different file.

## Current
`test_soil_data_and_config_factory.py` contains tests for the files `soil_data.py`, `layer_data.py`, and `soil_config_factory.py`.

## Solution

Put all tests for `soil_data.py` into `test_soil_layer.py`, `layer_data.py` into `test_layer_data.py`, and `soil_config_factory.py` into `test_soil_config_factory.py`.

_Note_: the files `test_soil_data.py` and `test_soil_config_factory.py` will need to be created, but `test_layer_data.py` already exists.

Isgotkowitz:
@Isgotkowitz rewrite this description to have clear mapping between code files and test files - Done

PR #860, opened 2023-10-16
Change access to `Weather`

Isgotkowitz:
This PR changes the way that `Weather` can be accessed by introducing the method `get_current_day_weather()`, which returns a `CurrentDayWeather` object. The `CurrentDayWeather` module and related methods have been moved out of the `Field` module (where the module was named `CurrentWeather`), as well as all associated tests.


ew3361zh (approved):
LGTM - curious how this will fit in with long-term plans around weather in the model. Nice work.

Isgotkowitz:
> LGTM - curious how this will fit in with long-term plans around weather in the model. Nice work.

Thanks Niko:)
The idea with this PR (and the one stacked on top of it) is to force every non-`Weather` module to access `Weather` in a controlled way (i.e. use a getter). Once this is done, we can freely modify and update the way `Weather` handles data internally without making any changes to other parts of the model.

RUFAS/current_weather.py
```diff
@@ -0,0 +1,107 @@
...
+
+    Attributes
+    ----------
+    incoming_light: float, optional, default=None
+        Incoming light radiation energy (MJ/m^2).
+    min_air_temperature: float, optional, default=None
+        Minimum air temperature for the day (C).
+    mean_air_temperature: float, optional, default=None
+        Average air temperature for the day (C).
+    max_air_temperature: float, optional, default=None
```
PooyaHekmati: How can "current" weather have min and max values? Current refers to a single point in time which gives scalar values for each parameter, min and max are defined over a set of data points.

RUFAS/current_weather.py
```diff
@@ -0,0 +1,107 @@
...
+    incoming_light: Optional[float] = None
+    min_air_temperature: Optional[float] = None
+    mean_air_temperature: Optional[float] = None
+    max_air_temperature: Optional[float] = None
+    daylength: Optional[float] = None
+    annual_mean_air_temperature: Optional[float] = None
+    snow_fall: float = 0.0
+    rainfall: float = 0.0
+    irrigation: float = 0.0
+    precipitation: float = 0.0
```
PooyaHekmati: why are these not in __init__ ?

RUFAS/current_weather.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __post_init__(self):
+        """Sets precipitation as snow_fall or rainfall depending on mean air temperature"""
+        if self.mean_air_temperature >= 0:
+            self.rainfall = self.precipitation
+        else:
+            self.snow_fall = self.precipitation
```
PooyaHekmati:
```suggestion
        is_freezing = self.mean_air_temperature < 0:
        if is_freezing:
            self.snow_fall = self.precipitation
        else:
            self.rainfall= self.precipitation
```

RUFAS/current_weather.py
```diff
@@ -0,0 +1,107 @@
...
+        Average air temperature for the day (C).
+    max_air_temperature: float, optional, default=None
+        Maximum air temperature for the day (C).
+    daylength: float, optional, default=None
+        Length of time from sunup to sundown on the day (hours).
+    annual_mean_air_temperature: float, optional, default=None
+        Average annual air temperature for the year (C).
+    snow_fall: float, default=0.0
+        Amount of snow that falls on the day (mm).
+    rainfall: float, default=0.0
```
PooyaHekmati: inconsistent naming; either rain_fall + snow_fall or rainfall + snowfall

RUFAS/current_weather.py
```diff
@@ -0,0 +1,107 @@
...
+        -------
+        int
+            Day length of the month (hour).
+
+        References
+        ----------
+        https://sunrise.maplogs.com/dane_county_wi_usa.15449.html
+        """
+        daylength = [9, 10, 11, 13, 14, 15, 15, 15, 13, 12, 10, 9]
+        return daylength[month-1]
```
PooyaHekmati: This has to come from inputs.

RUFAS/current_weather.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        days = [31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334, 365]
+        if is_leap_year(time.calendar_year):
+            days = [31, 60, 91, 121, 152, 182, 213, 244, 274, 305, 335, 366]
+        prev_month = 0
+        for day in days:
+            if prev_month < time.day <= day:
+                return days.index(day) + 1
+            else:
+                prev_month = day
```
PooyaHekmati: The code can be written more clearly. The name of the function is vague. This function has nothing to do with the weather; it belongs to utils.
```suggestion
  @staticmethod
  def day_to_month_conversion(time: Time) -> int:
      """
      Converts the day number into the corresponding month of the year.
      
      Parameters
      ----------
      time: Time
          Time object containing the current time of the simulation.
      
      Returns
      -------
      int
          The corresponding month of the year (1 for January, 2 for February, etc.).
      """
      # Define the cumulative day counts for non-leap and leap years
      non_leap_cumulative_days_in_months = [31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334, 365]
      leap_cumulative_days_in_months = [31, 60, 91, 121, 152, 182, 213, 244, 274, 305, 335, 366]
      
      cumulative_days_in_months = leap_cumulative_days_in_months if is_leap_year(time.calendar_year) else non_leap_cumulative_days_in_months 
      
      for month, day_count in enumerate(cumulative_days_in_months ):
          if time.day <= day_count:
              return month + 1 

```

PooyaHekmati (changes requested):
Overall looks good, but there are some comments that need to be addressed.

RUFAS/current_weather.py
Isgotkowitz: In this context, minimum and maximum refer to single points which are the lowest and highest temperatures reached on a single day respectively.

RUFAS/current_weather.py
Isgotkowitz: Because `@dataclass`es have their `init` methods generated by Python (described [here](https://docs.python.org/3.10/library/dataclasses.html)). There is a way to add a custom `init` (see [PEP 557](https://peps.python.org/pep-0557/#custom-init-method)), but I think that is not necessary in this use case.

That said, the base initialization method for dataclasses is very loose. Do you think it would be a good idea to have `CurrentWeather` initialization fail if it is passed an unexpected or non-keyword argument?

RUFAS/current_weather.py
Isgotkowitz: Consistent now

RUFAS/current_weather.py
Isgotkowitz: Suggested code changes have been added, and method has been moved to `Utility`.

Important note: I moved the test for this method to a new file `test_util.py`. In order to avoid creating more issues in the same vein as #649, I moved all the other existing tests for `Utility` from `test_misc.py` to `test_util.py`. If this change is undesirable it can be reverted.

RUFAS/current_weather.py
Isgotkowitz: What has to come from inputs? The list of average daylengths for each month, or the daylengths for every single day, something else?

A potential solution that eliminates the problem of Wisconsin daylength values being hardcoded without necessitating a change in the API would be to rewrite this function to calculate the daylength of any given day based on the latitude of one of the fields provided in the input data. There are numerous equations to do this, including SWAT equation 1:1.1.6 ((here)[https://swat.tamu.edu/media/99192/swat2009-theory.pdf]).

PooyaHekmati: Determining day length is not something that is related to weather. We either want to dynamically calculate it based on lat, long, and day of year (which won't be a function in weather data class, it sounds more like a utility) or we want to receive the day length from the input. Either way, hardcoding the day lengths of a particular geographical region in an unrelated part of the code base doesn't sound right.

RUFAS/current_weather.py
PooyaHekmati: Sounds good, `test_misc.py` needed to be broken down anyways.

RUFAS/current_weather.py
Isgotkowitz: > Either way, hardcoding the day lengths of a particular geographical region in an unrelated part of the code base doesn't sound right.

Agreed. To unblock this PR can an issue be written for it with the understanding that this is a mid to high priority, or should a different method of calculation be implemented in this PR? I think it would be better write an issue because it is not a super trivial task, and merits its own PR.

And because daylength is not actually a weather condition, should `CurrentDayWeather` be renamed to `CurrentDayConditions` (or something similar) to better reflect its purpose?

Isgotkowitz: Issue #881 has been written for calculating daylength with better resolution, and `CurrentDayWeather` is now `CurrentDayConditions`.

PooyaHekmati (approved):

Issue #812, opened 2023-09-29
Variables Not Output As Expected 


tomhuhh:
While working with the latest version of the Output manager, I noticed some variables are not output correctly. 

1) the number of animals in each parity remain static. 
2) No output of simulation day in the life cycle manager section. 
3) No methane emission output. 

## Expected Behavior

The number of animals and methane emissions should change over the simulation period. 

## Current Behavior

Number of animals for each parity 
![image](https://github.com/RuminantFarmSystems/MASM/assets/52800053/7a981213-c625-4f90-86a9-9d3051f089cf)

methane emission 
![image](https://github.com/RuminantFarmSystems/MASM/assets/52800053/7c12a17b-f61c-4e0f-83a3-b9b935320ea2)

## Possible Solution

1) Add output manager to the file where the output variable is generated. 
2) Update where the output variable is collected. 

## Steps to Reproduce

1.
2.
3.
4.

KFosterReed:
Hey team, I opened and then closed an issue that is similar to this. We need to clean up these outputting bugs so that we can meet our pilot testing goals.

ew3361zh:
@tomhuhh can you add a few specifics for this issue?

>1. the number of animals in each parity remain static.

What output variable are you using to track this?

>2. No output of simulation day in the life cycle manager section.

It looks like there are close to 40 values in the LifeCycleManager.daily_update.* group - can I assume this is the "life cycle manager section"? 

It's been a while since I worked in there but I think we'd originally talked about being able to use the index number of each value reported as interchangeable with the simulation day since it's a daily update. It shouldn't be hard if we just want to add day of simulation to this group though.

ew3361zh:
I am seeing sim_day in the info_maps of the LifeCycleManager.daily_update.* section (right at the top):
<img width="686" alt="Screenshot 2023-10-13 at 11 04 08 AM" src="https://github.com/RuminantFarmSystems/MASM/assets/70217952/cf2b106b-1178-4dd4-9acb-6e28da3b2cc1">

We could move that to values if that would be more helpful.

ew3361zh:
This was what I got when I plotted LifeCycleManager.daily_update.[calf_num, heiferI_num, heiferII_num, etc] against sim day:
<img width="902" alt="Screenshot 2023-10-13 at 11 29 17 AM" src="https://github.com/RuminantFarmSystems/MASM/assets/70217952/343d0eba-dbd3-41dc-a667-339f1b0668e8">

[...]

PR #861, opened 2023-10-16
Unexpected animal outcomes

ew3361zh:
This PR seeks to address the first two points in issue #812:
1. The number of animals in each parity remain static (`num_cow_for_parity`).
2. No output of simulation day in the life cycle manager section.

For part 1:
This seemed to hinge on the entire num_cow_for_parity dictionary being added at once. 
By separating the key-value pairs and adding them to the OM individually, the daily changes were correctly tracked.
This issue (#812) has helped us identify a bug in how OM adds complex data structures to the variables pool. It is further explained in issue #862.

For part 2:
`sim_day` was removed from the daily_update method info_map and instead added as a variable.

Additional note:

- In hopes of flattening the overall structure of the OM variables pool, I added each updated LifeCycleManager attribute individually rather than as a nested dictionary. If this is undesireable for any reason, please let me know. I think it should simplify getting at each piece of data updated in the daily_update method of LifeCycleManager.


ew3361zh:
@tomhuhh despite fixing the mechanism by which complex data structures are being added to the Output Manager (which was the reason `num_cow_for_parity` wasn't tracking correctly in the csv output files), there are still some other features this branch achieves which I'd like to have your feedback on:

1. Flattening out the LifeCycleManager daily update variables. I'd previously added these all to a dictionary and added the dictionary all at once. This PR adds each of those daily variables separately which makes the overall structure of the output pool a little flatter (less nesting) and perhaps easier to read/get data from. Take a look at the output from running a simulation with this PR and let me know.
2. Adds simulation day to the variables pool rather than keeping it in the info_map making it hopefully more easily accessible for your reporting needs.
3. Makes some overdue changes to the structure of the LifeCycleManager class which may not be noticeable right away but make more sense programmatically now that Output Handler isn't reliant on the prior LifeCycleManager class structure.

Please let me know what you think - we should get at least some of these changes implemented. Thanks!

tomhuhh (approved):
Great work @ew3361zh! The output structure is much more clear to me and makes it way easier for subsequent calculations. An example of the changes that I find helpful is below. 

`LifeCycleManager.daily_update.num_cow_for_parity_1
LifeCycleManager.daily_update.num_cow_for_parity_2
LifeCycleManager.daily_update.num_cow_for_parity_3
LifeCycleManager.daily_update.num_cow_for_parity_greater_than_3`

tomhuhh:
Question about adding the 'simulation day' data to the variable pool instead of the info_map- 

Would we like to apply it to other sections of the output, for example, milk yield and components, `Cow.milking_update.milk_data_at_milk_update`? I believe for consistency, it is beneficial to add the 'simulation day' data to all other sections if that is not too much additional work. Thanks!

ew3361zh:
> Question about adding the 'simulation day' data to the variable pool instead of the info_map-
> 
> Would we like to apply it to other sections of the output, for example, milk yield and components, `Cow.milking_update.milk_data_at_milk_update`? I believe for consistency, it is beneficial to add the 'simulation day' data to all other sections if that is not too much additional work. Thanks!

Thanks, Haowen. I think this could be done pretty easily if we have a list of the functions where this would be helpful. I can flatten the milk_data_update and add the sim_day to the Cow.milking_update function to start and look if there are any other similar unnecessarily nested dictionaries of data added to the Output Manager.

Do you know of any other functions where it would be helpful to add sim_day?

tomhuhh:
> Do you know of any other functions where it would be helpful to add sim_day?

Thanks, Niko! The `Cow.milking_update.milk_data_at_milk_update` function is the only one that I think can benefit the most from adding the 'simulation day' data. I will keep track of this while working with the model and possibly we can address this issue (if it becomes an issue) in a different PR. The current PR looks good to me. Thanks!

tomhuhh:
> Thanks, Niko! The `Cow.milking_update.milk_data_at_milk_update` function is the only one that I think can benefit the most from adding the 'simulation day' data. I will keep track of this while working with the model and possibly we can address this issue (if it becomes an issue) in a different PR. The current PR looks good to me. Thanks!

Just to make a note of my own note, as of now, I do not see the 'simulation day' data be included in the variable pool of the `Cow.milking_update.milk_data_at_milk_update` function yet. My assumption is that by the time this PR is merged into main, this will be updated. Thanks!

README.md
```diff
@@ -1,6 +1,7 @@
 [![Testing](https://github.com/RuminantFarmSystems/MASM/actions/workflows/testing_pytest.yml/badge.svg)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/testing_pytest.yml)
 [![Linting](https://github.com/RuminantFarmSystems/MASM/actions/workflows/lint_flake8.yml/badge.svg)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/lint_flake8.yml)
-[![Coverage](https://img.shields.io/badge/coverage-91%25-green)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/coverage_pytest-cov.yml)
+[![Coverage](https://img.shields.io/badge/coverage-90%25-green)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/coverage_pytest-cov.yml)
```
allisterakun, line 3: Why is the test coverage down to 90%?

ew3361zh: Good question - I believe it's because we're increasing the number of uncovered lines of code adding variables to the OM individually vs as a dictionary in a single line. I talked it over with @PooyaHekmati in our check-in Monday and he said it was okay. Testing these types of lines right now isn't a top priority.

allisterakun (approved):
LGTM

README.md
PooyaHekmati: True, the test suite is not being degraded here. This part of the code base doesn't have a good coverage anyways and it is something that we need to work on in the next quarter. Doesn't make sense to hold the PR for this.

Issue #753, opened 2023-09-07
Weather Data Reformat


allisterakun:
Currently, all the weather data are stored in separate 2D arrays of the following format:
```variable_name: [year_num][day]```
Where variable_name is the variable of interest (e.g., T_max => maximum temperature of the day), year_num is the difference between the year of interest and the starting year (e.g., for a simulation starting in 2019, the year 2020 would be index 1), and finally the Julian date.

I believe that the current weather data structure introduces unnecessary complexity, and it should be flattened or completely reformatted at one point.

Isgotkowitz:
I propose the following steps for addressing this issue:
- [x] Create method(s) that get all the information about the weather on the current day. #860 
  - Possibly with a `CurrentWeather` object?
- [x] Eliminate all instances of weather access via array indexing in the codebase. #864 
- [ ] Implement better internal structure for weather data.
  - This one is a biggie, most likely will require some designing before implementing.

At the current moment I think these tasks will require between 2-3 PRs to complete.

Isgotkowitz:
The priority of changing how `Weather` maintains data internally is low, so the final task should wait until sometime after the delivery of v1 in 2024.

ltp23:
This issue is blocked by #1383.

ew3361zh:
Handled in #1507.

PR #864, opened 2023-10-17
Change accesses to `Weather` throughout RuFaS

Isgotkowitz:
This PR fully integrates `get_current_weather()` from `Weather` into the rest of the codebase, and eliminates direct accesses to `Weather`'s attributes. The files where access was modified are `animal_manager.py`, `manure_handler_classes.py`, and `base_manure_treatment.py`, and have also had type annotations added to changed methods. All associated tests are updated as well.


RUFAS/weather.py
```diff
@@ -47,13 +47,13 @@ def __init__(self, weather_file, config):
...
 
         # initialize data sets
-        self.rainfall = []
-        self.T_max = []
-        self.T_min = []
-        self.T_avg = []
-        self.radiation = []
-        self.irrigation = []
-        self.T_avg_annual = []
+        self.__rainfall = []
```
JoeWaddell: Would this PR be the place to improve variable names (specifically ones with 'T', like T_max), or to update the format of the Docstrings in the weather file?

JoeWaddell (commented):
Nice straightforward PR. Changes make sense, and improves readability of the code overall.

JoeWaddell (approved):

RUFAS/weather.py
Isgotkowitz: It would, good thought.

ltp23 (approved):
Very good PR. Thanks, Ed. I like that we can now annotate the weather and time references. Thank you for updating their usage in the animal and manure module. 

One design concern I have is the use of 'private' dunder (double underscore) variables. They seem somewhat of an anti-pattern in Python because we are really going out of our way to make it inaccessible to the user. If someone wants to modify those supposedly private variables, they still can. My point is that if anything is not in the public API, it is expected that the user should not dig deeper than that.

Isgotkowitz:
> One design concern I have is the use of 'private' dunder (double underscore) variables. They seem somewhat of an anti-pattern in Python because we are really going out of our way to make it inaccessible to the user. If someone wants to modify those supposedly private variables, they still can. My point is that if anything is not in the public API, it is expected that the user should not dig deeper than that.

@ltp23 I agree, I think those double underscore arrays are very anti-pattern like. They should be changed at some point, but after our work together session today it seems like #753 will not be closed for a while, so the private arrays might be around for a while.

PR #866, opened 2023-10-17
Fix test_save_variables_to_csv_files_exceptions

PooyaHekmati:
It seems `test_save_variables_to_csv_files_exceptions()` is not setup properly. It 1) is not mocking the dependencies, 2) is not handling the right set of exceptions that mkdir can raise. This PR aims to fix that.

P.S. The PR seems big due to auto-formatting. The only code that was manually changed is `test_save_variables_to_csv_files_exceptions`.

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -546,8 +546,12 @@ def test_save_variables_to_csv_files_exceptions(
     """Unit test for the function _save_variables_to_csv_files() in the file output_manager.py"""
 
     with patch("pathlib.Path.mkdir") as mock_mkdir:
-        mock_mkdir.side_effect = OSError("test error")
-        with pytest.raises(OSError, match="test error"):
+        mock_mkdir.side_effect = OSError("test OS Error")
+        with pytest.raises(OSError, match="test OS Error"):
+            mock_output_manager._save_variables_to_csv_files({}, "filter", "path")
+
+        mock_mkdir.side_effect = FileNotFoundError("test File not Found Error")
+        with pytest.raises(OSError, match="test File not Found Error"):
             mock_output_manager._save_variables_to_csv_files({}, "filter", "path")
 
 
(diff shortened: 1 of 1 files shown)


tests/test_misc.py
```diff
@@ -514,13 +541,18 @@ def test_dict_to_file_csv_exception(mock_output_manager: OutputManager) -> None:
...
 
-    with pytest.raises((PermissionError, OSError)):
-        mock_output_manager._save_variables_to_csv_files({}, 'filter', invalid_path)
+    with patch("pathlib.Path.mkdir") as mock_mkdir:
+        mock_mkdir.side_effect = OSError("test OS Error")
+        with pytest.raises(OSError, match="test OS Error"):
+            mock_output_manager._save_variables_to_csv_files({}, "filter", "path")
+
+        mock_mkdir.side_effect = FileNotFoundError("test File not Found Error")
+        with pytest.raises(OSError, match="test File not Found Error"):
```
Isgotkowitz:
```suggestion
        with pytest.raises(FileNotFoundError, match="test File not Found Error"):
```
Makes the test a little more precise.

Isgotkowitz (approved):
Looks good Pooya, one suggestion to make the test more precise.

Which formatter did you use to apply these changes? It would be good if we got a standard formatter for RuFaS so we could work towards having the whole codebase being formatted the same.

ltp23 (approved):
LG8TM. The new structure for that test is more specific about expected error. Also, if there are more error types, we could use parametrization. Thanks. Pooya.

PooyaHekmati:
> Looks good Pooya, one suggestion to make the test more precise.
> 
> Which formatter did you use to apply these changes? It would be good if we got a standard formatter for RuFaS so we could work towards having the whole codebase being formatted the same.

I use black.

Issue #845, opened 2023-10-11
IM should only add data from input files to the pool if it is specified in the metadata.


ew3361zh:
IM should ensure input data that is NOT specified in the metadata is NOT being added to the pool.

## Expected Behavior
When loading data from the various input files, there may be data fields in the input files that is not specified in the metadata. This data should not be added to the IM pool.

## Current Behavior
Currently data in input files that is not specified by the metadata is being added to the IM pool because it is avoiding being checked.

e.g.

say we have a metadata file with this in "config_properties" section:
metadata = {"properties": {"config_properties": {"start_date": {...}, "end_date": {...}, "graphic_dir": {...}}}}

And the config input file has all these as well, plus one additional field:
config = {"start_date": {...}, "end_date": {...}, "graphic_dir": {...}, "extra_field_with_who_knows_what_data": {...}}

The metadata will validate and potentially fix the `start_date`, `end_date` and `graphic_dir` fields with the values from the config input file but if they are all valid/fixed, IM will add the entire config blob to the pool.

## Possible Solution
Not sure.

## Steps to Reproduce
1. Add an extra field to any input file:
![Screenshot 2023-10-11 at 4 57 39 PM](https://github.com/RuminantFarmSystems/MASM/assets/70217952/da9fec3b-e573-40b2-8f1c-c60c4e1bdeb3)

2. Make a call to IM.get_data() somewhere in the model.
<img width="876" alt="Screenshot 2023-10-11 at 4 57 10 PM" src="https://github.com/RuminantFarmSystems/MASM/assets/70217952/de991776-8e45-4b7b-8271-82da99629061">

3. It will return the value from that unchecked field. 
<img width="454" alt="Screenshot 2023-10-11 at 4 58 07 PM" src="https://github.com/RuminantFarmSystems/MASM/assets/70217952/b39dec66-02a7-42fd-9488-9dd3bfc751e0">

## Context (Environment)
This is a security hole.

ew3361zh:
Resolved with PR #867.

PR #867, opened 2023-10-17
Im pool fix

ew3361zh:
PR to address issue #845 - when loading data from the various input files, there may be data fields in the input files that is not specified in the metadata. This data should not be added to the IM pool.

- [x] Adds new method `_clean_input_data` which removes any key-value pairs from the input_data that doesn't have a corresponding key in the metadata against which to validate it.
- [x] Updates metadata files with missing fields present in input files that are being used by the model. 
- [x] Updates unit testing.

Tested with the same scenario in #845 and it did not add the secret monkeys to the IM pool.


PooyaHekmati (changes requested):
So, if I'm understanding correctly, after this PR, IM will first add everything from a file into its pool and then remove the ones that shouldn't be there. This is not correct, as mentioned in the IM design doc, IM is supposed to only add things that are 1) present in the metadata, 2) can satisfy the validation checks, 3) the default value if validation fails and a default value is provided. 
Simply adding everything and then taking out the ones that don't have a matching key in the metadata means 1) wasting computational power, 2) validation is not working.

ew3361zh:
> So, if I'm understanding correctly, after this PR, IM will first add everything from a file into its pool and then remove the ones that shouldn't be there. This is not correct, as mentioned in the IM design doc, IM is supposed to only add things that are 1) present in the metadata, 2) can satisfy the validation checks, 3) the default value if validation fails and a default value is provided. Simply adding everything and then taking out the ones that don't have a matching key in the metadata means 1) wasting computational power, 2) validation is not working.

The order of operations for IM with this PR is:
1. Load metadata
2. For each blob, load the input data from the appropriate file
3. Clean the input data
4. Validate/fix the newly cleaned input data
5. Add the data to the IM pool if the data is validated/fixed
6. Move to the next blob and repeat steps 2-5.

All these steps were previously present except for step 3 which is what this PR adds.

PooyaHekmati:
> > So, if I'm understanding correctly, after this PR, IM will first add everything from a file into its pool and then remove the ones that shouldn't be there. This is not correct, as mentioned in the IM design doc, IM is supposed to only add things that are 1) present in the metadata, 2) can satisfy the validation checks, 3) the default value if validation fails and a default value is provided. Simply adding everything and then taking out the ones that don't have a matching key in the metadata means 1) wasting computational power, 2) validation is not working.
> 
> The order of operations for IM with this PR is:
> 
> 1. Load metadata
> 2. For each blob, load the input data from the appropriate file
> 3. Clean the input data
> 4. Validate/fix the newly cleaned input data
> 5. Add the data to the IM pool if the data is validated/fixed
> 6. Move to the next blob and repeat steps 2-5.
> 
> All these steps were previously present except for step 3 which is what this PR adds.

I doubt that because we know we had missing fields in the weather metadata, while those fields ended up in the pool. This means that those fields were added to the pool without validation because how can you validate something without knowing how to validate it? And I don't see any changes to the validation logic in this PR.

The fix is not to load everything and then remove some, the fix is to load the correct one in the first place. In other words, the for-loop in line 313 in IM (main branch) should loop over the properties items not over data items. I remember we had the same case elsewhere where the looping was over the data but had to be over metadata. Please also check json validation.

ew3361zh:
> > > So, if I'm understanding correctly, after this PR, IM will first add everything from a file into its pool and then remove the ones that shouldn't be there. This is not correct, as mentioned in the IM design doc, IM is supposed to only add things that are 1) present in the metadata, 2) can satisfy the validation checks, 3) the default value if validation fails and a default value is provided. Simply adding everything and then taking out the ones that don't have a matching key in the metadata means 1) wasting computational power, 2) validation is not working.
> > 
> > 
> > The order of operations for IM with this PR is:
> > 
> > 1. Load metadata
> > 2. For each blob, load the input data from the appropriate file
> > 3. Clean the input data
> > 4. Validate/fix the newly cleaned input data
> > 5. Add the data to the IM pool if the data is validated/fixed
> > 6. Move to the next blob and repeat steps 2-5.
> > 
> > All these steps were previously present except for step 3 which is what this PR adds.
> 
> I doubt that because we know we had missing fields in the weather metadata, while those fields ended up in the pool. This means that those fields were added to the pool without validation because how can you validate something without knowing how to validate it? And I don't see any changes to the validation logic in this PR.

Right...that's what this PR is addressing. That's what the new function does and it does this before the validation process begins. We avoid the validating things it doesn't know how to validate by making sure the input_data keys and metadata property keys match (based on the metadata property keys being the authority of what keys should be allowed).

> The fix is not to load everything and then remove some, the fix is to load the correct one in the first place. 

Everything is not loaded to the IM pool immediately. The data file is loaded and then immediately cleaned before it's sent to validation. The `_populate_pool()` loop at 165 is looping through the metadata file blob paths and in each loop, it is first loading the data from the input file (it does NOT load this into the IM pool, it just loads it to be validated). The next step is the metadata properties for that input file are saved in a variable.

**no data has been loaded to the IM pool yet**

The next step is `self._clean_input_data(input_data, properties)` (line 178).

This new function, added in this PR, is cleaning the data. If there is any key-value pair which doesn't have a key in the metadata properties, it will be removed from the input_data.

Once that process is done, the validation process begins (lines 179-186).

So at the point where the input data is sent to be validated, it has already had key-value pairs not in the metadata removed.

If the data has been validated/fixed, it will return that information on either line 181 if it's a json file and 184 if it's a csv.

Only at that point, if the data is valid, will it be added to the IM pool (line 191).

>In other words, the for-loop in line 313 in IM (main branch) should loop over the properties items not over data items. I remember we had the same case elsewhere where the looping was over the data but had to be over metadata. Please also check json validation.

The for loop in 313 is validating each row of data in against the default value from the metadata. It goes column by column. Also at this point the input data has already had all columns removed that didn't have column headers that were present in the metadata properties.

ew3361zh:
I tested this with a weather file and added a new column of data that didn't have a corresponding metadata property for validation and it was not added to the IM pool.

allisterakun:
Hi @PooyaHekmati, after reading through the code and talking with @ew3361zh, I believe this fix makes sense. 
1. Each input file is first loaded into a local variable, `input_data`. 
2. `input_data` is then passed into `_clean_input_data()` function (maybe a different name?):
  i. It will recursively loop through all the keys in `input_data` to check if each key is in the corresponding metadata section.
  ii. The missing key will be deleted from `input_data` if it is not in the metadata.
  iii. After the loop, all the keys left in input_data would have their corresponding metadata information. 
3.  The "cleaned" `input_data` will then be sent to be validated (and fixed if needed) and then added to the pool (if valid).
So, with this fix, no data will be added to the pool before it has been "cleaned" and validated. 

Alternatively, we can move the "cleaning/filtering" process into the validation functions. So it would be something like this:
1. Each input file is first loaded into a local variable, `input_data`. 
2. `input_data` is sent to the corresponding function (JSON or CSV) to be validated:
  i. In the validation function, we recursively iterate through all the keys in the corresponding metadata section.
  ii. If we find the key in the `input_data`, we will go through the validation and fixing process. If the key is valid or is fixed, we add it to a new data structure (another local variable that holds all the validated keys).
  iii. If the key is not found in the `input_data`, we try to fix it using the `default` value. If it doesn't have a default value (critical data), we then report an error and halt the execution.
  iv. After going through all the keys in the corresponding metadata section, we finally add the validated data (held in the new local variable) into the pool.

ew3361zh:
- [x] new method name changed from `_clean_input_data()` -> `_filter_input_data_by_metadata()`
- [x] new method now returns filtered input_data rather than updating input_data by reference/side-effect
- [x] updates unit testing, docstrings

PooyaHekmati (approved):

allisterakun (approved):
Looks Good!

Issue #862, opened 2023-10-16
OM not correctly adding complex data structures to pool


ew3361zh:
Output Manager is not set up correctly for adding complex data structures to the pool - because of pass-by-reference, more complex data structures added to OM on a daily basis will end up looking like they've added the same values for each day.

## Expected Behavior
For data structures added to OM, the values should accurately represent their value at the time within the simulation they are added to OM.

## Current Behavior
When a complex data structure is added to the OM pool multiple times over the course of a simulation, it will always point back to the memory location rather than the updated value. This means the values of the data structure will end up being the same as the final time the data structure is altered by the model and it will look like the values are unchanged over the course of the simulation.

## Possible Solution
In the OM `_add_to_pool()` method, there should be an exhaustive check for whether the values being added are a complex or primitive data types. If they are primitive, they are passed by value and can be added without further processing. If they are more complex and are passed by reference, they need to first have a deep_copy made and have that be added to the pool. This will prevent overwriting.

## Steps to Reproduce
This came about from issue #812 in which Haowen identified the LifeCycleManager's `num_cow_for_parity` dictionary attribute being added on a daily basis and was not changing over the course of a simulation.

1. In LifeCycleManager.daily_update, I added the `num_cow_for_parity` dictionary to the OM variables pool.
2. In the same location in the code I created a primitive version of each of the fields in the `num_cow_for_parity` and added each one of them individually to the OM variables pool.
3. I created an output filter to create an output csv with just these variables.
4. The primitive-type variables changed throughout the 365 day simulation. The dictionary versions were the same each day and matched the values for the final day of the simulation as recorded by the primitive types.

## Context (Environment)
This is a bug. It will likely grow the output pool in size and the work of seraching for and making a deep_copy every time a complex data structure is found will likely add some processing time to the simulation.

ew3361zh:
Resolved by PR #868.

PR #868, opened 2023-10-18
OM add complex data structures to pool properly

ew3361zh:
This PR seeks to address issue #862 in which a bug was found that the OM was adding complex data structures by reference and therefore by the end of a simulation, the values of some complex structures were all referring back to the final updated version of that variable rather than the values of that variable at the points it was added to the OM.

To address this, the PR:

- [x] Checks if the value being added to the pool is a primitive or complex type.
- [x] Makes a deepcopy of the value before adding it to the pool if it is a complex type. If it is a primitive type, it goes ahead as before and just adds it to the pool.
- [x] Updates unit testing.


PooyaHekmati (approved):

Isgotkowitz (approved):
LGTM, very clean fix.

PR #880, opened 2023-10-26
Make decomposition temperature effect based on soil layer temperature.

### LINKED ISSUE #875: Temperature effect on decomposition is calculated with air and not soil temperature
In the carbon cycling module, one of the factors that effects decomposition is the "temperature effect on decomposition factor".

## Expected Behavior

Temperature factor should be calculated for each individual soil layer based on that layer's temperature.

## Current Behavior

There is one temperature factor set for the entire soil profile, based on the air temperature.

## Reasoning
The expected behavior relies on some guesswork.

The [psuedocode](https://docs.google.com/document/d/1yrlYhPAmNDweJMQnEp8I1H3nje39390lymnmW_YlmfE/edit) defines the method of calculation of this factor in equation [S.6.A.1]. In the psuedocode, it writes the temperature as _T<sub>avg</sub>_. It is not explicitly stated in the parameters section of [S.6.A.1] that this is the average air temperature of the day, but it is defined as the "mean air temperature for a given day" in the parameters section of equation [S.2.B.1]. And the implementation of the temperature factor in the pre-Soil Crop Redesign Carbon Cycling module uses the average air temperature to calculate this factor.

The implementation of the temperature factor in the pre-Soil Crop Redesign Carbon Cycling module states the equations [S.6.A.1] and [S.6.A.2] come from the [defac](https://3.basecamp.com/3486446/buckets/5296287/vaults/2740532358) spreadsheet. That spreadsheet contains the graph:
image
Because the x-axis is labeled as "average soil temperature" it I think it is reasonable to assume that the temperature effect on decomposition should be calculated with the temperature of the soil, and should be specific to each soil layer.

matthew7838:
### Summary
This PR is related to the carbon cycling module. One of the factors that affect decomposition is the `temperature effect on decomposition factor`, however, the current implementation is based on `avg_temp`, which is one temperature for all layers based on the average air temperature. This PR aims to make the `temperature effect on decomposition factor` layer-dependent.

### Changes
- The main routine method calls to `Decomposition.py` will not pass in the `avg_temp` parameter.
- The `_calc_temp_factor` call in the main routine will now be called for each layer and take in layer temperature as the parameter.
- Since `decomposition_temperature_effect` is now layer-dependent, it will be moved from soil data to layer data.
- Some minor fixes to the comments.
- Tests updated

### Notes
For further understanding of the incentives behind this change, please checkout issue #875 .


Isgotkowitz (approved):
LGTM, nice fix Matthew

JoeWaddell (approved):
This looks great, runs properly. Nice job!

PR #882, opened 2023-10-30
Fix biomass adjustment after harvest

Isgotkowitz:


Isgotkowitz:
Will write fix in fresh branch to avoid merge conflicts

PR #884, opened 2023-10-31
Biomass adjustment and accumulation bug fix

### LINKED ISSUE #837: Crop growth stops after crop has been planted for > 1 year
Some crops, instead of being killed every year, and left in the field and start growing again the next spring, right now crops in RuFaS that should have this behavior do not.

## Expected Behavior

Perennial crops should stop growing over the winter and start growing again in the spring.

## Current Behavior

__Edit__: part of the reason this errant behavior is occurring is because above ground biomass is not being removed from the crop when it gets harvested.

Crops stop growing in the winter and then continue to not grow in subsequent growing seasons.

Below is a graph showing this behavior in alfalfa.
alfalfa_biomass_bug

## Possible Solution

To address the lack of above ground biomass removal, redundant biomass trackers should be removed. This could mean one of the following options:
- There is one biomass tracker, and there is a fraction that determines how much biomass is above ground and how much is below (bad solution).
- Aboveground and root biomass are tracked separately, with new biomass being partitioned between the two trackers using some crop specific fraction.

## Steps to Reproduce

1. Write a crop input file that grows a crop over multiple years.
2. Run RuFaS with that crop input.
3. Observe results.
4.
5.

Isgotkowitz:
This PR addresses two issues that were preventing biomass from being accumulated and reduced properly:
1. After a non-lethal harvest event, biomass is removed the above ground and root pools properly based on how much total biomass is removed. This is achieved by the new method `_recalculate_biomass_distribution()` in `CropManagement`.
2. The `CropData` attribute `heat_fraction` is now a `@property` attribute. This makes more sense because it is set based on two other attributes contained in `CropData`, and having it as a `@property` allows it to be automatically adjusted when `accumulated_heat_units` is adjusted.

A more minor but still necessary change has been to adjust the biomass distribution of a crop when it transitions into dormancy.

Tests have been rewritten or adjusted to account for the changes made in other modules, and code cleaned up where applicable.
- All tests for `CropData` are now contained within `test_crop_data.py` and `test_crop_data_and_species_factory.py` has been renamed to `test_crop_species_data_factory.py` to reflect it single responsibility.
- TODOs have been removed.
- Docstrings in methods touched have been polished.

Closes #837

Contrast the behavior of biomass accumulation after the changes in this PR vs. the behavior outlined in issue #837:
second_corrected_alfalfa_biomass


JoeWaddell (approved):
Runs well, results appear correct. Also nice job keeping test coverage high, and expanding/reformatting docstrings where appropriate. LGTM!

matthew7838 (approved):
LGTM, great consistency with test coverage

PR #885, opened 2023-10-31
Animalmanager refactor m1

JoeWaddell:
Beginning of milestone 1 in the following AnimalManager Refactor Design Doc: https://docs.google.com/document/d/16j-8ylEl-jZ4ZP2fW5djm_R_ROIRlFN3DhckbH0Wkuw/edit?usp=sharing

Typing hints will need careful review from a SWE! Likewise new docstrings: the main text of which also needs careful review by SMEs.

In review, if SWEs want to add type hints for the complex types that I struggled with, please do! Or, simply point me in the direction of how to construct type hints for some of these classes that seem to cause circular imports if attempted directly. Maybe this would require more TypedDicts, or a similar approach?


ancheypeng (approved):
Looks good! The added documentation is really helpful

RUFAS/routines/animal/animal_manager.py
```diff
@@ -351,20 +375,27 @@ def init_nutrient_rqmts(self, weather: Weather, time: Time, feed):
...
         Calls each animal's method to calculate its nutrient requirements.
 
-        Args:
-            feed: instance of the feed class
-            temp: the temperature on the current day
+        Parameters
+        ----------
+        feed : Feed
+            instance of the feed class
+        temp : float
```
Fjordrunner: I would prefer `current_temp` here because unlike other arguments that have similar names, this isn't an instance of a class but rather a specific value that might change drastically from call to call.

RUFAS/routines/animal/animal_manager.py
```diff
@@ -1038,21 +1088,23 @@ def sum_daily_milk(self, cows) -> float:
...
         """
         Gathers all the pen history data for a given cow class type. Checks the current pen
         and pen composition of all animals of a given cow class, before then updating the
         pen history for that class using the update_pen_history() method
 
-        Args:
-            cow_class: list of instances of whatever cow type's pen history is being gathered
+        Parameters
+        ----------
+        cow_class : List
```
Fjordrunner: This needs to be improved. There might be multiple cow classes in this list, so the name isn't accurate, and the type hint here should be included in my opinion.

RUFAS/routines/animal/animal_manager.py
```diff
@@ -1159,20 +1216,23 @@ def generate_animal_output(self, animal_type, index):
...
-            num_animals: the number of each type of animal (calves, heiferIs,
-            heiferIIs, heiferIIIs, cows, sold_heifers, and culled_cows) for
-            which information will be collected and returned. If num_animals is
-            larger than the minimum length of the animal lists, then num_animals
-            will be set to the minimum length of the animal lists
+        Parameters
+        ----------
+        num_animals: int
+            the number of each type of animal (calves, heiferIs, heiferIIs, heiferIIIs, cows, sold_heifers, and
+            culled_cows) for which information will be collected and returned. If num_animals is largerthan the minimum
```
Fjordrunner: Fix typo to "larger than," also, maybe a bit too long but nothing major.

RUFAS/routines/animal/animal_manager.py
```diff
@@ -1159,20 +1216,23 @@ def generate_animal_output(self, animal_type, index):
...
+            length of the animal lists, then num_animals will be set to the minimum length of the animal lists.
+
+        Returns : Tuple[List[Tuple[AnimalBase, str, bool]], Dict[str, Dict | int]]
```
Fjordrunner, line 1284: Does it make sense to pair these two in a tuple? Maybe splitting the functions up here could be helpful, but I do see that both outputs get built simultaneously which means that maybe this needs to stay. However, this means that we might want to change the function name to make it more informative? Just spewing some thoughts here.

Fjordrunner (commented):
Overall this is definitely good PR, I added some minor. However, we should try to clean up these TODOs as much as we can before merging, which we can discuss in our meeting tomorrow.

RUFAS/routines/animal/animal_manager.py
JoeWaddell: I think I've updated the typing properly, and updated the docstring as well.

RUFAS/routines/animal/animal_manager.py
JoeWaddell: This is actually something I have noted to review in the next PR: this method is called by `get_life_cycle_output`, which is actually no longer called. I think it was deprecated by one of the previous animal manager or life cycle refactors, but I need to confirm this with @ltp23 before removing this and the other method (given the age of these sections of code, I think it's likely they won't be reused).

Fjordrunner: We can leave this for the next PR but this is definitely something that should be removed in the next PR if this code will never be used.

Fjordrunner (approved):
Looks good, using Any in the type hints for the two places we discussed in our meeting seems like the appropriate choice given the complexity of the dictionaries. If after speaking with Niko or Loi about the manure management json structure, they say that the example one is not realistic, then I think we should add a type hint to the `manure_management_scenarios` argument in the `init_pens` function.

PR #888, opened 2023-11-01
Update metadata files with new Nasem Comp with TDN input file name

Isgotkowitz:
The name of the Nasem Comp with TDN input file went from `NASEM_comp_with_TDN.csv` to `NASEM_Comp_with_TDN.csv` but was not changed in a the metadata files.

Note there is still something wrong with `user_defined_ration_md.json` that is causing it to not run, but that is beyond the scope of this PR.

### DIFF
--- a/RUFAS/scenario_manager.py
+++ b/RUFAS/scenario_manager.py
@@ -46,6 +46,6 @@ class MetadataPaths(TypedDict):
     path: Path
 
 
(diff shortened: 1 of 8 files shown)


PooyaHekmati (approved):
CC @allisterakun

ew3361zh (approved):

PR #889, opened 2023-11-01
Fix for user-defined ration metadata

JoeWaddell:
Renamed the metadata file to remove the _md suffix.

Replaced metadata properties: the user-defined metadata file had an old metadata properties section.

Update the feed json, as calf values were still the old IDs.


Isgotkowitz (approved):

KFosterReed (approved):
The updated input files work as expected!

Issue #834, opened 2023-10-09
Full NASEM integration into RuFaS


JoeWaddell:
## Summary
While we currently have NASEM equations for nutrient requirement calculations (see `animal_requirements.py`, and we will soon (see #794) have both NRC and NASEM feed databases in RuFaS, we still have some places in the code that require `DE`, and others that require `TDN`, neither of which are reported on a per feed basis.

In addition, the constraint methods (see `ration_optimizer.py`, previously `ration_NLP.py`) still use the NRC equations.

## Current behavior
The value in `config.nutrient_standard` value determines which set of requirement calculations are used.

To accommodate the use of the NASEM standard, we have copied in TDN values for feed items that are also found in the NRC_comp file, and have temporarily removed items without those values.

We are also currently using the DE_base value in place of DE, which is a rough estimate of DE.

## Proposed behavior
1) Where DE is required/reported, we need to add additional logic to calculate DE from DE_Base, for the cases when we use the NASEM_comp.csv 

2) Add new series of constraint methods in `ration_optimizer`, such that TDN and DE of individual items is no longer required, and we can use the full (e.g. unmodified) NASEM_comp.csv. The value in `config.nutrient_standard` value would determine which set of constraint methods must be used.

3) Add logic to only report TDN if using NRC nutrient standard.

Ed731551:
As DE values must be reported on a total diet basis instead of calculations for individual feeds, the way to tackle this is first to calculate the DE_Base by summing up one-by-one these from tabulated values for obtaining DE_Base for the total diet (here there is no effects of additional animal or diet-related factors) and then apply the BW, NDF and starch adjustments to that single value in order to get DE of the diet tailored for the targeted animal category.

PR #892, opened 2023-11-01
Expanding NASEM Comp with TDN

JoeWaddell:
Following advisement from Haowen and Edward, I've expanded the temporary NASEM_Comp_with_TDN.csv (see Issue #834) with items with a `Fd_Category` of `Vitamins/Minerals`, giving them a TDN of 0.0.

I've also updated `user_feeds.csv` to reflect where each feed item can be found (NASEM only, NRC only, or both).

Note that all of those have a `Feed_type` of Conc: I want to be sure this is valid, given the `Feed_type` in `user_feeds.csv`.


tomhuhh (approved):
LGTM! Thanks Joe!

KFosterReed:
Hey team, I was looking at some outputs from a pilot test tonight and realized that the relationship between the DM amount and As Fed amount was off which led me to the small changes in the ration driver... As fed (kg) = DM(kg)/(DM%/100) where as previously it was  As fed (kg) = DM(kg) * (DM%/100). I think I even successfully updated the unit test? 

Sneaking this in here probably isn't best practice but I thought adding it to an already small PR was acceptable if not just a minor sin.

KFosterReed (approved):

JoeWaddell:
The change from multiplication to division led to a divide by 0 error (for when available feeds aren't present in a given ration), so I made a quick fix here. Should be good to merge, but I'll wait until @KFosterReed or @tomhuhh can take a quick look again, just to double-check this fix on top of a fix!

KFosterReed (approved):
Good catch Joe!

Ed731551 (commented):
To me implemented logic makes sense. Thanks

Issue #856, opened 2023-10-13
Implement loading pool functionality to output manager


PooyaHekmati:
We are developing Graphics Generation which works downstream to the output manager. We will also implement post-processing. To allow users to use these features without needing to re-run the simulation, it would be nice to have a function in the output manager to reload the pool.

ew3361zh:
Would it work to have the option to forgo flushing the OM pools instead of flushing them and then reloading them?

ew3361zh:
Closed with PR #896.

PR #896, opened 2023-11-01
Reload om pool

ew3361zh:
This PR seeks to address issue #856 which creates a function to reload the data from the all_variables.json file produced from a simulation into a variable for the user to process further at their leisure.

- [x] function to search output folder for all_variables.json file
- [x] function to open this file and load it to a variable
- [x] unit testing
- [x] remove print statements
- [x] remove pandas import in main.py
- [x] remove call to function in main.py (only used for testing purposes)  

Questions:
1. What if there are multiple all_variables files in the output folder from multiple simulations? Always select most recent? This is the current behavior.
2. What if there are no all_variables files in the output folder? Currently returning `None`.


ew3361zh:
Interesting that this test is failing here but passing locally. I think it has to do with the os.path.getctime which isn't showing the 2nd file mocked as being created after the first one. Going to look into this further:

Failing in GitHub:
Screenshot 2023-11-02 at 4 04 02 PM

Passing locally:
Screenshot 2023-11-02 at 4 07 24 PM

main.py
```diff
@@ -137,6 +140,56 @@ def execute_simulations(
...
 
 
+def reload_pool(directory_path: str = "output/", file_pattern: str = "*all_variables*") -> Dict[str, Any]:
+    """Reloads the Output Manager variables pool.
+
+    Parameters
+    ----------
+    directory_path : str, optional
+        The path to the directory to be searched.
+    file_pattern : str, optional
```
Isgotkowitz:
```suggestion
    directory_path : str, optional, default="output/"
        The path to the directory to be searched.
    file_pattern : str, optional, default="*all_variables*"
```

main.py
```diff
@@ -137,6 +140,56 @@ def execute_simulations(
...
+
+
+def get_filepath(directory_path: str = "output/", file_pattern: str = "*all_variables*") -> str:
+    """Searches the given directory for a file containing the pattern provided.
+
+    Parameters
+    ----------
+    directory_path : str, optional
+        The path to the directory to be searched.
+    file_pattern : str, optional
```
Isgotkowitz:
```suggestion
    directory_path : str, optional, default="output/"
        The path to the directory to be searched.
    file_pattern : str, optional, default="*all_variables*"
```

Isgotkowitz (approved):
Looks great, only comments are suggestions for adding defaults to docstrings.

PooyaHekmati:
> Questions:
> 
> 1. What if there are multiple all_variables files in the output folder from multiple simulations? Always select most recent? This is the current behavior.
> 2. What if there are no all_variables files in the output folder? Currently returning `None`.

Why not getting the path to the file from the user? That way there are no confusion about how to handle missing or multiple files.
Also, I believe these functions belong to the output manager. Why are they in main.py?

ew3361zh:
> > Questions:
> > 
> > 1. What if there are multiple all_variables files in the output folder from multiple simulations? Always select most recent? This is the current behavior.
> > 2. What if there are no all_variables files in the output folder? Currently returning `None`.
> 
> Why not getting the path to the file from the user? That way there are no confusion about how to handle missing or multiple files. 

I think that would be ideal. From our initial discussion on this it sounded like it wasn't clear exactly where/how this functionality would be implemented.

> Also, I believe these functions belong to the output manager. Why are they in main.py?

I think I wasn't sure if we needed/wanted to reinstantiate OM for just the purpose of reloading the data from the variables pool but I can/will move them to OM.

PooyaHekmati:
> > > Questions:
> > > 
> > > 1. What if there are multiple all_variables files in the output folder from multiple simulations? Always select most recent? This is the current behavior.
> > > 2. What if there are no all_variables files in the output folder? Currently returning `None`.
> > 
> > 
> > Why not getting the path to the file from the user? That way there are no confusion about how to handle missing or multiple files.
> 
> I think that would be ideal. From our initial discussion on this it sounded like it wasn't clear exactly where/how this functionality would be implemented.

Yes, it wasn't clear and it still is not. In cases like this, we need to deliver something that has minimum implicit assumptions and it works, then based on the feedback that we get from users/SMEs, we can work on making it more user-friendly (similar to MVP vs V1). 
> > Also, I believe these functions belong to the output manager. Why are they in main.py?
> 
> I think I wasn't sure if we needed/wanted to reinstantiate OM for just the purpose of reloading the data from the variables pool but I can/will move them to OM.

Generally speaking, when you don't want to reinstantiate a class, you can define the functions using `@staticmethod` decorator. But here, you need to reinstantiate OM anyway, because how can you load data to the pool if there is no pool? Also, while trying reinstantiate and load the file, you can report the logs to OM itself.

main.py
ew3361zh: thanks, Ed, we've updated these functions so they don't take these args anymore.

main.py
ew3361zh: same as above

main.py
```diff
@@ -41,37 +42,44 @@ def run_rufas(
     exclude_info_maps: bool = False,
     only_run_validation: bool = False,
     graphics_dir: Path = Path(""),
+    vars_file_path: str = None,
```
PooyaHekmati: use Path instead of str

main.py
```diff
@@ -41,37 +42,44 @@ def run_rufas(
...
+    only_run_validation : bool, optional, default=False
+        Validate input data and don't run a simulation.
+    graphics_dir : Path, optional, default=Path("")
+        The directory for saving graphics.
+    vars_file_path : str, optional, default=None
+        The path to the variables file to be reloaded into the Output Manager's variables pool.
     """
+    if vars_file_path:
+        reload_var_pool(vars_file_path)
+        return
```
PooyaHekmati: move this block to after `sys.stdout.write("RuFaS: Ruminant Farm Systems Model 2023\n")` in line 76

main.py
```diff
@@ -41,37 +42,44 @@ def run_rufas(
...
     if clear_output:
         output_dir = Path(config.global_variables.OUT_DIR)
         keep_list = [".keep", "output_filters"]
         Utility.empty_dir(output_dir, keep=keep_list)
 
     sys.stdout.write("RuFaS: Ruminant Farm Systems Model 2023\n")
-    metadata_file_list: List[MetadataPaths] = METADATA_PATHS
+    metadata_files: List[MetadataPaths] = METADATA_PATHS
```
PooyaHekmati, line 78: why?

main.py
```diff
@@ -80,6 +88,19 @@ def run_rufas(
         )
 
 
+def reload_var_pool(vars_file_path: str) -> None:
```
PooyaHekmati: use Path instead of str

main.py
```diff
@@ -80,6 +88,19 @@ def run_rufas(
...
+    """Instantiates Output Manager and triggers reloading of the variables pool from the provided file path.
+
+    Parameters
+    ----------
+    reload_pool_path : str
+        The file path to the variables file to reload into the Output Manager's variables pool.
+    """
+    output_manager = OutputManager()
+    vars_file_path = Path("output/" + vars_file_path)
+    output_manager.reload_pool(vars_file_path)
```
PooyaHekmati: Don't modify `vars_file_path` here. This code assumes/forces the file is always in `output` dir. What if the user wants to have the file in somewhere else?
Also, we need to trigger some actions after loading the pool.
```suggestion
    output_manager.set_metadata_prefix("reloaded_data") #or something similar
    output_manager.reload_pool(vars_file_path)
    output_manager.dump_all_nondata_pools(
            r"output", exclude_info_maps, format_option
        )
        output_manager.save_variables(
            Path(r"output"),
            Path(r"output/output_filters/"),
            exclude_info_maps,
            produce_graphics,
            graphics_dir,
        )
```

main.py
```diff
@@ -265,6 +286,11 @@ def parse_gnu_args() -> argparse.Namespace:
         help="Only validate the data, don't run a simulation",
         action="store_true",
     )
+    parser.add_argument(
+        "-r",
+        "--reload-pool",
+        help="Reload the output manager's variables pool from provided path",
```
PooyaHekmati: what do you think about `-l` for load instead of reload?

RUFAS/output_manager.py
```diff
@@ -1002,3 +1002,28 @@ def flush_pools(self) -> None:
         self.warnings_pool: Dict[str, OutputManager.pool_element_type] = {}
         self.errors_pool: Dict[str, OutputManager.pool_element_type] = {}
         self.logs_pool: Dict[str, OutputManager.pool_element_type] = {}
+
+    def reload_pool(self, file_path: Path) -> None:
+        """Reloads the Output Manager variables pool from file path provided by user in gnu arg.
+
+        Parameters
+        ----------
+        file_path : str
```
```suggestion
        file_path : Path
```

RUFAS/output_manager.py
```diff
@@ -1002,3 +1002,28 @@ def flush_pools(self) -> None:
...
+                    "function": self.reload_pool.__name__,
+                    }
+        self.add_log("open_json_file", f"Attempting to open {str(file_path)}.", info_map)
+        try:
+            with open(file_path) as file:
+                self.variables_pool = json.load(file)
+                self.add_log("load_data_successful", f"Successfully loaded data from {str(file_path)}.",
+                             info_map)
+        except Exception as e:
+            raise e
```
PooyaHekmati: 1. when you are not doing anything with the exception and you just re-raise it, you don't need to make an object for it.
2. `reload_var_pool` in main.py does not have a try-catch block. Raising an exception here makes the program crash without letting the user know why that happens. Instead of raising, add an error to the errors pool.
3. Avoid generic exceptions as much as possible. The exceptions that `open` and `json.load` can raise is documented, catch them specifically. 
```suggestion
        except Exception:
            raise
```

PooyaHekmati (changes requested):

main.py
ew3361zh: The reason I did this was Path objects aren't falsy. It seemed cleaner to check the string path provided vs the Path object against Path("") if the user didn't provide anything. 

I can change the `if vars_file_path:` to either `if vars_file_path != Path(""):` 
or we could try `if vars_file_path.is_file()`

main.py
ew3361zh: If we move it there we could have a scenario where the user accidentally clears the output folder containing the variables file they want to reload though.

main.py
ew3361zh: It seemed redundant to call it a list with the type annotation immediately after specifying it was a list.

main.py
ew3361zh: I've got these added. Was making changes as you did this review.

main.py
ew3361zh: I like that.

RUFAS/output_manager.py
ew3361zh: I will add more specific exceptions for open and json.load.

main.py
PooyaHekmati: The information that `vars_file_path` is meant to carry is the path to the variables file. The answer to the question: should we even do this? is not a path to the variables file. Therefore, using `vars_file_path` as the trigger to call `reload_var_pool` is not correct.

main.py
PooyaHekmati:
```if clear_output:
    if reload_vars and vars_file_path is the same as output:
        om.add_error("can't clear the output")
    else:
        clear_output()
```

main.py
ew3361zh: To clarify: if we use the load_pool gnu arg to hold the vars_file_path, it's not good practice to also use it as a boolean. So how about if I switched the load_pool gnu arg to a boolean and then if it's True, getting the vars_file_path from the user at that point?

Screenshot 2023-11-08 at 5 14 04 PM

main.py
ew3361zh: Updated

ew3361zh:
Updates:

- [x] Addresses issue #902 (seemed like a good quick fit with other work done in PR)
- [x] Converts `clear_output` code into a function
- [x] Adds `is_file_in_dir` function to check if the provided file path from the user is in the output directory in cases where the user selects both the -c and -l gnu args.
- [x] Updates unit testing and docstrings accordingly

RUFAS/output_manager.py
```diff
@@ -1002,3 +1002,32 @@ def flush_pools(self) -> None:
         self.warnings_pool: Dict[str, OutputManager.pool_element_type] = {}
         self.errors_pool: Dict[str, OutputManager.pool_element_type] = {}
         self.logs_pool: Dict[str, OutputManager.pool_element_type] = {}
+
+    def reload_vars_pool(self, file_path: Path) -> None:
+        """Reloads the Output Manager variables pool from file path provided by user in gnu arg.
```
```suggestion
    def load_variables_pool_from_file(self, file_path: Path) -> None:
        """Loads the Output Manager variables pool from the file path provided by the user in gnu arg.
```

main.py
```diff
@@ -80,6 +89,87 @@ def run_rufas(
...
+    vars_file_path : Path, optional, default=None
+        Path to file used to load Output Manager vars pool.
+    """
+    info_map = {
+        "class": "No caller class",
+        "function": clear_output_dir.__name__,
+    }
+    output_manager = OutputManager()
+    output_dir = Path(config.global_variables.OUT_DIR)
+    is_in_dir = is_file_in_dir(output_dir, vars_file_path)
```
PooyaHekmati: can we give a better name to this variable to indicate is what in the directory?

main.py
```diff
@@ -80,6 +89,87 @@ def run_rufas(
...
+    if is_in_dir:
+        output_manager.add_error("Can't clear output directory", f"{vars_file_path} in output directory.", info_map)
+    else:
+        keep_list = [".keep", "output_filters"]
+        Utility.empty_dir(output_dir, keep=keep_list)
+        output_manager.add_log("Output directory cleared", "No conflicts to clearing output directory.", info_map)
+
+
+def is_file_in_dir(dir_path: Path = Path(config.global_variables.OUT_DIR), vars_file_path: Path = None) -> bool:
+    """Checks if a file path is in the provided directory.
```
PooyaHekmati: if this function checks "a" file path, then why is the name of the arg specific to vars?

main.py
```diff
@@ -80,6 +89,87 @@ def run_rufas(
...
+    output_manager.dump_all_nondata_pools(
+            r"output", exclude_info_maps, format_option
+        )
+    output_manager.save_variables(
+            Path(r"output"),
+            Path(r"output/output_filters/"),
+            exclude_info_maps,
+            produce_graphics,
+            graphics_dir,
+        )
```
PooyaHekmati: `save_variables` generates some logs and warnings as well, so it should be called before `dump_all_nondata_pools`.

PooyaHekmati (approved):

Issue #857, opened 2023-10-13
.gitignore is not working


PooyaHekmati:
There have been multiple cases where a file has been changed and synced, while it is mentioned in the gitignore. Investigate the problem.

Possibly related:
https://stackoverflow.com/questions/25436312/gitignore-not-working
https://gist.github.com/pavankjadda/2bb6fbdd8786e1f57fd7bcbcc666b51d
https://pavankjadda.dev/how-to-fix-gitignore-not-working-issue/

Isgotkowitz:
@PooyaHekmati could you give an example of a file that this is happening with?

PooyaHekmati:
> @PooyaHekmati could you give an example of a file that this is happening with?

An example is PR #751 deleted `output/output_filters/_json_all_variables.txt` while we had `output/output_filters/*` in the gitignore.

ltp23:
I think one working solution for me I just found is to do things one directory level at a time. 

For example

```
# Exclude everything inside the output directory
output/*

# Re-include the output_filters directory
!output/output_filters/

# Exclude everything inside the output_filters directory
output/output_filters/*

# Re-include the needed file
!output/output_filters/json_test.txt
```

Isgotkowitz:
I have tried the solution outlined in the links above. I think it has the desired effect but there are some important things to note about the behavior.

To expand, I tried this solution on branch_a and it made RUFAS/scenario_manager.py untracked without deleting it - which is great because now I can change it freely and not worry about it finding its way into my PRs! The only issue is that when I switch to another branch where I have not uncached everything, call it branch_b, RUFAS/scenario_manager.py disappears on branch_a. Once I clear the cache on branch_b though, I can switch back to branch_a, recreate RUFAS/scenario_manager.py, and switch between branch_a and branch_b with RUFAS/scenario_manager.py being kept.

In short, the linked solution works but requires clearing the cache on every branch you're working on and recreating any files that you have lost.

ew3361zh:
> I have tried the solution outlined in the links above. I think it has the desired effect but there are some important things to note about the behavior.
> 
[...]

PR #911, opened 2023-11-06
Ignore tracked files

ew3361zh:
This PR seeks to address issue #857. 
As is stated in the issue conversation, .gitignore is working as intended but is not the right tool for this specific job. 

Summed up: there are specific files that we want tracked in GitHub but for which we want to prevent users from pushing local changes accidentally. These are files like scenario_manager.py, metadata files, output filter files. For some of these cases, like the output filter files, it makes more sense to remove the files from being tracked in GitHub rather than having them protected.

Because there isn't a built-in access-control method to allow people to make untracked local changes, this PR develops a workaround:

- [x] Adds git action to list protected file paths and tells github actions to fail linting if any of these file paths is changed (at all).
- [x] Removes all output filter files from output_filters folder. 
- [x] Adds 1 file to protected list to start: scenario_manager.py. If there are others, please comment here and we can add them. Note: the default metadata and default input files will eventually be good candidates for being on the protected list but with incoming changes in the next few weeks, it does not make sense to add them now. If this PR's solution to the issue is approved, I will make an issue to add these input files at an appropriate future date.
- [x] Write a quick-start Output Manager section in the wiki which has some easy set-up steps for capturing and filtering output.


.github/workflows/lint_flake8.yml
```diff
@@ -34,8 +34,34 @@ jobs:
...
     - name: Lint with flake8
       if: ${{ env.FILES != '' }}
       run: |
         # stop the build if there are Python syntax errors or undefined names
         flake8 ${{ env.FILES }}
+    - name: Check for Unauthorized Changes
+      run: |
+        paths_to_protect=(
+          'RUFAS/scenario_manager.py'
+          'input/metadata/default_metadata.json'
```
Isgotkowitz: I don't think the default metadata file should be included here. Even though we really don't want people accidentally changing it, it seems that there will still be some deliberate and necessary changes to it in the next couple weeks.

.github/workflows/lint_flake8.yml
```diff
@@ -34,8 +34,34 @@ jobs:
                         tr " " "\n" |
                         grep -E "\.py$" |
                         tr "\n" " ")" >> "$GITHUB_ENV"
+        echo "OTHER_FILES=$( echo ${{ steps.files.outputs.added_modified }} |
+                        tr " " "\n" |
+                        grep -E "\.txt$|\.json$"  |
```
Isgotkowitz: Should CSVs be looked for here too?

Isgotkowitz (changes requested):
Looks good. I have two questions re. which files should be checked, which I'd like to discuss then will approve.

.github/workflows/lint_flake8.yml
ew3361zh: That's a fair point. After those changes are implemented, do you think it's a helpful file to have on this list? It seems like a file we don't want alterations to in general.

.github/workflows/lint_flake8.yml
ew3361zh: Good call. I can add those.

ew3361zh:
> Looks good. I have two questions re. which files should be checked, which I'd like to discuss then will approve.

@Isgotkowitz Wondered if you had any additional specific files you thought of that should be included in this protected files category?

.github/workflows/lint_flake8.yml
Isgotkowitz: > After those changes are implemented, do you think it's a helpful file to have on this list? It seems like a file we don't want alterations to in general.

Yes, after the changes have been made I think so. When we get to that point it might be important to include the default files that it uses too.

Isgotkowitz:
> > Looks good. I have two questions re. which files should be checked, which I'd like to discuss then will approve.
> 
> @Isgotkowitz Wondered if you had any additional specific files you thought of that should be included in this protected files category?

@ew3361zh if the default metadata is protected, it may be good to have all the input files used by the default metadata also be protected.

Isgotkowitz (approved):

PooyaHekmati (approved):

PR #915, opened 2023-11-07
Fix water buildup

### LINKED ISSUE #766: Evaluate soil hydrology
Evaluate the behavior of the Soil module with respect to the hydrological processes, including
- percolation between layers
- runoff from and infiltration of soil surface
- erosion of soil

This evaluation will be considered complete when approval is given by Pete, Eric and Xuesong.

### LINKED ISSUE #887: Water Content in Surface Layer of Soil Exceeding Saturation Point
Water content in the surface layer of the soil profile is going above saturation point and in some cases way beyond physical limits.

## Expected Behavior

The water content of a soil layer should not exceed its saturation capacity after it has gone through all hydrologic processes (infiltration, percolation, evaporation, etc).

## Current Behavior

Water content is reaching levels above saturation point and physical capacity for the surface soil layer.

Below are some graphs demonstrating this behavior from the barnyard simulation.
Regular soil field:
regular_soil_surface_layer_water_bug

Sandy soil field:
sandy_surface_water_content_bug

## Possible Solution

TBD

## Steps to Reproduce

1. Run the barnyard simulation
2. Observe results
3.
4.

Isgotkowitz:
This PR eliminates the issue of water building up in the surface soil layer during periods of high precipitation. It also concludes the evaluation of the RuFaS hydrology model.

This fix is accomplished with a method for percolating water that is only called when there is enough water to oversaturate the surface soil layer, `_percolate_excess_water()`, in the `Percolation` module. The `Infiltration` module has been modified so that it _only_ handles calculating the infiltration and leaves adding water to the soil entirely to the `Percolation` module. Testing has been added or updated for all modified methods in `Infiltration` and `Percolation`. Coverage for both modules remains at 100%

There are a number of other small changes made in this PR, including:
- Fixing the water characteristics for one layer of the soil profile in one of barnyard soil profile inputs.
- Adding a single test case to `test_cycle_water()` in `test_field.py`, bringing coverage of `field.py` up to 100%.
- Setting the starting value for runoff to 0.
- Updates the README with progress made.

Closes #887, closes #766


RUFAS/routines/field/soil/infiltration.py
```diff
@@ -74,7 +74,6 @@ def infiltrate(self, rainfall: float) -> None:
         self.data.accumulated_runoff = min(rainfall, self._determine_accumulated_runoff(rainfall, retention_parameter))
         infiltrated_water = max(0.0, rainfall - self.data.accumulated_runoff)
         self.data.infiltrated_water = infiltrated_water
-        self.data.soil_layers[0].water_content += infiltrated_water
```
matthew7838, line 77: Good change to segregate functionality in my opinion.

RUFAS/routines/field/soil/percolation.py
```diff
@@ -71,6 +77,33 @@ def percolate(self, has_seasonal_high_water_table: bool) -> None:
...
+        The amount of water allowed to infiltrate the soil on any given day is based on the available capacity of the
+        entire soil profile. So when there is an extreme amount of infiltration or there are multiple days of high
+        infiltration in a row, this method ensures that the excess water will be distributed appropriately throughout
+        the entire soil profile.
+
+        """
+        self.data.set_vectorized_layer_attribute("percolated_water", [0.0] * len(self.data.soil_layers))
+        water_remaining_to_percolate = self.data.infiltrated_water
+        for layer in self.data.soil_layers:
+            acceptable_percolation = layer.acceptable_percolation_amount
```
matthew7838, line 95: Don't think this line is needed since I don't see the value being modified anywhere.

matthew7838 (approved):
LGTM, one small comment. Great work maintaining test coverage.

RUFAS/routines/field/soil/infiltration.py
Isgotkowitz: Thanks:)

RUFAS/routines/field/soil/percolation.py
Isgotkowitz: I assume you mean line 95. It isn't needed in the code, but it made the test way easier to write. If you think it makes the code more confusing though, I can eliminate this line and rewrite the test.

matthew7838: Sorry. Yes I mean line 95, don't think it's worth the hassle.

tomhuhh (approved):
Thanks Ed for trusting me reviewing this PR. The code looks good to me.  I am attaching some results from the simulations that I ran earlier. From what Ed explained to me this afternoon, these results also make sense. 

image

image

image

Issue #900, opened 2023-11-02
Handle multiple graphs per filter file


PooyaHekmati:
Currently, for a new graph, a new filter file is needed. Explore the possibility of defining multiple graphs in a single filter file.

Issue #895, opened 2023-11-01
Split test_save_variables


PooyaHekmati:
The function `test_save_variables` in test_misc.py is too big. It has too many asserts, and too many lines of code. Split this function into smaller functions.

PR #926, opened 2023-11-09
Handle multi graphs

PooyaHekmati:
This PR addresses Issue #895 and Issue #900. 
The formatting of graph filters is extended to handle `multiple` graphs in a single json file. The code is backward compatible.
I extracted `_route_save_functions` from `save_variables` to make it more manageable and testable. So, both:
```
{
  "type": "stackplot",
  "filters": ["Cow\\.milking_update\\.milk_data_at_milk_update"],
  "variables": ["milk_lactose", "milk_protein", "milk_fat"],
  "title": "Flat Json Daily Milk Nutritions",
  "legend": ["lactose", "fat", "protein"]
}
```
and
```
{
  "multiple": [
    {
      "type": "plot",
      "filters": [
        "Pen\\.calc_total_manure\\.daily_aggregate_calf_(degradable|non_degradable)_volatile_solids"
      ],
      "title": "Daily Aggregate Calf Degradable Volatile Solids",
      "legend": ["Degradable Volatile Solids", "Non-degradable Volatile Solids"]
    },
    {
      "type": "stackplot",
      "filters": ["Cow\\.milking_update\\.milk_data_at_milk_update"],
      "variables": ["milk_lactose", "milk_protein", "milk_fat"],
      "title": "Daily Milk Nutritions",
      "legend": ["lactose", "fat", "protein"]
    }
  ]
}

```
formats are acceptable. If the key `multiple` is present, then the value is expected to be a list of json blobs.


tests/test_misc.py
```diff
@@ -1699,170 +1692,217 @@ def test_filter_variables_pool_exclude_regex_patterns(
...
-            graphics_dir=Path("graphics"),
+        mock_output_manager._route_save_functions(
+            "graph_file",
+            "save_path",
+            {"key": {"var": "value"}},
+            True,
+            graph_data,
+            "graphics_dir",
         )
         mock_generate_graph.assert_called_once_with(
```
Isgotkowitz:
```suggestion
        )
        assert mock_output_manager.add_warning.call_count == 1
        mock_generate_graph.assert_called_once_with(
```
This would ensure that `add_warning` is not somehow being called when it shouldn't be.

RUFAS/output_manager.py
```diff
@@ -761,59 +766,88 @@ def save_variables(
...
+            for _, supported_prefix in self.__supported_filter_types_prefixes.items():
+                if filter_file.startswith(supported_prefix):
+                    break
+            else:
+                self.add_warning(
+                    "invalid filter file prefix",
+                    f"{filter_file} prefix is not in {list(self.__supported_filter_types_prefixes.values())}",
                     info_map,
                 )
                 continue
```
Isgotkowitz, line 780:
```suggestion
                if filter_file.startswith(supported_prefix):
                    break
                else:
                    self.add_warning(
                        "invalid filter file prefix",
                        f"{filter_file} prefix is not in {list(self.__supported_filter_types_prefixes.values())}",
                        info_map,
                    )
                    continue
```
Should if and else statements be on the same level of indentation?

RUFAS/output_manager.py
```diff
@@ -761,59 +766,88 @@ def save_variables(
...
-            input_path = os.path.join(dir_path, filter_file)
-            filter_content = self._load_filter_file_content(input_path)
-            if "filters" not in filter_content.keys():
-                self.add_error(
-                    "Missing filters entry",
-                    f"'filters' does not exist in {filter_file}",
+            for _, supported_prefix in self.__supported_filter_types_prefixes.items():
+                if filter_file.startswith(supported_prefix):
+                    break
+            else:
```
JoeWaddell, line 774: I can't seem to get the warning below to trigger, testing with bad filter names. Is this else statement indented too far to the left? I think I'm misunderstanding part of the flow here.

JoeWaddell (approved):

RUFAS/output_manager.py
PooyaHekmati: It is working:
image
Yes, the else statement belongs to the for-loop. See https://book.pythontips.com/en/latest/for_-_else.html#else-clause

JoeWaddell: Thanks for the link: I was looking in the warnings file generated, not the console. Checking the console I see it fails as expected.
image

Isgotkowitz (changes requested):
When graph filter files are incorrectly constructed, can errors/warnings be recorded for them? I tried changing the names of keys and passing invalid types of plots, and there were no errors or warnings raised that the graph filters were incorrect.

Edit: just looked through replies to Joe's comments. Got error messages printing by adding `-v errors`. Can the error messages be added to the files as well as being printed to standard out?

Also left two suggestions, one for a test and another for code formatting, both take it or leave it.

RUFAS/output_manager.py
PooyaHekmati: the else belongs to the for loop not the if statement.

Isgotkowitz: Saw reply to Joe, disregard this. Neat pythonic trick to know about though.

PooyaHekmati:
> When graph filter files are incorrectly constructed, can errors/warnings be recorded for them? I tried changing the names of keys and passing invalid types of plots, and there were no errors or warnings raised that the graph filters were incorrect.

That's a good catch, but out of scope of this PR. Can you please make an issue and document what you did and what didn't work?

> Edit: just looked through replies to Joe's comments. Got error messages printing by adding `-v errors`. Can the error messages be added to the files as well as being printed to standard out?

Yes, that was happening because the functions were being called in the wrong order. It should be fixed now.

Isgotkowitz (approved):

allisterakun (changes requested):
All the functionalities are good; I found an issue handling multiple graphs in the same filter file.
To reproduce:
1. Delete the `title` for both graphs in `graph_filter_example.json`.
2. Run the simulation
As a result, only one graph will be saved to the graph output directory. This is because the auto-generated filename is the same for both graphs, so that the second graph would overwrite the first PNG file. When no title is specified for a graph, a default file name will be generated as 
```
timestamp: str = datetime.datetime.now().strftime("%d-%b-%Y_%a_%H-%M-%S")
filename = f"saved_graph_{filter_file_name}-{timestamp}.png"
```
Both graphs are in the same filter file => filter_file_name is the same,
the timestamp is only precise to seconds; multiple graphs in the same filter file will likely have the same timestamp.

Therefore, the identical filenames for multiple graphs in the same filter file will cause the later graphs to overwrite the previously saved png graphs.

PooyaHekmati:
> All the functionalities are good; I found an issue handling multiple graphs in the same filter file. To reproduce:
> 
> 1. Delete the `title` for both graphs in `graph_filter_example.json`.
> 2. Run the simulation
>    As a result, only one graph will be saved to the graph output directory. This is because the auto-generated filename is the same for both graphs, so that the second graph would overwrite the first PNG file. When no title is specified for a graph, a default file name will be generated as
> 
> ```
> timestamp: str = datetime.datetime.now().strftime("%d-%b-%Y_%a_%H-%M-%S")
> filename = f"saved_graph_{filter_file_name}-{timestamp}.png"
> ```
> 
> Both graphs are in the same filter file => filter_file_name is the same, the timestamp is only precise to seconds; multiple graphs in the same filter file will likely have the same timestamp.
> 
> Therefore, the identical filenames for multiple graphs in the same filter file will cause the later graphs to overwrite the previously saved png graphs.

Good catch, fixed

allisterakun (approved):
All good now!

PR #929, opened 2023-11-10
Om improvement

PooyaHekmati:
3 things are happening here:
1. The function `save_variables` is now doing more than just saving the content of the variables pool. I renamed it to `save_results` to better capture what it does and disambiguate it.
1. In my last PR, improved the logic that checks the filter filename prefixes to only load those that are supported. This seems out of the scope of `save_results`, so I moved this logic to where we are listing the filter files. After this change, `_list_txt_and_json_files_in_dir` is only listing filter files, so I renamed it to `_list_filter_files_in_dir`.
1. It also brings the output manager's test coverage back to 100%.


allisterakun (approved):
Looks Good!

tests/test_misc.py
```diff
@@ -1889,7 +1906,11 @@ def test_route_save_functions_graph(
             graph_data,
             "graphics_dir",
         )
-        assert mock_output_manager.add_warning.call_count == 1
+        mock_output_manager.add_warning.assert_called_once_with(
```
Isgotkowitz, line 1916: Nice, increased precision is always great to see.

Isgotkowitz (approved):
Looks great, nice job Pooya

Issue #609, opened 2023-06-28
Negative values in manure excretion results for lactating cows and growing heifers


ltp23:
During the calculations of manure excretion values, there are some attributes that come out negative. The equations used in these calculations involve the results from the ration report, so depending on the dynamics between those results and hard-coded constants, we can get negative values. 

## Expected Behavior

All the manure excretion values should be at least non-negative, as these values will be passed on to the manure module. 

## Current Behavior

An example list of excretion values for a lactating cow that shows this problem is given below.
<img width="646" alt="Screenshot 2023-06-28 at 12 41 42 PM" src="https://github.com/RuminantFarmSystems/MASM/assets/98855117/c2fa4af5-6443-4187-a9ae-54d523020825">

## Possible Solution

- Adjust the equations in the manure calculations to accommodate the results from the ration report
- Set sensible defaults or limits to ensure non-negative values

## Steps to Reproduce

From my observations, the two files `lactating_cow_manure_excretion.py` and `growing_heifer_manure_excretion.py` encounter this issue, so to reproduce that I would add some code like the following and then place a breakpoint inside the if-statement.

```
for key, value in manure_excretion_values.items():
    if value < 0:
        print(key, value)  # Place a breakpoint here
```

## Context (Environment)

Currently, any negative manure excretion values will be zeroed out when passing inside the manure module just so that the program can still run. This is definitely not ideal because the affected results from the manure module now would be unusable.

tomhuhh:
My hunch is that the inaccurate diet in the default setting results in the negative values in manure excretions. That said, I can  work on defining limits for the inputs of these equations. Thanks!

tomhuhh:
Current progress: I noticed the negative manure outputs are associated with 0 DMI. @JoeWaddell and I looked into this and so far we discovered some negative requirements for animals that seem to be newly purchased. Will continue troubleshooting.

ew3361zh:
Adding a note here that this PR is mostly resolved by PR #1133, however to address the TODO in the code of pen_manure.py, an exception should be raised in the post-init() - TODO is on line 80. Once that's resolved, the TODO should be removed and then this issue should be closed.

ew3361zh:
- @ew3361zh to review PR #1133 and update progress.
- Review this again after we update #657 as well.

Isgotkowitz:
Blocked by #1680

allisterakun:
Grooming session:
[...]

PR #933, opened 2023-11-10
Updating manure excretions output data passed to OM

### LINKED ISSUE #904: Incorrect number of rows in the aggregate heifer manure outputs
## Expected Behavior

The same number of rows in the aggregate heifer manure outputs compared to other animal categories. 

## Current Behavior

The number of rows for heifers are twice of that of other animal categories. Those integers do not seem to belong there. 

Screenshot 2023-11-03 at 10 48 00 AM

## Possible Solution

Start with investigating how aggregate heifer manure outputs are collected by the output manager.  

## Steps to Reproduce

1. Use the filter 'Pen.calc_total_manure'. 
2.
3.
4.

ltp23:
The PR aims to fix the git issue #904, which is about the inconsistent lengths of csv output columns related to manure excretions in the animal module, specifically for heifers. This happens when in an animal manager's daily update there are multiple pens with overlapping animal groups, the daily update's function call will add the corresponding number of duplicates. 

We happened to observe this with the heifer group, but I think had we had multiple pens of lactating cows, we would have probably seen something similar.

With the current fix, I temporarily store the per-animal-group manure excretion output across all the pen's calculations and then dump the data only once at the end. For the per-pen manure excretion output, this has not changed and still gets evaluated as we iterate over each pen.

The following filter can be used for this new update.
```
.*AnimalManager.daily_updates.daily_aggregate.*
.*Pen.calc_total_manure.*
```


tomhuhh (approved):
LGTM! I tested with the filter Loi provided and the outputs are of the same length across different animal categories. Thanks @ltp23 for your hard work!

RUFAS/routines/animal/pen.py
```diff
@@ -422,7 +422,7 @@ def update_animals(self, new_animals: List[Any],
         self.update_animal_combination(animal_combination)
         self.update_classes_in_pen()
 
-    def calc_manure(self, feed, methane_model): # noqa
+    def calc_manure(self, feed, methane_model):  # noqa
         """
         Calculates the total manure excretion of the animals in the pen,
```
Isgotkowitz: Can this docstring be updated to to NumPy style and the method signature type annotated.

RUFAS/routines/animal/pen.py
```diff
@@ -550,7 +550,7 @@ def daily_p_update(self):
             total_p_animal = max(total_p_animal, 0)
             self.avg_p_animal = total_p_animal / len(self.animals_in_pen)
 
-    def set_up_new_animal(self, animal, p_conc, feed, temp, num_animals_before_additions): # noqa
+    def set_up_new_animal(self, animal, p_conc, feed, temp, num_animals_before_additions):  # noqa
```
Isgotkowitz, line 555: Same here as above, can the method signature be type annotated and the docstring style changed to NumPy

Isgotkowitz (approved):
Code looks good, especially the testing. Two comments, both related to updating docstrings.

While looking through the output, I did have some questions:
- Why do pen specific manure excretion values not distinguish between the animal types that produced them while the aggregate does? Is it assumed that a single pen will only have a single type of cow in them?
- There are negative initial values for some aggregate lactating cow excretion values that seem like they should not be there. This includes the output variables:
  - AnimalManager.daily_updates.daily_aggregate_lactating_cow_urine
  - AnimalManager.daily_updates.daily_aggregate_lactating_cow_manure_mass
  - AnimalManager.daily_updates.daily_aggregate_lactating_cow_total_solids
  - AnimalManager.daily_updates.daily_aggregate_lactating_cow_degradable_volatile_solids
  - AnimalManager.daily_updates.daily_aggregate_lactating_cow_non_degradable_volatile_solids
  - AnimalManager.daily_updates.daily_aggregate_lactating_cow_potassium
Is this a known issue?

ltp23:
> * Why do pen specific manure excretion values not distinguish between the animal types that produced them while the aggregate does? Is it assumed that a single pen will only have a single type of cow in them?

Probably because of the current usage from the SMEs' perspective. Also, one of the complicating factors is the classification of animals as they mature through different stages and change pens. If there is a need for this approach, we can make that happen too.

> * There are negative initial values for some aggregate lactating cow excretion values that seem like they should not be there. This includes the output variables:
>   
>   * AnimalManager.daily_updates.daily_aggregate_lactating_cow_urine
>   * AnimalManager.daily_updates.daily_aggregate_lactating_cow_manure_mass
>   * AnimalManager.daily_updates.daily_aggregate_lactating_cow_total_solids
>   * AnimalManager.daily_updates.daily_aggregate_lactating_cow_degradable_volatile_solids
>   * AnimalManager.daily_updates.daily_aggregate_lactating_cow_non_degradable_volatile_solids
>   * AnimalManager.daily_updates.daily_aggregate_lactating_cow_potassium
>     Is this a known issue?

I think so. We have an open issue #609 that describes this. The negative values come from the unbalance between the calculated values from ration calculations and the formulas we use. However, one temporary, OK fix is to make sure each value is at least non-negative.

RUFAS/routines/animal/pen.py
ltp23: I have updated the docstring for this method, but it needs to be refactored later.

RUFAS/routines/animal/pen.py
ltp23: I will hold off on this one because it will probably be removed during the animal manager refactoring.

ltp23:
Thanks Haowen and Ed!!

PR #935, opened 2023-11-10
Add metadata prefix to graph file names

PooyaHekmati:
This PR adds metadata prefix to the generated graphs' file names, similar to what we get from the output manager.

### DIFF
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -652,7 +652,11 @@ class OutputManager(object):
                     else:
                         result = [json_content]
                 elif path.endswith(".txt"):
-                    list_of_elements = [element for element in filter_file.read().splitlines() if element]
+                    list_of_elements = [
+                        element
+                        for element in filter_file.read().splitlines()
+                        if element
+                    ]
                     result = [{"filters": list_of_elements}]
                 else:
                     raise Exception(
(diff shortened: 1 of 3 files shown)


JoeWaddell (approved):
Wonderful work!

RUFAS/output_manager.py
```diff
@@ -652,7 +652,11 @@ def _load_filter_file_content(self, path: str) -> List[Dict[str, str]]:
                     else:
                         result = [json_content]
                 elif path.endswith(".txt"):
-                    list_of_elements = [element for element in filter_file.read().splitlines() if element]
+                    list_of_elements = [
+                        element
+                        for element in filter_file.read().splitlines()
+                        if element
+                    ]
```
ew3361zh, line 659: I realize this is likely the formatter you're using but strange that this would be the preferred format for list comprehensions.

PooyaHekmati: I agree, sometimes the formatter does things that surprises me. Do you think we should change it back? Although, the next time I apply the auto formating it will become the same as is here.

ew3361zh: I think keeping it is fine. I can adjust :)

ew3361zh (approved):
lgtm!

PR #936, opened 2023-11-10
Animal Manager Refactor m2

JoeWaddell:
Here we're removing deprecated methods, implementing yet-unused methods, and have added a couple new methods. 

Important note: deprecated the use of the function `daily_animal_routine`, and instead call the class method `daily_updates` directly in `simulation_engine.py`. This necessitated some changes to tests.

Of particular note is some of the cleanup in the daily_updates method, specifically the new `collect_manure_excretions_output_data` method.

Also note that all calls to output manager will be handled differently in: https://github.com/RuminantFarmSystems/MASM/pull/947

Between this and #947, `daily_udpates` should be greatly streamlined.

This PR is meant to accomplish outlined objectives for Milestone 2 in the following AnimalManager Refactor Design Doc: https://docs.google.com/document/d/16j-8ylEl-jZ4ZP2fW5djm_R_ROIRlFN3DhckbH0Wkuw/edit?usp=sharing


RUFAS/routines/animal/animal_manager.py
```diff
@@ -1780,6 +1555,16 @@ def _add_animal_to_pen_and_id_map(
         )
         self.animal_to_pen_id_map[animal.id] = pen_with_min_stocking_density.id
 
+    def collect_manure_excretions_output_data(self, pen: Pen, feed: Feed, manure_excretions_output_data: Dict):
+        pen.classes_in_pen = self._get_classes_in_pen(pen)
```
ancheypeng: Is there a reason why the values of `classes_in_pen` and the return value of `_get_classes_in_pen` aren't the same? It's a bit confusing to set the value of an attribute to the return value of the getter function of the same name.

tests/animal_module_tests/test_animal_allocation.py
```diff
@@ -540,3 +543,24 @@ def test_allocate_animals_to_pens(mocker: MockerFixture) -> None:
...
+        config=mocker.MagicMock(),
+        feed=mocker.MagicMock(),
+        weather=mocker.MagicMock(),
+        time=mocker.MagicMock(),
+        )
+    # act
+    animal_manager.set_animal_grouping_scenario(scenario)
+
+    # assert
+    animal_manager.ANIMAL_GROUPING_SCENARIO = scenario
```
ancheypeng: is this supposed to be an assertion?

ltp23 (approved):
The code looks good to me! It's good that we are removing the dangling daily-animal-routine method. I think we also want to remove the State class at some point too.

RUFAS/routines/animal/animal_manager.py
ltp23: So there isn't an exact 1:1 matching between animal classes and animal types (e.g., Cow class objects can be classified as DryCow or LacCow types). When we want to update the `classes_in_pen` field, we would need to re-evaluate the animals in the pen. We can change the name `get` to something else so it doesn't sound like a normal, simple getter.

ancheypeng: Got it! I think changing the name of the function would be helpful

tests/animal_module_tests/test_animal_allocation.py
JoeWaddell: Haha yes that should be...updating now!

RUFAS/routines/animal/animal_manager.py
JoeWaddell: Good series of comments here: how does 'determine' look here?

ancheypeng (approved):
Looks good!

YijingGong (approved):
The code cleaning looks good to me! Also tried to run Rufas with some additional pens, and it works.

Issue #938, opened 2023-11-11
MetadataPaths class is misleading


PooyaHekmati:
The class `MetadataPaths` in the scenario manager only contains one path but both the name and the docstring suggest that it can have multiple. Fix the name and docstring.

Issue #940, opened 2023-11-11
Code repetition in main.py


PooyaHekmati:
the line `for metadata_file in metadata_files:` is repeated twice in main.py; while there is only one needed. Fix this issue by moving this for-loop to `run_rufas`. In other words, `run_validation` and `execute_simulations` should accept only one metadata.

PR #942, opened 2023-11-12
Supercharge RuFaS: Now with Turbo-Charged Parallel Processing! 🚀

PooyaHekmati:
Get ready for a significant speed boost in RuFaS! This PR introduces parallel processing, transforming our simulation and validation process from a single-file marathon to a multi-core sprint. Imagine a team of relay runners instead of a solo marathon runner - that's the kind of efficiency we're talking about!

**Personal Benchmark Achievement**:
On my laptop, which is armed with 16 CPU cores, I observed a dramatic drop in processing time. Running 6 metadata files took 48 seconds before, but now, it's 18 seconds! And the best part? Increasing the metadata files up to 16 doesn’t extend this time. It's like having a sports car that goes faster without burning extra fuel!

**Key Changes:**

1. **Parallel Processing Implementation**:
   - Embraced `multiprocessing` for `execute_simulation` and `run_validation`. This means handling multiple metadata files is now more like a symphony with each musician (read: CPU core) playing their part in harmony.
   - Adopted tuple packing for argument handling, ensuring smooth compatibility with `multiprocessing`'s `imap_unordered` method.

2. **Refinement and Issue Resolution**:

   - Polished argument parsing in `main()` for enhanced clarity. `store_false` actions for boolean flags are now in place, making command-line interactions more intuitive.
   - Addressed and resolved Issue #938 and Issue #940.
   - Fixed a missing link in `run_validation` where `output_manager.set_metadata_prefix(metadata_file["prefix"])` was missed.

**A Thought to Ponder**:
Considering the benefits, should we introduce a GNU arg to allow users the choice of enabling or disabling parallel processing? I’m keen to hear your perspectives on this.

**Testing and Feedback**:
I've tested these changes thoroughly on my end, and the results are promising. However, I invite you all to give it a go and share any insights, observations, or potential oversights.

Looking forward to your valuable feedback as we shift RuFaS into high gear!


JoeWaddell:
Realized I had commented on Slack, but not here: I tested this on multiple machines, and it worked flawlessly.

>  Considering the benefits, should we introduce a GNU arg to allow users the choice of enabling or disabling parallel processing? I’m keen to hear your perspectives on this.

I think this would be good for A) cases in which someone has a unique hardware setup that conflicts with this, and B) cases in which we may want to use a different type of parallel processing (e.g. another Python module).

These two reasons are pure speculation, but do you anticipate that this would continue to work on machines with e.g. 128 cores?

Issue #894, opened 2023-11-01
Separate OutputManager test from test_misc


PooyaHekmati:
The test cases for the output manager are currently in test_misc.py. Extract these test cases and put them in test_output_manager.py.

ew3361zh:
Closed with PR #943

PR #943, opened 2023-11-13
Move Output Manager testing into own test file

ew3361zh:
PR to address issue #894.

Takes all test cases in test_misc.py related to output_manager.py classes and functions and moves them into test_output_manager.py. This includes test cases related to the LogVerbosity class which is in output_manager.py.

This will help clean up test_misc.py and is better practice to keep test cases separated out like this.


PooyaHekmati (approved):

Isgotkowitz (approved):
Looks good, love seeing the big test files get broken down

PR #947, opened 2023-11-13
Animal Manager Refactor m3

JoeWaddell:
Milestone 3 in the following AnimalManager Refactor Design Doc: https://docs.google.com/document/d/16j-8ylEl-jZ4ZP2fW5djm_R_ROIRlFN3DhckbH0Wkuw/edit?usp=sharing

New class: AnimalReporter (looking for name improvements!)

This class serves two main purposes:
A) Simplify existing methods by removing calls to om.
B) Central location for calls to output manager for Animal Module will make it easier to add new output variables (expanding existing code, or noting what is already available).

All reporting methods have unit tests.

Note that the info_map for each of the reporting methods uses the location of the previous location of the information, e.g. where it was last modified. This should help reduce the number of keys in filters that require renaming.

Also note: I've commented out some imports, calls to om, and other test logic, as a temporary measure. If any of those need to be reinstated please comment!


RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -267,12 +268,24 @@ def update_milk_production_history(self, sim_day):
...
-        Args:
+        If milk production history has already been updated for the day,
+        the most recent entry is deleted before appending the latest values.
+        Once a cow reaches 305 days in milk, latest_milk_production_305days is updated.
+
+        Parameter
+        ---------
             sim_day: simulation day
         """
+        if len(self.milk_production_history) > 0 and self.milk_production_history[-1].simulation_day == sim_day:
```
ancheypeng, line 267: Is this a temporary fix or a permanent change? It seems a bit odd to have to delete the most recent entry before adding the new values. Is the `milk_production_history` updated elsewhere as well?

RUFAS/routines/animal/pen.py
```diff
@@ -422,7 +422,7 @@ def update_animals(self, new_animals: List[Any],
         self.update_animal_combination(animal_combination)
         self.update_classes_in_pen()
 
-    def calc_manure(self, feed: Feed, methane_model: str):  # noqa
+    def calc_manure(self, feed, methane_model: str):  # noqa
```
ancheypeng, line 414: why was the type hint removed here?

RUFAS/routines/animal/animal_reporter.py
```diff
@@ -0,0 +1,399 @@
...
+# TODO remove the below if can't resolve circular imports!
+# from RUFAS.routines.animal.pen import Pen
+# from RUFAS.routines.animal.animal_manager import AnimalManager
+from RUFAS.routines.animal.manure.general_manure import AnimalManureExcretions
+
+om = OutputManager()
+
+
+class AnimalReporter:
+    def report_daily_animal_population(animal_manager) -> None:
```
ancheypeng: add type hints for consistency (this also applies to all the functions in this file)

JoeWaddell: This (and Feed and Pen) actually (unfortunately) causes a circular import: and it's a pretty complex file to write an enum for. Do you have a suggestion for how to handle that?

RUFAS/routines/animal/life_cycle/cow.py
JoeWaddell: This might need to be a permanent change, since milk_production_history can be modified in more than one place under certain circumstances. This is just a way to ensure the latest history value is the most updated. 

I think it might be "temporary" in the sense that the MilkProductionHistory class itself might need a refactor to handle this potential problem.

RUFAS/routines/animal/pen.py
JoeWaddell: Curiously, Feed wasn't imported, and if it is imported it causes a circular import.

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -267,12 +268,24 @@ def update_milk_production_history(self, sim_day):
...
 
-        self.milk_production_history.append(MilkProductionHistory(sim_day, self.days_in_milk,
-                                                                  self.estimated_daily_milk_produced, self.days_born))
+        if self.days_in_milk == 305 and len(self.milk_production_history) > 305:
+            milk_history = [day.milk_production for day in self.milk_production_history[-305:]]
+            self.latest_milk_production_305days = np.sum(milk_history)
+
+        self.milk_production_history.append(
+            MilkProductionHistory(sim_day, self.days_in_milk, self.estimated_daily_milk_produced, self.days_born)
+        )
```
KFosterReed: I think the update to `self.milk_production_history` on Lines 286-288 should come before the calculation of the 305 day milk. Looking at the milk production history list at that point, the last record has a DIM of 304. Also, looking at the `milk_history` obeject, it has a length of 305 but I am thinking the first item is at zero DIM and has a zero value for milk production.

RUFAS/simulation_engine.py
```diff
@@ -34,6 +34,7 @@ def simulate(self) -> None:
         t_start_sim = timer.time()
 
         self._run_simulation_main_loop()
+        routines.animal.animal_reporter.AnimalReporter.report_end_of_simulation(self.state.animal_manager)
```
KFosterReed: Do we know if there are plans to have similar 'end of simulation' reports in other modules?

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -297,69 +297,6 @@ def daily_update(self, sim_day: int,
...
-        calving_to_preg_time_greater_than_3 = self.avg_calving_to_preg_time['greater_than_3']
-        om.add_variable("calving_to_preg_time_1", calving_to_preg_time_1, info_map)
-        om.add_variable("calving_to_preg_time_2", calving_to_preg_time_2, info_map)
-        om.add_variable("calving_to_preg_time_3", calving_to_preg_time_3, info_map)
-        om.add_variable("calving_to_preg_time_greater_than_3", calving_to_preg_time_greater_than_3, info_map)
-        avg_age_for_calving_1 = self.avg_age_for_calving['1']
-        avg_age_for_calving_2 = self.avg_age_for_calving['2']
-        avg_age_for_calving_3 = self.avg_age_for_calving['3']
-        avg_age_for_calving_greater_than_3 = self.avg_age_for_calving['greater_than_3']
-        om.add_variable("avg_age_for_calving_1", avg_age_for_calving_1, info_map)
```
krb78, line 357: Avg_age_for_calving_1 seems like it would be an important output for us to keep. Perhaps it is just streamlined somewhere else through these changes, I will continue to check, but if not, I wanted to flag this here :)

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -267,12 +268,24 @@ def update_milk_production_history(self, sim_day):
         Updates the animal's milk production history by appending a
         MilkProductionHistory object to the list.
 
-        Args:
+        If milk production history has already been updated for the day,
+        the most recent entry is deleted before appending the latest values.
+        Once a cow reaches 305 days in milk, latest_milk_production_305days is updated.
```
krb78, line 261: Not that it's particularly important here (and assuming I understand what we are talking about), but for a time at which we think about how RuFaS might "talk to" other dairy software systems perhaps a name of "currentlact_milk_production_305days" would be helpful and the prior lactation would be called "previous_milk_production_305days" as this is the lingo used by many in the industry.

krb78 (commented):
I want to ensure that we can get av age at first calving in the output somehow and only saw it deleted. maybe it'll be accessible post-processing but just want to double check before this is integrated. otherwise, my comments are minor and all is good :)

KFosterReed (commented):
This is mostly looking good! I left one comment about the Milk305 calculation and I started to look at the new outputs. I still want to check some of the outputs but I did confirm that all the different manure outputs are the same! (when we only have one pen per animal class for example the aggregate lac cow manure excretion from the AnimalManager is the same as the pen with the lactation cows and the AnimaltoManureModule outputs for that pen). So far so good!

I tried to look at some of the life cycle purchases, sales and exits but didn't get anything in the graphs. Probably user error and hope to come back to it soon :)

RUFAS/routines/animal/life_cycle/life_cycle.py
JoeWaddell: It was indeed streamlined elsewhere! Those data are recorded in the method `report_life_cycle_manager_data` in `animal_reporter.py`
The `info_map` in that method will still report this location, so it shouldn't change the key when using output filters.

RUFAS/routines/animal/life_cycle/cow.py
JoeWaddell: This is good to know: I think we could address this in a separate PR since this one is already huge. Would the idea be to add another attribute, so each cow has a "current" and "previous" 305days lactation value?

So as an example, a cow that is 306 days into its first lactation has a current value of whatever it was the day prior, and a previous value of NA. Then on its second lactation the "current" gets demoted to "previous", and "current" is updated once it is at or past 305 days?

RUFAS/simulation_engine.py
JoeWaddell: Good question: I met with Pooya about this specific issue, and he OKed putting this information here for now...but we might need a more elegant solution if it balloons across modules.

One method per module might not be excessive though: we can wrap this any new "end of simulation" methods into one method, like I did in `report_daily_reports`.

RUFAS/routines/animal/animal_reporter.py
ancheypeng: There is a method described [here](https://adamj.eu/tech/2021/05/13/python-type-hints-how-to-fix-circular-imports/) on how to fix circular imports. It's a bit hacky, but I'm not sure if there are any better ways to do it. The type hints aren't too important either, so leaving them out is also not a big issue.

RUFAS/routines/animal/life_cycle/cow.py
krb78: Yes! Your first thought is correct :) And sounds great to make it another smaller PR, no worries. This is minor in the grand scheme of what is needed to make RuFaS actually work, it's just a nice to have.

krb78 (approved):
Approving now that my questions are answered :)

RUFAS/routines/animal/life_cycle/cow.py
ltp23: We can invert the logic and only add when the list is empty or the current day differs from the previous day. 

```
if len(self.milk_production_history) == 0 or self.milk_production_history[-1].simulation_day != sim_day:
        self.milk_production_history.append(
            MilkProductionHistory(sim_day, self.days_in_milk, self.estimated_daily_milk_produced, self.days_born)
        )
```

RUFAS/simulation_engine.py
ltp23: I think the manure module could do something similar for certain tasks that are more time and memory-efficient to perform only once at the end of the simulation. So, if all modules agree on the same name, then the simulation engine can call all at once.

RUFAS/routines/animal/animal_reporter.py
```diff
@@ -0,0 +1,402 @@
...
+        Parameters
+        ----------
+        animal_manager : AnimalManager
+            Instance of Animalmanager class.
+
+        """
+        info_map = {
+            "class": "cow",
+            "function": "update_milk_production_history",
+        }
```
ltp23: I think it would be better to match the function name in the info map with the function it is in, so we know where we added the variable. I see other info maps in the class have similar patterns.

RUFAS/routines/animal/animal_reporter.py
```diff
@@ -0,0 +1,402 @@
...
+
+# TODO remove the below if can't resolve circular imports!
+# from RUFAS.routines.animal.pen import Pen
+# from RUFAS.routines.animal.animal_manager import AnimalManager
+from RUFAS.routines.animal.manure.general_manure import AnimalManureExcretions
+
+om = OutputManager()
+
+
+class AnimalReporter:
```
ltp23: Maybe AnimalModuleReporter instead? Overall, I really like the moving of all or almost all the output data to this module.

ltp23 (approved):

RUFAS/routines/animal/animal_reporter.py
JoeWaddell: I like this name idea! It makes sense to include Module to avoid confusion about whether this is for individual animals.

RUFAS/routines/animal/animal_reporter.py
JoeWaddell: To explain the reasoning for the current implementation: this way it allows the user see where the variable was last updated (note: this assumes that when folks make changes to the variable in the future, they also update it here). It also avoids the problem of how nearly all output methods in the Animal Module would now have the same class name, making that part of the naming scheme a little redundant. 

It's something I discussed with @PooyaHekmati, and it came up in similar discussions with @Isgotkowitz, so I'd be keen to hear what everyone thinks about the proposal below:
Do you think to split the difference we could/should have another field in the metadata that would be defined as the last updated location? So that way the key here would be the AnimalModuleReporter class, the specific function that sent the data to om, and also the name of the last-known update to that variable? Or am I overthinking this?

jadamchick (approved):
Looks great! I ran reports for most of the outputs we need for the pilot farms and compared values between this branch and main (for default settings), and everything reports out well and the values look identical between branches (so no unexpected shifts that I am picking up on).

PR #953, opened 2023-11-14
Culled, sold, and died descriptors

JoeWaddell:
updated "sold", "culled", and "died" labels to be more consistent and descriptive:
culled_heifer_num -> sold_heiferII_num
sold_heifer_num -> sold_heiferIII_num
culled_cow_num -> sold_and_died_cow_num

Created and added died_cow_num to om (is currently only reported within cull_reason_stats).


YijingGong:
I actually like the original names. The "culled" and "sold" actually indicate different things.  "culled" indicates there is a health issue related to those animals. 

culled_heifer_num: this is the number of heiferII that get culled because of reproduction failure (not pregnant for a certain amount of time).
sold_heifer_num: this is the number of heiferIII sold because of the limit of herd size (no health issue or death or other problems). 
culled_cow_num: yes, this is the number of cows culled and died. -> Maybe this one can be named "culled_and_died_cow_num"

Another option below if we want to use "sold" to make them list together:
culled_heifer_num -> sold_heiferII_num
sold_heifer_num -> sold_heiferIII_oversupply_num
culled_cow_num -> sold_and_died_cow_num

jadamchick:
Thanks @YijingGong! I really like your new suggestions, I think they are very clear without having to be super familiar with the inner workings of the model.

KFosterReed:
great discussion everyone! I like the use of the word 'sold' to replace 'culled' for the heifers and distinguishing between whether they are heiferII or heiferIIIs is a great idea. 

For the cow labels - if we are going to keep a variable that is the combined number of cows that were culled and diet, what if we call it: 'cow_herd_exits' or 'cow_herd_exit_num'?

Isgotkowitz (approved):
Names look good, are much clearer and reflective of the data that they hold.

YijingGong:
The names look good to me. I tested by running and checking the output csv. All mentioned variables are there and modified as desired.

YijingGong (approved):

PR #954, opened 2023-11-14
Daily herd milk fat and protein added

JoeWaddell:
In Cow.py:
Created and assigned to Cow:
- daily_milk_fat_kg
- daily_milk_protein_kg

In life_cycle.py:
created:
- daily_herd_milk_fat_kg
- daily_herd_milk_fat_percent
- daily_herd_milk_protein_kg
- daily_herd_milk_protein_percent

Added daily_herd_milk_fat_percent and daily_herd_milk_protein_percent to life_cyle_daily_update_herd_keys.


jadamchick:
This is useful because % fat and % protein are only reported on a per cow per day basis, and calculating a whole-herd average requires weighting by each cow's milk production.  This is a common metric used to track herd production and would be useful to have ready-to-go out of the simulation. They can then be used to easily calculate whole herd Energy Corrected Milk or Fat & Protein Corrected Milk in post.

Ed731551:
Adding to Julie's post, I would like to suggest a minor change. Instead of using kg for naming these variables, I would say it is better to use the word 'yield'. For example, In Cow.py (same applies for life_cycle.py), instead of 'daily_milk_fat_kg', I would prefer to use: 'daily_milk_fat_yield'. That terminology is of common use in Animal Science and directly reflects the purpose of these variables (yield of milk solids components: fat, protein, lactose). Either kg or kg/d are units for these variables.

Similar observation for %. Instead of '%' may be 'Comp.; for referring to composition. Just suggestions. Thanks.

tomhuhh:
Adding to Edward's comments on the naming conventions, I tend to agree with using `'yield' `and `'comp'` instead of `'kg' `and `'percent'` in the variable names. Thanks!

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -338,9 +341,13 @@ def milking_update(self, sim_day, calving_interval):
             daily_fat_correct_milk_production = \
                 0.4 * estimated_daily_milk_produced + \
                 0.15 * fat_percent * estimated_daily_milk_produced
+            self.daily_milk_fat_kg = fat_percent*estimated_daily_milk_produced
+            self.daily_milk_protein_kg = self.mPrt * self.estimated_daily_milk_produced
         else:
             fat_percent = 0
             daily_fat_correct_milk_production = 0
+            self.daily_milk_fat_kg = 0
+            self.daily_milk_protein_kg = 0
```
tomhuhh, line 376: I would suggest that we add here that the milk protein and lactose percentages should also be set back to zero if the cows are not lactating.

tomhuhh: Maybe this has been addressed in the life_cycle.py but I would address it here as well for consistency.

tomhuhh (commented):
Looks good to me with minor change suggestions.

jadamchick:
Hi all - either naming convention is fine with me. I personally think "kg" and "%" is more clear (I don't know that I would know what "comp" was referring to without someone telling me at least once that it is short for composition -- I would jump to "components", and think it was redundant to say "milk_fat_component"). But I trust you all on what you think is best. @tomhuhh @Ed731551 @JoeWaddell

RUFAS/routines/animal/life_cycle/cow.py
JoeWaddell: Current behavior is that `mPrt` and `lactose_milk` are never modified: they are hardcoded in `__init__`. I've opened an Issue to address this: https://github.com/RuminantFarmSystems/MASM/issues/962

JoeWaddell:
I favor having the units in the variable names themselves when possible, since it will A) help SWEs when adding methods, and B) help SMEs when evaluating raw outputs.

One option: what about unit suffixes in the code (for clarity), and longer, proper Animal Science names in output manager? e.g. the call to om would be something like:
`om.add_variable("Herd_milk_fat_yield_(kg/day)", daily_herd_milk_fat_kg, info_map)`


I think this is something both @KFosterReed and @PooyaHekmati would be able to comment on, since it deals with naming conventions we need to keep in mind moving forward. In the short term, whatever we decide here will apply to the same questions in #955

PooyaHekmati:
> I favor having the units in the variable names themselves when possible, since it will A) help SWEs when adding methods, and B) help SMEs when evaluating raw outputs.
> 
> One option: what about unit suffixes in the code (for clarity), and longer, proper Animal Science names in output manager? e.g. the call to om would be something like: `om.add_variable("Herd_milk_fat_yield_(kg/day)", daily_herd_milk_fat_kg, info_map)`
> 
> I think this is something both @KFosterReed and @PooyaHekmati would be able to comment on, since it deals with naming conventions we need to keep in mind moving forward. In the short term, whatever we decide here will apply to the same questions in #955

We definitely want to be as future-proof as possible, but that shouldn't come at the cost at the expense of blocking the progress today. If having the units as suffix in the key is highly desired, I have nothing against it. Just don't do it manually. Instead, you can do: `info_map["suffix"]="(kg/day)"` an OM takes care of the rest.

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -338,9 +341,13 @@ def milking_update(self, sim_day, calving_interval):
             daily_fat_correct_milk_production = \
```
ew3361zh, line 366: Do you think we could get some references on these formulas used for fat_percent and daily_fat_correct_milk_production? Seems like a lot of magic numbers (okay to make an issue about it if this isn't the PR to accomplish this in).

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -113,6 +113,8 @@ def __init__(self, args):
         self.milking = False
         self.days_in_milk = 0
         self.estimated_daily_milk_produced = 0
+        self.daily_milk_fat_kg = 0
+        self.daily_milk_protein_kg = 0
```
ew3361zh: Can we add the definitions of these somewhere?

JoeWaddell: Good call: I'll add the definitions in comments below for now, but have opened an Issue to handle this in a more uniform manner across the animal types (and also tackling the other comment you outlined). https://github.com/RuminantFarmSystems/MASM/issues/970

The Issue description I just wrote likely leaves something to be desired, so any SME/SWEs please feel free to chime in!

RUFAS/routines/animal/life_cycle/cow.py
JoeWaddell: Opened an Issue to resolve this: https://github.com/RuminantFarmSystems/MASM/issues/971

Also easy fodder for the Fixathon, if we don't get to it before then.

ew3361zh (approved):
LGTM. I would update the main comment/description to reflect the changes to the names (e.g. `daily_milk_fat_kg` -> `milk_fat_kg`) and I think the docstrings you added (and the ones we'll add later in the fixathon) will be great to address any questions as to what the variable represents.

RUFAS/routines/animal/life_cycle/cow.py
tomhuhh: Thanks for creating an issue Joe! However, that creates another potential issue because based on my past experience working with the data, even if the animal is not lactating, the code still assign them with milk protein and lactose values. By controlling those values here, we can potentially resolve that issue. What are your thoughts on this?

tomhuhh: Also, just to comment on the milk protein and lactose values, I believe from what I learnt from Kristan they are quite stable throughout the lactation. Thus, it is okay to assume they do not change. However, for sure, we can implement more dynamic equations to predict these values in the future.

tomhuhh (commented):
Thanks Joe! I added a few more comments regarding milk protein and lactose. Otherwise, I think this PR is good to go.

RUFAS/routines/animal/life_cycle/cow.py
JoeWaddell: So I added some logic and constants: please let me know if the descriptions are correct from an animal science perspective. Otherwise, I think it does what you've requested: adds logic to the milking update method, but should work here? The whole herd values now vary a little, since it's counting lactating and non-lactating animals (this should also be checked that it's expected behavior)

tomhuhh: Thanks Joe! I can confirm that for non-lactating animals, milk fat, protein and lactose are 0. So this is great! While checking this, I noticed that for over 360 lactating animals, there milk fat values are 0. What are your thoughts on this? 

To replicate this issue, you can run `python3 -c "from main import run_rufas; run_rufas(exclude_info_maps=True, clear_output=True)"` with the current branch and use the filter `Cow.milking_update`. 

Screenshot 2023-11-22 at 11 29 51 AM

Thanks!

RUFAS/routines/animal/animal_module_constants.py
```diff
@@ -69,6 +69,15 @@ class AnimalModuleConstants:
...
+    MILK_CRUDE_PROTEIN = 3.2
+    """Milk crude protein content, percentage."""
+
+    MILK_LACTOSE = 4.85
+    """Milk lactose content, percentage."""
+
+    MILK_TRUE_PROTEIN = 3.5
+    """Milk true protein content, percentage."""
+
     DMI_CONSTRAINT_PERCENT = 0.20
```
tomhuhh, line 81: These descriptions look good to me! Thank you Joe!

RUFAS/routines/animal/life_cycle/cow.py
JoeWaddell: I think I've figured out the problem: 1) I updated the code so we update self.fat_percent here, since it was being assigned elsewhere (in `update`).  2) Some individuals will still have a fat_percent equal to 0.0, since that's what the equation gives if days_in_milk = 1. We should consider whether this equation needs to be modified.

tomhuhh: Thanks Joe! That is a huge progress! The outputs reflect your description. Also, milk fat content should not be 0 on the first day. Typically, milk (colostrum) fat content from day 1 should be higher than milk fat content later on (https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6940821/#:~:text=Colostrum%20contains%20bioactive%20components%20with,vitamins%20and%20polyunsaturated%20fatty%20acids.). Thus, I agree that we should consider a fix.

tomhuhh:
After checking with @KFosterReed, we have decided to use milk fat estimated on DIM = 2 to represent that value on DIM =1 due to the limitation of the prediction equation as a temporary fix. The plot of the milk fat as a function of DIM also proves this point. @KFosterReed or I (if I can find it in the time being) will upload the documentation for this equation to Basecamp Animal module/life cycle folder. Thanks! 

image

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -274,6 +280,24 @@ def update_milk_production_history(self, sim_day):
...
+        Note that this equation produces 0.0 if days_in_milk is set to one,
+        so we've implemented a minimum days_in_milk value of 2.
+
+        Parameters
+        ----------
+        days_in_milk : int
+            Number of days in milk.
+        """
+        if days_in_milk == 1:
+            days_in_milk = 2
```
tomhuhh, line 296: This fix agrees with our discussion.

tomhuhh (approved):
LGTM! Thanks Joe!

PR #955, opened 2023-11-14
update "methane" to "enteric_methane" with units

JoeWaddell:
Updated "methane" keys related to enteric methane from "methane" to "enteric_methane_g" or "enteric_methane_kg" (pen_manure.py only).
Since they are all within manure sections of the code and output, it is helpful to clarify that these values are methane from enteric fermentation, not manure handling and storage.


Ed731551:
Agree. I think it is important to make that distinction.

RUFAS/routines/animal/manure/calf_manure_excretion.py
```diff
@@ -113,7 +113,7 @@ def manure_calculations(ration_formulation,
         phosphorus=manure_phosphorus_excreted,
         phosphorus_fraction=manure_phosphorus_fraction,
         potassium=0,
-        methane=methane_emission
+        enteric_methane_g=methane_emission
```
tomhuhh, line 116: I am not sure if we want to go with this naming format with the variables. If so, that means all the other variable names need similar modification. Could we add the unit to the output variable name instead?

RUFAS/routines/animal/manure/dry_cow_manure_excretion.py
```diff
@@ -193,7 +193,7 @@ def manure_calculations(ration_formulation,
         phosphorus=manure_phosphorus_excreted,
         phosphorus_fraction=manure_phosphorus_fraction,
         potassium=potassium,
-        methane=methane_emission
+        enteric_methane_g=methane_emission
```
tomhuhh, line 196: Same thoughts as above.

RUFAS/routines/animal/manure/dry_cow_manure_excretion.py
```diff
@@ -61,7 +61,7 @@ def manure_calculations(ration_formulation,
     nutrient_amounts, nutrient_concentrations = RationReporter.report_ration(
         ration_formulation, feed.available_feeds)
     dry_matter_intake = nutrient_amounts['dm']
-    ash_diet_content = nutrient_amounts['ash']
+    # ash_diet_content = nutrient_amounts['ash']
     CP_concentration = nutrient_concentrations['CP']
     potassium_concentration = nutrient_concentrations['potassium']
     ASH_concentration = nutrient_concentrations["ash"]  # TODO: Unused
```
tomhuhh: I think we can remove the `# TODO: Unused` because it is used.

RUFAS/routines/animal/manure/dry_cow_manure_excretion.py
```diff
@@ -61,7 +61,7 @@ def manure_calculations(ration_formulation,
     nutrient_amounts, nutrient_concentrations = RationReporter.report_ration(
         ration_formulation, feed.available_feeds)
     dry_matter_intake = nutrient_amounts['dm']
-    ash_diet_content = nutrient_amounts['ash']
+    # ash_diet_content = nutrient_amounts['ash']
```
tomhuhh, line 64: Agree that we can comment out this line of code.

RUFAS/routines/animal/manure/growing_heifer_manure_excretion.py
```diff
@@ -95,7 +95,7 @@ def manure_calculations(ration_formulation,
                       ) * GeneralConstants.GRAMS_TO_KG
 
     # Nitrogen excretion in feces, kg [A.3B.B.3]
-    fecal_nitrogen = manure_nitrogen - urine_nitrogen  # TODO: Unused
+    # fecal_nitrogen = manure_nitrogen - urine_nitrogen  # TODO: Unused
```
tomhuhh: Agree that this should be commented out because it is not output. However, it is good to know that we can calculate this value when we need to.

RUFAS/routines/animal/manure/lactating_cow_manure_excretion.py
```diff
@@ -124,7 +128,7 @@ def manure_calculations(ration_formulation,
     dry_matter_concentration = nutrient_concentrations['dm']
     ADF_concentration = nutrient_concentrations['ADF']
     CP_concentration = nutrient_concentrations['CP']
-    lignin_concentration = nutrient_concentrations['lignin']
+    # lignin_concentration = nutrient_concentrations['lignin']
```
tomhuhh, line 131: Agree with the change.

tomhuhh (commented):
Thanks for catching this Joe! I agree with the naming change from `'methane'` to `'enteric methane'.` However, I am uncertain if we want to have `'enteric methane g'` and` 'enteric methane kg'` separately. If we decide to report enteric methane in kg, I tend to think that we should make that conversion of units inside the `_manure_excretion.py` files.

jadamchick:
Hi everyone! I'm ambivalent about where it happens and trust your judgement, as long as somewhere it will be be indicated what the units are for the methane outputs, especially since right now they are being reported in kg in some places and in g in other places. I would 100% support if we want to add units to all key names, not just methane :) but my agenda right now was just to make sure we don't report enteric methane emissions to stakeholder farms that are off by a factor of 1000 or 1000000 bc of unit carelessness when handling the outputs (I'm a primary culprit for that type of thing).  @JoeWaddell  @tomhuhh

Ed731551:
In terms of herd, units should be in kg/d. On an individual animal basis g/d. Some people interested on estimating carbon footprint of the dairy farm might appreciate an annual report of the herd in metric tons per year. For consideration.

RUFAS/routines/animal/manure/calf_manure_excretion.py
JoeWaddell: Good question: it looks like the way `ManureManagerOutputHelper` is set up would preclude easily adding suffixes for all the outputs in this dataclass. However @ltp23 can likely clarify. 

I think I prefer the practice of adding SI unit abbreviations to variable names - when possible, since sometimes names are long enough (or are self-evident).

The drawback - as you mentioned - is that it is  be ust one of *many* names that would need to be updated, if want a unified system. I think making that change over time makes sense, though, and can be added to tech debt and Fixathon 2024 tasks!

Isgotkowitz:
> In terms of herd, units should be in kg/d. On an individual animal basis g/d. Some people interested on estimating carbon footprint of the dairy farm might appreciate an annual report of the herd in metric tons per year. For consideration.

@Ed731551 this is a great point to raise. I think specifying the time span of the data is really important too, as we make output reporting more complex and possibly give users options for the time span they want their data collected over.

RUFAS/routines/animal/manure/calf_manure_excretion.py
Isgotkowitz: > I think making that change over time makes sense, though, and can be added to tech debt and Fixathon 2024 tasks!

I agree, trying to make all those changes at once might really bog everyone down if it is attempted outside of something like the Fixathon. But before those changes begin to be implemented, can we develop a team-wide standard that everyone should follow when adding units to variable names and outputs?

RUFAS/routines/animal/manure/lactating_cow_manure_excretion.py
```diff
@@ -34,25 +35,28 @@ def methane_mitigation(NDF_concentration: float,
...
-        Methane mitigation method used to reduce enteric methane emissions, including "3-NOP", "Monensin", "Essential Oils", and "Seaweed". 
-    methane_mitigation_additive_amount: float 
-        The amount of methane mitigation feed additive that is added, mg/kg dry matter intake (DMI). The recommended dose for 3-NOP is between 40 and 100 mg/kg DMI, while that for monensin is between 16 and 36 mg/kg DMI. 
+    methane_mitigation_method: str
+        Methane mitigation method used to reduce enteric methane emissions, including "3-NOP", "Monensin",
+        "EssentialOils", and "Seaweed".
+    methane_mitigation_additive_amount: float
+        The amount of methane mitigation feed additive that is added, mg/kg dry matter intake (DMI).
+        The recommended dose for 3-NOP is
+        between 40 and 100 mg/kg DMI, while that for monensin is between 16 and 36 mg/kg DMI.
```
Isgotkowitz, line 44: I don't think these docstrings need to be as specific as they are. If possible, they should reference other parts of the code or documentation written that list all methane mitigation strategies/have a complete list of recommended ranges for different strategies.

Isgotkowitz (approved):
Looks good

KFosterReed (approved):
These name changes make sense to me. I am going to add an issue about the fecal N. I don't think it is currently used in the manure module or directed to the output reports. It is a useful value for outputting though not currently critical so I will make a low priority issue.

Issue #951, opened 2023-11-14
Move directory location settings in metadata to gnu args


ew3361zh:
Currently several settings for directory locations (csv_dir, output_dir, graphic_dir) are in the metadata config_properties section. These should be set by gnu args and removed from all the metadata files.

ew3361zh:
Addressed with PR #958

Issue #939, opened 2023-11-11
remove user input from main.py in run_load_vars_pool


PooyaHekmati:
`run_load_vars_pool` in main.py receives a user input. Convert that input to GNU arg.

ew3361zh:
Addressed with PR #958

PR #958, opened 2023-11-14
Move dir path args from metadata to gnu args

ew3361zh:
This PR moves all dir paths set in config input files and from user prompts to gnu args. This addresses issue #951 and #939.

Additional related cleanup present:
1. Ensure all metadata files share the same properties and the same validation guidelines. 
2. Ensure all input files shares the same fields (not necessarily the same values for those fields) required by the metadata.
3. Updates unit testing.


PooyaHekmati (approved):
Lots of improvements happening here!
Side note: I know it is not directly related to this PR, but I think `load_pool` and `vars_file_path` are not clear enough about what their intended uses are. I wish we could name them better.

allisterakun (approved):
Looks good!

PR #960, opened 2023-11-14
fix day length calculation

### LINKED ISSUE #881: Daylength values are hardcoded
The lengths of days are calculated based on the average daylengths of each month in Madison, WI. They should be specific to the location being simulated by RuFaS.

## Expected Behavior

Daylengths calculated should be specific to the location being simulated in RuFaS.

## Current Behavior

Daylengths are calculated based on the average daylength of each month in Madison, WI. See the method `determine_daylength` in `current_day_conditions.py` for details.

## Possible Solution

- Pull the latitude from one of the field inputs, and use it to calculate the daylength for any given day (most appropriate solution for v1).
- Make the daylength of each day an input (more appropriate for a future version of RuFaS).

## Possible Implementation

Use equation 1:1.1.6 from the [SWAT Theoretical Documentation](https://swat.tamu.edu/media/99192/swat2009-theory.pdf)

matthew7838:
### Summary
This PR will address issue #881 regarding how day length is calculated.

### Changes

- Calculations are now based on SWAT 1:1.1.6
- Added a helper method to calculate solar declination radians based on 1:1.1.2
- Fixed how the `determine_daylength` is called in `weather.py`

### Note
The current call of `determine_daylength` in `weather.py` has the latitude **hard-coded** to Madison, WI. This will only be a **temporary value** until the method to retrieve latitude is resolved.


RUFAS/current_day_conditions.py
```diff
@@ -57,23 +58,65 @@ def __post_init__(self):
             self.rainfall = self.precipitation
 
     @staticmethod
-    def determine_daylength(month: int) -> int:
+    def determine_daylength(day_number: int, geographic_latitude_radians: float, month: int,
+                            angular_velocity=0.2618) -> float:
```
Isgotkowitz: The API for this method should be changed. It is redundant and gives callers the option to change things that shouldn't be changed.

- Latitude should be passed as degrees, not radians, and converted with either a subroutine or inside the method body.
- Taking in the month and the day is redundant. Just the day number should be taken in, and used to either determine the month or used directly to determine whether there is no sunset or sunrise.
- Angular velocity should be hardcoded in the method body.

RUFAS/current_day_conditions.py
```diff
@@ -57,23 +58,65 @@ def __post_init__(self):
...
+                else:
+                    return 0
+            elif month == 12 or month <= 3:
+                if geographic_latitude_radians > 0:
+                    return 0
+                else:
+                    return 24
+        else:
+            day_length = ((2 * math.acos(-math.tan(solar_declination_radians) * math.tan(geographic_latitude_radians)))
+                          / angular_velocity)
```
Isgotkowitz:
```suggestion
        solar_declination_radians = CurrentDayConditions.calculate_solar_declination_radians(day_number)
        tangent_product = -math.tan(solar_declination_radians) * math.tan(geographic_latitude_radians)
        is_polar_solstice = math.abs(tangent_product) > 1
        if is_polar_solstice:
            is_summer = 6 <= month <= 9
            in_northern_hemisphere = geographic_latitude_radians > 0
            if is_summer and in_northern_hemisphere:
                return 24
            elif not is_summer and in_northern_hemisphere:
                return 0
            elif is_summer and not in_northern_hemisphere:
                return 0
            else:
                return 24
        else:
            day_length = (2 * math.acos(tangent_product) / angular_velocity
```

RUFAS/current_day_conditions.py
```diff
@@ -57,23 +58,65 @@ def __post_init__(self):
...
+
+        Returns
+        -------
+        float
+            Solar declination in radians
+
+        References
+        ----------
+        SWAT 1:1.1.2
+        """
```
Isgotkowitz:
```suggestion
        Helper method to determine the solar declination in radians.

        Parameters
        ----------
        day_number : int
            Day number of the year.

        Returns
        -------
        float
            Solar declination (radians).

        References
        ----------
        SWAT 1:1.1.2

        """
```

RUFAS/current_day_conditions.py
```diff
@@ -57,23 +58,65 @@ def __post_init__(self):
...
-        month: int
-            Month of the year.
+        day_number : int
+            Day number of the year
+        geographic_latitude_radians : float
+            geographic latitude in radians
+        angular_velocity: float default = 0.2618
+            angular velocity of earth
+        month : int
+            month of the simulation
```
Isgotkowitz:
```suggestion
        day_number : int
            Day number of the year.
        geographic_latitude_radians : float
            Geographic latitude in radians.
        angular_velocity: float default = 0.2618
            Angular velocity of earth (radians / hour).
        month : int
            Month of the simulation.
```

RUFAS/current_day_conditions.py
```diff
@@ -57,23 +58,65 @@ def __post_init__(self):
             self.rainfall = self.precipitation
 
     @staticmethod
-    def determine_daylength(month: int) -> int:
+    def determine_daylength(day_number: int, geographic_latitude_radians: float, month: int,
+                            angular_velocity=0.2618) -> float:
         """
-        Approximate day length of the month by using data from Madison, WI.
+        Calculates the day length for the field based on its day and latitude
```
Isgotkowitz:
```suggestion
        Calculates the day length for the field based on its day and latitude.
```

tests/test_current_day_conditions.py
```diff
@@ -18,12 +20,50 @@ def test_current_weather_snowfall(snow_fall: int, rainfall: int, actual: Current
...
 
 
-@pytest.mark.parametrize("months, expected", [
-    ([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12], [9, 10, 11, 13, 14, 15, 15, 15, 13, 12, 10, 9])
+@pytest.mark.parametrize("day_number, geographic_latitude_radians, month, polar_location, northern_hemisphere", [
+    (15, 0.752, 1, False, True),  # Madison example, no polar day or night possible
+    (365, 1.222, 12, True, True),  # Winter in polar circle
+    (180, 1.222, 8, True, True),  # Summer in polar circle
+    (365, -1.222, 12, True, False),  # Winter in polar circle (south hemisphere)
+    (180, -1.222, 8, True, False),  # Sumer in polar circle (south hemisphere)
```
Isgotkowitz: Expected daylength should be defined in the test cases. They will have to be pretty heavily approximated because the SWAT equation is not particularly precise. You can find the actual daylengths for different times and places [here](https://sunrise-sunset.org/us/madison-wi/2023/11)

RUFAS/current_day_conditions.py
```diff
@@ -57,23 +58,65 @@ def __post_init__(self):
...
+            angular velocity of earth
+        month : int
+            month of the simulation
 
         Returns
         -------
-        int
-            Day length of the month (hour).
+        float
+            The day length of  the field on specific day
```
Isgotkowitz:
```suggestion
            The length of the current day (hours).
```
This day length is not specific to any field.

Isgotkowitz (changes requested):
Good start Matthew. Changes requested on the method signature and test cases of `determine_daylength()`

RUFAS/current_day_conditions.py
matthew7838: Thanks. I forgot about the existence of that day-mont conversion method.

RUFAS/current_day_conditions.py
```diff
@@ -57,23 +60,62 @@ def __post_init__(self):
...
+        Returns
+        -------
+        float
+            The length of the current day (hours).
+
+        References
+        ----------
+        SWAT 1:1.1.6
+        """
+        angular_velocity = 0.2618
```
ew3361zh: Is this a constant?

matthew7838: yes as long as the farm's on earth.

ew3361zh: Haha...then should it go in and come from the C&S (earth) constants file?

matthew7838: Good point and I agree. I'm not quite aware of the existence of the file, could you point me to it? Thanks.

RUFAS/current_day_conditions.py
```diff
@@ -57,23 +60,65 @@ def __post_init__(self):
...
+        Calculates the day length for the field based on its day and latitude.
+
+        Parameters
+        ----------
+        day_number : int
+            Day number of the year.
+        geographic_latitude : float
+            Geographic latitude (degrees).
+        year : int
+            Year of the current simulation.
```
Isgotkowitz:
```suggestion
            Calendar year of the current simulation.
```

tests/test_current_day_conditions.py
```diff
@@ -18,12 +21,54 @@ def test_current_weather_snowfall(snow_fall: int, rainfall: int, actual: Current
...
+                elif is_winter:
+                    if northern_hemisphere:
+                        assert CurrentDayConditions.determine_daylength(
+                            day_number, geographic_latitude, year) == expected
+                    else:
+                        assert CurrentDayConditions.determine_daylength(
+                            day_number, geographic_latitude, year) == expected
+            else:
+                assert expected == pytest.approx(CurrentDayConditions.determine_daylength(
+                    day_number, geographic_latitude, year), 0.1)
```
Isgotkowitz: The suggested code should work for all non-polar tests and all summer/winter polar tests, but not polar tests outside of the extremes. Being able to handle any arbitrary, valid test case is a nice to have so it is up to you if you want to make that change.
```suggestion
            observed = CurrentDayConditions.determine_daylength(day_number, geographic_latitude, year)
            tolerance = 0.1 if not polar_location else 0.0
            assert expected == pytest.approx(CurrentDayConditions.determine_daylength(
                    day_number, geographic_latitude, year), tolerance)
```

tests/test_current_day_conditions.py
```diff
@@ -18,12 +21,54 @@ def test_current_weather_snowfall(snow_fall: int, rainfall: int, actual: Current
...
+                             northern_hemisphere: bool, expected: float, year: int) -> None:
     """Tests that correct day length were returned by the corresponding month"""
-    day_length = []
-    for month in months:
-        day_length.append(CurrentDayConditions.determine_daylength(month))
-    assert day_length == expected
+    with patch('RUFAS.current_day_conditions.CurrentDayConditions.calculate_solar_declination_radians',
+               wraps=CurrentDayConditions.calculate_solar_declination_radians) as mocked_radian_calculation:
+        with patch('RUFAS.util.Utility.day_to_month_conversion',
+                   wraps=Utility.day_to_month_conversion) as mocked_month_conversion:
```
Isgotkowitz, line 40: What is the `wraps` doing? I haven't seen it before.

RUFAS/weather.py
```diff
@@ -172,7 +172,8 @@ def get_current_day_conditions(self, time: Time) -> CurrentDayConditions:
         year = time.year
         day = time.day
         month = Utility.day_to_month_conversion(day, time.calendar_year)
-        daylength = CurrentDayConditions.determine_daylength(month)
+        # 0.752 as radian is hard-coded to Madison,WI
+        daylength = CurrentDayConditions.determine_daylength(day, 0.752, month)
```
Isgotkowitz: This be changed to 43.0723

RUFAS/current_day_conditions.py
Isgotkowitz: It should go in the `general_constants.py` file, located in the `RUFAS` directory. Units and origin should be noted in it as well

Isgotkowitz (approved):
Looks good Matthew. Approving, but there are some suggested changes. Only thing that actually needs to be changed is the hardcoded value for Madison's latitude.

RUFAS/current_day_conditions.py
```diff
@@ -57,23 +60,65 @@ def __post_init__(self):
...
     @staticmethod
-    def determine_daylength(month: int) -> int:
+    def determine_daylength(day_number: int, geographic_latitude: float, year: int) -> float:
+        """
+        Calculates the day length for the field based on its day and latitude.
+
+        Parameters
+        ----------
+        day_number : int
+            Day number of the year.
```
ew3361zh: Looking at other feedback - should we specify whether this is day number of the calendar year (and not of the simulation year)? Or even if a simulation starts on a day other than Jan 1, is the day number always the calendar day number?

RUFAS/current_day_conditions.py
```diff
@@ -57,23 +60,65 @@ def __post_init__(self):
...
         """
-        Approximate day length of the month by using data from Madison, WI.
+        Helper method to determine the solar declination in radians.
 
         Parameters
         ----------
-        month: int
-            Month of the year.
+        day_number : int
+            Day number of the year.
```
ew3361zh: Same as above - is it self-evident that day number of the year refers to calendar year? Or should we specify that in case it gets confused with day of the simulation year?

tests/test_current_day_conditions.py
```diff
@@ -18,12 +21,54 @@ def test_current_weather_snowfall(snow_fall: int, rainfall: int, actual: Current
...
+                    else:
+                        assert CurrentDayConditions.determine_daylength(
+                            day_number, geographic_latitude, year) == expected
+                elif is_winter:
+                    if northern_hemisphere:
+                        assert CurrentDayConditions.determine_daylength(
+                            day_number, geographic_latitude, year) == expected
+                    else:
+                        assert CurrentDayConditions.determine_daylength(
+                            day_number, geographic_latitude, year) == expected
```
ew3361zh: If you decide to keep this block (i.e. not make the change Ed suggested below), I'd switch these two around to avoid the if not and then kept it to just if/else. Seems simpler unless there's another condition we're leaving this open for (not being facetious). 
```suggestion
                if is_winter:
                    if northern_hemisphere:
                        assert CurrentDayConditions.determine_daylength(
                            day_number, geographic_latitude, year) == expected
                    else:
                        assert CurrentDayConditions.determine_daylength(
                            day_number, geographic_latitude, year) == expected
                else:
                    if northern_hemisphere:
                        assert CurrentDayConditions.determine_daylength(
                            day_number, geographic_latitude, year) == expected
                    else:
                        assert CurrentDayConditions.determine_daylength(
                            day_number, geographic_latitude, year) == expected
```

tests/test_current_day_conditions.py
ew3361zh: Should it be:
```suggestion
            observed = CurrentDayConditions.determine_daylength(day_number, geographic_latitude, year)
            tolerance = 0.1 if not polar_location else 0.0
            assert expected == pytest.approx(observed, tolerance)
```

ew3361zh (approved):
Very nice work, just a few comments, questions, suggestions but otherwise looks close to ready!

tests/test_current_day_conditions.py
matthew7838: It basically mocks the function without affecting its return value since I want to simply check these functions were called.

RUFAS/weather.py
matthew7838: Good catch

RUFAS/current_day_conditions.py
Isgotkowitz: I agree with being more explicit in the documentation - this day number should be documented as the "ordinal day", which is the fancy term for cumulative day number of the year.

> day number of the calendar year (and not of the simulation year)

The `Time` module only has one day counter and it is the same for both the calendar year and the simulated year. Right now, the simulation year and calendar year change at the same time.

tests/test_current_day_conditions.py
matthew7838: Thanks, Ed your suggestions are way cleaner. @ew3361zh I think you're right, that's how I'll adapt to this.

RUFAS/current_day_conditions.py
matthew7838: Sure, that should better be specified.

Issue #944, opened 2023-11-13
Enhance Soil Carbon Initialization routines


Isgotkowitz:
**Is your feature request related to a problem? Please describe.**
The current routine for initializing soil carbon levels is overly simplistic. After discussing with SMEs, more detailed initialization routines have been developed.

Current behavior:
After the total amount of carbon in the soil layer is calculated, it is split equally between the three carbon pools in each layer:
1/3 goes to active carbon, 1/3 goes to slow carbon, and 1/3 goes to passive carbon.
Metabolic and structural litter pools are not initialized.

**Describe the solution you'd like**
For the surface soil layer:
- the total amount of soil carbon is partitioned between the active and slow carbon pools with a split of 2% and 98% respectively.
- the initial residue in the simulation is split equally between the surface metabolic and structural litter pools.
  - Note that this even split is a big assumption, but a more realistic assumption would require knowing field management activity that occurred prior to the simulation starting, so this is assumption is at least a good place to start.

For subsurface soil layers:
- the total amount of soil carbon is partitioned between the active, slow, and passive carbon pools with a split of 2%, 54%, and 44% respectively.

**Describe alternatives you've considered**

**Additional context**
Numbers for the splits between active, slow, and passive carbon come from Xuesong, a carbon cycling expert. They are not empirical, because the different carbon pools are just abstractions made up for the purposes of modeling carbon dynamics well. The numbers listed in this issue description are the same as or similar to ones used in more comprehensive carbon cycling models, including SWAT-C, DSSAT, DayCent, etc.

ew3361zh:
Closed wwith PR #961

PR #961, opened 2023-11-15
Updates to soil carbon init method and testing

ew3361zh:
PR to address issue #944.

1. Updates calculations in `_initialize_carbon_pools()`
2. Adds residue to `_initialize_carbon_pools()` args
3. Updates testing.

Note:
- [x] Not sure if a different approach to testing whether residue is being accurately split between the structural and metabolic litter pools is necessary. Would need to think about how to mock the post_init with specific residue values. If you have any suggestions/thoughts, that would be helpful.
- [x] Also could use some help in how to update the Notes docstring section of the `_initialize_carbon_pools()` method. Seems like from the description in #944, this enhancement still isn't entirely precise but is at least a step in the right direction.


RUFAS/routines/field/soil/layer_data.py
```diff
@@ -424,12 +426,17 @@ def _initialize_carbon_pools(self, field_size: float) -> None:
...
-        self.slow_carbon_amount = one_third_soil_carbon
+        if self.top_depth == 0:
+            self.active_carbon_amount = 0.02 * total_soil_carbon_amount
+            self.slow_carbon_amount = 0.98 * total_soil_carbon_amount
+            self.structural_litter_amount = (1 / 2) * residue
+            self.metabolic_litter_amount = (1 / 2) * residue
+        else:
+            self.active_carbon_amount = 0.02 * total_soil_carbon_amount
+            self.slow_carbon_amount = 0.54 * total_soil_carbon_amount
+            self.passive_carbon_amount = 0.44 * total_soil_carbon_amount
```
Isgotkowitz:
```suggestion
            self.passive_carbon_amount = 0.44 * total_soil_carbon_amount
            self.structural_litter_amount = 0.0
            self.metabolic_litter_amount = 0.0
```
We want to really be explicit in making sure there is no belowground litter (for now) because some layers get post_init called on them once when they get initialized, then get modified, then get post_init called on them again.

tests/soil_crop_tests/soil_tests/test_layer_data.py
```diff
@@ -89,18 +89,28 @@ def test_post_init(top: float, bottom: float, concentration: float) -> None:
...
+        (1.4, 120, 4158.0, 91476.0, 112266.0, 0.0, 0.0),
+        (3.556, 120, 4158.0, 91476.0, 112266.0, 0.0, 0.0),
+        (0.88, 120, 4158.0, 91476.0, 112266.0, 0.0, 0.0),
+    ])
+def test_initialize_carbon_pools(field_size: float, top_depth: int, expected_active: float,
+                                 expected_passive: float, expected_slow: float,
+                                 expected_structural_litter: float, expected_metabolic_litter: float) -> None:
     """Tests that carbon pools in a soil layer are properly initialized."""
-    actual = LayerData(field_size=field_size, top_depth=120.0, bottom_depth=750.0, bulk_density=1.5,
+    actual = LayerData(field_size=field_size, top_depth=top_depth, bottom_depth=750.0, bulk_density=1.5,
```
Isgotkowitz:
```suggestion
    actual = LayerData(field_size=field_size, residue=residue, top_depth=top_depth, bottom_depth=750.0, bulk_density=1.5,
```
If residue is passed into the constructor it will be passed into `_intialize_carbon_pools()` and the results can be observed by looking at the structural and metabolic litter amounts.

RUFAS/routines/field/soil/layer_data.py
```diff
@@ -405,14 +405,16 @@ def _initialize_nitrogen_pools(self, field_size: float, residue: float) -> None:
...
         Parameters
         ----------
-        field_size: float
+        field_size : float
             Size of the field (ha).
+        residue : float
+            Amount of residue on the soil surface when this soil layer is initialized (kg / ha).
 
         Notes
         -----
```
Isgotkowitz, line 420: The notes section should contain a brief message stating that the splits for the initialization of carbon pools are not empirical but generally are the same as values used by other models, and the 50/50 split between litter pools is a heavy abstraction but a more accurate split cannot be predicted without knowing management practices prior to initialization.

Isgotkowitz (changes requested):
Excellent start Niko. Changes requested for checking that the litter pools are init'ed correctly in the test, and adding a comment noting where the numbers we used came from. Then I think it will be ready to go

tests/soil_crop_tests/soil_tests/test_layer_data.py
ew3361zh: What are some good typical residue #s to use (or does it matter)?

ew3361zh: Added 5, 10, and 19.5

Isgotkowitz (approved):
LG8TM nice job Niko

tests/soil_crop_tests/soil_tests/test_layer_data.py
Isgotkowitz: Any numbers work.

RUFAS/routines/field/soil/layer_data.py
```diff
@@ -405,31 +405,41 @@ def _initialize_nitrogen_pools(self, field_size: float, residue: float) -> None:
...
+        The splits for the initialization of carbon pools are not empirical but generally are
+        the same as values used by other models. The 50/50 split between litter pools is
+        a heavy abstraction but a more accurate split cannot be predicted without knowing management
+        practices prior to initialization.
 
         """
         soil_volume_in_cubic_meters = self.layer_thickness * (field_size * HECTARES_TO_SQUARE_MILLIMETERS) * \
             CUBIC_MILLIMETERS_TO_CUBIC_METERS
         soil_mass_in_kg = self.bulk_density * MEGAGRAMS_TO_KILOGRAMS * soil_volume_in_cubic_meters
+        total_soil_carbon_amount = (soil_mass_in_kg * (self.percent_organic_carbon_content / 100) / field_size)
```
matthew7838, line 430: `self.percent_organic_carbon_content ` is the only part that is still using percent instead of a fraction. I think we can convert them in this PR, there should be just some minor work.

matthew7838 (approved):
LGTM, a minor comment that I do think it's better to be addressed in this PR, but that's not stopping the approval.

RUFAS/routines/field/soil/layer_data.py
ew3361zh: Thanks, Matthew. I'm not as familiar with the C&S module - can you explain a little more what you are asking to happen here?

Isgotkowitz: @matthew7838 `LayerData` is still using clay, sand, silt, and rock contents as percents and not fractions, so I think any changes that need to be made would be in the soil input files (looks like there are some that need updating though). Seems like it would be better to leave those for another PR though, because it is not always easy to determine if those inputs are off or not.

ew3361zh: @Isgotkowitz will make an issue about this to ensure it is handled

PR #963, opened 2023-11-15
Quick debug in main

allisterakun:
A bug was introduced following @ew3361zh's #861.

In `RUFAS/routines/animal/life_cycle/cow.py`, line 361, when referring to the id of a cow, it should be `self.id` but not `self.cow_id`
image


ew3361zh (approved):
Thank you, good catch!

Isgotkowitz (approved):

JoeWaddell (approved):
Nice quick fix so that main runs again.

Note: if not already known, the key changes in the PR that prompted this one necessitate an update to the `graph_filter_example.json`, since it changed related om keys.

PR #972, opened 2023-11-17
Calculate dry matter fraction of crop yields

### LINKED ISSUE #946: Calculate and report dry matter mass of crop yields
In order to provide both end users and the Feed Storage module with needed information about crop yields, the dry matter mass of yields needs to be calculated and reported along with the total biomass of crop yields.

When a crop is harvested, the dry matter mass should be calculated by subtracting the mass of water held in the harvested crop from the total biomass in the yield, and this value should be stored and reported to the Output Manager.

**Additional context**
In order to resolve this issue, a tracker for the amount of water held by the crop may have to be introduced in `CropData` and updated daily by the Water Uptake and Water Dynamics submodules in the Crop module.

Isgotkowitz:
This PR adds dry matter fractions to the Crop Configurations specified in the `crop_data.py` file. It also expands how yields are recorded and stored in `crop_management.py`. Dry matter fractions for various crops are recorded on page 4 of [this](https://docs.google.com/document/d/1auWeT6XnItd8yLoWGkbSwJ2DkUULxv5ClA48Grm69IE/edit#heading=h.h0dz48r2a92t) Google Doc.

Before it was decided that hard-coded dry matter fractions would be used to calculate the dry matter mass of yields, I attempted to create a water balance tracker for crops. In the course of this, I found some issues in the way that water stress was being calculated and tracked:
- There were two trackers for the total amount of water that the crop was taking up on a daily basis. This has been changed so that there is only one tracker now, and it is being used consistently throughout the entire codebase.
- `water_dynamics.py` was using the cumulative evapotranspiration to calculate the water deficiency, instead of the cumulative transpiration (which is the same thing as water uptake in SWAT).

Docstrings have been updated and TODOs removed as appropriate in methods/files touched.


RUFAS/routines/field/crop/crop_data.py
```diff
@@ -493,6 +501,7 @@ class Corn(CropData):
 
     optimal_harvest_index: float = 0.50
     min_harvest_index: float = 0.30
+    dry_matter_percentage: float = 86.0907
```
ew3361zh, line 504: Is there any reason this and the CornSilage dry_matter_percentage attributes didn't get notes assigned to them similar to how the rest did?

RUFAS/routines/field/crop/crop_data.py
```diff
@@ -330,18 +328,26 @@ class CropData:
     """cumulative water demands not met by all previous layers"""
     actual_water_uptakes: Optional[List[float]] = None
     """the actual amount of water to be removed from the soil"""
-    total_water_uptake: float = 0.0
-    """Total amount of water the plant took from the soil on the current day (mm)"""
+    water_uptake: float = 0.0
+    """Total amount of water the plant took from the soil on the current day (mm). Note that SWAT considers the amount
+        of water taken up on any given day to be the actual amount of transpiration on that day."""
+    cumulative_water_uptake: float = 0.0
+    """Cumulative sum of water uptaked by the plant (mm)."""
```
ew3361zh: uptaken? uptook?

ew3361zh: Also curious how`cumulative_water_uptake` differs from `water_uptake`

RUFAS/routines/field/crop/crop_data.py
```diff
@@ -330,18 +328,26 @@ class CropData:
...
+    """
+    Percentage of fresh yield that is dry matter (unitless).
+    Note: this value is the default for Sorghum harvested as a grain.
+    """
     dry_down_fraction: float = 0.2
     """proportion of plant biomass that is lost to dry-down (unitless; [0, 1])"""
     optimal_phosphorus_fraction: float = 0.073
     """optimal proportion of the plant's biomass comprised of nitrogen for the current growth stage (unitless)"""
-    user_harvest_index: Optional[float] = None  # TODO: handle user input for this. - GitHub Issue #246
+    user_harvest_index: Optional[float] = None
```
ew3361zh, line 350: I couldn't find this in the metadata/input files - does it need to be in there?

RUFAS/routines/field/crop/crop_management.py
```diff
@@ -232,14 +229,15 @@ def _record_yield(self, field_name: str, field_size: float, year: int, day: int)
             Julian day on which this harvest occurred.
 
         """
-        mass_harvested = self.data.yield_collected
+        wet_mass_harvested = self.data.wet_yield_collected
+        dry_mass_harvested = self.data.dry_matter_yield_collected
```
ew3361zh: Just curious why you switch between `x_yield_collected` and `x_mass_harvested` (insert christmas joke here).

ew3361zh (approved):
Looks good. Just left a smattering of comments and questions but none felt big enough to require changes before approving.

RUFAS/routines/field/crop/crop_data.py
ew3361zh: e.g.

"""
    Note: this value comes from the xxxx configuration values.
"""

RUFAS/routines/field/crop/crop_data.py
Isgotkowitz: Good catch. `cumulative_water_uptake` is total amount of water that is taken up over the lifetime of the crop, `water_uptake` is the amount of water taken up on a single day. Will update comment

RUFAS/routines/field/crop/crop_data.py
Isgotkowitz: No - but it is written in a bunch of places in the code and tested, so I think it would be better to leave it in for now, and wait to mess with it until #246 comes up.

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: No reason, they both are pretty clear. Do you think it would be better to make them consistent?

RUFAS/routines/field/crop/crop_data.py
Isgotkowitz: Because corn silage is the only crop config that actually has its harvest type built into the name - but these notes should be in there for corn, because corn grain doesn't actually have "grain" in the name.

RUFAS/routines/field/crop/crop_data.py
```diff
@@ -863,6 +911,10 @@ class SugarBeet(CropData):
 
     optimal_harvest_index: float = 2.00
     min_harvest_index: float = 1.10
+    dry_matter_percentage: float = 22.5
+    """
+    Note: this value is set equal to the dry matter percentage for Potatoes. It is expected to be deprecated as well.
```
krb78, line 916: I think this is reasonable. I can ask a friend who works for the sugar beet industry what the DM% is of beets when they are harvested, if that's more helpful. I found a paper from New Zealand and don't imagine it to be much different in the US.

RUFAS/routines/field/crop/crop_management.py
krb78: I think it'd be better to have them be consistent - I'm a fan of "x_yield_collected"

RUFAS/routines/field/crop/crop_data.py
```diff
@@ -900,6 +952,10 @@ class Potato(CropData):
 
     optimal_harvest_index: float = 0.95
```
krb78, line 953: (Not that familiar with potato harvest), but is the reason the optimal_harvest_index is less than 1 for potatoes because they have much more above ground biomass left behind than sugar beets do? I thought they'd both be >1

krb78 (approved):
Just had some minor comments, Ed. I think these changes look great! :)

RUFAS/routines/field/crop/crop_data.py
Isgotkowitz: That is a great question - the SWAT database has both the optimal and minimum harvest indices listed as 0.95 for potatoes, but it seems incorrect to have them be less than one.

> is the reason the optimal_harvest_index is less than 1 for potatoes because they have much more above ground biomass left behind

I think it would still need to be > 1 though, because no matter how SWAT (and RuFaS) slices it, if the harvest index is < 1 then 100% of root biomass will be left in the ground.

When I get to updating the crop configurations to meet what we discussed last week, I am going to remove the crops we don't explicitly need for V1, including potatoes.

RUFAS/routines/field/crop/crop_data.py
Isgotkowitz: I am planning to remove Sugar beets from the available crops because they aren't explicitly needed for V1 and aren't planned to be supported by other modules (as far as I am aware), so I don't think that would be necessary. However, if you would like to have sugar beets in RuFaS I could leave them in and set the parameters based on what your friend recommends.

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: Changed. 

> `x_mass_harvested` (insert christmas joke here).

Stellar Christmas joke @ew3361zh

Isgotkowitz:
Documenting future plans: potatoes and sugar beets will be removed in subsequent PRs. They are not required for V1, and I discussed this change with Kaitlyn in Slack.

PR #974, opened 2023-11-17
Integrate carbon to tillage

### LINKED ISSUE #945: Integrate Carbon into the `TillageApplication` routines.
Carbon pools are not effected by tillage operations currently, which is problematic because tillage operations have a large effect on soil carbon levels.

The metabolic litter, structural litter, active carbon and slow carbon pools in every soil layer should be effected by the tillage in the same way as soil nitrogen and phosphorus.

The passive soil carbon should be effected the same way, but should only be mixed between subsurface pools.

**Additional context**
For the carbon pools that are mixed between all soil layers, it seems that they can be mixed by simply adding the names of those pools to the list `pools_to_till_in_soil` in the method `till_soil()` in the file `tillage_application.py`. Custom functionality may have to be introduced in order to mix only subsurface soil pools.

ew3361zh:
PR to address issue #945. No unit testing written yet.


RUFAS/routines/field/field/tillage_application.py
```diff
@@ -139,17 +144,23 @@ def _mix_soil_layers(self, pool_name: str, tillage_depth: float, mixing_fraction
         """
         redistribution_fractions = []
         total_to_mix_from_pools = 0
+        top_layer_offset = 0
+        is_passive_carbon_pool = pool_name == "passive_carbon_amount"
         for layer in self.soil_data.soil_layers:
+            if is_passive_carbon_pool and layer.top_depth == 0:
```
Isgotkowitz: I like this introduction of an offset into the `_mix_soil_layers()` method, but I think its functionality should be exposed in the method signature (i.e. add a boolean argument that says whether or not the surface soil layer should be tilled), instead of checking whether the pool that was passed was passive carbon. I don't anticipate any other pools being added that would not get tilled in the surface soil layer, but I think it would be cleaner to restrict the checking to `till_soil()`.

Isgotkowitz (commented):
Great start Niko. I think the check for passive carbon could be shifted around to make it the method more future proof

RUFAS/routines/field/field/tillage_application.py
ew3361zh: Sounds good. I've added it as an arg.

RUFAS/routines/field/field/tillage_application.py
```diff
@@ -106,12 +106,22 @@ def till_soil(self, tillage_depth: float, incorporation_fraction: float, mixing_
...
+                                 "structural_litter_amount",
+                                 "active_carbon_amount",
+                                 "slow_carbon_amount",
+                                 "passive_carbon_amount"]
+        pools_to_offset_top_layer = ["passive_carbon_amount"]
         for pool in pools_to_till_in_soil:
-            self._mix_soil_layers(pool, tillage_depth, mixing_fraction)
+            offset_top_layer = False
+            if pool in pools_to_offset_top_layer:
+                offset_top_layer = True
```
Isgotkowitz:
```suggestion
            offset_top_layer = pool in pools_to_offset_top_layer
```

RUFAS/routines/field/field/tillage_application.py
Isgotkowitz, line 1: Could you please add a note to the `Notes` section that specifically mentions that passive carbon is not mixed into the top soil layer?

Isgotkowitz (approved):
Looks great Niko, nice job.

RUFAS/routines/field/field/tillage_application.py
ew3361zh: much better

RUFAS/routines/field/field/tillage_application.py
ew3361zh: Will do!

matthew7838 (approved):
LGTM. Tests and everything are working fine.

PR #976, opened 2023-11-19
Report generation

PooyaHekmati:
This PR implements report generation. I wrote a comprehensive wiki for it, so instead of repeating myself here, I refer to https://github.com/RuminantFarmSystems/MASM/wiki/Report-Generator .


RUFAS/report_generator.py
Isgotkowitz, line 1: Can these aggregator methods be wrapped in some type of class?

RUFAS/output_manager.py
```diff
@@ -785,16 +786,20 @@ def save_results(
         )
         list_of_filter_files = self._list_filter_files_in_dir(filters_dir_path)
         for filter_file in list_of_filter_files:
+            info_map["filter file"] = filter_file
```
Isgotkowitz, line 789: Why `"filter file"` instead of `"filter_file"`?

RUFAS/report_generator.py
```diff
@@ -0,0 +1,292 @@
...
+            The aggregated report data as a list.
+
+        Raises
+        ------
+        ValueError
+            If the report data is empty or if the necessary aggregation keys are not found in filter_content.
+        """
+        selected_variables = filter_content.get("variables")
+        slice_start = filter_content.get("slice_start", 0)
+        slice_end = filter_content.get("slice_end", 0)
```
Isgotkowitz: What about defaulting to -1 instead, and possibly also removing the rule in `_prepare_report_data()` that 0 indicates the position of the final element? Making 0 map to the final position of a list definitely makes the code more resilient, but also makes it more confusing.
```suggestion
        slice_end = filter_content.get("slice_end", -1)
```

RUFAS/report_generator.py
```diff
@@ -0,0 +1,292 @@
...
+    "sum": sum_aggregator,
+    "subtraction": subtraction_aggregator
+}
+
+
+class ReportGenerator:
+    def generate_report(
+        self,
+        filtered_pool: Dict[str, Dict[str, List[Any]]],
+        filter_content: Dict[str, str | int],
```
Isgotkowitz: The key `"variables"` in this argument is supposed to point to a list of strings, if I understand the code correctly.
```suggestion
        filter_content: Dict[str, str | int | List[str]],
```

RUFAS/report_generator.py
```diff
@@ -0,0 +1,292 @@
...
+            The filtered data pool with each key mapping to its respective data element.
+
+        selected_variables : List[str]
+            Variables to be included from the filtered pool.
+
+        slice_start : int
+            Starting index for slicing data elements.
+
+        slice_end : int
+            Ending index for slicing; slices to end if set to 0.
```
Isgotkowitz: Should there be any kind of validation/warning mechanism that checks to make sure the slice start is actually in front of the slice end?

RUFAS/report_generator.py
```diff
@@ -0,0 +1,292 @@
...
+                for temp_key, temp_values in temp_data.items():
+                    if temp_key not in selected_variables:
+                        continue
+                    if temp_key in report_data:
+                        report_data[temp_key].extend(temp_values)
+                    else:
+                        report_data[temp_key] = temp_values
+            else:
+                if key in report_data:
+                    report_data[key].extend(filtered_pool[key]["values"][
```
Isgotkowitz: This is the only line of code not covered by the test suite.

RUFAS/report_generator.py
```diff
@@ -0,0 +1,292 @@
...
+        -------
+        Dict[str, List[Any]]
+            Processed data suitable for report generation, keyed by selected variables.
+
+        Raises
+        ------
+        KeyError
+            If selected_variables is None and the data within the pool requires variable selection.
+        """
+        report_data: Dict[str, List[Any]] = {}
```
Isgotkowitz, line 255:
```suggestion
        report_data: Dict[str, List[Any]] = {}
        slice_end = slice_end if slice_end != 0 else -1
```
Then the slice end doesn't have to be rechecked every time a slice is grabbed from the filtered pool.

Isgotkowitz (approved):
Nice job Pooya. Handful of comments, may or may not be appropriate to address them in this PR.

A helpful feature would be the ability to specify no vertical or horizontal aggregator function. I think this would be a benefit so that someone could collect the raw data and the sum, average, etc all in one pass, then if something looks off they can easily start scrolling through the data because it is already right there

RUFAS/output_manager.py
```diff
@@ -803,22 +808,52 @@ def save_results(
...
+                            if report_name not in reports.keys()
+                            else f"{report_name} {self._get_timestamp(True)}"
+                        ] = {
+                            "values": report_generator.generate_report(
+                                filtered_pool,
+                                filter_content,
+                            )
+                        }
+                    except (ValueError, KeyError) as e:
+                        self.add_error("report generation error", str(e), info_map)
```
ew3361zh, line 834: Should we raise the exception as we're doing with others (e.g. on lines 915-916 within `_save_variables_to_csv_files()`)?

ew3361zh (approved):
LGTM. Was working well with the example JSON file from the Wiki. I think getting some SME feedback would be helpful as they'll be some of the primary users on this but I think it's okay to get it merged and into use. I think examples of anything more they may need will come up through use. I tried a few combinations but lots of options available and nothing unexpected was happening with the tweaks I tried.

RUFAS/output_manager.py
PooyaHekmati: there is no reason behind it, both are fine.

RUFAS/output_manager.py
PooyaHekmati: If we raise an error here, then the rest of the filter file which might be intact won't be processed. I think it is better to give the user as much as we can instead of panicking on the first error.

RUFAS/report_generator.py
PooyaHekmati: Yes, and they will most likely be. But for now I think it is better to keep them here.

RUFAS/report_generator.py
PooyaHekmati: I agree that this is not very clean and I am open to improvements. But `-1` won't work. The `end` in python is always exclusive, meaning this code:
```
x=[1, 2, 3, 4, 5]
print(x[:-1])
```
prints:
```
[1, 2, 3, 4]
````

RUFAS/report_generator.py
PooyaHekmati: Answered above

RUFAS/report_generator.py
Isgotkowitz: Good catch, I should have tried it before I suggested it. What about using `None` instead of -1? (That one I did try, seems like it would do the trick)

PooyaHekmati: Yes, it works! thanks for the suggestion

RUFAS/report_generator.py
PooyaHekmati: Yes, but let's keep it unchecked for now and observe how users will react to the whole thing. I'm anticipating some changes.

Issue #979, opened 2023-11-20
Evaluate soil nitrogen dynamics


Isgotkowitz:
Evaluate nitrogen dynamics applied to field via chemical and manure fertilizer.

PR #983, opened 2023-11-20
Update percentages of organic carbon content in soil profile input files

### LINKED ISSUE #982: Review inputs to make sure `percent_organic_carbon_content` is being passed as percent and not as fraction.
Soil profile input files need to be double-checked and gut-checked to make sure that `percent_organic_carbon_content` is being passed as a percent and not a fraction.

Isgotkowitz:
Doubles checks all soil profile input files to make sure that the percent organic carbon content matches its input source.

Profile to source mappings:
- `ARL_soil.json` comes from [this excel sheet](https://3.basecamp.com/3486446/buckets/5296287/uploads/2928055772).
- `regular_barnyard_soil.json` comes from [this excel sheet](https://3.basecamp.com/3486446/buckets/5296287/uploads/2788739772).
- `testing_soil.json` is just a slight variation on `regular_barnyard_soil.json`.
- `default_soil.json` is a copy of `regular_barnyard_soil.json`.
- `default_soil_2.json` is a copy of `ARL_soil.json`.

Also, the `OM_percent` field has been removed. This parameter may or may not be brought back to replace `percent_organic_carbon_content` for RuFaS V1, and if it is resurrected it can easily be calculated using SWAT Theoretical documentation eqn. 4:1.1.4

SWAT eqn. 4:1.1.4 is used to determine the fraction of a soil layer that is organic carbon based on the amount of organic matter in the soil layer - and can be used in the reverse direction to estimate organic matter based on organic carbon content.

Closes #982

### DIFF
--- a/input/data/soil/ARL_soil.json
+++ b/input/data/soil/ARL_soil.json
@@ -18,7 +18,7 @@
             "saturated_hydraulic_conductivity": 9.17,
             "initial_temperature": 15.77575,
             "bulk_density": 1.34,
-            "percent_organic_carbon_content": 0.012,
+            "percent_organic_carbon_content": 2.32,
             "percent_clay_content": 21.95,
             "percent_silt_content": 63.42,
             "percent_sand_content": 14.63,
@@ -33,8 +33,7 @@
             "denitrification_threshold_water_content": 1.1,
             "residue_fresh_organic_mineralization_rate": 0.05,
             "denitrification_rate": 0.1,
-            "active_N_percent": 0.02,
-            "OM_percent": 0.04
+            "active_N_percent": 0.02
         },
         {
             "bottom_depth": 1041.4,
@@ -45,7 +44,7 @@
             "saturated_hydraulic_conductivity": 9.17,
             "initial_temperature": 14.50797297,
             "bulk_density": 1.42,
-            "percent_organic_carbon_content": 0.012,
+            "percent_organic_carbon_content": 0.348,
             "percent_clay_content": 27.27,
             "percent_silt_content": 59.09,
             "percent_sand_content": 13.64,
@@ -60,8 +59,7 @@
             "denitrification_threshold_water_content": 1.1,
             "residue_fresh_organic_mineralization_rate": 0.05,
             "denitrification_rate": 0.1,
-            "active_N_percent": 0.02,
-            "OM_percent": 0.006
+            "active_N_percent": 0.02
         },
         {
             "bottom_depth": 1168.4,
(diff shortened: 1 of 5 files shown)


ew3361zh (approved):
LGTM.

It does raise the interesting point of whether we need a more rigorous documentation of ALL our "default" inputs. Not sure if that is already happening somewhere.

input/data/soil/ARL_soil.json
```diff
@@ -18,7 +18,7 @@
             "saturated_hydraulic_conductivity": 9.17,
             "initial_temperature": 15.77575,
             "bulk_density": 1.34,
-            "percent_organic_carbon_content": 0.012,
+            "percent_organic_carbon_content": 2.32,
```
ew3361zh, line 21: I'm guessing from the linked ARL_soil.json spreadsheet that the OrgC% is not a percent and that's why this number doesn't match? Maybe why this number was updated to 2.32 based on the spreadsheet number showing 0.012 would be helpful to note in the PR description.

Isgotkowitz: You know I am not sure how/where they got the values for organic carbon percents in that spreadsheet. They certainly don't look right. But the organic carbon percent can be estimated from the organic matter percent with the equation $OrganicCarbonPercent=0.58*OrganicMatterPercent$. Since the values they have for organic matter in that spreadsheet seem more correct, I think for now it is better to opt for calculating the organic carbon percents using that equation.

Isgotkowitz:
> It does raise the interesting point of whether we need a more rigorous documentation of ALL our "default" inputs. Not sure if that is already happening somewhere.

@ew3361zh I think the input documentation should definitely get compiled into a more comprehensive and readable source. There is already some in the `"description"`s of properties in the metadata, but I don't know how consistently this is done for all the properties.

input/data/soil/default_soil_2.json
```diff
@@ -60,8 +59,7 @@
             "denitrification_threshold_water_content": 1.1,
             "residue_fresh_organic_mineralization_rate": 0.05,
             "denitrification_rate": 0.1,
-            "active_N_percent": 0.02,
-            "OM_percent": 0.006
+            "active_N_percent": 0.02
```
matthew7838, line 62: Is this percent or fraction? Since it's really small if it's in percent, I figured would be better check with you again.

matthew7838 (approved):
LGTM, one small comment I would like to hear your feedback.

input/data/soil/ARL_soil.json
ew3361zh: Sounds good. Is that equation what you note is coming from SWAT eqn. 4:1.1.4 in the PR description?

Isgotkowitz: Yes

input/data/soil/default_soil_2.json
Isgotkowitz: Good question, I am not sure yet. That input currently is not being used either, so I will look into whether or not it is needed while I work on #979

PR #984, opened 2023-11-20
IM runtime populate pool

allisterakun:
This PR: 
1. Added two public methods `add_dict_variable_to_pool()` and `add_tabular_variable_to_pool()`  and two private methods `_add_variable_to_pool()` and `_metadata_properties_exists` in `InputManager`:
    - `add_dict_variable_to_pool()`
        
            Adds a dictionary variable to the InputManager's pool after validating it against metadata.

            This function takes in a variable along with its name and a key to access its validation metadata.
            It validates the data against the provided metadata and adds the data to the InputManager pool if it is valid.
    
            Parameters
            ----------
            variable_name: str
                The name of the dictionary variable to be added.
            data : Dict[str, Any]
                The data of the variable, structured as a dictionary.
            properties_blob_key : str
                A key used to locate the metadata for validation of the variable.
            eager_termination : bool
                If True, raises a RuntimeError when the variable is invalid.
                If False, the function returns False without raising an error.
    
            Returns
            -------
            bool
                True if the variable is successfully validated and added to the pool.
                False if the variable is invalid and not added to the pool.
    
            Raises
            -------
            TypeError
                If `data` is not the expected type of Dict[str, Any].

    - `add_tabular_variable_to_pool()`
            
           Adds a tabular variable to the InputManager's pool after validating it against metadata.
    
            This function takes in a variable along with its name and a key to access its validation metadata.
            It validates the data against the provided metadata and adds the data to the InputManager pool if it is valid.
    
            Parameters
            ----------
            variable_name: str
                The name of the variable to be added.
            data : Dict[str, List[Any]] | List[Any]
                The data of the tabular variable, structured as a dictionary of lists or a list.
            properties_blob_key : str
                A key used to locate the metadata for validation of the variable.
            eager_termination : bool
                If True, raises a ValueError when the variable is invalid.
                If False, the function returns False without raising an error.
    
            Returns
            -------
            bool
                True if the variable is successfully validated and added to the pool.
                False if the variable is invalid and not added to the pool.
    
            Raises
            -------
            TypeError
                If `data` is not the expected type of Dict[str, List[Any]] | List[Any].
    - `_add_variable_to_pool()`
            
            Adds a variable to the pool after validating its data against specified metadata properties.
    
            Notes
            -----
            This function processes and validates the input data for a variable based on
            its metadata properties. It then adds the validated data to a pool. The function
            also provides an option for eager termination in case of invalid data.
    
            Parameters
            ----------
            variable_name : str
                The name of the variable to be added to the pool.
            data : Dict[str, Any]
                The data associated with the variable that needs validation and addition to the pool.
            properties_blob_key : str
                The key in the metadata properties against which the data is validated.
            eager_termination : bool
                Flag indicating whether the function should return early in case of invalid data.
            is_variable_dict : bool
                Weather the variable is a dictionary variable (rather than tabular).
    
            Returns
            -------
            bool
                 True if the variable is successfully added, False otherwise.
    
            Raises
            -------
            ValueError
                If eager_termination is True and the variable failed validation.
            
    - `_metadata_properties_exists()`
           
            Checks if specific properties exist in the metadata for a given variable.
    
            Notes
            -----
            This function is designed to verify the existence of specified properties
            within the metadata of a particular variable. It returns a boolean indicating
            the existence of the properties, and a KeyError in case of missing metadata
            or properties.
    
            Parameters
            ----------
            variable_name : str
                The name of the variable for which the metadata is to be checked.
            properties_blob_key : str
                The key representing the specific properties blob in the metadata to check.
    
            Returns
            -------
            bool
                True if the properties exist, False otherwise.
    
            Raises
            -------
            ValueError
                If no metadata is loaded in InputManager.__metadata.
            KeyError
                If no metadata properties can be found with the given `properties_blob_key`.
       

3. Wrote unit tests for all new functions.
4. Renamed the following methods in `InputManager`:
    - `_validate_json_element()` => `_validate_dict_element()`
    - `_validate_csv_element()` => `_validate_tabular_element()`


RUFAS/input_manager.py
```diff
@@ -145,15 +144,25 @@ def _populate_pool(self, eager_termination: bool) -> bool:
...
         -------
         bool
             True if data is valid, otherwise False.
         """
         info_map = {"class": self.__class__.__name__,
-                    "function": self._populate_pool.__name__,
+                    "function": self.populate_pool.__name__,
                     }
+        if variable_name is not None and data is not None:
+            self.__pool[variable_name] = data
```
PooyaHekmati: The variable has to be validated using metadata before being added to the pool. This is why we said we should rename the `_validate_json_element` and `_validate_csv_element` functions.

PooyaHekmati (changes requested):

RUFAS/input_manager.py
allisterakun: Got it! Will update!

RUFAS/input_manager.py
```diff
@@ -159,8 +158,8 @@ def _populate_pool(self, eager_termination: bool) -> bool:
         total_elements_counter = 0
         fixed_elements_counter = 0
 
-        data_type_to_loader_map = {"json": self._load_data_from_json,
-                                   "csv": self._load_data_from_csv}
+        data_type_to_loader_map: Dict[str, Callable] = {"json": self._load_data_from_json,
```
PooyaHekmati, line 161: thanks for adding type hints

RUFAS/input_manager.py
```diff
@@ -180,18 +179,21 @@ def _populate_pool(self, eager_termination: bool) -> bool:
...
+                                                                                  properties_blob_key,
+                                                                                  filtered_input_data,
+                                                                                  eager_termination,
+                                                                                  element_counter_and_validity)
 
                 fixed_elements_counter += element_counter_and_validity["fixed_elements"]
                 valid_elements_counter += element_counter_and_validity["valid_elements"]
                 total_elements_counter += element_counter_and_validity["total_elements"]
                 if element_counter_and_validity["is_valid"]:
+                    # POTENTIAL ISSUE!!!
```
PooyaHekmati: do you want to elaborate more?

RUFAS/input_manager.py
```diff
@@ -789,3 +793,40 @@ def flush_pool(self) -> None:
                     }
         self.__pool = {}
         om.add_log("Clear variable pool", "The pool is emptied.", info_map)
+
+    def add_variable_to_pool(self, variable_name: str, data: Dict[str, Any], properties_blob_key: str,
```
PooyaHekmati: what does this do?
add docstring, even if it is a one-liner

RUFAS/input_manager.py
```diff
@@ -789,3 +793,40 @@ def flush_pool(self) -> None:
...
+        valid_elements_counter = 0
+        invalid_elements_counter = 0
+        total_elements_counter = 0
+        fixed_elements_counter = 0
+
+        metadata_properties = self.__metadata["properties"][properties_blob_key]
+        for metadata_property in metadata_properties.keys():
+            element_counter_and_validity = {"fixed_elements": 0, "total_elements": 0, "valid_elements": 0,
+                                            "invalid_elements": 0, "is_valid": True}
+            element_counter_and_validity = self._validate_dict_element([metadata_property], properties_blob_key,
```
PooyaHekmati: what if data is not in dict?

PooyaHekmati (commented):
The PR seems to be on the right direction; however, the description states: `Updates InputManager._populate_pool to a public method InputManager.populate_pool.`; But I still see `_populate_pool` in the code. Also, the newly added function `add_variable_to_pool` is not mentioned in the description, is missing unit tests and docstring.

RUFAS/input_manager.py
allisterakun: It now can take in either a Dict or a List

RUFAS/input_manager.py
allisterakun: Done

RUFAS/input_manager.py
allisterakun: Will make a Git issue

allisterakun:
> The PR seems to be on the right direction; however, the description states: `Updates InputManager._populate_pool to a public method InputManager.populate_pool.`; But I still see `_populate_pool` in the code. Also, the newly added function `add_variable_to_pool` is not mentioned in the description, is missing unit tests and docstring.

All updated

RUFAS/input_manager.py
PooyaHekmati: #1014

RUFAS/input_manager.py
```diff
@@ -789,3 +793,92 @@ def flush_pool(self) -> None:
...
+
+        Returns
+        -------
+        bool
+            True if the variable is successfully validated and added to the pool.
+            False if the variable is invalid and not added to the pool.
+
+        Raises
+        -------
+        RuntimeError:
```
PooyaHekmati: Maybe value error is a better choice here?
https://docs.python.org/3/library/exceptions.html#ValueError

RUFAS/input_manager.py
```diff
@@ -789,3 +793,92 @@ def flush_pool(self) -> None:
...
+            If eager_termination is True and the variable failed validation.
+
+        The function logs the count of total, valid, fixed, and invalid elements of the variable during validation.
```
PooyaHekmati: Either move this line to the beginning of the docstring or add it under Notes

RUFAS/input_manager.py
```diff
@@ -789,3 +793,92 @@ def flush_pool(self) -> None:
...
+        fixed_elements_counter = 0
+
+        metadata_properties = self.__metadata["properties"][properties_blob_key]
+        for metadata_property in metadata_properties.keys():
+            element_counter_and_validity = {
+                "fixed_elements": 0,
+                "total_elements": 0,
+                "valid_elements": 0,
+                "invalid_elements": 0,
+                "is_valid": True
```
PooyaHekmati: maybe it's better to initialize it with False?

RUFAS/input_manager.py
```diff
@@ -789,3 +793,92 @@ def flush_pool(self) -> None:
...
+                input_data=data,
+                eager_termination=False,
+                element_counter_and_validity=element_counter_and_validity
+            ) if isinstance(data, Dict) else self._validate_tabular_element(
+                var_name=variable_name,
+                properties_blob_key=properties_blob_key,
+                input_data={variable_name: data},
+                eager_termination=False,
+                element_counter_and_validity=element_counter_and_validity
+            )
```
```suggestion
            if isinstance(data, Dict):
              element_counter_and_validity = self._validate_dict_element(
                  element_hierarchy=[metadata_property],
                  properties_blob_key=properties_blob_key,
                  input_data=data,
                  eager_termination=False,
                  element_counter_and_validity=element_counter_and_validity
              ) 
            elif isinstance(data, List):
              element_counter_and_validity = self._validate_tabular_element(
                  var_name=variable_name,
                  properties_blob_key=properties_blob_key,
                  input_data={variable_name: data},
                  eager_termination=False,
                  element_counter_and_validity=element_counter_and_validity
              )
           else:
             raise TypeError with a clear message, don't forget to update the docstring accordingly
```

RUFAS/input_manager.py
```diff
@@ -789,3 +793,92 @@ def flush_pool(self) -> None:
...
+                   f"{total_elements_counter=}", info_map)
+        om.add_log(f"Validation count for variable {variable_name}: total valid",
+                   f"{valid_elements_counter=}", info_map)
+        om.add_log(f"Validation count for variable {variable_name}: total fixed",
+                   f"{fixed_elements_counter=}", info_map)
+        om.add_log(f"Validation count for variable {variable_name}: total invalid",
+                   f"{invalid_elements_counter=}", info_map)
+
+        if invalid_elements_counter == 0:
+            self.__pool[variable_name] = data
```
PooyaHekmati: maybe first check if `self.__pool[variable_name]` has a value, if does, then add a warning that we are overwriting it.

RUFAS/input_manager.py
```suggestion
        RuntimeError
```

tests/test_input_manager.py
```diff
@@ -1855,3 +1874,139 @@ def test_flush_pool(mock_input_manager: InputManager) -> None:
...
+        with patch("RUFAS.output_manager.OutputManager.add_error") as mock_om_add_error:
+            result = mock_input_manager.add_variable_to_pool(variable_name=variable_name,
+                                                             data=data,
+                                                             properties_blob_key="properties1",
+                                                             eager_termination=False)
+
+    assert result is False
+    assert mock_om_add_log.call_count == 4
+    assert mock_om_add_error.call_count == 1
+    assert variable_name not in mock_input_manager._InputManager__pool
```
Isgotkowitz: If the variable name was already in the pool, maybe it would be worthwhile to add a check that the old data is still present.

RUFAS/input_manager.py
```diff
@@ -789,3 +793,92 @@ def flush_pool(self) -> None:
...
+        """
+        info_map = {"class": self.__class__.__name__,
+                    "function": self.add_variable_to_pool.__name__,
+                    }
+        valid_elements_counter = 0
+        invalid_elements_counter = 0
+        total_elements_counter = 0
+        fixed_elements_counter = 0
+
+        metadata_properties = self.__metadata["properties"][properties_blob_key]
```
Isgotkowitz, line 882: Would it be worthwhile to add a check here to make sure that metadata has actually been loaded before trying to add a variable on the fly?

Isgotkowitz (approved):
Code looks great, two comments on checking/testing edge cases and one on a docstring typo.

RUFAS/input_manager.py
PooyaHekmati: yes, also check if `properties_blob_key` does exist.

tests/test_input_manager.py
allisterakun: Good catch!

RUFAS/input_manager.py
allisterakun: Will do!

RUFAS/input_manager.py
allisterakun: Done

RUFAS/input_manager.py
allisterakun: Done

RUFAS/input_manager.py
allisterakun: Done

RUFAS/input_manager.py
allisterakun: Done

RUFAS/input_manager.py
allisterakun: Done

tests/test_input_manager.py
allisterakun: Done!

RUFAS/input_manager.py
```diff
@@ -789,3 +793,121 @@ def flush_pool(self) -> None:
                     }
         self.__pool = {}
         om.add_log("Clear variable pool", "The pool is emptied.", info_map)
+
+    def add_variable_to_pool(self, variable_name: str, data: Dict[str, Any] | List[Any], properties_blob_key: str,
+                             eager_termination: bool) -> bool:
```
Isgotkowitz: Will this work for more complex tabular data i.e. data from a CSV that has multiple tables? The way it is set up now, it could only accept a single "column" of data, which is problematic because that isn't the way most of the properties for tabular data are set up.

For example let's say I wanted to added another set of weather data, with the name "new_weather_data", properties blob key "weather_properties", and the actual data being `{"year": [1, 1, ...], "precip": [1.3, 0.0, 4.1, ...], ...}`. I think the method would accept this data, but would route the validation through `_validate_dict_element()` instead of `_validate_tabular_element()`, which would ultimately fail I think.

To work through another example, let's say that I only wanted to add a set of precipitation data. We'll call it "new_precip_data", with the properties key "weather_properties", and the actual data being `[0.0, 1.1, 3.55, ...]`. This adheres to all the type hints, but would fail because "weather_properties" expects there to be a "year" column, "jday" column, etc. And there isn't a way to specify a sub-property, like specifying the properties as "weather_properties.precip" would result in a KeyError.

Isgotkowitz (commented):
@allisterakun I think the way that tabular data is accepted and parsed by `add_variable_to_pool()` needs to be changed, but I will follow up and talk through this with you cause I might just be missing something.

RUFAS/input_manager.py
allisterakun: After conversation with @Isgotkowitz over Slack messages, I have decided to separate the current `add_variable_to_pool()` function into `add_dict_vairable_to_pool()` and `add_tabular_variable_to_pool()`. 
I am also looping in @PooyaHekmati for the information.

-------------------------------------------------------------------------------------------------------------------------------
Here is the conversation for documentation:
> **Ed Hansen**
  [9:44 AM](https://ruminantfarmsystems.slack.com/archives/D05KG2KPVJ5/p1701787461962509)
Hey Allister, I just reviewed #984 and I think I found a problem with the way tabular data is handled by add_data_to_pool(). I wrote it up in a comment, but let me know if you would like to meet and discuss it.
[9:44 AM](https://ruminantfarmsystems.slack.com/archives/D05KG2KPVJ5/p1701787494988279)
Also I will test it out quickly right now to actually verify that there is actually a problem - I might just be imagining things.

-------------------------------------------------------------------------------------------------------------------------------

> **Allister Liu**
  [9:45 AM](https://ruminantfarmsystems.slack.com/archives/D05KG2KPVJ5/p1701787544223749)
I think for complex tabular data, isn't it just the same as dictionary?
[9:46](https://ruminantfarmsystems.slack.com/archives/D05KG2KPVJ5/p1701787564996769)
b/c it will be a Dict[str, List[Any]]]

-------------------------------------------------------------------------------------------------------------------------------

> **Ed Hansen**
  [9:47 AM](https://ruminantfarmsystems.slack.com/archives/D05KG2KPVJ5/p1701787646397839)
Yes, but the issue is that it would get routed through the wrong validation method

-------------------------------------------------------------------------------------------------------------------------------

> **Allister Liu**
  [9:49 AM](https://ruminantfarmsystems.slack.com/archives/D05KG2KPVJ5/p1701787773313419)
Hmmmm I am wondering if the _validate_dict function can handle those Dict[str, List[Any]] data. If not, then I guess there is no way for us to distinguish a dict or a tabular without introducing an extra parameter. I guess I will have to make a new function for that?

-------------------------------------------------------------------------------------------------------------------------------

> Ed Hansen
  [9:56 AM](https://ruminantfarmsystems.slack.com/archives/D05KG2KPVJ5/p1701788180577269)
Probably. Unless it would be an acceptable solution to change the metadata in some way. For example the properties could be changed so that the tabular properties go from
```
"precip": {
        "description": "Amount of precipitation (mm H2O).",
        "type": "number",
        "minimum": 0,
        "default": 0
}
```
> to
```
"precip": {
  "description": "Amount of precipitation (mm H2O).",
  "type": "array",
  "properties": {
      "type": "number",
      "minimum": 0,
      "default": 0
}
```
> [9:58 AM](https://ruminantfarmsystems.slack.com/archives/D05KG2KPVJ5/p1701788280016779)
Or for each top level property, you could add a "type" - like this:
```
    "field_properties": {
      "format": "dict"
      "soil_specification": {...},
      "crop_specification": {...},
      ...
    },
    "weather_properties": {
      "format": "tabular"
      "year": {...},
      "jday": {...},
      ...
    },
```
>(edited)

>[9:58 AM](https://ruminantfarmsystems.slack.com/archives/D05KG2KPVJ5/p1701788327425139)
That might be well outside the scope of #984 though

-------------------------------------------------------------------------------------------------------------------------------

>Allister Liu
  [10:02 AM](https://ruminantfarmsystems.slack.com/archives/D05KG2KPVJ5/p1701788532152789)
Hmmm.... That's quite a big change actually. For this PR, I guess I will make a separate function add_tabular_variable_to_pool  and rename the current function to add_dict_variable_to_pool.
I do think treating everything as a dictionary and modifying the metadata for CSVs to add a "type": "array" would make sense. b/c at the end of the day when those CSVs are imported into IM, they are stored as Dict[str, List[Any]]. I will speak to Pooya about this and see what he thinks about it.

RUFAS/input_manager.py
```diff
@@ -789,3 +793,264 @@ def flush_pool(self) -> None:
...
+                The key representing the specific properties blob in the metadata to check.
+
+            Returns
+            -------
+            tuple
+                A tuple containing a boolean and an exception:
+                - The boolean is True if the properties exist, False otherwise.
+                - KeyError if the metadata is not loaded or the properties are not found,
+                  otherwise None.
+            """
```
Isgotkowitz: Shift docstring to the left by 1 tab.

RUFAS/input_manager.py
```diff
@@ -789,3 +793,264 @@ def flush_pool(self) -> None:
...
         om.add_log("Clear variable pool", "The pool is emptied.", info_map)
+
+    def _metadata_properties_exists(self, variable_name: str, properties_blob_key: str) -> (bool, KeyError):
+        """
+            Checks if specific properties exist in the metadata for a given variable.
+
+            This function is designed to verify the existence of specified properties
+            within the metadata of a particular variable. It returns a boolean indicating
+            the existence of the properties, and a KeyError in case of missing metadata
+            or properties.
```
Isgotkowitz: This should be housed in the "Notes" section of the docstring

RUFAS/input_manager.py
```diff
@@ -789,3 +793,264 @@ def flush_pool(self) -> None:
...
+                - The boolean is True if the properties exist, False otherwise.
+                - KeyError if the metadata is not loaded or the properties are not found,
+                  otherwise None.
+            """
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._metadata_properties_exists.__name__,
+                    }
+        if not self.__metadata:
+            om.add_error("No metadata loaded", "No metadata is loaded to the InputManager.", info_map)
+            return False, KeyError("No metadata loaded.")
```
Isgotkowitz: I think this is more of a `TypeError` or `ValueError` - maybe a good fix would be to return one of those error types here and make the return type hint be `(bool, TypeError | KeyError)` (or ValueError if that is more appropriate).

RUFAS/input_manager.py
```diff
@@ -789,3 +793,264 @@ def flush_pool(self) -> None:
...
+                The type of the variable (e.g., 'dict', 'tabular') to determine the validation method.
+
+            Returns
+            -------
+            tuple
+                A tuple containing a boolean and an exception:
+                - The boolean is True if the variable is successfully added, False otherwise.
+                - ValueError if invalid data is encountered and eager_termination is True,
+                  otherwise None.
+            """
```
Isgotkowitz: Shift docstring 1 tab to the left.

RUFAS/input_manager.py
```diff
@@ -789,3 +793,264 @@ def flush_pool(self) -> None:
...
+        return True, None
+
+    def _add_variable_to_pool(self, variable_name: str, data: Dict[str, Any], properties_blob_key: str,
+                              eager_termination: bool, variable_type: str) -> (bool, ValueError):
+        """
+            Adds a variable to the pool after validating its data against specified metadata properties.
+
+            This function processes and validates the input data for a variable based on
+            its metadata properties. It then adds the validated data to a pool. The function
+            also provides an option for eager termination in case of invalid data.
```
Isgotkowitz: This should go in the "Notes" section of the docstring.

RUFAS/input_manager.py
```diff
@@ -789,3 +793,264 @@ def flush_pool(self) -> None:
...
+            variable_name : str
+                The name of the variable to be added to the pool.
+            data : Dict[str, Any]
+                The data associated with the variable that needs validation and addition to the pool.
+            properties_blob_key : str
+                The key in the metadata properties against which the data is validated.
+            eager_termination : bool
+                Flag indicating whether the function should return early in case of invalid data.
+            variable_type : str
+                The type of the variable (e.g., 'dict', 'tabular') to determine the validation method.
```
Isgotkowitz: If `dict` and `tabular` are the only types that need to be worried about, maybe it would be better to make this input a boolean and call it `is_dict` or something.

RUFAS/input_manager.py
```diff
@@ -789,3 +793,264 @@ def flush_pool(self) -> None:
...
+                    element_counter_and_validity=element_counter_and_validity
+                )
+            elif variable_type == 'tabular':
+                element_counter_and_validity = self._validate_tabular_element(
+                    var_name=metadata_property,
+                    properties_blob_key=properties_blob_key,
+                    input_data=data,
+                    eager_termination=False,
+                    element_counter_and_validity=element_counter_and_validity
+                )
```
Isgotkowitz: What happens if the variable_type is neither `tabular` nor `dict`?

RUFAS/input_manager.py
```diff
@@ -789,3 +793,264 @@ def flush_pool(self) -> None:
                     }
         self.__pool = {}
         om.add_log("Clear variable pool", "The pool is emptied.", info_map)
+
+    def _metadata_properties_exists(self, variable_name: str, properties_blob_key: str) -> (bool, KeyError):
```
```suggestion
    def _metadata_properties_exists(self, variable_name: str, properties_blob_key: str) -> (bool, KeyError | None):
```
Directly denoting that the error value might be None would be good because that is one of the expected cases.

RUFAS/input_manager.py
```diff
@@ -789,3 +793,264 @@ def flush_pool(self) -> None:
...
+            return False, KeyError("No metadata loaded.")
+        if properties_blob_key not in self.__metadata["properties"]:
+            om.add_error("No metadata found", f"No metadata is found for variable '{variable_name}' with given "
+                                              f"properties_blob_key {properties_blob_key}.", info_map)
+            return False, KeyError(f"No metadata is found for variable '{variable_name}' with given properties_blob_key"
+                                   f" {properties_blob_key}.")
+        return True, None
+
+    def _add_variable_to_pool(self, variable_name: str, data: Dict[str, Any], properties_blob_key: str,
+                              eager_termination: bool, variable_type: str) -> (bool, ValueError):
```
```suggestion
    def _add_variable_to_pool(self, variable_name: str, data: Dict[str, Any], properties_blob_key: str,
                              eager_termination: bool, variable_type: str) -> (bool, ValueError | None):
```

RUFAS/input_manager.py
```diff
@@ -789,3 +793,264 @@ def flush_pool(self) -> None:
...
+                    input_data=data,
+                    eager_termination=False,
+                    element_counter_and_validity=element_counter_and_validity
+                )
+            elif variable_type == 'tabular':
+                element_counter_and_validity = self._validate_tabular_element(
+                    var_name=metadata_property,
+                    properties_blob_key=properties_blob_key,
+                    input_data=data,
+                    eager_termination=False,
```
Isgotkowitz: Why `eager_termination=False` instead of `eager_termination=eager_termination`?

RUFAS/input_manager.py
```diff
@@ -180,18 +179,21 @@ def _populate_pool(self, eager_termination: bool) -> bool:
...
+                                                                                  properties_blob_key,
+                                                                                  filtered_input_data,
+                                                                                  eager_termination,
+                                                                                  element_counter_and_validity)
 
                 fixed_elements_counter += element_counter_and_validity["fixed_elements"]
                 valid_elements_counter += element_counter_and_validity["valid_elements"]
                 total_elements_counter += element_counter_and_validity["total_elements"]
                 if element_counter_and_validity["is_valid"]:
+                    # ISSUE #1014 !!!
```
```suggestion
```

Isgotkowitz (changes requested):
Looks really good Allister. Changes requested on docstring formatting and making type hints a little more detailed. Also had some questions/suggestions in the code.

RUFAS/input_manager.py
allisterakun: I believe `ValueError` would be more appropriate, I will change it to `ValueError`

> **exception TypeError**
Raised when an operation or function is applied to an object of inappropriate type. The associated value is a string giving details about the type mismatch.
...
Passing arguments of the wrong type (e.g. passing a [list](https://docs.python.org/3/library/stdtypes.html#list) when an [int](https://docs.python.org/3/library/functions.html#int) is expected) should result in a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError), but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).

RUFAS/input_manager.py
allisterakun: This would be better!

RUFAS/input_manager.py
allisterakun: Changing `variable_type: str` to `is_dict: bool`

RUFAS/input_manager.py
allisterakun: Initially, I did this because `eager_termination` would be handled differently during runtime from the data-processing before the simulation. But giving it a second thought and looking at how `_validate_dict_element()` and `_validate_tabular_element()` handle `eagter_termination`, I do believe this (passing `eager_termination` to the validation functions rather than hard-coding it to `False`) would be a better solution.

RUFAS/input_manager.py
```diff
@@ -832,34 +835,36 @@ def _metadata_properties_exists(self, variable_name: str, properties_blob_key: s
...
+            Weather the variable is a dictionary variable (rather than tabular).
+
+        Returns
+        -------
+        tuple
+            A tuple containing a boolean and an exception:
+            - The boolean is True if the variable is successfully added, False otherwise.
+            - ValueError if invalid data is encountered and eager_termination is True,
+              otherwise None.
             """
```
```suggestion

        """
```

tests/test_input_manager.py
Isgotkowitz, line 1: There are a lot of statements in these tests that are not being run, which I think is at the root of why test coverage is going down overall in the README. Of those, I don't know exactly how many are being introduced in this PR, but the tests should be changed so that they are running fully.

RUFAS/input_manager.py
```diff
@@ -794,36 +793,40 @@ def flush_pool(self) -> None:
...
+            The key representing the specific properties blob in the metadata to check.
+
+        Returns
+        -------
+        tuple
+            A tuple containing a boolean and an exception:
+            - The boolean is True if the properties exist, False otherwise.
+            - KeyError if the metadata is not loaded or the properties are not found,
+              otherwise None.
             """
```
```suggestion
        
        """
```

RUFAS/input_manager.py
```diff
@@ -789,3 +792,270 @@ def flush_pool(self) -> None:
...
+
+        if not metadata_properties_exists:
+            raise metadata_error
+
+        add_variable_success, add_variable_value_error = self._add_variable_to_pool(
+            variable_name=variable_name,
+            data=data,
+            properties_blob_key=properties_blob_key,
+            eager_termination=eager_termination,
+            variable_type='dict')
```
Isgotkowitz: Needs to be updated to a bool argument

RUFAS/input_manager.py
```diff
@@ -789,3 +792,270 @@ def flush_pool(self) -> None:
...
+
+        if not metadata_properties_exists:
+            raise metadata_error
+
+        add_variable_success, add_variable_value_error = self._add_variable_to_pool(
+            variable_name=variable_name,
+            data=data,
+            properties_blob_key=properties_blob_key,
+            eager_termination=eager_termination,
+            variable_type='tabular')
```
Isgotkowitz: Needs to be updated to a bool argument

tests/test_input_manager.py
```diff
@@ -1849,3 +1872,707 @@ def test_flush_pool(mock_input_manager: InputManager) -> None:
...
+            assert result is None
+            assert mock_om_add_error.call_count == 1
+            mock_input_manager._metadata_properties_exists.assert_not_called()
+            mock_input_manager._add_variable_to_pool.assert_not_called()
+
+            mock_input_manager.add_dict_variable_to_pool = \
+                input_manager_original_method_states["add_dict_variable_to_pool"]
+            mock_input_manager._metadata_properties_exists = \
+                input_manager_original_method_states["_metadata_properties_exists"]
+            mock_input_manager._add_variable_to_pool = input_manager_original_method_states["_add_variable_to_pool"]
```
Isgotkowitz:
```suggestion
        with pytest.raises(TypeError):
            mock_input_manager.add_dict_variable_to_pool(
                variable_name=variable_name,
                data=data,
                properties_blob_key=properties_blob_key,
                eager_termination=False
            )

        assert mock_om_add_error.call_count == 1
        mock_input_manager._metadata_properties_exists.assert_not_called()
        mock_input_manager._add_variable_to_pool.assert_not_called()

        mock_input_manager.add_dict_variable_to_pool = \
            input_manager_original_method_states["add_dict_variable_to_pool"]
        mock_input_manager._metadata_properties_exists = \
            input_manager_original_method_states["_metadata_properties_exists"]
        mock_input_manager._add_variable_to_pool = input_manager_original_method_states["_add_variable_to_pool"]
```
Nothing is returned when an error is raised

tests/test_input_manager.py
```diff
@@ -1849,3 +1872,707 @@ def test_flush_pool(mock_input_manager: InputManager) -> None:
...
+
+
+@pytest.mark.parametrize('variable_name, data, properties_blob_key', [
+    ("var1", {}, "key1"),
+    ("var2", {"a": 1}, "key2"),
+    ("var3", {"a": "A", "b": 2, "c": True}, "key3"),
+])
+def test_add_dict_variable_to_pool_metadata_error(variable_name: str,
```
Isgotkowitz: Same comment from `test_add_dict_variable_to_pool_type_error`

tests/test_input_manager.py
```diff
@@ -1849,3 +1872,707 @@ def test_flush_pool(mock_input_manager: InputManager) -> None:
...
+                properties_blob_key=properties_blob_key,
+                eager_termination=True,
+                variable_type='dict'
+            )
+
+            mock_input_manager.add_dict_variable_to_pool = \
+                input_manager_original_method_states["add_dict_variable_to_pool"]
+            mock_input_manager._metadata_properties_exists = \
+                input_manager_original_method_states["_metadata_properties_exists"]
+            mock_input_manager._add_variable_to_pool = input_manager_original_method_states["_add_variable_to_pool"]
```
Isgotkowitz: Same comment from `test_add_dict_variable_to_pool_type_error`

tests/test_input_manager.py
```diff
@@ -1849,3 +1872,707 @@ def test_flush_pool(mock_input_manager: InputManager) -> None:
...
+            assert result is None
+            assert mock_om_add_error.call_count == 1
+            mock_input_manager._metadata_properties_exists.assert_not_called()
+            mock_input_manager._add_variable_to_pool.assert_not_called()
+
+            mock_input_manager.add_tabular_variable_to_pool = \
+                input_manager_original_method_states["add_tabular_variable_to_pool"]
+            mock_input_manager._metadata_properties_exists = \
+                input_manager_original_method_states["_metadata_properties_exists"]
+            mock_input_manager._add_variable_to_pool = input_manager_original_method_states["_add_variable_to_pool"]
```
Isgotkowitz: Same comment from `test_add_dict_variable_to_pool_type_error`

tests/test_input_manager.py
```diff
@@ -1849,3 +1872,707 @@ def test_flush_pool(mock_input_manager: InputManager) -> None:
...
+            mock_input_manager._metadata_properties_exists.assert_called_once_with(
+                variable_name=variable_name,
+                properties_blob_key=properties_blob_key)
+            mock_input_manager._add_variable_to_pool.assert_not_called()
+
+            mock_input_manager.add_tabular_variable_to_pool = \
+                input_manager_original_method_states["add_tabular_variable_to_pool"]
+            mock_input_manager._metadata_properties_exists = \
+                input_manager_original_method_states["_metadata_properties_exists"]
+            mock_input_manager._add_variable_to_pool = input_manager_original_method_states["_add_variable_to_pool"]
```
Isgotkowitz: Same comment from `test_add_dict_variable_to_pool_type_error`

tests/test_input_manager.py
```diff
@@ -1849,3 +1872,707 @@ def test_flush_pool(mock_input_manager: InputManager) -> None:
...
+                properties_blob_key=properties_blob_key,
+                eager_termination=True,
+                variable_type='tabular'
+            )
+
+            mock_input_manager.add_tabular_variable_to_pool = \
+                input_manager_original_method_states["add_tabular_variable_to_pool"]
+            mock_input_manager._metadata_properties_exists = \
+                input_manager_original_method_states["_metadata_properties_exists"]
+            mock_input_manager._add_variable_to_pool = input_manager_original_method_states["_add_variable_to_pool"]
```
Isgotkowitz: Same comment from `test_add_dict_variable_to_pool_type_error`

Isgotkowitz (changes requested):
I found that a number of statements in the tests were not being executed, and also the calls to `_add_variable_to_pool()` need to have arguments updated so that the string arguments for `dict` and `tabular` are changed to bools.

tests/test_input_manager.py
allisterakun: I just updated `test_input_manager.py` making sure all codes are run.
image

Isgotkowitz (approved):
LGTM, nice job Allister

RUFAS/input_manager.py
```diff
@@ -789,3 +792,273 @@ def flush_pool(self) -> None:
                     }
         self.__pool = {}
         om.add_log("Clear variable pool", "The pool is emptied.", info_map)
+
+    def _metadata_properties_exists(self, variable_name: str, properties_blob_key: str) -> (bool,
+                                                                                            KeyError | ValueError |
+                                                                                            None):
```
PooyaHekmati: In Python, the standard and recommended practice is to raise exceptions, rather than returning them, for error handling and signaling exceptional situations in code. This approach is aligned with the design and idiomatic use of Python as a language.

1. **Use of `raise` Statement**: The `raise` statement in Python is specifically designed to force a specific exception to occur. When an exception is raised, it terminates the flow of the program, allowing the exception to propagate up the call stack until it is handled. This mechanism is intended for handling errors or other exceptional conditions that disrupt the normal flow of control within a code block [ref1](https://victoria.dev/blog/do-i-raise-or-return-errors-in-python/)  [ref2](https://realpython.com/python-raise-exception/).

2. **Avoiding Silent Failures**: Raising exceptions helps to avoid situations where functions fail silently. For instance, if a function encounters an error condition and only returns an error object instead of raising an exception, the calling code might inadvertently treat this as a normal return value, potentially leading to unnoticed errors or undefined behaviors [ref](https://victoria.dev/blog/do-i-raise-or-return-errors-in-python/).

3. **Error Propagation**: In Python, when an exception is raised in a function and not handled within that function, it propagates up to the caller. If it remains unhandled, it continues to propagate until it reaches the main program, potentially halting the program with an exception traceback. This propagation mechanism is an integral part of Python's error-handling approach, ensuring that exceptions are either properly handled or result in a clear indication of an unhandled error [ref](https://realpython.com/python-raise-exception/#:~:text=If%20your%20code%20raises%20an,halts%20with%20an%20exception%20traceback).

4. **Python's Philosophy**: Python's design encourages the use of exceptions for error handling. This contrasts with some other programming languages, like C or Go, which might encourage returning error codes. The idiomatic use of exceptions in Python includes the use of `try...except` constructs for handling these exceptions, a common approach seen throughout Python's standard library and the language itself [ref1](https://realpython.com/python-raise-exception/) [ref2](https://realpython.com/python-raise-exception/#:~:text=In%20contrast%2C%20Python%20encourages%20you,have%20many%20examples%20of%20it)

Based on these principles, it would be advisable to change the code to raise exceptions instead of returning them. This change would align the code with Python's best practices, improve its readability and maintainability, and ensure that error conditions are handled appropriately and consistently.

**This applies to all other cases as well. The docstrings and test cases have to be updated accordingly as well.**

RUFAS/input_manager.py
```diff
@@ -789,3 +792,273 @@ def flush_pool(self) -> None:
                     }
         self.__pool = {}
         om.add_log("Clear variable pool", "The pool is emptied.", info_map)
+
+    def _metadata_properties_exists(self, variable_name: str, properties_blob_key: str) -> (bool,
```
PooyaHekmati: either `_metadata_properties_exist` or `_metadata_property_exists`

RUFAS/input_manager.py
```diff
@@ -789,3 +792,273 @@ def flush_pool(self) -> None:
...
+              otherwise None.
+        """
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._metadata_properties_exists.__name__,
+                    }
+        if not self.__metadata:
+            om.add_error("No metadata loaded", "No metadata is loaded to the InputManager.", info_map)
+            return False, ValueError("No metadata loaded.")
+        if properties_blob_key not in self.__metadata["properties"]:
+            om.add_error("No metadata found", f"No metadata is found for variable '{variable_name}' with given "
```
PooyaHekmati, line 833: why is this an error?

RUFAS/input_manager.py
```diff
@@ -789,3 +792,273 @@ def flush_pool(self) -> None:
...
+        ----------
+        variable_name : str
+            The name of the variable to be added to the pool.
+        data : Dict[str, Any]
+            The data associated with the variable that needs validation and addition to the pool.
+        properties_blob_key : str
+            The key in the metadata properties against which the data is validated.
+        eager_termination : bool
+            Flag indicating whether the function should return early in case of invalid data.
+        is_dict_variable : bool
```
```suggestion
        is_variable_dict : bool
```
We don't say `Is developer Allister`, we say `is Allister a developer`

PooyaHekmati (changes requested):
I hate to return the PR again, but there are changes needed.

RUFAS/input_manager.py
allisterakun: Because if there were no metadata properties found with the provided `properties_blob_key`, then there's no way we can validate the data.

RUFAS/input_manager.py
allisterakun: Done!

RUFAS/input_manager.py
```diff
@@ -789,3 +792,256 @@ def flush_pool(self) -> None:
...
+        else:
+            return True
+
+    def add_dict_variable_to_pool(self, variable_name: str, data: Dict[str, Any], properties_blob_key: str,
+                                  eager_termination: bool) -> bool:
+        """
+        Adds a dictionary variable to the InputManager's pool after validating it against metadata.
+
+        This function takes in a variable along with its name and a key to access its validation metadata.
+        It validates the data against the provided metadata and adds the data to the InputManager pool if it is valid.
```
Isgotkowitz, line 949: Added to notes section

RUFAS/input_manager.py
```diff
@@ -789,3 +792,256 @@ def flush_pool(self) -> None:
...
+                is_variable_dict=True)
+            return add_variable_success
+
+    def add_tabular_variable_to_pool(self, variable_name: str, data: Dict[str, List[Any]] | List[Any],
+                                     properties_blob_key: str, eager_termination: bool) -> bool:
+        """
+        Adds a tabular variable to the InputManager's pool after validating it against metadata.
+
+        This function takes in a variable along with its name and a key to access its validation metadata.
+        It validates the data against the provided metadata and adds the data to the InputManager pool if it is valid.
```
Isgotkowitz, line 1004: Add to notes section

RUFAS/input_manager.py
```diff
@@ -789,3 +792,256 @@ def flush_pool(self) -> None:
...
+        ----------
+        variable_name: str
+            The name of the dictionary variable to be added.
+        data : Dict[str, Any]
+            The data of the variable, structured as a dictionary.
+        properties_blob_key : str
+            A key used to locate the metadata for validation of the variable.
+        eager_termination : bool
+            If True, raises a RuntimeError when the variable is invalid.
+            If False, the function returns False without raising an error.
```
Isgotkowitz: Needs to be updated  - runtime errors aren't raised anymore.

RUFAS/input_manager.py
```diff
@@ -789,3 +792,256 @@ def flush_pool(self) -> None:
...
+        ----------
+        variable_name: str
+            The name of the variable to be added.
+        data : Dict[str, List[Any]] | List[Any]
+            The data of the tabular variable, structured as a dictionary of lists or a list.
+        properties_blob_key : str
+            A key used to locate the metadata for validation of the variable.
+        eager_termination : bool
+            If True, raises a ValueError when the variable is invalid.
+            If False, the function returns False without raising an error.
```
Isgotkowitz: I think either this docstring or the method behavior needs to be changed. Right now, the method doesn't raise a ValueError, it just doesn't catch the `ValueError` raised by `_add_variable_to_pool()` raises one.

Isgotkowitz (approved):
Approving, but there are some minor changes that need to be made to docstrings.

Nice job Allister

PooyaHekmati (approved):
[LGTM](https://lgtm.lol/i/125)

PR #987, opened 2023-11-22
Barnyard machine phoshorus dynamics evaluation

Isgotkowitz:


Isgotkowitz:
Does not need to be merged into `barnyard_phos_eval` branch. Only has changes to inputs necessary for testing.

PR #988, opened 2023-11-22
Adding ElementsCounter class to IM and refactoring helper methods for primitive types

ltp23:
Context:
I have a fairly big, mostly complete PR (#886) that needs to be broken down into smaller PRs. The current state of the input manager is that it can validate primitive data types including boolean, numeric, and string but it can't validate more complex, recursive types like objects and arrays. Also, when we replace faulty data with default data, we should try to revalidate the default data that is being used. 

Content:
In this first PR which is more about refactoring, I added helper methods to validate the three primitive data types: boolean, numeric, and string. Another new addition is a separate `ElementsCounter` class to help keep track of the number of elements the input manager processes. 

Two other standalone helper methods: 
- `_get_nested_dict_value()`: to replace the use of the `reduce()` later on and to generalize to extract nested pieces of data from both lists and dictionaries.
- `_convert_variable_path_to_str`: to convert the paths of input data trails that lead to errors into a string format for the output manager.

If interested, you can refer to the previously mentioned PR #886 to see how things are connected together in the end.

### DIFF
--- a/RUFAS/input_manager.py
+++ b/RUFAS/input_manager.py
@@ -884,6 +884,99 @@ class InputManager:
             return False, "Value greater than maximum"
         return True, ""
 
+    @staticmethod
+    def _is_str_value(variable_value: Any, variable_properties: Dict[str, Any]) -> Tuple[bool, str]:
+        """
+        Check if the variable value is a string.
+
+        Parameters
+        ----------
+        variable_value : Any
+            The value of the variable to check.
+        variable_properties : Dict[str, Any]
+            The properties for the variable of interest.
+
+        Returns
+        -------
+        Tuple[bool, str]
+            A tuple containing a boolean indicating if the check passed and a string containing the reason for failure.
+
+        """
+
+        if type(variable_value) is not str:
+            return False, "String variable is not a string."
+        return True, ""
+
+    @staticmethod
+    def _check_str_len_lower_bound(variable_value: str,
+                                   variable_properties: Dict[str, Any]) -> Tuple[bool, str]:
+        """
+        Check if the variable value is greater than the minimum length.
+
+        Parameters
+        ----------
+        variable_value : str
+            The value of the variable to check.
+        variable_properties : Dict[str, Any]
+            The properties for the variable of interest.
+
+        Returns
+        -------
+        Tuple[bool, str]
+            A tuple containing a boolean indicating if the check passed and a string containing the reason for failure.
+        """
+
+        if "minimum_length" in variable_properties and len(variable_value) < variable_properties["minimum_length"]:
+            return False, "String length less than minimum."
+        return True, ""
+
+    @staticmethod
+    def _check_str_len_upper_bound(variable_value: str,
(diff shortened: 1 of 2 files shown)


RUFAS/input_manager.py
```diff
@@ -789,3 +793,380 @@ def flush_pool(self) -> None:
...
+        value : int, optional
+            The amount by which to decrement the attribute (default is 1).
+
+        Raises
+        ------
+        Exception
+            If the given name is not an attribute of ElementsCounter.
+        """
+
+        self.update(name, getattr(self, name) - value)
```
Isgotkowitz, line 1141: Instead of having one method each for incrementing and decrementing all the different counters, would it be better to have the an increment and decrement method for each pool? It would mean writing a lot more methods, but may be worth the trade-off because it eliminates the risk of getting an `AttributeError` when `getattr()` is called.

Also for both the increment and decrement method, I think it would be good to validate the value that is being used to increment/decrement the counter.

RUFAS/input_manager.py
```diff
@@ -789,3 +793,380 @@ def flush_pool(self) -> None:
...
+        The count of all elements processed.
+    valid_elements : int
+        The count of all valid elements processed.
+    fixed_elements : int
+        The count of all elements that were fixed during processing.
+    invalid_elements : int
+        The count of all elements found to be invalid during processing.
+
+    Methods
+    -------
```
Isgotkowitz, line 1073: `decrement()` and `reset()` should be added here.

RUFAS/input_manager.py
```diff
@@ -789,3 +793,380 @@ def flush_pool(self) -> None:
...
+        Returns
+        -------
+        Tuple[bool, str]
+              A tuple containing a boolean indicating if the check passed and a string containing the reason for
+              failure.
+        """
+
+        if "pattern" in variable_properties and not re.fullmatch(variable_properties["pattern"], variable_value):
+            return False, "String does not match pattern."
+        return True, ""
```
Isgotkowitz, line 1054: Are you planning to refactor so that these helper methods replace the logic contained in `_num_type_validator()`, `_string_type_validator()` and `_bool_type_validator()`?

Isgotkowitz (commented):
Code is very clean, nice job Loi. Commenting right now - I don't fully understand how the string, bool, and number validation helper methods fit in with the refactor as described in #886. Once I get a better handle on it I will re-review.

RUFAS/input_manager.py
```diff
@@ -789,3 +793,380 @@ def flush_pool(self) -> None:
...
+        """
+        Reset all counts to zero.
+
+        Returns
+        -------
+        None
+        """
+
+        self.total_elements = 0
+        self.valid_elements = 0
```
ew3361zh, line 1153: I'm curious where you landed with how you're counting nested elements - I assume each nested element is counting towards the total elements number but is the outermost element being counted?

I'm not sure it matters in the end since we don't have a set number of elements we're expecting to have but the total_elements should equal te valid + invalid elements at the very least.

PR #993, opened 2023-11-27
Make supplementing manure application nutrient deficiencies optional

### LINKED ISSUE #949: Make automatically backfilling manure application nutrient deficiencies an optional feature.
The current behavior of the Crop and Soil module is to take in the desired amounts of nitrogen and phosphorus to be contained in a manure application, request those nutrients from the Manure module, then supplement any deficient nutrients with chemical fertilizers. This practice is a huge assumption to be making for farmers, so it should be made an optional one.

Add a boolean input to the field input file that indicates whether the user would like manure applications to be automatically supplemented with chemical fertilizer for applications on that field.

- Do not have the feature of automatic nutrients supplementation with chemical fertilizer at all.
- Always have the feature turned on.

**Additional context**

Isgotkowitz:
This PR makes the supplementation of manure nutrient deficiencies optional, and lets the user specify if they want these supplementations to occur or not.

_What does this mean?_ When users specify a manure application, its size is determined by two parameters: the minimum amounts of nitrogen and phosphorus they want in it. The Crop and Soil module passes these amounts to the Manure module, which then returns an amount of manure that either satisfies the requested nutrient amounts or gets as close to meeting them as possible. If the requested amounts of nitrogen and/or phosphorus were not met in the manure application, then RuFaS would automatically create the most efficient chemical fertilizer application to meet the unmet nutrient demands. This is made an optional feature by this PR.

Changes:
- `field_data.py` now contains the attribute `supplement_manure_nutrient_deficiencies`, which indicates whether or not automatic nutrient supplementation should occur. Also it reformats the documentation and removes attributes not being used.
- `field.py` only adds chemical fertilizer to deficient manure nutrient amounts when allowed to do so by the user. If there is a deficient manure application and it can't supplement it with chemical fertilizer, it raises a warning to the OutputManager.
- `field_manager.py` sets `supplement_manure_nutrient_deficiencies` during the initialization of fields.
- All field and metadata files have been adjusted for the change in input. The `nutrient_standard` property in `ARL_metadata.json` was not the same as the other metadata files, so that has been updated as well.

Test coverage of all files touched remains at 100%.

Closes #949


RUFAS/routines/field/field/field.py
```diff
@@ -452,9 +457,20 @@ def _execute_manure_application(self, requested_nitrogen: float, requested_phosp
...
         unmet_nitrogen_demand = max(0.0, requested_nitrogen - supplied_nitrogen)
         unmet_phosphorus_demand = max(0.0, requested_phosphorus - supplied_phosphorus)
+
         if unmet_nitrogen_demand == 0.0 and unmet_phosphorus_demand == 0.0:
             return
-        elif unmet_nitrogen_demand > 0.0 and unmet_phosphorus_demand == 0.0:
+
+        if not self.field_data.supplement_manure_nutrient_deficiencies:
+            warning_name = "nutrient_deficient_manure_application"
+            warning_message = f"Manure nitrogen deficient by {unmet_nitrogen_demand} kg, manure phosphorus " \
```
ew3361zh, line 466: Is there any benefit to breaking this message up into 2 separate potential warnings (i.e. one for if nitrogen demand is unmet and one for phosphorus demand being unmet)? Or is it more helpful to keep them paired?

RUFAS/routines/field/field/field_data.py
```diff
@@ -1,73 +1,87 @@
...
+        The longitude value of where the field is located (degrees).
+    minimum_daylength : float, default=6.33
+        Shortest day of the year for this watershed (hours).
+    dormancy_threshold_daylength : float, optional
+        Threshold daylength to initiate dormancy in a plant (hours).
+    current_residue : float, default=0.0
+        Total amount of residue on the current day (kg per hectare).
+    transpiration : float, default=0.0
+        Total amount of water lost to transpiration in the field on the current day (mm).
+    max_transpiration : float, default=0.0
```
ew3361zh, line 28: Curious about the distinguising guideline as to why some of these vars are `xxx_in_mm` and others don't have the unit as a part of the name.

input/metadata/ARL_metadata.json
```diff
@@ -153,7 +153,7 @@
       "nutrient_standard": {
         "type": "string",
         "default": "NASEM",
-        "pattern": "(NASEM|NRC)",
+        "pattern": "^(NASEM|NRC)$",
```
ew3361zh, line 156: I really thought we'd fixed all of these.

ew3361zh (approved):
Looks good and works well for me. Nice job! A couple comments about some choices made but overall seems good to go.

RUFAS/routines/field/field/field.py
Isgotkowitz: From a code cleanliness standpoint I think it is more helpful to keep them paired, because there is only one error message to be raised instead of checking which of the three permutations (only N unmet, only P unmet, both N and P unmet) it falls under.

From a user perspective I can see it going both ways: if only phosphorus is deficient then it makes sense to only include information in the warning about phosphorus, because that is the only behavior that is causing the warning. But explicitly stating that the nitrogen deficiency is 0 leaves no room for the user to look at the error message and wonder if the nitrogen demand was met.

RUFAS/routines/field/field/field_data.py
Isgotkowitz: For Crop and Soil, we have so far opted to be really explicit about units in docstrings and not include it variable names (for the most part), but there is not any stated guideline we are following as far as I know. Developing a RuFaS-wide variable naming standard would be a very worthwhile use of time though, because it would be great to have the variables here match the style and verbosity of other modules.

input/metadata/ARL_metadata.json
Isgotkowitz: Ik, they just keep popping up.

RUFAS/routines/field/field/field.py
```diff
@@ -452,9 +457,20 @@ def _execute_manure_application(self, requested_nitrogen: float, requested_phosp
...
+                              f"deficient by {unmet_phosphorus_demand} kg."
+            info_map = {"class": self.__class__.__name__, "function": self._execute_manure_application.__name__,
+                        "prefix": f"field='{self.field_data.name}'", "year": year, "day": day}
```
allisterakun: maybe move this to the top of the function? The same variable is defined on line 398 also. Could be reused.

allisterakun (approved):

PR #994, opened 2023-11-27
Updating nitrous oxide calculations for manure treatments

### LINKED ISSUE #952: Review the reporting of nitrous oxide from the manure module
Tasks:
- [x] Compost bedded pack barn N2O calculations
- [x] Open lots N2O calculations
- [x] Add those outputs to the OM at the pen level by default.

ltp23:
In this PR, I added more code for the daily update calculations for the two manure treatments - compost bedded pack barn (scenario id 8) and open lots (scenario id 9). 

So, to see the effects of these changes, you can change the manure management scenarios in the animal JSON files to 8 or 9. Currently, only those two manure treatment options have nitrous oxide data. 

To get the specific nitrous oxide data, you can add the expression `.*storage_nitrous_oxide.*` in a filter file. Currently, the data are reported at the pen level by default.


tomhuhh (commented):
LGTM! Thanks Loi! I ran a 5-year simulation with manure management scenario set to 8 for calves and 9 for other animal categories. There are definitely nitrous oxide outputs, but I will need some literature reference to tell if those numbers make biological sense. Will get back to you on that! Thanks again! 

image
 
P.S. We should probably tag a SDE on this PR as well.

tomhuhh:
My first question is that are we sure the results are on the pen level? Looking at the example that is given in the design doc, the estimate for nitrous oxide is 0.03 kg (I assume this is on the individual animal level) when using `open lots` as the manure treatment method. Assuming 85 lactating cows (100 cows in the herd) in pen 3 , that should give us 2.55 kg nitrous oxide in total. However, the output from pen 3 here is only ~30% of what it should be. Thanks!

ltp23:
> My first question is that are we sure the results are on the pen level? Looking at the example that is given in the design doc, the estimate for nitrous oxide is 0.03 kg (I assume this is on the individual animal level) when using `open lots` as the manure treatment method. Assuming 85 lactating cows (100 cows in the herd) in pen 3 , that should give us 2.55 kg nitrous oxide in total. However, the output from pen 3 here is only ~30% of what it should be. Thanks!

The way nitrous oxide is calculated is based on a fixed percentage of the amount of nitrogen in the animal manure. Specifically, the percentages are as follows

Compost bedded pack barn with tillage: 7%
Compost bedded pack barn without tillage: 1%
Open lots: 2%

In the example graph below, when the manure nitrogen is around 50kg, the amount of nitrous oxide is about 1 kg, or 2%. Though the scaling is kinda hard to see here. 

default_scenario_nitrous-oxide-emissions-29-Nov-2023_Wed_13-15-53

As a side note, while debugging, I noticed that the animal manure variables are still negative sometimes and the order of the colors and the legends on the graph can mismatch if not careful. Both of these issues will need to be addressed separately.

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -1594,19 +1591,77 @@ def total_nitrogen_loss_from_compost_bedded_pack_barn(
...
+        daily_nitrogen_input : float
+            The amount of nitrogen present in the manure excreted by animals (kg).
+
+        Returns
+        -------
+        float
+            The nitrogen lost to nitrous oxide emission (kg).
+
+        """
+        return 2e-2 * daily_nitrogen_input
```
Isgotkowitz: Do the constants used in these equations need references to literature or external documentation?

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
```diff
@@ -222,6 +223,17 @@ class Tillage(BaseManureHandler):
     pass
 
 
+class Harrowing(BaseManureHandler):
+    """A class that handles calculations related to harrowing.
+
+    Attributes:
+        All inherited from BaseManureHandler.
```
Isgotkowitz:
```suggestion
    Attributes
    ----------
    All inherited from BaseManureHandler.
```
Is there a plan to add harrowing-specific attributes at some point?

RUFAS/routines/manure/manure_manager.py
```diff
@@ -306,9 +308,10 @@ def _add_manure_nutrients(
         Returns
         -------
         None
-
         """
-        # TODO: With the introduction of different manure types, we should rename attributes in ManureTreatmentDailyOutput
+
+        # TODO: With the introduction of different manure types, we should
+        # rename attributes in ManureTreatmentDailyOutput
```
Isgotkowitz: Is there an issue or timeline associated with this TODO? If so, it should either be added in here or entirely documented in the issue and the TODO removed from the code.

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
```diff
@@ -22,9 +22,40 @@ class ManureTreatmentDailyOutput(LiquidManurePortionProtocol):
         liquid_manure_phosphorus: Amount of phosphorus excreted in manure, kg.
         liquid_manure_potassium: Amount of potassium in manure, kg.
         daily_final_manure_volume: Final manure volume after treatment, m^3.
-
-        # TODO: Document remaining attributes.
-
```
Isgotkowitz, line 27: Nice

RUFAS/routines/manure/manure_treatments/open_lots.py
```diff
@@ -0,0 +1,196 @@
...
+)
+from RUFAS.time import Time
+from RUFAS.weather import Weather
+
+
+class OpenLots(BaseManureTreatment):
+    """Class for the open lots manure treatment system.
+
+    Attributes:
+        All attributes inherited from BaseManureTreatment.
```
Isgotkowitz:
```suggestion
    Attributes
    ----------
    All attributes inherited from BaseManureTreatment.
```

RUFAS/routines/manure/manure_treatments/open_lots.py
```diff
@@ -0,0 +1,196 @@
...
+        manure_total_solids : float
+            The total mass of the manure (kg).
+        bedding_total_solids : float
+            The mass of the bedding material (kg).
+        manure_volatile_solids : float
+            The mass of manure volatile solids (kg).
+        days_since_last_tillage : float
+            The number of days since the manure-bedding mixture was last tilled.
+        lag : int
+            The lag time
```
Isgotkowitz: Is this time measured in days, hours, something else?

input/data/manure/manure_management.json
```diff
@@ -62,6 +62,13 @@
       "manure_handler": "tillage",
       "manure_separator": "none",
       "manure_treatment": "compost bedded pack barn"
+    },
+    {
+      "scenario_id": 9,
+      "bedding_type": "none",
+      "manure_handler": "harrowing",
+      "manure_separator": "none",
+      "manure_treatment": "open lots"
```
Isgotkowitz, line 71: Do the manure metadata properties need to be updated to accept this new scenario?

Isgotkowitz (approved):
Code looks very clean and test coverage is increased for the manure module overall, which is great. Couple of comments, related to formatting and possible future work needed.

It seems that this PR has more than one responsibility, and its large size makes it difficult to very thoroughly review. In the future I think it would be beneficial to split up changes of this scope.

tomhuhh:
> I noticed that the animal manure variables are still negative

I thought we applied a temporary fix to update all the negative values in animal manure outputs to 0 when it happens. I could be mistaken. I can look into this issue more next week. Thanks!

tomhuhh:
> In the example graph below, when the manure nitrogen is around 50kg, the amount of nitrous oxide is about 1 kg, or 2%.

I agree with the findings of this example. The code works as expected. However, I noticed that there might be a typo in the design document of the `'Open Lots' `methods. Please see below. I will send an email to Varma to confirm the value.   

Screenshot 2023-12-01 at 2 55 42 PM

tomhuhh (approved):

RUFAS/routines/manure/gas_emissions/calculator.py
ltp23: I will add those hard-coded values to one of the manure module constants files.

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
ltp23: Currently, all the manure handler classes don't seem to have specific attributes associated with them. They inherit from the same base manure handler class. So, I'm thinking during a future refactoring, these separate manure handler subclasses could be consolidated and removed.

RUFAS/routines/manure/manure_manager.py
ltp23: I will remove this TODO. My current solution for this is to combine the manure nutrients from all three types of manure - liquid, slurry, and solid.

RUFAS/routines/manure/manure_treatments/open_lots.py
ltp23: This one is measured in days. I'll update the docstring.

input/data/manure/manure_management.json
ltp23: Yes, I will update this new scenario in the default metadata file.

jadamchick (commented):
Quick review of the nitrous oxide outputs look good! Thanks for this Loi

Issue #905, opened 2023-11-03
Feed Library Metadata ranges are too restrictive


KFosterReed:
There are a few wonky things in the feed metadata validation: 

1. Some of the upper and lower bounds for the feed composition values in the metadata are too restrictive. 
2. For PAF the default value (0) is outside of the min (0.87) and max (1.04) range. 
3. The boolean values  `is_wetforage` and `is_fat`  are listed as a number type 

## Expected Behavior

The values in the current feed composition libraries shouldn't require input validation warnings and fixes

## Possible Solution

1.  For all composition values that have a unit of %DM or % of something else we should have 0 as a minimum and 100 as the maximum.
2. Have 0 be the minimum and 1.04 be the maximum
3. Not sure what the correct fix is here. Do we need to change the values in the inputs to true and false? Or is there a way to check if they are only 0 or only 1?
4. 
## Steps to Reproduce

1. Run the default or user_defined_ration metadata on main
5.
6.
7.

allisterakun:
Will open a PR and update the metadata!

Ed731551:
It seems to me that some variables of % of something else are out of expected range. Perhaps additional constraints may help?

KFosterReed:
@Ed731551 could you please be more specific? Which variables and what are you considering the 'expected range'? The purpose of the metadata validation is not to enforce that the parameters stay within 'normal' boundaries but rather that they stay within **feasible** boundaries. Because RuFaS is meant to be a research tool, we want to allow users (scientists especially) to test theories about how different inputs might affect the simulation. Even if they are not realistic in terms of what is commonly fed to cows, we would still want someone to create a new feed that is, for example, 100% NDF. This is why I would like the bounds of all variables that represent a percentage of something to be between 0 and 100.

Ed731551:
Sorry, I recognize that I was speculating about that. This requires a more careful revision. At this stage I do not have a good answer for this.

allisterakun:
The last issue (The boolean values `is_wetforage` and `is_fat` are listed as a number type) will be addressed in PR #913. It is still a work in progress as I need to ensure all references to `is_wetforage` and `is_fat` in the codebase are updated to expect a Boolean value.

As for the first two issues, @elle-andreen mentioned that she would address them in her metadata revisions in PR #722 .

Isgotkowitz:
[...]

PR #995, opened 2023-11-27
Update NASEM_Comp_with_TDN.csv

tomhuhh:
(1) Implemented the TDN & DE values calculated by Edward for whey feeds; 
(2) Utilize the TDN & DE values associated with dry citrus pulp for wet citrus pulp.


tomhuhh:
Also, I have reordered the table so it is numerical order in terms of feed IDs.

KFosterReed (approved):
After filling in a few missing composition values, the new whey feeds work and I think these values are defensible from a nutrition perspective since they are the same as similar feeds

Isgotkowitz (approved):
LGTM. Default scenario runs on my machine.

Since you are making changes to `NASEM_Comp_with_TDN.csv`, would this be a good time to change the invalid values in the file? There are two ADICP and three NDICP invalid values, and one blank WSC.

KFosterReed:
> LGTM. Default scenario runs on my machine.
> 
> Since you are making changes to `NASEM_Comp_with_TDN.csv`, would this be a good time to change the invalid values in the file? There are two ADICP and three NDICP invalid values, and one blank WSC.

For the invalid values, I made this issue [https://github.com/RuminantFarmSystems/MASM/issues/905](url) because it is the metadata ranges that are too restrictive. We should update the WSC value though to 0. @tomhuhh can you please do that before merging? Thanks!

PR #999, opened 2023-11-28
Add ordering to horizontal aggregation

PooyaHekmati:
Ordering is important for some aggregations such as division and subtraction. Since the output manager uses dict as the underlying data structure, it is not possible to maintain ordering there. Also, since the filtering is done using regex, ordering is not possible there either. Therefore, this PR aims to address the gap by introducing a new entry in the report filter files called `horizontal_order`. This new entry is NOT regex as it defeats the purpose. It has to be an exact match with the variables pool's keys.
I tested it with:
```json
{
  "multiple": [
    {
      "name": "Last year's daily % in Parity 1A",
      "filters": [
        "LifeCycleManager.daily_update.(num_cow_for_parity_1|milking_cow_num)"
      ],
      "horizontal_order": [
        "LifeCycleManager.daily_update.num_cow_for_parity_1",
        "LifeCycleManager.daily_update.milking_cow_num"
      ],
      "slice_start": -365
    },
    {
      "name": "Last year's daily % in Parity 1B",
      "filters": [
        "LifeCycleManager.daily_update.(num_cow_for_parity_1|milking_cow_num)"
      ],
      "horizontal_order": [
        "LifeCycleManager.daily_update.milking_cow_num",
        "LifeCycleManager.daily_update.num_cow_for_parity_1"
      ],
      "slice_start": -365
    }
  ]
}
```
The wiki has been updated accordingly.
The PR also adds the usage fixture in some other tests and makes some minor improvements in the report generator's code.


RUFAS/report_generator.py
```diff
@@ -182,34 +185,35 @@ def generate_report(
...
-        elif vertical_aggregator:
-            return [
+
+        if vertical_aggregator:
+            vertically_aggregated = [
                 vertical_aggregator(data_series)
                 for _, data_series in report_data.items()
             ]
 
+        if horizontal_aggregator and vertical_aggregator:
```
Fjordrunner, line 206: We could in theory move this if statement in the first one and then avoid having an addition if statement with an and clause, but that would result in a more nested first if statement, so don't change anything unless it would be explicitly preferred.

Fjordrunner (approved):
Looks good to me! I added a singular comment about the structure of an if statement but it is an extreme nitpick.

Isgotkowitz (approved):
LG8TM. Example works well, and error messages are very clear when something goes wrong.

jadamchick:
I hate to rain on the parade, but I'm getting identical values out of each filter (and my expectation is that one would be the reciprocal of the other). Maybe I'm misunderstanding what the output should be? 

No errors given related to report generation. 

The values (for both) are what I would expect out of Option B (num_cow_for_parity_1 / milking_cow_num). 

Screenshot of what I get when I open the report: 
image

jadamchick:
I hadn't updated my report_generation branch before running the new code.. Looks good now! I'll update the wiki with a relevant example.

jadamchick (commented):
Works great for me!

PooyaHekmati:
> I hadn't updated my report_generation branch before running the new code.. Looks good now! I'll update the wiki with a relevant example.

Thanks

Issue #989, opened 2023-11-22
Graph Generator erroring when stackplot option selected and filter is not a dictionary


ew3361zh:
## Expected Behavior
When the user selects the stackplot option, they should be able to use filters that are not dictionaries.

## Current Behavior
When the user selects the "stackplot" option and provides a filter that has multiple variables eg:

```
{
      "type": "stackplot",
      "filters": ["Cow\\.milking_update\\.milk_(lactose|protein|fat)"],
      "variables": ["milk_lactose", "milk_protein", "milk_fat"],
      "title": "Daily Milk Nutritions",
      "legend": ["lactose", "fat", "protein"]
}
```

Graph generator can't handle the filter and errors with this message:

"graph generation exception. need at least one array to concatenate"

## Possible Solution
After review with @PooyaHekmati, the source of the issue is with how matplotlib handles different types of graphs and the logic that can address this (short-term, at least) is in this highlighted portion of graph_generator.py > _draw_graph():

<img width="769" alt="Screenshot 2023-11-22 at 1 15 44 PM" src="https://github.com/RuminantFarmSystems/MASM/assets/70217952/1edfaaa2-0dbc-4e2a-b14d-cd9c529896a6">

## Steps to Reproduce
Try the above filter and run `python main.py -v = errors` to see the error printed in the console.

ew3361zh:
Closed with #1000

Issue #927, opened 2023-11-09
Improve error messaging when invalid graph filters passed


Isgotkowitz:
**Is your feature request related to a problem? Please describe.**
When graph filter files are incorrect, errors are raised and recorded but are not always very clear or specific when saying why the error occurred.

**Describe the solution you'd like**
When there is an issue in a graph filter file found, the error message should be as clear as possible about what cause the error.

**Describe alternatives you've considered**
Leave the error/warning messages as they are.

**Additional context**
Example:
Below is the body of a misformatted filter file titled "graph_filter_example.json" (the first graph has the key `types` instead of `type`)
```json
{
  "multiple": [
    {
      "types": "plot",
      "filters": [
        "Pen\\.calc_total_manure\\.daily_aggregate_calf_(degradable|non_degradable)_volatile_solids"
      ],
      "title": "Daily Aggregate Calf Degradable Volatile Solids",
      "legend": ["Degradable Volatile Solids", "Non-degradable Volatile Solids"]
    },
    {
      "type": "stackplot",
      "filters": ["Cow\\.milking_update\\.milk_data_at_milk_update"],
      "variables": ["milk_lactose", "milk_protein", "milk_fat"],
      "title": "Daily Milk Nutritions",
      "legend": ["lactose", "fat", "protein"]
    }
  ]
}
```
The error message generated when parsing this file is 
```json
"OutputManager._route_save_functions.graph generation exception": {
    "info_maps": [
        {
            "timestamp": "09-Nov-2023_Thu_13-58-57.381800"
        }
    ],
    "values": [
        "'type'"
    ]
}
```
It would be more helpful if it had a message like the following:
```json
"OutputManager._route_save_functions.graph generation exception": {
    "info_maps": [
        {
            "timestamp": "09-Nov-2023_Thu_13-58-57.381800"
        }
    ],
    "values": [
        "Key 'type' missing for graph 'Daily Aggregate Calf Degradable Volatile Solids' in file graph_filter_example.json."
    ]
}
```

ltp23:
@Isgotkowitz will take another look at this git issue.

Isgotkowitz:
11/15 Update: still a relevant issue. The above example is still holds. IMO this is a low priority issue though.

ew3361zh:
Am I correct in summing this up to be more about giving info about improperly formatting the filter file than about sending bad filters (i.e. this is about someone having good filters that are just not properly formatted but not about bad filters that are properly formatted)?

Issue #899, opened 2023-11-02
Graph generator features: handling dictionaries automatically


JoeWaddell:
Current (as of PR #779) dictionaries require information in `filters`, and also a separate list of inputs in the `variables` section of the graph filter.

## requested feature ##
If no `variables` are defined, graph generator will attempt to plot all information in the dictionary. This will be very useful for outputs that are dynamic, e.g. `ration_per_animal_for_pen_3`, as the keys in this dict will change with changes to the Feed input JSON, so this would be frustrating to juggle so many unique filters for Pilot Testers evaluating different farms.

A warning should be logged if the dict is ragged, or it's otherwise impossible to plot data in the dict (e.g. non-numeric data). Perhaps it should plot numeric data, and simply skip non-numeric data in the same dict. 

### related improvements ###
While these may be worth their own PR, I think it's relevant for automatic graph generation from dicts.

1) Have `legend` also generated automatically, from variable names in a dict. The reasoning is the same as the problem that arises from dynamic names (only in some dictionaries).

2) Adding functionality for a `variables_excluded` list. Simply to reduce the length of lists for dicts that are large, in cases where all but a handful of variables are needed.

ltp23:
@PooyaHekmati will reach out to @JoeWaddell regarding the details of this git issue.

PooyaHekmati:
I talked with Joe about this. This feature can be implemented by changing how GG handles keys in dictionaries. Right now it matches the keys by string comparison, in other words, the user has to know and provide the exact key in the dict, which is not always possible. To fix this, we can convert this to regex matching. Also, we can add filter by exclusion, similar to OM filtering mechanism.

PooyaHekmati:
This applies to Report Generator filters as well. Consider making/extracting the function such that it can be used there as well (shouldn't be a problem).

ew3361zh:
Closed by PR #1048

PR #1000, opened 2023-11-28
Graph generator stackplot fix

ew3361zh:
PR to address issue #989 in which graph generator errors when trying to plot a stackplot where the filter is not a dictionary.

To accomplish this:
1. Add functionality within the `_draw_graph()` to handle and direct stackplot data that is not in a dictionary.
2. Rename and update `_handle_type_based_plot()` -> `_handle_dict_plots()`
3. Move repeated functionality from `_draw_graph()` to `_handle_dict_plots()`
4. Updates example graph filter examples with new functionality (and with changes that are in main now which broke the example plot function filter as well).
5. Updates unit testing.


Isgotkowitz (approved):
Code LGTM. I was able to generate some novel stackplots with data from different variables and the example graphs were producing as expected.

ew3361zh:
Updates:
1. Creates new function in Output Manager `_prepare_plot_data()` that readies the data to be graphed before it's sent to Graph Generator.
2. Simplifies the logic in Graph Generator's `_draw_graph()` now that it is no longer manipulating the data that's sent to it.
3. Updates unit testing and docstrings.

Isgotkowitz (commented):
@ew3361zh I think I have identified an edge case that is causing a stack plot to be produced incorrectly. It might be more of an issue with the variable names though.

```json
    {
      "type": "stackplot",
      "filters": [
        "field='.*',layer_index='.*'.water_content"
      ],
      "variables": ["water_content", "water_content", "water_content", "water_content"],
      "title": "Soil Water Content Stack Plot",
      "legend": ["Layer 0 Water Content (mm)", "Layer 1 Water Content (mm)", "Layer 2 Water Content (mm)", "Layer 3 Water Content (mm)"]
    },
    {
      "type": "plot",
      "filters": [
        "field='.*',layer_index='.*'.water_content"
      ],
      "title": "Soil Water Content Line Plot",
      "legend": ["Layer 0 Water Content (mm)", "Layer 1 Water Content (mm)", "Layer 2 Water Content (mm)", "Layer 3 Water Content (mm)", "Layer 4 Water Content (mm)"]
    }
``` 

Edit: the top graph specification can be modified to be:
```json
    {
      "type": "stackplot",
      "filters": [
        "field='.*',layer_index='.*'.water_content"
      ],
      "variables": ["0'.water_content", "1'.water_content", "2'.water_content", "3'.water_content"],
      "title": "Soil Water Content Stack Plot",
      "legend": ["Layer 0 Water Content (mm)", "Layer 1 Water Content (mm)", "Layer 2 Water Content (mm)", "Layer 3 Water Content (mm)"]
    }
```
and it seems to work as expected.

RUFAS/output_manager.py
```diff
@@ -893,6 +894,51 @@ def _route_save_functions(
...
+            if is_data_in_dict:
+                if selected_variables:
+                    data_dict = Utility.convert_list_of_dicts_to_dict_of_lists(values)
+                    for selected_variable in selected_variables:
+                        if selected_variable in data_dict:
+                            prepared_pool.setdefault(selected_variable, []).extend(data_dict[selected_variable])
+                else:
+                    self.add_error("No selected variables. Can't plot this data set.",
+                                   "List your 'variables' in the graph output filter.", info_map)
+                continue
```
```suggestion
                if not selected_variables:
                    self.add_error("No selected variables. Can't plot this data set.",
                                   "List your 'variables' in the graph output filter.", info_map)
                    continue
                data_dict = Utility.convert_list_of_dicts_to_dict_of_lists(values)
                for selected_variable in selected_variables:
                    if selected_variable in data_dict:
                        prepared_pool.setdefault(selected_variable, []).extend(data_dict[selected_variable])
```
Not an important change, but reworking this branching statement to something similar to the above would make the code a little flatter.

RUFAS/output_manager.py
```diff
@@ -893,6 +894,51 @@ def _route_save_functions(
                     info_map,
                 )
 
+    def _prepare_plot_data(self, filtered_pool: Dict[str, pool_element_type],
+                           selected_variables: Optional[List[str]] = None
+                           ) -> Dict[str, List[int | float]]:
+        """Extracts the values from the filtered pool data and converts them a dict format
```
```suggestion
        """Extracts the values from the filtered pool data and converts them to a dictionary format
```

Isgotkowitz (approved):
Changes look good Niko. Left some suggestions on small things.

I got the stack plot specified in earlier comments to work, it just required being more specific about how variables were being specified in the `"variables"` column.

RUFAS/output_manager.py
ew3361zh: I can take a look. I think if the user hasn't added selected_variables to their filter file then we just want to break out of this loop (i.e. not continue) because we can't plot any of the data from that filter file (as of now). I think this will get addressed when I dive into issue #899.

RUFAS/output_manager.py
```diff
@@ -893,6 +894,51 @@ def _route_save_functions(
                     info_map,
                 )
 
+    def _prepare_plot_data(self, filtered_pool: Dict[str, pool_element_type],
```
PooyaHekmati: Are there any particular reasons why this function is in OM and not in GG?

RUFAS/output_manager.py
PooyaHekmati: dictionary is a data structure, not a format.

RUFAS/output_manager.py
PooyaHekmati: >Not an important change, but reworking this branching statement to something similar to the above would make the code a little flatter.

Checking for the unhappy path first is a good idea and one the best practices in writing code. It leads to a cleaner, more understandable, and maintainable code and is less prone to bugs.

ew3361zh: I've adjusted the code accordingly and put the unhappy path first (and just break the loop at that point since it's unnecessary to continue).

RUFAS/output_manager.py
ew3361zh: I looked at both locations and the deciding factor was that we could log errors to help users/devs debug issues with their filter data better when the function is in OM. (Since we can't import OM into GG because of the circular import issue).

PooyaHekmati: preparing the plot data is beyond the domain of the output manager (preparing the plot data is different from managing the outputs). It should be move to GG. To avoid circular imports, you can return the results/status to the caller (output manager) so it can log it. Since the function is accepting a set of inputs, you may want to make it asynchronous and use `yield`.

JoeWaddell (approved):
Tested with various filters I've been using, and this works nicely!

As a suggestion: since the milk lactose/protein/fat values are (currently) less dynamic than they will be in the future, you could replace the currently stackplot in the `graph_filter_example.json` with the animal population filter below.

```
{
  "type": "stackplot",
  "filters": [
    "AnimalManager.daily_updates.(num_calves|num_heiferIs|num_heiferIIs|num_heiferIIIs|num_lactating_cows|num_dry_cows)"
  ],
  "title": "Herd composition (stackplot)",
  "legend": ["num_calves","num_heiferIs","num_heiferIIs","num_heiferIIIs","num_lactating_cows","num_dry_cows"]
}
```

ew3361zh:
> Tested with various filters I've been using, and this works nicely!
> 
> As a suggestion: since the milk lactose/protein/fat values are (currently) less dynamic than they will be in the future, you could replace the currently stackplot in the `graph_filter_example.json` with the animal population filter below.
> 
> ```
> {
>   "type": "stackplot",
>   "filters": [
>     "AnimalManager.daily_updates.(num_calves|num_heiferIs|num_heiferIIs|num_heiferIIIs|num_lactating_cows|num_dry_cows)"
>   ],
>   "title": "Herd composition (stackplot)",
>   "legend": ["num_calves","num_heiferIs","num_heiferIIs","num_heiferIIIs","num_lactating_cows","num_dry_cows"]
> }
> ```

Thanks, Joe!

ew3361zh:
After looking further into the matter, I wasn't able to get yield to work properly with the `prepare_plot_data()` method in `graph_generator`.

BUT!

I think we don't need it here. The order of the filters getting to `prepare_plot_data()` is 

1. OM's `save_results()` method loads the filter file and then loops through the list of filters.
2. For each filter, OM's `_route_save_functions()` is called and sends the data to GG's `prepare_plot_data()`.
3. So `prepare_plot_data()` will only be handling a single filter at a time.
4. For each filter, if the data provided is in a dictionary, the filter file either has selected_variables or doesn't and there is one route for each case (filter the provided data by the selected_variables and prepare a log of the successful filter or prepare and error message to be sent back to OM).
5. If the data provided is not in a dictionary, we have 2 other routes to be able to prepare the data accordingly (and add log messages).
6. Therefore there shouldn't be a situation where there's a backlog of filters stuck in `prepare_plot_data()` and we're needing to return data before its finished going through all the filters.

ew3361zh:
Updates:

- [x] `prepare_plot_data` function moved to GG, returns prepared data and logs of data being prepared (logs/errors/warnings) for OM to add to its pools.
- [x] creates method in OM to handle these returned logs and to direct graph_generator to act accordingly (i.e. don't graph if there was an error preparing the data which means there's no graphable data).
- [x] updates unit testing.

Please see note left above (9 hours prior) for discussion on why use of yield didn't seem necessary because `prepare_plot_data` is only receiving and processing 1 filter at a time.

ew3361zh:
With this last bunch of code added, I think we're touching on a good portion of #927. @Isgotkowitz if you could take another look with that issue in mind, that would be helpful.

RUFAS/graph_generator.py
```diff
@@ -150,10 +151,84 @@ def generate_graph(
...
+            A tuple containing the formatted data that can more readily be plotted by
+            graph_generator and the logs, warnings, and errors to be reported to OutputManager.
+        """
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self.prepare_plot_data.__name__,
+        }
+        selected_variables = graph_details.get("variables")
+        title = graph_details.get("title")
+        log_pool = []
```
PooyaHekmati:
```suggestion
        log_pool : Dict[str, str|Dict[str, str]]= []
```

RUFAS/graph_generator.py
```diff
@@ -150,10 +151,84 @@ def generate_graph(
...
+        prepared_pool = {}
```
PooyaHekmati: I spent 5 minutes trying to infer the type of data that goes into this dict and I couldn't, chances are other people won't be able to do it as well.

RUFAS/graph_generator.py
```diff
@@ -150,10 +151,84 @@ def generate_graph(
...
+        for index, key in enumerate(filtered_pool.keys()):
+            values: List[Any] = filtered_pool[key]["values"]
+            is_data_in_dict = isinstance(values[0], dict)
+            if is_data_in_dict:
+                if not selected_variables:
+                    error = {"error": f"Can't plot {title} data set", "message": f"No selected variables for {key}.",
+                             "info_map": info_map}
+                    log_pool.append(error)
```
PooyaHekmati: Simplify, we add documentary variables when the statement is hard to understand. `error = error` doesn't make the code more readable.
```suggestion
                    log_pool.append({"error": f"Can't plot {title} data set", "message": f"No selected variables for {key}.",
                             "info_map": info_map})
```

RUFAS/graph_generator.py
```diff
@@ -150,10 +151,84 @@ def generate_graph(
...
+                    break
+                else:
```
PooyaHekmati: remove unnecessary `else` and its indentation: we first check for the unhappy path first and either raise or return or break, so for the rest of the function we can safely assume everything is in order and we don't need to indent the code.
`Flat is better than nested.`
```suggestion
```

RUFAS/graph_generator.py
```diff
@@ -150,10 +151,84 @@ def generate_graph(
...
+                    data_dict = Utility.convert_list_of_dicts_to_dict_of_lists(values)
+                    selected_variables_in_data = 0
```
PooyaHekmati:
```suggestion
                    selected_variables_counter = 0
```

RUFAS/graph_generator.py
```diff
@@ -150,10 +151,84 @@ def generate_graph(
...
+                    for selected_variable in selected_variables:
+                        if selected_variable in data_dict:
+                            prepared_pool.setdefault(selected_variable, []).extend(data_dict[selected_variable])
+                            log = {"log": f"Successfully added {title} data to prepared_pool",
```
PooyaHekmati: `added {title} data`?

RUFAS/graph_generator.py
```diff
@@ -150,10 +151,84 @@ def generate_graph(
...
+                                   "message": f"Data for {selected_variable} added to prepared_pool.",
+                                   "info_map": info_map}
+                            log_pool.append(log)
```
PooyaHekmati: simplify

RUFAS/graph_generator.py
```diff
@@ -150,10 +151,84 @@ def generate_graph(
...
+                            selected_variables_in_data += 1
+                        else:
+                            warning = {"warning": f"{selected_variable} not a valid key in provided data",
+                                       "message": f"{selected_variable} won't be graphed.",
+                                       "info_map": info_map}
+                            log_pool.append(warning)
```
PooyaHekmati: check for unhappy path first

RUFAS/graph_generator.py
```diff
@@ -150,10 +151,84 @@ def generate_graph(
...
+                    error = {"error": f"Can't plot {title} data set", "message": f"No selected variables for {key}.",
+                             "info_map": info_map}
+                    log_pool.append(error)
+                    break
+                else:
+                    data_dict = Utility.convert_list_of_dicts_to_dict_of_lists(values)
+                    selected_variables_in_data = 0
+                    for selected_variable in selected_variables:
+                        if selected_variable in data_dict:
+                            prepared_pool.setdefault(selected_variable, []).extend(data_dict[selected_variable])
```
PooyaHekmati: I spent 5 minutes trying to understand what is happening here, this is my doubtful understanding: if we already have the `selected_variable` in the `prepared_pool.keys()` we extend its values by `data_dict[selected_variable]`, otherwise, we first make an empty list and then extend it?
`Readability counts.`
`Simple is better than complex.`
`If the implementation is hard to explain, it's a bad idea.`

RUFAS/graph_generator.py
```diff
@@ -150,10 +151,84 @@ def generate_graph(
...
+                                   "message": f"Data for {selected_variable} added to prepared_pool.",
+                                   "info_map": info_map}
+                            log_pool.append(log)
+                            selected_variables_in_data += 1
+                        else:
+                            warning = {"warning": f"{selected_variable} not a valid key in provided data",
+                                       "message": f"{selected_variable} won't be graphed.",
+                                       "info_map": info_map}
+                            log_pool.append(warning)
+                    if selected_variables_in_data == 0:
```
PooyaHekmati: if this variable is used to determine the validity of the data, then it should be a `bool` rather integer.

RUFAS/graph_generator.py
```diff
@@ -150,10 +151,84 @@ def generate_graph(
...
+                        error = {"error": f"Can't plot {title} data set",
+                                 "message": "No filter-file variables found in data provided.",
+                                 "info_map": info_map}
+                        log_pool.append(error)
+            else:
+                if selected_variables:
```
PooyaHekmati: I don't understand. `selected_variables` is supposed to tell us which items in the dict we want to plot. If data is not in dict, then why we need to check for the presence of `selected_variables`?

RUFAS/graph_generator.py
```diff
@@ -150,10 +151,84 @@ def generate_graph(
...
+                    prepared_pool[selected_variables[index]] = values
+                    log = {"log": f"Successfully added {title} data to prepared_pool",
+                                  "message": f"Data for {selected_variables[index]} added.",
+                                  "info_map": info_map}
+                    log_pool.append(log)
```
PooyaHekmati: simplify

RUFAS/graph_generator.py
```diff
@@ -150,10 +151,84 @@ def generate_graph(
...
+                else:
+                    prepared_pool[key] = values
+                    log = {"log": f"Successfully added {title} data to prepared_pool",
+                                  "message": f"Data for {key} added.",
+                                  "info_map": info_map}
+                    log_pool.append(log)
```
PooyaHekmati: simplify

RUFAS/output_manager.py
```diff
@@ -887,12 +887,16 @@ def _route_save_functions(
             if produce_graphics:
                 try:
                     graph_generator = GraphGenerator(self.__metadata_prefix)
-                    graph_generator.generate_graph(
-                        filtered_pool,
-                        filter_content,
-                        filter_file,
-                        graphics_dir,
-                    )
+                    prepared_data, log_pool = graph_generator.prepare_plot_data(filtered_pool,
```
PooyaHekmati: can we move this call to `graph_generator.generate_graph()`? That way, GG takes care of its own business, instead of expecting others to prepare the data for it.

RUFAS/output_manager.py
```diff
@@ -902,6 +906,33 @@ def _route_save_functions(
                     info_map,
                 )
 
+    def _route_graph_generator_logs(self, log_pool: List[Dict[str, str | Dict[str, str]]]) -> bool:
```
PooyaHekmati: Let's give this function a more generic name and decouple it from the GG.

RUFAS/output_manager.py
```diff
@@ -902,6 +906,33 @@ def _route_save_functions(
...
+        for log in log_pool:
+            if "error" in log:
+                self.add_error(log["error"], log["message"], log["info_map"])
+                error_count += 1
+            elif "log" in log:
+                self.add_log(log["log"], log["message"], log["info_map"])
+            elif "warning" in log:
+                self.add_warning(log["warning"], log["message"], log["info_map"])
+
+        return error_count == 0
```
PooyaHekmati: `Readability counts.`
```suggestion
        is_valid = True
        for log in log_pool:
            if "error" in log:
                self.add_error(log["error"], log["message"], log["info_map"])
                is_valid = False
            elif "log" in log:
                self.add_log(log["log"], log["message"], log["info_map"])
            elif "warning" in log:
                self.add_warning(log["warning"], log["message"], log["info_map"])

        return is_valid 
```

PooyaHekmati (changes requested):
The PR is on the right track, however, I couldn't understand the details to approve it. It is more complex/complicated than needs to be. Also, unable to comment on the applicability of `yield` for the same reason.

PooyaHekmati:
@Isgotkowitz @JoeWaddell Since the PR already has your approval, please feel free to answer my comments.

Isgotkowitz:
> With this last bunch of code added, I think we're touching on a good portion of https://github.com/RuminantFarmSystems/MASM/issues/927. @Isgotkowitz if you could take another look with that issue in mind, that would be helpful.

I think the changes made are definitely steps in the right direction, but not enough to close 927.

RUFAS/graph_generator.py
```diff
@@ -150,10 +151,84 @@ def generate_graph(
         except Exception as e:
             raise e
 
+    def prepare_plot_data(self, filtered_pool: Dict[str, Dict[str, List[Any]]],
```
Isgotkowitz: I think this function could have the functionality of `_route_graph_generator_logs()` added to it, without adding much complexity. Instead of saving all the errors/warnings/logs and returning them, raise everything to the Output Manager where it finds it and track whether it caught any errors with a boolean.

Then when it reaches the end of the function, return `None` if it caught any errors. Otherwise return `prepared_pool`.

RUFAS/graph_generator.py
Isgotkowitz: I think it would be `List[Dict[str, str | Dict[str, str]]]`

RUFAS/graph_generator.py
Isgotkowitz: Based on the return type hint, it should be `Dict[str, List[int | float]]`. But this raises the question: what happens if the user tries to plot a variable that is a list of strings?

RUFAS/graph_generator.py
ew3361zh: @Isgotkowitz that is the correct type annotation for the log pool, I will add it (it's essentially a list of dicts of what the add_log/add_warning/add_error methods take as arguments: log name, log message, info map).

RUFAS/graph_generator.py
ew3361zh: Added the type hint here as well. I considered the functionality of type checking the values for each of the variables listed in the filter but when Pooya and I talked it over, we determined it was too much responsibility to take on. I will double check what happens if someone adds a variable that has string data in it.

RUFAS/graph_generator.py
ew3361zh: Makes sense. Updated.

RUFAS/graph_generator.py
ew3361zh: This is an effort to make the logs clearer by associating them with the title from the filter file being used to prepare the date. So for example in the graph_generator_example.json file, the stackplot example's title is "Hard composition (stackplot)". So the message would be "...added Hard composition stackplot data."

RUFAS/graph_generator.py
ew3361zh: If I understand this correctly, I think we'd need to import OM to GG which we can't do for circular import error concerns. If I'm misunderstanding, please let me know.

RUFAS/graph_generator.py
ew3361zh: I think we have 4 main possibilities:
1. The values are in a dict and there are selected variables.
2. The values are in a dict and there are no selected variables.
3. The values are not in a dict and there are selected variables.
4. The values are not in a dict and there are no selected variables.

I'm not sure if option 3 is a probable possibility but in the case it comes up, my thought was we'd take the variable name the user chose (maybe it differs from the variable name assigned when it was added to OM) and add the data to the prepared pool with the variable. As I'm looking at it again though, we're opening the possibility for a list index out of range error (i.e. the list of variables they want to plot is shorter than the list of variables in the filtered pool). Maybe this is better addressed when we've incorporated regex into the prepare pool function.

ew3361zh: I can remove it for now.

RUFAS/output_manager.py
ew3361zh: Agreed!

RUFAS/output_manager.py
ew3361zh: Yes

RUFAS/graph_generator.py
ew3361zh: Yes, your "doubtful" understanding is correct.

RUFAS/graph_generator.py
ew3361zh: I couldn't find an instance of a dictionary having a string value to add here to check what would happen.

RUFAS/graph_generator.py
```diff
@@ -130,30 +131,100 @@ def generate_graph(
...
 
         Raises
         ------
         Exception
             Generic exception raised by utility functions.
         """
         try:
+            prepared_data, log_pool = self._prepare_plot_data(filtered_pool, graph_details)
+
+            error_found = any("error" in log for log in log_pool)
```
PooyaHekmati: just a thought: what about naming the variable: `found_errors`?

RUFAS/graph_generator.py
```diff
@@ -172,56 +243,17 @@ def _draw_graph(
...
-        plot_function : Callable[..., None]
-            Matplotlib function for plotting.
-
-        Returns: None
-        """
-        values_tuple = tuple(data_dict[variable] for variable in selected_variables)
-        plot_function(list(range(len(values_tuple[0]))), values_tuple)
+        if graph_type in TUPLE_BASED_FUNCTIONS:
+            values_tuple = tuple(data[variable] for variable in selected_variables)
+            plot_function(list(range(len(values_tuple[0]))), values_tuple)
```
PooyaHekmati: please decompose this into multiple lines. The variable names will help to make it easier to understand and maintain.

RUFAS/graph_generator.py
```diff
@@ -130,30 +131,100 @@ def generate_graph(
...
+                    if not is_variable_in_data:
+                        log_pool.append({"warning": f"{selected_variable} not a valid key in provided data",
+                                         "message": f"{selected_variable} won't be graphed.",
+                                         "info_map": info_map})
+                    else:
+                        prepared_pool.setdefault(selected_variable, []).extend(data_dict[selected_variable])
+                        log_pool.append({"log": f"Successfully added {title} data to prepared_pool",
+                                         "message": f"Data for {selected_variable} added to prepared_pool.",
+                                         "info_map": info_map})
+                if all(selected_variable not in data_dict for selected_variable in selected_variables):
```
PooyaHekmati: use explanatory variable here

RUFAS/graph_generator.py
ew3361zh: That works for me!

PooyaHekmati (approved):

PR #1001, opened 2023-11-28
Evaluate and correct Soil nitrogen dynamics

Isgotkowitz:
This branch/PR was used to tweak the nitrogen cycling submodule in the Crop and Soil module until the results met the satisfaction of the SMEs. Changes include:
- Adding organic manure N into the stable and active pools instead of the fresh and active pools, and changing the split between them so much more goes into the active pool.
- More variables are now collected by the OutputGatherer.
- `LayerData` and `SoilData` now track N movement and emissions with more granularity.
- Active organic N mineralizes into the ammonium pool, instead of directly into the nitrate pool.
- An equation was fixed in the humus mineralization submodule that was incorrectly specified in the SWAT documentation.
- The defaults for the manure applications provided by the manure module have been changed to be more reflective of actual dairy manure composition.
  - It is now 70% organic N and 30% ammonium by default.
- The denitrification rate default of 1.4 was found to be way to high, so it has been changed to 0.05 for all soil profiles and that is now the default in the metadata. 

Tests have been updated/added as well - note one exception: the profile totaling methods for different N pools in the soil. It felt redundant to test all of these because they all work exactly the same way, but tests for them can be added easily if that is deemed necessary.

Note that this PR does not address leaching and runoff N. Pete Vadas and I were having trouble getting the current equations to work, so he proposed a much simpler mechanism for calculating these factors. I will be implementing them in a subsequent PR.

Addresses #979


RUFAS/routines/field/field/field.py
Isgotkowitz, line 1: Note to reviewers: adding the day and year in the data and not the info map makes it way easier to extract, and is super useful.

RUFAS/routines/field/manager/output_gatherer.py
```diff
@@ -110,7 +110,22 @@ def send_daily_variables(self) -> None:
             om.add_variable("grazing_inorganic_phosphorus_runoff", field.soil.data.grazing_inorganic_phosphorus_runoff,
                             info_map)
             om.add_variable("soil_phosphorus_runoff", field.soil.data.soil_phosphorus_runoff, info_map)
+            om.add_variable("nitrate_runoff", field.soil.data.nitrate_runoff, info_map)
+            om.add_variable("ammonium_runoff", field.soil.data.ammonium_runoff, info_map)
+            om.add_variable("eroded_fresh_organic_nitrogen", field.soil.data.eroded_fresh_organic_nitrogen, info_map)
```
krb78, line 115: Apologies if I missed this Ed, but I thought we weren't adding to the "fresh" pools anymore and were just using "stable" and "active"

RUFAS/routines/field/soil/layer_data.py
```diff
@@ -262,6 +262,8 @@ class LayerData:
     """Fresh organic nitrogen content of this soil layer (kg / ha)
         Note: all layers except the top layer are initialized with 0 fresh organic nitrogen."""
 
+    nitrous_oxide_emissions: float = 0.0
+    """Amount of nitrous oxide emitted from this soil layer on the current day (kg / ha)."""
     annual_denitrified_nitrogen_total: float = 0
```
krb78: not something you've changed, but curious why we go from N2O emissions/day to annual denitrified nitrogen? doesn't the process of denitrification turn NO3- to N2 gas?

RUFAS/routines/manure/manure_nutrients/nutrient_request_results.py
```diff
@@ -17,13 +17,13 @@ class NutrientRequestResults:
     total_manure_mass: float = 0.0
     """Total amount of manure that can be fulfilled (kg). Default to 0."""
 
-    organic_nitrogen_fraction: float = 0.4
+    organic_nitrogen_fraction: float = 0.7
     """Fraction of nitrogen that is present in organic form, between 0 and 1 (unitless). Default to 0.4."""
```
krb78: Do we need to change the descriptions on these to correct the defaults?

krb78 (approved):
I think this looks great, Ed. Just a few minor comments and probably some are me and my lack of familiarity 🤓

RUFAS/routines/manure/manure_nutrients/nutrient_request_results.py
Isgotkowitz: Yes we do, good catch

RUFAS/routines/field/soil/layer_data.py
Isgotkowitz: Yes it does - `annual_denitrified_nitrogen_total` is the cumulative sum of `nitrous_oxide_emissions`. `annual_denitrified_nitrogen_total` was written way back when I did not understand what the process of denitrification produced. I can change the name of the latter to make them clearer.

Sidenote: even though one is the cumulative sum of the other, they have different units (annual is in kg, daily is in kg/ha), they should probably be made the same right? Also, the same issue may exist for ammonia emissions - I can change this if it seems like a good idea.

RUFAS/routines/field/manager/output_gatherer.py
Isgotkowitz: Hmm it is possible - but what are we talking about adding? N from manure used to be added to the fresh pool, but that was a definite wrong. The fresh pool is for N from plant residue, but maybe for simplicity/V1 sake we should just be putting plant residue into the stable and/or active pools?

RUFAS/routines/field/soil/nitrogen_cycling/nitrification_volatilization.py
```diff
@@ -31,21 +31,28 @@ def __init__(self, soil_data: Optional[SoilData] = None, field_size: Optional[fl
...
+        documentation eqn. 3:1.3.2, 3).
 
         References
         ----------
         SWAT Theoretical documentation section 3:1.3
 
         """
+        self.data.set_vectorized_layer_attribute("volatilized_ammonium_emissions", [0.0] * len(self.data.soil_layers))
         for layer in self.data.soil_layers:
             if layer.temperature <= 5:
```
matthew7838, line 50: Just a simple thought to make the code cleaner, maybe just run everything inside the scope of `layer.tmp >5`.

RUFAS/routines/field/soil/nitrogen_cycling/nitrification_volatilization.py
```diff
@@ -84,17 +92,25 @@ def _calculate_nitrification_volatilization_temp_factor(temperature: float) -> f
...
+
         References
         ----------
         SWAT Theoretical documentation eqn. 3:1.3.1
 
         """
-        return 0.41 * ((temperature - 5) / 10)
+        unbounded_factor = 0.41 * ((temperature - 5) / 10)
+        bounded_factor = min(1.0, unbounded_factor)
+        return bounded_factor
```
matthew7838: I think this whole line can be written in one line.

tests/soil_crop_tests/soil_tests/nitrogen_cycling_tests/test_nitrification_volatilization.py
```diff
@@ -7,17 +7,16 @@
 
 
 # --- Static method tests ---
-@pytest.mark.parametrize("temp", [
-    20,
-    31.493,
-    14.334,
-    5.0193
+@pytest.mark.parametrize("temp,expected", [
```
matthew7838, line 10: I think it could make things clearer to give a bit more detail into what to expect.

tests/soil_crop_tests/soil_tests/nitrogen_cycling_tests/test_nitrification_volatilization.py
```diff
@@ -7,17 +7,16 @@
...
+    (20, 0.615),
+    (31.493, 1.0),
+    (14.334, 0.382_694),
+    (5.0193, 0.000_791_3)
 ])
-def test_calculate_nitrification_volatilization_temp_factor(temp: float) -> None:
+def test_calculate_nitrification_volatilization_temp_factor(temp: float, expected: float) -> None:
```
matthew7838, line 16: same suggestion.

matthew7838 (approved):
LGTM, some small suggestions. I'll trust Pete and you with the conceptioinal aspects.

RUFAS/routines/field/soil/nitrogen_cycling/nitrification_volatilization.py
Isgotkowitz: There isn't a strict definition of code cleanliness, but in Python flatter code should be preferred over nested code as outlined in [PEP 20](https://peps.python.org/pep-0020/), so I think it should be kept as-is for right now.

tests/soil_crop_tests/soil_tests/nitrogen_cycling_tests/test_nitrification_volatilization.py
Isgotkowitz: Could you please be more specific? What isn't clear in this test?

tests/soil_crop_tests/soil_tests/nitrogen_cycling_tests/test_nitrification_volatilization.py
Isgotkowitz: Same comment as above.

RUFAS/routines/field/soil/nitrogen_cycling/nitrification_volatilization.py
matthew7838: That makes sense, thanks for pointing that out!

RUFAS/routines/field/soil/layer_data.py
krb78: yes! I'd put them in the same units - I don't think it really matters which one you choose, but they should either be kg/ha or kg

RUFAS/routines/field/manager/output_gatherer.py
krb78: haha, we might have been referring to manure not plant residue and I could be confused (most likely)! 😆 Would it be blasphemy to only add to the stable and/or active pools? I'll admit I don't know how RuFaS moves things from the fresh pool to the stable and active pools, so while my gut says simplicity might be better at this stage, I think if it's already in and an important distinction for us to make we should probably keep it. (because I'm good at adding complexity and never answering a question, this has me thinking about how our corn yields are actually lowest in the year following alfalfa which leads me to believe that in fact the fresh pools are not working properly if they were supposed to give approximately the amount of N we needed to get the same amount of growth as the synthetic fertilizer we add in subsequent years 🤔 ). Could @KFosterReed help us or should we direct this question to Pete and/or Eric?

Isgotkowitz: > Would it be blasphemy to only add to the stable and/or active pools

No, I think this would be a very reasonable choice for V1 - but I leave that decision to the SMEs. Maybe a good solution would be to leave it as-is in this PR, and in the crop evaluation we can actually observe the dynamics of the fresh pools when they have lots of plant residue actually adding to them. At that point we will be able to make a more informed decision about whether or not we should keep them and maybe invest time in debugging them, or just change to putting the residue N in the active and stable pools.

RUFAS/routines/field/soil/layer_data.py
Isgotkowitz: Awesome - will do. I'm going to stick with kg/ha to try to keep them as consistent as possible with everything else, but they can always be changed later.

Issue #863, opened 2023-10-17
Bug: Sphinx documentation is not being updated


Isgotkowitz:
The documentation generated by Sphinx should be regenerated and written to the `docs/_build` directory everytime the Sphinx workflow is run.

## Expected Behavior

The Sphinx documentation should be getting updated every time that the GitHub Sphinx workflow is being run.

## Current Behavior

The Sphinx documentation has not been updated since August 7, 2023. When the Sphinx workflow runs, it can be seen that it is generating new html documentation but somehow this is not being written into the commit.

## Steps to Reproduce

1.
2.
3.
4.

## Possible Implementation

Change to the current workflow from 
```sh
cd docs
make html
```
to 
```sh
cd docs
make clean
make html
```

ew3361zh:
Closed with PR #1010

PR #1003, opened 2023-11-29
Fix issue with Sphinx not updating

ew3361zh:
PR to address issue #863


ew3361zh:
PR #1010 has a better-working script.

PR #1012, opened 2023-12-01
Remove credit lines from all py files

### LINKED ISSUE #910: Remove credit lines from the codebase
In many files, we have credit lines such as written by, email address, etc... . We need to remove these from the codebase.

ew3361zh:
This PR addresses issue #910.

Manually went through each of the files and removed the credit lines at the tops of each of those files.

For the files from which I removed the credit lines, I also fixed the flake8 warnings and violations.

### DIFF
--- a/RUFAS/errors.py
+++ b/RUFAS/errors.py
@@ -1,28 +1,20 @@
-"""
-RUFAS: Ruminant Farm Systems Model
-File name: errors.py
-Description: Defines custom errors for RUFAS
-Author(s): Kass Chupongstimun, kass_c@hotmail.com
-"""
-
-
 class UserInput(Exception):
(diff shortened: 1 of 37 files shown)


ltp23 (approved):
Thanks Niko for going over many files!! I'm wondering if we could write a local script to do that too.

ew3361zh:
> Thanks Niko for going over many files!! I'm wondering if we could write a local script to do that too.

Thanks, Loi. I was wondering the same thing but didn't want to miss any files that might have small differences in how they wrote the credits and also didn't want to write something too general that would erase the notes at the top of some of the files that were in triple-quotes.

Isgotkowitz (approved):

PR #1010, opened 2023-11-30
Fix issue with Sphinx Documentation not automatically updating

ew3361zh:
PR to address issue #863 where Sphinx documentation is not being updated automatically with the current GitHub Actions Workflow.

## There were a few issues going on:
1. The .rst files used to create the html sphinx documentation haven't updated since the first time they were generated. They need to be regularly regenerated to reflect changes in the python files and overall model structure. This PR adds an action to address this but this can be done manually if this seems like too much to do with every push to GitHub. It depends on how up-to-date we want the Sphinx documentation to be.
2. The html documentation was using the old .rst files because new ones hadn't been generated.
3. The changes made by the Sphinx GitHub Action weren't being pushed to the PR after they were run.

## What this PR does to address the above issues:
- Updates the Sphinx Github Action to: 
   - Remove current .rst files except for index.rst and modules.rst.
   - Regenerate Sphinx documentation for the project using the sphinx-apidoc command.
   - Make new html documents based on the updated documentation.
   - Add these Sphinx documentation changes to the PR.

### Notes
- This first PR will be enormous because Sphinx documentation hasn't been updated in almost a year and a lot has changed in the model in the last year. I branched off this branch and the changes were minimal so it's not adding 43k lines every time its running the action.
- This will likely add some documentation to many PRs. We will need to tell people to ignore these files in their reviews.
- The index.rst and modules.rst files caused issues in creating the Sphinx documentation when they were not present during the rebuild (i.e. I tried removing them with all the other .rst files and the sphinx-apidoc command failed because it was looking for these two files).

PR #1015, opened 2023-12-04
removes inline comments

ew3361zh:
This branch is off of the #1010 branch to see if the github action will correctly only add new differences.


ew3361zh:
No longer needed. Just created for testing purposes.

PR #921, opened 2023-11-08
Updating the descriptions in the default metadata file

ltp23:
In this PR, everyone is welcome to improve the descriptions in the `default_metadata.json` related to their work in RuFaS.

PR #1022, opened 2023-12-05
Capability to create updated Data Collection App schema inputs with RuFaS

### LINKED ISSUE #682: Identify discrepancies between Crop and Soil Meta data and DCA inputs
Last week Ed submitted a PR with the Crop and Soil meta data (PR#665). He mentioned that there are some discrepancies in the input names and the description/units between the meta data and the DCA inputs. 

I am creating this issue to highlight the need to resolve those issues. The meta data should be taken as the correct values and updates made to the DCA. If questions arise, please make note of them and hold them for when Ed is from vacation this Wednesday (7/26).

### LINKED ISSUE #968: keeping metadata and DCA schema in sync
Keeping metadata and DCA schema is a labor-intensive work. We need to figure out a way to make this easier, automating it as much as possible. @allisterakun has already written a script for this, which can be a good starting point.

Isgotkowitz:
This is the first iteration of the `SchemaGenerator`, which automates the process of creating input schema for the Data Collection App. It has a GNU arg that can be used to invoke when running RuFaS (`-gs` or `--generate-schemas`)

If you want to check out these schemas in action (which I would highly recommend and appreciate), see PR [54](https://github.com/RuminantFarmSystems/Data-Collection-App/pull/54) in the DCA repo.

### In this PR
- The `SchemaManager` and associated tests have been created.
- `main.py` (and associated tests) have been updated to accept the new GNU arg and run SchemaManager when told to do so.
- The `.gitignore` has been configured to track the directory `data_collection_app_schemas/` but not any of the schemas in it.
- I found some irregularities in the metadata properties and corrected them to be consistent with the rest of the properties. Also a reason why this PR looks so massive. None of the changes I have made should effect how the simulation actually runs, but that should be verified independently as much as possible.
- The non default metadata properties have been synched up with the default metadata properties. Changes were made in the properties in #921 in the default metadata that were not added to the rest of the metadata files.

### Notes
This PR is already massive, so I held off on making changes that will probably be necessary in the future. They include:
- The log verbosity for the OutputManager is set multiple times in main, but this seems like the kind of thing that should only happen once. I added another call to set it before the SchemaManager runs so that it is possible to see logs and errors produced when generating schemas.
- A future feature for the SchemaGenerator/main would be adding the capability to specify which file the metadata properties are pulled from (which determine what the input schema are), and where the schemas are written to.
- As Allister pointed this out in his review, adding a way to specify the grid columns would be very beneficial.
- Also as Allister pointed out, there should be a better name for elements of an array than just "<array title>_element". I think a good solution would be adding a `"title"` key to the metadata properties.

Closes #968


RUFAS/schema_manager.py
```diff
@@ -0,0 +1,216 @@
...
+        default = input_properties.get("default")
+        description = input_properties.get("description")
+
+        if default is not None:
+            schema["default"] = default
+        if description is not None:
+            schema["options"] = {}
+            schema["options"]["infoText"] = description
+
+        keys = [key for key in input_properties.keys() if key != "type" and key != "description" and key != "default"]
```
allisterakun: maybe have a list of keys to ignore?

RUFAS/schema_manager.py
```diff
@@ -0,0 +1,216 @@
...
+
+    """
+
+    @staticmethod
+    def setup_number_schema(title: str, input_properties: dict[str, Any]) -> dict[str, Any]:
+        schema = {
+            "title": title,
+            "type": "number",
+            "options": {
+                "grid_columns": 12,
```
allisterakun: Maybe have this to be passed in as an argument, so that we can choose to have the variable take up half of the page width or full page width

RUFAS/schema_manager.py
```diff
@@ -0,0 +1,216 @@
...
+        if description:
+            schema["options"]["infoText"] = description
+
+        return schema
+
+    @staticmethod
+    def setup_bool_schema(title: str, input_properties: dict[str, Any]) -> dict[str, Any]:
+        schema = {
+            "title": title,
+            "type": "boolean",
```
allisterakun:
```suggestion
            "type": "boolean",
            "format": "checkbox",
```

allisterakun: The default format is a dropdown list of `True` and `False`. I think using a checkbox to indicate a boolean variable is more intuitive and more user-friendly.

RUFAS/schema_manager.py
```diff
@@ -0,0 +1,216 @@
...
+            "options": {
+                "grid_columns": 12,
+                "format": "select2",
```
allisterakun:
```suggestion

```

allisterakun: The "format" option should not go into the "options"; it does not have effect if placed here.

RUFAS/schema_manager.py
```diff
@@ -0,0 +1,216 @@
...
+        description = input_properties.get("description")
+
+        if default is not None:
+            schema["default"] = default
+        if description is not None:
+            schema["options"]["infoText"] = description
+
+        element_properties = input_properties["properties"]
+        element_schema_creator = SchemaManager.DATA_TYPE_TO_SCHEMA_SETUP_MAP[element_properties["type"]]
+        element_title = title + "_element"
```
allisterakun: After conversation with @Isgotkowitz, we decided this would be a future improvement to avoid the name becoming "variableName_element_element" for nested arrays.

RUFAS/schema_manager.py
allisterakun:
```suggestion
        ignored_keys = ["type", "description", "default"]
        keys = [key for key in input_properties.keys() if key not in ignored_keys]

```

allisterakun (changes requested):
Looks great! Just some small issues.

main.py
```diff
@@ -133,6 +140,23 @@ def run_rufas(
...
+        schema_generator = SchemaGenerator()
+        schema_generator.generate_schemas(None, None)
+
+        log_title = "Completed schema generation"
+        log_message = "Main routine completed generating new input schemas for the Data Collection App."
+        info_map = {
+            "class": "main",
+            "function": "run_rufas"
+        }
+        output_manager.add_log(log_title, log_message, info_map)
```
allisterakun: Maybe put this into a separate function in main? Just to keep the top-level run_rufas() clean.

tests/test_schema_generator.py
```diff
@@ -0,0 +1,276 @@
...
+                }
+            }
+        }
+     })
+])
+def test_setup_array_schema(title: str, properties: dict[str, Any], schema: dict[str, Any]) -> None:
+    """Tests that array schema is setup correctly."""
+    actual = SchemaGenerator.setup_array_schema(title, properties)
+
+    assert actual == schema
```
allisterakun: Would it be a better practice to patch `setup_number_schema()`, `setup_bool_schema()`, and `setup_string_schema()` and check the number of calls to each method? (This would also mean having tests for array of numbers, array of bools, and array of strings)

tests/test_schema_generator.py
```diff
@@ -0,0 +1,276 @@
...
+                    }
+                }
+            }
+        })
+])
+def test_setup_object_schema(title: str, properties: dict[str, Any], schema: dict[str, Any]) -> None:
+    """Tests that object schema are setup correctly."""
+    actual = SchemaGenerator.setup_object_schema(title, properties)
+
+    assert actual == schema
```
allisterakun, line 447: Same thing here, I would suggest to patch `setup_number_schema()`, `setup_bool_schema()`, and `setup_string_schema()`. Passing the expected number of calls to each sub-function in as an input, and check for the number of calls to each function.

allisterakun (approved):
Looks GREAT, Ed!!! Wonderful job! Left some comments on unit tests, just some suggestions.

main.py
Isgotkowitz: Good idea

tests/test_schema_generator.py
Isgotkowitz: Yes, it would definitely be a better practice. I will work on setting up these tests.

Update: the way that data types are mapped to schema setup methods (with the `DATA_TYPE_TO_SCHEMA_SETUP_MAP` variable) is problematic, it required a bunch of workarounds to get the mocking correct. To do this, I moved `DATA_TYPE_TO_SCHEMA_SETUP_MAP` to be outside of `SchemaGenerator`, and mocked the methods it contains directly. In future iterations, this design should be revisited.

tests/test_schema_generator.py
Isgotkowitz: Same as above response.

input/metadata/default_metadata.json
```diff
@@ -234,8 +234,8 @@
         "breed": {
           "type": "string",
           "default": "HO",
-          "pattern": "(HO|JE)",
-          "description": "Breed (select one) -- The predominant breed of the herd (Holstein or Jersey)"
+          "pattern": "^(HO|JE)$",
+          "description": "Breed (select one Holstein/Jersey) -- The predominant breed of the herd (Holstein or Jersey)"
```
elle-andreen: I would just leave this as 'select one' as the breeds/options are given in the description, similar to other 'select one' variables

elle-andreen:
This is looking great to me thus far as well! I had no issues running the simulation on this branch and don't see any issues with what you've modified in the metadata and other files, other than the one very minor comment above. This is great groundwork as for eventually refining and optimizing the SchemaGenerator.

I have pasted some thoughts from other conversation and my comments PR 54 (in the DCA repo) to spur others' thoughts on this:
- Eventually, for SchemaGenerator to function optimally (and save us the most time), the metadata will need to undergo some reformatting. As soon as we reasonably can, we may want to try to establish guidelines for how existing and new metadata entries should be formatted so that SchemaGenerator can pull the necessary information for each input variable. I'm sure that some of that reformatting can be done automatically, but will likely also require SME's taking the time to ensure the necessary info for each variable is there and is clear/complete. Ideally this would have happened in conjunction with the metadata review process, but we deemed it too labor and time intensive for the moment. But we should keep it on our mental horizon!
- We need the DCA to align with the metadata, but we also need it to be readable and easy to use. The DCA may need addition things not found in the metadata, like extra information, maybe different input formats that are then converted under the hood to match what's required by the metadata, external links, etc. I have started thinking about what those 'extras' may be, and I'm sure folks involved in the pilot testing have as well. I'm also curious about what can and cannot be easily preserved with each automatic update. If we were to add additional formatting beyond what we have here, then every time we update the content of the metadata, will that formatting and other content be preserved? 
- I see this script brings the metadata default values to the DCA, which is great. Can/are we incorporating the min/max values given in the metadata as well? Either as part of the infotext/description, or as some kind of validation in coordination with the JSON (or maybe that's happening already!)

0luchi (approved):
Very well documented and runs the simulation with no issues. Looks good to me!

RUFAS/schema_generator.py
```diff
@@ -0,0 +1,343 @@
...
+            try:
+                new_schema = SchemaGenerator.setup_object_schema(key, properties[key])
+            except Exception as e:
+                error_title = "Schema generator raised exception"
+                error_message = f"Key: '{key}' raised exception: {str(e)}"
+                om.add_error(error_title, error_message, info_map)
+                continue
+
+            schema_name = key.replace("properties", "schema")
+            new_schema_file_name = f"{schema_name}.json"
```
ltp23, line 81: Based on what I see in the DCA, I think you would want the output file type to be `.js` instead of `.json`, right?

RUFAS/schema_generator.py
```diff
@@ -0,0 +1,343 @@
...
+            new_schema_file_path = Path.joinpath(schema_output_path, new_schema_file_name)
+
+            log_title = "Schema generator writing new schema"
+            log_message = f"Writing new schema in {new_schema_file_path}"
+            om.add_log(log_title, log_message, info_map)
+
+            with open(new_schema_file_path, "w") as outfile:
+                json.dump(new_schema, outfile, indent=2)
```
ltp23, line 89: And to save each schema into a variable,

```
outfile.write(f"var {schema_name} = ")
json.dump(new_schema, outfile, indent=2)
outfile.write(";")
```

RUFAS/schema_generator.py
```diff
@@ -0,0 +1,343 @@
...
+            schema["maximum"] = maximum
+        if default is not None:
+            schema["default"] = default
+        if description:
+            schema["options"]["infoText"] = description
+
+        return schema
+
+    @staticmethod
+    def setup_bool_schema(title: str, input_properties: dict[str, Any]) -> dict[str, Any]:
```
ltp23, line 136: Need some docstring here?

RUFAS/schema_generator.py
```diff
@@ -0,0 +1,343 @@
...
+        ValueError
+            If the Regex pattern used for validation does not adhere to the format "^(<option 1>|<option 2>|...)$".
+
+        Notes
+        -----
+        When a string input is taken, often it is to select from a preset group of options. This method is designed to
+        derive those options from the Regex pattern that is used to validate it.
+
+        """
+        pattern = "\\^\\(.*\\)\\$"
```
ltp23, line 255: Can we assume this is the expected pattern?

ltp23 (approved):
Awesome PR, Ed! I did check out how the generated schemas would look in the DCA and the functionality remained the same. I guess the next step would be to auto-populate the sections that involve importing the schema files and the schema variables into the index.html file.

RUFAS/schema_generator.py
Isgotkowitz: Yes, but even changing `.json` to `.js` would not ensure that the names of the newly generated schema line up perfectly with names of the old schema. For example, the old field schema file name is `field_management.js` but the new schema file name is `field_schema.json`.

RUFAS/schema_generator.py
Isgotkowitz: The issue with adding this in (like above) is that there isn't a clean way to map the schema name generated by `SchemaGenerator` to the name that was in the old schema files.

RUFAS/schema_generator.py
Isgotkowitz: Yes good catch.

RUFAS/schema_generator.py
Isgotkowitz: Great question. I think we have to. It seems like all string inputs are either names (which have no constraints and don't need this method to be used) or have a very rigid set of strings that are considered valid inputs. For the latter, we need to be able to extract that rigid set from the pattern in order to properly create the drop down menu of options in the input schema.

Two notes: there are two inputs that don't work with the above assumption, the start and end dates specified in the config input. I think these are special cases though, and a better solution would be to modify their structure instead of modifying the `SchemaGenerator` to accommodate them.

The second note is that I don't know if it is possible to embed a Regex pattern in a schema and have the DCA enforce it on inputs. I didn't find a way to do this but if it is possible we should opt for utilizing that.

Isgotkowitz:
> Awesome PR, Ed! I did check out how the generated schemas would look in the DCA and the functionality remained the same. I guess the next step would be to auto-populate the sections that involve importing the schema files and the schema variables into the index.html file.

Thank you for your review and comments @ltp23. You raised a lot of good points about how the SchemaGenerator could/should be improved, but I think most of those changes would be too much to try and include in this PR. I will create an issue laying out steps and tasks for SchemaManager improvement and will include your suggestions there.

Isgotkowitz:
Issues addressed by this PR will be addressed by #2046 (and more!)

PR #1023, opened 2023-12-05
[1/n][Feed Storage] Implement Storage structure

PooyaHekmati:
This PR implements the structure of the storage class based on the design doc https://docs.google.com/document/d/1C86lch2dkYBERR-qLDmE7L7n1gYQjETrIKUyrQpWrPs/edit?usp=sharing along with its tests. The logic will be populated in future PRs.

### DIFF
--- a/RUFAS/config.py
+++ b/RUFAS/config.py
@@ -1,3 +1,6 @@
+from typing import Dict
+
+
 def is_leap_year(year):
     """
     Description:
--- a/RUFAS/config.py
+++ b/RUFAS/config.py
@@ -1,3 +1,6 @@
+from typing import Dict
+
+
 def is_leap_year(year):
     """
     Description:
@@ -18,8 +21,7 @@ def is_leap_year(year):
 
 
 class Config:
-
-    def __init__(self, data):
+    def __init__(self, data: Dict[str, int | bool | str]):
         """
         Description:
             Object containing configuration information of the simulation
@@ -31,19 +33,21 @@ class Config:
 
         # gets a start/end date in the format year:julian-day. That way the program
         # can start in the middle of the year
-        self.start_full_date = data['start_date'].split(':')
-        self.end_full_date = data['end_date'].split(':')
+        self.start_full_date = data["start_date"].split(":")
+        self.end_full_date = data["end_date"].split(":")
         self.start_year = int(self.start_full_date[0])
         self.end_year = int(self.end_full_date[0])
         self.start_day = int(self.start_full_date[1])
         self.end_day = int(self.end_full_date[1])
 
         # set seed attributes
-        self.set_seed = data['set_seed']
-        self.seed = data['random_seed']
+        self.set_seed = data["set_seed"]
+        self.seed = data["random_seed"]
 
         # TODO: remove conditional once all json files have simulate_animals field
-        self.simulate_animals = data['simulate_animals'] if 'simulate_animals' in data else True
+        self.simulate_animals = (
+            data["simulate_animals"] if "simulate_animals" in data else True
+        )
 
         year_length = 365
         leap_year_length = 366
@@ -56,15 +60,13 @@ class Config:
         self.years = []
 
         for year in range(self.start_year, self.end_year + 1):
-
             if year == self.start_year == self.end_year:
                 days = [None for _ in range(1, self.start_day)]
                 days += [_ for _ in range(self.start_day, self.end_day + 1)]
             elif year == self.start_year:
                 days = [None for _ in range(1, self.start_day)]
                 if is_leap_year(year):
-                    days += (_ for _ in range(self.start_day,
-                                              leap_year_length + 1))
+                    days += (_ for _ in range(self.start_day, leap_year_length + 1))
                 else:
                     days += (_ for _ in range(self.start_day, year_length + 1))
             elif year == self.end_year:
--- a/RUFAS/config.py
+++ b/RUFAS/config.py
@@ -1,6 +1,3 @@
-from typing import Dict
-
-
 def is_leap_year(year):
     """
     Description:
--- a/RUFAS/routines/feed_storage/harvested_crop.py
+++ b/RUFAS/routines/feed_storage/harvested_crop.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 from enum import Enum
+from RUFAS.time import Time
 
 
 class CropCategory(Enum):
@@ -71,6 +72,10 @@ class HarvestedCrop:
         The category of the crop (enum).
     type : CropType
         The type of the crop (enum), a subdivision of crop category.
+    harvest_time : Time
+         The time at which the crop was harvested
+    storage_time : Time
+         The time at which the crop was stored
     fresh_mass : float
         The fresh mass of the crop in kg.
     dry_matter_percentage : float
@@ -102,6 +107,8 @@ class HarvestedCrop:
 
     category: CropCategory
     type: CropType
+    harvest_time: Time
+    storage_time: Time
     fresh_mass: float
     dry_matter_percentage: float
     dry_matter_digestibility: float
--- a/RUFAS/routines/feed_storage/storage.py
+++ b/RUFAS/routines/feed_storage/storage.py
@@ -81,7 +81,7 @@ class Storage:
         """
         pass
 
-    def calculate_dry_matter_loss_to_gas(self, dry_matter: float, time_in_silo: int):
+    def calculate_dry_matter_loss_to_gas(self, dry_matter: float, time_in_silo: int) -> float:
         """
         Calculates the dry matter loss to gas.
 
--- a/RUFAS/routines/feed_storage/storage.py
+++ b/RUFAS/routines/feed_storage/storage.py
@@ -142,7 +142,7 @@ class Storage:
         """
         pass
 
-    def calculate_bale_density(self, initial_dry_matter: float):
+    def calculate_bale_density(self, initial_dry_matter: float) -> float:
         """
         Calculates the density of a bale.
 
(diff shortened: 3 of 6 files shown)


RUFAS/config.py
```diff
@@ -31,19 +33,21 @@ def __init__(self, data):
...
-        self.set_seed = data['set_seed']
-        self.seed = data['random_seed']
+        self.set_seed = data["set_seed"]
+        self.seed = data["random_seed"]
 
         # TODO: remove conditional once all json files have simulate_animals field
-        self.simulate_animals = data['simulate_animals'] if 'simulate_animals' in data else True
+        self.simulate_animals = (
+            data["simulate_animals"] if "simulate_animals" in data else True
+        )
```
Isgotkowitz, line 50: Can this conditional and associated TODO be removed? It seems that since `simulate_animals` is now in the config properties, it is safe to assume that simulate animals will be in the data.

RUFAS/routines/feed_storage/harvested_crop.py
```diff
@@ -102,6 +107,8 @@ class HarvestedCrop:
 
     category: CropCategory
     type: CropType
+    harvest_time: Time
+    storage_time: Time
```
Isgotkowitz, line 111: As an alternative to using two full `Time` object, what about keeping four `int`s: calendar year of harvest, ordinal day of harvest, calendar year of storage, and ordinal day of storage?

RUFAS/routines/feed_storage/storage.py
```diff
@@ -0,0 +1,169 @@
...
+        Calculates the dry matter loss to gas.
+    calculate_dry_matter_loss_to_effluent(dry_matter: float, estimated_maximum_effluent: float, time_in_silo: int):
+        Calculates the dry matter loss to effluent.
+    calculate_protein_degradation():
+        Calculates protein degradation.
+    calculate_heat_generated():
+        Calculates the total sensible heat generated.
+    calculate_bale_density(initial_dry_matter: float):
+        Calculates the density of a bale.
+    recalculate_nutrient_fractions():
```
Isgotkowitz:
```suggestion
    receive_crop(crop: HarvestedCrop)
        Receives a harvested crop and adds it to the storage.
    process_degradations()
        Processes the degradations and losses of the stored crops.
    give_feed(amount: float, crop_type: str)
        Gives out a specified amount of feed of a certain crop type.
    calculate_dry_matter_loss_to_gas(dry_matter: float, time_in_silo: int)
        Calculates the dry matter loss to gas.
    calculate_dry_matter_loss_to_effluent(dry_matter: float, estimated_maximum_effluent: float, time_in_silo: int)
        Calculates the dry matter loss to effluent.
    calculate_protein_degradation()
        Calculates protein degradation.
    calculate_heat_generated()
        Calculates the total sensible heat generated.
    calculate_bale_density(initial_dry_matter: float)
        Calculates the density of a bale.
    recalculate_nutrient_fractions()
```

RUFAS/routines/feed_storage/storage.py
```diff
@@ -0,0 +1,169 @@
...
+        Recalculates the relative nutrient concentrations after dry matter loss.
+    """
+
+    def __init__(self):
+        self.stored: List[HarvestedCrop] = []
+        self.capacity = float("inf")
+
+    def receive_crop(self, crop: HarvestedCrop):
```
Isgotkowitz:
```suggestion
    def receive_crop(self, crop: HarvestedCrop) -> None:
```
Might be overkill

RUFAS/routines/feed_storage/storage.py
```diff
@@ -0,0 +1,169 @@
...
+        crop_type : CropType
+            The type of crop to give out.
+
+        Returns
+        -------
+        None
+        """
+        pass
+
+    def calculate_dry_matter_loss_to_gas(self, dry_matter: float, time_in_silo: int):
```
Isgotkowitz:
```suggestion
    def calculate_dry_matter_loss_to_gas(self, dry_matter: float, time_in_silo: int) -> float:
```

RUFAS/routines/feed_storage/storage.py
```diff
@@ -0,0 +1,169 @@
...
+        Returns
+        -------
+        float
+            The amount of dry matter lost to gas.
+        """
+        pass
+
+    def calculate_dry_matter_loss_to_effluent(
+        self, dry_matter: float, estimated_maximum_effluent: float, time_in_silo: int
+    ):
```
Isgotkowitz, line 104:
```suggestion
    ) -> float:
```

RUFAS/routines/feed_storage/storage.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        Calculates protein degradation.
+
+        Returns
+        -------
+        None
+        """
+        pass
+
+    def calculate_heat_generated(self):
```
Isgotkowitz:
```suggestion
    def calculate_heat_generated(self) -> float:
```

RUFAS/routines/feed_storage/storage.py
```diff
@@ -0,0 +1,169 @@
...
+            The amount of dry matter.
+        estimated_maximum_effluent : float
+            The estimated maximum effluent.
+        time_in_silo : int
+            Time in days the crop has been in the silo.
+
+        Returns
+        -------
+        float
+            The amount of dry matter lost to effluent.
```
Isgotkowitz, line 120: Units?

RUFAS/routines/feed_storage/storage.py
```diff
@@ -0,0 +1,169 @@
...
+        ----------
+        dry_matter : float
+            The amount of dry matter.
+        time_in_silo : int
+            Time in days the crop has been in the silo.
+
+        Returns
+        -------
+        float
+            The amount of dry matter lost to gas.
```
Isgotkowitz, line 98: Units in docstring

RUFAS/routines/feed_storage/storage.py
```diff
@@ -0,0 +1,169 @@
...
+class Storage:
+    """
+    Abstract class representing a general storage structure.
+
+    Attributes
+    ----------
+    stored : List[HarvestedCrop]
+        A list of HarvestedCrop objects representing the crops stored.
+    capacity : float
+        The maximum capacity of the storage, currently set to infinity.
```
Isgotkowitz, line 14: Units should be specified.

RUFAS/routines/feed_storage/storage.py
```diff
@@ -0,0 +1,169 @@
...
+        Calculates the total sensible heat generated.
+
+        Returns
+        -------
+        float
+            The total sensible heat generated in kJ/kg.
+        """
+        pass
+
+    def calculate_bale_density(self, initial_dry_matter: float):
```
Isgotkowitz:
```suggestion
    def calculate_bale_density(self, initial_dry_matter: float) -> float:
```

Isgotkowitz (changes requested):
Great start. Majority of comments related to type annotations and units. Requesting changes because I think it would be good to have units in docstrings.

RUFAS/routines/feed_storage/storage.py
PooyaHekmati: will be added at the implementation time

RUFAS/routines/feed_storage/storage.py
PooyaHekmati: will be added at the implementation time

RUFAS/routines/feed_storage/storage.py
PooyaHekmati: will be added at the implementation time

RUFAS/routines/feed_storage/harvested_crop.py
PooyaHekmati: What is the benefit of that?

RUFAS/config.py
PooyaHekmati: It's out of the scope of this PR. This will be addressed in issue 1017.

RUFAS/routines/feed_storage/harvested_crop.py
Isgotkowitz: Eliminates unneeded complexity. A timestamp is just a static recording of a time, it doesn't need to be a full object that has methods for modifying it.

Also, it is more space efficient, because keeping a copy of `Time` would need to be created to be stored in each field.

Isgotkowitz (approved):

KevinP-B (approved):
Harvest and storage times added to test

RUFAS/routines/feed_storage/storage.py
```diff
@@ -0,0 +1,169 @@
...
+from .harvested_crop import HarvestedCrop, CropType
+
+
+class Storage:
+    """
+    Abstract class representing a general storage structure.
+
+    Attributes
+    ----------
+    stored : List[HarvestedCrop]
```
JoeWaddell, line 11: Since this deals with just storage of harvested crops, is there a separate class that will handle storage for A) purchased crops, and B) feed additives?

RUFAS/routines/feed_storage/harvested_crop.py
PooyaHekmati: Some operations such as comparison, subtraction, addition, etc... can be defined on timestamps which need special handling. So it is not just a normal integer. We have to be consistent across the codebase on how we track and store time to prevent compatibility issues down the road. The space that an instance of time requires is nothing compared to other things that we store, I/O pools, for examples.

RUFAS/routines/feed_storage/storage.py
PooyaHekmati: That is indirectly an open question that we had in the design doc. For now, the focus here is on the harvested crops. We can always modify it of course. We could also consider bringing the purchase feed logic to this module (instead of handling it in the animal module)

PR #1024, opened 2023-12-06
[2/n][Feed Storage] Implement Hay structure 

PooyaHekmati:
This PR implements the structure of the hay class and its children based on the design doc https://docs.google.com/document/d/1C86lch2dkYBERR-qLDmE7L7n1gYQjETrIKUyrQpWrPs/edit?usp=sharing along with its tests. The logic will be populated in future PRs.

PR #1025, opened 2023-12-06
[3/n][Feed Storage] Implement Grain structure

PooyaHekmati:
This PR implements the structure of the hay class and its children based on the design doc https://docs.google.com/document/d/1C86lch2dkYBERR-qLDmE7L7n1gYQjETrIKUyrQpWrPs/edit?usp=sharing. The logic and tests will be populated in future PRs.


KevinP-B (approved):
Grain storage implemented for dry and high-moisture storage. Looks good.

RUFAS/routines/feed_storage/grain.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    Represents grain storage and manages its specific attributes and behaviors.
+
+    Inherits from Storage.
+    """
+
+    pass
+
+
+class Dry(Grain):
```
JoeWaddell, line 14: Should these class names be grain-specific? i.e. DryGrain - in case of other Feeds also using the name "Dry". Or will this be clear once instantiated?

RUFAS/routines/feed_storage/hay.py
```diff
@@ -0,0 +1,85 @@
...
+        Calculates the protein loss in the hay.
+
+        Returns
+        -------
+        None
+        """
+        pass
+
+
+class ProtectedIndoors(Hay):
```
JoeWaddell: Same question about names: once in use, will ProtectedIndoors be clearly differentiated from the same name for  other Feed types?

JoeWaddell (commented):
Essentially the same questions as #1023 and #1024 (and I can see that #1026 and #1027 will be the same), in addition to thoughts about whether class names need to be unique. Nice job laying the groundwork!

RUFAS/routines/feed_storage/grain.py
PooyaHekmati: hmmm, good question. I don't know, @KevinP-B is a better person to answer this. I am OK either way.

JoeWaddell: Cool: updating the names can happen in the PRs that add the methods themselves, so I'll approve this and the others (once I finish reviewing those).

JoeWaddell (approved):

PR #1029, opened 2023-12-06
Unit tests for simulation engine

### LINKED ISSUE #821: Increase Test Coverage for simulation_engine.py
The test coverage for simulation_engine.py is below 100% but the structure of classes.py, on which simulation_engine.py heavily depends, is too complex and difficult to mock.

## Expected Behavior
Currently test coverage for simulation_engine.py needs to be at 100%.

## Current Behavior
Currently test coverage for simulation_engine.py is 67%.

## Possible Solution
One of the tasks for Milestone 3 is to revisit classes.py. I think at that point, simulation_engine.py will also either be refactored or be easier to test (hopefully disentangled from relying too heavily on multiple classes from classes.py).

## Detailed Description
Part of the issue now is that SimulationEngine._initialize_simulation() depends on the Weather class which violates the max complexity linting guideline (it is complexity of 12 and max complexity is 10).

ltp23:
This PR removes the existing unit tests and add new ones for the simulation engine. I put all the tests in a separate file.


ew3361zh (approved):
Really nice work, Loi! Thorough testing and I think that's a great idea to move them into their own file.

RUFAS/simulation_engine.py
```diff
@@ -86,28 +100,36 @@ def _run_post_annual_routines(self) -> None:
...
+    @staticmethod
+    def _visualize_sim_progress(day: int, update_interval: int = 50) -> None:
         """
         Shows a rotating char on console to confirm simulation is alive.
 
-        Args:
-            day (int): day of year
-            update_interval (int): the interval at which the char symbol is updated
+        Parameters
+        ----
```
Isgotkowitz:
```suggestion
        Parameters
        ----------
```

RUFAS/simulation_engine.py
```diff
@@ -86,28 +100,36 @@ def _run_post_annual_routines(self) -> None:
...
+        day : int
+            day of year
+        update_interval : int
+            the interval at which the char symbol is updated. Default is 50.
```
Isgotkowitz:
```suggestion
            Ordinal day of the year.
        update_interval : int, default=50
            The interval at which the char symbol is updated.
```
Adding units if known for `update_interval` would be good.

tests/test_simulation_engine.py
```diff
@@ -0,0 +1,329 @@
...
+
+    # Arrange
+    mocker.patch.object(SimulationEngine, '__init__', return_value=None)
+    simulation_engine = SimulationEngine()
+
+    simulation_engine.time = mocker.MagicMock()
+    simulation_engine.time.to_str.return_value = to_str_return_value
+    simulation_engine.state = mocker.MagicMock()
+    simulation_engine.state.animal_manager.simulation_day = initial_simulation_day
+    patch_for_add_log = mocker.patch('RUFAS.simulation_engine.om.add_log')
```
Isgotkowitz, line 280: This is awesome, I had no idea you could patch Output Manager like this. Would this way be a best practice for checking that Output Manager methods are used correctly?

RUFAS/simulation_engine.py
```diff
@@ -21,31 +21,42 @@
 
 
 class SimulationEngine:
+    """
+    The SimulationEngine class is responsible for orchestrating the entire simulation
+    process for RuFaS. It manages the simulation's lifecycle, advancing time, executing daily
+    and annual routines, and logging simulation progress.
+    """
```
Isgotkowitz, line 44: Could you add a Methods and Attributes section here?

Isgotkowitz (approved):
Looks great Loi. Only comments are suggestions for improvements in docstrings

tests/test_simulation_engine.py
ltp23: I don't know if this is a best practice. I think it looks neat and works.

ltp23:
Thanks Niko and Ed for your reviews.

PR #1037, opened 2023-12-12
[Report Generator] No aggregation reports

PooyaHekmati:
This PR mainly focuses on allowing users to provide a report filter with no aggregation. Previously, we handled this case by raising an error. The user might be interested in this feature as it allows them to slice data, assign labels to it, and save it along with aggregated data in the same CSV file.
If data is being aggregated, we now append `hor_agg`, `ver_agg`, `ver_hor_agg`, or `hor_ver_agg` to the column name to denote the aggregation.
The PR also makes some minor improvements in terms of code organization in `RG.generate_report()`.

### DIFF
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -833,7 +833,7 @@ class OutputManager(object):
                                 if column_name not in reports.keys()
                                 else f"{column_name} {self._get_timestamp(True)}"
                             ] = {"values": value}
-                    except (ValueError, KeyError) as e:
+                    except KeyError as e:
                         self.add_error("report generation error", str(e), info_map)
                 else:
                     self._route_save_functions(
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -833,7 +833,7 @@ class OutputManager(object):
                                 if column_name not in reports.keys()
                                 else f"{column_name} {self._get_timestamp(True)}"
                             ] = {"values": value}
-                    except KeyError as e:
+                    except (KeyError, ValueError) as e:
                         self.add_error("report generation error", str(e), info_map)
                 else:
                     self._route_save_functions(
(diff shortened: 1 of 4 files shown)


jadamchick (approved):
Works great! 

I used this filter: 
{
    "name": "Animals sold in the last year",
    "filters": ["AnimalModuleReporter.report_sold_animal_information.(animal_id|animal_type|body_weight)"],
    "slice_start": -365
 }

And got exactly the info I wanted. 
And when I included that within a report with multiple filters, it also worked nicely and included these values logically in the csv output. 
The tags on the labels of the aggregated columns (eg "ver_agg") also work well. 

I did notice that when a filter leads to empty report data, it halts the production of all reports (it used to report this in the errors file but continue to generate all of the reports for which it could locate the data). I preferred the previous, so that I would still get the majority of the reports (and the logs, warnings, etc. files to help me troubleshoot if needed).

PooyaHekmati:
> I did notice that when a filter leads to empty report data, it halts the production of all reports (it used to report this in the errors file but continue to generate all of the reports for which it could locate the data). I preferred the previous, so that I would still get the majority of the reports (and the logs, warnings, etc. files to help me troubleshoot if needed).

Thanks for catching that. It should be fixed now, can you please give it a second look?

jadamchick:
> > I did notice that when a filter leads to empty report data, it halts the production of all reports (it used to report this in the errors file but continue to generate all of the reports for which it could locate the data). I preferred the previous, so that I would still get the majority of the reports (and the logs, warnings, etc. files to help me troubleshoot if needed).
> 
> Thanks for catching that. It should be fixed now, can you please give it a second look?

Yes, it works as expected now. Thank you!

Isgotkowitz (approved):
Code looks good to me. I tried creating some new reports and worked well - this is going to be a really handy feature.

PR #1041, opened 2023-12-12
Crop configuration changes

### LINKED ISSUE #959: Define default crop quality metrics for crop types in the Crop Module
Currently, there are few to no crop quality metrics being estimated or tracked for crops growing in RuFaS. This needs to be changed in order for the Crop and Soil module to be properly integrated into the Feed module, as well as for making sure that the Carbon Cycling module is functioning as well as possible for V1.

Define a table that maps general crop types to default crop quality metrics.

Implement and calibrate a growing degree day-based model for estimating crop quality metrics in a more dynamic way.

**Additional context**
There is a [design document](https://docs.google.com/document/d/1auWeT6XnItd8yLoWGkbSwJ2DkUULxv5ClA48Grm69IE/edit#heading=h.h0dz48r2a92t) for defining the table of quality metrics.

Isgotkowitz:
This PR changes the crops available to grow in RuFaS and the harvest parameters they use, so that we will be able to effectively evaluate the Crop module and interface precisely with the Feed Storage module.

### New Crops Available
The crop configurations available to users have been moved out of the file `crop_data.py`. They have been grouped by species into files within the new `crop_configurations` directory. The following are available (each bullet corresponds to a file):
- alfalfa (silage, baleage, hay)
- cereal rye (grain, silage, baleage, hay)
- corn (silage, grain)
- soybean (grain, hay)
- tall fescue (silage, baleage, hay)
- triticale (grain, silage, baleage, hay)
- winter wheat (grain, silage, baleage, hay)

All available crops inherit from their generic species, which inherit from `CropData`. Note that the documentation has been added or updated in places but is not uniform. I believe it is passable in its current state given the timeline this PR needs to be completed within.

### Changes to Code
- `crop_management.py`: has been updated so that yield nitrogen and phosphorus are correctly calculated based on the new harvest parameters, and a more detailed transfer of residue occurs between the Crop and Soil modules now.
- `dormancy.py`: the documentation has been rewritten to follow the NumPy style. It also has been updated to correctly account for the new harvest parameters, and properly transfer residue lost to the soil profile.
- `species_data_factory.py`: refactored so that it only creates the newly supported crops configurations.
- `field.py`: passes more information to `dormancy.py` now, and now has the `CarbonCycling` module process the residue from plants going into dormancy. 
- `residue_partition.py`: added a missing section to the documentation.
- Testing has been updated as appropriate. Coverage remains high for the Crop and Soil module overall.

### Changes to Inputs
- All metadata properties have had their properties updated so that only supported crop species are accepted.
  - There was also a manure handler option that wasn't in the non-default metadata properties - this has been fixed too.
- All crop inputs files (housed in `inputs/data/crop`) have been changed so that they use supported crops. These changes were made arbitrarily, which I think is ok because as far as I know no one is paying super close attention to the crop yields yet.

Closes #959


allisterakun (approved):
Great work Ed! It looks good to me! Very nice refactoring, the class hierarchy is much clearer.
All other changes are simple logic with precise docstring.

RUFAS/routines/field/crop/crop_configurations/alfalfa.py
```diff
@@ -0,0 +1,137 @@
...
+    lignin_dry_matter_percentage: float, default=7.419
+        Percentage of dry matter yield that is lignin (unitless).
+    yield_nitrogen_fraction: float, default=0.0327536
+        Fraction of dry matter crop yield that is nitrogen (unitless).
+    yield_phosphorus_fraction: float default=0.00351
+        Fraction of wet crop yield that is phosphorus (unitless).
+
+    Notes
+    -----
+    Alfalfa baleage currently has the same harvest and quality properties as Alfalfa silage.
```
krb78, line 97: This is ok for now - baleage can be a lot more variable in it's DM when harvested, but we can save those updates for V1.X :)

RUFAS/routines/field/field/field.py
```diff
@@ -909,17 +909,19 @@ def _make_custom_crop(**specs) -> Crop:
...
         Parameters
         ----------
         daylength : float
             Length of time from sunup to sundown on the current day (hours).
+        rainfall : float
+            Amount of rain that fell on the current day (mm).
 
-        Details
-        -------
+        Notes
```
krb78, line 923: how do crops come out of dormancy? Is it just the opposite parameters? Maybe I'll get there as I continue to read ;)

krb78 (approved):
I think this looks great, Ed. Generally, it's totally fine to have our default baleage to be silage for now. Baleage can be harvested at a more variable DM% than silage can be but someday we can allow users to change some of these defaults (I think). 

Generally, I think the structure of this is much simpler and will allow us to work toward adding more crops and varieties in an easier way in the future.

RUFAS/routines/field/crop/crop_configurations/alfalfa.py
Isgotkowitz: Hopefully we'll be able to build a dynamic dry matter tracker, so we won't even need it in the crop configuration.

RUFAS/routines/field/field/field.py
Isgotkowitz: To read the theory that we use, check [SWAT](https://swat.tamu.edu/media/99192/swat2009-theory.pdf) section 5:1.2. It doesn't say anything specifically about crops that should be dying instead of going into dormancy (a.k.a. annuals I think), so it may/may not be important for us to check that in the Crop evaluation.

PR #1042, opened 2023-12-13
Perform aggregation on referenced columns and scalar operations on each column

### LINKED ISSUE #997: Make constants in report generator filters
The report generator works by aggregating the data that comes from the output manager's variables pool. For some operations, however, it is needed to accept a constant value from the user. For instance, the user may want to divide something by 100 to get the percentage. Or they may want to multiply by 454 to convert lb to g. Make this possible by extending the list of possible keys (add constant?). One thing to consider is the difference between horizontal and vertical aggregation, how can we differentiate the constants for them?

### LINKED ISSUE #998: Add Cross referencing in the report generator
Consider the report filter below:
```json
{
  "multiple": [
    {
      "name": "first item",
      "filters": ["LifeCycleManager.daily_update.daily_milk_production"],
      "slice_start": -365
    },
    {
      "name": "second item",
      "filters": ["Pen\\.calc_total_manure\\.pen_\\d+_daily_urine_nitrogen"],
      "vertical_aggregation": "sum"
    }
  ]
}
```
Currently, the first and second items are handled separately. In this issue, we would like to make it possible to draw a connection between them so that it is possible to aggregate their results, for instance, `third item = first item + second item` or `third item = first item / second item`. 
Note that the current flow of the code might be blocking this. Consider shuffling things around: right now the output manager filters the pool and sends the results to the report generator if applicable. It might make more sense to start the flow from the report generator and have it call the output manager. (push vs pull)

ltp23:
This PR focuses on adding a couple of features to the report generator.

- Cross-referencing: Ability to refer to previously defined reports in the same report json config file (Issue #998).
- Constant columns: Adding a column that contains only some value. The typical use case for this is to do a horizontal aggregation operation with such a column first and then optionally do a vertical aggregation (Issue #997). 

```
   {
      "name": "Last Year's Average Daily Milk Production (gallons)",
      "filters": [
        "LifeCycleManager.daily_update.daily_milk_production"
      ],
      "constants": {
        "Liters to Gallons": 0.264172
      },
      "horizontal_first": true,
      "slice_start": -365
    },
    {
      "name": "Last Year's Average Daily Number of Milking Cows",
      "filters": [
        "LifeCycleManager.daily_update.milking_cow_num"
      ],
      "slice_start": -365
    },
    {
      "name": "Last Year's Average Daily Milk Production per Cow (gallons)",
      "cross_references": [
        "Last Year's Average Daily Milk Production (gallons)_hor_ver_agg",
        "Last Year's Average Daily Number of Milking Cows_ver_agg"
      ],
      "horizontal_aggregation": "division"
    }
```

The existing code for generating reports is still intact, just in case there is a need to go back to that version. However, I moved the code for handling the report generator in the OutputManager into a separate method.

---

Edit 1:
- Added issue numbers for the first two bullet points.
- Removed the padding section
- Changed the column titles a bit

---
Edit 2:
- Simplified the format for constants.
- Added suffixes to cross-reference columns.


RUFAS/output_manager.py
```diff
@@ -805,46 +804,31 @@ def save_results(
...
                     continue
-                filtered_pool = self._filter_variables_pool(
-                    filter_content["filters"], filter_file
-                )
+
+                filtered_pool = {}
+                if "filters" in filter_content.keys():
+                    filtered_pool = self._filter_variables_pool(
+                        filter_content["filters"], filter_file
+                    )
```
PooyaHekmati: Why is this if statement needed? what happens if filters is not in filter_content?

RUFAS/output_manager.py
```diff
@@ -805,46 +804,31 @@ def save_results(
...
+                        f"The `filters` can only be skipped if `references` entry is present "
+                        f"and the purpose is to generate a derived report.",
                         info_map,
                     )
                     continue
-                filtered_pool = self._filter_variables_pool(
-                    filter_content["filters"], filter_file
-                )
+
+                filtered_pool = {}
```
PooyaHekmati: add type annotation

RUFAS/output_manager.py
```diff
@@ -861,15 +845,107 @@ def save_results(
             )
             self._dict_to_file_csv(reports, report_file_path)
 
+    def _handle_report_generation(self,
+                                  filter_content: Dict[str, Any],
+                                  filtered_pool: Dict[str, Any] | None,
```
PooyaHekmati: wrong type hint, `filtered_pool` comes from `_filter_variables_pool` which returns `Dict[str, pool_element_type]`

RUFAS/output_manager.py
```diff
@@ -861,15 +845,107 @@ def save_results(
             )
             self._dict_to_file_csv(reports, report_file_path)
 
+    def _handle_report_generation(self,
```
PooyaHekmati: This method should be placed in ReportGenerator

RUFAS/output_manager.py
```diff
@@ -861,15 +845,107 @@ def save_results(
...
+        info_map : Dict[str, Any]
+            A dictionary containing logging information such as the class and function names.
+
+        reports : Dict[str, Dict[str, List[Any]]]
+            A dictionary to store the generated reports, keyed by their names.
+
+        Raises
+        ------
+        ValueError, KeyError
+            If there is an error in report generation, such as missing data or invalid configuration.
```
PooyaHekmati: This docstring doesn't give any extra information. Provide more details for each case separately.

---
Update after reading the code: this function doesn't raise anything

RUFAS/output_manager.py
```diff
@@ -861,15 +845,107 @@ def save_results(
...
+
+        self.add_log("init_report_generation", "Report Generation Started", info_map)
+        report_generator = ReportGenerator()
+        try:
+            report_name = self._generate_unique_report_name(filter_content, reports)
+
+            if "cross_references" in filter_content.keys():
+                self._check_for_missing_references(filter_content["cross_references"], reports)
+                reference_data = {ref: reports[ref] for ref in filter_content["cross_references"]}
+                filtered_pool.update(reference_data)
```
```suggestion
                reference_datal.update(filtered_pool)
```
this prevents potential side effects.

RUFAS/output_manager.py
```diff
@@ -861,15 +845,107 @@ def save_results(
...
+        """
+
+        base_name = filter_content.get("name", f"untitled_{self._get_timestamp(True)}")
+
+        if base_name in reports:
+            base_name = f"{base_name} {self._get_timestamp(True)}"
+
+        return base_name
+
+    @staticmethod
```
PooyaHekmati: why does this need to be static?

PooyaHekmati (changes requested):
Could not finish the review as the PR is not in a review-ready state. 
1. Clearly mention the issues that this PR is aiming to address.
2. Avoid packing other focal points in one PR (padding is out of the scope of this PR).
3. Docstrings are not informative and clear enough.
4. There are code organization issues here: misplaced methods (things that belong to RG are placed in OM, static methods are used with no justification)

RUFAS/output_manager.py
ltp23: There is an if statement above that that checks for the absence of the `filters` and `cross_references` keys. 

```
if (
        not isinstance(filter_content, dict)
        or ("filters" not in filter_content.keys() and "cross_references" not in filter_content.keys())
):
    self.add_error(...)
```

If this part checks out, then we know there must be either `filters` or `cross_references` or both keys in the `filter_content`.

This means it is okay to have both filtered data from the output manager (using the `filters` key) and previously aggregated data (using the `cross_references` key) in the same report.

If this seems unnecessary, I can just allow for one or the other option.

RUFAS/output_manager.py
ltp23: Okay, I will add this type to it `Dict[str, Dict[str, List[Any]]]`

RUFAS/output_manager.py
ltp23: I think this method should stay in the Output Manager so we can have easier access to all the `reports` and timestamp generation used in naming.

tomhuhh (commented):
Thanks Loi! Here is a summary of my comments on this PR from the output standpoint based on our discussion. 

1. For the 'Last 6 Month's Heifer Population Growth' value, if it is not a special request, I think we can remove it as an example because it can be misleading. Otherwise, we need to find ways to correctly calculate it. 
2. We also agreed to update the unit for methane. So instead of '90 to 30 days ago Methane Emission from Lactating Cows', I would write '90 to 30 days ago Methane Emission from Lactating Cows (grams)'.
3. Calculations for milk yield need to be revisited, either it is daily milk yield per cow or  annual milk production per cow. 

Thanks for the hard work!

RUFAS/output_manager.py
ltp23: Right, I will add a Notes section to describe when an error can occur and cause that report to be skipped. There won't be a `Raises` section because the error is logged by the Output Manager.

RUFAS/output_manager.py
ltp23: I think both versions are equivalent because `filtered_pool` and `reference_data` are created anew for every new report. Unless we make deep copies every time, both dictionaries can be modified in-place, which can corrupt the original data pool if not careful. The code in the report generator doesn't modify existing data and only creates new data, so it is okay not to do the deep copying. But if we want to be sure, deep copying may be necessary.

RUFAS/output_manager.py
PooyaHekmati: Please use explanatory variables to make the code more understandable and maintainable.

RUFAS/output_manager.py
PooyaHekmati: The domain of responsibility dictates where each method belongs to, not ease of access. This method is doing things that are not under the umbrella of Managing Outputs.

RUFAS/output_manager.py
ltp23: It doesn't need to be static. It could also be a normal instance method with `self` or a classmethod with `cls`. The general way that I use `staticmethod` here and in other places is when we don't need a specific instance object for that method to work on. For example, here, I could call this method directly without creating an `om` object — `OutputManager._check_for_missing_references(refs, reports)`. It can also be called like this `self._check_for_missing_references(...)` from another `self` method in OutputManager.

I understand this can be confusing by switching back and forth between different method styles. I'm open to keeping things simple by just using one style. 

Another thing to notice is static methods don't use the `self` variables, so these methods could also be moved somewhere else. But I don't think this is worth it if the methods are not that reusable in other contexts.

RUFAS/output_manager.py
PooyaHekmati: You have two dictionaries here and are appending them together. One of them is instantiated locally, the other one is an incoming arg. If you append the local one to the arg one (your current version), the caller receives this change as well, i.e., the function has a side effect. But if you append the arg to the local one, there won't be any side effects without needing deepcopy.
This code:
```Python
def append_local_to_arg(d):
    local_d={"2":2}
    d.update(local_d)
    
def append_arg_to_local(d):
    local_d={"3":3}
    local_d.update(d)
    
d={"1":1}
print(f"before any function calls {d=}")
append_local_to_arg(d)
print(f"after call to append_local_to_arg {d=}")
append_arg_to_local(d)
print(f"after call to append_arg_to_local {d=}")
```
Generates this output:
```
before any function calls d={'1': 1}
after call to append_local_to_arg d={'1': 1, '2': 2}
after call to append_arg_to_local d={'1': 1, '2': 2}
```

ltp23:
Thank you, Haowen!

> 1. For the 'Last 6 Month's Heifer Population Growth' value, if it is not a special request, I think we can remove it as an example because it can be misleading. Otherwise, we need to find ways to correctly calculate it.

I will remove this example. Maybe we can add another example later about this topic.

> 2. We also agreed to update the unit for methane. So instead of '90 to 30 days ago Methane Emission from Lactating Cows', I would write '90 to 30 days ago Methane Emission from Lactating Cows (grams)'.

I will update the title to include the units.

> 3. Calculations for milk yield need to be revisited, either it is daily milk yield per cow or  annual milk production per cow.

This example is leaning more toward the average daily milk yield per cow. I tweaked the column names a bit to improve clarity. What I wanted to do was to come up with an example that was fairly plausible and also able to demonstrate the use of cross-referencing between columns in the report.

ltp23:
> 1. Clearly mention the issues that this PR is aiming to address.

I will tag the issues in the descriptions.

> 2. Avoid packing other focal points in one PR (padding is out of the scope of this PR).

I will remove the code for padding for later.

> 3. Docstrings are not informative and clear enough.

I updated some docstring regarding the error handling.

> 4. There are code organization issues here: misplaced methods (things that belong to RG are placed in OM, static methods are used with no justification)

I added some more explanations to your specific questions. Since these are more organization and style questions without affecting functionality too much, they can be changed either here or later. Some choices I made were about convenience and accessibility to the existing code while others were more about coding style preferences.

RUFAS/output_manager.py
ltp23: Okay, I have moved the report generation code out of the OutputManager class and into the ReportGenerator class.

RUFAS/output_manager.py
```diff
@@ -113,11 +116,11 @@ def _pool_element_factory(self) -> pool_element_type:
...
-        self,
-        pool: Dict[str, pool_element_type],
-        key: str,
-        value: Any,
-        info_map: Dict[str, Any],
+            self,
+            pool: Dict[str, pool_element_type],
+            key: str,
+            value: Any,
+            info_map: Dict[str, Any],
```
Isgotkowitz, line 122: Do you have an autoformatter makes these changes? If so, could you let me know what it is please? I would like to propose one for the team to use so that we can start to make our code look really consistent.

RUFAS/output_manager.py
```diff
@@ -1,18 +1,21 @@
+# !/usr/bin/env python3
```
```suggestion
```
I think we are trying to eliminate headers/shebangs in the codebase.
Edit: with issue #910, all credit lines were removed from the codebase, but not necessarily all shebangs. It seems that they are used inconsistently across the RuFaS codebase. Is there a particular reason or benefit to having this one here? If not, I think it would be better to remove it.

RUFAS/report_generator.py
```diff
@@ -271,3 +296,339 @@ def _prepare_report_data(
             else:
                 report_data[key] = filtered_pool[key]["values"][slice_start:slice_end]
         return report_data
+
+    def handle_report_generation(self,
+                                 filter_content: Dict[str, Any],
+                                 filtered_pool: Dict[str, Dict[str, List[Any]]],
+                                 ) -> List[Dict[str, str | Dict[str, str]]]:
```
Isgotkowitz:
```suggestion
    def handle_report_generation(
            self,
            filter_content: Dict[str, Any],
            filtered_pool: Dict[str, Dict[str, List[Any]]],
    ) -> List[Dict[str, str | Dict[str, str]]]:
```

RUFAS/report_generator.py
```diff
@@ -271,3 +296,339 @@ def _prepare_report_data(
...
+            report_data = self._generate_single_report(data, filter_content)
+
+            self.reports[report_name] = {"values": report_data}
+        except (KeyError, ValueError) as e:
+            error_event = {
+                "error": "report_generation_error",
+                "message": f"Error generating report: {report_name}: {e}",
+                "info_map": info_map,
+            }
+            event_logs.append(error_event)
```
Isgotkowitz: Would it be possible/reasonably easy to separate the except statement into two, one for each error type? If so, I think that should be done here and the error message made a little more specific.

RUFAS/report_generator.py
```diff
@@ -271,3 +296,339 @@ def _prepare_report_data(
...
+            The aggregated report data as a list.
+
+        Raises
+        ------
+        ValueError
+            This type of error is raised in the following cases:
+                If the report data is empty.
+                If the type of aggregation is not supported. The supported types are: sum, average, product,
+                    subtraction, division, SD.
+                If there are no horizontal or vertical aggregation keys in the filter content.
```
Isgotkowitz:
```suggestion
        ValueError
            If the report data is empty.
            If the type of aggregation is not supported. The supported types are: sum, average, product,
                subtraction, division, SD.
            If there are no horizontal or vertical aggregation keys in the filter content.
```
This keeps the style consistent with what is requested in issue #1074

RUFAS/report_generator.py
```diff
@@ -271,3 +296,339 @@ def _prepare_report_data(
...
+
+    @staticmethod
+    def _prepare_report_data_with_constants(filtered_pool: Dict[str, Dict[str, List[Any]]],
+                                            filter_content: Dict[str, Any]) -> Dict[str, List[Any]]:
+        """
+        Processes and structures a filtered data pool for report generation.
+
+        This method organizes data from a filtered pool based on selected variables and slicing parameters.
+        It caters to different data structures within the pool, ensuring data is formatted appropriately
+        for report inclusion.
```
Isgotkowitz:
```suggestion
        Notes
        -----
        This method organizes data from a filtered pool based on selected variables and slicing parameters.
        It caters to different data structures within the pool, ensuring data is formatted appropriately
        for report inclusion.
```

RUFAS/report_generator.py
```diff
@@ -271,3 +296,339 @@ def _prepare_report_data(
...
+                }
+            ]
+
+        Parameters
+        ----------
+        report_data : Dict[str, List[Any]]
+            The data to which constants need to be added.
+        filter_content : Dict[str, Any]
+            A dictionary containing filter criteria, aggregation instructions, and scalar operation details.
+
```
Isgotkowitz, line 575: Raises section here.

RUFAS/report_generator.py
```diff
@@ -271,3 +296,339 @@ def _prepare_report_data(
...
+        return report_data
+
+    @staticmethod
+    def _add_constants_data(report_data: Dict[str, List[Any]],
+                            filter_content: Dict[str, Any]) -> None:
+        """
+        Add constants to the report data.
+
+        Notes
+        -----
```
Isgotkowitz:
```suggestion
        Examples
        --------
```

RUFAS/report_generator.py
```diff
@@ -271,3 +296,339 @@ def _prepare_report_data(
...
+        filter_content : Dict[str, Any]
+            The filter content for the report.
+
+        Returns
+        -------
+        str
+            If there is no `name` key in the filter content, the name is set to `untitled_<timestamp>`.
+            If the name is already present in the list of previously generated reports, the timestamp is appended
+            to the name to make it unique.
+            Otherwise, the name is set to the value of the `name` key in the filter content.
```
Isgotkowitz: I think the description of the returned value should be "The unique name for this report generated from its contents." and the current description of the returned value should be kept, but placed in a "Notes" section.

RUFAS/report_generator.py
```diff
@@ -271,3 +296,339 @@ def _prepare_report_data(
...
+
+        report_name = self._generate_unique_report_name(filter_content)
+
+        init_event = {
+            "log": "start_generate_report",
+            "message": f"Start generating report: {report_name}",
+            "info_map": info_map,
+        }
+        event_logs.append(init_event)
+        data = dict(filtered_pool)
```
Isgotkowitz: Why does this line need to be here? `filtered_pool` should already be a dictionary?

RUFAS/report_generator.py
```diff
@@ -133,11 +136,33 @@ def subtraction_aggregator(data: List[float]) -> float:
...
+    """
+
+    def __init__(self):
+        """
+        Initializes the ReportGenerator.
+        """
+
+        self.reports: Dict[str, Dict[str, List[Any]]] = {}
+
+    def clear_reports(self):
```
Isgotkowitz:
```suggestion
    def clear_reports(self) -> None:
```

RUFAS/report_generator.py
```diff
@@ -133,11 +136,33 @@ def subtraction_aggregator(data: List[float]) -> float:
...
+
     def generate_report(
-        self,
-        filtered_pool: Dict[str, Dict[str, List[Any]]],
-        filter_content: Dict[str, str | int | List[str]],
-    ) -> Dict[str, List[float]]:
+            self,
+            filtered_pool: Dict[str, Dict[str, List[Any]]],
+            filter_content: Dict[str, str | int | List[str]],
+    ) -> List[float]:
```
Isgotkowitz:
```suggestion
    ) -> Dict[str, List[float]]:
```
This should be changed back or the values being returned should be changed.

RUFAS/report_generator.py
```diff
@@ -271,3 +296,339 @@ def _prepare_report_data(
...
+        """
+
+        constant_config = filter_content.get("constants")
+        if not constant_config:
+            return
+
+        max_length = max([len(lst) for lst in report_data.values()])
+        for constant in constant_config:
+            if constant.get("name") in report_data:
+                raise ValueError(f"Constant name {constant.get('name')} already exists in report data.")
```
Isgotkowitz: Are constants always going to be numbers? If so, they should be validated somewhere - it is currently possible to make the ReportGenerator crash by passing certain non-number values. There are also some non-number values that it handles somewhat gracefully, like when a value is `None`/null - but it doesn't log any errors specific to this issue.

RUFAS/report_generator.py
```diff
@@ -271,3 +296,339 @@ def _prepare_report_data(
...
+            "constants": [
+                {
+                  "name": "Kilograms to Pounds",
+                  "value": 2.20462
+                },
+                {
+                  "name": "Pounds to Dollars",
+                  "value": 10
+                }
+            ]
```
Isgotkowitz: What would you think about changing the way constants are specified to be a dictionary with the key being the name and the value being the value? For example, the above would change to: 
```suggestion
            "constants": {
                "Kilograms to Pounds": 2.20462,
                "Pounds to Dollars": 10
            }
```
This would make the constants section a little more robust because 1) it would not be possible* to pass constants with the same name, and 2) it would not be possible* to forget to add a name or value to any constants (*these would be possible, except the JSON would be invalid - which I think would be a benefit but I can see why this could be considered a drawback).

The drawbacks include the above design being less future-proof (for example, if constants needed to have more elements than just "name" and "value", this design would have to be scrapped), and it also might be a lot of work at this point to change the structure. If either of these are the case, disregard this comment.

RUFAS/report_generator.py
```diff
@@ -271,3 +296,339 @@ def _prepare_report_data(
...
+            return [horizontal_aggregator(vertically_aggregated)]
+
+        if horizontal_agg_key:
+            return horizontally_aggregated
+
+        if vertical_agg_key:
+            return vertically_aggregated
+
+        raise ValueError(
+            f"Didn't find `horizontal_aggregation` or `vertical_aggregation` in {filter_content.get('name')}.")
```
Isgotkowitz: I think it would be beneficial to modify this function so that reports without vertical or horizontal aggregation are allowed through. This is the current behavior on main, and I find it extremely useful when working with reports.

Isgotkowitz (changes requested):
This is a great start on these issues Loi. Most of my comments are on minor things (formatting, docstrings, type annotations, etc).

I am requesting changes on 2 things:
1. There is almost no validation of constants right now, which makes it possible to crash the report generator or have it fail to report the reason why the report generation failed.
2. The current behavior on main allows a report to have no vertical or horizontal aggregation specified for a report. This is a very handy feature and it would be nice to have it preserved if possible.

PooyaHekmati:
@ltp23 what is the status here?

RUFAS/output_manager.py
ltp23: I think this one was auto-formatted by Pooya's IDE.

RUFAS/output_manager.py
ltp23: I think we can remove the shebang lines across the files because they are useful when we run them individually on a unix-based machine.

RUFAS/report_generator.py
ltp23: I can add the type of error in the log message. Because the way I handle those two error types is the same, I still have them combined though.

RUFAS/report_generator.py
ltp23: Okay. I will move the rules for naming into the Notes section.

RUFAS/report_generator.py
ltp23: This is to create a shallow copy so new columns won't be appended to the `filtered_pool`.

RUFAS/report_generator.py
ltp23: Right, the type here should be `Dict[str, List[float]]`.

RUFAS/report_generator.py
ltp23: I would assume the constants to be of type `int` or `float`. I'm not sure where the ideal place to validate these json specs files is. For now, I think we are doing the validation in the generator classes. 

Here, I will add a helper method to validate the `name` and `value` fields for each constant.

RUFAS/report_generator.py
ltp23: Right, I do see the merits and drawbacks of either approach. I like the dictionary approach for its succinctness. The array approach allows more info for each constant. Since the code for managing constants is isolated to this part of the code, we can change it later when we know more.

ltp23: I decided to change to your suggestion because with the current use case, the dictionary format is easier for the user and it didn't take too long to make the update.

RUFAS/report_generator.py
ltp23: I'm sorry about leaving out this existing functionality along with the suffixes for different aggregation types. I have put them back. 

Now with the additional feature of referencing previously generated columns, one thing to be careful of when using this is to not forget the appropriate suffixes like `_<column_name>`, `_hor_agg`, `_hor_ver_agg`, etc.

RUFAS/output_manager.py
Isgotkowitz: Sounds good. I will make an issue about this

Isgotkowitz:
@ltp23 would it be possible to change the behavior when no name is present? I think the prepending of `untitled_<number>-<timestamp>`, or just `_`, is unnecessary in many cases.

For example, reporting the time. It can be done effectively with this filter:
```json
{
  "multiple": [
    {
      "name": "",
      "filters": ["Time\\..*"]
    }
  ]
}
```
But in the behavior of Report Generator on `main`, it results in this report:
|_Time.day|_Time.year|_Time.calendar_year|
|-----------|-----------|----------------------|
|...|...|...|

And with the current behavior of Report Generator in this PR it results in
|untitled_11-Jan-2024_Thu_09-50-44.081395_Time.day|untitled_11-Jan-2024_Thu_09-50-44.081395_Time.year|untitled_11-Jan-2024_Thu_09-50-44.081395_Time.calendar_year|
|-----------|-----------|----------------------|
|...|...|...|

When it would probably be cleanest just to have the report be
|Time.day|Time.year|Time.calendar_year|
|-----------|-----------|----------------------|
|...|...|...|

This would obviously mean more work on the code though, and since it is not essential for addressing the issues in this PR it may be better to leave changing this behavior for another PR.

RUFAS/output_manager.py
```diff
@@ -797,54 +768,43 @@ def save_results(
...
+                        filter_content["filters"], filter_file
+                    )
                 if exclude_info_maps:
                     filtered_pool = self._exclude_info_maps(filtered_pool)
 
                 if filter_file.startswith(
-                    self.__supported_filter_types_prefixes["report"]
+                        self.__supported_filter_types_prefixes["report"]
                 ):
-                    self.add_log(
```
PooyaHekmati, line 827: don't remove this log.

RUFAS/output_manager.py
```diff
@@ -797,54 +768,43 @@ def save_results(
...
-                        for key, value in report.items():
-                            column_name = f"{report_name}_{key}"
-                            reports[
-                                column_name
-                                if column_name not in reports.keys()
-                                else f"{column_name} {self._get_timestamp(True)}"
-                            ] = {"values": value}
-                    except (KeyError, ValueError) as e:
-                        self.add_error("report generation error", str(e), info_map)
+                    log_pool = report_generator.handle_report_generation(filter_content, filtered_pool)
```
PooyaHekmati: why is `generate_report` renamed to `handle_report_generation`?

RUFAS/report_generator.py
```diff
@@ -133,10 +136,32 @@ def subtraction_aggregator(data: List[float]) -> float:
...
+    def clear_reports(self) -> None:
+        """
+        Clears the reports generated by the ReportGenerator.
+
+        Returns
+        -------
+        None
+        """
+
+        self.reports = {}
```
PooyaHekmati:
```suggestion
        self.reports: Dict[str, Dict[str, List[Any]]] = {}
```

RUFAS/report_generator.py
```diff
@@ -133,10 +136,32 @@ def subtraction_aggregator(data: List[float]) -> float:
...
+
     def generate_report(
```
PooyaHekmati, line 159: This method is now obsolete, why is it still in the codebase?

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
             else:
                 report_data[key] = filtered_pool[key]["values"][slice_start:slice_end]
         return report_data
+
+    def handle_report_generation(
```
PooyaHekmati: I like `generate_report` as the name better, it is shorter and less ambiguous.

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+            self,
+            filter_content: Dict[str, Any],
+            filtered_pool: Dict[str, Dict[str, List[Any]]],
+    ) -> List[Dict[str, str | Dict[str, str]]]:
+        """
+        Generates a single report based on filtered data and aggregation criteria.
```
PooyaHekmati: This can't be true because in the code you are calling `_generate_single_report` plus doing some other stuff. Therefore, this method is doing more than generating a single report.

ltp23:
Hi @Isgotkowitz, I simplified the column names as you suggested. I included your example in the example json file. Thank you for your suggestion.

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+            such as 'name', 'filters', 'cross_references', and aggregation instructions.
+
+        filtered_pool : Dict[str, Dict[str, List[Any]]]
+            The data pool from which reports are generated.
+
+        Returns
+        -------
+        List[Dict[str, str | Dict[str, str]]]
+            A list of log events.
+        """
```
PooyaHekmati:
```suggestion
        """
    Generates a report based on filtered data and aggregation criteria specified in filter_content. 
    This method logs the process of report generation and incorporates cross-references if provided.
    
    Note
    -----
    This method does not explicitly raise exceptions; it handles potential issues internally 
    and logs them.

    Parameters
    ----------
    filter_content : Dict[str, Any]
        Configuration for the report. This includes keys like 'name', 'filters', 'cross_references', 
        and aggregation instructions. The structure should align with the expected format for report generation.

    filtered_pool : Dict[str, Dict[str, List[Any]]]
        The data pool from which the report is generated. This should follow the structure of OutputManager's 
        variables pool, where each key represents a data category with its associated data.

    Returns
    -------
    List[Dict[str, str | Dict[str, str]]]
        A list of log events generated during the report creation process. Each log event is a dictionary 
        containing information such as 'log' type, 'message', and an 'info_map' with additional details.
```

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+
+        event_logs: List[Dict[str, str | Dict[str, str]]] = []
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self.handle_report_generation.__name__,
+        }
+
+        report_name = self._generate_unique_report_name(filter_content.get("name"))
+
+        init_event = {
```
PooyaHekmati:
```suggestion
        init_event_log = {
```

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+            report_data = self._generate_single_report(data, filter_content)
+
+            for col, values in report_data.items():
+                column_name = self._generate_unique_report_name(f"{report_name}_{col}"
+                                                                if len(report_name) > 0 else col)
+                self.reports[column_name] = {"values": values}
+
+        except (KeyError, ValueError) as e:
+            error_type = e.__class__.__name__
+            error_event = {
```
PooyaHekmati:
```suggestion
            error_event_log = {
```

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+
+        report_name = self._generate_unique_report_name(filter_content.get("name"))
+
+        init_event = {
+            "log": "start_generate_report",
+            "message": f"Start generating report: {report_name}",
+            "info_map": info_map,
+        }
+        event_logs.append(init_event)
+        data = dict(filtered_pool)
```
PooyaHekmati:
```suggestion
        data_shallow_copy = dict(filtered_pool)
```

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+            error_event = {
+                "error": "report_generation_error",
+                "message": f"Error generating report ({report_name}) => {error_type}: {e}",
+                "info_map": info_map,
+            }
+            event_logs.append(error_event)
+
+        return event_logs
+
+    def _generate_unique_report_name(self, report_name: str | None) -> str:
```
PooyaHekmati: `ensure_unique_report_name_with_timestamp` sounds more descriptive

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+            report_name = str(report_name)
+
+        if report_name in self.reports:
+            report_name = f"{report_name}_{Utility.get_timestamp(True)}"
+
+        return report_name
+
+    def _check_for_missing_references(self, references: List[str]) -> None:
+        """
+        Checks if all the referenced reports are present.
```
PooyaHekmati: are present in where?

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+        ------
+        KeyError
+            If any of the report references are missing.
+        """
+
+        missing_references = [ref for ref in references if ref not in self.reports]
+        if missing_references:
+            raise KeyError(f"Missing referenced reports: {', '.join(missing_references)}")
+
+    @staticmethod
```
PooyaHekmati: I don't see why this needs to be static.

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+        ValueError
+            If there is an error preparing the report data.
+            If the report data is empty.
+            If the type of horizontal or vertical aggregation is not supported.
+        """
+
+        try:
+            report_data = ReportGenerator._prepare_report_data_with_constants(filtered_pool, filter_content)
+        except ValueError as e:
+            raise ValueError(f"Error preparing report data => {e}")
```
PooyaHekmati: Don't format it.
```suggestion
        except ValueError:
            raise
```

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+
+        if not report_data:
+            raise ValueError(
+                f"filter {filter_content.get('filters')} in {filter_content.get('name')} led to empty report data."
+            )
+
+        try:
+            horizontal_agg_key, vertical_agg_key = ReportGenerator._validate_aggregation_keys(filter_content)
+        except ValueError as e:
+            raise ValueError(f"Error validating aggregation keys => {e}")
```
PooyaHekmati: Don't format it.
```suggestion
        except ValueError:
            raise
```

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+            vertical_aggregator = AGGREGATION_FUNCTIONS.get(vertical_agg_key)
+            vertically_aggregated = ReportGenerator._apply_vertical_aggregation(report_data, vertical_aggregator)
+
+        aggregate_report = ReportGenerator._package_report_data(horizontally_aggregated, vertically_aggregated,
+                                                                filter_content)
+
+        return aggregate_report if aggregate_report else report_data
+
+    @staticmethod
+    def _package_report_data(
```
PooyaHekmati:
```suggestion
    def _combine_aggregated_report_data(self,
```

RUFAS/output_manager.py
```diff
@@ -755,13 +726,13 @@ def _filter_variables_pool(
...
+            self,
+            save_path: Path,
+            filters_dir_path: Path,
+            exclude_info_maps: bool,
+            produce_graphics: bool,
+            graphics_dir: Path,
+            csv_dir: Path
     ) -> None:
         """
         Reads a text file containing a list of keys and filters the variables pool by those keys.
```
Isgotkowitz: Seems like this function is used to parse all kinds of filter files, not exclusively filters files in text format. This docstring should be updated to reflect and any other things in it that are out of date.

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+            horizontally_aggregated: List[float] | None,
+            vertically_aggregated: List[float] | None,
+            filter_content: Dict[str, Any]
+    ) -> Dict[str, List[float]] | None:
+        """
+        Packages the horizontally and vertically aggregated data into a dictionary.
```
PooyaHekmati:
```suggestion
        Combines horizontally and vertically aggregated data based on specified aggregation criteria from filter_content.
```

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+        if horizontal_agg_key:
+            return {"hor_agg": horizontally_aggregated}
+
+        if vertical_agg_key:
+            return {"ver_agg": vertically_aggregated}
+
+        return None
+
+    @staticmethod
+    def _validate_aggregation_keys(filter_content: Dict[str, Any]) -> tuple[str | None, str | None]:
```
PooyaHekmati:
```suggestion
    def extract_and_check_aggregation_keys(self, filter_content: Dict[str, Any]) -> tuple[str | None, str | None]:
```

RUFAS/output_manager.py
```diff
@@ -797,54 +768,43 @@ def save_results(
...
         list_of_filter_files = self._list_filter_files_in_dir(filters_dir_path)
+        report_generator = ReportGenerator()
         for filter_file in list_of_filter_files:
             info_map["filter file"] = filter_file
             input_path = os.path.join(filters_dir_path, filter_file)
             filter_contents = self._load_filter_file_content(input_path)
-            reports: Dict[str: Dict[str: List[Any]]] = {}
+            self.add_log("init_report_generation",
+                         f"Generating report(s) for file: {filter_file}",
+                         info_map)
```
Isgotkowitz: I think this log should be relocated to the if statement which calls `handle_report_generation()`. As it is now, the OutputManager will log that it is generating a report for a filter file even if that filter file doesn't start with `report`.

RUFAS/output_manager.py
```diff
@@ -797,54 +768,43 @@ def save_results(
...
                     )
                     continue
-                filtered_pool = self._filter_variables_pool(
-                    filter_content["filters"], filter_file
-                )
+
+                filtered_pool:  Dict[str, OutputManager.pool_element_type] = {}
+                if "filters" in filter_content.keys():
+                    filtered_pool = self._filter_variables_pool(
+                        filter_content["filters"], filter_file
```
Isgotkowitz, line 804: PyCharm is flagging the following type mismatch to me on this line: "Expected type 'list[str]', got 'str | int' instead"

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+        -------
+        tuple[str | None, str | None]
+            The horizontal and vertical aggregation keys.
+
+        Raises
+        ------
+        ValueError
+            If the type of horizontal or vertical aggregation is not supported. The supported types are:
+            sum, average, product, subtraction, division, SD.
+        """
```
PooyaHekmati:
```suggestion
       """
    Extracts horizontal and vertical aggregation keys from the filter content and validates them against
    supported aggregation types.

    Parameters
    ----------
    filter_content : Dict[str, Any]
        A dictionary containing filter criteria, aggregation instructions, and scalar operation details.
        It should include keys for 'horizontal_aggregation' and 'vertical_aggregation' if applicable.

    Returns
    -------
    tuple[str | None, str | None]
        A tuple containing the horizontal and vertical aggregation keys, respectively.
        Returns None for each key if it is not specified or if it is not among the supported types.

    Raises
    ------
    ValueError
        Raised if the specified horizontal or vertical aggregation type is not supported. Supported types are defined in the AGGREGATION_FUNCTIONS dictionary.
    """
```

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+            raise ValueError(f"Unsupported horizontal aggregation type: {horizontal_agg_key}")
+
+        vertical_agg_key = filter_content.get("vertical_aggregation")
+        if vertical_agg_key and vertical_agg_key not in AGGREGATION_FUNCTIONS:
+            raise ValueError(f"Unsupported vertical aggregation type: {vertical_agg_key}")
+
+        return horizontal_agg_key, vertical_agg_key
+
+    @staticmethod
+    def _apply_horizontal_aggregation(report_data: Dict[str, List[float]], loop_list: List[str],
```
PooyaHekmati:
```suggestion
    def _apply_horizontal_aggregation(self, report_data: Dict[str, List[float]], loop_list: List[str],
```

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+        ------
+        ValueError
+            If the data to be aggregated has different lengths.
+        """
+
+        lengths = [len(report_data[key]) for key in loop_list]
+        if len(set(lengths)) != 1:
+            raise ValueError("Can't aggregate data with different lengths")
+        max_length = max(lengths)
+        aggregated_data = []
```
PooyaHekmati:
```suggestion
        aggregated_data: List[float]= []
```

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+        for i in range(max_length):
+            temp_data = [report_data[key][i] for key in loop_list]
+            filtered_temp_data = list(filter(None.__ne__, temp_data))
```
PooyaHekmati:
```suggestion
            non_null_data_points = list(filter(None.__ne__, temp_data))
```

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+            aggregated_data.append(aggregator(filtered_temp_data))
+        return aggregated_data
+
+    @staticmethod
+    def _apply_vertical_aggregation(report_data: Dict[str, List[float]],
```
PooyaHekmati: same comments as `_apply_horizontal_aggregation` apply.

RUFAS/report_generator.py
```diff
@@ -1,4 +1,7 @@
+from __future__ import annotations
+
```
PooyaHekmati:
```suggestion
```
As per setup.py, RUFAS is intended to run using Python 3.10+.

PooyaHekmati (changes requested):
I could not review the whole PR as I ran out of time and energy. AVOID submitting massive PRs like this.

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+            "info_map": info_map,
+        }
+        event_logs.append(init_event)
+        data = dict(filtered_pool)
+
+        try:
+            if "cross_references" in filter_content.keys():
+                self._check_for_missing_references(filter_content["cross_references"])
+                cross_reference_data = {ref: self.reports[ref] for ref in filter_content["cross_references"]}
+                data.update(cross_reference_data)
```
PooyaHekmati:
```suggestion
                cross_reference_data.update(data)
```
with this change, do you still need the shallow copy?

RUFAS/report_generator.py
Isgotkowitz: What about checking if there is any horizontal or vertical agg keys at all, and if not then returning `report_data` immediately? I.e. checking the easiest return path first. That would allow the `_package_report_data()` method to be simplified.

RUFAS/report_generator.py
```diff
@@ -271,3 +296,479 @@ def _prepare_report_data(
...
+        return event_logs
+
+    def _generate_unique_report_name(self, report_name: str | None) -> str:
+        """
+        Make the given report name unique, if it is not, by appending a timestamp to it.
+
+        Notes
+        -----
+        The rules for generating the name are as follows:
+        1. If the report name is None or empty, the name is set to `untitled_<timestamp>`.
```
Isgotkowitz:
```suggestion
        1. If the report name is None, the name is set to `untitled_<timestamp>`.
```

Isgotkowitz (approved):
Addresses issues well, so I am approving. There are a few suggestions/comments that should be addressed before merging, but should all be simple fixes.

RUFAS/report_generator.py
ltp23: What I could improve here is moving the validation step for agg keys up before the report data prep step. 

I think the `_package_report_data()` is alright because I have different conditional checks. I could pass the two agg keys as direct arguments instead of through the `filter_content`.

RUFAS/output_manager.py
ltp23: This one is still somewhat unpredictable. Originally, we expected the type for each filter content object to be `Dict[str, str | int]`. Now the type of the values has evolved since, so it is no longer `str | int`.

RUFAS/output_manager.py
ltp23: I think the location for `add_log()` is alright here because this message is for the whole filter file and not just an individual filter content. But I can add a check and only log the message when it is a report filter file.

RUFAS/report_generator.py
ltp23: This is optional. The code should still work without the shallow copy. Because the `cross_reference_data` variable is initialized inside the `if` clause, I don't want to use it outside the scope.

RUFAS/report_generator.py
ltp23: Okay. I can remove this.

RUFAS/report_generator.py
Isgotkowitz: > What I could improve here is moving the validation step for agg keys up before the report data prep step.
> I could pass the two agg keys as direct arguments instead of through the filter_content.

I think both of these changes improve the code, but they are not what I am suggesting.

I am suggesting changing this:
```python
        horizontally_aggregated = None
        vertically_aggregated = None

        if horizontal_agg_key:
            horizontal_aggregator = AGGREGATION_FUNCTIONS.get(horizontal_agg_key)
            loop_list = filter_content.get("horizontal_order", report_data.keys())
            horizontally_aggregated = self._apply_horizontal_aggregation(report_data, loop_list,
                                                                         horizontal_aggregator)

        if vertical_agg_key:
            vertical_aggregator = AGGREGATION_FUNCTIONS.get(vertical_agg_key)
            vertically_aggregated = self._apply_vertical_aggregation(report_data, vertical_aggregator)

        aggregate_report = self._combine_aggregate_report_data(horizontally_aggregated,
                                                               vertically_aggregated,
                                                               filter_content)

        return aggregate_report if aggregate_report else report_data
```
to this:
```python
        if not horizontal_agg_key and not vertical_agg_key:
            return report_data

        horizontally_aggregated = None
        vertically_aggregated = None

        if horizontal_agg_key:
            horizontal_aggregator = AGGREGATION_FUNCTIONS.get(horizontal_agg_key)
            loop_list = filter_content.get("horizontal_order", report_data.keys())
            horizontally_aggregated = self._apply_horizontal_aggregation(report_data, loop_list,
                                                                         horizontal_aggregator)

        if vertical_agg_key:
            vertical_aggregator = AGGREGATION_FUNCTIONS.get(vertical_agg_key)
            vertically_aggregated = self._apply_vertical_aggregation(report_data, vertical_aggregator)

        aggregate_report = self._combine_aggregate_report_data(horizontally_aggregated,
                                                               vertically_aggregated,
                                                               filter_content)

        return aggregate_report
```
I tried this and it doesn't seem to break the code - but it is up to you whether or not this is a worthwhile change to make.

RUFAS/report_generator.py
```diff
@@ -133,95 +134,369 @@ def subtraction_aggregator(data: List[float]) -> float:
...
+            The unique name for the report.
         """
-        Generates a report based on filtered data and aggregation criteria.
 
-        This method processes filtered data and performs specified horizontal and/or vertical aggregations
-        based on the instructions provided in filter_content.
+        if report_name is None:
+            report_name = f"untitled_{Utility.get_timestamp(True)}"
+
+        if not isinstance(report_name, str):
```
PooyaHekmati: `report_name` is typed as `str`, how can it not be str?!

RUFAS/report_generator.py
```diff
@@ -133,95 +134,369 @@ def subtraction_aggregator(data: List[float]) -> float:
...
+
+        try:
+            horizontal_agg_key, vertical_agg_key = self._extract_and_check_aggregation_keys(filter_content)
+        except ValueError:
+            raise
+
+        try:
+            report_data = self._prepare_report_data_with_constants(filtered_pool, filter_content)
+        except ValueError:
+            raise
```
PooyaHekmati:
```suggestion
        try:
            horizontal_agg_key, vertical_agg_key = self._extract_and_check_aggregation_keys(filter_content)
            report_data = self._prepare_report_data_with_constants(filtered_pool, filter_content)
        except ValueError:
            raise
```

RUFAS/report_generator.py
```diff
@@ -133,95 +134,369 @@ def subtraction_aggregator(data: List[float]) -> float:
...
+
+        Returns
+        -------
+        List[float]
+            The vertically aggregated data as a list.
+        """
+
+        aggregated_data = []
+        for _, data in report_data.items():
+            filtered_data = list(filter(None.__ne__, data))
```
PooyaHekmati: non_null_data_points

RUFAS/report_generator.py
```diff
@@ -133,95 +134,369 @@ def subtraction_aggregator(data: List[float]) -> float:
...
+        aggregator : Callable[[List[float], float, str], float]
+            The aggregation function to be used.
+
+        Returns
+        -------
+        List[float]
+            The vertically aggregated data as a list.
+        """
+
+        aggregated_data = []
```
PooyaHekmati: add type hint

RUFAS/report_generator.py
```diff
@@ -270,4 +546,99 @@ def _prepare_report_data(
...
                 report_data[key] = filtered_pool[key]["values"][slice_start:slice_end]
+
+        try:
+            self._add_constants_data(report_data, filter_content)
+        except ValueError:
+            raise
+
         return report_data
+
+    def _add_constants_data(
```
PooyaHekmati:
```suggestion
    def _add_constants_to_report_data(
```

RUFAS/report_generator.py
```diff
@@ -270,4 +546,99 @@ def _prepare_report_data(
...
+        Parameters
+        ----------
+        report_data : Dict[str, List[Any]]
+            The data to which constants need to be added.
+        filter_content : Dict[str, Any]
+            A dictionary containing filter criteria, aggregation instructions, and scalar operation details.
+
+        Returns
+        -------
+        None
```
PooyaHekmati:
```suggestion
```

RUFAS/report_generator.py
```diff
@@ -270,4 +546,99 @@ def _prepare_report_data(
...
+
+        Raises
+        ------
+        ValueError
+            If the name or value of any constant is not valid.
+        """
+
+        constant_config = filter_content.get("constants")
```
PooyaHekmati:
```suggestion
        constants_config = filter_content.get("constants")
```

RUFAS/report_generator.py
```diff
@@ -270,4 +546,99 @@ def _prepare_report_data(
...
+        for name, value in constant_config.items():
+            report_data[name] = [value] * max_length
+
+    def _validate_constants(
+            self,
+            report_data: Dict[str, List[Any]],
+            constant_config: Dict[str, int | float]
+    ) -> None:
+        """
+        Validate constants before adding them to report data.
```
PooyaHekmati: As far as this method is concerned, adding constants to the report is not a thing. Instead, briefly explain what validation means in this context.

RUFAS/report_generator.py
```diff
@@ -270,4 +546,99 @@ def _prepare_report_data(
...
+            raise
+
+        max_length = max([len(lst) for lst in report_data.values()])
+        for name, value in constant_config.items():
+            report_data[name] = [value] * max_length
+
+    def _validate_constants(
+            self,
+            report_data: Dict[str, List[Any]],
+            constant_config: Dict[str, int | float]
```
PooyaHekmati:
```suggestion
            constants_config: Dict[str, int | float]
```

PooyaHekmati (changes requested):
In addition to the new comments, there are unresolved comments from the last review.

RUFAS/report_generator.py
PooyaHekmati: I don't understand what you mean by _optional_. A function has to have no side effects and avoid any unnecessary operations. If swapping the operands saves a shallow copy operation while ensuring there are no side effects, there is no reason to do the extra work of copying.

RUFAS/output_manager.py
ltp23: This is okay now. I removed the older version.

RUFAS/report_generator.py
ltp23: I changed the name `_generate_single_report` to `_generate_report_data`.

RUFAS/report_generator.py
ltp23: I see. I can add this check to exit early. Thanks, Ed.

RUFAS/report_generator.py
ltp23: I will remove this check because before I allowed the name to not be of type str and then converted it to str.

RUFAS/report_generator.py
```diff
@@ -133,95 +134,365 @@ def subtraction_aggregator(data: List[float]) -> float:
...
+        if missing_references:
+            raise KeyError(f"Missing referenced reports: {', '.join(missing_references)}")
+
+    def _generate_report_data(
+            self,
+            filtered_pool: Dict[str, Dict[str, List[Any]]],
+            filter_content: Dict[str, Any]
+    ) -> Dict[str, List[Any]]:
+        """
+        Generates a report based on filtered data and aggregation criteria, including scalar operations and constants.
```
PooyaHekmati: So, the name of the function states that we generate `report_data` here, while the docstring states that we generate a `report` here. What's going on? Which one are we doing? How can `report` and `report_data` be the same? They are inherently different things. What do you think about `perform_aggregation` or something like that for its name?

RUFAS/report_generator.py
PooyaHekmati: why do we need to emphasize on report being `single` here?

PooyaHekmati (changes requested):
Please address all comments before requesting a review:
https://github.com/RuminantFarmSystems/MASM/pull/1042/files#r1449241431
https://github.com/RuminantFarmSystems/MASM/pull/1042#discussion_r1450771190
https://github.com/RuminantFarmSystems/MASM/pull/1042#discussion_r1450774937

Also, the name and docstring of `generate_report` and `_generate_report_data` need to be improved. Probably, removing the word `single` from `generate_report`'s docstring and renaming `_generate_report_data` to something that better captures its functionality would fix the issue.

RUFAS/report_generator.py
ltp23: I added the following to the docstring

```
Validate each name-value pair in the constants config dictionary before adding
        each constant as a column to the report data.
```

PooyaHekmati: That doesn't address the comment. The comment is saying that this method has nothing to do with adding constants to report data, so there is no reason to mention that in its docstring. Instead, docstring has to elaborate what does validation mean here, which still doesn't.

RUFAS/report_generator.py
ltp23: If I adjust it like you suggested, it would look like this

```
if "cross_references" in filter_content.keys():
    self._check_for_missing_references(filter_content["cross_references"])
    cross_reference_data = {ref: self.reports[ref] for ref in filter_content["cross_references"]}
    cross_reference_data.update(filtered_pool)

report_data = self._generate_report_data(cross_reference_data, filter_content)
```

This should work and the functionality would be the same. My point was the `cross_reference_data` variable was defined inside the if-clause. But this is not a problem if you are more concerned about the extra copying. I will use your version.

PooyaHekmati: I still don't understand the hesitation and resistance to the change. If there are some objective problems with it and it doesn't work as intended, please elaborate. Otherwise, the suggested change saves an unnecessary operation.

RUFAS/report_generator.py
ltp23: The names are confusing for me too. I can change to `perform_aggregations`.

ltp23:
> why do we need to emphasize on report being single here?

What I mean here is the RG generates an individual report for each filter-content object. I changed the docstring to this 'Generates an individual report specified by the given filter content.'

RUFAS/report_generator.py
```diff
@@ -133,95 +134,366 @@ def subtraction_aggregator(data: List[float]) -> float:
...
-        self,
-        filtered_pool: Dict[str, Dict[str, List[Any]]],
-        filter_content: Dict[str, str | int | List[str]],
-    ) -> Dict[str, List[float]]:
+            self,
+            filter_content: Dict[str, Any],
+            filtered_pool: Dict[str, Dict[str, List[Any]]],
+    ) -> List[Dict[str, str | Dict[str, str]]]:
+        """
+        Generates an individual report specified by the given filter content.
```
PooyaHekmati:
```suggestion
        Generates a report specified by the given filter content.
```

RUFAS/report_generator.py
```diff
@@ -133,95 +134,366 @@ def subtraction_aggregator(data: List[float]) -> float:
...
+        if missing_references:
+            raise KeyError(f"Missing referenced reports: {', '.join(missing_references)}")
+
+    def _perform_aggregations(
+            self,
+            filtered_pool: Dict[str, Dict[str, List[Any]]],
+            filter_content: Dict[str, Any]
+    ) -> Dict[str, List[Any]]:
+        """
+        Generates a report based on filtered data and aggregation criteria, including scalar operations and constants.
```
PooyaHekmati:
```suggestion
        Fetches aggregation keys from the filter content and applies aggregation to the data.
```

PooyaHekmati (changes requested):
two edits to docstrings plus two unresolved comments:
https://github.com/RuminantFarmSystems/MASM/pull/1042/files#r1449241431
https://github.com/RuminantFarmSystems/MASM/pull/1042#discussion_r1450950546

RUFAS/report_generator.py
ltp23: I updated it to this to explain what rules are being checked. Or maybe this is too detailed.

```
Validates the names and values of the constants in the constants_config.

This method checks that each constant's name is not None, a non-empty string, and
is unique compared to the names of all the existing reports. It also verifies that
each constant's value is a number (int or float) and not None.

```

RUFAS/output_manager.py
ltp23: I haven't removed this log. There were two places I added init log comments.

1. In the OutputManager class, around lines 779-784

```
if filter_file.startswith(
        self.__supported_filter_types_prefixes["report"]
):
    self.add_log("init_report_generation",
                 f"Generating report for file: {filter_file}",
                 info_map)
```

2. Then for each filter-content, there is another init log. This is done in the ReportGenerator class. (Around lines 190-195)

```
init_event_log = {
    "log": "start_generate_individual_report",
    "message": f"Start generating individual report: {individual_report_name}",
    "info_map": info_map,
}
event_logs.append(init_event_log)
```

PooyaHekmati (approved):

PR #1048, opened 2023-12-14
Filter Graph Plot Data By Regex

ew3361zh:
PR to address issue #899.

To accomplish this, the PR:
1. Extracts the regex filtering functionality from OM's `_filter_variables_pool()` method to a new function in util.py.
2. Incorporates this new function back into OM's `_filter_variables_pool()` method to function as before, as well as into `ReportGenerator` and `GraphGenerator` in their respective data prep functions.
3. Updates `prepare_plot_data()` method in GG to now incorporate filtering by exclusion as well as graphing all variables in data set when no variables are listed in graph filter file.
4. Updates `generate_graph()` method to auto-generate graph-filter file legend when legend is left blank or is empty.
5. Updates unit testing.

Note:
- [x] This would add an optional `filter_by_exclusion` field to filter files for both graph generator and report generator. The default value would be False unless the variables field in the filter file is empty or missing in which case the value will be switched to True so that all data sent will be graphed. If this PR is approved with this field, I will add that info to the wikis for both graph generator and report generator.
- [x] To graph all variables without specifying any variables, the user can enter the generic regex pattern for "any character, 0-infinity times" which is `.*`. The resulting filter file would have this entry: 
```
"variables": [".*"]
```


JoeWaddell:
Testing this out, and the code looks nice and clean.

However, it doesn't seem to tackle #899 , at least not as I imagined (which is likely different from how I described it!). I'll try clarify my thoughts here.

Currently, if information is stored in a dictionary like this (see `AnimalManager._calc_ration_at_interval.ration_per_animal_for_pen_3`): 
image


Then the graph filter needs to be written like this: 
image

However, note that this dictionary is unique to not only each animal combination (see the `default_feed.json`), but (usually/often) for each farm. So if a user (for example the pilot testers) are comparing multiple farms, they need 4X filter files to plot data for all combinations. Likewise, any mismatch in the keys results in no figure. Also, since new pens can be filled dynamically with any animal type, it becomes impossible to predict which combination should be used to make the list of variables/legend fields.

Hence the idea that the information that comes from the variables and legend fields currently filled in manually in the filter could be simply populated from the keys in the dictionary.

In addition, the suggestion of using an exclusion filter would be to allow graph generator to plot all potential variables (e.g. any key in the dict) if *not* in the regex list of excluded names (in the above example, for instance, the user could exclude `dry_matter_intake_total`, and it would plot all other keys without needing to list them in the variables or legend fields).

Note: for my uses, this is not super-highly-urgent, as it's mostly solving an annoyance in making and proofing dozens of graph filters in cases when comparing multiple farms, or testing multiple diets. However, @jadamchick may be able to help prioritize this issue (specifically #899 ), since it might not actually be a concern (instead looking at the data in excel, for instance).  Likewise when we compare the diets Edward has created, it would require a filter for all 144!

jadamchick:
Thanks for the tag, Joe. It would be really helpful to have this function in place asap to help with reporting feed emissions by ingredient and by animal group for pilot farms. As it stands now, we need to write individual filters for each feed ingredient by animal group (i.e. if a feed is fed to 3 different animal groups, need to have 3 separate filters for it). If this can help to cut down that volume/labor, it would be great!

I'll quote here what I commented in the Feed Emissions PR (#1057) -- please let me know if you have more questions. 

I am able to use the report generator to slice and aggregate the feed emissions data. It is a little fiddly (need to specify each feed individually and make sure it is labeled in the title in order to know in the output which is which). Copying my example filter here FYI. It would be slick to be able to do this a little more smoothly, but this is functional for what we need.

```
{
    "multiple":[
        {
            "name": "footprint by ingredient - 51 lact",
            "filters": [
                "AnimalModuleReporter.report_daily_feed_emissions.pen_3_animal_LAC_COW_feed_emissions"
            ],
            "variables": "51", 
            "slice_start": -365, 
            "vertical_aggregation": "sum"
        },
        {
            "name": "footprint by ingredient - 104 lact",
            "filters": [
                "AnimalModuleReporter.report_daily_feed_emissions.pen_3_animal_LAC_COW_feed_emissions"
            ],
            "variables": "104", 
            "slice_start": -365,
            "vertical_aggregation": "sum"
        },
        {
            "name": "footprint by ingredient - total lact",
            "filters": [
                "AnimalModuleReporter.report_daily_feed_emissions.pen_3_animal_LAC_COW_feed_emissions"
            ],
            "variables": "feed_emissions_total", 
            "slice_start": -365,
            "vertical_aggregation": "sum"
        }
    ]
} 
```

JoeWaddell (approved):
This works great! As expected, and this makes it much easier to use any feed file with a single filter, now. The code is straightforward, as well. LGTM

RUFAS/graph_generator.py
```diff
@@ -238,25 +242,27 @@ def _prepare_plot_data(self, filtered_pool: Dict[str, Dict[str, List[Any]]],
...
+                if not filtered_data:
+                    log_pool.append({"error": f"Can't plot {title} data set",
+                                     "message": "No variables found in data provided.",
+                                     "info_map": info_map})
+                else:
+                    for filtered_key, filtered_value in filtered_data.items():
+                        if filtered_key in prepared_pool:
+                            prepared_pool[filtered_key].extend(filtered_value)
+                        else:
+                            prepared_pool[filtered_key] = filtered_value
```
allisterakun: I think we should check for `if not filtered_data: ` first, before handling the `non_int_float` data. Although I don't think this would be an issue, I do think it would be the correct logic to make sure the `filtered_data` actually has a value before checking the special cases like `non_int_float` data.

So the proposed logic would be:
```python
filtered_data = Utility.filter_pool(data_dict, selected_variables, filter_by_exclusion)
if not filtered_data:
    log_pool.append({"error": f"Can't plot {title} data set",
                     "message": "No variables found in data provided.",
                     "info_map": info_map})
else:
    non_int_float_keys = [
        key for key, value in filtered_data.items()
        if not (isinstance(value, (int, float)) or
                  (isinstance(value, list) and all(isinstance(item, (int, float)) for item in value)))
                  ]
    for key in non_int_float_keys:
        log_pool.append({"error": f"Can't plot {title} data set",
                         "message": f"{key} key contains data that is non-numerical and can't be graphed.",
                         "info_map": info_map})
    for filtered_key, filtered_value in filtered_data.items():
        if filtered_key in prepared_pool:
            prepared_pool[filtered_key].extend(filtered_value)
        else:
            prepared_pool[filtered_key] = filtered_value
```

allisterakun (approved):
Looks good, Niko! Just a minor comment,

RUFAS/graph_generator.py
ew3361zh: Nice catch. That makes sense to me, I will make that change.

PR #1050, opened 2023-12-14
Get data keys from Input Manager by property

### LINKED ISSUE #1006: Move functionality for getting data by property type from Input Manager from `FieldManager` to `InputManager`
Currently, the functionality to get data from the `InputManager` by that data's properties is implemented in the method `_get_field_blob_names()` in the `FieldManager` class.

This functionality should be generalized and moved into the `InputManager` class.

**Additional context**
When this functionality is moved to the `InputManager`, `_get_field_blob_names()` in `FieldManager` should be replaced with the functionality from `InputManager`. Also, the functionality should be integrated into the `Weather` module, in order to make daylength specific to the simulated location.

Isgotkowitz:
This PR adds the functionality the Input Manager to get top-level metadata keys from the "files" section of the metadata based on what their properties are (`get_data_keys_by_properties()`). It is used in two places (currently): Field Manager and Weather. Field Manager has had an old method deleted and replaced by this new method. Weather has had a new method added to utilize this functionality, which gets the latitude and then uses it to calculate the daylength.

All testing has been updated as necessary to maintain or expand coverage.

Closes #1006


tests/test_input_manager.py
```diff
@@ -307,8 +308,8 @@ def test_populate_pool_raises_keyerror(mock_input_manager: InputManager,
             with pytest.raises(KeyError):
                 mock_input_manager._populate_pool(eager_termination=True)
 
-                assert add_log.call_count == 0
-                assert add_warning.call_count == 0
+            assert add_log.call_count == 0
+            assert add_warning.call_count == 0
```
Isgotkowitz: Before outdenting these statements they were not being run.

RUFAS/input_manager.py
```diff
@@ -780,6 +780,54 @@ def get_metadata(self, metadata_address: str) -> Any:
...
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self.get_data_keys_by_properties.__name__,
+        }
+
+        try:
+            input_data = self.get_metadata("files")
+        except KeyError:
+            error_name = "Cannot find data"
+            error_message = "Could not find input metadata."
```
ew3361zh, line 846: We have a similar try/except block in the get_metadata method but there the error is actually raised (line 780 currently). How do you envision these working together and do we need both try/except blocks? 

We could remove raising that error in `get_metadata()` and just return None and then this function could be updated to react to what's returned instead of having this second try/except block which is doing a pretty similar thing.

RUFAS/input_manager.py
```diff
@@ -780,6 +780,54 @@ def get_metadata(self, metadata_address: str) -> Any:
             raise KeyError(f"Data not found: Cannot find \"{metadata_address}\", "
                            f"\"{parent_address}\" does not have attribute \"{invalid_key}\".")
 
+    def get_data_keys_by_properties(self, target_properties: str) -> list[str]:
+        """
+        Retrieves list of keys that point to data which have specified properties.
```
ew3361zh: I think I was having trouble originally envisioning which keys you were referring to needing when describing this functionality. Could we specify this any more? My understanding is that these are keys within the first level of the "files" portion of the metadata. Is that correct? I think that would be helpful for understanding what this function's purpose is within IM.

RUFAS/input_manager.py
```diff
@@ -780,6 +780,54 @@ def get_metadata(self, metadata_address: str) -> Any:
...
+
+        Parameters
+        ----------
+        target_properties : str
+            The name of the property that is being searched for.
```
```suggestion
            The name of the properties group that is being searched for.
```

RUFAS/input_manager.py
```diff
@@ -780,6 +780,54 @@ def get_metadata(self, metadata_address: str) -> Any:
...
+
+        Returns
+        -------
+        list[str]
+            List of keys that point to data in the Input Manager which adhere to the target property.
```
ew3361zh: I'm not sure how to make this clearer but it feels like we're missing some information. It's admitedly a very complicated and specific thing. I'm just throwing out some ideas here to try and encompass more of what I think should be in here for the user to understand what's returned by this function.

```suggestion
            List of keys that point to the data within the Input Manager's pool which adhere to the metadata's guidelines for the target properties.
```

RUFAS/input_manager.py
```diff
@@ -780,6 +780,54 @@ def get_metadata(self, metadata_address: str) -> Any:
...
+        """
+        data_keys = []
+
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self.get_data_keys_by_properties.__name__,
+        }
+
+        try:
+            input_data = self.get_metadata("files")
```
ew3361zh: Since this seems like a pretty specific call to `get_metadata()`, did you think about just adding the list of metadata "files" keys with their properties to the IM pool and calling `get_data()` to retrieve what's needed?

ew3361zh (changes requested):
Looks good, Ed, and I think would work fine as written but had a few questions about the way things were implemented and the clarity of the docstrings given how niche the request to IM seems to be.

ew3361zh:
Also, I'm seeing some command-line level complexity merge conflicts with Sphinx documentation which seems like a problem if it pops up at all regularly. Please let me know how that goes - I may need to go back and think about the approach with the Sphinx git-action.

RUFAS/input_manager.py
Isgotkowitz: > first level of the "files" portion of the metadata. Is that correct?
Yes. I tried to avoid using the term "files" though, because the metadata has been structured so that we can eventually expand the input streams to things besides just files (APIs, databases, etc)

I'm not sure what a better way would be to refer to the blobs that define the input data in metadata though.

RUFAS/input_manager.py
Isgotkowitz: But `files` isn't really data, it's metadata, right? Also it is not specified what `__pool` is supposed to be used for, but it seems like just data, because all the metadata is kept in `__metadata`. So adding the information from `files` into `__pool` would mix this up.

RUFAS/input_manager.py
Isgotkowitz: Your suggested change is good - it's much clearer that we are looking for keys that will be used to get information from the metadata. I will update this docstring, but will probably need more suggestions on it.

RUFAS/input_manager.py
ew3361zh: I get what you're saying here - it does seem like this helps make the case for separating out the files sections of the metadatas from the properties section. Because the list of what the field names are seems like data and not metadata but the way we have it set up now, I think you're right, it is technically metadata and could be confusing to add to the IM pool.

RUFAS/input_manager.py
Isgotkowitz: > How do you envision these working together and do we need both try/except blocks?

I added the error to OM in the above code because it seemed like bad practice to do nothing with a raised error. But I don't raise another Key Error because it is not unexpected that `get_data_keys_by_properties()` will return no data.

> We could remove raising that error in get_metadata() and just return None and then this function could be updated to react to what's returned instead of having this second try/except block which is doing a pretty similar thing.

I think `get_metadata()` should be left as-is. Changing its behavior would require carefully checking the rest of the codebase and updating any places that use it, which is beyond the scope of this issue.

RUFAS/input_manager.py
ew3361zh: A couple more thoughts:

What do you think about making that hardcoded "files" arg for `get_metadata()` an arg for `get_data_keys_by_properties` with maybe a default as "files"?

I also wondered about if using an Example section would be helpful here. e.g. show a small snippet of the current metadata and show what you're looking for and what's being sent as args.

RUFAS/input_manager.py
Isgotkowitz: Yes, it seems like it could argued for the keys in `files` being either metadata or data.

> did you think about just adding the list of metadata "files" keys with their properties to the IM pool and calling get_data() to retrieve what's needed?

When it comes the properties, I think those are definitely metadata and should not be placed in `__pool` (unless it is acceptable to have metadata in `__pool` - I am still unsure about that).

RUFAS/input_manager.py
Isgotkowitz: > I also wondered about if using an Example section would be helpful here. e.g. show a small snippet of the current metadata and show what you're looking for and what's being sent as args.

I think that would be really helpful, I'll add one.

> What do you think about making that hardcoded "files" arg for get_metadata() an arg for get_data_keys_by_properties with maybe a default as "files"?

I think that eventually the name will change from `"files"` to something else - but whatever that something else is, it will be probably be constant and there will only be one of them per simulation, so there wouldn't be any use-case for calling functions to change it.

RUFAS/input_manager.py
```diff
@@ -783,6 +783,82 @@ def get_metadata(self, metadata_address: str) -> Any:
             raise KeyError(f"Data not found: Cannot find \"{metadata_address}\", "
                            f"\"{parent_address}\" does not have attribute \"{invalid_key}\".")
 
+    def get_data_keys_by_properties(self, target_properties: str) -> list[str]:
+        """
+        Retrieves list of keys that point to data which have specified properties.
```
```suggestion
        Retrieves the list of metadata keys that point to data which have the target_properties.
```

RUFAS/input_manager.py
```diff
@@ -783,6 +783,82 @@ def get_metadata(self, metadata_address: str) -> Any:
...
+
+        Parameters
+        ----------
+        target_properties : str
+            The name of the properties group that is being searched for.
```
```suggestion
            The name of the metadata properties group that is being searched for.
```

RUFAS/input_manager.py
```diff
@@ -783,6 +783,82 @@ def get_metadata(self, metadata_address: str) -> Any:
...
+                    ...
+                },
+                ...
+            },
+            "properties": {...},
+            ...
+        }
+        ```
+        The the call `get_data_keys_by_properties("field_properties")` would be expected to return the list
+        `["field_1", "field_2"]`.
```
ew3361zh, line 828: thank you, this is really helpful

ew3361zh (approved):
Thanks, I think the changes have made the function's role clearer. I still don't love "files" being hardcoded - maybe we could extract it and give it a more generic name for now? Not a dealbreaker for me though.

A couple other small suggestions in wording that you can take or leave but thought it made things a little clearer still.

Isgotkowitz:
>  I still don't love "files" being hardcoded - maybe we could extract it and give it a more generic name for now? 

@ew3361zh just added this. I declared `PATH_TO_INPUTS = "files"` at the top of `input_manager.py`, and it is now used in `get_data_keys_by_properties()`. If you think the name could be made better though, let me know.

RUFAS/input_manager.py
```diff
@@ -783,6 +785,82 @@ def get_metadata(self, metadata_address: str) -> Any:
...
+            except KeyError:
+                error_name = "Data does not have properties."
+                error_message = f"{key} in metadata does not contain 'properties' value."
+                om.add_error(error_name, error_message, info_map)
+                continue
+
+            if properties == target_properties:
+                data_keys.append(key)
+
+        return data_keys
```
ltp23: I think using try/except with a for-loop looks somewhat strange. Design-wise, the check for the presence of the `properties` field should be the job of some other function. If you like a one-liner with list comprehension:

```
 return [key for key, data in input_data.items() if data.get('properties') == target_properties]
```

tests/test_input_manager.py
```diff
@@ -1862,6 +1863,60 @@ def test_get_metadata_raises_exception(dummy_metadata_path: str,
...
+def test_get_data_by_properties_no_data(mock_input_manager: InputManager,
+                                        input_manager_original_method_states: Dict[str, Callable]) -> None:
+    """Tests that error is handled properly when get_metadata() raises KeyError."""
+    mock_input_manager.get_metadata = MagicMock(side_effect=KeyError)
+
+    with patch("RUFAS.output_manager.OutputManager.add_error") as add_error:
+        actual = mock_input_manager.get_data_keys_by_properties("dummy_property")
+
+    assert add_error.call_count == 1
+    assert actual == []
```
ltp23, line 1875: I thought variables declared within a with-clause would have a block-level scope, but this works so that means they have a function-level scope. But I think it would be better style to indent the two assert statements.

ltp23 (approved):
I made a couple of comments. Overall, the PR looked good to me. Thanks Ed!

ltp23:
Maybe changing `PATH_TO_INPUTS` to something like `ROOT_LEVEL_PROPERTY/KEY`, since `files` is not really a path. Instead of declaring at the module level, you could make it a class variable.

Isgotkowitz:
> Maybe changing `PATH_TO_INPUTS` to something like `ROOT_LEVEL_PROPERTY/KEY`, since `files` is not really a path. Instead of declaring at the module level, you could make it a class variable.

I will change it to `ADDRESS_TO_INPUTS` - this will keep it's name consistent with the terminology used in `get_metadata()`. Also I will keep it as a module level variable for now, I discussed this design a couple days ago with Pooya and this is what he recommended.

RUFAS/input_manager.py
Isgotkowitz: I have implemented this change, there is no catching of KeyError's in `get_data_keys_by_properties()` - but I think this is ok. The method only looks in `__metadata` if there is something in it, and the only way metadata can be added is through `start_data_processing()`. That method always calls `_populate_pool()` which effectively validates the metadata for the purposes of `get_data_keys_by_properties()`.

PR #1053, opened 2023-12-15
Linting Script Filter for Added and Modified Files

### LINKED ISSUE #831: Linting scripts should only check added or modified Python files.
The `lint_branch_changes` scripts (both shell and batch) check all Python files that were touched in a PR, including ones that were deleted. This is unnecessary, because if a Python file is deleted it doesn't matter if there were any linting issues in it.

Add the argument `--diff-filter=A,M,R` - this will result in only added, modified, or renamed files being linted.

**Additional context**

allisterakun:
This PR addresses issue #831:
- Adds the `--diff-filter=AM` flag to the `git diff` operation to only get the list of added or modified files.
- Here is the reference documentation:

### [--diff-filter=[(A|C|D|M|R|T|U|X|B)…​[*]]](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---diff-filterACDMRTUXB82308203)
Select only files that are Added (A), Copied (C), Deleted (D), Modified (M), Renamed (R), have their type (i.e. regular file, symlink, submodule, …​) changed (T), are Unmerged (U), are Unknown (X), or have had their pairing Broken (B). Any combination of the filter characters (including none) can be used. When * (All-or-none) is added to the combination, all paths are selected if there is any file that matches other criteria in the comparison; if there is no file that matches other criteria, nothing is selected.

Also, these upper-case letters can be downcased to exclude. E.g. --diff-filter=ad excludes added and deleted paths.

Note that not all diffs can feature all types. For instance, copied and renamed entries cannot appear if detection for those types is disabled.


lint_branch_changes.sh
```diff
@@ -16,12 +16,12 @@ else
...
+changed_files=$(git diff --name-only --diff-filter=AM "$(git merge-base ${base_branch} HEAD)" | grep -E '\.py$')
 
 if [ -z "$changed_files" ]; then
     # Exit if there are no Python files to lint
     echo "No Python files modified on this branch yet."
     exit 0
 fi
 
-flake8 $changed_files
+flake8 "$changed_files"
```
ew3361zh, line 27: Curious what this is doing - script seems to work the same for me with or without the quotes.

ew3361zh (approved):
LGTM, nice work, Allister!

lint_branch_changes.sh
allisterakun: I did this because I got a warning from Pycharm:
image

Here is the reference: [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086)

Isgotkowitz (approved):
LGTM

ltp23 (approved):
LGTM.

I didn't know we had this problem to begin with :).

Issue #1033, opened 2023-12-11
Cow ED-TAI does not include Pre-Synch Hormone Delivery


YijingGong:
According to today's meeting (Dec 11, 2023) on reproduction. Kaitlyn mentioned that, the current ED-TAI in Rufas is not what farmers do in reality. 

## Expected Behavior

In reality, the majority of farms do ED during a presynchronization hormone delivery protocol.

For example, if the presynchronization hormone delivery protocol is "PreSynch"
1. At a certain DIM, inject one PGF
2. 14 days later, inject another PGF
3. monitor estrus after the second injection of PGF for 10 to 14 days
  3.1 If estrus is detected and the cow is inseminated - good
  3.2 If estrus is not detected by the end of 10 or 14 days, all cows that are not inseminated enter a user-chosen TAI program. 

<img width="645" alt="Screenshot 2023-12-11 at 3 10 58 PM" src="https://github.com/RuminantFarmSystems/MASM/assets/51036031/dc2a4d88-b986-4ea7-8504-4cb0be377ca8">

(ref: https://www.dcrcouncil.org/wp-content/uploads/2019/04/Dairy-Cow-Protocol-Sheet-Updated-2018.pdf)

We should be able to define different presynchronization hormone delivery protocols using the HormoneDeliverySchedule class. Possible presynchronization hormone delivery protocols include: "PreSynch", "Double OvSynch", “G-6-G“ according to DCRC guidelines. Potentially can also add user-defined protocols.

## Current Behavior

Current behavior is below:
1. At a certain DIM (current VWP input), cows started to be monitored for estrus 
2. Estrus is monitored for a certain period of time (controlled by current TAI_start_day)
  2.1 If estrus is detected and the cow is inseminated - good
  2.2 If estrus is not detected by the end of this period of time, all cows that are not inseminated enter a user-chosen TAI program. 

<img width="870" alt="Screenshot 2023-12-11 at 3 18 19 PM" src="https://github.com/RuminantFarmSystems/MASM/assets/51036031/c4f49ee9-2980-4e1e-9b07-b4d8aedfc5e6">

## Possible Solution

1. One Possible solution is to change the current structure of ED-TAI
2. Another possible way (compromising) is to keep the current structure, just add any hormone delivery events, and also make the TAI_start_day minus VWP equal to the length of the ED period, to mimic that we are doing ED during presynchronization.

## Possible Implementation

Maybe do this when next year's animal module re-designing.

[...]

PR #1055, opened 2023-12-15
Cow repro refactor

### LINKED ISSUE #1035: Heifer preg losses are at the same rate as cows
## Expected Behavior

In reality, we would expect heifer pregnancy losses to happen very rarely (relative to the rate of cow pregnancy loss). 

## Current Behavior

The preg check days and loss rates are used for both cows and heifers. The rates of loss are appropriate for cows. 
preg_loss_rate_1, preg_loss_rate_2, and preg_loss_rate_3

## Possible Solution

Create a separate set of inputs for heifer preg loss? This could either be at the same 3-check rhythm as cows, or a single risk value with a random time component. 

## Steps to Reproduce

1.
2.
3.
4.

### LINKED ISSUE #1039: Definition of Voluntary Waiting Period in cow TAI protocol is potentially confusing/unintuitive
## Expected Behavior

Voluntary waiting period should refer to the number of days after calving in which a cow is not eligible to be bred (i.e. if she comes into estrus during this time, she will not be bred). After the voluntary waiting period has passed, then breeding begins. 

## Current Behavior

In the ED and ED-TAI protocols, Voluntary Waiting Period reflects the usage above (as expected). 
In the TAI protocol, Voluntary Waiting Period refers to the time during which no hormone shots are given. But in reality, because hormone shots (presynch and TAI protocols) can last for several weeks, this pushes the actual time of insemination far later than it should be. 

## Possible Solution

Change Voluntary Waiting Program to have the same meaning as in the other two protocols, and change the name of the value it currently represents in the TAI protocol to something more appropriate (eg first hormone delivery or program start day, though it needs to be carefully defined/differentiated because 'program start day' is also used, with a distinct meaning, in ED-TAI). 

More generally, this flags that it might be worthwhile to do a thorough review of terminology used in the repro protocols to see where it is confusing relative to how those terms are used in the dairy industry, to maybe make it more user friendly. 

## Steps to Reproduce

1.
2.
3.
4.

ltp23:
This PR focuses on restructuring the different reproduction protocols for cows. 

In the metadata files, I have removed the `cow_repro_programs` section and the `cows` section is the one being used instead. 

There are many different possible protocol combinations to mix and match, but I think the resynch ones seem trickier to get right. There is an open issue #1033 related to the ED-TAI coupled with the presynch process. It seems like it is possible to include it here or in the next PR though I'm not sure. 

cow_repro_method: `ED`
presynch_protocol: `N/A`
resynch_protocol: `N/A`
repro_sub_protocol: `N/A`

cow_repro_method: `TAI`
presynch_protocol: `PreSynch`, `Double OvSynch`, `G6G`
resynch_protocol: `TAIbeforePD`, `TAIafterPD`, `PGFatPD`
repro_sub_protocol: `Ov Synch 48`, `Ov Synch 56`, `CoSynch 72`, `5d CoSynch`

cow_repro_method: `ED-TAI`
presynch_protocol: `N/A`
resynch_protocol: `TAIbeforePD`, `TAIafterPD`, `PGFatPD`.
repro_sub_protocol: `Ov Synch 48`, `Ov Synch 56`, `CoSynch 72`, `5d CoSynch`

---

Diagrams

[Link to folder](https://whimsical.com/cow-repro-protocols-KCGFZkgpcTz8gCh4GtEWDc)


input/data/animal/default_animal.json
```diff
@@ -1,196 +1,205 @@
...
+                    "estrus_service_rate": 1,
+                    "ed_conception_rate": 0.6,
+                    "cow_presynch_protocol": "Double OvSynch",
+                    "cow_TAI_protocol": "OvSynch 56",
+                    "ovsynch56_conception_rate": 0.6,
+                    "ovsynch48_conception_rate": 0.6,
+					"cosynch72_conception_rate": 0.6,
+                    "cosynch5d_conception_rate": 0.6,
+                    "cow_resynch_protocol": "TAIafterPD",
+                    "tai_program_start_day": 72
```
tomhuhh: Thanks Loi for working on this! I think we should be able to remove the duplicate requests for information related to cow repro protocols?

input/data/animal/default_animal.json
```diff
@@ -1,196 +1,205 @@
...
+				"preg_check_day_3": 200,
+				"preg_loss_rate_3": 0.017,
+
+				"cow_preg_check_day_1": 32,
+				"cow_preg_loss_rate_1": 0.02,
+				"cow_preg_check_day_2": 60,
+				"cow_preg_loss_rate_2": 0.096,
+				"cow_preg_check_day_3": 200,
+				"cow_preg_loss_rate_3": 0.017,
+
```
tomhuhh: Similarly here, should we remove the duplicate for input requests related to preg check?

tomhuhh (commented):
Thanks Loi! So far I have only reviewed the input file and left some comments there. I also tested the code with some example scenarios using 'ED-TAI' as the cow repro protocol. I think the results look promising. I'll review the rest of the code when I find more time, either this week or next week. Thanks again! 

image

input/data/animal/default_animal.json
ltp23: Thanks Haowen. I will remove the `cow_repro_programs` section.

input/data/animal/default_animal.json
ltp23: I added an extra set of preg-check parameters for cows but I didn't have the literature values for them. So currently, they have the same values as those of heifers.

RUFAS/routines/animal/life_cycle/animal_constants.py
```diff
@@ -40,9 +42,22 @@
 INJECT_CIDR = "inject CIDR"
 
 # presynch protocols
+PRESYNCH_PERIOD_START = "Presynch period started"
```
krb78, line 48: There are many different "presynch" protocols, could we label this one "presynch11" or "presynch12" as appropriate? (there are 14, 12, 11, and 10 day versions of this protocol, so labeling it as the one it is will help us and users)

RUFAS/routines/animal/life_cycle/hormone_delivery_schedule.py
```diff
@@ -63,32 +64,74 @@ class HormoneDeliverySchedule:
...
+            25: {
+                'set_presynch_end': True,
+                'set_tai_start': True,
+            },
+        },
+        CowReproProtocolEnum.PreSynch_DoubleOvSynch.value: {
+            0: {'deliver_hormones': ['GnRH']},
+            7: {'deliver_hormones': ['PGF']},
+            10: {'deliver_hormones': ['GnRH']},
+            16: {
```
krb78: If we have a presynch11 (the "presynch" protocol ends 11 days after the final PGF and the first GnRH of the ovsynch protocol begins on day 25), then I think the double ovsynch protocol is missing a day. The GnRH shot of the ovsynch (TAI protocol) should begin on day 17. I might misunderstand what day the GnRH of the TAI protocol is beginning, but based off of how G6G and presynch are set up, I think we are off by a day.

RUFAS/routines/animal/life_cycle/hormone_delivery_schedule.py
```diff
@@ -63,32 +64,74 @@ class HormoneDeliverySchedule:
...
+                'set_ai_day': True,
+                'set_conception_rate': True,
+                'set_tai_end': True
+            }
         },
-        'CoSynch 72': {
+        CowReproProtocolEnum.TAI_CoSynch_72.value: {
             0: {'deliver_hormones': ['GnRH']},
             7: {'deliver_hormones': ['PGF']},
             10: {'deliver_hormones': ['GnRH']},
```
krb78: In a cosynch72 protocol, the final GnRH shot should occur at the same time as breeding on day 11.

RUFAS/routines/animal/life_cycle/hormone_delivery_schedule.py
```diff
@@ -63,32 +64,74 @@ class HormoneDeliverySchedule:
...
         },
-        '5d CoSynch': {
+        CowReproProtocolEnum.TAI_5d_CoSynch.value: {
             0: {'deliver_hormones': ['GnRH']},
             5: {'deliver_hormones': ['PGF']},
             6: {'deliver_hormones': ['PGF']},
             8: {'deliver_hormones': ['GnRH']},
-            9: {'set_ai_day': True, 'set_conception_rate': True}
+            9: {
+                'set_ai_day': True,
```
krb78, line 120: In a 5 day cosynch, the final GnRH shot and the TAI happen on the same day (day 8)

input/data/animal/default_animal.json
```diff
@@ -1,196 +1,192 @@
...
+		"herd_init": false,
+		"breed": "HO"
+	},
+	"herd_initialization": {
+		"initial_animal_num": 500,
+		"simulation_days": 5000
+	},
+	"animal_config": {
+		"management_decisions": {
+			"breeding_start_day_h": 380,
```
krb78: Maybe I'll arrive at the answer to this question as I progress, but is this truly the date breeding begins in heifers or is it the date we start simulating estrus? If it's the latter (which it was prior to refactor) - can we update the name of the input to accurately show that?

input/data/animal/default_animal.json
```diff
@@ -1,196 +1,192 @@
...
+				"birth_weight_std_je": 1,
+				"target_heifer_preg_day": 420,
+				"mature_body_weight_avg": 740.1,
+        		"mature_body_weight_std": 73.5
+			}
+		},
+		"from_literature": {
+			"repro": {
+				"preg_check_day_1": 32,
+				"preg_loss_rate_1": 0.02,
```
krb78, line 81: so these are for heifers? could we call it "heifer_preg_loss_rate_1"?

input/data/animal/default_animal.json
krb78: I think that's fine, Loi. Across the industry this data isn't well established because truthfully heifers lose so few pregnancies. But we can determine updated defaults for that at a later time.

input/metadata/ARL_metadata.json
```diff
@@ -450,113 +450,38 @@
...
                 "type": "string",
                 "description": "Resynch Protocol -- Resynch protocol for cow before/at/after pregnancy check day in the TAI (Timed Artificial Insemination) program",
                 "default": "TAIafterPD",
-                "pattern": "^(TAIafterPD|TAIbeforePD|PGFatPD)$"
+                "pattern": "^(TAIafterPD|TAIbeforePD|PGFatPD|N/A)$"
               },
               "repro_sub_protocol": {
                 "type": "string",
-                "pattern": "^(OvSynch 56|OvSynch 48|CoSynch 72|5d CoSynch)$"
+                "description": "Sub-protocol for cow reproductive program. This ONLY applies to the TAI repro protocol.",
```
krb78: Question, Loi: what protocol for TAI is instituted in an "ED-TAI" herd? I thought it would be the same as this "repro_sub_protocol" in which case we actually need this "sub-protocol" to be able to be instituted for animals in the TAI-ED program. In this scenario, it will need to have a start date AFTER the VWP date. In a TAI only program, the sub-protocol will begin BEFORE the VWP date. Perhaps I just misunderstand how the ED-TAI group is working and it will become clearer as I progress through the code.

input/metadata/ARL_metadata.json
```diff
@@ -450,113 +450,38 @@
...
+                "pattern": "^(OvSynch 48|OvSynch 56|CoSynch 72|5d CoSynch|N/A)$",
+                "default": "OvSynch 56"
               },
               "repro_sub_properties": {
                 "type": "object",
                 "conception_rate": {
                   "type": "number",
-                  "description": "Conception Rate -- The conception rate for the cow reproductive program (default is 0.6)",
+                  "description": "Conception Rate -- The conception rate for the cow reproductive program (default is 0.6). This ONLY applies to the TAI repro protocol.",
```
krb78: just an FYI you have this written as 0.6 in the description but the code one line below says the default is 0.5. Also, like my question above, it seems this conception rate should apply to animals receiving TAI in either the TAI protocol or the ED-TAI protocol.

input/metadata/ARL_metadata.json
```diff
@@ -450,113 +450,38 @@
...
                 "type": "string",
-                "description": "Presynch Protocol -- The preSynch protocol (1st breeding) for cow in TAI (Timed Artificial Insemination) program",
+                "description": "Presynch Protocol -- The preSynch protocol (1st breeding) for cow in TAI (Timed Artificial Insemination) program. This ONLY applies to the TAI repro protocol.",
                 "default": "Double OvSynch",
-                "pattern": "^(Double OvSynch|Presynch|G6G)$"
+                "pattern": "^(Double OvSynch|PreSynch|G6G|N/A)$"
               },
               "resynch_protocol": {
                 "type": "string",
                 "description": "Resynch Protocol -- Resynch protocol for cow before/at/after pregnancy check day in the TAI (Timed Artificial Insemination) program",
```
krb78: A Resynch_protocol needs to be able to be instituted in both the TAI and the ED-TAI programs.

tests/animal_module_tests/test_repro_helper_modules.py
```diff
@@ -26,32 +26,48 @@
     }),
 
     # Valid cow protocols
-    ('cows', 'OvSynch 48', {
+    ('cows', CowReproProtocolEnum.TAI_OvSynch_48.value, {
         0: {'deliver_hormones': ['GnRH']},
         7: {'deliver_hormones': ['PGF']},
         9: {'deliver_hormones': ['GnRH']},
         10: {'deliver_hormones': ['GnRH']},
```
krb78: There shouldn't be 2 GnRH shots in this protocol on days 9 and 10, just one on day 9. Breeding occurs on day 10.

tests/animal_module_tests/test_repro_helper_modules.py
```diff
@@ -26,32 +26,48 @@
...
+            'set_conception_rate': True,
+            'set_tai_end': True
+        }
     }),
-    ('cows', 'OvSynch 56', {
+    ('cows', CowReproProtocolEnum.TAI_OvSynch_56.value, {
         0: {'deliver_hormones': ['GnRH']},
         7: {'deliver_hormones': ['PGF']},
         9: {'deliver_hormones': ['GnRH']},
         10: {'deliver_hormones': ['GnRH']},
```
krb78: Same here as above - breeding (TAI) should occur on day 10, not another GnRH shot.

tests/animal_module_tests/test_repro_helper_modules.py
```diff
@@ -26,32 +26,48 @@
...
+            'set_ai_day': True,
+            'set_conception_rate': True,
+            'set_tai_end': True
+        }
     }),
-    ('cows', 'CoSynch 72', {
+    ('cows', CowReproProtocolEnum.TAI_CoSynch_72.value, {
         0: {'deliver_hormones': ['GnRH']},
         7: {'deliver_hormones': ['PGF']},
         10: {'deliver_hormones': ['GnRH']},
```
krb78: The GnRH and TAI should both occur on day 10.

tests/animal_module_tests/test_repro_helper_modules.py
```diff
@@ -26,32 +26,48 @@
...
+            'set_conception_rate': True,
+            'set_tai_end': True
+        }
     }),
-    ('cows', '5d CoSynch', {
+    ('cows', CowReproProtocolEnum.TAI_5d_CoSynch.value, {
         0: {'deliver_hormones': ['GnRH']},
         5: {'deliver_hormones': ['PGF']},
         6: {'deliver_hormones': ['PGF']},
         8: {'deliver_hormones': ['GnRH']},
```
krb78: TAI should occur the same day (day 8) as the GnRH shot.

krb78 (changes requested):
Hi Loi! I think you did an awesome job with this! It's much clearer! I had a few questions that would be great for us to address before we merge this in (especially in regards to the ED-TAI protocol) and a few minor changes I think you can make easily (or I hope). Also, I hope it's straightforward enough that sometimes I only made a comment once, but there are a few places we need to make the update (ie the default value for Ovsynch56 conception rate). 

I saw a lot of code about Estrus Detection deleted, but is there some that was changed that we do need to review?

RUFAS/routines/animal/life_cycle/animal_constants.py
ltp23: Yes, we can add a specific label. Currently in the animal event log, it will look like this. When you run the program, one of the output files is `default_scenario_saved_variables_json_animal_events_for_cows<timestamp>.json` in the `output` folder. 
image

Also, we only have 3 options for presynch protocols: Presynch, Double OvSynch, and G6G, so we are missing one. I'm not sure how to map these to the 4 options you mentioned.

RUFAS/routines/animal/life_cycle/hormone_delivery_schedule.py
ltp23: This point is kinda subtle. Here, the `set_tai_start` flag means starting the TAI protocol on the following day. Before going into more details, I think the event log looks correct actually.

image

The code would be easier to work with when we set the presynch end and tai start flags on the same day. This would prevent the following scenario where the presynch period has just ended but the TAI period hasn't started yet, there is a possibility of re-entering another presynch. 

Ideally, we would want the stated day to be the day that the event is happening. There are several workarounds: 
- If we can assume another presynch cannot happen right after one another, then a simple check between the current day and the day the last presynch ended would work.
- Change from `set_tai_start` label to `set_tai_start_next_day` for clarity.
- Have a mini repro program state management system that helps keep track of the many possible transitions between repro programs. For example, at any one point, a cow's repro program can be in one or more of the following states: waiting, presynch, TAI, resynch, ED, etc.

RUFAS/routines/animal/life_cycle/hormone_delivery_schedule.py
ltp23: Yes, I will move that to day 11.

input/data/animal/default_animal.json
ltp23: I have been only using this parameter in heifers. When a heifer reaches this day, we start simulating estrus or start a TAI protocol. What name would you like to change it to?

input/data/animal/default_animal.json
ltp23: Based on this and the other comment about the parameters for cow pregnancy loss rates not so well-established and so minuscule, I think I would remove those newly added parameters and keep these the same for both heifers and cows.

input/metadata/ARL_metadata.json
ltp23: Yes, I will update the metadata descriptions to clarify this.

input/metadata/ARL_metadata.json
ltp23: Right, I will update this. This conception applies to any TAI procedure used in both TAI and ED-TAI protocols.

input/metadata/ARL_metadata.json
ltp23: In both TAI and ED-TAI protocols, the `repro_sub_protocol` specifies the specific TAI procedure to be used.

ltp23: In the TAI-only protocol, if there is a presynch period, then the presynch period will start when the number of days in milk is equal to the voluntary waiting period (currently set at 50).  The first day of the TAI procedure will follow immediately the day after the presynch period ends. So I think this is different from your description, and I'm missing the scenario where there is no presynch protocol set.

ltp23: For the ED-TAI protocol, if there is no estrus detected in the period between the end of the voluntary waiting period and the day before the tai program start day (between 51 and 69 days in milk), then a TAI procedure will start on day 70.

input/data/animal/default_animal.json
```diff
@@ -1,196 +1,192 @@
...
+				},
+				"cows": {
+					"estrus_detection_rate": 0.65,
+					"estrus_conception_rate": 0.6,
+					"presynch_protocol": "Double OvSynch",
+                    "resynch_protocol": "TAIbeforePD",
+                    "repro_sub_protocol": "OvSynch 56",
+					"repro_sub_properties": {
+						"conception_rate": 0.5,
+						"tai_program_start_day": 70
```
tomhuhh: I'm just curious, what is the difference between 'estrus_detection_rate' and 'estrus_conception_rate' of the animal category (i.e. either 'heifers' or 'cows') vs. 'conception_rate' and 'estrus_detection_rate' for the specific repro sub protocol? Should they refer to the same things? Thanks!

RUFAS/routines/animal/life_cycle/animal_constants.py
krb78: Hi Loi, I'm actually only requesting that we label "Presynch" to be "PresynchXX" with the XX being the number of days between the last PGF shot and the start of the Ovsynch56. Double Ovsynch and G6G are specific protocols in and of themselves, but "Presynch" has a number of days associated with it because there are 4 different versions of this protocol and we are only simulating one of the 4 options.

RUFAS/routines/animal/life_cycle/hormone_delivery_schedule.py
krb78: Ah, ok, thanks for clarifying this. Your event log does look correct. I think changing the name from set_tai_start to set_tai_start_next_day might be the easiest and most clarifying change we can make. :) From there, the reader/user will know what is happening :)

input/data/animal/default_animal.json
krb78: Thanks, Loi! I think if we changed this from breeding_start_day_h to estrus/repro_simulation_start_day_h (am I allowed to have a / in my name?) or breeding_protocol_start_day_h that would clarify things! I just want us to signify its not the day we start breeding, it's the day we start the breeding protocol.

input/data/animal/default_animal.json
krb78: I wouldn't remove the newly added parameters, Loi. We should keep them, they are very different than cows. If cows lose almost 10% of pregnancies between 32 and 60 days, heifers lose ~1% of pregnancies and virtually 0% of pregnancies between 60 and 200 days. We need to make this differentiation but the problem is, the rate of pregnancy loss is so low, it's not something that is well researched and reported on. We could look at some farm or DHIA repro records to base this number on.

input/metadata/ARL_metadata.json
krb78: In reality, many farms do a "presynchronization" to bring cows into estrus, then breed off heats, then do what we colloquially call a "clean up" or "back door" ovsynch. In this scenario, a protocol most similar to the one we have labeled as "presynch" would be instituted, but then cows that demonstrate estrus would be bred and those that don't demonstrate estrus would continue on to the TAI protocol (Ovsynch56 for example). 

Largely, I was just trying to ensure that RuFaS will allow for the VWP to be before or after the TAI program start day and that I understand the code well enough to understand what to put in as the VWP. Technically the VWP should be the earliest day at which we breed cows. Therefore it would be the following in each of the following scenarios: 
1) A TAI only (double ovsynch starts on day 45, ovsynch 56 starts on day 62, TAI occurs on day 72): VWP = 72
2) A ED-TAI program (estrus detection begins on day 50, we give cows 1 week to show estrus, ovsynch 56 begins on day 57, TAI is day 67): VWP = 50
3) An ED only program (estrus detection begins on day 50): VWP = 50. 

The above VWP are what someone familiar with dairy management would write for the VWP in each scenario. I just want to make sure that RuFaS does not always expect VWP to occur **before** the TAI protocol start date, because in the 1st scenario of TAI only, VWP occurs at the very end/upon completion of the protocols.

tomhuhh:
Hello Loi! If I set 'resynch_protocol' to 'TAIafterPD', the following error will show up. However, if I set it to 'TAIbeforePD', there is no error. I think this might need some further investigation. Thanks! 
Screenshot 2023-12-26 at 2 42 53 PM

tomhuhh:
Something might have changed since last week I reviewed this branch. In addition to the error mentioned above, the behavior of the herd structure is also not as good as last week's version. 

This is with the input provided by the farm except that 'resynch_protocol' is set to 'TAIbeforePD'. 

image

This is when I made the changes suggested by Kaitlyn a while ago while keeping  'resynch_protocol' as 'TAIbeforePD'.  The specific changes include: 

- TAI conception rate to 55%, DNB to 250 days, & estrus detection rate for cows to 0.9 
- Low MY cut-off - 30 kg 
- Also, I had to run longer for the herd structure to stablize. 

image

The performance of the herd structure is better with these assumptions. 

Thanks!

RUFAS/routines/animal/life_cycle/animal_constants.py
YijingGong: Kaitlyn was suggesting to just change the name of "PreSynch" to "PreSynch11" @ltp23

YijingGong:
The PR looks in good shape already. Good job @ltp23. Just a few comments:

1. cow ED
pure ED with N/A for presynch, repro_sub_protocol, and resynch_protocol. VWP = 50, detection rate = 0.65, conception rate = 0.6
The event log seems pretty correct:
"1379": [
            "simulation_day=47",
            "new birth, start milking",
            "1st estrus after calving: estrus day scheduled on day 1403"
          ],
          "1403": [
            "simulation_day=71",
            "estrus occurred before the end of voluntary waiting period",
            "estrus day scheduled on day 1420"
          ],
          "1420": [
            "simulation_day=88",
            "estrus occurred before the end of voluntary waiting period",
            "estrus day scheduled on day 1445"
          ],
          "1445": [
            "simulation_day=113",
            "estrus occurred",
            "estrus not detected",
            "estrus day scheduled on day 1459"
          ],
          "1459": [
            "simulation_day=127",
            "estrus occurred",
            "estrus detected",
            "AI day scheduled on day 1460"
          ],
          "1460": [
            "simulation_day=128",
            "AI performed",
            "inseminated with conventional",
            "failed conception at rate 0.5",
            "cow not pregnant",
            "estrus day scheduled on day 1482"
          ],
          "1482": [
            "simulation_day=150",
            "estrus occurred",
            "estrus detected",
            "AI day scheduled on day 1483"
          ],
          "1483": [
            "simulation_day=151",
            "AI performed",
            "inseminated with conventional",
            "successful conception",
            "cow pregnant"
          ],
But it seems the parity 1 cow % is a bit too high (~70%), any thoughts? Screenshot 2023-12-27 at 4 41 04 PM

2. cow TAI 
"PreSynch" + "OvSynch 56" + "TAIbeforePD", VWP = 69, tai_start_day = 45
"1217": [
            "simulation_day=63",
            "new birth, start milking"
          ],
          "1285": [
            "simulation_day=131", **<-- 1285 - 1217 = 68. The code currently starts the TAI program on the day of VWP, but we should have started it on the day of tai_start_day. VWP should be used to determine whether the ai_day should be scheduled (before WVP, no ai_day should be scheduled).**
            "Presynch period started: PreSynch",
            "inject PGF"
          ],
          "1299": [
            "simulation_day=145",
            "inject PGF"
          ],
          "1310": [
            "simulation_day=156",
            "Presynch period ended: PreSynch",
            "TAI period started: OvSynch 56"
          ],
          "1311": [
            "simulation_day=157",
            "inject GnRH"
          ],
          "1318": [
            "simulation_day=164",
            "inject PGF"
          ],
          "1320": [
            "simulation_day=166",
            "inject GnRH"
          ],
          "1321": [
            "simulation_day=167",
            "inject GnRH"
          ],
          "1322": [
            "simulation_day=168",
            "AI day scheduled on day 1322",
            "TAI period ended: OvSynch 56",
            "AI performed",
            "inseminated with conventional",
            "successful conception",
            "cow pregnant",
            "setting up resynch TAI protocol: **OvSynch 56**" **<-- It's doing the "TAIbeforePD" correctly, but the event log is wrong here**
          ],
          "1348": [
            "simulation_day=194",
            "inject GnRH"
          ],
          "1354": [
            "simulation_day=200",
            "pregnancy check 1: confirmed",
            "discontinued resynch TAI protocol after first preg check confirmed: **OvSynch 56**" **<-- same here**
          ],


3. cow ED-TAI
VWP = 50, tai_start = 80, "N/A"for presynch, "OvSynch 56" for repro_sub_protocol + "TAIafterPD" for resynch_protocol:
"1195": [
            "simulation_day=103",
            "AI performed",
            "inseminated with conventional",
            "failed conception at rate 0.5499999999999999",
            "cow not pregnant", 
            "estrus day scheduled on day 1213" **<-- starting from here, doing correctly the "TAIafterPD" resynch, but can we keep consistency, add a line as before saying "setting up resynch TAI protocol: TAIafterPD". Similarly for "PGFatPD"**
          ],
          "1213": [
            "simulation_day=121",
            "estrus occurred",
            "estrus detected",
            "AI day scheduled on day 1214"
          ],
Another case when I tested "PGFatPD":
"1546": [
            "simulation_day=142",
            "AI day scheduled on day 1546",
            "TAI period ended: OvSynch 56",
            "AI performed",
            "inseminated with conventional",
            "failed conception at rate 0.4",
            "cow not pregnant",
            "estrus day scheduled on day 1573"
          ],

When I was reviewing the code, I had a question. Likely this is for Kaitlyn and Kristan. Currently, the code has a function (_reduce_conception_rate_based_on_parity) to reduce the conception rate for higher parity cows, do we still want this function? Cuz I am thinking about the pilot farms, they are basically reporting what they observed. Maybe we don't want to account for the reduced fertility for higher parity cows? @KFosterReed @krb78

krb78:
I have 2 questions/thoughts. 

1) @YijingGong in your comments above, in scenario 2 where you are testing the resynch function. Do you just want RuFaS to write "TAIbeforePD" instead of "Ovsynch56" in the event log because technically, RuFaS is doing an Ovsynch56 protocol (and it should be doing this - so it is correct)? I'll defer to you and your expertise with this, because for me, I don't inherently see a problem with calling it "Ovsynch56" as that is the protocol we are executing, but if you wanted to be able to track resynchronization breedings differently than first breedings, I understand why you'd want your event log to read differently. 

2) For the pilot farms because the farmers are reporting their average conception rate, we should be setting the conception rate parity decrease to 0.0 instead of 0.026. However, if RuFaS is being used to run research simulations or other scenarios, I think we will still want this parity conception rate decrease to exist. Does that make sense to you @KFosterReed and @YijingGong?

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -684,586 +663,744 @@ def update(self, sim_day, calving_interval): # noqa
...
+        Notes
+        -----
+        This follows the TAIbeforePD protocol, in which the first shot of GnRH is given 6 days before the first
+        pregnancy check. The conception rate is decreased by the user-defined value. If the cow is pregnant
+        on the first pregnancy check, the TAI protocol is discontinued.
+
+        It is also important to note that the TAI protocol is set up in advance regardless of the outcome of the
+        insemination. This is because even if the insemination is successful, the cow can still fail the first
+        pregnancy check. In this case, the TAI protocol is already initiated 6 days before the first pregnancy check
+        in order to save time.
```
jadamchick: Just a note that this is also important because in real life, you don't know whether or not the insemination was successful until the first preg check (unless she already came into estrus). So even cows who will be confirmed pregnant will receive a GNRH shot 6 days before preg check when using the TAIbeforePD protocol.

jadamchick:
>        "1285": [
>        "simulation_day=131", **<-- 1285 - 1217 = 68. The code currently starts the TAI program on the day of VWP, but we should have started it on the day of tai_start_day. VWP should be used to determine whether the ai_day should be scheduled (before WVP, no ai_day should be scheduled).**
>        "Presynch period started: PreSynch",
>        "inject PGF"
>        ],

I agree with Yijing here - voluntary waiting period meaning needs to be consistent across all protocols as the day before which animals will not be bred. If it means something different in the TAI protocol (eg still means the day of the first hormone injection of presynch or ovsynch), that is confusing. 
For TAI protocol, voluntary waiting period should be the earliest day that AI is scheduled to occur.

jadamchick:
This is looking great! I haven't run it yet but after reviewing the code have a couple comments/clarifications: 

- See above about VWP meaning
- Clarification about TAI_program_start_day: Is this used for both TAI and ED-TAI protocols? The metadata says it is, but the metadata also says for VWP that "If TAI is chosen for the cow, it is days after parturition when the 1st hormonal injection is made in TAI programs." Which is used to indicate day of first injection for TAI protocol? 
- In the ideal world, TAI_program_start_day would refer to the day the first hormone injection is given for presynch (if specified) or ovsynch (if presynch not specified). However, for ED-TAI, I think it always refers to the first day of the Ovsynch protocol, not a presynch protocol? Can that be clarified in the metadata?
- Relatedly, I see that presynch is still not an option with ED-TAI, and assume it is beyond the scope of this refactor. In the future, would it be possible to do a presynch protocol with ED-TAI, and then allow estrus detection for a period of time before starting Ovsynch shots? (see "presynch" protocol on the reference sheet [here](https://www.dcrcouncil.org/wp-content/uploads/2019/04/Dairy-Cow-Protocol-Sheet-Updated-2018.pdf), with 10-14 days of estrus detection AI). @krb78 what are your thoughts?

I will update with more review after running it and looking at outputs. Thanks Loi!

YijingGong:
> I have 2 questions/thoughts.
> 
> 1. @YijingGong in your comments above, in scenario 2 where you are testing the resynch function. Do you just want RuFaS to write "TAIbeforePD" instead of "Ovsynch56" in the event log because technically, RuFaS is doing an Ovsynch56 protocol (and it should be doing this - so it is correct)? I'll defer to you and your expertise with this, because for me, I don't inherently see a problem with calling it "Ovsynch56" as that is the protocol we are executing, but if you wanted to be able to track resynchronization breedings differently than first breedings, I understand why you'd want your event log to read differently.
> 2. For the pilot farms because the farmers are reporting their average conception rate, we should be setting the conception rate parity decrease to 0.0 instead of 0.026. However, if RuFaS is being used to run research simulations or other scenarios, I think we will still want this parity conception rate decrease to exist. Does that make sense to you @KFosterReed and @YijingGong?

For Q1, you are right that Rufas is actually doing the Ovsynch56 protocol correctly. When I wrote the comments, I subconsciously connected the event log of "setting up resynch TAI protocol" with the options of the resynch protocols. 
Q2. Make sense to me.

RUFAS/routines/animal/animal_manager.py
```diff
@@ -1623,9 +1623,9 @@ def daily_updates(self, feed: Feed, weather: Weather, time: Time):
             self._handle_newly_added_animals([*animals_added, *calves_born], feed, current_temperature)
 
             self._record_animal_counts()
-            if time.is_last_day_of_simulation:
-                self._record_animal_events(self.heiferIIs)
+            if self.simulation_day >= 365 and self.simulation_day % 100 == 0:
                 self._record_animal_events(self.cows)
```
tomhuhh, line 1629: '365' seems like a magic value to me. Should we add it to the constant list?

RUFAS/routines/animal/life_cycle/heiferII.py
```diff
@@ -642,6 +649,15 @@ def execute_ed_protocol(self, sim_day: int) -> None:
         """
         Execute the ED protocol.
 
+        Notes
+        -----
+        The two main differences between how estrus detection is handled in the ED protocol and the
+        SynchED protocol are:
+        1. The estrus detection rate and conception rate are different.
```
tomhuhh, line 656: 'The estrus detection rates and conceptions rates are different. '

jadamchick (changes requested):
I ran this for one of the pilot testing herds; the necessary updates to animal.json and metadata.json were the only changes I made compared to earlier runs. 
This was using the ED protocol for both heifers and cows. 
I ran it for 5 years (1230-cow herd), and compared the results to compiled results from previous runs. 

With this new refactor, there were: 
- 2x as many Parity 1 animals (63% of the herd, v. 36% previously)
- Half as many dry cows (6% vs. 12% of the cow herd)
- 10x as many cows sold for low production in the last year (347 v. 31; 56% of total culls vs. 9%)

So I feel comfortable saying something is not working right with the ED protocol in this refactor. 

I looked at the logs and noticed a couple oddities: 
- when pregnancy loss occurs, it reschedules the next estrus to an earlier day in the simulation: 
image

- when insemination fails and the subsequent estrus is not detected, so the cow reaches pregnancy check and is checked "not pregnant", it reschedules estrus again (so is pushed out an extra ~20+ days) instead of maintaining the already-scheduled estrus: 
image

On a brighter note, I did not see any pgf injections given while the protocol was set to ED!

jadamchick:
> I have 2 questions/thoughts.
> 
>     1. @YijingGong in your comments above, in scenario 2 where you are testing the resynch function. Do you just want RuFaS to write "TAIbeforePD" instead of "Ovsynch56" in the event log because technically, RuFaS is doing an Ovsynch56 protocol (and it should be doing this - so it is correct)? I'll defer to you and your expertise with this, because for me, I don't inherently see a problem with calling it "Ovsynch56" as that is the protocol we are executing, but if you wanted to be able to track resynchronization breedings differently than first breedings, I understand why you'd want your event log to read differently.
> 
>     2. For the pilot farms because the farmers are reporting their average conception rate, we should be setting the conception rate parity decrease to 0.0 instead of 0.026. However, if RuFaS is being used to run research simulations or other scenarios, I think we will still want this parity conception rate decrease to exist. Does that make sense to you @KFosterReed and @YijingGong?

The second point caught my eye too -- because, I think, the parity decrease is hard-coded as a 0.05 penalty for each older parity. (So when looking through the logs of a herd where we inputted a 0.5 conception rate, there are many reports of "failed conception at rate 0.4" or "0.45").   If there is a way to control whether this penalty is implemented or not (eg so we can "turn it off" for pilot testing herds), that would be great. 
The 0.026 that Kaitlyn is referencing is the decrease applied for repeated breedings, I think?

tomhuhh (commented):
Thanks Loi! I added a couple minor comments on the code/docstring.

jadamchick:
This morning I ran this code with the same herd as above, using the TAI and then ED-TAI protocols. 

- TAI looked like it performed well. Outputs were similar to what I saw previously (without the refactor). Presynch shots consistently began on the day I specified as "voluntary waiting period" (see earlier comments about confusion around this name for this input). 

- ED-TAI is confusing. The outputs were poor (52% L1, high number of culls for low production, etc.). When I looked at the event logs, it appears that something is funky about when the first GnRH shot is occurring: a) it varies across a range of ~15 days, and b) it is later than either the "TAI program start day" or the "Voluntary waiting period" inputs. It also seems to always coordinate with a scheduled estrus (e.g., is it waiting for a scheduled estrus day to begin her first Ovsynch shot?)

For more detail, my inputs for ED-TAI run were: 
- ED-TAI
- presynch: Double Ovsynch (assuming this would be ignored, and it appears that it was in fact ignored)
- sub-protocol: Ovsynch 56
- program start day: 64 (assumed this would be the day of the first Ovsynch shot)
- voluntary waiting period: 74 (since this herd does first breeding entirely on TAI, then rebreeding on heat detection)

Results from several cows scrolled to at random in the event log: 
image


Let me know if you have any questions, I'm happy to discuss any of this if it would be easier to talk through!

ltp23:
Hi everyone, I have made some logic changes based on the design [diagrams](https://whimsical.com/cow-repro-protocols-KCGFZkgpcTz8gCh4GtEWDc)  and the feedback so far. Some additional changes. which I can revert or modify further, are:

- I added two true/false fields to control whether to decrease conception rate or not: `decrease_conception_rate_in_rebreeding` (I keep track of how many times the cow has entered the open state and multiply it some user-defined, hard-coded value), and `decrease_conception_rate_by_parity` (based on number of calves).
- Input parameter name changes
```
"cows": {
"estrus_detection_rate": 0.65,
"estrus_conception_rate": 0.6,
"presynch_program": "Double OvSynch",
"presynch_program_start_day": 50,
"ovsynch_program": "OvSynch 56",
"ovsynch_program_start_day": 66,
"ovsynch_program_conception_rate": 0.7,
"resynch_program": "TAIbeforePD"
}
```
If both presynch and ovsynch programs are used in TAI protocol, the correct way to set the ovsynch program start day would be on the last day of the presynch program or later. But if the ovsynch program start day is set before the end of the presynch program, then still, the ovsynch program can't start until the last day of presynch.  

- I added a list of possible repro states that a cow can be in based on what I have gathered so far. Some are more aligned with real-world scenarios than others. Typically, I would expect the cow to be in only one of those states at any point. The main use of this whole design is so that we can route to different functions depending on the current repro state. 
  - `NONE`: the empty state that acts as the starting state, or a useful intermediate state while the cow is transitioning between different states.
  - `WAITING_FULL_ED_CYCLE`: The typical state when we do estrus detection.
  - `WAITING_SHORT_ED_CYCLE`: This only applies when we choose the `PGFatPD` resynch method, in which the cow waits a maximum of 7 days for estrus. 
  - `WAITING_ED_DAILY`: This is used only in the period between `program_start_day` and `tai_program_start_day` when ED-TAI protocol is selected.
  - `IN_PRESYNCH`: Used in the duration of the presynch program.
  - `HAS_DONE_PRESYNCH`: After the cow has gone through a presynch program.
  - `IN_OVSYNCH`: When an ovsynch program is used. This can happen in both TAI-only and ED-TAI protocols.
  - `ESTRUS_DETECTED`: A very brief state in which we mark an estrus has been detected and then the following day an AI will be performed.
  - `AFTER_AI`: As the name suggests, after an AI has been performed, the cow enters this state. With a successful conception, she will enter the pregnant state.
  - `PREGNANT`: A cow enters or stays in this state when she is successfully conceived or when she passes a pregnancy check. If she fails a pregnancy check, she will exit this state.   
- For simplicity, in PGFatPD, all failed pregnancy checks will lead to a PGF injection, followed by a short estrus cycle detection, which is then followed by a TAI program if estrus not detected.

ltp23:
One thing I haven't figured out yet is why the number of cows with parity 3 or more just drops off after a while. 

default_scenario_herd-composition-16-Jan-2024_Tue_11-35-12

default_scenario_cow-population-by-parity-16-Jan-2024_Tue_11-35-12

RUFAS/routines/animal/animal_manager.py
ltp23: Right, Haowen. This is definitely temporary. I just wanted to log the animal events for cows at a regular interval. I think what I will do before merging is to call it once at the end of the simulation.

ltp23:
> With this new refactor, there were:
> 
> * 2x as many Parity 1 animals (63% of the herd, v. 36% previously)
> * Half as many dry cows (6% vs. 12% of the cow herd)
> * 10x as many cows sold for low production in the last year (347 v. 31; 56% of total culls vs. 9%)


Hi Julie, I still haven't been able to resolve the strange results for ED-only protocol. Even with the latest updates, the patterns you observed still exist.


> * when insemination fails and the subsequent estrus is not detected, so the cow reaches pregnancy check and is checked "not pregnant", it reschedules estrus again (so is pushed out an extra ~20+ days) instead of maintaining the already-scheduled estrus:

I can try to fix this one so that we don't reschedule a new estrus day and continue with the existing schedule. This makes me think that by extending the next estrus day, the cow may have gotten culled because of not getting pregnant within the breeding stage and having low milk production.

Update: I have updated it to not scheduling a new estrus day.

ltp23:
> Relatedly, I see that presynch is still not an option with ED-TAI, and assume it is beyond the scope of this refactor. In the future, would it be possible to do a presynch protocol with ED-TAI, and then allow estrus detection for a period of time before starting Ovsynch shots?

This is doable and I can pick up on this later.

ltp23:
Hi Haowen (@tomhuhh), with the latest changes and using the parameters you tried, I got the following

> * TAI conception rate to 55%, DNB to 250 days, & estrus detection rate for cows to 0.9
> * Low MY cut-off - 30 kg
> * Also, I had to run longer for the herd structure to stabilize.

`cow_repro_method`: ED-TAI
`do_not_breed_time`: 250
`estrus_detection_rate`: 0.9
`ovsynch_program_conception_rate`: 0.7
`resynch_program`: TAIbeforePD
default_scenario_cow-population-by-parity-18-Jan-2024_Thu_22-41-28
default_scenario_culling-cow-stats-18-Jan-2024_Thu_22-41-28
default_scenario_herd-composition-18-Jan-2024_Thu_22-41-28

jadamchick:
Thanks for all of the great + persistent work on this Loi! 
I ran several different combo's of repro programs and these are my observations: 

- ED: looks good! no red flags in the herd dynamics or the animal events
- TAI: also looks good! I would like to try more combinations of presynch and resynch programs. When I ran PresynchOvsynch paired with TAIafterPD, all seems to run smoothly. 
- **ED-TAI: a few issues yet....** 

1. It seems as though every single cow comes into estrus + is detected/bred for her first breeding (eg no one makes it to Ovsynch). I was looking at a herd with VWP=60, Ovsynch_start_day = 70, and estrus_detection = 0.6. So I would expect < 50% of cows to come into estrus during that 10-d window (10/21=0.48). And I would expect 40% of those who do come into estrus, not to be detected. That leaves roughly 70% of cows to proceed to Ovsynch. But when I looked through the events, I didn't see a single cow go into Ovsynch for her first AI event (they were only bred on TAI through resynch). 

2. Something funky is happening with animals who enter the herd during the time between calving ("new birth") and AI. They don't have any repro events, until DNB. (see screen shot for example). 
 
image



Lastly, I think there are a few metadata updates that could make things more clear for inputs: 
a- VWP: update the description to match the updated meaning, and include "N/A" option (to use for TAI protocol)
b- estrus_detection_rate and estrus_conception_rate: properties describes that these values are only relevant for ED programs. Is that true? 
c- presynch_program_start_day and ovsynch_program_start_day: it would be nice to have the option of "N/A" for when these are not in use 


Thanks Loi!! Let me know if you have any questions & we'll keep working through this together :)

KFosterReed:
Glad to see there is progress on this front! One comment in response to Julie's point 2 above is that other than during initialization, there should not be cows entering the herd at this time period (between calving and reproduction). My guess it is a misalignment between the cows in the herd database and the desired simulation. 

Two thoughts for fixing this are 1) look at the methods for herd initialization and modify something related to repro, 2) add an issue about this and ignore for now as it will sort it self out by the end of the simulation.  

@jadamchick, one workaround thought for you would be to initialize the herd when running the simulations (last time I checked it was much much quicker) in which case the repro methods will align and the ideal outcome is that these gaps in achieving pregnancy won't happen

ltp23:
> 1. It seems as though every single cow comes into estrus + is detected/bred for her first breeding (eg no one makes it to Ovsynch). I was looking at a herd with VWP=60, Ovsynch_start_day = 70, and estrus_detection = 0.6. So I would expect < 50% of cows to come into estrus during that 10-d window (10/21=0.48). And I would expect 40% of those who do come into estrus, not to be detected. That leaves roughly 70% of cows to proceed to Ovsynch. But when I looked through the events, I didn't see a single cow go into Ovsynch for her first AI event (they were only bred on TAI through resynch).

Hi Julie, I can answer the first point first. We might have glossed over this point. Currently, we do estrus detection **daily** during the period between VWP and OvSynch Start Day. This means it almost guarantees we will detect estrus the longer the interval is and the more often we check it. An alternative is to designate only one or two estrus days (doing some calculations) in this interval and detect estrus. This will likely lead to what you expect. Please let me know if you want this alternative.

ltp23:
> a- VWP: update the description to match the updated meaning, and include "N/A" option (to use for TAI protocol)
> b- estrus_detection_rate and estrus_conception_rate: properties describe that these values are only relevant for ED programs. Is that true?
> c- presynch_program_start_day and ovsynch_program_start_day: it would be nice to have the option of "N/A" for when these are not in use

I can make updates to these suggestions.

jadamchick:
Hi Loi! The estrus detection should be whatever the farm input is for % of estrus events that are detected and lead to an AI event. Happy to chat more tomorrow if you have more questions on that. Thanks!

krb78:
Hi Loi!

The way you currently have estrus being detected every day is correct. I think Julie is commenting that the % of cows found in estrus during a 10 day window seemed off given the length of the estrus cycle is 21 days with a standard deviation of 6.

Kaitlyn Briggs, DVM, MBA
C: 607.386.0040

If you receive an email from me outside of your own working hours, please do not feel pressure to respond outside of your own work schedule.
________________________________

jadamchick:
> Glad to see there is progress on this front! One comment in response to Julie's point 2 above is that other than during initialization, there should not be cows entering the herd at this time period (between calving and reproduction). My guess it is a misalignment between the cows in the herd database and the desired simulation.
> 
> Two thoughts for fixing this are 1) look at the methods for herd initialization and modify something related to repro, 2) add an issue about this and ignore for now as it will sort it self out by the end of the simulation.
> 
> @jadamchick, one workaround thought for you would be to initialize the herd when running the simulations (last time I checked it was much much quicker) in which case the repro methods will align and the ideal outcome is that these gaps in achieving pregnancy won't happen

Thanks @KFosterReed -- I wasn't sure how impactful this is/isn't. If we are running a herd for 4 years and only looking at the last year, then is it safe to assume that this subset of animals will no longer be relevant (since, as I understand it, they all enter on simulation day 1)? 
I will create an issue about this to flag it as something to revisit but not urgently needed during this refactor. 

Issue # 1157 created here: https://github.com/RuminantFarmSystems/MASM/issues/1157

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -685,585 +671,963 @@ def update(self, sim_day, calving_interval): # noqa
         self.update_milk_production_history(sim_day)
 
         if not self.do_not_breed:
```
KFosterReed, line 685: I am just reading through the code a bit to see if I can find any logic that might not make sense and it seems like this will start repro programs for cows that haven't yet calved? I ran it with a stop here and at least for cows entering the herd, the execute_ed_tai_protocol() function is being called before they have given birth

input/metadata/default_metadata.json
```diff
@@ -366,7 +366,7 @@
             "description": "Repro Management",
             "voluntary_waiting_period": {
               "type": "number",
-              "description": "Voluntary Waiting Period (days) -- If TAI (Timed Artificial Insemination) protocol is chosen for cows, voluntary waiting period (VPW) is days after parturition at which the 1st hormonal injection is made. If ED (Estrus Detection) or ED-TAI (combination of Estrus Detection and TAI) is chosen for cows, VWP is days after parturition when a cow is eligible to be bred during estrus",
+              "description": "Voluntary Waiting Period (days) -- If 'TAI' (Timed Artificial Insemination) is chosen for cows, it is the number of days after parturition when the first hormonal injection is made in a presynch program (if present) followed by a TAI program, and if a presynch program is not present, a TAI program will be moved up and start earlier. If 'ED' or 'ED-TAI' is chosen for cows, it is the number of days after parturition when estrus is started to be monitored.",
```
jadamchick: This should say something like: 
        When the cow's days in milk has reached this day, monitoring for estrus will begin. Used only in the ED and ED-TAI protocols. 

What value should be given when using TAI protocol? 0?

input/metadata/default_metadata.json
```diff
@@ -445,126 +455,54 @@
               "type": "object",
               "estrus_detection_rate": {
                 "type": "number",
-                "description": "Estrus Detection Rate -- The % of in-heat animals that would be detected in the ED (Estrus Detection) programs (default is 0.6)",
+                "description": "The percentage of in-heat animals that would be detected in the ED (Estrus Detection) programs (default is 0.6)",
```
jadamchick: could add something like: 
"The percentage of in-heat animals that would be detected in the ED (Estrus Detection) programs, including during the estrus detection portion of ED-TAI programs (default is 0.6)"

input/metadata/default_metadata.json
```diff
@@ -445,126 +455,54 @@
...
                 "default": 0.6,
                 "minimum": 0,
                 "maximum": 1
               },
               "estrus_conception_rate": {
                 "type": "number",
-                "description": "Estrus Conception Rate -- The % of inseminated animals that would be concepted in the ED (Estrus Detection) programs (default is 0.6)",
+                "description": "The percentage of inseminated animals that would be concepted in the ED (Estrus Detection) program (default is 0.6)",
```
jadamchick: same as above (include reference to ED-TAI). 
and maybe add something like "Conception rate for TAI protocols is specified under "Ovsynch_program_conception_rate" "

input/metadata/default_metadata.json
```diff
@@ -445,126 +455,54 @@
...
+                "description": "Cow's OvSynch program to be used in both TAI and ED-TAI protocols.",
+                "pattern": "^(OvSynch 48|OvSynch 56|CoSynch 72|5d CoSynch|None)$",
+                "default": "OvSynch 56"
               },
-              "ed_conception_rate": {
+              "ovsynch_program_start_day": {
                 "type": "number",
-                "description": "ED Conception Rate (percent) -- The percentage of serviced animals that become pregnant in the ED (Estrus Detection) protocol ",
-                "default": 0.6,
+                "description": "OvSynch program start day is the day of the first GnRH injection in an OvSynch program. It is used in both TAI and ED-TAI repro protocols.",
```
jadamchick: Add more description to describe how this works with relation to the presynch program (what you showed me today from the notes in the code seemed really helpful)

input/metadata/default_metadata.json
```diff
@@ -445,126 +455,54 @@
               "type": "object",
               "estrus_detection_rate": {
                 "type": "number",
-                "description": "Estrus Detection Rate -- The % of in-heat animals that would be detected in the ED (Estrus Detection) programs (default is 0.6)",
+                "description": "The fraction of in-heat cows that would be detected in the ED (Estrus Detection) protocol, including during the estrus detection portion of ED-TAI protocols. When TAI protocol is used, this value will be ignored, and it is recommended to set it to 0.",
                 "default": 0.6,
                 "minimum": 0,
                 "maximum": 1
               },
               "estrus_conception_rate": {
```
jadamchick: Can we rename "estrus_conception_rate" to something like "ED_conception_rate" to make it clear that this is the conception rate used for estrus detection?

input/metadata/default_metadata.json
```diff
@@ -445,126 +455,54 @@
...
-                "description": "5d CoSynch Conception Rate (percent) -- The conception rate for cows in the 5d CoSynch TAI (Timed Artificial Insemination) protocol",
-                "minimum": 0,
-                "maximum": 1,
-                "default": 0.6
-              },
-              "cow_resynch_protocol": {
+              "resynch_program": {
                 "type": "string",
-                "description": "Cow Resynch Protocol (select one) -- Resynch protocol for cows before/at/after pregnancy check day in the TAI (Timed Artificial Insemination) protocol",
+                "description": "Resynch program for cow before/at/after first pregnancy check day in the TAI and ED-TAI repro protocols. When the ED protocol is used, this value will be ignored, and it is recommended to set it to None.",
```
jadamchick: Can we please add one statement like: "If PGFatPD is chosen, the values for estrus detection and conception to that detected estrus following PGF are determined by "estrus_detection_rate" and "ED_conception_rate" (even for TAI protocol).

jadamchick (approved):
I approve!! What a work of perseverance, Loi :) 

TAI: I ran several farms with varying presynch, ovsynch, and resynch combinations for TAI and VWP=0. Everything seems to work as expected (including that OvSynch will not start until after Presynch is complete, even if the start day is specified as earlier). I noted that if PGFatPD is chosen for the resynch protocol, it will institute estrus detection, so it is important to have values filled in for ED detection + conception in that case.

ED-TAI: I ran several combinations of VWP, Ovsynch start day, and resynch protocols, and all worked as expected. Even when VWP was later than OvSynch start day, it appropriately skipped estrus detection before the first TAI and then started detecting after the first AI event occurred. 

ED: no hiccups noticed. 

Other eyes would be valuable at this point, since Loi has spent a lot of time addressing my observations and I'm sure I have blind spots that may have slipped past!  That said, I feel quite confident with what is here. Great work all. 

(I did make a couple suggested improvements to metadata for clarity on the TAI - PGFatPD situation above, but not dealbreakers).

RUFAS/routines/animal/animal_manager.py
```diff
@@ -1736,3 +1738,67 @@ def _record_heiferIIs_conception_rate(self):
...
+
+        info_map = {"class": self.__class__.__name__, "function": self._record_culling_stats.__name__}
+        om.add_variable('num_cows_by_death_cull',
+                        self.life_cycle_manager.cull_reason_stats_range[
+                            animal_constants.DEATH_CULL
+                        ],
+                        info_map)
+        om.add_variable('num_cows_by_low_prod_cull',
+                        self.life_cycle_manager.cull_reason_stats_range[
+                            animal_constants.LOW_PROD_CULL
```
krb78, line 1765: I just want to say that I'm happy we've added this 😃 some of the feedback we got from farms as they filled out the pilot testing paperwork was "where do we put these low production culls?" and I always had to explain it was accounted for, but it'll be helpful to have this as an OM variable :)

input/metadata/ARL_metadata.json
```diff
@@ -359,7 +359,7 @@
             "description": "Repro Management",
             "voluntary_waiting_period": {
               "type": "number",
-              "description": "Voluntary Waiting Period (days) -- If TAI (Timed Artificial Insemination) protocol is chosen for cows, voluntary waiting period (VPW) is days after parturition at which the 1st hormonal injection is made. If ED (Estrus Detection) or ED-TAI (combination of Estrus Detection and TAI) is chosen for cows, VWP is days after parturition when a cow is eligible to be bred during estrus",
+              "description": "Voluntary Waiting Period (days) -- When the cow's days in milk has reached this day, monitoring for estrus will begin. Used only in the ED and ED-TAI protocols. When TAI protocol is used, this value will be ignored, and it is recommended to set it to 0.",
```
krb78: can we say "monitoring for estrus and subsequent breeding if found, will begin"? Because we were sort of monitoring before but we don't do anything other than say "before VWP, wait for next estrus" if she comes into estrus before the VWP.

input/metadata/ARL_metadata.json
```diff
@@ -438,126 +448,54 @@
...
+                "pattern": "^(OvSynch 48|OvSynch 56|CoSynch 72|5d CoSynch|None)$",
+                "default": "OvSynch 56"
               },
-              "ed_conception_rate": {
+              "ovsynch_program_start_day": {
                 "type": "number",
-                "description": "ED Conception Rate (percent) -- The percentage of serviced animals that become pregnant in the ED (Estrus Detection) protocol ",
-                "default": 0.6,
+                "description": "When set properly, OvSynch program start day is the day of the first GnRH injection in an OvSynch program. When the OvSynch program follows a presynch program, the start day of the OvSynch program should be on or after the last day of the presynch program. Otherwise, the OvSynch program will be shifted to start on the last day of the presynch program. Because the OvSynch program is only supported in TAI protocol, this value will be ignored when other protocols such as ED or ED-TAI are used, and it is recommended to set it to 0.",
```
krb78: Wait, Loi, isn't ovsynch_program_start_day used in ED-TAI? Say we set VWP=50 and ovsynch_program_start_day to 70, that works and whatever cows we don't find in estrus between 50 and 70 days get started on an ovsynch program. I just tested this, so I know this is true and I'm hoping the language here is just a typo 😅

krb78 (approved):
Woof! Amazing job, @ltp23 🥳 🕺 💃 Not that I didn't believe Julie, but I also ran a half dozen iterations (real ones provided by Blake) and weird fake ones to see if I could break it haha. It seems to work well! Great job! You deserve a full week of rest after this project! Who knew cow reproduction was so complicated?! 😆  

I just had 2 quick comments on wording otherwise!

YijingGong (approved):
LGTM! Really good job Loi! One minor thing: I found the event log is empty when I put "start_date": "2009:244","end_date": "2010:264", but if I put a longer time such as 3 years or 5 years, I can see the animal events. 

I've tested:
1) cow ED:
Glad that this works:  "cows": {
					"estrus_detection_rate": 0.65,
					"ED_conception_rate": 0.6,
					"presynch_program": "N/A",
					"presynch_program_start_day": 0,
          			"ovsynch_program": "N/A",
					"ovsynch_program_start_day": 0,
					"ovsynch_program_conception_rate": 0,
          			"resynch_program": "N/A"
				}
2) TAI: zero estrus happening when resynch is TAIafterPD or TAIbeforePD
3) ED-TAI works well with VWP =50, presynch_start_day and ovsynch_start_day varied

ltp23:
> I found the event log is empty when I put "start_date": "2009:244","end_date": "2010:264", but if I put a longer time such as 3 years or 5 years, I can see the animal events.

Thanks Yijing, before merging, I will adjust to only output events on the last day of the simulation.

RUFAS/routines/animal/animal_manager.py
ltp23: During debugging, a high count of culled cows due to low milk production was a sign of something going wrong in the rebreeding.

input/metadata/ARL_metadata.json
ltp23: Good catch, Kaitlyn. Yes, this value is used in both TAI and ED-TAI.

PR #1057, opened 2023-12-15
Purchased feed emissions footprint report

Isgotkowitz:
This PR introduces the ability to calculate the emissions associated with purchased feed that is fed to animals simulated by RuFaS. It does this through of the `FeedEmissionsManager`, which is contained within the `AnimalManager`, and utilizes a CSV input that contains a mapping of emissions (kg) per feed dry matter (kg).

At the beginning of the simulation, a Feed Emissions Manager is initialized by the Animal Manager. When it is initialized, it:
- Pulls location information from a field input file.
- Then gets the FIPS (county) code associated with the location via the FCC's API for getting the county FIPS code by lat/long.
- Builds a mapping between RuFaS feed IDs and the emissions associated with each.

Everyday, when the amounts of feed are reported per pen, the amounts are passed to a reporting method that calculated the emissions per feed per pen. To get the feed emissions, use the filter:

```AnimalModuleReporter.report_daily_feed_emissions.pen_.*```

I have added the feed emissions data to the `input/data/EEE` directory, and wrote a new properties section in the metadata.

### _Notes_
- If the is no lat/long provided to the simulation, the location defaults to Madison, WI.
- **Updated** The FCC API fails a lot (~50% of the time with a 502 Bad Gateway error). `FeedEmissionsManager` requests the county code three times, and if it does not get an ok response it adds an error to the OutputManager and the county code defaults to 55025 (Dane County, WI). **Update** now it is working super well, not sure what was going on with it over the weekend.
- If a feed is not in the feed emissions data a Warning is raised to the Output Manager.
- It is assumed that feed emissions data is all in kg emissions / kg dry matter feed.

### CODE REVIEW

File: RUFAS/routines/animal/animal_module_reporter.py [hunk 1]
```diff
@@ -168,12 +168,44 @@ def report_daily_ration(animal_manager) -> None:
...
+            The unique number identifying a pen.
+        pen_animal_name : str
+            The name of the animal combination in a pen.
+        feed_emissions_manager : FeedEmissionsManager
+            The Feed Emissions Manager for the Animal Manager.
+
+        """
+        info_map = {
+            "class": "AnimalModuleReporter",
+            "function": "report_daily_feed_emissions",
```
PooyaHekmati: avoid hardcoded values

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -168,12 +168,44 @@ def report_daily_ration(animal_manager) -> None:
...
+        }
+
+        daily_feed_emissions = animal_manager.feed_emissions_manager.\
+            create_daily_purchased_feed_emissions_report(ration_total)
+        om.add_variable(
+            f"ration_daily_feed_emissions_totals_for_pen_{pen_id}_{pen_animal_name}",
```
PooyaHekmati: avoid formatting the data that is sent to downstream systems, unless they expect you to do so.
```suggestion
            f"pen_{pen_id}_animal_{pen_animal_name}_feed_emissions",
```
We don't have subtotal or weekly or monthly, so no need to specify. If clarity is needed, you can add them to `info_map`.

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
+import requests
+
+from RUFAS.input_manager import InputManager
+from RUFAS.output_manager import OutputManager
+
+im = InputManager()
+om = OutputManager()
+
+
+class FeedEmissionsManager:
```
PooyaHekmati: This class seems to _estimate_ or _calculate_ the emissions that are caused by the purchased feed, it doesn't _manage_ it.
```suggestion
class PurchasedFeedEmissionsEstimator:

```

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+
+class FeedEmissionsManager:
+
+    def __init__(self):
+        latitude, longitude = self._get_geographic_coordinates()
+
+        try:
+            self.county_code = self._get_county_code(latitude, longitude)
+        except requests.exceptions.RequestException:
+            self.county_code = 55025
```
PooyaHekmati: avoid magic numbers

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self.__init__.__name__
+        }
+        om.add_variable("FIPS_code", self.county_code, info_map)
```
PooyaHekmati: maybe call it `FIPS_county_code`?

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+        return emissions_dict
+
+    def _get_geographic_coordinates(self) -> (float, float):
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self._get_geographic_coordinates.__name__
+        }
+
+        latitude = 43.073
+        longitude = -89.401
```
PooyaHekmati: avoid magic numbers

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+
+        field_keys: list[str] = im.get_data_keys_by_properties("field_properties")
+
+        if not field_keys:
+            warning_name = "Default Feed Emissions data"
+            warning_message = "Could not obtain feed emissions geographic attributes, defaulting to Madison, WI."
+            om.add_warning(warning_name, warning_message, info_map)
```
PooyaHekmati:
```suggestion
            om.add_warning("Default Feed Emissions data", 
            "Could not obtain feed emissions geographic attributes, defaulting to Madison, WI.",
            info_map
            )

```

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+
+            return {"latitude": latitude, "longitude": longitude}
+
+        field_key = field_keys[0]
```
PooyaHekmati: why index 0? What is its significance?

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+
+        latitude_data_address = f"{field_key}.absolute_latitude"
+        latitude = im.get_data(latitude_data_address)
```
PooyaHekmati:
```suggestion
        latitude = im.get_data(f"{field_keys[0]}.absolute_latitude")
```

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+
+        longitude_data_address = f"{field_key}.longitude"
+        longitude = im.get_data(longitude_data_address)
```
PooyaHekmati:
```suggestion
        longitude = im.get_data( f"{field_keys[0]}.longitude")

```

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+        longitude = abs(longitude) * -1
```
PooyaHekmati: what?? Why?!

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+
+        return latitude, longitude
+
+    def _get_county_code(self, latitude: float, longitude: float) -> int:
+        endpoint = "https://geo.fcc.gov/api/census/block/find"
+        params = {"latitude": latitude, "longitude": longitude, "format": "json"}
+
+        responses = []
+        for _ in range(3):
```
PooyaHekmati:
```suggestion
        retries = 3
        for _ in range(retries):
```
avoid magic numbers

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+            response = requests.get(endpoint, params=params)
```
PooyaHekmati: add a log to OM before and after making the API call

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+            if response.status_code == 200:
+                answer = response.json()
+                break
```
PooyaHekmati:
```suggestion
                return int(answer["County"]["FIPS"])

```

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+            }
+            error_name = "Bad API responses"
+            error_message = f"Responses: {responses}"
+            om.add_error(error_name, error_message, info_map)
+            raise requests.exceptions.RequestException(
+                f"Bad API response: {[response.text for response in responses]}"
+            )
+
+        county_code = int(answer["County"]["FIPS"])
+        return county_code
```
PooyaHekmati: simply log the API events when they happen, there is no reason to cache them and then process them. It's actually better to not make API calls in rapid succession when there is a failure, give the server some time to change its state; logging API events right after the API call makes a slight delay between API calls.

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def _setup_feed_emissions(self) -> dict[str, float]:
```
PooyaHekmati: This function really needs docstrings to explain how it works

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+        emissions_dict = {}
+
+        for feed_id, amount_fed in daily_feed_totals.items():
+            if feed_id == "dry_matter_intake_total":
+                continue
+            if feed_id in self.missing_feed_ids:
+                continue
+            elif feed_id not in self.feed_emissions.keys():
+                om.add_warning(
+                    "Purchased Feed Emissions Warning",
```
PooyaHekmati:
```suggestion
                    "Missing Purchased Feed Emissions",

```

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+
+        total_emissions = 0.0
+        emissions_dict = {}
+
+        for feed_id, amount_fed in daily_feed_totals.items():
+            if feed_id == "dry_matter_intake_total":
+                continue
+            if feed_id in self.missing_feed_ids:
+                continue
+            elif feed_id not in self.feed_emissions.keys():
```
PooyaHekmati:
```suggestion
            if feed_id not in self.feed_emissions.keys():

```

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+                    f"Missing purchased feed emissions data for RuFaS feed {feed_id}.",
+                    info_map
+                )
+                self.missing_feed_ids.append(feed_id)
+                continue
+            emissions = amount_fed * self.feed_emissions[feed_id]
+            total_emissions += emissions
+            emissions_dict[feed_id] = emissions
+        emissions_dict["feed_emissions_total"] = total_emissions
+        return emissions_dict
```
PooyaHekmati:
```suggestion
            emissions_per_feed_id["feed_emissions_total"]+= emissions
            emissions_per_feed_id[feed_id] = emissions
        return emissions_per_feed_id

```

RUFAS/routines/animal/feed_emissions_manager.py
```diff
@@ -0,0 +1,137 @@
...
+            Maps the feed type to the amount of emissions generated in producing and delivering it (feed ID: kg CO2).
+
+        """
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self.create_daily_purchased_feed_emissions_report.__name__
+        }
+
+        total_emissions = 0.0
+        emissions_dict = {}
```
PooyaHekmati:
```suggestion
        emissions_per_feed_id={"feed_emissions_total": 0}

```

PooyaHekmati (changes requested):
The PR is on the right track but there are some changes needed before it can be merged onto main, even for a quick implementation. 
If the API is unreliable, since we already know the location of the pilot farms, can we plugin the county code into input files instead of making the API call at runtime? There is still value in having the automated API calls for scenarios other than the pilot farms, but since `Explicit is better than implicit` maybe we want to explore that avenue.
Also, until we fix the metadata structure, a change to `properties` mus be applied to all metadata files.

RUFAS/routines/animal/feed_emissions_manager.py
Isgotkowitz: On the very off chance that someone tries to simulate a farm located at a longitude in the Eastern hemisphere.
This is not handled well though, at the very least an error should be raised.

RUFAS/routines/animal/animal_module_reporter.py
Isgotkowitz: This is a little wonky because there is no `self`. I have changed it to
```python
        info_map = {
            "class": AnimalModuleReporter.__name__,
            "function": AnimalModuleReporter.report_daily_feed_emissions.__name__,
        }
```

RUFAS/routines/animal/feed_emissions_manager.py
Isgotkowitz: Because there might be multiple field keys returned - one for each simulated field. The only thing that is known is at that point in the execution is that `field_keys` is not an empty list, so it just picks the first one.

RUFAS/routines/animal/feed_emissions_manager.py
Isgotkowitz: Added a docstring, if it isn't clear I can hit it again.

Isgotkowitz:
Update: the API seems to be working perfectly now.

Isgotkowitz:
> If the API is unreliable, since we already know the location of the pilot farms, can we plugin the county code into input files instead of making the API call at runtime? There is still value in having the automated API calls for scenarios other than the pilot farms, but since `Explicit is better than implicit` maybe we want to explore that avenue. 

As noted in the update, the API seems to be working really well now. Not sure what was going on with it before. But I think passing in the county code is a good idea. I was thinking about this, and one hypothetical situation I thought of is a farmer whose farm is at the edge of a county, and feed is cheaper in the neighboring county, so that's where they buy their feed. In this case it would be good to allow them to directly set the county code.

> Also, until we fix the metadata structure, a change to `properties` mus be applied to all metadata files.

Done.

RUFAS/routines/animal/feed_emissions_manager.py
PooyaHekmati: I know, we have talked about it. But the one who will read and maintain the code won't. So the question is coming from that perspective rather mine.

RUFAS/routines/animal/feed_emissions_manager.py
PooyaHekmati: So, if user plugins 42.4955242, 76.4205981 (lat, long) which belong to a village called Toru Aygyr in Kyrgyzstan, we convert it to 42.4955242, -76.4205981, which happens to be near Ithaca, fetch Tompkins County's ID, get the feed emissions according to Tompkin County and keep going without even letting user know about what just happened? That doesn't sound like a correct strategy.

RUFAS/routines/animal/purchased_feed_emissions_estimator.py
```diff
@@ -0,0 +1,161 @@
...
+        info_map = {
+            "class": self.__class__,
+            "function": self._get_county_code.__name__
+        }
+
+        endpoint = "https://geo.fcc.gov/api/census/block/find"
+        params = {"latitude": latitude, "longitude": longitude, "format": "json"}
+
+        attempts = 3
+        for attempt_count in range(attempts):
```
PooyaHekmati:
```suggestion
        max_attempts = 3
        for attempt_count in range(max_attempts):

```

RUFAS/routines/animal/purchased_feed_emissions_estimator.py
```diff
@@ -0,0 +1,161 @@
...
+                "Feed Emissions API Call",
+                f"Calling the FCC's FIPS County Code API, attempt: {attempt_count}",
+                info_map
+            )
+            response = requests.get(endpoint, params=params)
+            if response.status_code == 200:
+                answer = response.json()
+                om.add_log(
+                    "Successful Feed Emissions API Call",
+                    f"Got successful response from the FCC's FIPS County Code API on attempt: {attempt_count}",
```
PooyaHekmati:
```suggestion
                    f"Got successful response from the FCC's FIPS County Code API on {attempt_count=}",

```

RUFAS/routines/animal/purchased_feed_emissions_estimator.py
```diff
@@ -0,0 +1,161 @@
...
+                    info_map
+                )
+                return int(answer["County"]["FIPS"])
+            om.add_error(
+                "Error Feed Emissions API Call",
+                f"FCC's FIPS County Code API attempt: {attempt_count} failed with status code: {response.status_code}.",
```
PooyaHekmati:
```suggestion
                f"FCC's FIPS County Code API {attempt_count=} failed with {response.status_code=}.",

```
you can shorten this f-strings like so.

PooyaHekmati (changes requested):
The main reason that I am requesting changes is the line `longitude = abs(longitude) * -1`; that is a very dangerous move.

> As noted in the update, the API seems to be working really well now. Not sure what was going on with it before. But I think passing in the county code is a good idea. I was thinking about this, and one hypothetical situation I thought of is a farmer whose farm is at the edge of a county, and feed is cheaper in the neighboring county, so that's where they buy their feed. In this case it would be good to allow them to directly set the county code.

Right now the focus is on getting results for pilot farms. Evaluating different purchase strategies is something that we will explore in the future. wrt the API, there is no guarrantee that it won't become unresponsive again. So I think it is important to allow the possiblity of passing the county code directly and not mandating _either let me call the API or I force Madison WI to your results_.

Isgotkowitz:
> The main reason that I am requesting changes is the line `longitude = abs(longitude) * -1`; that is a very dangerous move.

@PooyaHekmati Ok, what is the best way to move forward then? A couple options that come to mind are:
- Adjust the metadata properties so that the longitude has to be in Western hemisphere.
- Assume the user passed a valid longitude and use it as is.

> Right now the focus is on getting results for pilot farms. Evaluating different purchase strategies is something that we will explore in the future. wrt the API, there is no guarrantee that it won't become unresponsive again. So I think it is important to allow the possiblity of passing the county code directly and not mandating _either let me call the API or I force Madison WI to your results_.

Would the config file be a good place to put the county code?

Isgotkowitz:
> > The main reason that I am requesting changes is the line `longitude = abs(longitude) * -1`; that is a very dangerous move.
> 
> @PooyaHekmati Ok, what is the best way to move forward then? A couple options that come to mind are:
> 
> * Adjust the metadata properties so that the longitude has to be in Western hemisphere.
> * Assume the user passed a valid longitude and use it as is.

Assume the longitude is valid, and make sure to check for null response from API.

> > Right now the focus is on getting results for pilot farms. Evaluating different purchase strategies is something that we will explore in the future. wrt the API, there is no guarrantee that it won't become unresponsive again. So I think it is important to allow the possiblity of passing the county code directly and not mandating _either let me call the API or I force Madison WI to your results_.
> 
> Would the config file be a good place to put the county code?

Yes.

jadamchick (commented):
This is looking great! I had to add FIPS_county_code to the config file to get it to run, and looks like you just added that too.

- Eventually this CSV will contain all feeds in the feed library -- so we will need to add every feed individually to the metadata properties?
- Warnings look good for alerting when a feed is missing emissions data.
- I created an altered CSV (changed some values and added an additional feed) and used that easily; I had to add the new feed ID to the metadata properties in order for it to be utilized. Once I did that, it appropriately grouped those feed emissions in with the right pen's reporting and total.
- I'm sure you're already aware that we will eventually need to distinguish between which feeds are purchased and which are grown on-farm (identified in the input feed.json). For pilot farms right now, we are assuming all feeds are purchased. For the next iteration (when we incorporate crop & soil module), will need to be able to determine which are purchased to use the emissions factors from the CSV, and which are home grown (and calculate footprint using RuFaS). This works great for our scenario right now.
- I am able to use the report generator to slice and aggregate the feed emissions data. It is a little fiddly (need to specify each feed individually and make sure it is labeled in the title in order to know in the output which is which). Copying my example filter here FYI. It would be slick to be able to do this a little more smoothly, but this is functional for what we need.

```json
{
    "multiple":[
        {
            "name": "footprint by ingredient - 51 lact",
            "filters": [
                "AnimalModuleReporter.report_daily_feed_emissions.pen_3_animal_LAC_COW_feed_emissions"
            ],
            "variables": "51", 
            "slice_start": -365, 
            "vertical_aggregation": "sum"
        },
        {
            "name": "footprint by ingredient - 104 lact",
            "filters": [
                "AnimalModuleReporter.report_daily_feed_emissions.pen_3_animal_LAC_COW_feed_emissions"
            ],
            "variables": "104", 
            "slice_start": -365,
            "vertical_aggregation": "sum"
        },
        {
            "name": "footprint by ingredient - total lact",
            "filters": [
                "AnimalModuleReporter.report_daily_feed_emissions.pen_3_animal_LAC_COW_feed_emissions"
            ],
            "variables": "feed_emissions_total", 
            "slice_start": -365,
            "vertical_aggregation": "sum"
        }
    ]
} 
```

Isgotkowitz:
> * Eventually this CSV will contain all feeds in the feed library -- so we will need to add every feed individually to the metadata properties?

Yes, I think so. I have not worked extensively with the validation mechanism for tabular (CSV) data in the Input Manager, but I think it is both a requirement and a best practice to define properties for every column.

> * I'm sure you're already aware that we will eventually need to distinguish between which feeds are purchased and which are grown on-farm (identified in the input feed.json). For pilot farms right now, we are assuming all feeds are purchased. For the next iteration (when we incorporate crop & soil module), will need to be able to determine which are purchased to use the emissions factors from the CSV, and which are home grown (and calculate footprint using RuFaS). This works great for our scenario right now.

Yes. I am not sure how this will work, but at that point it may be good to move Purchased Feed Emissions Estimator closer to wherever feed is being "purchased".

PooyaHekmati:
> This is looking great! I had to add FIPS_county_code to the config file to get it to run, and looks like you just added that too. 
> 
> - Eventually this CSV will contain all feeds in the feed library -- so we will need to add every feed individually to the metadata properties? 

Yes, I know that is labor-intensive, we can/should improve it.

> - I'm sure you're already aware that we will eventually need to distinguish between which feeds are purchased and which are grown on-farm (identified in the input feed.json). For pilot farms right now, we are assuming all feeds are purchased. For the next iteration (when we incorporate crop & soil module), will need to be able to determine which are purchased to use the emissions factors from the CSV, and which are home grown (and calculate footprint using RuFaS). This works great for our scenario right now. 

Yes, that fits with the longer-term plan. Currently, there is a disconnect between the soil and crop module and the animal module. Once we finish implementing the feed storage module, that gap is bridged. Then we can/should move the purchasing feed logic from the animal module to the feed module. Quoting myself from a few days ago:
>_Just to have it on your radar, I am implementing a class called `FeedManager`. Right now it just does the inventory tracking and later on it will handle degradation as well (as we discussed). However, a longer term plan is to move purchasing feed logic to this class as well. The rationale is that the Animal module should be agnostic about the origin of the feeds. As far as Animal manager is concerned, the animals need to receive a certain amount of nutrition, where it comes from is a different question. This separation is a move toward SOLID principles and makes it possible/easier to experiment with different storage vs purchase strategies, incorporating econ and degradation._


> - I am able to use the report generator to slice and aggregate the feed emissions data. It is a little fiddly (need to specify each feed individually and make sure it is labeled in the title in order to know in the output which is which). Copying my example filter here FYI. It would be slick to be able to do this a little more smoothly, but this is functional for what we need.
> 
Yes, I added a comment to issue #899 to make sure this will ba taken care of as well.

RUFAS/routines/animal/purchased_feed_emissions_estimator.py
```diff
@@ -0,0 +1,191 @@
...
+            "class": self.__class__,
+            "function": self._get_county_code.__name__
+        }
+
+        county_code_from_input = im.get_data("config.FIPS_county_code")
+        if county_code_from_input is not None:
+            return county_code_from_input
+
+        om.add_warning(
+            "Feed Emissions Estimator found invalid data.",
```
PooyaHekmati: Knowing that county code is optional in the input data, would it be a warning when it is not provided? Maybe log suffices? Also, the missing data here is not invalid, right?

RUFAS/routines/animal/purchased_feed_emissions_estimator.py
```diff
@@ -0,0 +1,191 @@
...
+            "FIPS county code pulled from the Input Manager's pool was null.",
+            info_map
+        )
+
+        location = self._get_geographic_coordinates()
+        if not location:
+            om.add_warning(
+                "Using default FIPS county code",
```
PooyaHekmati: I think we should mention the cause in the subject and talk about the effect/consequence in the body.

RUFAS/routines/animal/purchased_feed_emissions_estimator.py
```diff
@@ -0,0 +1,191 @@
...
+                f"Could not find simulated location coordinates, defaulting to {DEFAULT_FIPS_COUNTY_CODE=}.",
+                info_map
+            )
```
PooyaHekmati, line 105: I think an exception should be raised here so that the caller uses the default value.

RUFAS/routines/animal/purchased_feed_emissions_estimator.py
```diff
@@ -0,0 +1,191 @@
...
+
+        latitude = location[0]
+        longitude = location[1]
```
PooyaHekmati, line 109: these can be added to info_map

RUFAS/routines/animal/purchased_feed_emissions_estimator.py
```diff
@@ -0,0 +1,191 @@
...
+
+        endpoint = "https://geo.fcc.gov/api/census/block/find"
```
PooyaHekmati, line 111: info_map

PooyaHekmati (approved):
Just consider the comments before merging.
[LGTM](https://lgtm.lol/i/811)

RUFAS/routines/animal/purchased_feed_emissions_estimator.py
Isgotkowitz: > would it be a warning when it is not provided? Maybe log suffices?

Yes, good point.

> Also, the missing data here is not invalid, right?

Yes and no. A null FIPS code is never a valid FIPS code, but the way the metadata properties are set up it is not an unexpected case to have a `null` FIPS code in the Input Manager.

RUFAS/routines/animal/purchased_feed_emissions_estimator.py
Isgotkowitz: Nice catch, that actually needs to be in there or it will crash.

jadamchick (approved):
Looks great! I'll keep playing with it with real pilot farms as we get them running and let you know if anything comes up. Thanks for getting this in place so quickly.

RUFAS/routines/animal/purchased_feed_emissions_estimator.py
PooyaHekmati: > > Also, the missing data here is not invalid, right?
> 
> 
> Yes and no. A null FIPS code is never a valid FIPS code, but the way the metadata properties are set up it is not an unexpected case to have a null FIPS code in the Input Manager.
> 

`missing data is not invalid` doesn't mean it is valid. In order to make the statement that something is either valid or invalid, that thing has to exist first. When it doesn't, it is both valid and invalid at the same time and it is neither of them as well, we just don't know. Imagine you are hungry, I tell you why don't you eat something? You say the food that I have is rotten. This is an invalid scenario. Another possibility is when you say I don't have any food. In this case, you don't care the possibility of food poisoning when there is no food.

PR #1058, opened 2023-12-17
Quickfix for linting script

Isgotkowitz:
Removed extra quotation marks from final line of linting shell script.

The only change in this PR is going from 
```sh
flake8 "$changed_files"
```
to
```sh
flake8 $changed_files
```

I don't understand the exact mechanics of how this shell script works, but when the quotation marks are added the script does not work correctly with multiple changed files - but does work with a single changed file, which is how I think this slipped through the cracks.


ew3361zh (approved):
Nice catch! Tested this on main (current version) with 2 files that had linting violations and the script wasn't working. This patch seemed to fix it - script ran and identified the linting issues with multiple files.

Isgotkowitz:
> Nice catch! Tested this on main (current version) with 2 files that had linting violations and the script wasn't working. This patch seemed to fix it - script ran and identified the linting issues with multiple files.

Thanks for checking it out Niko

ltp23 (approved):
Thanks Ed. 

I think the difference matters only when the filenames have spaces or something that would make the name parsing yield multiple parts.

The version with the double-quotes will treat the variable string as one file even if there are multiple parts in it. The version with no quotes will treat each white-space-delimited part in the variable string as a separate file.

Isgotkowitz:
> I think the difference matters only when the filenames have spaces or something that would make the name parsing yield multiple parts.

I think you're right Loi - in this case it seems like the newlines are being missed.

PR #1063, opened 2023-12-19
Crop Evaluation

### LINKED ISSUE #1030: Evaluate and correct Crop growth/yields
The growth of crops and the yields they produce needs to be evaluated. This includes evaluating:
- crop growth (biomass accumulation, leaf area index, root development, etc)
- water uptake
- nutrient uptake
- yield accuracy (mass, nutrient composition, moisture)

In order to properly accomplish this, a number of different scenarios need to be simulated and checked. See [this document](https://docs.google.com/document/d/1gAhYHYnwUTPvVEY6M0i5WoKNCRVVILL_w6lZJNvBaWU/edit) for details on the scenarios that need to simulated. Before these scenarios can be simulated, a number of different crop configurations need to be added, as outlined in the table on pages 4-5 of [this document](https://docs.google.com/document/d/1auWeT6XnItd8yLoWGkbSwJ2DkUULxv5ClA48Grm69IE/edit#heading=h.h0dz48r2a92t).

Isgotkowitz:
This PR completes the V1 crop evaluation of RUFAS. There have been many bug fixes made around the codebase, the most important one being in `crop_management.py` (wet and dry yields were being switched around). If there are any changes that you would like an explanation on, please leave a comment and I will be happy to explain. All changes have had associated tests updated.

All new input files associated with the crop evaluation have been preserved in the new directory `model_evaluation/soil_crop_evaluation/eval_inputs` directory. I recommend briefly skimming these files at most. Please note that they will not run as-is, I would need to update the names of files in the metadatas.

Closes #1030


RUFAS/routines/field/crop/crop_configurations/alfalfa.py
```diff
@@ -26,6 +26,8 @@ class Alfalfa(CropData):
     minimum_temperature: float = 4.0
     optimal_temperature: float = 25.0
 
+    potential_heat_units: float = 400.0
```
matthew7838, line 29: What's the reason for separating this attribute? I would assume it can go with those involved with temperature.

krb78 (approved):
I think all of this looks good! It was quite long 🤪 but a bunch of it were all our metadata files, so it didn't feel as long as it looked. I APPROVE 😸 🌽 

The one thing I would say Ed, and perhaps this will just be in the documentation (when it's done), but definitions for a lot of these would be helpful to the average user. (as examples: what exactly a PHU is, labile carbon vs active carbon, tillage incorporation vs mixing fraction)

Isgotkowitz:
> The one thing I would say Ed, and perhaps this will just be in the documentation (when it's done), but definitions for a lot of these would be helpful to the average user. (as examples: what exactly a PHU is, labile carbon vs active carbon, tillage incorporation vs mixing fraction)

Thank you @krb78 :)

There is already quite a bit of documentation within the base files and the metadata properties file - but compiling and reviewing all that documentation still needs to happen. I think it will also be important when documentation is reviewed that what is there is comprehensible.

For example, PHU is documented as "Total heat units require for maturity". This is an accurate definition, but may not be super helpful as-is for the average user.

RUFAS/routines/field/crop/crop_management.py
```diff
@@ -188,11 +190,10 @@ def cut_crop(self, collected_fraction: float = 0) -> None:
         self.data.leaf_area_index = self.data.leaf_area_index * (1 - fraction_cut)
         self.data.accumulated_heat_units = self.data.accumulated_heat_units * (1 - fraction_cut)
 
-        self.data.wet_yield_collected = self.data.cut_biomass * collected_fraction
-        self.data.dry_matter_yield_collected = self.data.wet_yield_collected * (self.data.dry_matter_percentage / 100)
+        self.data.dry_matter_yield_collected = self.data.cut_biomass * collected_fraction
+        self.data.wet_yield_collected = self.data.dry_matter_yield_collected / (self.data.dry_matter_percentage / 100)
```
matthew7838, line 194: I'm interested in the reason of why this is switched.

RUFAS/routines/field/crop/crop_management.py
```diff
@@ -320,22 +322,65 @@ def _transfer_residue(self, soil_data: SoilData, killed: bool) -> None:
...
+        Distributes nutrients from plant residue into the soil profile.
+
+        Parameters
+        ----------
+        soil_data : SoilData
+            Object that tracks the attributes of the soil profile that contains this crop.
+        root_residue_mass : float
+            Dry matter mass of residue that is roots (kg / ha).
+
+        """
```
matthew7838: Is there any reference to this method?

RUFAS/routines/field/crop/growth_constraints.py
```diff
@@ -164,9 +164,8 @@ def _determine_temperature_stress(air_temp: float, min_temp: float,
         if min_temp < air_temp <= optimal_temp:
             stress = 1 - exp(numerator / (air_temp - min_temp) ** 2)
 
-        elif optimal_temp < air_temp <= double_diff:
-            stress = 1 - exp(numerator / (double_diff - min_temp) ** 2)
-
+        elif optimal_temp < air_temp < double_diff:
+            stress = 1 - exp(numerator / (double_diff - air_temp)**2)
```
matthew7838:
```suggestion
            stress = 1 - exp(numerator / (double_diff - air_temp) ** 2)
```

RUFAS/routines/field/soil/carbon_cycling/residue_partition.py
```diff
@@ -172,9 +174,19 @@ def _add_litter_to_pools(self) -> None:
...
         if self.data.plant_root_residue != 0.0 and self.data.crop_root_depth != 0.0:
             self._add_subsurface_residue(self.data.plant_root_residue, self.data.crop_root_depth)
+            self._set_soil_structural_litter_decomposition_rate()
             self.data.plant_root_residue = 0.0
             self.data.crop_root_depth = 0.0
 
+    def _set_soil_structural_litter_decomposition_rate(self) -> None:
+        """
+        Sets the soil structural litter decomposition rate using the same methodology as the
+        `_determine_plant_structural_to_slow_or_active_rate`.
```
matthew7838, line 184:
```suggestion
        `_determine_plant_structural_to_slow_or_active_rate`.

```

matthew7838 (approved):
LGTM, good modification to update the test coverage. As for changes to some default init values, I believe it's some modification based on SMEs' feedback? If so nothing seems odd to me. Others are just some small comments and some interests for the reason of changes.

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: SWAT tracks and records biomass on a dry matter basis, but I had originally assumed that it was tracked on a wet basis. So instead of taking the biomass tracked by the crop and multiplying it by the dry matter fraction to get the dry matter amount of biomass, the biomass needs to be divided by the dry matter fraction in order to get the wet biomass amount.

RUFAS/routines/field/crop/crop_configurations/alfalfa.py
Isgotkowitz: Mostly because the top two temperature attributes are grouped in different "sections" of `CropData` and I thought it might be nice to preserve that. No real reason though.

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: No - but I will add a notes section explaining why it is present.

Issue #1062, opened 2023-12-19
Add option to use recycled water for pen manure cleaning


KFosterReed:
**Is your feature request related to a problem? Please describe.**
The manure module ManureHandler class calculates the amount of water used to clean the pen and adds that to the manure mass. On many farms, most, if not all of this water is recycled from the manure management system so is not 'new' water that is added every day. This causes the total manure volume (especially for systems that use flushing for the pen cleaning) to be unrealistically large. 

**Describe the solution you'd like**
 I would like to add a '`water recycling fraction`' parameter to the manure handler configuration inputs so that we can only add the fraction of water that is not recycled to the manure management system and have a more realistic volume of manure. This '`water recycling fraction`' would be used in the ManureHandler class function `calc_cleaning_water_volume_in_main_barn()`. The equation would be updated to: 
`cleaning_water_volume = num_animals * self.config.cleaning_water_use_rate*(1-water_recycling_fraction)`

**Describe alternatives you've considered**
An alternative would be to update the values for the clearning water use rates directly. This would be easier now but would not allow us to distinguish between those farms that recycle water and those that don't in the future. 

**Additional context**

ew3361zh:
@KFosterReed how is the `water recycling fraction` ascertained? 
Is it a constant? 
Or is it being populated based on some characteristics of the farm (e.g. certain manure handling/processing types being present on a farm would mean they would need to implement the fraction)?
Or some other way?

KFosterReed:
> @KFosterReed how is the `water recycling fraction` ascertained? Is it a constant? Or is it being populated based on some characteristics of the farm (e.g. certain manure handling/processing types being present on a farm would mean they would need to implement the fraction)? Or some other way?

I don’t think there is a standard recycling fraction so my preference would be to just add it as a user input in the relevant manure handler options. For example, CBPB (compost bedded pack barns) and Open lots wouldn’t have water recycling but all other manure handler types could potentially have that as an option. The default value would be  0% for all methods. An alternative option would be to just make it an input at the manure management scenario input level. 

@krb78 do you have any thoughts as to where it makes more sense to have this input?

krb78:
The most common recycling that occurs is 2 places: 1) the water in the plate cooler is used to provide drinking water or milking parlor wash water (but this would be in the EEE layer, not the manure module) and 2) if it is a flush dairy, the water used to flush is often recycled many times (some audits have shown water is used 4 times on a dairy farm with a flush system). 

[...]

PR #1065, opened 2023-12-19
Manure Model Debugging

KFosterReed:
I have been debugging the Manure module to address the following problems:
- Manure Total Ammoniacal N and ammonia emission conversion errors
- Total N mass balances along the manure management chain
- The amount of water added by the manure handler does not reflect the option for waste water recycling (see [Issue 1062](https://github.com/RuminantFarmSystems/MASM/issues/1062) )
- Manure density constant updates
- Set sludge removal efficiencies to 0 so that all nutrients remain in liquid until we have a functioning sludge management method


RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
```diff
@@ -125,10 +127,11 @@ def daily_update(self,
                 max(
                     0.0,
                     pen.manure.manure_total_ammoniacal_nitrogen
-                    - housing_ammonia_emission,
+                    - (housing_ammonia_emission*GeneralConstants.AMMONIA_TO_AMMONIA_NITROGEN),
```
elle-andreen, line 130: Hi Kristan, I was browsing the various manure module issues and PRs and noticed something in the commits here. To the best of my knowledge, the ammonia emissions calculations (for housing and outdoor storage at least) in calculator.py are estimating kg of ammonia **nitrogen** loss, not ammonia loss. Those calculations are based on [this publication](https://core.ac.uk/download/pdf/29274074.pdf). This would probably be best confirmed with Varma. If we are in fact calculating NH3-N here, we don't need this ammonia to ammonia N conversion factor, and we should probably update the docstring to make it more clear than we're calculating NH3-N and not NH3!

KFosterReed: Hi Elle! Thank you for looking into this! I did ask Varma to clarify this point in our meeting and he indicated the value was in units of ammonia, not ammonia-N but since that was a verbal confirmation on the spot, it is very possible he was mistaken. Since you are reviewing the documentation, could you please compare the coefficients in this paper and the ones in our documentation and code? THANK YOU!!

elle-andreen: Absolutely! I can notate things here for now to either be changed through this PR or in other smaller PRs. 😊

ew3361zh: @KFosterReed and @elle-andreen I wondered if this part of the code (and it's counterpart in slurry_storage_outdoor.py which uses the same constant) should be added to the PR I'm working on for Ammoniacal N? I saw this comment thread and it sounded like things were on hold so right now it's not a part of the PR but it wouldn't be difficult to add it in if this is the correct modeling.

KFosterReed: Hi Niko! Thanks for following up on this! Elle did confirm that the original equation is in the units of ammonia-N (not the full molecule of ammonia) so we can delete this constant.

PR #1079, opened 2023-12-26
Test coverage for Time.py

### LINKED ISSUE #898: Bring test coverage of `Time` up to 100%
The `Time` class is undertested.

## Expected Behavior

Every method of the `Time` class should be tested.

## Current Behavior

Only `record_time` has a dedicated test written.

## Possible Solution

Write tests for `Time` class.

matthew7838:
### Summary
This PR aims to bring a complete test suite to the `time.py` module.

### Changes

- Test coverage of  `time.py` is now fully covered.
- In `time.py`, `end_year()` has changed so that it will return true right on the day of the end year instead of returning true on the next `advance()` call.
- Some type hints were added for clarity.


RUFAS/time.py
```diff
@@ -69,8 +69,7 @@ def end_year(self):
             bool: True if it is the end of a year, False otherwise
         """
 
-        # if the day is > the length of the current year, then the year is over
-        return self.day > len(self.years[self.year - 1])
+        return self.day >= len(self.years[self.year - 1])
```
JoeWaddell: Will this not shorten the sim by a day, or had we been adding a day to the simulation already?

I'm looking at `simulation_engine._annual_simulation`, which will only run the `_daily_simulation` method if it's not `end_year`.

I think the proposed change here overlaps with `is_last_day_of_simulation`, correct? If so, it will just require rewriting the end_year checks in some of the unit tests. I might be missing something subtle here though, so let me know if that's the case!

JoeWaddell (commented):
Great job on updating config.py, and expanding the test coverage! Just one question about the difference in end_year and is_last_day_of_simulation.

RUFAS/time.py
matthew7838: Now I look into it again I think it makes sense, I'll revert the changes.

JoeWaddell (approved):
Nice work, LGTM

RUFAS/time.py
```diff
@@ -17,11 +17,11 @@ def __init__(self, config: Config):
 
         calendar_year: int = config.start_year
         # number of years
-        years: List[int] = config.years
+        years: List[List[int]] = config.years
```
allisterakun:
```suggestion
       
```

RUFAS/time.py
```diff
@@ -17,11 +17,11 @@ def __init__(self, config: Config):
...
 
         self.start_year: int = calendar_year
         self.calendar_year: int = calendar_year
-        self.years: List[int] = years
+        self.years: List[List[int]] = years
```
allisterakun:
```suggestion
        self.years: List[List[int]] = config.years
```

allisterakun (approved):
Looks good, good job, Matthew! Just a minor comment.

PR #1080, opened 2023-12-28
tested all functions in heiferI.py

0luchi:
Added test cases in test_animal_life_cycle.py for all the functions in heiferI.py
- Used mock.MagicMock() arguments
- Test cases are in test_heiferI() and test_update_heiferI()
- Corrected an error in line 58 of heiferI.py


ew3361zh:
I haven't done a full review yet but wanted to ask a couple questions first:

1. Is this addressing a particular issue to increase coverage for heiferI.py? If so, could you link the issue to this PR to ensure we know the full extent of the issue being addressed?
2. This looks like it increases test coverage for heiferI.py greatly from 27% (currently on main) to 86% (on this branch). Did you look at covering the other lines (35, 43, 45, 47-49, 84-85) to get to 100%?

0luchi:
@ew3361zh 
1. I don't think it is addressing an issue. I was just working based on this spreadsheet https://docs.google.com/spreadsheets/d/1wjdPo_20AtT84-809bf_-FXA8h6e00zMk8prx7_d3vc/edit?pli=1#gid=0
2. I will work on covering those lines. This is my first PR so I wanted to make sure I was doing the testing correctly first.

0luchi:
@ew3361zh 
I just improved coverage to 100%

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,114 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+    """Unit tests for all the functions in class HeiferI in file life_cycle.py."""
+
+    heiferI = mocker.MagicMock(autospec=HeiferI)
+    mocker.patch('RUFAS.routines.animal.life_cycle.heiferI.HeiferI.__init__', return_value=heiferI)
+    heiferI.body_weight = 1000.0
+    heiferI.mature_body_weight = 1200.0
+    heiferI.daily_growth = 1.0
+    heiferI.days_born = 300
+    temp = 30
+    animal_grouping_scenario = AnimalType.HEIFER_I
```
JoeWaddell: This needs to be the `AnimalGroupingScenario` class, not the `AnimalType` itself! Likewise this change will require the rolling back of the change in `heifer1.py`

RUFAS/routines/animal/life_cycle/heiferI.py
```diff
@@ -55,7 +55,7 @@ def set_nutrient_rqmts(self, temp, animal_grouping_scenario, nutrient_conc: dict
         animal_requirements = req.calc_rqmts(body_weight=self.body_weight,
                                              mature_body_weight=self.mature_body_weight,
                                              day_of_pregnancy=None,
-                                             animal_type=animal_grouping_scenario.get_animal_type(self),
+                                             animal_type=animal_grouping_scenario,
```
JoeWaddell: This should be rolled back, since the methods in `animal_requirements.py` need the AnimalGroupingScenario type, not AnimalType directly. It'ssomething we can consider refactoring, if you see a problem its causing here.

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,114 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+        "breeding_start_day_h": 400,
+        "heifer_repro_programs": {
+            "heifer_TAI_protocol": heifer_TAI_protocol,
+            "heifer_synchED_protocol": heifer_synchedED_protocol
+        },
+        "nutrient_standard": nutrient_standard
+    }
+    mocker.patch('RUFAS.routines.animal.life_cycle.life_cycle.AnimalBase.config', animal_base_config)
+
+    # testing set_nutrient_rqmts()
```
JoeWaddell: Each of these separate methods should be separated into different tests. It will increase the length of this file a little, but will be easier to maintain tests as methods are added/removed, since it'll ensure that individual tests aren't reliant on things that happened in other tests.

JoeWaddell (changes requested):
This is looking good! I've left some comments, mostly about breaking it up into separate unit tests, for maintainability. Should be straightforward, though it will lead to a longer file.

Also a couple questions about a change to the logic in heifer1.py, which I think will need to be reverted (and may require rewriting a couple sections of the tests).

RUFAS/routines/animal/life_cycle/heiferI.py
0luchi: Sorry about that. I was just a little confused with what the type of animal_grouping_scenario is because the calc_rqmts() function had the type annotation of AnimalType, but this clarifies things now thank you.

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,141 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
     assert life_cycle_manager.avg_heifer_culling_age == approx(0.0)
     assert life_cycle_manager.avg_cow_culling_age == approx(0.0)
     assert life_cycle_manager.avg_mature_body_weight == approx(0.0)
+
+
+def test_set_nutrient_rqmts(mocker: MockerFixture) -> None:
+    """Unit tests for all the functions in class HeiferI in file life_cycle.py."""
```
JoeWaddell: For each unit test here you'll want a docstring summary that matches the format of the others: in this case 

`Unit test for method set_nutrient_rqmts in file heiferI.py.`

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,141 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+        "heifer_repro_programs": {
+            "heifer_TAI_protocol": heifer_TAI_protocol,
+            "heifer_synchED_protocol": heifer_synchedED_protocol
+        },
+        "nutrient_standard": nutrient_standard
+    }
+    mocker.patch('RUFAS.routines.animal.life_cycle.life_cycle.AnimalBase.config', animal_base_config)
+
+    HeiferI.set_nutrient_rqmts(heiferI, temp, animal_grouping_scenario, nutrient_conc)
+    animal_req.assert_called()
```
JoeWaddell: Here I think you could add an `assert_called_with` to check that it used the expected values in the call.

Additionally, I think you could add another assertion here, namely that the mocked heiferI object correctly applies the requirement values from the animal_requirements object.

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,141 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+    heiferI = mocker.MagicMock(autospec=HeiferI)
+    mocker.patch('RUFAS.routines.animal.life_cycle.heiferI.HeiferI.__init__', return_value=heiferI)
+    heiferI.body_weight = 1000.0
+    heiferI.mature_body_weight = 1200.0
+    heiferI.daily_growth = 1.0
+    heiferI.days_born = 300
+    feed = mocker.MagicMock(autospec=Feed)
+    mock_numI = (mocker.MagicMock(autospec=int), mocker.MagicMock(autospec=int))
+    heiferI.calc_base_manure.return_value = mock_numI
+    HeiferI.calc_manure_excretion(heiferI, feed, "methane_model")
```
JoeWaddell, line 1555: I think a check that `manure_calculations` was also called would be good here.

JoeWaddell (changes requested):
Looking good! Just had a couple comments on docstring formatting and a couple extra assertions needed.

tests/animal_module_tests/test_animal_life_cycle.py
0luchi: I tried doing this, but I've been having trouble because manure_calculations does not belong to a class. I'm not sure of how to mock it

JoeWaddell: The confusing thing is that since Heifer I is mocked, we have to mock manure_calculations where it's imported, instead of from where it 'should' exist.
So it can be mocked as follows:
`thingy = mocker.patch('RUFAS.routines.animal.life_cycle.heiferI.manure_calculations', return_value=(None, None))`
That mocked method can then be checked via things like: `thingy.assert_called_once()`, and even better, `assert_called_with`, etc.
Of course we'd want better names than 'thingy'.

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,171 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+                                  body_condition_score_5=3,
+                                  previous_temperature=temp,
+                                  average_daily_gain_heifer=heiferI.daily_growth,
+                                  NDF_conc=0.3,
+                                  TDN_conc=0.7,
+                                  net_energy_diet_concentration=(15.625 * 0.64)/10.0,
+                                  days_born=heiferI.days_born)
+    nutrient_conc = {'dm': 7.0, 'NDF': 4.5, 'TDN': 7.8}
+    HeiferI.set_nutrient_rqmts(heiferI, temp, animal_grouping_scenario, nutrient_conc, 0.0, 0.0)
+    animal_req.assert_called()
```
Isgotkowitz:
```suggestion
```

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,171 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+    animal_req.assert_called_with(body_weight=heiferI.body_weight,
+                                  mature_body_weight=heiferI.mature_body_weight,
+                                  day_of_pregnancy=None,
+                                  animal_type=animal_grouping_scenario.get_animal_type(heiferI),
+                                  body_condition_score_5=3,
+                                  previous_temperature=temp,
+                                  average_daily_gain_heifer=heiferI.daily_growth,
+                                  NDF_conc=0.3,
+                                  TDN_conc=0.7,
+                                  net_energy_diet_concentration=(15.625 * 0.64)/10.0,
```
Isgotkowitz: I think it would be good to define the value `(15.625 * 0.64)/10.0` (it's just 1.0) in a variable and label it. Otherwise it is very confusing where these numbers are coming from. The way that `net_energy_diet_concentration` is calculated is also very weird - there are multiple paths of execution but if the defaults for `metabolizable_energy` and `previous_DMI` are used, then both paths lead to the same result. I think it would be good to pass in non-default and non-zero values for `metabolizable_energy` and `previous_DMI`, and check that the `net_energy_diet_concentration` is calculated to be what you expect it to be.

tests/animal_module_tests/test_animal_life_cycle.py
Isgotkowitz: Also, `self.p_excrt` and `self.manure_excretion` should be checked to make sure they were set correctly.

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,171 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+    p_urine = 0.000002 * heiferI.body_weight * 1000
+
+    heiferI.p_growth = (
+        (0.0012 + 0.004635 * (heiferI.mature_body_weight**0.22) * (heiferI.body_weight ** (-0.22)))
+        * heiferI.daily_growth
+        / 0.96
+        * 1000
+    )
+    p_absorb = p_urine + heiferI.p_maint_feces + heiferI.p_growth
+    p_req = p_absorb / 0.664
```
Isgotkowitz: This is entirely duplicating the logic of `phosphorus_rqmts`, which means that if there are any bugs in the method they will be propagated into the test. The values that you expect to be calculated for `p_maint_feces`, `p_growth`, and `p_req` should be hardcoded in the test (in an ideal world - in practice it can be difficult and sometimes impractical to do this for large unwieldy math equations).

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,171 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+        "heifer_repro_programs": {
+            "heifer_TAI_protocol": heifer_TAI_protocol,
+            "heifer_synchED_protocol": heifer_synchedED_protocol
+        },
+        "nutrient_standard": nutrient_standard
+    }
+    mocker.patch('RUFAS.routines.animal.life_cycle.life_cycle.AnimalBase.config', animal_base_config)
+    weight = HeiferI.get_non_preg_bw_change(heiferI)
+    calc_weight = 0.55 * 0.96 * heiferI.mature_body_weight - 0.96 * heiferI.body_weight  # divisor == 1
+    assert weight == calc_weight
```
Isgotkowitz:
```suggestion
    mocker.patch('RUFAS.routines.animal.life_cycle.life_cycle.AnimalBase.config', animal_base_config)
    
    actual_weight = HeiferI.get_non_preg_bw_change(heiferI)
    
    expected_weight = 0.55 * 0.96 * heiferI.mature_body_weight - 0.96 * heiferI.body_weight  # divisor == 1
    assert actual_weight == expected_weight
```
Grouping by setup, act, and assert helps to keep things more readable. Also, using the terms actual and expected for what the method actually does vs what you expect it to do also makes the test a lot more readable.

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,171 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+        * 1000
+    )
+    p_absorb = p_urine + heiferI.p_maint_feces + heiferI.p_growth
+    p_req = p_absorb / 0.664
+    assert heiferI.p_maint_feces == 8
+    assert heiferI.p_req == p_req
+
+
+def test_heiferI_get_non_preg_bw_change(mocker: MockerFixture) -> None:
+    """Unit test for the function get_non_preg_bw_change in file routines/animal/life_cycle/heiferI.py."""
```
Isgotkowitz: This test is only testing a single case: when `days_born` is exactly equal to `AnimalBase.config["target_heifer_preg_day"]`. It should be checking the other two cases, when days born is > and < the target heifer preg day. All that said, I think you should hold off on expanding this test until talking w/ @JoeWaddell about this method, because he is going to evaluate this method's logic to see if it needs to be changed.

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,171 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+    nutrient_standard = 'NRC'
+    animal_base_config = {
+        "heifer_repro_method": heifer_repro_method,
+        "target_heifer_preg_day": 399,
+        "breeding_start_day_h": 400,
+        "nutrient_standard": nutrient_standard
+    }
+    mocker.patch('RUFAS.routines.animal.life_cycle.life_cycle.AnimalBase.config', animal_base_config)
+    heiferI.update_body_weight_history = mocker.MagicMock()
+    heiferI.get_non_preg_bw_change = mocker.MagicMock()
```
Isgotkowitz, line 1640: It would be good to set the return value for `get_non_preg_bw_change()` and check that daily growth is being correctly set.

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,171 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+    heiferI.get_non_preg_bw_change.assert_called()
+    assert updated_val
+    assert heiferI.days_born == 399
+
+    heiferI.days_born = 398
+    updated_val = HeiferI.update(heiferI, sim_day)
+    assert not updated_val
+    assert heiferI.days_born == 399
+    heiferI.update_body_weight_history.assert_called()
+    heiferI.get_non_preg_bw_change.assert_called()
```
Isgotkowitz: The body weight change should be checked in this test as well.

Isgotkowitz (approved):
Overall this PR looks good. Several comments about improving tests, not all suggestions for improvements may be feasible though. After looking through those (and getting a second approval) I think this PR should be good to go.

Keep up the good work Oluchi!

JoeWaddell (approved):
Nice job updating these after all the comments! LGTM

PR #1081, opened 2024-01-02
Add Kill only harvest operation to available harvest operations.

### LINKED ISSUE #969: Add kill only harvest operation as a management practice available to users
## Current State
Right now, there are only two harvest operations available for users: harvest-and-kill and harvest-no-kill (`"default"` and `"no_kill"`) in the Crop Input. There is already code for a kill only harvest operation, but the inputs and initialization routines need to be updated to make this functionality available to users.

## Desired State
Allow users to specify a kill only harvest operation.

Some additional things to note:
- This seems like a prime opportunity to make the names of harvest operations better. Possibly "default" => "harvest_and_kill", "no_kill" => "harvest_no_kill"?
- The Data Collection App MUST GET UPDATED with the new available harvest operation.

Isgotkowitz:
This PR adds a crop management option so that crops may be killed _without_ also being harvested. Since this involved changing the available harvest management operations, I thought it would be a good opportunity to make some improvements related to how harvest operations are implemented. This includes:
- Changing the names. The elements of the enum and the accepted input options are 'harvest_kill', 'harvest_only', and 'kill_only', instead of 'default' and 'harvest_no_kill'.
- Mapping the user-input harvest operations (which are given as strings) to the elements of the `HarvestOperation` enum as early as possible (in this case, in the `CropSchedule` init method).

All tests have been updated appropriately. 

Closes #969


RUFAS/routines/field/crop/crop_management.py
```diff
@@ -52,14 +52,18 @@ def manage_harvest(self, harvest_op: HarvestOperation, field_name: str, field_si
         is_killed = False
```
allisterakun: This may be out of the scope of this PR, but I am wondering if having a local `is_killed` variable is necessary.
`self.data.is_alive` can be used in place on line 68 `self._transfer_residue(soil_data, !self.data.is_alive)`
In `self.kill()` function, it sets `self.data.is_alive` to false if the crop is killed. It would be unnecessary to explicitly set the local `is_killed` variable to True when killing the crop.

allisterakun (approved):
The changes all look good! Just a minor comment. Good job Ed!

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: @allisterakun good suggestion, I implemented this and cleaned up the mocking as a side effect.

krb78 (approved):
I think this looks great Ed! It's much clearer regarding what is happening at the end of a growing season.

PR #1082, opened 2024-01-03
Implement FeedManager and recieve_crop

PooyaHekmati:
This PR defines the structure of the `FeedManager` and implements its `recieve_crop` method.
[design doc](https://docs.google.com/document/d/1C86lch2dkYBERR-qLDmE7L7n1gYQjETrIKUyrQpWrPs/edit?usp=sharing)
It also enhances some existing docstrings.
Moves `sample_crop_data` to its own file to make it reusable and remove duplication.


RUFAS/routines/feed_storage/feed_manager.py
```diff
@@ -0,0 +1,137 @@
...
+            The type of crop to distribute.
+
+        Returns
+        -------
+        float
+            The actual amount of feed distributed.
+        """
+        pass
+
+    def query_available_feeds(
```
JoeWaddell, line 117: Great name for this method: will be clear in the ration formulation process.

RUFAS/routines/feed_storage/feed_manager.py
```diff
@@ -0,0 +1,137 @@
...
+            storage.process_degradations()
+
+    def give_feed(self, amount: float, crop_type: CropType) -> float:
+        """
+        Distributes feed to the Animal module based on the FIFO principle.
+
+        Parameters
+        ----------
+        amount : float
+            The amount of feed to distribute.
```
JoeWaddell, line 106:
```suggestion
            The amount of feed to distribute (kg).
```

RUFAS/routines/feed_storage/feed_manager.py
```diff
@@ -0,0 +1,137 @@
...
+
+        Parameters
+        ----------
+        crop_type : CropType, optional
+            The type of crop to query (default is None, which queries all types).
+
+        Returns
+        -------
+        Union[Dict[CropType, float], float]
+            The amount of available feed, either as a total or for a specific crop type.
```
JoeWaddell, line 131:
```suggestion
            The amount of available feed (kg), either as a total or for a specific crop type.
```

RUFAS/routines/feed_storage/feed_manager.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        pass
+
+    def purchase_feed(self) -> None:
+        """The purchase feed logic is currently in the Animal Module. We will move it to here."""
```
JoeWaddell, line 136: Excellent: looking forward to helping with that move!

RUFAS/routines/feed_storage/feed_manager.py
```diff
@@ -0,0 +1,137 @@
...
+        ----------
+        amount : float
+            The amount of feed to distribute.
+        crop_type : CropType
+            The type of crop to distribute.
+
+        Returns
+        -------
+        float
+            The actual amount of feed distributed.
```
JoeWaddell, line 113:
```suggestion
            The actual amount of feed distributed (kg).
```

RUFAS/routines/feed_storage/enums.py
```diff
@@ -33,35 +33,35 @@ class CropType(Enum):
     Attributes
     ----------
     WHEAT : str
-        Type of Small Grain.
+        Represents a type of type of Small Grain.
```
Isgotkowitz:
```suggestion
        Represents a type of Small Grain.
```
Same comment for other attributes in this docstring.

RUFAS/routines/feed_storage/feed_manager.py
```diff
@@ -0,0 +1,137 @@
+from typing import List, Dict, Union
+from enum import Enum
+from .harvested_crop import HarvestedCrop
+from .storage import Storage
+from .enums import CropType, CropCategory
+
+from .baleage import Baleage
+from .grain import Grain, Dry, HighMoisture
+from .hay import Hay, ProtectedIndoors, ProtectedTarped, ProtectedWrapped, Unprotected
+from .sileage import Sileage, Bag, Bunker, Pile
```
Isgotkowitz, line 10: Is this the preferred way of writing import statements, i.e. `.baleage` instead of `RUFAS.routines.feed_storage.baleage`?

RUFAS/routines/feed_storage/feed_manager.py
```diff
@@ -0,0 +1,137 @@
...
+                f"Crop of category '{harvested_crop.category}' is not compatible with storage type '{storage_type}'. "
+                f"Compatible storage types are: {', '.join([cls.__name__ for cls in compatible_storage_classes])}"
+            )
+
+        if storage_type not in self.active_storages:
+            self.active_storages[storage_type] = storage_type.value()
+
+        self.active_storages[storage_type].receive_crop(harvested_crop)
+
+    def process_degradations(self):
```
Isgotkowitz:
```suggestion
    def process_degradations(self) -> None:
```

RUFAS/routines/feed_storage/feed_manager.py
```diff
@@ -0,0 +1,137 @@
...
+    and querying available feeds.
+
+    Attributes
+    ----------
+    active_storages : List[Storage]
+        List of active storage units in the simulation.
+    """
+
+    def __init__(self):
+        self.active_storages: Dict[StorageType, Storage] = {}
```
Isgotkowitz: There is a mismatch between type annotations here. I understand why it would be `List[Storage]`, but I do not understand why it would be a `Dict[StorageType, Storage]`.

Isgotkowitz (changes requested):
Requesting changes on the way that `sample_crop_data` is implemented and shared between modules. I am all for increasing reusability and decreasing duplication, but I don't think it is worth the tradeoff of adding a bunch of `noqa`'s in this case. Everything else is minor comments.

JoeWaddell (approved):
Only minor comments here (and my suggestion to add units to docstrings might be premature) - this is a great start. Exciting to see this come together. @Isgotkowitz brought up great questions.

RUFAS/routines/feed_storage/feed_manager.py
PooyaHekmati: Yes, always (is not limited to Python) relative path takes priority over the absolute path. The reason is that with relative paths, the module or package is bundled, if we decide to rename `routines`, as an example, relative path is agnostic about it.

RUFAS/routines/feed_storage/feed_manager.py
PooyaHekmati: `Dict[StorageType, Storage]` is correct, it maps `StorageType(Enum)` to instances of `Storage`.

RUFAS/routines/feed_storage/feed_manager.py
PooyaHekmati: I understand why you would want to have the unit here; however, this class is agnostic about the units. We cannot guarantee that the amount received from the caller (S&C) is indeed in kg. I am concerned that being explicit about the unit here, makes it harder to debug, should something go wrong.

PooyaHekmati:
> Requesting changes on the way that `sample_crop_data` is implemented and shared between modules. I am all for increasing reusability and decreasing duplication, but I don't think it is worth the tradeoff of adding a bunch of `noqa`'s in this case. Everything else is minor comments.

I disagree, duplication is known to be the root of many, if not most, problems in software. But I removed the noqas.

RUFAS/routines/feed_storage/feed_manager.py
Isgotkowitz: > this class is agnostic about the units. We cannot guarantee that the amount received from the caller (S&C) is indeed in kg. I am concerned that being explicit about the unit here, makes it harder to debug, should something go wrong.

Can you please expand on how this class can be agnostic to units? I am not understanding how labeling units in different methods would make things harder to debug.

tests/test_feed_storage_module/test_feed_manager.py
```diff
@@ -0,0 +1,70 @@
...
+    HarvestedCrop
+        An instance of the HarvestedCrop class.
+    """
+    category = CropCategory.CORN
+    crop_type = CropType.WHOLE_PLANT
+    return HarvestedCrop(category=category, type=crop_type, **sample_crop_data)
+
+
+@pytest.fixture
+def feed_manager():
```
Isgotkowitz:
```suggestion
def feed_manager() -> FeedManager:
```

tests/test_feed_storage_module/test_feed_manager.py
```diff
@@ -0,0 +1,70 @@
...
+    return FeedManager()
+
+
+def test_receive_crop_success(feed_manager: FeedManager, harvested_crop: HarvestedCrop):
```
Isgotkowitz:
```suggestion
def test_receive_crop_success(feed_manager: FeedManager, harvested_crop: HarvestedCrop) -> None:
```

tests/test_feed_storage_module/test_feed_manager.py
```diff
@@ -0,0 +1,70 @@
...
+            harvested_crop=harvested_crop,
+            storage_type=StorageType.DRY,
+        )
+    except ValueError:
+        pytest.fail("Unexpected ValueError raised")
+
+
+def test_receive_crop_multiple(
+    feed_manager: FeedManager, harvested_crop: HarvestedCrop
+):
```
Isgotkowitz:
```suggestion
) -> None:
```

tests/test_feed_storage_module/test_feed_manager.py
```diff
@@ -0,0 +1,70 @@
...
+        )
+        assert StorageType.DRY in feed_manager.active_storages.keys()
+        assert StorageType.BUNKER in feed_manager.active_storages.keys()
+        assert len(feed_manager.active_storages[StorageType.BUNKER].stored) == 1
+        len(feed_manager.active_storages[StorageType.DRY].stored) == 2
+    except ValueError:
+        pytest.fail("Unexpected ValueError raised")
+
+
+def test_receive_crop_error(feed_manager: FeedManager, harvested_crop: HarvestedCrop):
```
Isgotkowitz:
```suggestion
def test_receive_crop_error(feed_manager: FeedManager, harvested_crop: HarvestedCrop) -> None:
```

tests/test_feed_storage_module/test_harvested_crop.py
```diff
@@ -86,13 +42,13 @@ def test_valid_category_type_combinations(
     ],
 )
 def test_invalid_category_type_combinations(
-    category: CropCategory, crop_type: CropType, sample_crop_data: Dict[str, float]
+    category: CropCategory, crop_type: CropType
 ):
```
Isgotkowitz:
```suggestion
) -> None:
```

tests/test_feed_storage_module/test_harvested_crop.py
```diff
@@ -86,13 +42,13 @@ def test_valid_category_type_combinations(
...
     with pytest.raises(ValueError):
         HarvestedCrop(category=category, type=crop_type, **sample_crop_data)
 
 
-def test_attributes(sample_crop_data: Dict[str, float]):
+def test_attributes():
```
Isgotkowitz:
```suggestion
def test_attributes() -> None:
```

tests/test_feed_storage_module/test_storage.py
```diff
@@ -78,36 +36,35 @@ def harvested_crop(sample_crop_data: Dict[str, float]) -> HarvestedCrop:
...
     with pytest.raises(Exception) as excinfo:
-        storage.receive_crop(harvested_crop, harvested_crop.harvest_time)
+        storage.receive_crop(harvested_crop)
     assert "exceeds the storage capacity" in str(excinfo.value)
 
 
-def test_receive_unacceptable_crop(
-    storage: Storage, sample_crop_data: Dict[str, float]
-):
+def test_receive_unacceptable_crop(storage: Storage):
```
Isgotkowitz:
```suggestion
def test_receive_unacceptable_crop(storage: Storage) -> None:
```

tests/test_feed_storage_module/test_harvested_crop.py
```diff
@@ -67,9 +25,7 @@ def sample_crop_data() -> Dict[str, float]:
         (CropCategory.GRASS, CropType.MEADOW_FESCUE),
     ],
 )
-def test_valid_category_type_combinations(
-    category: CropCategory, crop_type: CropType, sample_crop_data: Dict[str, float]
-):
+def test_valid_category_type_combinations(category: CropCategory, crop_type: CropType):
```
Isgotkowitz:
```suggestion
def test_valid_category_type_combinations(category: CropCategory, crop_type: CropType) -> None:
```

RUFAS/routines/feed_storage/feed_manager.py
Isgotkowitz: So the Feed Manager expects to only manage 0 or 1 instances of each type of Storage? I.e. it will only simulate one bunker, one hay storer, etc.

Isgotkowitz (approved):
Thank you for addressing requested changes. I have a follow up question about the typing of `active_storages` (in the comment/response above) and I also would like to get a better understanding of how units will work in this class, but otherwise looks good.

RUFAS/routines/feed_storage/feed_manager.py
KevinP-B: That should be correct, Ed. in situ, that isn't what we would see, but in silico, there is no reason we've yet thought of that we need multiple instances when a single instance storing feed of varying ages will suffice. Is that correct, Pooyah?

KevinP-B (approved):
Implementation of FeedManager and receive_crop look good at a lay level

RUFAS/routines/feed_storage/feed_manager.py
PooyaHekmati: Yes, that is correct. Representation of something is different from that thing itself.

RUFAS/routines/feed_storage/feed_manager.py
PooyaHekmati: This class is simply a (data) broker. It receives data from upstream (S&C) and passes it downstream (Storage classes and Animal module). There is no way that this class can verify the units, it just gets scalar values, so it cannot be used as a source of truth. Therefore, using it as a source of truth for units can lead to unexpected outcomes. For example, let's say an upstream process, purchase feed logic as an example, has passed weight in lb. Unable to verify that, Feed Manager has stored what it has received. An SME realizes that the outputs are not correct and tries to figure out why. They suspect that units might be wrong, and look at the docstring here and see kg when it is lb. This misrepresentation makes them rule out the possibility of units not being correct. But if we don't lie to them here about something that we cannot be confident about, the next logical step is to check the upstream caller, where they will look at it and find the problem.

JoeWaddell: Great explanation, you just merged this PR, but wanted to chime in that I think the reasoning is solid, and something I think we'll need to implement elsewhere in the model.

PR #1083, opened 2024-01-03
Remove `simulate_animals` in config

### LINKED ISSUE #1068: Remove simulate_animals from metadata
In PR 878 we removed Animal DB and updated how herd initialization works. After that change, we no longer use `simulate_animals` in the config file; however, it is still present in the input files and in the metadata. Fix this issue by removing the following section from the metadata and corresponding data from input files.
```
"simulate_animals": {
        "type": "bool",
        "description": "Whether or not to simulate animals during the simulation",
        "default": true
      }
```

allisterakun:
Closes #1068 
Removing all `simulate_animals` variables in metadata and all config input JSON files

### DIFF
--- a/input/data/config/ARL_config.json
+++ b/input/data/config/ARL_config.json
@@ -3,7 +3,6 @@
     "end_date": "2019:365",
     "random_seed": 0,
     "set_seed": false,
-    "simulate_animals": false,
     "nutrient_standard": "NASEM",
     "FIPS_county_code": 55021
 }
--- a/input/data/config/barnyard_config.json
+++ b/input/data/config/barnyard_config.json
@@ -3,7 +3,6 @@
     "end_date": "2016:243",
     "random_seed": 0,
     "set_seed": false,
-    "simulate_animals": false,
     "nutrient_standard": "NASEM",
     "FIPS_county_code": 55073
 }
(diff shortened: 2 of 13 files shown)


ew3361zh (approved):
I think it looks good and seems to fully address the issue so I'm approving however I have a couple questions:

1. There's still a reference to this field in config.py (line 49) which I think should be removed unless it's serving another purpose.
2. That brings up my second point which is I'm not clear if there's an option not to simulate animals anymore (not sure we need one but just curious). It seems like the way things are set up that this field is set in line 49 of config.py and will now always be true. Do we need an option not to simulate animals? If so, how should it be done? Or is it alreayd done somewhere else?

Isgotkowitz:
I think #1068 is not right. In the description, it states "In PR 878 we removed Animal DB and updated how herd initialization works. After that change, we no longer use simulate_animals in the config file". This is not exactly true; `simulate_animals` is still used in the `AnimalManager` to determine whether or not to simulate animals.

Furthermore, I think having `"simulate_animals"` as an option is a really important feature. Simulating animals makes the simulations run for a significantly longer period of time (default scenario is 10 sec w/ simulate animals set to false, 1 min 12 sec with it set to true), so it is extremely convenient to be able to turn off animal simulation when it is not needed.

CC @PooyaHekmati

allisterakun:
After some digging following @Isgotkowitz's comment, I found that 'simulate_animals` still serves a purpose in the codebase. I am closing the PR without any changes.

PR #1087, opened 2024-01-05
Expand runtime IM add_variable_to_pool capabilities

### LINKED ISSUE #1060: Expand capabilities for adding data to pool
Currently the methods for adding data to pool are only capable of adding data that can be validated against "full" properties (`"animal_properties"`, `"field_properties"`, etc.).

A method that allows data to be added to pool that can replace data which is located inside of a more complex data structure in the pool.

**Additional context**
There are already two methods for adding data to the Input Manager at run time, `add_dict_variable_to_pool()` and `add_tabular_variable_to_pool()` (in Input Manager). They rely on the hidden methods `_add_variable_to_pool()` and `_metadata_properties_exists()`. It may be possible to expand/refactor these methods to address this feature.

allisterakun:
This PR addresses issue #1060 :
- [x] Adds the option `modifiability to all metadata properties:
    - `modifiability` indicates if a variable is required at initialization and if it can be modified during runtime.
    - The `modifiability` property is a string with three possible options (basically an enum):
        - `"required_and_locked"`: The variable must be included in the input file at initialization and cannot be modified during runtime.
        - `"required_and_unlocked"`: The variable must be included in the input file at initialization and can be modified during runtime.
        - `"not_required_and_unlocked"`: It is optional to include the variable in the input file at initialization and the variable can be modified during runtime. Note: if a variable with `modifiability` of `"not_required_and_unlocked"` is also included in the input, the variable will still be validated against the metadata (and fixed if needed and possible).
- [x] Creates an enum class `Modifiability` in `RUFAS/input_manager.py` to represent the three possible states of modifiability.
- [x] Creates the following methods in `InputManager`:
    - [x] _get_variable_modifiability()
        ```
        def _get_variable_modifiability(self, variable_name: str, variable_properties: Dict[str, Any]) -> Modifiability:
        """
        Determines the modifiability status of a variable based on its properties and returns the corresponding enum
        value.

        This function looks for a 'modifiability' key within `variable_properties`. If present and its value is not
        empty, the function attempts to map this value to an enum member in Modifiability. If the value does not
        correspond to any enum members, a KeyError is raised after logging the error. If 'modifiability' is absent or
        its value is empty, the function defaults to Modifiability.NOT_REQUIRED_AND_UNLOCKED.

        Parameters
        ----------
        variable_name : str
            The name of the variable for which the modifiability status is being determined. Used for error logging.
        variable_properties : Dict[str, Any]
            A dictionary containing the properties of the variable, containing the desired 'modifiability' property.

        Returns
        -------
        Modifiability
            An enum member representing the variable's modifiability status.

        Raises
        ------
        KeyError
            If 'modifiability' in `variable_properties` does not match any enum member in Modifiability. The error
            message includes the invalid modifiability value and suggests valid values.
        """
       ```
    - [x] _is_input_required_upon_initialization()
        ```
        def _is_input_required_upon_initialization(self, variable_name: str, variable_properties: Dict[str, Any]) -> bool:
        """
        Determines whether a variable requires an input value upon initialization based on its modifiability status.

        This function utilizes the '_get_variable_modifiability' method to ascertain the modifiability status of the
        variable identified by 'variable_name' and described by 'variable_properties'. It then checks if the
        modifiability status is either 'REQUIRED_AND_LOCKED' or 'REQUIRED_AND_UNLOCKED', indicating that the variable
        must be initialized with a value.

        Parameters
        ----------
        variable_name : str
            The name of the variable being evaluated for its initialization requirements.
        variable_properties : Dict[str, Any]
            A dictionary containing the properties of the variable, which should include its modifiability status among
            others.

        Returns
        -------
        bool
            True if the variable's modifiability status necessitates an input value upon initialization,
            False otherwise.
        """
        ```
    - [x] _is_modifiable_during_runtime()
        ```
        def _is_modifiable_during_runtime(self, variable_name: str, variable_properties: Dict[str, Any]) -> bool:
        """
        Checks if a variable can be modified during runtime based on its modifiability status.

        This function determines the modifiability status of a variable using the '_get_variable_modifiability' method.
        It assesses whether the variable, identified by 'variable_name' and described by 'variable_properties', is
        allowed to be modified after initialization. A variable is considered modifiable during runtime if its
        modifiability status is either 'REQUIRED_AND_UNLOCKED' or 'NOT_REQUIRED_AND_UNLOCKED'.

        Parameters
        ----------
        variable_name : str
            The name of the variable to check for runtime modifiability.
        variable_properties : Dict[str, Any]
            A dictionary containing the properties of the variable, including details that determine its modifiability.

        Returns
        -------
        bool
            True if the variable is allowed to be modified during runtime, False otherwise.
        """
        ```
    - [x] _handle_missing_data()
        ```
        def _handle_missing_data(self, variable_properties: Dict[str, Any], var_name: str) \
            -> None:
        """
        Handles missing data for a variable, logging errors or warnings based on the context of initialization or
        runtime updates.

        This function determines if it's being called during the initialization phase and checks if the missing variable
        data is required at this stage using '_is_input_required_upon_initialization'. If required, it logs an error and
        raises a KeyError. If not, it logs a warning.

        For missing data during runtime updates, it logs an error and raises a KeyError, assuming the data is necessary
        for the update.

        Parameters
        ----------
        variable_properties : Dict[str, Any]
            Properties of the variable, potentially including its modifiability status.
        var_name : str
            The name of the variable with missing data.

        Raises
        ------
        KeyError
            Raised if the missing data is deemed necessary, either during initialization or for a runtime update.

        Notes
        -----
        - Relies on the caller function's name to determine if it's called during initialization, making the function
          sensitive to naming conventions of the initialization method.
        """
        ```
    - [x] _set_nested_value()
        ```
        def _set_nested_value(self, nested_dict: Dict[str, Any], element_hierarchy: List[str], value: Any) -> \
            Dict[str, Any]:
        """
        Sets a given value within a nested dictionary structure at a specified hierarchical level and returns the
        updated dictionary.

        Notes
        -----
        This function iteratively traverses through the levels of a nested dictionary, guided by a list of keys
        (element_hierarchy). At each level, it checks for the existence of the key. If a key is missing, a new nested
        dictionary is created at that key. The function sets the given value at the final key in the hierarchy.

        While this function modifies the input dictionary in place, it also returns the modified dictionary for
        convenience and chaining operations.

        Parameters
        ----------
        nested_dict : Dict[str, Any]
            The nested dictionary to be updated. This should be the top-level dictionary if the hierarchy spans multiple
            levels.
        element_hierarchy : List[str]
            A list of strings representing the keys that define the path through the nested dictionary to the location
            where the value should be set. Each element in the list corresponds to one level in the nested structure.
        value : Any
            The value to be set at the specified location within the nested dictionary structure.

        Returns
        -------
        Dict[str, Any]
            The updated nested dictionary after setting the specified value.

        Examples
        --------
        >>> nested_dictionary = {'a': {'b': {'c': 1}}}
        >>> updated_dict = self._set_nested_value(nested_dictionary, ['a', 'b', 'd'], 2)
        >>> print(updated_dict)
        {'a': {'b': {'c': 1, 'd': 2}}}

        Note
        ----
        Be cautious of the in-place modification of the input dictionary, which may lead to unintended side effects.

        In other words, this function operates through the side effect of altering the state of `nested_dict` outside
        its local scope. This behavior of in-place modification of the input dictionary introduces potential unintended
        changes.
        """
        ```
- [x] Modifies `InputManager._validate_tabular_element()` and `InputManager._validate_dict_element()`
    - Calls `InputManager._handle_missing_data()` when the data we validate is not included in the input.
- [x] Modifies `InputManager._add_variable_to_pool()`
    - [x] Expands the functionality to handle adding data to a nested structure:
        - [x] Checks if the `variable_name` parameter is a hierarchy (string separated by '.')
        - [x] Checks the variable modifiability to see if it is modifiable during runtime.

image


Isgotkowitz:
A couple things that come to mind while reading through the description and skimming the code:
1. Should conflicting modifiabilities be allowed? If my thinking is correct, this structure would be allowed:
```
"a": {
    "b": "val1"
}
```
with properties (omitting type, etc):
```
"a": {
    "modifiability": "required and locked",
    "properties: {
        "b": {
            "modifiability": "required and unlocked"
        }
    }
}
```
I'm pretty sure "b" would be modifiable at runtime, which is what we want, but this violates the modifiability which is explicitly defined for "a", which seems bad.
2. Why `_get_caller_function()` instead of adding an argument to "_add_variable_to_pool" which indicates if it is runtime or not?
3. What about adding `UNREQUIRED_AND_LOCKED` to `MODIFIABILITY`? A feature that would be nice to have is the ability to specify in the metadata properties a value can be not provided or `null` (I swear I remember us talking about it as a team but I can't find an issue for it). Adding that fourth option to `MODIFIABILITY` would make the code a little more future-proof in case we wind up implementing this feature.
4. Can you please expand on this: "In other words, this function operates through the side effect of altering the state of `nested_dict` outside its local scope. This behavior of in-place modification of the input dictionary introduces potential unintended changes." How can those unintended changes be avoided or logged? (Might be premature in asking this because I don't actually understand what those changes are yet)
5. `_set_nested_value()` can only set values which are nested _directly_ inside objects (dicts). What about setting a value "nested" in an array (i.e. setting a value at a certain index of an array)? If there isn't a use case for this, or even if there is one, it would probably be good to handle it in a different issue/PR but is worth thinking about right now.

allisterakun:
> A couple things that come to mind while reading through the description and skimming the code:
> 
> 1. Should conflicting modifiabilities be allowed? If my thinking is correct, this structure would be allowed:
> 
> ```
> "a": {
>     "b": "val1"
> }
> ```
> 
> with properties (omitting type, etc):
> 
> ```
> "a": {
>     "modifiability": "required and locked",
>     "properties: {
>         "b": {
>             "modifiability": "required and unlocked"
>         }
>     }
> }
> ```
> 
> I'm pretty sure "b" would be modifiable at runtime, which is what we want, but this violates the modifiability which is explicitly defined for "a", which seems bad. 2. Why `_get_caller_function()` instead of adding an argument to "_add_variable_to_pool" which indicates if it is runtime or not? 3. What about adding `UNREQUIRED_AND_LOCKED` to `MODIFIABILITY`? A feature that would be nice to have is the ability to specify in the metadata properties a value can be not provided or `null` (I swear I remember us talking about it as a team but I can't find an issue for it). Adding that fourth option to `MODIFIABILITY` would make the code a little more future-proof in case we wind up implementing this feature. 4. Can you please expand on this: "In other words, this function operates through the side effect of altering the state of `nested_dict` outside its local scope. This behavior of in-place modification of the input dictionary introduces potential unintended changes." How can those unintended changes be avoided or logged? (Might be premature in asking this because I don't actually understand what those changes are yet) 5. `_set_nested_value()` can only set values which are nested _directly_ inside objects (dicts). What about setting a value "nested" in an array (i.e. setting a value at a certain index of an array)? If there isn't a use case for this, or even if there is one, it would probably be good to handle it in a different issue/PR but is worth thinking about right now.

1. This should be allowed. Here is how I interpret the example you provided:
    - "b" is modifiable at runtime (as a sub-property of "a"), but parent dictionary "a" cannot be modified entirely. In other words, some (or all) of the sub-properties of a dictionary can be modified one at a time, but you cannot change the entire dictionary all at once.
2. I chose to use `_get_caller_function()` instead of introducing an argument because:
    1. Avoids passing the same argument through recursive function calls
    Here are two situations in which you can reach `_get_caller_function()`:
        1. `_populate_pool` => `_validate_dict_variable` => (recursively calls `_validate_dict_variable`) => `_handle_missing_data` => `_get_caller_function`
        2. `_add_variable_to_pool` =>`_validate_dict_variable` => (recursively calls `_validate_dict_variable`) => `_handle_missing_data` => `_get_caller_function`
   `_validate_dict_variable()`, `_validate_tabular_variable()`, and `_handle_missing_data` all have to take in an extra argument. The new argument would be set in `_populate_pool` and `_add_variable_to_pool` and be passed down through the recursive call chain to reach `_handle_missing_data`, where the situation is assessed.
    2. Dynamically determine the current state instead of hard-coding a Boolean value => What if a third possible state arises?
3. To specify in the metadata properties that a value can be not provided or `null` upon initialization, we can just set the modifiability to "NOT_REQUIRED_AND_UNLOCKED". If you don't require a variable upon initialization, it won't make sense to lock it and make it unmodifiable during runtime.
4. Basically, the `nested_dict` is a dictionary from the outer scope of the function (`_set_nested_value()`). Python passes the outer-scope object into `_set_nested_value()` by reference so that all the operations done to the `nested_dict` are directly applied to the outer-coped object. It is not considered good practice for a function to modify variables from outside its scope. However, the functionality of `_set_nested_value()` relies entirely on this side effect of manipulating the outer-scope variable.
5. I confirm that `_set_nested_value()` does not handle the situation where you want to modify a specific element within an array structure. And yes, I do believe this can be a separate issue.

Isgotkowitz:
> > A couple things that come to mind while reading through the description and skimming the code:
> > 
> > 1. Should conflicting modifiabilities be allowed? If my thinking is correct, this structure would be allowed:
> > 
> > ```
> > "a": {
> >     "b": "val1"
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > with properties (omitting type, etc):
> > ```
> > "a": {
> >     "modifiability": "required and locked",
> >     "properties: {
> >         "b": {
> >             "modifiability": "required and unlocked"
> >         }
> >     }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm pretty sure "b" would be modifiable at runtime, which is what we want, but this violates the modifiability which is explicitly defined for "a", which seems bad. 2. Why `_get_caller_function()` instead of adding an argument to "_add_variable_to_pool" which indicates if it is runtime or not? 3. What about adding `UNREQUIRED_AND_LOCKED` to `MODIFIABILITY`? A feature that would be nice to have is the ability to specify in the metadata properties a value can be not provided or `null` (I swear I remember us talking about it as a team but I can't find an issue for it). Adding that fourth option to `MODIFIABILITY` would make the code a little more future-proof in case we wind up implementing this feature. 4. Can you please expand on this: "In other words, this function operates through the side effect of altering the state of `nested_dict` outside its local scope. This behavior of in-place modification of the input dictionary introduces potential unintended changes." How can those unintended changes be avoided or logged? (Might be premature in asking this because I don't actually understand what those changes are yet) 5. `_set_nested_value()` can only set values which are nested _directly_ inside objects (dicts). What about setting a value "nested" in an array (i.e. setting a value at a certain index of an array)? If there isn't a use case for this, or even if there is one, it would probably be good to handle it in a different issue/PR but is worth thinking about right now.
> 
> 1. This should be allowed. Here is how I interpret the example you provided:
>    
>    * "b" is modifiable at runtime (as a sub-property of "a"), but parent dictionary "a" cannot be modified entirely. In other words, some (or all) of the sub-properties of a dictionary can be modified one at a time, but you cannot change the entire dictionary all at once.

Ok, sounds good. I just wanted to confirm that this would be acceptable behavior.

> 2. I chose to use `_get_caller_function()` instead of introducing an argument because:
>    
>    1. Avoids passing the same argument through recursive function calls
>       Here are two situations in which you can reach `_get_caller_function()`:
>       
>       1. `_populate_pool` => `_validate_dict_variable` => (recursively calls `_validate_dict_variable`) => `_handle_missing_data` => `_get_caller_function`
>       2. `_add_variable_to_pool` =>`_validate_dict_variable` => (recursively calls `_validate_dict_variable`) => `_handle_missing_data` => `_get_caller_function`
>          `_validate_dict_variable()`, `_validate_tabular_variable()`, and `_handle_missing_data` all have to take in an extra argument. The new argument would be set in `_populate_pool` and `_add_variable_to_pool` and be passed down through the recursive call chain to reach `_handle_missing_data`, where the situation is assessed.
>    2. Dynamically determine the current state instead of hard-coding a Boolean value => What if a third possible state arises?

Good thinking - I did not consider the possibility of third state.

> 3. To specify in the metadata properties that a value can be not provided or `null` upon initialization, we can just set the modifiability to "NOT_REQUIRED_AND_UNLOCKED". If you don't require a variable upon initialization, it won't make sense to lock it and make it unmodifiable during runtime.

I disagree. If the functionality was present for allowing a user to tell RuFaS that it wants an input to be null (implicitly or explicitly), preserving that truth in the input manager might be important and it should be possible. However a nullable input is currently not an option, and even if it was it probably would not make a difference whether it was locked or not, so it isn't worth worrying about.

> 4. Basically, the `nested_dict` is a dictionary from the outer scope of the function (`_set_nested_value()`). Python passes the outer-scope object into `_set_nested_value()` by reference so that all the operations done to the `nested_dict` are directly applied to the outer-coped object. It is not considered good practice for a function to modify variables from outside its scope. However, the functionality of `_set_nested_value()` relies entirely on this side effect of manipulating the outer-scope variable.

Gotcha

> 5. I confirm that `_set_nested_value()` does not handle the situation where you want to modify a specific element within an array structure. And yes, I do believe this can be a separate issue.

Sounds good. Seems like an issue that shouldn't have any time put into it unless a need/use-case for it comes up.

RUFAS/input_manager.py
```diff
@@ -280,6 +315,209 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+                return Modifiability.__getitem__('_'.join(variable_properties["modifiability"].strip().upper().split()))
+            except KeyError:
+                om.add_error("Unknown modifiability entry",
+                             f"Unknown modifiability value of {variable_properties['modifiability']} for variable "
+                             f"{variable_name}. Modifiability must be one of {Modifiability.values()}",
+                             info_map)
+                raise KeyError(f"Unknown modifiability value of {variable_properties['modifiability']} for variable "
+                               f"{variable_name}. Modifiability must be one of {Modifiability.values()}")
+        else:
+            return Modifiability.__getitem__("NOT_REQUIRED_AND_UNLOCKED")
```
```suggestion
        modifiability = variable_properties.get("modifiability")
        if not modifiability:
            return Modifiability.__getitem__("NOT_REQUIRED_AND_UNLOCKED")

        try:
            return Modifiability.__getitem__('_'.join(modifiability.strip().upper().split()))
        except KeyError:
            om.add_error(
                "Unknown modifiability entry",
                f"Unknown modifiability value of {modifiability} for variable {variable_name}. Modifiability must be "
                f"one of {Modifiability.values()}",
                info_map
            )
            raise KeyError(
                f"Unknown modifiability value of {modifiability} for variable {variable_name}. Modifiability must be "
                f"one of {Modifiability.values()}"
            )
```

RUFAS/input_manager.py
```diff
@@ -280,6 +315,209 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
 
         return filtered_input_data
 
+    def _get_variable_modifiability(self, variable_name: str, variable_properties: Dict[str, Any]) -> Modifiability:
+        """
+        Determines the modifiability status of a variable based on its properties and returns the corresponding enum
+        value.
+
+        This function looks for a 'modifiability' key within `variable_properties`. If present and its value is not
```
Isgotkowitz, line 401:
```suggestion
        Notes
        -----
        This function looks for a 'modifiability' key within `variable_properties`. If present and its value is not
```

RUFAS/input_manager.py
```diff
@@ -280,6 +315,209 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+        -------
+        str
+            The name of the function that called the current function.
+
+        Notes
+        -----
+        - The function uses the `inspect` module to access the current and previous frames in the call stack.
+        - It is designed to work within a single thread; behavior in multi-threaded environments may be unpredictable.
+        - Special care is taken to handle the call stack in a way that avoids creating references that could prevent
+          garbage collection of stack frames.
```
```suggestion

        Returns
        -------
        str
            The name of the function that called the current function.

        Notes
        -----
        This function examines the call stack to find the current function's name and then identifies the name of the
        function that called it. This is achieved by navigating the call stack frames and extracting the function names,
        ensuring to remove any duplicates to preserve the actual call sequence.

        - The function uses the `inspect` module to access the current and previous frames in the call stack.
        - It is designed to work within a single thread; behavior in multi-threaded environments may be unpredictable.
        - Special care is taken to handle the call stack in a way that avoids creating references that could prevent
          garbage collection of stack frames.
```

RUFAS/input_manager.py
```diff
@@ -280,6 +315,209 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+        """
+        current_function = inspect.currentframe().f_back.f_back.f_code.co_name
+
+        call_stack = inspect.stack()
+        call_stack_functions = list(dict.fromkeys([call_stack[i].function for i in range(len(call_stack))]))
+
+        return call_stack_functions[call_stack_functions.index(current_function) + 1]
+
+    def _is_input_required_upon_initialization(self, variable_name: str, variable_properties: Dict[str, Any]) -> bool:
```
Isgotkowitz, line 443: I think this method and `_is_modifiable_at_runtime()` could be eliminated and replaced with a little bit of extra logic in `Modifiability` and the spots where these two methods are called.

In `Modifiability`, two methods could be added: `get_required_during_initialization()` and `get_modifiable_at_runtime()`, which just return `(REQUIRED_AND_LOCKED, REQUIRED_AND_UNLOCKED)` and `(REQUIRED_AND_UNLOCKED, NOT_REQUIRED_AND_UNLOCKED)` respectively.

In the places where `_is_modifiable_at_runtime()` and `_is_input_required_upon_init()` are called, they can be replaced with
```python
is_modifiable_during_runtime = self._get_variable_modifiability() in Modifiability.get_modifiable_at_runtime()
```
and same thing for `is_input_required_upon_initialization`

RUFAS/input_manager.py
```diff
@@ -280,6 +315,209 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+
+        Raises
+        ------
+        KeyError
+            Raised if the missing data is deemed necessary, either during initialization or for a runtime update.
+
+        Notes
+        -----
+        - Relies on the caller function's name to determine if it's called during initialization, making the function
+          sensitive to naming conventions of the initialization method.
```
```suggestion

        Parameters
        ----------
        variable_properties : Dict[str, Any]
            Properties of the variable, potentially including its modifiability status.
        var_name : str
            The name of the variable with missing data.

        Raises
        ------
        KeyError
            Raised if the missing data is deemed necessary, either during initialization or for a runtime update.

        Notes
        -----
        This function determines if it's being called during the initialization phase and checks if the missing variable
        data is required at this stage using '_is_input_required_upon_initialization'. If required, it logs an error and
        raises a KeyError. If not, it logs a warning.

        For missing data during runtime updates, it logs an error and raises a KeyError, assuming the data is necessary
        for the update.
        
        - Relies on the caller function's name to determine if it's called during initialization, making the function
          sensitive to naming conventions of the initialization method.
```

RUFAS/input_manager.py
```diff
@@ -280,6 +315,209 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+                    "variable upon program initialization, setting the variable value to None.",
+                    info_map)
+                return
+        else:
+            om.add_error(
+                "Validation: key not found in data -- value required to update variable during runtime",
+                f"Key {var_name} not found in data. A value is required for to update variable during runtime.",
+                info_map)
+            raise KeyError(f"Key {var_name} not found in data. A value is required for to update variable "
+                           "during runtime.")
```
```suggestion
        if not is_initialization:
            om.add_error(
                "Validation: key not found in data -- value required to update variable during runtime",
                f"Key {var_name} not found in data. A value is required for to update variable during runtime.",
                info_map)
            raise KeyError(f"Key {var_name} not found in data. A value is required for to update variable "
                           "during runtime.")

        is_input_required_upon_initialization = self._is_input_required_upon_initialization(
            variable_name=var_name,
            variable_properties=variable_properties)
        if is_input_required_upon_initialization:
            om.add_error("Validation: key not found in input data -- input required upon initialization",
                            f"Key {var_name} not found in input data. Input value is required for this "
                            "variable upon program initialization.",
                            info_map)
            raise KeyError("Key {var_name} not found in input data. Input value is required for this "
                           "variable upon program initialization.")
        
       om.add_warning(
            "Validation: key not found in input data -- input not required upon initialization",
            f"Key {var_name} not found in input data. Input value is not required for this "
            "variable upon program initialization, setting the variable value to None.",
            info_map)
```

RUFAS/input_manager.py
```diff
@@ -949,22 +1200,79 @@ def _metadata_properties_exist(self, variable_name: str, properties_blob_key: st
...
+        Dict[str, Any]
+            The updated nested dictionary after setting the specified value.
+
+        Examples
+        --------
+        >>> nested_dictionary = {'a': {'b': {'c': 1}}}
+        >>> updated_dict = self._set_nested_value(nested_dictionary, ['a', 'b', 'd'], 2)
+        >>> print(updated_dict)
+        {'a': {'b': {'c': 1, 'd': 2}}}
+
```
Isgotkowitz, line 1625:
```suggestion

        >>> nested_dictionary = {'a': {'b': {'c': 1}}}
        >>> updated_dict = self._set_nested_value(nested_dictionary, ['a', 'b', 'c'], 2)
        >>> print(updated_dict)
        {'a': {'b': {'c': 2}}}
```

RUFAS/input_manager.py
```diff
@@ -949,22 +1200,79 @@ def _metadata_properties_exist(self, variable_name: str, properties_blob_key: st
...
+        Note
+        ----
+        Be cautious of the in-place modification of the input dictionary, which may lead to unintended side effects.
+
+        In other words, this function operates through the side effect of altering the state of `nested_dict` outside
+        its local scope. This behavior of in-place modification of the input dictionary introduces potential unintended
+        changes.
```
Isgotkowitz: This section should be moved into "Notes"

RUFAS/input_manager.py
```diff
@@ -990,14 +1300,42 @@ def _add_variable_to_pool(self, variable_name: str, data: Dict[str, Any], proper
...
+            if eager_termination:
+                om.add_error("IM Runtime Modification",
+                             f"{variable_name} is not modifiable during runtime.",
+                             info_map)
+                raise PermissionError("IM Runtime Modification Error: "
+                                      f"{variable_name} is not modifiable during runtime.")
+            else:
+                om.add_warning("IM Runtime Modification",
+                               f"{variable_name} is not modifiable during runtime.",
+                               info_map)
```
```suggestion
        if not is_modifiable_during_runtime and eager_termination:
            om.add_error("IM Runtime Modification", f"{variable_name} is not modifiable during runtime.", info_map)
            raise PermissionError(f"IM Runtime Modification Error: {variable_name} is not modifiable during runtime.")
        elif not is_modifiable_during_runtime:
            om.add_warning("IM Runtime Modification", f"{variable_name} is not modifiable during runtime.", info_map)
```

RUFAS/input_manager.py
```diff
@@ -13,6 +15,39 @@
...
+    REQUIRED_AND_LOCKED : str
+        Indicates the variable must be initialized with a value and cannot be modified thereafter.
+    REQUIRED_AND_UNLOCKED : str
+        Indicates the variable must be initialized with a value but can be modified during runtime.
+    NOT_REQUIRED_AND_UNLOCKED : str
+        Indicates the variable does not need to be initialized with a value and can be modified during runtime.
+    """
+    REQUIRED_AND_LOCKED: str = "required and locked"
+    REQUIRED_AND_UNLOCKED: str = "required and unlocked"
+    NOT_REQUIRED_AND_UNLOCKED: str = "not required and unlocked"
```
```suggestion
    REQUIRED_LOCKED : str
        Indicates the variable must be initialized with a value and cannot be modified thereafter.
    REQUIRED_UNLOCKED : str
        Indicates the variable must be initialized with a value but can be modified during runtime.
    UNREQUIRED_UNLOCKED : str
        Indicates the variable does not need to be initialized with a value and can be modified during runtime.
    """
    REQUIRED_LOCKED: str = "required and locked"
    REQUIRED_UNLOCKED: str = "required and unlocked"
    UNREQUIRED_UNLOCKED: str = "unrequired and unlocked"
```

Isgotkowitz (approved):
LGTM, nice job Allister. Left a bunch of suggestions that I think would improve the code, but all are take it or leave it.

RUFAS/input_manager.py
```diff
@@ -300,6 +343,214 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+        Returns
+        -------
+        Modifiability
+            An enum member representing the variable's modifiability status.
+
+        Raises
+        ------
+        KeyError
+            If 'modifiability' in `variable_properties` does not match any enum member in Modifiability. The error
+            message includes the invalid modifiability value and suggests valid values.
```
PooyaHekmati, line 422: This sounds more like a ValueError.
https://docs.python.org/3/library/exceptions.html#ValueError

PooyaHekmati: Do we even need to raise an exception? Can we use the default value and log a warning to let the user know about this?

RUFAS/input_manager.py
```diff
@@ -300,6 +343,214 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+        """
+        info_map = {"class": self.__class__.__name__,
+                    "function": self._get_variable_modifiability.__name__,
+                    }
+
+        modifiability = variable_properties.get("modifiability")
+        if not modifiability:
+            return Modifiability.__getitem__("UNREQUIRED_UNLOCKED")
```
```suggestion
        default = "UNREQUIRED UNLOCKED"
        modifiability = variable_properties.get("modifiability", default)
```

RUFAS/input_manager.py
```diff
@@ -300,6 +343,214 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+        -----
+        - The determination is based solely on the modifiability status of the variable. Other factors that might affect
+          initialization requirements are not considered by this function.
+        - This function depends on the 'Modifiability' enum and the '_get_variable_modifiability' method to function
+          correctly. Any changes to these components may affect the behavior of this function.
+        """
+        variable_modifiability = self._get_variable_modifiability(variable_name=variable_name,
+                                                                  variable_properties=variable_properties)
+        return (variable_modifiability == Modifiability.REQUIRED_AND_LOCKED) or \
+               (variable_modifiability == variable_modifiability.REQUIRED_AND_UNLOCKED)
```
```suggestion
        return variable_modifiability in Modifiability.get_required_during_initialization()
```

RUFAS/input_manager.py
```diff
@@ -300,6 +343,214 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+        Returns
+        -------
+        bool
+            True if the variable's modifiability status necessitates an input value upon initialization,
+            False otherwise.
+
+        Notes
+        -----
+        - The determination is based solely on the modifiability status of the variable. Other factors that might affect
+          initialization requirements are not considered by this function.
```
PooyaHekmati: like what other factors?

RUFAS/input_manager.py
```diff
@@ -300,6 +343,214 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+        - This function depends on the 'Modifiability' enum and the '_get_variable_modifiability' method to function
+          correctly. Any changes to these components may affect the behavior of this function.
```
```suggestion
```
All functions depend on the correctness of their dependencies, right?

RUFAS/input_manager.py
```diff
@@ -300,6 +343,214 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+        return call_stack_functions[call_stack_functions.index(current_function) + 1]
+
+    def _is_input_required_upon_initialization(self, variable_name: str, variable_properties: Dict[str, Any]) -> bool:
+        """
+        Determines whether a variable requires an input value upon initialization based on its modifiability status.
+
+        This function utilizes the '_get_variable_modifiability' method to ascertain the modifiability status of the
+        variable identified by 'variable_name' and described by 'variable_properties'. It then checks if the
+        modifiability status is either 'REQUIRED_AND_LOCKED' or 'REQUIRED_AND_UNLOCKED', indicating that the variable
+        must be initialized with a value.
```
PooyaHekmati, line 450: why hardcoding when we already have `Modifiability.get_required_during_initialization()`?

RUFAS/input_manager.py
```diff
@@ -300,6 +343,214 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+        - The function relies on the 'Modifiability' enum and the '_get_variable_modifiability' method. Changes to these
+          components may impact the function's behavior.
+        - This function does not consider other aspects of the variable that might affect its ability to be modified at
+          runtime, focusing solely on its declared modifiability status.
+        """
+        variable_modifiability = self._get_variable_modifiability(variable_name=variable_name,
+                                                                  variable_properties=variable_properties)
+        return (variable_modifiability == Modifiability.NOT_REQUIRED_AND_UNLOCKED) or \
+               (variable_modifiability == variable_modifiability.REQUIRED_AND_UNLOCKED)
+
```
PooyaHekmati: same as `_is_input_required`, we can use `Modifiability.get_modifiable_at_runtime` here.

RUFAS/input_manager.py
```diff
@@ -300,6 +343,214 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+        -----
+        This function determines if it's being called during the initialization phase and checks if the missing variable
+        data is required at this stage using '_is_input_required_upon_initialization'. If required, it logs an error and
+        raises a KeyError. If not, it logs a warning.
+
+        For missing data during runtime updates, it logs an error and raises a KeyError, assuming the data is necessary
+        for the update.
+
+        - Relies on the caller function's name to determine if it's called during initialization, making the function
+          sensitive to naming conventions of the initialization method.
```
PooyaHekmati: I don't understand this:
>making the functionn sensitive to naming conventions of the initialization method.

RUFAS/input_manager.py
```diff
@@ -300,6 +343,214 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+            Raised if the missing data is deemed necessary, either during initialization or for a runtime update.
+
+        Notes
+        -----
+        This function determines if it's being called during the initialization phase and checks if the missing variable
+        data is required at this stage using '_is_input_required_upon_initialization'. If required, it logs an error and
+        raises a KeyError. If not, it logs a warning.
+
+        For missing data during runtime updates, it logs an error and raises a KeyError, assuming the data is necessary
+        for the update.
```
```suggestion
```
you covered this in the riases section

RUFAS/input_manager.py
```diff
@@ -300,6 +343,214 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+
+        - Relies on the caller function's name to determine if it's called during initialization, making the function
+          sensitive to naming conventions of the initialization method.
+        """
+        caller_function = self._get_caller_function()
+        is_initialization = caller_function == self._populate_pool.__name__
```
PooyaHekmati: I know we talked about this and I have seen Ed's comment too. But it seems adding an arg to this function named something like `called_by_populate_pool` or `called_at_initialization` and defaulting it to false would simplify the process, making it threadsafe again.

RUFAS/input_manager.py
```diff
@@ -300,6 +343,214 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": caller_function
+        }
+        if not is_initialization:
+            om.add_error(
+                "Validation: key not found in data -- value required to update variable during runtime",
+                f"Key {var_name} not found in data. A value is required for to update variable during runtime.",
```
```suggestion
                "missing required data",
                f"Key {var_name} not found in data. A value is required for to update variable during runtime.",
```
don't repeat yourself.

RUFAS/input_manager.py
```diff
@@ -300,6 +343,214 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+                info_map)
+            raise KeyError(f"Key {var_name} not found in data. A value is required for to update variable "
+                           "during runtime.")
+
+        is_input_required_upon_initialization = self._get_variable_modifiability(
+            variable_name=var_name,
+            variable_properties=variable_properties) in Modifiability.get_required_during_initialization()
+        if is_input_required_upon_initialization:
```
PooyaHekmati: can we simplify this using `:=` operator?

RUFAS/input_manager.py
```diff
@@ -300,6 +343,214 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+            om.add_error("Validation: key not found in input data -- input required upon initialization",
+                         f"Key {var_name} not found in input data. Input value is required for this "
+                         "variable upon program initialization.",
+                         info_map)
```
PooyaHekmati: don't repeat yourself

RUFAS/input_manager.py
```diff
@@ -300,6 +343,214 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+        variable_modifiability = self._get_variable_modifiability(variable_name=variable_name,
+                                                                  variable_properties=variable_properties)
+        return (variable_modifiability == Modifiability.NOT_REQUIRED_AND_UNLOCKED) or \
+               (variable_modifiability == variable_modifiability.REQUIRED_AND_UNLOCKED)
+
+    def _handle_missing_data(self, variable_properties: Dict[str, Any], var_name: str) \
+            -> None:
+        """
+        Handles missing data for a variable, logging errors or warnings based on the context of initialization or
+        runtime updates.
```
```suggestion
    def _log_missing_data(self, variable_properties: Dict[str, Any], var_name: str) \
            -> None:
        """
        Handles logging for missing data for a variable, logging errors or warnings based on the context of initialization or
        runtime updates.
```

PooyaHekmati (changes requested):
Looks good overall, left some comments.
Didn't get to review the whole PR though.

PooyaHekmati (approved):

PR #1089, opened 2024-01-08
Soil N leaching and runoff calibration

### LINKED ISSUE #495: `leaching_runoff_erosion.py` does not have scientific primary source
# Problem
Many of the equations used in `nitrogen_cycling/leaching_runoff_erosion.py` have an unknown origin at this time. These include
- S.4.C.5
- S.4.C.6
- S.4.C.7
- S.4.C.8

# Solution
Comb the literature looking for where these equations may have originated from. Also, talk with SME's (Pete, Xuesong, etc) about why they're there and document in the code.

Note: this issue is linked to #486

### LINKED ISSUE #979: Evaluate soil nitrogen dynamics

Isgotkowitz:
This PR changes the way that nitrogen is removed from different water-related processes and adjusts empirical parameters in soil profile inputs.

More specifically, it introduces a single method (`_calculate_nitrogen_removed_by_water()`) that is used to calculate the amounts of nitrogen removed by leaching and runoff. This method replaces several previous methods, which is why there is such a high number of deletions in this PR. The default factors that govern denitrification and ammonium volatilization have also been adjusted in this PR (in both the metadata and all soil profiles currently in the codebase). It is important to note that these factors have only been tuned using a single dataset on a single soil profile, so it is important when running a different simulation to consider if these factors should be adjusted based on your soil profile.

Closes #979, closes #495.


RUFAS/routines/field/soil/nitrogen_cycling/leaching_runoff_erosion.py
```diff
@@ -10,6 +11,21 @@
...
+Reference:
+Vadas, P. A., & Powell, J. M. (2019). Nutrient mass balance and fate in dairy cattle lots with different surface
+materials. Transactions of the ASABE, 62(1), 131–138. https://doi.org/10.13031/trans.12901
+
+"""
+NITRATE_RUNOFF_COEFFICIENT = 1e-8
+AMMONIUM_RUNOFF_COEFFICIENT = 8e-6
+NITRATE_PERCOLATION_COEFFICIENT = 6e-8
+AMMONIUM_PERCOLATION_COEFFICIENT = 4e-6
+ACTIVE_ORGANIC_NITROGEN_PERCOLATION_COEFFICIENT = 5e-8
```
allisterakun, line 27: Just a thought, and this is not within the scope of this PR:

Why not put these constant coefficients into `RUFAS.routines.field.crop_and_soil_constants`?

allisterakun (approved):
The code looks good. Nice job, Ed!

RUFAS/routines/field/soil/nitrogen_cycling/leaching_runoff_erosion.py
Isgotkowitz: I think it makes sense to place them here because this is the only module they will ever be used in. All constants in the Crop and Soil constants file are either used in multiple places or are general by nature (i.e. they are for unit conversion).

krb78 (approved):
Looks great Ed! 🥇

PR #1092, opened 2024-01-08
Tested all functions in calf.py 

0luchi:
- Tested all functions in calf.py to improve coverage to 100%


tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,170 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+
+@pytest.mark.parametrize(
+    'semen_type, conventional_semen, sexed_semen, birth_rate, female_calf_rate',
+    [
+        ("conventional", -2.0, -3.0, -4.0, -5.0),
+        ("unconventional", 200.0, 500.0, 300.0, 400.0),
+    ]
+)
+def test_calf_init_values(mocker: MockerFixture, semen_type: str, conventional_semen: str,
+                          sexed_semen: str, birth_rate: str, female_calf_rate: str) -> None:
```
JoeWaddell, line 1525: Missing docstring here (and in others), the format should mimic the others in this file, so it tells the reader which method/function it's testing, and which file it came from. Both help with duplicate/similar method names, and also CTRL+SHIFT+F searching for method names.

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,170 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
     assert life_cycle_manager.avg_heifer_culling_age == approx(0.0)
     assert life_cycle_manager.avg_cow_culling_age == approx(0.0)
     assert life_cycle_manager.avg_mature_body_weight == approx(0.0)
+
```
JoeWaddell, line 1469: We'll need to add a test here for the `__init__` call as well.

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,170 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+    mock_args = mocker.MagicMock()
+    Calf.init_values(calf, mock_args)
+
+    # assert
+    if 0.5 < birth_rate:
+        assert calf.culled
+    if 0.5 < conventional_semen:
+        assert calf.gender == "male"
+    else:
+        assert calf.gender == "female"
```
JoeWaddell, line 1560: Looks like line 75 in calf.py isn't covered here: I think it'll require another line in the paramaterized args at the top, but it might be tricky.

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,170 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+    mock_args = mocker.MagicMock()
+    calf.events.init_from_string = mocker.MagicMock(autospec=AnimalEvents)
+
+    # act
+    Calf.assign_calf_values(calf, mock_args)
+
+    # assert
+    assert not calf.culled
+    assert not calf.sold
+    assert calf.gender == "female"
```
JoeWaddell, line 1584: Here you could hardcode args values, and then check that the calf correctly had those applied: e.g.
`test_args = {'birth_weight': 100, 'body_weight': 150, 'wean_weight': 101, 'mature_body_weight': 300}`
then after the act: `assert calf.birth_weight == 100`...etc

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,170 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+                 return_value=calf)
+    feed = mocker.MagicMock(autospec=Feed)
+    patch_base_manure = mocker.patch.object(
+        Calf, 'calc_base_manure', return_value=calf
+    )
+    Calf.calc_base_manure(calf, feed, "methane_model")
+    patch_base_manure.assert_called()
+
+
+def test_calf_phosphorus_rqmts(mocker: MockerFixture) -> None:
```
JoeWaddell, line 1646: By the way, nice job making these test names unique!

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,170 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+                 return_value=calf)
+    animal_config = {
+        "wean_day": wean_day,
+    }
+    mocker.patch('RUFAS.routines.animal.life_cycle.life_cycle.AnimalBase.config', animal_config)
+    calf.update_body_weight_history = mocker.MagicMock()
+    calf.get_non_preg_bw_change = mocker.MagicMock()
+    calf.days_born = days_born
+
+    # act
```
JoeWaddell, line 1688: One more set of checks here to assert that the wean_weight and daily_growth are set properly would round this out nicely!

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,170 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+def test_calf_calc_manure_excretion(mocker: MockerFixture) -> None:
+    # test calc_manure_excretion
+    calf = mocker.MagicMock(autospec=Calf)
+    mocker.patch('RUFAS.routines.animal.life_cycle.life_cycle.Calf',
+                 return_value=calf)
+    feed = mocker.MagicMock(autospec=Feed)
+    patch_base_manure = mocker.patch.object(
+        Calf, 'calc_base_manure', return_value=calf
+    )
+    Calf.calc_base_manure(calf, feed, "methane_model")
```
JoeWaddell: The act here isn't calling calc_manure_excretion, but the calc_base_manure method in animal_base.

JoeWaddell (changes requested):
Looking good, it needs a few more assertions and docstrings, but otherwise it's great to see the coverage jump up for calf.py
Nice work!

JoeWaddell (approved):
Nice job! LGTM: good expansion of test coverage for calf.py 

In future updates we'll likely move these methods to their own test file, but given that there are a few unit test PRs open, best to wait until those are merged.

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,237 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+                     "birth_weight": 0,
+                     "keep_female_calf_rate": 2.0,
+                     "mature_body_weight_avg": 1.0,
+                     "mature_body_weight_std": 1.0
+                     }
+    mocker.patch('RUFAS.routines.animal.life_cycle.life_cycle.AnimalBase.config',
+                 new_callable=mocker.PropertyMock, return_value=animal_config)
+    calf = Calf(args)
+
+    print(calf.birth_weight)
```
Isgotkowitz:
```suggestion
```

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,237 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+
+    if "body_weight" in args:
+        assert calf.gender == "female"
+        assert not calf.sold
+        assert calf.body_weight == args["body_weight"]
+        assert calf.wean_weight == args["wean_weight"]
+    else:
+        assert calf.gender == "male"
+        assert calf.sold
+        assert calf.birth_weight == 2.0
```
Isgotkowitz:
```suggestion
```
Birth weight is checked before getting into the if/else block.

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,237 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+    [
+        ({"id": 1, "breed": 2, "birth_date": 1, "wean_weight": 1,
+         "days_born": 1, "body_weight": 1.0, "events": "3: simulation_day=0, event", "mature_body_weight": 2.0,
+          "birth_weight": 2.0, "p_init": 1}),
+        ({"id": 2, "breed": 1, "birth_date": 1, "events": "3: simulation_day=0, event", "mature_body_weight": 2.0,
+         "wean_weight": 1, "days_born": 1, "birth_weight": 2, "p_init": 1}),
+    ]
+)
+def test_init_calf(mocker: MockerFixture, args: dict) -> None:
+    """Unit test for __init__() method in file routines/animal/life_cycle/calf.py."""
```
Isgotkowitz, line 1482: This method achieves good coverage, however it is doing more than just testing `__init__`. It is also indirectly testing the methods `assign_calf_values()` and `init_values()`. Ideally a unit test would not be doing this, though an integration test would. Unfortunately mocking methods that are called in `__init__` is (in my experience) a pain and this test works well, so it is not worth rewriting it in this case.

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,237 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+        "wean_day": wean_day,
+    }
+    mocker.patch('RUFAS.routines.animal.life_cycle.life_cycle.AnimalBase.config', animal_config)
+    calf.update_body_weight_history = mocker.MagicMock()
+    calf.get_non_preg_bw_change = mocker.MagicMock()
+    calf.days_born = days_born
+
+    # act
+    sim_day = 2
+    weaned = False
```
Isgotkowitz:
```suggestion
```

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,237 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+    weaned = Calf.update(calf, sim_day)
+    calf.days_born += 1
+    calf.update_body_weight_history.assert_called()
+    if calf.days_born == wean_day:
```
Isgotkowitz: Incrementing `calf.days_born` and asserting `calf.days_born == wean_day` is duplicating/undoing a lot of logic that is being tested. A better approach would be defining the expected return value in the test parameters, then checking that what you have defined as your expectation matches what you are actually seeing in the test behavior.

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,237 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+        assert weaned
+        assert calf.wean_weight == calf.body_weight
+    else:
+        assert not weaned
+        assert calf.daily_growth == calf.body_weight - calf.body_weight
```
Isgotkowitz: Daily growth is set whether or not the calf is weaned, but in this test it is only checked if the calf is not weaned. Also, this is asserting that when a calf is weaned it doesn't grow at all, which is not what the code does.

tests/animal_module_tests/test_animal_life_cycle.py
```diff
@@ -1463,3 +1466,237 @@ def test_reset_daily_stats(life_cycle_manager: LifeCycleManager) -> None:
...
+
+    calf = mocker.MagicMock(autospec=Calf)
+    mocker.patch('RUFAS.routines.animal.life_cycle.life_cycle.Calf',
+                 return_value=calf)
+    animal_config = {
+        "wean_day": wean_day,
+    }
+    mocker.patch('RUFAS.routines.animal.life_cycle.life_cycle.AnimalBase.config', animal_config)
+    calf.update_body_weight_history = mocker.MagicMock()
+    calf.get_non_preg_bw_change = mocker.MagicMock()
```
Isgotkowitz:
```suggestion
```

Isgotkowitz (approved):
Nice tests Oluchi. I noticed `test_update_calf()` in the testing logic doesn't seem to line up with the method logic. There are also some spots in `test_init_calf()` that should be cleaned up that I left comments on. Other than those places though, I think things are solid.

PR #1093, opened 2024-01-08
Fix OM create_directory testing

### LINKED ISSUE #1070: Fix `test_create_directory_successful` in `test_output_manager.py`
PR 967 has introduced a new function called `OutputManager.create_directory`. The test case `test_create_directory_successful` in `test_output_manager.py` is not set up correctly: The function's main action is to call `Path.mkdir`. The test case currently is not mocking this, instead, it is directly calling it, making an actual directory, and then checking how many logs and errors are reported to the OutputManager. There are several issues here:

1. The main action of the function is unchecked.
2. The test case has side effects -- every time the test case is run, it really makes a directory.
3. The test case has no teardown/clean up. After it is done testing, the directory that it created is still there.

Fix the test case by mocking `Path.mkdir` and check how it is called.

ew3361zh:
PR to address issue #1070.

This PR fixes 2 tests in test_output_manager.py both of which were creating dirs when testing was run:

1. `test_route_save_functions_graph` now mocks the call to `create_directory()` so that a graphics dir is not created when this test is run.
2. `test_create_directory_successful` now mocks Path.mkdir so a directory is not created when this test is run.

I did also try to delete the dirs that had been previously created but they were empty and I don't think were added to GitHub anyhow.


allisterakun (approved):
Look good! All tests behave as expected on my end. Good job, Niko!

Isgotkowitz (approved):
LGTM. Thank you for taking care of this Niko, that `new/dir/to/create` has really been bugging me.

PR #1094, opened 2024-01-08
Feedmanager query

PooyaHekmati:
This PR implements querying stored feed/crops functionality for Feed Manager.
[design doc](https://docs.google.com/document/d/1C86lch2dkYBERR-qLDmE7L7n1gYQjETrIKUyrQpWrPs/edit?usp=sharing)
Adds unit tests for different combinations of query.
The query can be done based on either `CropType`, `CropCategroy`, `StorageType`, or any combination of them.
The result is passed as a list of type `QUERY_RESULT_DATA_TYPE`, using `_query_result_factory`.


KevinP-B (approved):
Queries for feeds all look good.

RUFAS/routines/feed_storage/feed_manager.py
```diff
@@ -115,22 +127,66 @@ def give_feed(self, amount: float, crop_type: CropType) -> float:
...
         Parameters
         ----------
-        crop_type : CropType, optional
-            The type of crop to query (default is None, which queries all types).
+        query_crop_types : List[CropType], optional
+            The types of crop to query (default is None, which queries all types).
+        query_crop_categories : List[CropCategory], optional
+            The categories of crop to query (default is None, which queries all categories).
+        query_storage_types : List[StorageType], optional
+            The types of storage to query (default is None, which queries all types).
```
Isgotkowitz:
```suggestion
        query_crop_types : List[CropType], optional, default=None
            The types of crop to query (if None, all crop types are queried).
        query_crop_categories : List[CropCategory], optional, default=None
            The categories of crop to query (if None, all crop categories are queried).
        query_storage_types : List[StorageType], optional, default=None
            The types of storage to query (if None, all storages types are queried).
```
This would strictly meet the numpy docstring style for specifying defaults.

It may be worthwhile to add an examples section here to help people understand this method faster. I was able to figure it out alright, but it took me a little bit of time working through the test cases. Up to you.

Isgotkowitz (approved):
One suggestion on how to improve one of the docstrings. Otherwise looks great. Nice job Pooya.

JoeWaddell (approved):
Excellent work! The logic is easy to follow, and the tests cover it nicely.

RUFAS/routines/feed_storage/feed_manager.py
PooyaHekmati: Thanks for catching this. I would be happy to provide examples. Can you please tell me more about which parts were challenging to understand so I can focus on them?

Isgotkowitz: Yes - understanding how the different query parameters are combined with each other.

For example in the `query_available_feeds_combination()` I wasn't sure at first why you were only getting one type of corn and why no other crops were making it through. I had to think through the different permutations of type, category, and storage before it became really clear how the different parameters are used to create a filter.

So writing up that test as an example might be really helpful, where it explains that even though there are four different crops currently stored by the Feed Manager, only dry corn grain is returned because it is the only permutation in the query parameters that is valid and is present in the Feed Manager.

PR #1098, opened 2024-01-09
Integrate Feed Manager into the Field Manager

Isgotkowitz:
This PR integrates the Feed Manager into the Field Manager. This is accomplished by passing the Feed Manager into the Field Manager, where it is passed into individual fields and used by Crops to store harvested yields.

Things I want to emphasize/discuss:
- For storage types, all grains map to Dry, all silages map to Bunker, all baleages map to Baleage, and all hays map to Protected Tarped. 
- The Crop module uses a single set of default crop qualities that I set arbitrarily. There aren't any crop quality defaults in the FeedStorage module, so they have to be passed to a `HarvestedCrop` when it is initialized. I am not sure at the moment how this should be improved - whether crop specific crop qualities should be set in the Crop module or Feed Storage module.


PooyaHekmati:
Will review this shortly, just wanted to react to it and state that I'm happy to see this :)

RUFAS/routines/field/crop/crop_configurations/soybean.py
```diff
@@ -94,6 +101,8 @@ class SoybeanHay(Soybean):
     species: str = "soybean_hay"
     name: str = "soybean hay"
 
+    crop_type: CropType = CropType.FORAGE
```
krb78, line 104: This one feels odd, Ed. Mostly because all the silages and hays are also considered "forages" but you've called them all something more specific. Could we call this one "hay" - the only times I've seen this in practice it has been swept into windrows and then baled.

RUFAS/routines/field/crop/crop_configurations/triticale.py
```diff
@@ -23,6 +25,9 @@ class Triticale(CropData):
     plant_category: PlantCategory = PlantCategory("cool_annual")
     is_nitrogen_fixer: bool = False
 
+    crop_category: CropCategory = CropCategory.SMALL_GRAIN
+    crop_type: CropType = CropType.WHEAT
```
krb78, line 29: Is this supposed to be croptype.triticale?

krb78 (approved):
This looks good, Ed! I just had a few minor edits/questions. A few thoughts I have for the future (or perhaps questions about how this will be used) that are not just directed at you, but things for us to ponder as a group 👀   
1. Could a user plant 2 alfalfa fields, harvest 1 as silage in a bunker and 1 as baleage? If so, how would RuFaS decide what to feed from? (Perhaps a question for @PooyaHekmati and @KevinP-B). 
2. What should our instructions be for farmers (especially the smaller farmers) who still use silos? Should we tell them to just update the packing density and feed out rate, but choose baleage or silage based on DM at time of harvest? 
3. For the grains that are stored "dry" (and I think this question is a feed storage question, not an Ed question, so sorry for asking it here), do we have an option to store these in bins (that have very little shrink) vs commodity bays? Alternatively, what happens for farms that grow this feed and don't store it on their farm they rent space in a big bin (for example) - do we just instruct them to pretend it's being stored on their property in a bin?

RUFAS/routines/field/crop/crop_configurations/soybean.py
Isgotkowitz: The answer to that question will need to involve @PooyaHekmati and @KevinP-B because the `CropType` is an enum that is defined by the Feed Storage module - there isn't a `HAY` crop type currently available. The only two options for Soy right now are `FORAGE` and `GRAIN`, so it seemed that mapping Soybean hay to `FORAGE` was the most natural.

PooyaHekmati:
> 1. Could a user plant 2 alfalfa fields, harvest 1 as silage in a bunker and 1 as baleage?

The Feed Manager does support this, but I'm not sure about this S&C.

>If so, how would RuFaS decide what to feed from? (Perhaps a question for @PooyaHekmati and @KevinP-B).

We have not implemented that logic yet and all the feeds are considered purchased at this time. However, we had the similar question when designing the new feed module. There is a function called `give_feed` which will handle this and it will be able to handle different feeding strategies. We plan to start by implementing a FIFO (First-in-first-out) strategy. We should be able to add new strategies, maybe you want to see how LIFO (last-in-first-out) works. Or any other strategies that are beyond my imagination. I'd like to hear what your thoughts are on this.

> 3. For the grains that are stored "dry" (and I think this question is a feed storage question, not an Ed question, so sorry for asking it here), do we have an option to store these in bins (that have very little shrink) vs commodity bays? Alternatively, what happens for farms that grow this feed and don't store it on their farm they rent space in a big bin (for example) - do we just instruct them to pretend it's being stored on their property in a bin?

I think this is a question for @KevinP-B , but no, we don't have bins. We should consider adding it if it impacts the degradation.

RUFAS/routines/field/crop/crop_configurations/triticale.py
Isgotkowitz: Not for this version of the Feed Storage module - because the only supported Small Grains are `WHEAT`, `RYE`, `OAT`, and `RICE`.

KevinP-B:
The easiest way to model specific use cases for storage (bins, towers) in the short term would be to copy then heuristically modify an existing similar storage solution. For example, bins could be 1/2 the loss of a bunker and uprights are bunkers with 3% effluent loss.

That said, we can add those patch-y  or more robust solutions in the future.

As for off-site storage, there will be energy implications, but the spatial or geographic location of the pile does not factor into the storage modeling aside.

Pooyah, do stored crops have a location property? That might be useful.

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

Isgotkowitz:
> Could a user plant 2 alfalfa fields, harvest 1 as silage in a bunker and 1 as baleage?

@krb78 yes, it is possible.

RUFAS/routines/field/crop/crop_configurations/triticale.py
krb78: check, ok! Makes sense.

RUFAS/routines/field/crop/crop_configurations/soybean.py
krb78: I think that makes sense. but what doesn't make sense is that this seems to be the only one labeled as "forage" so it feels inconsistent with the other feeds. Above on line 24 it says:   crop_type: CropType = CropType.RYE and on line 104 for corn silage it says:  crop_type: CropType = CropType.SILAGE. So I think my real question is and ask to @PooyaHekmati and @KevinP-B is why not be consistent with soybean and that way in the future we could have "straw" along with "hay" and "grain" (if we so chose)?

krb78:
@PooyaHekmati I think FIFO is fairly standard, though farmers honestly do "willy nilly" things and sometimes don't stand by this (even though it's largely the intention). 

regarding the bins vs bunkers - perhaps a V1.X problem would be to add additional storage types. Something that is definitely of interest for users/farmers currently is how controlling shrink (with different storage types) changes environmental impact. I think even some simple updates to enable this would be greatly appreciated... because the truth is controlling shrink and caring about feed storage is something that impacts the bottom line and farmers are more interested in understanding this type of investment on the footprint than on understanding some of the more commonly discussed (feed additive, digester, cover crop) investments because they know controlling shrink and improving feed quality through storage matters for their cows and bottom line too. (just my unnecessary two cents haha)

RUFAS/routines/field/crop/crop_data.py
```diff
@@ -69,6 +83,13 @@ class CropData:
     is_alive: bool = True
     """is the crop currently alive in the field?"""
 
+    crop_category: CropCategory = CropCategory.SMALL_GRAIN
+    """Broad category into which this crop type falls."""
+    crop_type: CropType = CropType.GRAIN
+    """Sub type of this crop."""
+    storage_type: StorageType = StorageType.DRY
+    """The method of storage that will be used for this crop when harvested."""
```
PooyaHekmati, line 91: These docstrings don't look right.

RUFAS/routines/field/crop/crop_management.py
```diff
@@ -212,6 +219,48 @@ def _recalculate_biomass_distribution(self, roots_harvested: bool) -> None:
             self.data.above_ground_biomass = 0.0
             self.data.root_fraction = 1.0
 
+    def _store_harvested_crop(self, time: Time, field_size: float, feed_manager: FeedManager) -> None:
+        """
+        Compiles the details of a harvest of this crop into a HarvestedCrop instance.
```
PooyaHekmati: and passes it to the feed manager.

RUFAS/routines/field/crop/crop_management.py
```diff
@@ -212,6 +219,48 @@ def _recalculate_biomass_distribution(self, roots_harvested: bool) -> None:
...
+            dry_matter_percentage=self.data.dry_matter_percentage,
+            dry_matter_digestibility=DEFAULT_CROP_QUALITIES.get("dry_matter_digestibility"),
+            crude_protein_percent=DEFAULT_CROP_QUALITIES.get("crude_protein_percent"),
+            non_protein_nitrogen=DEFAULT_CROP_QUALITIES.get("non_protein_nitrogen"),
+            starch=DEFAULT_CROP_QUALITIES.get("starch"),
+            adf=DEFAULT_CROP_QUALITIES.get("adf"),
+            ndf=DEFAULT_CROP_QUALITIES.get("ndf"),
+            sugar=DEFAULT_CROP_QUALITIES.get("sugar"),
+            lignin=self.data.lignin_dry_matter_percentage,
+            ash=DEFAULT_CROP_QUALITIES.get("ash"),
```
PooyaHekmati, line 260: Why use defaults? Don't we have the actual values?

RUFAS/routines/field/field/field.py
```diff
@@ -114,6 +116,9 @@ def __init__(self, field_data: Optional[FieldData] = None, soil: Optional[Soil]
         self.manure_manager: ManureManager = manure_manager
         """:class:`ManureManager` instance from which manure is requested for application to the field."""
 
+        self.feed_manager: FeedManager = feed_manager or FeedManager()
+        """:class:`FeedManager` instance which receives harvested crops."""
```
PooyaHekmati, line 144: Didn't we fix the docstrings formatting already?

RUFAS/routines/field/field/field.py
```diff
@@ -39,7 +40,8 @@ def __init__(self, field_data: Optional[FieldData] = None, soil: Optional[Soil]
                  fertilizer_events: Optional[List[FertilizerEvent]] = None,
                  fertilizer_mixes: Optional[Dict[str, Dict[str, float]]] = None,
                  manure_events: Optional[List[ManureEvent]] = None,
-                 manure_manager: Optional[ManureManager] = None):
+                 manure_manager: Optional[ManureManager] = None,
+                 feed_manager: Optional[FeedManager] = None):
```
PooyaHekmati: Why is it optional? What happens if it is None?

RUFAS/routines/field/crop/crop_management.py
krb78: We will be using legitimate values from these crops, but representative values instead of what might actually be harvested in the field. These "defaults" are the same values in the feed library and will be the same ones used when we feed the cows. We decided to not use the actual values because we won't have all those capabilities worked out in RuFaS/SWAT and didn't want to delay launch even more to figure this one out. It'll be a version 1.X item 😉

RUFAS/routines/field/field/field.py
Isgotkowitz: No, issue #371 is still open. It had one PR that made some significant contributions to fixing the docstring formatting in a lot of files, but there are still modules that need to have their docstrings rewritten. Also that PR didn't address the misformatting of class docstrings.

RUFAS/routines/field/field/field.py
Isgotkowitz: If it is None, the field will instantiate its own `FeedManager`.

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: > Why use defaults? Don't we have the actual values?

I thought the FeedStorage module would be handling default quality metrics, which is why I wrote up the table of crop quality metrics to only have N, P, and lignin (see [this](https://docs.google.com/document/d/1auWeT6XnItd8yLoWGkbSwJ2DkUULxv5ClA48Grm69IE/edit#heading=h.h0dz48r2a92t)). Seems like it would be better to have crop quality defaults handled in the Crop module though - I can add those in but I think that should be in a subsequent PR.

RUFAS/routines/field/crop/crop_data.py
Isgotkowitz: Don't look right as in they are misformatted, or they are providing incorrect information, or they are redundant, or something else?

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: Good catch

RUFAS/routines/field/field/field.py
Isgotkowitz: Might be a good fixathon task

RUFAS/routines/field/field/field.py
PooyaHekmati: Would you like to document it?

RUFAS/routines/field/crop/crop_management.py
PooyaHekmati: @Isgotkowitz did you share and discuss that design doc with me?
@krb78 It is ok to postpone things like this as long as we document them somewhere (we usually make issues about them and mention the issue number as TODO in the code).

RUFAS/routines/field/crop/crop_data.py
PooyaHekmati: misformatted

Isgotkowitz: Would it be acceptable to leave fixing them for the fixathon, as part of issue #371?

RUFAS/routines/field/field/field.py
Isgotkowitz: Yes - I am not sure if this is the best approach to handling a Field instantiated without a FeedManager though. Is this something we could reasonably expect to happen in RuFaS? If not, then wouldn't it be better to raise an error and fail initialization if no FeedManager is passed?

RUFAS/routines/field/crop/crop_data.py
PooyaHekmati: yes

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: Shared with you - yes. It is in the "Ed" Google Drive folder that you shared with me. Discussed - also yes, but that document went through several iterations, so I am not sure that we discussed it in its current form.

I am not sure when/where we decided how the default crop quality metrics would be handled for V1, but I didn't clearly document it - that is my bad, I apologize.

RUFAS/routines/field/field/field.py
PooyaHekmati: I think raising an error is a better approach.

Isgotkowitz: I will do this and also apply the same logic to the `ManureManager`

Update: changing this code is important and is not particularly hard to do. The issue lies in the unit tests - there are a lot of initializations that would need to change (and I mean a lot - it would significantly increase the size and scope of this PR). Would an acceptable alternative be to raise an error to OM when a Field attempts to initialize without a FeedManager, and we can revisit the initialization routine for Field (and other modules in C&S which have the similar designs) during the Fixathon?

RUFAS/routines/field/crop/crop_management.py
krb78: Ah this makes sense! Document, document, document 🤓 (I don't know if you need this information for your documentation but...) We started discussing the default crop quality metrics with Kevin, Pooya, Kristan, Ed, and myself one evening (early/mid-December) and then the following Monday we worked with Eric and the other C&S SME's to double check that using the NASEM/NRC values would be acceptable as defaults.

RUFAS/routines/field/field/field.py
PooyaHekmati: Let's:
1. make an issue about this, we may or may not do it during the fixathon
2. log an error to OM when it happens.

RUFAS/routines/field/crop/crop_management.py
PooyaHekmati: Yes, I remember that. But my understanding is different from:
>I thought the FeedStorage module would be handling default quality metrics

As far as I remember, S&C is supposed to generate the crop objects, either calculating the quality metric or setting them to defaults. Feed manager takes it from there and does the degradation over time.

PooyaHekmati (approved):

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: Ok, S&C will handle the quality defaults - but first I want to confirm with SMEs (@KevinP-B @krb78) that it will be ok to use the values from the feed library table (where we got lignin, N and P)? I don't understand how those are recorded/calculated, but they may be problematic to use if they are numbers from feeds _coming out of_ storage instead of _going into_ storage (i.e. pre-degradation) - or if they are post-storage values, it will still be ok to use them cause we are in a crunch right now.

RUFAS/routines/field/field/field.py
Isgotkowitz: Written up in #1104.

KevinP-B:
Without a longer form discussion, I can say that the feed tables a pretty comprehensive and should include values we can use as default for pre and post storage, or even early harvest. That said, for feeds coming out of storage, you're correct that the length of storage is not necessarily or transparently accounted for. In those cases the butrient of interest is important. Lignin and phosphorous will not change meaningfully (for our purposes) during storage, for example.

Nitrogen is another story, though, that may deserve a phone chat.

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

krb78:
1.  I can't figure out how to respond in GitHub to Kevin's messages when he responds via email (yes, I was born in the 20th century hahahaha).
  2.  While Kevin is 100% correct, I think for the sake of RuFaS it doesn't matter in V1. Because we aren't actually feeding any of the feeds that are coming out of storage. Yes, the inventory is being tracked for feeding, but we are not using any of the post storage quality metrics in the feeding part of the animal module. In other words, RuFaS will be weirdly disjointed for V1, but I think it's ok. We will have corn silage with specific quality metrics entering storage --> storage will result in some changes (leaching, shrink, etc.) these changes and losses will be recorded in the output manager (I assume) --> when we feed inventory of these feeds will decrease in a FIFO manner but the quality metrics of these feeds will be determined by NASEM/NRC (so will at this point be the same as what went into storage). But being we don't have the ability to feed the post-storage feeds right now, I don't think it matters if the pre-storage and post-storage quality metrics are the same.

Kaitlyn Briggs, DVM, MBA
C: 607.386.0040

If you receive an email from me outside of your own working hours, please do not feel pressure to respond outside of your own work schedule.

PR #1102, opened 2024-01-11
Improves error messaging of invalid graph filter file keys

### LINKED ISSUE #927: Improve error messaging when invalid graph filters passed

ew3361zh:
PR to address issue #927 where invalid keys in a graph filter file were not adequately documented to OM's errors logs.

This PR:
1. Creates a new method in GG that checks the keys from the graph filter file first against a list of required graph filter keys (the 3 necessary for graphing, without which matplotlib can't graph anything regardless) and second against the optional graph filter keys.
2. For each invalid required key found, it creates an error message that is returned back to OM for logging.
3. For each invalid optional key found, it creates a warning message that is returned to OM for logging but still allows graphing to continue because the graph can still be made, it just may not have the legend or title or other optional feature because the key was improperly added to the filter file.
4. Updates unit testing.

Notes:
1. I don't know how volatile the valid_graph_filter_keys list will be. I don't imagine very but if there is a better solution to maintaining a list that people feel like may need to be regularly updated to validate against, please leave a comment.
2. The errors found by the `_validate_graph_filter` function aren't returned to OM for logging until after the data in the filtered graph data pool has been prepared by GG. I did this to combine all the potential logs from validating and preparing the data to be graphed before returning them to OM for logging so the user will have a fuller picture of all the issues they need to address to get their graph filter to work as expected.


RUFAS/graph_generator.py
```diff
@@ -160,9 +162,41 @@ def generate_graph(
...
+        Parameters
+        ----------
+        graph_details : Dict[str, str  |  List[str]]
+            A dictionary containing details/metadata about the graph.
+
+        Returns
+        -------
+        List[Dict[str, str] | Dict[str, Any]]
+            The logs, warnings, and errors to be reported to OutputManager.
+        """
```
```suggestion
        """
        Ensures all the filter keys are valid and if not, raise an error and report them back to Output Manager.

        Parameters
        ----------
        graph_details : Dict[str, str | List[str]]
            A dictionary containing details/metadata about the graph.

        Returns
        -------
        List[Dict[str, str] | Dict[str, Any]]
            The logs, warnings, and errors to be reported to OutputManager.
        
        """
```

RUFAS/graph_generator.py
```diff
@@ -160,9 +162,41 @@ def generate_graph(
...
+        valid_graph_filter_keys = ["type", "filters", "variables", "title", "legend", "align_labels", "aspect",
+                                   "canvas", "constrained_layout", "dpi", "edgecolor", "facecolor", "figheight",
+                                   "figsize", "figwidth", "frameon", "grid", "snap", "subplot_adjust", "tight_layout",
+                                   "title", "transform", "xlabel", "xticklabels", "xticks", "xlim", "ylabel",
+                                   "yticklabels", "yticks", "ylim", "yscale", "xscale", "zorder"]
```
Isgotkowitz: It looks like these keys come from a number of different sources, including `FIGURE_SETTERS`, `AXES_SETTERS`, and the structure of graphs that we have defined. Would it be possible to do something like:

```python
valid_graph_filter_keys = ["type", "filters", "variables", "title", "legend"] + FIGURE_SETTERS.keys() + AXES_SETTERS.keys()
```

This would make it a little more resilient to changes in which options are available.

RUFAS/graph_generator.py
```diff
@@ -160,9 +162,41 @@ def generate_graph(
...
+            "class": self.__class__.__name__,
+            "function": self._validate_graph_filter.__name__,
+        }
+        for filter_key in graph_details.keys():
+            if filter_key not in valid_graph_filter_keys:
+                graph_filter_validation_logs.append({"error": f"Can't plot {graph_details.get('title')} data set",
+                                                     "message": f"Invalid filter file key '{filter_key}'. "
+                                                     f"Please see wiki for full list of valid filter keys.",
+                                                     "info_map": info_map})
+        return graph_filter_validation_logs
```
Isgotkowitz, line 206: In addition to checking that all present keys are valid, this method should also check that the keys which are strictly necessary for producing a graph are present. I think this is only type and filters, and filters already has a pretty good error message raised when it is not present.

RUFAS/graph_generator.py
```diff
@@ -160,9 +162,41 @@ def generate_graph(
...
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self._validate_graph_filter.__name__,
+        }
+        for filter_key in graph_details.keys():
+            if filter_key not in valid_graph_filter_keys:
+                graph_filter_validation_logs.append({"error": f"Can't plot {graph_details.get('title')} data set",
+                                                     "message": f"Invalid filter file key '{filter_key}'. "
+                                                     f"Please see wiki for full list of valid filter keys.",
+                                                     "info_map": info_map})
```
Isgotkowitz: I don't know that pointing people towards the Wiki is the best thing to put in this message, because it means the wiki needs to stay updated the list of options present in at least axes setters and figure setters. That may be setting us up for more wiki maintenance than what we would like or would be reasonable.

RUFAS/graph_generator.py
```diff
@@ -141,9 +141,11 @@ def generate_graph(
...
         """
         try:
+            graph_filter_validation_logs = self._validate_graph_filter(graph_details)
             prepared_data, log_pool = self._prepare_plot_data(filtered_pool, graph_details)
+            all_logs = log_pool + graph_filter_validation_logs
 
-            found_errors = any("error" in log for log in log_pool)
+            found_errors = any("error" in log for log in all_logs)
             if found_errors:
                 return log_pool
```
```suggestion
                return all_logs
```
Function needs to return `all_logs` at the bottom as well.

Isgotkowitz (changes requested):
Requesting changes on two things: graph generator still does not log a helpful error message when the `"type"` key is missing, and the wrong log pool is being returned in `generate_graph()`.

RUFAS/graph_generator.py
ew3361zh: I think that could work. I wondered about dividing these into 2 categories - required ("type", "filters", and "variables") and optional (everything else). 

For the required fields, we add an error and specify the field that's missing.
For the optional fields, we add a warning and tell them which fields they have in the file that don't match any of the optional fields but I think their graphs would still be able to plot so maybe no need to halt that process? (note: I tested this with "title" -> "titled" and the graph still plotted, just with out a title).

In this way we could be more specific about the fields they absolutely must have.

RUFAS/graph_generator.py
ew3361zh: I guess you had the same idea I just typed above. I'm pretty sure "variables" should be required as well since if the filter is in a dictionary format, it has to have some variables against which to check or the graph won't be able to be plotted.

RUFAS/graph_generator.py
ew3361zh: My thought was that the list wouldn't change that much - I'm not sure how often matplotlib adds new keys to the FIGURE_SETTERS and AXIS_SETTERS. I'm also not sure how else people would easily check their filters list. Do you have an alternate idea for where to point them?

RUFAS/graph_generator.py
ew3361zh: Right. I thought I'd made that change. Thank you.

I may change this at the end of the function as well in case we end up going with logging warnings for invalid "optional" filters.

RUFAS/graph_generator.py
Isgotkowitz: That sounds good - especially not stopping the report from being generated if it has a non-critical mistake

RUFAS/graph_generator.py
Isgotkowitz: Are those lists exhaustive of matplotlib though? I figured they were a limited set of commands of the more useful/common matplotlib functions, but I don't actually know.

> Do you have an alternate idea for where to point them?

Maybe providing the path to the Graph Generator file? I don't know how acceptable that would be though.

ew3361zh: My current thought is the list of filters (which is in the GG wiki currently) isn't going to change often and if there is an update to what filters the GG can accommodate, it will be easy enough to update the wiki at that time.

RUFAS/graph_generator.py
```diff
@@ -156,13 +158,56 @@ def generate_graph(
...
+            A dictionary containing details/metadata about the graph.
+
+        Returns
+        -------
+        List[Dict[str, str] | Dict[str, Any]]
+            The logs, warnings, and errors to be reported to OutputManager.
+
+        """
+        required_graph_filter_keys = ["type", "filters", "variables"]
+        optional_graph_filter_keys = list(FIGURE_SETTERS.keys()) + list(AXES_SETTERS.keys())
```
```suggestion
        required_graph_filter_keys = ["type", "filters"]
        optional_graph_filter_keys = list(FIGURE_SETTERS.keys()) + list(AXES_SETTERS.keys() + ["variables"])
```
As I understand it, "variables" is only needed when one of the variables filtered out by the filter is a dictionary and you want to pull values out of that dictionary.

Isgotkowitz (approved):
I think this looks good Niko, nice job. Only comment is on one key that is marked as required but I think is actually optional.

RUFAS/graph_generator.py
ew3361zh: Right. That makes sense and we do have a check for those variables being present when the filter is a dictionary (and that would be logged to the OM). I can make this change. Thanks, Ed!

RUFAS/graph_generator.py
```diff
@@ -156,10 +158,53 @@ def generate_graph(
                 graph_details, filter_file_name, graphics_dir
             )
 
-            return log_pool
+            return all_logs
         except Exception as e:
             raise e
```
ltp23, line 164: I don't remember the exact syntax, but we can just simply do this
```
except:
   raise
```

ltp23 (approved):
It looks good to me. Thanks Niko.

RUFAS/graph_generator.py
ew3361zh: Thanks, Loi. It looks like making that change gives me a flake8 violation (E722) so I will leave it as is for now.

PR #1107, opened 2024-01-16
Removed snow melt warnings

### LINKED ISSUE #1004: Snow melt warnings should be removed
## Current Behavior
When the `Snow` submodule in the Crop and Soil module calculates the snow melt on the current day to be greater than the amount of snow present in the soil profile on the current day, it raises a warning to the OutputManager.

This behavior is problematic because it is unnecessary to raise warnings (or even add logs) when snow melt is calculated to be greater than the amount of snow present.

## Desired Behavior

Don't raise warnings when snow melt on a given day is calculated to be greater than what is physically possible. Just setting the upper bound on actual snow melt to be equal to the current amount of snow on the ground is sufficient (and is already being done).

## Solution

Remove the warnings raised to the OutputManager.

## Additional Context
To see the equation(s) governing snow melt, check SWAT Theoretical documentation section 1:2.5. It states that "Snow melt is controlled by the air and snow pack temperature, the melting rate, and the areal coverage of snow." It does not take into account how much snow is actually on the ground, which is why calculating the snow melt to be greater than what could possibly melt on any given day is not unexpected or unreasonable behavior.

0luchi:
Issue #1004


Isgotkowitz (approved):
LGTM. In the future, please add a description about what the PR is doing/how it is doing it. Because the substantive changes of this PR are so small, a single sentence in this case would do the trick. Also in PRs like this (high # of lines changed but small/minor substantive changes) I think it is good to let people know that most lines changed are just formatting - this helps people review quicker.

Nice job Oluchi

ew3361zh (approved):
LGTM, I would second Ed's comments though - PRs should normally have a description that explains the What - Why - How of addressing the issue.

PR #1110, opened 2024-01-17
Adding methane generation volume to anaerobic digester daily output

### LINKED ISSUE #1077: Adding methane volume to the output for anaerobic digestion manure treatment option
- [x] Add methane generation volume to anaerobic digestion
- [x] Update accumulated output for digester and lagoon option

ltp23:
This is a pretty simple PR.

Just adding a missing variable to the output of one of the manure treatment options - an anaerobic digester and digester/lagoon combo.

And updating the related unit tests.


Isgotkowitz (approved):
LGTM

RUFAS/routines/manure/manure_treatments/anaerobic_digestion_and_lagoon.py
```diff
@@ -91,4 +91,6 @@ def _daily_update_helper(self) -> ManureTreatmentDailyOutput:
             sim_day=self._sim_day,
         )
 
+        self._accumulate_daily_output(self.anaerobic_digestion_daily_output)
+        self._accumulate_daily_output(anaerobic_lagoon_daily_output)
```
tomhuhh, line 95: This makes a lot of sense to add the daily output from both anaerobic digestion and anaerobic lagoon to the accumulated outcome. 

Maybe this is obvious but I am just wondering why is 'anaerobic_digestion_daily_output' an attribute of self but 'anaerobic_lagoon_daily_output' is not. 

Thanks!

tomhuhh (commented):
Thanks Loi for your good work! 

(1) The code runs without difficulty.  
(2) I left a comment in the code regarding the digestion and lagoon daily outputs. 
(3) I am not an expert in this area, but I would say looking at the results when setting the manure treatment method as 'anaerobic digestion', the lactating cow pen has the highest methane generation volume which makes sense to me. 
(4) Should 'methane_generation_volume' also be an output from the 'anaerobic digestion and lagoon'? Is this addressed in a different PR? I understand this is outside the initial scope of this PR. I was just a bit confused. 

Thanks!

RUFAS/routines/manure/manure_treatments/anaerobic_digestion_and_lagoon.py
ltp23: Thanks Haowen, no problem. This is because in this case, the output data from the digester is an intermediate step so I had to save it like that. The output data from the lagoon as the last step will always be accessible.

tomhuhh: Thanks Loi!

tomhuhh (approved):

tomhuhh:
Hello Loi! Could you please help me understand point (4) in my comments? When I set the manure management scenario to be 'anaerobic digestion and lagoon', the 'methane_generation_volume' output is 0 regardless of animal category. 

>  (4) Should 'methane_generation_volume' also be an output from the 'anaerobic digestion and lagoon'? Is this addressed in a different PR? I understand this is outside the initial scope of this PR. I was just a bit confused.

Thanks!
Haowen

ltp23:
Thanks Haowen. for the question.

> (4) Should 'methane_generation_volume' also be an output from the 'anaerobic digestion and lagoon'? Is this addressed in a different PR? I understand this is outside the initial scope of this PR. I was just a bit confused.

I added a csv filter file as an example and set the lactating cows to be using scenario 6. Please let me know if it works for you.

tomhuhh:
Hello @ltp23! That makes sense. I think maybe I just did not use the correct filter? 

With the filter you included, I was able to get the methane volume outputs from anaerobic digestion. Is `AnaerobicDigestion_ManureTreatmentDailyOutput_Pen_3.methane_generation_volume` the same as `Anaerobic digestion and lagoon`? 

The filter that I used, however, did not work on `Anaerobic digestion and lagoon`. It did work on `Anaerobic digestion`. Please see below. 

`ManureTreatmentDailyOutput_Pen_0.methane_generation_volume
ManureTreatmentDailyOutput_Pen_1.methane_generation_volume
ManureTreatmentDailyOutput_Pen_2.methane_generation_volume
ManureTreatmentDailyOutput_Pen_3.methane_generation_volume`

Thanks!

ltp23:
> With the filter you included, I was able to get the methane volume outputs from anaerobic digestion. Is `AnaerobicDigestion_ManureTreatmentDailyOutput_Pen_3.methane_generation_volume` the same as `Anaerobic digestion and lagoon`?
> 
> The filter that I used, however, did not work on `Anaerobic digestion and lagoon`. It did work on `Anaerobic digestion`. Please see below.
> 
> `ManureTreatmentDailyOutput_Pen_0.methane_generation_volume ManureTreatmentDailyOutput_Pen_1.methane_generation_volume ManureTreatmentDailyOutput_Pen_2.methane_generation_volume ManureTreatmentDailyOutput_Pen_3.methane_generation_volume`

Thanks Haowen @tomhuhh ,
I updated the available options in the metadata for the manure treatments. So, right now, there is no option for `anaerobic digestion` by itself, and it has to be coupled with the lagoon treatment, hence the name `anaerobic digestion and lagoon` or `anaerobic digestion and lagoon with split`.

tomhuhh:
Thanks for the clarification! @ltp23

Issue #619, opened 2023-07-01
Implement a reasonable default diet 


tomhuhh:
## Expected Behavior
The crude protein (CP) composition is within reasonable range for a lactating diet cow. 

## Current Behavior
Low CP% (~10%) in the current diet. Also, a low DM% at 37% is observed. 

## Possible Solution
Add some limits to the feed ingredients to mimic a diet that Edward formulated. 

## Steps to Reproduce

1. Formulate a diet using the default setting and calculate the nutrient composition. 
2.
3.
4.

## Context (Environment)

When testing MONENSIN at a dose of 24 mg/kg DMI, we expected decrease in enteric methane emissions. However, methane emission increased due to low CP% (~10%).

Ed731551:
Equation for estimating reduction of CH4 yield emissions should work within the range of 15 to 19% CP. In case of diets outside this, I would suggest using the basal equation with only monensin addition as a predictor (recommended doses).

Equation as follows (Abstract - Modelers conference, 2023): 
% Reduction in CH4 yield emissions = 0.03223 - 0.00410 × Monensin

NOTE = Monensin equations (% of reduction) were developed from data of lactating cows only. Caution should be exercised when aiming to implement these to other animal groups. 

Nevertheless, a diet with 10% CP for a production level of 44 kg of milk per day (as far as I remember) does not make any sense to me. Worth checking details on how that diet was formulated.

tomhuhh:
Closing it as the issue is resolved by PR#1112.

PR #1112, opened 2024-01-17
Update default diet

tomhuhh:
This PR is meant to address the issue mentioned in issue #619. 

We created a new default diet for all animal categories based on the FarmES NorthEast regional diets with the following adjustments.
(1) Replaced the mineral mixes (301) with minerals 231 and 234.
(2) Replaced the byproduct mixes with almond hulls, soybean meals, and corn grains with a ratio of 35%, 31% and 34% based on their ratios in the close-up cow diet from Edward's NE Diet 15.
(3) We further slightly adjusted the ratios of the three byproducts to allow for higher CP content in the diets.
be965bae1

Below is a comparison among the diet candidates in terms of crude protein (CP) and dry matter (DM): 

CP:  
Screenshot 2024-01-17 at 3 39 59 PM

DM:
Screenshot 2024-01-17 at 3 40 33 PM

Updated on 1/18/24 with more description for why we want to do this: 
(1) The current default diet yields a CP% of around 10% for the lactating cow pen, which is lower than desired.
(2) Given the limitations in ingredients for pilot testing FarmES diets, many aren't suitable for a broader audience. 
(3) I am uncertain if there is any constraint in directly using FarmES diets for RuFaS's audience. 
(4) Edward's regional diets do not include diets for heifers.


allisterakun (approved):
Looks all good to me!

Ed731551:
@tomhuhh  diets make sense to me. I think you can proceed.

Ed731551 (approved):

Issue #1076, opened 2023-12-22
Investigate animal class improvements needed for report_animal_sold_information (calves, cows sold)


jadamchick:
## Expected Behavior

The ideal would be for `AnimalModuleReporter.report_sold_animal_information_sort_by_sell_day`and
`AnimalModuleReporter.report_sold_animal_information` to include calves sold, and to distinguish between cows sold and cows who died (even better: to not include cows who died at all in this report). 

## Current Behavior

Calves are not included. 
Cows reported are sold_and_died_cows, which lumps together all cows who left the herd through either path. 

## Steps to Reproduce

Report filter: 
```
   {
            "name": "Animals sold in the whole simulation",
            "filters": [
                "AnimalModuleReporter.report_sold_animal_information.(animal_id|animal_type|body_weight|sold_day)"
            ]
        }
```
2.
3.
4.

## Context (Environment)

The purpose for this report is to be able to allocate greenhouse gas emissions to milk vs. beef production, which requires knowing the weights and ages (eg calf, heiferII, heiferIII, cow) of all animals sold. 
Right now we can get at this information for calves and cows through: 
- calves: `"LifeCycleManager.daily_update.sold_calf_num"` and average calf birthweight (an input)
- cows: use this (sold_animal_information) report to know the average weight of cows sold, and subtract cow deaths (`LifeCycleManager.daily_update.cull_reason_stats.culled for death`) from total herd exits (died + sold) (`LifeCycleManager.daily_update.cow_herd_exit_num`). 

It would be simpler, smoother, and less prone to errors if those were included in this report.

allisterakun:
Please dig into the animal module to investigate how to report sold calves and how to distinguish between sold and died cows. 
If it is easy, please implement the solution; otherwise, make a separate issue.

PR #1114, opened 2024-01-18
Adding cows and calves sold counts to OM

ltp23:
Issue #1076 

Adding to the OM only sold cows and excluding cows that died. To clarify, culled cows for any reason other than because of death are sold.

Also adding calf sold count to the OM.

I will remove example report json files for this PR before merging.


ltp23:
This PR was included in the the PR #1121.

PR #1115, opened 2024-01-18
added num_cows_total to output manager

JoeWaddell:
Per request in Fixathon task 76, added `num_cows_total` to AnimalManager.daily_updates, in the `report_daily_animal_population` method in `AnimalModuleReporter`.


krb78 (approved):

jadamchick (approved):
Looks great! Gives the same values as when I use report generator to combine milking cows + dry cows. Can use it directly in report generator filters to save a step (and reduce number of outputs needed in the report).

PR #1122, opened 2024-01-18
Adding KeyError to ration_driver

JoeWaddell:
Draft PR to tackle Fixathon task 89. Adding a proper error message to missing Feed IDs in `ration_driver.py`


JoeWaddell:
The merge conflict comes from the default_feed file, which includes a temporary change to elicit the error. This should be reverted prior to merge. Similarly, error message should be edited when Feed module and other refactors are in place, as it describes current behavior.

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -1022,6 +1022,14 @@ def get_feed_data_from_feed_ids(self, feed_ids: Set[int]):
                 continue
             for feed_id in sorted(list(feed_ids)):
                 # Get the list index of the feed_id in self.feed_id list.
-                idx = self._feed_id_to_list_idx_dict[int(feed_id)]  # missing 54, 136, 26, 118, 139
-                result[key].append(vals[idx])
+                try:
+                    idx = self._feed_id_to_list_idx_dict[int(feed_id)]
+                    result[key].append(vals[idx])
```
Isgotkowitz, line 1027: There are three spots in this try block where a key error could occur. Is it possible to know which one is the most likely or expected place for a KeyError to occur? If not, would it be helpful to catch the different key errors and raise more specific error messages about them?

Isgotkowitz (approved):
Looks good Joe. One comment about which Key Error is being raised specifically.

Also, should there be a test case for this in a unit test somewhere?

RUFAS/routines/animal/ration/ration_driver.py
JoeWaddell: Excellent questions. We're anticipating some Feed-related changes as the Storage module comes online, so I think it best that I open an issue to solve when that is in-progress. I'll circle back to this because it might be worth doing now, if those changes are further off than expected.

JoeWaddell:
> Also, should there be a test case for this in a unit test somewhere?

Great point: I'll add that to this PR and re-request your review.

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -1022,6 +1022,16 @@ def get_feed_data_from_feed_ids(self, feed_ids: Set[int]):
...
+                except KeyError:
+                    info_map = {"class": self.__class__.__name__,
+                                "function": self.get_feed_data_from_feed_ids.__name__,
+                                }
+                    om.add_error("KeyError", f"Key {feed_id} not found in AvailableFeeds. "
+                                 f"Check that price was set in purchased_feeds in input feed json, "
+                                 f"and that it is included in the specified NRC or NASEM csv.",
```
ew3361zh, line 1034: Is it possible to know here whether this is conflicting with the NRC or NASEM csv and to report that specifically?

ew3361zh (changes requested):
I agree with Ed that I think a testing should be added for this error being raised. After that, as long as you're happy with the specificity level of the error message, I think it's good to go.

RUFAS/routines/animal/ration/ration_driver.py
JoeWaddell: It depends on which the user picked in their config file, since they can only use one at a time. Input Manager validation should handle this in the future.

Isgotkowitz (approved):
Nice

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -1022,6 +1022,16 @@ def get_feed_data_from_feed_ids(self, feed_ids: Set[int]):
...
-                idx = self._feed_id_to_list_idx_dict[int(feed_id)]  # missing 54, 136, 26, 118, 139
-                result[key].append(vals[idx])
+                try:
+                    idx = self._feed_id_to_list_idx_dict[int(feed_id)]
+                    result[key].append(vals[idx])
+                except KeyError:
+                    info_map = {"class": self.__class__.__name__,
+                                "function": self.get_feed_data_from_feed_ids.__name__,
+                                }
+                    om.add_error("KeyError", f"Key {feed_id} not found in AvailableFeeds. "
```
ew3361zh:
```suggestion
                    om.add_error("KeyError", f"Feed ID {feed_id} not found in AvailableFeeds. "
```

ew3361zh (approved):
LGTM! Just one suggestion which to make clearer what the key that isn't found actually is.

PR #1133, opened 2024-01-19
Fix Negative Manure Excretions 

tomhuhh:
This PR is to address the issue mentioned in #609

As a solution, we shuffled the order of manure excretion and ration formulation in the 'daily_updates' method. This way, the code should formulate the ration first and then calculate the manure excretions.


tomhuhh:
Consider using the filters below to check the manure and dry matter intake outputs on the pen level. 

`pen.calc_total_manure`
`AnimalManager._calc_ration_at_interval`

tomhuhh:
Consider using the code by Loi to check if there is any negative manure outputs on individual animal basis. 

```
for key, value in manure_excretion_values.items():
    if value < 0:
        print(key, value)  # Place a breakpoint here
```

RUFAS/routines/animal/animal_manager.py
```diff
@@ -1628,13 +1628,6 @@ def daily_updates(self, feed: Feed, weather: Weather, time: Time):
                 self._record_animal_events(self.cows)
                 self._record_heiferIIs_conception_rate()
 
-            manure_excretions_output_data = {}
-            for pen in self.all_pens:
-                self.collect_manure_excretions_output_data(pen, feed, manure_excretions_output_data)
-            self.calc_p_rqmts()
```
JoeWaddell, line 1634: I think calc_p_rqmts and daily_p_update need to remain in the previous order, since _update_phosphorus_concentrations comes after this block. With that change, I think it should be OK, assuming the manure excretion outputs match their previous values (do you have output filters that could be useful for diagnostics?)

tomhuhh: Thanks Joe! I've updated the location of the phosphorus calculations. That makes sense. 

You can find the filters below: (I used them in a csv file)

```
pen.calc_total_manure
AnimalManager._calc_ration_at_interval
```

JoeWaddell (approved):
LGTM: outputs are the same for the runs I did, with the exception of the (formerly erroneous, now believable) first day! Good improvement

ew3361zh:
So I ran a simulation with the provided filter on main and then on this branch and I'm getting a bunch of differences between the results for calc_total_manure by pen across the simulation for the following fields:
pen 0 daily inorganic phosphorus fraction
pen 0 daily organic phosphorus fraction
pen 0 daily phosphorus
pen 0 daily phosphorus fraction
pen 1 daily urea
pen 1 daily total ammoniacal nitrogen concentration
...
and actually too many to name here.

Not sure what a good next step would be.

tomhuhh:
Thanks @ew3361zh!  

I think we expect to see differences on the herd level in these outputs because previously there are negatives values in these outputs on the individual animal level. What kind of trend (increase or decrease?) did you notice?

JoeWaddell:
@ew3361zh I checked those same fields on my end but only found the first day as being different: were you spot-checking the CSVs generated, or are you using graphics/report generation?

ew3361zh:
@JoeWaddell @tomhuhh I ran it with CSV outputs and then put the two CSVs in a CSV comparison web app and it highlighted multiple fields throughout (many on day 1 but also a bunch of the same fields at different intervals throughout).

My next step was to try and make some graphs but it seems like it's different fields across all the pens so it might take a little work.

JoeWaddell:
@ew3361zh if it was always the same interval difference, then it should make sense: since the days ration is formulated (default 30d) it'll diverge (manure calculated after ration reformulated, now). If they're other intervals though, or intermittent, we've got a more complex problem!

ew3361zh:
@JoeWaddell @tomhuhh ok, now that I've graphed it I'm not finding the same differences. I will try a few others but I think we can hopefully just write it off to a weird CSV web app mistake or a Niko human error mistake.

I'm still seeing it with the csvs. To recreate, use Haowen's filters to create some filtered CSVs of the data we're looking at first in main and then in the negative-manure-fix branch and dump them side by side in one of the CSV comparison tools and let me know what you think: https://www.textcompare.org/csv/compare/ or https://extendsclass.com/csv-diff.html for example.

ew3361zh:
Screenshot 2024-01-22 at 9 13 54 AM

After running a couple more simulations, it seems like the main culprits all have to do with phosphorus and phosphorus fractions. The graph on the left is from main and the graph and the right is from this branch.

ew3361zh:
Screenshot 2024-01-22 at 9 14 52 AM
Screenshot 2024-01-22 at 9 15 04 AM
Screenshot 2024-01-22 at 9 16 23 AM
Screenshot 2024-01-22 at 9 16 32 AM

ltp23 (approved):
I think it makes sense to collect manure excretions data at the end of the day instead of at the start of the day (which would get the data from the previous day).

Unrelatedly, I noticed that this column remains negative on the main branch and this branch `AnimalManager._calc_ration_at_interval.avg_rqmts_pen_3_LAC_COW.NEpreg_requirement`. But other negative values on the first day now are gone.

Thanks Haowen!

tomhuhh:
Thank you @ew3361zh @JoeWaddell @ltp23 for taking your time and looking into this issue. After some investigation in the code, here is what I found- 

1.  There is discrepancy between the phosphorus outputs between main branch and the current branch. This makes sense because:  
(1) We changed the order of diet formulation vs. manure excretions; diet has an impact on the manure excretions - 'p_intake' is used in the calculations of fecal phosphorus, 'p_feces_excrt'. 
(2) The discrepancy is small ~2% on the pen level. 

2. Nice catch on the negative `NEpreg_requirement` value. I updated the equation to calculate the uterine weight because a component is missing.

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -717,7 +717,7 @@ def calculate_NASEM_energy_maintenance_requirements(
             )
             if days_in_milk is None:
                 days_in_milk = 0
-            uterine_weight = ((calf_birth_weight * 0.2288) * math.exp(-0.2 * days_in_milk)) + 0.204
+            uterine_weight = ((calf_birth_weight * 0.2288 - 0.204) * math.exp(-0.2 * days_in_milk)) + 0.204
```
ew3361zh, line 731: What are these numbers? Are they constants? I would recommend either adding these as constants or specifying what they represent in the code so what's happening in this line is clearer (not sure if it necessarily needs to be done in this PR or if a separate issue should be made - you can decide @tomhuhh).

magic_number_1 = 0.2288 (or constant_1 if it's something we'd use elsewhere in the codebase)
magic_number_2 = 0.204
magic_exp_1 = -0.2
e.g. uterine_weight = ((calf_birth_weight * magic_number_1 - magic_number_2_ * math.exp(magic_exp_1 * days_in_milk)) + magic_number_2

KFosterReed:
Hey Team! Using a few minutes to read through this conversation (I know it was merged) but wanted to highlight that I believe there should be negative NEpreg values related to uterine involution. Mostly a comment for @tomhuhh

Issue #1135, opened 2024-01-19
Accessing elements from available_feeds using keys that might not exist in the dictionary


tomhuhh:
It appears that the issue is related to accessing elements from `available_feeds` using keys that might not exist in the dictionary. Specifically, the key-value pair (202, 1) is causing an issue because '202' is not present in available_feeds. 

## Expected Behavior

The code should check if the key is present in available_feeds before attempting to access its values. 

## Current Behavior

The code directly accesses `available_feeds[key][nutr]`. 

## Possible Solution

To resolve this, we can add a conditional check if key in available_feeds: before accessing available_feeds[key].

## Steps to Reproduce

1. Update the calf's diet to only contain 216 and change its inclusion to 100%. 
2. Run the simulation and you should see the error. 
![Screenshot 2024-01-19 at 3 09 03 PM](https://github.com/RuminantFarmSystems/MASM/assets/52800053/b786b8b2-b734-4628-af38-aa196b47f183)
3.
4.

tomhuhh:
Closing this issue as it can be addressed by issue #1137

PR #1136, opened 2024-01-19
Add a conditional check if key in available_feeds: before accessing available_feeds[key].

tomhuhh:
Added a conditional check if key in available_feeds: before accessing available_feeds[key]. This is to address issue #1135.

### DIFF
--- a/RUFAS/routines/animal/ration/ration_driver.py
+++ b/RUFAS/routines/animal/ration/ration_driver.py
@@ -368,7 +368,6 @@ class RationReporter:
 
         for key, val in ration.items():
             nutrient_amount["dm"] += val
-            
             if key in available_feeds:
                 for nutr in nutrients:
                     # all values on a 100% dry matter basis
--- a/RUFAS/routines/animal/ration/ration_driver.py
+++ b/RUFAS/routines/animal/ration/ration_driver.py
@@ -368,28 +368,30 @@ class RationReporter:
 
         for key, val in ration.items():
             nutrient_amount["dm"] += val
-            for nutr in nutrients:
-                # all values on a 100% dry matter basis
-                if nutr == "DM":
-                    if available_feeds[key][nutr]:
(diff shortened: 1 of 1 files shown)


RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -368,28 +368,29 @@ def report_ration(cls, ration: Dict, available_feeds: Dict) -> Tuple[Dict, Dict]
...
-                        denom = 6.38
-                    # [A.2.A.1]
-                    else:
-                        denom = 6.25
-                    nutrient_amount[nutr] += (available_feeds[key]["CP"] / (denom * 100)) * val
-                else:
-                    if nutr == 'DE':
-                        if available_feeds[key]['DE'] != -1:
-                            nutrient_amount[nutr] += val * (available_feeds[key]['DE'] / 100)
+            if key in available_feeds:
```
ew3361zh, line 371: What if the key is not in the available_feeds? Do you just want it to be ignored silently? Or would it be helpful to add an else statement which adds a warning to OM noting the key doesn't exist?

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -368,28 +368,29 @@ def report_ration(cls, ration: Dict, available_feeds: Dict) -> Tuple[Dict, Dict]
 
         for key, val in ration.items():
             nutrient_amount["dm"] += val
```
ew3361zh, line 370: Do you still want the value for the key added to the nutrient amount's 'dm' if the key is invalid?

ew3361zh (commented):
A couple quick questions for you to consider - please just ask for my re-review after they've been addressed.

tomhuhh:
Closing this PR because it is not applicable anymore... Will check and create (if it does not exist) an issue regarding the fact that the calf ration is hard coded.

Issue #1142, opened 2024-01-22
Python KeyError: 'status' 


tomhuhh:
It seems that there is an attempt to delete the key 'status' from the ration_per_animal dictionary, but that key doesn't exist in the dictionary, resulting in a KeyError.

## Expected Behavior

The code runs with no difficulty even when the 'status' or 'objective' key does not exist. 

## Current Behavior

A 'KeyError' as shown below. 

![Screenshot 2024-01-22 at 8 29 04 AM](https://github.com/RuminantFarmSystems/MASM/assets/52800053/7c61d19f-e67c-433e-aa54-3215679b70c4)

## Possible Solution

Check if the 'status' key exists before attempting to delete it. 

```
if 'status' in ration_per_animal:
    del ration_per_animal['status']
```

## Steps to Reproduce

1. Run the input files of farm 13 in pilot testing cohort 2 on the main branch. 
2.
3.
4.

tomhuhh:
I was not able to replicate the error either. 

However, as @JoeWaddell verified, there is no future concerns associated with the error that I encountered because the `Del` statement deals with a copy of the `ration_per_animal`. 

Closing this issue and the related PR.

Issue #1137, opened 2024-01-19
Refactoring Calf Ration 


tomhuhh:
Currently the calf ration is hard coded. This causes trouble when there is only one of the two ingredients (202, 216) in the calf diet. 

## Expected Behavior

The calf's ration can be user defined. 

## Current Behavior

The code ignores what the user input as calf's ration. 

Here is the error you can get. 
![image](https://github.com/RuminantFarmSystems/MASM/assets/52800053/cdfabf77-7359-4dd1-a6eb-3369e25b41a2)

## Possible Solution

A way to get around with this is to set the inclusion rates as follows: for example, when 216 is the only ingredient, 202: 0, 216: 150. 

## Steps to Reproduce

1. Update the calf's diet to only contain 216 and change its inclusion to 100%.
2. Run the simulation and you should see the error as shown in the screenshot above. 
3.
4.

PR #1143, opened 2024-01-22
Update animal_module_reporter.py

tomhuhh:
This PR is to address issue #1142. It seems that there is an attempt to delete the key 'status' from the ration_per_animal dictionary, but that key doesn't exist in the dictionary, resulting in a KeyError.

Our solution is to check if the 'status' key exists before attempting to delete it. In this PR, we added the if statement for both 'status' and 'objective'. 

To validate the change, first grab the input files of farm 13 in pilot testing cohort 2 and run it on the main branch to repeat the error and then use this updated branch to run the same input files and you should expect no errors. 

Yet, we still need to review the code where the `ration_per_animal dictionary` is constructed or modified to understand why the 'status' key might be missing when it's expected. A comment from Joe: 'I think it's because it's inherited from a previous formulation or other animal'.

### DIFF
--- a/RUFAS/routines/animal/animal_module_reporter.py
+++ b/RUFAS/routines/animal/animal_module_reporter.py
@@ -166,8 +166,10 @@ class AnimalModuleReporter:
         }
         for pen in animal_manager.all_pens:
             ration_per_animal = pen.ration_per_animal.copy()
-            del ration_per_animal["status"]
-            del ration_per_animal["objective"]
+            if 'status' in ration_per_animal:
+                del ration_per_animal["status"]
+            if 'objective' in ration_per_animal:
+                del ration_per_animal["objective"]
             ration_total = {}
             ration_total["dry_matter_intake_total"] = 0
             for key in ration_per_animal.keys():


JoeWaddell (approved):
Good fix for catching edge case errors.

PooyaHekmati:
> Yet, we still need to review the code where the `ration_per_animal dictionary` is constructed or modified to understand why the 'status' key might be missing when it's expected. A comment from Joe: 'I think it's because it's inherited from a previous formulation or other animal'.

This makes me a bit uncomfortable here. We need to understand both upstream and downstream impacts before making the change. In other words, 1, why is this happening in the first place? Is it a signal that something bigger is going on and we are not aware of it? 2, what happens downstream after we make this change? how does it effect other things?

JoeWaddell:
@PooyaHekmati excellent points, I was too hasty in approving. I assumed it had to do with the hard-coded ration in calf.py #1137 (we had a conversation about this earlier), so my mistake! 

What I thought was happening is that folks wanted to hardcode in a different calf ration, which would throw an error here (if they didn't add the status and objective fields, which are only used in the optimization process). Calf ration formulation is a 'wish', so it's been backburnered. 

Now that I'm taking a closer look, @tomhuhh I cant' replicate the error on my end. Can you add the files to replicate it to this branch, so that we're all on the same page?

JoeWaddell (changes requested):
See previous comment, can't replicate the outlined error.

PooyaHekmati:
> @PooyaHekmati excellent points, I was too hasty in approving. I assumed it had to do with the hard-coded ration in calf.py #1137 (we had a conversation about this earlier), so my mistake!
> 
> What I thought was happening is that folks wanted to hardcode in a different calf ration, which would throw an error here (if they didn't add the status and objective fields, which are only used in the optimization process). Calf ration formulation is a 'wish', so it's been backburnered.
> 
> Now that I'm taking a closer look, @tomhuhh I cant' replicate the error on my end. Can you add the files to replicate it to this branch, so that we're all on the same page?

No problem at all @JoeWaddell, since PR 1000, we have started taking the Zen of Python more seriously. A few point from the Zen of Python:
_Errors should never pass silently._
_In the face of ambiguity, refuse the temptation to guess._
_Special cases aren't special enough to break the rules._

In cases like this PR, we have to be careful about the big picture. If it is possible, we have to work through the downstream and upstream systems first. If the issue is major and we can't afford to investigate the interconnected systems right now, we can snooze the alert for now by clearly documenting that xyz problem was observed and it was decided to silence it for now because of abc reason. The xyz problem needs to be revisited by doing this and that.

tomhuhh:
I was not able to replicate the error either. 

However, as @JoeWaddell verified, there is no future concerns associated with the error that I encountered because the `Del` statement deals with a copy of the `ration_per_animal`. 

Closing this PR and the related issue.

PR #1144, opened 2024-01-22
Manure total ammoniacal nitrogen

### LINKED ISSUE #1066: Manure Total Ammoniacal N conversions and use of variables are incorrect
## Expected Behavior
There are several inconsistencies in the manure nitrogen and ammonia N tracking. These calculations should use values of manure composition provided by the animal module and track the total mass of manure N, manure ammoniacal N, and ammonia N at each step along the way. Manure ammoniacal N is a subset of total manure N and ammonia N emissions are generated from manure ammoniacal N. 

## Current Behavior

1. The `PenManure` class does not divide `animal_manure["total_ammoniacal_nitrogen_concentration"]` by the number of animals in the pen to get the average total ammoniacal concentration before multiplying by the volume to get the mass of total ammoniacal nitrogen to create the `manure_total_ammoniacal_nitrogen` attribute. It also creates a separate `urine_total_ammoniacal_nitrogen` attribute that is used by the `housing_ammonia_emission` function in calculator.py however the `urine_total_ammoniacal_nitrogen` should not be needed or used.
2. Slurry and solid manure densities are incorrect
3. Ammonia emissions are not being adjusted for N content before removal from total ammoniacal N variables and are not being subtracted from total manure N at all
4. The method for estimating ammonia emissions from manure storage/treatment does not use the correct value as an input for the total ammoniacal N. 
- Update to point 4 above: I have confirmed that several of the Manure Treatment methods are using different total ammoniacal nitrogen values as the inputs to estimate ammonia emissions. This has also led me to understand that the different methods are also calculating the accumulated_total_ammoniacal_nitrogen slightly differently. A couple of things to change for consistency: 

- [ ] Each manure treatment method has a `'daily_update_helper` function. Within this function, the following operations should be conducted in this order: 1) `accumulate_daily_output`, 2) all emissions estimates (methane, then ammonia, then nitrous oxide), 3) the accumulated outputs for liquid_manure_total_ammoniacal_nitrogen, liquid_manure_total_solids, storage_methane, storage_ammonia, and storage_nitrous_oxide should be updated. 
- [ ] the inputs to the ammonia emissions estimation methods in all manure treatment methods should be the accumulated_output.liquid_manure_total_ammoniacal_nitrogen values. 

We will need to change the naming conventions somepoint soon because currently all nitrogen outputs from the manure handler are called liquid_manure_nitrogen or liquid_manure_total_ammoniacal_nitrogen leaving no room for non liquid nitrogen but I think we can update that later. I will create a google doc to share that describes the updated calculations for the ammonia emissions inputs and accumulated manure values in more detail. 

## Possible Solution

- [ ] Divide `animal_manure["total_ammoniacal_nitrogen_concentration"]` by `num_animals` [here](https://github.com/RuminantFarmSystems/MASM/blob/814258e2bec0d026d0c3c58b15b1b0daa46ace84/RUFAS/routines/manure/pen_manure/pen_manure.py#L111)
- [ ] Update `housing_ammonia_emission()` so that it uses `manure_total_ammoniacal_nitrogen` instead of  `urine_total_ammoniacal_nitrogen`
- [ ] In the manure_constants.py file update the manure densities to the following: 
    - [ ] LIQUID_MANURE_DENSITY = 1000
    - [ ]  SLURRY_MANURE_DENSITY = 990
    - [ ] SOLID_MANURE_DENSITY = 700
- [ ] More to come for 3&4

## Steps to Reproduce

1.
6.
7.
8.

## Possible Implementation

I have started a draft PR ([1065](https://github.com/RuminantFarmSystems/MASM/pull/1065)) for my debugging and testing purposes

ew3361zh:
PR to address issue #1066 - these are the changes implemented to address the issue:

### In gas_emission_constants.py
- SOLID_MANURE_THRESHOLD updated from 8.0 -> 0.08 
- SLURRY_MANURE_THRESHOLD updated from 5.0 -> 0.05
### In manure_constants.py 
- LIQUID_MANURE_DENSITY updated from 990.0 -> 1000
- SLURRY_MANURE_DENSITY updated from 1400 -> 900
- SOLID_MANURE_DENSITY updated from 1250 -> 700
### In calculator.py 
- urine ammoniacal nitrogen references updated to manure ammoniacal nitrogen
- dry_matter calculation updated from `manure_mass` / `total_solids` to `total_solids` / `manure_mass`
- removes unused `methane_emission_FOR_slurry_storage()` method
### In manure_handler_classes.py
- urine ammoniacal nitrogen references updated to manure ammoniacal nitrogen
### In anaerobic_digestion.py
- `calc_daily_sludge_output()` method removed
### In anaerobic_lagoon.py
- `calc_daily_sludge_output()` method removed
### In compost_bedded_pack_barn.py
- `accumulate_daily_update()` method moved to immediately after daily_output calculation.
- storage methane, storage ammonia, storage nitrous oxide now subtracted from accumulated output
### In manure_treatment_configs.py
- `total_solids_removed_efficiency_for_treatment` set to 0 in SLURRY_STORAGE_UNDERFLOOR_CONFIG and SLURRY_STORAGE_OUTDOOR_CONFIG.
- `sludge_accumulation_volume_fraction` set to 0 in DefaultManureTreatmentConfigFactory
### In open_lots.py
- `accumulate_daily_update()` method moved to immediately after daily_output calculation.
### In slurry_storage_outdoor.py
- `methane_loss` now calculated by GasEmissionsCalculator `methane_emission_from_slurry_storage()` method (previously `methane_emission_FOR_slurry_storage()`)
- `calc_ammonia_emission()` method no longer uses `barn_area` param and now takes `accumulated_manure_total_solids` param. This is to accommodate the switch from calculating `ammonia_loss` using the `GasEmissionsCalculator.ammonia_emission()` method -> `GasEmissionsCalculator.storage_ammonia_emission()`
### In slurry_storage_outdoor.py
- `methane_loss` now calculated by GasEmissionsCalculator `methane_emission_from_slurry_storage()` method (previously `methane_emission_FOR_slurry_storage()`)
### In pen_manure.py
- updates manure constant used for `manure_volume` calculation from MANURE_DENSITY -> SLURRY_MANURE_DENSITY
- updates calculation of total_ammoniacal_nitrogen to divide animal_manure["total_ammoniacal_nitrogen_concentration"] by `num_animals`

Updates unit testing for all these files and functions.


KFosterReed:
Hi @ew3361zh I am guessing you made the most recent commit based on a conversation with @elle-andreen because the original equation for the housing manure is based on urine not manure TAN. I think that is a good idea for consistency with the published work and the design docs, however, the important thing is what is becoming inputs to the housing ammonia equations. However, I just want to make sure that we are on the same page that we do not want to create or track a separate urine TAN value in the manure classes. 

also FYI I started reviewing this an hour or so ago and it is looking pretty good so far! Hope to have comments to you before Monday :) 

THanks for your work and happy Friday!

elle-andreen:
> Hi @ew3361zh I am guessing you made the most recent commit based on a conversation with @elle-andreen because the original equation for the housing manure is based on urine not manure TAN. I think that is a good idea for consistency with the published work and the design docs, however, the important thing is what is becoming inputs to the housing ammonia equations. However, I just want to make sure that we are on the same page that we do not want to create or track a separate urine TAN value in the manure classes.
> 
> also FYI I started reviewing this an hour or so ago and it is looking pretty good so far! Hope to have comments to you before Monday :)
> 
> THanks for your work and happy Friday!

You're correct! Niko just beat me to making the commit before I could make notes here regarding that change and a couple other items in this PR that he and I discussed this afternoon! I'll add those here shortly. Just to clarify from your comment - by manure classes do you mean the individual manure treatments? In other words, after we calculate housing ammonia emissions, we want to be subtracting the NH3-N from the pen **manure TAN** value, not creating and tracking a separate/additional urine TAN value that's utilized in the manure treatment file. Let me know if I'm following that right, I wasn't 100% clear from your comment whether we are good to go with using urine TAN for housing emissions or if we should revert (un-revert??) to using manure TAN as the input.

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -871,7 +822,7 @@ def _housing_specific_constant(
         elif manure_mass == 0.0 or total_solids == 0.0:
             return GasEmissionConstants.SOLID_AND_SEMI_SOLID_MANURE_HSC
 
-        dry_matter = manure_mass / total_solids
+        dry_matter = total_solids / manure_mass
```
elle-andreen: Currently, the housing specific constant (HSC) value used in calculating ammonia emissions from storage (starting at line 605) for most manure treatments is determined based on the %DM of the stored manure (line 825 here in calculator.py). This %DM is then compared with thresholds given in gas_emission_constants.py to assign an HSC value associated with slurry, solid/semisolid, or liquid manure. Kristan and I discussed that a simpler way to assign the HSC value is just basing HSC on manure treatment. In manure_manager.py, each manure treatment has been designated as producing either liquid or solid manure, originally to simplify nutrient requests from crop module. This liquid/solid classification could also be treated as an attribute of the pen, and all manure assigned to the that treatment would use the liquid or solid manure HSC value (LIQUID_MANURE_HSC = 4.1, SOLID_AND_SEMI_SOLID_MANURE_HSC = 10), regardless of actual DM. I believe this ammonia emission calculation is currently only used with manure treatments that are all designated as liquid manure anyway (i.e., solid manure-producing treatments calculate NH3-N with different equations). In the future we may be able to reinstitute %DM as a more precise determinant of HSC, so I think we should leave our solid and semi-solid manure HSC in place for now. Perhaps could just remove the slurry HSC value (because we're not producing an manure designated as slurry), and add whatever logic is needed to assign HSC by manure type rather than basing it on calculated %DM.

RUFAS/routines/manure/manure_treatments/slurry_storage_underfloor.py
```diff
@@ -113,9 +112,8 @@ def _daily_update_helper(self) -> ManureTreatmentDailyOutput:
             slurry storage underfloor system.
```
elle-andreen, line 104: (for line 92): I think this should be `ammonia_loss = GasEmissionsCalculator.storage_ammonia_emission`? As far as I know, slurry outdoor and slurry underfloor should utilize the same storage ammonia emission function in calculator.py, and slurry_storage_outdoor.py has already been updated in this PR to reference GasEmissionsCalculator.**storage**_ammonia_emission (line 244 in slurry_storage_outdoor.py) instead of ammonia_emission.

KFosterReed:
> > Hi @ew3361zh I am guessing you made the most recent commit based on a conversation with @elle-andreen because the original equation for the housing manure is based on urine not manure TAN. I think that is a good idea for consistency with the published work and the design docs, however, the important thing is what is becoming inputs to the housing ammonia equations. However, I just want to make sure that we are on the same page that we do not want to create or track a separate urine TAN value in the manure classes.
> > also FYI I started reviewing this an hour or so ago and it is looking pretty good so far! Hope to have comments to you before Monday :)
> > THanks for your work and happy Friday!
> 
> You're correct! Niko just beat me to making the commit before I could make notes here regarding that change and a couple other items in this PR that he and I discussed this afternoon! I'll add those here shortly. Just to clarify from your comment - by manure classes do you mean the individual manure treatments? In other words, after we calculate housing ammonia emissions, we want to be subtracting the NH3-N from the pen **manure TAN** value, not creating and tracking a separate/additional urine TAN value that's utilized in the manure treatment file. Let me know if I'm following that right, I wasn't 100% clear from your comment whether we are good to go with using urine TAN for housing emissions or if we should revert (un-revert??) to using manure TAN as the input.

So this is another great example of how we have to make a decision in how much 'RuFality' deviates from reality. Then intent behind changing the name in this function back to 'urine TAN' instead of 'manure TAN' is that the equations are supposedly based on the urine TAN exclusively. (This in itself is a bit of a deviation from reality because urine nitrogen is in the form of urea, creatinine, and purine derivatives, not ammoniacal N. It is not until the urine urea mixes with the urease in manure that ammonium (or TAN) is produced. However we will work with the framework of the published equations for now). At this point in the manure management chain, all of the TAN is coming from urine anyway, however, because we are tracking and simulating the manure (combination of feces and urine and then later bedding and cleaning water), we have only created a pool called 'manure TAN', which at this stage in the order of operation should be equivalent to urine TAN. 

A few things to note here: 
1. In this commit, Niko has only changed the name of the variables inside the function. This does not actually change the quantity of TAN being passed into the function so we are passing in 'manure TAN' but implementing the function with variable names called 'urine TAN'. (again, I think this could be a good choice to be in alignment with the published literature.)
2. Varma had created a separate variable called 'urine TAN' in `pen_manure.py` however, this is disconnected from our method to estimate TAN from the animal excretion methods. What we should not do is have two different methods to estimate the same thing (again here emphasizing that at this stage of the manure chain, manure TAN == urine TAN). This would (and was) created a lot of confusion about the discrepancies between the animal manure excretion estimates and the manure module outputs. My preference would be to delete the 'urine TAN' variable. After looking right now, the only place it is still used is in the anaerobic lagoon method. This is incorrect not only because we should be using 'manure TAN' but because the inputs to the manure treatment functions should be from the reception pit or separators. I will make a commit to change this soon. 


So to summarize, I think we can leave this because it does align with the equation we are using. However, I think we should make a note in the documentation for the housing ammonia emissions that we have made the simplifying assumption that 'urine TAN' = 'manure TAN' at this point.

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
```diff
@@ -123,8 +123,8 @@ def daily_update(self,
         housing_ammonia_emission = GasEmissionsCalculator.housing_ammonia_emission(
             num_animals=pen.num_animals,
             barn_area_per_animal=pen.barn_area_from_pen_type,  # m^2/animal
-            urine_total_ammoniacal_nitrogen=pen.manure.urine_total_ammoniacal_nitrogen,  # kg
-            urine=pen.manure.urine,  # kg
+            manure_total_ammoniacal_nitrogen=pen.manure.manure_total_ammoniacal_nitrogen,  # kg
```
KFosterReed: @elle-andreen - this is where the inputs are passed to ammonia emissions function. As it is, the function doesn't actually run because it is looking for `urine_total_ammoniacal_nitrogen` and we are passing in a variable named `manure_total_ammoniacal_nitrogen`

elle-andreen:
...
> 
> So to summarize, I think we can leave this because it does align with the equation we are using. However, I think we should make a note in the documentation for the housing ammonia emissions that we have made the simplifying assumption that 'urine TAN' = 'manure TAN' at this point.

Thanks Kristan, your comments really helped me fill in some missing pieces of info! I was not aware previously that not only do we not want to track urine TAN in any way in the manure module, we weren't even actually generating/passing a urine TAN value. I agree that if we are considering solely manure and not feces & urine fractions in the manure module, it is not ideal to have a urine TAN variable here. It sounds like it's the urine **mass** part of this equation that we cannot easily get around - we can assume urine TAN = manure TAN but cannot set urine mass = manure mass (either in the code, or in reality). I think what you've written/committed is a good comprise between 'Rufality' and the original publication that hopefully avoids further confusion. I added a short explanation of this in the documentation. I apologize for my part in the back and forth on this Niko!

KFosterReed:
> ...
> 
> > So to summarize, I think we can leave this because it does align with the equation we are using. However, I think we should make a note in the documentation for the housing ammonia emissions that we have made the simplifying assumption that 'urine TAN' = 'manure TAN' at this point.
> 
> Thanks Kristan, your comments really helped me fill in some missing pieces of info! I was not aware previously that not only do we not want to track urine TAN in any way in the manure module, we weren't even actually generating/passing a urine TAN value. I agree that if we are considering solely manure and not feces & urine fractions in the manure module, it is not ideal to have a urine TAN variable here. It sounds like it's the urine **mass** part of this equation that we cannot easily get around - we can assume urine TAN = manure TAN but cannot set urine mass = manure mass (either in the code, or in reality). I think what you've written/committed is a good comprise between 'Rufality' and the original publication that hopefully avoids further confusion. I added a short explanation of this in the documentation. I apologize for my part in the back and forth on this Niko!

No apologies accepted! This was an important distinction and I am grateful you brought it to our attention!

RUFAS/routines/manure/gas_emissions/calculator.py
KFosterReed: I think there are a couple of things going on here and we can solve a few problems with one simplifying fix that is related to Elle's suggestion. 

1. The calculations for DM% throughout the manure module need some review. I was tracking the total_solids and manure_mass variables over the weekend and while I am working towards a fix for those, I think the plan to dissociate the ammonia emissions from that variable for now is a good idea
2. Elle's comment that for the solid manure types, we are using a different method, helped a light bulb go off for me. She is correct that for open lots and compost bedded pack barns, we should be using a different method, which means we shouldn't have any 'Manure Handler' ammonia emissions at all and this function should always return a 0. 
3. I didn't realize this before but this function is ONLY used during manure treatments, not when the emissions are from the barn floor.
4. The last piece that was puzzling me was that we were leaning towards not designating a manure type (i.e. solid or liquid) until the reception pit. So I was struggling with having liquid manure types being assigned based on the pen.
5. Based on the above, this is where I have landed: Let's get rid of the check on the DM content of the manure in this function and make it based on if it is called from the ManureHandler or from a ManureTreatment. Then, in the ManureHandler, can we add a check before the `housing_ammonia_emission()` function is called and only call it if the pen_type is 'freestall' or 'tiestall'? See the issue 1172 that I created this morning for more about the pen types. 

The values returned would look like this: 
If the function is called from the Manure Handler: the housing specific constant is 260
If the function is called from a Manure Treatment: the housing specific constant is 4.1 because we are going to assume that it is a liquid manure (as Elle suggested earlier). 

An extra safe option would be to check the pen type and treatment types before returning the HSC from this function but I am not sure it is necessary?

(side note, should we call it `get_housing_specific_constant`? )

elle-andreen: > 1. The calculations for DM% throughout the manure module need some review. I was tracking the total_solids and manure_mass variables over the weekend and while I am working towards a fix for those, I think the plan to dissociate the ammonia emissions from that variable for now is a good idea

Yes, I did not explicitly say that my suggestion would avoid rely on DM% tracking because I haven't look at simulated DM mass or % outputs, but I'm fairly sure we don't have great estimates of evaporation, which in and of itself could affect moisture content and thus %DM in a big way!

> 2. Elle's comment that for the solid manure types, we are using a different method, helped a light bulb go off for me. She is correct that for open lots and compost bedded pack barns, we should be using a different method, which means we shouldn't have any 'Manure Handler' ammonia emissions at all and this function should always return a 0.

For this function, yes! CBPB pens each have their own ammonia emission estimation methods.

> 3. I didn't realize this before but this function is ONLY used during manure treatments, not when the emissions are from the barn floor.

Yep - for barn emissions, HSC is always set at 260 (for better or worse)

> 4. The last piece that was puzzling me was that we were leaning towards not designating a manure type (i.e. solid or liquid) until the reception pit. So I was struggling with having liquid manure types being assigned based on the pen.
> 5. Based on the above, this is where I have landed: can we make the `housing_specific_constant` function based on 1) if it is called from the ManureHandler or from a ManureTreatment and then 2) if it is called from the ManureHandler have an additional check to see what pen type it is coming from? side note, should we call it `get_housing_specific_constant`? )

From a non-SWE perspective, this sounds good to me! That single function can 'get' the HSC for us based on the ManureHandler/ManureTreatment and pen type contingencies. I think renaming the constant for the indoor emissions HSC (currently DEFAULT_HOUSING_SPECIFIC_CONSTANT) to something like 'indoor_housing_specific_constant' would be helpful as well.

ew3361zh: Thank you both! A few points of clarity to get these changes implemented:

1. `storage_ammonia_emission()` in calculator.py is used ONLY by SlurryStorageOutdoor and AnaerobicLagoon to calculate ammonia emissions. I will update the `storage_ammonia_emission()` function to only use the Liquid Manure HSC (4.1) as detailed above.
2. SlurryStorageUnderfloor though currently uses the calculator.py `ammonia_emission()` function to calculate ammonia emission. If we decide we want to keep this method of ammonia emissions calculations for SlurryStorageUnderfloor, I can update the logic to not specify the HSC its using. `calc_ammonia_emission()` in `SlurryStorageUnderfloor` could then send it the Liquid Manure HSC instead of using the DEFAULT HSC which `ammonia_emission()` currently has hard coded as a default value for the HSC it uses for calculations. 
3. However, as detailed in @elle-andreen table in the [google doc](https://docs.google.com/document/d/1jF62faHDByADKJ0oCs_E0vV3kuHJEYlzLWH3PQUFyZU/edit), it looks like SlurryStorageUnderfloor should be using the same method of ammonia emissions calculations as SlurryStorageOutdoor and AnaerobicLagoon (and they should all be the RO-S method). So is the RO-S method of ammonia calculation the `storage_ammonia_emission()` or the `ammonia_emission()` method?
4. I will keep `housing_ammonia_emission()` function as is using the DEFAULT HSC because it's only being used by ManureHandler classes.
5.  `_housing_specific_constant()` is where the dry matter calcluations are used to determine which HSC to return. If we make the changes in part 1 of this list, the function will no longer be used anywhere in RuFaS. Is it important to keep around? It sounds like maybe dry matter calculations need an overhaul throughout but wasn't sure if this was helpful to have around for reference even if it's not being used currently.

ew3361zh: > 5\. Based on the above, this is where I have landed: Let's get rid of the check on the DM content of the manure in this function and make it based on if it is called from the ManureHandler or from a ManureTreatment. Then, in the ManureHandler, can we add a check before the `housing_ammonia_emission()` function is called and only call it if the pen_type is 'freestall' or 'tiestall'? See the issue 1172 that I created this morning for more about the pen types.

One point of clarification here since I don't think I was super clear in my part 5 above - the `_housing_specific_constant()` is ONLY called once by any other function which is the `storage_ammonia_emission()` method used only by SlurryStorageOutdoor and AnaerobicLagoon.

We can get rid of the check on DM but because this function is only called one other place which is used only by Manure Treatment classes, it will always just return the Liquid Manure Constant. If we want it to evolve to be incorporated into some of the methods that have the DEFAULT_HSC hard coded in, I can look into that option.

KFosterReed: Hi Niko, 
thanks for laying all this out so clearly! After reviewing Elle's helpful table more closely. Here are some responses:

1. I agree that we should get rid of `ammonia_emission()` and only use the `storage_ammonia_emission()` function for SlurryStorageOutdoor, SlurryStorageUnderfloor, and AnaerobicLagoon. 
2. In the future, I think there is a path to combine `housing_ammonia_emission` and `storage_ammonia_emission` in the future but I think for now let's leave them separate. 
3. Similar to above, I think in the future, there is a way to have all HSC come from the `housing_specific_constant()` function but for now, let's delete it and have them fixed with constant input values to the two different ammonia emission functions. 
4. Can we rename the constants so it is clear one is for housing and one is for storage?

elle-andreen: I agree with deprecating housing_specific_constant. We may need it in the future but we don't right now!

I spent a while thinking about these changes this afternoon. It sounds like we're okay with hard-coding the HSCs for housing and storage ammonia emissions for now. If so, I thought the following might simply things (sorry for the image format, I was working on this in OneNote). This should incorporate most or all of the changes we've listed, hopefully it is not hard to follow:

The resistance equation itself ('return' in _ammonia_barn_resistance) is currently the same for all pen types and manure treatments that use these ammonia emission equations, it is just the HSC value that differs between housing and storage. We can repurpose the current resistance calculation function to calculate resistance based on a given temperature and HSC, regardless if those values are coming from the housing or storage ammonia emission function. 

1. In storage_ammonia_emission, set HSC equal to LIQUID_MANURE_HSC (4.1).
2. In housing_ammonia_emission, set HSC equal to DEFAULT_HOUSING_SPECIFIC_CONSTANT (260).
		> Rename DEFAULT_HOUSING_SPECIFIC_CONSTANT to 'housing_HSC' or similar
3. Rename the _ammonia_barn_resistance function to _ammonia_resistance, and update its respective uses in the housing_ammonia_emission and storage_ammonia_emission functions. 
		> Note that in storage_ammonia_emission, this value is renamed to 'storage_area_resistance'
4. Update the _ammonia_resistance function to be able to accept whatever HSC value either housing_ammonia_emission or storage_ammonia_emission passes to it (I think that's possible, right?)

image

ew3361zh: Thank you both again - I've instituted the following changes which I believe covers everything we've talked about but please let me know if you see something I've missed or would like any of these changes reverted.

- `LIQUID_MANURE_HSC` -> `STORAGE_HSC`
- `DEFAULT_HOUSING_SPECIFIC_CONSTANT` -> `HOUSING_HSC`
The latter is a little redundant but if there is a different name for either of these you like more, it's not difficult to change.
- `storage_ammonia_emission()` now uses `STORAGE_HSC`
- `housing_ammonia_emission` in the manure_handler_classes.py `daily_update()` method now only calculates ammonia emissions if the pen type is "freestall" or "tiestall" and otherwise is set to 0.0
- `SlurryStorageUnderfloor` now uses `storage_ammonia_emission()` to calculate ammonia loss instead of `ammonia_emission()`. This puts it in line with `AnaerobicLagoon` and `SlurryStorageOutdoor`
- `_housing_specific_constant()` method in calculator.py is removed
- `_ammonia_barn_resistance()` -> `_ammonia_resistance()` as outlined by @elle-andreen above.

KFosterReed:
Ok team! I think I finally got it. I ended up implementing a maximum storage ammonia emission production after going back to the original paper again because I was getting crazy amounts of storage ammonia in the first couple days of the simulation. This is what they said in the Rotz and Oenma paper that finally got the lightbulb to go off for me: 

"Daily loss of NH3−N was determined such that the cumulative loss up to a given date could not exceed the accumulated TAN in storage. This was particularly important in the early stages of loading, when a thin layer of manure on the bottom of the storage facility created maximum exposure for the loss of TAN"

Once that was implemented and we were actually tracking the manure TAN and manure total solids then I started getting errors in the nutrient request manager because the mass of N was > than the mass of the manure due to an odd subtraction and division by 1000 in setting the `liquid_total_manure_mass` part of the the `add_manure_nutrients()` function in `manure_manager.py`. Right now we are not tracking the losses of total manure mass in storage (from what I can tell).  But I think we should tackle total solids and volatile solids first and then move on to total mass and dry matter. The storage ammonia emissions more or less make sense to me even though (in the early phases especially) we are losing all of the TAN to ammonia. 

@elle-andreen let me know what you think!
@ew3361zh  I am sure there is probably a better/cleaner/simpler way to update the DailyOutput and AccumulatedManure variables at the end of the `daily_update_helper()` functions. I am OK with putting that in a future issue if you think that is a good idea.

elle-andreen:
This is looking great to me! In terms of the quantity and pattern of housing and storage NH3-N emissions, those look to me to be aligning well with our expected daily and cumulative emissions. When I run the simulation, am not seeing the very high (e.g., ammonia emission = manure TAN) initial emissions of ammonia that you're describing, but that may just be input-dependent. From what I can tell, we seem to be tracking and updating N loss correctly within the manure module at least (i.e., excluding any conversion errors or other weirdness between the animal and manure modules); both housing and storage ammonia emissions are being subtracted from the daily manure TAN values correctly.

One thing I will mention just for our future reference: I see in the last batch of commits that, in `def storage_ammonia_emission`, we've replaced the resistance calculation (`_ammonia_resistance`) with using the STORAGE_HSC value directly, which is as it should be! Sometime soon we should go back and rename STORAGE_HSC to something like STORAGE_ RESISTANCE and also deprecate other storage-related HSC's for now. I'll add it to the list of manure module issues to be opened.

I know we still have a couple issues to work out but I think within the scope of this PR, we should be good to go!

ew3361zh:
> @ew3361zh I am sure there is probably a better/cleaner/simpler way to update the DailyOutput and AccumulatedManure variables at the end of the `daily_update_helper()` functions. I am OK with putting that in a future issue if you think that is a good idea.

I think we could look at extracting some of that code into other helper functions but I know we talked Friday about wanting to align them in the order of operations they individually do as best we could. While it could potentially be straightforward, I think it could be good to do in a separate PR that's just focused on aligning those functions' formats and orders. I think a good goal for this PR is to get the outputs tracked the way we want them. Then at least when we're refactoring those helper functions, we'd be able to identify any issues that's disturbing the outputs being tracked to our liking.

I fixed some of the linting issues and can work on the testing once Elle has taken a look and agrees with the changes (or we figure out what else needs to get changed).

elle-andreen (approved):
Our poop is finally in a group 😄 Great job Niko!

I just went back through the changes here and everything is looking good to me as far as I can tell. My one question is regarding the sections we removed that pertain to sludge manure. Sludge manure is something that accumulates inside anaerobic digesters or lagoons, I believe we want to ignore it for now but eventually we probably need to bring it back into the manure chain. There's still plenty of spots in the code where sludge manure is referenced. If it seems likely we will eventually bring sludge back, should we remove these sections? Or just leave them be, knowing we're not using them right now?

KFosterReed:
> Our poop is finally in a group 😄 Great job Niko!
> 
> I just went back through the changes here and everything is looking good to me as far as I can tell. My one question is regarding the sections we removed that pertain to sludge manure. Sludge manure is something that accumulates inside anaerobic digesters or lagoons, I believe we want to ignore it for now but eventually we probably need to bring it back into the manure chain. There's still plenty of spots in the code where sludge manure is referenced. If it seems likely we will eventually bring sludge back, should we remove these sections? Or just leave them be, knowing we're not using them right now?

I wish there was a poop emoji I could put on your message Elle! 

My thoughts on the sludge question are similar to other parts of code that are being deprecated that we might bring back at some point... let's get rid of it now and bring it back if we need it. WE can always look at the old code if needed but removing it now I think makes it easier to streamline / fix/ debug in the short term so my vote is to remove anything that isn't being used

KFosterReed (approved):

PR #1146, opened 2024-01-22
Reformat the variable names of Crop and Soil output

### LINKED ISSUE #965: Inconsistency in crop and soil output data format
Crop and soil uses a class called OutputGathere to report its variables to the OutputManager. To do so, it uses prefixes in the info_map. There are few issues here:
1. According to the `OutputManager.add_variable` docstring, if a prefix is present in the info_map, it suppresses the automatic prefix generation. This means that almost everywhere in the codebase, the output keys are: `class.function.variable`; but crop and soil outputs are `field=<field_name>.variable`. Not only this is inconsistent with the rest of the system, it also doesn't respect the semantics.
2. OutputGathere populates info_map as: `info_map = {"class": self.__class__.__name__, "function": self.send_daily_variables.__name__}`. This is not a good practice. The `class` and `function` are reported in the info_map to make it possible to backtrack and find the origin of an output variable. In the case of crop and soil, the OutputGatherer is the messenger, not the source. So the class and function names should not be set to `OutputGatherer` and `send_daily_variables`. This obfuscates the results and is the reason why I didn't develop OutputManager such that it automatically gets the caller's info. If field_name is needed, it can be reported as an additional field in the info_map or as a suffix.

Isgotkowitz:
This PR changes how variables are recorded in the Crop and Soil module.

Previously, they were recorded in the form:
```
field='<field name>',<other meta information>.<variable name>
```
but are now recorded as:
```
<class name>.<function name>.<variable name>.field='<field name>',<other meta information>
```
All uses of the OutputManager in Crop and Soil have been checked to make sure that the above changes were made correctly, and all tests have been updated accordingly.

Additionally, the module `OutputGatherer` has been changed to `FieldDataReporter` to indicate that it specifically records data from the Field module (although it still does not indicate this perfectly). I messed up changing the name of the associated test file though (my apologies), so please know that `test_output_gatherer.py` is essentially the same as `test_field_data_reporter.py`, just with corrected tests.

Closes #965


PooyaHekmati (approved):

ew3361zh (changes requested):
This looks like it addresses part 1 of issue #965 nicely but not part 2 as all the data in the OM from C&S is still coming from a single class and function (`FieldDataReporter.send_daily_variables`). While it's clearer that the data is coming from the C&S module by renaming OutputGatherer to FieldDataReporter, the vital data of where each of these variables is originating within the codebase is still obscured. If the intention is to address the second part of the issue in a different PR, that's fine by me but I think it should be made clear and perhaps a separate issue should be created.

Screenshot 2024-01-23 at 11 11 38 AM

Isgotkowitz:
> This looks like it addresses part 1 of issue #965 nicely but not part 2 as all the data in the OM from C&S is still coming from a single class and function (`FieldDataReporter.send_daily_variables`). While it's clearer that the data is coming from the C&S module by renaming OutputGatherer to FieldDataReporter, the vital data of where each of these variables is originating within the codebase is still obscured. If the intention is to address the second part of the issue in a different PR, that's fine by me but I think it should be made clear and perhaps a separate issue should be created.
> 
> Screenshot 2024-01-23 at 11 11 38 AM

@ew3361zh good point, this PR does not address that aspect of the Crop and Soil output obfuscation. I will write up a separate issue for it, because figuring out a way to record daily data in a way that is consistent and centralized is at odds with the design of specifying the data in class.function.name and has implications outside of just the Crop and Soil module.

Edit: I have written up issue #1148 to address this.

ew3361zh (approved):

PR #1150, opened 2024-01-23
Add a "None" option for bedding types

### LINKED ISSUE #1132: Add 'None' option for bedding types
We would like to add an option to not apply bedding to the pens in the Manure Management methods. 

This would require adding adding a 'None' bedding type in the following places: 

1.  `DefaultBeddingConfigFactory` in bedding_classes.py
2. The metadata pattern options for `"bedding_type"` in the `"manure_management_scenario"` section
3. The metadata pattern options for `"bedding_type"` in the `"bedding_configs"` section
4. The manure management input json files `"bedding_configs"` section

As the name implies the values for all the configs will be 0 :)

**Additional context**

Isgotkowitz:
This PR adds an option for a "None" option for bedding types, as guided by the steps outlined in issue #1132. I am not sure exactly what this bedding option translates to in the real world, but in RuFaS it is just a bedding option that has no mass, no volume, and no nutrients.

For all 4 methods that this class inherits (calculating how much bedding is washed, mass of bedding, volume of bedding, and bedding dry matter solids) it returns 0 as the result in all cases. This seemed like the most logical approach to take, but is easy to modify if the behavior should be different.

I **_strongly_** encourage all reviewers with knowledge of the animal and manure module outputs to try using this new bedding option in a simulation and observing the results. I tried using it in the default scenario and it ran without crashing, but I have no idea how to assess the downstream effects of this bedding type.

Another note to reviewers about a possible but unrelated issue that I noticed: in the `calc_total_bedding_dry_solids()` method, the total mass of bedding is divided by the "bedding dry matter content". Looking at the definition for the bedding dry matter content, it seems to me like it is the fraction of dry matter in the bedding - which means that the dry solids in the bedding should be obtained by multiplying the mass by the dry matter content instead of dividing by it. But I do not have extensive experience with the Manure module, so my understanding may not be correct.

Update: the calculation of dry solids in bedding has been changed.

Closes #1132


jadamchick (approved):
Thanks Ed, this looks good to me from looking at the outputs, though I admittedly have limited knowledge of what goes on under the hood in the manure module. 

I ran this for c2f5, changing the bedding for pen 4 from "manure solids" to "none" (representing the 25% of the lact cow time that is spent milking or eating). Comparing the outcomes of this pen in new v. old scenarios (with "none" vs manure solids bedding), there was: 

- 100% less bedding mass & bedding volume
- some small changes in emissions (in both pens 3 & 4, in both directions) -- very small, much less than 1% fluctuations
- 2% lower storage ammonia from pen 4 in the new ("none") vs old ("manure solids") scenarios. I'm not sure if this could be attributed to the bedding or not. 

So with that, it looks good to me! Also tagging @krb78 here in case she has any specific thoughts or questions. 

@Isgotkowitz I agree with your suspicions about `calc_total_bedding_dry_solids` but also lack the expertise to say anything definitive.

krb78:
Appreciate the updates on the findings, Julie! :) I'm not too surprised by the super low changes in emissions, but think this is directionally appropriate and I'm good with that... in the grand scheme of things, no one should be using RuFaS to decide what bedding to use 😉

krb78:
I also think your calc_total_bedding_dry_solids guess is likely correct... But @ltp23 or @KFosterReed who spent a lot of time in the manure module lately probably can give us an answer.

Isgotkowitz:
@jadamchick @krb78 thank you for looking into this, I really appreciate it.

I will look further into `calc_total_dry_matter_solids()`

KFosterReed (approved):
Changes work as expected! I also checked implemented the proposed change to the bedding solids calculations and confirmed that it is both theoretically correct as well as correct in terms of how it is implemented now tracks the correct flow of mass in the manure system. Thanks Ed!

KFosterReed:
Oh, and as usual, I have broken a unit test without fixing it... once that unit test is fixed then my approval stands :)

Issue #1101, opened 2024-01-11
Inconsistency in the reported milk protein depending on the output filters used


tomhuhh:
The variable 'Cow.milking_update.milk_data_at_milk_update.milk_protein' might not be collecting the correct output as compared to the variable 'LifeCycleManager.daily_update.herd_milk_protein_percent'.

## Expected Behavior

There should be variations in the values of the variable 'Cow.milking_update.milk_data_at_milk_update.milk_protein' 

## Current Behavior

There is no variation in the values of the variable 'Cow.milking_update.milk_data_at_milk_update.milk_protein' 

## Possible Solution

Check where the variable 'Cow.milking_update.milk_data_at_milk_update.milk_protein' is updated.  

## Steps to Reproduce

1. Include these variables in your filter txt files and make sure they generate results first. 
2. Check and max and min for each variable. 
3.
4.

tomhuhh:
Rerunning the main branch and setting the filters as `Cow.milking_update.milk_data_at_milk_update`, `LifeCycleManager.daily_update.herd_milk_protein_percent`, I still found tiny discrepancies between the outputs in these two columns, which seems to me it might be a rounding error. It appears to me that it does not make sense to output the same information twice. Instead we could remove one variable if that is feasible to do. Please see the screenshot below. Thanks! 

<img width="618" alt="Screenshot 2024-01-30 at 2 50 37 PM" src="https://github.com/RuminantFarmSystems/MASM/assets/52800053/6ffc8754-c404-4fb3-8c16-54761f96ce7d">

tomhuhh:
I don't think this issue is relevant any more because RuFaS uses user input milk protein now. So I am closing it...

Issue #1097, opened 2024-01-09
Add milk fat and milk protein inputs to animal input data


jadamchick:
**Is your feature request related to a problem? Please describe.**
Milk fat and milk protein do not reflect actual farm production.

**Describe the solution you'd like**
Replace current dynamic fat percent method and hard-coded milk protein value with input values for average herd milk fat and milk protein. 

**Describe alternatives you've considered**
The ideal would be to have dynamic milk fat and protein equations reflective of animal physiology and nutrition. 
On the other end of the spectrum, we could not pay any attention to RuFaS' fat and protein outputs, and use farm-provided milk fat and milk protein %'s to calculate component-adjusted milk values post hoc. This requires assuming that milk composition does not meaningfully interact with RuFaS processes/outputs (intake, methane). 

**Additional context**

PR #1153, opened 2024-01-23
Add milk fat and milk protein percentages as user inputs

ltp23:
Issue #1097 

This PR adds two new user inputs for milk fat and milk protein percentages and directly replaces those two variables with the user inputs.

Before, the milk protein percentage was set as some default constant, and the milk fat percentage was calculated by an equation. 

This PR is probably also related this issue #1101. The filter `Cow.milking_update.milk_data_at_milk_update.milk_protein` is actually the milk protein percentage, which has always been a constant value.

---
Update: For verification purposes, I added several filters related to dry cows milk stats.


RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -405,7 +407,7 @@ def milking_update(self, sim_day, calving_interval):
             self.estimated_daily_milk_produced = max(0.0, estimated_daily_milk_produced)
             self.lactose_milk = AnimalModuleConstants.MILK_LACTOSE
             self.CP_milk = AnimalModuleConstants.MILK_CRUDE_PROTEIN
-            self.mPrt = AnimalModuleConstants.MILK_TRUE_PROTEIN
+            self.mPrt = self.get_user_defined_milk_protein_percent()
```
JoeWaddell, line 398: Should the `MILK_TRUE_PROTEIN` value should be removed from the constants file now?

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -415,11 +417,11 @@ def milking_update(self, sim_day, calving_interval):
 
         # calculate fat percent in milk and fat corrected milk production
         if self.milking:
-            self.fat_percent = self.calculate_fat_percent(self.days_in_milk)
+            self.fat_percent = self.get_user_defined_milk_fat_percent()
```
JoeWaddell, line 408: Should the `calculate_fat_percent` method be noted as deprecated with the intent that it would be re-implemented in the future?

output/output_filters/graph_milk_fat_and_protein.json
```diff
@@ -0,0 +1,28 @@
+{
```
JoeWaddell: I really appreciate the inclusion of the output filters! Since the variation here is due to dry cows being included in the total, should we include the change I made below (to confirm the dry cow thing) in this PR?
image

JoeWaddell (commented):
Nice clean implementation. Just some general comments/questions. As a more general question: are we OK to still add to the input JSONs, and if not, should we use the animal_constants file, instead?

output/output_filters/graph_milk_fat_and_protein.json
ltp23: I will let you decide on this, Joe. I can add it here since it is a small change.

RUFAS/routines/animal/life_cycle/cow.py
ltp23: I can remove it in this PR. Can you help me make a small git issue as a reminder to revisit this part? Thanks Joe.

RUFAS/routines/animal/life_cycle/cow.py
ltp23: Yes, I can remove this constant.

output/output_filters/graph_milk_fat_and_protein.json
ltp23: And yes it is helpful for me at least to look at some graphs during the review phase, but I will remove the filter files before merging.

input/data/animal/default_animal.json
```diff
@@ -23,7 +23,9 @@
       "heifer_repro_cull_time": 500,
       "do_not_breed_time": 185,
       "cull_milk_production": 30,
-      "cow_times_milked_per_day": 3
+      "cow_times_milked_per_day": 3,
+	    "milk_fat_percent": 3.6,
+	    "milk_protein_percent": 3.5
```
jadamchick: Let's do default of 3.2 for milk_protein_percent

input/metadata/default_metadata.json
```diff
@@ -313,6 +313,18 @@
...
+          "milk_fat_percent": {
+            "type": "number",
+            "description": "Milk Fat Percent (percent) -- The average or most common milk fat percentage in cow milk",
+            "default": 3.5,
+            "minimum": 0
+          },
+          "milk_protein_percent": {
+            "type": "number",
+            "description": "Milk Protein Percent (percent) -- The average or most common milk protein percentage in cow milk",
+            "default": 3.0,
```
jadamchick: default 3.2 for milk_protein_percent

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -415,11 +417,11 @@ def milking_update(self, sim_day, calving_interval):
...
             daily_fat_correct_milk_production = \
                 0.4 * estimated_daily_milk_produced + \
                 0.15 * self.fat_percent * estimated_daily_milk_produced
```
jadamchick, line 411: Just a note that we don't use this fat-corrected milk calcluation in RuFaS for anything; we report total milk and milk fat and milk protein so we can do the calculations (fat-corrected milk, energy-corrected milk, etc.) in post

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -324,9 +324,9 @@ def daily_update(
...
             cow.estimated_daily_milk_produced for cow in cows
         )
-        self.herd_milk_fat_kg = sum(cow.milk_fat_kg for cow in cows)
+        self.herd_milk_fat_kg = sum(cow.milk_fat_kg for cow in cows if cow.milking)
         self.herd_milk_fat_percent = self.herd_milk_fat_kg / self.daily_milk_production
-        self.herd_milk_protein_kg = sum(cow.milk_protein_kg for cow in cows)
+        self.herd_milk_protein_kg = sum(cow.milk_protein_kg for cow in cows if cow.milking)
         self.herd_milk_protein_percent = (
             self.herd_milk_protein_kg / self.daily_milk_production
         )
```
jadamchick: for herd_milk_fat_percent and herd_milk_fat_protein: need to multiply each * 100 to get a value in the same dimensions as we started with ("3.6%" instead of "0.036"): 

```
self.herd_milk_fat_kg = sum(cow.milk_fat_kg for cow in cows)
        self.herd_milk_fat_percent = self.herd_milk_fat_kg / self.daily_milk_production * 100 
        self.herd_milk_protein_kg = sum(cow.milk_protein_kg for cow in cows)
        self.herd_milk_protein_percent = (
            self.herd_milk_protein_kg / self.daily_milk_production
        ) * 100 
```


maybe these calculations (herd_milk_fat_percent & protein_percent) are redundant, since they are bringing us back to the input value that we specified in the animal.json... but it is a nice verification that all of the transformations along the way are working as expected.

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -415,11 +417,11 @@ def milking_update(self, sim_day, calving_interval):
...
-            self.milk_fat_kg = self.fat_percent*estimated_daily_milk_produced
+            self.milk_fat_kg = self.fat_percent * estimated_daily_milk_produced
             self.milk_protein_kg = self.mPrt * self.estimated_daily_milk_produced
```
jadamchick: need to divide each of these (for fat and for protein) by 100 to get a value in kg (since the input is "3.6", not "0.036")

```
self.milk_fat_kg = self.fat_percent*estimated_daily_milk_produced / 100             
self.milk_protein_kg = self.mPrt * self.estimated_daily_milk_produced / 100 
```

output/output_filters/graph_milk_fat_and_protein.json
jadamchick: I agree that herd milk protein should be the sum of milk protein for all milking cows -- but also that dry cows shouldn't have any value for milk protein kg (because their milk production should be zero?) . I tried to verify this but now that Cow.milking_update is only for lactation cows, I don't know how to check on milk production or milk protein production of dry cows :)

jadamchick (changes requested):
Thanks Loi! overall looks great, just a couple slips in transforming between "3.6%" and "0.036" that need to be addressed (see my comments). 

Just to summarize what is happening here -- I'm not sure it is actually spelled out anywhere, and maybe would be helpful? (I had to write this out in order to do the review): 

user input = herd milk fat (3.6%)
this gets assigned as each cow's milk fat percent
each cow's milk fat kg produced daily = cow milk production * cow milk fat percent (/100 if percent is expressed as 3.6)
herd milk fat kg produced daily = sum of cow milk fat kg over all cows (all cows milking, though dry cows should have values of 0)
herd milk fat % = herd milk fat kg / herd milk production (*100 if want this expressed as 3.6, not 0.036)

the calculated "herd milk fat %" should be the same as the user input value
(and ditto for protein)

tomhuhh:
As a general thought, also related to Julie and Joe's comments- do we want to completely rule out the possibility to calculate milk fat and protein using the equations that are associated with the outputs from the rest of RuFaS? Thus, from a design perspective, could we add a 'if' statement, so that if the user provides milk protein and fat components, we use them in the calculations; if not, we still go back to use the default milk protein values and milk fat equations? 

Additionally, do we want to add some randomness to the milk fat and protein even if we assume a constant value? By this I mean, if we know the standard deviation of milk fat and protein, can we randomize it for the milking cows in the herd. Theoretically it should not change the herd average, but this is where the stochastic part of the model comes in. I'm open to thoughts because I can also see the argument where this randomness is unnecessary compared to the rest of the model. 

Alternatively, as I don't have the full picture of it, if this is for the January deadline and for pilot testing, is it okay to only implement this logic in the pilot testing branch and make a note of it? If that is the case, the concerns that I have listed above can be ignored for the moment.  

Also, I would recommend for any important logic changes we make, we should try to document it in the relative design doc/pseudocode and make a note of it. 

Thanks!

jadamchick:
Great thoughts @tomhuhh -- and I think @KFosterReed is the person to speak to ultimate direction/vision. The ideal would definitely be to have milk fat and protein be dynamic based on cow and herd attributes changing day to day; however RuFaS has not had that capacity yet (the only equation in there was based on days in milk for fat, and protein was a fixed value). We figured this is the most expedient way to enable the milk fat + protein values to reflect herd-to-herd variation for now, with the intention to update those dynamic equations "someday". 

My vote is to keep the pilot testing branch as close to main branch as possible, and think that this could be implemented on both (since, using the current version, I've always had the herd value for milk fat come out to be 3.6% -- I think it could benefit from being able to modify that value?). But not my decision :) 

The question of randomness is a good one and I'll trust the experts on whether it really adds value at this point.

KFosterReed:
Thanks for the thoughtful comments all! My preference is to start (or re-start in this case) simple so let's go with only the option for static milk and fat based on user inputs. We can add it to the wish list for the Animal module and implement it in a more strategic way. (Also note, that the current equations for fat and protein are old and likely not representative anyway).

KFosterReed:
> Thanks for the thoughtful comments all! My preference is to start (or re-start in this case) simple so let's go with only the option for static milk and fat based on user inputs. We can add it to the wish list for the Animal module and implement it in a more strategic way. (Also note, that the current equations for fat and protein are old and likely not representative anyway).

I realize this may not have been clear - we can add the goal of having dynamic and stochastic fat and protein to the wish list :)

RUFAS/routines/animal/life_cycle/cow.py
ltp23: Right, I don't see that variable being used in any other calculations anywhere. I made a separate issue to remove it if we don't have plans to implement an improved version.

RUFAS/routines/animal/life_cycle/life_cycle.py
ltp23: I can multiply those by 100 and then in the life cycle file, I would divide the percentages by 100 before calculating amounts of milk fat and protein.

output/output_filters/graph_milk_fat_and_protein.json
ltp23: I added some filters to verify dry cows milk stats

```
LifeCycleManager.daily_update.dry_cows_daily_milk_production
LifeCycleManager.daily_update.dry_cows_milk_fat_kg
LifeCycleManager.daily_update.dry_cows_milk_protein_kg
```

RUFAS/routines/animal/life_cycle/cow.py
JoeWaddell: Opened an issue for this: https://github.com/RuminantFarmSystems/MASM/issues/1165

If you do wind up removing the calc in the PR, you could add the removed code to that Issue, to keep in case we want to restore it directly.

input/metadata/default_metadata.json
```diff
@@ -313,6 +313,18 @@
             "description": "Number of Milkings (per day) -- The average or most common number of times cows are milked per day (1, 2, or 3 times daily)",
             "default": 3,
             "minimum": 0
+          },
+          "milk_fat_percent": {
+            "type": "number",
+            "description": "Milk Fat Percent (percent) -- The average or most common milk fat percentage in cow milk",
+            "default": 3.5,
```
jadamchick, line 320: default for this should be 3.6 to match what is above (in the animal.json)

jadamchick (approved):
Looks great! runs smoothly, outputs match what I expect. The only quirk is that the graph seems to have an incorrect legend (attributing lots of kg of milk fat to dry cows), but I can see in the data that "LifeCycleManager.daily_update.dry_cows_milk_fat_kg" and      "LifeCycleManager.daily_update.dry_cows_milk_protein_kg" are all zeros so I'm confident that it's a graphing/labeling error. 

I would just reiterate that it's worthwhile to remove the calc_fat_percent from the code for now to avoid any confusion, and stash it in the issue for future reference.

JoeWaddell (approved):
LGTM: tested again, and since everyone is OK with the percentage format, this looks good to merge.

PR #1154, opened 2024-01-23
Update field module docstrings

harry-samuels:
This updates the class docstrings to Sphinx for the RUFAS/routines/field/field folder


RUFAS/routines/field/field/field_data.py
```diff
@@ -84,16 +90,6 @@ class FieldData:
...
-        """Initialize all attributes in FieldData object that need to be set based on other FieldData attributes.
-
-        Raises
-        ------
-        ValueError
-            If the watering amount is < 0.
-        ValueError
-            If the watering interval is < 0.
-
-        """
```
Isgotkowitz: For now I think the `__post_init__` docstring should be left where it is. This is a tough one because the NumPy docs style guide doesn't say anything about dataclasses or the post init method. But my reasoning for keeping the docstring here is that init is automatically generated by the dataclass, so we should let Python handle "documenting" that auto-generated init method however it likes and keep post init the same.

Also, for this docstring, could you please change the raises section to 
```
Raises
------
ValueError
    If the watering amount is < 0.
    If the watering interval is < 0.
```

Isgotkowitz (approved):
Looks good Harry. I left one comment that should be addressed, but otherwise the docs look really good. Thank you for taking this on!

Isgotkowitz (approved):

RUFAS/routines/field/field/tillage_application.py
```diff
@@ -4,21 +4,43 @@
...
-profile gets incorporated and/or mixed with the same logic.
-"""
-
 om = OutputManager()
 
 
 class TillageApplication:
+    """
+    This class contains all necessary methods for executing tillage operations on a field, based on SWAT Theoretical
+    documentation section 6:1.6 and the SurPhos plow.f file.
```
matthew7838, line 13: I think that we could move the based on .... part into a reference section

matthew7838 (approved):
LGTM, minor comments. Great job.

PR #1167, opened 2024-01-26
Separate planting and harvest skips in `CropSchedule`

### LINKED ISSUE #1169: Add planting skip and harvesting skip into `CropSchedule` instead of just having a single skip.
## Current State
There is one value in a crop schedule that determines how many years should pass between repetitions of a crop rotation in a field, called _pattern skip_.

A simple example: we want to plant and harvest in years 1, 2, and 3, and we want to repeat this rotation twice and skip two years between repetitions. `CropSchedule` would produce a full schedule that looks like this:
* planting years: 1, 2, 3, 5, 6, 7, 10, 11, 12
* harvesting years: 1, 2, 3, 5, 6, 7, 10, 11, 12

### What's the problem?
Say we have a stand of alfalfa, and we want to plant it in year 1, harvest it in years 1, 2 and 3 (only killing it in year 3), repeat this rotation twice and skip three years between repetitions of this rotation. What we want looks like this:
* planting years: 1, 7, 13
* harvesting years: 1, 2, 3, 7, 8, 9, 13, 14, 15

but what `CropSchedule` currently produces is a schedule that looks like this:
* planting years: 1, 5, 9
* harvesting years: 1, 2, 3, 7, 8, 9, 13, 14, 15

## Solution
Have separate inputs for the skip between planting repetitions and harvesting repetitions.
In the first example given above, instead of passing a skip of 2, 2 would be passed as both the planting and harvesting skip.
In the alfalfa example given above, the planting skip would be 5 and the harvesting skip would remain 3.

There are many ways to avoid error cases like the one demonstrated in the alfalfa example, but this requires modifying the internal logic of `CropSchedule` - which in turn results in more edge cases that produce incorrect crop schedules. Modifying the inputs is more robust and results in users being able to create more complicated crop schedules without having the logic of `CropSchedule` fail to produce the crop rotations they want.

Isgotkowitz:
This PR changes `CropSchedule` so that the skip between planting schedule repetitions and harvesting schedule repetitions can be differentiated.

While testing the available scenarios it became apparent that `no_animal.json` was a not a valid animal input file anymore. I have copied in the `default_animal.json` file and changed all the numbers of cows to 0.

Also, while reviewing the code, I noticed that the default harvest operation was not a valid option. I have changed it from "harvest" to "harvest_kill" to fix this.

Closes issue #1169. Please read this issue to understand the impact of changes made in this PR.


allisterakun (approved):
Straight-forward PR! Good job, Ed!

krb78 (approved):
Love it! Can't wait to use it! 🌽 👍

Issue #1129, opened 2024-01-19
Open Lots management configs need to be added to inputs


KFosterReed:
**Is your feature request related to a problem? Please describe.**
Open Lots are an option for the manure treatment method but there is no manure treatment config for this option in the manure input json file. In addition, the manure_handler recommended for the open lot treatment type is harrowing but there is not a manure handler config for this either. 

**Describe the solution you'd like**
We should have a manure handler config input for 'harrowing' and a manure treatment config for 'open lots' in the metadata and manure_management.json. 

The default configs for the harrowing configs are: 

` {
      "manure_handler_type": "harrowing",
      "daily_tillage_frequency": 0,
      "cleaning_water_recycle_fraction": 0.0
    }`

The default configs for the open lot are: 

`{"manure_treatment_type": "open lots",
      "total_solids_removal_efficiency_for_treatment": 0,
      "nitrogen_removal_efficiency_for_treatment": 0.0,
      "total_ammoniacal_nitrogen_removal_efficiency_for_treatment": 0.0,
      "phosphorus_removal_efficiency_for_treatment": 0.0,
      "potassium_removal_efficiency_for_treatment": 0.0,
      "volatile_solids_removal_efficiency_for_treatment": 0,
      "hydraulic_retention_time": 0,
      "sludge_accumulation_period": 0,
      "sludge_accumulation_volume_fraction": 0,
      "storage_time_period": 365,
      "freeboard_input": 0
    }`

**Describe alternatives you've considered**
We could leave these without explicit configs because the inputs are mostly 0 but having the configs in there is a lot more clear for users and supports future efforts to remove the default configs from the manure module.

**Additional context**

PR #1180, opened 2024-01-29
Adding manure module configs for open lots and harrowing

ltp23:
Issue #1129 

This PR does the following:

- Add a section in `manure_management.json` for the harrowing manure handler type. All the parameters are set to 0.
- Add a section in `manure_management.json` for the open lots treatment. I only kept the removal efficiency parameters and set most of them to 0 except the two for phosphorus and potassium. This is because I saw only these two removal efficiencies were used in the calculations. However, I used very small numbers for theses two (0.01) because I didn't know what values to use.
- Add descriptions in the metadata files for existing properties of manure treatment configs.


input/data/animal/default_animal.json
```diff
@@ -179,7 +179,7 @@
       "housing_type": "open air barn",
       "pen_type": "tiestall",
       "max_stocking_density": 1.2,
-      "manure_management_scenario_id": 2
+      "manure_management_scenario_id": 8
```
ew3361zh: Are these meant to replace the standard manure_management_scenario_ids used?

ew3361zh:
@ltp23 can you add what issue this is addressing to the description?

input/data/animal/default_animal.json
tomhuhh: My understanding is that those are temporarily used for reviewing purposes.

tomhuhh (commented):
LGTM. Thanks Loi! 

The code runs without difficulty for me, which means the added manure handling and storage methods work.  

The added descriptions in the metadata file seem to be fine. 

The only thing that I am uncertain about is the removal efficiency of phosphorus and potassium. I'm not sure if there is a design doc or  pseudocode doc that we can refer to. 

Thanks!

tomhuhh:
I would like to add that for manure handling methods such as harrowing and tillage, there is currently no calculation code available. I wonder if this is intentional? Thanks! 

Screenshot 2024-01-30 at 4 54 07 PM

input/data/animal/default_animal.json
ltp23: Right, they are there for reviewing purposes.

ltp23:
> The only thing that I am uncertain about is the removal efficiency of phosphorus and potassium. I'm not sure if there is a design doc or pseudocode doc that we can refer to.

Thanks Haowen. I don't have the design doc to see what the reference values are. I put in the small values for those two just to signify that those two are used in the calculations.

ltp23:
> I would like to add that for manure handling methods such as harrowing and tillage, there is currently no calculation code available. I wonder if this is intentional? Thanks!

Right, Haowen. All the manure handler classes inherit the base class and perform the same calculations, though the numbers can vary. These subclasses are empty currently because we don't have specific logic yet for each kind other than the calculations in the daily update function, which is done in the base class.

KFosterReed:
> > The only thing that I am uncertain about is the removal efficiency of phosphorus and potassium. I'm not sure if there is a design doc or pseudocode doc that we can refer to.
> 
> Thanks Haowen. I don't have the design doc to see what the reference values are. I put in the small values for those two just to signify that those two are used in the calculations.

Yes, these removal efficiencies should all be 0

tomhuhh (approved):
Thanks all for the clarifications!

ew3361zh (approved):
LGTM!

Issue #1090, opened 2024-01-08
Determine if "example_manure_management.json" file is needed.


ew3361zh:
The input file "example_manure_management.json" file in the input > data > manure folder may be unnecessary and can be removed. 

I believe I created it when we initially were looking at a different format for the manure management scenarios input files but the model wasn't set up to handle that format and it caused more problems than it solved.

I don't know if we're still looking at setting up the input files in a different way and if this file is useful to anyone to that end (or for any other reason).

Determine if this file is needed for any reason by anyone and if it needs to be in the codebase. If not, remove it.

allisterakun:
Confirm with the Subject Matter Experts about this file.

KFosterReed:
Unless @ltp23 has a reason to keep it, let's delete it

ew3361zh:
Spoke with Loi, he said the file is not needed by him. I will remove it.

PR #1182, opened 2024-01-30
Create multiple separators manure scenario option

### LINKED ISSUE #1103: Option to separate manure before anaerobic digestion and anaerobic lagoon
Currently there is a manure treatment option called anaerobic_digestion_and_lagoon which allows the user to specify that they send manure to anaerobic digestion and then after the digestion is finished, it goes into an anaerobic lagoon. If a manure separator is specified, the separation will occur between the digestor and the lagoon. We need to be able to have an option to have manure separation before the digestion AND after the digestion/before the lagoon

I would like another input to be added so that the user can specify multiple separators if needed and to have the inputs to the anaerobic_digestion_and_lagoon treatment type to have the option to come directly from the reception pit or from the manure separator. 

One alternative is to have the input for the 1st separation to be a boolean input and to have only one fixed option for the type of separation before digestion in the compound treatment case.
**Additional context**

ew3361zh:
PR to address issue #1103 to create the option for a manure scenario where there is manure separation before anaerobic digestion and then another manure separation after digestion but before anaerobic lagoon.

To accomplish this, the PR:

1. Creates a new input for `manure_separator_after_digestion` and adds that field to each of the current manure scenarios.
2. Adds a new manure scenario (10) which enacts both a `manure_separator` and a `manure_separator_after_digestion`.
3. Updates the metadatas accordingly.
4. Updates `AnimalManager` and `Pen` classes which initialize manure scenario data from InputManager.
5. Updates remaining `ANAEROBIC_DIGESTION_AND_LAGOON_WITH_SPLIT` references to the new agreed-upon `ANAEROBIC_DIGESTION_AND_LAGOON_WITH_SEPARATOR`.
6. Adds new attribute `manure_separators_after_digestion` to ManureManager and updates the class functions to handle its configuration within ManureManager.
7. Adds daily results of this new post-digestion separator to OM via ManureModuleOutputManagerHelper.
8. Updates `BaseManureTreatment` and `AnaerobicDIgestionAndLagoon` manure treatment classes to coordinate manure separator daily output according to whether this second separator is present.
9. Updates unit testing.


RUFAS/routines/manure/manure_manager.py
```diff
@@ -479,6 +493,18 @@ def _pen_daily_update(self, simulation_day: int, pen) -> None:
                 excluded_fields,
             )
 
+        if manure_separator_after_digestion_daily_output:
+            manure_separator_after_digestion_daily_output_prefix = {
+                "prefix": (manure_separator_after_digestion_daily_output.__class__.__name__
+                           + "_Pen_" + str(mm_pen.id)
+                           + "_" + mm_pen.animal_combination.name)
```
Isgotkowitz: Instead of using string concatenation, I think it would better to use f-strings here.

input/data/manure/manure_management.json
Isgotkowitz, line 1: Should `example_manure_management.json`, or is that file totally stale at this point?

Isgotkowitz (approved):
LG8TM. Two very minor comments. Tried running the default animal scenario with all pens using a manure management scenario with a separator after digestion and it seemed to run fine. Not sure where to begin with checking the outputs to verify that everything is working alright though.

RUFAS/routines/manure/manure_manager.py
```diff
@@ -633,13 +664,21 @@ def _handle_daily_update_for_compound_anaerobic_manure_treatment(
...
+            self.manure_separators[pen.id].daily_update(
+                manure_separator_daily_input=reception_pit_daily_output
+            )
+            if self.manure_separators[pen.id]
+            else None
+        )
         # Currently, calling the daily_update on the treatment only returns one final output
         manure_treatment_daily_output = self.manure_treatments[pen.id].daily_update(
             manure_handler_daily_output=manure_handler_daily_output,
             manure_treatment_daily_input=reception_pit_daily_output,
```
KFosterReed: I haven't started looking at the outputs yet but, looking at the code, I thought that what is passed into the `manure_treatment_daily_input` should be similar to the `handle_daily_updated_for_simple_manure_treatment`. Specifically, in that function, it pass in `receiption_pit_daily_output or separator_daily_output`. Is that possible here? I don't really understand how the code works with 'or' operation but I thin Loi can explain it if needed

input/data/manure/manure_management.json
ew3361zh: I opened an issue for investigating that a little while back #1090

RUFAS/routines/manure/manure_manager.py
ew3361zh: I can make that change, thanks Ed

RUFAS/routines/manure/manure_manager.py
ew3361zh: You may be right. I think that might be leftover from the previous iteration which didn't have the separator before digestion. I can make that update. The or functions to check the first value first and if that's zero or None, it will use the second value listed.

KFosterReed:
I am probably missing something but when I tried to run this calling manure management scenarios 7 and 10, I did not see an output for the separation after digestion. I tried to attach the variable names output file below to show what I am getting. Can you please point me in the right direction so I can review the outputs? 

Thanks!

[SLS_testing_variable_names_30-Jan-2024_Tue_20-21-40.txt](https://github.com/RuminantFarmSystems/MASM/files/14106013/SLS_testing_variable_names_30-Jan-2024_Tue_20-21-40.txt)

ew3361zh:
> I am probably missing something but when I tried to run this calling manure management scenarios 7 and 10, I did not see an output for the separation after digestion. I tried to attach the variable names output file below to show what I am getting. Can you please point me in the right direction so I can review the outputs?
> 
> Thanks!
> 
> [SLS_testing_variable_names_30-Jan-2024_Tue_20-21-40.txt](https://github.com/RuminantFarmSystems/MASM/files/14106013/SLS_testing_variable_names_30-Jan-2024_Tue_20-21-40.txt)

I think it has to do with how data is added to OM from the manure module via the ManureModuleOutputManagerHelper. I am working on a way to help distinguish the results a little better for analysis but I think the whole process of how data is added to OM from the manure module should be looked at. I will look at a temporary solution for this and update as soon as I find one. Should be soon.

ew3361zh:
@KFosterReed I added a special prefix for the data coming from the separator after digestion (ManureSeparatorAfterDigestionDailyOutput) so it will be distinct from the first separator which has the ManureSeparatorDailyOutput prefix:

Screenshot 2024-01-30 at 9 21 21 PM

Let me know if this doesn't do the trick!

KFosterReed (approved):
Fantastic work Niko! We will need more SME review of the specific partitioning of the nutrients during the manure treatment and separation phases but this works as expected in that it gives us the option to separate before digestion, after digestion, or both! 

Thank you!!

PR #1186, opened 2024-01-30
Simplify metadata

PooyaHekmati:
This PR simplifies the metadata by deduplicating the properties blob. The properties are now moved to their dedicated files under _input/metadata/properties_. A new **mandatory** entry has been added to the files blob to point to the properties file. After loading the metadata, IM attempts to load the properties. Once successful, it stores the properties details in `__properties_used` and deletes the properties entry from its files blob to prevent reloading the properties when populating the pool.
* Updated all metadata files on main. 
* Added `default.json` for properties. All metadata files are currently using this file, we can add new files if needs be.
* This is a temporary solution until the scenario manager comes online.


RUFAS/input_manager.py
```diff
@@ -79,6 +80,48 @@ def _load_metadata(self, metadata_path: str) -> None:
         except Exception as e:
             raise e
 
+    def _load_properties(self) -> None:
+        """
+        Loads properties data from a specified JSON file and updates the metadata.
+
+        This method reads the properties file path from the metadata, checks if the file exists, and then loads the
+        properties into the metadata. The original properties data in the metadata is first copied to a separate
+        attribute for future reference and then removed from the metadata files section.
```
Isgotkowitz, line 89:
```suggestion
        Notes
        -----
        This method reads the properties file path from the metadata, checks if the file exists, and then loads the
        properties into the metadata. The original properties data in the metadata is first copied to a separate
        attribute for future reference and then removed from the metadata files section.
```

Isgotkowitz (approved):
One comment on a docstring, otherwise LGTM, thank you for taking this on Pooya. Tried running all the simulations and after updating the `no_animal.json` structure to match the properties, all simulations ran well.

Could there be a better name for the properties than "default", maybe "properties_v1"? To me, default implies that it is just one option among many when this is not the case.

jadamchick (approved):
This is great! 
I was able to make another properties .json that copies the metadata used for pilot testing (identical to default but w/o the field + soil sections for now), and deleted the properties section from a pilot farm metadata.json, instead referencing the new properties file. Works great and appreciate the simplification.

ltp23 (approved):
This is great. It would help a lot with avoiding the duplication the metadata properties. I did put breakpoints here and there to see how it worked.  One thing I noticed was the removal of the `properties` object and saving it into the variable `__properties_used`.

Thanks Pooya.

ew3361zh (approved):
very nice, really appreciate these changes having just updated 7 different metadata properties files multiple times this past week. LGTM!

PooyaHekmati:
> Could there be a better name for the properties than "default", maybe "properties_v1"? To me, default implies that it is just one option among many when this is not the case.

That actually can be the case (although not desired), just not the case here in this case.

RUFAS/input_manager.py
PooyaHekmati: I don't think that's a note, it is the continuation of the docstring.

PR #1188, opened 2024-01-31
Update feeds library + c1f1 files

jadamchick:
1. update NASEM_Comp_with_TDN.csv so that it includes feeds 301-306 (restoring original state before updating to match what is on Main). 
2. update c1f1 animal.json and metadata.json files to be able to run (to ensure feed library is working correctly)


tomhuhh (approved):
Changes LGTM! The code runs with no difficulty.

PR #1100, opened 2024-01-10
Feed inputs metadata initial updates

elle-andreen:
This PR is part of the metadata review/updates and covers the feed inputs portion of the metadata (feed_properties, user_feeds_properties, NRC_Comp_properties, and NASEM_Comp_properties). 

Primary objectives are:
- Ensure descriptions are present (where necessary) and clearly describe input being parameterized
- Ensure units are present and correct
- Confirm min/max/default values are correct
- Confirm patterns are correct
- Address any other Feed Inputs issues noted by the team (see linked issues)

PR #1191, opened 2024-01-31
Update feed range values in metadata properties 

elle-andreen:
This is a mini version of PR #1100 that contains only the changes pertaining to feed constituent default/max/min values. This PR was created to, 1. enable a quick merge of the simple range value changes, while more complicated issues with the feed module metadata can be resolved in due time, and 2. circumvent the recent change in the structure of the metadata/properties/config files. 

Most feeds constituents (e.g., crude protein, neutral detergent fiber, calcium, etc.) are described on a percentage basis, either % dry matter or % of some other constituent. Multiple feeds expressed on a percentage basis had ranges that were too restrictive, i.e., not min=0 max=100. 

This PR:

1. Sets the range of all feeds expressed on a percentage basis to 0-100. 
2. Addressed a few incorrect or out-of-range default values. 
3. Fixes a few minor description issues.


jadamchick (changes requested):
Hi Elle -- this looks great! However, it won't run (even using all of the defaults). It gives an error that       `"Variable: 'CP.111' has invalid value: 134.1, and cannot be changed to a default value. Violates properties defined in metadata properties section 'NASEM_Comp_properties'."
`

When I sort the NASEM_Comp_with_TDN by CP values, it looks like feeds 244, 245, and 246 have CP values > 100. Based on my minimal nutrition knowledge, I think these are accurate (something to do with the calculation from amt of nitrogen). 
So the CP max will need to be increased, but probably also worth a look from Edward or someone with more nutrition savvy to make sure the rest of the % values are actually 0-100? (counter-intuitive, I know). 


Also, I'm not sure why this error is now stopping the simulation -- i.e., why it "cannot be changed to a default value".

JoeWaddell:
> Also, I'm not sure why this error is now stopping the simulation -- i.e., why it "cannot be changed to a default value".

This is because those are values without a default assigned: locally I added a default field, and it will now run. 
image

@jadamchick if you get a chance to try adding the (temporary) default field (since I think we won't want to assume a default value for such an important measure) on the pilot farms you ran into problems with, it should give us a better idea if this will solve all the related issues you ran into.

Also tagging @Ed731551 so he can chime in about why the crude protein max (and potentially others?) would have NASEM values over 100%

elle-andreen:
Thank you for bringing that to my attention, Julie! And for the temporary workaround Joe! 

Finally, a technical issue I actually have some subject-matter knowledge on 😂 You are correct in your line of thinking about crude protein - since most labs these days actually measure N content and multiply % N by 6.25, technically you can have CP values > 100%. I changed the maximum to 625 for CP, NDICP, and ADICP to accommodate a theoretical feed with 100% N. 

The only other constituent I can see that could have a maximum value > 100 would be kd (rate of ruminal degradation). I don't have the NASEM in front of me, but this value is usually expressed on a % DM per **hour** basis. So, if your ingredient degrades completely in the rumen in 30 minutes, your kd would be 200% per hour. I think we should either remove the maximum for this value, or make it something quite large, like 10,000. @Ed731551, what are your thoughts on that? As Joe said, are there any others in the metadata I may have missed that we should reconsider?

jadamchick:
I made a temporary fix of increasing the defaults for: CP, NDICP, ADICP, kd (as Elle suggested) (worth noting -- some of these had to be updated in both the NRC and NASEM sections). But now, just running the defaults, I get warnings about TDN- 

```
"Invalid data fixed: 'TDN.17' value changed from 127.36 to 0. Fix enabled by default value specified in 'NRC_Comp_properties'.",
      "Invalid data fixed: 'TDN.19' value changed from 102.67 to 0. Fix enabled by default value specified in 'NRC_Comp_properties'.",
      "Invalid data fixed: 'TDN.39' value changed from 163.51 to 0. Fix enabled by default value specified in 'NRC_Comp_properties'.",
      "Invalid data fixed: 'TDN.40' value changed from 176.33 to 0. Fix enabled by default value specified in 'NRC_Comp_properties'.",
      "Invalid data fixed: 'TDN.42' value changed from 147.4 to 0. Fix enabled by default value specified in 'NRC_Comp_properties'.",
      "Invalid data fixed: 'TDN.43' value changed from 183.97 to 0. Fix enabled by default value specified in 'NRC_Comp_properties'.",
      "Invalid data fixed: 'TDN.95' value changed from 101.01 to 0. Fix enabled by default value specified in 'NRC_Comp_properties'.",
      "Invalid data fixed: 'TDN.99' value changed from 122.34 to 0. Fix enabled by default value specified in 'NRC_Comp_properties'."
```


I'll run it with one of the pilot farms and the regional diets to see if anything else flags right off the bat. Stay tuned!

Ed731551:
Agree with @elle-andreen. Some feeds can have CP greater than 100% as calculation is N x 6.25 (urea for example).
Regarding second point raised, I would favor to remove the maximum values. 
On TDN values, I had a look on the NRC 2001 book and yes, some TDN tabulated values (e.g. some fat products) can be more than 100% assuming a high digestibility of those feeds at the maintenance level of intake (i.e., vegetable oil; TDN = 184 %). Hope that helps. Thank you.

jadamchick:
Having made the following updates locally (some had to updated in multiple sections of the metadata properties): 
- CP max to 625; added default of 0
- NDICP max to 625
- ADICP max to 625
- TDN max to 184
- KD max to 10000

The default feed file, first pilot farm (c1f1), and all of the default regional diets run w/o any warnings along these lines.

KFosterReed:
Hey team, where are we with this PR?

JoeWaddell:
> Hey team, where are we with this PR?

My instinct is that we could implement the maximum that Julie is using, since those have gone through the most diverse set of feed tests so far, and would be consistent with what Edward advised, as well.

elle-andreen:
I've updated the following ranges to:

CP max to 625
NDICP max to 625
ADICP max to 625
TDN max to 625
No maximum for Kd of N B2 fraction

@jadamchick, would you mind running this on a feed file or two to confirm you're not receiving warnings?

jadamchick:
> I've updated the following ranges to:
> 
> CP max to 625 NDICP max to 625 ADICP max to 625 TDN max to 625 No maximum for Kd of N B2 fraction
> 
> @jadamchick, would you mind running this on a feed file or two to confirm you're not receiving warnings?

Yep, will do today

input/metadata/properties/default.json
```diff
@@ -3089,51 +3087,51 @@
     },
     "ADICP": {
       "type": "number",
-      "description": "Acid detergent insoluble nitrogen multiplied by 6.25 (% DM).",
+      "description": "ADICP (% DM) -- Acid detergent insoluble nitrogen multiplied by 6.25.",
```
JoeWaddell, line 3090: Super-minor formatting quirk: is there a specific reason that (%DM) was moved to the beginning of the description (unlike the rest), or for the `--` instead of a single `-`?

JoeWaddell (approved):
LGTM, only the most minor of minor formatting comments.

jadamchick (approved):
Runs well on a handful of farms without any warnings!

input/metadata/properties/default.json
elle-andreen: It's a residual of the original PR covering the entire feed module! Not anything specific to this variable. The intention eventually is to standardize the location of the units and use a standard separator (i.e. "--") between the title and description, to hopefully utilize in Ed's script to convert the metadata properties to the DCA. But it's not a top priority right now, hence those changes haven't been applied uniformly. We can revert those changes here if we want.

PR #1193, opened 2024-02-01
Removes extraneous unused manure input file

### LINKED ISSUE #1090: Determine if "example_manure_management.json" file is needed.

ew3361zh:
PR to address issue #1090. This manure management input file is not used by anyone and can be removed.


allisterakun (approved):
LGTM

Isgotkowitz (approved):

Issue #957, opened 2023-11-14
Handle Missing Input File Data


ew3361zh:
Currently if there are fields present in an input file that are not in the metadata, the Input Manager will filter these excess fields out prior to adding the rest of the data from the input file to the Input Manager pool.

However, in the reverse situation, where an input file is missing fields that the metadata is expecting, Input Manager will crash and data validation will stop (and no scenario will be run).

We would like this to be better handled by Input Manager whether that is adding the default value present in the metadata if applicable or perhaps adding an error to the Output Manager and aborting the validation process or via another solution.

PR #1201, opened 2024-02-01
Handling missing keys in input data

ltp23:
#957 

This PR focuses on checking if user input json files have all the properties specified by the metadata files. Missing properties can be fixed by using the default values if they are given. For properties without default values and not provided by the user, they will recorded in the OM and the program will terminate early. 

Also note that the replacement with default values doesn't mean that those default values are valid. The real validation will be handled in a subsequent step. But first, we need to make sure the data is there before it can be validated.

The logic for handling nested scenarios can be confusing.


RUFAS/input_manager.py
```diff
@@ -280,6 +289,216 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
 
         return filtered_input_data
 
+    def _log_missing_keys_and_keys_with_default_values(
```
```suggestion
    def _log_missing_keys(
```
I think shortening the name better reflects that this function is single purpose.

RUFAS/input_manager.py
```diff
@@ -280,6 +289,216 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+        Notes
+        -----
+        This method is designed to check that all necessary data is present within a given input structure,
+        applying defaults where explicit values are not provided. It supports deeply nested structures through
+        recursion, handling both objects and arrays. It is very important to note that the method does not
+        validate if the default values are semantically correct. It only checks if the input data is complete
+        and fills in missing values with defaults where necessary.
+
+        Implementation Details
+        ----------------------
```
```suggestion

```
I think this should just be a separate paragraph in the "Notes" section.

RUFAS/input_manager.py
```diff
@@ -280,6 +289,216 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+
+        Notes
+        -----
+        Handling default values for arrays can be tricky due to their potential to nest and contain various
+        types of elements (objects, other arrays, or simple types). This method specifically addresses these challenges,
+        checking that each element within the array is appropriately processed according to its type,
+        as defined in the metadata properties.
+
+        Implementation Details
+        ----------------------
```
```suggestion

```

RUFAS/input_manager.py
```diff
@@ -280,6 +289,216 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+                if 'default' in property_details:
+                    input_data[property_key] = property_details['default']
+                    property_keys_with_default_values.append((property_key, property_details['default']))
+                else:
+                    missing_required_property_keys.append(property_key)
+                    continue
+
+            # Even if the property key is present or fulfilled with a default value,
+            # we still need to check if it's a nested object/array and add default values
+            # to missing properties in the nested object/array
```
```suggestion
```

RUFAS/input_manager.py
```diff
@@ -280,6 +289,216 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+                         f"Missing required property keys: {', '.join(missing_required_property_keys)}.",
+                         info_map)
+        if property_keys_with_default_values:
+            stringified_default_values = [f"({key} => {value})" for key, value in property_keys_with_default_values]
+            om.add_warning("Validation: missing required property keys",
+                           f"Default values used for required property keys that were missing: "
+                           f"{', '.join(stringified_default_values)}.",
+                           info_map)
+
+    def _add_default_values_to_missing_properties(
```
```suggestion
    def _add_default_values_for_missing_inputs(
```
What is missing is not properties but inputs.

input/metadata/properties/default.json
```diff
@@ -3472,7 +3482,7 @@
           "description": "Manure Separator Type -- The type of solid-liquid separator equipment to separate coarse fibrous solids/sand",
           "pattern": "^(screw press|rotary screen|none)$"
         },
-        "manure_treatment_methods": {
+        "manure_treatment": {
```
Isgotkowitz, line 3485: Why does this need to be changed?

RUFAS/input_manager.py
```diff
@@ -280,6 +289,216 @@ def _filter_input_data_by_metadata(self, input_data: Dict[str, Any],
...
+                        input_data,
+                        property_key,
+                        property_details,
+                    )
+                missing_required_property_keys.extend(nested_missing_required_property_keys)
+                property_keys_with_default_values.extend(nested_property_keys_with_default_values)
+
+        return input_data, missing_required_property_keys, property_keys_with_default_values
+
+    def _add_default_values_to_array_properties(
```
```suggestion
    def _add_default_values_to_array_inputs(
```
Same as above comment. The array properties already have defaults. It is the array inputs that need defaults added to them.

Isgotkowitz (commented):
I think this approach is acceptable only as a temporary solution. IM already has the code to recursively traverse the inputs, which it does. This solution duplicates the code to traverse the inputs, and uses it to traverse the inputs a second time. I have not delved too deeply into this issue yet, but I think a more permanent solution would be to modify `_fix_data()` so that it can handle data that is _missing_, not just data that is wrong.

I left a bunch of comments on the code, but most are only relevant if the original solution is pursued.

ltp23:
> I think this approach is acceptable only as a temporary solution. IM already has the code to recursively traverse the inputs, which it does. This solution duplicates the code to traverse the inputs, and uses it to traverse the inputs a second time. I have not delved too deeply into this issue yet, but I think a more permanent solution would be to modify `_fix_data()` so that it can handle data that is _missing_, not just data that is wrong.

Currently, we can handle recursive traversal of objects okay as long as there are no arrays along any paths. The fix_data() method simply replaces faulty values with default values, but these default values need to be validated also.

From what I can see, there are at least several steps that need a complete traversal of input data and metadata specs in tandem: checking if all the required properties are there, filtering out redundant input data that are not in metadata specs, checking if each property is valid (correct type, correct range of values, or correct option), and counting elements. Ideally, I think I would want to have the traversal mechanism separate from the task, whatever it is, that needs to be performed along the way (passing in the task as a function parameter or something like that). 

Incrementally, we would want to save on the number of times (a constant multiple, anyway) we need to do the traversal by combining tasks, but it shouldn't be a time-limiting factor. This is because of the likely shallow depth of the input data structure, which basically makes it a linear scan, and the data is already loaded only once and is relatively small (probably under 100k). At this point, I would emphasize more on the functionality that we can validate and count elements of all data types.

When handling the default values, there are several possible scenarios:
- We have a missing property so we create it with a given default value. The default values need to be validated at the point of usage or beforehand.
- We already have a property but it is not valid, so we want to replace the existing associated value with a given default value. Similarly, the default values need to be validated at the point of usage or beforehand.
- We can validate all the default values first so we know if we ever need to use them, they are already valid. But this may be a waste if we don't need to use them at all. Though, I think validating all default values beforehand has more upsides than downsides.

Overall, I'm okay with moving forward with this PR for two reasons: the work is isolated to those helper methods and can be removed later on, and it does handle replacing the missing properties with default values for all the data types.

JoeWaddell (approved):
I've been testing this and it works nicely! LGTM, with the caveat that I'm keen to see responses/changes following Ed's more specific comments.

RUFAS/input_manager.py
Isgotkowitz: Two more suggestions for this method:
1. I think it would be better to put its functionality into `_add_default_values_to_missing_properties()` and `_add_default_values_to_array_properties()`.
2. Each missing value should get it's own warning/error statement. When a bunch of information about multiple missing variables is compressed into one statement it makes it much harder to read.

Isgotkowitz (approved):
Works well, nice job Loi. Had some suggestions on `_log_missing_keys_and_default_values()` but they are take it or leave it. Also please address unresolved comments, but whether to implement them is up to you.

RUFAS/input_manager.py
ltp23: Thanks Ed. 

I will make changes for #2. I think for #1, my current preference is to make those functions unaware of how we want to handle the missing keys. I guess you want to log the missing keys right away so we don't have to return them? This approach is okay too but we will lose access to those missing keys later on, though I don't have immediate use for them besides writing them to logs.

input/metadata/properties/default.json
ltp23: Because this is the name of the property used in the `manure_management_scenarios` array in `manure_management.json`. It's been uncaught until now without checking array input data.

RUFAS/input_manager.py
Isgotkowitz: > my current preference is to make those functions unaware of how we want to handle the missing keys

How are `_add_default_values_to_missing_properties()` and `_add_default_values_to_array_properties()` supposed to be unaware of how missing keys are handled when they are the very methods handling missing keys?

ltp23: Right, those two methods add defaults to missing inputs in-place (a side-effect) and I want to separate that process from the way we handle the information about the missing keys like logging, which is another side effect.

Another point is whether we want to log the missing keys as they are processed or in one go at the end. They are functionally equivalent because there is no intervening step while this whole handling of missing keys is going on.

PR #1202, opened 2024-02-01
Removing a slow test case in test calc total manure

ltp23:
As the title says, one of the test cases for the `calc_total_manure` unit test uses too many mocked animals that slow down the test quite a bit. It is okay to do so because this test case doesn't improve the testing of the logic of the method that hasn't been accomplished by other test cases.


ew3361zh (approved):
Excellent, thanks Loi. This sped up running the test suite quite a bit. I agree I don't think we need to mock that many animals in this unit test.

JoeWaddell (approved):
LGTM, nice cleanup all round!

PR #1223, opened 2024-02-05
Adding lower DMI limit to heifer ration formulation

JoeWaddell:
Adds the lower DMI bound to the ration formulation process for heifers. This PR actually removes the existing removal, as it uses the same constraint as defined for lactating cows, which itself is defined as 20% (as currently defined in the animal module constants file) below the estimated DMI intake. 

Prior to this change DMI in pen 2 (when running the current default simulation) would drop as low as 6 kg per animal, and after this change it never drops below 12kg per animal.


tomhuhh (approved):
LGTM! Thanks Joe for the clarification!

PooyaHekmati (approved):

PR #1225, opened 2024-02-05
removal of feed quality file and associated methods

JoeWaddell:
Removal of calls to the deprecated feed_quality.csv, as well as the associated methods.


PooyaHekmati (commented):
lgtm
we probably need to remove feed_quality from other metadata files as well.

KevinP-B (approved):
Removal of calls to deprecated feed table. Looks solid! Thanks!

PooyaHekmati (approved):

PR #1238, opened 2024-02-06
Unpacking dictionary variables into columns in report generator

### LINKED ISSUE #1181: Make Report Generator aggregate column information clearer
From @jadamchick:

I've been using the new ability to generate report data selecting variables by regex, and it's great.

One question for you: when I want to report the data without any aggregation, it generates a report column for each variable in the dictionary and the variable name (for example: feed ID) is included in the column header. However, when I want to aggregate (for example, sum over the last year for each variable), it prints the results all in one column with no indication of the variable names.

Here are the filters I'm using (and I've already done it with individual feeds and pens to feel confident that it is selecting + aggregating the way I expect it to):

```
{
      "name": "DMI by feed, daily",
      "filters": [
      ],
      "variables": [".*"],
      "slice_start": -365
    },
    {
      "name": "DMI by feed, sum",
      "filters": [
      ],
      "variables": [".*"],
      "slice_start": -365
    },
    {
      "name": "Feed emissions by feed, daily",
      "filters": [
        "AnimalModuleReporter.report_daily_feed_emissions.*"
      ],
      "variables": [".*"],
      "slice_start": -365
    },
    {
      "name": "Feed emissions by feed, sum",
      "filters": [
        "AnimalModuleReporter.report_daily_feed_emissions.*"
      ],
      "variables": [".*"],
      "slice_start": -365
    },
```
And here are what the outputs look like for the two "sum" reports (there are 2 feeds for which we don't have emissions factors, which is why they are different lengths -- and which is part of the value of being able to summarize it like this!):
image (3)

## Potential Solutions (1 of the 3):
Option 3 sounds like it's the most preferable.
1. Print a neighboring column that lists the variable names, one in each row
2. Include all of the variable names in the column header of the aggregated data?
3. Have the aggregated values to each print in their own column, with that feed ID (variable name) included in the column header.

ltp23:
Closes #1181 

This PR aims to unpack the aggregated values from dictionary variables and spread them out in their separate columns. I try to maintain the existing functionality except for those cases that involve the use of `variables: [...]`, which may need some modifications if cross-referencing has been used.

I included two temporary example filter files provided by Julie and Haowen to test this out.


jadamchick (commented):
This is looking great! It does exactly as requested: reports each aggregated variable in its own column with its own label. 
I did have to update some of my filters that used cross-referencing with the new variable names, but that's an easy update. 

One question-- I don't know if this is an option or not, but hoping you can tell me either way: 
I would like to be able to report the % contribution of each feed ingredient's emissions to total feed emissions. 
I can get the absolute value of feed emissions for each ingredient with this filter: 
```
        {
            "name": "Feed emissions by ingredient",
            "filters": [
                "AnimalModuleReporter.report_daily_feed_emissions.*"
            ],
            "variables": [".*"],
            "slice_start": -365
        },
```

And I can get total feed emissions. 
But I can't figure out a way to then use each of those sums as the numerator in a secondary division. I tried this approach (below), but assume that the regex doesn't work for cross references.  
(It gives me the error-  KeyError: 'Missing referenced reports: Feed emissions by ingredient_.*_ver_agg'"). 

```
        {
            "name": "feed emissions % by ingred",
            "cross_references": [
                "Feed emissions by ingredient_.*_ver_agg",
                "Feed emissions by ingredient_feed_emissions_total_ver_agg"
            ],
            "horizontal_aggregation": "division"
        }
```

Do you have any ideas or know whether it is possible? It might just be outside of what report generator can do, and that is fine.

jadamchick:
Hi Loi -- just checking back on this. What can I do to help get it merged in as soon as possible? No worries if my question above isn't possible, just having the basic functionality as it is now will be very helpful for reports. Thanks!

ltp23:
> I would like to be able to report the % contribution of each feed ingredient's emissions to total feed emissions.

Hi Julie, the `cross_references` can't handle regex patterns yet. But also in this use case, doing horizontal division wouldn't work either because it would divide columns by one another, rather than dividing just one ingredient by the total. I don't know a good alternative at the moment besides making a separate report column per ingredient.

ltp23:
I have no more updates to this PR unless more reviews come in.

JoeWaddell (approved):
Behaves as expected, and I think adding the regex handling (which would be nice) merits a separate PR. Nice clean implementation, good work!

jadamchick (approved):
Works great!

PR #1241, opened 2024-02-06
Remove example report and update `ReportGenerator` wiki

### LINKED ISSUE #1011: Verify wiki for example report output filter
## Current Behavior
Due to changes to variables added to Output Manager, the report_example.json filter has several erroring filters:

```
      {
        "name": "First Month's Methane Emission from Pen 0",
        "filters": ["Pen\\.calc_total_manure\\.pen_0_daily_methane"],
        "slice_end": 30
      },
      {
        "name": "Last Month's Methane Emission from Pen 0",
        "filters": ["Pen\\.calc_total_manure\\.pen_0_daily_methane"],
        "slice_start": -30
      },
      {
        "name": "90 to 30 days ago Methane Emission from Pen 0",
        "filters": ["Pen\\.calc_total_manure\\.pen_0_daily_methane"],
        "slice_start": -90,
        "slice_end": -30
      }
```

Resulting in this error:
Screenshot 2023-12-01 at 12 09 41 PM

## Possible Solution
A similar issue came up within the graph generator example filter because the `calc_total_manure` method is no longer being called. Not sure if it's a long-term deprecation or a temporary one but if it's helpful to keep the same example data for these report examples, then replacing the filters in the above mentioned block with whatever the updated method that is collecting the pen daily methane would be the suggested fix.

Longer term I'm not sure the best way to keep these example filters clean other than if you see it broken, create an issue to fix it (or just fix it).

### LINKED ISSUE #1184: Update ReportGenerator Wiki
There have been some significant updates to ReportGenerator with PRs #1042 and #1048. The Wiki for Report Generator should be updated to reflect these updates.

Isgotkowitz:
This PR removes the example report `report_example.json`, as instructed in #1011.

The Report Generator wiki has been updated to include descriptions and examples of `constants`, `cross_references` and `filter_by_exclusion`. As part of reviewing this PR, please read through the wiki to see if it is missing anything or has any mistakes.

Closes #1011 closes #1184


ew3361zh (changes requested):
Really nice additions to the Wiki, Ed. Clear and concise and I think covers a lot of what we need in there. 

A few recommendations:
- Adding a section about the filter-by-exclusion option for filter files (see the `_prepare_report_data_with_constants()` method in report generator)
- Adding info about being able to use regex patterns for variables filtering (it's mentioned for the filters section but it would be good to mention for the variables as well)
- Add variables key-values to one or two of the report examples (could even just be using the regex `.*`)
- Currently the variables definition of the filter content section is capitalized and should probably be done lower-case to match the style of the other entries

Isgotkowitz:
> Really nice additions to the Wiki, Ed. Clear and concise and I think covers a lot of what we need in there.
> 
> A few recommendations:
> 
> * Adding a section about the filter-by-exclusion option for filter files (see the `_prepare_report_data_with_constants()` method in report generator)
> * Adding info about being able to use regex patterns for variables filtering (it's mentioned for the filters section but it would be good to mention for the variables as well)
> * Add variables key-values to one or two of the report examples (could even just be using the regex `.*`)
> * Currently the variables definition of the filter content section is capitalized and should probably be done lower-case to match the style of the other entries

@ew3361zh thank you, I went back and I think have addressed all the above points correctly.

ew3361zh (approved):
Looks beautiful. Nice work, Ed!

PooyaHekmati (approved):

PR #1242, opened 2024-02-06
Matching legends and colors in graph generator

### LINKED ISSUE #1008: Occasional mismatch between line colors and legends on graphs
Occasionally, the order of the legends does not match the line colors. In the following example, the legends are housing ammonia followed by manure nitrogen, but the colors are switched - housing ammonia supposed to be orange, and manure nitrogen blue. I think this is because the way the filters are looked up and somewhere along the way dictionaries are used in the process, and the order of dictionary keys is not guaranteed. 

```
{
  "multiple": [
    {
      "type": "plot",
      "filters": [
        "ManureHandlerDailyOutput_Pen_3.housing_ammonia",
        "AnimalModuleInputToManureModule_Pen_3.nitrogen"
      ],
      "title": "Housing Ammonia and Nitrogen",
      "legend": [
        "Pen 3 Housing Ammonia",
        "Pen 3 Nitrogen"
      ]
    }
  ]
}
```

default_scenario_housing-ammonia-and-nitrogen-29-Nov-2023_Wed_13-47-27

ltp23:
Closes #1008 

When the order of the filters in a graph json file differs from the order shown in the variable names file, there will be a mismatch between the legends and the line colors on the graphs.

For example, I have this graph json file. If you run this file on the main branch and this branch, you will see the order of the colors being switched. On the main branch, the manure nitrogen was recorded first and followed by housing ammonia, and the colors would be based on that data order, even though I wanted a different order in the legends.

```
{
  "multiple": [
    {
      "type": "plot",
      "filters": [
        "ManureHandlerDailyOutput_Pen_3_LAC_COW.housing_ammonia",
        "AnimalModuleInputToManureModule_Pen_3_LAC_COW.nitrogen"
      ],
      "title": "Housing Ammonia and Nitrogen",
      "legend": [
        "Pen 3 Housing Ammonia",
        "Pen 3 Nitrogen"
      ]
    }
  ]
}
```


RUFAS/graph_generator.py
```diff
@@ -150,10 +150,10 @@ def generate_graph(
                 return all_logs
 
             fig, _ = plt.subplots()
-            filtered_pool = {k: filtered_pool[k] for k in graph_details["filters"]
-                             if k in filtered_pool.keys()}
+            prepared_data = {k: prepared_data[k] for k in graph_details["filters"]
+                             if k in prepared_data.keys()}
```
ew3361zh: I believe this is duplicating the filtering that's happening already in `prepare_plot_data()` and I think may actually lead to issues of over-filtering.

For instance I was getting an error with using this filter which works on main:
```
{
      "type": "plot",
      "filters": [
        "AnimalManager\\._calc_ration_at_interval\\.ration_per_animal_for_pen_3"
      ],
      "title": "ration per animal for pen 3 no legend",
      "variables": ["234", "33", "110", "231", "51", "44", "2"],
      "filter_by_exclusion": true
    }
```
I think ordering the returned `prepared_data` dict after it's returned from `prepare_plot_data()` could potentially work.

ew3361zh (changes requested):
I think the extra filtering in `generate_graph()` isn't necessary. I think we just need to order/sort the prepared data dict that's returned from the prepare_plot_data function.

JoeWaddell:
This is throwing the error `graph generation exception. tuple index out of range` when using the following filter that works on main:
`{
    "type": "plot",
    "filters": [
        "LifeCycleManager.daily_update.(daily_milk_production|milking_cow_num)"
    ],
    "title": "Milk production",
    "legend": [
        "daily_milk_production",
        "milking_cow_num"
    ]
}`

Likewise it fails for the filters where I use a dict and the regex comprehension, for example using the filter (one of many in a "multiple" filter): 
`{
            "type": "plot",
            "filters": [
                "AnimalModuleReporter.report_daily_ration.ration_daily_feed_totals_for_pen_3_LAC_COW"
            ],
            "title": "ration pen 3 daily_feed totals",
            "variables": [".*"]
        }`

It could be that I've set these up incorrectly, but I wonder if the difference in functionality here compared to main could be related to what Niko pointed out.

Using the debugger I get the error below when inside `_route_save_functions` and try to run any of the lines inside the try block (using a breakpoint at the exception statement).
```
graph_generator = GraphGenerator(self.__metadata_prefix)
Traceback (most recent call last):
  File "C:\RUFAS\MASM-1\RUFAS\output_manager.py", line 867, in _route_save_functions
    log_pool = graph_generator.generate_graph(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\RUFAS\MASM-1\RUFAS\graph_generator.py", line 155, in generate_graph
    self._draw_graph(
  File "C:\RUFAS\MASM-1\RUFAS\graph_generator.py", line 307, in _draw_graph
    indices_list = list(range(len(values_tuple[0])))
                                  ~~~~~~~~~~~~^^^
IndexError: tuple index out of range

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
AttributeError: 'OutputManager' object has no attribute '__metadata_prefix'
log_pool = graph_generator.generate_graph(
                                filtered_pool,
                                filter_content,
                                filter_file,
                                graphics_dir,
                                )
Traceback (most recent call last):
  File "C:\RUFAS\MASM-1\RUFAS\output_manager.py", line 867, in _route_save_functions
    log_pool = graph_generator.generate_graph(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\RUFAS\MASM-1\RUFAS\graph_generator.py", line 155, in generate_graph
    self._draw_graph(
  File "C:\RUFAS\MASM-1\RUFAS\graph_generator.py", line 307, in _draw_graph
    indices_list = list(range(len(values_tuple[0])))
                                  ~~~~~~~~~~~~^^^
IndexError: tuple index out of range

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\RUFAS\MASM-1\RUFAS\graph_generator.py", line 155, in generate_graph
    self._draw_graph(
  File "C:\RUFAS\MASM-1\RUFAS\graph_generator.py", line 307, in _draw_graph
    indices_list = list(range(len(values_tuple[0])))
                                  ~~~~~~~~~~~~^^^
IndexError: tuple index out of range
```

ltp23:
Thanks Niko and Joe for those examples.

After brainstorming more possible failed cases, I realized that only when the user knows exactly what data they are going to get and in what order should they specify the order of the legend. This means that with the use of regexes and dictionary variables, the order of filtered data is, on the whole, unpredictable and therefore not recommended to be used with this feature. In these cases, it is best to let matplotlib handle the legend ordering automatically.

- When `filter_by_exclusion` is used, like in Niko's example, we generally can't tell what variables are remaining in order to do the legend ordering correctly.
- When `filter_by_exclusion` is set to `false` or not used but we have regex in the `variables` list, we can't also tell what we are going to get either. 
- Or when we include both regexes and exact filters,
- Or when one regex filter doesn't produce anything and another produces more than expected
- ...

The only valid case that seems to always work is the one where we specify the exact filters (with no special regex characters). This is limited in scope but agrees with how we would specify legend if we use matplotlib by itself.

ew3361zh:
> Thanks Niko and Joe for those examples.
> 
> After brainstorming more possible failed cases, I realized that only when the user knows exactly what data they are going to get and in what order should they specify the order of the legend. This means that with the use of regexes and dictionary variables, the order of filtered data is, on the whole, unpredictable and therefore not recommended to be used with this feature. In these cases, it is best to let matplotlib handle the legend ordering automatically.
> 
> * When `filter_by_exclusion` is used, like in Niko's example, we generally can't tell what variables are remaining in order to do the legend ordering correctly.
> * When `filter_by_exclusion` is set to `false` or not used but we have regex in the `variables` list, we can't also tell what we are going to get either.
> * Or when we include both regexes and exact filters,
> * Or when one regex filter doesn't produce anything and another produces more than expected
> * ...
> 
> The only valid case that seems to always work is the one where we specify the exact filters (with no special regex characters). This is limited in scope but agrees with how we would specify legend if we use matplotlib by itself.

I wonder how often people will specify the legend in those cases though - since they likely are likely using regex because they don't know or don't want to type out the full list of variables they're hoping to get.

I did add the lines right below where you added the dictionary comprehension which generate a legend automatically and I think would do it in the same order as the keys returned from the prepare_plot_data() method. I think maybe something to make sure they're both ordered/sorted before they're sent to be graphed would be helpful.

JoeWaddell:
> I wonder how often people will specify the legend in those cases though - since they likely are likely using regex because they don't know or don't want to type out the full list of variables they're hoping to get.

Precisely: the issue comes up with feed outputs and related variables, like feed emissions on a per ingredient basis. Feeds are different (sometimes overlapping) pen by pen, and vary farm to farm, so setting up filters for often-unique sets of variables & legend is intractable in many cases (e.g. what Julie and others are working on with numerous pilot farms). So the workflow there is to set ".*" for the variables, and let the legend do what it does.

Can the fix in this PR be modified such that it only applies when the legend is specified?

ltp23:
I added a small tweak such that the reordering of data only happens when a legend is used for literal filters (ie. no regexes or dictionaries involved). The logic for this is that when a regex or dictionaries are involved, the keys in the prepared data will differ from the filters. 

I would also add that a legend is helpful when you want to specify shorter, more user-friendly variable names.

ew3361zh:
I think this is a step in the right direction - I am still finding issues if I don't have the legend in the same order as the filters.

e.g. these two would hopefully produce the same graph but because I did the order of the legend differently in the two, the color issue still happens. This is an exaggerated example but I think the fact that the legend and filters have to be in the same order is restricting but I'm not sure how we could ever get that right since the legend and filters could be different.

```
{
      "type": "plot",
      "filters": [
        "AnimalManager._record_animal_counts.num_cows",
        "AnimalManager._record_animal_counts.num_cow_parity_1",
        "AnimalManager._record_animal_counts.num_cow_parity_2",
        "AnimalManager._record_animal_counts.num_cow_parity_3",
        "AnimalManager._record_animal_counts.num_cow_parity_4+"
      ],
      "title": "Cow Population by Parity",
      "legend": [
        "Total Cows",
        "Parity 1",
        "Parity 2",
        "Parity 3",
        "Parity 4+"
      ]
    },
    {
      "type": "plot",
      "filters": [
        "AnimalManager._record_animal_counts.num_cows",
        "AnimalManager._record_animal_counts.num_cow_parity_1",
        "AnimalManager._record_animal_counts.num_cow_parity_2",
        "AnimalManager._record_animal_counts.num_cow_parity_3",
        "AnimalManager._record_animal_counts.num_cow_parity_4+"
      ],
      "title": "Cow Population by Parity",
      "legend": [
        "Num Cows",
        "Parity 3",
        "Parity 1",
        "Parity 4+",
        "Parity 2"
      ]
    },
```

ltp23:
Hi Niko, in that example the order in the legend has to match the order of the filters. 

```
{
      "type": "plot",
      "filters": [
        "AnimalManager._record_animal_counts.num_cows",
        "AnimalManager._record_animal_counts.num_cow_parity_4+",
        "AnimalManager._record_animal_counts.num_cow_parity_2",
        "AnimalManager._record_animal_counts.num_cow_parity_1",
        "AnimalManager._record_animal_counts.num_cow_parity_3"
      ],
      "title": "Cow Population by Parity with Different Order",
      "legend": [
        "Total Cows",
        "Parity 4+",
        "Parity 2",
        "Parity 1",
        "Parity 3"
      ]
    },
```

To summarize, the use of legend is reliable only when the filters are string literals and I would mostly use it when I want shorter variable names.

Due to that limited use case, I will close this PR and the related issue without merging.

PR #1243, opened 2024-02-06
Updated Regional (NE/MW/W) Diets

harry-samuels:
I have updated three o the broader regional diets to be more complete.


KFosterReed:
@tomhuhh do you know what the plan for this PR is?

tomhuhh:
It seems that these diets are already in the pilot testing branch. So I think we should delete this branch. What are your thoughts on this? @harry-samuels Thanks!

PR #1244, opened 2024-02-07
Figuring out intermittent bug in a filter

Isgotkowitz:


RUFAS/routines/manure/manure_treatments/slurry_storage_outdoor.py
```diff
@@ -190,32 +190,30 @@ def freeboard_volume(self):
...
         """
 
         temperature_celsius = self._get_current_day_average_temperature_celsius()
         methane_loss = GasEmissionsCalculator.methane_emission_from_slurry_storage(
-            total_volatile_solids=accumulated_liquid_manure_total_solids,
+            total_volatile_solids=accumulated_liquid_manure_total_volatile_solids,
             temp=temperature_celsius,
         )
-        return methane_loss
+        return max(methane_loss, 0.0)
```
tomhuhh: @ltp23 I do not quite understand why you would compare methane_loss to 0? The logic that makes more sense to me is to compare the difference of accumulated volatile solids and 0 and take the maximum of the two. Please let me know what your thoughts are. Thanks!

Isgotkowitz:
@tomhuhh Do you think this PR can be closed? We can always recover the branch if we find that we need it later

tomhuhh:
Closing the PR...

Issue #1118, opened 2024-01-18
Update ManureManagerConfigHandler getter function args with enum members instead of strings


ew3361zh:
In MASM > RUFAS > routines > manure > IO_helpers > manure_manager_config_handler.py, update the arg for each of the getter functions with members from the appropriate enum classes.

This includes these four methods within the ManureManagerConfigHandler class:
`get_custom_bedding_config()`
`get_custom_manure_handler_config()`
`get_custom_manure_separator_config()`
`get_custom_manure_treatment_config()`

ew3361zh:
Consider making manure enums in single file or if that doesn't seem like a good solution create a folder within the manure routines to contain all enums for the manure class.

Isgotkowitz:
After working to address #1127 in PR #1250, I think this would be a bad change to make. Making the arg for these getters into an enum would severely reduce the configurability that the manure module currently has.

Isgotkowitz:
@Isgotkowitz come back here and explain the above comment, and what should happen with this issue.

Isgotkowitz:
In the methods listed above (except `get_custom_manure_handler_config()` which no longer exists), the string argument is converted to an Enum element and that element is used to find the appropriate configuration. This PR requests that an Enum element is passed instead of a string, but I think a better approach would be changing the configurations to be referenced with strings instead of Enum elements.

Manure Handler configs are looked up this way, but the string passed as an argument is the _name_ of the configuration that should be used, not the _type_. This is an important distinction because it allows multiple handlers of the same type to be used in a single simulation.

So instead of executing this issue, I propose the following steps be taken for the beddings, separators, and treatments:
- Add a "name" field to each configuration in the input.
- Enforce uniqueness of the "name" field when parsing inputs (i.e. when calling `_process_bedding_configs()`, `_process_manure_separator_configs()`, `_process_manure_treatment_configs()`).
- Remove the default configuration factories (this is what #1127 was about, it was just specific to manure handlers).
- Change the getter functions listed in the top comment to lookup configurations by name instead of type.

In short, the RuFaS should handle beddings, separators and treatments should be handled the same as handlers.

Isgotkowitz:
This can be folded into #1237

allisterakun:
Ed will write a new issue, closing this as unplanned.

PR #1250, opened 2024-02-08
Change how manure handler configurations are handled

### LINKED ISSUE #1127: Hard-coded default values for manure handler config
The `ManureHandlerConfig` and `DefaultManureHandlerConfigFactory` hardcode some values when initializing default manure handlers. 
```python
@dataclass
class ManureHandlerConfig:
    """Class for storing the configuration of a manure handler.

    Attributes
    ----------
    cleaning_water_use_rate : float
        Amount of cleaning water used per animal per day, L.
    minutes_per_cleaning : int
        Number of minutes needed per animal per cleaning, minutes.
    cleanings_per_day : int
        Number of cleanings per day.
    daily_tillage_frequency : int
        Number of times per day that compost bedding is tilled.
    cleaning_water_recycle_fraction : float
        Fraction of cleaning water that is from recycled (not fresh) water sources.
    """

    cleaning_water_use_rate: float = 0.0
    minutes_per_cleaning: int = 8
    cleanings_per_day: int = 2
    daily_tillage_frequency: int = 0
    cleaning_water_recycle_fraction: float = 0.0


class DefaultManureHandlerConfigFactory:
    """Class for creating default manure handler configurations."""

    FLUSH_SYSTEM_CONFIG = ManureHandlerConfig(
        cleaning_water_use_rate=757.0,
        cleaning_water_recycle_fraction=0.80,
    )
    MANUAL_SCRAPING_CONFIG = ManureHandlerConfig(
        cleaning_water_use_rate=10.0,
        cleaning_water_recycle_fraction=0.10,
    )
    ALLEY_SCRAPER_CONFIG = ManureHandlerConfig(
        cleaning_water_use_rate=10.0,
        cleaning_water_recycle_fraction=0.10,
    )
    TILLAGE_CONFIG = ManureHandlerConfig(
        daily_tillage_frequency=1,
    )
    HARROWING_CONFIG = ManureHandlerConfig(
        daily_tillage_frequency=0,
    )
```
## Steps to Reproduce

1.
2.
3.
4.

Isgotkowitz:
This issue closes #1127, removing the hard coded defaults for the manure handler configuration options. Additionally, it preserves the high level of configurability for manure handlers but adds more error handling to prevent users from using non-existent manure handlers.

I am pretty confident that similar changes could be made with respect to the bedding, separator, and treatment configurations (but I have not verified this yet!). This ties into issue #1118.

### Old behavior
What was possible with the old system:
```json
 "manure_management_scenarios": [
    {
      "scenario_id": 0,
      "manure_handler": "flush system",
      -- snip -- 
    },
    {
      "scenario_id": 1,
      "manure_handler": "more efficient flush system",
      -- snip --
    }
  ]
-- snip -- 
  "manure_handler_configs": [
    {
      "manure_handler_type": "flush system",
      "cleaning_water_use_rate": 757.0,
      -- snip --
    },
    {
      "manure_handler_type": "more efficient flush system",
      "cleaning_water_use_rate": 750.0,
      -- snip --
    }
  ]
```

### New behavior
How the above would be specified now:
```json
 "manure_management_scenarios": [
    {
      "scenario_id": 0,
      "manure_handler": "flush system",
      -- snip -- 
    },
    {
      "scenario_id": 1,
      "manure_handler": "more efficient flush system",
      -- snip --
    }
  ]
-- snip -- 
  "manure_handler_configs": [
    {
      "name": "flush system",
      "manure_handler_type": "flush system",
      "cleaning_water_use_rate": 757.0,
      -- snip --
    },
    {
      "name": "more efficient flush system",
      "manure_handler_type": "flush system",
      "cleaning_water_use_rate": 750.0,
      -- snip --
    }
  ]
```

### So what's the difference?
By introducing the "name" field on top of "manure_handler_type" in "manure_handler_configs", variations of the same system can be specified. For example, this means that any number of custom flush systems could be added to the input, but RuFaS will still be able to categorize and group them as flush systems.

The new system is also much less tolerant of users incorrectly writing their manure management input file. There are three cases that will crash the simulation:
1. "manure_handler_type" is not correctly specified to be one of the following: "flush system", "manual scraping", "alley scraper", "tillage" or "harrowing".
2. There is more than one manure handler configuration with the same name.
3. If the "manure_handler" value in a manure management scenario does not match any of the names in the specified in "manure_handler_configs".

Closes #1127


input/metadata/properties/default.json
```diff
@@ -3462,8 +3462,7 @@
         },
         "manure_handler": {
           "type": "string",
-          "description": "Manure Handling Method -- Method for cleaning barn alleyways.",
-          "pattern": "^(flush system|alley scraper|manual scraping|tillage|harrowing)$"
+          "description": "Manure Handling Method -- Method for cleaning barn alleyways."
```
KFosterReed: Does it make sense to make it clear in this description that the entry here needs to correspond to the "name" in the manure handler config? 

Perhaps something like: "Name of the Manure Handling Method -- Method for cleaning barn alleyways. Must match 'name' attribute of desired Manure Handler Config"

KFosterReed (approved):
Nice work Ed! 

I made one suggestion about the metadata description for the manure handler name in the manure scenario definition but other than that, this works as expected! I was able to run all the different manure handler types and specify multiple handlers of the same type and see that the configuration differences between two handlers of the same type resulted in the expected outcomes (i.e. differences in water use or tillage frequency). 

Once you have reviewed and updated the metadata as you think appropriate, this looks good to me!

input/metadata/properties/default.json
Isgotkowitz: Yes, good catch.

krb78 (approved):
I think it looks great, Ed! I ran it through and it works as it should :) Both for writing types that don't exist and making edits to config settings and resulting in changes in total manure output entering the reception pit. 🥳

RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
```diff
@@ -73,26 +74,34 @@ def get_custom_bedding_config(
...
         """
-        return self.custom_manure_handler_configs.get(
-            ManureHandlerType.get_type(manure_handler_type_name), None
-        )
+        try:
+            return self.manure_handler_configs[manure_handler_type_name]
+        except KeyError:
+            info_map = {"class": self.__class__.__name__, "function": self.get_manure_handler_config.__name__}
+            error_title = \
+                f"Attempted use a non-existent manure handler configuration called '{manure_handler_type_name}'."
```
Fjordrunner: I think this means to read "Attempted to use"?

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
```diff
@@ -347,23 +297,23 @@ class ManureHandlerFactory:
     @classmethod
     def get_instance(
```
Fjordrunner: I would argue that the function name could be improved to either include information about the instance being a BaseManureHandler instance, and that it is a specific subtype of said class.

Fjordrunner (approved):
Looks good to me, but added a naming suggestion as well as pointed out a potential typo.

jadamchick (approved):
Looks great! logical and clearly organized

RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
Isgotkowitz: Nice catch

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
Isgotkowitz: I have changed it to `get_manure_handler()`, and added all the children of `BaseManureHandler` as possible return types.

PR #1252, opened 2024-02-09
Graph reports

### LINKED ISSUE #996: Make graphs from reports
The report generator works by aggregating data. Visualizing the aggregated data is desired. Extend the list of possible keys in the report filter file such that it can accept keys that go into graph filter files as well. If graph-related keys are present, make a call to the graph generator and pass them along. Note that in this scenario, the report generator won't return the aggregated result to the output manager. If the user wants both visualization and numerical results, they would need to make a copy of the filter and remove the graph-related keys from the second copy.

ew3361zh:
Closes #996.

This PR allows for 2 new fields in a report filter file.
1. `graph_details` - will trigger the report data, once aggregated in report generator, to be graphed. The default setting will be for this data to only be graphed, so it won't be added to the report generated. `graph_details` is a dictionary and will only need the type of graph the user wants. 
e.g. report filter:
```
{
        "name": "Last Year's Average Cow Body Weight (kg)",
        "filters": [
          "LifeCycleManager.daily_update.avg_cow_body_weight"
        ],
        "slice_start": -365,
        "graph_and_report": true,
        "graph_details": {
          "type": "plot"
        }
      },
```
2. `graph_and_report` - a boolean which will trigger the data to be added to the report in addition to being graphed. If this is present and set to True, it will only work if the `graph_details` filter is also present. Otherwise, the data won't be graphed and a warning will be logged telling the user the data wasn't able to be graphed.
e.g. report filter to be graphed AND reported:
```
{
        "name": "DMI by feed daily",
        "filters": [
        ],
        "variables": [
          ".*"
        ],
        "slice_start": -365,
        "graph_and_report": true,
        "graph_details": {
          "type": "plot"
        }
      },
```

A good test file Joe created (just make sure the file you save it with starts with `report_`):
```
{
    "multiple": [
      {
        "name": "Last Year's Average Cow Body Weight (kg)",
        "filters": [
          "AnimalModuleReporter.report_life_cycle_manager_data.avg_cow_body_weight"
        ],
        "slice_start": -365,
        "graph_and_report": true,
        "graph_details": {
          "type": "plot"
        }
      },
      {
        "name": "Last Year's Average Cow Body Weight (kg)",
        "filters": [
          "AnimalModuleReporter.report_life_cycle_manager_data.avg_cow_body_weight"
        ],
        "slice_start": -365,
        "graph_and_report": true
      },
      {
        "name": "DMI by feed daily",
        "filters": [
          "AnimalModuleReporter.report_daily_ration.ration_daily_feed_totals_for_pen_3.*"
        ],
        "variables": [
          ".*"
        ],
        "slice_start": -365,
        "graph_and_report": false,
        "graph_details": {
          "type": "plot"
        }
      },
      {
        "name": "Last Year's Average Cow Body Weight (lbs)",
        "filters": [
          "AnimalModuleReporter.report_life_cycle_manager_data.avg_cow_body_weight"
        ],
        "constants": {
          "Kilograms to Pounds": 2.20462,
          "Some Adjustment Factor": 0.90
        },
        "horizontal_order": [
          "AnimalModuleReporter.report_life_cycle_manager_data.avg_cow_body_weight",
          "Kilograms to Pounds",
          "Some Adjustment Factor"
        ],
        "slice_start": -365,
        "graph_and_report": true,
        "graph_details": {
          "type": "plot"
        }
      }
    ]
}
```

To accomplish this, the PR:
1. Adds checks for the `graph_details` variable in Output Manager and adds the `__metadata_prefix` and `graphics_dir` info if it's present.
2. Checks for the presence of `graph_details` and `graph_and_report` variables in the filter file in Report Manager.
3. Creates a new function in Report Manager to prepare the aggregated report data to be graphed.

Other additions:
In graphing some of the report data, I noticed that the legend would often default to printing directly on top of the graph. If there are more than a few variables being graphed, it easily would cover significant portions of the graph. So this PR updates the standard legend placement to be outside the graph to the right. I tried "best" as the loc placement and 9/10 times would print the legend to the right of the graph but the 1/10 other times would be on top of the graph and would cover up important details. I think this is a good solution and makes the mechanism of legend placement a little more obvious to users who want to customize it further to their liking.


JoeWaddell:
This is looking good: one note is that the scenario isn't pasted into the filename: that would be a good addition to keep consistency across figures, and to sort out one scenario from the next more easily.

I am getting a bizarre bug, where data from one report seems to be leaking into another: looking at the report itself, it might be a report generator bug (or a problem with my filter, I'll paste it in below, too). It looks like columns are being stacked on top of one another in an unexpected way. I can share those on slack to avoid gumming up GitHub with more screenshots.
image

```
{
  "multiple": [
    {
      "name": "Last Year's Average Cow Body Weight (kg)",
      "filters": [
        "LifeCycleManager.daily_update.avg_cow_body_weight"
      ],
      "slice_start": -365,
      "graph_and_report": true,
      "graph_details": {
        "type": "plot"
      }
    },
    {
      "name": "DMI by feed daily",
      "filters": [
      ],
      "variables": [
        ".*"
      ],
      "slice_start": -365,
      "graph_and_report": true,
      "graph_details": {
        "type": "plot"
      }
    },
    {
      "name": "Last Year's Average Cow Body Weight (lbs)",
      "filters": [
        "LifeCycleManager.daily_update.avg_cow_body_weight"
      ],
      "constants": {
        "Kilograms to Pounds": 2.20462,
        "Some Adjustment Factor": 0.90
      },
      "horizontal_order": [
        "LifeCycleManager.daily_update.avg_cow_body_weight",
        "Kilograms to Pounds",
        "Some Adjustment Factor"
      ],
      "slice_start": -365,
      "graph_and_report": true,
      "graph_details": {
        "type": "plot"
      }
    }
```
the filter continues: the rest is the same as the `report_example.json` on main.

RUFAS/graph_generator.py
```diff
@@ -149,7 +149,8 @@ def generate_graph(
             if found_errors:
                 return all_logs
 
-            fig, _ = plt.subplots()
+            fig, _ = plt.subplots(figsize=(10, 6))
+            plt.subplots_adjust(right=0.65)
```
PooyaHekmati: why?

RUFAS/report_generator.py
```diff
@@ -202,12 +202,30 @@ def generate_report(
...
+                if not enable_graph_and_report:
+                    self.reports = Utility.filter_pool(self.reports, graph_data.keys(), filter_by_exclusion=True)
+            if enable_graph_and_report and not should_graph_report_data:
+                warning_event_log = {
+                    "warning": "report_generation_warning",
+                    "message": "Request to graph and report data not fulfilled "
+                    "- no graph_details present in report filter file.",
+                    "info_map": info_map,
+                }
+                event_logs.append(warning_event_log)
```
PooyaHekmati: is there a particular reason to have this logic in here? In OM line 868 we have: `elif filter_file.startswith(self.__supported_filter_types_prefixes["graph"]):`, can we reuse that logic? It would need some modifications, but most of it seems the same

RUFAS/output_manager.py
```diff
@@ -808,6 +808,10 @@ def save_results(
                 if filter_file.startswith(
                         self.__supported_filter_types_prefixes["report"]
                 ):
+                    if filter_content.get("graph_details"):
+                        filter_content["graph_details"]["graphics_dir"] = graphics_dir
+                        filter_content["graph_details"]["metadata_prefix"] = self.__metadata_prefix
+                        self.create_directory(graphics_dir)
```
PooyaHekmati, line 765: what if `produce_graphics` is false?

RUFAS/report_generator.py
```diff
@@ -220,6 +238,32 @@ def generate_report(
 
         return event_logs
 
+    def _prepare_report_data_to_be_graphed(
```
PooyaHekmati, line 244: this also can probably be handled more simply in OM

RUFAS/graph_generator.py
ew3361zh: This is part of what I wrote about in the PR description. The way we had it written before, the legend always (and unhelpfully) printed directly on top of the graph. My solution was to try having a standard location of the legend to the right of the graph. When I tried that, the legend would occasionally be cut-off because variable names being plotted were too long. These two lines of code adjust the size of the plots and allow for the legend to print without getting cut off. 

I'm not married to this part of the PR but I did find it helpful for viewing graphed data and it made results easier to read for all graphs (in my opinion).

RUFAS/output_manager.py
ew3361zh: Good question! I can add a check for that. I do think the gnu arg setting for graphics should supercede a filter file requesting graphics.

RUFAS/report_generator.py
ew3361zh: I will look into this option today an maybe we can discuss at our check-in.

input/data/config/default_config.json
```diff
@@ -1,6 +1,6 @@
 {
     "start_date": "2009:244",
-    "end_date": "2015:264",
+    "end_date": "2009:364",
```
allisterakun:
```suggestion
    "end_date": "2015:264",
```
Is this change intentional? Or should we revert back to the previous default?

RUFAS/output_manager.py
```diff
@@ -1003,10 +1002,10 @@ def flush_pools(self) -> None:
...
         Sets each pool to an empty dictionary.
         """
-        self.variables_pool = {}
-        self.warnings_pool = {}
-        self.errors_pool = {}
-        self.logs_pool = {}
+        self.variables_pool: Dict[str, OutputManager.pool_element_type] = {}
+        self.warnings_pool: Dict[str, OutputManager.pool_element_type] = {}
+        self.errors_pool: Dict[str, OutputManager.pool_element_type] = {}
+        self.logs_pool: Dict[str, OutputManager.pool_element_type] = {}
```
allisterakun:
```suggestion
        self.variables_pool = {}
        self.warnings_pool = {}
        self.errors_pool = {}
        self.logs_pool = {}
```
I think I took out the type hint before, b/c this is a Mypy error of re-definition for the pools.

RUFAS/output_manager.py
```diff
@@ -757,11 +753,17 @@ def save_results(
...
                 filtered_pool: Dict[str, OutputManager.pool_element_type] = {}
                 if "filters" in filter_content.keys():
-                    filtered_pool = self._filter_variables_pool(filter_content["filters"], filter_file)
+                    filtered_pool = self._filter_variables_pool(filter_content)
                 if exclude_info_maps:
                     filtered_pool = self._exclude_info_maps(filtered_pool)
 
                 if filter_file.startswith(self.__supported_filter_types_prefixes["report"]):
+                    if filter_content.get("graph_details"):
+                        print(f"here for {filter_content.get('name')}")
```
allisterakun:
```suggestion
```
Print statement

RUFAS/graph_generator.py
```diff
@@ -140,6 +143,20 @@ def generate_graph(
         Exception
             Generic exception raised by utility functions.
         """
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self.generate_graph.__name__,
+        }
+        if not produce_graphics:
+            print("no prod graphs?")
```
allisterakun:
```suggestion
```
This should be either deleted or put into the OM log

RUFAS/report_generator.py
```diff
@@ -221,6 +241,35 @@ def generate_report(
...
+            The name of the report to be graphed.
+        """
+
+        graph_generator = GraphGenerator(filter_content["graph_details"]["metadata_prefix"])
+        graph_details = {
+            **filter_content["graph_details"],
+            "title": filter_content["name"],
+            "filters": filter_content["filters"],
+        }
+        print(f"graphing {graph_details['title']}")
```
allisterakun:
```suggestion
```

allisterakun (approved):
Looks good, great job Niko!!!

Some minor suggestions!

input/data/config/default_config.json
ew3361zh: Good catch, thank you - just do this for testing to run shorter simulations.

RUFAS/output_manager.py
ew3361zh: Oh gotcha.

JoeWaddell (approved):
Nice job handling previous comments/concerns. Works as expected, and a nice addition to the reporting functionality!

PR #1258, opened 2024-02-09
Report feed inventory on last day to output manager

### LINKED ISSUE #1106: Report feed inventory to Output Manager
Now that we have integrated the Crop and Soil module into the Feed Manager, let's have some fun. The `FeedManager` is instantiated in `simulation_engine.py`. At the end of the simulation, use `FeedManager.query_available_feeds` and `OutputManager.add_variable` to report everything that FeedManager contains to the Output Manager.

ltp23:
Closes #1106 

This very small PR adds to the output manager the results of the call to `query_available_feeds()` in the simulation engine class.


PooyaHekmati (approved):

Isgotkowitz (approved):
LGTM. A future improvement could be to trim the output a little so that the values from the Enums are a littler cleaner, but that can depend on how users wind up utilizing this information.

PR #1259, opened 2024-02-11
Modified code to include accumulator function

### LINKED ISSUE #1211: Write an accumulator function for `calc_manure` in pen.py
As requested in a TODO by @ltp23 , the summation of manure amounts in `calc_manure` in pen.py may be enhanced by adding an accumulator function.

0luchi:
This PR addresses issue #1211 for the TODOx in the animal module
I created an accumulator helper function for `calc_manure` function named `manure_sums` to help with code enhancement in `pen.py`

The PR also contains unit test`test_calc_manure()` in `test_pen.py`


RUFAS/routines/animal/pen.py
```diff
@@ -407,9 +407,29 @@ def update_animals(self, new_animals: List[Any],
         self.update_animal_combination(animal_combination)
         self.update_classes_in_pen()
 
+    def manure_sums(self, manure, curr_manure, dict):
```
JoeWaddell: You'll need full typing here (I haven't checked the typing is correct, just showing the formatting here), e.g. `def manure_sums(self, manure: Dict[float, int], curr_manure: AnimalManureExcretions, dict) -> None:`

RUFAS/routines/animal/pen.py
```diff
@@ -407,9 +407,29 @@ def update_animals(self, new_animals: List[Any],
...
+        """
+        Accumulator helper function for calc_manure.
+        The function finds sums of manure components for each
+        animal in the pen and the total manure for each animal type.
+        Parameters
+        ----------
+            manure: Dict[float, int]
```
JoeWaddell: Each of the parameters should be justified with the main body of the docstring (e.g. one less indent than there is now), and there should be one space on either side of the `:` between the parameter and its typing.

RUFAS/routines/animal/pen.py
```diff
@@ -407,9 +407,29 @@ def update_animals(self, new_animals: List[Any],
         self.update_animal_combination(animal_combination)
         self.update_classes_in_pen()
 
+    def manure_sums(self, manure, curr_manure, dict):
+        """
+        Accumulator helper function for calc_manure.
+        The function finds sums of manure components for each
+        animal in the pen and the total manure for each animal type.
+        Parameters
```
JoeWaddell, line 449: There should be a space between the docstring summary and the start of the `Parameters` block.

RUFAS/routines/animal/pen.py
```diff
@@ -407,9 +407,29 @@ def update_animals(self, new_animals: List[Any],
...
+            curr_manure: AnimalManureExcretions
+                A dictionary that contains the manure excretion values as specified
+                in the AnimalManureExcretions class definition.
+            dict: Dict[float, int]
+                A dictionary that contains the manure excretion values for specific animals in the pen
+        """
+
+        for key in manure.keys():
+            manure[key] += curr_manure[key]
+            dict[key] += curr_manure[key]
```
JoeWaddell: I think here you need to return `manure`, so you can update it each time you call this method.

The second dict also needs a unique name.

RUFAS/routines/animal/pen.py
```diff
@@ -443,25 +463,16 @@ def calc_manure(self, feed, methane_model: str):  # noqa
...
         # find sums of manure components for each animal in the pen for
         # total manure in pen and total manure by animal type
-        # TODO: Write an accumulator function GitHub Issue # 1211
         for animal in animals:
             curr_manure = animal.manure_excretion
             if type(animal) == Calf:  # noqa
-                for key in manure.keys():
-                    manure[key] += curr_manure[key]
-                    calf_total[key] += curr_manure[key]
+                self.manure_sums(manure, curr_manure, calf_total)
```
JoeWaddell: With the other suggestion to ensure `manure` is a return from `manure_sums`, the call here would then be something like `manure, animal_type_dict = self.manure_sums(manure, curr_manure, calf_total)`

The `animal_type_dict` in this hypothetical would be the second dict returned by the new accumulator method, which is specific to the animal's type.

JoeWaddell (commented):
Various comments, though the main one is that when I was testing out my suggestions: I see realized that this method (Pen's `calc_manure`) isn't actually called! It is called by `Pen.calc_manure_excretion`, which is itself never called, though it appears to have been modified in the last few months.

@ltp23 may be able to comment if this method has been deprecated, or if it will be used again in the future.

RUFAS/routines/animal/pen.py
```diff
@@ -407,9 +407,29 @@ def update_animals(self, new_animals: List[Any],
...
+        Parameters
+        ----------
+            manure: Dict[float, int]
+                A dictionary that contains the the accumulated maniure excretion values for all animals
+            curr_manure: AnimalManureExcretions
+                A dictionary that contains the manure excretion values as specified
+                in the AnimalManureExcretions class definition.
+            dict: Dict[float, int]
+                A dictionary that contains the manure excretion values for specific animals in the pen
+        """
```
matthew7838:
```suggestion
                A dictionary that contains the manure excretion values for specific animals in the pen

        """
```
There should be an extra line between the last line of docstrings and the closing """

RUFAS/routines/animal/pen.py
```diff
@@ -407,9 +407,29 @@ def update_animals(self, new_animals: List[Any],
...
+
+        for key in manure.keys():
+            manure[key] += curr_manure[key]
+            dict[key] += curr_manure[key]
+
     def calc_manure(self, feed, methane_model: str):  # noqa
```
matthew7838, line 470: Also missing type hint here.

matthew7838 (commented):
LGTM, besides Joe's comments, I also have a few more. All minor comments.

matthew7838 (approved):

tests/animal_module_tests/test_pen.py
```diff
@@ -87,52 +87,7 @@ def pen_with_animals(pen: Pen, mock_animal_list: List[MagicMock]) -> Pen:
...
-            lazy_fixture("mock_animal_list"),
-        ),
-        (
-            lazy_fixture("pen_with_animals"),
-            lazy_fixture("mock_animal_list_ii"),
-            lazy_fixture("mock_animal_list_combined"),
-        ),
-    ],
-)
+@pytest.fixture
```
JoeWaddell: I think this fixture decorator needs to be removed, because this test isn't evaluated (and it's not used elsewhere as a fixture).

tests/animal_module_tests/test_pen.py
```diff
@@ -146,6 +101,67 @@ def test_add_new_animals(
     assert pen_values == expected_animals_in_pen
 
 
+@pytest.fixture
+def create_animals(mocker: MockerFixture) -> List[Calf | Cow | HeiferI | HeiferII | HeiferIII]:
```
JoeWaddell: Instead of `create_animals`, something like `animal_list` would be more accurate, so it doesn't make someone think it has to do with initialization.

Typing also needed to denote `-> None` as the return, to match other tests.

tests/animal_module_tests/test_pen.py
```diff
@@ -226,9 +242,70 @@ def test_calc_avg_stats():
     pass
 
 
-def test_calc_manure():
+def test_set_avg_nutrient_rqmts(pen: Pen):
```
JoeWaddell: Same re: `-> None` typing here and in other tests.

RUFAS/routines/animal/pen.py
```diff
@@ -438,9 +438,31 @@ def update_animals(self, new_animals: List[Any], animal_combination: AnimalCombi
         self.update_animal_combination(animal_combination)
         self.update_classes_in_pen()
 
+    def manure_sums(self, manure, curr_manure, animal_dict):
```
JoeWaddell: Typing needed here for inputs and return. Re: the typing in the docstring, the `manure` and `animal_dict` variables may be `str, float`, but it's hard to tell since this method isn't yet called. We can ask @ltp23 if this method is/should be deprecated, or if it is planned for use.

RUFAS/routines/animal/pen.py
```diff
@@ -438,9 +438,31 @@ def update_animals(self, new_animals: List[Any], animal_combination: AnimalCombi
...
+        Parameters
+        ----------
+            manure: Dict[float, int]
+                A dictionary that contains the the accumulated maniure excretion values for all animals
+            curr_manure: AnimalManureExcretions
+                A dictionary that contains the manure excretion values used to update the manure and animal dictionaries
+                in the AnimalManureExcretions class definition.
+            animal_dict: Dict[float, int]
+                A dictionary that contains the manure excretion values for specific animals in the pen
+
```
JoeWaddell, line 458: Needs Return section

RUFAS/routines/animal/pen.py
```diff
@@ -438,9 +438,31 @@ def update_animals(self, new_animals: List[Any], animal_combination: AnimalCombi
...
 
+    def manure_sums(self, manure, curr_manure, animal_dict):
+        """
+        Accumulator helper function for calc_manure.
+        The function finds sums of manure components for each
+        animal in the pen and the total manure for each animal type.
+        Parameters
+        ----------
+            manure: Dict[float, int]
+                A dictionary that contains the the accumulated maniure excretion values for all animals
```
JoeWaddell:
```suggestion
                A dictionary that contains the the accumulated manure excretion values for all animals
```

JoeWaddell (commented):
Looks much nicer! Good work. I added a few comments, but it's all typing and formatting.

The bigger question is for Loi (I tagged him in a comment), since the method here isn't being utilized right now, so we should get an idea if it's being deprecated or has a planned use case.

In either case, once the typing is tidied up, re-request my review and I'll approve!

JoeWaddell (approved):
LGTM, nice work!

Just one last super minor thing: the Parameters section in manure_sums is still indented one time too many, and should have a space between the docstring and `Parameters`

image

So e.g. there should be a blank line between 447 and 448, and line 450 (and all the others in that section) should be un-indented once.

PR #1260, opened 2024-02-11
changed function to use AnimalType enum 

### LINKED ISSUE #1209: Update calc_daily_walking_dist to use AnimalType enum
Update `calc_daily_walking_dist` in `pen.py` to use the AnimalType enum.

0luchi:
This PR addresses issue #1209 in accordance with the TODOs in the animal module


Isgotkowitz (changes requested):
This solution does not seem correct. When `self.classes_in_pen` contains `{'LacCow'}`, the statement `AnimalType.LAC_COW in self.classes_in_pen` evaluates to false even though it should be true. I think the issue is that the conditional statement is using Enum values but the set `self.classes_in_pen` is still using strings.

In addition to fixing the set type, implementing the unit test for `calc_daily_walking_dist()` in `test_pen.py` seems appropriate for this PR.

Issue #912, opened 2023-11-06
[GraphGenerator] Make Time an option for the x-axis of graphics generated by `GraphGenerator`


Isgotkowitz:
**Is your feature request related to a problem? Please describe.**
Often when generating graphs, it is useful for the x-axis to display time.

**Describe the solution you'd like**
`GraphGenerator` should provide an option for selecting how Time should be displayed in output graphs.
Possible options could include:
- The cumulative day count of the simulation (this is currently used).
- The ordinal day and calendar year of the simulation (ex. (2/2019)).
- The ordinal day and simulated year of the simulation (ex (2/1)).
**Describe alternatives you've considered**
Don't let the x-axis be changed.

**Additional context**
This is what graphs produced currently look like:
![image](https://github.com/RuminantFarmSystems/MASM/assets/43901812/b8a51430-794b-4aad-bc2d-547cff1ff36a)

Isgotkowitz:
Moving this back to TODO, because it is super wishful and is outranked by pretty much every other issue I have this sprint.

PR #1263, opened 2024-02-12
Adding ability to create pens on the fly

JoeWaddell:
This PR fixes the bug(s) preventing a pen to be added to the simulation. On main, if a pen needs to be created (e.g. a pen reaches capacity), then the simulation fails.

## New methods:
`duplicate_pen`
`data_padder_daily`
`data_padder_ration_interval`

## Deprecated methods:
`_create_default_pen` in animal_manager.py
Instead of using constants, new pens created during simulation need to match the pen that was full, so `duplicate_pen` fills this role.

## Existing methods modified:
Now not a private method, `configure_manure_manager_components` now uses a list of pens instead of entire Animal Manager, to remove circular dependency. The same had to happen for `__init__`, `daily_update`, and `simulate_daily_manure_manager`.

In addition, `configure_manure_manager_components` is called on lists of new pens in `daily_update`, to ensure that the pens have all the necessary manure info.

# Problems and open questions#
One big problem is that newly instantiated pens don't have data in OM for the days prior to their existence (which makes sense). However, this causes misalignment when comparing CSVs/reports/figures (see #1366). In this PR I've added methods to pad outputs in the reporter methods. This is hacky though, and would be better served by otherwise storing (and later accessing) the simulation day with each addition to the variables in OM. 

Also note: I haven't added the padding logic to all om calls that would necessitate it, because of problems with empty dicts, and the open question below.

It's an open question as to whether this padding logic should be kept, or removed before merging to main: this removal would mean that the misalignment could cause confusion before #1366 is solved. 

Also note: the un-padded/corrected misalignment would cause issues in interpreting the actual date once Issue #912 is instituted.


RUFAS/routines/animal/animal_manager.py
```diff
@@ -777,12 +777,61 @@ def _create_default_pen(
...
+    def _create_duplicate_pen(
+        cls,
+        pen_id: int,
+        animal_combination: AnimalCombination,
+        num_stalls: int,
+        max_stocking_density: float,
+        reference_pen: Pen,
+    ) -> Pen:
+        """
+        Create a default Pen object with the given parameters.
```
Isgotkowitz:
```suggestion
        Duplicate a Pen object using a handful of parameters and a reference Pen.
```

RUFAS/routines/animal/animal_manager.py
```diff
@@ -777,12 +777,61 @@ def _create_default_pen(
...
+            pen_type=reference_pen.pen_type,
+            manure_handling=reference_pen.manure_handling,
+            manure_separator=reference_pen.manure_separator,
+            manure_separator_after_digestion=reference_pen.manure_separator_after_digestion,
+            manure_storage=reference_pen.manure_storage,
+            animal_combination=animal_combination,
+            max_stocking_density=max_stocking_density,
+        )
+
     def _create_default_pens_for_potential_space_shortage(
```
Isgotkowitz:
```suggestion
    def _create_additional_pens_for_potential_space_shortage(
```
I think this name would better reflect this method's purpose.

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -7,11 +7,64 @@
...
+        if simulation_day > 0 and reference_variable in om.variables_pool:
+            if full_variable_to_add in om.variables_pool:
+                current_output_length = len(list(om.variables_pool[full_variable_to_add].values())[0])
+            else:
+                current_output_length = 0
+            length_difference = len(list(om.variables_pool[reference_variable].values())[0]) - current_output_length
+            if length_difference > 1:
+                short_variable_to_add = full_variable_to_add[full_variable_to_add.rfind(".") + 1 :]
+                print(f"padding: {short_variable_to_add}")
+                print(length_difference)
```
Isgotkowitz:
```suggestion
```

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -7,11 +7,64 @@
...
 class AnimalModuleReporter:
+    @staticmethod
+    def data_padder_daily(
+        reference_variable: str,
+        full_variable_to_add: str,
+        thing_to_add: Any,
+        simulation_day: int,
+        info_map: Dict[str, Any],
+    ) -> None:
+        if simulation_day > 0 and reference_variable in om.variables_pool:
```
Isgotkowitz: Add a docstring here.

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -7,11 +7,64 @@
...
+            if full_variable_to_add in om.variables_pool:
+                current_output_length = len(list(om.variables_pool[full_variable_to_add].values())[0])
+            else:
+                current_output_length = 0
+            length_difference = len(list(om.variables_pool[reference_variable].values())[0]) - current_output_length
+            if length_difference > 1:
+                short_variable_to_add = full_variable_to_add[full_variable_to_add.rfind(".") + 1 :]
+                print(f"padding: {short_variable_to_add}")
+                print(length_difference)
```
Isgotkowitz:
```suggestion
```

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -7,11 +7,64 @@
...
+                    )
+
+    @staticmethod
+    def data_padder_ration_interval(
+        reference_variable: str,
+        full_variable_to_add: str,
+        what_to_pad_with: Any,
+        simulation_day: int,
+        info_map: Dict[str, Any],
+    ) -> None:
```
Isgotkowitz: Add a docstring here.

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -559,12 +677,18 @@ def report_life_cycle_manager_data(life_cycle_manager, sim_day: int) -> None:
             dict(info_map, **{"units": cull_reason_stats_units}),
         )
 
-    def report_daily_pen_total(penlist):
+    @staticmethod
+    def report_daily_pen_total(simulation_day: int, penlist: List[Pen]) -> None:
```
Isgotkowitz:
```suggestion
    def report_daily_pen_total(simulation_day: int, pen_list: List[Pen]) -> None:
```

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -559,12 +677,18 @@ def report_life_cycle_manager_data(life_cycle_manager, sim_day: int) -> None:
...
         info_map = {
             "class": AnimalModuleReporter.__name__,
             "function": AnimalModuleReporter.report_daily_pen_total.__name__,
         }
+        classname = AnimalModuleReporter.__name__
+        funcname = AnimalModuleReporter.report_daily_pen_total.__name__
```
Isgotkowitz:
```suggestion
        classname = AnimalModuleReporter.__name__
        funcname = AnimalModuleReporter.report_daily_pen_total.__name__
        info_map = {
            "class": classname,
            "function": funcname,
        }
```

RUFAS/routines/manure/manure_manager.py
```diff
@@ -71,26 +74,30 @@ class ManureManager:
...
-        This class will replace the `ManureStorage` class.
-
-    Attributes:
-            manure_handlers: a dictionary that maps an animal pen's id to a ManureHandler object.
-            reception_pits: a dictionary that maps an animal pen's id to a ReceptionPit object.
-            manure_separators: a dictionary that maps an animal pen's id to a ManureSeparator object.
-            manure_treatments: a dictionary that maps an animal pen's id to a Treatment object.
+    Notes
+    -----
+    This class will replace the `ManureStorage` class.
```
Isgotkowitz:
```suggestion
```
I think the Manure storage class has already been replaced.

RUFAS/routines/manure/manure_manager.py
```diff
@@ -71,26 +74,30 @@ class ManureManager:
...
+    ----------
+    manure_handlers : Dict
+        a dictionary that maps an animal pen's id to a ManureHandler object.
+    reception_pits : Dict
+        a dictionary that maps an animal pen's id to a ReceptionPit object.
+    manure_separators : Dict
+        a dictionary that maps an animal pen's id to a ManureSeparator object.
+    manure_treatments : Dict
+        a dictionary that maps an animal pen's id to a Treatment object.
     """
```
Isgotkowitz:
```suggestion
    manure_handlers : Dict
        A dictionary that maps an animal pen's id to a ManureHandler object.
    reception_pits : Dict
        A dictionary that maps an animal pen's id to a ReceptionPit object.
    manure_separators : Dict
        A dictionary that maps an animal pen's id to a ManureSeparator object.
    manure_treatments : Dict
        A dictionary that maps an animal pen's id to a Treatment object.
    
    """
```

RUFAS/routines/manure/manure_manager.py
```diff
@@ -202,20 +208,24 @@ def _configure_manure_manager_components(self, animal_manager: AnimalManager) ->
                 custom_manure_treatment_config=custom_manure_treatment_config,  # type: ignore
             )
 
-    def daily_update(self, animal_manager: AnimalManager) -> None:
+    def daily_update(self, pen_list: List[Pen], simulation_day: int) -> None:
         """Calculates daily output data for each manure manager component for each animal pen.
 
         On the last day of the simulation, all the data generated daily by the manure manager
         components will be exported to a CSV file.
```
Isgotkowitz, line 214:
```suggestion
        Notes
        -----
        On the last day of the simulation, all the data generated daily by the manure manager
        components will be exported to a CSV file.
```

input/data/config/default_config.json
Isgotkowitz, line 1: Should these be reset before merging?

RUFAS/routines/manure/manure_manager.py
```diff
@@ -730,7 +740,7 @@ def _handle_daily_update_for_simple_manure_treatment(
         )
 
 
-def simulate_daily_manure_manager(manure_manager: ManureManager, animal_manager: AnimalManager) -> None:
+def simulate_daily_manure_manager(manure_manager: ManureManager, penlist: list[Pen], simulation_day: int) -> None:
```
Isgotkowitz:
```suggestion
def simulate_daily_manure_manager(manure_manager: ManureManager, pen_list: list[Pen], simulation_day: int) -> None:
```

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -100,7 +154,8 @@ def report_milk(pen, simulation_day: int) -> None:
 
             om.add_variable("milk_data_at_milk_update", milk_data_update, info_map)
 
-    def report_ration_interval_data(animal_manager, feed, simulation_day: int) -> None:
+    @staticmethod
+    def report_ration_interval_data(penlist: List[Pen], feed: Feed, simulation_day: int) -> None:
```
Isgotkowitz:
```suggestion
    def report_ration_interval_data(pen_list: List[Pen], feed: Feed, simulation_day: int) -> None:
```

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -394,6 +507,11 @@ def report_pen_manure_properties(pen) -> None:
             "enteric_methane_g": "g/day",
         }
         for manure_property, manure_value in pen.manure.items():
+            classname = AnimalModuleReporter.__name__
+            funcname = AnimalModuleReporter.report_pen_manure_properties.__name__
```
Isgotkowitz:
```suggestion
        classname = AnimalModuleReporter.__name__
        funcname = AnimalModuleReporter.report_pen_manure_properties.__name__
        for manure_property, manure_value in pen.manure.items():
```

RUFAS/routines/animal/animal_module_reporter.py
Isgotkowitz, line 1: Out of curiosity, why were the methods here converted to `staticmethod`s? If possible, it would be better to make instance or class methods.

Isgotkowitz (approved):
Left a bunch of suggestions, mostly for minor things though. Overall looks good.

My two cents w.r.t. the padding methods: I think it would be ok to keep them, but make it explicit in either #912 and/or #1366 that they should be removed when those issues get addressed.

RUFAS/routines/animal/animal_module_reporter.py
JoeWaddell: I updated the decorator to avoid this Mypy error: 
image

However, I will change them to class methods, since I think that's more appropriate here.

input/data/config/default_config.json
JoeWaddell: Definitely! Thanks for leaving the comment: it'll make for a good additional reminder.

jadamchick (approved):
Hi Joe, I appreciate the thoughtfulness going into this! 
I tested it out with a couple farms and it works well for the sake of the reports as I am generating them now (slicing everything to summarize over the last 365 days). I include pen counts in the report using a regex filter (`"filters": ["AnimalModuleReporter.report_daily_pen_total.*"],`) so I can see both a) that there were pens created above what I originally specified and b) that there is <1 animal in the original pens for those groups of animals, so it was just a need for the ramp-up period of the model but not actually reflecting an unrealistic number of animals once things have smoothed out. 

A couple thoughts- 
- this makes it especially important to ensure that all of the report filters I am using have regex to capture animal combo for anything I am trying to aggregate by calf/growing/etc. -- since filtering by pen number won't be reliable. (That's nicer anyway, so you don't have to adapt farm by farm, but even more crucial now to getting accurate outputs). 
- thinking about how pens are handled for key processes/routines within the model: do the additional pens have impacts on facilities/manure/energy calculations? (as I said above, they don't reflect the actual barn capacity needed to house a herd of this size, so don't want to penalize the farm's energy use or something by creating extra space to maintain)
- similarly thinking of within-model pen use: do the issues you outlined about sim-day alignment have any implications for how RuFaS uses pens information, apart from the outputs? (eg any impact on ration formulation, manure handling, etc.). Nothing comes to mind, but I'm not super familiar with all of the ins and outs

Overall looks good to me, though I'm admittedly not using the info right now in a format that would run into the issues you have outlined here and in 1366. Happy to chat more if you have specific questions/discussion points on that front.

tomhuhh:
Thanks Joe for tagging me on! Adding my 2-cent thoughts on the `num of stalls`. 

As `num of stalls` is used to calculate housing emissions downstream, I am hesitant to just create a new pen with the same number of stalls as the existing pen. For example, if the number of stalls is 100 while the number of additional animals is only 20, we have the risk of overestimating housing emissions. A temporary fix for this, per my discussion with Joe, is to set the number of stalls in the new pen as a multiplication of the number of additional animals and the max stocking density. This solution is not reflective of the real world scenario, but it generates final outcomes that make more sense mathematically. 

Also, I also have concerns about the newly generated pens. Once the animal number reduces back to within the limit of the original pen, the newly generated pens do not get removed by the current method. That can cause potential issues going back to what Julie mentions regarding energy and other uses. 

Thanks!
Haowen

JoeWaddell:
Following up from the meeting with Haowen, I changed the functionality such that the duplicate pen will only be the size necessary to house the excess individuals (multiplied by the stocking density), so that they aren't huge.

As Haowen pointed out, in general this behavior should be avoided in general: because excess animals would in reality be culled instead of overcrowded or moved to a temporary pen. However, since this isn't a universal rule, we opted to keep this in as a general fix for those edge cases. I'll open an issue to add logic that would remove excess pens once animal populations stabilize.

PR #1264, opened 2024-02-12
Using Black autoformatter on code pushed to Github

### LINKED ISSUE #1130: Set a team-standard formatter for use in the RuFaS development workflow
It is common for different programmers to have different formatting preferences when coding. This can become as issue when working on a team though, because differences in formatting choices can result in non-uniform code. Non-uniform code can cause unnecessary headaches or be a distraction when reading or reviewing code.

The team agrees on a way to standardize the formatting of RuFaS in a way that uses an automated tool. This tool will need to be easy to use and to adopt into workflows. Ideally, this formatter leaves as little room as possible for variations in coding style, and can be automatically applied to code that is submitted in PRs.

The [Black](https://black.readthedocs.io/en/stable/) formatter is a strong candidate for an autoformatter. It is widely used, and has good integrations with different IDEs as well as GitHub Actions. It is very aggressive, but is configurable enough that it can be made to play nice with Flake8.

There are many other autoformatters out there. Other commonly used ones are yapf and autopep8. All have their pros and cons.

An alternative to using an autoformatter would be to define a RuFaS-specific formatting standard and put the burden of formatting code correctly onto the people writing the code. This is, in my opinion, a bad option. It will result in people spending a lot more time fixating on formatting instead of scrutinizing the code and what it needs.

**Additional context**

ew3361zh:
Closes #1130.

Updates Github linting action to include an autoformatting using the Black formatter and pushes those changes to the PR.

This PR is particularly large because Black was run on the entire codebase to give RUFAS a fresh start as the new formatting Github Action is put into place with this PR.

To cut through the format-changes-only files, the three main files to review are:
1. `lint_flake8.yml`
2. `lint_branch_changes.bat`
3. `lint_branch_changes.sh`

Notes:
- Black doesn't format docstrings - it focuses on Python code and considers docstrings to be "content."
- There are a few instances where Black will reformat something and give precedence to line break readability over line-length. There's no way to prioritize which formatting choices it will institute over others. There is a way to have it ignore particular lines of code you won't want it to reformat so I've done that in a small handful of places.
- Black is designed to be unconfigurable beyond a few options (like line length), aiming for uniformity across all codebases. Flake8 is highly configurable, allowing developers to ignore specific rules, customize checks, and integrate other plugins for additional linting capabilities. To reconcile some of the differences which Black's formatting enforces without option to customize and what flake8 is looking for, I've updated both the flake8 local scripts as well as the flake8 Github action to **ignore** the following rules: 
1. E203 - no space before a `:` in some instances in which Black enforces adding that space for readability.
4. W503 - no line break before a binary operator (e.g. `+` or `*`) which Black enforces for readability.


.github/workflows/lint_flake8.yml
```diff
@@ -39,11 +39,27 @@ jobs:
...
+        git remote set-url origin https://${GITHUB_TOKEN}@github.com/${{ github.repository }}.git
+
+        CLEAN_BRANCH_NAME=${BRANCH_NAME/refs\/heads\//}
+        git push -f origin HEAD:$CLEAN_BRANCH_NAME
     - name: Lint with flake8
       if: ${{ env.FILES != '' }}
       run: |
         # stop the build if there are Python syntax errors or undefined names
-        flake8 ${{ env.FILES }}
+        flake8 ${{ env.FILES }} --ignore=E203,W503
```
PooyaHekmati: shouldn't this be in `https://github.com/RuminantFarmSystems/MASM/blob/main/.flake8`?

ew3361zh: Yes! Thank you, good catch.

.github/workflows/lint_flake8.yml
```diff
@@ -39,11 +39,27 @@ jobs:
                         tr " " "\n" |
                         grep -E "\.txt$|\.json$|\.csv$"  |
                         tr "\n" " ")" >> "$GITHUB_ENV"
+    - name: Format code with Black
+      run: |
+        black --line-length=120 ${{ env.FILES }}
```
Isgotkowitz, line 43: I suggest adding a `pyproject.toml` that looks like the following:
```
[tool.black]
target-version = ["py310", "py311"]
line-length = 120
include = '.*\.py$'
```
This allows black to format consistently whether it is being automated by the GitHub Action or someone is doing it locally, without having to specify the command line args each time. Also, having an established `pyproject.toml` in place would be nice in case we decide to add something like [isort](https://pycqa.github.io/isort/)

.github/workflows/lint_flake8.yml
```diff
@@ -27,7 +27,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install flake8
+        pip install flake8 black
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```
Isgotkowitz: Since flake8 is already in requirements.txt, can black be placed there as well and then remove this pip install line?

.github/workflows/lint_flake8.yml
```diff
@@ -39,11 +39,27 @@ jobs:
                         tr " " "\n" |
                         grep -E "\.txt$|\.json$|\.csv$"  |
                         tr "\n" " ")" >> "$GITHUB_ENV"
+    - name: Format code with Black
```
Isgotkowitz, line 41: I'm curious, why not use the official [Black GitHub Actions integration](https://black.readthedocs.io/en/stable/integrations/github_actions.html)?

RUFAS/input_manager.py
```diff
@@ -97,20 +106,29 @@ def _load_properties(self) -> None:
         Exception
             For any other unexpected errors during properties loading.
         """
-        info_map = {"class": self.__class__.__name__,
-                    "function": self._load_properties.__name__,
-                    }
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self._load_properties.__name__,
+        }
```
Isgotkowitz, line 117:
```suggestion
        info_map = {"class": self.__class__.__name__, "function": self._load_properties.__name__}
```
This is an example of how Black is really good, but has a few buried rules that can lead to unexpected formatting choices. The above change isn't actually necessary, I just wanted to point this out to people.

Isgotkowitz (approved):
Left some comments; only suggestion I will strongly advocate for is adding `pyproject.toml` and configuring Black there.

Also reiterating Pooya's comment, Flake8 configuration should only be done in .flake8 and not in any of the spots where it is actually invoked.

> There is a way to have it ignore particular lines of code you won't want it to reformat so I've done that in a small handful of places.

I think it needs to be emphasized that while this may be necessary at times, it is a bad practice and should be avoided as much as possible.

Nice job Niko! Can't wait to have this on main.

.github/workflows/lint_flake8.yml
ew3361zh: I think that could work!

RUFAS/input_manager.py
ew3361zh: It's a good point. I think Black focuses on its predetermined definition of readability to the nth degree. So it's not a necessary change, but does make the same choice across the board so things look consistent. I think I've commented on how strange some of the choices look but maybe once the RUFAS team has been using Black for long enough these will all look like the norm to us.

ew3361zh:
> Left some comments; only suggestion I will strongly advocate for is adding `pyproject.toml` and configuring Black there.
> 
> Also reiterating Pooya's comment, Flake8 configuration should only be done in .flake8 and not in any of the spots where it is actually invoked.

Ah, thanks for that clarification. I don't know that I got the latter part of that initially. I will test out removing it from the places where flake8 is invoked to ensure it works as intended and then push those updates.
 
> > There is a way to have it ignore particular lines of code you won't want it to reformat so I've done that in a small handful of places.
> 
> I think it needs to be emphasized that while this may be necessary at times, it is a bad practice and should be avoided as much as possible.

100%. It's ironic (maybe) that most of the cases where I had to implement having Black ignore lines of code were in the function definitions of functions that violated max-complexity and Black kept relocating the `# noqa` note.

.github/workflows/lint_flake8.yml
ew3361zh: I will look into this. Thanks, Ed!

.github/workflows/lint_flake8.yml
ew3361zh: I believe the big difference is the offical Black Github Actions Integration only checks for Black violations but won't actually format the code and implement the changes. In the issue and during our grooming session we'd discussed having Black actually run in github and push those formatting changes into the branch. But I can look and see if there are pieces in here that can be incorporated into the workflow we have. I like this part `- uses: psf/black@stable` in particular.

allisterakun (approved):
The git Action changes look good, easy to comprehend. 

Just a comment about maybe having a small section in wiki about black. I wasn't familiar with the tool and I had to read through their documentation about all the possible functionalities and customization options.

Issue #651, opened 2023-07-12
Integrate a static type checker into the Github workflow


michaelRichards99:
**Is your feature request related to a problem? Please describe.**
Type annotations are used extensively throughout our code base, as described by [PEP 484](https://peps.python.org/pep-0484/). According to the PEP 484 proposal, type annotations exist at runtime but there _is no type checking at runtime_. Instead, the proposal relies on external tools to provide type analysis to the Python source code. 

We have type annotations but are currently not checking them with a static type analysis tool.

Using a static type analyzer will be beneficial because it makes identifying certain classes of bugs much easier to track down.

**Describe the solution you'd like**
We should:
- [x] Investigate which static analyzer works best for RuFaS
- [ ] Add a Github workflow that runs the type analyzer as part of the CI pipeline

**Describe alternatives you've considered**
We can continue without a type analyzer, which is part of the benefit of using a dynamic language such as Python.

**Additional context**
The type analysis tool [mypy](https://mypy-lang.org/) indicates 222 existing type errors in the code base. Fixing these could be nontrivial.

michaelRichards99:
# Investigation of static type checkers for python:
## Pros and Cons
1. Mypy
	- Pros:
		- Most popular and most featureful.
		- Recommended by [RealPython](https://realpython.com/python-type-checking/), we use many of their resources.
		- Integrates well with both VScode and Pycharm
	- Cons:
		- see below
2. Pytype
	- Pros:
		- easy to integrate into existing code base because of how lenient it is 
	- Cons:
		- Not as customizable as the others
		- Doesn't integrate with Pycharm
3. Pyright
	- Pros:
		- Good overall option. similar in features to mypy.
	- Cons:
		- poor Pycharm integration
4. Pyre
	- Pros:
		- Comes with Pyre and Pysa. Pyre is the type checker and Pysa is a static analysis tool that can identify security vulnerabilities
	- Cons:
		- Poor Pycharm integration

### Drawbacks that all tools share
- Must migrate code base to be correctly typed.
	- depending on how strict the type checker tool is configured, the current MASM code base contains between 200-950 type errors.
- Takes developer time and effort

PooyaHekmati:
So, what is the conclusion? Which option is the winner?

michaelRichards99:
Mypy is the best option in my opinion. It has the best support across different IDEs, which will be beneficial for members on our team. It also comes with extensive documentation and plenty of external resources and support from the community.

[...]

PR #1278, opened 2024-02-14
Adding a type checker to workflows

ltp23:
| Feature               | MyPy   | PyType | PyRight | Pyre |
|-----------------------|--------|--------|---------|--------|
| Comprehensiveness, strictness | Comprehensive, strictness configurable  | Depends on how much code is type-annotated. Lenient, configuration options somewhat limited compared to mypy and pyright |  I would say it is more comprehensive than mypy under strict mode.   |  I can't tell if pyre has any unique advantages compared to other ones in terms of type annotation. However, one of its components, pysa, allows the detection of security vulnerabilities, which I don't think is a high priority for Rufas.
|Performance | Fast | Slow | Fast | Fast |
|Integration with PyCharm | Reasonably good | No | No | No |
|Intergation with VSCode | Yes | No | Yes, via Pylance | Yes, but much less common than mypy and pyright | 
|Customizability | Great | OK | Great | OK | 
|Estimated errors in Rufas | ~4500 | ~400. Not run fully yet. | ~13000 | ~1700 |
| My observations when used on Rufas | Under strict mode, lots of errors were caught and mypy expected type annotations almost everywhere. | Most lenient. Made inferences where possible, based on usage patterns, to allow gradual adoption. Very cool but the tradeoff was it took some time to do such analysis. | Errors were most granular, whereas in mypy similar errors would be grouped together. When I used it in PyCharm, the autocompletion feature worked! | Standard outputs like the other ones.

Because there is no automatic way to handle type conflicts and it is not really practical to go back and forth between the terminal and the editor to fix and verify things, having the type checker integrated into the IDE to allow visual feedback is quite important. As a result, I would opt for the mypy type checker, which is a safe, good middle-ground option in terms of comprehensiveness and performance. 

For PyCharm users, mypy seems to be the only one decently integrated. For VSCode users, both mypy and Pylance extensions are good. 

---
Update:

- I added a badge on the readme page to show the number of mypy errors.
- For any new PR, if the number of mypy errors increases compared to that of the main branch, the PR should be flagged with an error. For tricky and time-consuming type checking errors, we can use `# type: ignore` to temporarily bypass the type checker.


Isgotkowitz:
Closes #651

Isgotkowitz:
@ltp23 reading through the table comparing options, I agree - MyPy sounds like the best option.

> pysa, allows the detection of security vulnerabilities, which I don't think is a high priority for Rufas.

I would agree that this is not a high priority for RuFaS _right now_, but it may be in the future. However when we get to that point, it might be best to get a dedicated static analysis tool for security scanning (possibly Snyk).

A couple of suggestions for the implementation/integration into RuFaS: 
- Add the type checker run into the linting GitHub action. I think this will help minimize time setting up the overhead.
- The options for running the type checker should _only_ be set in the config file, not added in the GitHub Actions call (I assume it will still work correctly if not specified there).
- Configure the type checker in `pyproject.toml` if possible (it is for MyPy, didn't check the others).

A big unanswered question that I have related to the type checker (no matter which one is picked) is how will it be integrated into the workflow? If it finds an issue, should the PR be blocked? Should there just be a badge with the number of errors on the Readme? My two cents is that it would be good to keep the count on the Readme, and when you push a PR that count is not allowed to go higher.

ltp23:
> A big unanswered question that I have related to the type checker (no matter which one is picked) is how will it be integrated into the workflow? If it finds an issue, should the PR be blocked? Should there just be a badge with the number of errors on the Readme? My two cents is that it would be good to keep the count on the Readme, and when you push a PR that count is not allowed to go higher.

I think having a badge count is a good idea. I'm still not sure if blocking a PR based on type-checking errors is a good approach yet because it is not always straightforward to fix such errors. Though I have attempted to implement the blocking functionality based on the difference in the error count between a PR and the main branch.

ltp23:
> I would agree that this is not a high priority for RuFaS _right now_, but it may be in the future. However when we get to that point, it might be best to get a dedicated static analysis tool for security scanning (possibly Snyk).

It seems like it's not too much of a hassle to switch between type checkers so choosing mypy is definitely not set in stone.

ltp23:
> A couple of suggestions for the implementation/integration into RuFaS:
> 
> * Add the type checker run into the linting GitHub action. I think this will help minimize time setting up the overhead.

I think linting and type-checking are separate enough to have their own github actions. Also, flake8 and black linting can modify the contents of the changed files but the type checker won't. 

> * The options for running the type checker should _only_ be set in the config file, not added in the GitHub Actions call (I assume it will still work correctly if not specified there).

I'm not sure what you meant here. I thought we would want to automate the type checking process.

> * Configure the type checker in `pyproject.toml` if possible (it is for MyPy, didn't check the others).

This could be a refactoring task later. Mypy uses the config file `mypy.ini` so I think it could be combined with the toml file.

Isgotkowitz:
> > * The options for running the type checker should _only_ be set in the config file, not added in the GitHub Actions call (I assume it will still work correctly if not specified there).
> 
> I'm not sure what you meant here. I thought we would want to automate the type checking process.

MyPy is being configured in two places right now: `mypy.ini` and with the flags in `mypy_type_checker.yml`: `--install-types --non-interactive --strict`. It should only be configured once - in the config file, so that it runs the same whether it is in a GitHub action or someone is running it locally.

> > * Configure the type checker in `pyproject.toml` if possible (it is for MyPy, didn't check the others).
> 
> This could be a refactoring task later. Mypy uses the config file `mypy.ini` so I think it could be combined with the toml file.

MyPy is configurable with `pyproject.toml` - [see this](https://mypy.readthedocs.io/en/stable/config_file.html#). I think the configuration should be put in `pyproject.toml` in this PR, unless there is a pressing reason not to.

>> I would agree that this is not a high priority for RuFaS right now, but it may be in the future. However when we get to that >> point, it might be best to get a dedicated static analysis tool for security scanning (possibly Snyk).
>
> It seems like it's not too much of a hassle to switch between type checkers so choosing mypy is definitely set in stone.

I think you mean not set in stone. I disagree - it would be pretty significant to change a dependency like a type checker in a project with a lot of people, especially after it gets established in the workflow. But Mypy is really good and widely adopted, so it seems to me like a good plan to have type checking and we can add a static analysis tool aimed at security if/when we need one.

Isgotkowitz (changes requested):
See above reply to comments.

ltp23:
> MyPy is being configured in two places right now: `mypy.ini` and with the flags in `mypy_type_checker.yml`: `--install-types --non-interactive --strict`. It should only be configured once - in the config file, so that it runs the same whether it is in a GitHub action or someone is running it locally.

Not all options are configurable using the config file approach.

ltp23:
> I think you mean not set in stone. I disagree - it would be pretty significant to change a dependency like a type checker in a project with a lot of people, especially after it gets established in the workflow. But Mypy is really good and widely adopted, so it seems to me like a good plan to have type checking and we can add a static analysis tool aimed at security if/when we need one.

Sorry about the typo.
This can depend on whether we go from a stricter type checker to a more lenient one or the other way. The former choice would be pretty painless, I imagine, but the latter choice would require more type annotations. Mypy is an overall good option, but it is okay if we decide later on there is a better option. Setting things to strict mode helps in the long run.

ltp23:
> MyPy is configurable with `pyproject.toml` - [see this](https://mypy.readthedocs.io/en/stable/config_file.html#). I think the configuration should be put in `pyproject.toml` in this PR, unless there is a pressing reason not to.

I have merged the mypy.ini file into the toml file.

Isgotkowitz:
> Not all options are configurable using the config file approach.

Then I think it would be best to leave those options behind. Having an Action run a static analysis tool differently (in a substantive way) than it would run when invoked from the command line with no additional args can lead to a pipeline that we think works one way but actually works in another (this is what happened with flake8).

I suggest adding the following lines to `pyproject.toml`:
```
strict = true
exclude = ['^(?!.*(?:RUFAS|tests|main\.py|setup\.py)).*$']
```
When the above lines are added to `pyproject.toml`, the following two commands in the terminal:
```
$ mypy .
```
and
```
$ mypy --install-types --non-interactive --strict RUFAS tests main.py setup.py
```
find the same count of errors: "Found 4538 errors in 231 files (checked 332 source files)". I tried changing the Mypy runs in the Action to be `python -m mypy . ...`, and it produced some very unexpected results. So unfortunately the above would not be a silver bullet to unifying automated and manual Mypy runs, but I think it would help.

ltp23:
> the same count of errors: "Found 4538 errors in 231 files (checked 332 source files)".

This indicates that Mypy doesn't need to fetch external type files yet but it may not hold true later on. However, I will incorporate your suggestions to unify the two approaches.

I think the unexpected results you saw was probably because the main branch's toml file doesn't have a section for mypy configuration. My solution is temporarily allowing the `continue-on-error: true` when comparing the error counts between the two branches, so the toml file changes can be merged first. Then, I will make a follow-up PR to remove that.

Isgotkowitz (approved):
Thank you for addressing all my comments and suggestions Loi, this is an excellent start to having all of RuFaS properly type checked.

> My solution is temporarily allowing the continue-on-error: true when comparing the error counts between the two branches, so the toml file changes can be merged first. Then, I will make a follow-up PR to remove that.

This sounds like a great plan.

ew3361zh (approved):
Looks really good, Loi! I think it's fine to leave it as it stands now that it won't prevent merges to main until we've spent some more time with it running on our PRs.

I think it would be helpful to add an overview of this action and the usage of MyPy to the Wiki similar to the other 3 sections at the end: Using Flake8, Using Python Black Formatter, Using Sphinx. Just so people are clear on what's happening and how they can contribute to getting that number of errors down.

Really nice work though, I think this will be a helpful tool to continue to improve and clean up the codebase.

PR #1280, opened 2024-02-14
Deobfuscate Crop and Soil output

### LINKED ISSUE #1148: Deobfuscate Crop and Soil output
The daily and annual data recorded to the OutputManager from the Crop and Soil module do so (mostly) through a main collection module (formerly `OutputGatherer`, currently `FieldDataReporter`). This has certain benefits, but makes it significantly more difficult to determine where the output is _actually_ being generated, not just recorded.

Possible solutions:
- Set the info maps in `FieldDataReporter` to reflect where individual variables are actually generated/changed. This would make its behavior consistent with the `AnimalModuleReporter`.
- Break up the `FieldDataReporter` and have all variables recorded where they are generated.

This issue comes out of point #2 in issue #965.

allisterakun:
- add a new entry to info_map: a list of tuples called _data_origin_. Each tuple contains 2 elements: originating class, originating function
- field data reporter will still be reporting them but this will help point to where the data is being originally reported.

Closes #1148


RUFAS/routines/field/manager/field_data_reporter.py
```diff
@@ -34,578 +34,1412 @@ def send_daily_variables(self) -> None:
...
+                        info_map,
+                        **{
+                            "data_origin": [
+                                ("CarbonCycling", "_soil_carbon_aggregation"),
+                                ("LayerData", "_initialize_carbon_pools"),
+                            ]
+                        },
+                    ),
+                )
+                # same variable as above?
```
Isgotkowitz: Yup, that's an issue. Would you be ok with deleting "carbon_emissions" in this PR? Otherwise a separate issue can be written up for it.

Isgotkowitz (changes requested):
This is a great start Allister. Unfortunately there are interactions between biophysical modules that done in a very opaque way, and are very hard to find using F12 or Ctrl-B. This is a big issue, but one that is outside the scope of this PR.

Could you please check `field/tillage_application.py` and the following files in `crop`: `water_uptake.py`, `nitrogen_incorporation.py`, and `phosphorus_incorporation.py`? If you see a soil attribute being modified with `setattr()` or the function `set_vectorized_layer_attribute()` then it means the attribute is being written to, and this should be captured in `data_origin`.

PR #1281, opened 2024-02-14
Handle `IM._populate_pool()` invalid data

### LINKED ISSUE #1014: Handling invalid data in `IM._populate_pool()`
When validating a file in the InputManager, the `_populate_pool()` function first loads the data into a **Dictionary** (JSON file is straightforward, CSV file is loaded as `Dict[str, List[Any]]`). It will then iterate through every key in the loaded dictionary to run validation.

However, if the current block of data (current key) is valid, the current behavior is to add the entire loaded dictionary object into `InputManager.__pool`. This would result in potential invalid data being added to the pool.

Here is a snapshot of the code in `IM._populate_pool()`:
```

            ...

            input_data = data_loader(file_path)

            ...

            properties_blob_key = file_details["properties"]
            metadata_properties = self.__metadata["properties"][properties_blob_key]
            filtered_input_data = self._filter_input_data_by_metadata(input_data, metadata_properties)
            for metadata_property in metadata_properties.keys():
                
                ...

                if element_counter_and_validity["is_valid"]:
                    # POTENTIAL ISSUE!!!
                    self.__pool[file_blob_key] = filtered_input_data
```
The very last line would insert the entire input data into the InputManager pool for every successful validation of a section of the data, even the sections (future keys to be checked) that have not been validated.

## Expected Behavior

When validating the input data, all the valid data should be added into the InputManager pool, leaving out all the invalid keys.
For handling invalid data:
- if `eager_termination` is `False`
    - The invalid data should be left out of the InputManager pool, while all the other successfully validated data should be added.
- if `eager_termination` is `True`
    - The function `_populate_pool()` should return False upon encountering unfixable invalid data, this would in turn terminate the execution.

## Current Behavior

If any section of the input data is successfully validated, the entire input data dictionary object will be added to the InputManager pool. 
This would not be an issue for further simulation if the `eager_termination` is set to `True`, as the program would exit upon encountering unfixable invalid data. However, if `eager_termination` is set to `False` and some part of the input data is invalid and unfixable, it would still be added to the InputManager due to other sections being valid. This invalid data may cause issues further down the simulation.

## Possible Solution

Have a local Dictionary variable in the `_populate_pool()` function. Only the valid sections will be added to the local variable. At the end of the for loop, we can add the local variable to the InputManager pool so that all the invalid sections are left out.

## Steps to Reproduce

1. Example metadata:
```

```
2.
3.
4.

allisterakun:
Closes #1014 

Introduces a local variable `validated_data` as a buffer to temporarily hold the validated data in `InputManager._populate_pool()`. IM will only add the validated data after running validation for every file.


ew3361zh (approved):
Nice work, Allister! Did you consider any cases to expand testing to show this new functionality is acting as intended? I was thinking about it for a little while this morning and was struggling to come up with a good test case. Coverage is still at 100% though.

allisterakun:
> Nice work, Allister! Did you consider any cases to expand testing to show this new functionality is acting as intended? I was thinking about it for a little while this morning and was struggling to come up with a good test case. Coverage is still at 100% though.

Will make a test case for that! ==> DONE!

Isgotkowitz (approved):
LGTM

PR #1282, opened 2024-02-14
Adds a IM.__get_data_logs_pool

### LINKED ISSUE #1126: IM.get_data() log
## Expected Behavior

Keep a log of all the calls for `InputManager.get_data()`
## Current Behavior

No log
## Possible Solution

Create a pool in IM or OM to keep track of all the calls to get data

allisterakun:
Closes #1126 

- Added an `InputManager.__get_data_logs_pool` to keep track of all `IM.get_data()` calls.
- Renamed `OutputManager._generate_file_name` to `OutputManager.generate_file_name` to make it a public method.
- Introduced a new function `InputManager.dump_get_data_logs()` to dump the get data logs at the end of the simulation.
    ```
        """
        Dumps the stored get data logs to a JSON file at the specified path.

        Parameters
        ----------
        path : Path
            The directory path where the JSON file will be saved.

        Returns
        -------
        None
        """
    ```


RUFAS/input_manager.py
```diff
@@ -5,9 +5,13 @@
 import re
 
 import pandas as pd
+from path import Path
```
ew3361zh: Curious why not use the pathlib Path here?

allisterakun: It was the suggested default by PyCharm. I'll change it to pathlib for consistency! Great catch!

ew3361zh (approved):
Looks good overall! I had one comment about using pathlib Path but otherwise worked great and was wild to see that `IM.get_data()` was called 331k times over the course of the default scenario simulation.

RUFAS/input_manager.py
allisterakun: Done

RUFAS/input_manager.py
```diff
@@ -774,6 +779,10 @@ def get_data(self, data_address: str) -> Any:
         try:
             data_value = reduce(lambda d, key: d[key], element_hierarchy,
                                 self.__pool)
+
+            timestamp = Utility.get_timestamp(include_millis=True)
+            self.__get_data_logs_pool[timestamp] = f"InputManager.get_data() gets called for {element_hierarchy}."
```
```suggestion
            self.__get_data_logs_pool[timestamp] = f"InputManager.get_data() called for {element_hierarchy}."
```
I think this makes the message sound a little better.

RUFAS/input_manager.py
```diff
@@ -1187,3 +1196,20 @@ def add_tabular_variable_to_pool(self, variable_name: str, data: Dict[str, List[
...
+        Dumps the stored get data logs to a JSON file at the specified path.
+
+        Parameters
+        ----------
+        path : Path
+            The directory path where the JSON file will be saved.
+
+        Returns
+        -------
+        None
```
```suggestion
```
Not sure if there is a team standard set for this yet but I think it should be that the returns section isn't included if the method returns None. That is just my preference though, it is something we should decide on as a team.

Isgotkowitz (approved):
LGTM. Left two comments on minor things.

RUFAS/input_manager.py
allisterakun: Sounds reasonable! I kind of just copy and pasted the docstring from ChatGPT, will be more aware of those details!

PR #1284, opened 2024-02-14
Updated inputs (c1f1 and pilot testing templates) 

jadamchick:
Updated pilot testing template .jsons to reflect recent PRs. 
Updated c1f1 inputs so it would run.


allisterakun (approved):
Looks good!

PR #1286, opened 2024-02-14
Exclude main from automatic merge of Black formatting.

ew3361zh:
Currently the Github Action script to automatically run Black on the code will subsequently attempt to merge those formatted changes into the branch on which it is run.

This is in conflict with the protections we have on main from being able to merge without approvals.

When the original PR is merged into main, the code has already been formatted by Black. There won't be any further changes from another run of Black on this same code. 

Therefore, this PR removes the automatic-merge when the branch is main. This should prevent this error from happening and not lose any formatting done by Black.

Additionally came across an error where the script will fail if Black is asked to be run when there are no .py files changed. I've updated the script to not run when there are no .py files changed to prevent this error.

### DIFF
--- a/.github/workflows/lint_flake8.yml
+++ b/.github/workflows/lint_flake8.yml
@@ -1,7 +1,7 @@
 # This workflow will install Python dependencies and run lint with a single version of Python
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
 
-name: Linting
+name: Formatting and Linting
 
 on:
   push:
--- a/.github/workflows/lint_flake8.yml
+++ b/.github/workflows/lint_flake8.yml
@@ -42,6 +42,7 @@ jobs:
       run: |
         black --line-length=120 ${{ env.FILES }}
     - name: Commit and push formatting changes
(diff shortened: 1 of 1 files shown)


.github/workflows/lint_flake8.yml
```diff
@@ -39,9 +39,11 @@ jobs:
                         grep -E "\.txt$|\.json$|\.csv$"  |
                         tr "\n" " ")" >> "$GITHUB_ENV"
     - name: Format code with Black
+      if: ${{ env.FILES != '' && github.ref != 'refs/heads/main' }}
```
allisterakun, line 42: Maybe saving this `${{ env.FILES != '' && github.ref != 'refs/heads/main' }}` as an ENV variable?

allisterakun (approved):
Just a comment, otherwise it looks good! Good job, Niko!

PooyaHekmati (approved):
Let's also change the name of the action on line 4 to Formatting and Linting

.github/workflows/lint_flake8.yml
ew3361zh: I've tried this a few different ways and it's not working as intended. I may leave it for now since it's working as intended.

PR #1287, opened 2024-02-14
Heifer growth minimum

### LINKED ISSUE #1166: Heifer daily_growth is sometimes negative
Found in PR #1152 , as some animals were still producing unexpected negative values.

## Expected Behavior
Heifers aren't expected to lose weight.

## Current Behavior
Some values in `get_non_preg_bw_change` result in negative values.  Found in `heiferI.py`

For instance (found using the debugger), when:
 `self.mature_body_weight` is 735.203333, 
 `self.body_weight` is 457.4165980518182, 
 `AnimalBase.config["target_heifer_preg_day"]` is 420, 
and `self.days_born` is 450, 
the equation `(0.55 * 0.96 * self.mature_body_weight - 0.96 * self.body_weight) / abs(AnimalBase.config["target_heifer_preg_day"] - self.days_born)`
 results in -0.5787792534743794

###     Additional note
This method lacks a citation, pseudocode or otherwise. Seems to match [A.1A.C.6] in the Animal Life Cycle pseudocode doc.

## Possible Solution(s)
If it isn't as simple as a missing term or parentheses, additional logic may need to be added. e.g. instead of abs, if the difference is less than or equal to zero, the divisor should be 1. Alternatively, any results that are negative would be capped at 0, or some other minimum growth value.

## final additional note
This may also require similar logic changes in `get_bw_change` in `heiferII.py`

JoeWaddell:
Added new constant for animal module.

Added logic to use minimum growth rate for heifer constant in cases where growth is predicted to be negative (see Issue #1166 for longer writeup on the issue). I've set the constant to roughly the minimum growth rate exhibited, but what that should be prior to merge is an open question.

Negative growth on main only occurs in about 20-25 animals in the default simulation on main (6 yrs, `cow_num` of 100). They're also exclusively animals that had a pregnancy loss between 1st and 2nd pregnancy check. They usually exhibit negative growth rate for 15-20 days, e.g. until their next estrus day.


ltp23 (approved):
Looks good to me, Joe. I still don't know if there is a deeper issue here that needs more investigation.

jadamchick (commented):
Hi Joe,
Great investigation, this looks like a definite improvement from what was there. 
Is this on the list of things to be re-examined with the animal module re-design? Your comment that the heifers losing weight are those who lost a pregnancy is illuminating -- do they get treated as a "heiferI" in terms of growth, even though they are older/bigger animals? 
I also think it would be ideal for the user to be able to define what % of body weight they are targeting for heiferIII's when they calve in to be cows, instead of having that hard-coded in RuFaS; which might require more flexibility in some of this? 

In short: I think this fix is good for now; I also think this whole issue indicates some gaps between RuFaS and real life that maybe should be kept on the radar.  What do you think is the best way to handle that? 


Also - a reminder to add this to the change logs :)

JoeWaddell:
> do they get treated as a "heiferI" in terms of growth, even though they are older/bigger animals?

Exactly: so here we have to consider whether finding/developing a different equation for daily growth is worth it for this number of animals, or if a flat lower limit is sufficient. Like you mentioned: this is a great example of how we need to think about animal classes/types/graduation, because sometimes equations from the literature make assumptions that can't be applied across all life stages.

As for the 55%, that's a value from the literature, but any parameter is of course a candidate for being included in the constants file, since that acts as kind of an 'advanced inputs' collection. 

Also yes: I'll update the changelog, thanks for the reminder!

jadamchick (approved):

PR #1294, opened 2024-02-14
Update default_animal values

harry-samuels:
This pull request changes a few default values in the default_animal file to be accurate to the reference literature. Changes based off of this document: https://docs.google.com/spreadsheets/d/10w20Bcf9JggyHwFZ-kXc2iBbJen-S6akzY2nUCdJtuE/edit#gid=1207586583


input/data/animal/default_animal.json
```diff
@@ -37,40 +37,40 @@
         "milk_type": "whole"
       },
       "repro": {
-        "voluntary_waiting_period": 50,
+        "voluntary_waiting_period": 0,
```
krb78: I think you set this to 0 because it is our recommendation that VWP = 0 in TAI only protocols (and I believe that is the default).  However, I think it'd be more reasonable to leave at 50 because if we don't and someone doesn't update it because they are unsure of what input to put and instead know that they want to use ED or TAI-ED, we will start breeding cows at 20-30 days in milk (which is quite rare). Also, leaving at 50 would keep us consistent with dairy herd management software defaults (both DC305 and PC Dart use 50 as their default). I will add that if a farm is doing TAI only for first service, 50 is quite low and likely this will be 70+. However, I still think 50 is a good option as our default given the previously mentioned data.

input/data/animal/default_animal.json
```diff
@@ -37,40 +37,40 @@
...
-            "ED_conception_rate": 0.6,
+            "estrus_detection_rate": 0.6,
+            "ED_conception_rate": 0.5,
             "presynch_program": "Double OvSynch",
-            "presynch_program_start_day": 50,
+            "presynch_program_start_day": 44,
             "ovsynch_program": "OvSynch 56",
-            "ovsynch_program_start_day": 66,
-            "ovsynch_program_conception_rate": 0.7,
+            "ovsynch_program_start_day": 72,
```
krb78: How closely do we want this tied to the Double Ovsynch presynch program and/or VWP? A double ovsynch program lasts 27 days and if we wanted to start that on day 44, then the ovsynch-program-start-day should be day 61. Truthfully, I think all of this depends on what we set our default reproductive program to be in cows  and how "user proof" we want to make RuFaS...

krb78 (commented):
Nice work Haowen and Harry! I think there are just a few minor repro defaults we should consider updating or at least discussing - but it really depends on how "user proof" we'd like to make RuFaS.

input/data/animal/default_animal.json
KFosterReed: This and Kaitlyn's next comment raise great points about the default repro inputs. My thoughts are that it would make sense to switch the default repro program to ED-TAI instead of TAI only.

krb78: I would be a very strong proponent of that - I think it is the most common type of repro program used across the US!

input/data/animal/default_animal.json
```diff
@@ -16,15 +16,15 @@
   "animal_config": {
     "management_decisions": {
       "breeding_start_day_h": 380,
-      "heifer_repro_method": "TAI",
-      "cow_repro_method": "TAI",
+      "heifer_repro_method": "ED",
+      "cow_repro_method": "ED-TAI",
```
krb78, line 20: beautiful :)

krb78 (approved):
Looks great, Haowen! :) Thanks for making those edits!

input/data/animal/default_animal.json
ltp23: I would add that VWP is only relevant in ED and ED-TAI protocols. When TAI-only is selected, the VWP value will be ignored so it is nice to have it set to 0 to have the parameters logically consistent. Though I understand that if someone is to change the protocol to ED or ED-TAI from TAI-only, they may forget to reset the VWP value to something meaningful.

ltp23 (approved):
I did run it once and the herd structure looked normal to me. Thanks, Harry and Haowen.

PR #1296, opened 2024-02-15
Update input for manure total ammoniacal nitrogen

KFosterReed:
The previous method for estimating the mass of total ammoniacal nitrogen excreted by the animals was not in alignment with the values expected by the equations to predict ammonia emissions. Until we have a better method for predicting ammoniacal nitrogen excretion directly, we will replace the current methods with a simpler method that assumes a constant fraction of urine nitrogen is total ammoniacal nitrogen.

This PR removes the methods in `pen_manure.py` that were used to calculate the values needed for the previous method and updates the input for the `manure_total_ammoniacal_nitrogen` variable in the `PenManure` class.

### DIFF
--- a/RUFAS/routines/manure/pen_manure/pen_manure.py
+++ b/RUFAS/routines/manure/pen_manure/pen_manure.py
@@ -105,14 +105,6 @@ class PenManure:
             A PenManure object.
 
         """
-        manure_mass = animal_manure["manure_mass"]  # kg
-        manure_volume = (
-            manure_mass / ManureConstants.SLURRY_MANURE_DENSITY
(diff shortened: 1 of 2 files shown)


elle-andreen (approved):
Runs for me without issue! 

Just sticking this here strictly for future reference: manure TAN values are presently smaller than expected, which is likely connected to how the study that we referenced for estimating manure TAN sampled the manure as well as our methods for estimating manure volume. We still need to review the manure mass and volume equations in the future, but setting manure TAN mass equal to a constant fraction of N mass, and separating this value from manure volume, will ensure that a reasonable manure TAN value is being used for ammonia emission calculations in the meantime.

ew3361zh (approved):
LGTM! Nice job fixing the broken unit test!

PR #1298, opened 2024-02-15
Remove double counted animals from housing emissions equations

### LINKED ISSUE #1268: Housing emissions double counting
The number of animals/stalls are double counted in the `housing_methane_emission` & `housing_carbon_dioxide_emission` functions.

The `num_animals` is used when calculating housing methane. 
Screenshot 2024-02-13 at 10 27 13 AM

The `num_stalls` is used when calculating barn area. 
Screenshot 2024-02-13 at 10 30 57 AM

Similar issue also happens in the `housing_carbon_dioxide_emission` function. Please double check in the code file to make sure that I am not missing anything.  

## Expected Behavior

The barn area is a multiplication of stall area and the number of stalls, `num_stalls`.  So we do not further multiply it by `num_animals` when calculating housing emissions. 

## Current Behavior

The number of animals/stalls are double counted. 

## Possible Solution

Delete `num_animals` in the return statement of the  `housing_methane_emission` & `housing_carbon_dioxide_emission` functions. Please make sure to update the unit tests.  

## Steps to Reproduce

1. Simulate any manure management scenarios and make sure you have a filter that outputs housing methane and total methane. 
2. Check the ratio of housing methane to total methane and if it is above 5% (as suggested by the literature), there is an issue of overestimating housing methane. 
3.
4.

ew3361zh:
Closes #1268

This removes the num animals from the two housing emissions equations in `calculator.py`.

- [x] num_animals removed from `housing_methane_emission()` and `housing_carbon_dioxide_emission()` function params.
- [x] Updates args for places these functions are called within the codebase (`manure_handler_classes.py` and the test suite).
- [x] Updates some docstrings formatting and content.

### DIFF
--- a/RUFAS/routines/manure/gas_emissions/calculator.py
+++ b/RUFAS/routines/manure/gas_emissions/calculator.py
@@ -329,17 +329,15 @@ class GasEmissionsCalculator:
 
         .. math::
 
-            E_{CH_4} = num\\_animals \\times max(0, 0.13 * T_{barn}) \\times barn\\_area / 1000
+            E_{CH_4} = max(0, 0.13 * T_{barn}) \\times barn\\_area / 1000
 
         where:
 
             :math:`E_{CH_4}` is the methane housing emission in kg :math:`CH_4/day`,
 
-            :math:`T_{barn}` is the barn temperature in :math:`^{\\circ}C`,
+            :math:`T_{barn}` is the barn temperature in :math:`^{\\circ}C`, and
 
-            :math:`barn\\_area` is the barn area per animal based on housing type in :math:`m^2`, and
-
-            :math:`num\\_animals` is the number of animals in the pen.
+            :math:`barn\\_area` is the barn area per animal based on housing type in :math:`m^2`.
 
         Parameters
         ----------
@@ -375,17 +373,15 @@ class GasEmissionsCalculator:
 
         .. math::
 
-            E_{CO_2} = num\\_animals \\times max(0, 0.0065 + 0.0192 * T_{barn}) \\times barn\\_area / 1000
+            E_{CO_2} = max(0, 0.0065 + 0.0192 * T_{barn}) \\times barn\\_area / 1000
 
         where:
 
             :math:`E_{CO_2}` is the carbon dioxide housing emission in kg :math:`CO_2/day`,
 
-            :math:`T_{barn}` is the barn temperature in :math:`^{\\circ}C`,
+            :math:`T_{barn}` is the barn temperature in :math:`^{\\circ}C`, and
 
-            :math:`barn\\_area` is the barn area per animal based on housing type in :math:`m^2`, and
-
-            :math:`num\\_animals` is the number of animals in the pen.
+            :math:`barn\\_area` is the barn area per animal based on housing type in :math:`m^2`.
 
         Parameters
         ----------
--- a/RUFAS/routines/manure/gas_emissions/calculator.py
+++ b/RUFAS/routines/manure/gas_emissions/calculator.py
@@ -319,7 +319,7 @@ class GasEmissionsCalculator:
         return modified_hours * (max_temp - min_temp) / 2 + (max_temp + min_temp) / 2
 
     @classmethod
-    def housing_methane_emission(cls, num_animals: int, barn_area: float, barn_temp: float) -> float:
+    def housing_methane_emission(cls, barn_area: float, barn_temp: float) -> float:
         """
         Calculate housing methane emissions from manure handlers.
 
@@ -343,8 +343,6 @@ class GasEmissionsCalculator:
 
         Parameters
         ----------
-        num_animals : int
-            Number of animals in the pen (unitless).
         barn_area : float
             Barn area per animal based on housing type (:math:`m^2`).
         barn_temp : float
@@ -358,19 +356,16 @@ class GasEmissionsCalculator:
         Raises
         ------
         ValueError
-            If the number of animals or barn area is less than 0.
+            If the barn area is less than 0.
 
         """
-        if num_animals < 0:
-            raise ValueError("Number of animals must be greater than or equal to 0.")
-
         if barn_area < 0:
             raise ValueError("Barn area must be greater than or equal to 0.")
 
-        return num_animals * max(0.0, 0.13 * barn_temp) * barn_area / 1000
+        return max(0.0, 0.13 * barn_temp) * barn_area / 1000
 
     @classmethod
-    def housing_carbon_dioxide_emission(cls, num_animals: int, barn_area: float, barn_temp: float) -> float:
+    def housing_carbon_dioxide_emission(cls, barn_area: float, barn_temp: float) -> float:
         """
         Calculate carbon dioxide housing emission.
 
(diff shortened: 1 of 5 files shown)


RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -329,22 +329,18 @@ def housing_methane_emission(cls, num_animals: int, barn_area: float, barn_temp:
...
 
             :math:`E_{CH_4}` is the methane housing emission in kg :math:`CH_4/day`,
 
-            :math:`T_{barn}` is the barn temperature in :math:`^{\\circ}C`,
+            :math:`T_{barn}` is the barn temperature in :math:`^{\\circ}C`, and
 
-            :math:`barn\\_area` is the barn area per animal based on housing type in :math:`m^2`, and
-
-            :math:`num\\_animals` is the number of animals in the pen.
+            :math:`barn\\_area` is the barn area per animal based on housing type in :math:`m^2`.
```
elle-andreen: I think this should read similar to "...is the total barn area based on the pen type and number of stalls"? Based on the changes in manure_manager_pen.py

elle-andreen:
One more minor comment: in manure_manager_pen.py, we have a key `BarnArea` (line 140) that is surface area per **stall**, and a key `barn_area` (line 156) that is **total** surface area of the pen. Do we need `BarnArea` for anything? If not, I think we should update the docstring above there to specify that we are calculating **total** barn/pen surface area here, because right now we're describing defining the surface area of each stall.

ew3361zh:
> One more minor comment: in manure_manager_pen.py, we have a key `BarnArea` (line 140) that is surface area per **stall**, and a key `barn_area` (line 156) that is **total** surface area of the pen. Do we need `BarnArea` for anything? If not, I think we should update the docstring above there to specify that we are calculating **total** barn/pen surface area here, because right now we're describing defining the surface area of each stall.

It's a bit of complicated mapping that is happening there but I do believe we need both. Though maybe we need to define what they represent more clearly.

`BarnArea` (140) is a class-like tuple structure that allows you to store data in an ordered fashion. It has two fields: has_cows and no_cows.

In between we have the `barn_area_by_pen_type` dictionary which maps the string names of the `pen_types` to the variable instances of the BarnArea structure set up on lines 141-144.

`barn_area` (156) is a variable that holds an instance of the BarnArea tuple - specifically the one corresponding to self.pen_type after looking it up in the barn_area_by_pen_type dictionary. This variable is used to access the area values (has_cows or no_cows) based on the conditions ("Cow" in self.classes_in_pen) to calculate the total area needed by multiplying with self.num_stalls.

elle-andreen:
> It's a bit of complicated mapping that is happening there but I do believe we need both. Though maybe we need to define what they represent more clearly.
> 
> `BarnArea` (140) is a class-like tuple structure that allows you to store data in an ordered fashion. It has two fields: has_cows and no_cows.
> 
> In between we have the `barn_area_by_pen_type` dictionary which maps the string names of the `pen_types` to the variable instances of the BarnArea structure set up on lines 141-144.
> 
> `barn_area` (156) is a variable that holds an instance of the BarnArea tuple - specifically the one corresponding to self.pen_type after looking it up in the barn_area_by_pen_type dictionary. This variable is used to access the area values (has_cows or no_cows) based on the conditions ("Cow" in self.classes_in_pen) to calculate the total area needed by multiplying with self.num_stalls.

Thank you for the thorough explanation Niko! I definitely worded my comment poorly haha. I should have been more clear that by 'needing both', I meant to ask if the value of surface area per stall is utilized somewhere in the model, or do we only really need the total pen surface area value. 

I agree that we just need to more clearly define what they each represent. I think `def barn_area_from_pen_type` ultimately returns the total pen surface area, but the docstring is describing surface area per stall (or per animal, originally). Should we update it to describe the calculation of total pen surface area, or both per-stall surface area and total pen surface area?

elle-andreen:
One more question that is optional but hopefully helpful and relatively simple to implement - with calculating pen surface area by stall rather than by # of animals in the pen, we're now relying on the user to enter a realistic number of stalls for each pen. Just as an example, in the current default_animal.json, herd_num is 100 but the two lactating cow pens have 1000 stalls each. So you could end up with a proportionally very large barn surface area compared to your number of cows in the pen. 

I'm not sure if there has been a validation of some kind implemented somewhere for this. Could we have a check here or wherever is appropriate to confirm that num_stalls is within maybe +/- the user-defined stocking density range (I think default is 20%, and that's maybe too conservative) as num_animals? E.g., if num_stalls is 500 but num_animals is 300, an error or warning would be returned. It may not be that simple, as I'm not familiar with how cows are allocated to pens based on the inputted herd_num value, but hopefully you get what I mean!

ew3361zh:
> > It's a bit of complicated mapping that is happening there but I do believe we need both. Though maybe we need to define what they represent more clearly.
> > `BarnArea` (140) is a class-like tuple structure that allows you to store data in an ordered fashion. It has two fields: has_cows and no_cows.
> > In between we have the `barn_area_by_pen_type` dictionary which maps the string names of the `pen_types` to the variable instances of the BarnArea structure set up on lines 141-144.
> > `barn_area` (156) is a variable that holds an instance of the BarnArea tuple - specifically the one corresponding to self.pen_type after looking it up in the barn_area_by_pen_type dictionary. This variable is used to access the area values (has_cows or no_cows) based on the conditions ("Cow" in self.classes_in_pen) to calculate the total area needed by multiplying with self.num_stalls.
> 
> Thank you for the thorough explanation Niko! I definitely worded my comment poorly haha. I should have been more clear that by 'needing both', I meant to ask if the value of surface area per stall is utilized somewhere in the model, or do we only really need the total pen surface area value.
> 
> I agree that we just need to more clearly define what they each represent. I think `def barn_area_from_pen_type` ultimately returns the total pen surface area, but the docstring is describing surface area per stall (or per animal, originally). Should we update it to describe the calculation of total pen surface area, or both per-stall surface area and total pen surface area?

Ah! Sorry for misunderstanding. Yes, I think you're right about what it's returning there and I can update that.

I can't speak to whether the surface area per stall is utilized elsewhere in the model. Since it's hard-coded into the BarnArea NamedTuple locally in this function, I would guess that it's not but that's only a guess. 

I also did a search for `surface area` and `surface_area` and `stall` and couldn't find any references that fit.

elle-andreen (approved):
Good job team!

KFosterReed:
> One more question that is optional but hopefully helpful and relatively simple to implement - with calculating pen surface area by stall rather than by # of animals in the pen, we're now relying on the user to enter a realistic number of stalls for each pen. Just as an example, in the current default_animal.json, herd_num is 100 but the two lactating cow pens have 1000 stalls each. So you could end up with a proportionally very large barn surface area compared to your number of cows in the pen.
> 
> I'm not sure if there has been a validation of some kind implemented somewhere for this. Could we have a check here or wherever is appropriate to confirm that num_stalls is within maybe +/- the user-defined stocking density range (I think default is 20%, and that's maybe too conservative) as num_animals? E.g., if num_stalls is 500 but num_animals is 300, an error or warning would be returned. It may not be that simple, as I'm not familiar with how cows are allocated to pens based on the inputted herd_num value, but hopefully you get what I mean!

I think this is a great idea Elle but I would put it in the category as a 'nice to have' rather than a must have right now. Would you be up for making an issue and putting in the Animal Module project board as an enhancement?

tomhuhh (approved):
Thanks Niko! These changes make sense to me. I ran the code, and the simulation works with no difficulty. Additionally, the resulting housing methane emissions under the slurry storage outdoor condition looks reasonable. I think this is good to merge. Thanks!

PR #1299, opened 2024-02-15
Diesel consumption

### LINKED ISSUE #2357: [EEE] Reconcile Diesel Consumption PR Merge Conflicts
### Description:
The feature branch associated with PR #1299 is currently ~4800 commits behind dev and ~115 commits ahead. Due to the significant divergence, merging this branch into dev will likely result in numerous merge conflicts that need to be addressed manually.

### Objectives:

- Rebase or merge the dev branch into the PR branch to bring it up to date.
- Resolve all merge conflicts carefully to ensure that new features and changes in both branches are preserved and functional.
- Use planning spreadsheet for guidance on setup and intended functionality: https://docs.google.com/spreadsheets/d/1BP8QgoKnMbrG72aut2azpJVHyuD-M7RaNOy_XhEYsIA/edit?gid=1051621784#gid=1051621784
- Move EEE folder to same level as `biophysical` folder. Be aware of import needs from other modules. This may require moving some shared data structures to the `data_structures` folder. If this process proves to be too difficult, EEE can be moved to within `biophysical` but that is not the ideal if possible.

### Deliverables:
 
- A clean, conflict-free, up-to-date PR branch that runs the energy module successfully.
- A brief summary of how merge conflicts were resolved (especially if complex).
- (Lower priority) Update all associated unit and integration testing.

### Notes:
 
- Have a SME (Kristan or other) test the resulting PR for functionality and accuracy.
- This code was last updated in December. There may be issues found because of things being out of date or not working within the current model - use your best judgment based on what you know about the model and the best practices for coding. Trust your instincts. If you are really unsure, consult with other dev-team members including Pooya, who wrote the original code.

PooyaHekmati:
1. Implements `Tractor` and `Tractor Implement` for diesel consumption calculations.
2. Implements `Energy` for energy calculations.
3. Fixes a bug in `InputManager` where it was flushing the pool on new instantiations. 
4. Calculates the consumed diesel per operation and in total. The results can be used with `diesel_consumption_tractor_implement` filter.
5. The code in this PR is developed based on [this document](https://docs.google.com/spreadsheets/d/1BP8QgoKnMbrG72aut2azpJVHyuD-M7RaNOy_XhEYsIA/edit?usp=sharing).
6. The correctness of the values needs confirmation from SMEs. The consumption for Harvest events is thousands of times larger than other operations, a unit conversation is probably being wrong somewhere.


RUFAS/routines/EEE/tractor.py
```diff
@@ -0,0 +1,135 @@
...
+        Implements Helper Function 420  in EEE Functions file.
+        """
+        if herd_size < 0:
+            raise ValueError("Herd size must be a positive integer.")
+        if herd_size < 500:
+            return TractorSize.SMALL
+        elif herd_size < 2000:
+            return TractorSize.MEDIUM
+        else:
+            return TractorSize.LARGE
```
Isgotkowitz, line 98: Instead of having S&C say that a specific tractor size handled a field management operation, S&C should just provide  the size of the field the operation happened on, and then the tractor uses that determine what size tractor was used in the same manner that it does here.

RUFAS/routines/EEE/energy.py
```diff
@@ -0,0 +1,106 @@
...
+        """Runs all estimation functions and performs pre/post processing for them."""
+        base_info_map = {
+            "class": EnergyEstimator.__name__,
+            "function": EnergyEstimator.estimate_all.__name__,
+            "unit": "unitless",
+        }
+        crop_yield = 0  # TODO get the correct value
+        field_production_size = 0  # TODO get the correct value
+        herd_size = 0  # TODO get the correct value
+        operation_event = FieldOperationEvent.PLANTING  # TODO get the correct value
```
Isgotkowitz: For field-related emissions prediction, would it be doable to have separate methods for each FieldOperationEvent element (or maybe that is already the plan)?

RUFAS/routines/EEE/tractor.py
```diff
@@ -0,0 +1,135 @@
+from typing import List
+from .tractor_implement import TractorImplement
+from RUFAS.util import Utility
+from RUFAS.input_manager import InputManager
+from .enums import TractorSize, FieldOperationEvent, OperationType, CropType
+
+input_manager = InputManager()
+
+
+class Tractor:
```
Isgotkowitz, line 23: Looking at the init method here, it seems like there is going to be a new tractor initialized for every operation. Would it be better to just initialize the three tractors (S, M, and L) when EEE is initialized, and then keep reusing the same 3 tractor instances whenever calculating the emissions from an operation?

RUFAS/routines/EEE/tractor_implement.py
```diff
@@ -0,0 +1,126 @@
...
+    def calculate_drawbar_power(self, clay_percent: float) -> float:
+        """
+        Calculates Drawbar Power (kW) which is the power required by the implement for propulsion forward if it
+        requires a transfer of tractor power to its wheel drives for this purpose.
+        Implements Helper Function 414  in EEE Functions file.
+        """
+        functional_draft = self.calculate_functional_draft(clay_percent)
+        return functional_draft * self.field_speed_km_per_hr * 1.2 / 3600
+
+    def calculate_functional_draft(self, clay_percent: float) -> float:
```
Isgotkowitz: Clay percent will need to be pulled from OM as well.

RUFAS/routines/EEE/tractor.py
KFosterReed: For the time being we are setting a single tractor size for each farm based on the cow size. We will likely update this soon to be pegged to the total field size (i.e. the sum of the hectares across all fields) but I am seeking some external SME input on this and for now, cows is good enough :)

RUFAS/routines/EEE/energy.py
KFosterReed: I am not sure that I understand the question and also this may not be a good question for me to answer so feel free to ignore if it is not helpful but, my understanding is that we have one somewhat general equation for the amount of fuel needed by the tractor for different operations but that certain parts of it are turned on and off depending on the operation.

RUFAS/routines/EEE/tractor_implement.py
```diff
@@ -0,0 +1,126 @@
...
+        """
+        Assign a tractor implement based on the operation, tractor size, and crop type where applicable. Not all
+        operations are depended on crop type.
+        Implements Helper Function 419 in EEE Functions file.
+        """
+        dataset_raw = input_manager.get_data("tractor_dataset")
+        dataset = Utility.convert_dict_of_lists_to_list_of_dicts(dataset_raw)
+        for data_entry in dataset:
+            if (
+                data_entry.get("Crop Type or Tillage Implement").lower() == self.crop_type.value.lower()
```
KFosterReed: This seems like it will work for the Harvest and planting operations but not the nutrient application and tillings operations because the values for ID#s 872-901 in "Crop Type or Tillage Implement" will not match the self.crop_type variable 

Would it make sense to make a separate function for the other group of operations?

RUFAS/routines/EEE/tractor_implement.py
```diff
@@ -0,0 +1,126 @@
...
+            OperationType.MOWING,
+            OperationType.WINDROWING,
+        ]:
+            return 0.1 * self.field_speed_km_per_hr * self.width_m * self.field_efficiency
+        if self.operation_type in [
+            OperationType.COLLECTION,
+            OperationType.FERTILIZER_APPLICATION_BELOW_SURFACE,
+            OperationType.FERTILIZER_APPLICATION_SURFACE,
+            OperationType.LIQUID_MANURE_APPLICATION_BELOW_SURFACE,
+            OperationType.LIQUID_MANURE_APPLICATION_SURFACE,
```
KFosterReed: Ok, this is my fault. Can we move all of the OperationTypes _except for_ `OperationType.COLLECTION` to the other group? I am going to update the EEE sheet right now

PooyaHekmati: Yes, and it makes the code much simpler :D

RUFAS/routines/EEE/tractor_implement.py
PooyaHekmati: good catch

RUFAS/routines/EEE/tractor_implement.py
PooyaHekmati: Thanks, I changed the if statement to covert those cases as well. If it still doesn't work, we can add more functions.

KFosterReed:
Ok! Finally actually started to review this PR today. I'll start with the good news... 

Diesel use rates for planting and fertilizer application are in the right ballpark!

The not so great news is that estimates for manure application, tillage, and harvest are all way high (like 10^3 too high). I am going to start reviewing the methods in detail but in the meantime, I think the outputs could use some clean up. 

There are some seemingly extra variables being reported where the output is created but it is not populated with an output. Most of them don’t make sense. These are:

- crop_type_for_Fertilizer Application
- tillage_implement_for_Fertilizer Application
- crop_type_for_Tilling
- crop_type_for_Manure Application
- tillage_implement_for_Manure Application
- application_depth_for_Harvest
- tillage_implement_for_Harvest
- application_depth_for_Planting
- tillage_implement_for_Planting

Similarly, it only makes sense to report crop yield for harvest operations but there are crop yield variables reported for the other operations:

- crop_yield_for_Fertilizer Application
- crop_yield_for_Tilling
- crop_yield_for_Manure Application
- crop_yield_for_Planting

It's exciting progress even if it isn't quite right yet!

KFosterReed:
Hi Diesel team! I just pushed a change that I have been testing that does the following:

- changes crop_yield to be in units of tons per ha
- changes the crop_yield value for non-harvest operations to be 0 instead of 1
- changes the reported diesel consumption to be in units of liters per ha instead of liters per ton

In addition, I modified the equation for `field_efficiency` in cases of Collection/Harvest operations because the quotient was actually the reciprical of what it should be. These changes get the predicted results fairly close to the benchmarking values I am using for comparison. Planting in particular is within the expected range. Alfalfa harvest is 5-10 l too high and corn grain harvest is about 10x too high. 

One additional change that I identified that needs to be made relates to the method for estimating PTO for manure and fertilizer application. In particular: 

- the field capacity method in `field_capacity_ha_per_hr` should use the same functional equation as the collection operation type but instead of crop yield, it should use the fertilizer or manure DM application mass as an input
- Similarly in `calculate_needed_PTO` if the operation is fertilizer or manure application, we should actually replace the `crop_yield_ton_per_ha` variable with something like `dm_application_ton_per_ha`. This part of the equation doesn't matter for planting and tillage because the `G` parameter is 0 for those operations

KFosterReed:
Hello again Diesel team! Another update on what may be the longest running PR in GitHub's history...

It seems like our attempt to simplify the tractor size is leading to, on average, high specific fuel capacities/consumptions. Because of this, I decided to make a further simplifying assumption and use the average value for this parameter of 0.044 gal/hp/hr or 0.2234 L/kWh. This, in addition to changing crop yeild from 1 to 0 for non-harvest events leads to pretty reasonable results for planting, tillage, most harvests, and fertilizer application. 

There are a few pending tasks before we can actually close this chapter which I have added to an updated version of our [energy function document](https://docs.google.com/spreadsheets/d/1umygH9caaSpyX22MhQ_8Lcl4NUSW-U9o_DtN1PfHuVM/edit?usp=sharing): 
-  Add a third option for calculation in `field_capacity_ha_per_hr` for manure and fertilizer application. This is reflected in the sheet in the addition of HF 418c. For these operation types, the field capacity methd should use the same functional equation as the collection operation type but instead of crop yield, it should use the fertilizer or manure DM application mass as an input
- Similarly in `calculate_needed_PTO` if the operation is fertilizer or manure application, we should actually replace the `crop_yield_ton_per_ha` variable with something like `dm_application_ton_per_ha`. This part of the equation doesn't matter for planting and tillage because the G parameter is 0 for those operations.
- This would benefit from an SWE review of the changes I pushed today :)

KFosterReed:
We are getting closer and closer! With the updates to the fertilizer and manure application methods, the estimates for all practices in the freestall scenario look great! (will share a summary soon). However, testing on the open lot scenario revealed that there is a division by 0 error caused by the cover crop kill operation where the energy estimation methods expect there to be a yield value. The solution for this is to distinguish between harvest operations that just terminate the cover crop growth and leave the crop biomass on the field (and thus have 0 yield) and those that acctually collect the crop biomass. 

Allister already started to address this by pulling in the `harvest_type` variable and using it to determine the method used to calculate the field capacity but we also need to include this distinction in the method `determine_implement_parameters`. In this method previously, the implement parameters were selected based only on operation type, crop and tractor size. We need an additional selection criteria for if the operation is a kill only harvest, the implement parameters for the mowing operation should be selected (for the appropriate crop and tractor size). 

I added an updated tractor dataset to make sure that all the relevant crops have implement information for mowing operations so the next step is to update the `determine_implement_parameters`. @allisterakun let me know if you have any questions or need further information to help make this happen :)

KFosterReed (approved):
I approve this PR! 

Here is a summary of the diesel consumption estimates for the example farm field management practices in units of liters per hectare:
image


I think we are underestimating the manure application but it is not so far off and we can improve when we conduct a more in depth analysis. For now, It works and responds to the major input drivers as expected!

RUFAS/biophysical/field/crop/crop_management.py
```diff
@@ -171,7 +171,7 @@ def manage_harvest(
         if harvest_op in (HarvestOperation.KILL_ONLY, HarvestOperation.HARVEST_KILL):
             self.kill()
 
-        self._record_yield(field_name, field_size, time.current_calendar_year, time.current_julian_day)
+        self._record_yield(harvest_op, field_name, field_size, time.current_calendar_year, time.current_julian_day)
```
matthew7838: We should try not to abbreviate names.

RUFAS/EEE/energy.py
```diff
@@ -0,0 +1,391 @@
...
+    ) -> float:
+        """
+        General estimate of diesel fuel consumption for a given attachment type and tractor size.
+        Different practices use different types of tools/implements; the equation to estimate diesel fuel consumption
+        may be the same across practices, but different implements have different parameter values.
+
+        Parameters
+        ----------
+        crop_yield: float
+            Amount of crop yielded per hectares (metric ton/ha)
```
matthew7838:
```suggestion
            Amount of crop yielded per hectares (metric ton/ha).
```

RUFAS/EEE/energy.py
```diff
@@ -0,0 +1,391 @@
...
+        field_production_size: float
+            The filed area under production (ha)
```
matthew7838:
```suggestion
            The filed area under production (ha).
```

RUFAS/EEE/energy.py
```diff
@@ -0,0 +1,391 @@
...
+            The clay percentage of the field under production (unitless).
+        application_mass : float | None = None
+            The mass of a manure or fertilizer application (kg).
+        application_dm_content : float | None = None
+            The dry matter content of a manure or fertilizer application (kg).
+
+        Returns
+        -------
+        float
+            Diesel Consumption for Tractor-Implement (l/ha)
```
matthew7838:
```suggestion
            Diesel Consumption for Tractor-Implement (l/ha).
```

RUFAS/EEE/tractor_implement.py
```diff
@@ -0,0 +1,192 @@
...
+        ]:      # 418c
+            return (self.throughput / application_mass) * self.field_efficiency
+
+        return (
+            self.field_speed_km_per_hr
+            * GeneralConstants.KM_TO_M
+            * self.width_m
+            * self.field_efficiency
+            * GeneralConstants.SQUARE_METERS_TO_HECTARES
+        )       # 418a
```
matthew7838, line 133:
```suggestion
        ) 
```
These should either be removed or added to the docstrings.

RUFAS/EEE/tractor_implement.py
```diff
@@ -0,0 +1,192 @@
...
+        Calculates the Field Capacity for a specific crop, field operation and tractor implement.
+        Implements Helper Functions 418a and 418b in EEE Functions file.
+        """
+        if self.operation_type == OperationType.COLLECTION and self.harvest_type != HarvestOperation.KILL_ONLY:  # 418b
+            return (self.throughput / crop_yield_ton_per_ha) * self.field_efficiency
+        elif self.operation_type in [
+            OperationType.LIQUID_MANURE_APPLICATION_BELOW_SURFACE,
+            OperationType.LIQUID_MANURE_APPLICATION_SURFACE,
+        ]:      # 418c
+            return (self.throughput / application_mass) * self.field_efficiency
```
matthew7838, line 125: Correct me if I'm wrong. I think there's a chance that this application mass being none could lead to an error.

RUFAS/EEE/energy.py
```diff
@@ -0,0 +1,391 @@
...
+        for implement in tractor.implements:
+            crop_yield_ton_ha = crop_yield * GeneralConstants.KILOGRAMS_TO_MEGAGRAMS
+            if application_mass and application_dm_content:
+                application_mass_per_ha = (application_mass * GeneralConstants.KILOGRAMS_TO_MEGAGRAMS
+                                           / application_dm_content) / field_production_size
+            else:
+                application_mass_per_ha = None
+
+            total_power_needed_kW = self._calculate_total_power_needed(
+                tractor, implement, crop_yield_ton_ha, field_production_size, clay_percent, application_mass_per_ha
```
matthew7838, line 337: If we went through line 334, I think we will encounter an error in my other comment below.

RUFAS/EEE/tractor_implement.py
```diff
@@ -0,0 +1,192 @@
...
+        self.tillage_implement = tillage_implement
+        self.harvest_type = harvest_type if self.operation_event == FieldOperationEvent.HARVEST else None
+        constants = input_manager.get_data("EEE_constants.constants")
+        constants_by_ID = Utility.convert_list_to_dict_by_key(constants, "ID")
+        self.field_speed_km_per_hr: float = constants_by_ID[FIELD_SPEED_CONSTANT_ID][
+            "Value"
+        ]  # Constant 585 in EEE Functions file
+        self.field_efficiency: float = constants_by_ID[FIELD_EFFICIENCY_CONSTANT_ID][
+            "Value"
+        ]  # Constant 587 in EEE Functions file
```
matthew7838, line 57: Should be moved/removed. Same as above.

RUFAS/EEE/tractor_implement.py
```diff
@@ -0,0 +1,192 @@
...
+        self.determine_implement_parameters(application_depth)
+
+    def determine_implement_parameters(self, application_depth: float | None) -> None:
+        """
+        Assign a tractor implement based on the operation, tractor size, and crop type where applicable. Not all
+        operations depend on the crop type.
+        Implements Helper Function 419 in EEE Functions file.
```
matthew7838: This function is worth a more formal docstring. Split this into attributes, reference sections, etc.

RUFAS/EEE/tractor_implement.py
```diff
@@ -0,0 +1,192 @@
...
+                break
+
+    def field_capacity_ha_per_hr(
+        self, crop_yield_ton_per_ha: float | None, application_mass: float | None = None
+    ) -> float:
+        """
+        Calculates the Field Capacity for a specific crop, field operation and tractor implement.
+        Implements Helper Functions 418a and 418b in EEE Functions file.
+        """
+        if self.operation_type == OperationType.COLLECTION and self.harvest_type != HarvestOperation.KILL_ONLY:  # 418b
```
matthew7838, line 119:
```suggestion
        if self.operation_type == OperationType.COLLECTION and self.harvest_type != HarvestOperation.KILL_ONLY: 
```

RUFAS/EEE/tractor_implement.py
```diff
@@ -0,0 +1,192 @@
...
+            return (self.throughput / crop_yield_ton_per_ha) * self.field_efficiency
+        elif self.operation_type in [
+            OperationType.LIQUID_MANURE_APPLICATION_BELOW_SURFACE,
+            OperationType.LIQUID_MANURE_APPLICATION_SURFACE,
+        ]:      # 418c
```
matthew7838, line 124:
```suggestion
        ]: 
```

RUFAS/util.py
```diff
@@ -57,7 +58,11 @@ def convert_dict_of_lists_to_list_of_dicts(dict_of_lists: dict[str, list[Any]])
             A list of dictionaries with string keys and integer values.
 
         """
-        return [dict(zip(dict_of_lists.keys(), values)) for values in zip(*dict_of_lists.values())]
+        try:
+            return [dict(zip(dict_of_lists.keys(), values)) for values in zip(*dict_of_lists.values())]
+        except AttributeError:
+            print(dict_of_lists)
```
matthew7838: Is this intended? If for debugging should be removed.

tests/test_biophysical/test_crop_soil_field/field_tests/test_field.py
```diff
@@ -816,6 +822,7 @@ def test_record_planting(
 
         clay.assert_called_once()
     actual = field.om.variables_pool[f"Field._record_planting.crop_planting.field='{field_name}'"]
+    print(actual)
```
matthew7838:
```suggestion
```

matthew7838 (approved):
Left quite a few comments, not because it's a major problem, it's just that the PR is so large.

RUFAS/EEE/tractor_implement.py
allisterakun: Yes, but it is only relevant when the operation is LIQUID_MANURE_APPLICATION_BELOW_SURFACE or LIQUID_MANURE_APPLICATION_SURFACE. Which we should have a value for this variable

PR #1300, opened 2024-02-15
Housing emissions correction for CBPB and Open Lots

### LINKED ISSUE #1247: Housing emissions for CBPB and Open Lot pen types
Housing emissions from manure in RuFaS are defined as the gas emissions from manure that occur while manure is still in the pen with the animals, i.e., before manure is removed to manure treatment/storage. It's important to account for these emissions separately because they both represent a considerable portion of total emissions across the manure chain, and they represent a loss of nutrients from the daily manure mass that eventually enters storage. Of note is that the method to calculate housing emissions **depends on the pen type**. The methods for each pen type are as follows (using generic names just to demonstrate overlap between pen types):
- Method 1: Freestall & tiestall 
- Method 2: CBPB
- Method 3: Open Lot

Presently, in manure_handler_classes.py, the ManureHandlerDailyOutput is returning these 3 housing emissions variables:
	
```
	housing_methane_emission
	housing_carbon_dioxide_emission
	housing_ammonia_emission
```
	
These 3 values are calculated using corresponding functions in calculator.py. Presently, it appears the housing emissions (CH4, CO2, NH3) values are being calculated using **methods only applicable to freestall/tiestall** (e.g., applying method 1 to all pen types). Meaning, ManureHandlerDailyOutput is returning incorrectly calculated housing_emissions values for CBPB and Open Lot pens.

## Possible Implementation
1. We could add pen-type based logic similar to what is shown below (from Niko's TAN PR, not yet merged) to housing_methane_emission and housing_carbon_dioxide_emission in manure_handler_classes.py, to prevent incorrect housing emissions outputs of these variables being returned for CBPB and OL pens. 
image

For the future, we should also consider if/how housing emissions from CBPB and OL pens should be calculated and named, depending on if CBPB and OL are being used as _both_ the pen type and manure treatment method, or as _only_ the pen type followed by composting or field application.

Isgotkowitz:
This PR fixes the errant GHG emissions calculations for manure pen types that are not freestall or tiestall. If the pen is not a freestall or tiestall, there are no GHG emissions recorded for it.

Closes #1247


ew3361zh (approved):
LGTM!

elle-andreen (approved):
Looks great, thanks Ed!

Issue #1161, opened 2024-01-25
Delete messages printed to standard output when running simulations.


Isgotkowitz:
The messages printed in the terminal need to be improved in some ways and improved in others when running RuFaS simulations.

## Current behavior
When running RuFaS simulations, the current message(s) printed to standard out look like this:
```sh
$ python main.py -i -c
RuFaS: Ruminant Farm Systems Model 2023
Simulating...
/
Simulation Successful

/
Simulation Successful

/
Simulation Successful

$
```

## Proposed Improvements
1) It's 2024 now, not 2023.
2) After a simulation finishes successfully, it does not immediately go into the next simulation, it writes all the output files first. So repeating the "Simulating..." message when a new simulation actually starts would be helpful.
3) Either get the spinning progress wheel to work or get rid of it.
4) Add the simulation prefix from the `MetadataPaths` into the "Simulating..." message.

allisterakun:
Suggestion: remove all printed messages to stdout, report all messages to OM instead. Possibly adding a function in OM to handle all the status messages.

Isgotkowitz:
Planning meeting: add "Simulating" and "Simulation N% Successful" messages to logs. Add simulation n% successful messages at 10%* intervals. *Just a suggestion, would be a good place to start.

Also add a verbosity for credits, when true the "RuFaS 2024" credit line is printed.

ew3361zh:
Moving to Blocked until Pooya has merged in his PR refactoring ScenarioManager and main.py.

PR #1302, opened 2024-02-16
Show error and warning counts in the console

### LINKED ISSUE #1251: Display in the console the number of warnings and errors and/or links to the OM warning and error log files
I often forget to open the warning and error log files after the simulation finishes. So, I think having some form of a reminder would be very helpful. It seems incomplete to say that the simulation runs successfully but there are warnings and errors I'm not aware of right away. 

My proposed solution is to display the counts of warnings and errors at a minimum and also have an opt-in command line argument option to display the links (absolute or relative) to the location of the log files at the end of the simulation. 

To keep it even more minimal, only display the counts of warnings and errors if there is at least one warning/error.

ltp23:
Related to #1251 that I wrote.

When there is a warning or an error during the simulation, the console will display the counts and optionally show the links to files using the command line flag `-e` or `--exceptions`. I'm also debating whether to include this flag or not, since we already have the verbose flag (`-v`, `--verbose`) that prints out warnings and errors as the simulation progresses.

I haven't tested to see how things look when we run multiple scenarios.


ew3361zh:
It works great, Loi. Nice work and great idea for a feature! 

As a side note, I think I'd find it helpful to have it print that there were 0 errors and 0 warnings. Just my opinion though.

I think we had a discussion in the work-together meeting last week about how this summary of error and warning counts might fit well with a Verbosity setting that is just the core "Simulation running...Simulation complete" type messages. But we'd need to add a new Verbosity to that enum class and change the default Verbosity setting from none to this new member.

What do you think about folding both of these options into a new Verbosity enum?

Isgotkowitz (changes requested):
I like the addition of the counts of errors and warnings being printed at the end of the simulation.

However the `-e` flag is unnecessary in my opinion. The errors and warnings files are already labeled pretty clearly and placed right alongside all the other outputs, so it seems unnecessary to me to have their paths be added to stdout. I am requesting changes on this, but if you feel strongly about it I can come back and approve.

I disagree with @ew3361zh on folding #1161 into this PR. I think it would be better to keep it limited to #1251, and 1161 left for a future sprint - but that is up to @ltp23.

> I haven't tested to see how things look when we run multiple scenarios.

I ran multiple scenarios, IMO it looks fine. But you should test it yourself - don't just take my word for it.

ltp23:
> As a side note, I think I'd find it helpful to have it print that there were 0 errors and 0 warnings. Just my opinion though.

When there are 0 errors and 0 warnings, it will print out "No errors or warnings found." I made a recent change to show both error and warning counts when either count is not zero.

ltp23:
> I think we had a discussion in the work-together meeting last week about how this summary of error and warning counts might fit well with a Verbosity setting that is just the core "Simulation running...Simulation complete" type messages. But we'd need to add a new Verbosity to that enum class and change the default Verbosity setting from none to this new member.
> 
> What do you think about folding both of these options into a new Verbosity enum?

I don't have a strong preference on this so after we have defined the behaviors we want, we can make changes. Maybe the default log verbosity level should be set at `minimal`, so we can still see the initial messages and error and warning counts. Another thought I have is that the current None verbosity level should print out absolutely nothing.

ltp23:
> However the `-e` flag is unnecessary in my opinion. The errors and warnings files are already labeled pretty clearly and placed right alongside all the other outputs, so it seems unnecessary to me to have their paths be added to stdout. I am requesting changes on this, but if you feel strongly about it I can come back and approve.

I can remove this flag option because it still seems niche to me.

ltp23:
When one or multiple scenarios are used, I think the console should at least print out the name of each scenario based on its prefix.

RUFAS/output_manager.py
```diff
@@ -114,6 +114,8 @@ def __init__(self) -> None:
                     "function": self.__init__.__name__,
                 },
             )
+            self.error_log_file_paths: List[Path] = []
+            self.warning_log_file_paths: List[Path] = []
```
Isgotkowitz: Do these still need to be in here if the `-e` flag is removed?

Isgotkowitz (approved):
Looks good, nice job Loi.

> When one or multiple scenarios are used, I think the console should at least print out the name of each scenario based on its prefix.

Strongly agreed, it is part of #1161.

> Maybe the default log verbosity level should be set at `minimal`, so we can still see the initial messages and error and warning counts. Another thought I have is that the current None verbosity level should print out absolutely nothing.

I also agree with this. I think a good way of approaching this would be to make those changes to the OM so it supports the minimal option first, then tackle changing the output after that.

JoeWaddell (approved):
LGTM: tested under different circumstances, and it works nicely.

ltp23:
Thanks everyone for reviewing the PR!

PooyaHekmati (commented):
This is a good addition, however, the PR should have not been approved in its current state  because:

1. The description mentions `show the links to files` but the code does not do that.
2. `get_error_warning_counts` should be a method in `OutputManager` not a dangling function in main.
3. We need to have full control over what we print to console via `stdout`. I know that we already have few cases where we always print something, but the existing code should not be used as a guidance for new code. When adding new `stdout` usages, we have to be considerate of this.
4. > I haven't tested to see how things look when we run multiple scenarios.
5. The function `show_error_warning_counts` seems extraneous. It could be replaced with:
`sys.stdout.write(f"{error_count} error(s) and {warning_count} warning(s) found.\n")`

A follow up PR is needed to address these. Re-opening #1251 .

ltp23:
> 1. The description mentions `show the links to files` but the code does not do that.

After some feedback, it seemed redundant to show the file paths. I originally wanted to show them because when I ran the program in my terminal, I could click on the links to open the files, so I thought this would be nice. But not all terminals do that.

> 2. `get_error_warning_counts` should be a method in `OutputManager` not a dangling function in main.

Okay. I can move it into the output manager class.

> 3. We need to have full control over what we print to console via `stdout`. I know that we already have few cases where we always print something, but the existing code should not be used as a guidance for new code. When adding new `stdout` usages, we have to be considerate of this.

To enable this, we can store all the usages of stdout into an object and display them only under certain settings.

> 4. > I haven't tested to see how things look when we run multiple scenarios.

I had run the program with two scenarios and it looked fine, but I did make a comment about showing the name of each scenario.

> 5. The function `show_error_warning_counts` seems extraneous. It could be replaced with:
>    `sys.stdout.write(f"{error_count} error(s) and {warning_count} warning(s) found.\n")`

I will remove that function and use this instead.

PR #1305, opened 2024-02-16
Optimizing ration optimization

JoeWaddell:
Added logic to optimize to use previous month's ration - if it exists - in the `optimize` method, itself an integral part of the ration formulation process. 

The implementation on main uses a random series of values as the starting point for every formulation. Here, while the first formulation necessarily begins from the random set of values in the `scipy.minimize` method, subsequent optimizations uses starting values based on the previous ration. This essentially means that the minimize method already begins near the likely solution.

This saves time (EDIT: about 10% of runtime based on a handful of tests using large herd sizes), and also seems to avoid "swingy" formulation in growing and close up pens (note the jumps in DMI on main & default metadata). It doesn't appear to substantially affect the lactating pen. My intuition is that since growing and close up rations are easier to formulate (e.g. its easier to meet animal requirements), it can occasionally find high DMI solutions. This new method seems to keep the solution from "wandering" too far from the most common solution. 

This is just speculation however, based on the feed lists on main. In review we should try different sets of Feed inputs.


RUFAS/routines/animal/ration/ration_optimizer.py
```diff
@@ -884,7 +885,9 @@ def make_user_bounds(
         return tribounds
 
     def optimize(
-        self, animal_combination, available_feeds: Dict, ration_config: RationConfig
+        self, animal_combination: AnimalCombination,
+        ration_config: RationConfig,
+        previous_ration: Dict[str, float | str] = None
```
ltp23: I think this is missing the `| None`
```
previous_ration: Dict[str, float | str] | None = None
```
and some other places.

RUFAS/routines/animal/ration/ration_optimizer.py
```diff
@@ -899,6 +902,8 @@ def optimize(
             a DefaultDict of the AvailableFeeds class attributes defined in ration_driver.py
```
ltp23: There is an extra parameter in the docstring - `available_feeds`.

RUFAS/routines/animal/ration/ration_optimizer.py
```diff
@@ -912,10 +917,20 @@ def optimize(
...
-            x0.append(random.random() * 10)
+        if previous_ration:
+            prev_ration = previous_ration.copy()
+            del prev_ration['status']
+            del prev_ration['objective']
+            x0 = []
+            for key in prev_ration:
+                x0.append(prev_ration[key] / 3)
+                x0.append(prev_ration[key] / 3)
+                x0.append(prev_ration[key] / 3)
```
ltp23: I think to minimize copying, you could do something like this

```
x0 = [value / 3 for key, value in previous_ration.items() 
          if key not in ["status", "objective"] and isinstance(value, float)]
```

The `isinstance` part is not necessary but included to make the type checker happy because it can't tell if a value is of type str or float.

RUFAS/routines/animal/ration/ration_optimizer.py
```diff
@@ -912,10 +917,20 @@ def optimize(
...
+        else:
+            n = len(ration_config.price_list)
+            x0 = [1]
+            for _ in range(n - 1):
+                x0.append(random.random() * 10)
```
ltp23: An alternative if you like,

```
x0 = [1] + [random.random() * 10 for _ in range(n - 1)]
```

ltp23 (approved):
Thanks Joe for working on this! I left some minor code style comments, and I'm curious if this update helps speed things up here and there regarding the ration optimization because the logic is sound.

jadamchick:
Hi Joe,
I ran this on several pilot farms over the weekend; however I didn't realize until this morning that because the feed library in your branch lacks feeds 301-305, it would not really match their rations well (i.e. most of them call for at least one of those mixes) -- so it's hard to make judgements when comparing apples to oranges. 
I could compare for one herd (c2f2) and while the DM fed differs for several feeds, I suspect it isn't that important? But this really isn't my area of expertise to judge. The feeds in the table are those with a 5% or greater difference between the RO branch (average of 4 runs) and original ration (avg of 4 runs). 


<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/jms349/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/jms349/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">


category | RO as % of original | DM RO branch | DM original |   |  
-- | -- | -- | -- | -- | --
DMI tot | 100% |   |   |   |   |  
DMI by feed_108_ver_agg | 88% | 13451278.68 | 15336186 | legume hay |  
DMI by feed_230_ver_agg | 39% | 32928.49833 | 85016.47 | calcium carbonate |  
DMI by feed_231_ver_agg | 58% | 31346.69432 | 53989.4 | dicalcium phosphate dibasic
DMI by feed_232_ver_agg | 220% | 125327.4609 | 57094.61 | calcium phosphate monobasic
DMI by feed_239_ver_agg | 106% | 61085.38154 | 57443.74 | sodium chloride |  
DMI by feed_248_ver_agg | 107% | 65516.11788 | 61208.82 | calcium chloride anhydrous
DMI by feed_110_ver_agg | 110% | 21628959.85 | 19698696 | legume silage |  
  |   |   |   |   |   |  



</body>

</html>


A couple questions/thoughts: 
- do you have criteria/guidelines to use to judge whether the rations are "different" or not? or at least key outcomes to pay attention to? 
- how many herds do you want to test this on? I need to prioritize running herds to generate reports for the next few days, but can choose a few quick ones to sneak in later today (and will add the updated Feed Library to the branch I use when I run them).  
- alternatively, I can upload the Feed Library and a few pilot farm feed files onto this branch if you want to try using them to see the results for yourself (and can delete them before merging)

JoeWaddell:
Thanks for your thoughts on this @jadamchick! The key outcomes to look for differences is to see if the automatically formulated rations seem reasonable. That's a broad ask (sorry I have nothing more specific to suggest!), but that's because the only real difference when compared to main is that after the first formulated ration (which maintains the same behavior as found on main), it may tend to formulate rations similar to the first one - since the starting point of the optimization process uses the previously formulated ration. 

The testing I've done shows a more stable ration formulation process, so using a few different sets of feed IDs should be enough to stress test it. Also instead of adding your expanded Feed Library to this branch: you can actually just add the Feed Library to your local branch and then not revert the change instead of committing.

jadamchick:
Thanks Joe. When you say "automatically formulated ration", does that mean make sure that user_input=false? 

I'm not sure how to gauge "reasonable", but I can give you the results for a few more herds (as here) and you can make that judgment call :) 
And yes to keeping it on my local branch -- I was just offering to send you the files in case you wanted to run them instead of me ;)

JoeWaddell:
> Thanks Joe. When you say "automatically formulated ration", does that mean make sure that user_input=false?

Correct!

> I'm not sure how to gauge "reasonable", but I can give you the results for a few more herds (as here) and you can make that judgment call :) And yes to keeping it on my local branch -- I was just offering to send you the files in case you wanted to run them instead of me ;)

Ah, understood, thanks! We can coordinate that on Slack - if it's what's on the Pilot branch, I'll get your guidance on which I should use to test any future ration tweaks, too.

jadamchick (approved):
Runs smoothly for a variety of farms without crazy outrageous ration results.

Issue #1269, opened 2024-02-13
Volatile solids initialization in the manure module 


tomhuhh:
Currently the manure module only retrieves the total volatile solid (VS) output from the animal module. To obtain degradable VS (VSd) and nondegradable VS (VSnd), it uses an assumed ratio between VSd and VS which is `0.2/0.48 = 0.41` as suggested by the Chianese et al. 2009 and Sommer 2004 paper. 

However, this creates an disconnection between the outputs of animal module and the inputs of manure module. We potentially lose the opportunity to evaluate the impact of animal factors on manure gas emissions as predicted by VSd and VSnd. 

Instead, we should consider passing on the VSd and VSnd output from the animal module to the manure module and using these values in the calculations of, for example, storage methane.    

## Expected Behavior

I would expect the manure module to directly use the outputs of animal module, i.e. VSd, VSnd, when calculating emissions such as storage methane emissions.  

## Current Behavior

The manure module uses the VS output from the animal module and an assumed ratio between VSd and VS to calculate VSd and VSnd, and then calculates gas emissions such as storage methane emissions. 

This assumed ratio is from studies done in Europe, and the potential issue with that is that the European diets typically is higher in forage (i.e. alfalfa) which can lead to lower VSd to VS ratio. 

## Possible Solution
 

I would first check how the animal module outputs VSd and VSnd. Then I would add these two variables to the input attribute of the manure module. In the `calculator.py` file, I would then use VSd and VSnd for calculations of for example, storage methane.  

## Steps to Reproduce

1.
2.
3.
4.

elle-andreen:
Haowen and I created [this document](https://docs.google.com/document/d/1EfYQUIwoIrt05yWLauEefcdTlgkGkk0WX6EoVqrL6eQ/edit) that has additional detail on this issue. Please note as per the doc that issues 1269-1271 are interrelated and build upon one another. Let us know if there are further questions or clarifications needed!

PR #1306, opened 2024-02-17
Update methane emission from slurry storage

### LINKED ISSUE #1271: Storage methane emission calculations
Currently the `methane_emission_from_slurry_storage` is not correctly implemented. There seem to be some flaws with the approach proposed by Chianese et al. 2009  (https://www.researchgate.net/publication/43286301_Simulation_of_Methane_Emissions_from_Dairy_Farms_to_Assess_Greenhouse_Gas_Reduction_Strategies). More recent literature (Peterson et al., 2016, Peterson et al. 2024) refers back to the original paper (Sommer et al. 2004; https://doi.org/10.1023/B:FRES.0000029678.25083.fa) when they present the storage methane emission calculations. 

Here is a comparison between the two approaches. As you can see, on a daily basis, the Chianese approach overestimated the amount of storage methane by about 20 times.  

image

In addition, more recent literature (Peterson et al. 2016: 10.1371/journal.pone.0160968; Peterson et al. 2024: https://doi.org/10.1016/j.jenvman.2024.120233) suggest updates for constants used in the `methane_emission_from_slurry_storage` function. Please also check in the discussion section of the Peterson et al. 2016 paper for more details on why the updates were made.  

Screenshot 2024-02-13 at 11 37 53 AM

To summarize the changes in the constants used, we should use an Ea of 81 instead of 112.7 KJ/mol and lnA of 31.2 instead of 44.3.  Also to note, these two constants are great fit for sensitivity analysis of model constants.    

## Expected Behavior

The `methane_emission_from_slurry_storage` function correctly calculates daily storage methane emissions from accumulated manure storage (volatile solids). 

## Current Behavior

The approach used overestimated the daily storage methane emissions and some constants are outdated based on more recent publications from the same group.  

## Possible Solution

Update the  `methane_emission_from_slurry_storage` function based on the most recent publication (Peterson et al. 2024).  

## Steps to Reproduce

1. Simulate slurry storage and make sure your filter has storage methane and enteric methane.  
2. Compare storage methane to enteric methane. These two values should be somewhat within each other's range because they account for the two major sources for methane on dairy farms. 
3.
4.

ltp23:
Closes #1271 

[Docs](https://docs.google.com/document/d/1EfYQUIwoIrt05yWLauEefcdTlgkGkk0WX6EoVqrL6eQ/edit)

This PR updates the slurry storage methane emission calculations based on the suggestions from the design doc.

To see the complete results, I use this filter

```
{
  "multiple": [
    {
      "name": "Storage methane, kg",
      "filters": [
        "^(?:AnaerobicDigestion_)?ManureTreatmentDailyOutput_Pen_.*storage_methane$"
      ]
    }
  ]
}
```


RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -23,8 +23,8 @@ def methane_emission_from_slurry_storage(
 
         .. math::
 
-            E_{CH_4} = 24 \\cdot VS_{d} \\cdot b_{1} \\cdot e^{lnA - \\frac{E}{RT}}
-            + 24 \\cdot VS_{nd} \\cdot b_{2} \\cdot e^{lnA - \\frac{E}{RT}}
+            E_{CH_4} = 24/1000 \\cdot VS_{d} \\cdot b_{1} \\cdot e^{lnA - \\frac{E}{RT}}
+            + 24/1000 \\cdot VS_{nd} \\cdot b_{2} \\cdot e^{lnA - \\frac{E}{RT}}
```
tomhuhh, line 28: Hi @ltp23 ! This equation does not seem to be correct. Please refer to this document for the most up-to-date information- https://docs.google.com/document/d/1EfYQUIwoIrt05yWLauEefcdTlgkGkk0WX6EoVqrL6eQ/edit. To highlight what I am referring to, we should remove '24' in both locations. Thanks!

RUFAS/routines/manure/constants_and_units/gas_emission_constants.py
```diff
@@ -24,10 +24,10 @@ class GasEmissionConstants:
     Rate correcting factor for non-degradable volatile solids, used in calculation of slurry storage
     methane emission (unitless)."""
 
-    NATURAL_LOG_ARRHENIUS_CONSTANT: float = 43.33
+    NATURAL_LOG_ARRHENIUS_CONSTANT: float = 31.2
     """Natural log of the Arrhenius constant (unitless)."""
 
-    ACTIVATION_ENERGY: float = 112700.0
+    ACTIVATION_ENERGY: float = 81_000.0
     """
```
tomhuhh, line 28: I did not know we can use '_' inside the numbers. This is good to know!

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -92,17 +97,28 @@ def methane_emission_from_slurry_storage(
...
+        degradable_volatile_solids_fraction = degradable_volatile_solids / total_volatile_solids
+        non_degradable_volatile_solids_fraction = non_degradable_volatile_solids / total_volatile_solids
+
         methane_emission_from_degradable_volatile_solids = (
-            GasEmissionConstants.METHANE_EMISSION_COEFFICIENT
-            * degradable_volatile_solids
-            * GasEmissionConstants.DEGRADABLE_VOLATILE_SOLIDS_RATE_CORRECTING_FACTOR
-            * arrhenius_exponent
+            (
+                GasEmissionConstants.METHANE_EMISSION_COEFFICIENT
```
tomhuhh: As mentioned above, this should be deleted. We should also remove it from the constant file. Thanks!

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -92,17 +97,28 @@ def methane_emission_from_slurry_storage(
             non_degradable_volatile_solids,
         ) = cls._volatile_solid_components(total_volatile_solids)
 
+        degradable_volatile_solids_fraction = degradable_volatile_solids / total_volatile_solids
+        non_degradable_volatile_solids_fraction = non_degradable_volatile_solids / total_volatile_solids
```
tomhuhh: It is great that we are calculating the fractions of degradable and non-degradable volatile solids here. A further step is to also pass through the degradable and non-degradable volatile solids in the argument of this function. Thanks!

tomhuhh (changes requested):
Thanks Loi! This is a great start point. Please see my comments. I hope we can get this function sorted out in the next couple of days. Keep up the good work!

RUFAS/routines/manure/constants_and_units/gas_emission_constants.py
```diff
@@ -24,10 +24,10 @@ class GasEmissionConstants:
     Rate correcting factor for non-degradable volatile solids, used in calculation of slurry storage
     methane emission (unitless)."""
 
-    NATURAL_LOG_ARRHENIUS_CONSTANT: float = 43.33
+    NATURAL_LOG_ARRHENIUS_CONSTANT: float = 31.2
     """Natural log of the Arrhenius constant (unitless)."""
```
elle-andreen: We can update the units of this value from "unitless" to "g CH4/kg manure VS/h"

RUFAS/routines/manure/gas_emissions/calculator.py
ltp23: Just to clarify, do you want to remove both the `arrhenius_exponent` and the `correcting_factor` while keeping the `methane_emission_coefficient`?

RUFAS/routines/manure/gas_emissions/calculator.py
ltp23: I can do the fraction calculations inside the volatile components function.

RUFAS/routines/manure/gas_emissions/calculator.py
elle-andreen: We would remove only `GasEmissionConstants.METHANE_EMISSION_COEFFICIENT` (this is where the value of 24 that Haowen referred to above is coming from). `degradable_volatile_solids` would become `degradable_volatile_solids_fraction`. The other terms in lines 96-99 are fine as-is!

RUFAS/routines/manure/gas_emissions/calculator.py
tomhuhh: Thanks! That is a great idea. In addition, we need to figure out a way to pass through the `degradable_volatile_solids` and `non_degradable_volatile_solids` from the animal module and serve as the input for the volatile components function.

ltp23: This would need some redesign. Currently, the two components merge into one after reception pit.

KFosterReed: I want to add that the passage of the degradable and non degradable VS from the animal module will be addressed in a separate PR that addresses issue #1269.

tomhuhh (approved):
LGTM!

elle-andreen (approved):
Accounting for the daily update of VS not being altered yet, the daily methane values here look good! Thank you Loi!

ltp23:
Thanks everyone! So I think the next step is to try to pass along the degradable and non-degradable VS values down to the treatment method so that we can use them for calculating storage methane.

PR #1317, opened 2024-02-21
Reduce time spent in IM's get_data()

ltp23:
In this PR, I want to cache the results of repeated calls to IM's `get_data()` by using one of the built-in features, LRU caching and memoization, which could also be implemented from scratch. This caching helps reduce recomputation time but requires that data be not stale. The cache needs to be manually cleared if later on we decide to allow dynamic changes to the input data pool while the simulation is running.

The current implementation of `get_data()` has two bottlenecks: 1) retrieval of nested data (which is not too bad because the input data we have are no more than a constant number of levels deep (say, 10) and mostly shallow anyway) and 2) deep copying of returned data (more serious).

Although the number of calls to get_data() stays the same, only those that actually need to walk through the data pool will get logged, hence, the apparent significant reduction in the number of function calls. 

There are two data access patterns I see when retrieving the data from the input manager: 1) retrieve one big blob of data once, save it, look things up when needed, and 2) retrieve individual pieces of data directly. The first pattern is efficient if the data never changes but extra memory is needed. The second pattern will make lots of independent calls but always get fresh input data. I think both patterns are valid and the input manager should be robust enough to handle both styles.

### DIFF
--- a/RUFAS/input_manager.py
+++ b/RUFAS/input_manager.py
@@ -1094,7 +1094,7 @@ class InputManager:
         om.add_warning("Validation: data fixed", warning_message, info_map)
         return True
 
-    # @lru_cache(maxsize=128)
+    @lru_cache(maxsize=128)
     def get_data(self, data_address: str) -> Any:
         """
         Get the requested data from the pool.
--- a/RUFAS/input_manager.py
+++ b/RUFAS/input_manager.py
@@ -1,5 +1,5 @@
 from copy import deepcopy
-from functools import reduce
+from functools import reduce, lru_cache
 import json
 import os
 import re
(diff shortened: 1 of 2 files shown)


PooyaHekmati (changes requested):
The following code:
```python
from functools import lru_cache
import copy

class DataPool:
    def __init__(self):
        self.pool = {
            'key1': {'data': 100, 'nested': {'more_data': 200}},
            'key2': {'data': 300, 'nested': {'more_data': 400}}
        }

    def get_data(self, key):
        return copy.deepcopy(self.pool.get(key, None))

    @lru_cache(maxsize=128)
    def get_data_with_cache(self, key):
        return copy.deepcopy(self.pool.get(key, None))

if __name__ == "__main__":
    dp = DataPool()

    data1 = dp.get_data('key1')
    print("Original data1:", data1)
    data1['nested']['more_data'] = 999 
    print("Modified data1:", data1)
    data1_2 = dp.get_data('key1')
    print("data1_2:", data1_2)

    data2 = dp.get_data_with_cache('key2')
    print("\nOriginal data2:", data2)
    data2['nested']['more_data'] = 888 
    print("Modified data2:", data2)
    data2_2=dp.get_data_with_cache('key2')
    print("data2_2:", data2_2)

    data3 = dp.get_data_with_cache('key2')
    print("\nData3 fetched again:", data3)
    print("Comparing object IDs of data2 and data3 to check for shared objects:", id(data2['nested']) == id(data3['nested']))
```
gives the following output:

> Original data1: {'data': 100, 'nested': {'more_data': 200}}
> Modified data1: {'data': 100, 'nested': {'more_data': 999}}
> data1_2: {'data': 100, 'nested': {'more_data': 200}}
> 
> Original data2: {'data': 300, 'nested': {'more_data': 400}}
> Modified data2: {'data': 300, 'nested': {'more_data': 888}}
> data2_2: {'data': 300, 'nested': {'more_data': 888}}
> 
> Data3 fetched again: {'data': 300, 'nested': {'more_data': 888}}
> Comparing object IDs of data2 and data3 to check for shared objects: True

Therefore, LRU Cachig defeats the purpose of having deep copy in `get_data`, which in turn violates the SSOT requirement of Input Manager.

Side note 1: The Notes section of the docstring seems to be misplaced (shouldn't be before Parameters)
Side note 2: As we have talked about it in multiple meetings, it's been a while since @allisterakun implemented the capability of modifying `IM.pool` during runtime. That can make the cached data be incorrect. I don't see any mechanism in this PR to fix that.
Side note 3: @Isgotkowitz is working on improving IM, feel free to route suggestions his  way.

ltp23:
Right, the use of caching and modifying the returned copy are incompatible and will affect subsequent calls if the cache is not cleared when those changes are made. Since the requirement is to return a fresh copy for every call, caching becomes irrelevant. So I will close this PR without merging.

PR #1321, opened 2024-02-21
Report_daily_pen_totals

JoeWaddell:
Adding the total number of animals in each pen to OM. Added unit test for report method.


jadamchick (approved):
Works great, thanks Joe!

I can specify to report the pen count for a specific pen, or use the filter below to report each pen without having to know in advance how many there are. That the name contains both the pen number and animal combination is handy for quick reference in the report. 

```
        {
            "name": "Avg count by pen",
            "filters": ["AnimalModuleReporter.report_daily_pen_total.*"],
            "slice_start": -365
        },
```

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -404,6 +404,18 @@ def report_life_cycle_manager_data(life_cycle_manager, sim_day: int) -> None:
         )
         om.add_variable("cull_reason_stats", life_cycle_manager.cull_reason_stats, info_map)
 
+    def report_daily_pen_total(penlist):
+        info_map = {
+            "class": "AnimalModuleReporter",
+            "function": "report_daily_pen_total",
```
Isgotkowitz:
```suggestion
            "class": AnimalModuleReporter.__class__.__name__,
            "function": AnimalModuleReporter.report_daily_pen_total.__name__,
```
having the class and function specified like this will help to avoid the situation where the class or method name changes but the info map doesn't.

Isgotkowitz (approved):
LGTM, nice job Joe. One suggestion related to the info map.

PR #1322, opened 2024-02-21
OM Filter Handling Refactor

### LINKED ISSUE #996: Make graphs from reports

ew3361zh:
Closes #996.

This PR:
- Refactors `OM.save_results()` and creates handler functions for each type of filter file, replacing `OM.route_save_functions()`.
- Adds 2 new report filter fields:
1. `graph_details` - will trigger the report data, once aggregated in report generator, to be graphed. The default setting will be for this data to only be graphed, so it won't be added to the report generated. `graph_details` is a dictionary and will only need the type of graph the user wants. 
e.g. report filter:
```
{
        "name": "Last Year's Average Cow Body Weight (kg)",
        "filters": [
          "LifeCycleManager.daily_update.avg_cow_body_weight"
        ],
        "slice_start": -365,
        "graph_and_report": true,
        "graph_details": {
          "type": "plot"
        }
      },
```
2. `graph_and_report` - a boolean which will trigger the data to be added to the report in addition to being graphed. If this is present and set to True, it will only work if the `graph_details` filter is also present. Otherwise, the data won't be graphed and a warning will be logged telling the user the data wasn't able to be graphed.
e.g. report filter to be graphed AND reported:
```
{
        "name": "DMI by feed daily",
        "filters": [
        ],
        "variables": [
          ".*"
        ],
        "slice_start": -365,
        "graph_and_report": true,
        "graph_details": {
          "type": "plot"
        }
      },
```

Other additions:
In graphing some of the report data, I noticed that the legend would often default to printing directly on top of the graph. If there are more than a few variables being graphed, it easily would cover significant portions of the graph. So this PR updates the standard legend placement to be outside the graph to the right. I tried "best" as the loc placement and 9/10 times would print the legend to the right of the graph but the 1/10 other times would be on top of the graph and would cover up important details. I think this is a good solution and makes the mechanism of legend placement a little more obvious to users who want to customize it further to their liking.

- [ ] figure out whether/how to incorporate Pooya's refactor of `OM._filter_variables_pool()` in #1299 
- [ ] Tied into the first TODO is addressing duplicate code in each handler function for filtering the variables pool
- [x] Add diagram of changes/flow of data
- [ ] Update docstrings
- [ ] Update unit testing


ew3361zh:
Updated refactor flow chart:
save_results_refactor

ew3361zh:
covered in PR #1252

PR #1323, opened 2024-02-21
Update biogas output from Anaerobic Digestion

KFosterReed:
Methane generation in the Anaerobic Digestion should be predicted from the `methane_volume_via_Chen_equation` function in the manure module `calculator.py` file. Although two of the outputs (`methane_generation_volume` and `biogas_energy_content` ) were originating from this method, a third output, `biogas`, was being predicted using a different method. 

Although more review of this treatment method and methane prediction functions are needed, this small update aligns the 3 outputs to facilitate evaluation of the outputs from this manure treatment


ltp23 (approved):
It looks good to me. So, now the biogas energy content and biogas values will differ by a constant factor of 55 (the methane energy density constant).

Thanks Kristan.

tomhuhh (approved):
The change makes sense to me. 

The code runs with no difficulty. I also observed the same ratio (1:55) between biogas and biogas energy content. 

Here is the filter that I used to test it. 

````
ManureTreatmentDailyOutput_Pen_3_LAC_COW.biogas
ManureTreatmentDailyOutput_Pen_3_LAC_COW.biogas_energy_content
Accumulated_ManureTreatmentDailyOutput_Pen_3_LAC_COW.biogas
Accumulated_ManureTreatmentDailyOutput_Pen_3_LAC_COW.biogas_energy_content
````

Thanks Kristan!

PR #1325, opened 2024-02-22
Add units to Manure Module output 

### LINKED ISSUE #1290: Add units to Manure Module output
Title

allisterakun:
Closes issue #1290 
This PR adds units to all Manure module output variables.


RUFAS/routines/manure/manure_handlers/manure_handler_daily_output.py
```diff
@@ -61,32 +62,74 @@ class ManureHandlerDailyOutput(LiquidManurePortionProtocol):
...
+
     total_bedding_mass: float = 0.0
+    total_bedding_mass_unit: str = "kg"
+
     total_water_volume_in_milking_parlor: float = 0.0
-    total_daily_manure_volume: float = field(init=False)
+    total_water_volume_in_milking_parlor_unit: str = "m^3"
+
+    total_daily_manure_volume: Optional[float] = None
+    total_daily_manure_volume_unit: str = "m^3"
```
ltp23: If it still works, I would prefer to keep using `field(init=False)` because then when you create a manure handler output object, you can't accidentally specify a volume during initialization. It is meant to be calculated after initialization, whereas Optional means you can still specify one but choose not to.

RUFAS/routines/manure/manure_handlers/manure_handler_daily_output.py
```diff
@@ -106,3 +149,11 @@ def __post_init__(self) -> None:
...
         )
         self.liquid_manure_daily_volume = self.total_daily_manure_volume
+
+    @property
+    def units_dict(self) -> Dict[str, str]:
+        return {
+            k: v
+            for unit in ({k: v} for (k, v) in self.__dict__.items() if k.endswith("_unit"))
+            for (k, v) in unit.items()
+        }
```
ltp23: I think it would be easier to understand if you could rewrite this the normal way without using dictionary comprehension. Also, this method is reused in each of the daily output classes so it is better to move it somewhere else so it can accept a data class object as a parameter.

RUFAS/routines/manure/manure_nutrients/manure_nutrients.py
```diff
@@ -122,9 +139,10 @@ def __add__(self, other: ManureNutrients) -> ManureNutrients:
         summed_attributes = {
             field.name: getattr(self, field.name) + getattr(other, field.name)
             for field in fields(self)
-            if field.name != "manure_type"
+            if field.name != "manure_type" and not field.name.endswith("_unit")
```
ltp23, line 158: I think this is the same as `if field.name != 'manure_type_unit'`

ltp23 (commented):
Good work, Allister!

One extra thought I have is to add docstrings for the unit variables.

RUFAS/routines/manure/manure_handlers/manure_handler_daily_output.py
allisterakun: Ok! I will revert this!

RUFAS/routines/manure/manure_nutrients/manure_nutrients.py
allisterakun: Emmmm.... I don't think it is. This is saying that if the field name is not "manure_type", and the field name does not end with "_unit", then do something. 

So if the field name is "manure_type" or if the field name is "ABC_unit", then it will not execute the code below.

ltp23: Thanks Allister. That makes sense.

ltp23 (approved):

elle-andreen (approved):
All units look correct to me in the context of the manure module.

PR #1330, opened 2024-02-23
Adds units to non-biophysical variables sent to Output Manager

### LINKED ISSUE #1291: Add units to all outputs outside of biophysical modules
Add units to all outputs that do not come from the Animal, Manure, C&S or Feed Storage modules.

ew3361zh:
closes #1291 

Adding units to all om.add_variable() function calls to any .py files found that are not in MASM/RUFAS/routines/animal, MASM/RUFAS/routines/field, nor MASM/RUFAS/routines/manure.


RUFAS/routines/feed/feed.py
Isgotkowitz, line 1: I think this module is either deprecated or will be soon.

RUFAS/time.py
```diff
@@ -95,9 +95,9 @@ def record_time(self) -> None:
             "function": self.record_time.__name__,
             "prefix": "Time",
         }
-        om.add_variable("day", self.day, info_map)
-        om.add_variable("year", self.year, info_map)
-        om.add_variable("calendar_year", self.calendar_year, info_map)
+        om.add_variable("day", self.day, dict(info_map, **{"units": "simulation day"}))
```
```suggestion
        om.add_variable("day", self.day, dict(info_map, **{"units": "ordinal day"}))
```
I think this is a little more correct. "Simulation day" does not have a precise meaning, but we could strictly define it for RuFaS if we feel it is appropriate.

Isgotkowitz (approved):
LGTM. I am not sure what is going to happen with the feed module, so I didn't dig into what the units in that module are.

RUFAS/simulation_engine.py
```diff
@@ -71,13 +71,15 @@ def simulate(self) -> None:
         t_start_sim = timer.time()
         self._run_simulation_main_loop()
         AnimalModuleReporter.report_end_of_simulation(self.animal_manager, self.day_counter)
+        available_feeds_on_final_day = [
+            {k: v.value if isinstance(v, Enum) else v for k, v in feed.items()}
+            for feed in self.feed_manager.query_available_feeds()
+        ]
+        available_feeds_units = {k: "unitless" for feed in available_feeds_on_final_day for k in feed}
```
JoeWaddell: Looking at `query_available_feeds`, is the plan to update the units here once there are defined units for each feed in storage? I could imagine a way to add the unit for each feed to the unit dict, assuming a unit is associated with each item in feed_manager.

JoeWaddell (approved):
LGTM: just a small question that's is about future design.

RUFAS/simulation_engine.py
ew3361zh: That's a good thought. I hadn't seen that there was an amount category before which I think is currently `kg` for each feed. I'll update that but we could look at making it dynamic if different feeds and different storage methods end up having differing units. Nice catch!

PR #1339, opened 2024-02-26
Refactoring the method for displaying error and warning counts

ltp23:
In this small PR, I moved the method `get_error_and_warning_counts()` from `main.py` to inside the OutputManager class. This method is then called by the `simulate()` in the SimulationEngine class.

I also removed the method `show_error_warning_counts()` and kept the error count message to a minimum.

This is a follow-up PR to address additional comments from issue #1251 and the closed PR #1302

### DIFF
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -1139,30 +1139,3 @@ class OutputManager(object):
         errors_count = sum([len(value_dict["values"]) for value_dict in self.errors_pool.values()])
         warnings_count = sum([len(value_dict["values"]) for value_dict in self.warnings_pool.values()])
         return errors_count, warnings_count
-
-    def add_to_stdout_queue(self, msg: str) -> None:
-        """
-        Adds a message to the stdout queue.
-
-        Parameters
-        ----------
-        msg : str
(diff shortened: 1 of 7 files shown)


RUFAS/output_manager.py
```diff
@@ -114,6 +114,7 @@ def __init__(self) -> None:
                     "function": self.__init__.__name__,
                 },
             )
+            self._stdout_queue: List[str] = []
```
ew3361zh: Is this used anywhere?

ew3361zh (approved):
This looks good considering the feedback on #1302. I think the limited additional stdout is a good compromise for the time between now and when #1161 is addressed.

I wasn't sure the `self._stdout_queue` was needed as it's not referenced anywhere.

Nice work!

Isgotkowitz (approved):
LGTM, nice job Loi

RUFAS/output_manager.py
ltp23: I will remove that. I was working on managing the printout messages but then changed my mind and left it for another PR.

ltp23:
I made another issue #1340 to handle the console messages in a more controlled manner.

Issue #1335, opened 2024-02-23
[IM] Address issues and features in the Input Manager.


Isgotkowitz:
## Overview
There are some issues present in the implementation of the Input Manager that have impeded efforts to expand and improve its functionality. This issue outlines some ways in which the Input Manager can be refactored so that future features and improvements can be made more easily and in a more sustainable manner. Also, there are a number of features which could be implemented in the Input Manager that would greatly enhance its functionality and the functionality of RuFaS overall.

## Issues

- [x] #1487
- [x] #1488
- [x] #1489
- [x] #1490

## Features

- [x] #1273 
- [x] #1274 
- [x] #1331 
- [x] #1332 
- [x] #1333 
- [x] #1424

## Additional context
The problems with the Input Manager described here were investigated in issue #1249.

allisterakun:
During grooming:
@Isgotkowitz will break this into smaller issues, current children issues: #1409 and #1410.

Isgotkowitz:
Blocked because the only issue left is #1331, and that is currently blocked.

Isgotkowitz:
Grooming session: all issues in this issue have been completed, closing this issue

PR #1345, opened 2024-02-28
IM array and nested object validation, and element counting

### LINKED ISSUE #873: IM Element Counter Not Counting Correctly
IM element counter is not correctly functioning when json input data is being validated due to the recursive nature of the json validation method.

## Expected Behavior
The IM counts totals of valid elements, invalid elements, elements fixed, and total elements and reports the numbers to OM. The counts are currently not working correctly.

## Current Behavior
The `_validate_json_element` method is recursive to allow for checking nested dictionary elements. The dictionary tracking the values for the different elements to report to OM is instantiated in this method but gets reset when there is a nested variable that is validated.

## Possible Solution
One solution I quickly tested before writing this up was making the element trackers attributes of the IM and then adding a responsibility to the IM's `flush_pool()` method to also reset these values and it seemed to work well.

## Steps to Reproduce
Outlined by @JoeWaddell in PR #828. He was testing the new CLI verbose messaging and saw that an element was fixed during the validation process but the end result of the validation showed a count 0 elements fixed.

ltp23:
In this PR I modified/refactored several existing methods by creating second versions instead of doing so directly. I also added some new methods. I edited some metadata that got flagged. 

The total number of validated elements has increased from about 530K to 1.7mil elements.

I'm working on finalizing the remaining unit tests.

Attempt to close #873 and #683 

---

Update 1: I have modified in-place all the methods that needed to be changed for this PR. Updated all the relevant unit tests and wrote new ones.

---

Update 2: Fixed merged conflicts to make the PR work again. Compiled a list of follow-up tasks in issue #1401.

---

Update 3: Changed the default packing density to 200. Removed the active N percent and denitrification rate input variables.

---

Update 4: Removed issue #1401. Will replace with individual issues for each action item.


RUFAS/input_manager.py
```diff
@@ -848,6 +1066,46 @@ def _validate_dict_element(  # noqa
                     element_counter_and_validity["is_valid"] = False
                 return element_counter_and_validity
 
+    def _validate_json_element(  # noqa
```
```suggestion
    def _validate_json_element(
```
flake8 has no issues with this method, why is there noqa here?

ltp23: Removed now.

RUFAS/input_manager.py
```diff
@@ -284,48 +279,36 @@ def _populate_pool(self, eager_termination: bool) -> bool:
...
                         properties_blob_key,
                         filtered_input_data,
                         eager_termination,
-                        element_counter_and_validity,
+                        self.elements_counter,
+                    )
+                else:
+                    raise ValueError(
+                        f"Unsupported file type {file_type}, supported types are: {data_type_to_loader_map.keys()}"
                     )
```
```suggestion
```
This check is redundant. An invalid file type will be caught and handled when Input Manager attempts to load the file.

RUFAS/input_manager.py
```diff
@@ -563,338 +546,364 @@ def _add_default_values_to_array_inputs(
...
+    def _validate_input_by_type(
         self,
         variable_properties: Dict[str, Any],
-        var_name: str,
-        input_data_value: Any,
+        variable_path: List[str | int],
+        input_data: Dict[str, Any],
+        eager_termination: bool,
         properties_blob_key: str,
+        elements_counter: "ElementsCounter",
```
Isgotkowitz, line 741:
```suggestion
        elements_counter: ElementsCounter,
```
I think it would be better to move `ElementsCounter` to the top of the file, then type hit it normally.

RUFAS/input_manager.py
```diff
@@ -1030,6 +1039,106 @@ def _bool_type_validator(
...
+        """
+
+        for key in variable_path:
+            if isinstance(input_data, list) and 0 <= int(key) < len(input_data):
+                input_data = input_data[int(key)]
+            elif isinstance(input_data, dict) and isinstance(key, str) and key in input_data:
+                input_data = input_data[key]
+            else:
+                raise KeyError(f"There is an error at key {key} in the path {variable_path}")
+        return input_data
```
```suggestion
```
This method is defined twice.

RUFAS/input_manager.py
```diff
@@ -1741,3 +1812,161 @@ def dump_get_data_logs(self, path: Path) -> None:
...
+        ElementsCounter
+            A new ElementsCounter object with the counts of the two objects added together.
+        """
+
+        new_counter = ElementsCounter()
+        new_counter.valid_elements = self.valid_elements + other.valid_elements
+        new_counter.invalid_elements = self.invalid_elements + other.invalid_elements
+        new_counter.fixed_elements = self.fixed_elements + other.fixed_elements
+        new_counter._check_negative_counts()
+        return new_counter
```
Isgotkowitz: I think `ElementState` and `ElementCounter` are adding more than there need to be. Here are some changes that I think would make things simpler:
- Remove `decrement()` in `ElementCounter`. It isn't used at all, so it seems like it isn't needed. Also `_check_negative_counts()` could be removed.
- Remove `ElementState` and `increment()`. Instead have there be three methods in `ElementCounter`, one for adding a valid element, one for adding a fixed element, and one for adding an invalid element.

Alternatively, `ElementCounter` could be removed entirely and replaced with a few methods in Input Manager used to increment the counts of valid, fixed, and invalid elements.

RUFAS/input_manager.py
```diff
@@ -563,338 +546,364 @@ def _add_default_values_to_array_inputs(
...
                 )
-                om.add_warning(warning_name, warning_message, info_map)
                 return False
         return True
 
+    def _array_type_validator(
+        self,
+        variable_path: List[str | int],
+        variable_properties: Dict[str, Any],
+        input_data: Dict[str, Any],
```
Isgotkowitz, line 946: If the input data is a dict, it should be validated with the method dedicated to validating dict inputs.

RUFAS/input_manager.py
```diff
@@ -563,338 +546,364 @@ def _add_default_values_to_array_inputs(
...
+        """
+        Validates the container properties of an array input data element.
+
+        Parameters
+        ----------
+        variable_path : List[str | int]
+            The path to the variable being validated.
+        variable_properties : Dict[str, Any]
+            The metadata properties for the variable being validated.
+        input_data : Any
```
Isgotkowitz, line 889: The type should not be Any, it should be a list.

RUFAS/input_manager.py
```diff
@@ -563,338 +546,364 @@ def _add_default_values_to_array_inputs(
...
+        self,
+        variable_path: List[str | int],
+        variable_properties: Dict[str, Any],
+        input_data: Dict[str, Any],
+        eager_termination: bool,
+        properties_blob_key: str,
+        elements_counter: "ElementsCounter",
+    ) -> bool:
+        """
+        Validates an input data element of type object.
```
Isgotkowitz: This method is extremely similar to `_validate_dict_element()`. Please combine it with that method.

Isgotkowitz (changes requested):
Requested changes in the Input Manager. I think there are some things that could be done to make it simpler, though some of the suggestions may not apply after the changes from main are merged in.

RUFAS/input_manager.py
ltp23: It's been removed now.

RUFAS/input_manager.py
ltp23: Yes, this would help with typing inside the InputManager class, but within the ElementsCounter class, I would still need to use quotation marks around the type (like the add method).

RUFAS/input_manager.py
ltp23: It's been removed now. I transferred this code to an earlier PR, so I thought it wouldn't duplicate but it did, probably because of the differences in location.

RUFAS/input_manager.py
ltp23: I can remove `decrement()` and `_check_negative_counts()`. I would hold off on removing the class because I like some version of separation of concerns here and the decoupling helps with getting the counts from independent elements more easily, like when adding additional variables to the pool.

RUFAS/input_manager.py
ltp23: The type still needs to be checked inside the method to make sure it is a list.

RUFAS/input_manager.py
ltp23: I have tried the following and it works: remove the `dict_type_validator()` method, and use `_validate_input_by_type()` method directly in `_populate_pool()` and `_add_variable_to_pool()`. 

I think this approach makes sense because each element will be funneled through the `_validate_input_by_type()` method first and go to the appropriate type validator.

Changing the code is fairly simple, but fixing the broken tests and/or writing new ones in their place takes some non-trivial amount of work and I can't do that right now. So, I will leave that as one of the follow-up housekeeping tasks.

RUFAS/input_manager.py
ltp23: I would revisit this point when combining the two `_dict_type_validator()` and `_object_type_validator()`.

RUFAS/input_manager.py
Isgotkowitz: Ok, that sounds like a plan.

RUFAS/input_manager.py
```diff
@@ -756,259 +731,337 @@ def _add_default_values_to_array_inputs(
...
         Receives data loaded from json input file, recursively finds and then validates nested elements,
         attempts to fix any invalid elements, and tracks the number of valid, invalid, fixed,
         and total elements from the input file are checked.
 
         Parameters
         ----------
-        element_hierarchy : List[str]
-            A list of strings representing the path to the data being validated.
+        first_level_key : str
+            The name of the json data element just under the level of properties_blob_key.
```
```suggestion
            The name of the data element just under the level of properties_blob_key.
```
Two things here:
1. It would be incorrect to say that this is came from a JSON, it isn't known in the context of this function where it came from.
2. Could please reword this description or give some examples? I am not sure what you mean by this.

RUFAS/input_manager.py
```diff
@@ -756,259 +731,337 @@ def _add_default_values_to_array_inputs(
...
         if minimum_length is not None:
-            is_in_range = variable_properties["minimum_length"] <= len(input_data_value)
+            is_in_range = variable_properties["minimum_length"] <= len(input_data)
             if not is_in_range:
-                warning_name = "Validation: array length less than minimum"
-                warning_message = (
-                    f"Variable: '{var_name}' has length: {len(input_data_value)}, less than minimum "
-                    f"length: {minimum_length}. {properties_violation_message}"
+                om.add_warning(
+                    "Validation: array container length less than minimum",
```
```suggestion
                    "Validation: array length less than minimum",
```
"array container" makes it sounds like it is something that contains arrays.

RUFAS/input_manager.py
```diff
@@ -756,259 +731,337 @@ def _add_default_values_to_array_inputs(
...
         if maximum_length is not None:
-            is_in_range = len(input_data_value) <= variable_properties["maximum_length"]
+            is_in_range = len(input_data) <= variable_properties["maximum_length"]
             if not is_in_range:
-                warning_name = "Validation: array length greater than maximum"
-                warning_message = (
-                    f"Variable: '{var_name}' has length: {len(input_data_value)}, greater than "
-                    f"maximum length: {maximum_length}. {properties_violation_message}"
+                om.add_warning(
+                    "Validation: array container length greater than maximum",
```
```suggestion
                    "Validation: array length greater than maximum",
```

RUFAS/input_manager.py
```diff
@@ -1943,3 +1989,130 @@ def dump_get_data_logs(self, path: Path) -> None:
...
+        The element is valid.
+    INVALID : int
+        The element is invalid and cannot be fixed.
+    FIXED : int
+        The element is invalid initially but has been fixed.
+    """
+
+    VALID = 1
+    INVALID = 2
+    FIXED = 3
```
```suggestion
    VALID = "valid"
    INVALID = "invalid"
    FIXED = "fixed"
```
Definitely more redundant, but is clearer.

input/metadata/properties/default.json
```diff
@@ -3145,10 +3149,10 @@
           "pattern": "^(preservative|nutrient additive|absorbant)$",
           "default": "preservative"
         },
-        "package_density": {
+        "packing_density": {
           "type": "number",
-          "description": "Package Density -- The density of ensiled feed material, kg DM /m3",
-          "minimum": 200,
+          "description": "Packing Density -- The density of ensiled feed material, kg DM /m3",
+          "minimum": 14,
```
Isgotkowitz: Where is this number coming from?

input/metadata/properties/default.json
```diff
@@ -4745,6 +4810,20 @@
...
+          "minimum": 0.0,
+          "maximum": 100.0,
+          "default": 0.0
+        },
+        "denitrification_rate": {
+          "type": "number",
+          "description": "",
+          "minimum": 0.0,
+          "maximum": 1.0,
+          "default": 0.0
```
```suggestion
        },
```
These should be removed from inputs, instead of being allowed in.

input/metadata/properties/default.json
```diff
@@ -826,7 +826,7 @@
         "pen_type": {
           "type": "string",
           "description": "Pen Type -- The type of pen (freestall, tiestall, open lot, compost bedded pack barn)",
-          "pattern": "^{freestall|tiestall|open lot|compost bedded pack barn}$",
+          "pattern": "^freestall|tiestall|open lot|compost bedded pack barn$",
```
```suggestion
          "pattern": "^(freestall|tiestall|open lot|compost bedded pack barn)$",
```

Isgotkowitz (approved):

input/metadata/properties/default.json
JoeWaddell: Looks like from the conflict between the default and the former minimum, but confirmation that one or the other is correct from a SME would be best. I couldn't find it in the Feed design doc, and checking some sources online, 14 kg / m^3 seems low.

RUFAS/input_manager.py
JoeWaddell: Strings make more intuitive sense to me as well, but for my own edification, is there any benefit in using the existing ints?

JoeWaddell (commented):
Assuming @Isgotkowitz 's remaining comments are addressed prior to merge, this looks good. 

Given the size of this PR I agree with pushing some of the cleanup to Issues #1335 and #1401

ltp23:
> Looks like from the conflict between the default and the former minimum, but confirmation that one or the other is correct from a SME would be best. I couldn't find it in the Feed design doc, and checking some sources online, 14 kg / m^3 seems low.

There seems to be a mismatch between the metadata and the input data also. I saw that all the feed json files use 14 so I thought that I would update the metadata. But of course, this can be changed with further confirmation.

---

Update: I changed the default to 200.

input/metadata/properties/default.json
ltp23: Ok I will remove these in the metadata and input json files.

JoeWaddell (approved):
LGTM

PR #1355, opened 2024-03-01
Improve IM data retrieval in the life cycle module

### LINKED ISSUE #1320: Analyze `IM.get_data()` inquires
Currently, it appears to have over 300,000 calls to the `InputManager.get_data()` function for data inquires. Please analyze the situation and possibly save the data at local class/method to reduce repeated function calls to speed up the simulation.

ltp23:
In this PR, instead of calling IM's `get_data()` repeatedly from the Cow and LifeCycleManager classes, I replaced those calls with direct calls to the AnimalBase.config dictionary. 

I also added more unit tests to improve the test coverage for the Cow class. 

Closes #1320


allisterakun (approved):
LGTM! Good job, Loi

RUFAS/routines/animal/types/preg_check_config.py
```diff
@@ -0,0 +1,22 @@
...
+    List of expected keys for preg check configuration dictionary,
+    used in daily pregnancy check routine for HeiferII and Cow classes.
+
+    Each dictionary contains:
+        - 'day': The number of days from the AI day when the pregnancy check should occur.
+        - 'loss_rate': The probability of pregnancy loss at this check.
+        - 'on_preg_loss': The event message to log if pregnancy loss occurs at this check.
+        - 'on_preg': The event message to log if the cow remains pregnant at this check.
+        - 'on_not_preg' (optional): The event message to log if the cow is not pregnant, due to a failed conception,
+         at the start of the first check.
```
Isgotkowitz: I think the explanation of what the dictionary contains should be rewritten to be an "Attributes" section following the NumPy docstring guideline.

RUFAS/routines/animal/types/preg_check_config.py
```diff
@@ -0,0 +1,22 @@
+from typing import TypedDict
+
+
+class PregCheckConfig(TypedDict, total=False):
+    """
+    List of expected keys for preg check configuration dictionary,
+    used in daily pregnancy check routine for HeiferII and Cow classes.
+
+    Each dictionary contains:
+        - 'day': The number of days from the AI day when the pregnancy check should occur.
```
Isgotkowitz:
```suggestion
        - 'day': The number of days from the day of artificial insemination when the pregnancy check should occur.
```
I'm pretty sure this is what AI means in this context, I think it would be better to make it a little more verbose.

Isgotkowitz (approved):
Overall LG8TM Loi, nice job. I left two comments related to the class docstring on `PregCheckConfig`.

Also, I think it would be better to move `preg_check_config.py` into `life_cycle/`, instead of creating a new directory for it. Pregnancy is so closely related to repro that it makes sense to me to group it with the other repro files, and a directory named `types` doesn't really indicate at all what is in it or what it is relevant to.

ltp23:
> Also, I think it would be better to move `preg_check_config.py` into `life_cycle/`, instead of creating a new directory for it. Pregnancy is so closely related to repro that it makes sense to me to group it with the other repro files, and a directory named `types` doesn't really indicate at all what is in it or what it is relevant to.

Thanks Ed. I created a separate `types/` folder because I anticipated that to resolve lots of the type errors in the animal module, we would need to create many more type definition files.

ltp23:
Thanks everyone for reviewing this PR.

Issue #1206, opened 2024-02-02
Remove stocking_density from animal module


JoeWaddell:
The attribute `stocking_density` appears to duplicate the intended purpose of the method `current_stocking_density`. With careful revision, use of the `current_stocking_density` method should replace instances of `stocking_density`. Issue originally noted by @tomhuhh

PR #1358, opened 2024-03-03
Removed stocking Density 

0luchi:
This pull request addresses issue [#1206 ](https://github.com/RuminantFarmSystems/MASM/issues/1206)
It removes `pen.stocking_density` from the animal module and replaces every instance with `current_stocking_density`. It also updates the tests accordingly


RUFAS/routines/animal/animal_manager.py
```diff
@@ -1600,7 +1597,7 @@ def _add_animal_to_pen_and_id_map(
         animal_combination = self.ANIMAL_GROUPING_SCENARIO.find_animal_combination(animal)
         pen_with_min_stocking_density = min(
             self.pens_by_animal_combination[animal_combination],
-            key=lambda p: p.current_stocking_density,
+            key=lambda p: p.current_stocking_density(),
         )
```
JoeWaddell: The sim won't run with these parentheses, so locally I removed them and then it runs successfully. I suppose it's because `current_stocking_density` uses the property decorator, but I'm sure @ltp23 can clarify if that's the case.

JoeWaddell (commented):
I tried fixing the tests when removing the parentheses in `_add_animal_to_pen_and_id_map` but can't seem to get it right: likewise for including that commented out animal again. I already tagged Loi in the comment about the change to the logic, so hopefully he can give some insight into how to modify the tests.

RUFAS/routines/animal/animal_manager.py
ltp23: I agree. Properties don't need parentheses.

ltp23 (approved):

ltp23:
I added some changes to the add-animal test. I also changed a pen id in the pens fixture because a heiferII should go to a growing pen.

JoeWaddell (approved):
Great job @0luchi, and thanks @ltp23 for helping with advice and the test fix!

PR #1359, opened 2024-03-04
Emptying storage manure for anaerobic digestion and lagoon

### LINKED ISSUE #1343: Anaerobic digestion and lagoon manure storage update
This issue has similar behaviors compared to the issue #1341. However, the root cause could be a totally different direction. The `anaerobic lagoon` function works fine, which means it is able to empty the manure storage at the end of each storage time period on its own. However, when coupled with `anaerobic digestion`, the `_adjust_accumulated_output ` function in the `anaerobic_lagoon.py` file does not see to work any more.  

## Expected Behavior

The anaerobic lagoon gets emptied at the end of each manure storage time period when the `anaerobic digestion and lagoon` method is chosen as the manure treatment.    

## Current Behavior

The anaerobic lagoon does not get emptied at the end of each manure storage time period when the `anaerobic digestion and lagoon` method is chosen as the manure treatment.    

## Possible Solution

This issue requires more investigation and may end up with some code structure changes. 

## Steps to Reproduce

1. Select the  `anaerobic digestion and lagoon` method as the manure treatment and repeat the steps in issue #1341.  
2.
3.
4.

allisterakun:
This PR is a temporary fix to the issue "[Anaerobic digestion and lagoon manure storage update #1343](https://github.com/RuminantFarmSystems/MASM/issues/1343)".

### Context

Issue(s) closed by this pull request: #1343 

- [x] The [changelog](https://docs.google.com/spreadsheets/d/1U-3igxdstHTFb6k5dVcE5-x9IL0r6q18eUiwwytKZyA/edit#gid=0) has been updated

### What

AnaerobicLagoon._adjust_accumulated_output() is copied and pasted to be added to AnaerobicDigestionAndLagoon to empty the storage manure for each storage time period.

### Why

Temporarily fix the issue of manure storage not emptying.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 [![Testing](https://github.com/RuminantFarmSystems/MASM/actions/workflows/testing_pytest.yml/badge.svg)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/testing_pytest.yml)
 [![Linting](https://github.com/RuminantFarmSystems/MASM/actions/workflows/lint_flake8.yml/badge.svg)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/lint_flake8.yml)
 [![Coverage](https://img.shields.io/badge/coverage-95%25-brightgreen)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/coverage_pytest-cov.yml)
-[![Mypy](https://img.shields.io/badge/mypy-4686%20errors-red)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/mypy_type_checker.yml)
+[![Mypy](https://img.shields.io/badge/mypy-4687%20errors-red)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/mypy_type_checker.yml)
 
 # Vision
 To support research and sustainable decision-making in ruminant animal production through a state-of-the-art, open-source modeling environment that is continuously adapting as technology and scientific knowledge advance.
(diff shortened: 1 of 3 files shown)


tomhuhh (changes requested):
Thanks Allister! I like the idea of emptying the manure storage at the high level. 

However, the emptying function does not seems to work as expected. When I set the `storage_time_period` to 30 days and use scenario 5 in the default manure management input file as the manure management practice, here is the output I got. It seems that the manure is emptied every 120 days instead, which is the length of the `storage_time_period` for slurry storage. 

ADL

At this point, further troubleshoot is required.   

Thanks!

allisterakun:
Hi @tomhuhh,

After some troubleshooting, I have fixed the issue. `AnaerobicDigestionAndLagoon` will now use the `storage_time_period` specified for the **_"anaerobic lagoon"_**. 

Here is what I found out:
1. The `storage_time_period` property of **_"anaerobic digestion"_** (along with other manure treatment types that don't have a specified `storage_time_period` property in `manure_management.json`) will be set to a default value of 120 by **InputManager**.
2. I made the mistake of using the value of `storage_time_period` for **_"anaerobic digestion"_** (which is not specified in the input JSON file and set to the default value of 120 by IM) instead of **_"anaerobic digestion"_**.
3. I have fixed the above mistake: `AnaerobicDigestionAndLagoon` will now use the `storage_time_period` specified for the **_"anaerobic lagoon"_**.

elle-andreen (approved):
Thanks Allister! Both anaerobic digestion and lagoon and anaerobic lagoon seem to be emptying correctly now. They both take the storage time inputted for anaerobic lagoon in manure_management.json, as they should.

tomhuhh (approved):
Thanks Allister! This looks good to me. Thanks for addressing the issue so quickly! 

ADL_after

PR #1360, opened 2024-03-04
Add new Feed Emissions Library

KFosterReed:
The new library named `full_feeds_emissions_new.csv` contains revised estimates for the purchased feed emissions.
### Context

Issue(s) closed by this pull request: #

The [changelog](https://docs.google.com/spreadsheets/d/1U-3igxdstHTFb6k5dVcE5-x9IL0r6q18eUiwwytKZyA/edit#gid=0) has been updated: [ ]
Will add a change log when merged to main
### What

Updating inputs for feed emissions factors

### Why

Old values were incorrect

### How

added a new csv file to the EEE input directory

### Test plan

Replace the path for the `"purchased_feed_emissions"` metadata input to `"input/data/EEE/full_feeds_emissions_new.csv"`


jadamchick (approved):
Looks good! I ran it for c2f8. total feed emissions dropped 27x, and whey's contribution dropped from 89% to 10% of the feed emissions footprint. 

A couple questions- 
- this obviously has updates to the whey footprint, does it also update the BP blends containing whey? 
- will we eventually want to add this to Main as well? 
- can we delete the previous CSV and give this one the same name, so we don't need to update all the metadata paths?

KFosterReed:
Great news! 
Answers to the questions: 
- Yes! It also updates the blends and actually most of the regional EF provided by LEIF as well
- Yes, I wanted to make sure you had it first but I will also add it to main. 
- I think that would be OK but I think I would prefer to update the metadata with the new name because it will be easier to know (especially if it is on main), that we are using the correct file. There should be a way to find and replace all instances of 'full_feeds_emissions.csv' with 'full_feeds_emissions_new.csv' with a few clicks. I can make that change and push it to the pilot testing branch now if it is helpful?
-

jadamchick:
Sounds great! and sure, if it's not a hassle that would be helpful. thanks!

PR #1367, opened 2024-03-05
previous ration tests

JoeWaddell:
Added a few tests to `test_animal.py` related to changes to `ration_driver.py` made in #1305 

Minor change to `ration_optimizer.py` logic to avoid redundant initial assignment of `x0`.


tests/animal_module_tests/test_animal.py
```diff
@@ -2753,10 +2788,117 @@ def test_calculate_rqmts():
     pass
 
 
-# @pytest.mark.parameterize("udrm, om, expected", [MagicMock(), MagicMock(), True])
-def test_formulate_ration() -> None:
+def test_formulate_ration_is_udr_true(mocker: MockerFixture) -> None:
     """Unit test for function formulate_ration in file routines/animal/ration/ration_driver.py"""
     pass
+    # mock_animal_requirements
+    # req = RUFAS.routines.animal.ration.animal_requirements.AnimalRequirements()
```
ltp23: Removing it, maybe?

tests/animal_module_tests/test_animal.py
```diff
@@ -2753,10 +2788,117 @@ def test_calculate_rqmts():
...
+        pen=mocker.MagicMock(), available_feeds=mocker.MagicMock(), animal_grouping_scenario=mocker.MagicMock()
+    )
+    # Assert
+    assert actual == udrm_is_udr_expected
+
+
+def test_formulate_ration_hasattr(mocker: MockerFixture) -> None:
+    """Unit test for function formulate_ration in file routines/animal/ration/ration_driver.py"""
+    req = RUFAS.routines.animal.ration.animal_requirements.AnimalRequirements()
+    mocker.patch("RUFAS.routines.animal.ration.animal_requirements.AnimalRequirements.__new__", return_value=req)
```
ltp23: Should this be __init__ instead of __new__? Though I think they basically equivalent because init is called after new if I remember correctly.

ltp23 (approved):

tests/animal_module_tests/test_animal.py
JoeWaddell: When I tried patching in `__init__` instead, it gives me this error:
`FAILED tests/animal_module_tests/test_animal.py::test_formulate_ration_hasattr - TypeError: __init__() should return None, not 'AnimalRequirements'`

ltp23: Thanks Joe. That makes sense.

Confusingly, after experimenting a little bit more, all of the following three options would work:

```
# Patch where the class is used. This is more broad and replaces all the functionalities, including class methods, with whatever functionalities the mock object has.
mocker.patch("RUFAS.routines.animal.ration.ration_driver.animal_requirements.AnimalRequirements", return_value=req)
    
# Patch where the class is used. This option is more targeted and patches the instantiation process specifically, so class methods are spared.  
mocker.patch(
"RUFAS.routines.animal.ration.ration_driver.animal_requirements.AnimalRequirements.__new__", return_value=req
    )

# Like what you have. This option patches where the class is defined and has a more wide-ranging effect depending on how it is used in other places.
mocker.patch("RUFAS.routines.animal.ration.animal_requirements.AnimalRequirements.__new__", return_value=req)
```

In the confines of a unit test, they all seem to behave similarly and the effects of patching disappear after the test anyway. But I think I'm leaning more toward the second option.

Isgotkowitz (approved):
LGTM, nice job Joe

tests/animal_module_tests/test_animal.py
JoeWaddell: Thanks for the input and advice Loi! I'll go with your instinct here, I like your logic.

PR #1379, opened 2024-03-06
Modify IM get data 

### LINKED ISSUE #1334: [IM Change] Modify `get_data()` in IM to not throw KeyError
## Current Behavior
When the `get_data()` method is called in the Input Manager, it will throw a KeyError if it cannot find any data associated with the address that is passed to it.

## Desired Behavior
Create a new function `property_exists_in_pool()` (probably a better name), that accepts a data_address (string in the format of "a.b.c"), parses the data address and checks if the variable exists in  InputManager.__pool. This function will be a public method so that other modules can directly call this method to find out if a variable exists in the pool,

`get_data()` would call the new method to find out if a variable exists in InputManager.__pool first. if it doesn't exist, get_data() should return None and log an error to the Output Manager.

## Notes
Changing `get_data()` will probably be really simple. ~~But making sure this change is properly accounted for in all methods that call `get_data()` will probably be a ton of work.~~ I don't believe we need to change anything here, because all the current calls to `get_data()` already assume the variable exists in the pool.

ltp23:
In this PR I added a public method called `property_exists_in_pool()` in the InputManager class to check the existence of a data address in the input data pool. I also added a helper method `_extract_value_by_key_list()` that does the actual walk-through of the pool dictionary.

The original plan, as laid out in the description of issue #1334, was to modify the IM's `get_data()` method so that it calls the `property_exists_in_pool()` first and then returns the data. This seems redundant because we would have to go through the data twice, so I made both methods use the extra helper.

Lastly, the IM's `get_data()` now returns `None` when the data address is invalid instead of raising an error.

### Context

Issue(s) closed by this pull request: #1334 

The [changelog](https://docs.google.com/spreadsheets/d/1U-3igxdstHTFb6k5dVcE5-x9IL0r6q18eUiwwytKZyA/edit#gid=0) has been updated: [ ]


RUFAS/input_manager.py
```diff
@@ -1138,36 +1133,131 @@ def get_data(self, data_address: str) -> Any:
...
-                f'Cannot find "{data_address}", ' f'"{parent_address}" does not have attribute ' f'"{invalid_key}".',
-                info_map,
-            )
+        return None
 
-            raise KeyError(
-                f'Data not found: Cannot find "{data_address}", '
-                f'"{parent_address}" does not have attribute "{invalid_key}".'
-            )
+    def property_exists_in_pool(self, data_address: str) -> bool:
```
```suggestion
    def check_property_exists_in_pool(self, data_address: str) -> bool:
```

RUFAS/input_manager.py
```diff
@@ -1138,36 +1133,131 @@ def get_data(self, data_address: str) -> Any:
...
+
+        variable_path = data_address.split(".")
+        try:
+            self._extract_value_by_key_list(self.__pool, variable_path)
+            return True
+        except KeyError:
+            return False
+
+    def _extract_value_by_key_list(
+        self, input_data: List[Any] | Dict[str, Any], variable_path: Sequence[str | int]
```
ew3361zh, line 1196: Will the variable path ever have ints in it? i.e. do we need to accommodate for that in the Sequence type hint? Is this trying to accommodate for looking for specific list indices and does it accomplish that?

RUFAS/input_manager.py
```diff
@@ -1138,36 +1133,131 @@ def get_data(self, data_address: str) -> Any:
...
+        ...     ]
+        ... }
+        >>> var_path = ["manure_management_scenarios", 0, "bedding_type"]
+        >>> input_manager._extract_value_by_key_list(example_data, var_path)
+        'straw'
+        """
+
+        for key in variable_path:
+            if isinstance(input_data, list) and 0 <= int(key) < len(input_data):
+                input_data = input_data[int(key)]
```
ew3361zh, line 1255: Maybe a different name for the data found/being looked for?

ew3361zh (approved):
Looks good - I tested it with a bad data address and a good data address - the examples in the docstrings were very helpful for that, thank you, Loi!

A few comments/questions about naming of functions and variables and what the case would be for receiving an int in the data address but otherwise looks good to me!

Also, the changelog does still need to be updated.

allisterakun (approved):
Changes look good! Good job, Loi.

Just a minor test coverage issue (not from the changes in this PR) if you have some time to implement to bring IM coverage to 100%
image

RUFAS/input_manager.py
ltp23: The use of Sequence here is to fix some mypy errors. Functionally, it is equivalent to List.

It is possible that the variable path has a mix of strings and integers (normal integers or string integers). The use of integers is for array indexing, which is needed in the array validation PR.

RUFAS/input_manager.py
ltp23: It looks like that but it is actually necessary to use the same name for reassignment as it goes down the nested elements.

ltp23:
Thanks everyone for reviewing the PR.

PR #1384, opened 2024-03-06
[IM] Standardize metadata properties structure to simplify IM validation routines

### LINKED ISSUE #1346: [IM][Refactor] Make metadata properties uniform to simplify input validation.
### What
Currently, the IM loads _and validates_ data differently based on the format of the data source. This can be observed in the following code:
```python
data_type_to_loader_map: Dict[str, Callable] = {
    "json": self._load_data_from_json,
    "csv": self._load_data_from_csv,
}

    file_path = file_details["path"]

    try:
        file_type = file_details["type"]
        data_loader = data_type_to_loader_map[file_details["type"]]
        input_data = data_loader(file_path)
    except KeyError:
        raise KeyError(
            f"Faulty data type in {file_blob_key}," f"supported types are: {data_type_to_loader_map.keys()}"
        )

    -- snip --

    for metadata_property in metadata_properties.keys():
        -- snip -- 
        if file_type == "json":
            element_counter_and_validity = self._validate_dict_element(
                [metadata_property],
                properties_blob_key,
                filtered_input_data,
                eager_termination,
                element_counter_and_validity,
            )
        if file_type == "csv":
            element_counter_and_validity = self._validate_tabular_element(
                metadata_property,
                properties_blob_key,
                filtered_input_data,
                eager_termination,
                element_counter_and_validity,
            )
```

The IM should be changed so that it only has to differentiate between different input streams when calling `data_loader()`. After that point, when it is validating data, it should not have to take different paths of execution for different types of input streams.

### Why
By eliminating the separate validation paths, the code for the IM is made simpler and less complex. In this way, this issue closely aligns with #1335. Implementing this change will also have the side effect of making the metadata more flexible - more on this below.

### How
To have one set of methods that can validate input data from any input source, the structure of properties for different inputs must be the same. Once these properties have been rewritten, the `if` statements checking the data stream type can be removed.

Here is an example of how the economic properties could be rewritten:
```json
"emissions_properties": {
  "alfalfa_tillage_emissions": {
    "type": "number",
    "description": "The amount of total tonnes CO2-eq generated from alfalfa tillage",
    "minimum": 0,
    "default": 500
  },
  "corn_grain_tillage_emissions": {
    "type": "number",
    "description": "The amount of total tonnes CO2-eq generated from corn grain tillage",
    "minimum": 0,
    "default": 300
  },
  "corn_silage_tillage_emissions": {
    "type": "number",
    "description": "The amount of total tonnes CO2-eq generated from corn silage tillage",
    "minimum": 0,
    "default": 400
  }
}
```
changes to this:
```json
"emissions_properties": {
  "alfalfa_tillage_emissions": {
    "type": "array",
    "description": "The amount of total tonnes CO2-eq generated from alfalfa tillage",
    "properties": {
      "minimum": 0,
      "default": 500
    }
  },
  "corn_grain_tillage_emissions": {
    "type": "array",
    "description": "The amount of total tonnes CO2-eq generated from corn grain tillage",
    "properties": {  
      "type": "number",
      "minimum": 0,
      "default": 300
    }
  },
  "corn_silage_tillage_emissions": {
    "type": "array",
    "description": "The amount of total tonnes CO2-eq generated from corn silage tillage",
    "properties": {
      "type": "number",
      "minimum": 0,
      "default": 400
    }
}
```

A side effect of rewriting the metadata and changing the IM as described above would be that, in some cases, the same input data could come from different streams and would be validated against the same properties.

### Other
See [this comment](https://github.com/RuminantFarmSystems/MASM/issues/1249#issuecomment-1936632693) for more conversation on this issue.

Isgotkowitz:
This PR standardizes the metadata properties and removes some redundant functions in the Input Manager.

### Context

Issue(s) closed by this pull request: closes #1346 

Note that most of the changes in this PR are changes to the metadata properties. All other changes add up to 354 lines changed. I recommend understanding the changes being made in the default properties, but not reading through every single line.

### What

This PR simplifies input validation in the Input Manager by starting validation at a single point instead of routing input data into different methods. Additionally, the code to add runtime variables to the Input Manager has been simplified because it followed a very similar pattern to the initial validation routine.

In order to achieve this, the metadata properties for inputs that are usually passed in CSV format have been rewritten so they follow a consistent format with all the other metadata properties. This allows the Input Manager to validate input data in the same manner and with the same properties regardless of whether it originated from a CSV or JSON file.

### Why

To eliminate duplicative code. By simplifying the Input Manager in this way there is now only one way to validate a dictionary of arrays, instead of two.

Anyone who wants to write a new properties section in the metadata only has to learn one way of writing those properties, and don't have to figure which format they need to use based on the expected input data format.

### How

- The metadata properties have been rewritten, so that all properties follow the same format for the same types of data.
- `_validate_tabular_element()` has been deleted, and `_populate_pool()` and `_add_variable_to_pool()` now only call `_validate_dict_element()` to validate data.

### Test plan

All tests have been updated or deleted as appropriate. Coverage in the Input Manager remains high (99%), and with the only two missing lines being outside the scope of this PR.


ew3361zh (approved):
I think this looks good, Ed and the changes make sense - thanks for investigating and coming up with a good solution. 

My only concern is coordination with the IM Array element validation in #1345 which I haven't yet reviewed in depth (so maybe it's not an issue). I'm not sure the best way to ensure these efforts are being coordinated but I think it's worth having that conversation with @ltp23 if it hasn't happened already.

allisterakun (approved):
Changes look great, good job Ed! 
My only comment would be maybe renaming the `_validate_dict_element` function?

Isgotkowitz:
> I think this looks good, Ed and the changes make sense - thanks for investigating and coming up with a good solution.
> 
> My only concern is coordination with the IM Array element validation in #1345 which I haven't yet reviewed in depth (so maybe it's not an issue). I'm not sure the best way to ensure these efforts are being coordinated but I think it's worth having that conversation with @ltp23 if it hasn't happened already.

Thanks @ew3361zh. I will bring this up at today's worktogether session and make sure we're all on the same page.

Isgotkowitz:
> Changes look great, good job Ed! My only comment would be maybe renaming the `_validate_dict_element` function?

Thanks Allister. That would be an easy change to add in. I think `_dict_type_validator()` would be a good name - it would make it consistent with all the other validator functions.

PR #1385, opened 2024-03-06
Added warning to user-defined ration percentages

### LINKED ISSUE #1349: Add warning if user defined ration percentages do not sum to 100
We need to add a warning in the `UserDefinedRationManager` class to flag when the percentage totals do not sum to 100%. This should be a warning on a pen by pen basis, and the actual total should be reported to the user. 

This should presumably be a warning and not an error, since it is conceivable a user may want to feed over or under the DMI estimates.

However, it is possible that we may want to raise an error if it is beyond some extreme range, e.g. less than 50% predicted DMI, or above 200%.

JoeWaddell:
If the user-defined ration percentages do not total for an animal combination/pen, a warning is produced. Will close #1349

I've modified the default percentages in `default_feed.json` to generate the warning for all four animal combinations. This will be removed prior to merge to main.

In this review: we should consider if an error should be thrown instead of a warning. Alternatively, we could add an error if the difference from 100 is more than X percent.

For SWEs: we should assess if there is a way (or need) to have this warning only show once. Currently it is generated twice, e.g. Feed is initialized twice by default, so this should be reviewed (though is possibly moot given the ongoing Feed refactor/implementation).

I also made some changes in `user_defined_ration.py` to reduce related Mypy errors.


RUFAS/routines/feed/feed.py
```diff
@@ -112,24 +112,52 @@ def __init__(self, data: PurchasedFeedTypedDict):
...
         self.user_defined_ration_percentages = data["user_defined_ration_percentages"]
 
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self.__init__.__name__,
+        }
         udrm.calf_ration = {
             str(dict["feed_type"]): dict["ration_percentage"] for dict in self.user_defined_ration_percentages["calf"]
         }
+        if sum(udrm.calf_ration.values()) != 100.0:
```
ew3361zh: While I'm not particularly familiar with how to best test this (I haven't worked with ration formulation at all), I think it would be good to add some cushion to this to account for float rounding errors. I'm not sure the sensitivity, maybe +/- 0.01?

You could do something like:
target_sum = 100.0
tolerance = 0.01 (or whatever is an acceptable tolerance level)

if not math.isclose(sum(udrm.calf_ration.values(), target_sum, abs_tol=tolerance):
    om.add_warning...

jadamchick (approved):
Looks good to me, thanks Joe! 
I think warnings (vs errors) is a good idea, since my understanding of the original design is that you wanted to enable people to play with funky rations if desired. This serves the purpose of flagging in the case that there was a typo, misunderstanding of the input format, etc. without preventing the simulation if it was intentional.

To Niko's point about level of tolerance -- I think as long as it is just a warning and not an error, then it doesn't hurt to be precise.

RUFAS/routines/feed/feed.py
JoeWaddell: Added! Thanks for the suggestion, this is a great.

ew3361zh (approved):
I think this looks good, nice work, Joe!
As far as the warnings getting called twice I think it means the Feed class is being instantiated twice and from what I could find it was in simulation_engine.py and then again in herd_factor.py in the initialize_herd function. I haven't spent too much time in either of these portions of the code but it does seem like we shouldn't be initializing the Feed class twice (or should have a good reason to be doing it twice in one simulation). I think this is worth writing up a separate issue about to investigate further but I don't think it falls within the purview of this PR.

JoeWaddell:
> it does seem like we shouldn't be initializing the Feed class twice (or should have a good reason to be doing it twice in one simulation). I think this is worth writing up a separate issue about to investigate further but I don't think it falls within the purview of this PR.

My understanding is that since the Feed module is undergoing a thorough refactor (really, a near-total rewrite from scratch), that this will be resolved at that stage. In any case I'll open an Issue and highlight that the double instantiation was noticed here. Thanks!

PR #1389, opened 2024-03-07
Using regexes in cross-references field during report generation

### LINKED ISSUE #1257: Allow cross-references in report generator to use regexes
Currently, the `cross-references` feature in the report generator looks up previously generated reports literally and can't handle the case where the user specifies a regex pattern there.

Enable regex lookup in the report generator when performing cross-referencing.

ltp23:
This PR adds a new feature to the report generator that would allow using regex patterns in the "cross-references" field in the report json specification.

I also moved the call to stdout-write from simulation engine class to the main class, because I didn't see the errors from report generation included in the final tallies.

### Context

Issue(s) closed by this pull request: #1257 

The [changelog](https://docs.google.com/spreadsheets/d/1U-3igxdstHTFb6k5dVcE5-x9IL0r6q18eUiwwytKZyA/edit#gid=0) has been updated:
[ ] 

### Test plan

A simple example I used to test this feature out

```
{
  "multiple": [
    {
      "name": "Total housing methane, kg",
      "filters": [
        "^ManureHandlerDailyOutput_Pen_.*housing_methane$"
      ],
      "slice_start": -365,
      "horizontal_aggregation": "sum"
    },
    {
      "name": "Total storage methane, kg",
      "filters": [
        "^(?:AnaerobicDigestion_)?ManureTreatmentDailyOutput_Pen_.*storage_methane$"
      ],
      "slice_start": -365,
      "horizontal_aggregation": "sum"
    },
    {
      "name": "Total housing and storage methane, kg",
      "cross_references": [
        "Total.*methane.*"
      ],
      "horizontal_aggregation": "sum"
    }
  ]
}
```

One thing of note is the use of regex full-matching, vs. partial matching, to reduce the potential for false positives. The ignored characters should be accounted for by the special wildcard characters. For example, if I want to match the string "Total housing methane, kg_hor_agg", valid regex patterns, in this case, would be `Total.*methane.*` or `.*methane.*` whereas `methane` by itself is not enough.

### DIFF
--- a/RUFAS/report_generator.py
+++ b/RUFAS/report_generator.py
@@ -1,3 +1,4 @@
+import re
 from typing import Dict, List, Any, Callable
 
 from RUFAS.util import Utility
@@ -138,7 +139,7 @@ class ReportGenerator:
     A class to generate reports based on filtered data and aggregation criteria and store them in a dictionary.
     """
 
-    def __init__(self):
+    def __init__(self) -> None:
         """
         Initializes the ReportGenerator.
         """
@@ -197,7 +198,7 @@ class ReportGenerator:
         try:
             if "cross_references" in filter_content.keys():
                 self._check_for_missing_references(filter_content["cross_references"])
-                cross_reference_data = {ref: self.reports[ref] for ref in filter_content["cross_references"]}
+                cross_reference_data = self._get_reports_by_regex(filter_content["cross_references"])
                 cross_reference_data.update(filtered_pool)
                 report_data = self._perform_aggregations(cross_reference_data, filter_content)
             else:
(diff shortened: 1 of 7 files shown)


RUFAS/simulation_engine.py
```diff
@@ -93,9 +93,6 @@ def simulate(self) -> None:
             {"class": self.__class__.__name__, "function": self.simulate.__name__, "units": "days"},
         )
 
-        error_count, warning_count = om.get_error_and_warning_counts()
-        sys.stdout.write(f"{error_count} error(s) and {warning_count} warning(s) found.\n")
-
```
ew3361zh, line 98: Is this supposed to be removed?

main.py
```diff
@@ -440,6 +440,9 @@ def execute_simulations(
         input_manager.dump_get_data_logs(path=output_dir)
         output_manager.dump_all_nondata_pools(output_dir, exclude_info_maps, format_option)
 
+        error_count, warning_count = output_manager.get_error_and_warning_counts()
+        sys.stdout.write(f"{error_count} error(s) and {warning_count} warning(s) found.\n")
+
```
ew3361zh, line 445: I think some old code made its way into this PR. I believe we have this currently in simulation engine.

RUFAS/report_generator.py
```diff
@@ -266,10 +267,39 @@ def _check_for_missing_references(self, references: List[str]) -> None:
...
-        missing_references = [ref for ref in references if ref not in self.reports]
+
+        missing_references = []
+
+        for ref in references:
+            if not any(re.fullmatch(ref, report_name) for report_name in self.reports):
+                missing_references.append(ref)
+
         if missing_references:
             raise KeyError(f"Missing referenced reports: {', '.join(missing_references)}")
```
ew3361zh: Maybe we should update this since the reports referenced won't necessarily be full report names but rather patterns trying to match full report names. I think updating this to something like "Missing referenced reports matching the following pattern(s): ...").

ew3361zh (approved):
This looks good, Loi and was working as intended. I tested it with regex patterns that only partially matched instead of fully matched and it errored as expected.

There's some code that will likely be fixed when main is merged in but I noted it just to make sure.

Only one comment about updating the error message when there are missing references since we're matching patterns now instead of full report names.

Nice work!

RUFAS/simulation_engine.py
ltp23: Yes Niko, without moving these two lines to main, the error counts wouldn't include those generated during report and maybe graph generation.

main.py
ltp23: I moved these from simulation engine.

RUFAS/report_generator.py
ltp23: Good point, Niko. I will update this.

tomhuhh (approved):
I tested with the example filter that Loi provided and it worked as expected. 

Good work, Loi!

ltp23:
Thanks everone!

PR #1392, opened 2024-03-08
Emptying Manure Pits

### LINKED ISSUE #1272: Emptying manure pit
Currently under some manure treatment methods, for example `anaerobic lagoon` ,manure pit is emptied at the end of every storage period, and we're not tracking the removed manure. We should instead apply these manure to the field for the soil and crop module. 

## Expected Behavior

The manure module keeps track of the mass in the manure pit. When soil and crop module calls for manure for field application, the manure gets transferred to the soil and crop module. Then the storage period clock gets reset. For example, if the storage period is 180 days, on day 150, the soil and crop module calls for manure, the manure module checks its storage and assign soil and crop module with manure, then instead of emptying on day 180, it will then schedule to empty on day 330 (150 +180). 

## Current Behavior

Some manure treatments do not empty the manure pit at all, which means manure mass keeps increasing in the pit. 

Some other treatments empty the manure pit, for example, `anaerobic lagoon`, but the removed manure just disappears and is not tracked. 

## Possible Solution

Add a function to move manure from the manure module to the soil and crop module for field application for all manure treatment methods. 

Implement the storage period emptying functionality for all manure treatment methods. 

Synchronize these two methods by checking the manure removal to soil and crop module first, and then adjust the emptying schedule.  

## Steps to Reproduce

1. Run a manure treatment scenario, and make sure your filter includes accumulated volatile solids 
2. Plot volatile solids against simulation day.  
3.
4.

ew3361zh:
PR to consistently empty manure treatment pits on their designated storage time periods.

### Context

Issue(s) closed by this pull request: closes #1272

### What

This PR moves the repeated manure accumulation methods from the separate manure treatment child classes into the BaseManureClass. It also combines those accumulation methods into a single method (`_adjust_accumulated_output()`) which either adds the daily manure to the accumulated manure or empties the manure pit if the storage time period is complete.

[X] It also reports that emptied amount to the OM.

### Why

There was duplicate code in several of the child manure treatment classes that had the same functionality and other child manure treatment classes that didn't empty the manure pits at all on their scheduled storage time period. This makes that emptying more consistent across all manure treatments. The one exception is for AnaerobicDigestion which does not have a storage time period because we are assuming manure is flowing into and out of this treatment method in the same day. 

### How

The duplicate manure adjustment/accumulation code was moved into the BaseManureClass and called within each child class with that treatment class' daily output.

### Test plan

[X] Unit testing updated.

Results were tested running simulations using this graphing filter:
```
{
  "multiple": [
   {
      "type": "plot",
      "filters": ["Accumulated_ManureTreatmentDailyOutput_Pen_0_CALF\\.liquid_manure_total_solids",
      "Accumulated_ManureTreatmentDailyOutput_Pen_1_GROWING\\.liquid_manure_total_solids",
      "Accumulated_ManureTreatmentDailyOutput_Pen_2_CLOSE_UP\\.liquid_manure_total_solids",
      "Accumulated_ManureTreatmentDailyOutput_Pen_3_LAC_COW\\.liquid_manure_total_solids"],
      "title": "Total Solids",
      "legend": ["Pen0", "Pen1", "Pen2", "Pen3"]
    }
  ]
}

```


RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
```diff
@@ -334,12 +337,25 @@ def _get_current_day_rainfall(self) -> float:
         precipitation = current_conditions.precipitation
         return precipitation * GeneralConstants.MM_TO_M
 
-    def _accumulate_daily_output(self, manure_treatment_daily_output: ManureTreatmentDailyOutput) -> None:
-        """Accumulates the daily output of the manure treatment.
+    def _adjust_accumulated_output(self, manure_treatment_daily_output: ManureTreatmentDailyOutput) -> None:
+        """Adjust the accumulated output by either resetting it or adding the daily output to it.
+        The accumulated output will be reset on the first day of every storage time period.
```
```suggestion
        
        Notes
        -----
        The accumulated output will be reset on the first day of every storage time period.
```
Would it make more sense to have it be reset on the final day of every storage time period? I don't think it makes a very big difference, but this would prevent the manure treatment from being reset on the very first day of the simulation. If so, probably an issue for a different PR.

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
```diff
@@ -334,12 +337,25 @@ def _get_current_day_rainfall(self) -> float:
...
 
         """
-        self._accumulated_output += manure_treatment_daily_output
+        if self._sim_day % self.storage_time_period == 1:
+            ManureModuleOutputManagerHelper.add_dataclass_object(
+                self._accumulated_output,
+                info_maps={
+                    "class": self.__class__.__name__,
+                    "function": self._adjust_accumulated_output.__name__,
+                    "prefix": f"{self.__class__.__name__}_pit_emptying_amount_on_sim_day_{self._sim_day}",
```
Isgotkowitz: I think this prefix should be changed so that it doesn't include the sim day. By including the sim day in the prefix, it splits up pit emptyings from the same pit into different variables, making it difficult to see all the different pit emptyings from the same pit at once. Instead, including the pen id in the prefix may be more useful, because the Output Manager will separate the different pit emptyings from different pens into different variables, even if they come from the same type of pit.

Isgotkowitz (approved):
Ran the provided graph filter, it appears that the manure is being emptied routinely and correctly.

Two suggestions:
1. Changing the prefix of the variable that records pit emptyings. There is more on this in a comment below.
2. Using the storage time period that is already in the BaseManureTreatment class, instead of adding an extra attribute in some child classes to track it. Maintaining two different attributes to track the same thing is a bad practice, so it should be avoided. Also, having `self.storage_time_period` pop up in `BaseManureTreatment` is confusing because there is no declaration of it in that class.

Other than that, I think it looks good.

tomhuhh (changes requested):
Thanks Niko! The current code does what is requested in the PR. However, here are some suggestions on the output format below.

I would recommend referring to the output format of ration at each ration formulation interval to report the mass and nutrients of the manure being emptied at the end of each manure storage period. Please see an example below. 

Screenshot 2024-03-11 at 4 01 17 PM

The current output format (Please see below) can make the output list really long when we try to simulate multiple years. 

Screenshot 2024-03-11 at 4 05 19 PM
  

Hope this helps. 
Haowen

ew3361zh:
> I would recommend referring to the output format of ration at each ration formulation interval to report the mass and nutrients of the manure being emptied at the end of each manure storage period. Please see an example below.

Thanks, Haowen. It looks like removing the `simulation_day` from the prefix and adding it as an entry in the info_map. 

Are there other pieces that would be helpful?

If you have a better name than `...pit_emptying_amount`, please let me know. I didn't have something better at the time.

tomhuhh:
Hi Niko! The name of the variable `pit_emptying_amount` sounds ok to me. Maybe @KFosterReed has a better suggestion. 

Regarding the `simulation day ` in the prefix, I believe the change you suugested is in alignment with my thought too.  

Furthermore, for all the manure nutrients and mass such as `liquid_manure_nitrogen` and `liquid_manure_total_solids`, we could also output them as attributes (the way that `dm` & `CP` is reported). 

Thanks!
Haowen

KFosterReed:
Hey Team! So glad to see this coming along! I think 'pit_emptying_amount' is just fine. Although, I think just '_emptying_amount' is just as good and would make it just a little shorter?

ew3361zh:
> 2. Using the storage time period that is already in the BaseManureTreatment class, instead of adding an extra attribute in some child classes to track it. Maintaining two different attributes to track the same thing is a bad practice, so it should be avoided. Also, having `self.storage_time_period` pop up in `BaseManureTreatment` is confusing because there is no declaration of it in that class.

Thanks, Ed. I'll see if I can do something in the base class about this. I know part of the issue is that the base class is trying to accommodate for both the single treatment child classes and the amalgamated anaerobic_digestion_and_lagoon which needs to do both.

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
ew3361zh: It's in an interesting thought. I don't know if it's helpful to have the data or clear reset on the first simulation day? @tomhuhh do you have any opinion on the need for the emptying to occur on the first day of every storage time period or at the end? It's something I'd hesitate to change without knowing the reason why it was done like this in the first place in case there is a good reason the emptying should occur on the first day of the simulation.

ew3361zh:
> Hi Niko! The name of the variable `pit_emptying_amount` sounds ok to me. Maybe @KFosterReed has a better suggestion.

Sounds good, I've gone with `_emptying_amount_for_pen_{pen_number}` based on Ed and Kristan's suggestions.

> Regarding the `simulation day ` in the prefix, I believe the change you suugested is in alignment with my thought too.

Perfect, done!

> Furthermore, for all the manure nutrients and mass such as `liquid_manure_nitrogen` and `liquid_manure_total_solids`, we could also output them as attributes (the way that `dm` & `CP` is reported).

I'm not sure I understand this part. Can you explain a little more?

KFosterReed:
> 
> > Furthermore, for all the manure nutrients and mass such as `liquid_manure_nitrogen` and `liquid_manure_total_solids`, we could also output them as attributes (the way that `dm` & `CP` is reported).
> 
> I'm not sure I understand this part. Can you explain a little more?

So I haven't looked at or run this at all but I'll try to answer the question a bit... when the manure storage unit is emptied, we would like to report all the information that we have about that manure - this would be the total volume, total solids, nitrogen, phosphorus, volatile solids, etc. One thought could be to include everything in the liquid or solid manure protocol... is there a way of achieving this so that it adjusts when a new attribute is added to the protocol? (also, I don't think we have a solid manure protocol yet?)

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
elle-andreen: Presently I think there is no accumulated manure in any manure treatment at the start of the simulation, so there would be nothing to empty if we did empty on the first day of the simulation. Emptying also does not necessarily make the value of accumulated manure = 0, it just utilizes 0 as the value for _yeserday's_ accumulated manure. So the accumulated manure in storage on the emptying day would be 0 + that day's manure entering storage. I do think that the emptying should occur on the first day of the storage time period though, otherwise we would end up with a functional storage time period of n - 1 for the first storage time interval.

ew3361zh:
> > > Furthermore, for all the manure nutrients and mass such as `liquid_manure_nitrogen` and `liquid_manure_total_solids`, we could also output them as attributes (the way that `dm` & `CP` is reported).
> > 
> > 
> > I'm not sure I understand this part. Can you explain a little more?
> 
> So I haven't looked at or run this at all but I'll try to answer the question a bit... when the manure storage unit is emptied, we would like to report all the information that we have about that manure - this would be the total volume, total solids, nitrogen, phosphorus, volatile solids, etc. One thought could be to include everything in the liquid or solid manure protocol... is there a way of achieving this so that it adjusts when a new attribute is added to the protocol? (also, I don't think we have a solid manure protocol yet?)

What's sent to OM is a ManureTreatmentDailyOutput object (how the treatment classes track manure output) which contains a lot. If there are other things it should be tracking, we should probably make an issue to ensure anything added is getting added from the right source and calculated by the correct equation (if that applies).

Here is an example output of the emptying amount (I've since removed the "pit" part of this name):

AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.pen_id
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.simulation_day
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.liquid_manure_total_ammoniacal_nitrogen
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.liquid_manure_nitrogen
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.liquid_manure_total_solids
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.liquid_manure_total_volatile_solids
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.liquid_manure_phosphorus
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.liquid_manure_potassium
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.daily_final_manure_volume
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.liquid_manure_daily_volume
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.storage_methane
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.storage_ammonia
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.storage_nitrous_oxide
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.storage_nitrogen_leached
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.sludge_manure_total_solids
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.sludge_manure_total_volatile_solids
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.sludge_manure_nitrogen
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.sludge_manure_phosphorus
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.sludge_manure_potassium
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.sludge_manure_daily_volume
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.solid_manure_total_solids
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.solid_manure_total_volatile_solids
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.solid_manure_nitrogen
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.solid_manure_inorganic_nitrogen
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.solid_manure_organic_nitrogen
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.solid_manure_total_ammoniacal_nitrogen
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.solid_manure_phosphorus
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.solid_manure_water_extractable_inorganic_phosphorus
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.solid_manure_water_extractable_organic_phosphorus
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.solid_manure_non_water_extractable_inorganic_phosphorus
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.solid_manure_non_water_extractable_organic_phosphorus
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.solid_manure_potassium
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.solid_manure_daily_mass
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.biogas
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.biogas_energy_content
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.methane_generation_volume
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.heating_input_energy
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.evaporated_water
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.minimum_digester_volume
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.top_cover_volume
AnaerobicDigestionAndLagoon_pit_emptying_amount_pen_3.solid_manure_carbon_decomposition

tomhuhh (approved):
Thanks Niko! I think the current output format makes sense to me now, even though it is not necessary to report the gas emissions of the manure being emptied. 

Based on a previous comment by Kristan, items we would like to report about the emptied manure are, for example, total volume, total solids, nitrogen, phosphorus, volatile solids, etc. However, I also understand it can be tedious to hand pick and remove the gas emission ones.    

So I'm gonna approve it. 

Thanks!
Haowen

ew3361zh:
> Thanks Niko! I think the current output format makes sense to me now, even though it is not necessary to report the gas emissions of the manure being emptied.
> 
> Based on a previous comment by Kristan, items we would like to report about the emptied manure are, for example, total volume, total solids, nitrogen, phosphorus, volatile solids, etc. However, I also understand it can be tedious to hand pick and remove the gas emission ones.
> 
> So I'm gonna approve it.
> 
> Thanks! Haowen

Thanks, Haowen! I think the unwanted pieces can be easily filtered out in reporting as well if that's an issue.

PR #1396, opened 2024-03-11
Remove duplicate random seeding

### LINKED ISSUE #1319: Move `random` seed initialization into `execute_simulation()` or `run_rufas()` in main
Currently, the `random` config initialization happens in two places: `SimulationEngine._initialize_simulation()` and `initialize_herd()` in main. They are the same block of code. `HerdFactory` uses the random package for herd initialization, and it is run/initialized before the initialization of `SimulationEngine`.

## Expected Behavior

Only have the `random` config initialization at one single place for the entire simulation.
## Current Behavior

Duplicate code.
## Possible Solution

Move `random` seed initialization into `execute_simulation()` or `run_rufas()` in main

Isgotkowitz:
Removes duplicate seedings of randomization libraries.

### Context

Issue(s) closed by this pull request: closes #1319 

### What

There were previously two seedings of the random libraries used in RuFaS, one in the method `initialize_herd()` in `main.py` and another in the initialization routine of `SimulationEngine`.

### Why

Duplication is bad, and often goes hand in hand with difficult and problematic bugs. We don't want those bugs in the codebase, especially not when it effects something as crucial as the reproducibility of experiments.

### How

A new method (`set_random_seed()`) has been added to `main.py`, and is the only place where seeding actually happens in a RuFaS simulation. Please note that if the input dictates that there should be no random seeding, the seeding libraries will "seeded" with the system time. This is important because it will prevent seeding from being carried over between simulations.

Also note that the default animal input has been modified. This was done because something about changing the seeding location changed the running of the animal module such that it tried to create new pens on the fly, which currently doesn't work on the `main` branch. This issue is being addressed in PR #1263. 

### Test plan

A dedicated unit test has been written for `set_random_seed()`, and the unit tests for `intialize_herd()` and `_initialize_simulation` have been updated. Coverage for the main routine is at 99%, with one line being missed (and which doesn't need to be tested).


main.py
```diff
@@ -274,8 +273,39 @@ def run_validation(
...
 
 
+def set_random_seed(input_manager: InputManager) -> None:
+    """
+    Sets the random seed for this simulation, if one is provided.
+
+    Parameters
+    ----------
+    input_manager : InputManager
+        The Input Manager instance that contains all input data.
```
ew3361zh, line 283: I'm curious why you're passing IM as an argument vs instantiating it.

main.py
```diff
@@ -405,10 +429,9 @@ def execute_simulations(
         is_data_valid = input_manager.start_data_processing(str(metadata_file["path"]), True)
         if is_data_valid:
             output_manager.add_log("Validation complete", "Data is valid. \nSimulating...\n", info_map)
-            simulation_config = input_manager.get_data("config")
+            set_random_seed(input_manager)
```
ew3361zh, line 432: I'm also curious why you decided to opt to keep the seed setting function in main vs simulation engine.

ew3361zh (approved):
Nice work! Works as intended, testing looks good.

I had a couple questions about decisions you made - not necessarily because I think they're the wrong decisions but just to have the record of why you made the decisions.

main.py
Isgotkowitz: I tried instantiating the IM within the method, and it somehow cleared out all the pools after the method returned which totally screwed up the simulation. So this way was an easy workaround.

main.py
Isgotkowitz: Because the `initialize_herd()` needs those libraries seeded, and that method runs in main, before the Simulation Engine is initiated.

main.py
ew3361zh: That's interesting. I wonder why that's happening. Nice workaroundl

JoeWaddell (approved):
Code looks great, and testing shows it works as expected (I tried with a dozen or so seeds, they were repeatable; also tried not setting seed, and it was random).

PR #1398, opened 2024-03-12
Update storage methane eq

tomhuhh:
This PR acts as a quick fix for the storage methane equation. We changed from the Sommer 2004 equation to Chianese 2009 equation after consulting Alan Rotz. 

More details on the comparison between the two equations can be found in issue #1271.   

### Context

### What

More details can be found in issue #1271. We added the hour to day conversion factor of 24 to convert the hourly CH4 emission to daily emission following the Chianese 2009 equation. 

### Why

We noticed that the estimated storage methane is lower than what has been reported in the literature. After some troubleshooting and consultation with Alan Rotz, we decided to update the equation in the code. 

### How

We updated the `methane_emission_from_slurry_storage` function in the `calculator.py` file.  

### Test plan

Output and plot `ManureTreatmentDailyOutput_Pen_3_LAC_COW.storage_methane`.

### DIFF
--- a/RUFAS/routines/manure/constants_and_units/gas_emission_constants.py
+++ b/RUFAS/routines/manure/constants_and_units/gas_emission_constants.py
@@ -9,6 +9,10 @@ from RUFAS.routines.manure.manure_treatments.manure_treatment_types import (
 
 class GasEmissionConstants:
     """Constants used in gas emission calculations."""
+    HOUR_TO_DAY_CONVERSION_FACTOR = 24 
+    """
+    The conversion factor from hours to days.
+    """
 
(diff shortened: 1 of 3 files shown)


allisterakun (approved):

elle-andreen (approved):
I just ran this and I am still seeing really high storage methane values from slurry storages, like 2000+ kg/cow/year. That being said, I know we still need to correct issues with VS initialization and accounting for VS loss from methane, and those are surely contributing to values being so off. But just on the basis of evaluating the equations and constants used, I think this looks correct!

KFosterReed:
Thanks @tomhuhh! @elle-andreen, I am sure you are already on top of this based on our conversation last week but lets make sure the most recent documentation reflects this change (if it was ever changed in the first place :) ).

PR #1399, opened 2024-03-12
Add water from manure application to soil

### LINKED ISSUE #1348: Change Code for Irrigation Calculations in Crop and Soil
We need to modify the Crop and Soil code in order to take into account the volume of liquid manure applied to the field when calculating irrigation needs. This is due by May 1st 2024.

## Expected Behavior

The volume of water needed for irrigation should be less after we substract both manure liquids and precipitations. 

## Current Behavior

Currently, in the crop and soil irrigation calculations only substract precipitations. Therefore irrigation needs can be overestimated. 
## Possible Solution

Substract the liquid manure applied to a field to better estimate irrigation needs.   
## Steps to Reproduce

To calculate the volume of irrigation needs per unit of surface area, we currently:
1) we ask users to input the total volume of water needed per unit of surface area for a given time period, field size, and crop type.
2) we then use weather data to calculate the volume of precipitations for that same unit of surface area, 
3) finally, we substract the volume of precipitations per unit of surface area from the total volume of water needed per unit of surface area. 

## Context (Environment)

we need a more accurate estimation of irrigation needs, it is an important cost of operation.

## Detailed Description

To substract the volume of water (liquid) contained in manure, we propose to:
1) retrieve the volume of liquid manure applied to the field from the manure module/field connector.
2) divide the volume of liquid manure applied by the field size to obtain the volume of manure liquid applied per unit of surface area. 
3) substract the volume of manure liquid applied per unit of surface area from the current irrigation calculations in volume per unit of surface area.

Isgotkowitz:
This PR takes water from liquid manure applications and adds it to the soil. It also factors this water in when calculating how much water needs to be used to meet the irrigation demand.

### Context

Issue(s) closed by this pull request: closes #1348 

### What

When there is a lot of water added to a field via a manure application, that water is now added to the soil profile the same as water from irrigation or precipitation. This water added via manure application is also factored into the calculations for how much water needs to be applied to meet the irrigation schedule that has been specified for the field.

### Why

Manure applications can contain significant amounts of water, which increase soil moisture levels and effectively act as irrigation. It is important to capture this in RuFaS because manure farmers utilize very liquid-y manure applications as part of their irrigation schedule.

### How

There are two new methods, `_add_manure_water()` and `_get_manure_water()`, which act as setter and getter for water added via manure application, respectively. `_add_manure_water()` checks that it is appropriate to apply the water from a manure application to a field, and if so records it in `FieldData`. `_get_manure_water()` grabs the manure water from `FieldData` and empties that pool, records the water it grabbed, and then returns the water for it to be applied in the `_cycle_water()` method. Manure water has also been added as an argument to `_get_watering_amount`, which is used to manage how much irrigation should be applied to a field.

### Test plan

New tests have been written for `_add_manure_water()` and `_get_manure_water()`. The places where these methods are called (`_cycle_water()` and `_execute_manure_application()`) check that they are called correctly. `_get_watering_amount()` is now tested to check that it correctly accounts for manure water.


RUFAS/routines/field/field/field.py
```diff
@@ -737,6 +740,35 @@ def _record_manure_application(
...
+        """
+
+        if manure_type is ManureType.SOLID:
+            return
+
+        water_amount_in_l = manure_application.total_manure_mass * (1 - manure_application.dry_matter_fraction)
+
+        water_amount_in_mm = self.field_data.convert_liters_to_millimeters(
+            water_amount_in_l, self.field_data.field_size
+        )
```
ltp23, line 769: I think this method does more than just converting units so maybe change the name a little bit?

ltp23 (approved):
The PR looks good to me, Ed. I like that you guarded against negative values.

RUFAS/routines/field/field/field.py
Isgotkowitz: Do you mean `convert_liters_to_millimeters()` or `_add_manure_water()`?

RUFAS/routines/field/field/field.py
```diff
@@ -1328,13 +1360,15 @@ def _cycle_water(self, current_conditions: CurrentDayConditions, time):
...
+        manure_water = self._get_manure_water()
         watering_amount = self._determine_watering_amount(
             rainfall=current_conditions.rainfall,
+            manure_water=manure_water,
             year=time.year,
             day=time.day,
             irrigation=current_conditions.irrigation,
         )
-        total_precipitation = current_conditions.rainfall + watering_amount
+        total_precipitation = current_conditions.rainfall + watering_amount + manure_water
```
krb78: should we really call this precipitation? or perhaps something like total-water?

krb78 (approved):
Looks great and runs as expected :) Good work, Ed! Thanks for adding this - it'll make a big difference when we do our C&S on our western dairies!

RUFAS/routines/field/field/field.py
Isgotkowitz: Good question. RuFaS treats water from irrigation the same way that it treats actual precipitation, which I guess we tried to be reflected in the name - but total_water would be better, I'll make that change.

RUFAS/routines/field/field/field.py
ltp23: Oh I meant the `convert_liters_to_millimeters` method.

Isgotkowitz: My bad Loi, I didn't see this until right after I merged the PR. What makes you say that it does more than just converting units?

PR #1404, opened 2024-03-13
Processing data_origin keys in the info_maps

### LINKED ISSUE #1240: Have `OutputManager` handle `data_origin` key in info map
Issues #1148 and #1160 both instruct the use of a key `"data_origin"` in the info map to indicate where the value being recorded is updated, created, and/or modified. The information associated with this key should be handled specially when the Output Manager sees it.

What happens with key specifically when the Output Manager has yet to see it has not been figured out yet.

**Additional context**

ltp23:
In this PR, I wanted to do a pre-processing step for the data origins of some variables in the output manager by making them more visible. I added an extra key called "detailed_data_origins" alongside the existing ones - "info_maps" and "values". I paired up each data origin with its corresponding value.

This won't apply to variables that don't have data origins info.  
### Context

Issue(s) closed by this pull request: closes #1240 

### What & Why

It seems that info about data origins is helpful but it is not obvious how we can make them more accessible or usable. So I think the goal of this PR is to get something going and it helps to have all the data origins info in one place.

### How

I formatted the new keys following this pattern: `[origin_class.origin_function]->[current_key]`, where `current_key` has this format `downstream_class.downstream_function.variable_name`. I could extract the variable name but it is still uncertain to me whether I can assume this pattern if suffixes somehow get involved. Similarly, the use of prefixes affects whether we can reliably replace the downstream class and functions with the origin counterparts.

If there are multiple data origins, each origin will point directly toward the current key, instead of daisy-chaining them, though this could be changed, e.g, `[[origin_class_1.origin_function_1]->[current_key], [origin_class_2.origin_function_2]->[current_key]]`

Without doing extra work, this pre-processing work is already compatible with the excluding info maps command-line option (`-i`). When info maps are excluded, there is no data about data origins to be processed.

### Test plan

I tried the following json specs file to see how things looked. Currently, this only works for json output format.

```
AnimalModuleReporter.report_daily_animal_population.num_animals
AnimalModuleReporter.report_ration_interval_data.avg_rqmts_pen_1_GROWING
```

### DIFF
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -6,7 +6,7 @@ import sys
 from copy import deepcopy
 from enum import Enum
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Union, Tuple
 
 import pandas as pd
 from deprecated.sphinx import deprecated
@@ -55,7 +55,7 @@ class LogVerbosity(Enum):
             return False
         return False
 
-    def __str__(self) -> bool:
+    def __str__(self) -> str:
         if self.value == "none":
             return "NONE"
         return self.value[:-1].upper()
@@ -379,6 +379,7 @@ class OutputManager(object):
         self.add_log("save_dict_file_try", f"Attempting to save to {path}.", info_map)
         try:
             with open(path, "w") as json_file:
+                data_dict = self._add_detailed_data_origin(data_dict)
                 if minify_output_file:
                     json.dump(
                         Utility.make_serializable(data_dict, max_depth=3),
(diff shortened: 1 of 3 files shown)


ew3361zh (approved):
Tested this and was working as designed. There are a couple uncovered continue statements in the function that I think would be easy to add in (hopefully) but the test suite otherwise looked good.

The output looked as expected.
Really nice work, Loi!

RUFAS/output_manager.py
```diff
@@ -379,22 +379,106 @@ def dict_to_file_json(self, data_dict: Dict[str, Any], path: str, minify_output_
...
+        ...    [("[AnimalManager.daily_updates]->[AnimalModuleReporter.report_daily_animal_population.num_animals]",
+        ...     193)],
+        ...    [("[AnimalManager.daily_updates]->[AnimalModuleReporter.report_daily_animal_population.num_animals]",
+        ...     194)]
+        ... ]
+        """
+
+        for key in data_dict:
+            if not isinstance(data_dict[key], dict):
+                continue
```
ew3361zh, line 451: This line showed up as uncovered.

RUFAS/output_manager.py
```diff
@@ -379,22 +379,106 @@ def dict_to_file_json(self, data_dict: Dict[str, Any], path: str, minify_output_
...
+
+            sub_data_dict = data_dict[key]
+            if "info_maps" not in sub_data_dict or "values" not in sub_data_dict:
+                continue
```
ew3361zh, line 455: And this line as well showed up as uncovered by the current testing.

JoeWaddell (commented):
Is the duplication of the information in the `values` section intended? Likewise the repetition of the new keys? In the example below, the detailed_data_origins names are all identical, and the dicts it shows below duplicate those in the values section above.
image

ltp23:
> Is the duplication of the information in the `values` section intended? Likewise the repetition of the new keys? In the example below, the detailed_data_origins names are all identical, and the dicts it shows below duplicate those in the values section above. 

I think the new keys are okay for now but a different string format is probably more useful at some point. I can control the level of nestedness for dictionary values to save memory. If we want to have all the data origins and their corresponding values in one place for later use than duplication, or deepcopying even, seems necessary.

JoeWaddell (approved):
This seems like a good first step in making `data_origins` more useful. 

The redundancy is still an open question for me, specifically that the same `detailed_data_origins` key is reported for each value, and for each of those keys the associated value is reported again (e.g. in both the `values` and `detailed_data_origins` sections). It seems that the only problem is that it makes for a larger file, so it's not a major concern at this point. Likewise, from this PR's description, and a short chat with @ltp23, those concerns will be addressed in subsequent PRs.

Issue #1368, opened 2024-03-05
Fix update of VS in anaerobic digestion and lagoon treatment


elle-andreen:
This issue is related to Issue #1270: Volatile solids updates in the manure module. Issue 1270 refers primarily to slurry storage outdoor and underfloor methods, whereas this issue will cover anaerobic digestion treatment file, as there are some differences in how volatile solids (VS) should be updated.

## Expected Behavior
In the anaerobic digestion portion of the ADL treatment, we expect that a portion of the manure VS loaded into the digester will be destroyed through the anaerobic fermentation process that occurs inside the digester. Therefore, the VS content of the manure (effluent) leaving the digester and subsequently entering the anaerobic lagoon should be lower than the manure from the reception pit. This ultimately results in lower methane emissions from the anaerobic lagoon.

Additionally, if a separator is employed between the digester and lagoon (i.e., anaerobic digestion and lagoon with separator treatment), the VS content of manure entering the separator (and other nutrients, though this issue focuses on VS) should be updated to reflect VS loss during anaerobic digestion as well. The diagram below depicts both situations.

For example, say a pen's manure VS on one day is 100 kg (already accounting for any emissions lost in housing). Assume we lose 35% of the total VS through anaerobic digestion - the VS content of the manure leaving the digester is now 65 kg. Next, we optionally use a rotary screen separator and remove 40% of the VS - the VS content of the manure leaving the separator is now 39 kg. Thus, the manure entering the anaerobic lagoon on this day contains 39 kg of VS. If we had not used the separator, the manure VS would be 65 kg. 

## Current Behavior
Currently, there is no loss of VS or TS accounted for in the anaerobic digestion treatment file. Additionally, the individual VS fractions are not tracked and updated separately.

## Possible Solution
Please see [the first page of this doc ](https://docs.google.com/document/d/1xKSRiwmKHE0tPmK0C0dkyzXynBt-DBimWSkwsSt4KAU/edit)for further detail on how VS loss in anaerobic digestion should be calculated and updated. The subsequent pages contain some information and subject matter discussion. 

The PR to address this issue should:
1. Institute tracking of degradable and non-degradable volatile solids (VSd and VSnd) in the anaerobic digestion treatment file, and removes tracking of total volatile solids.
2. Institute a conversion factor to convert from CH4 generated to CH4+CO2 generated, i.e., total VSd loss/reduction during anaerobic digestion.
[...]

PR #1413, opened 2024-03-14
Anaerobic Digestion And Lagoon Volatile solids update

### LINKED ISSUE #1408: Update inputs to anaerobic lagoon in the anaerobic_digestion_and_lagoon.py
We want to update the inputs to anaerobic lagoon in the `anaerobic_digestion_and_lagoon.py` so that they are taking in the accumulated outputs and not the daily inputs.  

## Expected Behavior

Anaerobic lagoon takes in accumulated values instead of daily values as the input. 

## Current Behavior

Anaerobic lagoon takes in daily values as the input. 

image

## Possible Solution

Update the input path. 

## Steps to Reproduce

1. 
2.
3.
4.

allisterakun:
The PR is to update the input for anaerobic lagoon in the anaerobic digestion and lagoon treatment from daily input to accumulated input 

### Context

Issue(s) closed by this pull request: closes #1408 

### What

In the current anaerobic digestion and lagoon method, the anaerobic lagoon processes takes the daily output from anaerobic digestion or separator as the input. Instead, it should take the accumulated values as the input to calculate the gas emissions.  

### Why

Gas emissions are a result of the accumulated volatile solids in the lagoon. 

### How

We reordered the functions, so the _adjust_accumulated_output function first takes in daily output from anaerobic digestion or separator (if the system has a separator after the digester) and updates the accumulated values. Then we use the accumulated values as input for the daily update function of anaerobic lagoon. Lastly, we update the accumulated values again using the output from the anaerobic lagoon.


tomhuhh:
Tagging both of you now @ltp23 @elle-andreen 

I think the unit test still needs to be updated.

tomhuhh:
The separator after the anaerobic digester does not have the attribute clone.

**Expected Behavior**
When the manure_separator_after_digestion in the manure_management.json file is assigned, the model should process with no difficulty.

**Current Behavior**
When the manure_separator_after_digestion in the manure_management.json file is assigned (scenarios 7 & 10), the model crushes and gives an attribute error of AttributeError: 'ManureSeparatorDailyOutput' object has no attribute 'clone'.

```
File "/Users/.../GitHub/MASM/RUFAS/routines/manure/manure_treatments/base_manure_treatment.py", line 382, in _adjust_accumulated_output
    self._accumulated_output = manure_treatment_daily_output.clone()
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'ManureSeparatorDailyOutput' object has no attribute 'clone'
```

The code on the main branch works fine.

tomhuhh:
To solve the issue above, we should add some code as follows to the `manure_separator_daily_output` file. 

```
 from dataclasses import astuple

 def clone(self) -> ManureSeparatorDailyOutput:
        """Returns a clone of this object.

        Returns:
            ManureSeparatorDailyOutput object with the same attributes as this object.

        """
        return ManureSeparatorDailyOutput(*astuple(self))
```

However, to make this work, I think we should align the outputs in the `manure_separator_daily_output` with those in `manure_treatment_daily_output`. 

@ltp23 (and/or @allisterakun), I wonder if this is possible? Should we solve the issues of passing degradable and non degradable volatile solids instead of total volatile solids through anaerobic digestion first? 

Thanks!

KFosterReed:
Hey Team! Thank you for bearing with me today! I think we at least found our way to an improvement as the VS are being tracked and updated as expected both with and without a separator.

This is what I get when I use the manure scenario 7 for pen 3 in the default animal json: 
image

Unfortunately, I am now getting 0 biogas outputs from the anaerobic digestion. Perhaps we lost the reporting of the anaerobic digestion daily updates with the new changes?

The last thing is I would like to know if it is possible to update the VS coming out of the Anaerobic Digestion treatment so that the amount of methane produced is subtracted from the daily output?

elle-andreen:
> The last thing is I would like to know if it is possible to update the VS coming out of the Anaerobic Digestion treatment so that the amount of methane produced is subtracted from the daily output?

Linking [issue 1368](https://github.com/RuminantFarmSystems/MASM/issues/1368) here that is in part related to this question/need.

tomhuhh:
I wanted to add that when running scenario 7 on the main branch, I also got 0 for biogas output. So I would conclude that it is not due to the fix that we implemented yesterday.

RUFAS/routines/manure/manure_treatments/anaerobic_digestion_and_lagoon.py
```diff
@@ -76,13 +76,21 @@ def _daily_update_helper(self) -> ManureTreatmentDailyOutput:
...
         self._adjust_accumulated_output(anaerobic_lagoon_daily_output)
+
+        self._accumulated_output.liquid_manure_total_degradable_volatile_solids = (
+            self._anaerobic_lagoon._accumulated_output.liquid_manure_total_degradable_volatile_solids
+        )
+        self._accumulated_output.liquid_manure_total_non_degradable_volatile_solids = (
+            self._anaerobic_lagoon._accumulated_output.liquid_manure_total_non_degradable_volatile_solids
+        )
+
+        anaerobic_lagoon_daily_output.biogas = self.anaerobic_digestion_daily_output.biogas
```
allisterakun, line 94: This is a hotfix for the issue, but I don't believe this is a good coding/design practice to manually adjust some attributes.

tomhuhh:
Thanks Allister! After the fix that Allister implemented, I was able to generate biogas outputs with scenario 7. 

I ran scenarios 0, 5, 7, and 10. Results are shown below. 

Scenario 5 (digester, no separator)
Biogas, kg/yr: 44,192 
Manure CH4, kg CO2-eq/kg FPCM: 0.36  
Total GHG, kg CO2-eq/kg FPCM: 1.33 

Scenario 7 (digester, separator after digester)
Biogas, kg/yr: 44,192
Manure CH4, kg CO2-eq/kg FPCM: 0.22  
Total GHG, kg CO2-eq/kg FPCM: 1.18

Scenario 10 (digester and separator before and after)
Biogas, kg/yr: 44,192
Manure CH4, kg CO2-eq/kg FPCM: 0.15   
Total GHG, kg CO2-eq/kg FPCM: 1.11

As a control group - Scenario 0 (no digester, no separator): 
Biogas, kg/yr: 0
Manure CH4, kg CO2-eq/kg FPCM: 0.59 
Total GHG, kg CO2-eq/kg FPCM: 1.56

elle-andreen:
> Unfortunately, I am now getting 0 biogas outputs from the anaerobic digestion. Perhaps we lost the reporting of the anaerobic digestion daily updates with the new changes?

@KFosterReed are you looking at biogas in `AnaerobicDigestion_ManureTreatmentDailyOutput` or the regular lagoon `ManureTreatmentDailyOutput`? I think actual `biogas` values are only reported in the `AnaerobicDigestion_ManureTreatmentDailyOutput` because that is the only place we're generating biogas (biogas meaning CH4 from a digester). The regular `ManureTreatmentDailyOutput` does include the `biogas` variable but always reports 0, since no biogas (by our definition right now) is made in the lagoon. 

Do you want the `biogas` and other values from `AnaerobicDigestion_ManureTreatmentDailyOutput` to be included in the anaerobic lagoon `ManureTreatmentDailyOutput` so that all the outputs from this combo treatment are reported in one singular output, or are you okay with the two outputs being separate?

elle-andreen:
> I ran scenarios 0, 5, 7, and 10. Results are shown below.

This is not exactly related to Allister's fix but I also wanted to note that unless you manually set them to 0, I believe the (undesired) removal efficiencies for volatile solids and other variables are still in place for anaerobic digestion in manure_treatment_configs. So the values you see are (maybe) not exactly reflective of what you would get if we were updating manure volatile solids according to CH4/VS loss in the digester, rather than using a fixed removal efficiency. That being said, the fixed removal efficiency for VS (0.4) is not terribly different from what the average VS removal would be if we were basing VS loss on our current methane volume estimation method (more like 0.3-0.35).

KFosterReed:
> > Unfortunately, I am now getting 0 biogas outputs from the anaerobic digestion. Perhaps we lost the reporting of the anaerobic digestion daily updates with the new changes?
> 
> @KFosterReed are you looking at biogas in `AnaerobicDigestion_ManureTreatmentDailyOutput` or the regular lagoon `ManureTreatmentDailyOutput`? I think actual `biogas` values are only reported in the `AnaerobicDigestion_ManureTreatmentDailyOutput` because that is the only place we're generating biogas (biogas meaning CH4 from a digester). The regular `ManureTreatmentDailyOutput` does include the `biogas` variable but always reports 0, since no biogas (by our definition right now) is made in the lagoon.
> 
> Do you want the `biogas` and other values from `AnaerobicDigestion_ManureTreatmentDailyOutput` to be included in the anaerobic lagoon `ManureTreatmentDailyOutput` so that all the outputs from this combo treatment are reported in one singular output, or are you okay with the two outputs being separate?

I was looking at the `AnaerobicDigestion_ManureTreatmentDailyOutput` when there was no biogas being produced. I haven't tried to re-run it since the fix was made and the results @tomhuhh shared look great! 

@elle-andreen makes some great points about the removal of the VS in the anaerobic digestion component. If we can address issue 1368 in this PR then I think we will have a fully operational combo treatment.

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
```diff
@@ -368,3 +369,46 @@ def clone(self) -> ManureTreatmentDailyOutput:
...
+            pen_id=manure_separator_daily_output.pen_id,
+            simulation_day=manure_separator_daily_output.simulation_day,
+            liquid_manure_total_ammoniacal_nitrogen=(
+                manure_separator_daily_output.liquid_manure_total_ammoniacal_nitrogen
+            ),
+            liquid_manure_nitrogen=manure_separator_daily_output.liquid_manure_nitrogen,
+            liquid_manure_total_solids=manure_separator_daily_output.liquid_manure_total_solids,
+            liquid_manure_phosphorus=manure_separator_daily_output.liquid_manure_phosphorus,
+            liquid_manure_potassium=manure_separator_daily_output.liquid_manure_potassium,
+            # TODO: Check if this is correct
```
elle-andreen: I believe this is correct, so this TODO could probably be removed before merging, if you want.

input/data/animal/default_animal.json
```diff
@@ -181,7 +181,7 @@
 			"housing_type": "open air barn",
 			"pen_type": "freestall",
 			"max_stocking_density": 1.2,
-			"manure_management_scenario_id": 5
+			"manure_management_scenario_id": 10
```
elle-andreen: I would maybe revert this back to scenario 5, unless there is an explicit reason not to

elle-andreen (approved):
Thanks Allister! I left two minor comments but this portion of the digestion->lagoon and digestion->separator->lagoon treatments is functioning as desired now. 👍

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
```diff
@@ -380,5 +380,7 @@ def _adjust_accumulated_output(self, manure_treatment_daily_output: ManureTreatm
                     },
                 )
             self._accumulated_output = manure_treatment_daily_output.clone()
+            # TODO: Add the following logic
+            # Accumulate the output with repeated calls
```
Isgotkowitz: Does this TODO need to be hit in this PR? If not then adding in some sort of note about when it should be taken care of would be good.

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
```diff
@@ -368,3 +369,46 @@ def clone(self) -> ManureTreatmentDailyOutput:
 
         """
         return ManureTreatmentDailyOutput(*astuple(self))
+
+    @staticmethod
+    def from_manure_separator_daily_output(
```
Isgotkowitz: Maybe it would be good to add "get_output_..." to the beginning of this method name, so it starts with a verb

Isgotkowitz (approved):
I think this looks good. I ran the simulation and collected biogas outputs. It looks like it is only being produced in Pen 3 (from lactating cows), not sure if that is correct or not.

I noticed the outputs `AnaerobicDigestion_ManureTreatmentDailyOutput_Pen_3_LAC_COW.biogas` and `ManureTreatmentDailyOutput_Pen_3_LAC_COW.biogas` seem to be exactly the same, but `AnaerobicDigestion_ManureTreatmentDailyOutput_Pen_3_LAC_COW.biogas_energy_content` had values while `ManureTreatmentDailyOutput_Pen_3_LAC_COW.biogas_energy_content` had only zeros. Maybe some info is being missed when it is copied over?

Issue #1381, opened 2024-03-06
[CI/CD] Git Actions house keeping 


PooyaHekmati:
Below is the current list of actions that we have. There are duplicated and deprecated actions in it. Clean up this list by eliminating the actions that are no longer in use. Give good names to the remaining ones. Consider consolidating actions if possible.
![image](https://github.com/RuminantFarmSystems/MASM/assets/51127592/3b43714b-2b46-4e6b-bddc-0de0048bba11)

ltp23:
- [x] Clear the workflows runs that are no longer relevant, using Github CLI API requests for bulk delete.
- [x] Delete `testing_pytest.yml` in a PR.
- [x] Combine pytest and mypy actions.
- [x] Write issue for combining the linting action to the combined actions yml file. #1451 
- [x] Write issue for combining the build-sphinx action to the combined actions yml file. #1452 
- [x] Write issue for the `set-output` command being deprecated soon. #1453 
- [x] Write issue for automating the cleanup of old action results. #1454 

---

The command I used for step 1 was

```
# Replace the workflow name. List the last 800 runs and keep the latest 300.
# The rate limit for this was 5000 requests per hour. It cost more than 1 request per deletion.

gh run list --limit 800 --workflow 'Test and Coverage' --json databaseId,createdAt | jq -r 'sort_by(.createdAt) | .[:-300] | .[] | .databaseId | @sh' | xargs -I{} gh run delete "{}"
```

PR #1415, opened 2024-03-15
[CI] Making MyPy runnable locally

### LINKED ISSUE #1380: [CI/CD] Make mypy locally runnable
We have integrated mypy in our CI/CD pipeline on GitHub. It is beneficial to make it easy to run mypy locally as well so that we can check the code before submitting PRs. Also, we should unify any configs that is being used. This is very similar to what we did with flake8. @Isgotkowitz may have some transferrable experience here.

With the addition of the mypy type checker to the github workflows, there should be an entry page in the wiki to describe its basic functionality, the error count badge on the readme page, and the later enforcement of making sure each new PR does not introduce more type-checking errors.

Isgotkowitz:
MyPy is configured so that it is more useful/usable and is added to scripts that make it easier to run MyPy locally.

### Context

_NOTE_: I will update the Wiki to make it clearer about how to run MyPy locally, but I will do this after the PR has been reviewed and I'm confident about the details that I put in the wiki.

Issue(s) closed by this pull request: closes #1380

- It seems like the changelog doesn't need to updated for this one - nothing about this PR should effect external users. The [changelog](https://docs.google.com/spreadsheets/d/1U-3igxdstHTFb6k5dVcE5-x9IL0r6q18eUiwwytKZyA/edit#gid=0) has been updated. 

### What

- Configured MyPy's `follow-imports` argument to be "silent" in `pyproject.toml` (the default is "normal")
- Renamed the `lint_branch_changes` scripts (shell and batch) to `check_changes` and made them run MyPy on all changed Python files in addition to Flake8.

### Why

By changing `follow-imports` to be "silent" instead of "normal", MyPy will only check within a file for errors and will not follow import statements. This reduces the intractable wall of errors that MyPy produces while checking a single file into a much more manageable list. This is in line with the guidance in the [MyPy docs](https://mypy.readthedocs.io/en/stable/running_mypy.html#following-imports):

> If you are planning on adding type hints to a large, existing code base, we recommend you start by trying to make your  entire codebase (including files that do not use type hints) pass under [--follow-imports=normal] (https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-follow-imports). This is usually not too         > difficult to do: mypy is designed to report as few error messages as possible when it is looking at unannotated code.

> Only if doing this is intractable, we recommend passing mypy just the files you want to type check and use [--follow-imports=silent](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-follow-imports). Even if mypy is unable to perfectly type check a file, it can still glean some useful information by parsing it (for example, understanding what methods a given object has). See [Using mypy with an existing codebase](https://mypy.readthedocs.io/en/stable/existing_code.html#existing-code) for more recommendations.

Eventually, after we have reduced the number of MyPy errors in the entire codebase to 0 (or just something more manageable), we could change the `follow-imports` setting to "normal" and then work through all those errors.

Re: changing the scripts, they make it easier to check all places where MyPy errors may have been introduced.

### Test plan

I tried making trivial changes to a Python file in the `RUFAS` directory, and running both the shell and batch scripts to check that they work as expected. I have also tried running Mypy directly on a single file at a time to make sure it runs the same as in the script.

I advise reviewers to pull this branch locally and try running the scripts after making a trivial change. Additionally, I recommend running Mypy on a single file before and after changing `follow_imports = "silent"` to `follow_imports = "normal"` to observe the effect on the error output.


ltp23 (approved):
This looks good to me! Good job, Ed!

I verified it in two ways: switching between the normal and silent options; running mypy like normal and grepping for the changed files only. The results from two approaches were almost exactly the same.

A more involved follow-up task to explore is to see if it is possible and valid to filter out the errors not related to the changed lines.

Isgotkowitz:
> This looks good to me! Good job, Ed!
> 
> I verified it in two ways: switching between the normal and silent options; running mypy like normal and grepping for the changed files only. The results from two approaches were almost exactly the same.

Thank you Loi. I appreciate you checking the correctness of the scripts.

> A more involved follow-up task to explore is to see if it is possible and valid to filter out the errors not related to the changed lines.

That would certainly be a more involved follow-up task. An alternative could be to start enforcing the same rules as we do for Flake8: no errors at all in any Python files changed in a PR.

allisterakun (approved):
Changes look good. Great job, Ed!

Just one comment: since we have combined flake8 and mypy into a single script locally, do you think it's a good idea to combine the lining and mypy Git actions into one (so that we can save on the running time)?

Isgotkowitz:
> Changes look good. Great job, Ed!
> 
> Just one comment: since we have combined flake8 and mypy into a single script locally, do you think it's a good idea to combine the lining and mypy Git actions into one (so that we can save on the running time)?

Thanks Allister. I do think it is a good idea, but it is outside of the scope of this PR - but it is covered/mentioned in #1381

Issue #1099, opened 2024-01-10
`Time` should be single-source of truth for tracking and measuring time in RuFaS


Isgotkowitz:
There are multiple places in the code where time is kept/tracked/utilized in various forms. This is problematic as it makes the code non-standard and bug prone. All time-related uses should use the `Time` object.

## Expected Behavior

All cases where time needs to be managed or tracked should utilize the `Time` class.

## Current Behavior

Time is tracked and passed around in multiple different forms. For example, instead of passing a `Time` instance to demonstrate what the time is, two ints are used - one representing the ordinal day and the other the calendar year.

## Possible Solution

After refactoring the `Time` class, comb the codebase to look for places where time is being represented in a form that is not a `Time` instance, and change it to use `Time`.

allisterakun:
`Time` will probably need to be a singleton to enforce that different modules are all on the same page.

ew3361zh:
If desired, re-incorporate a simulation progress bar while working on this task.

ew3361zh:
Investigate this issue and create child issues to break this up and solve this issue.

ew3361zh:
Wrote up 2 issues to break this down a little further into more tangible steps: #1522 and #1533. I added some initial estimated T-shirt sizes to those but those can be debated at an upcoming grooming session.

PR #1418, opened 2024-03-15
Delete stdout msgs

### LINKED ISSUE #1161: Delete messages printed to standard output when running simulations.

ew3361zh:
PR to remove the stdout write messages and add a log verbosity to print the credits.

### Context

Issue(s) closed by this pull request: closes #1161

### What

1. Update LogVerbosity class to include a member for printing the credits (i.e. Ruminant Farm Systems...). Date has been removed from credit line.
2. Changes default LogVerbosity to credits.
3. Updates stdout messages to logs sent to OM (which will in turn link them with the metadata prefix used in the scenario in which they are logged).
4. Adds function in OM to print credits when LogVerbosity is any setting but `None`.
5. Removes progress spinning wheel from printing out to terminal. Not replaced with another progress bar - was a little complicated and because the Time class is being refactored soon, I think that will be better suited to be paired with that task #1099.
6. Updates unit testing.

### Why

These changes will reduce clutter printing out in the terminal during a default simulation. Logging the messages that used to print out will make it clearer as to which metadata scenario file they are being run from (which will make it easier to keep track of simulations of sequential scenarios).

### How

The LogVerbosity class was slightly refactored to accomodate the new member and is hopefully clearer as to what's happening with the different members including one another. The rest is fairly straightforward.

### Test plan

Testing was developed for the new OM function `print_credits()` and updated across simulation_engine, main, and output_manager.


RUFAS/output_manager.py
```diff
@@ -1220,3 +1227,11 @@ def get_error_and_warning_counts(self) -> tuple[int, int]:
...
         warnings_count = sum([len(value_dict["values"]) for value_dict in self.warnings_pool.values()])
         return errors_count, warnings_count
+
+    def print_credits(self) -> None:
+        """
+        Prints out the RuFaS credits when LogVerbosity is set to any level except None.
+        """
+        if self.__log_verbose >= LogVerbosity.CREDITS:
+            current_year = date.today().year
+            sys.stdout.write(f"RuFaS: Ruminant Farm Systems Model {current_year}\n")
```
Isgotkowitz: Would it maybe be better to not print the date at all? Right now there isn't anything 2024 specific to this "version" of RuFaS. But to me, it almost seems to be implied that this is the 2024 version of RuFaS.

main.py
```diff
@@ -428,7 +422,9 @@ def execute_simulations(
         output_manager.set_metadata_prefix(metadata_file["prefix"])
         is_data_valid = input_manager.start_data_processing(str(metadata_file["path"]), True)
         if is_data_valid:
-            output_manager.add_log("Validation complete", "Data is valid. \nSimulating...\n", info_map)
+            output_manager.add_log(
+                "Validation complete", f"Data is valid. \nSimulating...{metadata_file['prefix']}" " scenario", info_map
```
Isgotkowitz:
```suggestion
                "Validation complete", f"Data is valid. Simulating {metadata_file['prefix']} scenario.", info_map
```
I think the log message looks a little cleaner like this.

Isgotkowitz (approved):
Looks great, nice job Niko.

RUFAS/output_manager.py
ew3361zh: True, I can remove that. Maybe we eventually put a version number if anything.

main.py
ew3361zh: Updated!

ltp23 (approved):
It looks good to me, Niko!

I tested the different log verbosity levels and they worked as expected. I like that now we have more control over what is printed out to the console.

PR #1427, opened 2024-03-19
User defined ration percentage sum warning fix

JoeWaddell:
Quick fix for mistake I made in PR #1385. 

The mistake: I hadn't pasted in the correct names for the duplicated code checking the sums. Since the rest of those error messages were correct, I fooled myself when testing that all error messages were flagged at the same time. Quick fix, now works as expected.


jadamchick (approved):
looks good. When I fiddle with each of the rations, it gives an error appropriate to that animal combo & ration. 

edit: reminder to add it to the change log!

ew3361zh (approved):
Looks good, Joe, good catch!
Changelog still needs to be updated.

Issue #1366, opened 2024-03-05
Misalignment in graphs (and reports) due to instantiation of new variables during simulation


JoeWaddell:
## Status
This is a problem encountered in #1263, specific to the creation of new pens, and it may extend to other outputs.
When a new output is added to output manager, it causes misalignment in graphs and other outputs, because an assumption that the first entry is on simulation day 0 is incorrect.

**A super-simplified example of the problem**
Assume one runs a simulation for 15 days.
The simulation starts with 4 pens.
A 5th pen is instantiated on day 10 (see logic in #1263).
At the end of the simulation, each output related to pens 0-3 have a length of 15 (or 16, if we're counting day 0).

*The problem:* any outputs related to pen 4 has 10 fewer days, and so there's a misalignment in those outputs.

**Example output from a longer sim**
Note the crude arrows showing shorter line lengths.
![default_animalmodulereporter-daily_pens-05-Mar-2024_Tue_11-10-38](https://github.com/RuminantFarmSystems/MASM/assets/109082924/8caad026-a6e3-4261-b2db-cb788faad1e6)

Note below the hidden rows to show the "ragged" nature of these new pen outputs (this is without the "padding" I instituted in 1263).
![image](https://github.com/RuminantFarmSystems/MASM/assets/109082924/dac78bde-f539-46f0-9a26-85b7e9ffe236)

So without some note of what the time is at the time of collection, we lose information. (Note in "Yet another solution" that we can already store that in info_map, but that's harder for an end user to access, and lost completely in graphs).

## Intention
We need to ensure that the user is certain of when data was logged in output manager.

**What the above *should* look like** (thanks @Isgotkowitz for the suggestion to show some examples of ideal behavior)
- There are two options I see for the reports/CSVs:
    - Each variable would have its x-axis value printed next to it, so that there would be double the number of columns in that example. 
![image](https://github.com/RuminantFarmSystems/MASM/assets/109082924/01703a25-4fbd-45b4-b9f5-4fc38fd1118e)

    - To avoid the duplication inherent in the previous bullet point, there could be a single x-axis that all results are aligned with, e.g. the first column is simulation_day, and every entry in a column is aligned to that axis.
    - However, this would have a problem when printing a mix of monthly and daily variables, since there would be enormous gaps, and would be annoying for those mainly working in Excel.
[...]

PR #1428, opened 2024-03-19
Optionally auto-include simulation day to info map when adding variable to OM pool

ltp23:
This PR adds an option in the config json to enable the auto-inclusion of the current simulation day in the info map when the OM's `add_variable()` method is called.

### Context

### What & Why

To address the two issues #912 and #1366 more effectively, I think it is helpful or even necessary to associate each value added to the output manager with its corresponding simulation day. 

Currently in the `graph_generator.py` when you plot a variable (a time-series), the `_draw_graph()` method at line `361` (`plot_function(value)`) assumes the x-axis to be the indices of the value series, which is a happy coincidence that it matches the concept of simulation day in RuFaS. 

Also, the list of matplotlib plot functions in `MATPLOTLIB_PLOT_FUNCTIONS` is not all working because some of them require a supply of both x- and y-series.

### How

I don't have all the steps laid out yet to solve the two issues listed above, but I think this PR is a good first step to at least get a hold of the time data needed to pair up with the corresponding value data.

I added a flag in the `default_config.json` to turn on/off this auto-inclusion feature: `"auto_generate_simulation_day": false`.

If the info map already has the simulation day in it, then that will take precedence over the autogenerated one.

In order to synchronize the time in the simulation engine with that in the output manager, I give the output manager a very simple callback function from the simulation engine so that the OM can retrieve the current simulation day when needed. The detail of the callback may need to be modified a bit later when the `Time` class is refactored so that one place is responsible for keeping track of the simulation day. 

### Test plan

- You can try to turn on and off the config flag.
- Pick a variable that doesn't have a simulation day in its info map and check out the results.


RUFAS/output_manager.py
```diff
@@ -1227,6 +1234,18 @@ def get_error_and_warning_counts(self) -> tuple[int, int]:
...
 
+    def set_get_simulation_day_callback(self, callback: Callable[[], int] | None) -> None:
+        """
+        Set the callback for getting the simulation day.
+
+        Parameters
+        ----------
+        callback : Callable[[], int]
+            The callback for getting the simulation day.
+        """
```
Isgotkowitz, line 1245:
```suggestion
        
        """
```

RUFAS/simulation_engine.py
```diff
@@ -54,6 +54,9 @@ def __init__(self) -> None:
         self.day_counter: int = 0
         self._initialize_simulation()
 
+        if im.get_data("config.auto_generate_simulation_day"):
+            om.set_get_simulation_day_callback(lambda: self.day_counter)
```
Isgotkowitz, line 58: Instead of using the day counter in `SimulationEngine` to track the time, what about using the `index` in the `Time` class? Even though the name is worse, they track the same thing, and it would be beneficial because it moves us closer to realizing `Time` as the single source of truth for tracking time in RuFaS simulations/

RUFAS/output_manager.py
```diff
@@ -172,7 +173,13 @@ def add_variable(self, name: str, value: Any, info_map: Dict[str, Any]) -> None:
             Has no effect on manual prefix overrides.
         info_map["suffix"] : str, optional
             If present, gets appended to the key
+        info_map["simulation_day"] : int, optional
+            If present, overrides any autogenerated simulation day
```
Isgotkowitz, line 177:
```suggestion
            If present, overrides any autogenerated simulation day

```

input/data/config/default_config.json
```diff
@@ -5,5 +5,6 @@
     "set_seed": true,
     "simulate_animals": true,
     "nutrient_standard": "NASEM",
-    "FIPS_county_code": 55025
+    "FIPS_county_code": 55025,
+    "auto_generate_simulation_day": false
```
Isgotkowitz, line 9: Instead of adding to the inputs about whether or not this should be added, maybe it would be better to always record the day in the info maps and make it optional to use when generating a graph or report?

RUFAS/output_manager.py
```diff
@@ -120,6 +120,7 @@ def __init__(self) -> None:
                     "function": self.__init__.__name__,
                 },
             )
+            self._get_simulation_day_callback: Callable[[], int] | None = None
```
Isgotkowitz, line 123: I think this attribute should be reset in the method `flush_pools`, just to make the code a bit more robust.

RUFAS/output_manager.py
```diff
@@ -1227,6 +1234,18 @@ def get_error_and_warning_counts(self) -> tuple[int, int]:
         warnings_count = sum([len(value_dict["values"]) for value_dict in self.warnings_pool.values()])
         return errors_count, warnings_count
 
+    def set_get_simulation_day_callback(self, callback: Callable[[], int] | None) -> None:
```
Isgotkowitz, line 1237: Instead of passing a function in that gets the simulation day, what about just making Time an attribute of `OutputManager` that is passed by Simulation Engine at the beginning of a simulation? Then the index can just be read out of the Time class. IMO this would be simpler than using lambda expressions.

input/metadata/properties/default.json
```diff
@@ -40,6 +40,11 @@
       "maximum": 56045,
       "default": null,
       "description": "Unique 5-digit code that represents a specific U.S. county."
+    },
+    "auto_generate_simulation_day": {
+        "type": "bool",
+        "description": "Whether or not to automatically associate a simulation day with each output value if one is not provided",
+        "default": false
```
Isgotkowitz, line 47: I think this should default to true instead.

tests/test_output_manager.py
```diff
@@ -2018,3 +2018,25 @@ def test_add_detailed_data_origin(input_data: Dict[str, Dict[str, Any]], expecte
 
     # Assert
     assert result == expected
+
```
Isgotkowitz, line 2021: The changes to `add_variable()` still need to be tested.

Isgotkowitz (changes requested):
I think this is an excellent start towards addressing #912 and #1366. I am requesting changes on only one thing: the change to the `add_variable()` does not appear to be tested. Once that gets fixed, I can come back and approve.

I have also left a bunch of other comments with suggestions, but address or ignore them as you see fit.

PooyaHekmati:
This PR is only about the tech part of the code and does not involve any logic related to the animal module or other sciency things, therefore unassigning @JoeWaddell as a reviewer. Such PRs should not concern SMEs.

ltp23:
Closed because this feature is not needed.

PR #1444, opened 2024-03-25
Update full_feeds_emissions.csv

tomhuhh:
Replaced the full_feeds_emissions.csv with the file Kristan uploaded to the pilot testing branch.

We updated the full_feeds_emissions.csv with the most recent feed emission factors. 

### Context

This is related to a PR #1360 on the pilot testing branch.   

### What

The feed emission factors of certain feeds are outdated and show invalid results. 

### Why

The updated file provides more up-to-date feed emission factors. 

### How

The uploaded file updates the invalid and/or outdated emission factors for feeds.


tomhuhh:
Notes from Kristan: 

`Know that even the updated feed emissions library is still going to have higher feed emissions than previous estimates`

jadamchick (changes requested):
Thanks Haowen. I think the file should be uploaded as a separate csv with a new name rather than replacing the existing file. We discussed this in PR1360 and decided that having two distinct files was the best way to go. 
Kristan's comment: 

" I would prefer to update the metadata with the new name because it will be easier to know (especially if it is on main), that we are using the correct file. There should be a way to find and replace all instances of 'full_feeds_emissions.csv' with 'full_feeds_emissions_new.csv' with a few clicks."

tomhuhh:
Thanks Julie! I added the previous version back and named it `full_feeds_emissions_before_Mar24.csv` to distinguish it from the current version. That way, we do not need to update all the metadata files. Thanks again for bringing this to my attention!

jadamchick (commented):
Great! 
My only thought now is concern because of the naming discrepancy between this branch and pilot testing. 
On pilot_testing_Dec2023, we can easily update the name of "full_feed_emissions.csv" to be "full_feeds_emissions_before_March24.csv", so the two outdated files are the same (on main & on pilot_testing). 
But it would be really nice if the name of the up-to-date csv they are actually using is also the same. I'm thinking of at least 2 scenarios where this might matter: 
a) when someone is using a branch based off of main (eg a PR for review) and we want to test it using pilot farms. The metadata path will need to be changed any time we want to use branches based on one branch vs. the other. 
b) when syncing/merging Main into pilot_testing, I'm not sure how it will handle the different files. I assume it will add the new "full_feed_emissions.csv" to pilot_testing, so then we will either need to manage identical files with different names or just update the naming on pilot_testing to match what is on Main, which will require updating all of the metadata files on pilot_testing to the new file name. 

I think I'm asking if it would be possible to match the file names that are already established on pilot_testing? Since there are only a handful of metadata files on Main and it would be a quick update. 
Happy to discuss more and you might have a better solution in mind!

Isgotkowitz (approved):
Ran the default scenario with the updated feed emissions, worked like a charm. Still missing feed data for 216 and 176, but I think that's probably not a big deal.

Once the file names get work out I think it will be good to go.

jadamchick (approved):
Looks good. Thanks Haowen!

PR #1445, opened 2024-03-25
[IM] Encoding changed to UTF-8 from UTF-16

Isgotkowitz:
This is an actual fix to the Input Manager loading CSV files with an incorrect encoding in the method `_load_data_from_csv`. Instead of using UTF-16 in Pandas `read_csv` method, Python's `open` method has been modified to use UTF-8

### Context

### Test plan

- On a Windows machine
- Switch to this branch
- Modify the weather file in `default_metadata.json` to use `barnyard_100yr.csv`.
- Run the simulation.
- Also, to verify that other encodings _do not_ work, try changing the encoding in `open` to UTF-16 and/or removing the encoding in `open` and trying both UTF-8 and 16 in the Pandas `read_csv` call in Input Manager's `_load_data_from_csv()` method.

### DIFF
--- a/RUFAS/input_manager.py
+++ b/RUFAS/input_manager.py
@@ -253,8 +253,8 @@ class InputManager:
         }
         om.add_log("open_csv_file", f"Attempting to open {file_path}.", info_map)
         try:
-            with open(file_path, "r") as csv_file:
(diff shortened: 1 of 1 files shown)


allisterakun (approved):
Simple change! LGTM!

PooyaHekmati (approved):

PR #1450, opened 2024-03-25
Modify removal of manure nutrients when exceeded amounts requested

### LINKED ISSUE #1449: "Remove more nutrients than available: dry_matter" error needs to be a warning, not an error
When I run a simulation (currently c1f2), this farm applies only manure to fields (no fertilizer). I get an error message that stops the simulation telling me I don't have enough manure. 

[25-Mar-2024_Mon_12-48-21.894464][ERROR][c1f2_wcrops] Dumping all logs from main.py because of error 'Remove more nutrients than available: dry_matter'. This terminal error occurred during runtime. Traceback (most recent call las
t)

## Expected Behavior
I'd prefer if this only gave a warning and didn't stop the simulation. This is all the manure I have available (**insert oh shit**)

## Current Behavior
Right now the error ends the simulation, which is zero fun. 

## Possible Solution
Not sure how to do this but a "warning" >> "error" in this scenario 😄 

## Steps to Reproduce
1. Run the c1f2 metadata with crops file. with the herd starting in 1997 ending in 2020.
2. Wait 1hr...

ltp23:
### Context

Issue(s) closed by this pull request: closes #1449 

### What & Why

When the requested manure nutrients exceed what is currently available in the manure nutrient manager, an error will be raised. This PR converts that to a warning and adds the warning to the output manager. It is also necessary to match the returned values with the amounts of nutrients actually removed.

The error that Kaitlyn saw had to do the over-removal of the dry matter component in the nutrient request results. So when that happens, I want to cap that at what is currently available.

### How

- Add a warning when the requested amounts are larger than the available amounts.
- Update unit tests.

### Test plan

- Test on different scenarios.
- Have Kaitlyn test it out on the farm that had the error.


krb78 (approved):
After many hours of running the scenario, I didn't get any errors in trying to apply manure, so I think we are golden ✔️ 😄  Thanks, Loi!

RUFAS/routines/manure/manure_nutrients/manure_nutrient_manager.py
```diff
@@ -268,17 +272,38 @@ def _remove_nutrients(self, results: NutrientRequestResults, manure_type: Manure
             If any of the nutrients in the results is greater than what is currently available in the manager.
 
         """
-        for attr in ["nitrogen", "phosphorus", "total_manure_mass", "dry_matter"]:
-            if getattr(self._nutrients_by_manure_type[manure_type], attr) < getattr(results, attr):
-                raise ValueError(f"Remove more nutrients than available: {attr}")
 
-        current_nutrients = self._nutrients_by_manure_type.get(manure_type)
+        if manure_type not in self._nutrients_by_manure_type:
+            raise ValueError(f"Invalid manure type: {manure_type}. Supported types are: {ManureType}")
```
Isgotkowitz, line 277: This ValueError is not covered in testing.

Isgotkowitz (approved):
The check of the ManureType isn't covered by the tests, but other that that everything looks good. Nice job Loi

RUFAS/routines/manure/manure_nutrients/manure_nutrient_manager.py
ltp23: Thanks Ed, I have updated the test.

PR #1455, opened 2024-03-26
Jersey herd hotfix pilot testing dec2023

KFosterReed:
### Context
The birthweight and bodyweight of Jersey animals are not updated quickly enough when the initial herd population is a Holstein herd.

### What
 This PR provides a Jersey specific animal population file `jersey_animal_population.json` and updates the metadata properties file so that it does not throw an error due to the repro inputs of the Jersey herd. 

### Why
To more accurately simulate Jersey herd bodyweights

### How
To use this jersey herd population at initialization, change the file path referenced in the `"animal_population"` field to the new `"jersey_animal_population.json"` file instead of the default file.

### Test plan
Simulate a jersey herd with the new animal population file and compare the bodyweight outcomes with the simulation when using the default herd


jadamchick (approved):
Runs smoothly! Ran it for c2f24, a Jersey herd with input mature body weight of 499 kg. 
From one run, the average MBW was 480 kg (compared to 597 previously). 
Other results that seem different (& I'm assuming are connected to body weight, not to the ration/manure updates?): 

- lact cow DMI 17.1 kg/cow/d (was 20.2)
- milk allocation factor 0.93 (was 0.90) -- bc the weight of market animals is less with no change to amount of milk produced
- lact cow enteric methane/cow/d 0.315 (was 0.371)
- feed emissions kg CO2e/whole herd (adjusted) 0.68 (was 0.77) -- bc they're eating less 

there may be other manure differences too, but I think it's mixed in with updates to manure functionality so tough to tease out here

KFosterReed:
That's great to hear! I noticed the changes to the feed intake in the default branch too and it makes sense that they are eating less because they have a smaller bodyweight and so will be producing less methane and manure as well. I'm going to merge it with this branch and then comment on the bodyweight issue to provide some more direction/push to get it in the main branch as well

PR #1456, opened 2024-03-26
[Feed] [Back to Draft while Effluent calculation is fixed!] Implement `Sileage` logic [1/2]

### LINKED ISSUE #1376: [Feed] Populate logic in `RUFAS/routines/feed_storage/sileage.py`
Populate the logics of the following methods in `RUFAS/routines/feed_storage/sileage.py`:

- [ ] `calculate_protein_loss`
- [ ] `calculate_dry_matter_loss_to_gas`
- [ ] `calculate_dry_matter_loss_to_effluent`

[Design doc is here](https://docs.google.com/document/d/1C86lch2dkYBERR-qLDmE7L7n1gYQjETrIKUyrQpWrPs/edit)

* For questions, reach out to SME: @KevinP-B and/or SWE @PooyaHekmati . Their reviews are needed for any linked PRs.
* Size the PRs reasonably, not all the methods have to be implemented in a single PR.
* docstrings, function signatures, and unit tests need to be updated accordingly.
* I suggest reading and understanding the design doc before attempting to write the code.

Isgotkowitz:
Implements the logic specific to ensiled crops in the `Sileage` class.

### Context

Issues addressed by this pull request: addresses #1376

### What

The logic added is specific to calculating effluent loss from silage. Silage does lose dry matter to fermentation, but does so by calling the Storage class to handle that loss. It also adds a helper method and extra attributes to Harvested Crop, which are used in the calculation of effluent loss.

### Why

Silage is the only storage type among the main four (silage, baleage, hay, and grain) that experiences effluent loss. Thus, it is the only class that needs to implement logic related to effluent loss.

### How

Silage goes through the following process to calculate and report effluent for each crop:
- Calculates the maximum effluent that is lost from a crop.
- Calculates the dry matter lost to effluent on the current day.
- Calculates the loss coefficients that determine how the crude protein amounts and non-protein nitrogen percentages change in the stored crop due to effluent loss.
- Calculates and sets the crude protein and non-protein nitrogen percentages in the crop after loss.
- Resets the mass attributes in the ensiled crop.
- Reports the total dry matter loss to effluent.
- Calls `Storage` to handle dry matter loss to fermentation and record the state of the ensiled crops.

### Test plan

All methods added have been unit tested to ensure complete code coverage.

### Open Questions
- Do any of Silage's children need to be tested? They do not use different logic at all from Silage, so I assume they do not.
- Do Silage's children need to have their capacity set? Right now they only take "size" in their init method.

### NOTE
After speaking w/ Kevin this morning, a few issues came up with some of the silage logic (and the name). This PR is good to be reviewed, but there will be further changes needed before silage is complete - made in the subsequent PR #1471.


RUFAS/routines/feed_storage/sileage.py
```diff
@@ -19,55 +45,160 @@ def __init__(self, capacity: float = float("inf")):
...
 
         Returns
         -------
         float
-            The amount of dry matter lost to effluent, specific to Sileage.
+            The amount of dry matter the crop loses to effluent in kg.
+
+        """
+        time_in_silo = crop.days_stored(time)
+        return (0.1035 * estimated_maximum_effluent) * (0.1 * time_in_silo) / crop.dry_matter_mass
```
Isgotkowitz, line 143: This logic needs a small addition. I plan on doing this in a subsequent PR, but if reviewers would prefer that I get everything in one go, let me know.

tests/test_feed_storage_module/test_sileage.py
```diff
@@ -17,36 +40,196 @@ def test_acceptable_crops(sileage: Sileage):
...
+
+@pytest.mark.parametrize(
+    "dry_matter,max_effluent,days,expected",
+    [
+        (100.0, 25.0, 8, 0.0207),
+        (350.0, 40.0, 2, 0.002365714),
+        (30.0, 55.0, 12, 0.2277),
+        (400.0, 12.0, 1, 0.0003105),
+        (100.0, 0.0, 4, 0.0),
+    ],
```
Isgotkowitz, line 144: @KevinP-B these expected loss numbers seem a little low to me, what do you think?

RUFAS/routines/feed_storage/sileage.py
KevinP-B: I'm fine with it being addressed in #1471. For reference, estimated_max_effluent is returning a percent of dry matter here as written. As per our convo, there will need to be some changes to how this is implemented before review.

PR #1460, opened 2024-03-27
Exclude info maps when adding elements to OM pool with GNU arg `-i` set

ltp23:

### What & Why

### How
- [ ] Set the exclude info maps flag at the start of the simulation so that all the calls to `_add_to_pool()` in the OM will ignore the info maps if the flag was set to `True`.

#### On main branch
on-main-branch

#### On the current branch

### DIFF
--- a/.github/workflows/combined_pytest_mypy.yml
+++ b/.github/workflows/combined_pytest_mypy.yml
@@ -103,16 +103,16 @@ jobs:
           if [ -n "$URL" ]; then
             PYTEST_COVERAGE="Current Coverage: ${{ steps.coverage.outputs.coverage_percentage }}%"
             echo "$PYTEST_COVERAGE"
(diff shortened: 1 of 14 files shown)


PooyaHekmati (changes requested):
Unable to review, The PR is mixing different issues; please respect the single responsibility principle. Also, it is missing change log entry.

PooyaHekmati (approved):
Looks good, however, I am concerned that when `-i` is passed, we don't append anything to `pool[key]["info_maps"]`, not even empty dict; this may be problematic downstream. So it is worth asking SMEs to run this code with various filters to test it.
Also, the changelog entry `Exclude info maps when adding elements to OM pool with GNU arg -i set` can be `Won't save info maps in memory when adding elements to OM pool when GNU arg -i set`

allisterakun (approved):
LGTM

PR #1461, opened 2024-03-27
Farm specific lactation curve

### LINKED ISSUE #1393: Milk production does not change as expected when "Jersey" lactation parameters are updated
## Urgency Level

- [X] Medium: the issue affects functionality but has a workaround

## Status
 

- What is the current behavior?

When "breed" = "JE", there is no change to milk output when changes are made to the second set of brackets for "wood_l". 

When the parameters in the second set of brackets are unchanged and those in the first brackets are updated, the milk output changes (even while breed is still set to "JE"). 

this indicates that RuFaS is using the inputs in the first set of brackets to determine milk production, even when breed is set to "JE". 

To recreate this: (I did these same steps on both Main and Pilot_testing_Dec2023 and had identical results). 

1- run the default metadata with default_animal.json. Record total herd milk and milk/cow/d

 
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>

<div style='direction:ltr'>

Total   milk, kg_ver_agg | Milk   per cow/d, kg_hor_ver_agg
-- | --
1311464 | 41.38274

</div>

</body>

</html>

(I used the filters:            
```
{
            "name": "Total milk, kg",
            "filters": ["AnimalModuleReporter.report_life_cycle_manager_data.daily_milk_production"],
            "slice_start": -365
           },
           {
            "name": "Milk per cow/d, kg",
            "filters": [
                "AnimalModuleReporter.report_life_cycle_manager_data.daily_milk_production",
                "AnimalModuleReporter.report_daily_animal_population.num_lactating_cows"
            ],
            "horizontal_order": [
                "AnimalModuleReporter.report_life_cycle_manager_data.daily_milk_production",
                "AnimalModuleReporter.report_daily_animal_population.num_lactating_cows"
            ],
            "horizontal_first": "true",
            "slice_start": -365
           },
```

2. update "breed" to "JE", with no other changes. Record the total + per cow milk production. 

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>

<div style='direction:ltr'>

Total   milk, kg_ver_agg | Milk   per cow/d, kg_hor_ver_agg
-- | --
1309829 | 41.33305

</div>

</body>

</html>

3. With Breed still set to "JE", set the second set of parameters for "wood_l" to 0's 

image

Record total milk and milk/cow/d - again unchanged

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>

<div style='direction:ltr'>

Total   milk, kg_ver_agg | Milk   per cow/d, kg_hor_ver_agg
-- | --
1304592 | 41.18463

</div>

</body>

</html>

4. Set the second set of parameters back to defaults, and update the first set of parameters to 0's

image

Record total milk and milk/cow/d: 

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>

<div style='direction:ltr'>

Total   milk, kg_ver_agg | Milk   per cow/d, kg_hor_ver_agg
-- | --
15843.85 | 0.494662

</div>

</body>

</html>

-->

## Intention

It is expected that milk for Jersey herds will change when parameters in the second set of brackets change. 

This is only necessary because it is user expectation. I'm not actually sure why they need separate spots for inputs, unless RuFaS uses different equations for the different breeds but I'm not aware of that. 

- Why is this change necessary?

It is very frustrating when there is a disconnect between the inputs I think I am giving RuFaS (2nd set of brackets) and the effect on the outputs (i.e. RuFAS is ignoring my input and using a different set of inputs).  

You can achieve the needed milk production control by using the first set of brackets even for Jersey inputs, as long as the user knows that is the necessary workaround.

### LINKED ISSUE #1881: [Animal] Implement a module for setting farm-specific lactation curve parameters
Implement a module for setting farm-specific lactation curve parameters.

## Status

The process for setting the lactation curve has already been spelled out (see the documentation [here](https://3.basecamp.com/3486446/buckets/5296287/vaults/7108678280)) and a first iteration of the module have already been implemented in the `farm_specific_lactation_curve` branch.

## Intention

Finish the `LactationCurve` class so that users can have RuFaS determine lactation curve parameters specifically for their farms.

## Request

Finish the `LactationCurve` class and ensure that it is working properly by writing a unit test suite and working with SMEs to verify that milk yields are predicted as expected.

### Target deadline

End of sprint 20 - August 14th, 2024.

harry-samuels:
This pull request implements farm specific lactation curve parameter calculation based on temporal, location, and management data as well as reported milk yield.

### Context

Issue(s) closed by this pull request: closes #1881 

### What

This PR adds the class `LactationCurve`, which adjusts and generates the parameters for Wood's equation to estimate daily milk production. The adjustments to these parameters include temporal, spatial, and management components that requires minimal additional input from users.

### Why

This PR improves the accuracy of milk production estimations.

### How

`LactationCurve` gathers several inputs, including the location, herd composition, and annual milking yields from the config and animal inputs. It also grabs the new lactation inputs dedicated to setting and adjusting the lactation curve parameters.

After gathering all these inputs, `LactationCurve` calculates and adjusts Wood's l, m, and n lactation curve parameters for parities 1, 2, and 3+. When new `Cow`s are initialized, they get their l, m, and n values from `LactationCurve`, which is stored as a class attribute of `AnimalBase`.

### Test plan

An extensive test suite has been written for `LactationCurve` and run to make sure that LactationCurve performs as expected.

Also, the predicted milk yields from RuFaS will be checked against the values that were calculated when the underlying methodology for setting the Lactation Curve parameters was being developed.

Edit 8/8/2024 ~11am EDT: actual results are currently being compiled into a google sheet to be compared against expected results. Actual results will be calculated using relative differences, I (Ed) am pretty sure the expected values were computed with absolute differences so that should be considered when evaluating outputs. Will add a link to the google sheet here when it is complete.

Edit again: [link](https://docs.google.com/spreadsheets/d/1RQLw6wBqVwr-GsXExek1t5MlHtq2ipen25lM0M_YOFg/edit?usp=sharing) to the Google sheet with expected vs. actual results. During testing, there were a number of parity 1, 2 and 3+ fractions that had to be rounded in order to make sum to exactly 1.0. This seems like it would likely be the source of differences between expected and actual results if any. The exception to this is farm 9, there seems to be something off for the results of parities 2 and 3.

Edit again again: @YijingGong and I (Ed) found the the issue with farm 9: the problem was that the range that parameter l was fitted in was too restrictive ([-10, 10)). We bumped this up to a range of [-20, 20] and this did the trick.

### Open Questions
- Should there be different standard deviations for l, m, and n by parity? There were previously in the inputs, I am not sure if they should be carried over. Edit: the answer is yes, and these have been implemented.

### Next Steps
There are a number of next steps that can be taken, however given the size this PR has already reached I think it would be better to save these for a follow up issue/PR.

Several parts of `Cow` are either redundant or no longer necessary.
- The breed index and parity index are no longer needed - they should be removed.
- `calculate_daily_milk_produced` should use the method in `LactationCurve`, instead of reimplementing Wood's equation.
- All uses of `determine_param_value` should be replaced with calls to the new method in `Utility`.

The method `scipy.integrate.quad` is used in `calc_305_day_milk_yield`, and this method is slow (noticeably so - I can see the test cases ticking off one by one). Right now this isn't a problem, because this method is only called a handful of times per simulation. But this may not always be the case, so there should be a faster integration method used. I found that vectorized integration with NumPy was much faster, and provided accuracy comparable to `quad`.

Month adjustments are currently not made in `LactationCurve`, but we have everything in place to make them happen. So that is one option for future expansion.


tomhuhh:
Good work Harry and Catherine! 

I performed some tests by changing the following input parameters in the `default_animal.json` and `default_config.json` files. 

```
“cow_num”, 
“parity_percentage”, 
“annual milk yield”, 
“cow_times_milked_per_day”, 
‘end_date’, 
“FIPS_county_code” 
```
The results are in alignment with the results from the standalone code for farms c2f2 and c2f27.

To get this PR ready for a final review, here are some of my suggestions: 

1. Update all the input animal and metadata files. This PR involves structural changes in the input animal and metadata files. It is essential to make sure those files are updated across all files. 
2. Include the unit testing for the new feature. 
3. I noticed that there are a couple constants that could be moved into the constant file. For example, '305/365' could be named as something like 'assumed_milking_cow_percentage'.  

Keep up the good work! Thanks! 
Haowen

input/data/animal/LACTATION_CURVE_ADJUSTMENT_CONSTANTS.json
```diff
@@ -0,0 +1,108 @@
...
+		"Northeast": [1.04, -1.99, -1.13],
+		"Northern Plains": [-0.26, 0.19, -0.79],
+		"New York": [0.67, -1.21, -0.45],
+		"Pennsylvania": [1.15, -0.96, 0.06],
+		"Southeast": [-2.00, 2.59, 2.60],
+		"Southern Plains": [-0.51, -1.02, -0.93],
+		"West Coast": [1.09, 0.53, 0.52],
+		"Wisconsin": [1.4, -0.83, -0.2],
+		"2x/d": [-0.74, 0.090, 0.15],
+		"3x/d": [0.74, -0.090, -0.15]
```
Isgotkowitz: Do you anticipate this list of parameter adjustments ever changing? If so, it would be good to consider an input structure that would be easier to maintain.

input/metadata/properties/default.json
```diff
@@ -93,6 +93,22 @@
...
+          "description": "Percentage of total herd made up by given parity group, either 1, 2, or 3+",
+          "minimum": 0,
+          "maximum": 1
+        }
+      },
+      "annual_milk_yield_lbs": {
+        "type": "number",
+        "description": "The total milk yield generated by the farm in one year (in lbs). If this information is not available, it can be input as null",
+        "minimum": 0,
+        "default": null
```
Isgotkowitz: Is this determining how much milk will be determined, regardless of how cows are managed?

input/metadata/properties/default.json
```diff
@@ -1521,6 +1498,360 @@
...
+        "description": "adjustment values for each wood coefficient for cows milked thrice a day",
+        "properties": {
+          "type": "number",
+          "description": "wood coefficient adjustment"
+        }
+      }
+    },
+    "fips_region": {
+      "type": "array",
+      "description" : "The corresponding region for each state/location that can be described by the FIPS code. FIPS codes that do not correspond to a region in the dataset are the empty string.",
```
Isgotkowitz: Corresponding to what?

RUFAS/routines/animal/life_cycle/animal_base.py
```diff
@@ -21,6 +25,121 @@ def set_config(config):
         AnimalBase.config["nutrient_standard"] = im.get_data("config.nutrient_standard")
         AnimalBase.config["breed"] = im.get_data("animal.herd_information.breed")
 
+    @staticmethod
```
Isgotkowitz: Instead of adding these new methods into AnimalBase, why not create a new module dedicated to handling the lactation curve logic. Then AnimalBase can just call a top level routine in that module to get the milking parameters.

Also, there are a lot of things missing from these methods. There are no type hints, return types, docstrings or tests. These will need to be added.

input/data/animal/LACTATION_CURVE_ADJUSTMENT_CONSTANTS.json
Isgotkowitz, line 1: These are not constants, these are inputs.

Isgotkowitz (commented):
This is headed in the right direction, but there are a number of things that need to be done before this feature is ready. I suggest breaking it up into smaller PRs so that each piece may be fully implemented in a more digestible manner.

input/data/animal/LACTATION_CURVE_ADJUSTMENT_CONSTANTS.json
JoeWaddell: To echo this, I agree it would be better to further categorize the levels, like was done with the `fips_region`.

E.g. another section for year, another for parity, etc.

RUFAS/routines/animal/life_cycle/animal_base.py
```diff
@@ -21,6 +25,121 @@ def set_config(config):
...
+        adjustment_dict = im.get_data("lactation.adjustment_dict")
+
+        for category in [lactation_group, year, month, region, milking_frequency]:
+            if category:
+                parameter_a += adjustment_dict[category][0]
+                parameter_b += adjustment_dict[category][1] * 1e-2
+                parameter_c += adjustment_dict[category][2] * 1e-4
+
+        if MY_305d == None:
+            MY_305d = AnimalBase.calc_integral_wood_curve(parameter_a, parameter_b, parameter_c)
```
JoeWaddell: This section (where `MY_305d` is `None`) could be its own helper method.

RUFAS/routines/animal/life_cycle/animal_base.py
```diff
@@ -21,6 +25,121 @@ def set_config(config):
...
+            y = AnimalBase.get_y_values_wood_curve(t, parameter_a, parameter_b, parameter_c)
```
JoeWaddell: Is there a future need for `y`?

RUFAS/routines/animal/life_cycle/animal_base.py
```diff
@@ -21,6 +25,121 @@ def set_config(config):
...
+    @staticmethod
+    def calc_integral_wood_curve(parameter_a, parameter_b, parameter_c):
+        result, _ = quad(AnimalBase.get_y_values_wood_curve, 1, 305, args=(parameter_a, parameter_b, parameter_c))
+        return result
+
+    @staticmethod
+    def get_wood_parameters(
+        lactation_group=None, year=None, month=None, region=None, milking_frequency=None, MY_305d=None
+    ):
+        parameter_a = 19.9
```
JoeWaddell: Are the values (e.g. 19.9, 24.7, 33.76) ones that should be put in another file, e.g. in the LACTATION_CURVE json? Or do they come straight from the literature/ongoing work?

RUFAS/routines/animal/life_cycle/animal_base.py
```diff
@@ -21,6 +25,121 @@ def set_config(config):
...
+                parameter_b += adjustment_dict[category][1] * 1e-2
+                parameter_c += adjustment_dict[category][2] * 1e-4
+
+        if MY_305d == None:
+            MY_305d = AnimalBase.calc_integral_wood_curve(parameter_a, parameter_b, parameter_c)
+            y = AnimalBase.get_y_values_wood_curve(t, parameter_a, parameter_b, parameter_c)
+            return parameter_a, parameter_b, parameter_c, MY_305d
+
+        else:
+            min_diff = float("inf")
```
JoeWaddell: This section (where `MY_305d` is not `None`) could also be its own helper method.

RUFAS/routines/animal/life_cycle/animal_base.py
```diff
@@ -21,6 +25,121 @@ def set_config(config):
...
+
+        # Assuming Y = 1632 and Z = 2196 based on the given assumptions
+        P2_MY305_adj = im.get_data("lactation.parity_milk_adjustments.MY305_P2_adjustment")
+        P3_MY305_adj = im.get_data("lactation.parity_milk_adjustments.MY305_P3_adjustment")
+
+        if annual_MY_lbs != None:
+            total_avg_305 = annual_MY_lbs * 305 / (365 * num_milking_cows * 2.205)
+
+            # Extracting percentage distribution for each lactation group
+            percent_P1 = parity_percentages[0] * 100
```
JoeWaddell: This shows as not being used: is this for future reference, or an output to be logged elsewhere?

RUFAS/routines/animal/life_cycle/animal_base.py
```diff
@@ -21,6 +25,121 @@ def set_config(config):
...
+        milking_freq = im.get_data("animal.animal_config.management_decisions.cow_times_milked_per_day")
+        if milking_freq >= 2.5:
+            milking_freq = "3x/d"
+        else:
+            milking_freq = "2x/d"
+
+        # calculate lactation group yield
+
+        # Assuming Y = 1632 and Z = 2196 based on the given assumptions
+        P2_MY305_adj = im.get_data("lactation.parity_milk_adjustments.MY305_P2_adjustment")
```
JoeWaddell: In general I'd recommend more readable names, if possible (e.g. writing out parity and milk year in this variable name).

Same for other variables below, all the P# names

JoeWaddell (commented):
A few comments, please let me know if you need clarification! Looking good, but I have some thoughts on some structural changes and a few smaller items.

input/data/animal/LACTATION_CURVE_ADJUSTMENT_CONSTANTS.json
YijingGong: > Do you anticipate this list of parameter adjustments ever changing? If so, it would be good to consider an input structure that would be easier to maintain.

It does not change, they are values from a published paper (https://www.sciencedirect.com/science/article/pii/S0022030222004362). Unless we, as developers, want to change it, for example, to add years after 2016, etc. <- but for this purpose, I remember we have some plan for this extrapolation, not sure if it is in this PR though.

input/metadata/properties/default.json
YijingGong: > Is this determining how much milk will be determined, regardless of how cows are managed?

It is the observed annual milk yield (whatever reported in the Excel sheet when the farm data is collected)

RUFAS/routines/animal/life_cycle/animal_base.py
YijingGong: Agree. Better to put in a constant file or at least need some reference

RUFAS/routines/animal/life_cycle/animal_base.py
```diff
@@ -23,6 +29,154 @@ def set_config(config):
...
+                    MY_305d_best_estimate = MY_305d_vary
+            return parameter_a_best_estimate, parameter_b, parameter_c, MY_305d_best_estimate
+
+    @staticmethod
+    def set_lactation_curve_parameters():
+        region_dict = im.get_data("lactation.fips_region")
+
+        year = im.get_data("config.end_date")[:4]
+        if int(year) > 2016:
+            year = "2016"
```
YijingGong: Looks like there's no extrapolation here yet. That's ok. Just make sure to include it in the PR description

RUFAS/routines/animal/life_cycle/animal_base.py
```diff
@@ -23,6 +29,154 @@ def set_config(config):
...
+            FIPS_state_code = int(FIPS_code / 1000)
+        region = None
+        if FIPS_state_code != None:
+            for code_region in region_dict:
+                if code_region["code"] == FIPS_state_code:
+                    region = code_region["region"]
+                    print(str(FIPS_state_code))
+                    print(region)
+                    print(code_region["code"])
+                    print(code_region["region"])
```
YijingGong: Do we want those prints?

RUFAS/routines/animal/life_cycle/animal_base.py
```diff
@@ -23,6 +29,154 @@ def set_config(config):
...
+        )
+        AnimalBase.lactation_parameters[2] = AnimalBase.get_wood_parameters(
+            lactation_group="2", year=year, region=region, milking_frequency=milking_freq, MY_305d=P2_305
+        )
+        AnimalBase.lactation_parameters[3] = AnimalBase.get_wood_parameters(
+            lactation_group="3", year=year, region=region, milking_frequency=milking_freq, MY_305d=P3_305
+        )
+        """ print("Params 1: " + str(AnimalBase.lactation_parameters[1]))
+        print("Params 2: " + str(AnimalBase.lactation_parameters[2]))
+        print("Params 3: " + str(AnimalBase.lactation_parameters[3]))"""
```
YijingGong: Same here. Do we want those prints while running Rufas?

RUFAS/routines/animal/life_cycle/cow.py
```diff
@@ -281,16 +281,23 @@ def set_lactation_curve_params(self) -> None:
         Currently only set up for wood model.
         """
         if self.lactation_curve == "wood":
+
+            parity_key = self.parity_index
+            # this is a temporary fix for the negative parity_index issue
+            #'parity_key' should not be needed if self.parity_index is always an int between 0 and 2
+            if parity_key < 0:
+                parity_key = 0
```
YijingGong: Just out of my curiosity, why can parity be negative?

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,278 @@
...
+                MY_305d=parity2_305,
+            ),
+            self.get_wood_parameters(
+                lactation_group="3",
+                year=self.year,
+                region=self.region,
+                milking_frequency=self.milking_freq,
+                MY_305d=parity3_305,
+            ),
+        )
```
YijingGong: I get confused. Why does this lactation_curve.py have similar codes as what is in the animal_base.py? Do you want to move these here and forget to delete the code in animal_base.py?

input/data/animal/default_animal.json
```diff
@@ -94,12 +99,42 @@
...
+						3.77e-5,
+						5.82e-5,
+						5.54e-5
+					],
+					[
+						6.60e-5,
+						6.69e-5,
+						6.53e-5
+					]
+				]
```
YijingGong: Not sure if we still want to have these STD in the input file. My feeling is that farmers will likely never change these values.

input/metadata/properties/default.json
```diff
@@ -86,6 +86,22 @@
         "default": "HO",
         "pattern": "(HO|JE)",
         "description": "Breed (select one) -- The predominant breed of the herd (Holstein or Jersey)"
+      },
+      "parity_percentages": {
+        "type": "array",
+        "description": "Percentage of total herd represented by parity 1, 2, and 3+ (length 3)",
```
YijingGong: description can be "Percentage of the animal population in parity 1, 2, and 3 and beyond. The array length needs to be 3"

input/metadata/properties/default.json
```diff
@@ -86,6 +86,22 @@
...
+        "properties": {
+          "type": "number",
+          "description": "Percentage of total herd made up by given parity group, either 1, 2, or 3+",
```
YijingGong: description can be "Percentage of the animal population in parity 1, 2, or 3 and beyond"

YijingGong (changes requested):
I checked the code. Comments are left in line. Then I tried to run this branch, trying to see if the Wood's lactation curve parameter values are as expected. I ran across this error: 
Screenshot 2024-07-16 at 3 31 51 PM
I am just using the default inputs.

RUFAS/routines/animal/life_cycle/animal_base.py
```diff
@@ -3,13 +3,19 @@
 from RUFAS.routines.animal.life_cycle.body_weight_history import BodyWeightHistory
 from RUFAS.routines.animal.life_cycle.pen_history import PenHistory
 from RUFAS.input_manager import InputManager
+import numpy as np
+from scipy.integrate import quad
+
+
+im = InputManager()
```
ew3361zh: As a heads up, we are trying to no longer instantiate IM as a global variable - it was causing some issues with instantiating IM before we expected it to. Please see guidance and best practices in the wiki: 

https://github.com/RuminantFarmSystems/MASM/wiki/Input-Manager#getting-data-from-input-managers-data-pool

input/metadata/properties/default.json
yangcatherine: Corresponding to in the M. Li et al. lactation curve paper. Updated this just now.

input/data/animal/LACTATION_CURVE_ADJUSTMENT_CONSTANTS.json
yangcatherine: This was updated and the constants were categorized.

RUFAS/routines/animal/life_cycle/animal_base.py
yangcatherine: These are now included in the lactation curve inputs json file

RUFAS/routines/animal/life_cycle/cow.py
yangcatherine: Line 276 is "self.parity_index = 2 if self.calves - 1 > 2 else self.calves - 1" but self.calves is initialized to 0 so I believe we ran into problems with that. Thank you for raising my attention to this though, I will bring it up at the next meeting.

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,407 @@
...
+            Wood's lactation curve parameter l.
+        m_param: float
+            Wood's lactation curve parameter m.
+        n_param: float
+            Wood's lactation curve parameter n.
+
+        Returns
+        -------
+        numpy.float64
+            Milk yield on the provided day (kg).
```
Isgotkowitz, line 238: Kg needs to be double checked here

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,407 @@
...
+            Wood's lactation curve parameter l.
+        m_param: float
+            Wood's lactation curve parameter m.
+        n_param: float
+            Wood's lactation curve parameter n.
+
+        Returns
+        -------
+        float
+            305 day milk yield for a cow with the given lactation curve (kg).
```
Isgotkowitz, line 264: Kg needs to be double checked here

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,407 @@
...
+
+        Returns
+        -------
+        dict[str, float]
+            Wood's parameters l, m, and n for the specified parity.
+
+        """
+        parity = min(3, parity)
+
+        params = self.parity_to_parameter_mapping[parity]
```
Isgotkowitz, line 288: Add in check to prevent parity 0 cows from crashing the simulation?

input/metadata/properties/default.json
```diff
@@ -86,6 +86,23 @@
         "default": "HO",
         "pattern": "(HO|JE)",
         "description": "Breed (select one) -- The predominant breed of the herd (Holstein or Jersey)"
+      },
+      "parity_percentages": {
+        "type": "array",
```
Isgotkowitz: Change this to be dictionary- also these are fractions, not percentages.

input/metadata/properties/default.json
```diff
@@ -512,45 +529,6 @@
...
-              "properties": {
-                "type": "number",
-                "description": "Parameter n in the Wood’s equation. It describes the rate of decline after the peak",
-                "minimum": 0
-              }
-            }
-          },
           "wood_l_std": {
             "type": "array",
             "description": "Standard deviation of parameter l ",
```
Isgotkowitz: Remove standard dev inputs

input/metadata/properties/default.json
```diff
@@ -1556,6 +1534,438 @@
...
+		  "parameter_m_std_dev": {
+        "type": "number",
+        "description": "The standard deviation for of Wood's parameter m."
+      },
+      "parameter_n_std_dev": {
+        "type": "number",
+        "description": "The standard deviation for of Wood's parameter n."
+      }
+    },
+    "milking_cow_percentage": {
```
Isgotkowitz: This is fraction, not a percentage

RUFAS/routines/animal/life_cycle/lactation_curve.py
Isgotkowitz: After running the simulation a couple time, this has not been an issue. So Ig it would be better to not add a check, and if something goes wrong with it we'll know pretty quickly

Isgotkowitz:
Update: after a discussion with @YijingGong, the logic of `_estimate_305_day_milk_yield_by_parity` has been changed to use relative differences instead of absolute ones (i.e. the 305 day milk yield of P2 cows is 18% higher than the 305 day milk yield of P1 cows instead of the 305 day milk yield of P2 cows being 1632 kg greater than the 305 day milk yield of P1 cows). This required a change to the values used to calculate the 305 day milk yields relative to each other, which has also been made.

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -37,6 +37,15 @@
...
 
 
+"""
+These constants regulate the range and precision used to fit Wood's l lactation parameter to estimated 305 day milk
+yields.
+"""
+UPPER_BOUND = 20
+LOWER_BOUND = -20
+STEP_SIZE = 0.01
+
```
tomhuhh, line 47: Should we add these to a constant sheet?

tomhuhh (approved):

tomhuhh:
This looks great! Thanks Ed! 

I tested by changing milking frequency from 3x to 2x. Results can be found below. 

3x: 

l:  18.14
m:  0.2329
n:  0.0023420000000000003
l:  24.479999999999997
m:  0.2246
n:  0.003539
l:  24.339999999999996
m:  0.2523
n:  0.0039380000000000005

2x: 
l:  16.66
m:  0.2347
n:  0.002372
l:  23.0
m:  0.2264
n:  0.003569
l:  22.86
m:  0.2541
n:  0.003968

Thanks!

RUFAS/routines/animal/life_cycle/lactation_curve.py
Isgotkowitz: Like a documentation sheet, or a constants module in the code?

tomhuhh: I was picturing a constants module in the code but whatever works the best with the rest of the submodule sounds good to me.

RUFAS/general_constants.py
```diff
@@ -7,6 +7,8 @@ class GeneralConstants:
         MM_TO_CM: The conversion factor from millimeters to centimeters.
 
         GRAMS_TO_KG: The conversion factor from grams to kilograms.
+        LBS_TO_KG: The conversion factor from pounds to kilograms.
```
KFosterReed: I don't see where either of these conversions are used, which is great because there shouldn't be any conversions betwen metric and English units in the biophysical model! If that is truly the case, let's delete these from the constants file

RUFAS/routines/animal/life_cycle/lactation_curve.py
KFosterReed: The referenced work was conducted in kg so the parameters drawn from it are also predicting milk yield in kg and I believe this is correct

RUFAS/routines/animal/life_cycle/lactation_curve.py
allisterakun: We have the `animal_constants.py` and  `general_constants.py` holding animal-level or the entire simulation-level constants. From my understanding, if a variable only really applies to a submodule file, we can keep that in the submodule to avoid complicating the constant file.

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+    parity_to_parameter_mapping : dict[int, dict[str, float]]
+        Maps the parity (1, 2, and 3+) to the associated sets of l, m and n parameters.
+    parity_to_std_dev_mapping : dict[int, dict[str, float]]
+        Maps parities (1, 2, and 3+) to the standard devations of Wood's l, m, and n parameters.
+
+    """
+
+    def __init__(self, time: Time) -> None:
+        im = InputManager()
+        self.om = OutputManager()
```
allisterakun, line 81: @matthew7838, please advise if we should initiate a local om instance in each function or if it's fine to have this `self.om` initialized at the class init.

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+
+        lactation_inputs = im.get_data("lactation")
```
allisterakun:
```suggestion
        lactation_inputs: dict[str, any] = im.get_data("lactation")
```

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+        all_year_adjustments = lactation_inputs["adjustments"]["year"]
+        year_adjustments = self._get_year_adjustments(all_year_adjustments, time)
+
+        fips_code = im.get_data("config.FIPS_county_code")
```
allisterakun:
```suggestion
        fips_code: int = im.get_data("config.FIPS_county_code")
```

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+        Maps parities (1, 2, and 3+) to the standard devations of Wood's l, m, and n parameters.
+
+    """
+
+    def __init__(self, time: Time) -> None:
+        im = InputManager()
+        self.om = OutputManager()
+
+        lactation_inputs = im.get_data("lactation")
+        all_year_adjustments = lactation_inputs["adjustments"]["year"]
```
allisterakun:
```suggestion
        all_year_adjustments: dict[str, dict[str, float]] = lactation_inputs["adjustments"]["year"]
```

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+        year_adjustments = self._get_year_adjustments(all_year_adjustments, time)
+
+        fips_code = im.get_data("config.FIPS_county_code")
+        all_region_adjustments = lactation_inputs["adjustments"]["region"]
```
allisterakun:
```suggestion
        all_region_adjustments: dict[str, dict[str, float]] = lactation_inputs["adjustments"]["region"]
```

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+        region_mapping = lactation_inputs["state_to_region_mapping"]
```
allisterakun:
```suggestion
        region_mapping: dict[str, str] = lactation_inputs["state_to_region_mapping"]
```

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+        region_adjustments = self._get_region_adjustments(all_region_adjustments, region_mapping, fips_code)
+
+        animal_inputs = im.get_data("animal")
```
allisterakun:
```suggestion
        animal_inputs: dict[str, Any] = im.get_data("animal")
```

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+        animal_milking_frequency = animal_inputs["animal_config"]["management_decisions"]["cow_times_milked_per_day"]
```
allisterakun:
```suggestion
        animal_milking_frequency: int = animal_inputs["animal_config"]["management_decisions"]["cow_times_milked_per_day"]
```

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+        all_milking_frequency_adjustments = lactation_inputs["adjustments"]["milking_frequency"]
```
allisterakun:
```suggestion
        all_milking_frequency_adjustments: dict[str, dict[str, float]] = lactation_inputs["adjustments"]["milking_frequency"]
```

RUFAS/routines/animal/life_cycle/lactation_curve.py
allisterakun: Is this an int or a float?

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+        milking_frequency_adjustments = self._get_milking_frequency_adjustments(
+            all_milking_frequency_adjustments, animal_milking_frequency
+        )
+
+        parity_adjustments = lactation_inputs["adjustments"]["parity"]
```
allisterakun:
```suggestion
        parity_adjustments: dict[str, dict[str, float]] = lactation_inputs["adjustments"]["parity"]
```

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+
+        base_wood_parameter_l = lactation_inputs["parameter_mean_values"]["parameter_l_mean"]
+        base_wood_parameter_m = lactation_inputs["parameter_mean_values"]["parameter_m_mean"]
+        base_wood_parameter_n = lactation_inputs["parameter_mean_values"]["parameter_n_mean"]
```
allisterakun:
```suggestion
        base_wood_parameter_l: float = lactation_inputs["parameter_mean_values"]["parameter_l_mean"]
        base_wood_parameter_m: float = lactation_inputs["parameter_mean_values"]["parameter_m_mean"]
        base_wood_parameter_n: float = lactation_inputs["parameter_mean_values"]["parameter_n_mean"]
```

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+            [parity_adjustments["3"], year_adjustments, region_adjustments, milking_frequency_adjustments],
+        )
+
+        self.parity_to_parameter_mapping = {
+            1: self.parity_1_parameters,
+            2: self.parity_2_parameters,
+            3: self.parity_3_parameters,
+        }
+
+        self.parity_to_std_dev_mapping = {
```
allisterakun:
```suggestion
        self.parity_to_std_dev_mapping: dict[int, float] = {
```

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+            1: lactation_inputs["parameter_standard_deviations"]["1"],
+            2: lactation_inputs["parameter_standard_deviations"]["2"],
+            3: lactation_inputs["parameter_standard_deviations"]["3"],
+        }
+
+        annual_milk_yield = animal_inputs["herd_information"]["annual_milk_yield"]
```
allisterakun:
```suggestion
        annual_milk_yield: float = animal_inputs["herd_information"]["annual_milk_yield"]
```

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+            "n": Utility.generate_random_number(params["n"], std_deviations["parameter_n_std_dev"]),
+        }
+
+    def _adjust_lactation_curve_to_milk_yield(
+        self, animal_inputs: dict[str, Any], lactation_curve_inputs: dict[str, dict[str, Any]]
+    ) -> None:
+        """
+        Adjust the lactation parameters using predicted milk yields for the different parities of cows on the farm.
+        """
+        num_milking_cows = animal_inputs["herd_information"]["cow_num"] * lactation_curve_inputs["milking_cow_fraction"]
```
allisterakun:
```suggestion
        num_milking_cows: float = animal_inputs["herd_information"]["cow_num"] * lactation_curve_inputs["milking_cow_fraction"]
```

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+        annual_milk_yield = animal_inputs["herd_information"]["annual_milk_yield"]
+        parity_1_percentage = animal_inputs["herd_information"]["parity_fractions"]["1"]
+        parity_2_percentage = animal_inputs["herd_information"]["parity_fractions"]["2"]
+        parity_3_percentage = animal_inputs["herd_information"]["parity_fractions"]["3"]
```
allisterakun:
```suggestion
        parity_1_percentage: float = animal_inputs["herd_information"]["parity_fractions"]["1"]
        parity_2_percentage: float = animal_inputs["herd_information"]["parity_fractions"]["2"]
        parity_3_percentage: float = animal_inputs["herd_information"]["parity_fractions"]["3"]
```

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+        }
+
+    def _adjust_lactation_curve_to_milk_yield(
+        self, animal_inputs: dict[str, Any], lactation_curve_inputs: dict[str, dict[str, Any]]
+    ) -> None:
+        """
+        Adjust the lactation parameters using predicted milk yields for the different parities of cows on the farm.
+        """
+        num_milking_cows = animal_inputs["herd_information"]["cow_num"] * lactation_curve_inputs["milking_cow_fraction"]
+        annual_milk_yield = animal_inputs["herd_information"]["annual_milk_yield"]
```
allisterakun:
```suggestion
        annual_milk_yield: float = animal_inputs["herd_information"]["annual_milk_yield"]
```

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,425 @@
...
+        parity_1_percentage = animal_inputs["herd_information"]["parity_fractions"]["1"]
+        parity_2_percentage = animal_inputs["herd_information"]["parity_fractions"]["2"]
+        parity_3_percentage = animal_inputs["herd_information"]["parity_fractions"]["3"]
+        parity_2_milk_yield_adjustment = lactation_curve_inputs["parity_milk_yield_adjustments"][
+            "parity_2_305_day_milk_yield_adjustment"
+        ]
+        parity_3_milk_yield_adjustment = lactation_curve_inputs["parity_milk_yield_adjustments"][
+            "parity_3_305_day_milk_yield_adjustment"
+        ]
```
allisterakun:
```suggestion
        parity_2_milk_yield_adjustment: float = lactation_curve_inputs["parity_milk_yield_adjustments"][
            "parity_2_305_day_milk_yield_adjustment"
        ]
        parity_3_milk_yield_adjustment: float = lactation_curve_inputs["parity_milk_yield_adjustments"][
            "parity_3_305_day_milk_yield_adjustment"
        ]
```

allisterakun (approved):
It generally looks good! I left some suggestions for type hints.

KFosterReed (approved):
Great work team! I tested this with different annual milk production values, different herd parity inputs, and different county codes and the milk production changes as expected!

RUFAS/routines/animal/life_cycle/lactation_curve.py
matthew7838: This is perfect. We initialize it every time method for static methods, for non-static we should initialize it in a constructor like the one here.

tomhuhh:
As suggested by Kristan (@KFosterReed), I cross-checked the parameters entered in the LACTATION_CURVE_ADJUSTMENT_INPUTS.json file, and everything aligns with the data reported in Li et al. (2022).

Great job! Thanks!

RUFAS/routines/animal/life_cycle/lactation_curve.py
Isgotkowitz: This is the approach that guided my decision. If we find that any of these values are actually needed in more places in `life_cycle/` or `animal/`, we can always move them into a constants file as needed.

RUFAS/general_constants.py
Isgotkowitz: Nice catch Kristan, these are unused and I will take them out.

RUFAS/routines/animal/life_cycle/lactation_curve.py
Isgotkowitz: Float, but rounded to an int, will add it into the code.

RUFAS/routines/animal/life_cycle/lactation_curve.py
Isgotkowitz: Thanks Kristan

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,431 @@
...
+            Mapping of the parity (1, 2, 3+) to the estimated 305 day milk production of an individual cow of that
+            parity (kg).
+
+        """
+        milk_yield_305_day_all_cows = annual_milk_yield * (305 / GeneralConstants.YEAR_LENGTH)
+        milk_yield_305_day = milk_yield_305_day_all_cows / num_milking_cows
+
+        if not (parity_fractions_sum := sum([parity_1_frac, parity_2_frac, parity_3_frac])) == 1.0:
+            self.om.add_error(
+                f"Fractions of milking cows that are parity 1, 2 and 3+ sum to {parity_fractions_sum}, not 1.0",
```
YijingGong, line 379: Curious about this. I used the Farm 2 (c2f2) in the testing data where the herd parity is ("parity_fractions": {
      "1": 0.269305,
      "2": 0.2386925,
      "3": 0.49121
    },), and it gives me this error `[13-Aug-2024_Tue_12-25-34.866581][ERROR][default] Fractions of milking cows that are parity 1, 2 and 3+ sum to 0.9992075, not 1.0. Using 0.386, 0.281 and 0.333 as the fractions of parity 1, 2 and 3+ cows in the milking herd, respectively`, which is expected based on this code. 
However, your testing results of Farm2 match to my results. I am thinking should we add some "error range" for it, for example, if the difference is within 0.01, it's ok? Or could you help me understand how your testing cases work even if they did not add up to exactly 1.

RUFAS/routines/animal/life_cycle/lactation_curve.py
```diff
@@ -0,0 +1,431 @@
...
+        }
+
+    def _adjust_lactation_curve_to_milk_yield(
+        self, animal_inputs: dict[str, Any], lactation_curve_inputs: dict[str, dict[str, Any]]
+    ) -> None:
+        """
+        Adjust the lactation parameters using predicted milk yields for the different parities of cows on the farm.
+        """
+        num_milking_cows: int = (
+            animal_inputs["herd_information"]["cow_num"] * lactation_curve_inputs["milking_cow_fraction"]
```
YijingGong, line 303: So this says we don't have the ability to input the milking cow number, we always have a fixed proportion of milking cows (0.8356). I guess that's ok in our case where cohort farms did not report the # of milking cows. But again, I'm curious how did you do the testing where I passed you the # of milking cows?

YijingGong (approved):
Great job! I just have two small questions for my own curiousity. Approved it already.

RUFAS/routines/animal/life_cycle/lactation_curve.py
Isgotkowitz: When we ran into this in the tests, we would tweak the parity fractions until they added up to exactly 1.0. Usually this meant tweaking the numbers by a couple thousandths or ten thousandths, and the difference between the actual and expected results seemed small enough (like, a few kg in the 305 day milk yield) that this wasn't a big deal.

Isgotkowitz: An error range could definitely be added in, and 0.01 sounds like a good cutoff for it. Would that be something that needs to go in this PR, or would it be ok for a follow up PR?

RUFAS/routines/animal/life_cycle/lactation_curve.py
Isgotkowitz: For test cases, we calculated the milking cow fraction as the number of milking cows divided by the number of adult cows.

Isgotkowitz: > we always have a fixed proportion of milking cows (0.8356)

This proportion is not fixed, it is an input. So we would change it for each test case. Would it be preferred to have the milking cow # as an input instead?

YijingGong: Ohh perfect! I did not realize "lactation_curve_adjustment_inputs.json" is an input file as well. I thought it is a constant file.

RUFAS/routines/animal/life_cycle/lactation_curve.py
YijingGong: I personally think it would be good to have it in this PR, since it's closely related and it should be fairly straightforward, right?

Isgotkowitz: Yes, I will put it in right now. I will implement it so that if the difference is < 0.01 it will use the numbers as-is. This will be the most straightforward solution, but an alternative would be to round the numbers up so that they are exactly equal to 1.0.

Isgotkowitz: Added

PR #1470, opened 2024-03-28
Add black and flake8 actions to single combined yml gh action file

### LINKED ISSUE #1328: Failing Github Actions Not Preventing Merges
The Black formatting PR merge is overriding failed Github Actions as checks to prevent merging erroring code into main.

## Expected Behavior
The merging of the Black-formatted code into the PR should not override the result of the other Github Actions running.

## Current Behavior
It looks like with the order of operations with the Black formatting Github Action auto-merging into the PR, it overrides any other failed Github Actions.

e.g. this PR has a failing test but because it has 2 approvals it appears as though it will allow merging into main.
Screenshot 2024-02-22 at 4 37 48 PM

## Possible Solution
Trying to change the order of operations to running the other Github Actions (linting/unit-testing) on the Black-formatted merge.

ew3361zh:
The Black formatting and flake8 linting actions have been added to the combined Github Action yml file.

### Context

Issue(s) closed by this pull request: closes #1328

### What

1. The linting and Black formatting actions previously in "lint_flake8.yml" have been moved to the new combined actions file "combined_format_lint_test_mypy.yml".
2. "lint_flake8.yml" has been deleted.
3. The order of actions in the single combined github actions file is split into 2 main jobs (see diagram for additional clarity):
    Job 1, Action 1: check for changes to protected files.
    Job 1, Action 2: format code with Black.
    Job 1, Action 3: commit formatted code to PR.
    
    
    Job 2, Action 1: Fetch most recent commit to branch (i.e. either the Black formatted code newly committed in Job 1 
    or the original commit from the user if no Black formatting occurred).
    Job 2, Action 2: Lint latest commit with flake8.
    Job 2, Action 3: Run Pytest and coverage on latest commit.
    Job 2, Action 4: Run MyPy on latest commit.
    Job 2, Action 5: Update coverage and MyPy badges on ReadMe

4. flake8 is now run on the entire codebase during job 2. The reason is that we've run Black on the entire codebase now so there isn't a large backlog of flake8 violations - there should be none, in fact. All new violations of flake8 should be caught and dealt with (and there likely won't be any since Black will be run on any files changed).

### Why

This new order of Github Actions will ensure that the PR's code is formatted using Black and then automatically merged into the PR which means we won't need to worry about people running Black locally.
These changes also ensure that the other actions are all being run on the Black-formatted code rather than the original unformatted code pushed to the PR. This will help catch any unexpected formatting changes by Black which could affect the code both in how it runs and how it looks. If Black is not run (because no .py files were changed), Job 2 will still run but will run on the original commit from the user.

### How

The newly updated single GH Action yml file contains all the jobs we want run on code in a PR. Splitting the yml file into 2 main jobs - format and everything else - ensures that the code being linted, tested, and MyPy-ed is the Black-formatted code.
Proposed Github Actions Flow Changes

### Test plan

I added intentional flake8 violations to the codebase where Black should catch and fix them. If Job 2 (containing the flake8 action) was running on the original commit, it would flag that error and fail because the original commit has not been fixed by Black yet. If Job 2 is finding the commit from Black, it will not flag any errors because Black will have fixed the error. I ran it with the Black action off to intentionally fail on the flake8 action (which it did), then turned the Black action back on to have it catch and fix that error and then to ensure that Job 2 was finding the Black-formatted commit.


.github/workflows/combined_format_lint_test_mypy.yml
```diff
@@ -28,6 +27,104 @@ jobs:
...
+      - name: Get all added or modified files
+        id: get_files
+        run: |
+          echo "FILES=$( echo ${{ steps.files.outputs.added_modified }} |
+            tr " " "\n" |
+            grep -E "\.py$" |
+            tr "\n" " ")" >> "$GITHUB_ENV"
+          echo "OTHER_FILES=$( echo ${{ steps.files.outputs.added_modified }} |
+            tr " " "\n" |
+            grep -E "\.txt$|\.json$|\.csv$"  |
```
Isgotkowitz:
```suggestion
            grep -E "^(?!.*\.py$).*"  |
```
I think we want to capture everything that is not a .py file, which could include a lot of different file types besides just txt, JSONs and CSVs.

.github/workflows/combined_format_lint_test_mypy.yml
```diff
@@ -28,6 +27,104 @@ jobs:
...
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install coverage mypy
+          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```
Isgotkowitz:
```suggestion
          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi 
```
Isn't the installation of Mypy and Coverage handled by this install?

.github/workflows/combined_format_lint_test_mypy.yml
```diff
@@ -28,6 +27,104 @@ jobs:
           python -m pip install --upgrade pip
           pip install coverage mypy
           if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```
Isgotkowitz: Same comment as below, isn't the installation of Coverage and Mypy handled by the installation of requirements?

Isgotkowitz (commented):
I made some suggestions about possible changes to how requirements are set up and how the list of changed files is handled.

I tested this out and it seemed to correctly catch that there was a flake8 violation, but the way it did it was really weird. It is in PR #1504, and it might be good to talk through what it did because it might indicate a problem.

Also, is it not possible to combine the jobs in such a way that there only needs to be one installation of requirements? Or is it just not possible to have Black run and commit in the same job as the linting and testing?

Overall though, seems to be functioning correctly and come back and approve once we work through that odd behavior.

.github/workflows/combined_format_lint_test_mypy.yml
ew3361zh: Good call.

.github/workflows/combined_format_lint_test_mypy.yml
ew3361zh: dependencies don't persist through github jobs even within the same action. But you're right, we can do away with the independent mypy and coverage install since they're in the requirements.txt file. Good catch!

.github/workflows/combined_format_lint_test_mypy.yml
ew3361zh: Actually we don't have coverage in the requirements. I will add it.

Isgotkowitz (approved):
LGTM, functions as expected. Nice job Niko

.github/workflows/combined_format_lint_test_mypy.yml
```diff
@@ -26,8 +25,104 @@ jobs:
...
+          BRANCH_NAME: ${{ github.head_ref || github.ref }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git config --global user.name 'github-actions[bot]'
+          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
+          git add -A
+          git commit -m "Apply Black Formatting" --allow-empty
+          git remote set-url origin https://${GITHUB_TOKEN}@github.com/${{ github.repository }}.git
+          CLEAN_BRANCH_NAME=${BRANCH_NAME/refs\/heads\//}
+          git push -f origin HEAD:$CLEAN_BRANCH_NAME
```
ltp23: It looks like these two steps can be combined by sharing the conditional.

.github/workflows/combined_format_lint_test_mypy.yml
```diff
@@ -26,8 +25,104 @@ jobs:
...
+          echo "Branch name: $BRANCH_NAME"
+          LATEST_SHA=$(git rev-parse origin/"$BRANCH_NAME")
+          echo "Latest SHA: $LATEST_SHA"
+          echo "LATEST_SHA=$LATEST_SHA" >> $GITHUB_ENV
+          echo "::set-output name=sha::$LATEST_SHA"
+      
+      - name: Checkout the latest commit
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ steps.fetch-latest-sha.outputs.sha }}
```
ltp23, line 107: I think these two steps can be simplified by using the built-in sha variable.

```
- name: Checkout the latest commit
   uses: actions/checkout@v4
   with:
     ref: ${{ github.sha }}
```

ltp23 (approved):
I added two comments for simplifying some steps. Other than those, the PR looks good to me. Thanks Niko!

.github/workflows/combined_format_lint_test_mypy.yml
ew3361zh: For some reason using that wasn't always getting the latest commit and dynamically searching for the latest commit did work.

PR #1472, opened 2024-03-29
Simplify info maps

ltp23:

### What & Why

### How
- [ ] Set the exclude info maps flag at the start of the simulation so that all the calls to `_add_to_pool()` in the OM will ignore the info maps if the flag was set to `True`.

#### On main branch
on-main-branch

#### On the current branch


ltp23:
I created this duplicate PR because I thought there were some weird changes to json file changes.

PR #1475, opened 2024-03-29
No fert update

krb78:
just a quick "hot fix" to update the no_fertilizer input json so it runs and doesn't give errors. 

I had to add the "mix names"


KFosterReed (approved):
Hot Friday Afternoon Fix!

Issue #1463, opened 2024-03-27
Errors occur when new animal population files are created. 


KFosterReed:
We are not able to create a new animal population file directly from initializing a new herd for two reasons:

1. The animal population metadata properties are out of date
2. Once the meta data properties are fixed, then there are occasional runtime errors related to either the ReproStateEnum or the body_weight_history.

## Expected Behavior
We should be able to save the initialized herd population file as described in the RuFaS Command Line Arguments and then use that saved population file as the input to initialize a new herd

## Current Behavior

There are two barriers to this success: 

1. The metadata properties for the repro programs need to be updated
2. There are occasional errors that arise that stop the simulation. These two errors are in the attached files:

[c2f9_jersey_errors_27-Mar-2024_Wed_12-03-59.json](https://github.com/RuminantFarmSystems/MASM/files/14782111/c2f9_jersey_errors_27-Mar-2024_Wed_12-03-59.json)
[c2f9_jersey3_errors_27-Mar-2024_Wed_13-48-29.json](https://github.com/RuminantFarmSystems/MASM/files/14782112/c2f9_jersey3_errors_27-Mar-2024_Wed_13-48-29.json)

## Steps to Reproduce

To replicate the first metadata error
1. Run a simulation and save the animals (--save_animals) and specify a directory that you can use as an input file (--save_animals_dir)
2. Reference that filepath in the animal_population filed in the metadata file

To replicate the second, runtime error
4. On the JerseHerd_hotfix_pilot_testing_Dec2023 branch, run the c2f28_metadata or c2f9_metadata files

## Context (Environment)
This is preventing us from running farms that have Jersey breed animals which is critical for our success in the FARM ES collaboration

I believe there were some changes made in the repro refactor that changed some of the repro metadata inputs and potentially the animal reproduction attributes that are related to this. 

## Detailed Description
When an animal population file is saved, the animals in that population should be able to initialize a new simulation without manually modifying the metadata or the population file.

ew3361zh:
Allister is most familiar with this topic.
Loi has made some initial progress on addressing this issue within PR #1464.

Isgotkowitz:
@Isgotkowitz the "Trying to exit the PREGNANT repro state" error can reproduced in c2f28 with the random seed 2401192515 and looking at cow with id 2150

Current fix idea: when a cow is init'ed, check if it is pregnant. If it is, have the `_repro_state_manager` switch to pregnant state immediately.

[...]

Issue #1052, opened 2023-12-15
Heifer2 repro refactor


allisterakun:
#877 did not fully refactor the reproduction for HeiferII. Only the input data and the corresponding metadata are restructured, but the `HeiferII` class is not updated to reflect the change:
`HeiferII` class is still expecting `'tai_method_h'` and `'synch_ed_method_h'` values (along with other properties) upon initialization (in both `HeiferII.init_values()` and `HeiferII.assign_heiferII_values()` functions, which is being called in `HeiferII.__init()`):
```python
    def init_values(self, args):
        """
        Initialize repro program values
        """
        self.repro_program = args["repro_program"]

        # Estrus variables
        self.estrus_count = 0
        self.estrus_day = 0

        # TAI variables
        self.tai_method_h = args["tai_method_h"]
        self.tai_program_start_day_h = 0

        # synch_ED variables
        self.synch_ed_method_h = args["synch_ed_method_h"]
        self.synch_ed_program_start_day_h = 0
        self.synch_ed_estrus_day = 0
        self.synch_ed_stop_day = 0

        self.conception_rate = 0
        self.ai_day = 0
        self.abortion_day = 0
        self.days_in_preg = 0
        self.preg = False

        self.gestation_length = 0
        self.p_gest_for_calf = 0
        self.calf_birth_weight = 0
        self._hormone_schedule = None
        self._TAI_conception_rate = 0.0

    def assign_heiferII_values(self, args):
        """
        Assign the repro program with given values
        """
        self.repro_program = args["repro_program"]

        # Estrus variables
        self.estrus_count = args["estrus_count"]
        self.estrus_day = args["estrus_day"]

        # TAI variables
        self.tai_method_h = args["tai_method_h"]
        self.tai_program_start_day_h = args["tai_program_start_day_h"]

        # synch_ED variables
        self.synch_ed_method_h = args["synch_ed_method_h"]
        self.synch_ed_program_start_day_h = args["synch_ed_program_start_day_h"]
        self.synch_ed_estrus_day = args["synch_ed_estrus_day"]
        self.synch_ed_stop_day = args["synch_ed_stop_day"]

        self.conception_rate = args["conception_rate"]
        self.ai_day = args["ai_day"]
        self.abortion_day = args["abortion_day"]
        self.days_in_preg = args["days_in_preg"]
        self.gestation_length = args["gestation_length"]
        self.p_gest_for_calf = args["p_gest_for_calf"]
[...]

PR #1477, opened 2024-04-01
[Animal Herd] Changing metadata properties and `Calf` & `HeiferII` methods to make herd generation possible

### LINKED ISSUE #1410: [AnimalModule] Change the `abortion_day` field from None to 0 during initialization of heifer and cow objects
## Urgency Level

- [x] Medium: the issue affects functionality but has a workaround

## Status

Currently, there are a lot of warnings from the input manager caused by the replacement of `None` values in the `abortion_day` field with 0. 

## Intention

Change the default value for `abortion_day` field during the initialization of heifers and cows.

Isgotkowitz:
This PR addresses the first part of issue #1463: incorrect metadata properties for an animal population. It also makes small changes to `Calf` and `HeiferII` to ensure that animal populations generated do not throw warnings.

### Context

Issue(s) closed by this pull request: addresses #1463, #1052. Closes #1410.

### What

The two main issues addressed in this PR are incorrect values/types of values used in the Animal Module, and a mismatch between repro methods properties of "animal_population_properties" and "animal_properties".

### Why

Without properly aligning the animal population properties and animal properties, users will not be able to properly generate and use animal populations.

### How

For fixing the incorrect values and types being used (which either make population generation fail or results in a ton of warnings), the following steps have been taken.
- Cast the result of SciPy's `truncnorm.rvs` as a float (in both `Calf` and `HeiferII`). The result of `truncnorm.rvs` is usually a NumPy float, which the InputManager flags as invalid because it only accepts Python floats and ints for numerical inputs.
- Remove the resetting of `abortion_day` in the `__init__` method of `HeiferII`. HeiferII's init method always calls either `assign_heiferII_values` or `init_values`, both of which set the abortion day. HeiferII resets this value to None though, which isn't necessary.

For fixing the mismatch in the metadata properties, the following steps have been taken:
- The `repro_program` inputs for HeiferIIs and IIIs in the animal population properties now accept "SynchED", the same as `heifer_repro_method` in the animal properties.
- The `repro_program` inputs for Cows and Replacements in the animal population properties now accept "ED-TAI", the same as `cow_repro_method` in the animal properties.
- The `synch_ed_method_h` and `tai_method_h` inputs for HeiferIIs, IIIs, Cows, and Replacements in the `animal_population_properties` section have been removed and replaced with `repro_sub_protocol`. This input matches the `repro_sub_protocol` in `animal_properties`.
- `HerdFactory` sets the TAI method and SynchED methods to "tai_method_h" and "synch_ed_method_h" respectively, instead of setting them both to "repro_sub_protocol".
- The value representation of cows and replacements now includes the "repro_sub_protocol", and is included in the docstring.
- HeiferIIs:
  - `repro_sub_protocol` is now an attribute of HeiferIIs.
  - Instead of having the `tai_method_h` and `synch_ed_method_h` attributes set to values specific to these sub protocols, they are both set to `repro_sub_protocol`. This is a hacky fix, a more permanent one will be to eliminate `tai_method_h` and `synch_ed_method_h` entirely from the code base and only use `repro_sub_protocol`.
  - `repro_sub_protocol` is included in the value representation of HeiferIIs.
- HeiferIIIs init docstring has also been updated.
- Herd Factory and Life Cycle Manager sets the `repro_sub_protocol` attribute of HeiferIIs when transitioning HeiferIs to HeiferIIs.

### Test plan

Unit tests in three files broke. `test_animal.py` was fixed by adding `"repro_sub_protocol"` as a value in a dict used to construct mock cows, and a similar change was made in `test_pen.py`. `test_convert_heiferI_to_heiferII` was updated to check that `repro_sub_protocol` is being passed properly into HeiferIIs when they are created.

I also fixed a number of Mypy errors in `test_animal.py`.

To test and verify these changes, the following steps have been run:
```sh
$ python main.py -v warnings -I -s -S input/data/animal/
```
This correctly generates, writes, and uses an animal population.
After this has run, the newly generated animal population file is specified in the metadata and the following command is run:
```sh
$ python main.py -v warnings
```
Neither of these commands should log errors or warnings related to the animal population generation or the animal population files.

I strongly encourage generating a Jersey herd, and observe results with different combinations of Jersey/Holstein populations and Jersey/Holstein animal inputs to see if RuFaS is modeling the breeds correctly. Also, I encourage mixing up the repro programs used for both cows and heifers.


input/metadata/properties/default.json
```diff
@@ -297,10 +297,15 @@
               "minimum": 0,
               "maximum": 1
             },
-            "repro_sub_protocol": {
+            "tai_method_h": {
               "type": "string",
-              "pattern": "^(5dCG2P|5dCGP|2P|CP|N/A)$",
-              "default": "5dCG2P"
+              "description": "Heifer Reproduction Program",
+              "pattern": "^(md5CG2P|md5CGP)$"
```
allisterakun: Can we please confirm that the sub-TAI method is indeed [md5CG2P, md5CGP]? B/c I think this was updated to [5dCGP, 5dCG2P] in animal refactor PR by Loi, and last week, Haowen said it was [5dCGP, 5dCG2P].

Also, I think we would have an N/A or None option for these sub-methods according to the SMEs, but please confirm this also

allisterakun (approved):
Looks good, just one comment that needs to be confirmed!

input/metadata/properties/default.json
Isgotkowitz: > Can we please confirm that the sub-TAI method is indeed [md5CG2P, md5CGP]? B/c I think this was updated to [5dCGP, 5dCG2P] in animal refactor PR by Loi, and last week, Haowen said it was [5dCGP, 5dCG2P].

After last night's meeting, this is confirmed - it is 5dCGP and 5dCG2P

input/data/animal/default_animal.json
```diff
@@ -48,7 +48,8 @@
         "heifers": {
             "estrus_detection_rate": 0.9,
             "estrus_conception_rate": 0.6,
-            "repro_sub_protocol": "5dCG2P",
+            "tai_method_h": "md5CG2P",
+            "synch_ed_method_h": "2P",
```
ltp23: This would cause an error when you set the `heifer_repro_method` property (same file, in the `management_decisions` section) to `TAI` or `SynchED`. Internally, the code is looking for the `repro_sub_protocol` property. 

During the repro refactoring, I aimed to avoid using protocol-specific properties because they didn't always apply, so I used a more generic name like `repro_sub_protocol`. My suggestion is to remove these protocol-specific properties and it should be safe to do so.

input/metadata/properties/default.json
jadamchick: It would be nice to have a "N/A" or "none" option (I think "none" or "None" would align with the cow presynch/ovsynch/resynch options), so it's really obvious when you have the inputs set up for ED vs. TAI vs. SynchED.

input/data/animal/default_animal.json
```diff
@@ -48,7 +48,8 @@
...
             "repro_sub_properties": {
                 "conception_rate": 0.6,
                 "estrus_detection_rate": 0.9
```
jadamchick: How is it going to handle the sub_properties conception & detection rates now? (when will it choose sub_properties rates instead of the ones listed above)? 
My vote would be to only have a single set of estrus detection & conception rates instead of 2, since we can only choose one repro method (there isn't a need in the heifer set-up to have two different sets of parameters).

input/metadata/properties/default.json
ltp23: This is tied to my previous comment about having a more generic property name, which would simplify the code and the user input. 

The downside of that approach is that the input validation currently still lacks the ability to check for compatibility between the main protocol and the subprotocol. I think the next step forward should be planning for how to do cross-validation. In the meantime, we have to make the assumption that the user reads the metadata description, and chooses the right options.

input/data/animal/default_animal.json
Isgotkowitz: I will look into why there are two separate inputs here.

input/metadata/properties/default.json
Isgotkowitz: > The downside of that approach is that the input validation currently still lacks the ability to check for compatibility between the main protocol and the subprotocol. I think the next step forward should be planning for how to do cross-validation. In the meantime, we have to make the assumption that the user reads the metadata description, and chooses the right options.

I agree that cross-validation is the way to check for incorrect input combinations. However not currently having it does not mean we "have" to assume that users will input the data correctly - there should still be mechanism to check for and reject invalid combinations.

input/data/animal/default_animal.json
Isgotkowitz: So instead of expanding `repro_sub_protocol` into `tai_method_h` and `synch_ed_method_h` in `animal_properties`, `synch_ed_method_h` and `tai_method_h` should be collapsed into `repro_sub_protocol` for HeiferIIs, IIIs, Cows and Replacements in `animal_population_properties` - if I am understanding correctly?

jadamchick (commented):
I'll wait to give approval until all of the repro stuff is sorted out, but here are some initial comments: 

I used the default_animal.json to generate one population exactly as-is, and one population with no changes except for breed = JE. 

Then I ran 4 simulations for each of the following (with set seed = false, 6-yr simulations). 

- Animal_population: generated Holstein; animal.json breed: Holstein
- Animal_population: generated Holstein; animal.json breed: Jersey
- Animal_population: generated Jersey; animal.json breed: Jersey 

 The only attributes I expect RuFaS to treat differently based on breed are a) milk production and b) calf birth weight. 
 
Results: average values for the last year of the simulation 
 
image



A few observations: 

1. It ran smoothly with no errors when using the generated animal populations, even running them a dozen or more times.  
2. There are clear differences between the Jersey and Holstein herds in appropriate variables; there were no differences in inputs that RuFaS considers breed-agnostic and I did not vary between scenarios (eg mature body weight). 
3. Something about the animal population is still impacting the herd even 6 years into the simulation. At 40-50% herd turnover, I expect very few if any animals to remain in the herd during this last year who are from the original (day-1) population. However, RuFaS is still buying 10-20 heifers/year (despite also selling 10-20 heifers/yr lol), so if those replacement animals are drawn from this animal_population, then probably 10-20% (or more) of the cow herd is reflecting the attributes of that population rather than the attributes of the input animal.json. 
(This isn't a problem. But it's something really important to understand so we can work with/around it appropriately (and yet one more reason to prioritize getting those culling updates implemented!).)

input/data/animal/default_animal.json
krb78: Yes, that's how I understand it, Ed.

ltp23:
I was able to test out the animal population generation on the default scenario and used the generated file to run the simulation. This ran without errors and the herd structure looked normal/good to me. Next, I still need to see how it runs on the two farms that Kristan mentions in the issue. 

Using the default scenario as an example, one thing that is still puzzling to me and beyond the scope of this PR is the significant drop in the animal number from 10,000 to around 5,000 after the herd initialization ran for 5,000 days. The effect of this is somewhat masked by the use of sampling with replacement.

Isgotkowitz:
@jadamchick I reran the simulation with a mix of breeds in the animal input and in the herd population, and got the following results:

|Population|Breed|Milk/cow|Calf birth weight|
|-|-|-|-|
|HO|HO|40.6|43.9|
|HO|JE|32.8|33.8|
|JE|JE|27.6|27.1|
|JE|HO|36.5|37.9|

Which seems to indicate that the model is handling different breeds properly.

jadamchick (commented):
I think this does what it sets out to achieve: I am able to generate a new herd from a specific set of pilot farm inputs and then use that herd as the animal_population for future runs without encountering any errors (yet). 
I tried each of the heifer repro methods (ED, TAI, SynchED) and it runs smoothly without errors for each of them. 

My only hesitation is that I don't feel able to evaluate whether those heifer repro methods and/or their input parameters are operating as expected. I tried using a couple report filters to look at conception rate by method and heifers being culled for repro, under varying repro methods and specifications (sometimes putting in 0's for estrus_detection_rate and estrus_conception_rate, or for the subprotocol detection_rate and conception_rate). There are consistently higher repro culls when 0's are put in for the estrus_detection_rate and estrus_conception_rate (even for TAI and SynchED); and there are CR outputs reported for 2 or 3 methods within a given simulation (even though only 1 can be specified). 

I would guess that these issues are outside of the scope of this PR, but also don't feel 100% comfortable saying that the updates you've made to repro inputs are working as expected, because I'm not sure what "baseline/normal" should look like at this point. 

My filters: 
```
        {
            "name": "CR by method",
            "filters": ["AnimalManager._record_heiferIIs_conception_rate.heiferII_.*_conception_rate"],
            "slice_start": -365
        },
        {
            "name": "Heifers sold repro culls",
            "filters": [
                "AnimalModuleReporter.report_life_cycle_manager_data.sold_heiferII_num"
            ],
            "slice_start": -365
        },

```

Let me know your thoughts / happy to discuss more about this. Thanks!

Isgotkowitz:
> I think this does what it sets out to achieve: I am able to generate a new herd from a specific set of pilot farm inputs and then use that herd as the animal_population for future runs without encountering any errors (yet). I tried each of the heifer repro methods (ED, TAI, SynchED) and it runs smoothly without errors for each of them.
> 
> My only hesitation is that I don't feel able to evaluate whether those heifer repro methods and/or their input parameters are operating as expected. I tried using a couple report filters to look at conception rate by method and heifers being culled for repro, under varying repro methods and specifications (sometimes putting in 0's for estrus_detection_rate and estrus_conception_rate, or for the subprotocol detection_rate and conception_rate). There are consistently higher repro culls when 0's are put in for the estrus_detection_rate and estrus_conception_rate (even for TAI and SynchED); and there are CR outputs reported for 2 or 3 methods within a given simulation (even though only 1 can be specified).
> 
> I would guess that these issues are outside of the scope of this PR, but also don't feel 100% comfortable saying that the updates you've made to repro inputs are working as expected, because I'm not sure what "baseline/normal" should look like at this point.
> 
> My filters:
> 
> ```
>         {
>             "name": "CR by method",
>             "filters": ["AnimalManager._record_heiferIIs_conception_rate.heiferII_.*_conception_rate"],
>             "vertical_aggregation": "average",
>             "slice_start": -365
>         },
>         {
>             "name": "Heifers sold repro culls",
>             "filters": [
>                 "AnimalModuleReporter.report_life_cycle_manager_data.sold_heiferII_num"
>             ],
>             "vertical_aggregation": "sum",
>             "slice_start": -365
>         },
> ```
> 
> Let me know your thoughts / happy to discuss more about this. Thanks!

Thank you Julie, I appreciate your thoroughness in this review. I am not sure how to go about evaluating the outputs when changing the repro methods either, but a first way to establish a baseline would be to compare the directionality of output changes in the default simulation compared to the one here. If this branch seems to be matching main, then I think it should be merged.

After that, I think bringing in @KFosterReed, @YijingGong and @krb78 to help evaluate the current behavior would be good. Also bringing in @allisterakun to help plan out changes, if any, that should be made to the way repro is handled in both the code and the inputs.

Isgotkowitz:
@jadamchick I ran the default herd with each repro program using the pilot testing protocol (4 sims, randomly seeded, results averaged) on both main and this branch. Here are the results, using the filter you provided.

Main:
|Repro Program|Overall CR|ED CR|TAI CR|SynchED CR|Culls|
|-|-|-|-|-|-|
|ED|0.602|0.602|0.0|0.0|5.25|
|TAI|0.614|0.613|0.614|0.0|6|
|SynchED|0.371|0.592|0.0|0.0|7|

herd-fix:
|Repro Program|Overall CR|ED CR|TAI CR|SynchED CR|Culls|
|-|-|-|-|-|-|
|ED|0.612|0.612|0.0|0.0|7.25|
|TAI|0.588|0.589|0.587|0.0|4.5|
|SynchED|0.384|0.623|0.0|0.0|7|

For the conception rates, it does seem that directionality is only correct for the Overall CR in SynchED. But as for the others, I can't gauge how significant the differences are. 

It is far outside my range of expertise, but the culling results for all scenarios seemed to have relatively high levels of variability so I am not very confident in the averages here.

jadamchick:
Hi Ed! I think from those results it is safe to say that herd-fix and main are behaving similarly (I would consider everything to be within the same ballpark given expected variability).
I don't think that Synch-ED is working correctly on either (why is CR so much lower, and why is there no CR reported for SynchED CR even when SynchED method is specified) -- though that could be if you didn't change the repro sub-protocol method to one of the SynchED methods (2P or CP). 

I was still wondering which protocols are reading/responding to which sets of input conception_rates (the main/ED ones vs. the subprotocol one). I ran a set of tests to look at that comparing between herd-fix and main, just to make sure they are doing the same thing whatever that may be. I'd say the answer is yes, they are behaving the same -- so, approved! 

image

Some observations that I'll write up in its own issue- 

- ED: when entering inputs for subprotocol only -- the overall and ED CR's reported are untouched, but a boatload of heifers are sold for repro culls (i.e. fail to get pregnant)
- TAI: when entering inputs for only subprotocol or only ED -- the TAI CR is half to 2/3 of what is expected. Only when CR's are entered for subprotocol only, there are a bunch of repro culls (but not as many as in the ED scenario)
- SynchED: for all three inputs scenarios, TAI CR is lower than input. Only in the "subprotocol only" scenario, many many repro culls.

jadamchick (approved):

PR #1491, opened 2024-04-02
Enable json filter files with .json extension

### LINKED ISSUE #1432: [OM] Enable json filter files with .json extension
This issue is the first part of #1416 
- [x] Change the output json key from `detailed_data_origins` to `detailed_values`
- [x] Currently the way to produce a json output is to make a filter file in the `.txt` format. We probably want to keep this option to minimize disruption and allow the `.json` format as well. With the filter file in the `.json` format, it is easier to make customization.

ltp23:
### Context

Issue(s) closed by this pull request: closes #1432 

### What & Why
This PR is part 1 of a series of PRs aiming to improve the handling of data origins info. To allow more customizations, first, we want to be able to incorporate user settings. Currently, this is mainly targeted towards json filter files.

### How
Before, to generate json output, we would use a filter file with this name format `json_<example_name>.txt`. Now, in addition to this, the output manager also accepts the extension `.json`. For example, `json_<example_name>.json` should also work.

I also added a `detailed_values` in the config files flag to turn on/off the detailed_values property.

### Test plan

For example, I would create a filter file with the name `json_example.json`. The following two formats are equivalent and acceptable. 

The first format is simpler and makes sure that there is only one `filters` key.

The second format is more familiar as it is consistent with how to specify graphs and reports but there is no enforcing of only one object inside the array.  

```
{
  "filters": [
    ".*"
  ],
  "name": "All Variables"
}
```

```
{
  "multiple": [
    {
      "filters": [
        ".*"
      ],
      "name": "All Variables"
    }
  ]
}
```


RUFAS/output_manager.py
```diff
@@ -105,6 +105,7 @@ def __init__(self) -> None:
             self.warnings_pool: Dict[str, OutputManager.pool_element_type] = {}
             self.errors_pool: Dict[str, OutputManager.pool_element_type] = {}
             self.logs_pool: Dict[str, OutputManager.pool_element_type] = {}
+            self._detailed_values_flag: bool = False
```
Isgotkowitz: Add to attributes section of class docstring

RUFAS/output_manager.py
```diff
@@ -421,32 +422,33 @@ def dict_to_file_json(self, data_dict: Dict[str, Any], path: str, minify_output_
         self.add_log("save_dict_file_try", f"Attempting to save to {path}.", info_map)
         try:
             with open(path, "w") as json_file:
-                data_dict = self._add_detailed_data_origin(data_dict)
+                data_dict = self._add_detailed_values(data_dict)
+                max_depth = 4
```
Isgotkowitz: Instead of defining this in the method, could it be defined at the top of the file as a constant, with something explaining what it's doing.

Isgotkowitz (approved):
Two suggestions on minor things, but functions just as expected. Nice job Loi!

RUFAS/output_manager.py
ltp23: I can add a class constant called `JSON_OUTPUT_MAX_RECURSIVE_DEPTH` for this.

RUFAS/output_manager.py
```diff
@@ -88,10 +95,13 @@ class OutputManager(object):
         Contains errors reported to the output manager
     logs_pool : Dict[str, Dict[str, List[Dict[str, Any]]]
         Contains logs reported to the output manager
+    _detailed_values_flag : bool
```
```suggestion
    _include_detailed_values : bool
```

RUFAS/output_manager.py
```diff
@@ -88,10 +95,13 @@ class OutputManager(object):
...
+        Flag to include detailed values in the json output files
```
PooyaHekmati: missing verb in `Flag to include detailed values in the json output files` 
if this flag is true, what does it mean? are details already included or are you going to include them later?

RUFAS/output_manager.py
```diff
@@ -488,15 +500,12 @@ def _add_detailed_data_origin(self, data_dict: Dict[str, Any]) -> Dict[str, Any]
         ... ]
         """
 
-        for key in data_dict:
-            if not isinstance(data_dict[key], dict):
-                continue
+        if not self._detailed_values_flag:
+            return data_dict
 
+        for key in data_dict:
```
```suggestion
        for key, sub_data_dict in data_dict.values():
```

RUFAS/output_manager.py
```diff
@@ -488,15 +500,12 @@ def _add_detailed_data_origin(self, data_dict: Dict[str, Any]) -> Dict[str, Any]
...
             sub_data_dict = data_dict[key]
```
```suggestion
```

RUFAS/output_manager.py
```diff
@@ -508,19 +517,42 @@ def _add_detailed_data_origin(self, data_dict: Dict[str, Any]) -> Dict[str, Any]
...
+                detailed_values.append(detailed_origin_for_value)
 
-            sub_data_dict["detailed_data_origins"] = detailed_data_origins
+            sub_data_dict["detailed_values"] = detailed_values
 
         return data_dict
 
+    def _can_add_detailed_values(self, sub_data_dict: Dict[str, Any]) -> bool:
+        """
+        Checks if the provided sub_data_dict is suitable for adding detailed values.
```
```suggestion
        Checks if the provided sub_data_dict is compatible for adding detailed values.
```
?

RUFAS/output_manager.py
```diff
@@ -508,19 +517,42 @@ def _add_detailed_data_origin(self, data_dict: Dict[str, Any]) -> Dict[str, Any]
...
+
+        Parameters
+        ----------
+        sub_data_dict : Dict[str, Any]
+            The dictionary to check for suitability.
```
```suggestion
            The dictionary to check for compatiblity.
```

RUFAS/output_manager.py
```diff
@@ -620,11 +652,21 @@ def _list_to_file_txt(self, data_list: List[str], path: str) -> None:
...
         Returns a file name using the given base_name and timestamp.
+
+        Parameters
+        ----------
+        base_name : str
+            The base name of the file.
+        extension : str
+            The extension of the file.
+        include_millis : bool
+            Flag to include milliseconds in the timestamp.
```
```suggestion
```
I don't think this is necessary.

RUFAS/output_manager.py
```diff
@@ -958,6 +1002,42 @@ def _route_save_functions(
...
+            The name of the filter file being processed.
+        save_path : Path
+            The directory path where the JSON file will be saved.
+        filtered_pool : Dict[str, pool_element_type]
+            The pool of filtered data to be saved.
+        filter_content : Dict[str, Union[str, int]]
+            Additional content from the filter that might influence the file naming.
+        """
+
+        if filter_file.endswith(".json") and "name" not in filter_content:
```

RUFAS/output_manager.py
```diff
@@ -1293,3 +1373,15 @@ def print_errors_warnings_logs_counts(self) -> None:
...
+        """
+        Sets the flag for adding detailed values to the output files.
+
+        Parameters
+        ----------
+        flag : bool
+            The flag to add detailed values to the output files.
+        """
+
+        self._detailed_values_flag = flag
```
```suggestion
    def set_detailed_values_flag(self, flag: bool) -> None:
        """
        Sets the flag for adding detailed values to the output files.
        """

        self._detailed_values_flag = flag
```

PooyaHekmati (changes requested):
1. in json filters all filters are in the `filters` entry. this pr is assuming multiple blobs.
2. the changelog entry is not clear enough
3. the PR is not backward compatible.
4. it is not clear how json filters are handled.

RUFAS/output_manager.py
ltp23: I changed it to this

```
_include_detailed_values : bool
        Set to True to include detailed values in the json output files after the simulation
```

Adding the detailed value property happens after the simulation when we are at the step of making json output files.

RUFAS/output_manager.py
ltp23: Ok. I can rename that.

RUFAS/output_manager.py
ltp23: I changed it to this

```
    def _can_add_detailed_values(self, sub_data_dict: Dict[str, Any]) -> bool:
        """
        Checks if the provided sub_data_dict has the necessary structure and data to add detailed values.

        The sub_data_dict should meet the following requirements:
        - It must be a dictionary.
        - It must contain the keys "info_maps" and "values".
        - The length of the "info_maps" list and the "values" list must be equal.

        Parameters
        ----------
        sub_data_dict : Dict[str, Any]
            The dictionary to check for compatibility with adding detailed values.

        Returns
        -------
        bool
            True if the sub_data_dict meets the requirements for adding detailed values, False otherwise.
        """
```

Isgotkowitz: That looks like a good docstring but I think the "sub_data_dict" part and bullets beneath should go into the `Notes` section.

ltp23:
> 1. in json filters all filters are in the `filters` entry. this pr is assuming multiple blobs.

With the latest updates, the following two formats are equivalent and acceptable. 

The first format is simpler and makes sure that there is only one `filters` key.

The second format is more familiar as it is consistent with how to specify graphs and reports but there is no enforcing of only one object inside the array. If necessary, I can manually enforce it by checking that the array of the `multiple` key has a length of 1.

```
{
  "filters": [
    ".*"
  ],
  "name": "All Variables"
}
```

```
{
  "multiple": [
    {
      "filters": [
        ".*"
      ],
      "name": "All Variables"
    }
  ]
}
```

> 2. the changelog entry is not clear enough

I changed it to this "Enable json filter files with .json extension to allow more customizations later such as formatting detailed values."

> 3. the PR is not backward compatible.

Let me know if there are other things to make it more backward compatible.

> 4. it is not clear how json filters are handled.

I don't think I have made any changes to the way json filters are handled (`dict_to_file_json()` method). This PR is more about preparing user-created filter files and their contents before feeding them into that method.

RUFAS/output_manager.py
ltp23: That was when there were multiple blobs and they had no name fields, so a suffix of milliseconds timestamp would be needed to differentiate between files.

I have simplified it to the following, assuming only one file will be created.

```
        if "name" in filter_content:
            base_name = f"saved_variables_{filter_content['name']}"
        else:
            base_name = f"saved_variables_{filter_file}"

        file_name = self.generate_file_name(base_name, "json")
```

allisterakun (approved):
LGTM! Good job, Loi!

PooyaHekmati (approved):

PR #1496, opened 2024-04-03
Removed exclude-info-maps paramters in main functions

ltp23:
### Context

Issue(s) closed by this pull request: closes #

- [ ] The [changelog](https://docs.google.com/spreadsheets/d/1U-3igxdstHTFb6k5dVcE5-x9IL0r6q18eUiwwytKZyA/edit#gid=0) has been updated.

PR #1497, opened 2024-04-03
Exclude info maps when adding elements to OM pool with GNU arg -i set (reopened)

### LINKED ISSUE #1535: Exclude info maps when adding elements to OM pool with GNU arg -i set
To save some memory at runtime, we can skip storing info maps by setting the gnu -i flag, but the current implementation in the OM still saves those data during the simulation and only drops them at the end if not needed.

- [x] Set the exclude info maps flag at the start of the simulation so that all the calls to `_add_to_pool()` in the OM will ignore the info maps if the flag was set to `True`.
- [x] Check the reloading feature still working as expected.

ltp23:

### What & Why

### How
- [x] Set the exclude info maps flag at the start of the simulation so that all the calls to `_add_to_pool()` in the OM will ignore the info maps if the flag was set to `True`.

#### On main branch
on-main-branch

#### On the current branch

### Test plan

The point of this PR is to improve memory usage internally, so the existing functionalities regarding the outputs should remain the same.

Let say I have a json filter file that has a catch-all regex filter
```
.*
```

To run the program and exclude the info maps data,
```
python main.py -i
```


tomhuhh (approved):
Thank you Loi! I think overall this is a very neat implementation. 

I didn't know how to test this most efficiently. So I decided to increase the number of cows to 5000 in the default file and run a simulation using the command `python main.py -i `. Normally it would take me 40 to 60 minutes to run a herd of this size on my computer but it only took me about 20 minutes on this branch so I'm quite happy about the outcome. 

Thanks, and please let me know if there is a better way to test this improvement in model performance or if you need anything else from me.

main.py
```diff
@@ -129,6 +129,7 @@ def run_rufas(
     """
 
     output_manager = OutputManager()
+    output_manager.set_exclude_info_maps_flag(exclude_info_maps)
```
JoeWaddell, line 132: Now that this method is in place, does `run_load_vars_pool` actually need the `exclude_info_maps` input arg, or could it just check it's own (e.g. `OutputManager` flag that's been set?)

On further review, I see that this question applies to quite a few methods here.

ltp23: Hi Joe, there are several functions in `main.py` that have the `exclude_info_maps` parameter. I did make a PR to remove those (#1496), but to minimize the changes in `main.py`, I put it off at the moment.

jadamchick (approved):
Thanks Loi! I ran this with one of my main reports and compared outputs with and without the -i flag, and they were identical. Looks good to me!

PR #1438, opened 2024-03-21
Adding option to include time in graphs and reports

ltp23:
### Context

Issue(s) closed by this pull request: closes #
May close #912 
Hopefully this PR partially addresses issue #1366 

- [ ] The [changelog](https://docs.google.com/spreadsheets/d/1U-3igxdstHTFb6k5dVcE5-x9IL0r6q18eUiwwytKZyA/edit#gid=0) has been updated.

### What

This PR does two main things:
- Allows time formatting in the x-axis if there is simulation days data associated with the reported variable.
- Allows the addition of an index/simulation day column next to each data column.

### Why

It is helpful at times to have the time context for the report data.

### How

We have been making an implicit assumption about the x-axis to represent simulation days because in the GG we only pass in the data with no information about time.  The next step here would be to create a separate list of day values to be passed along with the data. 

1. In config files, like `default_config.json`, two new properties are added:
    a. `"time_format_string": "%m/%d/%y"` (e.g., 01/01/24). There are other possibilities.
    b. `"add_formatted_time": true`  (Default is `false`): This would add a property in the info maps `"formatted_time": "01/01/24"`
2. In graph filter files, two new options are added:
    a. `"time_step_unit": "day"`. Pick one from `"day", "month", or "year"`.
    b. `"time_step_value": 365`. If `"year"` is selected above, then this property should be left out. 
3. In report filter files, a new option is added: `"indexing": true` (Default is `false` and can be left out).
    - This would add either an index column (0, 1, 2, …) or a date column (09/01/09, 09/02/09, …) next to each data column.
4. The complicating factor in this PR is when the property `"graph_and_report": true` is present in a report filter file. This allows the generation of both graphs and reports at the same time, and the capabilities of both the graph and report generators still need to be preserved.
5. To summarize, anything related to actual dates (like those on the x-axis in graphs and date indexing in the reports), the following are required:
    1. `simulation_day` property in the info maps for the reported variable.
    2. `1b` is `true`, i.e., `"add_formatted_time": true` 
    3. `Item 2`, or `Item 3`, or both.

### Test plan

1. Check that the existing functionalities of the graph and report generators are unaffected.
2. I have tested the following plot functions: `plot`, `scatter`, `stackplot`. I will write a separate issue to do more reading on how to add more options.

For making graphs

```
{
  "multiple": [
    {
      "type": "plot",
      "filters": [
        "AnimalManager._record_animal_counts.num_animals",
        "AnimalManager._record_animal_counts.num_heiferIIs",
[...]

PR #1499, opened 2024-04-03
[1/n] Track simulation day in Time class and set up time in classes

### LINKED ISSUE #1536: [Time] Track simulation day in Time class and set up time in classes
This issue focuses on consolidating the logic of keeping track of the simulation day in the Time class and making the following classes — SimulationEngine, OutputManager, Rreportgenerator, GraphGenerator, have a reference to the Time object for later use. 

- [x] Rename the `self.index` attribute to `self.simulation_day` in the Time class.
- [x] Remove the `self.day_counter` attribute in the SimulationEngine class.
- [x] Make the above-mentioned classes store a reference to the same Time object that was created in the SimulationEngine class.
- [x] Update unit tests.

ltp23:
### Context

### What & Why

This PR aims to reduce the size of a bigger PR, #1438, by focusing on consolidating the logic of keeping track of the simulation day in the Time class and making the following classes — SimulationEngine, OutputManager, Rreportgenerator, GraphGenerator, have a reference to the Time object for later use. 

The overall goal is to incrementally resolve the two issues #912 and #1366.

### How

Main steps
- [x] Rename the `self.index` attribute to `self.simulation_day` in the Time class.
- [x] Remove the `self.day_counter` attribute in the SimulationEngine class.
- [x] Make the above-mentioned classes store a reference to the same Time object that was created in the SimulationEngine class.
- [x] Update unit tests.


RUFAS/output_manager.py
```diff
@@ -1261,3 +1262,15 @@ def print_errors_warnings_logs_counts(self) -> None:
...
+        """
+        Sets the time object for the output manager.
+
+        Parameters
+        ----------
+        time_obj : Time
+            The time object to be set.
+        """
+
+        self.time = time_obj
```
```suggestion
```
a public attribute doesn't need a setter function.

RUFAS/report_generator.py
```diff
@@ -139,12 +139,13 @@ class ReportGenerator:
...
     """
 
-    def __init__(self) -> None:
+    def __init__(self, time=None) -> None:
         """
         Initializes the ReportGenerator.
         """
 
         self.reports: Dict[str, Dict[str, List[Any]]] = {}
+        self.time = time
```
PooyaHekmati, line 155: why is this needed?

RUFAS/graph_generator.py
```diff
@@ -103,8 +103,9 @@ class GraphGenerator:
     NOTE: This class is not multi-thread safe!!!
     """
 
-    def __init__(self, metadata_prefix: str = "") -> None:
+    def __init__(self, metadata_prefix: str = "", time=None) -> None:
         self.metadata_prefix = metadata_prefix
+        self.time = time
```
PooyaHekmati, line 113: why is this needed?

RUFAS/time.py
```diff
@@ -80,24 +83,12 @@ def _calc_sim_length(self) -> int:
...
 
-    def to_str(self):
+    def advance(self) -> None:
         """
-        Description:
-            Returns a string representation of the current time.
-        Returns:
-            str: a String representation of the current time in the simulation
-                in the format "Year: <year> Day: <day>"
+        Advances the time in the simulation by 1 day and automatically detects end of months and years.
```
PooyaHekmati: I don't see any logic for handling month. Also, the year doesn't seem to be incremented.

RUFAS/time.py
```diff
@@ -106,12 +97,13 @@ def advance(self):
...
         """
-        Description:
-            Returns a bool signifying the end of a year.
-        Returns:
-            bool: True if it is the end of a year, False otherwise
+        Returns a bool signifying the end of a year.
+        Returns
+        -------
+        bool
+            True if it is the end of a year, False otherwise
```
PooyaHekmati: vague docstring.
the logic is not clear either.

RUFAS/time.py
```diff
@@ -157,3 +149,40 @@ def is_last_day_of_simulation(self):
...
+
+    def __str__(self) -> str:
+        """
+        Return a string representation of the current time.
+
+        Returns
+        -------
+        str
+            A string representation of a Time object.
+        """
```
PooyaHekmati:
```suggestion
```

RUFAS/time.py
```diff
@@ -157,3 +149,40 @@ def is_last_day_of_simulation(self):
...
+
+        return str(
+            {
+                "simulation_day": self.simulation_day,
+                "simulation_year_day": "Year: {}, Day: {}".format(self.year, self.day),
```
PooyaHekmati: use f-string; `format` is bad.

PooyaHekmati (changes requested):
In addition to in-code comments, `self.calendar_year` and `self.year` attributes in the Time class need better names. Why do we even have two kinds of year?

RUFAS/time.py
ltp23: I will remove the docstring for the str method.

RUFAS/time.py
ltp23: I will use f-string instead.

RUFAS/time.py
ltp23: I will clarify the docstring.

RUFAS/time.py
ltp23: I didn't see any mention of how months should be handled in the time file.

In the `advance()` method, there is a section that increments the year.

```
self.simulation_day += 1

if self.end_year():
    self.day = 1
    self.year += 1
    self.calendar_year += 1
else:
    self.day += 1
```

RUFAS/report_generator.py
ltp23: In later PRs, it is needed to convert simulation days into actual dates in reports.

RUFAS/graph_generator.py
ltp23: In later PRs, it is needed to convert simulation days into actual dates in reports.

RUFAS/output_manager.py
ltp23: Ok. I will remove the setter.

ltp23:
> `self.calendar_year` and `self.year` attributes in the Time class need better names. Why do we even have two kinds of year?

The `self.calendar_year` refers to the real year (e.g., 2009), whereas the `self.year` is the current simulation year (counting from 1). Maybe a better name for the latter would be `simulation_year`?

RUFAS/graph_generator.py
```diff
@@ -103,8 +103,9 @@ class GraphGenerator:
     NOTE: This class is not multi-thread safe!!!
     """
 
-    def __init__(self, metadata_prefix: str = "") -> None:
+    def __init__(self, metadata_prefix: str = "", time=None) -> None:
```
Isgotkowitz, line 111: Can time be imported and type annotated here?

RUFAS/output_manager.py
```diff
@@ -121,6 +121,7 @@ def __init__(self) -> None:
                     "function": self.__init__.__name__,
                 },
             )
+            self.time = None
```
Isgotkowitz, line 140: This attribute needs type hinting and a docstring. Also, does it need a setter?

RUFAS/report_generator.py
```diff
@@ -139,12 +139,13 @@ class ReportGenerator:
     A class to generate reports based on filtered data and aggregation criteria and store them in a dictionary.
     """
 
-    def __init__(self) -> None:
+    def __init__(self, time=None) -> None:
```
Isgotkowitz, line 149: Also needs type hinting and docstring update.

RUFAS/simulation_engine.py
```diff
@@ -88,9 +90,16 @@ def simulate(self) -> None:
...
+            "simulation_day_final_value",
+            str(self.time),
+            {
+                "class": self.__class__.__name__,
+                "function": self.simulate.__name__,
+                "units": {
+                    "simulation_day": MeasurementUnits.UNITLESS.value,
+                    "simulation_year_day": MeasurementUnits.UNITLESS.value,
+                },
+            },
```
Isgotkowitz: Instead of recording this on only one day and doing so outside of the `Time` class, what about adding it to the method `record_time` in the `Time` class?

That change might be a little outside the scope of this PR though.

RUFAS/simulation_engine.py
```diff
@@ -126,7 +134,7 @@ def _advance_time(self, print_day: Optional[bool] = False) -> None:
             "print_day": print_day,
         }
         if print_day:
-            simulating_day_log = f"simulating day: {self.time.to_str()}"
+            simulating_day_log = f"simulating day: {self.time}"
             om.add_log("simulation_day", simulating_day_log, info_map)
```
Isgotkowitz: Maybe it would be better to just remove this recording entirely, and leave all recording of time to the `record_time` method in `Time`? Especially since this recording of time is never actually used.

RUFAS/time.py
Isgotkowitz: Currently months aren't explicitly tracked in `Time` at all - the day field is the ordinal day of the year, so it is agnostic to whatever the month is.

RUFAS/time.py
```diff
@@ -160,3 +159,31 @@ def is_last_day_of_simulation(self):
...
+        Convert the simulation day to a date object that is relative to the start date of the simulation.
+
+        Parameters
+        ----------
+        simulation_day : int
+            The simulation day to convert to a date object.
+        Returns
+        -------
+        datetime.date
+            The date object that corresponds to the simulation day.
```
```suggestion
            The simulation day to convert to a date object.
        
        Returns
        -------
        datetime.date
            The date object that corresponds to the simulation day.

```

Isgotkowitz (commented):
I think this makes some good steps forward in improving the use and structure of `Time` in RuFaS. I left a couple of comments on, mostly minor stuff. I think `SimulationEngine` could be further refactored to put have less time-related functionality, but it may be more appropriate to save those changes for another PR more dedicated to refactoring SimEngine.

Isgotkowitz (approved):

RUFAS/graph_generator.py
ltp23: Due to the circular imports error, I can't use the Time annotation.

RUFAS/output_manager.py
ltp23: I can add a docstring but not a type hint because of circular imports error. It is a public attribute so it can be set directly without a setter.

RUFAS/simulation_engine.py
ltp23: I have added a `simulation_day` variable in the `record_time` method. So, now we have the following four variables

```
Time.day
Time.year
Time.calendar_year
Time.simulation_day
```

RUFAS/simulation_engine.py
ltp23: I have removed the `print_day` block.

allisterakun (approved):
Thank you Loi! Good PR

PooyaHekmati (approved):

PR #1500, opened 2024-04-04
Add Acceptable Measurement Units Enum

### LINKED ISSUE #1288: Create an Enum for acceptable units
Not having units attached to outputs is a problem. Issues #1253 and #1254 have been written to address this. However, when adding units, it would be good to be able to ensure that something is in fact a _valid_ unit.

Create an Enum that contains every single unit used by RuFaS outputs. This will probably need to be a large Enum, and will probably be fluid for some time as units and outputs are added.
Please update all the add_variable calls to include the units using the newly created Enum. Also, update the OM.add_variable to reject the variable if the received variable's unit is not the type of the newly created Enum.

The temporary fix (which is planned to be used for 1253 and 1254) is to use strings to specify what the unit is. The problem with this is that it would be difficult to verify that something is a proper string without have a long list of acceptable strings or a complicated Regex expression or both. By using Enum elements to indicate units, all that is needed to verify that something is a valid unit is to check whether it is a member of the Enum.

**Additional context**

ew3361zh:
This PR creates a new Enum class MeasurementUnits which contains every unit currently used in RuFaS.

### Context

Issue(s) closed by this pull request: closes #1288 

### What

The new Enum class for valid units used in RuFaS is called MeasurementUnits. It contains all current units used in the model.

There is also a new function in OutputManager to ensure that when a variable is added to OM, its `info_map` contains the `units` key and the value for that key is a valid member of the MeasurementUnits Enum.

#### Note:
While the line count is high for this PR, that's almost entirely because of the number of units throughout the model that needed to be changed/updated to incorporate the new units Enum.

#### Reviewers please check:
1. The enum values - these will be the standard units so the way they're written here is important. If you think there's a better/clearer way to write any of these units, please make the suggestion!
2. Spot check the units used. There are a lot but if you could keep an eye out for any units that don't seem appropriate for the variable it's used with, that would be very helpful.

### Why

This will help unify and make consistent all the units used across RuFaS and ensure that variables added to OM have units added.

### How

`OM.add_variable()` now has a check for the `units` key being in the info_map and will raise a KeyError if it's not present. This will ensure no new variables are added without having `units` present.
Once the `units` key is present, its value(s) are checked to ensure they are valid members of the MeasurementUnit Enum. This is done in the new OM function `_validate_units()`. It is a recursive function because there were a number of cases with nested units (mostly within the C&S Module). If that nested units structure is not desired, that should be addressed in a separate PR since it will affect outputs for C&S which is beyond the scope of this PR.

### Test plan

New code was run on all current metadata scenarios and test coverage was updated accordingly for new functionality and the new Enum.


allisterakun (approved):
Great work, Niko! Just wondering if we are still having different units for simulation day/year vs calendar day/year

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -88,31 +89,41 @@ def report_daily_animal_population(cls, animal_manager) -> None:
...
+            dict(info_map, **{"units": MeasurementUnits.ANIMALS}),
         )
         om.add_variable(
             "num_dry_cows",
             len([cow for cow in animal_manager.cows if not cow.is_lactating]),
-            dict(info_map, **{"units": "animals"}),
+            dict(info_map, **{"units": MeasurementUnits.ANIMALS}),
+        )
+        om.add_variable(
+            "num_cows_total", len(animal_manager.cows), dict(info_map, **{"units": MeasurementUnits.ANIMALS})
```
JoeWaddell: here and in a couple others, I'm not sure when we should use `COWS` and when the more general `ANIMALS`.

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -510,140 +521,214 @@ def report_life_cycle_manager_data(cls, life_cycle_manager: LifeCycleManager, si
...
+        )
+        om.add_variable(
+            "sold_heiferII_num",
+            life_cycle_manager.sold_heiferII_num,
+            dict(info_map, **{"units": MeasurementUnits.ANIMALS}),
+        )
+        om.add_variable(
+            "cow_herd_exit_num",
+            life_cycle_manager.cow_herd_exit_num,
+            dict(info_map, **{"units": MeasurementUnits.ANIMALS}),
```
JoeWaddell: Same here and below re use of `COWS` or `ANIMALS`

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -510,140 +521,214 @@ def report_life_cycle_manager_data(cls, life_cycle_manager: LifeCycleManager, si
...
-            "herd_milk_protein_kg", life_cycle_manager.herd_milk_protein_kg, dict(info_map, **{"units": "kg/day"})
+            "herd_milk_protein_kg",
+            life_cycle_manager.herd_milk_protein_kg,
+            dict(info_map, **{"units": MeasurementUnits.KILOGRAMS_PER_DAY}),
         )
         om.add_variable(
             "herd_milk_protein_percent",
             life_cycle_manager.herd_milk_protein_percent,
-            dict(info_map, **{"units": "unitless"}),
+            dict(info_map, **{"units": MeasurementUnits.UNITLESS}),
```
JoeWaddell:
```suggestion
            dict(info_map, **{"units": MeasurementUnits.PERCENT}),
```

RUFAS/routines/field/manager/field_data_reporter.py
```diff
@@ -34,620 +35,800 @@ def send_daily_variables(self) -> None:
...
                     "soil_active_decompose_carbon",
                     layer.soil_active_decompose_carbon,
-                    dict(info_map, **{"units": "kg/ha"}),
+                    dict(info_map, **{"units": MeasurementUnits.KILOGRAMS_PER_HECTARE}),
                 )
                 om.add_variable(
                     "soil_overall_carbon_fraction",
                     layer.soil_overall_carbon_fraction,
-                    dict(info_map, **{"units": "unitless"}),
+                    dict(info_map, **{"units": MeasurementUnits.UNITLESS}),
```
JoeWaddell:
```suggestion
                    dict(info_map, **{"units": MeasurementUnits.FRACTION}),
```

RUFAS/routines/field/manager/field_data_reporter.py
```diff
@@ -34,620 +35,800 @@ def send_daily_variables(self) -> None:
...
-                    "total_nitrogen_uptake", crop.data.total_nitrogen_uptake, dict(info_map, **{"units": "kg/ha"})
+                    "total_nitrogen_uptake",
+                    crop.data.total_nitrogen_uptake,
+                    dict(info_map, **{"units": MeasurementUnits.KILOGRAMS_PER_HECTARE}),
                 )
                 om.add_variable(
                     "optimal_nitrogen_fraction",
                     crop.data.optimal_nitrogen_fraction,
-                    dict(info_map, **{"units": "unitless"}),
+                    dict(info_map, **{"units": MeasurementUnits.UNITLESS}),
```
JoeWaddell:
```suggestion
                    dict(info_map, **{"units": MeasurementUnits.FRACTION}),
```

RUFAS/output_manager.py
```diff
@@ -173,9 +174,40 @@ def add_variable(self, name: str, value: Any, info_map: Dict[str, Any]) -> None:
...
         """
+        units = info_map.get("units")
+        if units is None:
+            raise KeyError("'units' was not found in info_map for call to 'add_variable()'")
+        self._validate_units(units)
+
         key = self._generate_key(name, info_map)
         self._add_to_pool(self.variables_pool, key, value, info_map)
 
+    def _validate_units(self, units: Dict[str, Any] | str):
```
JoeWaddell: needs typing for return

RUFAS/output_manager.py
```diff
@@ -173,9 +174,40 @@ def add_variable(self, name: str, value: Any, info_map: Dict[str, Any]) -> None:
         info_map["suffix"] : str, optional
             If present, gets appended to the key
         """
+        units = info_map.get("units")
+        if units is None:
```
JoeWaddell, line 178: Can/should this be ignored if info_maps is excluded? It might only be necessary when #1497 is merged, since maybe there's not a good way to check yet (1497 adds a nice attribute in `OutputManager` itself).

tests/test_output_manager.py
```diff
@@ -425,25 +426,90 @@ def test_add_log(
...
+        ),
+    ],
+)
+def test_validate_units(
+    mock_output_manager: OutputManager,
+    output_manager_original_method_states: Dict[str, Callable],
+    units: Dict[str, MeasurementUnits | Dict[str, MeasurementUnits]],
+    expected_exception: ValueError,
+    expected_message: str,
+):
```
JoeWaddell:
```suggestion
) -> None:
```

JoeWaddell (approved):
Just a few smallish questions/comments/suggestions. Overall looks good: definitely improves readability, and should help guard against typos creeping in! Good call.

RUFAS/output_manager.py
ew3361zh: Good question - I think it will be fine either way since the call to `_add_to_pool()` is where the info maps are actually not added and that happens after this check. I think this check is important to make sure that every call to `add_variable()` has "units" as a key in the info_map but it shouldn't be concerned with whether or not info_maps are being added to the pool (ideally).

RUFAS/routines/animal/animal_module_reporter.py
ew3361zh: Tagging in @jadamchick for her thoughts.

RUFAS/routines/animal/animal_module_reporter.py
ew3361zh: Tagging in @jadamchick for her thoughts.

PR #1502, opened 2024-04-05
Adding C1f1 crop&soil inputs

krb78:
This PR adds many input files to the pilot testing branch for c1f1. 

It is not closing any issues, but we are looking forward to running the first three farms to compare to the purchased feed outputs. 

I've run this through without simulating animals to make sure that the crops are being planted and there are no errors with the input files.


jadamchick (approved):
hooray! can't wait to see how it performs

Isgotkowitz (approved):
Runs! I noticed there were two RuntimeWarnings related to dividing up an invalid value. If these issues seem to be have an effect on the output then let me know and I will look into them

PR #1505, opened 2024-04-09
[RG] Improve the logic of handling the `horizontal_first` property

### LINKED ISSUE #1498: [RG] Debugging report generator errors related to horizontal aggregation when columns have different lengths
I'd like to collect examples of errors caused by the report generator. It would be very helpful if you could provide the following:
- Filters that are not working
- Current branch and scenario
- Error messages

ltp23:
### Context

Issue(s) closed by this pull request: closes #1498 

### What & Why

The errors encountered in the mentioned issue are primarily caused by performing horizontal aggregation when the columns in the data table have different sizes. This type of error is expected if no padding is applied to enforce consistent column lengths.

However, when both horizontal and vertical aggregation are specified, a good workaround is to perform the vertical aggregation first, followed by the horizontal aggregation. By performing vertical aggregation first, each column is collapsed into a single value, resulting in all columns having a uniform size of one. This approach is effective when the aggregation function is sum. For other functions, a good rule of thumb is to think through the math first.

Currently, in the report generator when double aggregation is used, there is a faulty logic related to the order of doing the aggregations. Horizontal aggregation is always performed first (which can cause errors) and followed by vertical aggregation. Then, depending on the value of `horizontal_first`, the second aggregator function will be applied to the result of the first aggregation. This seems like a good approach at first by preparing for separate aggregations and then using the results for the next step. What we should do instead is to perform horizontal aggregation first only when the `horizontal_first` property is true.

Currently, the report generator's logic for handling double aggregation has a flaw related to the order in which the aggregations are performed. Horizontal aggregation is always executed first (which can lead to errors), followed by vertical aggregation. Then, based on the value of horizontal_first, a vertical aggregator function will be applied to the horizontally aggregated results in the previous step, or vice versa. Because of the potential error caused by horizontal aggregation, what we can do instead is to perform horizontal aggregation first only when the `horizontal_first` property is set to true.

### How

Adjust the order of aggregation based on whether the `horizontal_first` property is true or false. When double aggregation is specified, only perform horizontal aggregation first when `horizontal_first` property is true. Otherwise, do the vertical aggregation first.

### Test plan

To test the changes, I used the following three examples:

- Retrieve the data columns as-is, without any aggregation, to see what the data table looks like.
- Perform horizontal aggregation first and then vertical aggregation. This may or may not work depending on the structure of the data table.
- Improve upon the second example by performing vertical aggregation first, followed by horizontal aggregation.

```
{
  "multiple": [
    {
      "name": "Total housing methane data, kg",
      "filters": [
        "^ManureHandlerDailyOutput_Pen_.*housing_methane"
      ],
      "slice_start": -365
    },
    {
      "name": "Total housing methane horizontal vertical, kg",
      "filters": [
        "^ManureHandlerDailyOutput_Pen_.*housing_methane"
      ],
      "horizontal_aggregation": "sum",
      "horizontal_first": true,
      "slice_start": -365
    },
    {
      "name": "Total housing methane vertical horizontal, kg",
      "filters": [
        "^ManureHandlerDailyOutput_Pen_.*housing_methane"
      ],
      "horizontal_aggregation": "sum",
      "horizontal_first": false,
      "slice_start": -365
    }
  ]
}
```

Updated version of the filter file provided by Julie 

[report_pilots 3Mar.json](https://github.com/RuminantFarmSystems/MASM/files/15015637/report_pilots.3Mar.json)


RUFAS/report_generator.py
```diff
@@ -633,7 +629,7 @@ def _prepare_report_data_with_constants(
 
         try:
             self._add_constants_to_report_data(report_data, filter_content)
-        except ValueError:
+        except ValueError:  # NOT COVERED
```
PooyaHekmati: don't add comments; especially vague comments like `not covered`; what does it even mean?

RUFAS/report_generator.py
```diff
@@ -400,78 +400,74 @@ def _perform_aggregations(
...
-            return {"hor_agg": horizontally_aggregated}
-
-        if vertical_agg_key:
-            if filter_content.get("variables") is not None or len(vertically_aggregated) > 1:
-                return {f"{key}_ver_agg": value for key, value in vertically_aggregated.items()}
-            elif len(vertically_aggregated) == 1:
-                return {"ver_agg": list(vertically_aggregated.values())[0]}
-
-        return None
+        horizontal_first = filter_content.get("horizontal_first", True)
```
PooyaHekmati: shouldn't the default be false?

RUFAS/report_generator.py
```diff
@@ -400,78 +400,74 @@ def _perform_aggregations(
...
+            )
+            aggregate_report = {"hor_agg": horizontally_aggregated}
 
-        aggregate_report = self._combine_aggregate_report_data(
-            horizontally_aggregated, vertically_aggregated, filter_content
-        )
+        elif vertical_agg_key:
+            vertical_aggregator = AGGREGATION_FUNCTIONS[vertical_agg_key]
+            vertically_aggregated = self._apply_vertical_aggregation(aggregate_report, vertical_aggregator)
+            if filter_content.get("variables") is not None or len(vertically_aggregated) > 1:
```
PooyaHekmati: avoid cryptic and complex conditions, follow the clean code practices that we have been trying to do for years.

PooyaHekmati (changes requested):
In addition to in-code comments, there are undocumented changes to animal module.

RUFAS/report_generator.py
Isgotkowitz: Also, "horizontal_first" needs to be checked that it is actually a bool - and if not should log an error or warning. There are a lot of reports, even some examples in the Wiki, which use strings instead of bools.

Isgotkowitz (commented):
Think this PR is headed in the right direction Loi. One thing I noticed while looking at some filters locally and in the the Wiki is that the type of `"horizontal_first"` is not consistent and is not checked. I think that needs to be addressed in the code as well as in the Wiki. Otherwise, I think it will be good after addressing Pooya's comments - please re-request my review after you have addressed them.

RUFAS/report_generator.py
ltp23: I will remove that. I put it there earlier because that line wasn't covered by unit tests, but it is fixed now.

RUFAS/report_generator.py
ltp23: Yes, the default should be False.

ltp23: Hi Ed, I have added a check for it. Actually, it is kind of a subtle bug. The stringified version of true and false both return True in conditionals because they are non-empty strings. With the latest updates, if it is possible to convert to boolean, I would do that and give a warning so it won't disrupt the existing reports filters too much.

RUFAS/report_generator.py
ltp23: I added names to the conditions, maybe it is a bit clearer now.

```
has_dict_variables = filter_content.get("variables") is not None
has_multiple_columns = len(vertically_aggregated) > 1

if has_dict_variables or has_multiple_columns:
    aggregate_report = {f"{key}_ver_agg": value for key, value in vertically_aggregated.items()}
else:
    aggregate_report = {"ver_agg": list(vertically_aggregated.values())[0]}
```

RUFAS/report_generator.py
```diff
@@ -241,6 +243,66 @@ def generate_report(
 
         return event_logs
 
+    def _check_horizontal_first_property(
+        self,
+        filter_content: Dict[str, Any],
+        event_logs: List[Dict[str, str | Dict[str, str]]],
+        individual_report_name: str,
+    ) -> None:
```
Isgotkowitz: A couple suggestions on this method.
1. Have this method extract the value of `horizontal_first` in addition to checking it? This leads into point 2
2. Instead of calling this in `generate_report`, what about replacing the `filter_content.get("horizontal_first", False)` call in `_handle_horizontal_and_vertical_aggregation` with a call to this method?
3. Instead of allowing both strings and bools in the reports, just allow bools. If someone passes in a string as the value for "horizontal_first" log it as an error and make a note that it is defaulting to `False`.

I think these suggestions would help simplify the code and enforce simpler, more standardized writing of reports.

Isgotkowitz (approved):
Left a comment with a couple suggestions regarding the `_check_horizontal_property_method`, please consider them. Otherwise, LGTM and performs aggregation as expected. Nice job Loi

jadamchick (approved):
This seems to work well. I was able to recreate the errors on pilot testing branch (by lowering stall # and stocking density so a new pen would be created even during the last 365d) and then use the same seed on this branch. On this branch, the report generation errors were limited to the test filter I put in (with horizontal_first = true) and to cross references looking for variables whose names had now changed. 

It would be good to give a heads up before merging this into main, since it does change the names of variables affected (so any cross references can't find it now). 
For example: 
old: LAC-COW enteric methane, kg_hor_ver_agg
new: LAC-COW enteric methane, kg_ver_hor_agg

Also a note that the other type of error referenced in #1498 does still occur ~1/4 times (when the conditions are set to make this one also appear). It stops the simulation completely and says .Dumping all logs from main.py because of error 'list index out of range'
I will create a separate issue for that and provide more context for how to replicate it later today.

ltp23:
Thanks Julie (@jadamchick). The default option for the `horizontal_first` property as stated on the wiki has always been `false`. I think in one of the PRs I switched it to true in the code, and now I'm switching it back to false so it can be consistent with what is in the wiki description. This does mean that those filters that have been using the default option will have to change from `_hor_ver_agg` suffix to `_ver_hor_agg`. 

To make the filters more robust, it would be a good idea to explicitly state the true or false value of the `horizontal_first` property. There used to be a feature that allowed regexes when doing cross-referencing (e.g., `LAC-COW enteric methane, kg_.*_agg`) but it is not working now for some reason.

I think if you could make a PR on how to replicate the index-out-of-range error, that would be great. I couldn't reliably reproduce it.

jadamchick:
> 
> I think if you could make a PR on how to replicate the index-out-of-range error, that would be great. I couldn't reliably reproduce it.

yes! the new issue is described (including a specific metadata/seed combo) here: https://github.com/RuminantFarmSystems/MASM/issues/1564

RUFAS/report_generator.py
ltp23: Thanks Ed. 

I have simplified it based on your suggestions: 
- The `horizontal_first` property accepts booleans only.
- Making the method return a boolean.
- Call the method directly in the `_handle_horizontal_and_vertical_aggregations()`.

ltp23:
> In addition to in-code comments, there are undocumented changes to animal module.

Those extra changes are no longer needed because they have been addressed by another PR #1537.

PooyaHekmati (approved):

PR #1518, opened 2024-04-10
[GG] Option to omit prefixes from auto-populated legend keys

### LINKED ISSUE #1338: Remove class and method name from auto-populated legend keys
When using regex style filters in making a graph (see below), the automatically populated legend uses the full key name from output manager. This is likely too verbose in most cases. Adding an option to use just the final part of the key name would be ideal. E.g. instead of using `AnimalModuleReporter.report_daily_pen_totals.number_of_animals_in_pen_3_LAC_COW` in the legend, it would just use just `number_of_animals_in_pen_3_LAC_COW`.

**Example filter**
When using: 
```"filters": [
        "AnimalModuleReporter.report_daily_pen_total.*"
      ],
      "title": "AnimalModuleReporter daily_pens"
```

The resulting figure looks like this: 
image

While legend labels like `number_of_animals_in_pen_3_LAC_COW` would still overlap much of the figure (I saw that in another PR @ew3361zh noted a fix for the overlap, which will be excellent), it would be better than the current behavior.

**Additional considerations**
This behavior is particularly troublesome when dealing with similarly grouped keys (like this example) that are dynamic - in this case, the number of pens is not predetermined from the start (though there is a number of starting pens), so we cannot provide a series of specific keys and legends to plot before running a simulation.

Some option for the automatically generated legend, e.g. verbose (as it currently is), minimal (the suggestion above), or None, may be useful.

allisterakun:
This PR introduces a new `omit_legend_prefix` flag in the graph filter file to give the user the option to omit the prefixes for auto-generated legend keys
### Context

Issue(s) closed by this pull request: closes #1338.

### What

Introduces a new `omit_legend_prefix` flag in the graph filter file to give the user the option to omit the prefixes for auto-generated legend keys. This flag will default to False unless otherwise specified.

### Why

To address the issue of the auto-generated legends being too long.

### How

By giving the user the option to omit the prefixes of the auto-generated legends.

### Test plan

Before
image
After
image
The above graph is generated with the following filter running on the default scenario:
```JSON
{
  "type": "plot",
  "filters": [
    "AnimalModuleReporter.report_daily_pen_total.*"
  ],
  "title": "AnimalModuleReporter daily_pens",
  "omit_legend_prefix": true
}
```


Isgotkowitz (changes requested):
The variable name is not included when there is a suffix - try using this filter:
```json
{
  "multiple": [
    {
      "type": "plot",
      "filters": [
        ".*\\.snow_(melt|content).field='.*'"
      ],
      "variables": ["snow_content", "snow_melt"],
      "title": "Snow Dynamics",
      "omit_legend_prefix": true
    }
  ]
}
```

RUFAS/graph_generator.py
```diff
@@ -174,7 +174,10 @@ def generate_graph(
             self._draw_graph(graph_details["type"], prepared_data, prepared_data.keys())
             legend = graph_details.get("legend")
             if not legend:
-                graph_details["legend"] = list(prepared_data.keys())
+                if graph_details.get("omit_legend_prefix", False):
+                    graph_details["legend"] = list(key.split(".")[-1] for key in prepared_data.keys())
```
JoeWaddell: The only issue here is that if a decimal point is in a variable name, e.g. if there was some hypothetical `growth_above_1.5kg` key. 

It doesn't appear we have any like that currently, but if a variable is introduced in the future, it would break this functionality. 

Could we count from the other direction instead, or are the prefixes more variable/tough to clip off?

allisterakun:
1. len() == 1:
    1. name
2. len() == 2:
    1. prefix.name
3. len() == 3:
    1. Class.Method.Name
    2. prefix.name1.name2
    3. prefix.name.suffix
4. len() == 4:
    1. Class.Method.name.suffix
    2. prefix.name1.name2.name3
    3. prefix.name1.name2.suffix
5. len() == 4+:
    1. Class.Method.name1.name2...suffix
    2. Class.Method.name1.name2.name3…
    3. prefix.name1.name2.name3.name4…
    4. prefix.name1.name2.name3…suffix

```python
        combined_var_name_list = combined_var_name.split('.')

        if len(combined_var_name_list) == 1:
            # no prefix and no suffix
            return combined_var_name_list[0]

        elif len(combined_var_name_list) == 2:
            # prefix.name
            return combined_var_name_list[1]

        elif len(combined_var_name_list) >= 3:
            # has suffix
            if '=' in combined_var_name_list[-1]:
                slice_end = -1
                # class.method.*.suffix
                if re.match('([A-Z][a-z0-9]+)+', combined_var_name_list[0]):
                    slice_start = 2
                    return ".".join(combined_var_name_list[2:-1])
                # prefix.*.suffix
                else:
                    slice_start = 1
                    return ".".join(combined_var_name_list[1:-1])
            # no suffix
            else:
                slice_end = len(combined_var_name_list)
                # class.method.*
                if re.match('([A-Z][a-z0-9]+)+', combined_var_name_list[0]):
                    slice_start = 2
                    return ".".join(combined_var_name_list[2:])
                # prefix.*
                else:
                    slice_start = 1
                    return ".".join(combined_var_name_list[1:])
```
||
v
simplified logic
```python
combined_var_name_list = combined_var_name.split('.')
        if len(combined_var_name_list) == 1:
            # no prefix and no suffix
            return combined_var_name_list[0]
        elif len(combined_var_name_list) == 2:
            # prefix.name
            return combined_var_name_list[1]

        elif len(combined_var_name_list) >= 3:
            # class.method.* or prefix.*
            slice_start = 2 if re.match('([A-Z][a-z0-9]+)+', combined_var_name_list[0]) else 1
            # *.suffix or no suffix
            slice_end = -1 if '=' in combined_var_name_list[-1] else len(combined_var_name_list)

            return ".".join(combined_var_name_list[slice_start:slice_end])
```

RUFAS/graph_generator.py
```diff
@@ -188,6 +197,36 @@ def generate_graph(
 
         return all_logs
 
+    def _generage_legend_keys(self, combined_var_name: str) -> str:
```
```suggestion
    def _generate_legend_keys(self, combined_var_name: str) -> str:
```

RUFAS/graph_generator.py
```diff
@@ -188,6 +197,36 @@ def generate_graph(
...
+
+        Parameters
+        ----------
+        combined_var_name: str
+            The combined variable name to be processed.
+
+        Returns
+        -------
+        str
+            The striped variable name.
```
```suggestion
            The stripped variable name.
```

RUFAS/graph_generator.py
```diff
@@ -188,6 +197,36 @@ def generate_graph(
...
+            return combined_var_name_list[0]
+        elif len(combined_var_name_list) == 2:
+            # prefix.name
+            return combined_var_name_list[1]
+
+        elif len(combined_var_name_list) >= 3:
+            # class.method.* or prefix.*
+            slice_start: int = 2 if re.match("([A-Z][a-z0-9]+)+", combined_var_name_list[0]) else 1
+            # *.suffix or no suffix
+            slice_end: int = -1 if "=" in combined_var_name_list[-1] else len(combined_var_name_list)
```
Isgotkowitz: I think these different cases should be explained in the notes section of the docstring, not with inline comments.

RUFAS/graph_generator.py
```diff
@@ -172,9 +173,17 @@ def generate_graph(
             plt.subplots_adjust(right=ratio_of_graph_to_legend)
             filtered_pool = {k: filtered_pool[k] for k in graph_details["filters"] if k in filtered_pool.keys()}
             self._draw_graph(graph_details["type"], prepared_data, prepared_data.keys())
-            legend = graph_details.get("legend")
-            if not legend:
-                graph_details["legend"] = list(prepared_data.keys())
+
+            if not graph_details.get("legend"):
+                if graph_details.get("omit_legend_prefix", False):
```
Isgotkowitz: I think it would help readability a little if these two conditions were put in there own variables.

Isgotkowitz (approved):
Works on the various filters I tried. Left some on cleaning code up a little and fixing some typos. Also would be good to identify where the new Mypy errors are being introduced and fix them so the number of errors isn't going up.

With respect to the fix, I think it is good for now but relying on string parsing is not a good long term solution. A better one would be to get info maps into the graph generator so that the class name, function name, prefix, suffix, and anything else that may be added gets removed precisely. That may or may not require significant changes to the Graph generator and other parts of the code though.

RUFAS/graph_generator.py
JoeWaddell: Agreed: some explanation of what other edge cases you're handling here (e.g. it would be good to explain why you're searching for `([A-Z][a-z0-9]+)+` and `=`).

tests/test_graph_generator.py
```diff
@@ -216,6 +216,28 @@ def test_generate_graph_exception(graph_generator: GraphGenerator) -> None:
...
+        ("dummy_prefix.dummy_var.dummy_var2.dummy_var3", "dummy_var.dummy_var2.dummy_var3"),
+        ("dummy_prefix.dummy_var.dummy_var2.field='field'", "dummy_var.dummy_var2"),
+        ("DummyClass.dummy_method.dummy_var.dummy_var2.field='field'", "dummy_var.dummy_var2"),
+        ("DummyClass.dummy_method.dummy_var.dummy_var2.dummy_var3", "dummy_var.dummy_var2.dummy_var3"),
+        ("dummy_prefix.dummy_var.dummy_var2.dummy_var3.dummy_var4", "dummy_var.dummy_var2.dummy_var3.dummy_var4"),
+        ("dummy_prefix.dummy_var.dummy_var2.dummy_var3.field='field'", "dummy_var.dummy_var2.dummy_var3"),
+    ],
+)
+def test_generage_legend_keys(combined_var_input: str, expected_output: str, graph_generator: GraphGenerator) -> None:
+    actual_output = graph_generator._generage_legend_keys(combined_var_input)
```
JoeWaddell:
```suggestion
    actual_output = graph_generator._generate_legend_keys(combined_var_input)
```

JoeWaddell (commented):
As a general observation and suggestion here: in reading the code (and testing!), I see the 'field=field' type suffix is being removed in this method.

This is misleading since the filter option only refers to the prefix. I could see renaming the filter option to `omit_legend_prefix_and_suffix`, but you could also separate the two out e.g. `omit_legend_prefix` and a new `omit_legend_suffix`, if it's not a big bother, since maybe there's utility in keeping the suffix (though that's more a question for @Isgotkowitz )

RUFAS/graph_generator.py
```diff
@@ -203,7 +285,9 @@ def _validate_graph_filter(
             The logs, warnings, and errors to be reported to OutputManager.
         """
         required_graph_filter_keys = ["type", "filters"]
-        optional_graph_filter_keys = list(FIGURE_SETTERS.keys()) + list(AXES_SETTERS.keys()) + ["variables"]
+        optional_graph_filter_keys = (
+            list(FIGURE_SETTERS.keys()) + list(AXES_SETTERS.keys()) + ["variables", "omit_legend_prefix"]
```
JoeWaddell, line 297: Should include `omit_legend_suffix` as well, right?

RUFAS/graph_generator.py
```diff
@@ -188,6 +202,74 @@ def generate_graph(
...
+
+        elif len(combined_var_name_list) >= 3:
+            if omit_legend_prefix:
+                slice_start: int = 2 if re.match("([A-Z][a-z0-9]+)+", combined_var_name_list[0]) else 1
+
+            if omit_legend_suffix:
+                slice_end: int = -1 if "=" in combined_var_name_list[-1] else len(combined_var_name_list)
+
+            return ".".join(combined_var_name_list[slice_start:slice_end])
+
```
JoeWaddell, line 280: To avoid the MyPy error about lacking a return statement (since this doesn't cover a combined_var_name_list length of 0 or less), should there be a final `else` statement that doesn't do any prefix/suffix removal, and logs that there was an unexpected length?

tests/test_graph_generator.py
```diff
@@ -217,6 +217,98 @@ def test_generate_graph_exception(graph_generator: GraphGenerator) -> None:
...
+        ),
+        (
+            "dummy_prefix.dummy_var.dummy_var2.dummy_var3.field='field'",
+            False,
+            True,
+            "dummy_prefix.dummy_var.dummy_var2.dummy_var3",
+        ),
+    ],
+)
+def test_generage_legend_keys(
```
JoeWaddell:
```suggestion
def test_generate_legend_keys(
```

JoeWaddell (approved):
Works well (tested multiple options with modified calls to om), and the code is clean. Easy approval assuming the comments are handled.

Also: reminder to update the changelog to reflect that there's an option to omit the suffix, as well. Likewise for any documentation that needs to be updated.

Isgotkowitz:
> As a general observation and suggestion here: in reading the code (and testing!), I see the 'field=field' type suffix is being removed in this method.
> 
> This is misleading since the filter option only refers to the prefix. I could see renaming the filter option to `omit_legend_prefix_and_suffix`, but you could also separate the two out e.g. `omit_legend_prefix` and a new `omit_legend_suffix`, if it's not a big bother, since maybe there's utility in keeping the suffix (though that's more a question for @Isgotkowitz )

Yes, the suffix can be very important when differentiating between the same variable but in different fields, soil layers, crops, etc.

PR #1520, opened 2024-04-10
Add disclaimer message to all output txt, JSON, and CSV files

### LINKED ISSUE #1462: Add disclaimer to the outputs
Add the following disclaimer after the credits and to the first line of all output files: `Under construction, use the results with caution`

allisterakun:
This PR added a disclaimer message to all output txt, JSON, and CSV files and printed out the message to the console when the `LogVerbosity` is set to any level except None.
### Context

Issue(s) closed by this pull request: closes #1462. 

### What

Introduced a new function, `OutputManager._write_disclaimer()`, to write a line of disclaimer.
```python
    def _write_disclaimer(self, file_pointer: TextIO) -> None:
        """
        Writes the predefined disclaimer message to a given file.

        Parameters
        ----------
        file_pointer: TextIO
            A file-like object (supporting the `.write()` method) that points to the file where the disclaimer should
            be written.

        Example
        -------
        >>>
        ... with open('output.txt', 'w') as f:
        ...    self._write_disclaimer(f)
        """
```

Introduced a new function, `OutputManager.print_disclaimer()`, to print the disclaimer message to stdout if the `LogVerbosity` is set to any level except None.
```python
    def print_disclaimer(self) -> None:
        """
        Prints out the disclaimer message when LogVerbosity is set to any level except None.
        """
```

Introduced a new function, `OutputManager._remove_disclaimer_from_input_file`, to handle the issue of disclaimer message breaking the loading pool.
```python
    def _remove_disclaimer_from_input_file(self, file_path: Path) -> str:
        """
        Read the input data file, remove the disclaimer message is it is present on the first line.

        Parameters
        ----------
        file_path : Path
            The path to the file to be loaded to the variables pool.

        Returns
        -------
            The input file content joined into a string
        """
```

### Why

To add and print a disclaimer message.

### How

Modifying all data to file methods to call this function to add the disclaimer.

### Test plan

All affected tests are updated.


ew3361zh (changes requested):
Looks good for the most part. A couple of things that need to get updated:

1. According to the issue, the disclaimer should print out after the credits (for all LogVerbosity >= credits).
2. One way we use json files is to rerun a set of results back through OM to do further processing (graphs, reports, filtering, etc) and if that first line is there in a json file used for this purpose, it will error. What do you think about adding something to the end of the disclaimer in json files that says to remove it if the user needs the results for reprocessing? Or adding it to json files in a way that wouldn't affect a user running the `--load-pool` gnu arg option with the file? Maybe there's a better solution as well - just my initial thought.

RUFAS/output_manager.py
```diff
@@ -660,7 +681,8 @@ def _dict_to_file_csv(self, data_dict: Dict[str, Any], path: str) -> None:
             csv_columns.extend(csv_column_data)
 
         df = pd.concat(csv_columns, axis=1)
-
+        df.insert(loc=0, column="DISCLAIMER", value=DISCLAIMER_MESSAGE)
+        print(df)
```
ew3361zh: remove print statement

RUFAS/output_manager.py
```diff
@@ -1220,6 +1243,25 @@ def flush_pools(self) -> None:
         self.errors_pool = {}
         self.logs_pool = {}
 
+    def _remove_disclaimer_from_input_file(self, file_path: Path) -> str:
```
ew3361zh: This function doesn't look like it's used at all.

ew3361zh (commented):
This is looking good and was working but there's a new function `_remove_disclaimer_from_input_file()` that isn't used at all. Is this for potential txt input files?

I think for json input files we'd discussed maybe not even needing to remove the disclaimer since it's a dictionary k-v pair and will be filtered out. I don't think we have the capacity to load csv/txt files for further processing at this time so maybe not necessary to have this new function?

I think if this function gets accounted for then I'm good to approve.

ew3361zh (approved):
LGTM, nice work, Allister!

RUFAS/output_manager.py
```diff
@@ -396,6 +398,24 @@ def _get_prefix(self, caller_class: str, caller_function: str) -> str:
...
+        file_pointer: TextIO
+            A file-like object (supporting the `.write()` method) that points to the file where the disclaimer should
+            be written.
+
+        Example
+        -------
+        >>>
+        ... with open('output.txt', 'w') as f:
+        ...    self._write_disclaimer(f)
+        """
```
ltp23: This example is not working yet and has the side-effect of creating a new file. A workaround is to use a throwaway string that acts as a file.

```
Examples
--------
>>> output_manager = OutputManager()
>>> import io
>>> file_like_string = io.StringIO()
>>> output_manager._write_disclaimer(file_like_string)
>>> assert file_like_string.getvalue() == DISCLAIMER_MESSAGE + "\\n"
```

RUFAS/output_manager.py
```diff
@@ -660,6 +681,7 @@ def _dict_to_file_csv(self, data_dict: Dict[str, Any], path: str) -> None:
             csv_columns.extend(csv_column_data)
 
         df = pd.concat(csv_columns, axis=1)
+        df.insert(loc=0, column="DISCLAIMER", value=DISCLAIMER_MESSAGE)
```
ltp23: This will cause repeated disclaimer messages in the first column.

```
DISCLAIMER,Column1,Column2  ...
This is a disclaimer message,Value1,Value2   ...
This is a disclaimer message,Value3,Value4   ...
...
```

I think we may want to keep only one value.

```
df = pd.concat(csv_columns, axis=1)
disclaimer_column = [DISCLAIMER_MESSAGE] + [""] * (len(df) - 1)
disclaimer_df = pd.DataFrame({"DISCLAIMER": disclaimer_column})
df = pd.concat([disclaimer_df, df], axis=1)
```

ltp23 (approved):
Looks good to me. I left a couple of comments. Thanks Allister!

PR #1528, opened 2024-04-11
Migrating AnimalRequirements-specific unit tests to dedicated test file.

Fjordrunner:
This PR simply moves the unit tests for functions within the `animal_requirements.py` file to a dedicated test file that was created in a former PR. These unit tests formerly lived in `test_animal.py`.  All unit tests in both affected files are still passing.


ew3361zh (approved):
LGTM! 

I was curious about the MyPy error number increasing and I think it's just a version issue with MyPy. I'm not sure why main would be behind on the version since we're not specifying a version in the requirements. But for MyPy 1.8 (old version) I got a number matching main and for the updated version 1.9 I got the number matching this branch.

In short, I don't think this code increases the MyPy error number but there are 2 MyPy errors in the test_animal_requirements.py file that you could look at addressing.

JoeWaddell (approved):
Thanks for moving these! Agreed that the MyPy errors within (from tests prior to MyPy and more detailed typing) can be fixed in subsequent PRs since the size of this one would make it a challenge to identify subtle changes, though an easy fix or two (like adding a None return to `test_set_requirements`) could be done when updating to main (which could be what Niko was referring to).

PR #1537, opened 2024-04-11
Data_padder_units

### LINKED ISSUE #1531: Method data_padder needs to accept "units"
`data_padder` needs units.

## Current Behavior

Currently crashes, since `om.add_Variable` doesn't have the required units.

## Possible Solution

I'll add the units to the individual calls, and the method itself.

## Steps to Reproduce

1. Alerted to this by @tomhuhh , who Ran `c2f12_metadata.json` on the `lactation_curve_simulations_March` branch.
2. Was able to replicate, get the following error. 
image

## Context (Environment)

Quickest fix in this case is increasing pen sizes so overflows don't happen, permanent fix is adding units to the method and its calls.

JoeWaddell:
### Context

Issue(s) closed by this pull request: closes #1531

### What

Adds unit input parameter to `data_padder`

### Why

See tagged Issue, the method didn't include units.

### How

Added new arg, in lieu of adding the entire info_map itself, though that would be an appropriate fix as well.

### Test plan

Updated existing unit tests.

### Needs help
Typing for the units: since the calls to units in the codebase are fairly diverse in their format/typing, I wasn't sure how to capture that succintly. The current `Dict[str, str | MeasurementUnits] | MeasurementUnits | str` seems to fail in some cases (even a Dict[str, str]), so I think maybe I'll need to use `Sequence` or something similar?
image

### DIFF
--- a/RUFAS/routines/animal/animal_module_reporter.py
+++ b/RUFAS/routines/animal/animal_module_reporter.py
@@ -29,7 +29,7 @@ class AnimalModuleReporter:
         thing_to_add: Any,
         simulation_day: int,
         info_map: Dict[str, Any],
-        units: Dict[str, str] | MeasurementUnits | str
+        units: Dict[str, str | MeasurementUnits] | MeasurementUnits | str
     ) -> None:
         """
         Pads a variable in OutputManager for entries that it "missed" relative to another variable.
@@ -57,6 +57,9 @@ class AnimalModuleReporter:
             The day of the simulation.
         info_map: Dict[str, Any]
             The info_map to use when padding.
+        units: Dict[str, str | MeasurementUnits] | MeasurementUnits | str
+            Units for the variable being added, in the format provided in the main call to add_variable,
+            (e.g. the one following the call of data_padder).
 
         """
         if simulation_day > 0 and reference_variable in om.variables_pool:
@@ -347,8 +350,8 @@ class AnimalModuleReporter:
             del ration_per_animal["status"]
             del ration_per_animal["objective"]
             ration_total = {}
-            ration_total["dry_matter_intake_total"] = 0
-            ration_total["byproducts_total"] = 0
+            ration_total["dry_matter_intake_total"] = 0.0
+            ration_total["byproducts_total"] = 0.0
             for key in ration_per_animal.keys():
                 if key != "status" and key != "objective":
                     ration_total[key] = pen.ration_per_animal[key] * len(pen.animals_in_pen)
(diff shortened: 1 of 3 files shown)


RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -29,6 +29,7 @@ def data_padder(
         thing_to_add: Any,
         simulation_day: int,
         info_map: Dict[str, Any],
+        units: Dict[str, str | MeasurementUnits] | MeasurementUnits | str,
```
ew3361zh: Because of the way units are nested within info_maps and not wanting to extract and change them from their enums to their enum values later for reporting, I ended up using the enum values (which are the actual strings for the units). When handling units is finished, we may need to go back and change these to the Enums but in the meantime, these type hints should be for strings.
```suggestion
        units: Dict[str, str] | str,
```

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -56,6 +57,9 @@ def data_padder(
             The day of the simulation.
         info_map: Dict[str, Any]
             The info_map to use when padding.
+        units: Dict[str, str | MeasurementUnits] | MeasurementUnits | str
```
ew3361zh:
```suggestion
        units: Dict[str, str] | str
```

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -243,6 +248,7 @@ def report_ration_interval_data(cls, pen_list: List[Pen], feed: Feed, simulation
                 0,
                 simulation_day,
                 info_map,
+                MeasurementUnits.KILOGRAMS,
```
ew3361zh:
```suggestion
                MeasurementUnits.KILOGRAMS.value,
```

ew3361zh (approved):
Looks good, Joe. I have a couple changes that need to be implemented for consistency's sake in that I went using the values for the enums rather than the enums themselves.

I think I missed a couple in my revision though which I wasn't able to comment on here because it wasn't in parts of the file you touched.

line 256  in `animal_module_reporter.py` should be:

```
dict(info_map, **{"units": MeasurementUnits.KILOGRAMS.value})
```

and line 284 should be:

```
ration_per_animal_units = {key: MeasurementUnits.KILOGRAMS.value for key in ration_per_animal.keys()}
```

tomhuhh:
Thanks @JoeWaddell ! Could we also update the upper limit for the stocking density to 3 or 5 in this PR? Thanks!

tomhuhh (approved):
Looks good to me! I tested with some extremely low number of stalls across different pens and the code gave me no error.

Thanks Joe!

PR #1538, opened 2024-04-11
[Main] Total random seed control

### LINKED ISSUE #1523: Record random seed when `set_seed` is set to false.
Randomization is used heavily in RuFaS, and seeding the randomization libraries used is optional. This is an important feature, but can make it very difficult to reproduce crashes caused by intermittent bugs. 

If `set_seed` in the Config input is set to `false`, generate a random number, log and/or record it, then use that number to seed the random libraries used.

Capture the "state" of the random libraries, record it, and set up some input/method of passing that state in and configuring the random libraries with it. [This StackOverflow post](https://stackoverflow.com/questions/32172054/how-can-i-retrieve-the-current-seed-of-numpys-random-number-generator) has a good explanation of how/why to do this for Numpy's random library. A similar method would have to be developed for Python's standard `random` library as well.  

**Additional context**
If the first solution (generating a seed randomly, then seeding with it) is used, SMEs need to be consulted to ensure that the "randomness" of that seed is appropriate for use in RuFaS.

Isgotkowitz:
This PR records the seed used for randomization libraries in RuFaS when one is not provided in the inputs.

### Context

Issue(s) closed by this pull request: closes #1523 

### What

When a simulation is configured to _not_ seed random libraries with the seed provided in the inputs, a random seed is generated randomly, recorded, and used to seed the random libraries.

Also, an upper bound (${2}^{32} - 1$) has been added for `"seed"` in the config input.

### Why

When running many random simulations, bugs can crop up intermittently. Without know the initial state of the random libraries (i.e. what they were seeded with) were in that simulation, it makes it very difficult to reproduce that bug. This PR fixes this problem by recording and logging the random seed used even if the user does not provide one.

The upper bound for `"seed"` was added because Numpy's `random.seed` method has an upper bound.

### How

If the user configures a simulation to run randomly (`set_seed` in the config input set to false), an integer is randomly generated in the range $[0, {2}^{32} - 1]$. This integer is logged and recorded as a variable, then used to seed the Numpy and Python `random` libraries.

### Test plan

All changes to code were made in the `set_seed` method of `main`, so only the unit test for `set_seed` was updated. Also tested by running several simulations with `set_seed` set to `false` to observe the logging of the generated random seed.

### DIFF
--- a/main.py
+++ b/main.py
@@ -10,7 +10,7 @@ import argparse
 import random
 import traceback
 from pathlib import Path
-from typing import List
+from typing import List, Any
 
 import numpy
 
@@ -27,7 +27,7 @@ NUMPY_RANDOM_SEED_LOWER_BOUND = 0
 NUMPY_RANDOM_SEED_UPPER_BOUND = 2**32 - 1
 
 
-def main():
+def main() -> None:
     cmd_arguments = parse_gnu_args()
     if cmd_arguments.load_pool:
         load_pool = True
@@ -297,7 +297,7 @@ def set_random_seed(input_manager: InputManager) -> None:
     set_seed = input_manager.get_data("config.set_seed")
 
     om = OutputManager()
-    info_map = {"class": "No caller class", "function": set_random_seed.__name__}
+    info_map: dict[str, Any] = {"class": "No caller class", "function": set_random_seed.__name__}
 
     if set_seed:
         seed = input_manager.get_data("config.random_seed")
--- a/main.py
+++ b/main.py
@@ -10,7 +10,7 @@ import argparse
 import random
 import traceback
 from pathlib import Path
-from typing import List, Any
+from typing import List
 
 import numpy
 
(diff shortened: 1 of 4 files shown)


main.py
```diff
@@ -291,15 +297,19 @@ def set_random_seed(input_manager: InputManager) -> None:
     """
     set_seed = input_manager.get_data("config.set_seed")
 
+    om = OutputManager()
+    info_map: dict[str, str | MeasurementUnits] = {"class": "No caller class", "function": set_random_seed.__name__}
```
allisterakun, line 301: Just wondering, if having space in the class name "No caller class" would later be problematic for OM.

allisterakun (approved):
Good job, Ed! Just on comment

jadamchick (approved):
Looks great! I ran it several times with one of the pilot testing farms. The simulation ran smoothly and the log file recorded the seed used each time. Results were comparable to those from running the same farm on pilot testing branch. One run threw an error and ended early because of user error (referenced the wrong feed library); the log file also recorded the seed used for that run.

main.py
Isgotkowitz: Fortunately it does not, though the output is not very pretty. The result of filtering out the generated seed looks like a CSV with this:
|No caller class.set_random_seed.generated_random_seed|
|-|
|363401316|

Issue #1516, opened 2024-04-10
[OM] Include units in reports when specified


ltp23:
This is part 7 of the issue #1416 

- [ ] Include units in the reports when specified by the user.
- [ ] To make the formatting of units more consistent, consider reusing the code for formatting units.
- [ ] Make sure the functionality `graph with report` still work like before.

ltp23:
@JoeWaddell @tomhuhh @jadamchick 

Please provide suggestions/preferences (if any) on how you want to display units in reports.

tomhuhh:
Hi Loi! Thanks for bringing this to our attention. It seems to me that the variables that I've been working with all have their units specified (either in variable names or info maps), even though some may not be correct. Do you have any specific examples of the options for displaying units? Also, are you referring to the report generated by the txt filter, the report generated by the json filter, or the graphic report? Thanks!

ltp23:
Thanks @tomhuhh. I made a separate issue for each output format type (#1434, #1514, #1515, #1516) , but you can make suggestions on any of them. Julie did mention a good example of auto-including the units in the report column names. I will try to always somehow include a flag to turn the features on and off.

allisterakun:
During grooming session:
If info_map is not excluded, the units will be passed to the ReportGenerator in the info_map section, and we can extract the units from there.
If info_map is excluded, no units will be passed to the ReportGenerator. Therefore, we won't put units on graphs.

allisterakun:
> During grooming session: If info_map is not excluded, the units will be passed to the ReportGenerator in the info_map section, and we can extract the units from there. If info_map is excluded, no units will be passed to the ReportGenerator. Therefore, we won't put units on graphs.

This issue is being sized as L because through the process of aggregation, new complex units can emerge. The RG code should be able to handle this. 
For example, if variables `foo` and `bar` are aggregated through division, the units of `foo / bar` will be `x/y` where `x` is the unit of `foo`, and `y` is the unit of `bar`.

PR #1542, opened 2024-04-13
Include units in csv column names

### LINKED ISSUE #1514: [OM] Include units in csv column names
This is part 5 of the issue #1416 

- [x] Include units directly in the csv column names.
- [x] Remove extra columns like `.prefix`, `.units`.

Currently, it is already possible to include units when using csv filtering.

For example, with this filter

```
Weather.average_annual_temperature
```

I got the following

```
Weather.average_annual_temperature,Weather.average_annual_temperature.units,Weather.average_annual_temperature.prefix
8.178051643192488,°C,Weather
```

ltp23:
### Context

Issue(s) closed by this pull request: closes #1514 

### What & Why
When using csv filtering (`csv_<example>.txt`), there are additional columns such as `.prefix` and `.units` that seem redundant and use extra memory storage. This PR aims to remove those unnecessary columns and directly incorporate the units into the column names.

When the exclude-info-maps flag `-i` is used, that will disable this feature because information about units is stored in the info maps.

This feature doesn't apply to reports yet (which are also in the csv format) because there are some other considerations such as missing info maps and different combinations of aggregation operations affecting the final units. There is an open issue regarding this #1516.

### How

- [x] Look for the units key in the `info_maps` dictionary if it exists. Ignore other keys in the `info_maps`.
- [x] Format the units somehow. This PR only does a simple appending of the units string to the variable name `simple_variable_name_<units>` or `<dictionary_variable_name>.subkey_<units>`.
- [x] Do some code refactoring and write/update unit tests
- [x] There are some missing units in the animal module reporter and they need to be added.

### Test plan

To test things out, I would create a filter file `csv_examples.txt` and select simple and dictionary variables.

```
Weather.average_annual_temperature
^AnaerobicDigestion_ManureTreatmentDailyOutput_.*LAC_COW.liquid_manure_total_volatile_solids$
AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_.*LAC_COW
```

### DIFF
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -11,9 +11,9 @@ from typing import Any, Dict, List, Union, Tuple
 import pandas as pd
 from deprecated.sphinx import deprecated
 
-from RUFAS.units import MeasurementUnits
 from RUFAS.graph_generator import GraphGenerator
 from RUFAS.report_generator import ReportGenerator
+from RUFAS.units import MeasurementUnits
 from RUFAS.util import Utility
 
 
@@ -537,24 +537,61 @@ class OutputManager(object):
             A list of (column_name, column_data) tuples.
 
         """
-        column_list = []
-        mandatory_fields = ["values", "info_maps"] if "info_maps" in data_dict else ["values"]
-        for field in mandatory_fields:
-            data_list = data_dict[field]
-            if data_list and isinstance(data_list[0], dict):
-                csv_column_lists: Dict[str, List[Any]] = {subkey: [] for item in data_list for subkey in item.keys()}
-                for nested_dictionary in data_list:
-                    for subkey, value in nested_dictionary.items():
-                        csv_column_lists[subkey].append(value)
 
-                for subkey in csv_column_lists.keys():
-                    column_title = f"{variable_name}.{subkey}"
-                    column_list.append(pd.Series(csv_column_lists[subkey], dtype=object, name=column_title))
-            else:
-                column_title = f"{variable_name}"
-                column_list.append(pd.Series(data_list, dtype=object, name=column_title))
+        column_list = []
+        units = data_dict["info_maps"][0]["units"] if data_dict.get("info_maps", []) else None
+        data_list = data_dict["values"]
+        if data_list and isinstance(data_list[0], dict):
+            csv_column_lists: Dict[str, List[Any]] = {subkey: [] for item in data_list for subkey in item.keys()}
+            for nested_dictionary in data_list:
+                for subkey, value in nested_dictionary.items():
+                    csv_column_lists[subkey].append(value)
+
+            for subkey in csv_column_lists.keys():
+                column_title = f"{variable_name}.{subkey}{self._get_units_substr(variable_name, units, subkey)}"
+                column_list.append(pd.Series(csv_column_lists[subkey], dtype=object, name=column_title))
+        else:
+            column_title = f"{variable_name}{self._get_units_substr(variable_name, units)}"
+            column_list.append(pd.Series(data_list, dtype=object, name=column_title))
 
         return column_list
+
+    def _get_units_substr(
+        self, variable_name: str, units: str | Dict[str, str] | None, subkey: str | None = None
+    ) -> str:
+        """Get the units substring for a column title.
+
+        Parameters
+        ----------
+        variable_name : str
+            The name of the variable or group of variables associated with the units.
+        units : str | Dict[str, str] | None
+            The units associated with the data.
+        subkey : str | None, optional
+            The subkey to retrieve the units for, if units is a dictionary. Default is None.
+
+        Returns
+        -------
+        str
+            The formatted units substring for the column title.
+        """
+
+        if not isinstance(units, dict):
+            return f"_{units}" if units else ""
+
+        if subkey not in units:
+            if subkey is not None:
+                self.add_error(
+                    "units_key_error",
+                    f"Key '{subkey}' not found in the units dictionary for variable '{variable_name}'.",
+                    info_map={
+                        "class": self.__class__.__name__,
+                        "function": self._get_units_substr.__name__,
+                    },
+                )
+            return ""
+
+        return f"_{units[subkey]}"
 
     def _dict_to_file_csv(self, data_dict: Dict[str, Any], path: str) -> None:
         """Saves a dictionary to a csv file.
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -606,7 +606,7 @@ class OutputManager(object):
         if subkey is None:
             self.add_error(
                 "units_subkey_missing",
-                f"Subkey is required when units is a dictionary for variable '{variable_name}'.",
+                f"Variable {variable_name} has the 'units' property as a dictionary, please specify a key.",
                 info_map={
                     "class": self.__class__.__name__,
                     "function": self._get_units_substr.__name__,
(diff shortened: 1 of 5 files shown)


ew3361zh (approved):
Nice work, Loi. I tested this with a few filters and then one of the currently nested units dicts and it accounted for them all well.

I wondered if adding the units at the end might be more readable if they were added in parantheses?

e.g.

`ManureTreatmentDailyOutput_Pen_1_GROWING.daily_final_manure_volume_m^3`
->
`ManureTreatmentDailyOutput_Pen_1_GROWING.daily_final_manure_volume_(m^3)`

Testing looked good as well.

RUFAS/output_manager.py
```diff
@@ -549,25 +549,62 @@ def _dict_to_csv_column_list(self, variable_name: str, data_dict: Dict[str, List
...
+
+        Parameters
+        ----------
+        variable_name : str
+            The name of the variable or group of variables associated with the units.
+        units : str | Dict[str, str] | None
+            The units associated with the data.
+        subkey : str | None, optional
+            The subkey to retrieve the units for, if units is a dictionary. Default is None.
+
```
ew3361zh, line 584: I think an example could be helpful here to understand what the subkey is.

ltp23: I have added some examples in the docstring.

ltp23:
> I wondered if adding the units at the end might be more readable if they were added in parantheses?
> 
> e.g.
> 
> `ManureTreatmentDailyOutput_Pen_1_GROWING.daily_final_manure_volume_m^3` -> `ManureTreatmentDailyOutput_Pen_1_GROWING.daily_final_manure_volume_(m^3)`

Thanks Niko. I have just made a change and put the units in parentheses.

RUFAS/output_manager.py
```diff
@@ -549,25 +549,72 @@ def _dict_to_csv_column_list(self, variable_name: str, data_dict: Dict[str, List
...
+                    "units_key_error",
+                    f"Key '{subkey}' not found in the units dictionary for variable '{variable_name}'.",
+                    info_map={
+                        "class": self.__class__.__name__,
+                        "function": self._get_units_substr.__name__,
+                    },
+                )
+            return ""
+
+        return f"_({units[subkey]})"
```
```suggestion
        return f" ({units[subkey]})"
```
Instead of appending with an underscore, I think it would be cleaner to just the units enclosed in parentheses. This makes the column a little more readable, especially for non-devs who don't spend all their reading and writing things in snake case.

RUFAS/output_manager.py
```diff
@@ -549,25 +549,72 @@ def _dict_to_csv_column_list(self, variable_name: str, data_dict: Dict[str, List
...
+            if subkey is not None:
+                self.add_error(
+                    "units_key_error",
+                    f"Key '{subkey}' not found in the units dictionary for variable '{variable_name}'.",
+                    info_map={
+                        "class": self.__class__.__name__,
+                        "function": self._get_units_substr.__name__,
+                    },
+                )
+            return ""
```
Isgotkowitz: If the units aren't present for a variable, I think that merits logging an error/or warning

Isgotkowitz: If the subkey is not in the units, I think that should be its own case that causes an error to be logged. And likewise for the case when the subkey is "None".

RUFAS/output_manager.py
```diff
@@ -549,25 +549,72 @@ def _dict_to_csv_column_list(self, variable_name: str, data_dict: Dict[str, List
...
-                    for subkey, value in nested_dictionary.items():
-                        csv_column_lists[subkey].append(value)
-
-                for subkey in csv_column_lists.keys():
-                    column_title = f"{variable_name}.{subkey}"
-                    column_list.append(pd.Series(csv_column_lists[subkey], dtype=object, name=column_title))
-            else:
-                column_title = f"{variable_name}"
-                column_list.append(pd.Series(data_list, dtype=object, name=column_title))
+        units = data_dict["info_maps"][0]["units"] if data_dict.get("info_maps", []) else None
```
Isgotkowitz, line 554: This leaves the door open for a pretty serious issue: what if the same variable is recorded with varying units? It seems unlikely that this would ever happen, but currently it is possible and we should consider whether or not it should be allowed. And if it is decided that it shouldn't be allowed then it is an opportunity for some large reductions in the memory footprint of info maps, because the units would only have to be recorded once and not in every single info map.

Note that this issue is well outside the scope of this PR, I just want to bring it up so that we can start considering solutions.

Isgotkowitz (commented):
Looks good Loi! Wrote some filters and they were behaving as expected. I left some suggestions and thoughts, but they are take it or leave it.

I am holding off on approving this just because I think the second approval should come from a SME, and this PR already has one dev review

RUFAS/output_manager.py
ltp23: Okay. I will add a whitespace between the variable name and the units.

RUFAS/output_manager.py
ltp23: Both CSV filtering and report generation share the same method of making CSV files. Currently, at the point of making CSVs from reports, the info maps have been stripped so this logging would cause lots of warnings. I haven't yet worked out how to incorporate units into the reports gracefully considering the existing usage by pilot testers.

ltp23: I can rearrange the conditions a little bit to make things clearer.

jadamchick (commented):
Hi Loi, I like it! 
For some reason it didn't seem to work for the first set of filters I tried (relevant to manure excretion outputs) -- though it looks like those units exist in the code (and when I run the same CSV on main). 
```
AnimalModuleReporter.report_animal_module_manure.daily_aggregate_.*_total_ammoniacal_nitrogen_concentration
AnimalModuleReporter.report_animal_module_manure.daily_aggregate_.*_urine_nitrogen
AnimalModuleReporter.report_animal_module_manure.daily_aggregate_.*_manure_nitrogen
AnimalModuleReporter.report_animal_module_manure.daily_aggregate_.*_phosphorus
AnimalModuleReporter.report_animal_module_manure.daily_aggregate_.*_phosphorus_fraction
AnimalModuleReporter.report_animal_module_manure.daily_aggregate_.*_total_solids
AnimalModuleReporter.report_animal_module_manure.daily_aggregate_.*_degradable_volatile_solids
AnimalModuleReporter.report_animal_module_manure.daily_aggregate_.*_non_degradable_volatile_solids
```

but when I tried with a different set, it seemed to work fine (including the cull reason units which are also within a dictionary). 

```
^AnimalModuleReporter.report_life_cycle_manager_data.*
^AnimalModuleReporter.report_daily_animal_population.*
```

Any thoughts on why the units didn't come through on the set for `report_animal_module_manure`?

Let me know if that is something you want to address and get a re-review before approval, or if you think it is something unrelated to this PR.

RUFAS/output_manager.py
ltp23: Because I'm using the hardcoded index 0, I'm assuming that the units should stay consistent for all the add-variables calls to the same variable. But this can change if we want to accommodate the case where units can be dynamic. 

I also think we could perform a check when we add a variable to the OM pool to make sure the current units are the same as the last added units.

ltp23:
Hi Julie (@jadamchick), there is a small error when reporting units for several variables but I have fixed it now.

```
AnimalModuleReporter.report_animal_module_manure.daily_aggregate_.*
AnimalModuleReporter.report_pen_manure_properties.pen_.*
SimulationEngine.simulate.available_feeds_on_final_day
```

or use the catch-all filter

```
.*
```

Please let me know if it also works on your end.

jadamchick (approved):
Works great now!

PR #1547, opened 2024-04-15
Account for Precipitation in Outdoor Manure Storages

### LINKED ISSUE #1422: [Manure] Account for precipitation in manure volume for SlurryStorageOutdoor
## Urgency Level

- [x] Medium: the issue affects functionality but has a workaround

## Status

Currently there is no accounting for daily precipitation in the calculation of total manure volume for SlurryStorageOutdoor.

## Intention

Precipitation is a key component to take into account for calculating manure volume for the outdoor manure treatment processes.

## Request

Currently we do track precipitation for AnaerobicLagoon manure volume with its `_adjust_final_manure_volume()` method and we should be doing a similar process for SlurryStorageOutdoor. That would be a good place to start and consider moving that logic into BaseManureTreatment since it's used by at least 2 different treatment child classes.

ew3361zh:
Try to standardize how outdoor manure treatments account for precipitation when calculating `daily_final_manure_volume` and N2O emissions based on manure cover.

### Context

Issue(s) closed by this pull request: closes #1422

### What

- Updates `_adjust_final_manure_volume()` method in `AnaerobicLagoon` to account for whether there is a cover in the config.

- Adds `_adjust_final_manure_volume()` method to `SlurryStorageOutdoor` class and ensured it was also taking into account the cover status from the config.

- Adds `manure_cover` option for `crust` which follows the same behavior for handling precipitation for the outdoor treatments/storages as `no cover` and follows the same behavior for handling N2O emissions as `cover`.

Updates unit testing.

### Why

Because both `AnaerobicLagoon` and `SlurryStorageOutdoor` are outdoor manure storages, they need to take into account precipitation when calculating `daily_final_manure_volume` and they should do it in a similar fashion.

Adding the cover option now allows distinction between a naturally occurring manure crust cover and an impermeable human-made cover.

### How

One thing to note in the how is that I opted not to add this method (`_adjust_final_manure_volume()`) to the `BaseManureTreatment` class despite it being shared by these 2 manure treatment child classes. It's not shared/used by any of the other manure treatment child classes so it seemed like more of a violation of the Interface Segregation Principle than adding it to the base class would be supporting the Open-Closed Principle.

### Test plan

Testing was updated to ensure coverage extended to the method's use in `SlurryStorageOutdoor` as well as maintaining accuracy of coverage for existing usage with `AnaerobicLagoon` for the new `crust` option.

Testing was added for the N2O calculation in `gas_emissions_constants.py`.

### DIFF
--- a/RUFAS/routines/manure/manure_treatments/anaerobic_lagoon.py
+++ b/RUFAS/routines/manure/manure_treatments/anaerobic_lagoon.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import math
 from typing import Tuple
 
+from RUFAS.routines.manure.enums.ManureCoverEnum import ManureCoverEnum
 from RUFAS.routines.manure.constants_and_units.gas_emission_constants import GasEmissionConstants
 from RUFAS.routines.manure.constants_and_units.manure_constants import ManureConstants
 from RUFAS.routines.manure.gas_emissions.calculator import (
@@ -208,7 +209,10 @@ class AnaerobicLagoon(BaseManureTreatment):
             The adjusted final manure volume.
 
         """
-        return current_day_final_manure_volume + self.precipitation_volume
+        if self.config.manure_cover == ManureCoverEnum.NO_COVER.value:
+            return current_day_final_manure_volume + self.precipitation_volume
+        else:
+            return current_day_final_manure_volume
 
     @property
     def sludge_accumulation_volume(self) -> float:
--- a/RUFAS/routines/manure/manure_treatments/anaerobic_lagoon.py
+++ b/RUFAS/routines/manure/manure_treatments/anaerobic_lagoon.py
@@ -209,7 +209,7 @@ class AnaerobicLagoon(BaseManureTreatment):
             The adjusted final manure volume.
 
         """
-        if self.config.manure_cover == ManureCoverEnum.NO_COVER.value:
+        if self.config.manure_cover in [ManureCoverEnum.NO_COVER.value, ManureCoverEnum.CRUST.value]:
             return current_day_final_manure_volume + self.precipitation_volume
         else:
             return current_day_final_manure_volume
--- a/RUFAS/routines/manure/manure_treatments/anaerobic_lagoon.py
+++ b/RUFAS/routines/manure/manure_treatments/anaerobic_lagoon.py
@@ -369,10 +369,7 @@ class AnaerobicLagoon(BaseManureTreatment):
 
         """
         if self._current_pen is not None and self._current_pen.num_animals is not None:
-            return (
-                self._current_pen.num_animals
-                * GasEmissionConstants.DEFAULT_STORAGE_AREA_PER_ANIMAL
-            )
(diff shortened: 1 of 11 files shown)


allisterakun (approved):
LGTM! Good job, Niko

RUFAS/routines/manure/manure_treatments/anaerobic_lagoon.py
```diff
@@ -227,7 +231,10 @@ def flushing_volume(self) -> float:
             Flushing water volume, m^3
 
         """
-        return self._manure_handler_daily_output.cleaning_water_volume
+        if self._manure_handler_daily_output:
+            return self._manure_handler_daily_output.cleaning_water_volume
+        else:
+            return 0.0
```
ew3361zh, line 237: Addresses MyPy error for case when `self._manure_handler_daily_output` is `None`.

RUFAS/routines/manure/manure_treatments/slurry_storage_outdoor.py
```diff
@@ -69,7 +70,10 @@ def treatment_volume(self) -> float:
             The minimum treatment volume, m^3.
 
         """
-        return self.wastewater_volume * self.storage_time_period
+        if isinstance(self.storage_time_period, int):
+            return self.wastewater_volume * self.storage_time_period
+        else:
+            return 0.0
```
ew3361zh, line 76: Addresses MyPy error for case when `self.storage_time_period` is `None`.

RUFAS/routines/manure/manure_treatments/slurry_storage_outdoor.py
```diff
@@ -227,7 +251,7 @@ def calc_ammonia_emission(
         num_animals: int,
         accumulated_manure_volume: float,
         accumulated_manure_total_ammoniacal_nitrogen: float,
-    ) -> Tuple[float, float]:
+    ) -> float:
```
ew3361zh, line 260: MyPy error correction - should return a `float` here, not a tuple of floats.

elle-andreen (approved):
This is looking great to me! For others' reference, here is the updated behavior of covers and crusts:

N2O emission logic:
- If manure_cover = "cover" or "crust", N2O emission factor is used (i.e. N2O is produced). Value of emission factor is the same regardless if the input is "cover" or "crust", the cover type just determines if the factor is used or not.
- If manure_cover = "no cover", the emission factor is not used (i.e. no N2O produced).

Precipitation logic:

- If manure_cover  = "no cover" or "crust", precipitation volume is added to storage. Precipitation addition is reported in the ManureTreatmentDailyOutput value. Natural manure crusts are considered permeable to rainwater. 
- If manure_cover  = "cover", precipitation volume is not added to storage. Though not all manure covers are completely rain impermeable, we have made a simplification in considering them to completely prevent rain entry into manure storage. 

Note that slurry underfloor and the anaerobic digestion portion of anaerobic digestion and lagoon do not receive precipitation, as they are currently considered to be always covered.

PR #1552, opened 2024-04-17
[IM] Unify IM validator method signatures

### LINKED ISSUE #1487: [IM] Unify signatures and behavior of validator methods in Input Manager
There are several validator routines in the Input Manager used to validate different types of data, and although they share many similar behaviors they do not share a common method signature.

## Expected Behavior

All validator methods (`_string_type_validator`, `_num_type_validator`, `_bool_type_validator`, `_array_type_validator`, and `_object_type_validator`) have the same method signature. The routing of inputs into these methods and the way that data is handled after being routed through them should be made uniform as well.

## Current Behavior

The methods `_string_type_validator`, `_num_type_validator` and `_bool_type_validator` share one signature, and `_bool_type_validator` and `_array_type_validator` share another. The effect is that this leads to different behavior after validating different types of inputs.

## Possible Solution

Change the signatures of `_string_type_validator`, `_num_type_validator` and `_bool_type_validator` to match `_bool_type_validator` and `_array_type_validator`. Also modify the behavior in `_validate_input_by_type` of replacing invalid data with defaults so that all data types are replaced with default data, if the data passed is found to be invalid.

Isgotkowitz:
This PR unifies the method signatures of the five data validator methods.

### Context

Issue(s) closed by this pull request: closes #1487 

- N/A, this PR does not change the functionality of the Input Manager at all -- The [changelog](https://docs.google.com/spreadsheets/d/1U-3igxdstHTFb6k5dVcE5-x9IL0r6q18eUiwwytKZyA/edit#gid=0) has been updated.

### What

The method signatures for validating the five different types of data that the metadata accepts (numbers, strings, bools, objects, and arrays) have been unified so that they all have the same arguments. This allows the wrapper method `_validate_input_by_type` to route input data into the appropriate method using a single mapping, instead of two.

### Why

By not distinguishing between complex and primitive data types in `_validate_input_by_type`, the code is made much simpler and more streamlined.

### How

The signatures have been changed for `_bool_type_validator`, `_number_type_validator`, and `_string_type_validator` have been changed to match the signatures for `_array_type_validator` and `_object_type_validator`.

### Test plan

The unit tests for `_bool_type_validator`, `_number_type_validator`, `_string_type_validator`, and `_validate_input_by_type` have been updated. Also, all available metadatas have been run and checked to ensure that Input Manager finds the same issues and produces the same messages.

### Open Questions
- How should the argument `called_during_initialization` be used? It is currently passed from method to method, but not actually used.
- Is it ok to pass arguments to a method that aren't used? It seems like an anti pattern, but in this case makes the code simpler in some ways.


RUFAS/input_manager.py
```diff
@@ -828,61 +828,61 @@ def _validate_input_by_type(
...
-            return complex_type_to_validator_map[data_type](
-                variable_path,
-                variable_properties,
-                input_data,
-                eager_termination,
-                properties_blob_key,
-                elements_counter,
-                called_during_initialization,
-            )
+        fixable_data_types = ["string", "number", "bool"]
```
ew3361zh: I would think about adding these as instance attributes rather than having it defined locally within this function.
Similar to what we do with OM's `__supported_filter_types_prefixes`

```
self.__supported_filter_types_prefixes: Dict[str, str] = {
                "csv": "csv_",
                "graph": "graph_",
                "json": "json_",
                "report": "report_",
            }
```

ew3361zh (approved):
LGTM, nice work, Ed!

We discussed the questions posed in the PR:
How should the argument called_during_initialization be used? It is currently passed from method to method, but not actually used.

- I believe we decided this needs to be addressed but is not necessarily within the scope of this PR.

Is it ok to pass arguments to a method that aren't used? It seems like an anti pattern, but in this case makes the code simpler in some ways.

I think the clarity added at the higher level function is worth any potential wasted variables in the lower level functions. And no MyPy nor flake8 warnings are raised with this behavior so I think it's at least not considered bad by those entities.

RUFAS/input_manager.py
Isgotkowitz: Declared as a module constant!

allisterakun (approved):
Excellent job, Ed!

Just note the `called_during_initilization` variable being passed through and not used. I have put this into issue #1565

PR #1555, opened 2024-04-18
Generate a CSV of metadata properties used to validate the input data

### LINKED ISSUE #1512: [IM] Add feature to generate list of inputs
`OutputManager` automatically produces a file containing all the output variables. Here we want to replicate the functionality for inputs.

ew3361zh:
### Note
We're experiencing some of the issue with commits from other PRs getting pulled into unrelated PRs. The only files actually changed by this PR:
- `input_manager.py`
- `test_input_manager.py`
- `main.py`
- `test_main.py`

New functionality added to input manager that will generate a CSV of the metadata properties used to validate the input data.

### Context

Issue(s) closed by this pull request: closes #1512

- [x] Update Wiki gnu args

### What

This PR adds 4 new functions to IM:
1. `dump_metadata_properties()` which is called from `main.py` which oversees the process to convert the metadata properties to CSV. 
2. `_parse_metadata_properties()` which recursively traverses the metadata properties to flatten it into a list of dictionaries containing each of the elements in the properties.
3. `_create_record()` which is a helper function to format the flattened properties while also reducing duplicate code used in `_parse_metadata_properties()`.
4. `_check_property_type_primitive()` which is another helper function for `_parse_metadata_properties()` that determines whether a property is at the level of a primitive property or if there is more nested objects within it.

This process is only triggered when the user uses the `-a`/`--audit-input-data` gnu arg (formerly known as `-o`/`--only-run-validation`). This gnu arg still validates the input data but now also produces a CSV of the metadata properties.

The CSV of the metadata properties has 8 columns:
`properties_group``name`, `type`, `description`, `pattern`, `default`, `maximum`, `minimum`.

Each of these is collected for every element even if it doesn't apply to that element to ensure standardization in adding the elements to the CSV.

With arrays, there was a choice of it being 2 elements or 1 element:
2 elements version:
`fertilizer_schedule_properties_pattern_repeat,array,Number of times that this fertilizer application schedule should be repeated.,,,,`
`fertilizer_schedule_properties_pattern_repeat,number,,,0,,0`

1 element version:
fertilizer_schedule_properties_pattern_repeat,number,Number of times that this fertilizer application schedule should be repeated.,,0,,0

I opted for the second version which meant grabbing the `description` field from the parent `array` to the nested primitive `number` element.

### Why

This was a feature requested by multiple partners/stakeholders.

### How

The functionality was added to input manager which seemed like the best fit. But adding the feature to run during every simulation seemed like adding unwanted processing time. It was discussed in our dev-team group meeting to add as an extension of the `only-run-validation` gnu arg because it fit well with the process of validating the data.

### Test plan

All new functions in IM have been tested and IM coverage is at 100%. Additionally I expanded coverage for the call to this new functionality from main.py.

To test the results, in the command-line/terminal type:
`python main.py -a`

It will generate a CSV file saved in the output folder with the name `_InputManager_metadata_properties_date.csv`


RUFAS/input_manager.py
```diff
@@ -2009,6 +2028,116 @@ def dump_get_data_logs(self, path: Path) -> None:
...
+            The separator used between parts of the data entry names, by default '_'.
+
+        Returns
+        -------
+        List[Dict[str, Any]]
+            A list of flattened data entries from the json file.
+        """
+        records = []
+        for property_key, property_value in data.items():
+            if isinstance(property_value, dict):
```
allisterakun, line 2071: There is not an else statement opposite to this if. So... is it necessary?

RUFAS/input_manager.py
```diff
@@ -2009,6 +2028,116 @@ def dump_get_data_logs(self, path: Path) -> None:
...
+                                    sep,
+                                )
+                            )
+                    elif self._check_property_type_primitive(property_value):
+                        name = prefix + sep + property_key
+                        record = self._create_record(property_value, name)
+                        records.append(record)
+                        break
+                    elif property_value.get("type") == "object":
+                        self._parse_metadata_properties(property_value, prefix + sep + property_key, sep)
```
allisterakun, line 2100: Maybe extract this into a separate function?

allisterakun, line 2100: If these are checking the property_value, should they be moved up a level of indentation?

allisterakun (approved):
Generally looks good! Just wanted to confirm if the indentation in `_parse_metadata_properties()` are as intended! See the comments for detail.

RUFAS/input_manager.py
ew3361zh: Right, this is to account for fields like the outer description in structures like this when looping through the structure:

```
"cows": {
      "type": "array",
      "description": "cows data",
      "properties": {
        "type": "object",
        "id": {
          "type": "number",
          "description": "Unique integer identifier.",
          "minimum": 1
        },
```
When looping through something like this, I decided to create an entry for `cows_properties_id` but not one for the outer array of `cows`, so in that case, I don't need the "type": "array" and "description": "cows data" and if it's not a dictionary, I chose to essentially ignore it.

KFosterReed:
This works exactly as described! I was able to generate the input csv as well as see the changes reflected in the csv when I changed the metadata properties json. Great work Niko and team!!

KFosterReed (approved):
Please update the Wiki when appropriate but other than that I am extremely happy with this!

RUFAS/input_manager.py
ew3361zh: These get triggered when the nested_value is not a dictionary but we need to take an action on whether to continue the inner loop based on what the type is for the outer loop value.

For example, when looping through this part of the metadata properties:

```
"animal_properties": {
    "herd_information": {
      "type": "object",
      "description": "Herd Demographics",
      "calf_num": {
        "type": "number",
        "description": "Number of Calves (head) -- The initial number of pre-weaned calves",
        "default": 8,
        "minimum": 0
      },
      "heiferI_num": {
        "type": "number",
        "description": "Number of HeiferI's (head) -- The initial number of heifers that are weaned but not yet bred",
        "default": 44,
        "minimum": 0
      },
```
The first loop through will see that the property_value (the value for "animal_properties" is a dict and will move into the inner loop to start looping through {"herd_information": {...}}.

The first iteration through the inner loop will find that the nested value (the value for "herd_information") is also a dict.

The first check in that inner loop is whether the nested_value has a "type" that is primitive ("bool", "string", "number") or if it's an array of any of these primitive types. In this case, that's not true because the "type": "object".

We now know that the "type" is "object" and that we need to recursively loop through deeper in that inner structure (the value for "herd_information").

So we send the value dict for "herd_information" back through the function, we see that the outer loop will check on the values for each of the elements in the "herd_information" dict. The first value is "object" and we don't need to go any further which is why there's no elif/else statement for the `if isinstance(property_value, dict):`.

The next value is "Herd Demographics". We don't care about this either.
The next value is the dict for "calf_num". We do care about this. Now the inner check on whether the value for any of the values in the `calf_num` dict is another dict.
At the first check, we know the nested value is not a dict (it's the string "number").
So this jumps down to the `elif self._check_property_type_primitive(property_value):` check. We check the property_value, which is the `calf_num` dict, for whether its type is any of those primitive types, and if it is, we can create a record for it and not loop through any further in the inner loop (the key-value pairs for the `calf_num` dict).

So in fewer words, those two elif statements are there to catch the recursive calls to the `_parse_metadata_properties()` and save some looping through each key-value pair of nested dictionaries but need to be at that level because they should only get triggered if the nested_value is not a dictionary. Hopefully this adds some context and explanation as to why it's set up the way it is.

PR #1556, opened 2024-04-18
Combine badge scripts and update readme badges

### LINKED ISSUE #1517: [GH Actions] Combine the two update-badge scripts
To simplify the code for github workflows even further, we can combine all the scripts for updating the badge together. This also helps with avoiding make more than one commit to the README page during a workflow run.

There is an active branch `combine-badge-scripts` that already combines the two mentioned scripts. Another simplifying step that has been done is to be able to call the script without passing in any argument.

This issue can follow after the update for the Black formatting action is completed.

### LINKED ISSUE #1554: [GH Actions] Fix the testing and linting badges on the README page
## Urgency Level

## Status

The two badges for testing and linting are not showing properly because they are not being updated somehow by the badge scripts.

## Intention

We want those badges to be functional.

## Request

- [ ] Make the testing badge show Pytest (maybe).
- [ ] Make the linting badge show Black formatter.

ltp23:
### Context

Issue(s) closed by this pull request: closes #1554, #1517 

- [ ] The [changelog](https://docs.google.com/spreadsheets/d/1U-3igxdstHTFb6k5dVcE5-x9IL0r6q18eUiwwytKZyA/edit#gid=0) has been updated. **Not needed for PRs on GH actions**

### What

There are two badges (Testing, Linting) on the readme page that are not working yet. Before, they were linked to their own yaml file but now those files are gone. I want to replace them with two other badges - Flake8 and Pytest. I also want to set the initial states of the badges to `unknown` and then update them at the end.

### Why

It looks more presentable to have all the badges working. In addition, the two update-badge scripts (test coverage and mypy) are fairly simple to combine so the badge update process can be streamlined further.

### How

- Combine the two scripts `update_badge.sh` and `update_mypy_badge.sh` into one.
- Set the colors and statuses for the flake8 and pytest badges in the yaml because the statuses are more accessible there.
- Let the flake8 and pytest errors temporarily slide so the respective badges can be updated before they are stopped.

### Test plan

- Add a flake8 error and a pytest error separately to see that the respective badges show `failed` and `passed` otherwise.
- Check to see if all the badges are set to the `unknown` state at the start of the workflow and then update them accordingly at the end of the workflow.

image


.github/workflows/combined_format_lint_test_mypy.yml
```diff
@@ -26,11 +30,37 @@ jobs:
...
+          mypy_badge_markdown_str="[![Mypy](${mypy_badge_url})]"
+          sed -i "s|\[\!\[Mypy\]\(.*\)\]|${mypy_badge_markdown_str}|" ./README.md
+
+      - name: Commit & Push Badge Updates
+        uses: EndBug/add-and-commit@v7
+        with:
+          message: "Set all badges to unknown at the start"
+          add: "README.md"
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          branch: ${{ github.head_ref }}
```
Isgotkowitz: Why do these need to be set to unknown at the start? If it is possible to update them to unknown or failing at the end based on the final results of the action, I think that should be done instead.

Isgotkowitz (approved):
Looks good, I think. The rich diff for the Readme looks a bit wonky, but I that is just GitHub being funky.

If possible, I think it would be good to avoid the extra resetting and committing of the README at the start of the workflow - but only if it is possible to do that without making things super complex.

ew3361zh (approved):
I agree with Ed on doing away with the extra update of the badges at the start if possible. That might shave some time off the total actions run. Tested it and it looked good though. Nice work, Loi!

allisterakun (approved):
Nice job, Loi! LGTM!
Also agreeing on Ed's point about the extra update of the badges at the start

ltp23:
Thanks guys. I will remove the initial update.

PR #1557, opened 2024-04-18
Create script to compare metadata properties

### LINKED ISSUE #1511: [IM] Add feature to compare two metadata properties
In a metadata, we reference a `properties` file. Here we want to add a function that makes it possible to compare two different properties files. Since these are `json` files, consider using existing tools and libraries to make it easier and faster.

ew3361zh:
This PR adds a new independent script to compare 2 metadata properties json files.

### Context

Issue(s) closed by this pull request: closes #1511

### What

Using the DeepDiff package, running this script takes in 2 file paths for metadata properties files and compares them.
The differences are saved in a .txt file in the output folder.

### Why

This was a feature requested by stakeholders/partners.

### How

DeepDiff is a robust and customizable package designed to compare different python data structures like dictionaries. I've used a few of the features including `verbose_level` and `ignore_order`. `verbose_level` controls how verbose the results are and at level 2 it seems to gather the level of depth we would need (values of properties changed along with keys). `ignore_order` is a nice feature that would help the comparison in cases where we've moved properties or groups of properties around but don't necessarily need to know that we've moved them around as long as they're still present.

`deepdiff` package is added to `requirements.txt`.

### Test plan

I have tested this by copying the `default.json` metadata properties file and making changes throughout and looking at the results. Unit testing has been added to cover the code in `compare_metadata_properties.py`.


Isgotkowitz (commented):
I think this is making good steps towards autogenerating a list of changed inputs. There are a couple things though that I think would make it easier to use:
- instead of calling the inputs file 1 and file 2, have them be named something like old file and new file. This makes it a lot easier to understand which inputs have actually been added and which have been removed.
- tweaking the output format to be a little less JSON-y. For example, a current output is this:
```
Comparing: 'old_properties.json' to 'input/metadata/properties/default.json'
{ 'dictionary_item_added': { "root['config_properties']['random_seed']['maximum']": 4294967295},
  'values_changed': { "root['animal_properties']['pen_information']['properties']['max_stocking_density']['maximum']": { 'new_value': 3,
                                                                                                                         'old_value': 2}}}
```
It would help the readability a lot if it looked more like this:
```
Comparing changes going from 'old_properties.json' to 'input/metadata/properties/default.json'

Items added: 
"root['config_properties']['random_seed']['maximum']": 4294967295},

Values changed:
"root['animal_properties']['pen_information']['properties']['max_stocking_density']['maximum']": {'new_value': 3, 'old_value': 2}
```
Basically the less indentation the better

ew3361zh:
> * instead of calling the inputs file 1 and file 2, have them be named something like old file and new file. This makes it a lot easier to understand which inputs have actually been added and which have been removed.

For me having the time of the file determine which placement it has in the command adds a layer of confusion. My preference would be to have the user determine their starting place and then what they'd like to compare that to. I'd be curious what @allisterakun thinks though. It could just be that my brain doesn't like it but it makes more sense to do it this way to everyone else.

> * tweaking the output format to be a little less JSON-y.

Can definitely work on this, thanks!

Isgotkowitz (approved):
Nice job Niko, to me this strikes a good balance between readability and precision in listing the changes. Two suggestions to consider before approving:

1. Having someone who isn't a dev try this out. I thought this was put into a pretty readable format, but it would be good to have someone who is not as familiar with JSON and the metadata look at the outputs too.
2. Write the items removed in addition to the ones changed and added.

allisterakun (approved):
The comparison script looks good! Excellent job, Niko!

Out of curiosity: why are there changes to the CS module code and input files?  Is this due to GitHub merge issues?

ew3361zh:
> Out of curiosity: why are there changes to the CS module code and input files? Is this due to GitHub merge issues?

That is exactly what's going on. Hopefully we get out of this cycle of unwanted commits soon. I think once this batch of PRs created during this last sprint are all closed and merged in we should break it.

PR #1570, opened 2024-04-21
Add CIDR count to heiferII

### LINKED ISSUE #1217: Add CIDR_count to heiferII
LifeCycleManager method `_extract_repro_stats_from_heiferII` includes placeholder (e.g. commented out) logic that would allow CIDR count tracking for Heifer II objects. However, this needs to be added to HeiferII functionality.

Likely involves reviewing CIDR_injection history on heiferII objects, or counting from their event logs.

0luchi:
### Context

Issue(s) closed by this pull request: closes #1217

### What
This PR adds `CIDR_count` to `heiferII.py` by adding it as a property and updating it when there is a `CIDR_injection`

### Why
This is important because it enables `CIDR_count` tracking for `lifecycle_manager.py` 

### How
I set `CIDR_injections` equal to `CIDR_count` as I assumed that every injection increments the CIDR level by 1 from [Repreoduction Refactoring design doc](https://docs.google.com/document/d/1Uro29xXOIfrNWv3jg5uy3FuC1FEQR3CGQaHnoyuKF8Q/edit)

### Test plan
I uncommented existing the existing test case in `test_animal_life_cycle.py` to confirm the functionality


RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -652,7 +652,7 @@ def _extract_repro_stats_from_heiferII(self, heiferII: HeiferII) -> None:
             heiferII: The heiferII to extract the stats from.
 
         """
-        # self.CIDR_count += heiferII.CIDR_count  # TODO: add CIDR count to heiferII GitHub Issue #1217
+        self.CIDR_count += heiferII.CIDR_count
```
JoeWaddell, line 655: I think this should be a new attribute for a `LifeCycle` object, `CIDR_count_h` (in addition to the existing `CIDR_count`), to fit alongside the others below. The `_h` indicates that it's the count for heifers.

So here instead of assigning the value in `CIDR_count` to `CIDR_count_h`, it would add the value in `heiferII.CIDR_injections`.

Then in  `_extract_repro_stats_from_cow` you'd need to add similar logic (operating on a `cow` object, and assigning them to the `CIDR_count` attribute in LifeCycle)

RUFAS/routines/animal/life_cycle/heiferII.py
```diff
@@ -103,6 +103,7 @@ def __init__(self, args):
         self.GnRH_injections = 0
         self.PGF_injections = 0
         self.CIDR_injections = 0
+        self.CIDR_count = self.CIDR_injections
```
JoeWaddell, line 106: I think this should just be initalized to 0, just to avoid confusion.

JoeWaddell (changes requested):
This is a tricky one to wrap my head around, and I think I wrote the issue as far simpler than it really is, my mistake! So I left a couple comments that hopefully make sense. I think it just needs to mimic the logic for the other injections, since `_execute_hormone_delivery_schedule` in `cow.py` uses the heiferII `_execute_hormone_delivery_schedule`...it's a little confusing, so don't take my suggestions as infallible: we can always ask Loi to help if we need to!

KFosterReed:
@JoeWaddell can you confirm if this branch, PR and the related issue can be deleted?

JoeWaddell:
> @JoeWaddell can you confirm if this branch, PR and the related issue can be deleted?

Confirmed! This functionality was added in the Refresh, for future reference see `_update_heifer_reproduction_statistics` in `herd_manager.py`, and the resetting of injections in `reset_daily_statistics` in `reproduction.py`, and also resetting of the CIDR_count in `herd_statistics.py`

PR #1571, opened 2024-04-22
Fixed recording of phosphorus from surface soil layer

### LINKED ISSUE #1566: Available Phosphorus Pool needs to be reset to 0 after "watering" (rainfall or irrigation)
There are two pools of phosphorus on the soil surface (available and recalcitrant) and phosphorus doesn't appear to be moving out of them quite properly into the soil. 

## Expected Behavior
There are two pools of phosphorus on the soil surface (available and recalcitrant. When phosphorus is applied 75% should go to the available pool (this is happening fertilizer.py line 164) and 25% goes to the recalcitrant pool. When it rains the FIRST time, all of the available_phosphorus should be released into the soil and the available_phosphorus pool should go to 0 (because all of it goes to the soil). In all subsequent rainfalls, the recalcitrant_phosphorus pool is the one that is impacted. The second rainfall 40% of the recalictrant_phosphorus goes into the soil and in all subsequent rainfalls 7.5% of the remaining phosphorus in the recalcitrant pool should go into the soil. 

## Current Behavior
Currently, it seems we don't set the available_phosphorus_pool to 0 when we move the phosphorus from this pool to the soil. We put phosphorus in the soil, but it never leaves the available_phosphorus_pool and is resulting in a lot of runoff. 

It also appears that while 

## Possible Solution
Adding a reset of the available_phosphorus_pool to 0 when a rainfall/irrigation event occurs would be great. During rainfall events only the recalcitrant phosphorus pool should be the phosphorus that is accessible to becoming solubilized. It seems like we just have a generic "phosphorus_pool" instead of just the recalcitrant_phosphorus_pool that can become solubilized.

The second piece, I can't seem to identify in the code at all is that for the partitioning of the recalcitrant pool to the soil in subsequent rainfall events. I think we need to add this partitioning, 40% on the second rainfall following application and 7.5% for every additional rainfall event. 

Lastly, irrigation needs to be considered a "rainfall" event, not just true precipitation that falls from the sky. 

## Steps to Reproduce
I'm working through c1f2_wcrops_metadata when I found this error. 

## Context (Environment)
My phosphorus runoff is 500x what it should be! 

## Detailed Description
I'm proposing we 1) broaden rainfall to be irrigation or rainfall 2) "zero" the available_phosphorus_pool after the first rainfall/irrigation event as all of it should go to the soil layers below the surface and 3) partition the recalcitrant_phosphorus_pool for each rainfall/irrigation event following the first one such that the size of that pool decreases and less is available for solubilization and runoff. 

## Possible Implementation
I think these items will occur in the fertilizer.py file 

## Target Deadline
Monday, around noontime, if possible.

Isgotkowitz:
Fixes the runoff phosphorus amount being recorded by the field size (converting kg/ha to kg when it should not have been). This change has been verified by checking both the docstrings (and equations) of the methods and variables that calculate and record phosphorus runoff.

### Context

Issue(s) closed by this pull request: closes #1566 

### Test plan

This change has been tested with the C1F2 pilot farm simulations, resulting in a massive decrease in the recorded amounts of phosphorus runoff. Also, unit tests have been corrected.

### DIFF
--- a/RUFAS/routines/field/soil/phosphorus_cycling/soluble_phosphorus.py
+++ b/RUFAS/routines/field/soil/phosphorus_cycling/soluble_phosphorus.py
@@ -76,8 +76,8 @@ class SolublePhosphorus:
                 self.data.soil_layers[0].layer_thickness,
             )
             self.data.soil_layers[0].labile_inorganic_phosphorus_content -= phosphorus_runoff
-            self.data.soil_phosphorus_runoff = phosphorus_runoff * field_size
-            self.data.annual_soil_phosphorus_runoff += phosphorus_runoff * field_size
+            self.data.soil_phosphorus_runoff = phosphorus_runoff
+            self.data.annual_soil_phosphorus_runoff += phosphorus_runoff
 
         for layer_index in range(len(self.data.soil_layers)):
             current_layer = self.data.soil_layers[layer_index]
--- a/tests/soil_crop_tests/soil_tests/phosphorus_cycling_tests/test_soluble_phosphorus.py
+++ b/tests/soil_crop_tests/soil_tests/phosphorus_cycling_tests/test_soluble_phosphorus.py
@@ -203,8 +203,8 @@ def test_daily_update_routine(runoff: float, field_size: float) -> None:
     if runoff > 0:
         incorp._determine_phosphorus_runoff_from_top_soil.assert_called_once_with(runoff, field_size, 3.4, 1.4, 20)
         assert incorp.data.soil_layers[0].labile_inorganic_phosphorus_content == (3.4 - 0.9 - 1.2)
-        assert incorp.data.annual_soil_phosphorus_runoff == (0.9 * field_size)
-        assert incorp.data.soil_phosphorus_runoff == (0.9 * field_size)
+        assert incorp.data.annual_soil_phosphorus_runoff == 0.9
+        assert incorp.data.soil_phosphorus_runoff == 0.9
     else:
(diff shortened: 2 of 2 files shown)


krb78 (approved):
good eyes on catching this one, Ed 👀

PooyaHekmati:
I'm confused, the chang log says the units of have changed, but I don't see any changes in the units in the code here.

krb78:
true, units in the code haven't changed. We had inadvertently multiplied total kg phosphorus applied by field size, so the output was not kg/ha as the units are intended to be. This fixes that so we are actually reporting the units we claim to be reporting :)

ew3361zh (approved):
LGTM

PooyaHekmati:
> true, units in the code haven't changed. We had inadvertently multiplied total kg phosphorus applied by field size, so the output was not kg/ha as the units are intended to be. This fixes that so we are actually reporting the units we claim to be reporting :)

Thanks for the explanation, I changed the changelog to: `Corrected recording of phosphorus runoff from top soil layer from kg to kg/ha.`

Isgotkowitz:
> > true, units in the code haven't changed. We had inadvertently multiplied total kg phosphorus applied by field size, so the output was not kg/ha as the units are intended to be. This fixes that so we are actually reporting the units we claim to be reporting :)
> 
> Thanks for the explanation, I changed the changelog to: `Corrected recording of phosphorus runoff from top soil layer from kg to kg/ha.`

Thank you Pooya

PR #1572, opened 2024-04-22
Creating manure for field applications when no manure is simulated.

### LINKED ISSUE #1521: [CS][Pilot Testing] Create manure applications when no animals are simulated.
When animals are not simulated, there is no manure produced and thus no manure available for field applications. This means that to properly simulate manure applications to fields, animals need to be simulated - which significantly increases the runtime and overhead when a user only wants to model fields.

The Crop and Soil module checks whether or not animals are being simulated. If they are, Crop and Soil requests manure from the Manure module (like it currently does). If they are not, then it constructs a manure application based on the user-configured manure application request (i.e. creates manure out of thin air).

When creating manure applications, ratios of dry matter to total mass, N to dry matter mass, etc can be pulled from U of M's [ManureDB](http://manuredb.umn.edu/?a=Dairy) - but ratios should be double checked by SMEs.

Instead of having the Crop and Soil module create manure out of thin air, the logic for checking whether or not animals are simulated and creating manure applications could be located in the Manure module.

**Additional context**
This feature was requested by @krb78 who is currently simulating fields for pilot testing, and this feature would be a major help for her. She should be consulted during the implementation and review process.

A very temporary version of this feature was implemented a while back during the evaluation of the Crop and Soil module. See the method `_construct_evaluation_manure_application` in `field.py` in commit 6daa1218afaa3b701c35399ae5d88dc44c69d09d to see what it looked like. It may be helpful to look at this feature.

Isgotkowitz:
Manure applications are now created and applied to fields, even when no animals/manure production is simulated.

### Context

Issue(s) closed by this pull request: closes #1521 

### What

Previously, manure applications to fields could only be simulated if there were animals being simulated. This PR changes that, so if no animals are simulated then the `FieldManureSupplier` will be created and used to generate manure applications based on the requested amounts of nitrogen and phosphorus.

### Why

This will significantly enhance the ability for pilot testers to evaluate RuFaS effectively, by allowing fast simulations of field outcomes without losing simulated manure applications.

### How

The class `FieldManureSupplier` has been created. This is a drop-in replacement for the `ManureManager` in `Field` instances, from which manure is requested. It only has one method, `request_nutrients`, which has the same interface as `ManureManager.request_manure`. It takes in a request for manure, calculates the amount of manure that would be need to meet the requested nitrogen, then the requested phosphorus, and picks the smaller amount of manure. It packages the selected manure amount with the appropriate amounts of nutrients and sends it out to be applied to a field.

When `FieldManager` is initialized, it looks at whether animals are being simulated. If so, then it selects the `ManureManager` as the manure supplier. If it doesn't, then it instantiates the `FieldManureSupplier` and uses that.

### Test plan

Unit tests have been written to ensure that `FieldManager` correctly selects the right manure supplier, and that `FieldManureSupplier` properly creates both liquid and solid manure applications.

All unit tests that have broke have been fixed.

Also, changing `simulate_animals` to False in various scenarios and checking that manure applications are actually being created and used.


krb78 (approved):
I had to do a double take this worked so well I wasn't sure I'd actually turned simulate_animals "off" 😆 Nice work, Ed. It works beautifully and delivers solid and liquid manure as expected.

RUFAS/routines/field/manager/field_manure_supplier.py
```diff
@@ -0,0 +1,109 @@
...
+The Field Manure Supplier uses ratios of dry mass to wet mass, Nitrogen to dry mass, and Phosphorus to dry mass to
+project total amounts of manure applied. These ratios were derived from the ManureDB
+(http://manuredb.umn.edu/?a=Dairy&y=2024&y=2023&y=2022&y=2021&y=2020) for the year of 2023. The liquid ratios were
+calculated using the data collected for manure designated as "Slurry" and "Liquid", and the solid ratios for manure
+designated as "Solid".
+"""
+
+
+"""Factor for converting dry matter mass of liquid manure to wet mass."""
+LIQUID_MANURE_DRY_MASS_TO_WET_MASS = 21.739
```
ltp23, line 17: Is this a good place to put this and some other constants? I don't know if you have some files dedicated to storing the constants.

RUFAS/routines/field/manager/field_manure_supplier.py
```diff
@@ -0,0 +1,109 @@
...
+
+        Notes
+        -----
+        This method calculates the total mass of manure that would be applied for each of the requested nutrients, then
+        selects the smallest mass and uses it to construct the amount of manure that is actually returned. If one of the
+        requested nutrient masses is 0, that nutrient is not considered when formulating the manure result.
+
+        """
+        type_to_constants_map = {
+            ManureType.LIQUID: {
```
ltp23: This map can be converted to a class variable so it doesn't need to be recreated every time this method is called. This reminds me to do something similar in the cow preg check method.

RUFAS/routines/field/manager/field_manager.py
```diff
@@ -90,16 +93,43 @@ def annual_update_routine(self) -> None:
...
+
+        Returns
+        -------
+        ManureManager | FieldManureSupplier
+            The Manure Manager if animals are to be simulated, otherwise a Field Manure Supplier instance.
+
+        """
+        info_map = {"class": self.__class__.__name__, "function": self._get_manure_supplier.__name__}
+
+        animals_simulated = im.get_data("config.simulate_animals")
```
ltp23, line 109: Can you check to see how often this gets called? If there are too many, we would need to stash it somewhere.

ltp23 (approved):
The PR looks good to me. Thanks Ed!

RUFAS/routines/field/manager/field_manure_supplier.py
Isgotkowitz: I think so, because there isn't any other module that currently needs these constants. But if another module comes along that needs them, then they should/would be moved to the crop and soil constants file.

RUFAS/routines/field/manager/field_manure_supplier.py
Isgotkowitz: Good idea!

RUFAS/routines/field/manager/field_manager.py
Isgotkowitz: Just the once, though the `config` input is queried ~ a dozen times, which shouldn't hamper performance much but still isn't ideal.

Isgotkowitz:
@krb78 I just tried running the default scenario with and without animals, and found some interesting and potentially notable differences in the manure applications:

With animals simulated:
|Year|Day|Dry Matter Mass|Dry Matter Fraction|N|P|
|-----|----|------------------|---------------------|--|--|
|327|2009|            4082              |            0.083                  |  355 |    11|
|304|2012|            1982              |                0.0854              |  163 |  5.44  |
|324|2013|               2997           |                   0.085           |  244 | 8.195   |

Without animals simulated
|Year|Day|Dry Matter Mass|Dry Matter Fraction|N|P|
|-----|----|------------------|---------------------|--|--|
|327|2009|            3066.66|0.046                |  146.66 |    60|
|304|2012|            1482.21|                0.046                |  70.88 |  28.9  |
|324|2013|               2197.7|                   0.046                |  105.1 | 42.9   |

None of the differences seem too problematic to me, as long as there is a way to get some extra nitrogen into the soil when animals aren't simulated.

krb78:
Hi Ed! Interesting finding here - I'm actually going to flag this for Julie (@jadamchick) because I know this week she's starting to look into N:P:K ratios in manure coming out of the animal module and manure module. I'm actually more intrigued by the default scenario setting - I'm thinking perhaps separated liquid was added to the field because those would have a lot lower P than regular manure would. It's just VERY strange to me that there would be any dairy manure that would be 30x the amount of N to P without having had some manure separation/nutrient partitioning treatment occurring (and I know the default scenario does have a digester and a SLS in at least one of the manure scenarios). The strange thing about it being separated liquid is that the DM is lower... oof another brain buster haha. Well, I'll just say that I think the manureDB numbers are much more realistic in a typical unseparated scenario and the bottom scenario is exactly the reason lots of farmers purchase synthetic N fertilizer because we reach P caps before we have enough N for our corn, but certainly cause for pause, though not for extreme worry!

Issue #589, opened 2023-06-21
Improve lactation reduction method


JoeWaddell:
## Context and the problem(s) ##
When a pen's mean requirements can't be met by a ration optimization attempt, lactation is reduced for all animals in the pen. 
The goal is to lower requirements such that a ration can be formulated that meets said requirements.

#### Problem 1: Overshooting necessary reduction ###
If the lactation reduction step is too large, we can overshoot the reduction. This means the drop in lactation will be more severe that required.

#### Problem 2: Underestimating the necessary reduction  ###
Smaller steps mean we can more accurately reach the balance of requirements and ration nutrition (e.g. smaller incremental changes mean we overshoot by a much smaller margin). 
However, small steps mean more optimization attempts, which means more computational time. 

# What we need #
Some method should be implemented to estimate the total reduction required, and get close to the requirement/nutrition balance: and then make small steps until requirements are satisfied. 

## Current implementation(s) ##
This logic is roughly in place in the code already. It's currently in ration_driver.py, in ration_formulation

NEl_con = net energy for lactation constraint, e.g. the restriction placed on the optimization method in ration_NLP.py, and can be used (perhaps) as a proxy for the nutritional/requirement deficit. However, these specific values have been shown to be overly aggressive in some circumstances. 

`if NEl_con < -0.5:`
`                reduction = 3 * (-NEl_con)`
`           else:`
`                reduction = 1.5`

tomhuhh:
Hello @JoeWaddell! Do you consider this issue to be closed? Thanks!

JoeWaddell:
This one I think is still unresolved, because we could implement a better way to estimate the reduction 'step', based on an animal's requirements and deficit. It's more of a "wish list" type issue, for future reference.

PR #1573, opened 2024-04-22
Light_ration_refactor_2

JoeWaddell:
This PR introduces the `handle_failed_constraints()` function to `ration_driver.py` as well as an accompanying unit test. This is the first step in a ration submodule refactor within the Animal Module. An accompanying unit test was added, and the files `ration_driver.py` and `ration_optimizer.py` were cleaned up as well.

### Context
- [x] Removed `: Any` typing for items that otherwise cause circular imports (set to `Any` for now to reduce visual messiness in VS Code).
 
### What
The function `handle_failed_constraints()` was added to reduce the lengths of `formulate_ration()` and `get_user_defined_ration()`, since both functions have to handle the case where certain constraints are failed. This function, in the case that certain constraints are violated, logs a failure summary to the output manager recording the failure.

### Why
This PR was introduced to reduce the complexity of methods within `ration_driver.py`, as well as to improve readability within the Python file. Additionally, some code was cleaned up to follow best Python practices, resulting in the shortening of unnecessarily long code.

### How
This function extracts common logic into the new function `handle_failed_constraints()`, and removes unnecessary code throughout the ration submodule of the Animal Module, following best Python practices.

### Test plan
- [x] Updated affected unit tests, added tests for new methods

A unit test was added for the new function `handle_failed_constraints()`. Also, some other unit tests were affected in the 
 code cleanup process, but all tests still pass.


RUFAS/routines/animal/ration/ration_optimizer.py
```diff
@@ -947,7 +947,10 @@ def optimize(
             bnds = []
             bnds = [(0, (lim / 3) + 0.0001) for lim in ration_config.feed_limit_list]
 
-        if str(animal_combination) in ["AnimalCombination.LAC_COW"]:
+        print(animal_combination)
```
jadamchick: will want to take these print statements out?

jadamchick (commented):
Looking good! I like the direction this is headed. Some initial questions.. 

I ran this with the c2f3 ration under 2 scenarios: a) set to user_input = FALSE and b) user_input = TRUE with 0.20 tolerance and 5kg max milk reduction. 

In the first scenario (user_input = FALSE), it did not report any failed constraint attempts. 
In the second, I've attached the output for you to look at. 
[c2f3_UDR_.2_5_july16_saved_variables_csv_feed_emissions.txt_16-Jul-2024_Tue_14-48-31.csv](https://github.com/user-attachments/files/16284915/c2f3_UDR_.2_5_july16_saved_variables_csv_feed_emissions.txt_16-Jul-2024_Tue_14-48-31.csv)

A few things I noted: 
- it reports the handle_failed_constraints whether or not it actually failed (i.e. even for num_reattempts=0)
- for pen 2, there was a single instance where it says that there was a failed constraint (calcium). Should I assume that it reformulated the ration once, and the second attempt (first reattempt) was "successful"? That was my initial assumption, but I'm not sure given that: a) the ration reported for that interval is identical to the ration attempted in the failed_constraints column, and b) in that case I would maybe expect another line in the handle_failed_constraints report for reattempt=1 with failed_constraints = 0? 
- for pen 3, at every interval, it failed 20 reattempts, which looks to be when it reached the max milk reduction (always somewhere around 4.9kg). When I looked at the ration reported for the first interval, it doesn't line up with any of the attempted rations listed in the failed_constraints report, so I wondered if maybe it defaulted to the fixed user ration (though I didn't verify that).. but also see that it reports milk reduction of -4.9kg also. So I'm not sure. 


So, summary questions/comments: 
- it would be nice if the behavior of including/not including this report when there are no reattempts is consistent across UDR=T and UDR=F scenarios 
- it would be nice to know what is the outcome/resolution if there are failed constraints even at the last reattempt listed for a given interval (eg the failed calcium constraint instance for pen 2, and after 20 failed reattempts [to the max milk reduction] for pen 3)

[](url)

JoeWaddell:
> In the first scenario (user_input = FALSE), it did not report any failed constraint attempts. In the second, I've attached the output for you to look at. [c2f3_UDR_.2_5_july16_saved_variables_csv_feed_emissions.txt_16-Jul-2024_Tue_14-48-31.csv](https://github.com/user-attachments/files/16284915/c2f3_UDR_.2_5_july16_saved_variables_csv_feed_emissions.txt_16-Jul-2024_Tue_14-48-31.csv)

I love your thorough reviews/testing!

> A few things I noted:
> 
>     * it reports the handle_failed_constraints whether or not it actually failed (i.e. even for num_reattempts=0)

This should be an easy fix - the order of operations here is a little screwy, and the naming is a little confusing. I'll improve things, and when reviewing documentation of how ration formulation works, I expect to improve clarity as well.

The thinking here was that it would report the first failed ration, wherein it hasn't yet reattempted to formulate a ration. What we could do is rename `num_reattempts` to simply `attempt_number`, so instead of starting with 0, we start with 1. E.g., if it fails just once, it'll report `attempt 1`, instead of `num_reattempts 0`. Does that sound more intuitive?

>     * for pen 2, there was a single instance where it says that there was a failed constraint (calcium). Should I assume that it reformulated the ration once, and the second attempt (first reattempt) was "successful"? That was my initial assumption, but I'm not sure given that: a) the ration reported for that interval is identical to the ration attempted in the failed_constraints column, and b) in that case I would maybe expect another line in the handle_failed_constraints report for reattempt=1 with failed_constraints = 0?



Since formulation should only be reattempted if the requirements can be dropped - which is currently restricted to reducing milk production in lactating cows - other animal classes simply fall back on using their last formulated ration. So if it fails, it still logs it, but uses the previous month's ration, assuming it's close.

However, A) this isn't ideal, and B) that is impossible on the first month - e.g. if you can't formulate even a single ration for a non lactating cow pen. I opened and issue about this a while back (See https://github.com/RuminantFarmSystems/MASM/issues/577), so it's definitely something we need to address. In the PR to fix the NRC library, I also added in an error so the simulation doesn't get fail if this unlikely (but possible) condition is met: https://github.com/RuminantFarmSystems/MASM/pull/1835

>     * for pen 3, at every interval, it failed 20 reattempts, which looks to be when it reached the max milk reduction (always somewhere around 4.9kg). When I looked at the ration reported for the first interval, it doesn't line up with any of the attempted rations listed in the failed_constraints report, so I wondered if maybe it defaulted to the fixed user ration (though I didn't verify that).. but also see that it reports milk reduction of -4.9kg also. So I'm not sure.

It does - it will default to the proportions in the user defined ration, using the reduced animal requirements (e.g. lactation reduction). Right now we're reducing lactation in increments of half a kilo, which is why in that case it's approx 4.9 for the pen - it won't go beyond the reduction maximum, so it'll never quite be precisely the value set by the user.

We could certainly reduce the reduction increment so it's closer, but that will mean longer simulation times if reduction is required. We have an issue to see if there's a better solution to skip the intermediate reductions, e.g. predicting the amount needed to meet requirements: https://github.com/RuminantFarmSystems/MASM/issues/589

> So, summary questions/comments:
> 
>     * it would be nice if the behavior of including/not including this report when there are no reattempts is consistent across UDR=T and UDR=F scenarios
> 
>     * it would be nice to know what is the outcome/resolution if there are failed constraints even at the last reattempt listed for a given interval (eg the failed calcium constraint instance for pen 2, and after 20 failed reattempts [to the max milk reduction] for pen 3)

This should be the case already, but I'll triple-check it - I think it goes back to the numbering and nomenclature of reattempts v/ attempts.

jadamchick:
Thanks Joe. 
My impression from the outputs of the UDR=True run is that it was reporting something in the constraints_failed report even when there was no failure/reformulation -- mainly based on the fact that it happened at every single ration interval, always with reattempt_number = 0, and always with an empty constraints_failed_dict [].  (Eg see columns BU, BV, BW in the csv above)
This did not occur in the UDR=False run. 

I think most of the other questions & your responses make sense, and mostly boil down to having some good documentation so the user knows what type of logic / flow chart is being followed :) I know that is in the works! 
I don't think the reduction increment needs to be less, 0.5 kg makes sense to me. 

Let me know if you want to chat more on this!

JoeWaddell:
> Thanks Joe. My impression from the outputs of the UDR=True run is that it was reporting something in the constraints_failed report even when there was no failure/reformulation -- mainly based on the fact that it happened at every single ration interval, always with reattempt_number = 0, and always with an empty constraints_failed_dict []. (Eg see columns BU, BV, BW in the csv above) This did not occur in the UDR=False run.

Thanks for this: I didn't realize the behavior was slightly different with UDR=True and False! I'll update the code such that: A) it won't report failed constraints if it did not in fact fail (as is happening when UDR=False), and B) Will change the name from `reattempt` to `attempt`, to ensure the numbering is more intuitive. 

In addition I'll write up a flow chart and present that here, since it should be useful for the ongoing documentation updates.

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -56,95 +60,149 @@ def formulate_ration(
...
+            },
+        }
+        om.add_variable(
+            f"failed_constraint_summary_for_pen_{pen.id}",
+            fail_summary,
+            dict(info_map, **{"units": fail_summary_units}),
+        )
 
     @staticmethod
     def calc_milk_average(pen) -> float:
```
allisterakun:
```suggestion
    def calc_milk_average(pen: Pen) -> float:
```

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -163,8 +221,7 @@ def calc_milk_average(pen) -> float:
     def reduce_milk_production(cls, pen, reduction: float) -> None:
```
allisterakun:
```suggestion
    def reduce_milk_production(cls, pen: Pen, reduction: float) -> None:
```

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -1003,10 +994,11 @@ def get_forage_NDF(
...
 
         """
+        forage_NDF_item = 0.0
+
         if feed_item_info["feed_type"] == "Forage":
             forage_NDF_item = feed_item_info["NDF"] * kg_fed
-        else:
-            forage_NDF_item = 0.0
+
         return forage_NDF_item
```
allisterakun:
```suggestion
        return feed_item_info["NDF"] * kg_fed if feed_item_info["feed_type"] == "Forage" else 0
```
Since the logic here is so simple, what do you think of using the short notation?

allisterakun (approved):
Great work, Joe! Thank you for the type hints additions! 
This PR looks great! Just left some suggestions!

JoeWaddell:
The promised flowchart! Still in progress, but should reflect what's in the code at a high level. Also should clarify when and why lactation reduction happens (or doesn't), when things are being reported, and how we might want to add to the reduction/reporting methods (e.g. adding a lactation reduction max to the automated formulation...or at least some minimum average production in the pen that should raise an error or warning, see https://github.com/RuminantFarmSystems/MASM/issues/1859).

ration_formulation

I'll upload to basecamp in its raw .svg format once I get some feedback, or as folks request!

RUFAS/routines/animal/ration/ration_driver.py
JoeWaddell: Good improvement!

elle-andreen (approved):
All looks good here to me, the behavior I'm seeing in the reports matches what's described in the PR. I was able to generate formulation failures in both automatically formulated rations and user-defined rations, and the constraint failure reports were included in both as expected. The contents of the failed constraint output are sufficient and useful in identifying where the ration could be modified to reduce/prevent production reductions. Thanks for your work on this Joe!

jadamchick (approved):
Looks great Joe! Thanks for the flow chart, that is incredibly helpful. 
I re-ran the same UDR farm as last time, and now there is no failed constraint report with empty values for pens 1 or 2, so that is good. 

Two minor comments: 
* The naming update: if starting at 1 instead of 0, should it be labeled as "attempt" instead of "reattempt"? i.e. if it only fails once, the failure report is technically describing 'attempt 1' rather than 'reattempt 1', I think? 
* I had the following error. It doesn't seem to have impacted the report, but FYI: 
image

PR #1574, opened 2024-04-23
Quick fix for missing body weight history in ration driver

### LINKED ISSUE #1564: 'List index out of range' error ends simulation for some stall # / stocking density scenarios
## Current Behavior
The attached error occurs for some but not all simulations when stall numbers and stocking density are low enough to require new pens to be created on the fly.  See below for a specific metadata + seed combo. 

It ends the simulation and does **not** continue to the next simulation listed in scenario manager. 

[for error_c2f7_pt branch_19apr_errors_19-Apr-2024_Fri_11-12-49.json](https://github.com/RuminantFarmSystems/MASM/files/15042406/for.error_c2f7_pt.branch_19apr_errors_19-Apr-2024_Fri_11-12-49.json)

## Steps to Reproduce

1. Branch pilot_testing_Dec2023, metadata for c2f7
2. Make the following changes to c2f7_animal.json: 
-  Calf_num: 4
- Pen 1: number_of_stalls: 45; max_stocking_density: 1.2
- Pen 2: max_stocking_density: 1.2
- Pen 3: number_of_stalls: 45; max_stocking_density: 1.2 

3. Make the following changes to c2f7_config.json: Set seed: true; random_seed: 4195961855

## Context (Environment)

Though it only seems to happen in very specific/sporadic scenarios, it is challenging because it ends all simulations. I seem to have been able to avoid it by increasing stall number and/or stocking density in those cases, but haven't investigated the root cause.

ltp23:
### Context

Issue(s) closed by this pull request: closes #1564 

### How

- In `ration_driver`, when reporting failed constraints, if body weight history is missing, I set the simulation day to -1. I'm open to different alternatives. I also denested the `pen_requirements` key in the `fail_summary_units` by removing it.
- In `animal_module_reporter`, when padding data, there needs to be an exact mapping between the units and the values.

### DIFF
--- a/RUFAS/routines/animal/animal_module_reporter.py
+++ b/RUFAS/routines/animal/animal_module_reporter.py
@@ -69,7 +69,7 @@ class AnimalModuleReporter:
                 current_output_length = 0
             length_difference = len(list(om.variables_pool[reference_variable].values())[0]) - current_output_length
             if length_difference > 1:
-                short_variable_to_add = full_variable_to_add[full_variable_to_add.rfind(".") + 1 :]
+                short_variable_to_add = full_variable_to_add[full_variable_to_add.rfind(".") + 1:]
                 for _ in range(0, length_difference - 1):
                     om.add_variable(
                         short_variable_to_add,
@@ -238,7 +238,7 @@ class AnimalModuleReporter:
             AnimalModuleReporter.data_padder(
                 f"{classname}.{funcname}.ration_nutrient_amount_pen_0_CALF",
                 f"{classname}.{funcname}.ration_nutrient_amount_pen_{pen.id}_{pen.animal_combination.name}",
-                {},
+                {variable: 0 for variable in nutrient_amount.keys()},
                 simulation_day,
                 info_map,
                 nutrient_amount_units,
@@ -277,7 +277,7 @@ class AnimalModuleReporter:
             AnimalModuleReporter.data_padder(
                 f"{classname}.{funcname}.avg_rqmts_pen_0_CALF",
                 f"{classname}.{funcname}.avg_rqmts_pen_{pen.id}_{pen.animal_combination.name}",
-                {},
+                {variable: 0 for variable in pen.avg_nutrient_rqmts.keys()},
                 simulation_day,
                 info_map,
                 avg_nutrient_rqmts_units,
@@ -291,7 +291,7 @@ class AnimalModuleReporter:
             AnimalModuleReporter.data_padder(
                 f"{classname}.{funcname}.ration_per_animal_for_pen_0_CALF",
                 f"{classname}.{funcname}.ration_per_animal_for_pen_{pen.id}_{pen.animal_combination.name}",
-                {},
+                {variable: 0 for variable in ration_per_animal.keys()},
                 simulation_day,
                 info_map,
                 ration_per_animal_units,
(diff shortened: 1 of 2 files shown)


jadamchick (approved):
Thanks Loi - this runs smoothly for me and seems to avoid the error. I don't have much insight into the approach to the fix, so will let the SWE's weigh in on the approach. 
Also reminder to update the changelog :)

JoeWaddell:
Were you able to find out the reason some animals were lacking a body weight history? Was it just animals that were purchased on the same day as this check, or at the initialization of the sim?

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -238,7 +238,7 @@ def report_ration_interval_data(cls, pen_list: List[Pen], feed: Feed, simulation
             AnimalModuleReporter.data_padder(
                 f"{classname}.{funcname}.ration_nutrient_amount_pen_0_CALF",
                 f"{classname}.{funcname}.ration_nutrient_amount_pen_{pen.id}_{pen.animal_combination.name}",
-                {},
+                {variable: 0 for variable in nutrient_amount.keys()},
```
Isgotkowitz, line 241:
```suggestion
                {key: val for key, val in nutrient_amount.items()},
```
I think this would be more correct than adding in 0. Same goes for all the other instances of adding in values.

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -283,33 +283,36 @@ def get_user_defined_ration(  # noqa
...
+            "avg_milk_production_reduction_pen": MeasurementUnits.KILOGRAMS.value,
         }
         if failed_constraints is not None:
             for constr in failed_constraints:
                 constraints_failed_list.append(constr["fun"].__name__)
             animal_list = list(pen.animals_in_pen.values())
+            if not animal_list or not animal_list[0].body_weight_history:
+                simulation_day = -1
+            else:
+                simulation_day = animal_list[0].body_weight_history[-1].simulation_day
```
Isgotkowitz, line 308: Why is this only checked for the first animal in the animal list? What if another animal in the list does have a bodyweight history?

Also, `simulation_day` is a recording of when this ration failure is happening, so it would definitely be incorrect to say that it is failing on day -1. It would be better to pass `Time` in and use the simulation day in there.

Isgotkowitz (changes requested):
It is good that this issue has a solution, but I think it would be better to find a more permanent solution. Some things to consider:
- It seems that the ration driver needs the simulation day to record when it fails. This means two things: it should report the correct simulation day (not -1) and it should get the simulation day from `Time`, which should be the single source of truth for time in a simulation.
- There is no explanation of why some animals have no body weight history, or any fix for it. Is it ok that some animals have no body weight history, or is it indicative that something is not working correctly upstream?
- There needs to be testing done.

RUFAS/routines/animal/animal_module_reporter.py
ltp23: I don't think this is correct because we want to pad the missing days with 0s when pens are created on the fly.

RUFAS/routines/animal/ration/ration_driver.py
ltp23: An alternative is to find latest simulation day among all the animals in the list.

Passing in a Time object and getting simulation day from it may or may not be what we want. The Time object will store the current day but the latest simulation days in the body weight histories may not coincide with the current day.

Though, having access to the current simulation day is good if the animal list is empty and we can use it as a fallback value.

PR #1577, opened 2024-04-23
[Manure] Correct methane double-accumulation in anaerobic lagoon treatment

### LINKED ISSUE #1529: [Manure] Correct methane double-accumulation in anaerobic lagoon treatment
This is a relatively low priority (but also probably low effort level) PR as the problem here only really affects a single accumulated output. 

The value of `Accumulated_ManureTreatmentDailyOutput_Pen_.*.storage_methane` should be equal to the prior day's accumulated value plus the current day's daily value for `storage_methane`. Presently, only in the anaerobic lagoon treatment, the current day's value is being double-counted.

image

However, this double-counting only occurs in the actual `storage_methane` variable, and the values for `Accumulated_ManureTreatmentDailyOutput_Pen_.liquid_manure_total_volatile_solids` or other variables that are utilized in gas emission calculations are not affected. I.e., outputs from this treatment other than `storage_methane` should be correct.

allisterakun:
Correct methane double-accumulation in anaerobic lagoon treatment
### Context

Issue(s) closed by this pull request: closes #1529 

### What

Correct methane double-accumulation in anaerobic lagoon treatment

### Why

Correct methane double-accumulation in anaerobic lagoon treatment

### How

Delete the extra accumulation of methane storage in `_update_methane_emission()`

### Test plan

Looking at the daily vs accumulated storage methane values. Take the difference between the current day's and the previous day's accumulated storage methane value and check if it is the same as the daily storage methane value.
image


ltp23 (approved):
LGTM. Thanks Allister.

Interesting that this only surfaces now.

elle-andreen (approved):
Methane accumulation is now functioning as intended, and individual nutrients (total solids and the volatile solids fractions) are still being updated/accumulated correctly. Thanks Allister!

PR #1578, opened 2024-04-23
[Manure] Remove DefaultFactory for bedding/increase configurability of manure bedding

### LINKED ISSUE #1525: [Manure] Remove default factory for manure beddings and increase configurability
Sub issue of #1524, please read that issue because it goes in depth for what needs to happen to close this issue.

Steps to close this issue:
- Remove the `DefaultBeddingConfigFactory`
- Add a name field as a unique identifier for each bedding configuration.
- Change the manure management scenarios to look at the name field when finding the bedding of that manure management scenario.

The fix executed in #1250 for issue #1127 can be used as a template for how to address this issue.

Isgotkowitz:
This PR moves all configuration of bedding into the manure management input file, and increases the level of configurability that is possible for beddings.

### Context

Issue(s) closed by this pull request: closes #1525 

### What

This PR removes the default bedding configurations that were previously hardcoded into `bedding_classes.py`, and the default factory that used them. It also gives users the capability to configure multiple beddings of the same type.

The input for manure bedding now works in the following way:
- For each bedding config, there is a "name" field which _must_ be unique to that bedding configuration.
- The manure management scenario's field "bedding_type" now refers to the name of the bedding configuration, not the type.

### Why

Having configurations for bedding hardcoded into the manure module was problematic because it made it possible for users to change their input files in such a way that beddings would be automatically created and configured without alerting the user at all.

### How

To achieve the changes to manure inputs, the following changes have been made in the code:
- `BeddingFactory` no longer relies on the `DefaultBeddingConfigFactory` - all configs used to manufacture beddings must come from user input.
- The `BeddingType` is now a regular Enum, meaning that if a value is looked for in the enum but isn't found, an error will be raised. The previous behavior was to silently default to the `SAND` type.
- `ManureManagerConfigHandler._process_bedding_configs` checks to make sure that there are no duplicate configurations for the same "name" bedding option, and raises a ValueError if it finds duplicate configurations. It also now returns of mapping between the unique name of a bedding config and the config, instead of a mapping between the bedding type and the config.
- `ManureManagerConfigHandler.get_bedding_config` now looks for bedding configs based on the name, not on the type of the bedding config. If it does not find the bedding config with the given name, it will raise a KeyError.

### Test plan

All unit tests have been updated and removed as appropriate. Tests have been added to verify correct behavior of `_process_bedding_configs` and `get_bedding_config` in error cases.

Also, simulations have been run to ensure that the simulation crashes if any of the three error conditions are met in the manure management input file:
- Two or more bedding configurations have the same "name".
- A manure management scenario references a bedding configuration name that does not exist in the bedding configuration section.
- A bedding configuration is defined to have a type that is not one of the 6 available options (sand, straw, sawdust, manure solids, CBPB sawdust, none)

### Notes
The way that bedding configurations are specified and handled now matches the behavior of manure handlers.


allisterakun (approved):
Great PR, good job Ed! Simulation runs with no issue!

krb78 (approved):
Looks good to me! I tested naming things the same, making new bedding items (is this the right word?) and changing their configs. RuFaS seemed to give errors as expected and then did respond appropriately to differences when looking at outputs (when there were no errors).

PR #1579, opened 2024-04-24
Enhance json output data with origin label options units in `detailed_values` property

### LINKED ISSUE #1433: [OM] Improve the handling of data origins info part 2
This issue is the second part of #1416.

- [ ] To enable/disable the special handling of data origins info, allow its specification in the filter file by adding an option like: `"origin_label": "selected_option"` (where "selected_option" is one of the following — `true and report origins`, `true origin`, `report origin`, `none or nothing or disabled` (default)). This would enable the decoupling from the command-line option `-i` that turns on/off info maps.
- [ ] Probably create an Enum class that handles those different options. 
- [ ] Change the data origin-value pairs from `[["[true.origin]->[report.origin]", value]]` to a stringified version that matches the specified option like this `"[true.origin]->[report.origin]: stringied_value units"`.
- [ ] When this whole feature is used, consider whether it is possible to remove the `values` field in the json output. Be extra careful with this since it may affect the graph and report generators.
- [ ] Update the wiki entry for output manager to provide more information about how to enable the detailed data origins info and what that entails.

ltp23:
This is a follow-up PR to PR #1491 so it should be merged after it to reduce diff size.

### Context

Issue(s) closed by this pull request: closes #1433 

### What

This PR introduces different options for formatting the origin labels:
- `true and report origins`: Includes both the true origin (class and function) and the report origin, separated by `->`.
Format: `[true_origin_class.true_origin_function]->[report_origin]: value (units)`
Example: [AnimalManager.daily_updates]->[AnimalModuleReporter.report_daily_animal_population.num_animals]: 193 (animals)

- `true origin`: Includes only the true origin (class and function).
Format: `[true_origin_class.true_origin_function]: value (units)`
Example: [WeatherManager.daily_temperature]: avg = 25.5 (°C), min = 18.2 (°C), max = 32.1 (°C)

- `report origin`: Includes only the report origin.
Format: `[report_origin]: value (units)`
Example: [WeatherManager.avg_temperature]: 25.5 (°C)

- `none`: Does not include any origin information, only the value and units.
Format: `value (units)`
Example: 193 (animals)

Units are also included now along with their associated values. The units are specified in the info_maps list of each sub-dictionary and are added to the detailed value strings in parentheses after the value.

### Why

The detailed values feature provides additional context to the output data by including the origins and units of each value.

The origin label formatting options allow users to customize the level of detail included in the detailed values. 

### How

- The `_format_detailed_value_str` method is called to format the detailed value string based on the selected origin label format (true and report origins, true origin, report origin, or none).
- The `_get_origin_label` method retrieves the origin label from the filter content and validates it against the supported options, returning the corresponding OriginLabel enum member or OriginLabel.NONE for invalid values.

### Test plan

Updating unit tests to cover different scenarios:
- Adding detailed values with different origin label formats
- Handling dictionary values and units
- Retrieving the correct origin label based on the filter content

A example json filter file

```
{
  "filters": [
    "Weather.average_annual_temperature",
    "AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_0_CALF"
  ],
  "name": "Selected Variables",
  "origin_label": "true and report origins"
}
```


ew3361zh:
This is handled in PR #2066. Closing as duplicate.

Issue #1332, opened 2024-02-23
[IM][New Feature] Validate metadata properties before validating input data


Isgotkowitz:
There is currently no validation of the properties against which input data is validated, which is allowing buggy inputs to make it into RuFaS simulations.

## Expected Behavior

The properties of input data defined by the metadata should be validated to make sure there are no bugs or errors in them.

## Current Behavior

When IM loads properties from the properties file provided in the metadata, it does not check the properties for any kind of errors.

## Possible Solution

Introduce a set of mechanisms that check the metadata properties to make sure they are valid.

Some example errors that should be checked for in metadata properties:
- Minimums that are greater than maximums.
- Defaults that are invalid.
- Properties having missing or extra keys.

This is not an exhaustive list of possible errors though!

Also, the depth of the properties should be checked so that it does not go deeper than allowed - see [this comment](https://github.com/RuminantFarmSystems/MASM/issues/1249#issuecomment-1937047862) and [this comment](https://github.com/RuminantFarmSystems/MASM/issues/1249#issuecomment-1942738576) from #1249. This issue is also written up in #1333.

## Context (Environment)

There are already a number of errors in the default metadata properties, including instances of all three of the above errors, so this is not a hypothetical issue.

## Possible Implementation

Immediately after loading the metadata properties from the file that they are provided in, the Input Manager should begin validating them. Only after determining that the metadata is valid should the IM move on to loading and validating the input data.

PR #1584, opened 2024-04-26
Validate Metadata and Properties

### LINKED ISSUE #1333: [IM Feature] Limit depth of nested inputs
There is no limit to how deeply nested the structure of input data can be.

## Desired Behavior

A mechanism within the IM prevents inputs from being considered valid if they are "deeper" than a certain amount. See "Possible Solution" for an explanation of depth.

## Current Behavior

The IM (in theory) will validate an input fully no matter how deep it goes. In reality it is not actually validating elements of arrays, regardless of their depth.

## Possible Solution

Define the depth of a value as the number of arrays or objects that must be stepped into in order to reach the value (value is either a string, boolean, or number), and the depth of an object (or dictionary) is set to the depth of its deepest value. Assume that the depth starts at a count of 0 and is already inside an object.

Examples of the above system of applied to data:
`{"a": 1}`, `{"b": "value"}`, and `{"c": true}` have depth = 0.

`{"x": [1, 2, 3]}` and `{"y": {"i": 1, "j": true}}` have depth = 1.

```json
{
  "a" : [
    {"b": "c"},
    {"e": "f"}
  ]
}
```
and `{"a": [[1, 2, 3], [4, 5, 6]]}` have depth = 2.

```json
{
  "a": [
    {
      "e": [1, 2, 3],
      "f": true
    },
    {"g": 4}
  ]
}
```
has depth = 3.

## Context (Environment)

The issue of IM not being able to fully validate inputs because they are too "deep" for the current implementation to handle has already come up - elements of arrays are not always validated, including elements of objects. By explicitly putting depth limits on inputs we can be more confident that the implementation of IM will be able to correctly validate the inputs given to it.

## Possible Implementation

A good way to handle bounding depth of inputs would be not actually check the depths of inputs that passed to IM, but the depths of the properties against which inputs are validated. Using this approach would mean that this issue is really a subtask of #1332.

ew3361zh:
IM now validates metadata and properties before validating input data.

### Context

Issue(s) closed by this pull request: closes #1333 #1332

### What

Adds validation of properties:
1. `_validate_metadata()` method which validates the top level metadata (`files` section) - checks for required keys and checks the values for `path` and `type` to ensure they are valid.
2. `_validate_properties()` method which validates the depth of properties against established `metadata_depth_limit` which is based off the current deepest metadata properies.
3. For each set of properties for an array, bool, number, or string type, there is further validation that all the required keys are present for that property and that none of the keys present is invalid or unexpected. Additionally the values for defaults, minimums, maximums, patterns, etc are validated against the provided values to ensure there are no conflicts.

### Why

These validations can ensure the metadata and the associated properties used to validate the input data are themselves also valid.

### How

New functionality added as described above was added within the IM's `start_data_processing()` method to establish validity of the metadata before the properties are loaded and to establish validaty of the properties before the input data is loaded. Errors and logs are added to OM.

Additionally, some validity errors were raised during the process of setting up these validity checks and those were addressed within input data and properties sections. I did not exhaustively check all the other metadatas but can do so before merging.

### Test plan

Testing has been updated for much of the new functionality but some additional testing is still needed 100% coverage for IM. I'd suggest testing changes to required keys in both the metadata and properties. Also looking at changes to the depth of the properties (i.e. adding depth) or adjusting the `metadata_depth_limit` lower to see violations.


RUFAS/input_manager.py
```diff
@@ -2028,6 +2041,31 @@ def dump_get_data_logs(self, path: Path) -> None:
         file_path = os.path.join(path, file_name)
         om.dict_to_file_json(self.__get_data_logs_pool, file_path)
 
+    def _check_max_depth(self, data: Any, current_depth: int = 0) -> Any:
+        """
+        Recursively find the maximum depth of nested dictionaries.
+        """
```
Isgotkowitz: Docstring

RUFAS/input_manager.py
```diff
@@ -338,6 +343,15 @@ def _populate_pool(self, eager_termination: bool) -> bool:
...
             self._log_missing_keys(missing_required_property_keys, property_keys_with_default_values)
             filtered_input_data = self._filter_input_data_by_metadata(input_data, metadata_properties)
+
+            input_max_depth = self._check_max_depth(filtered_input_data)
+            if input_max_depth > self.input_depth_limit:
+                om.add_warning(
+                    "Max input depth exceeded",
+                    f"Max depth of input file {file_path} exceeds limit of {self.input_depth_limit}",
+                    info_map,
+                )
```
Isgotkowitz: Instead of checking the actual input data, it would be much more efficient to check the metadata properties for 2 reasons: 
- The depth checking time will not fluctuate with the number of input files.
- Instead of checking every single array element, only the properties of that element would be checked.

RUFAS/input_manager.py
```diff
@@ -2028,6 +2041,31 @@ def dump_get_data_logs(self, path: Path) -> None:
         file_path = os.path.join(path, file_name)
         om.dict_to_file_json(self.__get_data_logs_pool, file_path)
 
+    def _check_max_depth(self, data: Any, current_depth: int = 0) -> Any:
```
Isgotkowitz: I think this returns an int, not Any.

RUFAS/input_manager.py
```diff
@@ -2028,6 +2041,31 @@ def dump_get_data_logs(self, path: Path) -> None:
...
+        """
+        Recursively find the maximum depth of nested dictionaries.
```
Isgotkowitz: If possible/feasible, it would be worth considering if this would run more efficiently with an iterative approach instead of a recursive one.

RUFAS/input_manager.py
```diff
@@ -2028,6 +2041,31 @@ def dump_get_data_logs(self, path: Path) -> None:
...
+                    for item in value:
+                        depth = self._check_max_depth(item, current_depth + 1)
+                        max_depth = max(max_depth, depth)
+
+        elif isinstance(data, list):
+            for item in data:
+                depth = self._check_max_depth(item, current_depth + 1)
+                max_depth = max(max_depth, depth)
+
+        return max_depth
```
Isgotkowitz: Instead of going all the way down until the bottom of the dict is reached, why not check if the max depth exceeds the limit and just fail as soon as it hits it?

RUFAS/input_manager.py
```diff
@@ -83,6 +83,7 @@ def __init__(self) -> None:
             self.__pool: Dict[str, Any] = {}
             self.__get_data_logs_pool: Dict[str, str] = {}
             self.elements_counter = ElementsCounter()
+            self.input_depth_limit = 5
```
Isgotkowitz: This should be an attribute in the docstring.

Isgotkowitz (changes requested):
I think this is a good start towards having inputs that are not too deep, but there are more efficient routes that could be taken to realizing this goal. The big thing I am requesting changes on is checking the depth of the metadata properties, instead of checking the depths of the actual inputs - but if you feel that this would be a worse approach to take then we should take some time to discuss it.

RUFAS/input_manager.py
ew3361zh: It's brought up a few questions for me which I can also check in with Pooya on but if you have thoughts, I'd be curious:
1. What are we expecting users to do with a failure on depth of the metadata properties file? 
2. What might be an appropriate failure message? 
3. Do we want to think of this to align with the "eager_termination" flag? - i.e. if someone is validating their input data, they'd probably want to know all the parts that violated the max depth but then if they're running a simulation, we'd have eager_termination on and it would fail at the first instance.

RUFAS/input_manager.py
ew3361zh: Looks like IM has no attributes in the docstring so I'll add them all.

RUFAS/input_manager.py
ew3361zh: Possibly. Recursing made more sense for input files. I'll need to think through what might work best with our metadata properties structure.

RUFAS/input_manager.py
Isgotkowitz: > It's brought up a few questions for me which I can also check in with Pooya on but if you have thoughts, I'd be curious:
> 
> 1. What are we expecting users to do with a failure on depth of the metadata properties file?

Rewrite the metadata properties. Since this is a system that will work with what we currently have it won't come up until the input properties are being modified, which will happen when someone is working on a new feature.

> 2. What might be an appropriate failure message?

Similar to error messages for invalid data, something like `"Properties for input 'path.to.input' have greater than x level of nesting. These properties must be rewritten to be pass."`

> 3. Do we want to think of this to align with the "eager_termination" flag? - i.e. if someone is validating their input data, they'd probably want to know all the parts that violated the max depth but then if they're running a simulation, we'd have eager_termination on and it would fail at the first instance.

I think that would be a nice feature.

Also, definitely good to check with Pooya about this stuff. Everything I have said has been based on my recollections of conversations we had about this issue, but it would be good to get everyone on the same page about what the behavior is going to be.

allisterakun:
During group review session:
- We went through and tested the current code, and it seems to work with no issues.
- @ew3361zh will contact @PooyaHekmati to confirm how we want to traverse the metadata properties so that we can accurately represent the depth (length of path) of a variable in the input data.
- Potentially, we want to make a general function to traverse the metadata properties for validation and have this feature be one of the checks. Other checks may include: making sure all variables have a "type" property defined, making sure the properties of object and array variables follow the standard structure, validating if the provided default value of a variable is valid, checking if there are any extra unknown keywords...

RUFAS/input_manager.py
```diff
@@ -2028,6 +2044,271 @@ def dump_get_data_logs(self, path: Path) -> None:
...
+        optional_keys = {"title", "description"}
+        valid_keys = required_keys | optional_keys
+        valid_types = {"json", "csv"}
+        for key, data in metadata_files.items():
+            if not required_keys <= data.keys():
+                missing_keys = required_keys - data.keys()
+                om.add_error("Metadata Validation", f"Missing required keys '{missing_keys}' in {key}", info_map)
+                raise ValueError
+            invalid_keys = data.keys() - valid_keys
+            if invalid_keys:
```
allisterakun, line 2064:
```suggestion
            if invalid_keys := (data.keys() - valid_keys):
```

RUFAS/input_manager.py
```diff
@@ -2028,6 +2044,271 @@ def dump_get_data_logs(self, path: Path) -> None:
...
+            "function": self._validate_metadata.__name__,
+        }
+        metadata_files = self.__metadata["files"]
+        required_keys = {"path", "type", "properties"}
+        optional_keys = {"title", "description"}
+        valid_keys = required_keys | optional_keys
+        valid_types = {"json", "csv"}
+        for key, data in metadata_files.items():
+            if not required_keys <= data.keys():
+                missing_keys = required_keys - data.keys()
```
allisterakun, line 2060:
```suggestion
            if missing_keys := (required_keys - data.keys()):
```

RUFAS/input_manager.py
```diff
@@ -2028,6 +2044,271 @@ def dump_get_data_logs(self, path: Path) -> None:
...
+                    "Metadata depth exceeds maximum allowed depth of " f"{self.metadata_depth_limit} at path {path}",
+                    info_map,
+                )
+                raise ValueError(
+                    "Metadata depth exceeds maximum allowed depth of " f"{self.metadata_depth_limit} at path {path}"
+                )
+
+            if depth > max_depth:
+                max_depth = depth
+                deepest_path = path
```
allisterakun, line 2121:
```suggestion
        current_max_depth = 0
        deepest_path = []

        type_to_validator_map: Dict[str, Callable] = {
            "number": self._metadata_number_validator,
            "array": self._metadata_array_validator,
            "bool": self._metadata_bool_validator,
            "string": self._metadata_string_validator,
        }
        while stack:
            current_obj, depth, path = stack.pop()

            if depth > self.metadata_depth_limit:
                om.add_error(
                    "Max metadata depth exceeded.",
                    "Metadata depth exceeds maximum allowed depth of " f"{self.metadata_depth_limit} at path {path}",
                    info_map,
                )
                raise ValueError(
                    "Metadata depth exceeds maximum allowed depth of " f"{self.metadata_depth_limit} at path {path}"
                )

            if depth > current_max_depth:
                current_max_depth = depth
                deepest_path = path
```
maybe rename this local variable to avoid confusion with `self.metadata_depth_limit`

RUFAS/task_manager.py
```diff
@@ -110,6 +110,7 @@ def start(
                 self.output_manager,
                 "TASK_MANAGER",
             )
+            print("right before the exception")
```
allisterakun, line 113:
```suggestion
```
remove print statement

input/metadata/properties/default.json
```diff
@@ -3482,7 +3482,7 @@
         "type": "number",
         "minimum": 0.87,
         "maximum": 1.04,
-        "default": 0
+        "default": 0.87
```
allisterakun, line 3485:
```suggestion
        "default": 0
```
restore the original value

input/metadata/properties/default.json
```diff
@@ -4878,7 +4878,6 @@
     "field_size": {
       "type": "number",
       "description": "Size of the field.",
```
allisterakun, line 4880:
```suggestion
      "description": "Size of the field.",
      "units": "ha",
```
Restore the previous state

allisterakun (commented):
Generally looks good. As Ed mentioned, I do believe this PR should be split into at least two PRs.

input/metadata/properties/default.json
ew3361zh: This one is an incorrect default which falls outside the minimum-maximum range. I confirmed the new value with Joe but will document this in the new PR for validating properties.

input/metadata/properties/default.json
ew3361zh: Confirmed change that should be happening in metadata properties - "units" is not a valid key and we shouldn't be tracking it in the metadata. Will document this though in the new PR.

ew3361zh:
Divided up into separate PRs #1637 and #1640.

PR #1585, opened 2024-04-26
Update pilot testing manure and soil inputs

Isgotkowitz:
Among other changes made to the main branch, the inputs for manure and soil have changed. This PR pulls in the changed inputs and properties, and updates the pilot farm inputs to work with the new properties. There should be no significant impact to final outcomes after incorporating these changes.

#### Important Things to Check:
That the culling outputs remain the same. The structure of culling probability inputs has been updated but all the actual inputs _should_ be updated correctly.


jadamchick (approved):
Hi Ed, 
Thanks for this, it looks good and runs smoothly for me. I merged it into a copy of the pilot testing branch (no issues merging) and ran a few farms at random; their culling/turnover outputs all looked similar. 
I did notice a new warning cropped up: 
`"Default value used for required property key that was missing: include_detailed_values => False.",`

It looks we like we need to include this new input in the config files? I don't know enough context to say whether it matters right now or not. We can ignore it / go ahead and merge if it's not material to running these farms.

Isgotkowitz:
> Hi Ed, Thanks for this, it looks good and runs smoothly for me. I merged it into a copy of the pilot testing branch (no issues merging) and ran a few farms at random; their culling/turnover outputs all looked similar. I did notice a new warning cropped up: `"Default value used for required property key that was missing: include_detailed_values => False.",`
> 
> It looks we like we need to include this new input in the config files? I don't know enough context to say whether it matters right now or not. We can ignore it / go ahead and merge if it's not material to running these farms.

Nice. The `include_detailed_values` flag was introduced in #1491, it just determines whether or not extra information is included with output from JSON filters. I set it to `false` for all the configs because I don't think it would really add anything, but it can be changed very easily.

PR #1586, opened 2024-04-26
Updates manure volume used in anaerobic lagoon NH3-N emission

### LINKED ISSUE #1546: [Manure] Incorrect manure volume used in anaerobic lagoon NH3-N emission
## Expected Behavior
A component of the equation predicting ammonia emission from liquid manure storages (anaerobic lagoon or slurry storage) is the total mass of manure in storage ([see equation M.3.A.10 on page 14](https://3.basecamp.com/3486446/buckets/5296287/cloud_files/7195916818)). This value should be obtained from the accumulated manure treatment output, not the daily output. See the slurry_storage_outdoor.py file for a correct implementation of this functionality.

## Current Behavior
Currently, in anaerobic_lagoon.py, the **daily** manure volume is utilized in the equation for storage ammonia emission. 

image

The value for `manure_volume` in this equation should be updated to be `accumulated_output.liquid_manure_daily_volume`. 

The expected outcome of this change is that predicted ammonia emissions from the anaerobic lagoon will decrease considerably (~50%), and will be more in-line with estimated emissions from slurry storage outdoor, where the function was being implemented correctly. The decrease occurs because the total manure mass value is in the denominator of the equation - we are moving from a smaller (daily) to a larger (accumulated) value, so the final value will decrease.

ew3361zh:
### Context

Issue(s) closed by this pull request: closes #1546

### What

Updates manure volume used to calculate AnLag NH3-N emission to `self._accumulated_output.liquid_manure_daily_volume`. This matches the behavior in `slurry_storage_outdoor.py`.

### Why

In AnaerobicLagoon, the function used to calculate NH3-N had been incorrectly using `daily_output.daily_final_manure_volume` leading to the massive over-estimates in expected ammonia emissions from the anaerobic lagoon.

### How

Swapped out the daily manure volume for the accumulated output liquid daily volume. The equation in GasEmissionsCalculator was already expecting the `manure_volume` attribute to be "total volume of the manure produced by the animals in the storage area."

### Test plan

Testing was updated to expect the new accumulated manure volume. Having @elle-andreen review outputs to ensure it is changing as expected.


elle-andreen (approved):
The volume and NH3-N emission from the lagoon are being calculated as expected now. I added some quick figures below just to illustrate the change in storage ammonia from the lagoon. In the righthand updated panels, lagoon NH3-N emissions are much more in line with slurry storage emissions, whereas previously they were several times greater. 

Thanks for helping uncover and correct this Niko! 

image

allisterakun (approved):
Straight-forward changes! Thank you, Niko!

PR #1587, opened 2024-04-26
[Manure] Remove urine TAN factor

elle-andreen:
The [changelog](https://docs.google.com/spreadsheets/d/1U-3igxdstHTFb6k5dVcE5-x9IL0r6q18eUiwwytKZyA/edit#gid=0) has been updated.

### Context
As part of the manure module output evaluation, April noted the TAN values in the manure module were lower than expected. She recommended we remove the previously-used conversion factor (0.45) converting urinary N to TAN and just set TAN equal to urinary N.

### What
This PR removes the urine TAN factor (0.45) from all files where it was previously used. Manure and urine TAN are now equal to urinary N.


tomhuhh (changes requested):
Thanks Elle for tagging me on this PR! These changes make sense to me. I used the following filter just to understand the ratio of urine N in manure N. My results showed that on average urine N is about 51.1% of manure N, which is consistent with what April reported.  

```
AnimalModuleReporter.report_pen_manure_properties.pen_3_daily_urine_nitrogen
AnimalModuleReporter.report_pen_manure_properties.pen_3_daily_manure_nitrogen
```

However, I noticed that TAN is also calculated in the animal module with a different approach as shown below (can be found in all the `_manure_excretion.py` files). The TAN values look quite different from the urine N values. Also, these TAN values from the output of the animal module do not seem to be used by the manure module. I wonder if we should just remove these lines of code.  

   ```
 tan_percent_of_urea = 48.2 - 2.9 * urine_urea_nitrogen_concentration
  
    total_ammoniacal_nitrogen_concentration = (tan_percent_of_urea / 100) * urine_urea_nitrogen_concentration
```
Everything else looks great to me. 

Thanks!

elle-andreen:
That is a good point Haowen! I think it's likely that we could remove those calculations for now. If I recall correctly, we used the original urine N to TAN conversion this this PR is moving because the calculations you noted in your comment were not lining up well with expected urine or manure TAN values. @KFosterReed, would you be fine with removing the actual TAN calculations (that are not being used) in addition to removing the TAN conversion factor?

KFosterReed:
If it is easy to remove them without causing downstream dependency problems, let's do it!

elle-andreen:
After investigating in the code, it is not as straightforward as I would have originally imagined to just remove the TAN calculations @tomhuhh was referring to. I was not able to just remove `total_ammoniacal_nitrogen_concentration` entirely from the manure excretion files in the animal module and still have the model run appropriately. I think this is because the calf TAN excretion is estimated a little differently vs. the other animal classes, but frankly I am not super familiar with the different excretion prediction methods. I think it would be best to check in with Joe/Animal module folks on this, and then create a separate issue to remove these functions if desired.

tomhuhh (approved):
Thanks Elle for following up on this PR. I'll create a separate issue for updating the TAN equations in the animal module to maintain consistency with the manure module.

Isgotkowitz (approved):
LGTM, nice job Elle

PR #1589, opened 2024-04-29
[IM] Fixed modifiability check during initialization

### LINKED ISSUE #1565: [IM] Modifiability check removed
## Urgency Level

- [x] Urgent: a critical feature or fix is needed

## Status

The modifiability check for missing input data was removed in PR #1345. 
This results in `IM._log_missing_data()` being uncalled.

## Intention

The supposed behavior is:
IM would loop through metadata and try to extract each property from the input data for validation. If a variable specified in the metadata is not included in the input data, we should check for its modifiability to see if it is required upon initialization and handle it accordingly.

## Request

Use a _try_ _except_ block when retrieving variables from the input data, and handle the missing data situation.

allisterakun:
The modifiability check was accidentally removed. This PR fixes it for the `IM.populat_pool()` call during initialization.
### Context

Issue(s) closed by this pull request: closes #1565 

### What

Introduced a new `IM._extract_input_data_by_key_list()` function:
```python
    def _extract_input_data_by_key_list(self,
                                        input_data: List[Any] | Dict[str, Any],
                                        variable_path: Sequence[str | int],
                                        variable_properties: Dict[str, Any],
                                        called_during_initialization: bool) -> Any:
        """
         Extracts a value from the input data based on a specified path and handles missing data by calling
         InputManager._log_missing_data().

        Parameters
        ----------
        input_data : List[Any] | Dict[str, Any]
            The input data containing the value to be extracted.
        variable_path : List[str | int]
            A list of keys to be used to extract the value from the input data.
        variable_properties : Dict[str, Any]
            The metadata properties for the variable being validated.
        called_during_initialization: bool
            Boolean variable indicating whether the function is being called during initialization.

        Returns
        -------
        Any
            The value extracted from the input data if found.
            None if not found.
        
        Notes
        -----
        This function navigates through the given input data (which can be a list or a dictionary) following the path
        specified in `variable_path`. If the path leads to a value, it is returned.
        If a KeyError occurs during this process (i.e., a key or index is missing in the path), the function extracts
        the variable name by finding the last string element in the `variable_path` array and handles this missing data
        by calling InputManager._log_missing_data().
        """
```
When retrieving the input data, the validator functions will call `IM._extract_input_data_by_key_list()` instead of calling `IM._extract_value_by_key_list()` to handle the possible missing data situiation.

### Why

The missing input data situation is improperly handled in the IM validation process; the modifiability is not checked.

### How

By calling the new `IM._extract_input_data_by_key_list()` function, if encountering a missing data situation, it will call `InputManager._log_missing_data()` to check the modifiability and properly handle the situation.

### Test plan

All affected unit tests are updated. 
Wrote new tests for the new `InputManager._extract_input_data_by_key_list()` function to bring the coverage for IM to 100%.
image


RUFAS/input_manager.py
```diff
@@ -1299,6 +1309,54 @@ def _extract_value_by_key_list(
...
+    def _extract_input_data_by_key_list(
+        self,
+        input_data: List[Any] | Dict[str, Any],
+        variable_path: Sequence[str | int],
+        variable_properties: Dict[str, Any],
+        called_during_initialization: bool,
+    ) -> Any:
+        """
+         Extracts a value from the input data based on a specified path and handles missing data by calling
+         InputManager._log_missing_data().
```
```suggestion
        Extracts a value from the input data based on a specified path and handles missing data by calling
        InputManager._log_missing_data().
```

RUFAS/input_manager.py
```diff
@@ -1299,6 +1309,54 @@ def _extract_value_by_key_list(
...
+        specified in `variable_path`. If the path leads to a value, it is returned.
+        If a KeyError occurs during this process (i.e., a key or index is missing in the path), the function extracts
+        the variable name by finding the last string element in the `variable_path` array and handles this missing data
+        by calling InputManager._log_missing_data().
+        """
+        result = None
+        try:
+            result = self._extract_value_by_key_list(input_data, variable_path)
+        except KeyError:
+            var_name: str = [name for name in reversed(variable_path) if type(name) == str][0]
```
Isgotkowitz: Instead of just using the variable name, I think it would be better to use `_convert_variable_path_to_str` and have the result it produces be the var name. This makes errors that are logged downstream much easier to understand.

Isgotkowitz (changes requested):
If the `modifiability` of an input is `unrequired unlocked` in the metadata properties and is missing at initialization, the Input Manager will still say that the input is required at runtime. Which I believe is not the correct behavior, but I am not actually sure. 

Also, there aren't any uses of modifiability in the metadata properties, and no mention of it in the Wiki page. I think this would be a good opportunity to revisit its use and documentation.

README.md
```diff
@@ -1,7 +1,7 @@
 [![Flake8](https://img.shields.io/badge/Flake8-passed-brightgreen)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
 [![Pytest](https://img.shields.io/badge/Pytest-passed-brightgreen)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
 [![Coverage](https://img.shields.io/badge/Coverage-95%25-brightgreen)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
-[![Mypy](https://img.shields.io/badge/Mypy-4024%20errors-red)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
+[![Mypy](https://img.shields.io/badge/Mypy-4025%20errors-red)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
```
Isgotkowitz: Whatever the new error is, it should be corrected.

RUFAS/input_manager.py
```diff
@@ -348,7 +348,7 @@ def _populate_pool(self, eager_termination: bool) -> bool:
                     eager_termination=eager_termination,
                     properties_blob_key=properties_blob_key,
                     elements_counter=self.elements_counter,
-                    called_during_initialization=False,
+                    called_during_initialization=True,
```
ew3361zh, line 351: If we're setting it to True here, I think we should probably remove the default False value for it in _log_missing_data() (and probably not have a default value for it unless there's a good reason).

allisterakun: Will update that!

ew3361zh: I'm also curious about it having a default value of False in all the `type_to_validator` functions.

ew3361zh:
I'm curious if it would work to add `called_during_initialization` to the args and then call `_extract_value_by_key_list()` instead of making this wrapper `_extract_input_data_by_key_list()` function?

allisterakun:
> I'm curious why it wouldn't work to just call `_extract_value_by_key_list()` instead of making this wrapper `_extract_input_data_by_key_list()` function?

Because `_extract_value_by_key_list()` is also used by other parts of the OM (e.g., extracting metadata). Instead of having a try-except block for all calls to `_extract_value_by_key_list()` for extracting input data, I made this wrapper `_extract_input_data_by_key_list()` function to handle the exception of missing input data specifically.

Isgotkowitz (approved):
LGTM, works as expected.

RUFAS/input_manager.py
Isgotkowitz: This hasn't been changed, is this deliberate or should this be changed?

ew3361zh (approved):
LGTM, I had one suggestion which is to remove the default values for `called_during_initialization` in the validate by type functions since I think we have it set correctly now at the higher level and the defaults at the lower level could lead to unindended results. Nice work!

PR #1591, opened 2024-04-29
Wrangle time and weather to generate series of dates and weather conditions

Isgotkowitz:
This PR adds the functionality to generate series of dates and get series of `CurrentDayConditions` from `Weather`.

### Context

- [ ] Yes
- [x] No

### What

Two utility methods are added to handle dates: one converts an ordinal date to a `datetime.date`, and the other generates a series of `datetime.date`s based on a given date and how the series should start/end around it.

The functionality has been added to `Weather` so that instead of just getting the weather for the current day, a series of weather conditions can be gathered over any arbitrary timespan.

### Why

This feature is critical to making the feed storage module fully functional, because feed storage will need to both get series of past and future weather data in order to properly process degradations and project future feed states.

### How

Heavy use of Python's `datetime` library are used for generating and handling time 

Series of weather data are gathered by using indexing relative to the years in which the weather data is held. This is a very precarious approach, more on this point below.

### Test plan

Unit tests have been added for the utility methods.

A unit test the weather series generation has been written, however it is not exhaustive because the structure of weather data makes it really difficult to efficiently test realistic weather data sets. I have generated weather series manually in the middle of RuFaS simulations and verified that they are getting correct series, so I am confident that it is working well enough to go ahead with this PR. But I will plan on refactoring Weather as part of the next sprint so that the internals of this method can be simplified and made much more robust.

### Edits
- 1
  - @allisterakun pointed out that `convert_ordinal_date_to_month_date` essentially contained the functionality of `day_to_month_conversion` method, so that method has been deprecated and replaced with `convert_ordinal_date_to_month_date` in the `Weather` and `CurrentDayConditions` classes.
  - Funny enough, this actually fixes a bug that was present in the code but had no effect on final outputs because it only comes up when simulating latitudes close to the north and south poles.


RUFAS/util.py
```diff
@@ -504,3 +505,56 @@ def is_leap_year(year: int) -> bool:
...
+        """
+        if starting_offset > ending_offset:
+            raise ValueError(f"Starting offset ({starting_offset=}) is greater than ending offset ({ending_offset=}).")
+
+        time_series = [date + datetime.timedelta(day) for day in range(starting_offset, ending_offset + 1)]
+
+        return time_series
+
+    @staticmethod
+    def convert_ordinal_date_to_month_date(year: int, day: int) -> datetime.date:
```
allisterakun, line 476: I found that there is another `day_to_month_conversion()` in the `Utility` class that sort of overlaps with the functionality of this new method. That method converts the Julian/ordinal day into the corresponding month. Consider deprecating `Utility.day_to_month_conversion()` and change all occurrences to call the new method and use the month in the returned `Datetime` object.

allisterakun (approved):
Great PR, Ed! 
LGTM, just one minor comment!

RUFAS/util.py
Isgotkowitz: Good idea!

RUFAS/util.py
```diff
@@ -504,3 +444,56 @@ def is_leap_year(year: int) -> bool:
...
+        """Generates a datetime.date based on a year and ordinal day."""
+        maximum_day = (
+            GeneralConstants.YEAR_LENGTH if not Utility.is_leap_year(year) else GeneralConstants.LEAP_YEAR_LENGTH
+        )
+        if not 1 <= day <= maximum_day:
+            raise ValueError(f"Invalid day: {day} of year {year}.")
```
ew3361zh:
```suggestion
            raise ValueError(f"Invalid day: {day} of year {year} must be between 1 and {maximum_day}.")
```

RUFAS/weather.py
```diff
@@ -181,6 +181,62 @@ def get_current_day_conditions(self, time: Time) -> CurrentDayConditions:
 
         return current_conditions
 
+    def get_conditions_series(self, time: Time, starting_offset: int, ending_offset: int) -> list[CurrentDayConditions]:
```
ew3361zh, line 184: This function seems a little overly complex and not single-responsibility but I think it sounds like it's part of a larger refactor of Weather which needs to happen and I'm okay with this as a short-term holdover to get functionality working in feed storage.

ew3361zh (approved):
Nice work, Ed. It seems to be functioning as intended. I didn't have any issues running it or testing it. I thought the `get_conditions_series()` was a little complex and could be broken into smaller functions but it seems like it's a part of a larger plan to restructure weather and that handling weather in its current form so that other things aren't blocked is a higher priority here.

RUFAS/weather.py
Isgotkowitz: It might be good to break it up then - it is definitely not supposed to be more than a short-term holdover but stuff happens and just in case the weather reformat doesn't happen in the very near future, it would be good to have this method written as well as possible. Are there any distinct functionalities that stick out to you when looking at this method which you think could be put into subroutines?

PR #1592, opened 2024-04-29
Write to file the usage counts of filters

### LINKED ISSUE #1559: [OM] Create methods to preprocess output filters
One potential way to save on memory usage at runtime is to preprocess the filter files to keep track of what variables from the variables pool would be requested later. This can provide a shortcut to storing only those variables during runtime. Naturally, if the filters are exhaustive and cover all the variables, then this approach wouldn't save anything.

For this issue, I want to gauge the usefulness of this approach by first creating one or more methods to record and count the frequency of all the requested variables and save the results to a file for further analysis.

ltp23:
### Context

Issue(s) closed by this pull request: closes #1559 

### What

The goal of this PR is to keep track of what variables and how often they are being used by the filter files and then write this info to a file.

### Why

The resulted file can help inform later optimization tricks such as storing only those variables needed by the filter files and caching the results of duplicate filters.

### How

- Save the filter counts in the attribute OM's `_filters_counter`.
- Write to file at the end of the simulation along with other dumping methods. All of them are called inside the `dump_all_nondata_pools()` method.
- The resulted file is named something like this `default_filters_usage_<timestamp>.json`.

### Test plan

- Write some filter files and see what filters are recorded.
- Also note that we are counting the usage frequencies of actual variables. 

---

Edit 1:

- One thing of note is that the counts of filtered variables make no distinction between normal filtering and filtering by exclusion. For example, if in one filter file variable A is included and in another filter file that same variable A is not excluded, the usage count here for variable A would be 2. The output file doesn't have any information about what proportion of the counts comes from which method of filtering.


RUFAS/output_manager.py
```diff
@@ -1233,6 +1240,19 @@ def dump_errors(self, path: Path) -> None:
         file_path = os.path.join(path, self.generate_file_name("errors", "json"))
         self.dict_to_file_json(self.errors_pool, file_path)
 
+    def dump_filters_usage_data(self, path: str) -> None:
```
Isgotkowitz: Might be better to use `Path` here instead of string, to keep it consistent with the other methods for dumping logs, errors, warnings

RUFAS/output_manager.py
```diff
@@ -931,6 +937,7 @@ def filter_variables_pool(self, filter_content: Dict[str, Any]) -> Dict[str, poo
             filter_patterns=filter_content.get("filters", []),
             filter_by_exclusion=filter_by_exclusion,
         )
+        self._filters_counter.update(filtered_pool.keys())
```
allisterakun: Can we move this counter-updating process into line 1053 to avoid potential issues with the OM VM issues?

RUFAS/output_manager.py
```diff
@@ -138,6 +143,7 @@ def __init__(self) -> None:
                 },
             )
             self.time = None
+            self._filters_counter: Counter[str] = collections.Counter()
```
allisterakun: Maybe rename the `_filter_counter` to something like `_filtered_variable_counter` to clarify that this counter is counting the filtered variables instead of the actual filter.

allisterakun (approved):
Straightforward PR, nick work, Loi!

RUFAS/output_manager.py
Isgotkowitz: Reiterating the point Allister brought up, it would be better to use the "variables" instead of "filters" because that is what is really being counted - this applies to all things related to counting variables in this PR.

Isgotkowitz (approved):
Works as expected, LGTM. Made suggestions on some small things.

Regarding the use of exclusion vs. inclusion filters, I think it is fine how it is but it would be good to explicitly note its behavior so that nobody gets confused or has to guess how it works.

ltp23:
> Regarding the use of exclusion vs. inclusion filters, I think it is fine how it is but it would be good to explicitly note its behavior so that nobody gets confused or has to guess how it works.

Thanks Ed. I updated the following comment in the description.

Edit 1:

- One thing of note is that the counts of filtered variables make no distinction between normal filtering and filtering by exclusion. For example, if in one filter file variable A is included and in another filter file that same variable A is not excluded, the usage count here for variable A would be 2. The output file doesn't have any information about what proportion of the counts comes from which method of filtering.

ltp23:
Thank you guys for reviewing!

RUFAS/output_manager.py
ltp23: Yes I will change it to `_filtered_variables_counter`.

RUFAS/output_manager.py
ltp23: Yes I will move it to `save_results`.

PR #1594, opened 2024-04-29
Improve Graph Generation Filter Handling

### LINKED ISSUE #1532: Graph generation exception too vague
## Urgency Level

## Expected Behavior

Graph generation exception should give the name of the filter that caused the exception. 

## Current Behavior

image
and 
image

Neither give an indication of which filter is causing the problem.

## Possible Solution

Reporting the filter name in either the error log or in console would greatly speed up debugging.

## Steps to Reproduce

1. Unsure which is causing a problem! If necessary I can send all my active filters to reproduce elsewhere, otherwise tag me in the PR and I'll try it locally.

## Context (Environment)

Some graph filters were used for testing reasons, but I could see a user setting up multiple filters and then being confused which is failing (especially when many are in place, and when variable/output names change, or when graph options are tweaked, etc).

ew3361zh:
When GraphGenerator is unable to graph data based on a filter, the exceptions generated are sometimes less than informative and helpful (the message is "graph generation exception. string indices must be integers"). This PR makes an effort to try and prevent some of the more common errors that users may encounter that were leading to these errors.

### Context

Issue(s) closed by this pull request: closes #1532

### What

2 of the more common errors that generate a non-specific error from GraphGenerator are:
1. Having invalid characters in the title of the graph making the graph-saving process error.
2. Selecting a graph type that is not from the list of currently known working graphs: `plot` and `stackplot`.

If GG encounters either of these errors, it will raise the "graph generation exception. string indices must be integers" exception and the user likely will not know which filter triggered it (without reviewing which graphs actually were produced) nor why that filter triggered it.

This PR addresses both of these issues. We can't be exhaustive on what errors will be produced by GG but these two seem more common and worth addressing since they're also easily fixable.

### Why

By handling some of the more common cases users have been encountering where they're getting graphing errors, the model is more usable, friendly, and robust.

### How

1. In the TUPLE_BASED_FUNCTIONS global var in graph_generator.py, the "scatter" type is added which enables it to function properly.
2. In util.py, a new function has been added to filter out characters that typically will lead to errors in saving files. The graph's title (if it has a title), will be filtered using this function.

Further examination will be needed to determine if all other graph types are enabled and fully functioning. This will be covered in issue #1602.

### Test plan

Testing was added to cover these new functinalities.
To test, I would try running a simulation with -v errors and having a graph filter that has a `type` that uses `scatter` and review the error logged to OM. Secondly I'd add a `title` to the graph filter that contained one of the prohibited characters ("<", ">", ":", "/", "\"", "|", "\\", "?", "*", ".") and see how the resulting graph looks (and there should be no errors for including one or multiple of these characters).


JoeWaddell (approved):
This all looks great: I've tested it with a number of filters and gotten the expected results. Nice work! Clean, readable code.

Isgotkowitz (approved):
LGTM, functions as expected - although I can't get any scatter plots to work. This might just be user error though.

PR #1595, opened 2024-04-30
Remove default factory and increase configurability for manure separators

### LINKED ISSUE #1526: [Manure] Remove default factory for manure separators and increase configurability
Sub issue of https://github.com/RuminantFarmSystems/MASM/issues/1524, please read that issue because it goes in depth for what needs to happen to close this issue.

Steps to close this issue:

- Remove the `DefaultManureSeparatorConfigFactory`
- Add a name field as a unique identifier for each separator configuration.
- Change the manure management scenarios to look at the name field when finding the separator of that manure management scenario.

The fix executed in https://github.com/RuminantFarmSystems/MASM/pull/1250 for issue https://github.com/RuminantFarmSystems/MASM/issues/1127 can be used as a template for how to address this issue.

Isgotkowitz:
This PR makes the manure management input file fully responsible for configuring separators in the manure module, and increases the level of configurability for separators in the manure module.

### Context

Issue(s) closed by this pull request: closes #1526 

- [x] Yes
- [ ] No

### What

The configurability of the manure separators has been increased by making it possible to configure multiple separators of the same "type" - so now it is possible to have multiple screw presses and rotary screens. It also removes the default configurations from inside the manure module, so only the manure separators that are defined in the input file will be available to use in a simulation.

### Why

Before, it was possible to define multiple separator configurations of the same type, but only one of them would be used - and this behavior would not be flagged to the user at all. It would also be possible for a user to specify a separator that they did not provide a configuration for, and the Manure Module would use a default configuration without alerting to the user that this was happening.

### How

The `DefaultManureSeparatorConfigFactory` has been removed, which was previously responsible for handling the default configurations.

A "name" field has been added to the separator configurations, which is a unique identifier used to distinguish between the different separator configurations. When defining a manure management scenario, the user must reference the name field when defining the manure separators used in that scenario (for both pre- and post-digestion separators). 

Importantly, a user can also specify "none" as the name of a separator used in a manure management scenario, but cannot define a separator configuration with the name "none". This is because "none" is a special case that represents no separation occurring, and the user should not be able to think they are configuring the "none" separator when in fact they are not.

All other changes made in this PR are necessary changes to support the switching from referencing manure separator configurations by their type to their name.

### Test plan

All unit tests related to these changes have been updated.

The following filter has been written to compare the separator outputs from the default scenario on `main` and this branch (no differences were found in outputs - as expected):
```json
{
  "multiple": [
    {
      "name": "Manure Separator Output",
      "filters": [
        "ManureSeparatorDailyOutput_Pen_2_CLOSE_UP..*"
      ],
      "vertical_aggregation": "average"
    }
  ]
}
```

The manure management input has been changed to make sure RuFaS fails to run in the following cases:
1. The manure separator in the management scenario references a separator that is not configured.
2. Two manure separator configurations have identical "name" fields.
3. A manure separator is configured with the name "none".
4. A manure separator is configured with a type that is currently not-allowed/unsupported (something other than "rotary screen" or "screw press")

### Open Questions
- Should more options be added for separator types? Currently, there are only two separator types available but running switching between them has no effect on the outputs.
  - See the file `RUFAS/routines/manure/manure_separators/manure_separator_classes.py` to see this.


RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
```diff
@@ -110,23 +110,39 @@ def get_manure_handler_config(self, manure_handler_type_name: str) -> ManureHand
...
+        if manure_separator_name == "none":
+            return None
+        try:
+            return self.manure_separator_configs[manure_separator_name]
+        except KeyError:
+            info_map = {"class": self.__class__.__name__, "function": self.get_manure_separator_config.__name__}
+            error_title = (
+                f"Attempted to use a non-existent manure separator configuration called '{manure_separator_name}'."
+            )
+            error_message = "Raising ValueError."
```
allisterakun: I think that the `error_title` should be a brief summary of the error, while the `error_message` should contain the details.
```suggestion
            error_title = "Unknow Manure Separator Name"
            error_message = (
                f"Attempted to use a non-existent manure separator configuration called '{manure_separator_name}'."
            )
```

RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
```diff
@@ -220,29 +236,35 @@ def _process_manure_handler_configs(cls, manure_handler_configs: List[Dict]) ->
...
-            )
-        return manure_separator_config_by_manure_separator_type
+        info_map = {"class": cls.__name__, "function": cls._process_manure_separator_configs.__name__}
+
+        available_manure_separator_configs: Dict[str, ManureSeparatorConfig | None] = {}
+        for config in manure_separator_configs:
+            name = config.pop("name")
+            if name in available_manure_separator_configs:
+                error_name = f"Manure separator '{name}' has multiple configurations."
+                error_message = "Raising ValueError."
```
allisterakun: Same as above
```suggestion
                error_name = "Duplicate Manure Separator Configuration"
                error_message =  f"Manure separator '{name}' has multiple configurations."
```

RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
```diff
@@ -220,29 +236,35 @@ def _process_manure_handler_configs(cls, manure_handler_configs: List[Dict]) ->
 
     @classmethod
     def _process_manure_separator_configs(
-        cls, manure_separator_json_configs: List[Dict]
-    ) -> Dict[ManureSeparatorType, ManureSeparatorConfig]:
-        """Returns a dictionary of manure separator config objects, with the key being the manure separator type.
+        cls, manure_separator_configs: List[Dict]
```
allisterakun:
```suggestion
        cls, manure_separator_configs: List[Dict[str, Any]]
```

RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
```diff
@@ -220,29 +236,35 @@ def _process_manure_handler_configs(cls, manure_handler_configs: List[Dict]) ->
...
+    ) -> Dict[str, ManureSeparatorConfig | None]:
+        """
+        Returns a dictionary of manure separator config objects, with the key being the name of the separator config.
 
         Parameters
         ----------
-        manure_separator_json_configs : List[Dict]
+        manure_separator_configs : List[Dict]
```
allisterakun:
```suggestion
        manure_separator_configs : List[Dict[str, Any]]
```

RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, manure_manager_config):
         self.manure_handler_configs = self._process_manure_handler_configs(
             manure_manager_config["manure_handler_configs"]
         )
-        self.custom_manure_separator_configs = self._process_manure_separator_configs(
+        self.manure_separator_configs = self._process_manure_separator_configs(
```
allisterakun, line 48: This is just a take-or-leave suggestion, but consider renaming this property to clearly indicate that it refers to the available/possible manure separators.
It took me a while to realize that this does not refer to the separator that the user defined in the scenarios.

RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
```diff
@@ -220,29 +236,35 @@ def _process_manure_handler_configs(cls, manure_handler_configs: List[Dict]) ->
 
     @classmethod
     def _process_manure_separator_configs(
```
allisterakun, line 236: Same take-or-leave suggestion: consider renaming this function to clearly indicate that it is "processing" (maybe a better verb?) the available/possible manure separators.

allisterakun (approved):
Great PR, Ed!
Pulled the branch and confirmed that it works as expected.
Just some minor take-or-leave comments.

input/metadata/properties/default.json
```diff
@@ -4415,13 +4415,11 @@
         },
         "manure_separator": {
           "type": "string",
-          "description": "Manure Separator Type -- The type of solid-liquid separator equipment to separate coarse fibrous solids/sand",
-          "pattern": "^(screw press|rotary screen|none)$"
+          "description": "Name of the manure separator to be used in this scenario. Must match 'name' attribute of the Manure Separator Config to be used unless no separator ('none') is used."
```
elle-andreen: I think this description and the one below could be slightly more general, I would word it as "Name of the manure separator used in a manure management scenario. Each named separator must have a Manure Separator Config defined, unless no separator ('none') is used."

input/metadata/properties/default.json
```diff
@@ -4507,10 +4505,15 @@
...
       "properties": {
         "type": "object",
+        "name": {
+          "type": "string",
+          "description": "Name of the unique manure separator config. This can be any string besides 'none'.",
+          "pattern": "^(?!none$)(.*)$"
+        },
         "manure_separator_type": {
           "type": "string",
           "description": "Manure Separator Type -- The type of solid-liquid separator equipment to separate coarse fibrous solids/sand",
```
elle-andreen: Completely optional but it might be helpful to add to the description: "Typical removal efficiency values for two types of separators (rotary screen and screw press) are included in the default manure management input file", just so we have it stated somewhere that these are intentional, appropriate defaults, not just placeholder-type values.

elle-andreen (approved):
I think this is a good step forward for separator configurability! We have discussed this elsewhere but I do think eventually implementing broader mechanical vs. gravity/non-mechanical separator would be very helpful, since, as you mentioned, the separator "type" right now does not really mean anything. I also am not sure if providing the default values for rotary screen and screw press separators in the input file is the best long-term place for them, but we can address that later. For now, I was able to define multiple separators and have them function as expected without any issues! Thanks Ed!

RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
Isgotkowitz: Nice

RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
Isgotkowitz: For now, I will leave them as-is - only because if I changed this for separators, I would want change it for beddings and handlers as well and this PR is large enough that I think it would be better to not make anymore changes that are strictly necessary.

RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
Isgotkowitz: Also same as above.

input/metadata/properties/default.json
Isgotkowitz: I think it might be better to leave this out, but state somewhere that this is the case for all configurations in the `manure_management.json` that is currently on the main branch.

PR #1603, opened 2024-05-02
[Util] Expanding logic to handle nested Dicts and Lists

PooyaHekmati:
### Context

Issue(s) closed by this pull request: closes N/A

- [ ] Yes
- [x] No

### What

Updates `convert_flat_dict_to_nested_dict` to `flatten_keys_to_nested_structure` and the functionality of `deep_merge` in Utility to allow handling Lists within Dicts. Also some minor clean ups to Utility.

### Why

Two new functions were introduced to the `Utility` class in PR #1599; they only handled Dicts; however, we need to be able to handle Lists as well for tasks such as Sensitivity Analysis.

### How

Extends the logic in  `Utility.convert_flat_dict_to_nested_dict` to be able to handle Lists well; then renames the function to `Utility.flatten_keys_to_nested_structure`. Does the same to `Utilty.deep_merge`; however, the name is still adequet.

Also, removes the methods in `Utility` that are no longer in use.

Finally, brings `test_make_serializable_recursive` from `test_misc` to `test_util` where it belongs.

### Test plan

Test Suite

### DIFF
--- a/RUFAS/util.py
+++ b/RUFAS/util.py
@@ -37,30 +37,53 @@ class Utility:
         return result
 
     @staticmethod
-    def convert_flat_dict_to_nested_dict(flat_dict: Dict[str, Any]) -> Dict[str, Any]:
+    def flatten_keys_to_nested_structure(input_dict: Dict[str, Any]) -> Dict[str, Dict | List]:
         """
-        Convert a flat dictionary with dot-separated keys into a nested dictionary structure.
+        Convert a dictionary with flat, dot-separated keys into a nested structure composed of
+        dictionaries and lists based on the keys. Numeric segments in the keys indicate list indices,
+        while non-numeric segments indicate dictionary keys.
 
         Parameters
         ----------
-        flat_dict : Dict[str, Any]
-            A dictionary with string keys that include dots to signify nested levels.
+        input_dict : Dict[str, Any]
+            A dictionary where the keys are strings that may include dots to signify hierarchical
+            levels in the resulting nested structure. Numeric key segments result in list creations,
+            and non-numeric segments result in dictionary creations.
 
         Returns
         -------
-        Dict[str, Any]
-            A nested dictionary where the structure is determined by splitting the keys on dots.
+        Dict[str, Union[Dict, list]]
+            A nested structure of dictionaries and lists derived by interpreting the flat dictionary keys.
         """
-        nested_dict = {}
-        for flat_key, value in flat_dict.items():
+        nested_structure = {}
+        for flat_key, value in input_dict.items():
             keys = flat_key.split(".")
-            current = nested_dict
-            for key in keys[:-1]:
-                if key not in current:
-                    current[key] = {}
-                current = current[key]
-            current[keys[-1]] = value
-        return nested_dict
+            current = nested_structure
+            for i, key in enumerate(keys[:-1]):
+                next_key_is_digit = keys[i + 1].isdigit() if i + 1 < len(keys) else False
+
+                if key.isdigit():
+                    key = int(key)
+                    while len(current) <= key:
+                        current.append([] if next_key_is_digit else {})
+                    current = current[key]
+                else:
+                    if isinstance(current, list):
+                        current = current[-1]
+                    if key not in current:
+                        current[key] = [] if next_key_is_digit else {}
+                    current = current[key]
+
+            last_key = keys[-1]
+            if last_key.isdigit():
+                last_key = int(last_key)
+                while len(current) <= last_key:
+                    current.append(None)
+                current[last_key] = value
+            else:
+                current[last_key] = value
+
+        return nested_structure
 
     @staticmethod
     def deep_merge(dict1: Dict[str, Any], dict2: Dict[str, Any]) -> None:
--- a/RUFAS/util.py
+++ b/RUFAS/util.py
@@ -22,7 +22,7 @@ class Utility:
             A dictionary where keys are unique keys from input dictionaries,
             and values are lists of corresponding values from input dictionaries.
         """
-        result = {}
+        result: Dict[str, List[Any]] = {}
 
         for item in list_of_dicts:
             for key, value in item.items():
--- a/RUFAS/util.py
+++ b/RUFAS/util.py
@@ -1,12 +1,8 @@
 import datetime
-import json
 import re
 import shutil
-import sys
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Tuple, Optional
-
-from RUFAS import errors
 
 
 class Utility:
@@ -37,7 +33,7 @@ class Utility:
         return result
 
     @staticmethod
-    def flatten_keys_to_nested_structure(input_dict: Dict[str, Any]) -> Dict[str, Dict | List]:
+    def flatten_keys_to_nested_structure(input_dict: Dict[str, Any]) -> Dict[str, Any]:
         """
         Convert a dictionary with flat, dot-separated keys into a nested structure composed of
         dictionaries and lists based on the keys. Numeric segments in the keys indicate list indices,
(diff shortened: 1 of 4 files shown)


RUFAS/util.py
```diff
@@ -37,123 +33,87 @@ def convert_list_of_dicts_to_dict_of_lists(list_of_dicts: List[Dict[str, Any]])
         return result
 
     @staticmethod
-    def convert_flat_dict_to_nested_dict(flat_dict: Dict[str, Any]) -> Dict[str, Any]:
+    def flatten_keys_to_nested_structure(input_dict: Dict[str, Any]) -> Dict[str, Any]:
         """
-        Convert a flat dictionary with dot-separated keys into a nested dictionary structure.
+        Convert a dictionary with flat, dot-separated keys into a nested structure composed of
+        dictionaries and lists based on the keys. Numeric segments in the keys indicate list indices,
```
JoeWaddell, line 39: Wonderfully clear explanation!

RUFAS/util.py
```diff
@@ -37,123 +33,87 @@ def convert_list_of_dicts_to_dict_of_lists(list_of_dicts: List[Dict[str, Any]])
         return result
 
     @staticmethod
-    def convert_flat_dict_to_nested_dict(flat_dict: Dict[str, Any]) -> Dict[str, Any]:
+    def flatten_keys_to_nested_structure(input_dict: Dict[str, Any]) -> Dict[str, Any]:
```
JoeWaddell, line 36: This name could be revised, since 'flatten' implies the method's action instead of an adjective. Maybe `flat_keys_to_nested_structure` or `flattened_keys_to_nested_structure`?

JoeWaddell (approved):
This looks great, and works as intended. Wonderful job!

The recursive nature of `deep_merge` - and the way it now handles lists and dicts differently - turned my brain inside out as I was reading the code and then testing different possibilities, but it looks good, and has handled everything I've thrown at it.

RUFAS/util.py
```diff
@@ -37,123 +33,87 @@ def convert_list_of_dicts_to_dict_of_lists(list_of_dicts: List[Dict[str, Any]])
...
 
     @staticmethod
-    def read_json_file(file_path: Path) -> Dict[Any, Any]:
+    def deep_merge(target: Dict[Any, Any], updates: Dict[Any, Any]) -> None:
         """
-        Description:
-            Reads and interprets the JSON file at the given path. Compiles the
-            information into dictionaries used to instantiate simulation objects.
+        Recursively merges 'updates' into 'target'. Supports deep merging for dictionaries and lists, including lists
+        that contain dictionaries and dictionaries that contain lists.
```
ew3361zh, line 88: I think it would be helpful to note in the description that merging lists here will overwrite the indices of the target with the indices of the updates rather than appending them to the original list.

RUFAS/util.py
```diff
@@ -37,123 +33,87 @@ def convert_list_of_dicts_to_dict_of_lists(list_of_dicts: List[Dict[str, Any]])
         return result
 
     @staticmethod
-    def convert_flat_dict_to_nested_dict(flat_dict: Dict[str, Any]) -> Dict[str, Any]:
+    def flatten_keys_to_nested_structure(input_dict: Dict[str, Any]) -> Dict[str, Any]:
         """
-        Convert a flat dictionary with dot-separated keys into a nested dictionary structure.
+        Convert a dictionary with flat, dot-separated keys into a nested structure composed of
+        dictionaries and lists based on the keys. Numeric segments in the keys indicate list indices,
+        while non-numeric segments indicate dictionary keys.
```
ew3361zh, line 40: It might be worth adding an Examples section to the docstrings that shows the functionality in action.

tests/test_util.py
```diff
@@ -281,3 +252,82 @@ def test_deep_merge() -> None:
...
+
+    b = {
+        "a": {"i": [11, 12, 13]},
+        "b": {"i": {"c": 15}, "j": {"d": [8, {"x": [19, 110]}]}},
+    }
+
+    expected = {
+        "a": {"i": [11, 12, 13], "j": {"c": 3, "d": 4}},
+        "b": {"i": {"c": 15, "d": 6}, "j": {"c": 7, "d": [8, {"x": [19, 110], "y": 11}, 12]}},
+    }
```
ew3361zh, line 271:
```suggestion
    a = {
        "a": {"i": [1, 2, 3, 4, 5], "j": {"c": 3, "d": 4}},
        "b": {"i": {"c": 5, "d": 6}, "j": {"c": 7, "d": [8, {"x": [9, 10], "y": 11}, 12]}},
    }

    b = {
        "a": {"i": [11, 12, 13]},
        "b": {"i": {"c": 15}, "j": {"d": [8, {"x": [19, 110]}]}},
    }

    expected = {
        "a": {"i": [11, 12, 13, 4, 5], "j": {"c": 3, "d": 4}},
        "b": {"i": {"c": 15, "d": 6}, "j": {"c": 7, "d": [8, {"x": [19, 110], "y": 11}, 12]}},
    }

```
Just a thought that this could help explain how the function handles lists.

ew3361zh (approved):
Overall looks great and seems to work as intended. 

I think some use-case examples in the PR description would be helpful as I'm not familiar with the SA work being done and why these changers are necessary (and there being no issue associated with it makes tracking the need for these changes more important).

I also added a few comments/suggestions that I think would improve understanding of these changes and what the functions are doing (and why).

JoeWaddell:
> I think some use-case examples in the PR description would be helpful as I'm not familiar with the SA work being done and why these changers are necessary (and there being no issue associated with it makes tracking the need for these changes more important).

Good point about documenting the need for this change: this was a necessary to handle inputs in which there are lists of dictionaries. Take `default_animal.json`, and `pen_information` inside that. There it's a list of dicts, and so to mutate data in the fourth pen (e.g. designating it as one of the `SA_input_variables` in `TaskManager`), we'd need to designate a key such as: `animal.pen_information.3.max_stocking_density`.

ltp23 (approved):
The two methods work as expected. Although I would love to see the code less deeply nested, the solutions do require that unless we break it down into smaller helper methods.

RUFAS/util.py
```diff
@@ -37,123 +33,87 @@ def convert_list_of_dicts_to_dict_of_lists(list_of_dicts: List[Dict[str, Any]])
...
+                    if key not in current:
+                        current[key] = [] if next_key_is_digit else {}
+                    current = current[key]
+
+            last_key = keys[-1]
+            if last_key.isdigit():
+                last_key = int(last_key)
+                while len(current) <= last_key:
+                    current.append(None)
+                current[last_key] = value
```
ltp23, line 78: Line 78 (current[last_key] = value) is repeated and can be moved outside the if-clause.

RUFAS/util.py
PooyaHekmati: Function names should always start with a verb.

RUFAS/util.py
PooyaHekmati: I think the name `target` and the fact that the function doesn't return anything are sufficient pieces of information for this.

RUFAS/util.py
JoeWaddell: Shouldn't it then be `unflatten`, since it's converting the input from a flattened structure to a nested structure?

RUFAS/util.py
PooyaHekmati: Examples and how-to-drive-the-codes should be obtained from the tests. We only add examples to docstrings when tests are not capable of providing that information and the function logic is complex. In this case, the unit test gives a clear example.

tests/test_util.py
PooyaHekmati: b.j.d already covers the case where update is shorter than target. Let's keep a.i as is to cover the case where target is shorter than update.

RUFAS/util.py
PooyaHekmati: Yes it can but it reduces readibilty. This logic is already complex enough.

PR #1606, opened 2024-05-02
Adding str() method to unit enum class

### LINKED ISSUE #1549: [Units] Add __str__ method to MeasurementUnits class
A `__str__` method is the better implementation vs. using the `.value` at the end of each instance of `MeasurementUnits` throughout the model.

This would allow using `MeasurementUnits` as a type hint in methods that pass it as an argument and would strengthen the check on whether the units accompanying a variable to OM is actually an approved `MeasurementUnits` unit.

## Expected Behavior

When referencing a measurement unit, it should be able to be referenced like this:
```
units = {"mass": MeasurementUnits.KILOGRAMS}
```
And when that gets added to OM and used in reports/graphs/etc, it will look like "kilograms".

## Current Behavior

Currently there is no __str__ method so every instance of MeasurementUnits looks like this:
```
units = {"mass": MeasurementUnits.KILOGRAMS.value}
```

## Possible Solution

The __str__ method should be added to the MeasurementUnits enum:
```
def __str__(self):
    return self.name
```
And then all references should be updated including updating type hints in functions.

ltp23:
### Context

Issue(s) closed by this pull request: closes #1549 

- [ ] Yes
- [x] No

### What

In this PR, I want to add a str() method for the units enum class so that when a units enum member is stringified, its value is returned instead of the default cumbersome dictionary string.

### Why

It'd be nice to simplify the usage of the units enum class by just when listing units in info maps.

### How

- Added a `__str__()` method in the `MeasurementUnits` enum class.
- Removed the use of `.value` when listing the units in the info maps.
- The `_validate_units()` method will convert a MeasurementUnits enum member into its 
corresponding string value.

### Test plan

- Make sure units still show up in csv and json output files.


RUFAS/output_manager.py
```diff
@@ -210,40 +210,52 @@ def add_variable(self, name: str, value: Any, info_map: Dict[str, Any]) -> None:
...
+
+        if type(units) is not MeasurementUnits:
+            raise TypeError(
+                f"The following unit does not have the type MeasurementUnits: {units} (type {type(units)})."
+            )
+
+        try:
+            return str(units)
+        except ValueError:
+            raise ValueError(f"'{units}' is not a valid MeasurementUnits value.")
```
ew3361zh: Can this ValueError be reached?

RUFAS/output_manager.py
```diff
@@ -953,11 +965,16 @@ def filter_variables_pool(self, filter_content: Dict[str, Any]) -> Dict[str, poo
         results: Dict[str, OutputManager.pool_element_type] = {}
         counter: int = 0
         for key in filtered_pool.keys():
+            sliced_info_maps: List[Dict[str, Any]] = (
+                filtered_pool[key]["info_maps"][slice_start:slice_end] if "info_maps" in filtered_pool[key] else []
+            )
```
ew3361zh, line 973: Why is this in this PR and what is it doing?

RUFAS/output_manager.py
```diff
@@ -210,40 +210,52 @@ def add_variable(self, name: str, value: Any, info_map: Dict[str, Any]) -> None:
...
-                self._validate_units(unit)
-        else:
-            try:
-                MeasurementUnits(units)
-            except ValueError:
-                raise ValueError(f"'{units}' is not a valid MeasurementUnits value")
+            return {key: self._validate_units(unit) for key, unit in units.items()}
+
+        if type(units) is not MeasurementUnits:
+            raise TypeError(
```
ew3361zh, line 258: Also log error to OM

ew3361zh:
A `.value` for a unit still present.

`animal_module_reporter.py` line 255:
```
AnimalModuleReporter.data_padder(
                f"{classname}.{funcname}.MEdiet_pen_0_CALF",
                f"{classname}.{funcname}.MEdiet_pen_{pen.id}_{pen.animal_combination.name}",
                0,
                simulation_day,
                info_map,
                MeasurementUnits.KILOGRAMS.value,
            )
```

ew3361zh:
A bunch in the`fail_summary_units` dict in `ration_driver.py` line 284:

```
fail_summary_units = {
            "simulation_day": MeasurementUnits.SIMULATION_DAY.value,
            "reattempt number": MeasurementUnits.UNITLESS.value,
            "constraints_failed_dict": MeasurementUnits.UNITLESS.value,
            "ration_attempted": MeasurementUnits.UNITLESS.value,
            "pen requirements": {
                "NEmaint_requirement": MeasurementUnits.MEGACALORIES.value,
                "NEa_requirement": MeasurementUnits.MEGACALORIES.value,
                "NEg_requirement": MeasurementUnits.MEGACALORIES.value,
                "NEpreg_requirement": MeasurementUnits.MEGACALORIES.value,
                "NEl_requirement": MeasurementUnits.MEGACALORIES.value,
                "MP_requirement": MeasurementUnits.GRAMS.value,
                "Ca_requirement": MeasurementUnits.GRAMS.value,
                "P_req": MeasurementUnits.GRAMS.value,
                "DMIest_requirement": MeasurementUnits.KILOGRAMS.value,
                "avg_BW": MeasurementUnits.KILOGRAMS.value,
                "avg_milk_production_reduction_pen": MeasurementUnits.KILOGRAMS.value,
            },
        }
```

ew3361zh:
`feed.py` line 208:

```
nutrient_dict_units = {
                "carbon": MeasurementUnits.KILOGRAMS.value,
                "nitrogen": MeasurementUnits.KILOGRAMS.value,
                "phosphorus": MeasurementUnits.KILOGRAMS.value,
                "dry_matter": MeasurementUnits.KILOGRAMS.value,
                "crude_protein": MeasurementUnits.KILOGRAMS.value,
                "carbon_loss": MeasurementUnits.KILOGRAMS.value,
                "crude_protein_loss": MeasurementUnits.KILOGRAMS.value,
            }
```

ew3361zh:
All references in `manure_handler_daily_output.py`, `manure_nutrients.py`, `manure_separator_daily_output.py`, `manure_trreatment_daily>output.py`, `pen_manure.py`, `reception_pit.py` need to be updated as well.

RUFAS/output_manager.py
ltp23: Right, I will remove the ValueError check because a ValueError would have been raised at the point of usage from some place else.

RUFAS/output_manager.py
ltp23: Okay. I will add an error log entry for this.

RUFAS/output_manager.py
ltp23: So if you run the main branch, you should notice the units are not included in the output. The units are located inside the info maps but the info maps are not being filtered along with their corresponding values.

RUFAS/output_manager.py
```diff
@@ -210,40 +210,58 @@ def add_variable(self, name: str, value: Any, info_map: Dict[str, Any]) -> None:
...
+        -------
+        Dict[str, Any] | str
+            The validated and stringified units.
+
         Raises
         ------
+        TypeError
+            If any unit or nested unit does not have the type MeasurementUnits.
         ValueError
             If any unit or nested unit is not a valid MeasurementUnits value.
```
```suggestion
```

RUFAS/output_manager.py
```diff
@@ -210,40 +210,58 @@ def add_variable(self, name: str, value: Any, info_map: Dict[str, Any]) -> None:
...
         key = self._generate_key(name, info_map)
-        self._add_to_pool(self.variables_pool, key, value, info_map)
+        self._add_to_pool(self.variables_pool, key, value, {**info_map, "units": units})
 
         if isinstance(value, dict):
             for k, v in value.items():
                 self._variables_usage_counter[f"{key}.{k}"] = 0
 
-    def _validate_units(self, units: Dict[str, Any] | str) -> None:
+    def _validate_units(self, units: Dict[str, Any] | MeasurementUnits) -> Dict[str, Any] | str:
```
Isgotkowitz: Instead of having the behavior of this method change, I think it would be better to keep it as just a validator for units and put the responsibility of stringifying the units be on whatever library is writing it to a text file. This does get a little more tricky to get the units printing correctly though (because of the JSON library) but the following steps can be taken:
- have `MeasurementUnits` inherit `str`
- add the argument `ensure_ascii=False` to the `json.dump` method calls in OutputManager.

This is just a suggestion though, not a necessary change. If you keep it like this, I would just suggest changing the name to "stringify_units" - because it would be a little more reflective of what it is actually doing.

Isgotkowitz (approved):
Looks good, and functions as expected. Before merging, I think the code that changes the info maps filtering should be taken out and left to #1638.

I left one suggestion about changing the behavior of validate_units back to its original behavior, but that is take it or leave it.

ew3361zh (changes requested):
Hey Loi, updates to the PR are looking good. 

I'm still seeing the `.value` for MeasurementUnits instances in a few places that I'd mentioned in comments prior. Are they not supposed to be changed for some reason?
1. Line 255 in `animal_module_reporter.py`.
2. Line 306 in`ration_driver.py` for each of the units in the `fail_summary_units` dict.
3. Line 208 in `feed.py` for each of the units in the `nutrient_dict_units` dict.

RUFAS/output_manager.py
ltp23: Actually, I already removed the ValueError check.

RUFAS/output_manager.py
ltp23: `MeasurementUnits` is an Enum class, so switching to it to inheriting from the str class is quite a change and would lose the automatic membership check or some workaround would be needed. 

Whenever we add a variable to the variables pool, we recursively validate and stringify the corresponding units. I think it is more efficient to intertwine those two steps like we have here and only do it once. This string conversion is done here before all downstream, post-simulation output processing. 

The tradeoff here is that there is consistency in how units are handled but we would lose some customization later on because units are no longer of the original enum type.

In the meantime, I will at least rename it to stringify-units and I agree calling it validate-units seems to miss the stringifying step.

ltp23:
Hi @ew3361zh, thanks for catching those. I have removed those loose .value calls. The reason they were uncaught was that those variables were only added depending on the current states of the simulation.

ew3361zh (approved):

PR #1607, opened 2024-05-02
Hardcode init function name in info maps

### LINKED ISSUE #1394: Mypy error raised when adding function name with `__init__` in `info_map`
## Urgency Level

- [x] Low: the issue doesn't have a significant immediate impact

## Status

Mypy error raised when adding function name with `__init__` in `info_map`

image

## Intention

Either this shouldn't be flagged as an error, or we write the `self.__init__.__name__` in a different way.

## Request

One option is "telling" mypy to ignore this error throughout the codebase, but another is simply changing the way we enter data when using `__init__`, e.g. either manually writing the resulting string, or baking the method into OutputManager itself (e.g. recognizing where it was called automatically).

ltp23:
### Context

Issue(s) closed by this pull request: closes #1394 

- [ ] Yes
- [x] No

### What

We want to find a way to fix this mypy error: Accessing `__init__` on an instance is unsound, since instance.init could be from an incompatible subclass. This error suggests that calling the `__init__` method directly on an instance of a subclass can have ambiguous behavior. If the subclass overrides the init method in the base class, calling the init method will get the subclass' version. Otherwise, we will get the base class version. 

However, the way we are calling the init method to get its name here (basically just the string `__init__`) should have nothing related to the safeguard warning above from mypy. 

### Why

Without hardcoding the init string name, the code still works, so this is more of a tidying up of things.

### How

- Use a hard-coded string instead of `self.__init__.name`.


Isgotkowitz (approved):
Nice

allisterakun (approved):
Simple PR, good job, Loi!

PR #1609, opened 2024-05-03
Some test coverage for Task manager

PooyaHekmati:
Let's add some test cases for the Task Manager. It's not as comprehensive as it should be, but we have to start from somewhere.

- [ ] Yes
- [x] No


ew3361zh (approved):
LGTM

tests/test_task_manager.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+@pytest.mark.parametrize(
+    "input_str, expected",
+    [
+        ("Herd Initialization", TaskType.HERD_INITIALIZATION),
+        ("simulation single run", TaskType.SIMULATION_SINGLE_RUN),
+        ("simulation multi RUN", TaskType.SIMULATION_MULTI_RUN),
+        ("sensitivity analysis", TaskType.SENSITIVITY_ANALYSIS),
+        ("input data Audition", TaskType.INPUT_DATA_AUDITION),
```
Isgotkowitz: I did not see this before, but "audition" isn't quite the right word here, "audit" would be more appropriate. However this a little outside the scope of this PR.

Isgotkowitz (approved):
Tests here look good! However they are adding a lot of Mypy errors, I think it would be good to find a way to write these tests without increasing the error count or having to silence all of them.

PR #1615, opened 2024-05-07
Update replacement animals to heiferIIIs and use a separate male calf rate in herd init

### LINKED ISSUE #1395: [Herd Generation] Herd generation fails when number of simulated days and initial animal count are below certain threshold
If the number of days that a herd is simulated for is low enough, the Herd Factory will crash

## Expected Behavior

RuFaS should be given any valid combination of initial animals in a herd and number of days to simulate and generate a replacement herd based on those inputs.

## Current Behavior

When the initial number of animals and days to simulate them for are low enough, the Herd Factory crashes.

## Steps to Reproduce

1. Navigate to the `main` branch
2. Set `"initial_animal_num"` to 100 and `"simulation days"` to 4865 in the `"herd_initialization"` object in `default_animal.json`.
3. Run the default scenario with the flags `-I -s -t -v errors`
4. Observe results. There should be an error produced that is similar to:

herd_generation_error

### LINKED ISSUE #1621: Update replacement animals to heiferIIIs and use a separate male calf rate in herd init
This issue was created after the PR had been made.

Tasks to accomplish:
- Change the type of replacement animals from cows to heiferIIIs in herd factory and life cycle modules.
- Reduce the frequency of assigning heiferIIIs to the replacement group because this would affect the stability of the herd size and distribution over time.
- Create a separate parameter for male calf rate so that it is used only during herd initialization. The male calf rate indirectly determines how many calves we start with. To be more exact, male calves will be sold so the complement of that rate (1 - male calf rate) will give us a rough estimate of the number of female calves. And then during the herd generation process, this male calf rate will be used again to determine the gender of the newborn calves. The reason we want a separate parameter is that a very low rate is much more preferred and effective in herd generation, whose main purpose is to make cows and heifers.

ltp23:
### Context

Issue(s) closed by this pull request: closes #1395, #1621 

- [ ] Yes
- [x] No

### What & Why

Currently, the type of replacement animals in the herd generation and the life cycle modules is Cow and we should change that to the more appropriate type of HeiferIII.

During the herd initialization, currently, we are not incorporating the simulation day info and using the value of 0 instead, when calling the update methods on each animal, so we should do that also. 

### How

Main changes
- In `herd_factory.py`,
  - Add a parameter for `simulation_day` in each of the animal update methods: `_calves_update()`, `_heiferIs_update()`, etc.
  - A change in the logic for deciding who to put in the replacement group. Before, if a cow has just graduated from being a heiferIII and the current simulation day is 3000, she will be assigned to the replacement group. Now that replacement animals are heiferIIIs, I use a custom logic that randomly decides whether a heiferIII will remain in the herd or in the replacement group (1% chance, for example). A low rate for the latter is probably better because once a heiferIII is taken out of the herd, she won't reenter so the herd size may decrease over time but this can be mitigated by setting the following male calf rate low.
  - A simple but very important parameter to make the initialization process run well is the male calf rate. This parameter determines two things: 1) the starting calf number 2) how likely calves will be kept for the next generation. The _lower_ the male calf rate the better because male calves will get sold. When this rate is 0, all the starting calves will be female and all the newborn calves will also be female. The fate of a female calf is only mildly affected by two other variables `keep_female_calf_rate` and `still_birth_rate`.
- In `animal_population.py`, change all the references of cows as replacements to heiferIIIs.
- Add a new animal population json file — `animal_population_updated.json`.
- Update all the metadata files's `animal_population` property to point to the `animal_population_updated.json` file.
- Remove specific details about repro subprotocols stored on heifers and cows because they are not being used in the repro part of the life cycle.

Small changes
- Fix a typo in the name of the method `handle_herd_initialization` in `TaskManager`.
- There is a small error with the regex pattern for checking the save path for herd init output json.

### Test plan

An example of herd initialization task

```
{
      "task_type": "HERD_INITIALIZATION",
      "output_prefix": "Default Metadata Herd Init",
      "log_verbosity": "errors",
      "random_seed": 42,
      "init_herd": true,
      "save_animals": true,
      "save_animals_directory": "input/data/animal"
}
```

In the `herd_initialization` section of animal.json

```
"herd_initialization": {
    "male_calf_rate_for_herd_init_only": 0.1,
    "heiferIII_conversion_rate_to_replacement": 0.01,
    "initial_animal_num": 10000,
    "simulation_days": 5000
},
```


RUFAS/routines/animal/life_cycle/herd_factory.py
```diff
@@ -343,6 +347,18 @@ def _random_sample_with_replacement_by_type(
         animal_num = im.get_data(ANIMAL_NUM_KEY[animal_type])
 
         post_animals = []
+
+        if len(pre_animals) == 0 and animal_num > 0:
```
Isgotkowitz, line 351: If `pre_animals` is less than `animal_num`, maybe it would be good to raise a warning indicating that some animals will be repeated in the animal population.

input/data/animal/barnyard_animal.json
```diff
@@ -10,6 +10,8 @@
 		"breed": "HO"
 	},
 	"herd_initialization": {
+		"male_calf_rate_for_herd_init_only": 0.1,
+    	"heiferIII_conversion_rate_to_replacement": 0.01,
```
Isgotkowitz, line 14:
```suggestion
		"male_calf_rate_for_herd_init_only": 0.1,
    	        "heiferIII_conversion_rate_to_replacement": 0.01,
```

input/data/animal/no_animal.json
```diff
@@ -10,6 +10,8 @@
     "breed": "HO"
   },
   "herd_initialization": {
+  	"male_calf_rate_for_herd_init_only": 0.1,
+    "heiferIII_conversion_rate_to_replacement": 0.01,
```
Isgotkowitz, line 14:
```suggestion
    "male_calf_rate_for_herd_init_only": 0.1,
    "heiferIII_conversion_rate_to_replacement": 0.01,
```

input/metadata/properties/default.json
```diff
@@ -92,6 +92,20 @@
     "herd_initialization": {
       "type": "object",
       "description": "Animal generation related inputs",
+      "male_calf_rate_for_herd_init_only": {
+        "type": "number",
+        "description": "The probability of a new calf being male. This is only used for the initial herd generation and is not used for the rest of the simulation. A low value is recommended because male calves will get sold, whereas a female calf is much less likely to be removed from the herd. Two other variables that have mild effects on the count of female calves are keep_female_calf_rate and still_birth_rate.",
```
Isgotkowitz, line 97:
```suggestion
        "description": "The probability of a new calf being male during herd generation only. A low value is recommended because male calves will get sold, whereas a female calf is much less likely to be removed from the herd.",
```
These are useful notes about keep_female_calf_rate and still_birth_rate, but they should be in their own descriptions.

Isgotkowitz (changes requested):
A couple of things that should be addressed in this PR:
- It does not close #1395. The error message is very helpful when herd generation attempts to gather animals which are not there and setting the male calf keep rate for herd initialization is helpful, but it does not explicitly prevent this from happening.
- It seems wrong to put the simulation day in for animals that are generated in herd initialization. Simulation day means something very specific in RuFaS - it is the number of days into a RuFaS simulation that something happened. Putting in the day of herd generation would be incorrect, especially if there is no way to indicate that an animal was created before the simulation actually started.
- The removal of repro subprotocols in heifers and cows seems like a significant change, is it strictly necessary to make all of these changes in order to convert replacements from HeiferIIIs to Cows?

Overall, I think this PR has too many responsibilities - it should be split up so that each issue it addresses has its own PR. Also there is not an explanation of why replacements should be HeiferIIIs instead of Cows and no explanation of how to see or verify the effect of this change in animal outputs.

ltp23:
> * It does not close [[Herd Generation] Herd generation fails when number of simulated days and initial animal count are below certain threshold #1395](https://github.com/RuminantFarmSystems/MASM/issues/1395). The error message is very helpful when herd generation attempts to gather animals which are not there and setting the male calf keep rate for herd initialization is helpful, but it does not explicitly prevent this from happening.

The main reason for the herd initialization to fail as described in that issue is that the parameter `male_calf_rate_conventional_semen` is too high (0.53). For the herd init process, this rate determines how many calves we start with (47% of the initial animal number), and the number of calves sold and disappeared from the next generation. Another factor is if too many animals become replacements, there will no animals left at the end of the process to sample.

ltp23:
> * It seems wrong to put the simulation day in for animals that are generated in herd initialization. Simulation day means something very specific in RuFaS - it is the number of days into a RuFaS simulation that something happened. Putting in the day of herd generation would be incorrect, especially if there is no way to indicate that an animal was created before the simulation actually started.

I think there is minimal distinction between a simulation day during the herd init process and a simulation day in the normal simulation. The herd init process is just a focused mini-simulation that deals with the animal life cycle exclusively, and then the normal simulation is a continuation of that. 

Actually, you can tell whether a simulation day refers to a day during the herd init process or in the normal simulation by looking at the animal events history. 

```
# This branch
...
"1260": [
  "simulation_day=4812",
  "pregnancy check 2"
],
"1400": [
  "simulation_day=4952",
  "pregnancy check 3"
],
"1417": [
  "simulation_day=4969",
  "dry"
],
"1447": [
  "entered herd through initialization"
],
"1479": [
  "simulation_day=32",
  "new birth, start milking",
  "number of calves born so far: 3",
  "1st estrus after calving: estrus day scheduled on day 1509",
  "Current repro state(s): fresh"
],
"1509": [
  "simulation_day=62",
  "estrus occurred before the end of the voluntary waiting period",
  "estrus day scheduled on day 1526"
],
"1526": [
  "simulation_day=79",
  "estrus occurred before the end of the voluntary waiting period",
  "estrus day scheduled on day 1550"
],
...
```

```
# Main branch
# Simulation day is omitted when it is 0.
...
"1964": [
  "pregnancy check 2"
],
"2104": [
  "pregnancy check 3"
],
"2112": [
  "entered herd through initialization"
],
"2121": [
  "simulation_day=9",
  "dry"
],
"2175": [
  "simulation_day=63",
  "new birth, start milking",
  "number of calves born so far: 5",
  "1st estrus after calving: estrus day scheduled on day 2198",
  "Current repro state(s): fresh"
],
"2198": [
  "simulation_day=86",
  "estrus occurred before the end of the voluntary waiting period",
  "estrus day scheduled on day 2221"
],
...
```

input/data/animal/animal_population.json
jadamchick, line 1: It looks as though you opted to replace the existing animal_population.json rather than adding a new file named "updated" and changing the file paths. Why did you need to replace this, and how is this population different than the previous one?

input/data/animal/default_animal.json
```diff
@@ -10,7 +10,9 @@
     "breed": "HO"
   },
   "herd_initialization": {
-    "initial_animal_num": 10000,
+    "male_calf_rate_for_herd_init_only": 0.1,
+    "heiferIII_conversion_rate_to_replacement": 0.01,
+    "initial_animal_num": 1000,
```
jadamchick, line 15: Not sure if you intentionally changed this from 10000 to 1000?

jadamchick (commented):
Hi Loi, 
I agree that there is a lot here for one PR, and I don't have a great sense of how to test the changes made in a targeted way. Do you have thoughts on how to test: 
- replacements entering as heiferIII instead of cow
- implications of simulation day info included for update methods 
- implications of the new animal_population.json -- is this population different than the previous one ?
- implications of changing the details about the repro subprotocols 

I ran a few different scenarios to make sure it wasn't breaking anything that previously worked: 
1. runs okay with default herd
2. runs okay with random pilot testing herd (c2f4)
3. DOES NOT run okay with Jersey pilot testing herds (c2f19, c2f28); these herds run fine on pilot_testing_Dec2023 branch (UTD with Main as of this morning). 
When I merge this branch into a copy of pilot_testing_Dec2023, they give the following error: 
[c2f28 jersey- pt explore branch 15may 2nd run_errors_15-May-2024_Wed_16-20-22.json](https://github.com/RuminantFarmSystems/MASM/files/15326722/c2f28.jersey-.pt.explore.branch.15may.2nd.run_errors_15-May-2024_Wed_16-20-22.json) 
Note that the Jersey herds use a different animal population file, fwiw. 


Specific to the question of why make the change to replacements entering as heiferIII's instead of cows: I don't know why this was previously cows, since the documentation I have seen (and RuFaS outputs) refer to purchasing heiferIII's when replacement animals are needed. And therefore I don't feel that I have a full sense of what impacts to expect when this update is made. 

One thing I do anticipate/assume - that there will be fewer instances of cows in the herd with sporadic attributes (eg repro protocol or similar) that came from the generated population instead of the animal.json. Because they will be entering as heiferIII's, then they will be assigned all of their "cow attributes" when they calve in as first lactation cows. But that is not something I ever empirically observed, just intuited/assumed as I learned more about how RuFaS was working. 
My assumption now is that: cows who were purchased as replacements when they were heiferIII's will have 2 attributes from the generated population (that are assigned at birth and remain unchanged throughout life): breed and mature body weight. All other attributes will be assigned from the animal json. Breed will have 2 impacts: lactation curve parameters, and her calves' birth weight.

ltp23:
I will go over the feedback more in-depth over the weekend. The bits and pieces I know I will comment first. I'm also working on a smaller PR to keep only the essentials from this PR.

input/data/animal/default_animal.json
ltp23: Yes, I can revert the initial number to 10000. With the new additional parameters, the herd init process is more efficient so 1000 is good enough. The general consideration is how diverse we want the herd to be so we can sample from it.

input/data/animal/animal_population.json
ltp23: At first, I wanted an updated version to be there along with the current `animal_population.json`. Then, it is somewhat confusing to have the two coexisting. The main difference is that the new version has heiferIIIs as replacement animals whereas before they were cows.

ltp23:
> One thing I do anticipate/assume - that there will be fewer instances of cows in the herd with sporadic attributes (eg repro protocol or similar) that came from the generated population instead of the animal.json. Because they will be entering as heiferIII's, then they will be assigned all of their "cow attributes" when they calve in as first lactation cows. But that is not something I ever empirically observed, just intuited/assumed as I learned more about how RuFaS was working.
> My assumption now is that: cows who were purchased as replacements when they were heiferIII's will have 2 attributes from the generated population (that are assigned at birth and remain unchanged throughout life): breed and mature body weight. All other attributes will be assigned from the animal json. Breed will have 2 impacts: lactation curve parameters, and her calves' birth weight.

I agree with this observation. Replacement animals entering as HeiferIIIs help make the cow population more uniform because when heiferIIIs transition to the cow stage, they will take on the attributes as specified by animal json with the two main exceptions as you mentioned.

PR #1617, opened 2024-05-07
Update regional diets

harry-samuels:
<Update to the close_up and growing feed lists of the MW/W/E Diets-->

- [ ] Yes
- [ ] No

### What
New feeds have been added to make the diets more complete


input/metadata/properties/default.json
```diff
@@ -3044,7 +3044,7 @@
       "properties": {
         "type": "number",
         "minimum": 1,
-        "maximum": 300,
+        "maximum": 1000,
```
Isgotkowitz, line 3053: Why is this being changed? Same question for all of the increases in feed numbers.

input/data/feed/feed_MW_Diet_15.json
```diff
@@ -0,0 +1,326 @@
...
+      "storage_type": "bunker",
+      "moisture": "direct_cut",
+      "additive": "preservative",
+      "packing_density": 14,
+      "inoculation": "heterofermentative",
+      "bunk_type": "open_floor",
+      "ventilation": true,
+      "removal_rate": 6,
+      "initial_dry_matter": 0
+    }
```
Isgotkowitz: The storage type, moisture, and packing density are not correctly specified and are being changed to defaults. Same for the other two feed files.

Isgotkowitz (changes requested):
Some options for the storages are not being specified correctly and are being replaced with defaults. Also it is unclear why the metadata properties are being updated in this PR.

input/metadata/properties/default.json
tomhuhh: Hi Ed! Thanks for reviewing this. I personally think a maximum of 300 feed ingredients is not practical, and a higher threshold should be set. In the case of the pilot piloting branch, I believe there is a total of 306 ingredients.

tomhuhh: However, this change may also need to happen on the main branch level. Thanks!

jadamchick: On pilot testing, we've increased the max to 306 because there are 306 feeds in the library. Right now there isn't a plan to merge feeds 301-305 into Main because they are specific to FARM-ES. I think 306 is adequate, unless there are plans to add more feeds? (though I don't see any harm to a higher ceiling, either)

jadamchick (changes requested):
Hi Harry, 
It looks like the metadata properties default.json is outdated and would revert some updates that have been made (i.e. it doesn't align with the default metadata properties json on either Main or Pilot_testing). Can you please make sure this is UTD with Main? (or even with pilot testing, which would incorporate the max feed_properties value of 306 already there)

tomhuhh:
Deprecating this PR because it is outdated. Updated PR can be found here: https://github.com/RuminantFarmSystems/MASM/pull/1667

@Isgotkowitz @jadamchick I have already tagged you on the new PR. Thanks everyone for your efforts!

PR #1575, opened 2024-04-23
[OM] Make use of VM in OutputManager

allisterakun:
Users can periodically dump the current variable pool and save a chunk into a JSON file. This will allow us to break down a larger simulation pool into smaller chunks.

### Context

Issue(s) closed by this pull request: closes #1513

Users can set the `chunkification` flag to Ture in the tasks input JSON file to enable the functionality.
Then users can specify one of the following in the tasks input JSON file for each task:
1. "save_chunk_threshold_call_count": 
      ```json
      {
        "type": "number",
        "description": "The threshold adds variable call count for saving the output variable pool chunk.",
        "default": 0
      }
      ```
      If this variable is specified, the OM will keep track of the number of calls to the `OM.add_variable()` function. Once it reaches the threshold number, the current variable pool will be dumped, and the other metrics will be ignored.
2. "maximum_memory_usage": 
      ```json
      {
        "type": "number",
        "description": "The maximum memory usage in Bytes.",
        "default": 0
      }
      ```
      If `save_chunk_threshold_call_count` is not specified but this variable, the OM will set `OM.max_pool_size` to the specified `maximum_memory_usage` amount. The OM will estimate the current pool size every time the `OM.add_variable()` function is called and check against the max pool size. If the threshold is reached, the current variable pool will be dumped.

      The OM will accumulate the average size addition per add_variable call to estimate the current pool size.

      If multiple workers are defined, the `TaskManager` will treat each task equally and evenly distribute the max_memory_usage to all tasks. This means if the max_memory_usage is 8 * (1024 **3) Bytes and there are 8 workers; each worker will get 1 * (1024 **3) Bytes of max memory
3. "maximum_memory_usage_percent": 
      ```json
      {
        "type": "number",
        "description": "The maximum percentage of memory use.",
        "default": 80
      }
      ```
      
      If neither `save_chunk_threshold_call_count` nor `maximum_memory_usage` is specified, we will use `maximum_memory_usage_percent` to determine the maximum memory usage by simply multiplying it with the available memory. The OM will estimate the current pool size the same way as the last section, and this percentage number is also evenly distributed across all workers.

Here is a logic diagram:
image
https://whimsical.com/om-chunkification-32XPBmYkA5ncJmVytzLQdT

### Test plan

Updated and added new unit tests to cover all functions modified/added.

I changed the `input/data/tasks/default_task.json` to the following scenarios, they all run with no errors. 
```JSON
{
  "parallel_workers": 4,
  "tasks": [
    {
[...]

PR #1631, opened 2024-05-08
[OM] Filter Multiple Pools

allisterakun:
Creates a function to apply the same filter to multiple save pool chunks and aggregate the result.

### Context

- [ ] Yes
- [x] No

### What & How

1. New properties in the `OutputManager`:
    1. `OM.chunkification: bool`
        1. This boolean flag indicates if the OM would save the variable into chunks of JSON files at `OM.saved_pool_chunks_path`.
    2. `OM.saved_pool_chunks_num : int`
        1. This is a counter to keep track of the number of chunks saved.
    3. `OM.saved_pool_chunks_path: Path`
        1. This variable specifies the path to which the chunks of the variable pool will be saved.
2. New methods in the `OutputManager`:
    1. `OM._save_current_variable_pool()`
        ```python
        def _save_current_variable_pool(self) -> None:
                """
                Save the current variable pool into a JSON file. Flush the variable pool and reset the pool size.
                """
        ```
    2. `OM._sort_saved_chunk_files()`
        ```python
        def _sort_saved_chunk_files(self) -> List[Path]:
                """
                Get a list of all saved chunks of the output variable pool by retrieving all JSON files under
                the saved_pool_chunks_path. Then sort the files according to their file name to preserve the order.
                """
        ```
    3. `OM.filter_saved_pools()`
        ```python
        def filter_saved_pools(
                self, filter_content: Dict[str, Any], list_of_dumped_files: List[Path]
            ) -> Dict[str, OutputManager.pool_element_type]:
                """
                Filters saved pools of data by applying specific filter criteria.
                This method iterates over JSON files in the saved pool directory. It then loads each file and applies the
                filter by calling the `filter_variables_pool()` method. The results are aggregated into a single dictionary,
                combining entries under the same key by extending lists of info_maps and values.
                Parameters
                ----------
                filter_content : (Dict[str, Any])
                    A dictionary specifying the criteria used to filter the variables pools.
                list_of_dumped_files: List[Path]
                    A list containing all chunks of the output variable pool to be filtered.
                Returns
                -------
                Dict[str, OutputManager.pool_element_type]:
                    A dictionary containing the aggregated filtered pool elements after applying the filter to all JSON files
                    under the saved_pool_chunks_path directory.
                """
        ```
3. Changes the logic in `OM.save_results()`
    1. If `OM.chunkification` is True, the current variable pool will first be dumped into a JSON file and saved to `OM.saved_pool_chunks_path` by calling `OM._save_current_variable_pool()`.
    2. If `OM.chunkification` is True, we will get a list of saved pool chunks and sort them according to their file name by calling `OM._sort_saved_chunk_files()`. We will pass this list of sorted files, along with the filter content, to  `OM.filter_saved_pools()` to apply the filter to all saved pools.

### Why

This function allows applying the same filter to multiple pool chunks. This enables us to break down larger simulations into smaller ones and apply the filter at the end to evaluate the results.


RUFAS/output_manager.py
```diff
@@ -233,6 +238,28 @@ def add_variable(self, name: str, value: Any, info_map: Dict[str, Any], first_in
...
+        """
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self._save_current_variable_pool.__name__,
+        }
+
+        self.create_directory(self.saved_pool_chunks_path)
+        saved_pool_file_name = self.generate_file_name(f"saved_pool_{self.saved_pool_chunks_num}", "json")
+        saved_pool_file_path = Path.joinpath(self.saved_pool_chunks_path, saved_pool_file_name)
+        self.dict_to_file_json(data_dict=self.variables_pool, path=saved_pool_file_path, minify_output_file=False)
```
```suggestion
        self.dict_to_file_json(data_dict=self.variables_pool, path=saved_pool_file_path, minify_output_file=True)
```
Would minifying these files be better? It would mean less space taken up by the files, and might help save a little time reloading them.

tests/test_output_manager.py
```diff
@@ -1778,6 +1790,10 @@ def test_save_results(
     route_save_functions = mocker.patch.object(mock_output_manager, "_route_save_functions")
     add_error = mocker.patch.object(mock_output_manager, "add_error")
     mock_output_manager.time = MagicMock()
+    mock_output_manager.chunkification = chunkification
+    mock_output_manager._save_current_variable_pool = MagicMock()
+    mock_output_manager._sort_saved_chunk_files = MagicMock()
+    mock_output_manager.filter_saved_pools = MagicMock(return_value={})
```
Isgotkowitz: Instead of setting these equal to MagicMock's, could you patch them with `mocker.patch.object`. The benefit of doing it like this is that Mypy will not take issue with it.

RUFAS/output_manager.py
```diff
@@ -233,6 +238,28 @@ def add_variable(self, name: str, value: Any, info_map: Dict[str, Any], first_in
...
+        self.add_log(
+            "save_current_variable_pool",
+            f"Saved the current variable pool to {saved_pool_file_path}",
+            info_map,
+        )
+        self.variables_pool = {}
+        self.current_pool_size = sys.getsizeof(self.variables_pool.__repr__())
```
Isgotkowitz, line 261: Would it be more convenient to have `current_pool_size` be an `@property`?

tests/test_output_manager.py
```diff
@@ -2612,3 +2639,203 @@ def test_set_exclude_info_maps_flag(flag_value: bool) -> None:
...
+    mock_add_log.assert_called_once_with("save_current_variable_pool", log_message, info_map)
+
+    assert output_manager.variables_pool == {}
+    assert output_manager.current_pool_size == sys.getsizeof(output_manager.variables_pool.__repr__())
+    assert output_manager.saved_pool_chunks_num == 1
+
+    output_manager.create_directory = output_manager_original_method_states["create_directory"]
+    output_manager.generate_file_name = output_manager_original_method_states["generate_file_name"]
+    output_manager.dict_to_file_json = output_manager_original_method_states["dict_to_file_json"]
+    output_manager.add_log = output_manager_original_method_states["add_log"]
```
Isgotkowitz:
```suggestion
```
Because these methods are patched with the MockerFixture, they do not need to be reset.

tests/test_output_manager.py
```diff
@@ -2612,3 +2639,203 @@ def test_set_exclude_info_maps_flag(flag_value: bool) -> None:
...
+
+
+def test_sort_saved_chunk_files(mock_output_manager: OutputManager, tmpdir) -> None:
```
Isgotkowitz, line 2722: Type annotation for tmpdir, if possible.

RUFAS/output_manager.py
```diff
@@ -1016,7 +1043,55 @@ def filter_variables_pool(self, filter_content: Dict[str, Any]) -> Dict[str, poo
...
+            for key, value in temp_filtered_pool.items():
+                if key in filtered_pool.keys():
+                    filtered_pool[key]["info_maps"].extend(value["info_maps"])
+                    filtered_pool[key]["values"].extend(value["values"])
+                else:
+                    filtered_pool[key] = value
+
+        return filtered_pool
+
+    def save_results(  # noqa: C901
```
Isgotkowitz, line 1159: Would it be possible to break up this function so that it doesn't become overly complex?

Isgotkowitz (commented):
Overall, I think this looks good. I tried running the same simulation on both this branch and main, and there was no difference in outputs. Left some comments here and there, but I saw nothing serious that needs to be addressed.

The test suite looks sound, however I think there are a lot of mocks in it that are causing increases in the number of Mypy errors. Ideally that number wouldn't go up at all, but if this PR needs to go through then that is something that can be dealt with later.

RUFAS/output_manager.py
```diff
@@ -133,6 +133,11 @@ def __init__(self) -> None:
                 "report": "report_",
             }
             self.__log_verbose: LogVerbosity = LogVerbosity.CREDITS
+
+            self.chunkification: bool = False
+            self.saved_pool_chunks_num: int = 0
+            self.saved_pool_chunks_path: Path | None = None
```
Isgotkowitz, line 140: Could these be added to the attributes section of the docstring?

tests/test_output_manager.py
```diff
@@ -2665,3 +2685,186 @@ def test_set_exclude_info_maps_flag(flag_value: bool) -> None:
...
+        tmpdir.join("saved_pool_3_dummy_timestamp.json"),
+    ]
+
+
+def test_filter_saved_pools(
+    mock_output_manager: OutputManager,
+    tmpdir,
+    output_manager_original_method_states: Dict[str, Callable],
+    mocker: MockerFixture,
+):
```
Isgotkowitz, line 2744:
```suggestion
) -> None:
```

Isgotkowitz (approved):
LG8TM, nice job Allister.

RUFAS/output_manager.py
```diff
@@ -1062,6 +1137,8 @@ def save_results(
         )
         list_of_filter_files = self._list_filter_files_in_dir(filters_dir_path)
         report_generator = ReportGenerator(self.time)
+        if self.chunkification:
+            self._save_current_variable_pool()
```
ew3361zh, line 1206: I think I'm confused on how `_save_current_variable_pool()` and `save_results()` will work together - can you explain the end vision for that process? 
Right now with `_save_current_variable_pool()` nested within `save_results()`, it's only going to save a chunk when you save the full results. Are we planning to have `save_results()` called every time we want a chunk saved? 

If so it seems like we'd want to move out the rest of the logic about graphing and report generation somewhere else.

I know Pooya is planning some larger changes with RG and OM so maybe this is working concert with those plans but I'm a little confused at how chunkification is going to work with the way `save_results()` is set up.

ew3361zh:
Overall this looks pretty good Allister. 

I tested this by turning on `self.chunkification` and adding a Path for `self.saved_pool_chunks_path` and comparing the results with main and the feature branch.

I have one outstanding question about how that chunkification process is supposed to work (left as a comment) and to what degree it's supposed to be set up in this PR. But I know the difference will hopefully be seen when we're saving the results multiple times within a longer simulation.

I'm good to approve once that question is addressed.

ew3361zh:
I'm still not sure how this will work with the way `OM._save_current_variable_pool()` is called within `OM.save_results()`.

Currently there is only one place `OM.save_results()` is used which is in `TM.handle_post_processing()`.

If `OM.save_results()` is only called once at the end of a simulation, how will this enable this logic `OM._save_current_variable_pool()` to save more than one pool?

RUFAS/output_manager.py
ew3361zh: Agree with Ed these should be added to the docstring attributes.

ew3361zh (approved):
In tandem with what will happen in #1575, I think this is a good initial step towards the stated goal of these 2 PRs. Thanks for talking through this with the dev team at our wt.

PR #1636, opened 2024-05-08
Validate Top Level Metadata

### LINKED ISSUE #1634: [IM][New Feature] Validate top-level metadata before validating metadata properties
There is currently no validation of the top-level metadata (i.e. the `files` section of the metadata) which can allow buggy inputs to make it into RuFaS simulations.

The top-level properties of the metadata should be validated to make sure there are no bugs or errors in them.

Introduce a set of mechanisms that check the metadata properties to make sure they are valid.

- Ensure required keys are present (`type`, `path`, `properties`) and that the values are validated as best as possible (e.g. the type is an acceptable type - json/csv and the path is an existing path).

- Ensure other optional keys present are from a list of acceptable valid keys (i.e. the required keys + valid optional keys like `title`, `description`).

Raise an exception, log the error to OM, and stop the input validation process if there are any problems found.

ew3361zh:
Introduces new functionality to validate top-level properties of the metadata.

### Context

Issue(s) closed by this pull request: closes #1634

- [ ] Yes
- [x] No

### What

PR 1/3 to break up PR #1584 into more manageable sized pieces.
This PR focuses on validating the top-level metadata properties. It checks:
1. Are the required keys present? Required keys for the top level metadata being `type`, `path`, and `properties`.
2. Are there valid values for required keys?
3. Are there extra unexpected/invalid keys?

### Why

The top-level properties of the metadata should be validated to make sure there are no bugs or errors in them.

### How

New function `_validate_metadata()` added to IM. It does the following: 
- Checks for missing required keys
- Checks for unexpected extra keys
- Validates values for required keys.

In each case, it will raise an error and log the error to OM.

After discussion in retro session 5/8/24 about making changes to acceptable input types easy, I added them as a module attribute in the IM:

`VALID_INPUT_TYPES: set[str] = {"json", "csv"}`

### Test plan

Test suite was updated to cover the new function.
To test, I tried making some changes to the top level metadata (`default_metadata.json` or similar) including:
- adding a new key-value pair
- replacing one of the paths with a bad path to a non-existant file.
- switching the type to something other than `json`/`csv`
- removing the properties key
- changing the properties key to `""`


RUFAS/input_manager.py
```diff
@@ -82,6 +95,7 @@ def __init__(self) -> None:
             self.__metadata: Dict[str, Any] = {}
             self.__pool: Dict[str, Any] = {}
             self.__get_data_logs_pool: Dict[str, str] = {}
+            self.__valid_input_types: Set[str] = {"json", "csv"}
```
Isgotkowitz: Would it be better to declare this a module attribute instead? I don't think there is a right or wrong answer to this question, but it feels pretty similar to `FIXABLE_INPUT_DATA_TYPES` which is declared as a module constant. And it might be good to have constants like these grouped together.

RUFAS/input_manager.py
```diff
@@ -2086,6 +2101,42 @@ def dump_get_data_logs(self, path: Path) -> None:
         file_path = os.path.join(path, file_name)
         om.dict_to_file_json(self.__get_data_logs_pool, file_path)
 
+    def _validate_metadata(self) -> None:
+        """Checks that top-level metadata has valid and required keys and values."""
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self._validate_metadata.__name__,
+        }
+        metadata_files = self.__metadata["files"]
```
```suggestion
        metadata_files = self.__metadata[ADDRESS_TO_INPUTS]
```

RUFAS/input_manager.py
```diff
@@ -2086,6 +2101,42 @@ def dump_get_data_logs(self, path: Path) -> None:
...
+                om.add_error("Metadata Validation", f"Missing required keys '{missing_keys}' in '{key}'", info_map)
+                raise ValueError
+            if invalid_keys := (data.keys() - valid_keys):
+                om.add_error("Metadata Validation", f"Invalid keys '{invalid_keys}' in '{key}'", info_map)
+                raise ValueError
+
+            if data["type"] not in self.__valid_input_types:
+                om.add_error(
+                    "Metadata Validation",
+                    f"Invalid type '{data['type']}' in '{key}'. Expected {self.__valid_input_types}",
```
```suggestion
                    f"Invalid type '{data['type']}' in '{key}'. Expected one option from {self.__valid_input_types}",
```

RUFAS/input_manager.py
```diff
@@ -2086,6 +2101,42 @@ def dump_get_data_logs(self, path: Path) -> None:
...
+            "class": self.__class__.__name__,
+            "function": self._validate_metadata.__name__,
+        }
+        metadata_files = self.__metadata["files"]
+        required_keys = {"path", "type", "properties"}
+        optional_keys = {"title", "description"}
+        valid_keys = required_keys | optional_keys
+        for key, data in metadata_files.items():
+            if missing_keys := (required_keys - data.keys()):
+                om.add_error("Metadata Validation", f"Missing required keys '{missing_keys}' in '{key}'", info_map)
```
```suggestion
                om.add_error("Metadata Validation", f"Missing required keys {list(missing_keys)} in '{key}'", info_map)
```
I think this makes the message a little cleaner, and same idea could be applied to the log message below.

Isgotkowitz (approved):
Left some suggestions for alternative ways to define constants and format logs, but otherwise LGTM! Functions as expected and all error cases were failing validation when I tried them.

RUFAS/input_manager.py
```diff
@@ -14,13 +14,18 @@
...
-
 """
 Set enumerating the input data types that the Input Manager will attempt to fix while validating input data.
 """
 FIXABLE_INPUT_DATA_TYPES: set[str] = {"string", "number", "bool"}
 
+"""
+Set enumerating the input data formats the Input Manager can accept.
+"""
+VALID_INPUT_TYPES: Set[str] = {"json", "csv"}
```
allisterakun: Just curious, is there a specific reason that you chose Set over List?

allisterakun (approved):
Good job, Niko! LGTM!

RUFAS/input_manager.py
ew3361zh: Good question - I like sets for things like this because they automatically don't allow duplicates and are more computationally efficient than lists for membership tests and combining/subtracting.

PR #1637, opened 2024-05-09
Check max metadata properties depth

### LINKED ISSUE #1333: [IM Feature] Limit depth of nested inputs

ew3361zh:
PR 2/3 breaking up PR #1584 to validate top level metadata and metadata properties. This one focuses on validating depth of metadata properties. Stacked on PR #1636.

### Context

Issue(s) closed by this pull request: closes #1333

- [ ] Wiki for command line args updated (waiting until approval).
      
- [ ] Yes
- [x] No

### What

Adds a new function _validate_properties() in InputManager which is called within the `start_data_processing()` method. It checks the depth of the most nested value within the metadata properties. It attempts to measure max depth of the metadata properties in a way that reflects the depth it would take to reach the most nested value in the related input the properties would be validating.

e.g. it would take 7 steps in the default_animal.json file to reach the deepest nested value which is for the property `cull_day_prob`:
['animal_properties', 'animal_config', 'from_literature', 'culling', 'unknown_cull', 'cull_day_prob', 0]
The metadata properties depth will be measured at 7 following this path:
['animal_properties', 'animal_config', 'from_literature', 'culling', 'unknown_cull', 'cull_day_prob', `properties`]

The way the depth check is set up currently will not count the nested properties (max, min, description, etc) for `cull_day_prob` as part of the depth but it will ensure there is no further nested dict within that properties section.

It measures the current max depth of the default.json metadata properties and adds an attribute to IM to hold that value:`metadata_depth_limit`. This value is used as the guideline against which other metadata properties are measured in this part of the properties validation process.

Adds a new gnu arg `-m` which allows overriding the default `metadata_depth_limit`.

Note: there are several further validation functions by primitive type within this `_validate_properties()` method that are all established with `pass`. They are just placeholdered here but will be the central part of the 3/3 PR breaking up #1584.

### Why

This is a tool that can help ensure new/updated/changed metadata properties files don't exceed the current depth or can help guide reworking the structure of the metadata properties if the current depth is deemed to be too high.

### How

Using an iterative DFS algorithm, this method traverses the metadata properties but stops short of adding the deepest non-nested values to better mimic the structure and depth of the input data it's validating.
Both the path and depth are logged to OM and if a metadata properties is loaded to IM that exceeds the `metadata_depth_limit`, it will log an error to OM and raise an error, stopping the validation process.

### Test plan

- Testing suite was updated.
- To test this feature, I added (and removed) nested dictionaries from the animal properties focusing on the current max depth path as noted above and ran a simulation with verbosity of `logs`. It failed when I added another level of nested data and when I entirely removed the deepest section, it successfully found the next deepest section of the properties without issue.
- To test the new gnu arg: I ran `python main.py -i -c -m 4 -v errors` to set the metadata depth to 4 knowing it would cause a failure. Then set to 10 to make sure it would work and it did.


RUFAS/input_manager.py
```diff
@@ -2086,6 +2105,109 @@ def dump_get_data_logs(self, path: Path) -> None:
...
+            if data["properties"] is None or data["properties"] == "":
+                om.add_error("Metadata Validation", f"Properties section empty or None in '{key}'", info_map)
+                raise ValueError
+
+        om.add_log("Metadata Validation", "Top level metadata is valid.", info_map)
+
+    def _validate_properties(self) -> None:
+        """Iteratively traverses the metadata properties to check the max depth and routes
+        primitive properties to be validated by type.
+        Raises
```
```suggestion
        primitive properties to be validated by type.
        
        Raises
```

RUFAS/input_manager.py
```diff
@@ -2086,6 +2105,109 @@ def dump_get_data_logs(self, path: Path) -> None:
...
+            current_obj, depth, path = stack.pop()
+
+            if depth > self.metadata_depth_limit:
+                om.add_error(
+                    "Max metadata depth exceeded.",
+                    "Metadata depth exceeds maximum allowed depth of " f"{self.metadata_depth_limit} at path {path}",
+                    info_map,
+                )
+                raise ValueError(
+                    "Metadata depth exceeds maximum allowed depth of " f"{self.metadata_depth_limit} at path {path}"
```
```suggestion
                    f"Metadata depth exceeds maximum allowed depth of {self.metadata_depth_limit} at path {path}",
                    info_map,
                )
                raise ValueError(
                    f"Metadata depth exceeds maximum allowed depth of {self.metadata_depth_limit} at path {path}"
```

RUFAS/input_manager.py
```diff
@@ -2086,6 +2105,109 @@ def dump_get_data_logs(self, path: Path) -> None:
...
+                current_max_depth = depth
+                deepest_path = path
+
+            if isinstance(current_obj, dict):
+                for key, value in current_obj.items():
+                    if isinstance(value, dict):
+                        stack.append((value, depth + 1, path + [key]))
+                        value_type = value.get("type")
+                        if value_type in type_to_validator_map:
+                            type_to_validator_map[value_type](path + [key], value)
```
Isgotkowitz, line 2209: If the value type is not in the type_to_validator_map, that should be an error

RUFAS/input_manager.py
```diff
@@ -2086,6 +2105,109 @@ def dump_get_data_logs(self, path: Path) -> None:
...
+
+        stack = [(self.__metadata["properties"], 0, [])]
+        current_max_depth = 0
+        deepest_path = []
+
+        type_to_validator_map: Dict[str, Callable] = {
+            "number": self._metadata_number_validator,
+            "array": self._metadata_array_validator,
+            "bool": self._metadata_bool_validator,
+            "string": self._metadata_string_validator,
```
Isgotkowitz, line 2183: It would be good to have a dedicated validator method for objects too - it would allow us to write in more explicit rules about what object properties should and should not contain.

RUFAS/input_manager.py
```diff
@@ -2086,6 +2105,109 @@ def dump_get_data_logs(self, path: Path) -> None:
...
+                    "Metadata depth exceeds maximum allowed depth of " f"{self.metadata_depth_limit} at path {path}"
+                )
+
+            if depth > current_max_depth:
+                current_max_depth = depth
+                deepest_path = path
+
+            if isinstance(current_obj, dict):
+                for key, value in current_obj.items():
+                    if isinstance(value, dict):
```
Isgotkowitz, line 2205: Is there a way this could be made a little bit more specific about which keys should be pushed to the stack? The only types that will need to have their elements put on the stack are objects and arrays. For objects, all keys in the properties need to be checked except "type" and "description", and only the "properties" key needs to be added for arrays.

RUFAS/input_manager.py
ew3361zh: Weird. Thank you.

RUFAS/input_manager.py
ew3361zh: How about if it's not in the type_to_validator_map and it's not `None`, then we raise an error? Because none of the "animal_properties", "config_properties", etc, have a `type` field. And we're checking for `type` in the individual validator-by-type functions so we wouldn't need to worry about those here.

(and this is assuming we move forward with a validator function for "object" types).

RUFAS/input_manager.py
ew3361zh: I will give this a shot. Here it will just be a placeholder so it will have to be fleshed out in the next stacked PR #1640.

RUFAS/input_manager.py
ew3361zh: I think I may be misunderstanding but I'm not quite sure where this applies that it would be more helpful to be looking for and to add a specific key within an object or an array to add to the stack. I feel like that makes too many assumptions about a metadata properties being set up correctly whereas this function is trying to unbiasedly validate the depth to try and catch violations and prevent incorrectly set-up metadata properties from being used.

Do you have an example of where the current set up of adding the key for a nested dict would incorrectly be added to the stack? I think that could help me understand your concern here better.

RUFAS/input_manager.py
Isgotkowitz: > How about if it's not in the type_to_validator_map and it's not None, then we raise an error?

That sounds good. Maybe a long term solution would be to have those top level properties be typed explicitly as objects?

RUFAS/input_manager.py
Isgotkowitz: > Do you have an example of where the current set up of adding the key for a nested dict would incorrectly be added to the stack?

I don't think any keys are being added incorrectly, this more like an idea for making the validation process hyper-efficient. If an array's properties are being checked, then the only key in the properties that needs to be added to the stack will be "properties" because that is where it will be possible that the depth increases. Same idea for objects, except every key in the object's properties will need to be added to the stack except for "type" and "description". So instead of iterating over the `items()` in the properties, the keys in the properties could just be automatically added based on what they are named and the type of the properties that contains them.

> I feel like that makes too many assumptions about a metadata properties being set up correctly whereas this function is trying to unbiasedly validate the depth to try and catch violations and prevent incorrectly set-up metadata properties from being used.

This idea would definitely make a lot of assumptions about metadata structure, and the only benefit would be a slight time savings, if any. So consider this suggestion just food for thought

RUFAS/input_manager.py
```diff
@@ -2130,6 +2141,83 @@ def _validate_metadata(self) -> None:
 
         om.add_log("Metadata Validation", "Top level metadata is valid.", info_map)
 
+    def _validate_properties(self) -> None:
+        """Iteratively traverses the metadata properties to check the max depth and routes
+        primitive properties to be validated by type.
```
```suggestion
        properties to be validated by type.
```

RUFAS/input_manager.py
```diff
@@ -2130,6 +2141,83 @@ def _validate_metadata(self) -> None:
...
+                for key, value in current_obj.items():
+                    if isinstance(value, dict):
+                        stack.append((value, depth + 1, path + [key]))
+                        value_type = value.get("type")
+                        if value_type in type_to_validator_map:
+                            type_to_validator_map[value_type](path + [key], value)
+                        # else:
+                        #     if value_type != "object":
+                        #         print(value_type)
+                        #         print(key)
```
```suggestion
```

Isgotkowitz (approved):
LGTM. I think there are a lot of different avenues to be explored for making the validation process more precise and/or efficient, but this is an excellent start and lays all the groundwork for making any future improvements.

I am little curious, could you explain why you went with specifying an int in order to override the default depth, instead of something like a bool that determines whether or not metadata validation happens at all?

RUFAS/input_manager.py
ew3361zh: I think that could work (or their own designation type we could check for). Would want to run it by @PooyaHekmati first since it would be a pretty noticeable change in the metadata properties.

ew3361zh:
> LGTM. I think there are a lot of different avenues to be explored for making the validation process more precise and/or efficient, but this is an excellent start and lays all the groundwork for making any future improvements.
> 
> I am little curious, could you explain why you went with specifying an int in order to override the default depth, instead of something like a bool that determines whether or not metadata validation happens at all?

My initial thinking was use cases for expanded or restricted metadata depth checks but not an either/or on doing the validation at all.

allisterakun (approved):
Looks good, Great work, thank you Niko!

PR #1550, opened 2024-04-16
[OM] [Refactor] Remove duplicate filtering function in GG and RG

allisterakun:
This PR removes the duplicate filtering functions in GG and RG by moving the filter process upstream into the OM
### Context

Issue(s) closed by this pull request: closes #1387.

- [x] The [changelog](https://docs.google.com/spreadsheets/d/1U-3igxdstHTFb6k5dVcE5-x9IL0r6q18eUiwwytKZyA/edit#gid=0) has been updated.

### What

1. Removes `GraphGenerator._prepare_plot_data()` and `ReportGenerator._prepare_report_data_with_constants()` as they have the same functionality. 
2. Extends `OutputManager.filter_variables_pool()` with the functionalities from the above functions. 
3. Creates a `_log_non_int_float_data()` function in GraphGenerator to log the non-numerical data, which was originally handled in the removed  `GraphGenerator._prepare_plot_data()`.

### Why

To remove duplicate filtering functions in GG and RG by moving the filtering process upstream into the OM.

### How

Move the filtering process into the OM.

### Test plan

All affected unit tests have been updated.
New unit tests are written for the extended `OM.filter_variables_pool()` function to account for the new scenarios that result from the addition of functionalities.
Tested with different filters (graph, report, CSV, JSON), and they all work as expected.

PR #1638, opened 2024-05-09
Put back info maps in filter variables pool method

ltp23:
### Context

- [ ] Yes
- [x] No

### What

Currently, info maps data are missing when filtering variables. One important piece of information is the units so without the info maps, we can't display the units in the csv output.

### Why

Two existing PRs are affected by this (#1579, #1606) so it is more convenient to have this one merged first. This may have been removed by a recent refactor PR #1550.

### How

- Extract info maps along with variables being filtered.

### Test plan

Make a filter file like this one and run the simulation on the main branch and this branch and check for the presence of the info maps.

```
{
  "filters": [
    ".*"
  ],
  "name": "All Variables",
}
```
Also, turn on and off the exclude-info-maps flag also in the tasks json file.

```
"tasks": [
    {
      "task_type": "SIMULATION_SINGLE_RUN",
      "output_prefix": "Task 1",
      "log_verbosity": "errors",
      "exclude_info_maps": false
    }
  ]
```


Isgotkowitz (commented):
I am still fuzzy on what the exact behavior is supposed to be, but when I write three filters (report, JSON, and CSV) for the same set of variables I get the info maps in some but not others. Should it be possible to get the info maps in CSV and report outputs?

allisterakun (approved):
Straight-forward changes, LTGM. Good job, Loi!

ltp23:
> I am still fuzzy on what the exact behavior is supposed to be, but when I write three filters (report, JSON, and CSV) for the same set of variables I get the info maps in some but not others. Should it be possible to get the info maps in CSV and report outputs?

It is probably easier when you compare the results from the main branch and this branch. On the main branch, if I do json filtering, I can't see the info maps and with csv filtering, I can't see the units because the units are part of the info maps.

I don't know if there is a request for making info maps show up in csv and report csv files.

KFosterReed:
@PooyaHekmati can you confirm if this branch and PR can be deleted?

KFosterReed:
@allisterakun can you confirm if this branch and PR can be deleted?

PR #1640, opened 2024-05-09
Validate metadata properties

### LINKED ISSUE #1332: [IM][New Feature] Validate metadata properties before validating input data

ew3361zh:
PR 3/3 breaking up PR #1584 to validate top level metadata and metadata properties. This one focuses on validating metadata properties and properties keys.

### Context

Issue(s) closed by this pull request: closes #1332

- [ ] Yes
- [x] No

### What

This PR adds functionality to validate the metadata properties of the types (array, string, bool, object, and number) as well as those properties' keys. This occurs after validating the top level metadata and loading the properties.

Additionally, as the code was being tested and run, a handful of violations were found within `default.json` and `tasks_properties.json` and those issues have been corrected and, where applicable, cleared with the appropriate SME.

### Why

This functionality ensures there are no unexpected, misspelled, or invalid keys within the properties section for a type property. For example, with a metadata number type property, it should have a `type` key and the full list of valid keys it should have is: `{"type", "description", "minimum", "maximum", "default"}`. If there is a default, that default should be greater than the minimum (if minimum exists) and less than the maximum (if it exists). There are a number of other validations that should occur as well and all are attempted to be covered in this PR to ensure the metadata properties are themselves valid and we're not fixing input data with bad default data.

### How

Within the new `_validate_properties()` method, there is a validate-by-type function map similar to what's used with the input data but specific to the structure of the metadata properties. When a properties section for one of these types is reached, it will be dispatched by type to have both its keys and values validated.

### Test plan

The strategy I took was to start changing and adding keys in the metadata properties to try and get errors to trigger.

e.g. take this entry in field_properties:

```
"absolute_latitude": {
      "type": "number",
      "description": "The absolute latitude of the center of this field.\nUnits: degrees.",
      "minimum": 0.0,
      "maximum": 90.0,
      "default": 43.5
    },
```

Change `minimum` to `manimus` -> will raise error.
Change the value for `maximum` to < 43.5 -> will raise error.
Add a key-value pair: `"flying_monkeys": 234` -> will raise error.

You can try similar changes with each type - bool, string, array, number. Object types currently only have validation set up for including the `type` key and at least one other key that's not `type` or `description`.

For testing object types successfully having at least one extra key, you can remove all keys from an object except `type` and `description`.

Additionally, check for nullability:
1. for `initial_labile_inorganic_phosphorus_concentration`, the default is currently `null`. I added `"nullable": "true"` to the properties for this input to support this default being `null`.
2. Try setting `nullable` to `false` or removing the nullable property altogether. It will fail in both instances because the default is `null`.


RUFAS/input_manager.py
```diff
@@ -67,6 +67,21 @@ def get_modifiable_at_runtime(cls) -> List["Modifiability"]:
...
+    __pool : Dict[str, Any]
+        The input data pool.
+    __get_data_logs_pool : Dict[str, str]
+        The logs for each instance of data being fetched from Input Manager.
+    __valid_input_types : Set[str]
+        The currently valid input types we expect in the top-level metadata.
+    elements_counter : ElementsCounter()
+        An instance of the class to keep track of the number of elements in each state during validation.
+    metadata_depth_limit : int
+        The maximum depth allowed for a metadata properties structure.
```
allisterakun: As we discussed earlier at the retro (or maybe planning) meeting, this should be removed

RUFAS/input_manager.py
```diff
@@ -82,7 +97,9 @@ def __init__(self) -> None:
             self.__metadata: Dict[str, Any] = {}
             self.__pool: Dict[str, Any] = {}
             self.__get_data_logs_pool: Dict[str, str] = {}
+            self.__valid_input_types: Set[str] = {"json", "csv"}
```
allisterakun: Just curious, is there a specific reason that you chose Set instead of List (or other data structure)?

allisterakun (approved):
Good job, Niko! Just two minor comments.

RUFAS/input_manager.py
```diff
@@ -331,7 +331,7 @@ def _populate_pool(self, eager_termination: bool) -> bool:
 
         """
 
-        data_type_to_loader_map: Dict[str, Callable[[str], Dict[str, Any]]] = {
+        data_type_to_loader_map: Dict[str, Callable[[Path], Dict[str, Any]]] = {
```
ew3361zh, line 334: Fixes a mypy error.

input/metadata/properties/default.json
```diff
@@ -4884,7 +4884,6 @@
     "field_size": {
       "type": "number",
       "description": "Size of the field.",
-      "units": "ha",
```
ew3361zh, line 4887: Removed because it's not a valid property key for number type.

input/metadata/properties/default.json
```diff
@@ -3488,7 +3488,7 @@
         "type": "number",
         "minimum": 0.87,
         "maximum": 1.04,
-        "default": 0
+        "default": 0.87
```
ew3361zh, line 3463: Confirmed change with @JoeWaddell - default was previously below the minimum value as defined in the properties here.

RUFAS/input_manager.py
```diff
@@ -2221,25 +2221,212 @@ def _validate_properties(self) -> None:
...
-        """Validator function for array type properties in metadata."""
-        pass
+    def _metadata_number_validator(self, key_path: list[str], value: dict[str, Any]) -> None:
+        """Validates number type properties in metadata."""
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self._metadata_number_validator.__name__,
+        }
+        required_number_property_keys = {"type"}
+        valid_number_property_keys = {"type", "description", "minimum", "maximum", "default", "nullable"}
```
```suggestion
        required_number_property_keys = {"type"}
        optional_number_property_keys = {"description", "minimum", "maximum", "default", "nullable"}
        valid_number_property_keys = union(required_number_property_keys, optional_number_property_keys)
```
I think this syntax would work, but I didn't check it - also there may be a cleaner way to do this. The idea though is that it would be beneficial to write out the required_number_property_keys once.

Edit: I didn't read the full code before I wrote this comment. Maybe it would be better to add the `union` step in `_validate_metadata_properties_keys` instead of writing this step into every individual validator function.

RUFAS/input_manager.py
```diff
@@ -2221,25 +2221,212 @@ def _validate_properties(self) -> None:
...
+            if maximum is not None and default > maximum:
+                om.add_error(
+                    "Invalid metadata default.",
+                    f"Invalid 'default' for '{key_path}': 'default' {default} is " f"greater than 'maximum' {maximum}",
+                    info_map,
+                )
+                raise ValueError
+        if maximum is not None and minimum is not None and maximum < minimum:
+            om.add_error(
+                "Invalid metadata array length range.",
```
```suggestion
                "Invalid range of acceptable numbers.",
```

RUFAS/input_manager.py
```diff
@@ -2221,25 +2221,212 @@ def _validate_properties(self) -> None:
...
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self._metadata_number_validator.__name__,
+        }
+        required_number_property_keys = {"type"}
+        valid_number_property_keys = {"type", "description", "minimum", "maximum", "default", "nullable"}
+        self._validate_metadata_properties_keys(
+            required_number_property_keys, valid_number_property_keys, value, key_path
+        )
+        default = value.get("default")
```
Isgotkowitz: I think the way that the default is collected needs to be changed. If the default key is present and its value is null, that is a different case than the default key not being present at all.

RUFAS/input_manager.py
```diff
@@ -2221,25 +2221,212 @@ def _validate_properties(self) -> None:
...
+                    info_map,
+                )
+                raise ValueError
+            if maximum is not None and not isinstance(maximum, (int, float)):
+                om.add_error(
+                    "Invalid metadata default maximum.",
+                    f"Invalid 'maximum' for '{key_path}': " f"Expected a number but got {type(maximum)}",
+                    info_map,
+                )
+                raise ValueError
```
Isgotkowitz: I think these checks of the max and min need to happen outside of the if block they are wrapped in. Currently they are only checked if there is a default present, and but they should be checked to make sure they are valid numbers regardless of whether or not there is a default present.

RUFAS/input_manager.py
```diff
@@ -2221,25 +2221,212 @@ def _validate_properties(self) -> None:
...
-        """Validator function for string type properties in metadata."""
-        pass
+    def _metadata_string_validator(self, key_path: list[str], value: dict[str, Any]) -> None:
+        """Validates string type properties in metadata."""
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self._metadata_string_validator.__name__,
+        }
+        required_str_property_keys = {"type"}
+        valid_str_property_keys = {"type", "description", "pattern", "default", "nullable"}
```
Isgotkowitz: Same as above

RUFAS/input_manager.py
```diff
@@ -2221,25 +2221,212 @@ def _validate_properties(self) -> None:
...
+        self._validate_metadata_properties_keys(required_str_property_keys, valid_str_property_keys, value, key_path)
+        default = value.get("default")
```
Isgotkowitz: Also same as above for getting the default.

RUFAS/input_manager.py
```diff
@@ -2221,25 +2221,212 @@ def _validate_properties(self) -> None:
...
+        pattern = value.get("pattern")
```
Isgotkowitz: Similar to getting the default, the getting of the pattern should be changed to make sure that the case of the pattern key being present and set to null is treated differently from the pattern key not being present. Also, if the pattern key is present, it should be checked that it is a valid Regex pattern - `re.compile` does this in a pretty exhaustive way I think.

RUFAS/input_manager.py
```diff
@@ -2221,25 +2221,212 @@ def _validate_properties(self) -> None:
...
-        """Validator function for array type properties in metadata."""
-        pass
+    def _metadata_array_validator(self, key_path: list[str], value: dict[str, Any]) -> None:
+        """Validates array type properties in metadata."""
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self._metadata_array_validator.__name__,
+        }
+        required_array_property_keys = {"type", "properties"}
+        valid_array_property_keys = {"type", "properties", "description", "minimum_length", "maximum_length", "default"}
```
Isgotkowitz: Currently, the Input Manager will only attempt to fix data (i.e. use the default) if it is in `FIXABLE_INPUT_DATA_TYPES`. Which means that even if the properties specify a default for an array or object and an array or object is passed that is invalid, the default will not be used (I ran into this very issue in #1687, and the solution may just be making the object and array types fixable). As long as arrays and objects are not fixable, I think it should not be allowed for the properties to specify defaults for them because it is misleading for people who may read or write metadata properties without knowing the exact logic of the Input Manager.

RUFAS/input_manager.py
```diff
@@ -2221,25 +2221,212 @@ def _validate_properties(self) -> None:
...
+                    f"Invalid 'minimum_length' for '{key_path}': " f"Expected a number but got {type(minimum_length)}",
+                    info_map,
+                )
+                raise ValueError
+            if maximum_length is not None and not isinstance(maximum_length, (int, float)):
+                om.add_error(
+                    "Invalid metadata default array maximum length.",
+                    f"Invalid 'maximum_length' for '{key_path}': " f"Expected a number but got {type(maximum_length)}",
+                    info_map,
+                )
```
Isgotkowitz: Similar to the above comment on max's and min's in the number validator method, these max and min lengths should be check regardless of whether or not a default is present.

RUFAS/input_manager.py
```diff
@@ -2221,25 +2221,212 @@ def _validate_properties(self) -> None:
...
+        }
+        if missing_required_keys := required_properties_keys - properties.keys():
+            om.add_error(
+                "Metadata Validation",
+                f"Missing required keys {sorted(missing_required_keys)} for {path}. Required"
+                f" keys are {sorted(required_properties_keys)}.",
+                info_map,
+            )
+            raise ValueError
+        property_type = properties.get("type", "Unknown type")
```
Isgotkowitz, line 2200:
```suggestion
        property_type = properties.get("type")
```
If the type is missing or invalid, won't it get caught in `_validate_properties` and not make it to this line?

RUFAS/input_manager.py
```diff
@@ -2221,25 +2221,212 @@ def _validate_properties(self) -> None:
...
+        if property_type == "object":
+            return
```
Isgotkowitz: If the property is for an object, can/should it be assumed that there should be at least one key that is _not_ listed in the required or valid properties keys?

Isgotkowitz (changes requested):
I think this is well on the way towards being ready, but there are some cases not covered which should be addressed before this PR is merged:
- With frequent use of the `.get()` function to retrieve a value associated with a key, there is no distinguishing between when the properties do not include the key vs. when they do include the key and the value is set to null/None.
- The validity of the `nullable` key is not checked. If the key is present, then its value must be a boolean. Also, the nullability of a value should be considered when checking the default.

RUFAS/input_manager.py
ew3361zh: I like adding it to the `_validate_metadata_properties_keys()` function to minimize duplicate code. And it looks like sets have their own `.union()` method for joining sets together (and automatically removing duplicates).

RUFAS/input_manager.py
ew3361zh: You bring up an interesting point.
I made this diagram to try to organize the order of operations. Let me know what you think.
Screenshot 2024-05-30 at 11 25 55 AM

matthew7838: This graph will not check the default if the leftmost path is followed, which contradicts with 

> **If there is a default, that default should be greater than the minimum (if minimum exists) and less than the maximum (if it exists).**

 in the description.  However, I think this graph still makes logical sense and I'm not sure which is preferred.

RUFAS/input_manager.py
ew3361zh: Good call.

matthew7838 (changes requested):
Looks like things are on track. There are a few things. 

1. It seems like even though the value is valid, it will still not pass the validator if the input has an invalid default. Is this an intended behavior?
2. The docstrings are missing parameters.
3. As Ed mentioned, the case of key's existence should be differentiated. I would suggest using something like `in` to check that before calling the `get`.

RUFAS/input_manager.py
ew3361zh: Thanks Ed. I'm going to check with SMEs about the instances of there being defaults for arrays (and objects if any of those exist).

RUFAS/input_manager.py
ew3361zh: Yea sorry this is confusing because this is from my map to overview the IM validation process so it's not at the level of detail (yet) that will show that check happening. Thank you for pointing that out though - I will definitely have that clear in the final version (maybe as a sub-map of this main map so it doesn't get too crowded).

ew3361zh:
> Looks like things are on track. There are a few things.
> 
> 1. It seems like even though the value is valid, it will still not pass the validator if the input has an invalid default. Is this an intended behavior?

Yes, this behavior is intended to ensure the metadata itself is valid which includes having a default that is itself fulfilling the guidelines set by the metadata.

> 2. The docstrings are missing parameters.

For lower level functions (sub functions deep within a particular class), we're not adding full docstrings because it's too cumbersome. Ed and I had a discussion about how we're judging this and it's not exact but I think if a function is a couple levels deep within a class, that's my current guideline for whether it needs full docstrings or just an overview docstring.

matthew7838 (approved):
Thanks for the explanation. I will approve the PR, but I agree that Ed raised some important points that need to be resolved before we proceed with the merge.

RUFAS/input_manager.py
ew3361zh: It would right now with the way things are set up but I thought adding this check would make this function less dependent on that initial check in `_validate_properties()` in case there is later on a different way this function gets called.

RUFAS/input_manager.py
ew3361zh: Added a check for this so there should be at least one additional key that's not in the `valid_properties_keys` set if it's an object type property.

input/metadata/properties/default.json
```diff
@@ -1574,7 +1575,6 @@
           "type": "array",
           "description": "Julian day(s) of year to harvest",
           "minimum_length": 0,
-          "default": [],
```
ew3361zh, line 1577: Checked with @Isgotkowitz on removing these default arrays.

input/metadata/properties/default.json
```diff
@@ -4818,13 +4791,15 @@
...
+          "default": null,
+          "nullable": true
         },
         "initial_soil_ammonium_concentration": {
           "type": "number",
           "description": "Concentration of ammonium in the layer at the beginning of the simulation.\nUnits: milligrams / kilogram soil.",
           "minimum": 0.0,
-          "default": null
+          "default": null,
+          "nullable": true
```
ew3361zh, line 4802: Added nullability for defaults that were `null`.

input/metadata/properties/tasks_properties.json
```diff
@@ -154,7 +154,7 @@
         "SA_input_variables": {
           "type": "array",
           "description": "The input variables to perform Sensitivity Analysis on",
-          "default": [{}],
```
ew3361zh, line 157: Confirmed with @JoeWaddell that this default was not needed.

RUFAS/input_manager.py
```diff
@@ -2221,25 +2225,252 @@ def _validate_properties(self) -> None:
...
+        except re.error:
+            om.add_error(
+                "Invalid metadata string properties pattern.",
+                f"Invalid 'pattern' for '{key_path}': 'pattern' value '{pattern}' is not "
+                "a valid regex pattern.",
+                info_map,
+            )
+            raise ValueError(f"Invalid 'pattern' for '{key_path}': 'pattern' value '{pattern}' is not "
+                             "a valid regex pattern.")
+        if default != "" and default is not None and not has_no_default:
```
Isgotkowitz, line 2084:
```suggestion
        if default != "" and default is not None and not has_no_default:
```
If the default is an empty string, shouldn't it still be checked against pattern to make sure that it is a valid default?

Isgotkowitz (approved):
Only spotted two things: when a string default is "" (left comment on this below), and checking "nullable" - I think it is good to do `.get("nullable", False)`, but the value should still be checked to make sure that it is a bool.

Other than that though, LGTM. Nice job Niko

input/metadata/properties/default.json
Isgotkowitz: :regular salute: (GitHub doesn't have an emoji for this)

RUFAS/input_manager.py
ew3361zh: Unfortunately an empty string doesn't match a regex pattern that specifically requires at least one character to be present. This mostly came up because of the default for `output_pool_path` in `tasks_properties.json` is an empty string. It needs to have a pattern because if a path is given, it needs to be a valid path. So I'm not sure the best way around it.

```
"output_pool_path": {
          "type": "string",
          "pattern": "^(?:[a-zA-Z]:[\\/])?(?:[a-zA-Z0-9._-]+[\\/])*(?:[a-zA-Z0-9._-]+\\.json)$",
          "description": "The path to a JSON files to load the Output Manager's pool from for post processing",
          "default": ""
        },
```

Any suggestions?

RUFAS/input_manager.py
Isgotkowitz: I think it looks good, the only thing that needs to change is the validation of the default - that should be in the map describing how metadata properties are validated. By the time it gets to validating input data, all defaults should have been validated.

PR #1593, opened 2024-04-29
Phosphorus requirements fix

JoeWaddell:
### Context

Issue(s) closed by this pull request: closes #1588

- [x] The [changelog](https://docs.google.com/spreadsheets/d/1U-3igxdstHTFb6k5dVcE5-x9IL0r6q18eUiwwytKZyA/edit#gid=0) has been updated.

### What

Implemented a new value in the animal input JSON as suggested in Issue #1588

Corrected the `P_preg` section of phosophorus requirement calc to reflect the structure reported in the NASEM R code.

Corrected unit conversion per changes in #1583

Updated typing and docstrings, and unit test.

### Why

See tagged Issue: a new buffer value for Phosphorus intake was necessary to achieve realistic outputs. [Another followup PR ](https://github.com/RuminantFarmSystems/MASM/pull/1641) will make additional changes, if needed.

### Test plan

New buffer value methodology thoroughly examined by @KFosterReed , @jadamchick , and @krb78 using primarily (but not limited to) Pilot Test farm C2F7 using different buffer values.

Following report filter was used in analysis: 

      {
          "name": "Pen 3 daily DMI per cow, kg/cow",
          "filters": ["AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_3_LAC_COW"],
          "variables": ["dm"],
          "vertical_aggregation": "average",
          "slice_start": -365
      }, 
      {
          "name": "Pen 3 daily phosphorus intake per cow, kg/cow",
          "filters": ["AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_3_LAC_COW"],
          "variables": ["phosphorus"],
          "vertical_aggregation": "average",
          "slice_start": -365
      },
      {
            "name": "Pen 3 dietary phosphorus, %",
            "cross_references":[
                "Pen 3 daily phosphorus intake per cow, kg/cow_phosphorus_ver_agg",
                "Pen 3 daily DMI per cow, kg/cow_dm_ver_agg"
            ],
            "constants": {
                "By 0.01": 0.01
            },
            "horizontal_aggregation": "division",
            "horizontal_order": [
                "Pen 3 daily phosphorus intake per cow, kg/cow_phosphorus_ver_agg",
                "Pen 3 daily DMI per cow, kg/cow_dm_ver_agg", 
                "By 0.01"
            ]
      }, 
      {
          "name": "Pen 3 daily manure phosphorus, g/cow/d",
[...]

PR #1641, opened 2024-05-09
Phosphorus_process

JoeWaddell:
### Context

This is related to #1593, which indicated we may need to revise some of the process based phosphorus equations.

Notably `phosphorus_rqmts` in each animal file, and `calc_base_manure` in animal_base.py

- [ ] Yes
- [x] No

### What

Uses phosphorus requirement as calculated in an animal's process-based phosphorus calculations. Alongside the calculation made via NRC or NASEM methods, the optimizer then uses the maximum of these, to ensure we aren't making an underestimate.

### Why

Noted some phosphorus problems in #1593, this came from there.

### How

Simple addition of attribute to an animal.

### Test plan

Updated unit tests, and SMEs checked that the process based requirements are reasonable values, and may enhance our diet formulation.


krb78 (approved):
LTGM Joe! :) I ran a few pilot farms and the %P in the diet for lactating cows was all above >0.4% which I think fits with lactating cow diets we actually see today :) works great!

allisterakun (approved):
Simple and straight-forward changes. Looks great! Thank you, Joe!

Issue #1533, opened 2024-04-11
[Time] Update All Uses of `Time` to new `Time` class


ew3361zh:
## Urgency Level

- [ ] Urgent: a critical feature or fix is needed
- [ ] High: the issue significantly impacts the user experience
- [ ] Medium: the issue affects functionality but has a workaround
- [ ] Low: the issue doesn't have a significant immediate impact
- [x] Wishful: the issue should not be prioritized over any others
- [ ] Unsure: the urgency level has not been or cannot be assessed at this time 

## Status

This issue is blocked by #1522 but also the PR solving that issue shouldn't be merged until this issue is resolved as well because we need all modules/files/etc to be using the same version of Time throughout the model.

## Intention

There are multiple places in the code where time is kept/tracked/utilized in various forms. This is problematic as it makes the code non-standard and bug prone. All time-related uses should use the refactored Time object accomplished with the completion of issue #1522.

## Request

After the `Time` class has been refactored and the regulation of Time through the SimulationEngine is more standardized, look through the full model for places where calendar day, calendar year, simulation day, simulation year (and any other variants) and update them to be referencing the `Time` class. Ensure testing is updated as well.

PR #1643, opened 2024-05-09
Time refactor

### LINKED ISSUE #1522: [Time] Refactor Time class
Adapt the current `Time` class to be passed around the model and regulated/advanced via SimulationEngine. `Time` should be the single-source of truth for tracking and measuring time in RuFaS.

## Urgency Level

- [x] Wishful: the issue should not be prioritized over any others

## Status

This issue is really 2 issues that should be done together because they depend on one another:
Step 1:
- [ ] Rework SimulationEngine to be the central manager of the Time class between other modules. This likely needs to be folded into what's being done with #1411.
- [ ] Issue #1533 - update all references to time to use the newly refactored `Time` class (advise the PR addressing this issue to be branched off the PR addressing issue #1522 (this issue) after the PR solving 1522 is approved).

## Intention

All cases where time needs to be managed or tracked should utilize the newly refactored `Time` class.  

## Request

1. The `Time` class should be refactored to be clearer and references to it throughout the model should be more uniform (e.g. not passing parts of the Time class but the whole class so references are clearer as to where that data is coming from and how it's regulated).
2. Currently Time is only advanced in `SimulationEngine` in its `_advance_time()` method and `_run_post_annual_routines()` method. We want to be sure `Time` is not advanced anywhere else.
3. Update all current references to `Time` to ensure compatability and proper function with the newly refactored `Time` class.
4. Testing should be updated.

#### Additional Information

If refactoring Time AND updating all references to time (those currently using the current `Time` class and those using alternate avenues of simulation/calendar time) seems like too much for one single issue, we can break this up into two issues that could be done in tandem. 

In the first issue the refactor happens. 
The second issue is then branched off that refactor before the PR to address the first issue is merged into main. 
This will help ensure it's properly introduced and tested and accounted for.

Seems good to have 1 person do everything outlined in this issue within one Sprint if possible.

allisterakun:
Restructured the Time class data:
1. Got rid of the two-dimensional array: years: List[List[int]]
    1. The `years` variable is turned into a property for now (along with `start_year_int`, `start_day`, `end_year_int`, `end_day`), as the Weather class is still heavily dependent on this structure. These properties will be removed following the refactoring of the Weather class as they are no longer needed.
2. Use a single DateTime object to keep track of the current date instead.
3. Created the following properties for ease of use:
    1. current_julian_day: the current Julian day of the current year in integer.
    2. current_month: the current month in integer.
    3. current_simulation_year: the current simulation year in integer.
    4. current_calendar_year: the current calendar year in integer.
    5. simulation_day: the current simulation day in integer.
    6. year_start_day: the first Julian day of the current year in integer.
    7. year_end_day: the last Julian day of the current year in integer.
4. Updated all other functions in the Time class accordingly.
5. Updated all references to the Time class.

Other changes:
1. Moved `_record_animal_events()`, `_record_heiferIIs_conception_rate()`, and `_record_cows_conception_rate()` from `AnimalManager` into `AnimalModuleReporter` as they serve the purpose of reporting the herd data at the end of the simulation.
### Context

Issue(s) closed by this pull request: closes #1522 & #1533.

- [x] Yes
- [ ] No

### What

See description.

### Why

See description.

### How

See description.

### Test plan

1. Update all affected tests.
2. Run the simulation to check if all the days within the simulation range are reached. This also fixes the bug where there was a day missing for the previous Time class.


RUFAS/time.py
```diff
@@ -15,38 +16,28 @@ def __init__(self):
...
+        self.config_data: Dict[str, str | int | bool] = im.get_data("config")
+        self.start_date: datetime = datetime.datetime.strptime(self.config_data["start_date"], "%Y:%j")
+        self.end_date: datetime = datetime.datetime.strptime(self.config_data["end_date"], "%Y:%j")
 
-        self._init_time_config()
-        self.calendar_year: int = self.start_year_int
-        self.year: int = 1  # current year
-        self.simulation_length = self._calc_sim_length()
+        self.leap_year_length: int = GeneralConstants.LEAP_YEAR_LENGTH
+        self.year_length: int = GeneralConstants.YEAR_LENGTH
```
Isgotkowitz: Instead of keeping these as attributes, it might be better to just use the constants instead.

RUFAS/time.py
```diff
@@ -15,38 +16,28 @@ def __init__(self):
...
+        self.current_date: datetime = self.start_date
+        self.simulation_length_days: int = (self.end_date - self.start_date).days
+        self.simulation_length_years: int = self.end_date.year - self.start_date.year + 1
+        self.simulation_day: int = 0
```
Isgotkowitz: Instead of maintaining simulation day as a separate attribute, what about making it a property and taking the difference of the current date and end date?

RUFAS/time.py
```diff
@@ -61,42 +52,54 @@ def _init_time_config(self) -> None:
...
-                if Utility.is_leap_year(self.start_year_int):
-                    sim_length += self.leap_year_length - self.start_day
-                else:
-                    sim_length += self.year_length - self.start_day
-            else:
-                sim_length += len(self.years[i])
+    @property
+    def start_year_int(self) -> int:
+        start_full_date: list[str] = self.config_data["start_date"].split(":")
+        return int(start_full_date[0])
```
```suggestion
        return self.start_date.year
```

RUFAS/time.py
```diff
@@ -15,38 +16,28 @@ def __init__(self):
         """
         This object is responsible for creating and tracking time in the simulation.
         """
+        self.config_data: Dict[str, str | int | bool] = im.get_data("config")
+        self.start_date: datetime = datetime.datetime.strptime(self.config_data["start_date"], "%Y:%j")
+        self.end_date: datetime = datetime.datetime.strptime(self.config_data["end_date"], "%Y:%j")
```
Isgotkowitz, line 21: I don't think the config data needs to be saved as an attribute, because all the time-relevant information is recorded in the start and end dates

RUFAS/time.py
```diff
@@ -61,42 +52,54 @@ def _init_time_config(self) -> None:
...
+    @property
+    def current_julian_day(self) -> int:
+        return int(self.current_date.strftime("%j"))
+
+    @property
+    def current_month(self) -> int:
+        return self.current_date.month
+
+    @property
+    def current_simulation_year(self) -> int:
```
Isgotkowitz: For all these `@property` methods that aren't about the current time, could you please add docstrings? There are some that get kinda confusing because there are multiple start and end days and years

RUFAS/time.py
```diff
@@ -113,24 +116,16 @@ def end_year(self) -> bool:
         As soon as the current day is greater than the number of days in the year,
         the day is reset to 1 and the year is incremented by 1.
```
Isgotkowitz: I think this docstring is out of date.

tests/test_time.py
```diff
@@ -405,63 +26,138 @@ def test_time_initialization(mock_config: Dict[str, Any], mocker: MockerFixture)
     time = Time()
 
     mock_im_get_data.assert_called_once_with("config")
-    assert time.start_year_int == mock_config["start_year_int"] and time.calendar_year == mock_config["calendar_year"]
-    assert time.leap_year_length == mock_config["leap_year_length"]
-    assert time.year_length == mock_config["year_length"]
-    assert time.day == 2
+    assert time.config_data == mock_config
+    assert time.start_date == datetime.datetime.strptime("1999:2", "%Y:%j")
+    assert time.end_date == datetime.datetime.strptime("2000:1", "%Y:%j")
```
Isgotkowitz:
```suggestion
    assert time.start_date == datetime.datetime(year=1999, month=1, day=2)
    assert time.end_date == datetime.datetime(year=2000, month=1, day=1)

```

tests/test_time.py
```diff
@@ -405,63 +26,138 @@ def test_time_initialization(mock_config: Dict[str, Any], mocker: MockerFixture)
...
-    """Tests that Time instances correctly determine if the simulation has ended."""
-    # case when year is less than the length of years list
-    time = mock_time
-    for _ in range(366):
-        assert not time.end_simulation()
+    year_end_days = [365] * 266 + [366] * 366 + [365] * 1095 + [10] * 10
+    for end_day in year_end_days:
+        assert time.year_end_day == end_day
         time.advance()
-    assert time.end_simulation()
```
Isgotkowitz: A lot of these tests contain a lot of magic numbers and loops that are not intuitive to understand. If possible, I think it would be better to have them parametrized so that the test inputs are just the underlying date, and the expected result.

Isgotkowitz (approved):
This is a big leap forward in Time, nice job Allister. I tested this out by comparing the time output on this branch to the one on the main branch, and this fixes some very obvious bugs which is great to see.

I made suggestions on a bunch of things, all are take it or leave it.

There are also some extra changed files that are not related to the time refactor, but I'm sure those will go away when this branch is updated.

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -956,41 +957,206 @@ def report_daily_reports(cls, animal_manager, available_feeds: Dict[str, Dict[st
...
-        total_days : int
-            The total number of days in the simulation
+        life_cycle_manager : LifeCycleManager
+            Instance of LifeCycleManager class.
+        time : Time
+            The Time object with the current time information.
+        heiferIIs : List[HeiferII]
+            The list of HeiferIIs.
+        cows : List[Cow]
+            The list of Cows
```
ew3361zh, line 975: Can you add the reason for this refactor/rearrangement of code in animal_module_reporter and animal_manager to the PR description?

RUFAS/time.py
ew3361zh: Yea I'm also curious if the `config_data` needs to be saved for some reason.

RUFAS/time.py
```diff
@@ -15,41 +16,35 @@ def __init__(self):
...
-                self.day = self.years[0][i]
-                break
-
-        self.simulation_day = 0
-
-    def _init_time_config(self) -> None:
+    def advance(self) -> None:
         """
-        Initializes the time configuration for the instance by parsing the config data from InputManager pool.
+        Advances the time in the simulation by 1 day and automatically detects end of years.
```
ew3361zh: How does this method automatically detect the end of the year?

```suggestion
        Advances the time in the simulation by 1 day and automatically detects ends of years.
```

ew3361zh (approved):
I think this looks good, Allister. I have a few comments asking for further explainations in comments. 

I think the PR needs more description filled out for the what, why, how especially since it's considered for a major version upgrade.

PR #1647, opened 2024-05-12
Update Urine N Equations

tomhuhh:
This PR updated equations to calculate urine N for lactating, dry, and heifers. Specifically, instead of directly predicting urine N, we used the difference between manure N and fecal N. This is because we have more confidence in the predictions for fecal N than urine N.  

### Context

This PR is suggested by @KFosterReed. 

- [ ] Yes
- [x] No

### What

Previously, for example, the lactating cow urine N is predicted following Reed 2015: 
```
urine_nitrogen = (
        12.0
        + 0.333
        * (dry_matter_intake * GeneralConstants.KG_TO_GRAMS)
        * (CP_concentration * GeneralConstants.PROTEIN_TO_NITROGEN)
        / 100
```

In this update, the prediction is updated to the equations below following Reed 2015: 
```
fecal_nitrogen = (
        - 18.5 
        + 10.1
        * dry_matter_intake
    ) * GeneralConstants.GRAMS_TO_KG
urine_nitrogen = manure_nitrogen - fecal_nitrogen
```

### Why

This update is due to the fact that we have more confidence with the predictions of fecal N. 

### How

Please refer to the example given above. 

### Test plan

You can test this PR by including urine N in your output filter, comparing the results before and after the changes.


RUFAS/routines/animal/manure/dry_cow_manure_excretion.py
```diff
@@ -127,17 +127,17 @@ def manure_calculations(
...
-    urine_nitrogen = (
-        14.3
-        + 0.510
+    # Nitrogen excretion in feces, kg [A.3B.B.2]
+    fecal_nitrogen = (
+        0.345
+        + 0.317
         * (dry_matter_intake * GeneralConstants.KG_TO_GRAMS)
         * (CP_concentration * GeneralConstants.PROTEIN_TO_NITROGEN)
         / 100
```
Isgotkowitz: A couple of questions about this `/ 100`:
- Can 100 be replaced with a constant? It would help make this equation more readable.
- Eqn. A.3B.B.2 divides by 1000, not 100. Which one of these is correct?

RUFAS/routines/animal/manure/lactating_cow_manure_excretion.py
```diff
@@ -175,17 +175,11 @@ def manure_calculations(
...
-    # Urine nitrogen, kg [A.3E.B.2]
-    urine_nitrogen = (
-        12.0
-        + 0.333
-        * (dry_matter_intake * GeneralConstants.KG_TO_GRAMS)
-        * (CP_concentration * GeneralConstants.PROTEIN_TO_NITROGEN)
-        / 100
-    ) * GeneralConstants.GRAMS_TO_KG
+    # Fecal nitrogen, kg [A.3B.B.2]
+    fecal_nitrogen = (-18.5 + 10.1 * dry_matter_intake) * GeneralConstants.GRAMS_TO_KG
```
Isgotkowitz, line 180: Does this need bounds placed on it? To protect against the case where dry matter intake is so low that you get a negative amount of fecal nitrogen.

Isgotkowitz (approved):
Left one comment on protecting against an edge case that may or may not be worth it. Also found a discrepancy between the scientific documentation and the code, it is very science-y so I trust you and @KFosterReed will figure out what needs to happen.

Otherwise looks good, nice job Haowen.

RUFAS/routines/animal/manure/dry_cow_manure_excretion.py
tomhuhh: Great observation, Ed! I believe the number '100' is used is to convert `CP_concentration` which is a percent to a numerical number. So for example, if dietary CP concentration is 16%, the variable `CP_concentration` is 16 instead of 0.16. 

The `1000` you noticed is instead the constant`GeneralConstants.KG_TO_GRAMS`.  

That said, I can add `100` as a constant in the constant file. Thanks!

Isgotkowitz: Phew awesome, I got a scared when I was looking at those side by side. There is the "PERCENTAGE_TO_FRACTION" constant in general constants file which is for situations just like this.

RUFAS/routines/animal/manure/lactating_cow_manure_excretion.py
tomhuhh: This is a good point, Ed. In the original paper, the DMI range for lactating cows was from 3.94 to 29.4 kg/d. In this equation, the DMI will need to go below 1.83 kg/d for the fecal nitrogen to be negative, which is very rare. I'll discuss this further with Kristan. Thank you!

RUFAS/routines/animal/manure/dry_cow_manure_excretion.py
tomhuhh: Perfect!

elle-andreen (approved):
Updated N calculations appear to be functioning correctly. As expected, I am seeing a slight increase in ammonia emissions with these changes (see graphs). Ammonia N emission responsiveness to diet N % should increase further once PR 1587 is merged as well. Thanks Haowen! 
image
image

Issue #1397, opened 2024-03-11
Generating only desired output files 


tomhuhh:
**Is your feature request related to a problem? Please describe.**
Is it possible to filter out the other output files and only generate the desired output file(s) as highlighted below?  

<img width="915" alt="Screenshot 2024-03-07 at 11 20 56 AM" src="https://github.com/RuminantFarmSystems/MASM/assets/52800053/77d49bb9-af08-4fc5-bde0-b488a3a62e34">

This feature can also benefit the users when sometimes the user forgets that they have multiple filters (with undesired ones) to avoid generating output files for all the filters.  

This can save runtime too because some of these output files take a long time to generate. 

To conclude, it may be beneficial to include all these output files for development purposes, but it seems impractical in real world application.   

**Describe the solution you'd like**
When initializing the simulation, there is an option to select which output files the user would like to generate. 

**Describe alternatives you've considered**
N/A

**Additional context**

Isgotkowitz:
@tomhuhh I am working on this issue. I think we need to keep the overhead files (errors, logs, warnings, var names, etc.) because they help provide important context, and it is good to always have them generated even when you think you're not going to need them.

Would an acceptable fix to this issue be to put the different types of outputs of into different folders? I.e. reports go into a report folder, jsons go into a jsons folder, and logs/error/warnings go into their own folder.

Isgotkowitz:
Moving this issue to blocked until the task manager is finished.

Isgotkowitz:
Now that task manager is finished, options can be added to tasks for what the different output directories should be for JSONs, reports, and simulation messages.

tomhuhh:
Thanks for following up on this, Ed! I have a small concern about the space the simulation outcomes might take up on my computer, given the multiple rounds and the files they generate. Perhaps we could offer users the option to select which files they'd like to output for each simulation scenario? I think that might help manage the storage better. Thanks!

Isgotkowitz:
> Thanks for following up on this, Ed! I have a small concern about the space the simulation outcomes might take up on my computer, given the multiple rounds and the files they generate. Perhaps we could offer users the option to select which files they'd like to output for each simulation scenario? I think that might help manage the storage better. Thanks!

When you say "select which files they'd like to output for each simulation scenario" do you mean which filters they would like to have files produced for?

tomhuhh:
[...]

PR #1648, opened 2024-05-13
Create directories for different output types

Isgotkowitz:
This PR creates gives users the ability to specify directories for reports, JSONs outputs, and nondata outputs in the same manner that graphics and CSV directories are specified.

### Context

Issue(s) closed by this pull request: 

- [x] Yes
- [ ] No

### What

For any given task, the following directories can now be defined:
- "JSON_directory": this is where JSON outputs created from filter files will be placed.
- "report_directory": this is where reports will be placed.
- "nondata_output_directory": this is where logs, variable names, and other similar RuFaS outputs will be placed.

### Why

Some users/developers on RuFaS find it very helpful to have different outputs sorted into different directories, so they can easily hone in on the few outputs they are interested in. Others find it very helpful to have all outputs placed into a single directory so they are easier to gather and iterate over. The changes in this PR allow both types of users to have what works best of them by configuring the tasks to their liking.

### How

The task properties have been modified so that they now accept paths to place reports, JSONs, and nondata outputs in.
These paths are passed into various methods of OutputManager, which uses them when saving the different outputs.

#### Other Changes
There are a number of other noteworthy changes in this PR.
- A new tasks input has been written, `default_task.json`, and this task is now run by default instead of `sample_tasks.json`. The default task now runs the default scenario once with the random seed set to 42 - in other words it mimics the behavior of scenario manager before it was deprecated.
- The unit tests in OutputManager have been updated to make heavier use of the MockerFixture and instead of passing strings in many cases, `Path`s are passed instead - this brings the number of type errors down quite a bit.
- The pattern for acceptable output directories in the task manager's metadata has been changed so that it will accept relative as well as absolute Posix-compliant paths (i.e. it will accept `/home/yourhome/rufas/output/` as well as `output/`).

### Test plan

The test suite has been run with the updated unit tests to verify that changes are working as they should. Also, the default tasks and sample tasks have been run several times with different output directories specified to verify that output is correctly placed into the specified directories.


input/metadata/properties/tasks_properties.json
```diff
@@ -56,34 +56,46 @@
...
           "default": "output/output_filters/"
         },
         "CSV_directory": {
           "type": "string",
-          "pattern": "^([a-zA-Z]:)?[\\\\/](?:[a-zA-Z0-9_\\-\\.]+[\\\\/])*[a-zA-Z0-9_\\-\\.]*$",
+          "pattern": "^([a-zA-Z]:)?[\\\\/]?(?:[a-zA-Z0-9_\\-\\.]+[\\\\/])*[a-zA-Z0-9_\\-\\.]*$",
           "description": "The directory for the csv output files to be saved",
           "default": "output/CSVs/"
         },
+        "JSON_directory": {
```
PooyaHekmati: I'm worried that `json` could be confused with its broader use. maybe change to `json_output_directory`?

input/metadata/properties/tasks_properties.json
```diff
@@ -56,34 +56,46 @@
...
+          "pattern": "^([a-zA-Z]:)?[\\\\/]?(?:[a-zA-Z0-9_\\-\\.]+[\\\\/])*[a-zA-Z0-9_\\-\\.]*$",
           "description": "The saving directory for graphics",
           "default": "output/graphics/"
         },
+        "nondata_output_directory": {
+          "type": "string",
+          "pattern": "^([a-zA-Z]:)?[\\\\/]?(?:[a-zA-Z0-9_\\-\\.]+[\\\\/])*[a-zA-Z0-9_\\-\\.]*$",
+          "description": "The directory for saving log message files, variable names, and other simulation outputs to.",
+          "default": "output/nondata_output/"
+        },
```
PooyaHekmati:
```suggestion
        "logs_directory": {
          "type": "string",
          "pattern": "^([a-zA-Z]:)?[\\\\/]?(?:[a-zA-Z0-9_\\-\\.]+[\\\\/])*[a-zA-Z0-9_\\-\\.]*$",
          "description": "The directory for saving log message files, variable names, and other simulation outputs to.",
          "default": "output/logs/"
        },
```
thoughts?

input/data/tasks/default_task.json
```diff
@@ -0,0 +1,12 @@
+{
+  "parallel_workers": 8,
```
PooyaHekmati: 8 is a high value for default, let's reduce it to 4 or even 2

PooyaHekmati (approved):
LGTM overall, some minor comments in the code.

input/metadata/properties/tasks_properties.json
Isgotkowitz: I think it makes more sense, but my worry was that people might get confused because there are log-level logs, but also error logs, warning logs, and other things that are useful to group with logs (get data logs, variable names, etc) - but "logs" is cleaner so I'll change it to that.

input/metadata/properties/tasks_properties.json
Isgotkowitz: I am worried about that too, and the same issue is present for the CSV output directory.

input/data/tasks/sample_tasks.json
```diff
@@ -1,5 +1,5 @@
 {
-  "parallel_workers": 8,
+  "parallel_workers": 4,
```
allisterakun, line 2:
```suggestion
  "parallel_workers": 8,
```
Revert this?

allisterakun (approved):
LGTM, good job, Ed!
Just one minor comment: the names of the logs, warnings, and errors are inconsistent throughout the code. They are called "logs" or "non-data" 
```python
output_manager.dump_all_nondata_pools(args["logs_directory"]
```
Do we want to choose one name for consistency?

input/data/tasks/sample_tasks.json
Isgotkowitz: Pooya suggested lowering the number of workers, possibly to down to two, so I will keep it like this for now.

Isgotkowitz:
> LGTM, good job, Ed! Just one minor comment: the names of the logs, warnings, and errors are inconsistent throughout the code. They are called "logs" or "non-data"
> 
> ```python
> output_manager.dump_all_nondata_pools(args["logs_directory"]
> ```
> 
> Do we want to choose one name for consistency?

Good question. Pooya suggested using the term "logs" instead of "nondata" so I went with it - you're right that it is inconsistent, but I think it is small enough that it will not be a big deal.

PooyaHekmati:
> > LGTM, good job, Ed! Just one minor comment: the names of the logs, warnings, and errors are inconsistent throughout the code. They are called "logs" or "non-data"
> > ```python
> > output_manager.dump_all_nondata_pools(args["logs_directory"]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do we want to choose one name for consistency?
> 
> Good question. Pooya suggested using the term "logs" instead of "nondata" so I went with it - you're right that it is inconsistent, but I think it is small enough that it will not be a big deal.

These are two different domains with different contexts. Just because something is called in a certain way in the code doesn't mean it has to have the same name in the in/output.

input/metadata/properties/tasks_properties.json
PooyaHekmati: > I am worried about that too, and the same issue is present for the CSV output directory.

Changing the CSV makes them consistent indeed, but can change how people are using RuFaS.

Isgotkowitz:
Since this PR has two approvals and is in pretty good shape, I am going to merge it. However, it does not fully address #1397 and I think it would be better to write a separate PR that addresses instead of enlarging this PR more. I will write a follow-up PR that will allow users to do the following things:
- Suppress the writing of log files from tasks. This option will be specified in the tasks.
- Suppress the writing of log files from the Task Manager. This option will be specified as a GNU argument.
- Change the directory that log files from the Task Manager are written to. This option will be specified as a GNU argument.

PR #1649, opened 2024-05-13
Pilot testing CS C2F1

Rebecavb:
Adding the inputs for farm 1 cohort 2 for the soil and crop module


jadamchick (approved):
Thanks Rebeca!

PR #1651, opened 2024-05-13
Update path parameters to using Path type

### LINKED ISSUE #1600: [IO] Make the type of path parameters consistent
## Status

In `InputManager` and OutputManager, some methods type path parameters as `str` while others type them as `Path` objects. 

## Intention

Pick one type (preferrably Path from pathlib) and change all the annotations of the other type to this type. 

## Request

- If using Path objects, it would be easier to join them together with the `/` notation.

ltp23:
### Context

Issue(s) closed by this pull request: closes #1600 

- [ ] Yes
- [x] No

### What

In `InputManager` and `OutputManager`, some methods type path parameters as `str` while others type them as `Path` objects. 

### Why

- Making all the path parameters have a consistent type would make their expected behaviors easier to predict every time we see a path parameter.

### How

- Update any path parameters that have type `str` to use `Path` type instead.
- Concatenate path parts using the forward-slash shorthand `/` instead of `os.path.join`.

### Test plan

- Since this is more of a refactoring PR, the testing plan is just to make sure no existing behaviors break.


RUFAS/input_manager.py
```diff
@@ -109,13 +109,13 @@ def pool(self, incoming_pool: Dict[str, Any]) -> None:
         """The setter method for __pool"""
         self.__pool = incoming_pool
 
-    def start_data_processing(self, metadata_path: str, eager_termination: bool = True) -> bool:
+    def start_data_processing(self, metadata_path: Path, eager_termination: bool = True) -> bool:
```
ew3361zh, line 112: It looks like there is a reference to this function in task_manager.py's `start()` method where it's sending `metadata_path.as_posix()` rather than keeping it as a Path object. Do you think that's worth updating in this PR as well?

RUFAS/task_manager.py
```diff
@@ -370,14 +370,14 @@ def handle_single_simulation_run(args: Dict[str, Any], output_manager: OutputMan
...
-        """Validates input data saves metadata properies to CSV."""
+        """Validates input data saves metadata properties to CSV."""
         info_map = {
             "class": TaskManager.__name__,
             "function": TaskManager.handle_input_data_audit.__name__,
             "units": MeasurementUnits.UNITLESS,
         }
         output_manager.add_log("Validation start", f"Validating data for {args['metadata_file_path']}...", info_map)
-        is_data_valid = input_manager.start_data_processing(args["metadata_file_path"], eager_termination)
+        is_data_valid = input_manager.start_data_processing(Path(args["metadata_file_path"]), eager_termination)
```
ew3361zh, line 380: It looks like you updated this one but not the one on line 101 (also in task_manager.py). Any reason not to update both?

ew3361zh (approved):
LGTM, Loi. Left a couple comments about updating the second reference to IM's `start_data_processing()` in task_manager.py which currently has the `metadata_path.as_posix()` rather than a Path object.
Nice work!

allisterakun (approved):
Looks good, Loi! No further issues/comments besides Niko's comment about task_manager

RUFAS/input_manager.py
ltp23: For this one, I can remove the `as_posix()` call as it returns a string. 

```
is_data_valid = self.input_manager.start_data_processing(metadata_path)
```

RUFAS/task_manager.py
ltp23: Right, I missed the other one.

PR #1656, opened 2024-05-14
Rebeca c2f4 soil and crop inputs

Rebecavb:
Crop and soil inputs for the farm 4 cohort 2.


jadamchick (approved):
woohoo, another one in! 
In general you won't want to include tasks jsons in these merges, but it looks like the change is just a deleted empty line so I'm not concerned in this case

PR #1661, opened 2024-05-16
Allow Nullability in Metadata 

### LINKED ISSUE #1274: Introduce nullability in metadata
Adds an optional property of metadata, nullability:
- A boolean value indicating if a `null` or `None` value is acceptable for the variable upon initialization.
- This property should default to `False` unless otherwise specified.
- The check for nullability should be an extra step in the validation process whenever encountering a `null` or `None` value for a variable.

ew3361zh:
Introduces checks for nullability in metadata properties for bools, strings, and number types.

### Context

Issue(s) closed by this pull request: closes #1274

- [ ] Yes
- [x] No

### What

There is a need for some input values to be null and in order for that to work, we needed to adjust the checks within the type-validator functions in IM. Nullability is set as False by default in these checks if it is not present. Nullability checks were not added for array or object types.

To allow nullability to a particular value, its corresponding metadata properties section will have to have `nullable` set to `true` e.g.:

```
"breeding_start_day_h": {
          "type": "number",
          "description": "Heifer Breeding Start Day (days) -- Days old when RuFaS initiates estrus protocol (first cycle to occur ~21 days later)",
          "default": 380,
          "minimum": 0,
          "nullable": true,
        },
```

### Why

Input values having the option to be null has been requested by multiple stakeholders.
It was not added for object and array types because we don't currently "fix" those types and it seemed like it could cause issues/unexpected results to allow nullability for those types.

### How

Within each validator function (again for just bools, strings, and number types), there is a check added for nullability in the case where the input value is None. This now returns true before any other check is done. This essentially supercedes any other properties for the value i.e. minimums/maximums/etc won't be taken into consideration if nullability is set to `true` and the value is `None`.

### Test plan

I ran four scenarios:
1. Setting a current value in an input file to null that doesn't have nullability and has no default. This errored during the validation process.
2. Setting a current value in an input file to null to whose properties I added `{"nullable": true}`. This (in the case I chose) errored during the simulation because the value the simulation expected couldn't be `None`. So this was as expected.
3. Setting a current value outside the range set by the metadata properties and setting the default to `None` while setting nullability to true. This errored during the simulation (but after validation) as expected.

### DIFF
--- a/RUFAS/input_manager.py
+++ b/RUFAS/input_manager.py
@@ -1107,6 +1107,10 @@ class InputManager:
         input_data_value = self._extract_input_data_by_key_list(
             input_data, variable_path, variable_properties, called_during_initialization
         )
+
+        if variable_properties.get("nullable", False) and input_data_value is None:
+            return True
+
         variable_path_str = self._convert_variable_path_to_str(variable_path)
 
         info_map = {
@@ -1118,6 +1122,7 @@ class InputManager:
         properties_violation_message = (
             f"Violates properties defined in metadata properties section" f" '{properties_blob_key}'."
         )
+
         if type(input_data_value) is not float and type(input_data_value) is not int:
             warning_string = "Validation: value is not a number"
             warning_message = (
@@ -1163,6 +1168,10 @@ class InputManager:
         input_data_value = self._extract_input_data_by_key_list(
             input_data, variable_path, variable_properties, called_during_initialization
         )
+
+        if variable_properties.get("nullable", False) and input_data_value is None:
+            return True
+
         variable_path_str = self._convert_variable_path_to_str(variable_path)
         info_map = {
             "class": self.__class__.__name__,
@@ -1171,6 +1180,7 @@ class InputManager:
         properties_violation_message = (
             f"Violates properties defined in metadata properties section" f" '{properties_blob_key}'."
         )
+
         if type(input_data_value) is not str:
             warning_name = "Validation: string variable is not a string"
             warning_message = (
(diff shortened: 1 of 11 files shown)


RUFAS/input_manager.py
```diff
@@ -1118,6 +1118,10 @@ def _number_type_validator(
         properties_violation_message = (
             f"Violates properties defined in metadata properties section" f" '{properties_blob_key}'."
         )
+
+        if variable_properties.get("nullable", False) and input_data_value is None:
+            return True
+
```
Isgotkowitz: It may be better to move this even further up in the method, right after the input_data_value is gathered.

RUFAS/input_manager.py
```diff
@@ -1445,6 +1455,13 @@ def _fix_data(
             )
             om.add_error("Validation: invalid data not able to be fixed", error_message, info_map)
             return False
+        if variable_properties.get("nullable", False) is False and variable_properties.get("default") is None:
+            error_message = (
+                f"Variable: '{element_path}' has invalid value: {variable_parent[element_hierarchy[-1]]}"
+                f". Its nullability is set to False and default value is null. {properties_violation_message}"
+            )
+            om.add_error("Validation: invalid data not able to be fixed", error_message, info_map)
+            return False
```
Isgotkowitz: This is more checking the validity of the metadata properties that it is fixing the data, so I think it would be more appropriate to have it handled by the process that validates the metadata.

Isgotkowitz (approved):
Nice job Niko! Left comments on a few small things, but overall everything looks great. Only I thing I strongly encourage changing is moving the validation of the default out of `_fix_data` and leaving that for the validation of the metadata properties.

allisterakun (approved):
Nice job, Niko! This PR is reviewd together during a group-review session. See details in Ed's comments.

PR #1674, opened 2024-05-20
Update order of percolation routines

KFosterReed:
This PR changes the order of operations of the soil water percolation and infiltration routines. Previously soil water would infiltrate the soil and then percolate through the layers. This causes some challenges with the soil water mass balance accounting. 

Now percolation will happen at the begining of the day which will essentially percolate yesterday's soil water and then new water will be applied and infiltrated into the soil and through the soil layers. 

### Context

- [ ] Yes
- [x] No


RUFAS/routines/field/soil/percolation.py
```diff
@@ -77,19 +78,21 @@ def percolate(self, has_seasonal_high_water_table: bool) -> None:
...
             layer_above = self.data.soil_layers[layer_number - 1]
             percolated_water = layer_above.percolated_water
             if layer_number == deepest_layer + 1:
                 self.data.vadose_zone_layer.water_content += percolated_water
             else:
                 self.data.soil_layers[layer_number].water_content += percolated_water
 
-    def _percolate_infiltrated_water(self) -> int | None:
+
+    def percolate_infiltrated_water(self) -> int | None:
```
Isgotkowitz, line 89: I think this method doesn't need to return anything now, because it doesn't matter what the last layer percolated out of was.

PR #1675, opened 2024-05-20
Suppress log files and specify log directory for tasks and the Task Manager

### LINKED ISSUE #1397: Generating only desired output files

Isgotkowitz:
This PR adds two features to more effectively handle outputs: a switch to prevent logs from being written to output files, and the ability to specify where logs from the Task Manager are written to.

Note that I will update the wiki to reflect the changes that have been made in this PR, for both the GNU args and the change to the task metadata properties.

### Context

Issue(s) closed by this pull request: closes #1397 

- [x] Yes
- [ ] No

### What

Users will be able to specify if they want the logs generated by individual tasks and the Task Manager to be written to log files. Logs in this case means all the logs added to the Output Manager, and also the metadata used, variable names, and `get_data` usage counts.

Users can also specify where the Task Manager writes log files to - making its behavior and capabilities more similar to those of individual tasks.

### Why

As discussed in #1397, it is handy to be able to sort and limit which output files are written when running a simulation (or multiple simulations).

### How

To specify whether or not to write logs to a file, users put a bool argument in a task called "suppress_log_files" which defaults to False. That will determine whether or not logs are written to files for that individual task. To stop the Task Manager from writing its logs to files, a new GNU argument has been added.

Another GNU argument has been added which allows users to specify the path to the directory they want the Task Manager's logs to be written to. The behavior of this option is very similar to the way that the log directory is set for individual tasks.

### Test plan

The test suite has been updated and rerun. Also, I have:
- Raised exceptions in both the Task Manager and in the default simulation. I checked that logs are automatically dumped into files, overriding the user's specification of whether or not log files should be written.
- Run the default simulation with "suppress_log_files" turned on and off to verify that logs are not written and written to files, respectively.
- Run the default simulation with the "-s" GNU args specified and unspecified to verify that logs from the Task Manager are not written and written to files, respectively.
- Specified a directory in the GNU args for Task Manager's logs to be written to, and checked that the logs were in fact written there.


ew3361zh:
This mostly looks good, Ed. I have a few questions/thoughts though before approving:

1. I think per our discussion in the dev-team work-together yesterday, it would be helpful to have an option to suppress all logs for all tasks, not just TM. Are there any hesitations on your part for making that the option?
2. I noticed when I raised an error in IM, that the `-s` didn't seem to work as intended. TM errors were still logged as usual. What I did was update line 854 in IM's `_validate_input_by_type()` to be `if "type" in variable_properties:` thereby ensuring the KeyError on 855 would be raised.
3. I'm still a little confused if this is addressing Haowen's issue since he seemed concerned with producing results for filters he forgot were in his output_filters directory. I think this is a good step towards reducing clutter of outputs and am not quite sure how we could address the issue effectively as described. So just want to be sure Haowen is satisfied with the steps taken to address his concerns.

Isgotkowitz:
> This mostly looks good, Ed. I have a few questions/thoughts though before approving:
> 
> 1. I think per our discussion in the dev-team work-together yesterday, it would be helpful to have an option to suppress all logs for all tasks, not just TM. Are there any hesitations on your part for making that the option?

I would be hesitant about it, because the same result is achievable through a combination of the `-s` GNU arg and "suppress_log_files" in the task specifications. And even though it is more of a pain to have no logs by combining these two methods, I think that is actually a good thing because it nudges people towards having the logs.

> 2. I noticed when I raised an error in IM, that the `-s` didn't seem to work as intended. TM errors were still logged as usual. What I did was update line 854 in IM's `_validate_input_by_type()` to be `if "type" in variable_properties:` thereby ensuring the KeyError on 855 would be raised.

That is deliberate - if there is an uncaught error and a task or the Task Manager crashes, that is indicative of a serious bug and one where we would want all the information we can get about how to reproduce and debug it. For example, if an exception happens in a randomly seeded simulation and the verbosity is set so that logs are not written to the console there would be no other way to recover the seed except through reading the written log files.

> 3. I'm still a little confused if this is addressing Haowen's issue since he seemed concerned with producing results for filters he forgot were in his output_filters directory. I think this is a good step towards reducing clutter of outputs and am not quite sure how we could address the issue effectively as described. So just want to be sure Haowen is satisfied with the steps taken to address his concerns.

I will leave the final word on this to @tomhuhh. I interpreted issue #1397 as taking issue with both the extra reports and the logs always being written. The logs always being written can be changed, and the task manager can filter different simulations with different sets of filters, but ultimately there will be an output for every filter that is given to a task.

ew3361zh (approved):
LGTM. Raised some concerns and questions in the comments and Ed addressed them well (pending Haowen's approval). But code-wise this looks good and is well thought out.

allisterakun (approved):
LGTM, good job Ed!

tomhuhh (approved):
LGTM! Works as expected. Thanks Ed for working on this!

PR #1676, opened 2024-05-20
Incorporates Compare Properties Functionality into IM/TM

### LINKED ISSUE #1511: [IM] Add feature to compare two metadata properties

ew3361zh:
Moves `compare_metadata_properties()` functionality from its own file into IM/TM.

### Context

Issue(s) closed by this pull request: closes #1511

- [ ] Yes
- [x] No

### What

Original functionality is described in PR #1557. That still holds but is now incorporated in TM and IM and the test coverage has also been migrated over to IM testing.

### Why

This was a feature requested by stakeholders/partners and it was requested to be added in as a task to be run via IM.

### How

Same as in #1557 with these new additions:

1. New task `COMPARE_METADATA_PROPERTIES` has been added to TM. 
2. It is essentially an orphan functionality of IM because it doesn't need nor want any other functionalities to interfere.
3. The task to run this will look like this:

```
{
      "task_type": "COMPARE_METADATA_PROPERTIES",
      "properties_file_path": "input/metadata/properties/default.json",
      "comparison_properties_file_path": "input/metadata/properties/default_copy.json",
      "output_prefix": "compare",
      "log_verbosity": "errors"
    }
```

### Test plan

Create a copy of the default metadata properties file and make some changes to it (delete whole branches, delete single entries, change values, change keys).
Add a task similar to the above example into tasks json file you're working with and be sure to update the testing properties file path for comparison.


RUFAS/input_manager.py
```diff
@@ -2278,6 +2279,64 @@ def _create_record(self, data_entry: Dict[str, Any], name: str) -> Dict[str, Any
             "minimum": data_entry.get("minimum", ""),
         }
 
+    def compare_metadata_properties(self, properties_file_path: Path, comparison_properties_file_path: Path) -> None:
+        """
+        Compares two metadataa properties json files using the DeepDiff package and saves the results in a txt file.
```
```suggestion
        Compares two metadata properties json files using the DeepDiff package and saves the results in a text file.
```

RUFAS/input_manager.py
```diff
@@ -2278,6 +2279,64 @@ def _create_record(self, data_entry: Dict[str, Any], name: str) -> Dict[str, Any
...
+        data2 = deepcopy(self.meta_data)
+
+        diff = DeepDiff(data1, data2, ignore_order=True, verbose_level=2)
+
+        file_name = (
+            "diff_results_"
+            + os.path.basename(str(properties_file_path))
+            + "_vs_"
+            + os.path.basename(str(comparison_properties_file_path))
+        )
```
```suggestion
        first_file_path = os.path.basename(str(properties_file_path))
        second_file_path = os.path.basename(str(comparison_properties_file_path))
        file_name = f"diff_results_{first_file_path}_vs_{second_file_path}"
```
I think f-strings make this a little cleaner.

RUFAS/input_manager.py
```diff
@@ -2278,6 +2279,64 @@ def _create_record(self, data_entry: Dict[str, Any], name: str) -> Dict[str, Any
...
+        try:
+            om.add_log("Save metadata diff try", f"Attempting to save to {file_name}", info_map)
+            with open(f"output/{file_name}.txt", "w") as file:
```
Isgotkowitz: Instead of hardcoding the directory, maybe it would be better to use one of the directories that is passed either through the tasks or the GNU args. Possibly the output directory that is passed to `TaskManager` with the `-o` argument.

input/metadata/properties/tasks_properties.json
```diff
@@ -22,6 +22,18 @@
...
+          "pattern": "^(?:[a-zA-Z]:[\\/])?(?:[a-zA-Z0-9._-]+[\\/])*(?:[a-zA-Z0-9._-]+\\.json)$",
+          "description": "The path to the first metadata properties JSON file to load and compare to the other metadata file loaded",
+          "default": "input/metadata/properties/default.json"
+        },
+        "comparison_properties_file_path": {
+          "type": "string",
+          "pattern": "^(?:[a-zA-Z]:[\\/])?(?:[a-zA-Z0-9._-]+[\\/])*(?:[a-zA-Z0-9._-]+\\.json)$",
+          "description": "The path to the second metadata properties JSON file to load and compare to the other metadata file loaded",
+          "default": "input/metadata/properties/default.json"
+        },
```
Isgotkowitz, line 36: Would it be good to have these inputs be nullable?

Isgotkowitz (approved):
LGTM. Left a few comments and suggestions, but all pretty minor. Nice job Niko.

RUFAS/input_manager.py
ew3361zh: Much better, thank you

input/metadata/properties/tasks_properties.json
ew3361zh: I guess that makes sense here - we actually don't want a default at all really because then the fnf errors won't ever trigger. I will make that adjustment.

RUFAS/input_manager.py
ew3361zh: Ah, yes, this was a remnant of this being its own file without the OM or TM being involved. Good catch.

input/metadata/properties/tasks_properties.json
ew3361zh: I'm rethinking this now. We don't necessarily want this to be nullable because there's not a situation where someone would be using this feature and would put in null for one of the two paths, I don't think.

And it looks like the fnf is raised if the path is bad. I'm going to leave this as is for now but will bring it up for discussion in the work-together today.

ew3361zh: Maybe a good solution would be to remove the patterns for both of these instead?

Isgotkowitz: I think it would be better to leave the patterns in, I think that added layer of checking user input is a good thing to have. Also leaving the default in as-is seems like a good idea too, because you're right - there is no situation where null should be passed in when comparing two metadata files.

RUFAS/input_manager.py
```diff
@@ -2278,6 +2279,64 @@ def _create_record(self, data_entry: Dict[str, Any], name: str) -> Dict[str, Any
...
+        self.meta_data = {}
+        self._load_metadata(comparison_properties_file_path)
+        properties2 = self.meta_data
+
+        diff = DeepDiff(properties1, properties2, ignore_order=True, verbose_level=2)
+
+        first_file_path = os.path.basename(str(properties_file_path))
+        second_file_path = os.path.basename(str(comparison_properties_file_path))
+        file_name = f"diff_results_{first_file_path}_vs_{second_file_path}"
+        print(str(output_directory))
```
allisterakun: Remove this print statement

allisterakun (commented):
LGTM, just an extra print statement I think we should remove!

allisterakun (approved):
LGTM, just an extra print statement I think we should remove!

PR #1677, opened 2024-05-20
Add option to record info maps a single time in Output Manager

### LINKED ISSUE #1560: [OM] Allow the option to deduplicate info maps objects
Currently, we only have two options to manage the info maps. By default, info maps are included in their entirety along with each variable. We can also switch this off totally by using the `-i` gnu arg flag.

It is also possible to allow a third option that would only keep one copy of an info map object per variable, assuming all the info map objects are identical for that variable.

Suggested option names: NONE, ALL, ONE - or - EXCLUDE ALL, INCLUDE ALL, INCLUDE ONE.

Isgotkowitz:
This PR provides an avenue for deduplicating info maps by adding an option to `add_variable`.

### Context

Issue(s) closed by this pull request: closes #1560

- [ ] Yes
- [x] No

### What

An option has been added to Output Manager's `add_variable` method which allows calling methods to specify if all info maps should be recorded for a variable, or if it should just the first one.

### Why

RuFaS has a large memory footprint, in some cases so large that it prevents the simulation from being able to complete. Info maps, while important, are often composed of information that is duplicated in every info map. By providing the option to save only a single info map in the Output Manager's pool, the useful aspects of the info maps can (mostly) be kept while not using up tons of memory.

### How

The option "record_all_info_maps" has been added to `add_variable` and `_add_to_pool` in the Output Manager, and defaults to true. When true, it records all info maps as normal (unless info maps are excluded). When false, it will only record the info map in the Output Manager's pool when that variable is recorded for the first time. After that, it will disregard the info map.

### Test plan

The test suite for Output Manager has been updated and run. Also, a random variable has been picked (I chose `infiltrated_water` in `field_data_reporter.py`), and ran the default simulation twice, changing the argument for `record_all_info_maps`. In the JSON output, the info map for `infiltrated_water` was recorded only once when `record_all_info_maps` was set to false. In the CSV and report outputs there was no difference.

### Important
Deduplicating info maps using this option is only a good route to take if it can be assumed while writing the code that the info maps will be the same for every recording of a given variable. This assumption holds true for many variables in RuFaS right now, but may not as simulations and the information recorded in them become more complex.


RUFAS/output_manager.py
```diff
@@ -169,12 +171,20 @@ def _add_to_pool(
             The value to be added to the pool.
         info_map : Dict[str, Any]
             The info map to be added to the pool.
+        record_all_info_maps : bool, default True
+            If true, records all info maps passed for every variable. If false, records only the first info map passed
+            for a variable.
```
```suggestion
            If true, records all info maps passed for a variable. If false, records only the first info map passed for that variable.
```

RUFAS/output_manager.py
```diff
@@ -206,14 +216,18 @@ def add_variable(self, name: str, value: Any, info_map: Dict[str, Any]) -> None:
             Has no effect on manual prefix overrides.
         info_map["suffix"] : str, optional
             If present, gets appended to the key
+        record_all_info_maps : boo, default True
+            If true, records all info maps passed for every variable. If false, records only the first info map passed
+            for a variable.
```
```suggestion
            If true, records all info maps passed for a variable. If false, records only the first info map passed for that variable.
```

ew3361zh (approved):
I like this feature! Nice work, Ed. Worked as intended when I tested it with a couple different vars.

I wonder if the var name is clear enough though?

I think the functionality it's providing is a little masked behind the name i.e. the option to include a single info map doesn't seem intuitive to setting `record_all_info_maps` to `False`. Does that mean no info maps will be recorded? Some info maps? I understand it's clearer with the definition so maybe that's enough (and I'm fine with that if others agree it's enough as is) but I think my preference would be to have it be something like `record_first_info_map_only` or `record_single_info_map` and set the default to False (and update the rest of the associated logic).

RUFAS/output_manager.py
```diff
@@ -206,14 +216,18 @@ def add_variable(self, name: str, value: Any, info_map: Dict[str, Any]) -> None:
             Has no effect on manual prefix overrides.
         info_map["suffix"] : str, optional
             If present, gets appended to the key
+        record_all_info_maps : boo, default True
+            If true, records all info maps passed for every variable. If false, records only the first info map passed
```
JoeWaddell:
```suggestion
        record_all_info_maps : bool, default True
            If true, records all info maps passed for every variable. If false, records only the first info map passed
```

JoeWaddell (approved):
Great quality of life change! Elegant implementation, good test coverage. I will have to make some tweaks to the `data_padder` method, so I opened the following PR: #1701

Isgotkowitz:
> I like this feature! Nice work, Ed. Worked as intended when I tested it with a couple different vars.
> 
> I wonder if the var name is clear enough though?
> 
> I think the functionality it's providing is a little masked behind the name i.e. the option to include a single info map doesn't seem intuitive to setting `record_all_info_maps` to `False`. Does that mean no info maps will be recorded? Some info maps? I understand it's clearer with the definition so maybe that's enough (and I'm fine with that if others agree it's enough as is) but I think my preference would be to have it be something like `record_first_info_map_only` or `record_single_info_map` and set the default to False (and update the rest of the associated logic).

You're right Niko, that arg name is not very clear - I will change it to `record_first_info_map_only` (or maybe just `first_info_map_only`).

Isgotkowitz:
> Great quality of life change! Elegant implementation, good test coverage. I will have to make some tweaks to the `data_padder` method, so I opened the following PR: #1701

Thanks Joe:)

PR #1678, opened 2024-05-21
C2f27 crops

krb78:
pushing c2f27 crops to pilot testing branch. 

need to run without water and temp stress for all crops, run without N stress for corn silage and wheat to achieve yields as well.


jadamchick (approved):
Approved!
I created a separate config file for the w_crops version, and updated the w_crops metadata file path. This is a little more convenient than having to keep switching config file settings based on whether we are running the farm for animals or crops. 
Is there a reason that the config (for crops) starts in 2015? it looks like all of the crop inputs start in 2017. I ran it and it seems to run fine as is, so mostly just curious.

PR #1686, opened 2024-05-21
Update Expiring Github Dependency

### LINKED ISSUE #832: `set-output` command in GitHub linting action dependency is being deprecated
GitHub Actions is deprecating the `set-output` command ([read this](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)), which is used in the linting action workflow. The dependency is [get-changed-files](https://github.com/jitterbit/get-changed-files) by jitterbit.

## Expected Behavior

The GitHub linting action should run without error.

## Current Behavior

The GitHub linting action runs fine, but produces errors.

## Possible Solution

For jitterbit's implementation, then make the changes and maintain it internally. This may be extremely easy, because someone has already done this in a [PR](https://github.com/jitterbit/get-changed-files/pull/70).

## Steps to Reproduce

1. Click on the linting "Details" button from any linting run, and expand the "Setting up dependency for getting changed files".
2. Scroll down and you will see the warnings.
3.
4.

### LINKED ISSUE #1453: [GH Actions] Fix the warnings related to nodejs and `set-output` command when running actions
## Urgency Level

- [x] Low: the issue doesn't have a significant immediate impact

## Status

When running github actions, I see the following warnings. 

image

## Intention

The actions are still running okay but we probably want to update the versions of things before they get deprecated. 

## Request

- [ ] Fix nodejs-related warnings.
- [ ] Fix `set-output` deprecation warnings.
- [ ] Update all the yaml files that we currently have.

### LINKED ISSUE #1657: Add `default_task.json` to .gitignore and the list of protected files
After #1648 is merged, there will be a `default_task.json` input that will run the default scenario.

## Status

`default_task.json` is a normal file, that can be changed through a PR like any other input file.

## Intention

Protect the `default_task.json` file so that it will not be changed accidentally through a pull request.

## Request

Add the following path to the gitignore file: `input/data/tasks/default_task.json`.
Modify the GitHub Actions workflow `combined_format_lint_test_mypy.yml` so that it will fail if `default_task.json` is modified.

ew3361zh:
Updates an expiring Github dependency we're using as well as updating some of the soon-to-be-deprecated GH action commands and updating a package to update the README to the latest version.

### Context

Issue(s) closed by this pull request: closes #1453 closes #1657

Ed's edit: closes #832

- [ ] Yes
- [x] No

### What

The jitterbug package we used in our github actions to collect all the changed files is expiring soon and currently is giving us warnings on some of the soon-to-be-deprecated Github commands it uses (`set-output`). This PR seeks to replace that with a longer-term solution of manually gathering those files.

The EndBug/add-and-commit package was also using the `set-output` GH command but they have a newly updated version (9) which I've updated us to use here and it fixed the warnings.

Also removes another use of `set-output` which was unneeded.

### Why

There have been warnings that these features/packages are being deprecated and fixing them before they are deprecated will help our GH actions continue to run as expected.

### How

Rather than switching to another dependency, the GH action is updated to now manually get the changed files (both Python files and non-Python files).

Removed other usages of `set-output` by updating dependencies to their latest versions or checking for the need for our own use of `set-output` and removing it or updating it where applicable.

### Test plan

To test:
- Changed a .py file and ensured modified files were being accurately found (they were).
- Added that changed .py file to ensure protected paths were still able to be found (they were).
- Check that other functionalities dependent on that list of modified files were still functioning as intended (Black, e.g.).
- Checked that README was still being properly updated.


.github/workflows/combined_format_lint_test_mypy.yml
```diff
@@ -30,32 +30,27 @@ jobs:
...
+
+          OTHER_FILES=$(echo "$ALL_CHANGED_FILES" | grep -v '\.py$' | tr '\n' ' ')
+          echo "OTHER_FILES=$OTHER_FILES" >> $GITHUB_ENV
 
       - name: Check for Unauthorized Changes
         run: |
-          paths_to_protect=(
-            'RUFAS/scenario_manager.py'
-          )
+          paths_to_protect=()
```
allisterakun: Do we want to protect the metadata properties file? Any other files/directories?

.github/workflows/combined_format_lint_test_mypy.yml
```diff
@@ -212,12 +206,11 @@ jobs:
           ./.github/workflows/update_coverage_mypy_badges.sh
 
       - name: Commit & Push Badge Updates
-        uses: EndBug/add-and-commit@v7
+        uses: EndBug/add-and-commit@v9
         with:
           message: "Update badges on README"
           add: "README.md"
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          branch: ${{ github.head_ref }}
```
allisterakun, line 220: Just curious, why is this line removed?

allisterakun (approved):
Generally, it looks good; just a couple of questions. Good job, Niko!

.github/workflows/combined_format_lint_test_mypy.yml
Isgotkowitz: Yes! `default_task.json` - #1657

.github/workflows/combined_format_lint_test_mypy.yml
Isgotkowitz, line 1: There are still some warnings being produced when GH Actions run - I believe it is related to `uses actions/setup-python@v4`. If it is updated to v5, I think it will make them go away.

Isgotkowitz (commented):
LGTM - nice job Niko!

Isgotkowitz (commented):
Whoops meant to approve

Isgotkowitz (approved):

.github/workflows/combined_format_lint_test_mypy.yml
ew3361zh: With the update to the `EndBug` package they did, `branch` was no longer a valid arg. We've also already checked out `github.head_ref` earlier in the same job.

.github/workflows/combined_format_lint_test_mypy.yml
ew3361zh: Thanks, Ed, that seems to have done the trick!

PR #1687, opened 2024-05-21
Consolidate code for input validation in the Input Manager

### LINKED ISSUE #1488: [IM] Consolidate code for checking for missing data with other data validation code
All code that handles the validation of input data should be consolidated into one set of methods.

## Expected Behavior

The checking of input data identifying missing data should be an aspect of validating input data, not a separate process.

## Current Behavior

Input data is checked recursively to see if there is any missing data with the method `_add_default_values_to_missing_inputs`. After this method is finished running, the input data is handed off to `_validate_input_data_by_type` to which recursively checks the input for validity.

## Possible Solution

Adjust the logic of `_validate_input_data_by_type` and it's validation subroutines so that it handles missing data and validates it appropriately.

## Other Considerations
One thing to keep in mind while addressing this issue is to determine if input data that is explicitly said to have a value of `None` should be considered missing or not.

Isgotkowitz:
This PR consolidates the code for handling missing or extra keys in input data into the main routine for validating and fixing data.

### Context

Issue(s) closed by this pull request: closes #1488 

- [ ] Yes
- [x] No

### What

There were several methods for handling missing and extra data in inputs which existed outside the main routine for validating and fixing faulty data. These methods have been removed, and their functionality moved into the main routine that handle data validation (i.e. `_validate_input_by_type` and the type-specific subroutines it calls).

### Why

Having data fixing happen in multiple places was problematic because it made the code much more convoluted and difficult to understand than it needed to be. For example, replacing data with a default should occur in only one place. If it isn't, the door is opened to inconsistent ways in which data is replaced by defaults, if the logic for replacing data with defaults needs to change it has to change in many places not one, etc.

### How

The following methods for replacing missing data with inputs and removing extra keys have been removed:
- `_filter_input_array_data_by_metadata`
- `_filter_input_data_by_metadata`
- `_log_missing_keys`
- `_add_default_values_to_missing_inputs`
- `_add_default_values_to_array_inputs`

To ensure that extra keys are not added to the data, logic has been added to `_object_type_validator` to identify, remove, and log any keys in an object that are not specified in the properties for that object.

`_fix_data` has been modified so that if a key is missing from a dictionary, the invalid value that is trying to be fixed for that key is `None`.

Also, I have removed the keys "tai_method_h" and "synch_ed_method_h" from HeiferIIs, IIIs, Cows, and Replacements in both `animal_population.json` and the `repr` methods for these classes. These are keys are not specified in the animal population properties, and with the new warning for extraneous keys, the console gets blown up with warnings if these are left in. The animal population and module is otherwise unchanged though. Note that currently herd generation is currently failing on main, so it would appear that these changes are not causing herd generation to fail.

### Test plan

Lines added have had unit tests added and run to verify that their behavior is correct.

To check that data with no corresponding property is added to the Input Manager, dummy data was added to an input file. To verify that it was removed, the warning starting with "Validation: object contains extraneous data." was checked for. Also, a call to `get_data` was added in SimulationEngine to try and get that dummy data - which should result in `None` being returned and an error being logged saying that the data was not found.

To verify that missing data is correctly identified, I have tried removing different data from different levels to check that its absence is identified, and that it will populate the missing data with a default if there is one.


input/metadata/properties/tasks_properties.json
```diff
@@ -138,6 +138,7 @@
           "type": "array",
           "description": "The input variables to perform Sensitivity Analysis on",
           "default": [{}],
```
Isgotkowitz, line 157: I'm not sure how to best handle this: the default here will never be used, because arrays are not a "fixable" data type. One option is to make it nullable (like I've done below), another would be to switch the status of arrays to be fixable.

tests/test_input_manager.py
```diff
@@ -483,20 +373,14 @@ def test_populate_pool_partial_invalid(
...
-    mocker.patch.object(input_manager, "_log_missing_keys")
     mocker.patch("RUFAS.input_manager.om.add_warning")
     mocker.patch("RUFAS.input_manager.om.add_log")
 
     # Act
     result = input_manager._populate_pool(eager_termination=False)
 
     # Assert
-    assert result
+    assert result is False
```
Isgotkowitz, line 394: I'm not sure why the previous expected result was supposed to be true (as in, I understand why the code worked but I don't understand the logic behind it). It seems to me that the correct behavior when finding any invalid data in the entire pool would be to mark the entire pool as invalid, which is what this test now looks for.

RUFAS/input_manager.py
```diff
@@ -1091,6 +813,17 @@ def _object_type_validator(
...
                 return False
+
+        extraneous_keys = [key for key in object_value.keys() if key not in variable_properties.keys()]
+        for key in extraneous_keys:
+            om.add_warning(
+                "Validation: object contains extraneous data",
+                f"Variable: '{variable_path_str}' contains data at key '{key}' that is not specified in the metadata "
+                f"properties. {properties_violation_message}",
+                info_map,
+            )
```
Isgotkowitz, line 836: I don't know if logging warnings for every key removed is the best way to go, but to me it feels right to let users know that RuFaS is not actually using data they are giving it. I had a conversation about this with @PooyaHekmati a while ago and I remember him being in favor of not logging any warnings, as it might really clog up the output.

RUFAS/input_manager.py
```diff
@@ -1473,9 +1206,13 @@ def _fix_data(
             om.add_error("Validation: invalid data not able to be fixed", error_message, info_map)
             return False
 
-        original_invalid_value = variable_parent[element_hierarchy[-1]]
+        if type(variable_parent) is list:
+            original_invalid_value = variable_parent[element_hierarchy[-1]]
+        else:
+            original_invalid_value = variable_parent.get(element_hierarchy[-1])
```
Isgotkowitz, line 1212: If a key is missing from the input, then it will raise a key error if the square bracket syntax is used to retrieve its value. But if `.get()` is used, it will return `None` as the original invalid value when a key is missing, then attempt to fix it in the same way it would any other invalid value.

matthew7838 (approved):
Thanks for the explanation. I will approve the PR, but I agree that Ed raised some important points that need to be resolved before we proceed with the merge.

input/metadata/properties/tasks_properties.json
ew3361zh: I spoke with @JoeWaddell about this since I was running into it in my PR #1640 and he wrote:
"If someone is using sensitivity analysis methods those variables are so critical that we can't rely on a default anyway."

So I think I'd opt to remove the default.

ew3361zh (approved):
This looks good, well implemented. I tested out adding some new vars to the metadata properties to see what would happen and it errored as expected. Also added some vars to one of the input files and made sure it wasn't present after the check and it returned "None" and also wasn't listed in the containing structure.

I think some failing tests got merged into main but that doesn't seem to have anything to do with the changes here.

matthew7838 (approved):
LGTM

PR #1709, opened 2024-05-23
c2f10 crop and soil inputs

Rebecavb:
Cohort 2 farm 10 inputs

### DIFF
--- a/input/data/tasks/default_task.json
+++ b/input/data/tasks/default_task.json
@@ -3,9 +3,9 @@
   "tasks": [
     {
       "task_type": "SIMULATION_SINGLE_RUN",
-      "metadata_file_path": "input/metadata/c2f10_wcrops_metadata.json",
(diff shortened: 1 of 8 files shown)


jadamchick (approved):
Hi Rebeca! Looks good, the only thing is can you please not include the updates to default_task.json?

input/data/tasks/default_task.json
jadamchick, line 1: can you please revert this back to the default before merging? thanks!

krb78 (approved):
Looks great! The yields aren't what they should be even with the irrigation - but this might get fixed (I know hydrology of the soil layers is still being worked out).

Rebecavb:
> Hi Rebeca! Looks good, the only thing is can you please not include the updates to default_task.json?

Hi Julie, yes absolutely. Thanks for the review.

PR #1719, opened 2024-05-28
Update Soil Denitrification parameters

KFosterReed:
This is a two line change that fixes:

- [x] A double conversion typo from soil carbon concentration to percent

### Context

- [ ] Yes
- [x] No


Isgotkowitz (approved):
@KFosterReed nice catch. I am approving, the change to denitrification looks good to go. However the test has not been updated, that should be done before this gets merged. Let me know if you would like my help with this.

Also I think the message in the change log could be more specific, like "Removes an incorrect conversion from a fraction to percent when calculating denitrification."

Edit: I have not looked at the results from this change. I am curious, have you taken a look at them? It would be good to double check that they don't have a radical effect on N cycling before this gets merged into the main.

KFosterReed:
> @KFosterReed nice catch. I am approving, the change to denitrification looks good to go. However the test has not been updated, that should be done before this gets merged. Let me know if you would like my help with this.
> 
> Also I think the message in the change log could be more specific, like "Removes an incorrect conversion from a fraction to percent when calculating denitrification."
> 
> Edit: I have not looked at the results from this change. I am curious, have you taken a look at them? It would be good to double check that they don't have a radical effect on N cycling before this gets merged into the main.

Great question! Yes, I actually just did that today and the changes are even more in line with what would be expected: less nitrogen stress on crops due to more N available, more ammonia emissions, more N percolation through layers, N2O emissions much much much closer to expected values.

KFosterReed:
@Isgotkowitz re: test updates. I didn't change anything inside a function and didn't see any failed tests in the logs so I thought that maybe this was a silent bug? If not, I am not sure where / what test needs to be updated so help would be so appreciated!

Isgotkowitz:
> @Isgotkowitz re: test updates. I didn't change anything inside a function and didn't see any failed tests in the logs so I thought that maybe this was a silent bug? If not, I am not sure where / what test needs to be updated so help would be so appreciated!

To run the test suite, you have to run `pytest tests` in the command line - no testing will be run during a regular RuFaS simulation (e.g. running `python main.py`). You can do this locally, or you can check the results in GitHub by clicking on the red X that appears instead of the green check mark.

matthew7838 (approved):
LGTM along with the test fix.

Isgotkowitz:
Keeping this PR up to date so it will be ready to merge when Kristan gets back from vacation

PR #1720, opened 2024-05-28
Updates Github Action to Fetch Main Mypy Error Count From README

### LINKED ISSUE #1710: Remove second Mypy run from GitHub Action
There is an unnecessary second Mypy run on PRs by the GitHub Action `combined_format_lint_test_mypy.yml`.

## Status

To determine the net change in the Mypy error count between the feature branch and the base branch, the Action check's out the main branch and runs Mypy on it (see the step named "Run mypy on the main branch"). This second run of Mypy is unnecessary, because the number of Mypy errors on the main branch has already been recorded in the README.

## Intention

Make the GitHub Action faster/more efficient by removing the Mypy run on main branch.

## Request

Change the "Run mypy on the main branch" task to read the number of Mypy errors in the README on the main branch, instead of rerunning Mypy.

ew3361zh:
Rather than running a second full mypy on main, we instead will get the current mypy count from the README on main which will already have the most updated main mypy error count.

### Context

Issue(s) closed by this pull request: closes #1710

- [ ] Yes
- [x] No

### What

This changes the method for getting and setting the mypy error count for main to use for comparison to the PR branch. Instead of running mypy on the entire main branch, we can extract the mypy error count in the README file on main.

### Why

This will reduce the Github Actions runtime by only running mypy on the PR branch and doing a much quicker fetch of the mypy error count on main from the README.

### How

This replaces the previous `Run mypy on the main branch` GHA step with an action to get the mypy error count from main from the README on main. 
Here is what the new action does:
1. checkout main.
2. get the error count from the README file using grep/regex.
3. set the extracted mypy error count from the README to `mypy_main_branch_error_count.txt` as it was in the previous version of the action that ran mypy on main.

### Test plan

I will use the newly updated Github Actions wiki technique of setting up a test-main branch to ensure this worked as intended. Carried out in #1721:
1. Set up test-main branch.
2. Switched all refs to main in GH Action to test-main.
3. Created a new branch off test-main called test-test-main.
4. Added 2 mypy errors to the code and pushed those changes to the test-test-main PR.
5. Merged in test-test-main to test-main.
6. Successfully updates the README with the updated mypy error count.

### DIFF
--- a/.github/workflows/combined_format_lint_test_mypy.yml
+++ b/.github/workflows/combined_format_lint_test_mypy.yml
@@ -151,7 +151,13 @@ jobs:
         if: github.ref != 'refs/heads/main'
         run: |
           git checkout main
+          # Print the relevant part of README for debugging
+          cat README.md
           error_count=$(grep -oP 'Mypy-\K\d+(?= errors)' README.md)
+          if [ -z "$error_count" ]; then
+            echo "Error count not found in README.md"
+            exit 1
+          fi
           echo "Mypy error count on main branch: $error_count"
           echo $error_count > mypy_main_branch_error_count.txt
 
--- a/.github/workflows/combined_format_lint_test_mypy.yml
+++ b/.github/workflows/combined_format_lint_test_mypy.yml
@@ -147,12 +147,13 @@ jobs:
         id: set_current_branch_error_count
         run: echo "error_count=$(cat mypy_current_branch_error_count.txt)" >> $GITHUB_OUTPUT
 
-      - name: Run mypy on the main branch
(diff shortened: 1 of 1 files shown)


matthew7838 (approved):
LGTM, a small but powerful change.

ew3361zh:
After talking this through with @PooyaHekmati, this isn't a sound idea as the README is not a protected file - it's easily changeable and therefore the MyPy error count recorded in it is not a reliable count. Additionally, after reviewing the runtime for the Github Actions both with and without this second MyPy run, there is little to no difference so it's not saving us any GHA minutes.

PR #1723, opened 2024-05-28
Fixing percolation order

matthew7838:
This PR changes the order of operations of the soil water percolation and infiltration routines. Previously soil water would infiltrate the soil and then percolate through the layers. This causes some challenges with the soil water mass balance accounting.

Now percolation will happen at the begining of the day which will essentially percolate yesterday's soil water and then new water will be applied and infiltrated into the soil and through the soil layers.

### Context

- [ ] Yes
- [x] No

### Why

Original order causes challenges with the soil water mass balance accounting.

### Test plan

Use this filter and assess the mass balance of the soil water additions and losses: 

[csv_soil_water.txt](https://github.com/RuminantFarmSystems/MASM/files/15490241/csv_soil_water.txt)

### DIFF
--- a/RUFAS/routines/field/field/field.py
+++ b/RUFAS/routines/field/field/field.py
@@ -1389,8 +1389,10 @@ class Field:
 
         remaining_evapotranspirative_demand = self._evaporate_from_crop_canopies(full_evapotranspirative_demand)
 
-        self.soil.infiltration.infiltrate(water_reaching_soil)
         self.soil.percolation.percolate(self.field_data.seasonal_high_water_table)
+        self.soil.infiltration.infiltrate(water_reaching_soil)
+        self.soil.percolation.percolate_infiltrated_water()
+
         self.soil.soil_erosion.erode(
             self.field_data.field_size,
             0.02,
--- a/RUFAS/routines/field/soil/percolation.py
+++ b/RUFAS/routines/field/soil/percolation.py
@@ -86,7 +86,7 @@ class Percolation:
             else:
                 self.data.soil_layers[layer_number].water_content += percolated_water
 
-    def percolate_infiltrated_water(self) -> int | None:
+    def percolate_infiltrated_water(self) -> None:
         """
         Percolates infiltrated water into the soil profile.
 
@@ -106,7 +106,6 @@ class Percolation:
         """
 
         water_remaining_to_percolate = self.data.infiltrated_water
-        top_layer_to_percolate = None
         for index, layer in enumerate(self.data.soil_layers):
             acceptable_percolation = layer.acceptable_percolation_amount
             if water_remaining_to_percolate > acceptable_percolation:
@@ -116,11 +115,9 @@ class Percolation:
             else:
                 layer.water_content += water_remaining_to_percolate
                 water_remaining_to_percolate = 0.0
-                top_layer_to_percolate = index + 1
                 break
         if water_remaining_to_percolate > 0.0:
             self.data.vadose_zone_layer.water_content += water_remaining_to_percolate
-        return top_layer_to_percolate
 
     # --- Static methods ---
     @staticmethod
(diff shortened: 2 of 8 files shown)


RUFAS/routines/field/soil/percolation.py
```diff
@@ -107,19 +104,19 @@ def _percolate_infiltrated_water(self) -> int | None:
         that is put below a layer as being percolated by it.
 
         """
-        self.data.set_vectorized_layer_attribute("percolated_water", [0.0] * len(self.data.soil_layers))
+
         water_remaining_to_percolate = self.data.infiltrated_water
         top_layer_to_percolate = None
```
Isgotkowitz:
```suggestion
```
Since this routine is now running after the main percolation routine, there is nothing that needs to be returned by this method.

RUFAS/routines/field/field/field.py
```diff
@@ -1389,8 +1389,10 @@ def _cycle_water(self, current_conditions: CurrentDayConditions, time) -> None:
 
         remaining_evapotranspirative_demand = self._evaporate_from_crop_canopies(full_evapotranspirative_demand)
 
-        self.soil.infiltration.infiltrate(water_reaching_soil)
         self.soil.percolation.percolate(self.field_data.seasonal_high_water_table)
+        self.soil.infiltration.infiltrate(water_reaching_soil)
+        self.soil.percolation.percolate_infiltrated_water()
```
Isgotkowitz, line 1394: The test for `_cycle_water` needs to be updated.

RUFAS/routines/field/soil/soil.py
```diff
@@ -154,8 +154,8 @@ def daily_soil_water_routine(
         profile depend on how much water enters and moves through the soil profile.
 
         """
-        self.infiltration.infiltrate(rainfall, weighting_coefficient, potential_evapotranspiration)
         self.percolation.percolate(has_seasonal_high_water_table)
+        self.infiltration.infiltrate(rainfall, weighting_coefficient, potential_evapotranspiration)
         self.evaporation.evaporate(maximum_soil_evaporation)
```
Isgotkowitz: The second percolation routine should be called here (and the test will need to be updated).

Isgotkowitz (changes requested):
I did a quick scan of the water content and percolation outputs, and I think they look good. Holding off on approving until the following things are addressed though: 
1. `percolate_infiltrated_water` should not return anything now.
2. All changes need to be tested.
3. The changelog needs to be updated.

Also, reading the PR description I was not able to tell how the changes made were evaluated (i.e. the Test Plan section was not filled in). Please include that when writing a PR description.

RUFAS/routines/field/soil/soil.py
matthew7838: To clarify, you mean `percolate_infiltrated_water()`?

KFosterReed (approved):
I looked at the soil water balance between the water added in rainfall, irrigation, and snowmelt, the infiltrated water, and the percolated water in this updated branch and it aligns with what we were seeing before that the external SME team approved.

RUFAS/routines/field/soil/percolation.py
```diff
@@ -77,19 +74,19 @@ def percolate(self, has_seasonal_high_water_table: bool) -> None:
...
+        for layer_number in range(1, layer_count + 1):
             layer_above = self.data.soil_layers[layer_number - 1]
             percolated_water = layer_above.percolated_water
             if layer_number == deepest_layer + 1:
                 self.data.vadose_zone_layer.water_content += percolated_water
             else:
                 self.data.soil_layers[layer_number].water_content += percolated_water
 
-    def _percolate_infiltrated_water(self) -> int | None:
+    def percolate_infiltrated_water(self) -> None:
```
Isgotkowitz, line 89: Docstring should be changed as well.

Isgotkowitz (approved):
Thank you for addressing changes Matthew. Found one spot where a docstring needs to be updated, other than that LGTM

Issue #1714, opened 2024-05-23
Homegrown feed emissions estimation methods


KFosterReed:
**Is your feature request related to a problem? Please describe.**
I have to run to a meeting soon but wanted to get this out before the sprint planning meeting: 
[https://docs.google.com/document/d/1aGeSL-8RXqgZjQPQsoRVvXVlPH-E7QivZcCnnbt16Q8/edit?usp=sharing](https://docs.google.com/document/d/1aGeSL-8RXqgZjQPQsoRVvXVlPH-E7QivZcCnnbt16Q8/edit?usp=sharing)

**Describe the solution you'd like**
Ideally this can be implemented in the EEE module

**Describe alternatives you've considered**

**Additional context**
I will provide a report filter and excel file that includes examples of the exact desired calculations by the end of the day today. 

**Target Deadline**
May 31st

KFosterReed:
I am added an example CSV file to calculate emissions using outputs from an example report that is also attached. These example results were obtained by running the `homegrown_feed_test_metadata` scenarion on the 'Homegrown_feed_emissions' branch
[Template for Homegrown Feed Emissions Calcs.xlsx](https://github.com/RuminantFarmSystems/MASM/files/15436489/Template.for.Homegrown.Feed.Emissions.Calcs.xlsx)
[report_homegrown_feed_emissions.json](https://github.com/RuminantFarmSystems/MASM/files/15436496/report_homegrown_feed_emissions.json)

krb78:
I made some comments/edits to the spreadsheet but it didn't appear to be a live link, so here's a [live excel sheet](https://rosedmi-my.sharepoint.com/:x:/g/personal/kaitlyn_briggs_dairy_org/EcDwYdnlb9NKr_9xfyZqto8B9n-Ynk-G9pclGWscLFqFwA?e=Pb9J7C)

KFosterReed:
There was a duplicate feed ID number in the emissions.py file that had me scratching my head for a bit but I corrected it and then reviewed the corn silage, corn grain, and alfalfa silage actual purchased feed and emissions outputs... so far so good!

PR #1725, opened 2024-05-28
Calculate actual purchased feed totals and related emissions totals

Isgotkowitz:
This PR implements post-processing routines in `emissions.py` to estimate the emissions footprint of farm grown feeds as well as computing the purchased feeds used (and associated emissions) after adjusting for farm grown feeds.

### Context

Issue(s) closed by this pull request: addresses #1714 

- [x] Yes
- [ ] No

### What

`emissions.py` in the EEE module has been filled in to compute homegrown feed emissions and adjusted purchased feed emissions.

It also makes a few changes to how crop yields and synthetic fertilizer applications are recorded, to enable post processing.

### Why

Emissions from (and prevented by) farm-grown feeds are a huge part of a farm's emissions footprint, so it is critical that they are accounted for in RuFaS.

### How

`emissions.py` (roughly) goes through the following steps:
- Collects crop harvests and synthetic fertilizer applications that occurred in the last year of the simulation.
- Collects and aggregates the amounts of purchased feeds that were fed to animals in the last year of the simulation.
- Subtracts the homegrown feed amounts from their purchased feeds equivalents, records the updated purchased feed amounts, then computes and records the emissions associated with the updated purchased feed amounts.
- Collects the emissions and carbon stock changes from the simulated fields.
- Computes and records the emissions associated with individual crop harvests based on their mass relative to other harvests from the same field.

### Test plan

I tested these routines by running the homegrown emissions task in #1731, and comparing the Emissions module's output to expected outputs I computed by hand. Please see that PR for a more detailed description of how the results were obtained and observed.

Also I updated broken unit tests, but did not write any unit tests for the new Emissions routines. Due to the urgency of this task and the complexity of the routines, I didn't feel there was enough time to appropriately unit test everything.


RUFAS/simulation_engine.py
```diff
@@ -89,6 +90,9 @@ def simulate(self) -> None:
         total_simulation_time = t_end_sim - t_start_sim
         total_simulation_time_log = f"Total simulation time is: {total_simulation_time}"
         om.add_log("total_simulation_time", total_simulation_time_log, info_map)
+        om.add_log("Starting processing of emissions", "", info_map)
```
ew3361zh: maybe
```suggestion
        om.add_log("Emissions Processing", "Starting processing of emissions.", info_map)
```

RUFAS/simulation_engine.py
```diff
@@ -89,6 +90,9 @@ def simulate(self) -> None:
...
+        self.emissions.calculate_emissions()
+        om.add_log("Completed processing of emissions", "", info_map)
```
```suggestion
        om.add_log("Emissions Processing", "Completed processing of emissions.", info_map)
```

RUFAS/routines/EEE/emissions.py
```diff
@@ -0,0 +1,359 @@
...
+
+        Examples
+        --------
+        >>> a = {'one': {'values': [1, 2, 3]}, 'two': {'values': [4, 5, 6]}]}
+        >>> _transform_outputs_to_list_of_dicts(a)
+        [{'one': 1, 'two': 4}, {'one': 2, 'two': 5}, {'one': 3, 'two': 6}]
+
+        """
+        keys = data.keys()
+        values_list = [data[key]["values"] for key in keys]
```
ew3361zh, line 175: What happens if the `values_list`s are different lengths? Will yields for different crops/fertilizer applications/etc always have the same amount of data?

I think it will just take the shortest list of data and zip those together so we might potentially lose some data here without knowing it. I might add a warning/error here to OM.

RUFAS/routines/EEE/emissions.py
```diff
@@ -0,0 +1,359 @@
...
+        }
+        fertilizer_apps = om.filter_variables_pool(fertilizer_filter)
+        processed_apps = self._transform_outputs_to_list_of_dicts(fertilizer_apps)
+
+        time_filter = {
+            "name": "Time Filter",
+            "filters": ["Time.(day|calendar_year)"],
+            "slice_start": SLICE_START,
+            "slice_end": SLICE_END,
+        }
```
ew3361zh: It might be helpful to add a description entry for each of these filters to help explain what information they're grabbing and why. I tested adding a `"description": "Filter that gets the cutoff date used for xyz",` for one of the filters and it didn't affect the ability of the filter to run as expected.

RUFAS/routines/EEE/emissions.py
```diff
@@ -0,0 +1,359 @@
...
+                if feed_id in CROP_SPECIES_TO_PURCHASED_FEED_ID[crop]
+            }
+            for key, value in homegrown_alternatives.items():
+                amount_used = min(amount, value)
+                amount -= amount_used
+                homegrown_alternatives[key] -= amount_used
+
+            actual_purchased_feeds[feed_id] = amount
+
+        return actual_purchased_feeds
```
ew3361zh: I think this could be split into 2 separate functions - one to calculate homegrown totals and one to get the actual purchased feeds. This would make eventual testing easier.

RUFAS/routines/EEE/emissions.py
```diff
@@ -0,0 +1,359 @@
...
+        self._calculate_homegrown_feed_emissions(homegrown_feeds, fertilizer_applications)
+
+    def _calculate_purchased_feed_emissions(self, homegrown_feeds: list[dict[str, Any]]) -> None:
+        info_map = {"class": self.__class__.__name__, "function": self._calculate_purchased_feed_emissions.__name__}
+        purchased_feeds = self._gather_ration_feed_totals()
+        actual_purchased_feed_totals = self._calculate_actual_purchased_feeds(homegrown_feeds, purchased_feeds)
+        om.add_variable(
+            "actual_purchased_feed_totals",
+            actual_purchased_feed_totals,
+            dict(info_map, **{"units": MeasurementUnits.KILOGRAMS}),
```
ew3361zh, line 69: I think the fact that we're still adding variables to OM here makes a case for this to be part of SimulationEngine instead of the TM.handle_post_processing as we discussed at the work together.

RUFAS/routines/EEE/emissions.py
```diff
@@ -0,0 +1,359 @@
...
+                emissions_info = {
+                    "crop_type": crop["crop"],
+                    "nitrous_oxide_emissions": crop["nitrous_oxide_emissions"],
+                    "ammonia_emissions": crop["ammonia_emissions"],
+                    "carbon_stock_change": crop["carbon_stock_change"],
+                    "nitrogen_fertilizer_used": crop["nitrogen_fertilizer_used"],
+                    "phosphorus_fertilizer_used": crop["phosphorus_fertilizer_used"],
+                    "field_name": crop["field_name"],
+                }
+                om.add_variable(f"homegrown_{crop_type}_emissions", emissions_info, info_map)
```
ew3361zh: I would also break this up into multiple functions to make it easier to test and keep it to single responsibility.
I know this is an urgent task so if it makes sense to make an issue about doing this so the functionality is available for now, I think that would be okay.

Potential sub-functions:
- group_feeds()
- aggregate_fertilizer_applications()
And then a function that calls those, calls on self._calculate_emissions_by_field() same as now and maybe even a separate function to log the emissions data to OM.

ew3361zh (approved):
I think for the purposes and immediacy of the issue that this is a great job. It's complex and lots of processing is going on but I didn't see anything egregious - I think there's some refactoring and testing down the road but to get the core functionality in place, this LGTM. Nice work, Ed!

KFosterReed:
This still looks good to me! However, as I briefly mentioned in person yesterday, we have one additional request to add to this PR which is to include the manure N applications in the same way as we did the fertilizer N methods. The relevant output is: `Field._record_manure_application.manure_application.field=[field_name].nitrogen`. As with the other methods in this PR, the challenge is to aggregate by field and year.

RUFAS/routines/EEE/emissions.py
```diff
@@ -0,0 +1,450 @@
...
+EMBEDDED_NITROGEN_FERTILIZER_EMISSIONS_FACTOR = 5.32
+EMBEDDED_PHOSPHORUS_FERTILIZER_EMISSIONS_FACTOR = 3.07
+
+
+"""
+These are constants for calculating the upstream emissions of synthetic nitrogen and phosphorus fertilizer. Their units
+are in kg CO2e / kg N and kg CO2e / kg P, respectively.
+"""
+UPSTREAM_NITROGEN_FERTILIZER_EMISSIONS_FACTOR = 3.11
+UPSTREAM_PHOSPHORUS_FERTILIZER_EMISSIONS_FACTOR = 1.84
```
Isgotkowitz: @krb78 2 q's on these constants:
- I want to confirm that the units for the embedded emissions are in fact kg CO2e / kg nutrient,
- Is there a reference I can put in for the upstream emissions factors? You mentioned GWP 28 in the comment on the Google doc, but I am not sure if that is related to the reference Kristan had for the embedded emissions factors.

RUFAS/routines/EEE/emissions.py
```diff
@@ -0,0 +1,448 @@
...
+            crop["phosphorus_fertilizer_used"] = fertilizer_applications["phosphorus"] * fraction_of_total_mass_grown
+            crop["phosphorus_fertilizer_embedded_CO2_emissions"] = (
+                fertilizer_applications["phosphorus"]
+                * fraction_of_total_mass_grown
+                * EMBEDDED_PHOSPHORUS_FERTILIZER_EMISSIONS_FACTOR
+            )
+            crop["phosphorus_fertilizer_upstream_CO2_emissions"] = (
+                fertilizer_applications["phosphorus"]
+                * fraction_of_total_mass_grown
+                * UPSTREAM_PHOSPHORUS_FERTILIZER_EMISSIONS_FACTOR
```
Isgotkowitz: @krb78 can you confirm that this is the right way that these constants should be used?

RUFAS/routines/EEE/emissions.py
Isgotkowitz: Good point, I have added an error for now. But in the future it should be modified so that it will log specifically what data was missing.

RUFAS/routines/EEE/emissions.py
Isgotkowitz: I partly did this. I wrapped first part of this method, the one for aggregating crop yields by type, into its own method. I this will strike a nice balance between single-responsibility and not changing too much in this PR.

RUFAS/routines/EEE/emissions.py
Isgotkowitz: Good point. Following the structure laid out in #1299, the call to estimate emissions will be left in SimulationEngine.

RUFAS/routines/EEE/emissions.py
krb78: yes! kg CO2e/kg nutrient. How much will you hate me if I say: let's call it IPCC AR5 for now, but we change back with Kristan when she is back to double check that is correct. Unless did you write IPCC 2021 or did she? (I ask because there isn't an IPCC report associated with 2021... AR6 was 2022 and AR5 was 2014)

krb78: wait! I take all of this back haha

well not ALL of it - but those upstream_emission factors are from me 🤪 the potassium one would be 1.30 and the reference is BASF’s Eco-efficiency analysis tool representative of U.S. national values (pulled from the Rotz et al 2021 Environmental assessment of United States dairy farms)

RUFAS/routines/EEE/emissions.py
krb78: You are going to hate me Ed, this is not 100% correct. Will it work given how crops grow and where we tend to apply fertilizer? yes. But it's not truly correct :( What we actually need is to break these by the crop that is being grown at a specific date. As an example, the fertilizer applied for corn needs to be associated with the corn crop specifically, not spread among the triticale and the corn if double cropping is occurring. Is it a big deal, no, because corn provides so much of the mass harvested from the field compared to triticale...

krb78: Your math is otherwise correct though. it is the application of the nutrient * the embedded emission.

krb78:
Ed, this is looking really good! I tested it on 4 different pilot farms with all sorts of rotations and it worked beautifully. We just need to decide how easy or not it is to make the adjustment to the synthetic fertilizers. I'm going to hold off on my approval, even though I think we are really really close.

RUFAS/routines/EEE/emissions.py
Isgotkowitz: > What we actually need is to break these by the crop that is being grown at a specific date. As an example, the fertilizer applied for corn needs to be associated with the corn crop specifically, not spread among the triticale and the corn if double cropping is occurring.

Ok. That will be a trickier problem to figure out, but I'm glad you mentioned it so we can keep it in mind.

Isgotkowitz:
> Ed, this is looking really good! I tested it on 4 different pilot farms with all sorts of rotations and it worked beautifully. We just need to decide how easy or not it is to make the adjustment to the synthetic fertilizers. I'm going to hold off on my approval, even though I think we are really really close.

That is excellent to hear @krb78. I am going to make the adjustment in a PR stacked on top of this one for the fertilizers, and I will reach out to you on slack when I have it ready.

RUFAS/routines/EEE/emissions.py
Isgotkowitz: Kristan wrote IPCC 2021, GWP 100 in the [Google doc](https://docs.google.com/document/d/1aGeSL-8RXqgZjQPQsoRVvXVlPH-E7QivZcCnnbt16Q8/edit) describing the homegrown feed emissions estimation method.

Also, for the potassium one I will add it in with the stacked PR on this one, that is dedicated to add K to the fertilizer applications.

Isgotkowitz:
Adding that PR #1750 adds the rest of the functionality required for the Emissions Estimator.

krb78 (approved):
This is great, Ed. It runs as expected. I made issue "Fertilizer Emissions for specific crops #1753" to address the fertilizer allocation because i dont think it's worth our effort right now. Thanks for being a champ and working on this so much and all of my last minute changes/requests!

PR #1733, opened 2024-06-01
handle empty ration in animal module reporter

### LINKED ISSUE #1734: Elegantly handle ration refomulation on a per pen basis
If a pen starts the simulation as empty, a ration is correctly not formulated. This causes problems in AnimalModuleReporter, hence changes in #1733 

Further downstream effects are caused if the pen is populated prior to the next ration formulation interval, since manure calculation methods check if a pen is populated. This means that a ration would "correctly" be empty (e.g. not formulated), but the pen itself would be populated.

As such, we need to add a method to reformulate the ration on a pen by pen basis (currently in `_calc_ration_at_interval` it does so for all pens. 

## Possible Solution

What I propose is to is break up the current method that formulates a ration for all pens such that a helper method operates ration formulation on a single pen. 
That's good to do regardless for code clarity, and reducing the number of indentations in that part of the code.
This way nothing will change under normal circumstances, and it'll allow me to add in a simple bit of logic to check if a pen is populated and was not on the previous day, in which case I'll use the new helper method to reformulate for just that pen.

Then in the helper method to add a new animal to a pen: prior to when the animal is actually added, if the pen.is_populated attribute is False, then we flag it as requiring a ration reformulation (at the end of the daily update call in animal manager, in case more than one animal is added to that pen on that same day). Then reset the flag after we reformulate it using the new helper method.

## Steps to Reproduce

1. Start a simulation with the number of calves and heifers set to 0. Prior to merging #1733 this will cause an error in AnimalModuleReporter. Once that is fixed, however, it causes further problems in the Manure module. The proposed fix here should alleviate that.

JoeWaddell:
### Context

Issue(s) closed by this pull request: closes #1734 

- [ ] Yes
- [x] No

### What

Added check against `is_populated` prior to reporting ration information for a given pen.

Added new method `reformulate_ration_for_single_pen` in `AnimalManager`.
Added new `Pen` property, `needs_ration_formulation`, which checks if there is no formulated ration but the pen is populated (flags for reformulation).

Deprecated AnimalManager's `calc_avg_growth`, now that we're doing it on a pen by pen basis.

### Why

Empty ration correctly supplied for an empty pen, however if an empty pen becomes populated prior to the next formulation interval, then errors cascade through the rest of the simulation (notably in the Manure module), because there is a populated pen with no ration supplied!

### Test plan

Behavior is identical to main, which is the intent! Also does not fail if pens start as empty (overriding metadata defaults), or if are empty at the time of a typical formulation interval.
Cleaned up the code, so should be easier to follow now.

Updated existing unit tests. Added unit tests for new methods.


RUFAS/routines/animal/animal_manager.py
```diff
@@ -422,20 +442,25 @@ def calc_nutrient_rqmts(self, feed: Feed, current_temperature: float) -> None:
...
             in cases where a ration cannot be formulated such that it meets animal requirements
 
+        Parameters
+        ----------
+        pen : Pen
+            Pen object.
+
         """
         for pen in self.all_pens:
             if pen.animal_combination.name == "LAC_COW" or pen.animal_combination.name == "CLOSE_UP":
```
Isgotkowitz: Does this need to be updated so that it is only resetting milk production in the pen passed into the method?

RUFAS/routines/animal/animal_manager.py
```diff
@@ -1536,17 +1563,30 @@ def daily_updates(self, feed: Feed, weather: Weather, time: Time) -> None:
...
+                        self.cows,
+                        feed,
+                        current_temperature,
+                    )
+                    self._calc_ration_at_interval(feed, pen)
+                    AnimalModuleReporter.report_ration_interval_data(pen, feed, self.simulation_day)
+                    pen.calc_avg_growth()
                     if pen.animal_combination.name == "LAC_COW":
                         for animal in list(pen.animals_in_pen.values()):
                             animal.update_milk_production_history(self.simulation_day)
```
Isgotkowitz: What about replacing the `for pen in self.all_pens: ...` with `[self.reformulate_ration_single_pen(pen, current_temperature, feed) for pen in self.all_pens]`. 

This would be cleaner, but much much more importantly it would deduplicate the logic for the ration of a single pen. Currently it happens in two places: this for loop  and `reformulate_ration_single_pen`.

RUFAS/routines/animal/animal_manager.py
```diff
@@ -1813,3 +1853,44 @@ def _record_culling_stats(self) -> None:
...
+
+        Parameters
+        ----------
+        current_temperature : float
+            Current temperature.
+        feed : Feed
+            Feed object used in ration formulation.
+        """
+        for pen in self.all_pens:
+            if not pen.ration and pen.is_populated:
```
Isgotkowitz: I should have mentioned this earlier when we were talking about it, but I think it would be good to wrap `not pen.ration and pen.is_populated` into an `@property` method in `Pen`. That way it could be accessed like a regular flag, but would always be correct.

RUFAS/routines/animal/animal_manager.py
```diff
@@ -1536,17 +1563,30 @@ def daily_updates(self, feed: Feed, weather: Weather, time: Time) -> None:
             self.record_pen_history()
 
             if self.end_ration_interval():
```
Isgotkowitz, line 1535: Depending on how much change you feel comfortable introducing in this PR, I think a lot of the logic in this method could be reduced down to
```python
if self.end_ration_interval():
    self.clear_pens()
    self.allocate_animals_to_pens()
[self.reformulate_ration_single_pen(pen) for pen in self.all_pens if pen.needs_ration_reformulation or self.end_ration_interval()]
```

JoeWaddell: Love it! testing it now, will push soon

RUFAS/routines/animal/animal_manager.py
```diff
@@ -1105,54 +1119,55 @@ def _calc_ration_at_interval(self, feed: Feed) -> None:
...
-                    if key == "status":
-                        ration_per_pen[key] = ration_per_animal[key]
-                    else:  # feeds and price
-                        ration_per_pen[key] = ration_per_animal[key] * num_animals
-
-                pen.ration = ration_per_pen
-                pen.ration_per_animal = ration_per_animal  # Important
+        if pen.is_populated:
+            pen.subset_class_feeds(feed)
+            pen_specific_feed_data = available_feeds.get_feed_data_from_feed_ids(pen.allocated_feeds)
```
Isgotkowitz:
```suggestion
        if not pen.is_populated:
            return
        pen.subset_class_feeds(feed)
        pen_specific_feed_data = available_feeds.get_feed_data_from_feed_ids(pen.allocated_feeds)
        ...
```
Not something significant, but this would reduce the amount of indentation in the method.

RUFAS/routines/animal/animal_manager.py
```diff
@@ -1813,3 +1824,28 @@ def _record_culling_stats(self) -> None:
...
+        pen : Pen
+            Pen that requires ration reformulation.
+        current_temperature : float
+            Current temperature.
+        """
+        calves = [animal for animal in pen.animals_in_pen.values() if type(animal).__name__ == "Calf"]
+        heiferIs = [animal for animal in pen.animals_in_pen.values() if type(animal).__name__ == "HeiferI"]
+        heiferIIs = [animal for animal in pen.animals_in_pen.values() if type(animal).__name__ == "HeiferII"]
+        heiferIIIs = [animal for animal in pen.animals_in_pen.values() if type(animal).__name__ == "HeiferIII"]
+        cows = [animal for animal in pen.animals_in_pen.values() if type(animal).__name__ == "Cow"]
```
Isgotkowitz, line 1676:
```suggestion
        calves = [animal for animal in pen.animals_in_pen.values() if type(animal) is Calf]
        heiferIs = [animal for animal in pen.animals_in_pen.values() if type(animal) is HeiferI]
        heiferIIs = [animal for animal in pen.animals_in_pen.values() if type(animal) is HeiferII]
        heiferIIIs = [animal for animal in pen.animals_in_pen.values() if type(animal) is HeiferIII]
        cows = [animal for animal in pen.animals_in_pen.values() if type(animal) is Cow]
```

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -350,8 +354,9 @@ def report_daily_ration(cls, animal_manager, available_feeds: Dict[str, Dict[str
         }
         for pen in animal_manager.all_pens:
             ration_per_animal = pen.ration_per_animal.copy()
-            del ration_per_animal["status"]
-            del ration_per_animal["objective"]
+            for non_numeric_key in ["status", "objective"]:
```
Isgotkowitz, line 358: Since this set of keys is being used in more than one place, it wouldn't be a bad idea to make it a module constant so it is used consistently throughout the module

tests/animal_module_tests/test_animal_manager.py
```diff
@@ -1925,30 +1924,27 @@ def test_reset_milk_production_reduction(pens_with_mock_animals) -> None:
...
             assert animal.milk_production_reduction == 100.1
 
     # call the function once on the list of pens
-    AnimalManager.reset_milk_production_reduction(pen_list)
-
-    # then assert that all animals in all pens are still 100.1
-    for pen in pen_list.all_pens:
+    for pen in mock_animalmanager.all_pens:
+        AnimalManager.reset_milk_production_reduction(mock_animalmanager, pen)
+        # then assert that all animals in all pens are still 100.1
```
Isgotkowitz:
```suggestion
```

tests/animal_module_tests/test_animal_manager.py
```diff
@@ -2683,3 +2679,65 @@ def test_calc_animal_space_shortage(num_animals: int, max_spaces: int, expected:
     mock_animal_manager.all_pens[0].max_stocking_density = 1
     actual = mock_animal_manager._calc_animal_space_shortage(num_animals, mock_animal_manager.all_pens)
     assert actual == expected
+
+
+def test_reformulate_ration_single_pen(mocker: MockerFixture) -> None:
+    mocker.patch("RUFAS.routines.animal.animal_manager.AnimalManager.__init__", return_value=None)
```
Isgotkowitz:
```suggestion
def test_reformulate_ration_single_pen(mocker: MockerFixture) -> None:
    """Tests reformulate_ration_single_pen in AnimalManager."""
    mocker.patch("RUFAS.routines.animal.animal_manager.AnimalManager.__init__", return_value=None)
```

RUFAS/routines/animal/pen.py
```diff
@@ -325,6 +325,21 @@ def is_populated(self) -> bool:
...
+        """
+        Returns whether pen needs a ration formulated.
+
+        This is currently written to cover the case in which a ration was not formulated due to the pen being empty,
+         and on subsequent days is populated.
+
+        Returns
+        -------
+        bool
+            True if pen needs ration formulation.
```
Isgotkowitz:
```suggestion
        Returns whether pen needs a ration formulated. This covers the case in which a ration was
        not formulated due to the pen being empty, but was populated in subsequent days.
```
I think the returns section isn't necessary, the method name is really clear and the first line of the docstring covers it well.

RUFAS/routines/animal/animal_manager.py
```diff
@@ -1093,7 +1107,7 @@ def _update_phosphorus_concentrations(self) -> None:
             animals = self.animals_by_type[animal_type]
             self.phosphorus_concentration_by_animal_class[animal_type] = self._calc_phosphorus_concentration(animals)
 
-    def _calc_ration_at_interval(self, feed: Feed) -> None:
+    def _calc_and_update_ration(self, feed: Feed, pen: Pen) -> None:
```
Isgotkowitz: Maybe this function could be named something like `handle_pen_ration`? The "and" in `calc_and_update_ration` makes it sound like it has a dual purpose.

RUFAS/routines/animal/animal_manager.py
```diff
@@ -422,25 +442,29 @@ def calc_nutrient_rqmts(self, feed: Feed, current_temperature: float) -> None:
...
             latest_pen = cow.pen_history[-1].pen
             cow.set_nutrient_rqmts(
                 self.ANIMAL_GROUPING_SCENARIO,
                 nutrient_conc=self.all_pens[latest_pen].ration_nutrient_conc,
             )
 
-    def reset_milk_production_reduction(self) -> None:
+    def reset_milk_production_reduction(self, pen: Pen) -> None:
         """
         Resets reduction value for milk production to 0.0 for all animals in all pens
```
Isgotkowitz:
```suggestion
        Resets reduction value for milk production to 0.0 for all animals in the given pen.
```

RUFAS/routines/animal/animal_manager.py
```diff
@@ -422,25 +442,29 @@ def calc_nutrient_rqmts(self, feed: Feed, current_temperature: float) -> None:
                 previous_DMI=self.all_pens[latest_pen].dry_matter_intake,
             )
```
Isgotkowitz, line 424: Since all the heifers have the same routine, what would you think of collapsing their routines into something like this:
```python
for heifer in heiferIs + heiferIIs + heiferIIIs:
    latest_pen = heifer.pen_history[-1].pen
    heifer.set_nutrient_rqmts(
        current_temperature,
        self.ANIMAL_GROUPING_SCENARIO,
        nutrient_conc=self.all_pens[latest_pen].ration_nutrient_conc,
        metabolizable_energy=self.all_pens[latest_pen].MEdiet,
        previous_DMI=self.all_pens[latest_pen].dry_matter_intake,
    )
```

RUFAS/routines/animal/animal_manager.py
```diff
@@ -377,22 +378,41 @@ def avg_pen_dist(self) -> Tuple[float, float]:
...
         )
 
-    def calc_nutrient_rqmts(self, feed: Feed, current_temperature: float) -> None:
+    def calc_nutrient_rqmts(
+        self,
+        calves: List[Calf],
+        heiferIs: List[HeiferI],
+        heiferIIs: List[HeiferII],
+        heiferIIIs: List[HeiferIII],
+        cows: List[Cow],
```
Isgotkowitz, line 388: Maybe it would be better to have this method take in a `Pen` instead of a bunch of separate lists? Looking at the code, this method is currently only called in `reformulate_ration_single_pen`, meaning this method is essentially called on a pen by pen basis. I think this would make the method signature and `reformulate_ration_single_pen` a little cleaner, but would limit how this code could be used in the future.

Isgotkowitz (approved):
LGTM, nice job Joe. Left a bunch of comments, mostly on minor things (docstrings, names, etc.) - they are take em or leave em. There are few other suggestions for changes to the code, but it is up to you whether you feel they are appropriate for this PR.

RUFAS/routines/animal/animal_manager.py
JoeWaddell: My thought here was this the flexibility in this method could be useful if we want to reformulate on other subsets of animals (e.g. multiple pens at once, or just one of the types), but I'm open to changing it to reduce the number of arguments, too.

RUFAS/routines/animal/animal_manager.py
JoeWaddell: I like this, but ran into problems mocking against a direct type check - and this will only be a temporary measure once all animals are a single type, and we'll have to check against an attribute instead.

RUFAS/routines/animal/animal_manager.py
Isgotkowitz: No I think that is a totally valid reason to keep it as is.

tomhuhh:
To follow up on this issue, I performed a test last week and encountered an error, the detail of which has been shared with Joe.  

The test involves setting the number of calves to 0 in the default animal file and adjust the user input of true from false to true. 

Between Joe and myself, we concluded that a fix needs to be applied to align the rows between different columns in the output csv file. When there is no animals, the amount of ingredient fed should be 0 instead of being skipped.

JoeWaddell:
> Between Joe and myself, we concluded that a fix needs to be applied to align the rows between different columns in the output csv file. When there is no animals, the amount of ingredient fed should be 0 instead of being skipped.

This should be fixed in an upcoming PR: https://github.com/RuminantFarmSystems/MASM/pull/1778 , so we can consider it outside the scope of this PR for now!

tomhuhh (approved):
Okay, thanks Joe! Since the error happens during the output generation process, it makes sense to be handled in a separate PR. I'll give an approval for this PR. Thanks!

PR #1736, opened 2024-06-03
c2f15 inputs

Rebecavb:
c2f15 inputs


jadamchick (commented):
Hey Rebeca! The files look fine to merge, though I think we might want to extend the config to start a little earlier and include a couple pattern repeats, just to give it time to run a bit longer. I think we had said somewhere in the 2014 ballpark. Thoughts?

Rebecavb:
Hi @jadamchick I just updated the rotations for starting in 2014-2015-2016 with a pattern repeat of 2 so we can finish the rotations in 2022. Does this seem correct?

jadamchick (approved):
looks good, thanks Rebeca!

PR #1737, opened 2024-06-03
Disentangle the Animal and Manure modules

### LINKED ISSUE #1624: Disentangle the Animal and Manure modules
The connection between the Animal and Manure modules should be disentangled so that they meet the requirements outlined in #1622.

## Suggested Solution

An intermediate data structure should be defined which can package manure and manure processing information from a `Pen` and be passed to the Manure Manager. This novel data structure will be shared by the Manure and Animal modules, so it must not be defined in either of them. The information required to be in this data structure can be found in `ManureManagerPen` (the exact information contained can and probably will have to be tweaked).

There should be a method added to the `AnimalManager` which will provide a list of these data structures based on the current list of `Pens` that it contains. This list of data structures should be passed to the `ManureManager` when it is initialized in `SimulationEngine`, and `SimulationEngine` should pass this list of data structures into the daily routine of `ManureManager`. `ManureManagerPen` should be modified to work with these data structures.

Note that `ManureManager`'s daily routine is not called directly in `SimulationEngine`'s `_daily_simulation`, but through a classless wrapper function in `manure_manager.py`. This wrapper function should be removed and `ManureManager`'s daily routine should be called directly. Also, the `AnimalManager`'s `simulation_day` is passed to the `ManureManager` - this should be changed so that `Time`'s `simulation_day` is passed instead.

## Current Behavior

The pens of the Animal module are passed into the Manure module at initialization of the Manure module and on a daily basis to collect manure from the pen and initialize manure management scenarios based on the pen's specified manure management scenario.

Isgotkowitz:
This PR separates the Animal and Manure modules so that they only pass information to each other at the level of `SimulationEngine`, and removes all statements that import modules from the Animal module into the Manure module and vice versa.

### Context

Issue(s) closed by this pull request: closes #1624 

- [ ] Yes
- [x] No

### What

The changes introduced consist of creating new data structures or moving existing data structures from the Manure and Animal modules into the new directory `shared_structures`, and using these structures to communicate information between the Manure and Animal modules in `SimulationEngine`.

This PR maintains exactly the same communications as before (with one notable exception, see below), while making minimal changes in the existing code. This means that there are many improvements which can and will be made to how the Animal and Manure modules communicate, but these will be made in subsequent PRs.

### Why

By disentangling the Animal and Manure modules, it is easier to look at exactly how they connect to one another and see exactly what information is passed between them. This will make it easier to identify bugs and problematic designs.

It also makes the implementations of the Animal and Manure modules less coupled, so it is easier to make changes in one without having to make changes in another.

### How

Previously, the `ManureManager` would accept a list of `Pen` objects from the `AnimalManager` and extract manure information from each pen. It would do this via a classless wrapper method called `simulate_daily_manure_manager`, which has been removed. Now, the `SimulationEngine` tells `AnimalManager` to extract the manure information from each `Pen` in a data structure called `PenManureData` (implemented in `shared_structures`), and the resulting list of `PenManureData`s are passed to the `ManureManager`. This happens on a daily basis, and during the initialization of a simulation.

The `AnimalCombinations` enum has been moved from the Animal module to `shared_structures`. The `AnimalManureExcretions` typed dict has also been removed from `general_manure.py` in the Animal module and reimplemented, exactly the same as before, in `shared_structures`. Many of the files changed are simply changing the import statements for these structures.

### Important Bug Fix
There is a notable difference in manure outputs related to lactating cows. This is because there was a bug in the way that lactating cows were counted (see my comment in `manure_manager_pen.py` for details). The result is that the following outputs have changed for the default scenario outputs:
- AnaerobicLagoon_emptying_amount_pen_3.liquid_manure_total_ammoniacal_nitrogen
- AnaerobicLagoon_emptying_amount_pen_3.liquid_manure_nitrogen
- AnaerobicLagoon_emptying_amount_pen_3.daily_final_manure_volume
- AnaerobicLagoon_emptying_amount_pen_3.liquid_manure_daily_volume
- AnaerobicLagoon_emptying_amount_pen_3.storage_ammonia
- AnaerobicLagoon_emptying_amount_pen_3.storage_nitrous_oxide
- AnaerobicDigestionAndLagoon_emptying_amount_pen_3.liquid_manure_nitrogen
- AnaerobicDigestionAndLagoon_emptying_amount_pen_3.daily_final_manure_volume
- AnaerobicDigestionAndLagoon_emptying_amount_pen_3.liquid_manure_daily_volume
- AnaerobicDigestionAndLagoon_emptying_amount_pen_3.storage_ammonia
- AnaerobicDigestionAndLagoon_emptying_amount_pen_3.storage_nitrous_oxide

### Test plan

Unit tests have been written or updated to verify changes. Additionally, the following filter has been used to verify that final results are unchanged between this branch and `main` (note that this filters compares summed outputs, so it is not precise as it could be):
```json
{
    "multiple": [
        {
            "name": "Animal to Manure Module",
            "filters": ["AnimalModuleInputToManureModule.*"],
            "variables": [".*"],
            "vertical_aggregation": "sum"
        },
        {
            "name": "Pen 1 Related Outputs",
            "filters": [".*pen_1.*"],
            "variables": [".*"],
            "vertical_aggregation": "sum"
        },
        {
            "name": "Pen 2 Related Outputs",
            "filters": [".*pen_2.*"],
            "variables": [".*"],
            "vertical_aggregation": "sum"
        },
        {
            "name": "Pen 3 Related Outputs",
            "filters": [".*pen_3.*"],
            "variables": [".*"],
            "vertical_aggregation": "sum"
        },
        {
            "name": "Pen 0 Related Outputs",
            "filters": [".*pen_0.*"],
            "variables": [".*"],
            "vertical_aggregation": "sum"
        }
    ]
}
```
The results are identical, except for the outputs mentioned in the section above. To verify that these differences are only the result of the miscounting of lactating cows, remove `.values()` from line 919 of `pen.py` in the feature branch, rerun the default simulation, and compare the outputs to those from main.

### Edit
After speaking with @PooyaHekmati, I have made some slight adjustments to how the shared types are organized. Instead of having everything in a `shared_structures` directory, the data structures are placed in a directory named `data_structures`. The enum that is shared is placed in a file called `enums.py` that is at the same level as `SimulationEngine`. As we move forward with the disentanglement, it may be the case that a lot of enums are extracted and moved to `enums.py`. After a certain point, if there are lot of enums in `enums.py`, we can break off each into its own file and group them in a directory.


tests/animal_module_tests/test_pen.py
```diff
@@ -776,3 +777,61 @@ def test_calc_total_manure(
...
+    "animal_count,combination,expected",
+    [(3, AnimalCombination.LAC_COW, 3), (10, AnimalCombination.CALF, 0), (8, AnimalCombination.GROWING, 0)],
+)
+def test_count_lactating_cows(
+    mocker: MockerFixture, pen: Pen, animal_count: int, combination: AnimalCombination, expected: int
+) -> None:
+    """Tests _count_lactating_cows in Pen."""
+    mocker.patch.object(Cow, "__init__", return_value=None)
+    animals = {key: Cow() for key in range(animal_count)}
+    pen.animals_in_pen = animals
```
Isgotkowitz, line 792: @JoeWaddell this is how you can get around the typing of mocked object instances. You patch the init method of the desired type to be mocked, then you can instantiate them with no restrictions and they will be effectively mocked.

JoeWaddell: Thanks! Running into other unexpected errors, perhaps due to the nature of how the `Cow` class currently inherits from others

RUFAS/routines/manure/manure_manager.py
Isgotkowitz, line 1: I fixed a few Mypy errors in this file, in an attempt to make the net change to the error count negative for this PR.

RUFAS/routines/manure/pen_manure/manure_manager_pen.py
```diff
@@ -49,60 +47,35 @@ class ManureManagerPen:
...
-        -------
-        int
-            The number of lactating cows in the pen.
-
-        """
-        num_lac_cows = 0
-        if animal_combination is AnimalCombination.LAC_COW:
-            for animal in animals_in_pen:
-                if type(animal) is Cow:
-                    num_lac_cows += 1
```
Isgotkowitz, line 104: The bug in this method is right here. `animals_in_pen` was a dictionary, not a list. Iterating over a dictionary apparently just iterates over the keys of a dictionary, so instead of `animal` being an `AnimalBase` object it was the `int` id of the animal. Thus the type of `animal` was never `Cow` and the count of lactating cows was always 0.

This bug has been fixed in the reimplementation of this method in `pen.py`, but fixing this error is the cause of change for manure outputs in the default scenario.

tomhuhh:
Thanks Ed! In the slurry storage scenarios, I found the following output variables for lactating cows to be different between main and this branch. 

Pen 3 Related Outputs_SlurryStorageOutdoor_emptying_amount_pen_3.liquid_manure_total_ammoniacal_nitrogen_ver_agg

Pen 3 Related Outputs_SlurryStorageOutdoor_emptying_amount_pen_3.liquid_manure_nitrogen_ver_agg

Pen 3 Related Outputs_SlurryStorageOutdoor_emptying_amount_pen_3.storage_ammonia_ver_agg

Pen 3 Related Outputs_SlurryStorageOutdoor_emptying_amount_pen_3.storage_nitrous_oxide_ver_agg

This agrees with the findings from the anaerobic digestion and lagoon scenarios. 

I am also curious why it only affects `emptying amount`. My next step is to explore different filters and see if other output variables are impacted by this bug. That said, I believe the overall direction of this PR is promising. 

Thanks!
Haowen

tomhuhh (approved):
I believe the disentangling part of the PR is working given the overlaps in the values of the output variables between main and this branch. 

As I mentioned above, I plan to test different filters regarding the bug associated with lactating cows. If something else comes up during this process, I think a separate issue and PR can solve the problem. 

I'll leave the assessment of the code efficiency to SWE experts.  

Overall, nice job, Ed!

RUFAS/routines/animal/pen.py
```diff
@@ -909,3 +910,29 @@ def remove_animal(self, animal_id: int) -> None:
         """
         del self.animals_in_pen[animal_id]
         self.ration = self._calc_new_ration(len(self.animals_in_pen))
+
+    def _count_lactating_cows(self) -> int:
+        """Returns the count of lactating cows currently held in this pen."""
+        if self.animal_combination is not AnimalCombination.LAC_COW:
+            return 0
+        num_lac_cows = len([animal for animal in self.animals_in_pen.values() if type(animal) is Cow])
```
ew3361zh, line 918: Just for clarification - it looks like the definition of the LAC_COW AnimalCombination is 

```
LAC_COW: int
    Represents a pen with lactating cows only.
```

So do we need the list comprehension? (i.e. are there circumstances where the above definition is not true?)

RUFAS/simulation_engine.py
```diff
@@ -102,9 +102,8 @@ def _run_simulation_main_loop(self) -> None:
     def _daily_simulation(self) -> None:
         """Executes the daily simulation routines."""
         self.animal_manager.daily_updates(self.feed, self.weather, self.time)
-        simulate_daily_manure_manager(
-            self.manure_manager, self.animal_manager.all_pens, self.animal_manager.simulation_day
-        )
+        all_pen_manure_data = self.animal_manager.collect_pen_manure_data()
+        self.manure_manager.daily_update(all_pen_manure_data, self.animal_manager.simulation_day)
```
ew3361zh, line 106: Do you still want to shift this to be the Time `simulation_day` (as is mentioned in the issue)?

```
Also, the AnimalManager's simulation_day is passed to the ManureManager - this should be changed so that Time's simulation_day is passed instead.
```

ew3361zh (approved):
This looks great, Ed! I ran it and tried comparing results - nothing jumped out to me as vastly different between main and this branch.

I was initially going to say a diagram could help make clear all the changes (vs. the PR description) but after I spent some time with it I think the changes are clear on their own with the current description.

I left a couple comments/questions but nothing that would prevent me from approving the PR. Nice work!

RUFAS/simulation_engine.py
Isgotkowitz: I actually tried that - and it really threw the manure outputs for a loop. So I decided that it would be better to address that bug in its own PR, because it is probably going to require a good amount changes just to fix it.

RUFAS/routines/animal/pen.py
Isgotkowitz: > are there circumstances where the above definition is not true?

That's a great question, but I'm not familiar enough with the animal module to know if that is the case. For this PR though, I think it would be better to try and preserve as much of the original behavior as possible (though fixing the obvious bug that existed in this method before).

Tagging @JoeWaddell to see if you think the logic of this method should be changed.

PR #1738, opened 2024-06-04
Manure Nitrogen Sensitivity Analysis

tomhuhh:
### Context

- [ ] Yes
- [ ] No

### DIFF
--- a/RUFAS/routines/manure/constants_and_units/manure_constants.py
+++ b/RUFAS/routines/manure/constants_and_units/manure_constants.py
@@ -1,3 +1,5 @@
+from typing import NamedTuple
+
 class ManureConstants:
     """
     A class to store constants for manure management.
@@ -71,3 +73,16 @@ class ManureConstants:
     """
     The fraction of total solids in Liquid manure. Temporary placeholder until upstream solids tracking is fixed.
     """
+    
+    BarnArea = NamedTuple("BarnArea", [("has_cows", float), ("no_cows", float)])
+    freestall = BarnArea(has_cows=3.5, no_cows=2.5)
+    tiestall = BarnArea(has_cows=1.2, no_cows=1.0)
+    bedded_pack = BarnArea(has_cows=5.0, no_cows=3.0)
+    open_lot = BarnArea(has_cows=5.0, no_cows=3.0)
+
+    barn_area_by_pen_type = {
+        "freestall": freestall,
+        "tiestall": tiestall,
+        "compost bedded pack barn": bedded_pack,
+        "open lot": open_lot,
+    }
--- a/RUFAS/routines/manure/constants_and_units/manure_constants.py
+++ b/RUFAS/routines/manure/constants_and_units/manure_constants.py
@@ -1,5 +1,6 @@
 from typing import NamedTuple
 
+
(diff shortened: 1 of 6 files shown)


RUFAS/routines/manure/pen_manure/manure_manager_pen.py
```diff
@@ -105,7 +106,7 @@ def count_lactating_cows(cls, animal_combination: AnimalCombination, animals_in_
         return num_lac_cows
 
     @property
-    def barn_area_from_pen_type(self) -> float:
+    def barn_area_from_pen_type(self, barn_area_by_pen_type) -> float:
```
Isgotkowitz: The method signature doesn't need to be changed, you can reference the constants directly from the constants module within the method body.

RUFAS/routines/manure/manure_manager.py
```diff
@@ -116,6 +119,9 @@ def __init__(self, pen_list: List[Pen], weather: Weather, time: Time, manure_man
         self._daily_output_per_pen = []
         self._manure_nutrient_manager = ManureNutrientManager()
         self.configure_manure_manager_components(pen_list)
+        self.barn_area = ()
+
+    barn_area = im.get_data("manure_management.barn_area_configs")
```
Isgotkowitz: The barn area constants will already be in the `manure_manager_config` argument, so you can pull them directly out of there and not call IM.

RUFAS/routines/manure/manure_manager.py
```diff
@@ -116,6 +119,9 @@ def __init__(self, pen_list: List[Pen], weather: Weather, time: Time, manure_man
         self._daily_output_per_pen = []
         self._manure_nutrient_manager = ManureNutrientManager()
         self.configure_manure_manager_components(pen_list)
+        self.barn_area = ()
```
Isgotkowitz: I think the first thing that should be done in this method is to configure the barn area constants - otherwise there might somehow be a call to `barn_area_by_pen_type` before the barn areas have been set.

RUFAS/routines/manure/constants_and_units/manure_constants.py
```diff
@@ -71,3 +74,16 @@ class ManureConstants:
...
+    freestall = BarnArea(has_cows=3.5, no_cows=2.5)
+    tiestall = BarnArea(has_cows=1.2, no_cows=1.0)
+    bedded_pack = BarnArea(has_cows=5.0, no_cows=3.0)
+    open_lot = BarnArea(has_cows=5.0, no_cows=3.0)
+
+    barn_area_by_pen_type = {
+        "freestall": freestall,
+        "tiestall": tiestall,
+        "compost bedded pack barn": bedded_pack,
+        "open lot": open_lot,
```
Isgotkowitz, line 87: I would recommend setting these to `None` - because if for whatever reason they aren't set correctly, they will cause a crash that lets you know they weren't set correctly.

tomhuhh:
This was initially created for the sensitivity analysis poster project, but I believe it is no longer relevant, so I am closing it. However, the code that moves the constant barn area to the input list might still be useful for future reference.

PR #1739, opened 2024-06-04
Implement remaining `Hay` logic

### LINKED ISSUE #1377: [Feed] Populate logic in `RUFAS/routines/feed_storage/hay.py`
Populate the logics of the following methods in `RUFAS/routines/feed_storage/hay.py`:

- [ ] `bale_density`
- [ ] `bale_size`
- [ ] `calculate_protein_loss`

[Design doc is here](https://docs.google.com/document/d/1C86lch2dkYBERR-qLDmE7L7n1gYQjETrIKUyrQpWrPs/edit)

* For questions, reach out to SME: @KevinP-B and/or SWE @PooyaHekmati . Their reviews are needed for any linked PRs.
* Size the PRs reasonably, not all the methods have to be implemented in a single PR.
* docstrings, function signatures, and unit tests need to be updated accordingly.
* I suggest reading and understanding the design doc before attempting to write the code.

Isgotkowitz:
Completes the implementation of Hay degradation logic.

### Context

Issue(s) closed by this pull request: closes #1377

- [ ] Yes
- [x] No

### What

This PR implements the final pieces of logic needed for `Hay` and its children to process all dry matter and nutrient loss that occurs.

It also adds some needed constants to the general constants file.

### Why

Hay loses dry matter in a manner that is very different from silage and baleage, so it requires unique implementations for the subroutines that actually calculate dry matter loss.

### How

The routine `_calculate_additional_dry_matter_loss` has been implemented, and incorporated into `calculate_dry_matter_loss_to_gas` in `Hay`.

It also adds hay-type specific attributes for dry matter and nutritive loss in the `ProtectedWrapped`, `ProtectedTarped`, and `Unprotected` child classes.

### Test plan

Unit tests have been added to make sure that `Hay` children are init'ed correctly and that `_calculate_additional_dry_matter_loss` gives correct outputs over variety of weather conditions.


RUFAS/routines/feed_storage/hay.py
```diff
@@ -152,6 +163,41 @@ def _calculate_subsequent_dry_matter_loss_to_gas(self, crop: HarvestedCrop, time
...
+        float
+            Loss of dry matter that occurred over the specified period of weather conditions in kg.
+
+        References
+        ----------
+        .. [1] Feed Storage Scienitific Documentation, equation 1.2.8
+
+        """
+        if self.additional_dry_matter_loss_coefficient == 0.0:
+            return 0.0
```
allisterakun, line 193: If my understanding is correct, this would correspond to the protected indoor hay storage. Can we maybe mention this in the docstring? Just to explain why we're having this check.

allisterakun (approved):
Good job, Ed! It looks great! A simple comment about adding to the docstring.

Also, unrelated to this PR: 
If we're already checking for `days_store == 0` in lines 83-84, is rechecking the same condition in lines 124-125 still necessary?

RUFAS/routines/feed_storage/hay.py
Isgotkowitz: Yes, I will make a note of this in the docstring. I think it would be best not to explicitly tie it to protected indoor storage though, because that type might not always have zero additional loss - or it might not be the only type that has 0 additional loss. It is more a function of this equation.

KevinP-B (approved):
It looks like this addresses the logic we discussed. Looks like it matches the scientific documentation nicely. Thanks, Ed!

Isgotkowitz:
> Good job, Ed! It looks great! A simple comment about adding to the docstring.
> 
> Also, unrelated to this PR: If we're already checking for `days_store == 0` in lines 83-84, is rechecking the same condition in lines 124-125 still necessary?

Excellent question, it is not necessary. It makes it more resilient in case something changed in the code and it somehow reached that method on the zeroth day of storage, but that would be a big issue if that was happening so I will take it out.

PR #1742, opened 2024-06-04
Include var units in graph when requested

### LINKED ISSUE #1515: [OM] Include units in graphs when specified
This is part 6 of the issue #1416 

Part 6 
- [ ] Include units in graphs when specified by the user.
- [ ] To make the formatting of units more consistent, consider reusing the code for formatting units.

### LINKED ISSUE #1702: Graph legend for dictionaries fails to auto populate
Previous behavior as documented in wiki
image

Current behavior with same variables info (and no legend supplied)
image

Behavior while testing #1742 with a slightly different filter (for only the numeric feed keys): 
image

This was using the following filter:
```
{
            "type": "plot",
            "filters": [
                "AnimalModuleReporter.report_daily_ration.ration_daily_feed_totals_for_pen_3_LAC_COW"
            ],
            "title": "ration fed",
            "variables": ["\\d*"],
            "omit_legend_prefix": true,
            "display_units":true
        }
```

## Expected Behavior

When using regex for the variables and legends (as in the example from the Wiki), it should correctly auto-populate the legend for each variable.

## Current Behavior

However, it currently prints a single value in the legend, and it's the regex entry itself.

## Steps to Reproduce

1. 
```{
            "type": "plot",
            "filters": [
                "AnimalModuleReporter.report_ration_interval_data.ration_per_animal_for_pen_3_LAC_COW"
            ],
            "title": "ration per_animal_for_pen_3_LAC_COW",
            "variables": [
                ".*"
            ]
        }
```

ew3361zh:
This provides an option to add the units for a variable when graphing it.

### Context

Issue(s) closed by this pull request: closes #1515 closes #1702

- [ ] Yes
- [x] No

### What

When users would like to have the units for a variable or variables they're graphing, they can now add `"display_units": true` to their graph filters. If the user has not specified this value in their graph filter, it will default to `True` (i.e. it will add units to the variables graphed).

This will add the variable units from the associated info_map that was passed to GG with the data to be graphed.

Important to note: 
1. If the user selects to exclude info maps for a particular task for which they want to graph outputs, this function will not be able to work and will log a warning notifying the user of the inability to add the units because the info_maps were excluded but the graphs themselves will still be produced (just without units on the vars).

Logging an error at this point in the graphing process would cause the graphing process to halt which seems less desireable since we're defaulting `display_units` to `True`.

2. This does not currently work with graphing data from reports. There is a separate issue for adding units to report data when requested #1516 and it seemed like handling units for report data sent to be graphed would be better dealt with there.

### Why

This will be helpful in knowing the units of the data being graphed.

### How

A new function was added to GG: `_add_var_units()`. This function gets called if the user has added `"display_units": true` to their graph filter.

The function iterates through the filtered pool data sent to GG and places the units of the variable at the end of the variable name. For example, if a variable is "Animal_milk_produced" and the units are "gal/day", the function will convert this to "Animal_milk_produced (gal/day)".

One special case accounted for is if the user selected "omit_legend_suffix" in their graph filter which chops off the end of a variable name. This was initially removing the units from the end as well so I added some logic to get the units from the name and readd them back to the variable name after the suffix has been removed.

The logic for building the graph legend (where units will be added) was extracted to its own function as it was causing the `generate_graph()` function to violate the complexity flake8 guideline.

### Test plan

Try a graph filter with `"display_units": true` and look at the resulting graph.
Try a graph filter with both `"display_units": true` and `"omit_legend_suffix": true` and see the units still present in the resulting graph.
Try a graph filter with `"display_units": true` and use the `-i -v errors` gnu args to see the error raised when info maps are excluded and the user attempts to graph the data with units are desired to be included.

Example filter including a C&S var which has a suffix that will get removed:
```
{
  "multiple": [
    {
      "type": "plot",
      "filters": [
          ".*\\.snow_content.field='.*'"
      ],
      "title": "Snow Content",
      "display_units": true,
      "omit_legend_prefix": true,
      "omit_legend_suffix": true
    },
    {
      "type": "plot",
      "filters": [
        "^ManureTreatmentDailyOutput_Pen_0_CALF.daily_final_manure_volume",
        "Weather.precipitation"
      ],
      "title": "Precipitation / Anaergobic Lagoon Daily Output Volume",
      "display_units": true
    }
  ]
}
```


RUFAS/graph_generator.py
```diff
@@ -167,9 +167,13 @@ def generate_graph(
             return all_logs
         try:
             graph_filter_validation_logs = self._validate_graph_filter(graph_details)
-            prepared_data: Dict[str, List[Any]] = {key: filtered_pool[key]["values"] for key in filtered_pool.keys()}
-            non_numeric_data_logs = self._log_non_numerical_data(filtered_pool, graph_details)
-            all_logs = non_numeric_data_logs + graph_filter_validation_logs
+            var_units_logs: list[dict[str, str | dict[str, str]]] = []
+            updated_pool = filtered_pool
+            if graph_details.get("display_units", False):
+                updated_pool, var_units_logs = self._add_var_units(filtered_pool)
```
Isgotkowitz: If might be better to default to true?

RUFAS/graph_generator.py
```diff
@@ -214,6 +218,42 @@ def generate_graph(
...
+                updated_data[new_var_name] = details
+            else:
+                all_logs: List[Dict[str, Collection[str]]] = [
+                    {
+                        "error": "Can't add units to variables for graphing",
+                        "message": "'info_maps' unavailable to get units, check setting for exclude_info_maps.",
+                        "info_map": info_map,
+                    }
+                ]
+                return filtered_pool, all_logs
```
Isgotkowitz: For the conditional logic in this, would it be fair to assume that if one variable in the filtered pool has info maps then they all do? If so then the check for info maps being present could be placed before the for loop

RUFAS/graph_generator.py
```diff
@@ -214,6 +218,42 @@ def generate_graph(
...
+        """
+        updated_data = {}
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self._add_var_units.__name__,
+        }
+        for var_name, details in filtered_pool.items():
+            if "info_maps" in details and details["info_maps"]:
+                unit = details["info_maps"][0]["units"]
+                new_var_name = f"{var_name} ('{unit}')"
```
```suggestion
                new_var_name = f"{var_name} ({unit})"
```
I think the quotes aren't really necessary here.

RUFAS/graph_generator.py
```diff
@@ -214,6 +218,42 @@ def generate_graph(
...
+            The updated data with units added and logs if info_maps aren't found to get units.
+        """
+        updated_data = {}
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self._add_var_units.__name__,
+        }
+        for var_name, details in filtered_pool.items():
+            if "info_maps" in details and details["info_maps"]:
+                unit = details["info_maps"][0]["units"]
```
Isgotkowitz: Not something that can or should be addressed in this PR but is still worth pointing out: it is possible that somewhere upstream, a value for a variable is added to the Output Manager with different units than other values for that variable, in which case just taking the first unit would be problematic.

RUFAS/graph_generator.py
```diff
@@ -214,6 +218,42 @@ def generate_graph(
...
+                new_var_name = f"{var_name} ('{unit}')"
+                updated_data[new_var_name] = details
+            else:
+                all_logs: List[Dict[str, Collection[str]]] = [
+                    {
+                        "error": "Can't add units to variables for graphing",
+                        "message": "'info_maps' unavailable to get units, check setting for exclude_info_maps.",
+                        "info_map": info_map,
```
Isgotkowitz: I think it would be beneficial to pass in the title of the graph and add that in here, just to make the error message a little more specific.

Isgotkowitz (approved):
This adds units, with a few things to note:
- If "legend" is included, units will not be included in the legend.
- If "variables" is included, units will not be included in the legend.
- If "omit_legend_suffix" is false for variables with a suffix (i.e. variables from Crop and Soil), units will be added twice.

For "legend" and "variables" being present, I don't know if those are required cases for units to be added, so I will approve.

This solution for adding units back in to the legend relies in part on string parsing, and there are other parts of GG that rely on string parsing as well. I think it would be worthwhile to revisit the current setup of GG (and info maps) to see if we could remove that dependence on string parsing.

> If it would be preferable to only add a warning and allow the graphing to continue without adding the units, that would not be difficult to implement.

I think it would be good to leave it as an error, and allow the graphing to continue anyway - but that is just my two cents.

RUFAS/graph_generator.py
ew3361zh: I had the same thought but maybe will get some SME input on this.

JoeWaddell: The general rule I apply to figures is that units are mandatory for fully understanding them: so I'd agree that we should default to true!

I could see cases where the axis names are self-evident (including the unitless types), but I imagine that's a less common occurrence.

RUFAS/graph_generator.py
ew3361zh: That would work! It should check for the negative case first anyhow. Good suggestion.

RUFAS/graph_generator.py
ew3361zh: Can you expand on this a little? Are you saying units would be added incorrectly somewhere along the line? Maybe a check in OM to ensure all units are the same within an info map?

This makes me think we'd also potentially run into this issue with the logic recently added to OM._add_to_pool() with regards to handling the `first_info_map_only` flag. Probably best dealt with there.

RUFAS/graph_generator.py
ew3361zh: Done!

ew3361zh:
> I think it would be good to leave it as an error, and allow the graphing to continue anyway - but that is just my two cents.

@Isgotkowitz unfortunately I think this is an either/or situation. If we leave this as an error there's a check before we reach `_draw_graph()` that would return to OM without graphing anything. If it's logged as a warning though, it wouldn't trigger that (and there are several other possible errors being raised within this block so it would have to be a wonky exception to the current check which I don't think would look very good codewise.

I think I agree with you though that we should allow graphing to continue, especially since we're defaulting "display_units" to True.

JoeWaddell:
Per discussion in person with @ew3361zh , including units even if `variables` and/or `legend` is supplied would be preferred: that way a user could easily grab info from e.g. a dictionary without having to manually enter the units in the legend. That may be beyond the scope of this PR, but IMO would be the ideal behavior for `display_units`.

RUFAS/graph_generator.py
Isgotkowitz: > Are you saying units would be added incorrectly somewhere along the line? 

Not necessarily incorrectly, but inconsistently. I.e. kilograms on one day, pounds on the next.

> Maybe a check in OM to ensure all units are the same within an info map?

Yes, this will probably be the place where this is checked for eventually.

> This makes me think we'd also potentially run into this issue with the logic recently added to OM._add_to_pool() with regards to handling the first_info_map_only flag. Probably best dealt with there.

I agree. We might be able to make some more assumptions and do more of the info map deduplication automatically.

Isgotkowitz:
> > I think it would be good to leave it as an error, and allow the graphing to continue anyway - but that is just my two cents.
> 
> @Isgotkowitz unfortunately I think this is an either/or situation. If we leave this as an error there's a check before we reach `_draw_graph()` that would return to OM without graphing anything. If it's logged as a warning though, it wouldn't trigger that (and there are several other possible errors being raised within this block so it would have to be a wonky exception to the current check which I don't think would look very good codewise.
> 
> I think I agree with you though that we should allow graphing to continue, especially since we're defaulting "display_units" to True.

Oh dang, that stinks. Logging a warning and continuing sounds like a good solution then.

ew3361zh:
Updates:
- When configuring legend in the `_set_graph_legend()` method, the presence of `omit_legend_prefix` and `omit_legend_suffix` are checked for first. Next will set the legend with the variable names from the `variables` entry in the filter file. Finally if neither of those is pressent, the keys from the prepared data will be used to populate the legend. (previously 1 and 2 were switched).
- I added a line-break at the point of the "." in all variable names. It seemed like some were still too long even with the prefix/suffix removed. If this is not desireable or is confusing to look at, it's easy to remove.
- For the variables where the units are added to the info map as a dictionary, logic is added to the `_add_var_units()` method to extract those units and add them to the appropriate variable.
- Logic is added to the `_add_var_units()` method to handle if the user has supplied a regex filter to the "variables" entry in the graph filter file.
- A slight update to the error message in OM if the GG sends back a bad filter directing the user to look at the Wiki for how to properly set up a graph filter file.
- The default setting for "display_units" is True.
- Unit testing updated.

Updated test cases:
- This graph filter should work to generate a graph of all the ration elements with the appropriate units:
```
{
    "multiple": [
        {
            "type": "plot",
            "filters": [
                "AnimalModuleReporter.report_ration_interval_data.ration_per_animal_for_pen_3_LAC_COW"
            ],
            "title": "ration per_animal_for_pen_3_LAC_COW",
            "variables": [
                ".*"
            ]
        }
    ]
  }
```
- Try the original test case filter with `omit_legend_prefix`/`omit_legend_suffix` set to both true and false. In particular check the `false` setting to get a look at how the graph legends look with the line-breaks.


MyPy count is being worked on.

RUFAS/util.py
```diff
@@ -386,7 +386,7 @@ def filter_dictionary(
         }
 
     @staticmethod
-    def remove_special_chars(input_string: str) -> str:
+    def remove_special_chars(input_string: str | list[str] | None) -> str:
```
ew3361zh, line 389: MyPy error fix (fun used in GG)

ew3361zh:
possibly also closes #1702 @JoeWaddell ?

RUFAS/graph_generator.py
```diff
@@ -247,8 +243,99 @@ def generate_graph(
...
+                                "message": f"Unit for '{var_name}' not found in units dictionary. "
+                                "Using default 'not available'.",
+                                "info_map": info_map,
+                            }
+                        )
+                else:
+                    unit = unit_info
+
+                new_var_name = f"{var_name} ({unit})"
+                updated_data[new_var_name] = details
```
```suggestion
        for var_name, details in filtered_pool.items():
            unit_info = details["info_maps"][0]["units"]
            if isinstance(unit_info, dict):
                unit = unit_info.get(var_name, "not available")
                if unit == "not available":
                    logs.append(
                        {
                            "warning": "Missing unit information",
                            "message": f"Unit for '{var_name}' not found in units dictionary. "
                            "Using default 'not available'.",
                            "info_map": info_map,
                        }
                    )
            else:
                unit = unit_info

            new_var_name = f"{var_name} ({unit})"
            updated_data[new_var_name] = details
```

RUFAS/graph_generator.py
```diff
@@ -247,8 +243,99 @@ def generate_graph(
...
+        if omit_legend_prefix or omit_legend_suffix:
+            graph_details["legend"] = list(
+                self._generate_legend_keys(key, omit_legend_prefix, omit_legend_suffix) for key in prepared_data.keys()
+            )
+        elif selected_variables := graph_details.get("variables"):
+            graph_details["legend"] = selected_variables
+        else:
+            graph_details["legend"] = list(prepared_data.keys())
+
+        graph_details["legend"] = [s.replace(".", ".\n") for s in graph_details["legend"]]
```
Isgotkowitz: > I added a line-break at the point of the "." in all variable names. It seemed like some were still too long even with the prefix/suffix removed. If this is not desireable or is confusing to look at, it's easy to remove.

I think it does make it a little more confusing and less visually appealing. An alternative could be to add the argument `bbox_inches='tight'` to the `savefig` call in `_save_graph`. I don't fully understand all the nuances of what it does, but it resizes the graph so that legend is always fully in the plot, even when it is quite long.

Only downside is that it trims a lot of the whitespace from around the graph, but I think it is a worthwhile tradeoff.

RUFAS/util.py
Isgotkowitz: Fun is right.

Isgotkowitz (approved):
All graphs are working well, nice job Niko. Units aren't auto populated when `"legend"` key is provided, but I think that is reasonable - that key gives the user absolute control over the legend, so I think it is reasonable to make it their responsibility to know what they're graphing and what the units are.

Left a couple suggestions for improvements, one substantive one really minor.

> MyPy count is being worked on.

:applause:

ew3361zh:
> Units aren't auto populated when `"legend"` key is provided, but I think that is reasonable - that key gives the user absolute control over the legend, so I it is reasonable to make it their responsibility to know what they're graphing and what the units are.

In discussions with @JoeWaddell that was our thinking as well.

RUFAS/graph_generator.py
ew3361zh: Ooh, just tried the `bbox_inches="tight"` - I like it. I'll swap these out. Nice find.

JoeWaddell:
> possibly also closes #1702 @JoeWaddell ?

It does! Nice work on  this one - and confirming the discussion about behavior when legend is supplied by user.

RUFAS/graph_generator.py
ew3361zh: I like this, thanks Ed.

JoeWaddell (approved):
Great work: Ed has given nice feedback on the code, and after our discussion Friday in person, I think I've covered all the other feedback I had re: code and functionality. 

Below is an example of what's so nice about this PR: a figure with mixed units, and the user only needs to supply `".*"` for `variables`, and flag `omit_legend_prefix` as `true`!  image

Issue #1751, opened 2024-06-10
[Manure] Typo: Incorrect Barn Area Values for Lactating Cows in `barn_area_from_pen_type` 


tomhuhh:
## Expected Behavior

The system should use the default barn area values specific to lactating cows when calculating metrics for this animal category.

## Current Behavior

The barn area values for non-lactating animals are being used for all animal categories, including lactating cows, instead of the default values specific to lactating cows due to a typo in the if statement of method `barn_area_from_pen_type` in the file `manure_manager_pen.py`.

## Possible Solution

Review and correct the logic in the code where barn area values are assigned to ensure that lactating cows receive the appropriate default values. 

## Steps to Reproduce

1. Add a print statement:` print("classes in pen", self.classes_in_pen) ` in the method `barn_area_from_pen_type` in the file `manure_manager_pen.py`.
2.
3.
4.

## Context (Environment)

This issue impacts the accuracy of the barn area allocation for lactating cows, leading to potential errors in simulation outcomes. It is critical to ensure that lactating cows receive the appropriate barn area values to maintain accuracy of the RuFaS model. 

## Detailed Description

The application is expected to use specific default barn area values for lactating cows. However, due to a bug, the barn area values intended for non-lactating animals are being applied across all animal categories. This results in inaccurate barn area allocations, particularly for lactating cows, which require different space considerations compared to non-lactating animals.

## Target Deadline

PR #1752

tomhuhh:
Closing the issue as it is resolved by PR #1752

PR #1752, opened 2024-06-10
[Manure] Fix Incorrect Barn Area Assignments for Lactating Cows in `barn_area_from_pen_type`

tomhuhh:
This pull request fixes a bug that caused the barn area values for non-lactating animals to be used for all animal categories, including lactating cows. The changes ensure that the default barn area values for lactating cows are correctly applied.

This is to fix issue #1751 

### Context

- [ ] Yes
- [x] No

### What

- Correcting the logic to ensure that default barn area values for lactating cows are applied.
- Updating unit tests to verify the correct assignment of barn area values for each animal category.

### Why

These changes are crucial to ensure accurate barn area allocation for lactating cows, which is essential for simulation outcome accuracy. 

### How

- Reviewing and correcting the code where barn area values are assigned to different animal categories.
- Updating and adding unit tests to ensure the correctness of the barn area assignment logic.

### Test plan

- Running unit tests to ensure that the correct barn area values are assigned to lactating cows.
- Manually verifying that the application correctly differentiates between lactating and non-lactating animals and applies the appropriate barn area values.


ew3361zh:
Thanks, Haowen, this looks good. I just have a couple questions before approving:
1. So if there are DryCows in the classes_in_pen, then it shouldn't count as the barn_area.has_cows, right? It should only count if they are LacCows?
2. If (1) above is true, I think it would be helpful to update the docstrings and some of the variable names in the function for clarity.

e.g. from the docstrings:
```
Get the barn area based on the pen type and whether there are cows in the pen.
```
Should become:
```
Get the barn area based on the pen type and whether there are lactating cows in the pen.
```

I think I would also update that in the table ("Has Cows" -> "Has Lac Cows") and ("No Cows" -> "No Lac Cows").

Finally, I would update the lines:
```
BarnArea = NamedTuple("BarnArea", [("has_cows", float), ("no_cows", float)])
        freestall = BarnArea(has_cows=3.5, no_cows=2.5)
        tiestall = BarnArea(has_cows=1.2, no_cows=1.0)
        bedded_pack = BarnArea(has_cows=5.0, no_cows=3.0)
        open_lot = BarnArea(has_cows=5.0, no_cows=3.0)
```

to 

```
BarnArea = NamedTuple("BarnArea", [("has_lac_cows", float), ("no_lac_cows", float)])
        freestall = BarnArea(has_lac_cows=3.5, no_lac_cows=2.5)
        tiestall = BarnArea(has_lac_cows=1.2, no_lac_cows=1.0)
        bedded_pack = BarnArea(has_lac_cows=5.0, no_lac_cows=3.0)
        open_lot = BarnArea(has_lac_cows=5.0, no_lac_cows=3.0)
```

And this section:
```
if "LacCow" in self.classes_in_pen:
            return barn_area.has_cows * self.num_stalls
        return barn_area.no_cows * self.num_stalls
```

Would be:

```
if "LacCow" in self.classes_in_pen:
            return barn_area.has_lac_cows * self.num_stalls
        return barn_area.no_lac_cows * self.num_stalls
```

elle-andreen:
I agree with Niko's two comments/suggestions above regarding the use of "cows" vs. "no cows" and the docstring updates. Something else that I would consider adding is more clarity in the doc \string that these surface area values are actually (to my knowledge) the area of the stall **soiled with manure**, not the area of the stall itself. These values, and the housing emission equations they are used in, came from IFSM (as you probably know Haowen) and that is the interpretation they are presented with. 

Additionally, I think we should make an issue to revisit the soiled surface area calculations in general. I know this PR is to fix the existing bug, not to necessarily revisit the scientific concepts, so I don't think it needs to be addressed here. But, I think the way we are using these values & equations here does not align with how they were originally presented in IFSM. Especially for calves and growing animals, our housing CH4 and NH3-N estimations are currently very high, e.g. many days all of the VS or TAN excreted is lost through gas emissions. This could be related to the equations themselves, as they are empirical rather than mechanistic, but I think the way we are determining the surface area component of the equation is also contributing to the problem. I would be happy to write up that issue!

tomhuhh:
Thank you for your valuable insights @ew3361zh and @elle-andreen! 

Elle, if you could create an issue that you mentioned, it would be great! Thanks!

RUFAS/routines/manure/pen_manure/manure_manager_pen.py
```diff
@@ -137,24 +137,24 @@ def barn_area_from_pen_type(self) -> float:
             "compost bedded pack barn", or "open lot".
         """
 
-        BarnArea = NamedTuple("BarnArea", [("has_cows", float), ("no_cows", float)])
-        freestall = BarnArea(has_cows=3.5, no_cows=2.5)
-        tiestall = BarnArea(has_cows=1.2, no_cows=1.0)
-        bedded_pack = BarnArea(has_cows=5.0, no_cows=3.0)
-        open_lot = BarnArea(has_cows=5.0, no_cows=3.0)
+        ExposedManureSurfaceArea = NamedTuple("BarnArea", [("has_lac_cows", float), ("no_lac_cows", float)])
```
ew3361zh:
```suggestion
        ExposedManureSurfaceArea = NamedTuple("ExposedManureSurfaceArea", [("has_lac_cows", float), ("no_lac_cows", float)])
```

ew3361zh (approved):
Looks good Haowen, nice work!

1 thing to fix is the named typle should match the ExposedManureSurfaceArea name - I added it as a suggested change.

Another suggestion would be to either update the description of the PR or add a comment listing the additional changes you've made so we have a clear record of it.

elle-andreen (approved):
These changes look good to me! I have created an issue (#1766) with some more info and suggestions on how to revise the surface area concept/problem as a whole. But I think within the scope of this PR, things look good!

PR #1754, opened 2024-06-11
Fix herd generation and add GNU arg to specify task manager metadata

### LINKED ISSUE #1659: Make path to Task Manager Metadata a GNU argument
A GNU argument to specify the path to the Task Manager Metadata should be added.

## Status

The path to the task manager metadata which determines which tasks are run is hardcoded in `main.py`.

## Intention

It should be easy for users to specify different task manager metadata's so that they can conveniently switch between different sets of tasks.

## Request

Add a GNU argument that allows users to specify the task manager metadata containing the tasks they want to run. If the user doesn't specify anything with this flag, the default task manager metadata should be run.

### LINKED ISSUE #1745: Fix herd generation
Herd generation is broken and is not being tested regularly.

## Status

Herd generation currently does not work on the `main` branch. It crashes with the following:
image

## Intention

Getting herd generation running again.

## Request

Fix herd generation so that it runs on the default metadata when the task is changed to "HERD_INITIALIZATION".

Also, write a new task file that includes running herd initialization.

### Target deadline

By the end of sprint 16 (6/20).

Isgotkowitz:
This PR fixes herd generation so that it will work again, and adds a GNU arg to allow users to specify which task manager metadata they want to run from the command line.

Note: I will update the GNU Args Wiki when this PR is ready to be approved and I'm about to merge.

### Context

Issue(s) closed by this pull request: closes #1745, closes #1659

- [x] Yes
- [ ] No

### What

Fixing herd generation only required updating the call to `update` on `Cow`s in `HerdFactory`. The method had been updated a while ago to stop `update` from returning a bunch of unnecessary information.

Adding a GNU arg was as simple as adding another arg through to the ArgParser in `main.py`. It defaults to the metadata path that was previously hardcoded into `main.py`, so unless the option is invoked there is no difference in the behavior of RuFaS.

### Why

Herd generation needs to work.

Being able to specify the path to the task metadata from the command line will make it much easier for users to switch between running different sets of tasks, saving lots of time and file-reverting.

### How

pass

### Test plan

Unit tests for herd factory and main have been updated to verify the changes.

To test herd generation and the new GNU arg at the same time, the following can be run:
```sh
python main.py -p input/metadata/herd_init_metadata.json
```
or 
```sh
python main.py --path-to-metadata input/metadata/herd_init_metadata.json
```

Then running the following command to ensure that the default scenario is still run by default:
```sh
python main.py
```

Also, the default scenario has been run with the newly generated herd population input to make sure that the new herd is valid.


input/data/tasks/herd_init_task.json
```diff
@@ -0,0 +1,13 @@
+{
+  "parallel_workers": 4,
+  "tasks": [
+    {
+      "task_type": "HERD_INITIALIZATION",
+      "metadata_file_path": "input/metadata/default_metadata.json",
+      "output_prefix": "herd_init",
+      "save_animals": true,
+      "log_verbosity": "logs",
```
ew3361zh: Since logs will print out a lot of unnecessary warnings in addition to the logs, I think this would be better left at errors unless you have a strong reason otherwise.

```suggestion
      "log_verbosity": "errors",
```

ew3361zh (approved):
Herd generation is working again!
Task Manager metadata override was also working.
I do think it would be helpful to have better distinction between TM metadatas and non-TM metadatas in the metadata folder since if someone tries to use a path to any of the non-TM metadatas as an arg for this new GNU, it will crash.

input/data/tasks/herd_init_task.json
Isgotkowitz: Nope, no strong reason, I just like having a big wall of colorful text to read when I run RuFaS

JoeWaddell (approved):
Clean changes, straightforward implementation of new argument. Agreed that some way to better differentiate between metadata types would be ideal, even if its no more complex than prefixes we use so the files sort out visually, as an indication of which is used where.

Isgotkowitz:
> I do think it would be helpful to have better distinction between TM metadatas and non-TM metadatas in the metadata folder since if someone tries to use a path to any of the non-TM metadatas as an arg for this new GNU, it will crash.

> Agreed that some way to better differentiate between metadata types would be ideal, even if its no more complex than prefixes we use so the files sort out visually, as an indication of which is used where.

Thanks for the reviews @ew3361zh and @JoeWaddell. Going to merge this PR for now, with the metadatas as-is. But I will bring up separation of metadatas with Pooya tonight, and if he is alright with it I will make another PR and separate them.

Isgotkowitz:
Wiki is updated!

PR #1685, opened 2024-05-21
Modify Animal module manure N equations to match Manure module 

elle-andreen:
Manure total ammoniacal nitrogen (TAN) values are utilized predominantly in the Manure module, but are calculated in the animal manure excretion files in the Animal module. Previously, manure TAN concentration (mg/L) was calculated in the Animal module, then converted to total manure TAN (kg) in the Manure module in `pen_manure.py`. This was changed several months ago in PR #1296, which modified manure TAN (kg) to be a constant fraction of urine N (i.e., the constant `urine_TAN_factor`), instead of utilizing the manure TAN calculation in the Animal module. A more recent PR (#1587) further modified how manure TAN is calculated for all animal classes by removing `urine_TAN_factor` in `pen_manure.py`, thus setting manure TAN equal to urine TAN. 

With the recent changes to how TAN is estimated, the Animal module excretion files should be updated so that Animal and Manure module TAN values are in agreement. 

### Context

Issue(s) closed by this pull request: closes #1666 

Related PRs: #1296 and #1587 

- [x] The [changelog](https://docs.google.com/spreadsheets/d/1U-3igxdstHTFb6k5dVcE5-x9IL0r6q18eUiwwytKZyA/edit#gid=0) has been updated.
      
How big are the changes in the PR? Would you nominate it for a major version upgrade?
- [ ] Yes
- [X] No

### What
- All instances of `total_ammoniacal_nitrogen_concentration` in Animal module manure excretion files were modified to `manure_total_ammoniacal_nitrogen`, and were set equal to `urine_N`. Units were also updated from to kg. 
- `tan_percent_of_urea` (used in the previous estimation of urine/manure TAN) is no longer in use and was removed. 
- `manure_total_ammoniacal_nitrogen` is now called in `pen_manure.py` in order to import manure TAN values into the manure module directly, rather than utilizing a urine N value as before. 
- Manure TAN (kg) and other values in Manure module outputs are unchanged. 
- Tests were updated.

### Why
Values for manure TAN concentration in the Animal module were different from manure TAN values in the Manure module, which is undesirable. Animal module TAN predictions needed to be deprecated and replaced with updated methods as in the Manure module. 

### How
See "What" section. Misalignment between modules was corrected by updating Animal module calculations to match those in use in the Manure module. 

### Test plan
[...]

PR #1756, opened 2024-06-11
Revert "Modify Animal module manure N equations to match Manure module "

Isgotkowitz:
Reverts RuminantFarmSystems/MASM#1685


ew3361zh (approved):
thanks Ed

allisterakun (approved):
Thanks Ed!

PR #1759, opened 2024-06-13
Add tasks to run all available metadatas

### LINKED ISSUE #1746: Write set of tasks to test all available simulations on `main`
Title

## Status

The available simulations besides `default_metadata.json` are not included in any task manager metadatas - and are probably not being run very often.

## Intention

By writing a task that will use all these metadatas, we make it easier for developers and reviewers to check that their changes are still working for all available simulations. This encourages a more thorough testing process and allows us to catch more breaks before they happen.

## Request

Write a new tasks file that runs the following metadatas:
- ARL
- barnyard
- default
- multi crop
- swat
- testing

Also, write a task manager metadata file which runs the new tasks file.
### Additional Information
The non-default metadata scenarios might not work. If they are broken, please fix them so that they work.

Consider combining the new tasks file with the tasks file that will be written to test herd initialization (see #1745).

Isgotkowitz:
Creates a new set of tasks that runs all available simulations.

### Context

Issue(s) closed by this pull request: closes #1746 

- [ ] Yes
- [x] No

### What

This PR adds a set of tasks that run all the available simulation scenarios: default, ARL, barnyard, multi crop, swat, and testing.

It also removes some inputs from various input files that are no longer needed/necessary.

### Why

Having this set of tasks easily available makes it easier for people developing and reviewing code to test RuFaS and make sure that it is not breaking.

### How

The six available scenarios have been added as simulation single runs in the new tasks file `check_tasks.json`, which is referenced in the new `check_tasks_metadata.json`.

Inputs that were invalid or unneeded have been removed from `no_animal.json`, `sandy_barnyard_soil.json`, and `purchased_feed.json`, and a couple of the crop schedule files.

### Test plan

Run `check_tasks_metadata.json` and check that all tasks are completed.


matthew7838 (approved):
For the metadata, LGTM and it's running fine. However, adding some short explanations for those removed data might be clearer. My main question is why there's now calf_num and other attributes in ` no_animal.json`, looks a bit counterintuitive from my understanding.

Isgotkowitz:
> For the metadata, LGTM and it's running fine. However, adding some short explanations for those removed data might be clearer. My main question is why there's now calf_num and other attributes in ` no_animal.json`, looks a bit counterintuitive from my understanding.

@matthew7838 thank you for the review, it definitely is counterintuitive why no animals now has animals in it. There is/was a bug causing the simulation to crash if the animal counts were too low, so the minimums were raised from 0 to be 2 - 6. Now the only way to simulate no animals is to set the `simulate_animals` flag to false. Do you think it would be better to just delete `no_animals.json` altogether?

input/data/crop/ARL_rotation.json
```diff
@@ -10,10 +10,7 @@
             "harvest_years": [1989],
             "harvest_days": [305],
             "harvest_operations": ["harvest_kill"],
-            "harvest_type": "scheduled",
-            "planting_order": "1st",
-            "extracted": true
```
allisterakun, line 15: Do we need to update the metadata properties if these are taken out?

input/data/feed/purchased_feed.json
```diff
@@ -22,8 +22,8 @@
 
     "storage_options": [
         {
-            "storage_type": "bunker",
-            "moisture": "direct_cut",
+            "storage_type": "Bunker Silo",
+            "moisture": "Direct Cut",
```
allisterakun, line 26: Do we nee to update the regex pattern in metadata properties for this also?

input/data/tasks/check_tasks.json
allisterakun, line 1: I know thinking of names is hard, but maybe a different name? b/c "check_tasks" is a bit ambiguous, I thought it was literally just checking the tasks at first glance.

allisterakun (approved):
Great work, Ed! It looks good overall! Just left a few comments.

matthew7838:
> > For the metadata, LGTM and it's running fine. However, adding some short explanations for those removed data might be clearer. My main question is why there's now calf_num and other attributes in ` no_animal.json`, looks a bit counterintuitive from my understanding.
> 
> @matthew7838 thank you for the review, it definitely is counterintuitive why no animals now has animals in it. There is/was a bug causing the simulation to crash if the animal counts were too low, so the minimums were raised from 0 to be 2 - 6. Now the only way to simulate no animals is to set the `simulate_animals` flag to false. Do you think it would be better to just delete `no_animals.json` altogether?

@Isgotkowitz I do think that it's better to delete the file. I'm pretty sure that someone will raise the same question in the future. I think we should at least change the name to `minimum animal` or something, but at that point, I'm not sure if there's a purpose to keep the file.

input/data/crop/ARL_rotation.json
Isgotkowitz: No, these were legacy inputs that had been around for a while and never had the extraneous keys removed

input/data/feed/purchased_feed.json
Isgotkowitz: No, I don't think so, these are being updated so that they match the patterns in the metadata. What would they need to be updated to?

input/data/tasks/check_tasks.json
Isgotkowitz: Maybe "available_simulation_tasks"?

PR #1760, opened 2024-06-14
inputs c2f2

Rebecavb:
crop and soil inputs for cohort 2 farm 2


jadamchick (approved):
Looks good and runs smoothly! I added the purchased feed LUC emissions blob to the metadata file so it will run with the feed emissions updates.

Rebecavb:
> Looks good and runs smoothly! I added the purchased feed LUC emissions blob to the metadata file so it will run with the feed emissions updates.

Perfect. Thanks Julie

PR #1762, opened 2024-06-17
Remove default `ManureTreatment` factory

### LINKED ISSUE #1425: Can no longer "choose own adventure" for manure separator
## Expected Behavior
Be able to name and define the specs for a variety of manure separator systems beyond just the two listed (rotary screen and screw press). 

## Current Behavior
Because there are only 2 patterns listed in the metadata properties for manure_separator, it will not accept anything named other than these two. 

## Possible Solution

I wonder if the solution for manure handlers in PR #1250 would work here, allowing different names/variations of separator. You might still have to specify "type" as one of the two listed, but at least could name it according to its true identity (and update the specs accordingly). 

## Steps to Reproduce

1. In manure_management.json, copy/paste one of the manure_separator blobs, rename the type to "weeping wall", and specify manure_separator in one of the manure management scenarios as "weeping wall"
2.
3.
4.

## Context (Environment)

Many pilot farms have inputs for manure separators that are different from the two listed -- either because they represent another separator type entirely, or a combination of multiple separators used together.  These farms will not run without changing the separator inputs. 

For now, you can work-around by just altering the specs of one of the listed separators while keeping the name "screw press" or "rotary screen", but seems confusing/error-prone when what is really meant is "weeping wall" or another separator type or combination. 

There is no way to indicate in the .json that "where this says screw press, I really mean weeping wall."

### LINKED ISSUE #1524: [Manure] Remove Default Factories and Increase Configurability
Defaults for manure beddings, separators, and treatments are hardcoded into factory classes, and do not allow for configuration of multiple beddings, separators, or treatments of the same type.

## Expected Behavior

All configuration of manure processing components should come from the manure management input file. Users should be able to configure and use multiple beddings, separators, and treatments of the same type (i.e. configure and use two different Compost Bedded Pack Barn beddings).

## Current Behavior

If a user removes the configuration for a bedding, separator, or treatment but a manure management scenario calls for that bedding, separator, or treatment, then the scenario will use the defaults that are hardcoded for that bedding, separator, or treatment without any issue or warning to the user.

If a user adds two configurations for the same type of bedding, separator, or treatment, then the Manure module will only use one of them without alerting the user.

## Possible Solution

Use the same approach from PR #1250.
- Remove the default factories.
- Add a name field as a unique identifier for each bedding, separator, and treatment.
- Change the manure management scenarios to look at the name field when finding the components of that manure management scenario.

This solution will need to be applied for:
- [x] #1525
- [x] #1526
- [x] #1527

## Context (Environment)

This problem was brought up originally for manure handlers in issue #1127 and fixed in #1250. The pattern for fixing that issue should be applied when addressing this issue.

These issues should be addressed as part of #1237

### LINKED ISSUE #1527: [Manure] Remove default factory for manure treatments and increase configurability
Sub issue of https://github.com/RuminantFarmSystems/MASM/issues/1524, please read that issue because it goes in depth for what needs to happen to close this issue.

Steps to close this issue:
- Remove the `DefaultManureTreatmentConfigFactory`
- Add a name field as a unique identifier for each treatment configuration.
- Change the manure management scenarios to look at the name field when finding the treatment of that manure management scenario.

The fix executed in https://github.com/RuminantFarmSystems/MASM/pull/1250 for issue https://github.com/RuminantFarmSystems/MASM/issues/1127 can be used as a template for how to address this issue.

Isgotkowitz:
Removes `DefaultManureTreatmentConfigFactory` and adjusts the Manure module's routines so that treatments are only configured with inputs from the user.

Note that "treatments" in RuFaS currently encompass both manure treatments and storages. We (the devs) are aware of this and are planning out ways to address this issue.

### Context

Issue(s) closed by this pull request: closes #1527, closes #1524, closes #1425.

Note: I think #1425 should have already been closed, because it is specific to separators, but with this PR and the closing of #1524 it is now possible to choose your own adventure for all the different manure processing components.

- [ ] Yes
- [x] No

### What

Previously, the Manure module would only allow for one treatment to be configured for each type of treatment. If a user specified a treatment type that was not configured in the input, the Manure module would use a configuration from the `DefaultManureTreatmentConfigFactory`.

This PR makes manure treatments more customizable, and places more guard rails around the configuration and initialization in the Manure module so that the user is given more precise control over how treatments are configured. There are also more logs/warnings to keep the user more aware of what is happening.

### Why

The Manure module was capable of simulating a lot more than it might have been configured to do by the manure management input file, but this was a problematic design because it hid values and configuration information from the user.

### How

The `DefaultManureTreatmentConfigFactory` has been removed from the Manure module. Now, the only configurations available for use in the manure management scenarios are those that are defined in the manure management input file.

To allow users to simulate multiple different configurations of the same "type" of treatment, a "name" field has been added to the configuration of each treatment. The manure input metadata properties have been updated to account for this change, and provide more detailed descriptions about how the configuration of treatments works.

### Combined treatment/storage Options
There are two combined treatment-storage options: anaerobic digestion and lagoon, and anaerobic digestion and lagoon with separator. They are both a combination of anaerobic digestion and an anaerobic lagoon (duh), and because of this neither of these options are _directly_ configured. Rather the `ManureManagerConfigHandler` will save the last digestion and lagoon configs it sees and use them to configure both the combined options.

A log tells the user which digestion and lagoon configs were used, and if no digestion or lagoon configs are provided then a warning will be logged that the combined digestion-lagoon options could not be configured. If the use of one of these treatment options is still attempted, the simulation will crash.

Obviously, this behavior is not ideal, but a longer-term fix would be more appropriate for after this PR - IMO, if reviewers feel differently please let me know.

### Test plan

Unit tests have been updated and added to test new functionality.

RuFaS has been tested to ensure it crashes as expected with the following changes to the manure management input file:
- A manure management scenario references a treatment that is not defined.
- A manure treatment is defined with the name "anaerobic digestion and lagoon" or "anaerobic digestion and lagoon with storage"
- The same name is used in multiple manure treatment configurations.
- A manure treatment is defined with an invalid `manure_treatment_type`.

The behavior and outcomes for the default scenario are unchanged. This has been verified by running the default scenario on this branch and main, and observing that there are no differences between the following report on both branches:
```json
{
    "multiple": [
        {
            "name": "Animal to Manure Module",
            "filters": ["AnimalModuleInputToManureModule.*"],
            "variables": [".*"],
            "vertical_aggregation": "sum"
        },
        {
            "name": "Pen 1 Related Outputs",
            "filters": [".*pen_1.*"],
            "variables": [".*"],
            "vertical_aggregation": "sum"
        },
        {
            "name": "Pen 2 Related Outputs",
            "filters": [".*pen_2.*"],
            "variables": [".*"],
            "vertical_aggregation": "sum"
        },
        {
            "name": "Pen 3 Related Outputs",
            "filters": [".*pen_3.*"],
            "variables": [".*"],
            "vertical_aggregation": "sum"
        },
        {
            "name": "Pen 0 Related Outputs",
            "filters": [".*pen_0.*"],
            "variables": [".*"],
            "vertical_aggregation": "sum"
        }
    ]
}
```


RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
```diff
@@ -265,52 +276,80 @@ def _process_manure_separator_configs(
...
+        dict[str, Union[ManureTreatmentConfig, Tuple[ManureTreatmentConfig, ManureTreatmentConfig]]]
             A dictionary of manure treatment config objects, with the key being the manure treatment type.
 
+        Notes
+        -----
+        There are two cases which require a combination of anaerobic digestion and anaerobic lagoon - anaerobic
+        digestion and lagoon, and anaerobic digestion and lagoon with separation. In these cases, we return a tuple of
+        two manure treatment configs, the first being the config for the digester and the second being the config for
+        the lagoon. If a config is not provided for either the digester or the lagoon, then  the simulation will crash
+        if the user specified the one of the digestion-lagoon combinations to be used.
```
elle-andreen: Just a typo - extra "the" after "specified"

input/metadata/properties/default.json
```diff
@@ -4506,10 +4505,15 @@
       "type": "array",
       "properties": {
         "type": "object",
+        "name": {
+          "type": "string",
+          "description": "The unique identifier of the manure treatment configuration.",
+          "pattern": "^(?!anaerobic digestion and lagoon|anaerobic digestion and lagoon with separator$)(.*)$"
```
elle-andreen, line 4511: It's not clear to me why the two combo treatments are pattern options for manure_treatment_configs ? Should these not also/instead be options for manure_treatment_type?

RUFAS/routines/field/soil/nitrogen_cycling/denitrification.py
```diff
@@ -59,7 +59,7 @@ def denitrify(self) -> None:
                 layer.nitrate_content,
                 layer.denitrification_rate_coefficient,
                 layer.nutrient_cycling_temp_factor,
-                layer.soil_overall_carbon_fraction * 100,
+                layer.soil_overall_carbon_fraction,
```
elle-andreen: Is this an incidental change or meant to be part of this PR?

Isgotkowitz: That is incidental, no idea how it got there

input/metadata/properties/default.json
Isgotkowitz: This part is very confusing. The regex pattern means that the name of the treatment being configured can be anything, _except_ "anaerobic digestion and lagoon" or "anaerobic digestion and lagoon with separator".

Because these two options use both a digester and lagoon, they can't be configured with a single treatment configuration - they need two. Which is why it wasn't possible before for users to configure "anaerobic digestion and lagoon" and "anaerobic digestion and lagoon with separator" directly, even though the input file made it _seem_ like that was the case.

With the changes to the metadata, I wanted to make it as difficult as possible for the user to get the wrong idea that they are directly configuring "anaerobic digestion and lagoon" and "anaerobic digestion and lagoon with separator", which is why users can reference them in the manure management scenario configurations, but cannot use them as the "name" or "manure_treatment_type" in the treatment configurations.

RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
Isgotkowitz: Nice catch

input/metadata/properties/default.json
elle-andreen: That makes sense, thank you for the detailed clarification! I figured it had something to do with 'under the hood' functionality. I agree that this is a good addition to prevent users from trying to directly configure combo treatments. You probably had this in mind as you were working on changes, but I think this also sets the code up well to eventually 'unlink' the combo treatments options, and enable something like anaerobic digestion with slurry storage.

input/data/manure/manure_management.json
```diff
@@ -292,6 +295,7 @@
       "manure_cover": "cover"
     },
     {
+      "name": "anaerobic digestion",
       "manure_treatment_type": "anaerobic digestion",
       "total_solids_removal_efficiency_for_treatment": 0.45,
```
elle-andreen, line 300: Haowen, Kristan and I have been meaning to remove all of these "removal efficiency" configurations for a while now. At one time all the manure treatments had non-zero values for removal efficiencies, and they've been slowly removed or set to 0 over time, but anaerobic digestion is a holdout I guess. I think removing them all is outside the scope of this specific PR, but since this PR does deal with removing unneeded default configurations, maybe total solid, volatile solids, and total ammoniacal nitrogen removal efficiencies for anaerobic digestion could be set to 0 for now. When a removal efficiency is modified, it don't actually result in any changes (to my knowledge), so if they're not getting removed, setting them to 0 would be preferable I think.

Isgotkowitz: Good to know! I will try out changing the efficiencies for anaerobic digestion and look into how these values are used. If they're not being used at all/have no effect on outputs, I will change the efficiencies for anaerobic digestion and write up an issue for removing them.

elle-andreen (approved):
My questions/comments have been clarified and all looks good to me - thanks Ed!

input/data/manure/manure_management.json
Isgotkowitz: So I tried modifying some of the removal efficiencies, and it seems like they still have an effect on outputs - both for anaerobic digestion and the other treatments.

input/metadata/properties/default.json
Isgotkowitz: > but I think this also sets the code up well to eventually 'unlink' the combo treatments options, and enable something like anaerobic digestion with slurry storage.

Definitely. I'm not quite sure how that transition would take place, but this is bringing us a step closer to it.

RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
```diff
@@ -265,52 +276,80 @@ def _process_manure_separator_configs(
         return available_manure_separator_configs
 
     @classmethod
-    def _process_manure_treatment_configs(cls, manure_treatment_json_configs: List[Dict]) -> Dict[
-        ManureTreatmentType,
-        Union[ManureTreatmentConfig, Tuple[ManureTreatmentConfig, ManureTreatmentConfig]],
-    ]:
+    def _process_manure_treatment_configs(
+        cls, manure_treatment_configs: list[dict[str, Any]]
```
allisterakun, line 285: Just curious, for the type hints, is there a specific reason that we are using `dict` and `list` rather than `Dict` and `List` here?

RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
```diff
@@ -265,52 +276,80 @@ def _process_manure_separator_configs(
...
-                **json_manure_treatment_config
+        digester_combination_config: str | None = None
+        lagoon_combination_config: str | None = None
+        for config in manure_treatment_configs:
+            name = config.pop("name")
+            if name in available_manure_treatment_configs:
+                error_name = "Duplicate manure treatment configurations"
+                error_message = f"Manure treatment '{name}' has multiple configurations"
+                om.add_error(error_name, error_message, info_map)
+                raise ValueError(error_message)
```
allisterakun, line 326: Should this `ValueError` also be noted in the function docstrings?

allisterakun (approved):
Great work, Ed! There are no issues with me from the SWE perspective. I just left two comments about type hints and docstring.

RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
Isgotkowitz: No, it is purely my preference. Since Python 3.9 the type hints for stuff like `list` and `dict` are the same as `typing`'s `List` and `Dict` as far as I can tell, but I'm not at all an expert in this stuff so there might be some key differences that I am missing.

The only benefit I can think of is that using the built-in type hints potentially reduces the numbers of import statements that have to be made. But more important than that would be having all code in RuFaS either using `typing` library types for everything, or using all the built-in type hints as much as possible. But that has not been decided on for a team standard yet.

RUFAS/routines/manure/IO_helpers/manure_manager_config_handler.py
Isgotkowitz: Yes it should, nice catch.

PR #1767, opened 2024-06-18
Animal Health Module Setup

### LINKED ISSUE #1692: [Animal][AnimalHealth] Setup AnimalHealth Module Structure
See [Animal Health Submodule Design Doc](https://docs.google.com/document/d/1h2XPmNgGgCAKs8xTmxoHuBXZlzSxd7D2XvKjx_2vJhA/edit?usp=sharing) for details.
We will aim to simulate **mastitis** and **lameness** for now.

### Risk Determination ###
With **a series of logic checks**, RuFaS will determine if the animal is at risk of developing a particular disease on a particular simulation day.
If an animal is determined to be at risk for a certain disease, we will continue to the next stage of incidence simulation; otherwise, nothing happens, and the simulation continues as normal.

- Create a base function for the risk determination function, as the logic of risk determination for different diseases differs.

### Simulate Incidence ###
1. **Incidence rate**
    1. Hard-coded / user-input **baseline incidence rate**
    2. Pre-defined risk factors
        1. Intrinsic or extrinsic characteristics that modulate the animal’s likelihood of developing disease
        2. **Relative risk**
            1. a scalar multiplier representing the relative increase or decrease in the animal’s likelihood of developing disease due to a risk factor
    Calculate the incidence rate for an animal of a specific disease by combining the baseline incidence rate with the relative risk of each risk factor (that applies to the animal).
2. Compare a random RNG-generated value with the calculated incidence rate to determine if the animal will develop the disease within the at-risk period (life cycle stage, DIM, etc.)
3. **Disease start date**
    1. RuFaS will determine the disease start date (within the at-risk period) according to a simulation curve (probability mass function)

- Again, only set up the base methods and general structure for incidence simulation, as these differ for different diseases.
- Also, create necessary user inputs for the information needed from the user.

### Simulate Disease ###
1. **Disease length**
    1. According to epidemiological data. 
2. Disease outcome
    1. Death
        1. According to the case fatality risk
    2. Recovery
        1. An animal is eligible to recover only after n days
     3. Cull
         1. Removed (sold) from herd
     4. Remain diseased
         1. **Effects while diseased (immediate effects)**
             1. Before recovery, while the animal is still sick, the values utilized by RuFaS in simulating the animal’s daily life will be modified according to immediate effects. Examples of immediate effects include:
                 1. % reductions in hardcoded values.
                     1. E.g.: Cows with ketosis will reduce daily DMI by 50%*
                 2. New hardcoded reference values.
                     1. E.g.: ADG for calves with pneumonia is 0.3 kg/day*
 
- Again, only set up the base methods and general structure for disease simulation, as these differ for different diseases.
- Also, create necessary user inputs for the information needed from the user.

### Simulate Recovery ###
If the animal does not die during the duration n, it will recover and be tagged as “recovered” from the disease. 
1. **Effects during recovery (intermediate effects)**
    After recovery, some values may still be modified according to intermediate effects. These affect animals post-recovery during the same life cycle stage as the disease event. Examples of intermediate effects include: 
    1. % reductions in hardcoded values.
        1. E.g.: After recovering from scours, a calf’s ADG is 0.5 kg/day* until she is weaned
    2. New hardcoded reference values.
        1. E.g.: A cow that recovered from a DA will produce 10%* less milk during her lactation
2. **Permanent effects (lasting effects)**
    After recovery, values may be modified according to lasting effects in subsequent life cycle stages. Examples include:
    1. % reductions in hardcoded values.
        1. E.g. For cows that recovered from pneumonia as calves, give 2%* less milk during their first lactation
    2. New hardcoded reference values.
        1. E.g.: Conception rate is 0.33* for heifers that scoured as calves

- Again, only set up the base methods and general structure for recovery simulation, as these differ for different diseases.
- Also, create necessary user inputs for the information needed from the user.

### Output ###
At the end of every simulation, RuFaS should generate a report of the animal’s health. Output values for each disease include: 
1. Total animal-days of [disease]
2. Incidence of [disease]
3. Lost milk production due to [disease]
4. Decrease in feed efficiency due to [disease]
5. Increase in kgCO2eq/kgMilk due to [disease]
6. Decrease in income due to [disease]

- Again, only set up the base methods and general structure for output generation and reporting, as these differ for different diseases.

image

**Target Deadline**
By when should this issue be done.

ew3361zh:
The initial set up of the Animal Health module.

### Context

Issue(s) closed by this pull request: closes #1692

- [ ] Yes
- [x] No

### What

Sets up the following 4 classes:
- Disease
- DiseaseIncidence
- DiseaseRecovery
- AnimalHealthReporter

Each class has a series of methods that cover the outline from the design doc and issue. It's a little unclear from the design doc how this will be managed especially considering the large refactoring of the Animal Module on the horizon.

### Why

This skeleton structure is needed for further work and development of the Animal Health module.

### How

The structure as outlined in the Animal Health design doc as well as in the issue this PR closes was followed as best I could understand and the pieces were put into place accordingly.

### Test plan

There is no testing in place as these are just skeleton functions and frameworks.


RUFAS/routines/animal/animal_health/disease.py
```diff
@@ -0,0 +1,69 @@
+from RUFAS.routines.animal.life_cycle.animal_base import AnimalBase
+
+
+class Disease:
+    """
+    Abstract class representing disease simulation.
+    """
```
Isgotkowitz: Should this class inherit from ABC? I don't know how strict the definition is for an abstract class, but it might be the case that this is not one because it doesn't have any abstract methods.

RUFAS/routines/animal/animal_health/animal_health_reporter.py
```diff
@@ -0,0 +1,72 @@
+from RUFAS.routines.animal.animal_health.recovery import DiseaseRecovery
+from RUFAS.routines.animal.animal_health.incidence import DiseaseIncidence
+from RUFAS.routines.animal.animal_health.disease import Disease
+
+
+class AnimalHealthReporter:
+    # modeling after AnimalModuleReporter class - this could likely be folded into that if desired.
```
Isgotkowitz: My gut reaction for this is that we should be going for an all-or-nothing approach - either all the animal outputs are recorded in one place or the outputs get recorded where they are calculated/used. But this might be the perfect balance between grouping the recording of outputs together and keeping the recording of outputs close to where they are calculated.

Either way, looking forward to seeing the final design - I think it will be really helpful in creating a template that we could apply to other modules.

RUFAS/routines/animal/animal_health/disease.py
```diff
@@ -0,0 +1,69 @@
...
+    def __init__(self) -> None:
+        # get info about specific disease set by user
+        # im = InputManager()
+        # self.duration = im.get_data("disease.specific.duration")
+        # self.outcome_distribution = im.get_data("disease.specific.outcome_distribution")
+        # self.days_to_recovery = im.get_data("disease.specific.days_to_recovery")
+        # other disease-specific data: risk factors, incidence factors, etc.
+        pass
+
+    def start_disease(self, animal: AnimalBase, simulation_day: int) -> None:
```
Isgotkowitz: Instead of passing in the simulation day, would it be better to pass in a `Time` instance?

RUFAS/routines/animal/animal_health/disease.py
```diff
@@ -0,0 +1,69 @@
...
+        """Modifies daily life values of animal based on disease immediate effects.
+
+        Parameters
+        ----------
+        animal : AnimalBase
+            The diseased animal.
+        """
+        # pass
+
+    def _end_disease(self, animal: AnimalBase, outcome: str) -> None:
```
Isgotkowitz:
```suggestion
    def _end_disease(self, animal: AnimalBase, outcome: str) -> None:
```
Could this outcome be an enum instead of a string? I looked through the design doc, and it seems that there are a number of different outcomes that could happen for a cow, but it still might be better to define them through enums instead of strings.

RUFAS/routines/animal/animal_health/incidence.py
```diff
@@ -0,0 +1,82 @@
...
+        -------
+        float
+            The incidence rate of the disease.
+        """
+        # im = InputManager()
+        # im.get_data("location.of.baseline_incidence_rate") - get the user-input baseline incidence rate for disease
+        # combine relative risk factors with baseline_incidence_rate
+        pass
+
+    def _will_develop_disease(incidence_rate: float) -> bool:
```
Isgotkowitz: I would like to clarify, this method tells us whether an animal will or will not develop a disease, but not _when_ it will actually develop that disease?

RUFAS/routines/animal/animal_health/incidence.py
```diff
@@ -0,0 +1,82 @@
...
+        risk_factors : float
+            The relative risk factors the animal has.
+
+        Returns
+        -------
+        float
+            The incidence rate of the disease.
+        """
+        # im = InputManager()
+        # im.get_data("location.of.baseline_incidence_rate") - get the user-input baseline incidence rate for disease
```
Isgotkowitz: Not very important because this isn't actually implemented yet, but would it better to pull this data in `init` and store as an attribute instead of calling it from the IM every time this method runs? I know that was an issue with some things in the past, and minimizing the number of IM calls had a not insignificant benefit to performance.

Isgotkowitz (approved):
This is an excellent start to the Animal Health module Niko. I left a few comments and questions, but saw nothing specific that needs to changed right now.

A pattern I noticed a lot of was methods taking in a single AnimalBase instance at a time. What would you think of changing this so that lists of AnimalBase instances are taken instead? I ask because performance is already an issue in the Animal module, and as I understand it, it would be faster to have a single method call that takes in a list of things to process instead of iterating over a list of things and calling a method on every single thing.

RUFAS/routines/animal/animal_health/disease.py
allisterakun: Second that

allisterakun (changes requested):
I'm very sorry in advance. Niko, you'll probably hate me for this, but we have a circular import scenario due to my unclear design. 

The `Animal` class (which is `AnimalBase` atm) should import the `AnimalHealth` sub-module, and some daily routine "update/check" function for animal health should be run in `Animal.daily_routines()` (which is the update function rn).

I will discuss this more during our work together on Monday, 6/25, and I would appreciate your help with the animal module design in general. It would be a small change for this PR. But it does change how I pictured the Animal module in general.

Again, my apologies in advance!

RUFAS/routines/animal/animal_health/disease.py
ew3361zh: That's a good call. I think this is left over from my initial design where I had the disease class doing a lot more and I thought there would be more specific outcomes that would need to occur for each disease type.

RUFAS/routines/animal/animal_health/animal_health_reporter.py
ew3361zh: I will move it to the AnimalModuleReporter class for now.

RUFAS/routines/animal/animal_health/disease.py
ew3361zh: Done!

RUFAS/routines/animal/animal_health/disease.py
ew3361zh: Definitely. I like this idea.

RUFAS/routines/animal/animal_health/incidence.py
ew3361zh: That's correct. Though I was a little confused when trying to suss out the exact interaction of the at-risk-period and the incedence-rate. Perhaps there is more interaction than I've set up here or a better way of relating the two.

allisterakun (approved):
Great job, Niko!

RUFAS/routines/animal/animal_health/outcomes.py
```diff
@@ -0,0 +1,34 @@
...
+        Animal continues to be afflicted by the disease.
+
+
+    """
+
+    HEALTHY = "healthy"
+    DEATH = "death"
+    RECOVERY = "recovery"
+    CULL = "cull"
+    REMAIN_DISEASED = "remain_diseased"
```
Isgotkowitz: This is really splitting hairs, but could/should the enum terms be tweaked to make them a little more grammatically consistent? For example, changing them so that they all complete the sentence "the animal is _________". This might change death to dead, recovery to recovering, cull to culled, and remain diseased to diseased (of course there are many other possibilities though).

input/data/animal_health/example_animal_health.json
```diff
@@ -0,0 +1,29 @@
...
+        },
+        "immediate_effects": {
+            "effect_1": {
+                "name": "milk_production_decrease",
+                "percent": 5 
+            },
+            "effect_2": {
+                "name": "miscarriage_likelihood_increase",
+                "percent": 3.5
+            }
```
Isgotkowitz, line 14:
```suggestion
            "milk_production_percentage_decrease": 5 ,
            "miscarriage_likelihood_percentage_increase": 3.5
```
I'm not sure what other data fields might need to be included for each effect, if they are not just limited to name and percent then my suggestion probably wouldn't be very helpful.

RUFAS/routines/animal/animal_health/animal_health_status.py
```diff
@@ -0,0 +1,23 @@
...
+    def __init__(self) -> None:
+        # starting list for attribute data needed from Animal object:
+        # id
+        # breed
+        # body_weight
+        # daysBorn (age)
+        # animal_type (cow, calf, heiferI, etc)
+        # is_lactating
+        # parity, calving_interval, days_in_preg, other repro status data?
+        # other attributes will likely be needed
```
Isgotkowitz, line 20: AnimalHealthStatus will be a great way to package and share data about animals between different disease modules - I am curious to see how it will be integrated with the Animal object.

RUFAS/routines/animal/animal_health/animal_health.py
```diff
@@ -0,0 +1,32 @@
...
+                    if will_develop_disease:
+                        disease_start_date = disease.determine_at_risk_period(animal_health_status)
+                        animal_health_status.disease_start_date = disease_start_date
+        elif animal_health_status.status == DiseaseOutcomes.RECOVERY:
+            if animal_health_status:  # some way to determine if the animal is in the same life stage when it recovered
+                # Disease.intermediate_effect()
+                pass
+            else:
+                # Disease.lasting_effect()
+                pass
```
Isgotkowitz:
```suggestion
        elif animal_health_status.status == DiseaseOutcomes.RECOVERY:
            if animal_health_status:  # some way to determine if the animal is in the same life stage when it recovered
                # Disease.intermediate_effect()
                pass
            else:
                # Disease.lasting_effect()
                pass
        else:
            for disease in self.diseases:
                animal_at_risk = disease.assess_disease_risk(time, animal_health_status)
                if animal_at_risk:
                    incidence_rate = disease.calculate_incidence_rate()
                    will_develop_disease = disease.will_develop_disease(incidence_rate)
                    if will_develop_disease:
                        disease_start_date = disease.determine_at_risk_period(animal_health_status)
                        animal_health_status.disease_start_date = disease_start_date
```

Isgotkowitz (approved):
LGTM, this is laying excellent groundwork for the Animal Health submodule. Made a couple of comments but none of them are suggesting substantive things.

RUFAS/routines/animal/animal_health/outcomes.py
ew3361zh: I like it

input/data/animal_health/example_animal_health.json
ew3361zh: Yea, I don't know either. I might leave it this way for now just because I think there are a variety of effects and I'm not sure the flexibility we'll need/want for reading/implementing them.

PR #1775, opened 2024-06-20
Essential Amino Acid Requirement Implementation

### LINKED ISSUE #1791: Essential Amino Acid Requirements
Implement the requirements for the Essential Amino Acids (EAA).

allisterakun:
### Context

This PR implements the calculation for the essential amino acid requirements for animals. These calculations are part of the animal requirement calculations, which are run at the beginning of the simulation and once every ration period.

### What

- `AminoAcidCalculator`: This class contains the methods for amino-acid-related calculations. 
    -  Currently, only the Essential Amino Acid requirement calculations are implemented. 
    - The Non-essential Amino Acid requirement calculations and other possible amino-acid-related calculations will be implemented in the future. 
- Added an attribute to `AnimalBase` for `essential_amino_acid_requirement` that indicates the daily EAA requirements for the animal.
- Added the calculated EAA requirements to the animal requirements. 
    - The average pen EAA requirement is calculated and reported along with other animal requirements at the beginning of the simulation and every ration interval.

### Why

Calculate the EAA requirement for each animal and report the average pen EAA requirement for all animals in each pen.

### How

Following the formula in the AA_requirements design doc provided by @Rebecavb.

### Test plan

Implemented unit tests for all calculation functions.
@Rebecavb is going to review the output to evaluate the result.
However, upon review, we found that one of the inputs for the EAA requirements calculation, `gravid_uterine_weight_gain`, can sometimes be a negative number, which leads to negative EAA requirements. This is outside the scope of this PR; @allisterakun is making a new issue about the investigation.
To review the reported daily pen average EAA requirements output, run the simulation with the following filter:
```
^.*AnimalModuleReporter.report_ration_interval_data.avg_rqmts_pen.*$
```


RUFAS/routines/animal/ration/amino_acid.py
```diff
@@ -0,0 +1,325 @@
...
+        Parameters
+        ----------
+        lactating : bool
+            If the animal is lactating.
+        body_weight : float
+            Body weight (kilograms).
+        frame_weight_gain : float
+            Frame weight gain refers to the accretion of both fat and protein in carcass (grams per day).
+        gravid_uterine_weight_gain : float
+            Daiy energy Requirement associated to increased gain of reproductive tissues as pregnancy advances (Mcal/d).
```
Rebecavb:
```suggestion
            Daily energy Requirement associated to increased gain of reproductive tissues as pregnancy advances (Mcal/d).
```

RUFAS/routines/animal/ration/amino_acid.py
```diff
@@ -0,0 +1,325 @@
...
+        dry_matter_intake_estimate : float
+            Estimated dry matter intake according to empirical prediction equation within NASEM (2021) (kg/d).
+        milk_true_protein : float
```
matthew7838, line 167: Is it possible to modify the name so it can be more differentiable between contents and amount?

RUFAS/routines/animal/ration/amino_acid.py
```diff
@@ -0,0 +1,325 @@
...
+        CPMFP: float
+            Crude protein in metabolic fecal protein, g
+        NPMFP: float
+            Net protein requirement for metabolic fecal protein, g
+        NPGrowth: float
+            Net protein requirement for body frame weight gain, g
+        NPGest: float
+            Net protein requirement for pregnancy, g
+        NPMilk: float
+            Net protein in milk, or milk true protein yield, g
```
matthew7838: Better to have the same format for units as above.

RUFAS/routines/animal/ration/amino_acid.py
```diff
@@ -0,0 +1,325 @@
...
+        amino_acid : str
+            The type of amino acid to calculate.
+        NPMilk: float
+            Net protein in milk, or milk true protein yield, g
+
+        Returns
+        -------
+        float
+            The net amino acids yield in milk (grams per day).
+        """
```
matthew7838: Same with this formatting, also, I think having units in both the description and the return is a bit redundent.

tests/animal_module_tests/ration/test_amino_acid.py
```diff
@@ -0,0 +1,209 @@
...
+        ("lysine", 1.2),
+        ("methionine", 1.2),
+        ("phenylalanine", 1.2),
+        ("threonine", 1.2),
+        ("thryptophan", 1.2),
+        ("valine", 1.2),
+    ],
+)
+def test_calculate_scurf(amino_acid: str, NPscurf: float) -> None:
+    amino_acid_calculator = AminoAcidCalculator()
```
matthew7838: docstrings

tests/animal_module_tests/ration/test_amino_acid.py
```diff
@@ -0,0 +1,209 @@
...
+
+    expected_result = NPscurf * AMINO_ACID_COMPOSITION[amino_acid]["scurf"] / 100
```
matthew7838: When possible, we should pass in the manual calculated expected result as an argument into the test signature and assert it.

matthew7838 (changes requested):
Looks solid! Caught some things which shouldn't be too complicated to fix. 

- There are some inconsistencies in the docstrings and missing docstrings for the tests. 
- Some suggestions for the assertions in the tests. 
- Is there a source for the calculations? If so, it should be added to the docstrings.

RUFAS/routines/animal/ration/amino_acid.py
allisterakun: These input variables are from other parts of the animal module. I kept the original names to maintain consistency and avoid confusion.

RUFAS/routines/animal/ration/amino_acid.py
```diff
@@ -0,0 +1,350 @@
...
+            The type of amino acid to calculate.
+        NPscurf: float
+            Net protein requirement for scurf (g / day).
+
+        Returns
+        -------
+        float
+            The net demand for the specific amino acid in scurf excretion (g / day).
+        """
+        return NPscurf * AMINO_ACID_COMPOSITION[amino_acid]["scurf"] / 100
```
matthew7838, line 267: `GeneralConstants.PERCENTAGE_TO_FRACTION` might come in handy in this case.

tests/animal_module_tests/ration/test_amino_acid.py
```diff
@@ -0,0 +1,247 @@
...
+        ("valine", 256.7, 0.826253125),
+    ],
+)
+def test_calculate_endogenous_urinary_excretion(amino_acid: str, body_weight: float, expected: float) -> None:
+    """
+    Unit test for AminoAcidCalculator._calculate_endogenous_urinary_excretion()
+    """
+    amino_acid_calculator = AminoAcidCalculator()
+    actual_result = amino_acid_calculator._calculate_endogenous_urinary_excretion(amino_acid, body_weight)
+    assert actual_result == approx(expected, abs=0.001)
```
matthew7838, line 57:
```suggestion
    assert actual_result == approx(expected)
```
I Have tried this on a few tests and without setting the abs value it is still fine. The default is 1e-6, so keeping the default is also more accurate.

matthew7838 (approved):
LGTM, very minor comments but if those numbers are SMEs approved I think you're good to merge.

Rebecavb (approved):

PR #1776, opened 2024-06-24
Add utility function for expanding data

### LINKED ISSUE #1366: Misalignment in graphs (and reports) due to instantiation of new variables during simulation

Isgotkowitz:
This PR lays the groundwork for padding and aligning data temporally in graphs and reports.

This PR will be followed up with a PR that incorporates the padding capabilities into the Graph Generator and Report Generator. - PR #1778 

### Context

Issue(s) closed by this pull request: addresses #1366 

- [ ] Yes
- [x] No

### What

In order to align values from different variables, the simulated day that a variable is recorded on is now automatically added to the info map, though some options are provided to override this behavior. A new utility method has been introduced so that it can be 

### Why

The "when" of data is very important, and if that can't be captured in outputs then users will lose key information which will limit interpretation of results. #1366 gets into the why as well.

### How

The Output Manager's `add_variable` and `_add_to_pool` now provide the option to add the simulation day to the info map for the caller function. This functionality has not be made an option for logs, warnings, or errors - but easily could be.

A new utility method - `pad_temporal_data` has been added. This is the heart of the padding functionality. It is designed handle variables pulled out of the OutputManager's pool, and gives users options for the fill value and how values should be filled in after the last known value for a variable. Check the unit test for some good examples of how this function works.

### Test plan

Updated unit tests for the Output Manager and wrote new unit tests for `pad_temporal_data` - both suites are passing.

To test the end to end functionality, I wrote a JSON filter that only contained `Time.simulation_day` and checked that the simulation day in the info map matched the value recorded for the simulation day.

### Edit/Update
After speaking with Pooya, it was decided that defaulting to adding the simulation day to the info map of every variable added to the OM's pool was not a good idea at this time, given that it will expand the memory footprint and latency of recording variables when we are already short on memory and simulations take a long time. The data expansion capabilities will remain unchanged however.

### DIFF
--- a/RUFAS/util.py
+++ b/RUFAS/util.py
@@ -84,6 +84,71 @@ class Utility:
         return nested_structure
 
     @staticmethod
+    def pad_temporal_data(data_to_pad: dict[str, dict[str, list[Any]]]) -> dict[str, dict[str, list[Any]]]:
+        """
+        Pads data based on the simulation day(s) it was recorded on, relative to when other data was recorded.
+
+        Parameters
+        ----------
+        data_to_pad : dict[str, dict[str, list[Any]]]
+            The data to be padded. The top level key is a variable name, and points to a dictionary that contains the
+            keys "values" and optionally "info_maps".
+
+        Returns
+        -------
+        dict[str, dict[str, list[Any]]]
+            The padded data, so that gaps in the data are filled in with the last know value or None.
+
+        TODO: doesn't work if data has None as original value. Untested if there are multiple different values for one
+        simulation day.
+
+        """
+        all_simulation_days = []
+        for key, value in data_to_pad.items():
+            info_maps = value.get("info_maps")
+            if info_maps is None:
+                raise TypeError(f"Variable '{key}' has no info maps.")
+            if len(info_maps) != len(value["values"]):
+                raise ValueError(f"Variable '{key}' does not have matching number of values and info maps.")
+            if not all("simulation_day" in info_map.keys() for info_map in info_maps):
+                raise ValueError(f"Variable '{key}' does not have simulation day value in every info map.")
+            all_simulation_days += [info_map["simulation_day"] for info_map in info_maps]
+
+        filtered_simulation_days = sorted(set(all_simulation_days))
+        first_day = filtered_simulation_days[0]
+        last_day = filtered_simulation_days[-1]
+
+        padded_data: dict[str, dict[str, list[Any]]] = {}
+        for key, data in data_to_pad.items():
+            padded_variable_data: dict[str, list[Any]] = {"values": [], "info_maps": []}
+            zipped_data = zip(data["values"], data["info_maps"])
+            indexed_data = {data[1]["simulation_day"]: data for data in zipped_data}
+            last_day_of_original_data = max(indexed_data.keys())
+            last_value = None
+            for day in range(first_day, last_day_of_original_data + 1):
+                if day in indexed_data.keys():
+                    last_value = indexed_data[day]
+                    padded_variable_data["values"].append(last_value[0])
+                    padded_variable_data["info_maps"].append(last_value[1])
+                    padded_variable_data["info_maps"][-1]["simulation_day"] = day
+                else:
+                    if last_value is None:
+                        padded_variable_data["values"].append(last_value)
+                        padded_variable_data["info_maps"].append({"simulation_day": day})
+                    else:
+                        padded_variable_data["values"].append(last_value[0])
+                        padded_variable_data["info_maps"].append(last_value[1].copy())
+                        padded_variable_data["info_maps"][-1]["simulation_day"] = day
+
+            for day in range(last_day_of_original_data + 1, last_day + 1):
+                padded_variable_data["values"].append(None)
+                padded_variable_data["info_maps"].append({"simulation_day": day})
+
+            padded_data[key] = padded_variable_data
+
+        return padded_data
+
+    @staticmethod
     def deep_merge(target: Dict[Any, Any], updates: Dict[Any, Any]) -> None:
         """
         Recursively merges 'updates' into 'target'. Supports deep merging for dictionaries and lists, including lists
--- a/RUFAS/util.py
+++ b/RUFAS/util.py
@@ -3,6 +3,7 @@ import re
 import shutil
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Tuple, Optional
+import numpy as np
 
 from .general_constants import GeneralConstants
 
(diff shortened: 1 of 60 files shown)


RUFAS/util.py
```diff
@@ -83,6 +84,82 @@ def flatten_keys_to_nested_structure(input_dict: Dict[str, Any]) -> Dict[str, An
...
+            keys "values" and optionally "info_maps".
+        fill_value : Any, default numpy.nan
+            Value that is used to pad the front of the data values, and optionally the back.
+        pad_tail_values : bool, default False
+            Whether tail values should be padded with the last known value from a data set, or with the fill value.
+
+        Returns
+        -------
+        dict[str, dict[str, list[Any]]]
+            The padded data, so that gaps in the data are filled in with the last know value or None.
```
JoeWaddell:
```suggestion
            The padded data, so that gaps in the data are filled in with the last known value or None.
```

RUFAS/util.py
```diff
@@ -83,6 +84,82 @@ def flatten_keys_to_nested_structure(input_dict: Dict[str, Any]) -> Dict[str, An
 
         return nested_structure
 
+    @staticmethod
+    def pad_temporal_data(
```
JoeWaddell: Should it be clear that this is a daily padder in the name? If not possible in the name, then the description should probably make it clear(er) that it pads daily, since some information isn't reported on a daily interval.

Related, feel free to disregard: I'm assuming we shouldn't make this more complex and accept some interval argument (maybe in day increments), but it's a thought.

tests/test_util.py
```diff
@@ -311,6 +312,195 @@ def test_flatten_keys_to_nested_structure_dict_w_list() -> None:
...
+    tail_pad: bool,
+    expected: dict[str, dict[str, list[Any]]],
+) -> None:
+    """Tests the utility method pad_temporal_data."""
+    actual = Utility.pad_temporal_data(data_to_pad, fill_value=fill_value, pad_tail_values=tail_pad)
+
+    assert actual == expected
+
+
+def test_pad_temporal_data_errors() -> None:
```
JoeWaddell: Should you add a test/error for cases in which one is trying to pad a nonexistent field?

RUFAS/util.py
```diff
@@ -83,6 +84,82 @@ def flatten_keys_to_nested_structure(input_dict: Dict[str, Any]) -> Dict[str, An
...
+        Pads data based on the simulation day(s) it was recorded on, relative to when other data was recorded.
+
+        Parameters
+        ----------
+        data_to_pad : dict[str, dict[str, list[Any]]]
+            The data to be padded. The top level key is a variable name, and points to a dictionary that contains the
+            keys "values" and optionally "info_maps".
+        fill_value : Any, default numpy.nan
+            Value that is used to pad the front of the data values, and optionally the back.
+        pad_tail_values : bool, default False
```
JoeWaddell: Per in person discussion, another flag to add `pad_gap_values` would be great, to allow the user to use the `fill_value` instead of the last known value in sections missing data (e.g. data only provided for days 1 and 100 of a 100 day simulation). Roughly follows behavior of `pad_tail_values` (maybe change to `pad_end_values`?)

Can be separate PR if too big for this one.

JoeWaddell (approved):
Looks great! Clear code, and good test coverage. A few name suggestions, and suggestion for new functionality/tests.

RUFAS/util.py
Isgotkowitz: Made it clearer (I hope) in the description that data will be padded on a daily step for all values in a certain range.

> Related, feel free to disregard: I'm assuming we shouldn't make this more complex and accept some interval argument (maybe in day increments), but it's a thought.

I think it would be good to hold off on that functionality for right now, but that could definitely be feature added in the future.

Another feature that might become handy would be to pad "intermittently". I don't know a good way to describe this, so I'll just given an example:

Say `pad_temporal_data` gets variables x and y, with x being recorded on days 1, 5, and 10 and y being recorded on days 2, 7, and 9.

Currently, x and y will be padded so that they both have values for all days in the range [1, 10]. But a feature that pads "intermittently" would only add values to x and y so that they both have values for all days in the set {1, 2, 5, 7, 9, 10}.

tests/test_util.py
Isgotkowitz: What do you mean by a non-existent field?

JoeWaddell: Ah yes, sorry for the lack of clarity: I was thinking of an empty dictionary.

RUFAS/util.py
Isgotkowitz: Changed `pad_tail_values` to `fill_end_values`, and added the flag `fill_gap_values`. If `fill_gap_values` is true, then gaps between original values will be filled with the last know original value. If false, then gaps will be filled with the `fill_value`.

tests/test_util.py
Isgotkowitz: Yes! That should be checked against.

Isgotkowitz:
After speaking with @PooyaHekmati, I am going to change the naming and docstrings surrounding this method to use the term "fill" instead of "pad" - this is because "padding" refers to adding data before or after existing data points, and "expanding" refers to adding data between existing data points. The functionality introduced does both, so I think "fill" is a good catch-all term. 

Tagging @JoeWaddell just to keep you updated.

tests/test_util.py
Isgotkowitz: Checked against!

RUFAS/util.py
```diff
@@ -83,6 +84,98 @@ def flatten_keys_to_nested_structure(input_dict: Dict[str, Any]) -> Dict[str, An
...
+
+        all_simulation_days = []
+        for key, value in data_to_expand.items():
+            info_maps = value.get("info_maps")
+            if info_maps is None:
+                raise TypeError(f"Variable '{key}' has no info maps.")
+            if len(info_maps) != len(value["values"]):
+                raise ValueError(f"Variable '{key}' does not have matching number of values and info maps.")
+            if not all("simulation_day" in info_map.keys() for info_map in info_maps):
+                raise ValueError(f"Variable '{key}' does not have simulation day value in every info map.")
```
allisterakun, line 144: For these errors, do we need to add an error to the OM before raising the exception? Or maybe you'd think that's unnecessary?

RUFAS/util.py
```diff
@@ -83,6 +84,98 @@ def flatten_keys_to_nested_structure(input_dict: Dict[str, Any]) -> Dict[str, An
...
+        filtered_simulation_days = sorted(set(all_simulation_days))
+        first_day = filtered_simulation_days[0]
+        last_day = filtered_simulation_days[-1]
+
+        expanded_data: dict[str, dict[str, list[Any]]] = {}
+        for key, data in data_to_expand.items():
+            expanded_variable_data: dict[str, list[Any]] = {"values": [], "info_maps": []}
+            original_units = data["info_maps"][0]["units"]
+            zipped_data = zip(data["values"], data["info_maps"])
+            indexed_data = {data[1]["simulation_day"]: data for data in zipped_data}
```
allisterakun, line 156: This is definitely not related to this PR, but what are your thoughts on enforcing some kind of data structure for all the add_variable data to follow?

allisterakun (approved):
Looks great, good job Ed!

allisterakun (approved):

RUFAS/util.py
Isgotkowitz: Unfortunately we cannot import the OM into the Utility class because it will cause a circular import error.

RUFAS/util.py
Isgotkowitz: I think that would be great, because all the parsing of dictionaries and zipping together data passed into the expand data method seems very fragile to me.

JoeWaddell (approved):
Great changes - method is useful, and implementation is easy to follow. Good test coverage. Easy approval!

Only final thoughts have to do with the naming of `fill_gap_values` and `fill_end_values`. I'm probably overthinking it, but a naive reader of the method might assume those are values to be used in gaps/the end, and not a boolean choice to use the `fill_value` or not. Likewise, `fill_gap_values` as a question would make me think that False wouldn't actually fill gaps (or the end).

All that said, I don't have a great suggestion for an alternative. `fill_gap_values` could be `use_fill_value_in_gaps` instead, but that's wordier and possibly still unclear (and would also require the flipping of the True/False logic, since here True would use the `fill_value`, and False would use the last known value). Some of this could maybe the decided in the subsequent PRs that use this functionality, e.g. https://github.com/RuminantFarmSystems/MASM/pull/1778

Isgotkowitz:
> Great changes - method is useful, and implementation is easy to follow. Good test coverage. Easy approval!
> 
> Only final thoughts have to do with the naming of `fill_gap_values` and `fill_end_values`. I'm probably overthinking it, but a naive reader of the method might assume those are values to be used in gaps/the end, and not a boolean choice to use the `fill_value` or not. Likewise, `fill_gap_values` as a question would make me think that False wouldn't actually fill gaps (or the end).
> 
> All that said, I don't have a great suggestion for an alternative. `fill_gap_values` could be `use_fill_value_in_gaps` instead, but that's wordier and possibly still unclear (and would also require the flipping of the True/False logic, since here True would use the `fill_value`, and False would use the last known value). Some of this could maybe the decided in the subsequent PRs that use this functionality, e.g. #1778

Good suggestion, thank you @JoeWaddell. I will change the arg names to be more explicit and update the tests accordingly.

tests/test_input_manager.py
Isgotkowitz, line 1: I put this in here to fix a Flake8 violation that snuck onto main somehow.

PR #1777, opened 2024-06-25
Task extraction

### LINKED ISSUE #1713: [Task Manager] Extract Functions from TM.task to reduce complexity and extendability.
## Expected Behavior

TM.task should not violate the flake8 max complexity constraint. 

Additionally, it should be better suited for extension for any future additional tasks that TM needs to run.

## Current Behavior

With recent additions of another task, it exceeds the allowed max complexity.

## Possible Solution

Extract the current functions and look for ways to make it more flexible to additional tasks being set up in TM.

matthew7838:
This PR extracts function into a function map so that TaskManagee.task will not violate the complexity constraint of flake8. Improves future extendability.

### Context

Issue(s) closed by this pull request: closes #1713 

- [ ] Yes
- [x] No

### What

- Tasks are now mapped into two dictionaries, one related to functions that do not need data validations, and one post validations.
- All tasks are moved into a static method and ensure that it has uniform signatures.
- Wrapper method `call_handler` is created to pass into tasks.
- Baseline test for `task ` is updated.
- All other related tasks' methods are tested.

### Why

Violates flake8 complexity restrain.

### How

There will be no more addition of condition statements when trying to add new tasks in the future with a uniform call of the wrapper method.

### Test plan

- Tests were added.
- Run task_manager metadata with sample_tasks.json

### DIFF
--- a/RUFAS/task_manager.py
+++ b/RUFAS/task_manager.py
@@ -322,12 +322,15 @@ class TaskManager:
         }
         task_id = args["task_id"]
         output_manager = OutputManager()
-        pre_validation_handlers = {TaskType.INPUT_DATA_AUDIT: TaskManager._input_data_audit_tasks,
-                                   TaskType.COMPARE_METADATA_PROPERTIES:
-                                       TaskManager._compare_metadata_properties_tasks}
-        post_validation_handlers = {TaskType.HERD_INITIALIZATION: TaskManager._herd_init_tasks,
-                                    TaskType.SIMULATION_SINGLE_RUN: TaskManager._simulation_engine_run_tasks,
-                                    TaskType.POST_PROCESSING: TaskManager._postprocessing_tasks}
+        pre_validation_handlers = {
+            TaskType.INPUT_DATA_AUDIT: TaskManager._input_data_audit_tasks,
+            TaskType.COMPARE_METADATA_PROPERTIES: TaskManager._compare_metadata_properties_tasks,
+        }
+        post_validation_handlers = {
+            TaskType.HERD_INITIALIZATION: TaskManager._herd_init_tasks,
+            TaskType.SIMULATION_SINGLE_RUN: TaskManager._simulation_engine_run_tasks,
+            TaskType.POST_PROCESSING: TaskManager._postprocessing_tasks,
+        }
         try:
             output_manager.run_startup_sequence(
                 LogVerbosity(args["log_verbosity"]),
@@ -344,8 +347,14 @@ class TaskManager:
 
             handler = pre_validation_handlers.get(task_type)
             if handler:
-                TaskManager.call_handler(handler, args=args, input_manager=input_manager, output_manager=output_manager,
-                                         task_id=task_id, produce_graphics=produce_graphics)
+                TaskManager.call_handler(
+                    handler,
+                    args=args,
+                    input_manager=input_manager,
+                    output_manager=output_manager,
+                    task_id=task_id,
+                    produce_graphics=produce_graphics,
+                )
                 return
             # if args["task_type"] == TaskType.INPUT_DATA_AUDIT:
             #     TaskManager.handle_input_data_audit(args, input_manager, output_manager, False)
@@ -372,8 +381,14 @@ class TaskManager:
 
             handler = post_validation_handlers.get(task_type)
             if handler:
-                TaskManager.call_handler(handler, args=args, input_manager=input_manager, output_manager=output_manager,
-                                         task_id=task_id, produce_graphics=produce_graphics)
+                TaskManager.call_handler(
+                    handler,
+                    args=args,
+                    input_manager=input_manager,
+                    output_manager=output_manager,
+                    task_id=task_id,
+                    produce_graphics=produce_graphics,
+                )
                 return
 
             # if args["task_type"] == TaskType.HERD_INITIALIZATION:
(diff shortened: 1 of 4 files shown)


RUFAS/task_manager.py
```diff
@@ -314,6 +326,17 @@ def task(args: Dict[str, Any], produce_graphics: bool, metadata_depth_limit: int
         }
         task_id = args["task_id"]
         output_manager = OutputManager()
+        print(output_manager)
```
ew3361zh: remove print statement.

RUFAS/task_manager.py
```diff
@@ -314,6 +326,17 @@ def task(args: Dict[str, Any], produce_graphics: bool, metadata_depth_limit: int
...
+
+        pre_validation_handlers = {
+            TaskType.INPUT_DATA_AUDIT: TaskManager._input_data_audit_tasks,
```
ew3361zh: I'm not sure I get this distinction between pre_validation_handlers and post_validation_handlers. INPUT_DATA_AUDIT is used for data validation.

RUFAS/task_manager.py
```diff
@@ -509,3 +533,65 @@ def set_random_seed(random_seed: int | None, output_manager: OutputManager) -> N
...
         output_manager.add_log("Random seed used", f"Seeded libaries with {random_seed=}", info_map)
+
+    @staticmethod
+    def _input_data_audit_tasks(
+        args: Dict[str, Any],
+        input_manager: InputManager,
+        output_manager: OutputManager,
+        task_id: Any,
+        produce_grahics: bool,
+    ) -> None:
```
ew3361zh, line 543: Add docstring function definition.

RUFAS/task_manager.py
```diff
@@ -509,3 +533,65 @@ def set_random_seed(random_seed: int | None, output_manager: OutputManager) -> N
...
+        TaskManager.handle_input_data_audit(args, input_manager, output_manager, False)
+        TaskManager.handle_post_processing(args, input_manager, output_manager, task_id)
+
+    @staticmethod
+    def _compare_metadata_properties_tasks(
```
ew3361zh: What do you think about making all these static methods have action verbs in their titles? e.g. _handle_compare_metadata_properties_tasks(...), _handle_herd_init_tasks(...), etc

ew3361zh (approved):
Nice work, Matthew, I think this is in good shape and paves a good way forward for adding new tasks in a manageable and understandable way.

I have a few comments mostly about function naming and adding docstrings but I think the structure looks sound.

RUFAS/task_manager.py
matthew7838: Good catch

matthew7838: After taking a second look, I think I did follow the original code.  it replaces `            if args["task_type"] == TaskType.INPUT_DATA_AUDIT:
                TaskManager.handle_input_data_audit(args, input_manager, output_manager, False)
                TaskManager.handle_post_processing(args, input_manager, output_manager, task_id)
                return` . In the original code, this was also there before the line `  is_data_valid = TaskManager.handle_input_data_audit(args, input_manager, output_manager, True)`

ew3361zh: Ah, I think I just didn't understand which "validation" it was before since it was doing validation itself. If @allisterakun is okay with this naming as well, I'm fine with keeping it as is. Thanks Matthew.

tests/test_task_manager.py
```diff
@@ -224,3 +225,237 @@ def test_input_data_audit(
...
+            mock_handle_post_processing,
+            args=args,
+            input_manager=mock_input_manager,
+            output_manager=mock_output_manager,
+            task_id=task_id,
+            produce_graphics=produce_graphics,
+        )
+
+        # Verify that the handler was called with the correct filtered arguments
+        print(mock_handle_post_processing.call_args)
```
allisterakun:
```suggestion

```

tests/test_task_manager.py
```diff
@@ -224,3 +225,237 @@ def test_input_data_audit(
...
+
+    mock_handler = mocker.patch.object(TaskManager, "call_handler", return_value=None)
+    mock_handle_input_data_audit = mocker.patch.object(TaskManager, "handle_input_data_audit", return_value=True)
+    mock_set_random_seed = mocker.patch.object(TaskManager, "set_random_seed", return_value=None)
+    task_manager.task(args, produce_graphics, 10)
+    mock_im_init.assert_called_once_with(10)
+
+    if pre_validate:
+        mock_handler.assert_called_once()
+        # print(mock_handler.call_args)
```
allisterakun:
```suggestion
```

tests/test_task_manager.py
```diff
@@ -224,3 +225,237 @@ def test_input_data_audit(
...
+    else:
+        mock_handle_input_data_audit.assert_called_once()
+        mock_set_random_seed.assert_called_once()
+        mock_handler.assert_called_once()
+        # print(mock_handler.call_args)
```
allisterakun:
```suggestion
```

RUFAS/task_manager.py
allisterakun: Since we're changing all the "validation" to "input data audit," I suggest keeping the naming consistent.

allisterakun (approved):
Looks good, Matthew! Just remove some comments and print statements in the tests.

PR #1778, opened 2024-06-25
Align data in graphs and reports

### LINKED ISSUE #1366: Misalignment in graphs (and reports) due to instantiation of new variables during simulation

Isgotkowitz:
This PR integrates the new data padding capabilities in `Utility` into the graph and report generators.

### Context

Issue(s) closed by this pull request: closes #1366

- [ ] Yes
- [x] No

Note: I will update the Wiki when this PR is approved and it is ready to merge.

### What

This PR modifies the `filter_variables_pool` method in the Output Manager so that it will make use of the two new keys in 
that can be specified in Graph and Report filters: `expand_data`, `expand_gap_values`, `expand_end_values`.

### Why

This allows users to _see_ when data was recorded in relation to other data.

### How

`filter_variables_pool` now checks filter contents for `expand_data`, if it finds it to be true then it uses `expand_data_temporally` in the Utility class to expand the filtered data before slicing it and sending it to the Report Generator. In order to keep this method from becoming overly complex, the functionality for counting variables and unpacking/filtering dictionary keys has been moved into a subroutine, `_parse_filtered_variables`. The order of operations in `filter_variables_pool` is now: break down dictionaries and count variables -> pad data if specified -> slice data.

`expand_data`, `fill_gap_values`, and `fill_end_values` have been added to the Graph Generator as optional keys. Interestingly, slices can be specified for graph filters and will work, but are not officially supported as graph keys - maybe this should be changed.

### Test plan

I have updated the OM's test suite, and written a test for `_parse_filtered_variables`. All lines changed are covered by the test suite, but it is important to note that a lot of the tests for `filter_variables_pool` are integration tests. For now I think this ok, because the OM seems quite well tested.

To test the new functionality of data padding, I have been comparing the results in the below graph and report filters. It appears that everything is functioning as expected, with the exception of graphing with `fill_gap_values` being false (the graphs in this case seem to go haywire) - but I think this is acceptable behavior right now, for the graphing capabilities we have available.
Graph:
```json
{
    "multiple": [
        {
            "title": "Unpadded Population and feed totals",
            "type": "plot",
            "filters": [
                "AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_3_LAC_COW",
                "AnimalModuleReporter.report_daily_pen_total.number_of_animals_in_pen_3_LAC_COW"
            ],
            "slice_start": -365,
            "variables": ["DE"]
        },
        {
            "title": "Padded Population and feed totals",
            "type": "plot",
            "filters": [
                "AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_3_LAC_COW",
                "AnimalModuleReporter.report_daily_pen_total.number_of_animals_in_pen_3_LAC_COW"
            ],
            "variables": ["DE"],
            "slice_start": -365,
            "expand_data": true,
            "mask_values": true,
            "use_fill_value_in_gaps": true,
            "use_fill_value_at_end": false
        },
        {
            "title": "Tail Padded Population and feed totals",
            "type": "plot",
            "filters": [
                "AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_3_LAC_COW",
                "AnimalModuleReporter.report_daily_pen_total.number_of_animals_in_pen_3_LAC_COW"
            ],
            "variables": ["DE"],
            "slice_start": -365,
            "expand_data": true,
            "mask_values": true,
            "use_fill_value_in_gaps": true,
            "use_fill_value_at_end": true
        },
        {
            "title": "Tail Padded Population and feed totals, gaps not filled",
            "type": "plot",
            "filters": [
                "AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_3_LAC_COW",
                "AnimalModuleReporter.report_daily_pen_total.number_of_animals_in_pen_3_LAC_COW"
            ],
            "variables": ["DE"],
            "slice_start": -365,
            "expand_data": true,
            "use_fill_value_in_gaps": false,
            "use_fill_value_at_end": true
        }
    ]
}
```
Report:
```json
{
    "multiple": [
        {
            "name": "Unpadded Population and feed totals",
            "filters": [
                "AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_3_LAC_COW",
                "AnimalModuleReporter.report_daily_pen_total.number_of_animals_in_pen_3_LAC_COW"
            ],
            "variables": ["DE"],
            "slice_start": -365,
            "expand_data": false,
            "graph_and_report": true,
            "graph_details": {
                "title": "Unpadded Population and feed totals",
                "type": "plot"
            }
        },
        {
            "name": "Padded Population and feed totals, gaps filled",
            "filters": [
                "AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_3_LAC_COW",
                "AnimalModuleReporter.report_daily_pen_total.number_of_animals_in_pen_3_LAC_COW"
            ],
            "variables": ["DE"],
            "slice_start": -365,
            "expand_data": true,
            "use_fill_value_in_gaps": true,
            "use_fill_value_at_end": false,
            "graph_and_report": true,
            "graph_details": {
                "title": "Padded Population and feed totals",
                "type": "plot",
                "mask_values": true
            }
        },
        {
            "name": "Tail Padded Population and feed totals, gaps filled",
            "type": "plot",
            "filters": [
                "AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_3_LAC_COW",
                "AnimalModuleReporter.report_daily_pen_total.number_of_animals_in_pen_3_LAC_COW"
            ],
            "variables": ["DE"],
            "slice_start": -365,
            "expand_data": true,
            "use_fill_value_in_gaps": true,
            "use_fill_value_at_end": true,
            "graph_and_report": true,
            "graph_details": {
                "title": "Tail Padded Population and feed totals, gaps filled",
                "type": "plot",
                "mask_values": true
            }
        },
        {
            "name": "Tail Padded Population and feed totals, gaps not filled",
            "type": "plot",
            "filters": [
                "AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_3_LAC_COW",
                "AnimalModuleReporter.report_daily_pen_total.number_of_animals_in_pen_3_LAC_COW"
            ],
            "variables": ["DE"],
            "slice_start": -365,
            "expand_data": true,
            "use_fill_value_in_gaps": false,
            "use_fill_value_at_end": true,
            "graph_and_report": true,
            "graph_details": {
                "title": "Tail Padded Population and feed totals, gaps not filled",
                "type": "plot"
            }
        },
        {
            "name": "Padded Population and feed totals, gaps not filled",
            "type": "plot",
            "filters": [
                "AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_3_LAC_COW",
                "AnimalModuleReporter.report_daily_pen_total.number_of_animals_in_pen_3_LAC_COW"
            ],
            "variables": ["DE"],
            "slice_start": -365,
            "expand_data": true,
            "use_fill_value_in_gaps": false,
            "use_fill_value_at_end": false,
            "graph_and_report": true,
            "graph_details": {
                "title": "Padded Population and feed totals, gaps not filled",
                "type": "plot"
            }
        }
    ]
}
```
Edit: Joe found an error in a few graphs where data was not being properly graphed when gaps were filled with fill values. To fix this, I added a masking option to graphs which will prevent matplotlib from trying to graph None/NaN values. This change is primarily located in the `_draw_graph` and `_mask_values` methods of Graph Generator. Masking is controlled by an option called `mask_values` in graph filters, and I have added this where appropriate to the above graph and report filters.


RUFAS/output_manager.py
```diff
@@ -976,22 +976,75 @@ def filter_variables_pool(self, filter_content: Dict[str, Any]) -> Dict[str, poo
...
+        selected_variables : List[str] | None
+            List of key names to select or exclude from variables containing dictionaries.
+        filter_name : str
+            Name of the filter used to collect variables for the filtered pool.
+        use_filter_name : bool
+            Whether to use the filter name when constructing the key name for data pulled from a dictionary.
+        filter_by_exclusion : bool
+            Whether keys in dictionaries should be filtered by exclusion.
+
+        """
```
matthew7838: Missing returns

matthew7838 (approved):
LGTM, the testing looks comprehensive, and the results also look good on my side.

RUFAS/output_manager.py
Isgotkowitz: Nice catch

JoeWaddell (commented):
Using the provided graphics filter, the one labeled `Tail Padded Population and feed totals` doesn't graph the DE values. The other two look to work as expected (though I had to add the `slice_start` argument to the last one in the filter).

image

RUFAS/graph_generator.py
```diff
@@ -527,6 +540,9 @@ def _draw_graph(
         selected_variables : Optional[List[str]]
             If it is present and the data is a list of dicts,
             it selects the variables to be plotted.
+        mask_values : bool, default False
+            Whether data that will be plotted with non-tuple based functions should be masked to remove None or NaN
+            values.
```
matthew7838, line 546: Should non-tuple-based functions be forced to be masked? If so, maybe we should ignore the `mask_values` and enforce the masking regardless when plotting with non-tuple based functions.

matthew7838 (approved):
LGTM, found a small thing that I would like to know about your inputs

RUFAS/graph_generator.py
Isgotkowitz: That is a great question. I haven't worked much with the tuple-based functions, but I'm pretty sure a lot of them don't work super well and are just generally very finicky, so I was hoping to leave them alone. I will play around with them a bit though and see if masking would be appropriate

Isgotkowitz: So looking at it just now, it seems that masking data for tuple-based plotting functions would be tricky but not undoable. That said, I think it would be better to make that a separate issue and have it be worked on at a later point

JoeWaddell (approved):
Had to make a tweak locally for it to work as expected: providing a `fill_value` argument in the graph filter didn't make its way to the filter variables pool. Quick change though, and this works nicely!
image

Not in the screenshot, but probably obvious: this required adding numpy to the `output_manager.py` imports

Isgotkowitz:
I will add in the fill-value option to both the report and graph filters.

PR #1735, opened 2024-06-03
Revert "Migrating AnimalRequirements-specific unit tests to dedicated test file."

Isgotkowitz:
Reverts RuminantFarmSystems/MASM#1528

Some tests were failing when this PR was merged. These changes should be reverted, fixed, then they can be re-reviewed and merged.

PR #1780, opened 2024-06-25
Refactor of test_animal.py

Fjordrunner:
This PR moves certain unit tests in the animal module to an animal_requirements-specific test file. Additionally, this PR cleaned up various MyPy errors throughout the general test_animal.py file. 

This PR is in response to PR #1735, which reverted a bugged PR that attempted to move unit tests to an animal_requirements-specific test file.

- [ ] Yes
- [x] No

### What
Some unit tests were migrated to an animal_requirements.py-specific test file. Additionally, multiple changes were made throughout the model, mostly incorporating type hints, to reduce the overall amount of Mypy errors in the test_animal.py file.

### Why
This PR increases our test modularity, reduces the length of the test_animal.py file, and reduces the amount of MyPy errors by adhering to best coding practices.

### Test plan
Multiple changes were made to test functions, resulting in many of them failing. All failing tests have been assessed and are now passing.


RUFAS/routines/animal/life_cycle/animal_base.py
```diff
@@ -167,7 +168,7 @@ def set_p_purchased(self) -> None:
         # (A.1G.C.1) from P tracking
         self.p_animal = 0.0072 * self.body_weight * GeneralConstants.KG_TO_GRAMS
 
-    def update_pen_history(self, curr_pen, curr_day, classes_in_pen):
+    def update_pen_history(self, curr_pen: int, curr_day: int, classes_in_pen: Set[str]):
```
Isgotkowitz:
```suggestion
    def update_pen_history(self, curr_pen: int, curr_day: int, classes_in_pen: Set[str]) -> None:
```

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -158,13 +158,17 @@ def calc_pen_requirements(
...
+                attribute_name,
+                calc_method_to_function_map[calc_method](arg, *stats_args),
+            )
 
     def set_requirements(self, pen, animal_grouping_scenario, recalc: bool):
         """
         Calculates the average requirements utilizing cow_requirements.py and an
         input pen to generate the average requirements across a pen. It then
-        populates the corresponding class variables.
+        populates the corresponding class variables.s
```
Isgotkowitz:
```suggestion
        populates the corresponding class variables.
```

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -337,7 +346,12 @@ def recalculate_requirements(self, pen, animal_grouping_scenario, requirements_l
             requirements_lists["BW"].append(animal.body_weight)
         return requirements_lists
 
-    def use_existing_requirements(self, pen, animal_grouping_scenario, requirements_lists: Dict):
+    def use_existing_requirements(
+        self,
+        pen,
+        animal_grouping_scenario,
+        requirements_lists: Dict[str, List[int]],
+    ):
```
Isgotkowitz, line 354: Can any type hints be added here?

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -232,7 +236,12 @@ def set_requirements(self, pen, animal_grouping_scenario, recalc: bool):
 
         pen.set_milk_avgs(self.avg_milk, self.avg_CP_milk, self.avg_milk_production_reduction)
 
-    def recalculate_requirements(self, pen, animal_grouping_scenario, requirements_lists: Dict):
+    def recalculate_requirements(
+        self,
+        pen,
+        animal_grouping_scenario,
+        requirements_lists: Dict[str, List[int]],
+    ):
```
Isgotkowitz, line 244: Can any more type hints be added here?

RUFAS/routines/animal/clustering_pen_grouping.py
```diff
@@ -165,3 +168,4 @@ def grouping(cow_list, pens, stocking_density):
 
     # returning the dictionary of groupings
     return grouping_data
+    return grouping_data
```
Isgotkowitz:
```suggestion
```

RUFAS/routines/animal/ration/calf_ration.py
```diff
@@ -310,3 +314,4 @@ def calc_intake(
         }
 
         return animal_intake
+        return animal_intake
```
Isgotkowitz:
```suggestion
```

RUFAS/routines/animal/ration/user_defined_ration.py
```diff
@@ -92,3 +96,4 @@ def make_ration_from_user_values(
         ration["status"] = "Optimal"
         ration["objective"] = 0.0
         return ration
+        return ration
```
Isgotkowitz:
```suggestion
```

RUFAS/routines/animal/ration/ration_optimizer.py
```diff
@@ -897,7 +898,7 @@ def make_user_bounds(
     def optimize(
         self, animal_combination: AnimalCombination,
         ration_config: RationConfig,
-        previous_ration: Dict[str, float | str] | None = None
+        previous_ration: Mapping[str, float | str] | None = None
```
Isgotkowitz:
```suggestion
        previous_ration: Dict[str, float | str] | None = None
```
Instead of changing this to `Mapping`, I suggest keeping it a `Dict` and adding the following at line 934: `value = int(value)`. The benefit is that it keeps the typing of `previous_ration` more consistent with the way it is typed in other places, and doesn't add the error associated with `Mapping`.

tests/animal_module_tests/test_animal.py
```diff
@@ -1187,6 +1376,7 @@ def test_update_body_weight_history(cow_fixture: AnimalBase) -> None:
         cow_fixture.update_body_weight_history(sim_day)
     # asserts for each update
     for idx, history in enumerate(histories):
+        print(idx, history)
```
Isgotkowitz:
```suggestion
```

tests/animal_module_tests/test_animal.py
```diff
@@ -2632,28 +2877,30 @@ def test_report_ration() -> None:
...
-def eq_constraint(x, ration_config):
-    return np.sum(x) - 10  # This 'eq' constraint checks if the sum of x is equal to 10
-
-
-def ineq_constraint(x, ration_config):
-    return np.sum(x) - 10  # This 'ineq' constraint checks if the sum of x is greater than 10
+def dummy_constraint(x: npt.NDArray[np.int16], _: MagicMock):
+    return np.sum(x) - 10
+    # This dummy constraint checks if the difference of the sum of x and 10
+    #  which is_constraint_violated() then handles
```
Isgotkowitz: I think there is some missing info here.

tests/animal_module_tests/test_animal.py
```diff
@@ -3255,6 +3514,8 @@ def test_get_DE(
         return_value=0.5,
     )
     actual = RationReporter.get_DE(kg_fed, feed_item_info, ration_report, body_weight)
+    print(actual)
+    print(expected)
```
Isgotkowitz:
```suggestion
```

tests/animal_module_tests/test_animal.py
```diff
@@ -3293,10 +3554,10 @@ def test_get_DE(
...
 )
 def test_get_ME(
-    kg_fed: float,
-    feed_item_info: dict,
-    ration_report: dict,
-    body_weight: float,
+    kg_fed: int,
+    feed_item_info: Dict[str, str | int],
+    ration_report: str,
+    body_weight: str,
```
Isgotkowitz: The type hints for these params do not match `get_ME` expects.

tests/animal_module_tests/test_animal.py
```diff
@@ -3318,10 +3579,10 @@ def test_get_ME(
...
 )
 def test_get_NE_maintenance_and_activity(
-    kg_fed: float,
-    feed_item_info: dict,
-    ration_report: dict,
-    body_weight: float,
+    kg_fed: int,
+    feed_item_info: Dict[str, str | bool],
+    ration_report: str,
+    body_weight: str,
```
Isgotkowitz: Same issue as above, these param types do not match what is expected by `get_NE_maintenance_and_activity`

tests/animal_module_tests/test_animal.py
```diff
@@ -3382,10 +3643,10 @@ def test_get_NE_maintenance_and_activity(
...
 )
 def test_get_NE_lactation(
-    kg_fed: float,
-    feed_item_info: dict,
-    ration_report: dict,
-    body_weight: float,
+    kg_fed: int,
+    feed_item_info: Dict[str, str | bool | int],
+    ration_report: str,
+    body_weight: str,
```
Isgotkowitz: Same as issues above, these param types do not match what `get_NE_lactation` expects

tests/animal_module_tests/test_animal.py
```diff
@@ -3443,10 +3704,10 @@ def test_get_NE_lactation(
...
 )
 def test_get_NE_growth(
-    kg_fed: float,
-    feed_item_info: dict,
-    ration_report: dict,
-    body_weight: float,
+    kg_fed: int,
+    feed_item_info: Dict[str, str | bool],
+    ration_report: str,
+    body_weight: str,
```
Isgotkowitz: Same as above, param type hints do not match what is expected by `get_NE_growth`.

tests/animal_module_tests/test_animal.py
```diff
@@ -3493,10 +3754,10 @@ def test_get_NE_growth(
...
 )
 def test_get_calcium(
-    kg_fed: float,
-    feed_item_info: dict,
-    ration_report: dict,
-    body_weight: float,
+    kg_fed: int,
+    feed_item_info: Dict[str, str | int],
+    ration_report: str,
+    body_weight: str,
```
Isgotkowitz: Same as above, param type hints do not match what is expected by `get_calcium`.

tests/animal_module_tests/test_animal.py
```diff
@@ -3538,10 +3799,10 @@ def test_get_calcium(
...
 )
 def test_get_phosphorus(
-    kg_fed: float,
-    feed_item_info: dict,
-    ration_report: dict,
-    body_weight: float,
+    kg_fed: int,
+    feed_item_info: Dict[str, str | int],
+    ration_report: str,
+    body_weight: str,
```
Isgotkowitz: Same as above, type params do not match what is expected by `get_phosphorus`

tests/animal_module_tests/test_animal.py
```diff
@@ -3607,19 +3868,20 @@ def test_get_fat_percentage(
...
 )
 def test_get_forage_NDF(
-    kg_fed: float,
-    feed_item_info: dict,
-    ration_report: dict,
-    body_weight: float,
+    kg_fed: int,
+    feed_item_info: Dict[str, str | int],
+    ration_report: str,
+    body_weight: str,
```
Isgotkowitz: Same as above, the type hints of these params do not match what is expected by `get_forage_NDF`

tests/animal_module_tests/test_animal.py
```diff
@@ -3704,14 +3966,24 @@ def test_get_metabolizable_protein(mocker: MockerFixture) -> None:
...
     body_weight = 100
 
     mocker.patch(
         "RUFAS.routines.animal.ration.ration_driver.RationReporter.get_TDN_discount",
         return_value=1,
     )
     expected = 171.1937767245963
-    actual = RationReporter.get_metabolizable_protein(ration, available_feeds, ration_report, body_weight)
+    actual = RationReporter.get_metabolizable_protein(ration, mock_avail_feeds, ration_report, body_weight)
+    assert np.isclose(actual, expected, rtol=1e-3)
```
Isgotkowitz, line 3988:
```suggestion
```

Isgotkowitz (changes requested):
Good work, Simon, love seeing that error count go down. Left comments on a lot of small things, address as you feel appropriate. Changes requested are specific to a handful of tests in `test_animal.py` where the type hints of the test parameters do not match the method being tested, each has a comment.

I'm also curious, are you running the formatter manually? I saw a couple of formatting changes that I was unable to get Black to reproduce.

RUFAS/routines/animal/life_cycle/animal_base.py
Fjordrunner: Good catch, thank you! This is now added.

RUFAS/routines/animal/ration/animal_requirements.py
Fjordrunner: Changed.

RUFAS/routines/animal/clustering_pen_grouping.py
Fjordrunner: This doubling of return statements happened a few other places...didn't catch this one, thank you! Changed.

RUFAS/routines/animal/ration/calf_ration.py
Fjordrunner: Done.

RUFAS/routines/animal/ration/user_defined_ration.py
Fjordrunner: Done.

RUFAS/routines/animal/ration/ration_optimizer.py
Fjordrunner: Good point, thank you! I think adding Mapping removed a Mypy error in the test_animal.py file but I did wince a bit when I added it.

tests/animal_module_tests/test_animal.py
Fjordrunner: Done.

tests/animal_module_tests/test_animal.py
Fjordrunner: Done.

tests/animal_module_tests/test_animal.py
Fjordrunner: Removed the double assert at the end here.

RUFAS/routines/animal/ration/animal_requirements.py
Fjordrunner: Adding type hints for AnimalGroupingScenario and Pen objects results in circular imports. There are some clever workarounds to these by using the "importlib" library, but I just spoke with Pooya, and since these files' structures are going to get heavily restructured soon, we will let this be for now.

RUFAS/routines/animal/ration/animal_requirements.py
Fjordrunner: See my response above. Good point though, it was definitely something I wanted to address!

tests/animal_module_tests/test_animal.py
Fjordrunner: I rewrote it as this, let me know if you have any more suggestions!


    # This dummy constraint returns the difference of the sum of x and 10
    #  which is_constraint_violated() then handles to return whether or not a constraint
    #  has been violated. The two constraints will either check to see whether the
    #  aforementioned difference is 0, in the case of an equality constraint, or greater
    #  than 0, in the case of an inequality constraint.

tests/animal_module_tests/test_animal.py
Fjordrunner: Depending on whether or not you want the values of the tests changed as well, this has been fixed. Thank you!

tests/animal_module_tests/test_animal.py
Isgotkowitz: Looking at it now, I think it would be simpler and cleaner just to give it a docstring like
```python
"""Simple constraint used to test is_constraint_violated()."""
```
because it is not very complicated, and doesn't need anything more than a one-liner.

Isgotkowitz (approved):
LGTM, nice job Simon

RUFAS/routines/animal/ration/ration_optimizer.py
```diff
@@ -929,6 +930,7 @@ def optimize(
             x0 = []
             prev_ration = previous_ration.copy()
             for key, value in prev_ration.items():
+                value = int(value)
```
JoeWaddell: Not sure why this is being changed to an int - just to round it as a starting place? My instinct is that keeping it as the previous float should be fine.

This also causes the simulation to fail: as it can't finish this optimize method since it's trying to convert strings to ints (in the cases where value is `status` or `objective`) . Removing it causes the simulation to complete, so I think the simple removal is sufficient.

RUFAS/routines/animal/ration/ration_optimizer.py
```diff
@@ -1096,7 +1098,7 @@ def attempt_optimization(
 
     @staticmethod
     def is_constraint_violated(
-        solution_x: npt.NDArray,
+        solution_x: npt.NDArray[np.float32],
```
JoeWaddell: Is there are reason some of these are typed as `float32` instead of `float64`?

RUFAS/routines/animal/clustering_pen_grouping.py
```diff
@@ -1,26 +1,29 @@
...
 from scipy.stats import percentileofscore
 
 
-def norm(x):
+def norm(x: List[int | float]) -> npt.NDArray[np.float64 | np.int64]:
     """
     Helper function to normalize a list of values and return that normalized
     list.
 
     Args:
```
JoeWaddell: This could be updated to `Parameters` with the dashes below, likewise the indentation of the arguments/parameters should be updated to follow format elsewhere.

JoeWaddell (changes requested):
Left a few comments, biggest concern is that the simulation fails unless you remove a specific new line - see my comment in ration_optimizer, I think the new line isn't necessary, so once it's removed should be OK

RUFAS/routines/animal/ration/ration_optimizer.py
Fjordrunner: Good point, switching that throughout the file now.

RUFAS/routines/animal/clustering_pen_grouping.py
Fjordrunner: Good point, I'll change that now.

RUFAS/routines/animal/ration/ration_optimizer.py
Fjordrunner: This was changed in an effort to steer away from using the Mapping type hint as opposed to the Dict one, but in hindsight, the optimal solution here is just to maintain the old code completely and live with an extra Mypy warning. Good catch!

JoeWaddell (approved):
Looks great! Nice work tidying things. To test, I compared it to main using both user defined and automated ration formulation, and behavior is identical.

Issue #1507, opened 2024-04-09
[Weather] Write up design doc for proposed changes to simplify Weather data structure.


ew3361zh:
As noted in #753, the structure of the weather data is too complex for ease of use with RuFaS in its current foramt. It should be refactored but may need a design document to cover all the affected aspects.

## Urgency Level

- [ ] Urgent: a critical feature or fix is needed
- [x] High: the issue significantly impacts the user experience
- [ ] Medium: the issue affects functionality but has a workaround
- [ ] Low: the issue doesn't have a significant immediate impact
- [ ] Wishful: the issue should not be prioritized over any others
- [ ] Unsure: the urgency level has not been or cannot be assessed at this time 

## Status

Currently, all the weather data are stored in separate 2D arrays of the following format:
variable_name: [year_num][day]
Where variable_name is the variable of interest (e.g., T_max => maximum temperature of the day), year_num is the difference between the year of interest and the starting year (e.g., for a simulation starting in 2019, the year 2020 would be index 1), and finally the Julian date.

## Intention

The current weather data structure introduces unnecessary complexity, and it should be flattened or completely reformatted.

## Request

Start with a design document.
As a starting point, here are some suggestions from ChatGPT as to how to approach the refactor of the current data structure:

## 1. Flattening to a Single DataFrame
One effective approach is to flatten your 2D arrays into a single pandas DataFrame, where each row represents a day, and columns represent the different weather variables for that day. Additional columns for the year, Julian day, and a constructed date column can make analysis more intuitive.

Steps to Flatten:

Convert 2D Arrays to DataFrame: Convert each 2D array into a DataFrame with year and day as indices.
Merge DataFrames: Merge these individual DataFrames into a single DataFrame where each variable becomes a column.
Add Date Information: Calculate the actual date from the year and day information for easier time series analysis.

Pros:

Simplifies access to the data by consolidating it into a single structure.
Facilitates operations like filtering, aggregation, and visualization.

Cons:

Memory Usage: A single DataFrame containing all variables for all days and years can become large, especially for long-term simulations or high-resolution data, potentially leading to high memory usage.
Performance: Operations on a very large DataFrame can be slower, especially if not properly indexed.
Complexity in Handling Different Data Types: If your weather data includes various types (e.g., temperatures as floats, weather conditions as strings), managing these within a single DataFrame can sometimes complicate processing and visualization.

## 2. Using MultiIndex DataFrame
[...]

PR #1781, opened 2024-06-26
Weather.py refactor

### LINKED ISSUE #1764: [Weather] Refactor Weather class according to design doc.
Once the design doc from issue #1507 if finalized, use it to refactor the Weather class.

## Expected Behavior

See design-doc:
https://docs.google.com/document/d/1QUTRhsaMCJZRekaHSY05hPyxMoRl-Xw4Wq9xrHXwn4A/edit

## Current Behavior

See design-doc for details.

## Possible Solution

See design doc for details.

matthew7838:
Changes the structure of the `Weather` data class.

### Context

Issue(s) closed by this pull request: closes #1764 

- [ ] Yes
- [x] No

### What

- The data will now be a dictionary. A datetime timestamp is the key and the corresponding CurrentWeatherCondition will serve as the value.
- A float variable initialized to hold the mean_annual_temperature after data creation.
- Added helper methods in `Time` to help convert the day.
- A helper method to check the sufficient of the weather data provided.
- Methods related to getting conditions will simply use the key to retrieve data from the dictionary.

### Why

The current format introduces unnecessary complexity in trying to not only extract Weather data for a particular date or period of time but also in trying to assemble Weather class instances from input data. This can (and has) led to data access errors most notably within the Crop and Soil module. 

### How

The structure of Weather objects is flattened so that they’re easier to access, work with, and understand.

### Test plan

1. Tests created
2. Run with different metadata and make sure that they're still running as expected.


RUFAS/weather.py
```diff
@@ -16,33 +15,23 @@ class Weather:
     """
     The `Weather` class manages all weather data used to run a single simulation.
 
+    Parameters
+    ----------
+    weather_file : dict[str:List[Any]]
```
allisterakun:
```suggestion
    weather_file : dict[str, List[Any]]
```
just to be consistent with the rest of the codebase

RUFAS/weather.py
```diff
@@ -16,33 +15,23 @@ class Weather:
...
-    __min_daily_temperature : list
-        Minimum temperatures of days (°C).
-    __mean_daily_temperature : list
-        Mean temperatures of days (°C).
-    __radiation : list
-        Energy from the sun (MJ/m^2).
-    __irrigation : list
-        Amounts of irrigation applied to fields on given days (mm).
-    __mean_annual_temperature : float
+    weather_data : dict[datetime:CurrentDayCondition]
```
allisterakun:
```suggestion
    weather_data : dict[datetime, CurrentDayCondition]
```

RUFAS/weather.py
```diff
@@ -357,3 +270,42 @@ def _get_latitude() -> float:
...
+        Returns
+        -------
+        None
+
+        """
+        years_list = weather_file["year"]
+        days_list = weather_file["jday"]
+        current_date = time.start_date
+
+        while current_date != time.end_date:
```
allisterakun: Since we know the start and the end date, can we use a for loop instead?

RUFAS/weather.py
```diff
@@ -357,3 +270,42 @@ def _get_latitude() -> float:
...
+            current_date_year = current_date.timetuple().tm_year
+            current_date_jday = current_date.timetuple().tm_yday
+            if (current_date_jday in days_list) and (current_date_year in years_list):
```
allisterakun, line 309: If we are validating by checking if we have a piece of weather data for all the simulation dates, how do we account for duplicate data? For example, what should we do with two different weather data entries for the same date?

RUFAS/weather.py
```diff
@@ -158,28 +100,24 @@ def get_current_day_conditions(self, time: Time) -> CurrentDayConditions:
...
-                incoming_light=self.__radiation[year - 1][day - 1],
-                min_air_temperature=self.__min_daily_temperature[year - 1][day - 1],
-                mean_air_temperature=self.__mean_daily_temperature[year - 1][day - 1],
-                max_air_temperature=self.__max_daily_temperature[year - 1][day - 1],
-                annual_mean_air_temperature=self.__mean_annual_temperature,
-                precipitation=self.__precipitation[year - 1][day - 1],
-                irrigation=self.__irrigation[year - 1][day - 1],
-                daylength=daylength,
+            self.weather_data[time.current_date].daylength = daylength
+            self.weather_data[time.current_date].annual_mean_air_temperature = self.mean_annual_temperature
```
allisterakun, line 122: Since we are using the same latitude for all calls, and the CurrenDayConditions is essentially queried every day  throughout the simulation (`AnimalManager.daily_updates()`), can we move these assignment into `__init__()`? I think this would make more sense than generating on the fly.

allisterakun (changes requested):
Looks pretty good, Matthew! There are just some potential changes in the comments. 

However, please remove the following properties from the `Time` class (https://github.com/RuminantFarmSystems/MASM/issues/1507#issuecomment-2127449078). These functions were to be deprecated during the `Time` class refactor, but they were only kept because they were used in the old `Weather` initialization.
- `Time.years`
- `Time.start_year_int`
- `Time.start_day`
- `Time.end_year_int`
- `Time.end_year`

RUFAS/weather.py
matthew7838: This is a good point. However, I think the current implementation would serve better for an extension if we moved on to a more geographically precise solution in the future.

RUFAS/weather.py
matthew7838: I think this should be fine. later on, with the transformation into a dictionary we will only keep the latest read of weather data from that day. I think it's a safe assumption for now to assume it is the newest data and should follow closely with the old code.

RUFAS/time.py
```diff
@@ -198,6 +121,27 @@ def convert_simulation_day_to_date(self, simulation_day: int) -> datetime.date:
...
+        ----------
+        year: int
+            Year of the date.
+        day: int
+            Number of days into the year.
+
+        Returns
+        -------
+        datetime
+            The date time object from the provided inputs
```
ew3361zh:
```suggestion
            The datetime object from the provided inputs.
```

RUFAS/time.py
```diff
@@ -198,6 +121,27 @@ def convert_simulation_day_to_date(self, simulation_day: int) -> datetime.date:
         actual_date = self.start_date + datetime.timedelta(days=simulation_day - 1)
         return actual_date
 
+    @staticmethod
+    def convert_year_jday_to_date(year: int, day: int) -> datetime:
+        """
+        Converts the year and its day of the year to a datetime object
```
ew3361zh:
```suggestion
        Converts the year and its day of the year to a datetime object.
```

RUFAS/weather.py
```diff
@@ -16,33 +15,23 @@ class Weather:
     """
     The `Weather` class manages all weather data used to run a single simulation.
 
+    Parameters
+    ----------
+    weather_file : dict[str, List[Any]]
+        The weather dictionary read from the provided weather file.
```
ew3361zh: I think make this more general since it won't necessarily always be a file we're pulling weather from.

```suggestion
        The weather dictionary read from the provided weather input source.
```

RUFAS/weather.py
```diff
@@ -16,33 +15,23 @@ class Weather:
...
-        Minimum temperatures of days (°C).
-    __mean_daily_temperature : list
-        Mean temperatures of days (°C).
-    __radiation : list
-        Energy from the sun (MJ/m^2).
-    __irrigation : list
-        Amounts of irrigation applied to fields on given days (mm).
-    __mean_annual_temperature : float
+    weather_data : dict[datetime, CurrentDayCondition]
+        A dictionary that map a date to the corresponding CurrentDayCondition.
```
ew3361zh:
```suggestion
        A dictionary that maps a date to the corresponding CurrentDayCondition.
```

RUFAS/weather.py
```diff
@@ -357,3 +270,43 @@ def _get_latitude() -> float:
...
+
+            if (current_date_jday in days_list) and (current_date_year in years_list):
+                continue
+            else:
+                info_map = {
+                    "class": "Weather",
+                    "function": Weather.check_adequate_weather_data.__name__,
+                }
+                om.add_error(
+                    f"Dumping all logs from main.py because of error '{ValueError}'",
```
ew3361zh:
```suggestion
                    "Inadequate weather data.",
```

RUFAS/weather.py
```diff
@@ -200,42 +138,17 @@ def get_conditions_series(self, time: Time, starting_offset: int, ending_offset:
...
-            elif year == starting_year_index:
-                start_day = starting_day_index
-                end_day = len(self.__mean_daily_temperature[year]) - 1
-            elif year == ending_year_index:
-                start_day = 0
-                end_day = ending_day_index
-            else:
-                start_day = 0
-                end_day = len(self.__mean_daily_temperature[year]) - 1
+        condition_list = []
```
ew3361zh: And other refs to this var.
```suggestion
        conditions_list = []
```

RUFAS/weather.py
```diff
@@ -357,3 +270,43 @@ def _get_latitude() -> float:
...
+
+    @staticmethod
+    def check_adequate_weather_data(weather_file: dict, time: Time) -> None:
+        """
+        Checks that there is enough weather data to cover the whole simulation time.
+
+        Parameters
+        ----------
+        weather_file: dict
+            A dictionary form of the weather file.
```
ew3361zh: I think these should be updated to be more generic since we won't necessarily be using a file as our weather data source.

RUFAS/weather.py
ew3361zh: Yea I'm not sure what we'd do about duplicate data. It does seem like something to at least log as a warning to OM. One or two duplicates seems marginal but if there's a whole second set of data that's duplicated, it seems like it could have a real impact on the simulation if it's different.

RUFAS/weather.py
ew3361zh: Good question, Allister - I think I agree with Matthew here on the possibility for extension when we consider a potential farm or simulation with multiple latitudes and wanting to be able to calculate daylight differently for each of them. We may end up seeing that the queries are too resource consuming and it's better to just make a set of weather data with CurrentDayConditions for each set of latitudes. It seemed like overall a 10-year set of weather data was not that memory-intensive so 2 sets wouldn't be much either.

ew3361zh (approved):
I think this looks great, Matthew! Left some comments and suggestions on docstrings mostly but the structure looks really good imo and I think you did a good job implementing the design doc and taking some of the open ended questions into account.

A couple questions: 
1. I noticed you didn't have to update any of the files/modules that use weather data/current day conditions. Did you look through those references and were there any improvements to how weather data is used you might suggest?
2. I know we had an issue with not enough weather data for some C&S simulations. Were you able to test those and ensure this was catching those shortcomings effectively?

RUFAS/weather.py
matthew7838: Sounds like a good measure to log a warning.

allisterakun (approved):
Looks good! Great job, Matthew!

PR #1784, opened 2024-06-27
Feed storage end-to-end testing

### LINKED ISSUE #1768: Write end-to-end testing routine for `FeedStorage`
Add an end-to-end testing routine that will test how stored crops loss mass and nutrients over time.

## Status

Currently, `END_TO_END_TESTING` is an type of task that can be run, but there is nothing implemented for it in `TaskManager`.

The Feed Storage module's routines for processing mass and nutrient losses in stored crops are tested, but only with unit tests.

## Intention

End-to-end testing is desired so that the behavior of stored crops can be observed over longer, more realistic simulation periods. It will also provide an easier and more systematic way to catch problems when making changes to the Feed Storage module.

## Request

Modify the `TaskManager` so that it will run an end-to-end testing routine on the Feed Storage module. This routine should consist of loading in some weather and configuration data, as well as some pre-computed results to compare against.

The first iteration of end-to-end testing for Feed Storage should be very simple - this is just a start to end-to-end testing. It would suffice to create an instance of each final child of `Storage`, store a `HarvestedCrop` instance in each instance, and process the degradations in each on a weekly or monthly basis over a relatively short period of time (100 days to a year).

All outputs from the `Storage` child instances should have corresponding outputs in the precomputed end-to-end testing results. These outputs should be compared to check for equality, or for equality within a specified tolerance.

### Additional Information
When implementing this, there will be no precomputed results to compare against, so those will have to be created. These initial results should be checked with Kevin Panke-Buisse (@KevinP-B) to make sure that they are correct.

Isgotkowitz:
This PR adds the capability to run end-to-end testing on the Feed Storage module.

### Context

Issue(s) closed by this pull request: closes #1768 

### What

This PR adds several inputs that are used to setup simulations to run end-to-end testing on and to verify that the simulation produced the expected results. This includes some novel inputs used to setup the Feed Storage module, so the metadata properties have been added to as well. The Feed Storage module has had some small changes made so that `FeedStorage.process_degradations` will function properly and will have some crops to degrade for testing purposes.

TaskManager has been changed so that it now runs end-to-end testing when an end-to-end testing task is passed to it. SimulationEngine has been modified so that it will process the degradations of stored crops in the Feed Storage module every 15 days.

### Why

End-to-end will be an excellent addition to our testing arsenal. It will allow people working on RuFaS to make changes and find out quickly and reliably if said changes had an effect on the outputs.

### How

When task manager sees an end-to-end testing task, it will route operations to a dedicated function `_handle_end_to_end_testing`, which then calls `_handle_simulation_engine_run_tasks` in the same manner that a normal, single simulation run would.

After the simulation is finished, the new module `EndToEndResultsComparer` is called, which compares actual results against the expected results, which are located in the `input/data/end_to_end_testing/` directory. The actual and expected results are compared using the `deepdiff` library and all differences between the actual and expected results are reported to the OutputManager.

When end-to-end testing is running, the `SimulationEngine` has been changed so that it will setup the FeedManager with a Harvested Crop in each storage type currently available, and then degrade the crops every 15 days.

A few tweaks had to be made in the Feed Storage module that the simulation would work. These changes are minor, except for hardcoding the `bale_size` in `Hay` - but introducing a method for calculating or setting this should be given it's own PR.

### Test plan

Unit tests have been added and run for the updated code in the Feed Manager, Simulation Engine, and Task Manager. Also, end-to-end testing has been run with a new set of dedicated end-to-end input files, which verify that the Feed Storage module is working as expected.

The expected end-to-end results will be checked with a SME to ensure that they are currently reflect correct results - updated as of 7/22, I still need to confirm the validity of the Feed Storage outputs with Kevin.

Update: given the timeline with which we would like to get this PR merged, I think it would be best to establish the end-to-end testing routines then revisit what the expected results should be. This can happen for each individual module in isolation, allowing this PR to move forward and each module to move forward with end-to-end testing at its own pace.

### Expansion
The end-to-end testing mechanism can be very easily expanded to cover different parts of RuFaS. In order to add another domain to end-to-end testing, three things need to be done: 1) a new E2E filter file must be written that contains both the filter for getting the actual results and the expected results, 2) the file `end_to_end_testing_result_paths.json` must be added to in order to point the End-to-End testing results to the proper output files, and 3) a comparison filter file must be added that will get the results of end to end testing for that particular domain.

Also, the current end-to-end testing routine only compares JSON outputs. This should be expanded to cover all the output types - CSVs, reports, and graphs.


RUFAS/end_to_end_tester.py
```diff
@@ -0,0 +1,153 @@
...
+from .routines.feed_storage.storage import Storage
+from .routines.field.crop.crop_data import DEFAULT_DRY_MATTER_DIGESTIBILITY
+
+
+class EndToEndTester:
+    """Executes an end-to-end test on RuFaS."""
+
+    def __init__(self) -> None:
+        self.im = InputManager()
+        self.om = OutputManager()
```
ew3361zh: Curious why we'd need a 3rd instance of a separated IM and OM from the TM IM/OM and the task IM/OM.

RUFAS/end_to_end_tester.py
```diff
@@ -0,0 +1,153 @@
...
+
+        self.compare_results()
+
+    def simulate(self) -> None:
+        """Runs a limited RuFaS simulation."""
+        while not self.time.current_date > self.time.end_date:
+            self.feed_manager.process_degradations(self.weather, self.time)
+            self.weather.record_weather(self.time)
+            self.time.record_time()
+            self.time.advance()
```
ew3361zh: How do you think this could be extended for e2e testing for other modules? I like the idea of being able to isolate each module and run a limited simulation just in that section (though I'm not sure how feasible it is with any of the other modules at this point tbh). Were you thinking this could be an abstract class eventually where the specific module(s) being e2e tested could passed to this class and the appropriate sections initialized?

RUFAS/end_to_end_tester.py
```diff
@@ -0,0 +1,153 @@
...
+            err_title = "Could not obtain expected end-to-end testing results"
+            self.om.add_error(err_title, err_msg, info_map)
+            raise ValueError(err_title)
+
+        try:
+            assert actual_results == expected_results
+        except AssertionError:
+            err_title = "Actual end-to-end testing results different from expected results"
+            self.om.add_error(err_title, err_msg, info_map)
+            raise AssertionError(err_title)
```
ew3361zh: rather than a hard actual_results == expected_results what do you think about using the deepdiff package to produce a record of the differences and record that to an output file? I feel like more often than not we'll want to get a detailed record of where the actual is differing from the expected and try to track down what in our code we've done to account for that difference.

ew3361zh:
This seems like a good start for what I understand e2e testing to want to accomplish. I'm not sure if the intention was to try and get the expected to match the actual results at this stage but it wasn't working for me (i.e. they didn't match and it errored out)- likely something I'm not setting up or executing right.

Shared some thoughts about what you're hoping to accomplish with this PR as far as thinking ahead for future module e2e testing as well and maybe some more nuanced and detailed outputs for the comparison of the results.

Overall nice start to this, Ed!

RUFAS/end_to_end_tester.py
Isgotkowitz: They are the same instances as the task IM/OM, although I just had a meeting with Pooya where we talked about the design of end-to-end testing, and I need to make some changes which will change a lot of this code.

RUFAS/end_to_end_tester.py
Isgotkowitz: When I wrote this I was thinking it would be a temporary measure before we were ready to do end-to-end testing on the entire model, at which point we would use `SimulationEngine` to handle all of this stuff. But just had a meeting with Pooya, and he said it would be better to just tweak `SimulationEngine` so that we can use it for end-to-end testing right now.

RUFAS/end_to_end_tester.py
Isgotkowitz: That's a great idea. I will add that in

Isgotkowitz:
> This seems like a good start for what I understand e2e testing to want to accomplish. I'm not sure if the intention was to try and get the expected to match the actual results at this stage but it wasn't working for me (i.e. they didn't match and it errored out)- likely something I'm not setting up or executing right.
> 
> Shared some thoughts about what you're hoping to accomplish with this PR as far as thinking ahead for future module e2e testing as well and maybe some more nuanced and detailed outputs for the comparison of the results.
> 
> Overall nice start to this, Ed!

Thanks Niko - it is good to know that it isn't working for you. I just had a meeting with Pooya and we talked a lot about end-to-end testing, the conclusion of that conversation being that I need to rework the structure pretty significantly. When I do I will be sure to put in some features which provide that more nuanced and detailed output comparison, that will be much more helpful than just a binary result when someone is trying to figure out what broke the end-to-end testing.

Isgotkowitz:
Update: didn't get the end-to-end testing filter correctly tracked.

RUFAS/simulation_engine.py
```diff
@@ -41,6 +41,9 @@ class SimulationEngine:
         handlers, reception pits, manure separators, and manure storage treatments.
     field_manager: FieldManager
         The FieldManager object that manages all fields in the simulation.
+    run_end_to_end_testing : bool
+        Indicates if a simulation is being run for end-to-end testing. Defaults to False, is set to True if end-to-end
+        testing inputs are found in the Input Manager.
```
PooyaHekmati: It is not clear if this is a signal to run e2e or if it is an indicator that e2e is running. The variable name suggests the former while the docstring suggests the latter. 

---------
edit after reading the rest of the code: maybe rename it to `end_to_end_test_run` or `is_end_to_end_test_run` or something like that?

RUFAS/simulation_engine.py
```diff
@@ -102,6 +106,9 @@ def _run_simulation_main_loop(self) -> None:
 
     def _daily_simulation(self) -> None:
         """Executes the daily simulation routines."""
+        process_degradations_today = self.time.current_julian_day % 15 == 0
+        if self.run_end_to_end_testing and process_degradations_today:
+            self.feed_manager.process_degradations(self.weather, self.time)
```
PooyaHekmati:
```suggestion        
        if self.run_end_to_end_testing:
            process_degradations_today = self.time.current_julian_day % 15 == 0
            if process_degradations_today:
                self.feed_manager.process_degradations(self.weather, self.time)
```
gives a minor performance boost when e2e is not running.

RUFAS/simulation_engine.py
```diff
@@ -176,3 +183,9 @@ def _initialize_simulation(self) -> None:
         self.manure_manager = ManureManager(all_pen_manure_data, self.weather, self.time, manure_class_config)
 
         self.field_manager = FieldManager(manure_manager=self.manure_manager, feed_manager=self.feed_manager)
+
+        run_end_to_end_testing = self.im.get_data("end_to_end_testing_inputs")
+        if not run_end_to_end_testing:
+            return
+        self.feed_manager.setup_stored_feeds(run_end_to_end_testing, self.time)
+        self.run_end_to_end_testing = True
```
PooyaHekmati:
```suggestion
        if run_end_to_end_testing is not None:
            self.feed_manager.setup_stored_feeds(run_end_to_end_testing, self.time)
            self.run_end_to_end_testing = True        
```
I think this is more readable. It is also less susceptible to bugs.

RUFAS/task_manager.py
```diff
@@ -428,6 +427,83 @@ def handle_single_simulation_run(args: Dict[str, Any], output_manager: OutputMan
         simulator.simulate()
         output_manager.add_log("Simulation completed", "Simulation completed", info_map)
 
+    @staticmethod
+    def _handle_end_to_end_testing(
+        args: Dict[str, Any],
+        input_manager: InputManager,
+        output_manager: OutputManager,
+        task_id: Any,
```
PooyaHekmati:
```suggestion
        task_id: str,
```
Other places in the TM where task_id is type annotated as Any are wrong, they should all be strings. Using Any as a type is rare and should be done in exceptional cases only.

RUFAS/task_manager.py
```diff
@@ -428,6 +427,83 @@ def handle_single_simulation_run(args: Dict[str, Any], output_manager: OutputMan
...
+        produce_graphics: bool,
+    ) -> None:
+        """Runs end-to-end testing routine."""
+        info_map = {"class": TaskManager.__name__, "function": TaskManager._handle_end_to_end_testing.__name__}
+
+        output_manager.add_log("Starting end-to-end testing simulation", "", info_map)
```
PooyaHekmati: This is not the correct usage of OM logging system. What you are putting the message where the name is supposed to be and leaving the message empty.

RUFAS/task_manager.py
```diff
@@ -428,6 +427,83 @@ def handle_single_simulation_run(args: Dict[str, Any], output_manager: OutputMan
...
+    @staticmethod
+    def _handle_end_to_end_testing(
+        args: Dict[str, Any],
+        input_manager: InputManager,
+        output_manager: OutputManager,
+        task_id: Any,
+        produce_graphics: bool,
+    ) -> None:
+        """Runs end-to-end testing routine."""
+        info_map = {"class": TaskManager.__name__, "function": TaskManager._handle_end_to_end_testing.__name__}
```
PooyaHekmati: Capture task_id and produce_graphics in the info_map as well; that's what it's meant for.

RUFAS/task_manager.py
```diff
@@ -428,6 +427,83 @@ def handle_single_simulation_run(args: Dict[str, Any], output_manager: OutputMan
...
+
+        output_manager.add_log("Starting end-to-end testing simulation", "", info_map)
+
+        TaskManager._handle_simulation_engine_run_tasks(args, input_manager, output_manager, task_id, produce_graphics)
+
+        output_manager.add_log("Completed end-to-end testing simulation", "", info_map)
```
PooyaHekmati: 1. same as above
2. no, not yet; you have not made the comparison yet.

RUFAS/task_manager.py
```diff
@@ -428,6 +427,83 @@ def handle_single_simulation_run(args: Dict[str, Any], output_manager: OutputMan
...
+
+    @staticmethod
+    def _compare_simulation_outputs_to_expected_outputs(output_manager: OutputManager) -> None:
+        """Compares outputs from a simulation to the results expected for that simulation."""
+        info_map = {
+            "class": TaskManager.__class__.__name__,
+            "function": TaskManager._compare_simulation_outputs_to_expected_outputs.__name__,
+        }
+        path_to_actual_results = None
+        for path in Path("output/JSONs/").iterdir():
```
PooyaHekmati: This is the default path, what if the result is somewhere else?

RUFAS/task_manager.py
```diff
@@ -428,6 +427,83 @@ def handle_single_simulation_run(args: Dict[str, Any], output_manager: OutputMan
...
+            )
+            return
+        with open(path_to_actual_results, "r") as results:
+            actual_results = json.load(results)
+        with open("input/data/end_to_end_testing/end_to_end_results.json", "r") as e_to_e_results:
+            expected_results = json.load(e_to_e_results)
+
+        diff = DeepDiff(expected_results, actual_results, ignore_order=True, verbose_level=2)
+
+        results_dir = Path("output/end_to_end_testing_results")
```
PooyaHekmati: why do we need to make a directory?

RUFAS/task_manager.py
```diff
@@ -428,6 +427,83 @@ def handle_single_simulation_run(args: Dict[str, Any], output_manager: OutputMan
...
+        results_dir.mkdir(parents=True, exist_ok=True)
+        results_path = Path(f"{results_dir}/end_to_end_test_results.txt")
+        with open(results_path, "w") as results_file:
+            if diff == {}:
+                results_file.write("End-to-end testing successful.\n")
+                results_file.write("No differences found between actual and expected outputs.\n")
```
PooyaHekmati: why??

RUFAS/task_manager.py
```diff
@@ -428,6 +427,83 @@ def handle_single_simulation_run(args: Dict[str, Any], output_manager: OutputMan
...
+                }
+                for k, v in sections.items():
+                    if k not in diff:
+                        continue
+                    results_file.write(v)
+                    for item, value in diff[k].items():
+                        results_file.write(f"{item}: {value}\n")
+                    results_file.write("\n")
+
+                output_manager.add_warning("End-to-end testing unsuccessful", "", info_map)
```
PooyaHekmati: This is an error not a warning. 
Also, this is not the correct usage.

RUFAS/task_manager.py
```diff
@@ -428,6 +427,83 @@ def handle_single_simulation_run(args: Dict[str, Any], output_manager: OutputMan
...
+                    "iterable_item_removed": "Actual results missing items:\n",
+                }
+                for k, v in sections.items():
+                    if k not in diff:
+                        continue
+                    results_file.write(v)
+                    for item, value in diff[k].items():
+                        results_file.write(f"{item}: {value}\n")
+                    results_file.write("\n")
+
```
PooyaHekmati: Use OM instead of developing a reporting system in parallel.

RUFAS/simulation_engine.py
Isgotkowitz: `is_end_to_end_test_run` sounds good to me

RUFAS/task_manager.py
Isgotkowitz: Roger that. I will add a section to the OM wiki that goes into more detail about how the different fields of logs/warnings/errors should be used. Edit: actually I wrote an issue about writing that section of the OM wiki.

RUFAS/task_manager.py
Isgotkowitz: > 2. no, not yet; you have not made the comparison yet.

End-to-end testing has not been completed, but the simulation that will be used for end-to-end testing has been completed as this point - this might be too confusing to put in the log messages though.

RUFAS/task_manager.py
Isgotkowitz: I thought it would be nice to have if/when we get to the point where we are producing a lot of different files for end-to-end testing - reports, CSVs, multiple different comparison files, etc. Edit: will take this out, but can always add it in later if that is the way we want to go.

RUFAS/task_manager.py
Isgotkowitz: So that there is always a results file and it is always explicit about what the results of end-to-end testing were. The benefit being that if we wanted to automate end-to-end testing, the script checking the results always looks at the same file and it will be able to match either "End-to-end testing successful" or "End-to-end testing unsuccessful"

Edit: rewrote how results are reported, but idea is the same. Script would always checks the value pointed to by key "end_to_end_testing_results"

RUFAS/task_manager.py
Isgotkowitz: Roger that

RUFAS/task_manager.py
Isgotkowitz: Changed so that it will look in the output JSON dir that is passed!

RUFAS/task_manager.py
Isgotkowitz: Wrote a new system for reporting results. It is much more barebones than the previous version.

Isgotkowitz:
Update: @PooyaHekmati I working to fix the Mypy errors that have been added. They should not mess with your review, but if I unearth something significant that needs to be changed I will leave an update.

RUFAS/task_manager.py
```diff
@@ -428,6 +427,69 @@ def handle_single_simulation_run(args: Dict[str, Any], output_manager: OutputMan
...
+        diff = DeepDiff(expected_results, actual_results, ignore_order=True, verbose_level=2)
+
+        testing_results_file_name = output_manager.generate_file_name("end_to_end_testing_results", "json")
+        testing_results_path = json_output_path.joinpath(Path(testing_results_file_name))
+        no_diff = diff == {}
+        if no_diff:
+            output_manager.add_log("End-to-end testing", "End-to-end testing successful", info_map)
+        else:
+            output_manager.add_error("End-to-end testing", "End-to-end testing unsuccessful", info_map)
```
Isgotkowitz: Make log messages different structurally so they are easier to differentiate

RUFAS/task_manager.py
```diff
@@ -428,6 +427,69 @@ def handle_single_simulation_run(args: Dict[str, Any], output_manager: OutputMan
...
+        with open(path_to_actual_results, "r") as results:
+            actual_results = json.load(results)
+        with open("input/data/end_to_end_testing/end_to_end_results.json", "r") as e_to_e_results:
+            expected_results = json.load(e_to_e_results)
+
+        diff = DeepDiff(expected_results, actual_results, ignore_order=True, verbose_level=2)
+
+        testing_results_file_name = output_manager.generate_file_name("end_to_end_testing_results", "json")
+        testing_results_path = json_output_path.joinpath(Path(testing_results_file_name))
+        no_diff = diff == {}
```
Isgotkowitz: Change name and add type hint to improve readability

RUFAS/output_manager.py
```diff
@@ -1199,7 +1214,8 @@ def _route_save_functions(
             "class": self.__class__.__name__,
             "function": self._route_save_functions.__name__,
         }
-        if filter_file.startswith(self.__supported_filter_types_prefixes["json"]):
+
+        if filter_file.startswith(self._filter_prefixes["json"]):
```
Isgotkowitz: Use .get() with some "off-limits" or something else as the default, so that key errors are avoided - maybe "better than a key error".

Criteria:
- someone reading code, should tell them it is for sake of getting around problem, is not supposed to work when default is used.
- minimize chances of someone actually prefixing their filter file with the default by chance/on accident

RUFAS/task_manager.py
```diff
@@ -436,6 +438,82 @@ def handle_single_simulation_run(args: Dict[str, Any], output_manager: OutputMan
...
+    def _compare_simulation_outputs_to_expected_outputs(args: Dict[str, Any], output_manager: OutputManager) -> None:
+        """Compares outputs from a simulation to the results expected for that simulation."""
+        info_map = {
+            "class": TaskManager.__class__.__name__,
+            "function": TaskManager._compare_simulation_outputs_to_expected_outputs.__name__,
+        }
+        path_to_actual_results = None
+        json_output_path = args["json_output_directory"]
+        for path in json_output_path.iterdir():
+            matches = re.match(
```
Isgotkowitz: Change name to `is_a_match` or `is_matching` or anything is-...

RUFAS/task_manager.py
```diff
@@ -436,6 +438,82 @@ def handle_single_simulation_run(args: Dict[str, Any], output_manager: OutputMan
...
 
+    @staticmethod
+    def _handle_end_to_end_testing(
+        args: Dict[str, Any],
+        input_manager: InputManager,
+        output_manager: OutputManager,
+        task_id: str,
+        produce_graphics: bool,
+    ) -> None:
+        """Runs end-to-end testing routine."""
```
Isgotkowitz, line 456: create module dedicated to e2e testing and move these routines there.

Isgotkowitz: Done

RUFAS/routines/feed_storage/hay.py
```diff
@@ -57,7 +57,7 @@ def bale_size(self) -> float:
         float
             The diameter of the hay bale (meters).
         """
-        pass
+        return 1.5
```
PooyaHekmati: either get this from IM or define a constant; magic numbers are big no no

RUFAS/simulation_engine.py
```diff
@@ -178,3 +193,10 @@ def _initialize_simulation(self) -> None:
         )
 
         self.field_manager = FieldManager(manure_manager=self.manure_manager, feed_manager=self.feed_manager)
+
+        # TODO: remove the below code after Animal and Feed Storage modules are connected - #1878
+        is_end_to_end_test_run = self.im.get_data("end_to_end_testing_inputs")
```
PooyaHekmati: Task manager should populate this based on the task type.

RUFAS/task_manager.py
```diff
@@ -436,6 +436,37 @@ def handle_single_simulation_run(args: Dict[str, Any], output_manager: OutputMan
...
+        info_map = {
+            "class": TaskManager.__name__,
+            "function": TaskManager._handle_end_to_end_testing.__name__,
+            "task_id": task_id,
+            "produce_graphics": produce_graphics,
+        }
+
+        output_manager.add_log("End-to-end testing", "Starting end-to-end testing simulation.", info_map)
```
PooyaHekmati:
```suggestion
        output_manager.add_log("End-to-end testing", "Starting simulation for end-to-end testing.", info_map)
```
It implies that we are simulating e2e

RUFAS/end_to_end_tester.py
```diff
@@ -0,0 +1,91 @@
...
+from typing import Any
+
+from RUFAS.output_manager import OutputManager
+from RUFAS.units import MeasurementUnits
+
+import json
+import re
+
+
+TestResultPaths = namedtuple("TestResultPaths", ["domain_prefix", "expected_results_path", "actual_results_path"])
```
PooyaHekmati:
```suggestion
TestResultPathType = namedtuple("TestResultPaths", ["domain_prefix", "expected_results_path", "actual_results_path"])
```
You are defining the type here; the name implies that a collection or iterable is being defined.

RUFAS/end_to_end_tester.py
```diff
@@ -0,0 +1,91 @@
...
+Maps a RuFaS domain that is supported for end-to-end testing to information needed to collect actual and expected
+results for testing that domain, and the prefix used when saving the test results to the Output Manager.
+"""
+DOMAINS = {
+    "Feed Storage": TestResultPaths(
+        "FeedStorageResults",
+        "input/data/end_to_end_testing/e2e_json_feed_storage_filter.json",
+        "end-to-end-testing_saved_variables_e2e_feed_storage_",
+    )
+}
```
PooyaHekmati: More descriptive name plus type annotation 
```suggestion
TEST_DOMAINS: Dict[str, TestResultPathType] = {
    "Feed Storage": TestResultPaths(
        "FeedStorageResults",
        "input/data/end_to_end_testing/e2e_json_feed_storage_filter.json",
        "end-to-end-testing_saved_variables_e2e_feed_storage_",
    )
}
```

Also, the key of this dict seems redundant, `domain_prefix` serves the same purpose. 
```suggestion
TEST_DOMAINS: List[TestResultPathType] = [
    TestResultPaths("FeedStorageResults",
        "input/data/end_to_end_testing/e2e_json_feed_storage_filter.json",
        "end-to-end-testing_saved_variables_e2e_feed_storage_",
    )
]
```
Furthermore, why not getting these from IM instead of hardcoding?

RUFAS/end_to_end_tester.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class EndToEndTester:
```
PooyaHekmati: This name is misleading. It suggests that this class does the end to end testing, but it only is a utility for the actual testing that happens in TM. Either the logic of `TM._handle_end_to_end_testing` has to be moved in this class; or the name has to be changed to reflect its utility nature.

RUFAS/end_to_end_tester.py
Isgotkowitz: > Also, the key of this dict seems redundant, domain_prefix serves the same purpose.

They could pretty easily be substituted for one another. That said, I do like having them separate because one makes for a really log message that reads really nicely and the other makes for a variable output prefix that looks really good.

Edit: scratch that, it makes them easier to incorporate into the inputs if they don't have a key.

Isgotkowitz: > Furthermore, why not getting these from IM instead of hardcoding?

I imagine that once we get the model up and running, these domains will be pretty set and we won't need to update them often. In which case it makes more sense to me to have them hardcoded, as opposed to being a modifiable input file. Other than that though, no good reason. I'll update it to use an input file and the input manager

Isgotkowitz:
Update: I have renamed several of the data structure to be more specific and function-appropriate. Unit tests have been added and updated appropriately.

input/data/end_to_end_testing/end_to_end_inputs.json
Isgotkowitz, line 1: This file is only necessary while the Animal and Feed Storage modules are disconnected, after they become connected this file can be deprecated along with some of the changes introduced - #1878

RUFAS/e2e_test_results_comparer.py
```diff
@@ -0,0 +1,95 @@
...
+
+        for path_set in test_result_path_sets:
+            om.add_log(
+                f"End-to-end testing for {path_set.domain}",
+                "Collecting and comparing actual and expected results",
+                info_map,
+            )
+            path_to_actual_results = None
+            for path in json_output_path.iterdir():
+                is_a_match = re.match(f"{str(json_output_path)}/{path_set.actual_results_path}.*", str(path))
```
PooyaHekmati: Why do you need to call `str()` inside an f-string? I think it should work without it

RUFAS/e2e_test_results_comparer.py
```diff
@@ -0,0 +1,95 @@
...
+                if is_a_match:
+                    path_to_actual_results = path
+                    break
+            else:
+                om.add_error(
+                    f"Could not find actual end-to-end testing results for {path_set.domain}",
+                    "End-to-end testing failed.",
+                    info_map,
+                )
```
PooyaHekmati:
```suggestion
                om.add_error(
                    "End-to-end testing failed.",
                    f"Could not find actual end-to-end testing results for {path_set.domain}",
                    info_map,
                )
```
Also, add 'path_set.domain` to `info_map`

RUFAS/e2e_test_results_comparer.py
```diff
@@ -0,0 +1,95 @@
...
+            "class": E2ETestResultsComparer.__class__.__name__,
+            "function": E2ETestResultsComparer.compare_actual_and_expected_test_results.__name__,
+        }
+        test_result_path_sets = E2ETestResultsComparer._get_test_result_paths()
+
+        for path_set in test_result_path_sets:
+            om.add_log(
+                f"End-to-end testing for {path_set.domain}",
+                "Collecting and comparing actual and expected results",
+                info_map,
```
PooyaHekmati, line 45: add 'path_set.domain` to `info_map`

RUFAS/e2e_test_results_comparer.py
```diff
@@ -0,0 +1,95 @@
...
+                continue
+            with open(path_to_actual_results, "r") as results:
+                actual_results = json.load(results)
+            with open(f"{path_set.expected_results_path}", "r") as e_to_e_results:
+                filter_and_results = json.load(e_to_e_results)
+                expected_results = filter_and_results["expected_results"]
+
+            diff = DeepDiff(expected_results, actual_results, ignore_order=True, verbose_level=2)
+
+            is_difference_in_results: bool = diff == {}
```
PooyaHekmati:
```suggestion
            is_difference_in_results:bool= False if diff == {} else True
```
Readability matters

RUFAS/e2e_test_results_comparer.py
```diff
@@ -0,0 +1,95 @@
...
+            if is_difference_in_results:
+                om.add_log(
+                    f"End-to-end testing succeeded for {path_set.domain}",
+                    "No differences found between actual and expected end-to-end testing results.",
+                    info_map,
```
PooyaHekmati, line 73: add `path_set.domain` to `info_map`

RUFAS/e2e_test_results_comparer.py
```diff
@@ -0,0 +1,95 @@
...
+                )
+            else:
+                om.add_error(
+                    f"Failed end-to-end testing for {path_set.domain}",
+                    "Identified differences between actual and expected results.",
+                    info_map,
```
PooyaHekmati, line 79: add `path_set.domain` to `info_map`

RUFAS/e2e_test_results_comparer.py
```diff
@@ -0,0 +1,95 @@
...
+            is_difference_in_results: bool = diff == {}
+            if is_difference_in_results:
+                om.add_log(
+                    f"End-to-end testing succeeded for {path_set.domain}",
+                    "No differences found between actual and expected end-to-end testing results.",
+                    info_map,
+                )
+            else:
+                om.add_error(
+                    f"Failed end-to-end testing for {path_set.domain}",
```
PooyaHekmati:
```suggestion
                    f"End-to-end testing failed for {path_set.domain}",
```
Consistency matters

RUFAS/output_manager.py
```diff
@@ -149,6 +159,16 @@ def __init__(self) -> None:
...
+            self.is_end_to_end_testing_run: bool = False
+            self.is_first_post_processing: bool = True
+
+    @property
+    def _filter_prefixes(self) -> dict[str, str]:
+        """Returns the appropriate set of acceptable filter prefixes."""
+        if not self.is_end_to_end_testing_run:
+            return self.__supported_filter_types_prefixes
+        else:
+            return self.__end_to_end_testing_filter_prefixes
```
```suggestion
        if self.is_end_to_end_testing_run:
            return self.__end_to_end_testing_filter_prefixes
        else:
            return self.__supported_filter_types_prefixes
```
avoid unnecessary negation

RUFAS/routines/feed_storage/feed_manager.py
```diff
@@ -8,6 +8,8 @@
 from .grain import Grain, Dry, HighMoisture
 from .hay import Hay, ProtectedIndoors, ProtectedTarped, ProtectedWrapped, Unprotected
 from .silage import Silage, Bag, Bunker, Pile
+from ...time import Time
+from ...weather import Weather
```
PooyaHekmati: avoid relative imports that are not on the same level or one higher

PooyaHekmati (approved):

RUFAS/e2e_test_results_comparer.py
```diff
@@ -0,0 +1,95 @@
...
+                    "Identified differences between actual and expected results.",
+                    info_map,
+                )
+            diff.update({"end_to_end_testing_passing": is_difference_in_results})
+            info_map.update({"units": MeasurementUnits.UNITLESS, "prefix": path_set.domain})
+            for comparison_type, difference in diff.items():
+                om.add_variable(comparison_type, difference, info_map)
```
ew3361zh: Curious about why these are being added as variables and not logs?

RUFAS/output_manager.py
```diff
@@ -149,6 +159,16 @@ def __init__(self) -> None:
             )
             self.time = None
             self._variables_usage_counter: Counter[str] = collections.Counter()
+            self.is_end_to_end_testing_run: bool = False
+            self.is_first_post_processing: bool = True
+
+    @property
+    def _filter_prefixes(self) -> dict[str, str]:
```
ew3361zh, line 166: Nice, I like this.

RUFAS/output_manager.py
```diff
@@ -1311,6 +1331,16 @@ def _route_save_functions(
                     f"Graphic generation is disabled, skipping {filter_file=}",
                     info_map,
                 )
+            return
+        is_comparison = filter_file.startswith(self._filter_prefixes.get("comparison", "Better than a key error."))
+        if is_comparison and not self.is_first_post_processing:
```
ew3361zh, line 1336: Anything to log here if this statement is false?

RUFAS/e2e_test_results_comparer.py
```diff
@@ -0,0 +1,95 @@
...
+        test_result_path_sets = E2ETestResultsComparer._get_test_result_paths()
+
+        for path_set in test_result_path_sets:
+            om.add_log(
+                f"End-to-end testing for {path_set.domain}",
+                "Collecting and comparing actual and expected results",
+                info_map,
+            )
+            path_to_actual_results = None
+            for path in json_output_path.iterdir():
```
ew3361zh, line 48: This was erroring for me `FileNotFoundError: [Errno 2] No such file or directory: 'output/JSONs'`.

Maybe user error - not sure if I'm supposed to set up the e2et output folder in the task or if this dir needs to be created and isn't.

RUFAS/e2e_test_results_comparer.py
Isgotkowitz: They are added as variables so that they can be filtered out into different files. So if we wanted to add e2e testing for the animal module, we would add the differences found as Animal-specific variables, then filter them out into an Animal results specific file using an Animal specific filter.

RUFAS/output_manager.py
Isgotkowitz: Great question, if we're going with exactly the old behavior then no because `_route_save_functions` didn't do anything to make sure the file it was working with actually had a valid filter prefix. But maybe it should! I will investigate it and see if it would make sense to log something here

RUFAS/e2e_test_results_comparer.py
Isgotkowitz: Hmm how were you running e2e testing?

ew3361zh:
```
{
  "parallel_workers": 4,
  "tasks": [
    {
      "task_type": "END_TO_END_TESTING",
      "metadata_file_path": "input/metadata/end_to_end_testing_metadata.json",
      "output_prefix": "e2et",
      "log_verbosity": "errors"    }
  ]
}
```

Isgotkowitz: Ahh, I think you have to use the output prefix "end-to-end-testing", otherwise the regex patterns for finding the actual results will not be able to find the right output files.

RUFAS/output_manager.py
Isgotkowitz: I don't think a message needs to be logged here. If a filter file is found that does not have a correct prefix, a warning will be logged about it in `_list_filter_files_in_dir` and I think this is sufficient

RUFAS/e2e_test_results_comparer.py
Isgotkowitz: Wait it may be more than that. This is the task that I have to run end to end testing with: 
```json
{
  "parallel_workers": 4,
  "tasks": [
    {
      "task_type": "END_TO_END_TESTING",
      "metadata_file_path": "input/metadata/end_to_end_testing_metadata.json",
      "output_prefix": "end-to-end-testing",
      "filters_directory": "input/data/end_to_end_testing",
      "log_verbosity": "logs",
      "exclude_info_maps": false,
      "random_seed": 42
    }
  ]
}
```
Sorry I should have included instructions for actually running end-to-end testing

ew3361zh: Hmm, reran with the output prefix as both "end-to-end-testing" and "end_to_end_testing" and got the same error.

ew3361zh: > ```json
> {
>   "parallel_workers": 4,
>   "tasks": [
>     {
>       "task_type": "END_TO_END_TESTING",
>       "metadata_file_path": "input/metadata/end_to_end_testing_metadata.json",
>       "output_prefix": "end-to-end-testing",
>       "filters_directory": "input/data/end_to_end_testing",
>       "log_verbosity": "logs",
>       "exclude_info_maps": false,
>       "random_seed": 42
>     }
>   ]
> }
> ```

That did it, thank you!

ew3361zh (approved):
Nice work, Ed! I'm satisfied as far as the initial setup of the e2e testing structure and process. I think seeing some actual results will be instructive and will refine the process further but this does seem open to extension and I think is a great start for having e2e established in RUFAS.

RUFAS/e2e_test_results_comparer.py
Isgotkowitz: I'm not sure why I had this in here, but you're right it works without it.

Isgotkowitz:
Update: while setting the expected results for Feed Storage equal to the current expected results (will assess in more detail in another issue/PR), I found that the Feed Storage module wasn't being properly initialized during end-to-end testing simulations.

This was because the SimulationEngine would initialize the FeedManager before a setter could indicate that end-to-end testing was being run. The solution was to make `is_end_to_end_run` an argument to the constructor of Simulation Engine instead of having it default to False and having a setter for it.

PR #1785, opened 2024-06-28
sobol and morris samplers

JoeWaddell:
Adds samplers for Morris and Sobol sampling from SALib.

### Context

- [ ] Yes
- [x] No

### What

Adds two new samplers.

### Why

Expands the sensitivity analysis methodology native to RuFaS.

### How

New options added in metadata, additional user inputs in the `sampler` field of a Sensitivity Analysis task.

### Test plan

Joe has been testing the sampler performance and the associated analytical methods. The Sobol sampler is the most thorough, but requires more runs. Morris is better than the Saltelli sampler (deprecated in SALib), but doesn't include interaction effects, as does the Sobol method.


RUFAS/task_manager.py
```diff
@@ -251,12 +253,23 @@ def _expand_sensitivity_analysis_args(self, multi_run_args: Dict[str, Any]) -> L
...
             sampled_values = saltelli_sampler.sample(
                 parsed_SA_input_variables,
                 multi_run_args["saltelli_number"],
                 skip_values=multi_run_args["saltelli_skip"],
             )
+        elif multi_run_args["sampler"] == "sobol":
+            sampled_values = sobol_sampler.sample(
+                parsed_SA_input_variables,
+                multi_run_args["saltelli_number"],
+                skip_values=multi_run_args["saltelli_skip"],
```
PooyaHekmati: If number and skip are used in sobol and morris as well, should we rename them to capture this? Right now it's confusing that we are calling sobol but passing saltelli args.

PooyaHekmati (approved):

input/metadata/properties/tasks_properties.json
```diff
@@ -151,18 +151,18 @@
...
+          "description": "The number of points in the sampling sequence to skip for Sensitivity Analysis",
           "minimum": 0,
           "default": 0
         },
-        "saltelli_number": {
+        "sampler_n": {
           "type": "number",
-          "description": "The saltelli sampling value for Sensitivity Analysis",
+          "description": "The number of samples to generate for Sensitivity Analysis. Ideally a power of 2 and <= 'skip_values'",
           "minimum": 2,
```
tomhuhh, line 190: If I understand it correctly, should this value be greater than or equal to the skip value instead?

tomhuhh (commented):
Great work, Joe! This addition make sense to me! I've yet to test the updates yet, but I'll keep you posted. Meanwhile, I made a minor comment. Always excited to learn more about the sensitivity analysis. Thanks for including me on this one!

tomhuhh (approved):
Nice job as always, Joe! Thanks for correcting the typo flagged earlier. 

I tested with both "sobol" and "morris" methods using the configurations as follows. In this case, the "morris" method required 6 simulations while the "sobol" method required 12 simulations with the same configurations (i.e. skip_values and sampler_n). This is consistent with what Joe described in the context of this PR.   

```
{
  "parallel_workers": 4,
  "tasks": [

    
    {
      "task_type": "SENSITIVITY_ANALYSIS",
      "output_prefix": "morris",
      "log_verbosity": "errors",
      "sampler": "morris",
      "SA_load_balancing_start": 0,
      "SA_load_balancing_stop": 1,
      "skip_values": 0, 
      "sampler_n": 2, 
      "SA_input_variables": [
        {
          "variable_name": "animal.herd_information.calf_num",
          "lower_bound": 6,
          "upper_bound": 10,
          "data_type":"int"
        },
        {
          "variable_name": "animal.herd_information.cow_num",
          "lower_bound": 98,
          "upper_bound": 102,
          "data_type":"int"
        }
      ]
    }
  ]
}
```

Issue #1787, opened 2024-06-28
[Report Generator] Associate Report Filter Constants with Enum


ew3361zh:
When issue #1516 is complete, the user will have the option to include the units in the column headers for the data aggregated in the report.

To make this as accurate and consistent as possible, we also need to ensure any report aggregation done with constants is done correctly. Importantly, we want to be able to note how this will affect the units of the data this constant is being used with in report generator.

To achieve this, in report generator when we're extracting the constant from the filter file, we should have it match with the constant from the Enum class of GeneralConstants (MASM > RUFAS > general_constants.py) and some thought should be given to how each would affect a unit it's interacting with.

Discussed with @PooyaHekmati during our check-in.

ew3361zh:
Tagging in @KFosterReed, @JoeWaddell, @jadamchick, @krb78 for any thoughts/concerns/feedback.

PR #1798, opened 2024-07-03
Add Units to Reports

### LINKED ISSUE #1516: [OM] Include units in reports when specified

ew3361zh:
Takes variable units and adds them to the column headers of reports. Default is on but provides user option to turn this feature off in filter file settings.

**Note: an extensively large example report filter is provided for testing purposes but will be removed prior to merging. This report file `report_filter_julie.json` accounts for 1364 of the total lines added here.**

### Context

Issue(s) closed by this pull request: closes #1516

- [ ] Yes
- [x] No

### What

Adds new functionality to include units in the report column header in the format similar to how units appear in the variable-names report: `report_name_(units)`.

Allows for combinations and simplifications/reductions of units based on the aggregation function called. 
For example: 
- if the `product` aggregation is called on two sets of data with `kg` and `kg/day` as their respective units, the resulting unit for the aggregated data would be `kg^2/day`.
- if the `division` aggregation function is called on two sets of data with `m^3/day` and `day/m` as their respective units, the resulting unit for the aggregated data would be `m^2`.

Additionally accounts for report data sent to be graphed so that we're not double-adding units or adding units in GraphGenerator when the user has requested not to add units.

Accounts for combining cross-referenced report units.

Overview diagram of the new way ReportGenerator will work with adding units:
https://whimsical.com/reportgenerator-overview-PV7meFcHSUMxtmtfbUHZWs

### Why

It's helpful to know what units the aggregated data are in.

### How

In RG:
- Units are added automatically to report data from info_maps. User can turn this off by setting `display_units` to `false`.
- When the aggregation occurs, the units are extracted from each of the variable names and combined according to the aggregation operation being performed on the data.
- For "product" and "division" aggregations, the units are combined into new complex units and accordingly reduced when like units are found in numerators and denominators.
- For all other operations, a check is made to ensure the units of the data being combined are the same and a warning is logged to OM if they are not but the aggregation is allowed to continue.

In MeasurementUnits class:
Units-specific logic to:
- extract units from var-names
- parse the units to get exponent values
- combine exponents
- simplify complex units with like units in numerator/denominator
- convert unit dictionaries back into strings for adding to report titles

### Test plan

An extensive example report is added for testing purposes to see the complex units resulting from the combinations `report_filter_julie.json`.
Additionally, the user is encouraged to try their own reports and aggregations - the units for each variable are available in the variable names file generated with each simulation.
Unit testing was fixed.

### TODO
- [x] Add warning for unlike units for addition/subtraction vars.
- [x] Expand unit test coverage to new functionality.
- [x] Address MyPy error increases
- [x] Add flag option for user to NOT reduce units in cases they don't want to.
- [x] Remove example report filter file.


Isgotkowitz:
@ew3361zh when this PR is ready to be merged, please update issue #1787 with any specific tasks that still need to be addressed and update the t-shirt size, if necessary.

jadamchick (commented):
Hi Niko! Thanks for the patience on getting back to you with this. 
I'm attaching the report from running the default herd, with an additional column where I corrected/clarified the units if what it provided was either wrong or confusing. 
[default_report_units_branch_report_report_filter_julie.json_05-Aug-2024_Mon_12-43-57.csv](https://github.com/user-attachments/files/16501907/default_report_units_branch_report_report_filter_julie.json_05-Aug-2024_Mon_12-43-57.csv)


A few notes on patterns I noticed- 
- it seemed to take "unit" / "unit" to produce "unit^2" -- eg animals^2 where in reality it is a percent
- the notation is confusing for some of the denominators -- eg "kg/animals^-1" -- does that mean animals/kg or kg*animals or kg/animals? (In other words, how is the ^-1 getting applied? and/or is it redundant with the fraction bar)
- When there is an operation with the constant, and the constant has a unit, that unit doesn't get incorporated. I wouldn't really expect it to, but it does mean that the auto-units are incorrect in those cases... 
- A couple inconsistencies that are probably more upstream questions about how units are specified: 
   - when unit of time is/isn't included. since in the reports, some are aggregated to be the sum for the entire year while others are an average per-day value (for example), it seems like something worth thinking about 
   - when "kg of __" is specified. For example, it looks like the purchased feed emissions are labeled as kg CO2-eq while the vast majority are just "kg" (when they could be kg CH4, kg N2O, kg body weight, kg DMI, etc.)

Happy to chat more about this if it would be helpful. Thanks for working on it!

ew3361zh:
> Hi Niko! Thanks for the patience on getting back to you with this. I'm attaching the report from running the default herd, with an additional column where I corrected/clarified the units if what it provided was either wrong or confusing. [default_report_units_branch_report_report_filter_julie.json_05-Aug-2024_Mon_12-43-57.csv](https://github.com/user-attachments/files/16501907/default_report_units_branch_report_report_filter_julie.json_05-Aug-2024_Mon_12-43-57.csv)
> 
> A few notes on patterns I noticed-
> 
> * it seemed to take "unit" / "unit" to produce "unit^2" -- eg animals^2 where in reality it is a percent

I think I did something along the way to invert how those units should be combined - for example the instance of animals^2 was an animals/animals unit and therefore should have cancelled one another out resulting in `unitless`. I will address this and update the PR. Thanks!

> * the notation is confusing for some of the denominators -- eg "kg/animals^-1" -- does that mean animals/kg or kg*animals or kg/animals? (In other words, how is the ^-1 getting applied? and/or is it redundant with the fraction bar)

Let me take another look at the results that are doing this after I fix the combine_units() function.

> * When there is an operation with the constant, and the constant has a unit, that unit doesn't get incorporated. I wouldn't really expect it to, but it does mean that the auto-units are incorrect in those cases...

This is going to be addressed in issue #1787.


> * A couple inconsistencies that are probably more upstream questions about how units are specified:
>   
>   * when unit of time is/isn't included. since in the reports, some are aggregated to be the sum for the entire year while others are an average per-day value (for example), it seems like something worth thinking about
>   * when "kg of __" is specified. For example, it looks like the purchased feed emissions are labeled as kg CO2-eq while the vast majority are just "kg" (when they could be kg CH4, kg N2O, kg body weight, kg DMI, etc.)
>
> Happy to chat more about this if it would be helpful. Thanks for working on it!

I think it would be helpful to chat on these two and look at some specific examples and talk about what you think would be a helpful output in these cases.

Thank you for reviewing, Julie! I will follow up with you once I've made the changes to address the combine-units error.

ew3361zh:
Updates:
1. Var unit combination methods are corrected to properly reduce like-units.
2. Vars that should have been unitless percents are now unitless percents.

Still working on:
1. Finishing unit testing fixes and expanding coverage for new functionality.
2. Going to move unit combination and reduction functions into MeasurementUnits Enum class.
3. Add docstrings, type-hints, reduce MyPy errors.

RUFAS/report_generator.py
```diff
@@ -406,7 +408,7 @@ def _perform_aggregations(
         self,
         filtered_pool: Dict[str, Dict[str, List[Any]]],
         filter_content: Dict[str, Any],
-    ) -> Dict[str, List[float]]:
+    ) -> tuple[dict[str, list[any]], list[dict[str, str | dict[str, str]]]]:
```
allisterakun: I am just curious about the purpose of using a tuple here. Is it purely to address MyPy errors and type hints? Or is there a specific reason?

RUFAS/report_generator.py
```diff
@@ -453,44 +459,153 @@ def _perform_aggregations(
...
             has_dict_variables = filter_content.get("variables") is not None
             has_multiple_columns = len(vertically_aggregated) > 1
-
-            if has_dict_variables or has_multiple_columns:
-                aggregate_report = {f"{key}_ver_agg": value for key, value in vertically_aggregated.items()}
+            if filter_content.get("display_units", True):
+                if has_dict_variables or has_multiple_columns:
+                    aggregate_report = {self._update_key(key): value for key, value in vertically_aggregated.items()}
+                else:
+                    units = re.search(r"\(.*\)", next(iter(report_data)))
```
allisterakun, line 511: Is it worth extracting the search for units using regex into a function?

allisterakun: Also, from my understanding, we are relying fully on the "." in the variable name string to figure out the unit. Are there any potential issues (like having dots as part of the variable name)?

RUFAS/report_generator.py
```diff
@@ -453,44 +459,153 @@ def _perform_aggregations(
...
         elif vertical_agg_key:
             vertical_aggregator = AGGREGATION_FUNCTIONS[vertical_agg_key]
             vertically_aggregated = self._apply_vertical_aggregation(aggregate_report, vertical_aggregator)
 
             has_dict_variables = filter_content.get("variables") is not None
             has_multiple_columns = len(vertically_aggregated) > 1
-
-            if has_dict_variables or has_multiple_columns:
-                aggregate_report = {f"{key}_ver_agg": value for key, value in vertically_aggregated.items()}
+            if filter_content.get("display_units", True):
```
allisterakun: Since this is used in the horizontal and vertical agg, can we extract the "display_units" from `filter_content` as a local variable at the beginning of the function and use that instead? So we don't have to extract the "display_units" on the fly for multiple times.

RUFAS/report_generator.py
```diff
@@ -453,44 +459,153 @@ def _perform_aggregations(
...
+                }
+                warning_event_log = {
+                    "warning": "Report Generator Units Warning",
+                    "message": f"Report units do not match for operation {operation}.",
+                    "info_map": info_map,
+                }
+                event_logs.append(warning_event_log)
+            combined_numerator = numerator1.copy()
+            combined_denominator = denominator1.copy()
+
```
allisterakun, line 623: Since we are not using the `AGGREGATION_FUNCTIONS`, do you think we should add a safety measure (else) at the end to handle the situation for unexpected operations that are not in ["sum", "subtraction", "average", "SD", "product", "division"]. If we decide to introduce new operations in the future and forget to update this function, troubleshooting would be much easier.

RUFAS/report_generator.py
```diff
@@ -507,26 +622,34 @@ def _handle_horizontal_and_vertical_aggregations(
...
+                aggregate_report = {"hor_ver_agg": [vertical_aggregator(horizontally_aggregated)]}
         else:
             vertically_aggregated = self._apply_vertical_aggregation(aggregate_report, vertical_aggregator)
             ver_hor_aggregated = []
             for elements in zip(*vertically_aggregated.values()):
                 ver_hor_aggregated.append(horizontal_aggregator(list(elements)))
-            aggregate_report = {"ver_hor_agg": ver_hor_aggregated}
-        return aggregate_report
+            aggregate_units, event_logs = self._aggregate_units(vertically_aggregated, horizontal_aggregator)
+            if filter_content.get("display_units", True):
```
allisterakun: Same suggestion; since "display_units" is used more than once in this function, why not save this as a local variable at the beginning of the function?

allisterakun (approved):
Great job, Niko! It generally looks good! Left some comments.
Also, maybe mention that you removed the is_leap_year() function from the `Schedule` class in field. And maybe update the TODO checklist at the end of the PR description.

krb78 (approved):
Honestly, this looks really awesome! It seems like Allister gave some stellar code related feedback. I did find a glitch, but I don't think it's anything in what you did, I assume it's elsewhere in the code because so many of these look great. When I ran this filter through, it gave me actual purchased feeds in kg (which is correct) and then actual purchased feed emissions in kg CO2-eq/kg DM (which is incorrect, it should just be kg CO2-eq):             "name": "Actual Purchased Feeds",
            "filters": [
                ".*actual_purchased_feed_totals.*"
            ],
            "variables": [".*"]
        },
        {
            "name": "Actual Purchased Feed Emissions",
            "filters": [
                ".*actual_purchased_feed_emissions.*"
            ],
            "variables": [".*"]
        }

Again, I don't think this is you, but this PR led me to realize I think we messed up somewhere in the labeling elsewhere in the code (so thanks - I think haha).

RUFAS/report_generator.py
ew3361zh: In discussions with Pooya, one of his requests was to log a warning to OM when the units for two sets of data were different when they're added, subtracted, or averaged by RG. That check happens at `_combine_units()` but the list of warnings, logs, and errors gets returned to OM from `generate_report()` so essentially this return (and the others like it) are just passing that list of warnings back up to `generate_report()` along with the previous return which was the aggregated data.

ew3361zh:
> Honestly, this looks really awesome! It seems like Allister gave some stellar code related feedback. I did find a glitch, but I don't think it's anything in what you did, I assume it's elsewhere in the code because so many of these look great. When I ran this filter through, it gave me actual purchased feeds in kg (which is correct) and then actual purchased feed emissions in kg CO2-eq/kg DM (which is incorrect, it should just be kg CO2-eq): "name": "Actual Purchased Feeds", "filters": [ "._actual_purchased_feed_totals._" ], "variables": ["._"] }, { "name": "Actual Purchased Feed Emissions", "filters": [ "._actual_purchased_feed_emissions._" ], "variables": ["._"] }
> 
> Again, I don't think this is you, but this PR led me to realize I think we messed up somewhere in the labeling elsewhere in the code (so thanks - I think haha).

Thanks for reviewing, Kaitlyn! Yes, it looks like from the variable_names file, all `purchased_feed_emissions` are logged to OutputManager with the unit (`kg CO2` / `kg DM`). If you want to change that, I can create an issue for it - it wouldn't be too difficult. But take a look at RUFAS > routines > animal > purchased_feed_emissions_estimator.py - that's where all the purchased feed emissions are added with the (`kg CO2` / `kg DM`) unit.

ew3361zh:
> Great job, Niko! It generally looks good! Left some comments. Also, maybe mention that you removed the is_leap_year() function from the `Schedule` class in field. And maybe update the TODO checklist at the end of the PR description.

Thanks, Allister! Just real quick re: `is_leap_year()` that's from a PR Matthew merged into `dev` yesterday. Another one of those where I'm not sure why it's bleeding over into this PR which has nothing to do with it.

RUFAS/report_generator.py
ew3361zh: I think this is the only place this extraction check is used for now so I'm not sure it needs extraction.

As far as the `.` interfering with regex patterns - the key for this search is that it is only looking for characters between two parantheses and I don't believe we have any other parts of a variable name that use that structure currently.

PR #1804, opened 2024-07-05
Resolve division by zero error in C&S

### LINKED ISSUE #1799: [C&S] Division by Zero when No Crop to Harvest
In certain situations, like when there is no weather data, C&S won't grow the crops. However, due to its current structure, it still does the Harvest operation. This leads to uncaught Division by Zero exception when the Harvest operation tries to find the ratio of the crop being over and under the ground (there is no crop so it leads to 0/0). 

Add try-except clause to this operation and catch the division by zero exception. When it happens, log a warning to OM to let the user know about what happened. 

Keep @KFosterReed , @krb78 , and @PooyaHekmati updated about the progress here.

This is a follow up to #1088

matthew7838:
Potential zero division error when calculating the fraction cut, adding try-catch to take cure of such scenarios.

### Context

Issue(s) closed by this pull request: closes #1799

- [ ] Yes
- [x] No

### What

Simple try-catch and adding warnings when zero division error happens.

### Why

This will prevent the process from failing and notify the user when a division error happens.

### How

Adds warning, keeps running.

### Test plan

Tests modified.
**NOTE** There's no concrete way to test this right now as the cause of this issue is resolved with refactored weather.py. However, after talking with Pooya it's agreed that we should still catch this and log a warning to prevent potential failures in the future.

### DIFF
--- a/RUFAS/routines/field/crop/crop_management.py
+++ b/RUFAS/routines/field/crop/crop_management.py
@@ -180,13 +180,16 @@ class CropManagement:
 
         try:
             fraction_cut = self.data.cut_biomass / self.data.biomass
-        except ZeroDivisionError as e:
+        except ZeroDivisionError:
             info_map = {"class": self.__class__.__name__, "function": self.cut_crop.__name__}
             warning_name = "Harvesting in crop management causes zero division error"
             warning_message = (
-                f"Variable: biomass in CropData has value: '{self.data.biomass}"
+                f"A zero division error occurred in the harvesting process of crop management when calculating "
+                f"fraction cut."
+                f"The variable 'biomass' in CropData has an invalid value: '{self.data.biomass}'. "
             )
             om.add_warning(warning_name, warning_message, info_map)
+
         self.data.biomass -= self.data.cut_biomass
         self._recalculate_biomass_distribution(roots_harvested)
 
--- a/RUFAS/routines/field/crop/crop_management.py
+++ b/RUFAS/routines/field/crop/crop_management.py
@@ -179,30 +179,6 @@ class CropManagement:
 
         try:
             fraction_cut = self.data.cut_biomass / self.data.biomass
-            self.data.biomass -= self.data.cut_biomass
-            self._recalculate_biomass_distribution(roots_harvested)
-
-            self.data.leaf_area_index = self.data.leaf_area_index * (1 - fraction_cut)
-            self.data.accumulated_heat_units = self.data.accumulated_heat_units * (1 - fraction_cut)
-
-            self.data.dry_matter_yield_collected = self.data.cut_biomass * collected_fraction
-            self.data.wet_yield_collected = self.data.dry_matter_yield_collected / (
-                self.data.dry_matter_percentage / 100
-            )
-
-            self.data.yield_residue = self.data.cut_biomass * (1 - collected_fraction)
-
-            if self.data.do_harvest_index_override:
-                self.data.yield_nitrogen = self.data.optimal_nitrogen_fraction * self.data.wet_yield_collected
-                self.data.yield_phosphorus = self.data.optimal_phosphorus_fraction * self.data.wet_yield_collected
-                self.data.residue_nitrogen = self.data.optimal_nitrogen_fraction * self.data.yield_residue
-                self.data.residue_phosphorus = self.data.optimal_phosphorus_fraction * self.data.yield_residue
-            else:
-                self.data.yield_nitrogen = self.data.yield_nitrogen_fraction * self.data.dry_matter_yield_collected
-                self.data.yield_phosphorus = self.data.yield_phosphorus_fraction * self.data.dry_matter_yield_collected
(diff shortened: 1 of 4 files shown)


RUFAS/routines/field/crop/crop_management.py
```diff
@@ -169,36 +169,51 @@ def cut_crop(self, collected_fraction: float = 0) -> None:
...
+        except ZeroDivisionError:
+            info_map = {"class": self.__class__.__name__, "function": self.cut_crop.__name__}
+            warning_name = "Zero division error in crop management"
+            warning_message = (
+                f"A zero division error occurred in the harvesting process of crop management when calculating "
+                f"fraction cut."
+                f"The variable 'biomass' in CropData has an invalid value: '{self.data.biomass}'. "
+            )
+
+            om.add_warning(warning_name, warning_message, info_map)
```
ew3361zh, line 191: I'd add this to OM as an error rather than a warning.

tests/soil_crop_tests/crop_tests/test_crop_management.py
```diff
@@ -540,3 +540,35 @@ def test_distribute_residue_nutrients(
...
+
+    with patch("RUFAS.output_manager.OutputManager.add_warning") as add:
+        crop.cut_crop(0.5)
+        crop.determine_biomass_cut_from_whole_plant.assert_called_once()
+        info_map = {"class": crop.__class__.__name__, "function": crop.cut_crop.__name__}
+        warning_name = "Zero division error in crop management"
+        warning_message = (
+            f"A zero division error occurred in the harvesting process of crop management when calculating "
+            f"fraction cut."
+            f"The variable 'biomass' in CropData has an invalid value: '0'. "
```
ew3361zh:
```suggestion
        warning_message = (
            "A zero division error occurred in the harvesting process of crop management when calculating "
            "fraction cut."
            "The variable 'biomass' in CropData has an invalid value: '0'. "
```

tests/soil_crop_tests/crop_tests/test_crop_management.py
```diff
@@ -540,3 +540,35 @@ def test_distribute_residue_nutrients(
...
+        yield_nitrogen_fraction=0.12,
+        yield_phosphorus_fraction=0.0092,
+        above_ground_biomass=75.0,
+    )
+
+    crop = CropManagement(data)
+    crop._recalculate_biomass_distribution = MagicMock()
+    crop.determine_biomass_cut_from_whole_plant = MagicMock(return_value=0)
+
+    with patch("RUFAS.output_manager.OutputManager.add_warning") as add:
```
ew3361zh: We've been trying to move away from doing these `with patch(...) as xyz:` statements since they've produced some unexpected results based on where you check the patch being called.

The pytest version is a little more predictable using the MockerFixture:
`patch_for_add_warning = mocker.patch("RUFAS.output_manager.OutputManager.add_warning")`

ew3361zh: Also, I think the issue with this test not passing has to do with how there's an om instantiated at the top of the file. If you reference that OM for mocking, it seems to work:

`patch_for_add_warning = mocker.patch("tests.soil_crop_tests.crop_tests.test_crop_management.om.add_warning")`

ew3361zh:
Overall looks like a good approach, Matthew - if you fix the test and change the warning to an error, I'd be good to approve. Please just re-request me when those are addressed and I'll take another look.

RUFAS/routines/field/crop/crop_management.py
matthew7838: Adding warnings is suggested by Pooya, I've checked with him.

ew3361zh: If we're adding warnings do we need to have any substitutes/defaults for the data yields/residues that aren't calculated?

Where is the division by zero error happening? Is it just this line?
`fraction_cut = self.data.cut_biomass / self.data.biomass`

If so, do we need to move the rest of the code out of the block, add a default `fraction_cut` and allow the rest of the code to still execute?

matthew7838: The cause of issue #1799 is because of unnecessary harvest happening when not enough weather data is provided. This code is updating things so I would think maybe we should just not update any number under such circumstances. However, I think this is something we should check with @PooyaHekmati. Maybe need SME's opinion on what to set for the substitutes/defaults.

allisterakun (approved):
Looks good!

RUFAS/routines/field/crop/crop_management.py
PooyaHekmati: At this point, we don't want to change any equations or numbers here. We just want to catch it, warn the user, and continue the process.

ew3361zh (approved):
LGTM, Matthew!
Please ensure the MyPy error count doesn't increase over main but otherwise should be good to go.

PR #1807, opened 2024-07-10
More_constraints

### LINKED ISSUE #1783: [Animal] Automatic ration formulation generating unrealistic rations
## Expected Behavior
RuFaS has the capability to generate a ration according to user-specified ingredients and inclusion rates and ranges (user-defined ration), or, RuFaS can automatically generate a ration (automatic formulation) from just a set of default or user-provided feeds. The goal of automatic formulation is to achieve the least-cost combination of feeds that meet animal requirements as predicted by RuFaS. Beyond meeting these two objectives, there are also minimum bounds for most components (e.g., min DMI, min CP supply), as well as maximums for some nutrients/values.

Recently, purchased feed prices in the default feeds file were set uniformly to 0.1, effectively removing price differentiation from the automatic formulation process. This can result in rations that technically meet animal requirements, but are unrealistic in practice, e.g., financially infeasible, gross over or undersupply of some nutrients, etc. 

## Current Behavior
The current default feeds file, if run using automatic formulation, results in a ration that is on average 90% DM, 40% CP, 25% NDF (the minimum allowed value), and < 2% starch.

## Possible Solution
We should re-implement appropriate purchased feed prices in the default feed file, at least temporarily, to prevent the above issue from occurring when automatic ration balancing is used.  

Beyond this, we should consider some other aspects that may require separate issues/solutions:
1. Should we implement additional bounds (e.g., for CP) for nutrients in automatic ration formulation?
2. When and where might we need to provide generic/default feed prices?

JoeWaddell:
### Context

Should partially resolve #1783 

- [ ] Yes
- [x] No

### What

Fixes to `MP_supply` calculations, following changes by Kristan on the `ration_protein_debugging` branch.

Added a new upper limit to MP supply in `ration_optimizer.py`.

Added logic to handle per ingredient minimum value set in Feed library.

Added clipping logic to restrict attempted ration in `ration_optimizer.py` to reduce number of `scipy.optimize.minimize` warnings.

### Why

Automated ration formulation was producing unrealistically protein-rich rations. In part this is due to ration prices

### Test plan

Used `input/data/feed/NASEM_Comp_with_TDN_testbed.csv` for testing purposes.
[NASEM_Comp_with_TDN_testbed.csv](https://github.com/user-attachments/files/16511504/NASEM_Comp_with_TDN_testbed.csv)
This library has modified minimum and limit values, such that we can constrain the ration formulation process more than what's on default.

Needs to be evaluated by SMEs experienced with ration formulation on farms, to see where we should set the new protein constraint, and if we should suggest limits in the Feed library, e.g. a minimum of 1kg for forage, upper limit of 15kg (though this kind of decision may be outside the scope of this PR, or what RuFaS itself should provide/suggest). See: https://github.com/RuminantFarmSystems/MASM/issues/1888

### Additional notes
The warning sometimes generated by scipy is an open issue for that project: https://github.com/scipy/scipy/issues/19362
Note that others found similar errors raised in a closed issue, as well: https://github.com/scipy/scipy/issues/11403
This warning seems to not impact output, though I'm unsure how to suppress it when running RuFaS.


RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -1110,7 +1114,7 @@ def get_metabolizable_protein(
             1000 * 0.13 * TDN_total_actual, 1000 * 0.85 * sum(RDP_diet)
         )
 
-        MP_supply = MP_bact + sum(RUP_diet) * 0.0001 + 0.4 * 11.8 * DMI_estimate
+        MP_supply = MP_bact + sum(RUP_diet) * 0.1 + 0.4 * 11.8 * DMI_estimate
```
KFosterReed: One thing I couldn't understand was why this 0.1 was here :/ I can't find any support for it in the not-pseudocode.  It also makes me wonder if we are still off by a factor of 10 and on line 1106, it should be: 

` RUP_diet.append(kg_fed * RUP_list[i]*0.01 * dRUP_diet[i])`

If I am folllowing the logic:
RUP_list is the percent of the feed that is RUP (calculated as the diffrence between total diet CP% and diet RDP%)

dRUP_diet is a list of the intestinal digestibilities of RUP (i.e. what percent of the feed CP that makes it to the small intestine is digested and absorbed by the animal)

KFosterReed: Having the above conversation with myself helped me understand and now I think the unit conversions for RUP are really fixed - still the same numbers but the conversions are separated and in the places where it makes more sense to me and hopefully to others :)

elle-andreen:
For the new/modified protein constraints, to make sure I'm understanding these properly - am I correct in saying:
 - The lower MP supply constraint is equal to the MP requirement (e.g., ration must meet the requirement)
 - The upper MP supply constraint is equal to the one "full" MP requirement above the MP supplied? (E.g. MP requirement x 2 - MP supply)

I will say I think basing this constraint on MP requirement and supply is a good move, protein is a tricky nutrient (although they all are, really) to build guidelines around. As is, if I swap in the default feed prices from PR 1836, I don't think the constraint is coming into play, I'm still seeing 16-18% CP values there (which are perfectly fine).

RUFAS/routines/animal/ration/ration_optimizer.py
```diff
@@ -627,7 +628,7 @@ def protein_constraint(  # noqa
         ))
         # [A.Cow.E.14]-[A.Heifer.E.14]
         # Dietary RUP (kg)
-        ration_config.RUP_diet = sum(
+        ration_config.RUP_diet = 1000 * sum(
             np.multiply(
```
tomhuhh, line 620: We could use a constant variable here?

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -1109,19 +1113,22 @@ def get_metabolizable_protein(
...
         )
 
         # MP bact calcs
         MP_bact = 0.64 * min(
-            1000 * 0.13 * TDN_total_actual, 1000 * 0.85 * sum(RDP_diet)
+            GeneralConstants.KG_TO_GRAMS * 0.13 * TDN_total_actual, 0.85 * sum(RDP_diet)
         )
 
-        MP_supply = MP_bact + sum(RUP_diet) * 0.0001 + 0.4 * 11.8 * DMI_estimate
+        MP_supply = MP_bact + sum(RUP_diet) + 0.4 * 11.8 * DMI_estimate
```
tomhuhh, line 1158: I'm not quite following the logic here to completely remove the multiplier. In Kristan's revision, my understanding is that she multiplied the multiplier by 1000 to convert 0.0001 to 0.1.

tomhuhh (commented):
Thanks Joe for looping me in. 

Suggested some minor changes for now. Will keep an eye on this PR and review it when it is made official. 

Just curious, what is the reference/rationale behind the suggested upper constraints for protein? This is related to Elle's comments. I definitely need to revisit NASEM for reviewing these changes.

RUFAS/routines/animal/ration/ration_driver.py
JoeWaddell: The intent was to make the change upstream, e.g. in the RUP calculations - note the additional conversion of kilograms to grams. It could still be that there's a missing or extra step, so thank you for taking a close look at each step like this!

RUFAS/routines/animal/ration/ration_optimizer.py
JoeWaddell, line 620: So in reviewing the documentation (https://3.basecamp.com/3486446/buckets/5296287/google_documents/7108678721, since the more recently updated one: https://3.basecamp.com/3486446/buckets/5296287/google_documents/7108678956 has apparently been deleted!), I'm struggling to understand the logic we have in the code right now, specifically re: the conversion of RUP and dRUP to fractions, when they were percentages.

This happens both here and in the `get_metabolizable_protein` method, which should follow the same logic (they are being applied to different objects, hence different methods - though streamlining this would be ideal, see https://github.com/RuminantFarmSystems/MASM/issues/1869)

In the documentation it shows that RUPdiet should be in kg, as should feed, and dRUP is a percentage of RUP. So I'm unsure why we should convert RUP and dRUP to fractions, or why RUP should be apparently in % in this case. Perhaps related to the change we've made to convert RUP to grams (to be in the same unit as MPbact), but I'm unsure, and think other SMEs on this PR may be able to clarify, but perhaps an error that could have led to some of the MP problems we've noted here.

RUFAS/routines/animal/animal_typed_dicts.py
```diff
@@ -253,7 +253,9 @@ class AvailableFeedsTypedDict(TypedDict):
     N_B: List[float]
     CP: List[float]
     dRUP: List[float]
+    lactating_cow_minimum: List[float]
     lactating_cow_limit: List[float]
+    dry_cow_minimum: List[float]
```
Isgotkowitz, line 260: Looks like the documentation is lacking for a lot of these typed dicts, but I would still strongly encourage starting an attributes sections and adding these new attributes to it.

RUFAS/routines/animal/ration/ration_config.py
```diff
@@ -210,6 +211,7 @@ def __init__(
         self.N_B_list = N_B__list
         self.CP_list = CP__list
         self.dRUP_list = dRUP__list
+        self.feed_minimum_list = feed_minimum__list
```
Isgotkowitz, line 222: Same comment on the units in this file.

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -1109,19 +1109,23 @@ def get_metabolizable_protein(
         RDP_diet = []
         RUP_diet = []
         for i, kg_fed in enumerate(ration.values()):
-            RDP_diet.append(RDP_list[i] * kg_fed * 0.01)
-            RUP_diet.append(kg_fed * RUP_list[i] * dRUP_diet[i])
+            RDP_diet.append(GeneralConstants.KG_TO_GRAMS * RDP_list[i]
+                            * kg_fed * GeneralConstants.PERCENTAGE_TO_FRACTION)
+            RUP_diet.append(GeneralConstants.KG_TO_GRAMS * kg_fed
+                            * RUP_list[i] * GeneralConstants.PERCENTAGE_TO_FRACTION
+                            * dRUP_diet[i] * GeneralConstants.PERCENTAGE_TO_FRACTION)
```
Isgotkowitz, line 1147: These calculations seem involved enough that I think they merit their own line/variable, i.e. 
```python
rdp_diet_amount = GeneralConstants.KG_TO_GRAMS * RDP_list[i] * kg_fed * GeneralConstants.PERCENTAGE_TO_FRACTION
RDP_diet.append(rdp_diet_amount)
```

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -1171,6 +1175,10 @@ def __init__(self) -> None:
         self.lactating_cow_limit: List[float] = []
         # dry cow feed limits
         self.dry_cow_limit: List[float] = []
+        # lactating cows feed limits
+        self.lactating_cow_minimum: List[float] = []
+        # dry cow feed limits
+        self.dry_cow_minimum: List[float] = []
```
Isgotkowitz: Are there known units associated with these minimums?

RUFAS/routines/animal/ration/ration_optimizer.py
```diff
@@ -641,7 +642,26 @@ def protein_constraint(  # noqa
...
             ration_config.MPbact + ration_config.RUP_diet + 0.4 * 11.8 * DMI
         )
-        return ration_config.MP_supply - (ration_config.MP_requirement / 1000)
+        return ration_config.MP_supply - (ration_config.MP_requirement)
+
+    @staticmethod
+    def protein_constraint_upper(decision_vector: np.ndarray,
+                                 ration_config: RationConfig) -> float:
+        """
+        Sets up the upper bound for the protein requirement constraint in the NLP. Uses MP_supply as calculated in
```
Isgotkowitz: Is NLP non-linear programming?

RUFAS/routines/animal/ration/ration_optimizer.py
```diff
@@ -641,7 +642,26 @@ def protein_constraint(  # noqa
...
+
+        Parameters
+        ----------
+        ration_config: RationConfig object
+            Attributes are animal requirement and feed supply information required for optimization
+
+        Returns
+        -------
+        float
+
```
Isgotkowitz:
```suggestion
        float
            Upper protein bound for ration formulation (<units>).

```

RUFAS/routines/animal/ration/ration_optimizer.py
```diff
@@ -945,8 +965,15 @@ def optimize(
             )
             x0 = [np.mean(bnd) for bnd in bnds]
         else:
-            bnds = []
-            bnds = [(0, (lim / 3) + 0.0001) for lim in ration_config.feed_limit_list]
+            # bnds = []
+            # bnds = [(0, (lim / 3) + 0.0001) for lim in ration_config.feed_limit_list]
+            bnds = list(zip(
+                [(lim / 3) for lim in ration_config.feed_minimum_list],
+                [(lim / 3) for lim in ration_config.feed_limit_list]))
```
Isgotkowitz:
```suggestion
            lower_bounds = [(lim / 3) for lim in ration_config.feed_minimum_list]
            upper_bounds = [(lim / 3) for lim in ration_config.feed_limit_list]
            bnds = list(zip(lower_bounds, upper_bounds))
```
Why is there a / division by 3 here? It would be good to stick this value in variable and name it to make it clearer,  if possible.

RUFAS/routines/animal/ration/ration_optimizer.py
```diff
@@ -1024,14 +1051,16 @@ def attempt_optimization(
...
+            feed_limit_list = self.triple_values_in_list(available_feeds["dry_cow_limit"])
+            feed_minimum_list = self.triple_values_in_list(available_feeds["dry_cow_minimum"])
+            lactating = False
         else:
-            feed_limit_list = self.triple_values_in_list(
-                available_feeds["dry_cow_limit"]
-            )
+            feed_limit_list = self.triple_values_in_list(available_feeds["dry_cow_limit"])
+            feed_minimum_list = self.triple_values_in_list(available_feeds["dry_cow_minimum"])
             lactating = False
```
Isgotkowitz: The `elif` and `else` branches have the same logic, is there something that needs to be added to one of them?

RUFAS/routines/animal/ration/ration_optimizer.py
```diff
@@ -641,7 +642,26 @@ def protein_constraint(  # noqa
...
+        ----------
+        ration_config: RationConfig object
+            Attributes are animal requirement and feed supply information required for optimization
+
+        Returns
+        -------
+        float
+
+        """
+        return (ration_config.MP_requirement * 2) - ration_config.MP_supply
```
Isgotkowitz: Why is `MP_requirement` doubled?

RUFAS/routines/animal/ration/ration_optimizer.py
```diff
@@ -641,7 +642,26 @@ def protein_constraint(  # noqa
         ration_config.MP_supply = (
             ration_config.MPbact + ration_config.RUP_diet + 0.4 * 11.8 * DMI
         )
-        return ration_config.MP_supply - (ration_config.MP_requirement / 1000)
+        return ration_config.MP_supply - (ration_config.MP_requirement)
```
Isgotkowitz, line 633: Is there a units conversion that still needs to happen here? Or was it wrong before to be dividing by 1000?

Isgotkowitz (commented):
Looks like this PR is coming along nicely. I left comments on some things that I think would help make the code more readable, but unfortunately I have nothing more substantive to offer than that at this time.

RUFAS/routines/animal/ration/ration_optimizer.py
JoeWaddell: That was a pseudo-arbitrary upper limit bound- e.g. protein content should be no more than double the minimum, which might be reasonable. Definitely open to diverse SME input, and I'll need to update the docstring to explain!

RUFAS/routines/animal/ration/ration_optimizer.py
JoeWaddell: The original division by 1000 was happening when it shouldn't have: was converting to kg when supply was in grams.

RUFAS/routines/animal/ration/ration_optimizer.py
JoeWaddell: Excellent question! https://github.com/RuminantFarmSystems/MASM/pull/1846

Older code, and the reasoning has been deprecated. I'll have to flesh out the description in 1846, which was a test to ensure that the tripling wasn't happening for valid/extant reasons.

RUFAS/routines/animal/ration/ration_optimizer.py
JoeWaddell: Correct! Can update the docstring accordingly.

RUFAS/routines/animal/ration/ration_driver.py
JoeWaddell: The limits and minimums are kg, good call.

RUFAS/routines/animal/ration/ration_optimizer.py
JoeWaddell: Nope: I'll reduce the code here...I had left it in as a template for future additions (e.g. lists for other animal types), but I don't see that on the horizon, so better to avoid confusion!

RUFAS/routines/animal/ration/ration_optimizer.py
tomhuhh: Hi Joe! I think this makes sense to me. This is how I interpret it. 

`np.multiply(ration_config.RUP_list, 0.01)` gives us a list of RUP in fractions. For example, if for ingredient i, the RUP content (% DM) is 60%, this equation converts 60 to 0.6. Similarly,  `np.multiply(ration_config.dRUP_list, 0.01)` converts digestible RUP (dRUP, % RUP) from percentage to fraction as you described. This equation, `np.multiply(np.multiply(ration_config.RUP_list, 0.01), np.multiply(ration_config.dRUP_list, 0.01),)` calculates the fraction of dRUP on DM basis. So if dRUP is 60% of RUP, then dRUP is 60% * 60% (=36% = 0.36) of DM. Finally this equation, ```np.multiply(decision_vector, np.multiply(np.multiply(ration_config.RUP_list, 0.01), np.multiply(ration_config.dRUP_list, 0.01),),)``` calculates dRUP in kg. Hope this clarifies.

tomhuhh: `decision_vector` in my interpretation is the DM intake for ingredient i.

RUFAS/routines/animal/ration/ration_optimizer.py
Isgotkowitz: In addition to putting it in the docstring, it would also be good to stick 2 in its own variable with a descriptive name - possibly something like "protein_upper_limit_generation_factor" (or something better).

elle-andreen:
Wanted to drop some thoughts here after yesterday's animal module meeting:

Firstly, I think it may be worth waiting to institute new constraints until the amino acid requirements and supply functionality is implemented (unless the timeline is too long?), I think that will present a whole new major framework to work with, and may both solve and create issues that could have downstream effects on any constraints we add here.

I had not really experimented with the auto formulation, so I wanted to put in a wider variety of ingredients that I felt could be assembled into a reasonable ration in multiple ways, and see what happened. Here is an output with the feeds, the prices I used, the inclusion amount of each feed over time, and the ration composition on average. 
[Test ration output.xlsx](https://github.com/user-attachments/files/16443916/Test.ration.output.xlsx)


Things that came to mind based on this output:
- The nutritive composition of the ration here is not too far off (45% NDF, 22.5% CP, 5.8% starch) but the ingredient composition is. The ration is about 25% forage DM, 15% energy feeds (corn, molasses, cookie meal), and 60% wet brewers grains (lol). It seemed to pull in the minimum amount of forage (about 7-8 pounds of silage and/or hay), make most of the rest up with brewers grains, and then meet the lacking energy requirements with corn/molasses once it hit the upper constraint for NDF%. 
- A simple new constraint that comes to mind from this is dietary dry matter %. A TMR should be about 50% DM, though if animals are exclusively grazing or grazing and receiving silage/concentrate separately, the net DM could be lower. Maybe a lower bound of 35% and upper bound of 65%, depending on others' thoughts? I don't know exactly if/how this would improve things other than just not generating unrealistically wet rations like the above (29% DM). 

- The best general idea I've had so far for constraints improvement is to try implementing % inclusion rate bounds by feed category. We have constraints for semi-categorical things like forage NDF, but they are fixed. We could categorize each feed in the library into something like:

Forages 
Energy feeds
Protein feeds
By-product feeds 
Supplements (e.g., amino acids)

and require the user to specify a range (% DM) for each category. I think this would at least ensure a feed from each category is present in the ration, instead of the optimizer being able to do something like the above, and would prevent it from using such a huge quantity of non-forage feeds. We would have to assess how easy or difficult of a time the optimizer has formulating within these bounds, and it could make utilizing user feeds with the auto formulation more challenging. I would be interested in knowing others' thoughts and, if it's not too difficult to just add that functionality into the optimizer, it would be fun to play around with that and see what happens.

JoeWaddell:
> I think it may be worth waiting to institute new constraints until the amino acid requirements and supply functionality is implemented (unless the timeline is too long?),

I think the timeline for implementing the AA requirement/supply is too far out, as we don't want this branch to get "stale" and harder to merge without conflicts.

For the other constraints you suggest: great ideas! I think they'd be worth entirely separate Issues/PRs

The one re: dietary dry matter % would be straightforward enough to implement general bounds like you suggest.

For the one where the user would define a DM range for each category, I think that might get too far into user-defined ration territory, unless we come up with general bounds that would be applicable near-universally. That said, some intermediate step between user-defined and fully automatic would be an interesting approach, so I think it's a great concept that you'd laid out!

KFosterReed:
This is looking much better! Also, thanks to @elle-andreen for the great review and suggestions! I ran it with a couple of the feeds in the default diets and, while not something you would feed a real cow, the compositions are quite reasonable and the recipes are better! Unlike Elle's example, my scenarios were hitting the upper bounds for forage inclusion (I had forages at $0.01/kg DM). 

I would agree with Joe that I don't want to add even more constraints to this more constraints PR :) I think the total diet DM% could be a useful one. 

My suggestion for either this PR or an improvement to shortly follow would be to modify the default upper and lower limits to be more reasonable based on feed / category. For example, this scenario was calling for 6 kilos of whey which is too much. Also, it is not unreasonable to have more than 10 kilos of corn silage in a diet. 



image

JoeWaddell:
Thank you for the review! Glad it looks like it's improving things.

> My suggestion for either this PR or an improvement to shortly follow would be to modify the default upper and lower limits to be more reasonable based on feed / category.

Heartily agreed! The minimums and limits I added were to really challenge it, so I was going to 1) remove that modified Feed library prior to merge, and 2) (following approval) open an Issue to set minimums/limits where reasonable. 

> For example, this scenario was calling for 6 kilos of whey which is too much.

...*whey* too much?

RUFAS/routines/animal/ration/ration_config.py
```diff
@@ -176,6 +177,8 @@ def __init__(
             RUP degradability in each feed (% of RUP).
         limit__list : list, optional
             Limiting upper bounds for each feed (kg).
+        feed_minimum_list : list, optional
```
Isgotkowitz:
```suggestion
        feed_minimum__list : list[float], default []
```
I think it would be better to not refer to this argument as "optional" because this term in Python is a little ambiguous. "optional" can mean that `None` is an acceptable value for this parameter, or it can mean that the caller function doesn't have to specify a value for this argument. In this case, it is only the latter and I think this is indicated more explicitly by saying what the default is.

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -707,7 +707,7 @@ def get_DE(
         DE_act = feed_item_info[de_key] * RationReporter.get_TDN_discount(
             ration_report, body_weight
         )
-        return DE_act
+        return DE_act * kg_fed
```
Isgotkowitz, line 707: Are the units of the returned value still Mcal / kg, or are they just Mcal now?

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -734,11 +734,15 @@ def get_ME(
         -------
         float
             metabolizable energy of feed i, Mcal/kg.
+        Notes
```
Isgotkowitz, line 735:
```suggestion
        
        Notes
```

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -734,11 +734,15 @@ def get_ME(
         -------
         float
             metabolizable energy of feed i, Mcal/kg.
```
Isgotkowitz: Same question as above, is this still Mcal/kg or is it just Mcal now?

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -780,10 +784,15 @@ def get_NE_maintenance_and_activity(
         float
             Net energy of feed i, Mcal/kg.
```
Isgotkowitz: I'm very much beating a dead horse here, but again Mcal/kg or just Mcal?

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -865,10 +878,15 @@ def get_NE_growth(
...
 
+        Notes
+        -----
+        Division of ME_item by kg_fed is to get the ME for each item on a per kg basis,
```
Isgotkowitz:
```suggestion
        Division of ME_item by kg_fed is to get the metabolizable energy for each item on a per kg basis,
```
I'm not sure if this is correct or not, but expanding abbreviations like this would really help non-SMEs understand what is going on.

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -1124,19 +1142,23 @@ def get_metabolizable_protein(
...
 
         TDN_total_actual = TDNtotal * RationReporter.get_TDN_discount(
             ration_report, body_weight
         )
 
         # MP bact calcs
         MP_bact = 0.64 * min(
-            1000 * 0.13 * TDN_total_actual, 1000 * 0.85 * sum(RDP_diet)
+            GeneralConstants.KG_TO_GRAMS * 0.13 * TDN_total_actual,
+            GeneralConstants.KG_TO_GRAMS * 0.85 * sum(RDP_diet)
```
Isgotkowitz: These two calculations feel a little magic-numbery, I think breaking up this `min()` statement into the something like the following would help:
```python
metabolizable_protein_TDN_amount = GeneralConstants.KG_TO_GRAMS * 0.13 * TDN_total_actual
metabolizable_protein_RDP_amount = GeneralConstants.KG_TO_GRAMS * 0.85 * sum(RDP_diet)
MP_bact = 0.64 * min(metabolizable_protein_TDN_amount, metabolizable_protein_RDP_amount)
```

RUFAS/routines/animal/ration/ration_optimizer.py
```diff
@@ -622,12 +623,12 @@ def protein_constraint(  # noqa
         # [A.Cow.E.13]-[A.Cow.E.13]
         # Metabolizable bacterial protein production (g)
         ration_config.MPbact = float(0.64 * min(
-            1000 * 0.13 * ration_config.TDNact_diet,
-            1000 * 0.85 * ration_config.RDP_diet,
+            GeneralConstants.KG_TO_GRAMS * 0.13 * ration_config.TDNact_diet,
+            GeneralConstants.KG_TO_GRAMS * 0.85 * ration_config.RDP_diet,
         ))
```
Isgotkowitz: Same suggestion for breaking this up into individual calculations from the other implementation of computing MPbact

RUFAS/routines/animal/ration/ration_optimizer.py
Isgotkowitz: This still needs to be addressed

Isgotkowitz (approved):
Code changes LGTM, nice job Joe. Found a number of spots where a value is multiplied by an amount in kg before being returned but the docstring lists their units as Mcal / kg - this didn't seem right to me but I might just be missing something.

Re: the Scipy warning, they can be captured like this:
```python
import warnings
...
with warnings.catch_warnings(record=True) as caught_warnings:
    result = minimize(
        self.objective,
        x0,
        method="SLSQP",
        bounds=bnds,
        constraints=self.heifer_constraints,
        args=arguments,
    )
    for warning in caught_warnings:
        om.add_warning(
            f"Captured warning during optimization of type {warning.category.__name__}",
            f"{warning.message}. Warning generated in {warning.filename}",
            {"class": self.__class__.__name__, "function": self.optimize.__name__, "full_warning": warning}
        )
```

> ...whey too much?

Love it

RUFAS/routines/animal/ration/ration_driver.py
JoeWaddell: You're correct, thanks for the catch!

RUFAS/routines/animal/ration/ration_driver.py
JoeWaddell: I'm hesitant to do that since we'd be discarding `rdp_diet_amount` on each iteration in the enumerate loop, so I'm not sure what we'd gain (since they're on their own lines already).

RUFAS/routines/animal/ration/ration_optimizer.py
JoeWaddell: It'll have to be unitless, and with a different explanation for each constraint...since here the return indicates if the constraint is passed or not. Non-negative if the supply is below the upper limit, in this case

RUFAS/routines/animal/ration/ration_config.py
JoeWaddell: Excellent suggestion!

RUFAS/routines/animal/animal_typed_dicts.py
JoeWaddell: Good call: I've opened a new issue regarding this, since I think it'll be best to do these all at once, and with consideration to what might change in ongoing Feed refactors

https://github.com/RuminantFarmSystems/MASM/issues/1917

KFosterReed (approved):
This is a big improvement and when lower limits are added, it produces pretty reasonable rations!

JoeWaddell:
> This is a big improvement and when lower limits are added, it produces pretty reasonable rations!

Great! I opened an issue about that here: https://github.com/RuminantFarmSystems/MASM/issues/1888

PR #1808, opened 2024-07-11
Add Units to Var Names File

### LINKED ISSUE #1749: [Output Manager] Add units to variable names in variable names file
Users need to know what units are associated with a particular variable. It would make sense to have it added to the var names file. 

Have the units listed for each variable in the file generated by `OM.dump_variable_names_and_contexts()`. I think it would be helpful to have them added regardless of format option but the person addressing this issue can look at the feasibility and necessity of that (i.e. maybe it doesn't make sense to have the units with every format option in case that leads to more confusion because of how the vars are displayed).

ew3361zh:
Adds variable units to variable names file generated by OM.

### Context

Issue(s) closed by this pull request: closes #1749

- [ ] Yes
- [x] No

### What

For every simulation the OM generates a file with a list of all the variables added to the OM variables pool. With units data now available in info_maps, that units data can be added to the variables name file.

### Why

The information of what units a variable is logged to OM with is vital to the accuracy and relevancy of the outputs from the simulation. Having a quick resource to spot check what units a variable has can ensure the data is being added and reported in the ways we expect. 

### How

Within the OM's `dump_variable_names_and_contexts()` method which is how the data for the variable names file is assembled and sent to be saved the formatting has been updated to include the variable units for each of the format (`variable_name_style`) options (inline, basic, verbose, and block).

### Test plan

A lot of the decisions of how to add the units for each `variable_name_style` are subjective style preferences. When testing, please take a look at each option.

This is done in the `default_task` (or whatever task file you're using) by specifying the `variable_name_style`. Please run a simulation or multiple single simulations within one task with each of the style options (the default is `basic`).
We have `verbose`, `inline`, `block`, and `basic`.

Look at the resulting var names file(s) in output > logs > `output_prefix.variable_names_datetime.txt`.


Isgotkowitz (approved):
I think the units look great in all the different formatting options - however while testing out this code I found that a crash occurs if `exclude_info_maps` is true. Obviously in this case, the units can't be output because they're not recorded, so the code needs to adjusted for that but otherwise I think it is good to go.

If anything really substantive needs to be changed please let me know and I can re-review.

ew3361zh:
> I think the units look great in all the different formatting options - however while testing out this code I found that a crash occurs if `exclude_info_maps` is true. Obviously in this case, the units can't be output because they're not recorded, so the code needs to adjusted for that but otherwise I think it is good to go.
> 
> If anything really substantive needs to be changed please let me know and I can re-review.

I addressed this. I had mistakenly been running the exclude info maps for TM only and thinking it worked. Thanks, Ed!

JoeWaddell (commented):
The PR as it was looked great, clean code, and functions as expected (and is a great addition, since this should provide clarity to users).

However: looks like we may be dealing with some of the unintentional merge problems again, as there are a bunch of unrelated animal health files now included: maybe #1767 was rolled back unintentionally? I'll hold on on approval until that's resolved/clarified

ew3361zh:
> The PR as it was looked great, clean code, and functions as expected (and is a great addition, since this should provide clarity to users).
> 
> However: looks like we may be dealing with some of the unintentional merge problems again, as there are a bunch of unrelated animal health files now included: maybe #1767 was rolled back unintentionally? I'll hold on on approval until that's resolved/clarified

Hey @JoeWaddell, I think this is just a Github tracking issue - AnimalHealth (#1767) is still a part of main. There are certain actions that you can take in Github to squash or consolidate commits to make PRs cleaner but when you do this, it rewrites the commit history by assigning different commit hashes for some reason. So then when another PR like mine has not done the same squashing, it falls out of sync with the commit history on main despite the code itself being the same.

These usually resolve when the next PR gets merged in.

JoeWaddell (approved):
With the clarification on the apparent animal_health inclusion in this PR, LGTM!

PR #1809, opened 2024-07-11
Extract validation logic

### LINKED ISSUE #1800: [1/2][IM] Validating Outputs: Extract validation logic from IM
We want to reuse the validation logic from IM to validate outputs. As the first step, extract the validation logic from the Input Manager and put it in a new class. Make calls to that class from Input Manager. This new validation class is agnostic about the source of the data. It receives data and validation rules (currently known as metadata), performs the validation, and returns the result. It has to be maintainable, extendable, and scalable. All SOLID principles are expected to be implemented in it. Most likely, its methods need to be static.

matthew7838:
Extracts the validation logic from the InputManager.
### Context

Issue(s) closed by this pull request: closes #1800

- [ ] Yes
- [x] No

### What

- Every method that relates to the validation process in [This graph](https://whimsical.com/input-manager-data-validation-overview-T6TuSktxPUYe3BRUURhaDm) has been extracted and put into a separate file named `data_validator.py`.
- Elements and ElementCounter have been moved to a file named `data_validator.py`.
- Modifiability has been moved to a new file named `data_validator.py`.
- All validation methods are now static
- Some methods to extract data have also been moved to `data_validator.py` to avoid circular imports.
- Updated tests and fixed the usage of patching with `with()`.

### Why

`input_manager.py` is way too complicated and has too many classes. This change should follow closer to the SOILD principle.

### How

These changes will make the IM closer to a single-purpose class. The split of major functions (validation) will also make the code more maintainable. Extraction of classes in `input_manager.py` also prevents circular imports.

### Test plan

- Ran with a complete test suite.
- Ran with default metadata and it still works the same.


allisterakun (approved):
Great work, Mathew! This cleans up the IM! 
Just one take-or-leave suggestion: can we maybe put all the IM-related Python files into a folder so that we don't end up with many .py files lying around under the main RUFAS folder?

matthew7838:
> Great work, Mathew! This cleans up the IM! Just one take-or-leave suggestion: can we maybe put all the IM-related Python files into a folder so that we don't end up with many .py files lying around under the main RUFAS folder?

I think this is a good idea, I will double-check with Pooya about this in 1:1 since it involves a bit of structure change.

PooyaHekmati:
Hi @matthew7838 
Does this need to be draft still?

matthew7838:
> Hi @matthew7838 Does this need to be draft still?

No, my bad for forgetting to bring it out of draft.

PooyaHekmati (changes requested):
1. The file organization is not clear. Having `elements.py` and `modifiability.py` under RUFAS is confusing. Besides their un-informative and un-descriptive names, why do they need to be in their own files? why not having everything in one file?
2. The purpose of this logic extraction is to make it accessible by the other sub-systems (OM). This PR; however, does not facilitate that. It still calls the logic `input validation`. As stated in issue #1800 :
>This new validation class is agnostic about the source of the data. 

and

> It has to be maintainable, extendable, and scalable.

3. Based on the following:
>It receives data and validation rules (currently known as metadata), performs the validation, and returns the result.

these two function calls are not correct as they don't return anything:
```InputValidator.validate_metadata(self.__metadata)```
and 
```InputValidator.validate_properties(self.__metadata, self.metadata_depth_limit)```

File: RUFAS/data_validator.py [hunk 1]
```diff
@@ -0,0 +1,1477 @@
...
+import re
+from enum import Enum
+from functools import reduce
+from typing import Dict, Any, Callable, List, Union, Sequence, Tuple
+
+from RUFAS.output_manager import OutputManager
+
+
+#  log error here, leave termination for IM
+#  Move elements counters...etc to here
```
allisterakun:
```suggestion
```
remove comments

RUFAS/data_validator.py
```diff
@@ -0,0 +1,1477 @@
...
+        """Iteratively traverses the metadata properties to check the max depth and routes
+        properties to be validated by type.
+
+        return
+        ------
+        Tuple[bool, str]
+            boolean to indicate the validation status, error message in str if there's error that should be raised by
+            the caller.
+        """
+        om = OutputManager()
```
allisterakun, line 193: Do we have to create an instance like this for all the static methods that report to the OM every time they are called? Is there a workaround? I tried deleting all the `om = OutputManager()` in each function and just have one definition of a global `om = OutputManager()` at the beginning of the file. The simulation seems to run without crashing and the logs seem to be reported correctly. So... just something to look into!

allisterakun (approved):
It generally looks look, great job, Matthew! I left a comment about removing the repeating om definition.

Also, please consider updating the [data validation diagram](https://whimsical.com/input-manager-data-validation-overview-T6TuSktxPUYe3BRUURhaDm) and the PR description before merging.

RUFAS/data_validator.py
matthew7838: I've discussed this with Pooya, we will stick to this way for now. We want to avoid global initialization and OM is a singleton; therefore, initializing them like this won't impact the performance.

RUFAS/data_validator.py
```diff
@@ -0,0 +1,1473 @@
...
+                return False, f"Invalid path '{data['path']}' in '{key}'"
+
+            if data["properties"] is None or data["properties"] == "":
+                om.add_error("Metadata Validation", f"Properties section empty or None in '{key}'", info_map)
+                return False, f"Properties section empty or None in '{key}'"
+
+        om.add_log("Metadata Validation", "Top level metadata is valid.", info_map)
+        return True, ""
+
+    # Validate input by type related
```
Isgotkowitz, line 617:
```suggestion
```

RUFAS/data_validator.py
```diff
@@ -0,0 +1,1473 @@
...
+    @staticmethod
+    def _metadata_object_validator(key_path: list[str], value: dict[str, Any]) -> Tuple[bool, str]:
+        """Validates object type properties in metadata."""
+        required_object_property_keys = {"type"}
+        optional_object_property_keys = {"description"}
+        valid, message = DataValidator._validate_metadata_properties_keys(
+            required_object_property_keys, optional_object_property_keys, value, key_path
+        )
+        if not valid:
+            print(message)
```
Isgotkowitz:
```suggestion
```

RUFAS/data_validator.py
```diff
@@ -0,0 +1,1473 @@
...
+    @classmethod
+    def get_required_during_initialization(cls) -> List["Modifiability"]:
+        return [Modifiability.REQUIRED_LOCKED, Modifiability.REQUIRED_UNLOCKED]
+
+    @classmethod
+    def get_modifiable_at_runtime(cls) -> List["Modifiability"]:
+        return [Modifiability.REQUIRED_UNLOCKED, Modifiability.UNREQUIRED_UNLOCKED]
+
+
+class DataValidator:
```
Isgotkowitz, line 179: Could you please add a docstring for `DataValidator`? It is well documented and named otherwise, so a one-liner should be sufficient IMO

Isgotkowitz (approved):
Compared simulation results and logs produced between this branch and `dev`, seemed to be exactly the same except for `InputManager` being changed to `DataValidator` - as expected.

The only substantive thing that should be addressed before merging are the test suites, because the coverage has gone down a bit. In particular, `_log_missing_data`, `_is_data_required_upon_initialization` and `_get_variable_modifiability` appear to not be covered at all. Ideally though both `DataValidator` and `InputManager` would have 100% test coverage.

Nice job Matthew!

PooyaHekmati (approved):

PR #1813, opened 2024-07-11
Add outputs required by FARMES

### LINKED ISSUE #1812: Add Manure Nutrient Request to Outputs and Homegrown Feed Emissions reports
The FARM ES team would like these outputs added to the report filter they are using: 
- total manure nitrogen requested for application to homegrown feeds in the last year of the simulation
- total yields of the homegrown feeds in the last year of the simulation
- total soil carbon stock change in the last year of the simulation

Kristan has implemented this in a testing branch and will create a draft PR shortly. She would appreciate SWE support for reviews, unit testing, and code clarity support!

**Additional context**

**Target Deadline**
July 17th

KFosterReed:
This PR does the following:

- Adds the amount of manure nutrients that were requested to the  `record_manure_application` method in `field.py`
- Adds the amounts of homegrown feeds calculated in to the OM from `calculate_total_homegrown_feed_amounts_by_crop_type` in `emissions.py`
- Changes the units of the `carbon_stock_change` variable in the `calculate_homegrown_feed_emissions` method from kg/ha to kg/field

### Context

Issue(s) closed by this pull request: closes #1812 

Integration with the FARMES platform is relying heavily on the use of the report function to summarize the results they will use for their reporting. They have requested that we provide the following outputs directly from the report filter they will use:
- the total manure N that was requested for application for each crop in addition to the actual manure N that was applied
- the total yields of each homegrown feed _from the last year of the simulation_
- the total carbon stock change associated with each feed instead of the kg of carbon per ha

- [ ] Yes
- [x] No

### Test plan

Running the FARMES_fakefarm1_wcrops_metadata file and the FARMES_July2 report filter should produce a report with the following variables:

- "Homegrown Feed Emissions - Corn Silage_manure_nitrogen_requested_ver_agg"
- "Homegrown Feed Totals_EmissionsEstimator._calculate_total_homegrown_feed_amounts_by_crop_type.homegrown_feed_totals.corn_silage"
- And a value of over 16,000 for the output variable  "Homegrown Feed Emissions - Corn Silage_carbon_stock_change_ver_agg"


RUFAS/routines/field/field/field.py
```diff
@@ -639,11 +639,26 @@ def _execute_manure_application(
...
+            field_coverage=field_coverage,
+            nitrogen=requested_nitrogen,
+            phosphorus=requested_phosphorus,
+            potassium=None,
+            application_depth=application_depth,
+            surface_remainder_fraction=surface_remainder_fraction,
+            year=year,
+            day=day,
+            output_name="manure_request",
+        )
```
ew3361zh, line 660: If I'm correctly understanding the intent of adding this, I am going to indent this to be added only if the else statement is triggered (otherwise I think we'd be adding an empty manure application both if there's `manure_supplied` or not).

KFosterReed: Hmmm, I would have to test it out to be sure but I think I want it outside of the else statement because I want the requested manure nutrients to be reported both in the event that manure is applied (i.e. when the if statement on L 588 is triggered) and wwhen there was no manure that can be applied (i.e. `manure_supplied = 0` and the else statement on L644 is triggered.

One thing I thought about when putting this in is that although the `record_manure_application` method works perfectly for reporting the manure requests, the name of the function is a little misleading.

ew3361zh: There is a call to `self._record_manure_application()` on line 631 that's in that if-statement from line 588 (`if manure_supplied is not None`) so if we leave the second call on line 648 outside the else statement, that means it will trigger in addition to that line 588 call when manure_supplied is not None (so there would be 2 calls to self._record_manure_application() and the second one would record dry_matter_mass of 0, etc even if manure_supplied is not None).

KFosterReed: Correct! So in the first call to `self._record_manure_application()` it is recording the `manure_supplied` and in the second call, it is recording the `manure_requested`. In the case where manure is applied and `manure_supplied` is not `None`, there should be 2 calls to the `self._record_manure_application()`. Does that make sense?

ew3361zh: Aah, yes, I get it now. It is a bit confusing with the function name and what we're recording but the `output_name` field can help with that distinction for now. I will move it back outside the else statement!

krb78 (approved):
Works great and looks great! :) Thanks for all the hard work on this!

RUFAS/routines/EEE/emissions.py
```diff
@@ -52,18 +52,20 @@ def __init__(self) -> None:
...
-        self._calculate_homegrown_feed_emissions(homegrown_feeds, fertilizer_applications, manure_applications)
+        self._calculate_homegrown_feed_emissions(
+            homegrown_feeds, fertilizer_applications, manure_applications, manure_requests
+        )
 
     def _calculate_purchased_feed_emissions(self, homegrown_feeds: list[dict[str, Any]]) -> None:
         info_map = {"class": self.__class__.__name__, "function": self._calculate_purchased_feed_emissions.__name__}
         purchased_feeds = self._gather_ration_feed_totals()
-        actual_purchased_feed_totals = self._calculate_actual_purchased_feeds(homegrown_feeds, purchased_feeds)
+        (actual_purchased_feed_totals, homegrown_feed_totals) = self._calculate_actual_purchased_feeds(
```
Isgotkowitz: `homegrown_feed_totals` is never used - should it be, or does it need to get removed?

RUFAS/routines/EEE/emissions.py
```diff
@@ -209,7 +223,7 @@ def _calculate_actual_purchased_feeds(
 
             actual_purchased_feeds[feed_id] = amount
 
-        return actual_purchased_feeds
+        return actual_purchased_feeds, homegrown_totals
```
Isgotkowitz: The `homegrown_totals` variable at the end of this method does not contain the totals of homegrown feeds, it contains the totals of homegrown feeds that were left over after feeding the animals. Just want to confirm that this is the correct number to be returning here.

RUFAS/routines/EEE/emissions.py
```diff
@@ -219,9 +233,23 @@ def _calculate_total_homegrown_feed_amounts_by_crop_type(
...
+        om.add_variable(
+            "homegrown_feed_totals",
+            homegrown_totals,
+            dict(
+                {
+                    "class": self.__class__.__name__,
+                    "function": self._calculate_total_homegrown_feed_amounts_by_crop_type.__name__,
+                },
+                **{"units": MeasurementUnits.KILOGRAMS},
+            ),
```
```suggestion
            {
                "class": self.__class__.__name__,
                "function": self._calculate_total_homegrown_feed_amounts_by_crop_type.__name__,
                "units": MeasurementUnits.KILOGRAMS,
            },
```

RUFAS/routines/EEE/emissions.py
```diff
@@ -309,6 +338,12 @@ def _calculate_homegrown_feed_emissions(
             aggregated_manure_apps[field_name]["nitrogen"] += app["nitrogen"]
             aggregated_manure_apps[field_name]["phosphorus"] += app["phosphorus"]
 
+        aggregated_manure_requests = {key: {"nitrogen": 0.0, "phosphorus": 0.0} for key in all_fields}
+        for app in manure_requests:
+            field_name = app["field_name"]
```
```suggestion
        for request in manure_requests:
            field_name = request["field_name"]
```

RUFAS/routines/EEE/emissions.py
```diff
@@ -309,6 +338,12 @@ def _calculate_homegrown_feed_emissions(
             aggregated_manure_apps[field_name]["nitrogen"] += app["nitrogen"]
             aggregated_manure_apps[field_name]["phosphorus"] += app["phosphorus"]
```
Isgotkowitz, line 336:
```suggestion
fields_with_manure_requests = {request["field_name"] for request in manure_requests}
all_fields = list(fields_with_manure_requests | fields_with_crops)
```
This handles the edge case where there was a field that grew crops but did not request any manure.

RUFAS/routines/field/field/field.py
```diff
@@ -639,11 +639,26 @@ def _execute_manure_application(
                 surface_remainder_fraction=surface_remainder_fraction,
                 year=year,
                 day=day,
+                output_name="manure_application",
             )
         else:
             supplied_nitrogen = 0.0
             supplied_phosphorus = 0.0
 
+        self._record_manure_application(
```
Isgotkowitz, line 648: I read through the comment on this conversation - I think using the `_record_manure_application` is not great because, as mentioned in that thread, the naming gets confusing. What about instead calling `add_variable` with the manure request data right here?

input/data/tillage_schedule/no_till.json
Isgotkowitz, line 1: Does these changes need to be reverted?

input/data/manure_schedule/no_manure.json
Isgotkowitz, line 1: Does this change need to be reverted?

input/data/fertilizer_schedule/no_fertilizer.json
Isgotkowitz, line 1: Does this change need to be reverted?

RUFAS/routines/EEE/emissions.py
```diff
@@ -336,6 +372,7 @@ def _calculate_homegrown_feed_emissions(
                 "potassium_fertilizer_used": MeasurementUnits.KILOGRAMS,
```
Isgotkowitz, line 368: The units of `carbon_stock_change` need to be changed to kilograms

Isgotkowitz (approved):
Found a couple of small things that I think need to be tweaked or confirmed before this gets merged, but overall looks great. Nice job @KFosterReed and @ew3361zh.

One thing I am curious about in the filters is the nitrate percolation - the filters are hardcoded to only look at nitrate percolation from layer 3. Is there a significance to layer 3 (which is really the fourth layer) or is this intended to capture whatever nitrates are leaching out of the bottom of the deepest soil layer?

RUFAS/routines/EEE/emissions.py
KFosterReed: Good call! My first attempt to report the homegrown feed amounts was via this method and I forgot to remove this. I will remove it here and in the return of  `calculate_actual_purchased_feed`

RUFAS/routines/field/field/field.py
KFosterReed: I think the reason I didn't use `o.add_variable` is part laziness and part code succinctness? Let me try and see if I can get it to work now...

PR #1835, opened 2024-07-17
NRC-fix

JoeWaddell:
### Context

- [ ] Yes
- [x] No

### What

This updates the metadata properties so that the NRC feed library can be used.

This PR also updates the activity requirement, to correct a unit conversion problem that causes the NRC calculations specifically to be drastically higher than expected.

### Test plan

Tested using both NASEM and NRC. Updated and rearranged unit tests to follow logic in the activity method.

Should be checked against pseudocode to ensure the conversion correction is now correct.


RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -1890,14 +1891,15 @@ def energy_activity_rqmts(self, body_weight: float, housing: str, distance: floa
...
+            net_energy_activity: float = distance_km * 0.00045 * body_weight + net_energy_activity1
             return net_energy_activity
         elif nutrient_standard == "NASEM":
             if housing == "Barn":
-                net_energy_activity = distance * 0.00035 * body_weight
+                net_energy_activity = distance_km * 0.00035 * body_weight
             elif housing == "Grazing":
                 nonpasturekgDMI: float = 1.0
-                net_energy_activity = distance * body_weight * 0.75 * (600 - 12 * nonpasturekgDMI) / 600
+                net_energy_activity = distance_km * body_weight * 0.75 * ((600 - 12 * nonpasturekgDMI)) / 600
```
allisterakun:
```suggestion
                net_energy_activity = distance_km * body_weight * 0.75 * (600 - 12 * nonpasturekgDMI) / 600
```

input/metadata/properties/default.json
```diff
@@ -3353,6 +3353,14 @@
         "pattern": "^(Aminoacids|Forage|Conc|Milk|Mineral|Vitamins|Starter|No)$"
       }
     },
+    "Fd_Category": {
```
allisterakun, line 3356: Where in the code are we using the newly added "Fd_Category"?

allisterakun (approved):
Looks good, Joe! Just some comments!

input/metadata/properties/default.json
JoeWaddell: That's used in `report_daily_ration` in `ration_reporter.py`, to determine if the feed item is categorized as a By-Product or not.

RUFAS/routines/animal/ration/animal_requirements.py
```diff
@@ -1878,6 +1878,7 @@ def energy_activity_rqmts(self, body_weight: float, housing: str, distance: floa
             National Academic Press, Chapter 3 "Energy", pp. 30-31, 2021.
 
         """
+        distance_km = distance * GeneralConstants.M_TO_KM
```
elle-andreen, line 1883: We should update the docstring above (line 1860) to match here (distance -> distance_km, and update the units, just to prevent the same confusion we already went through!)

input/metadata/properties/default.json
elle-andreen: Just to double check - this just adding properties for a variable that already existed but was missing properties, right? It's an 'input' but is already present for each feed in the default NASEM & NRC feed libraries

elle-andreen (approved):
All looks as it should here to me, thanks Joe! I am seeing requirements much more in line with what we're seeing using the NASEM standard. When I ran a quick simulation, I was seeing very similar numbers using the NRC standards on this branch to those using NASEM on dev (both around 41 Mcal/d).

input/metadata/properties/default.json
JoeWaddell: Correct! It was always there, just not in the metadata so it didn't survive the validation step.

RUFAS/routines/animal/ration/animal_requirements.py
JoeWaddell: Updated the docstring - seems like it might be easier to input the distances in meters, as Input Manager is already tuned to do, so I made a note that the conversion happens in the code to align with the cited equations.

PR #1836, opened 2024-07-17
suggested feed prices for default feed file

### LINKED ISSUE #1834: Suggested Feed prices for default ration
As mentioned in #1783, automated ration formulation is giving unrealistic rations. The primary culprit is pricing: with flat pricing for feed items, a strange ration is formulated. A simple PR could fix this by adding "default" prices to the `default_feed.json` in `input/data/feed/default_feed.json`.

For SMEs: please leave any suggestions for prices below, otherwise I'll be reaching out to a couple folks when opening an associated PR.

JoeWaddell:
### Context

Issue(s) closed by this pull request: closes #1834, #2087

- [ ] Yes
- [x] No

### How

Referenced older feed files and used input from other SMEs, e.g. prior to #1112


KFosterReed:
[user_feeds_w_feed_prices.xlsx](https://github.com/user-attachments/files/16339773/user_feeds_w_feed_prices.xlsx)

I copied the user_feeds csv and added some suggested realistic prices for most of the feeds in the default feed list along with notes of how I got them... paying particular attention to units in this one!

elle-andreen:
Using the values Joe entered in the PR's input file, vs. the prices Kristan gave, I am seeing
image

This is at the same DMI, productivity, etc. I would say these rations are "okay" but they are still very simplistic. They're also low in starch, especially under Kristan's pricing scheme (sorry Kristan 😅), but not totally unrealistically low. Fiddling with the price of corn grain got me a scenario where the formulation just flipped back and forth between including either corn silage vs. corn grain, so a lot of fluctuations in starch. So I think the downfalls of the current ration using these prices are at least partly attributable to the fact that we're still operating under relatively limited constraints on the ration/nutrient composition. I think the sources Kristan pulled prices from/the other logic are reasonable, and the prices and their proportions seem more or less appropriate based on my limited knowledge of feed pricing. Pending others' thoughts, I would be fine instituting either pricing scheme as a fix to get the default ration functioning again!

One other thing: I am seeing values above 1000 for `forage_NDF (percent)` in the ration supply reports, it may not be in the scope of this PR, but I wonder if there is a units issue there. The `forage_NDF_percent (percent of DM)` values look correct (40-45% of DM)

Isgotkowitz (approved):
LGTM

tomhuhh (commented):
<p>Great work, Joe!</p>
<p>I checked the feed prices in the <code inline="">default_feed.json</code> file against the Excel sheet and found no errors. I tested it with and without the user-defined ration feature enabled (using all default settings) and focused on diets for lactating cows.</p>
<p>When comparing diets created using non-linear programming (NLP) and the user-defined ration (UDR), I averaged the last 13 rows of the results and noticed some clear differences. The feed prices seem correct, assuming Kristan has verified them, but the NLP method produces a different diet compared to the UDR. This issue needs further investigation to understand why these differences happen, though it doesn't seem to prevent us from using this set of prices, assuming they come from a validated source.</p>
<p>I also checked the <code inline="">default_feed_2.json</code> file and saw that prices are missing for ingredients with these IDs: 15, 31, 42, 66, 197, and 210. Adding these prices and testing that diet as well could give more complete results.</p>
<p>Here’s a summary of the comparison between the two diet formulation methods:</p>

Ingredient ID | User-Defined Ration (%) | Non-Linear Programming(%)
-- | -- | --
2 | 11.7 | 0.0
44 | 11.2 | 0.0
51 | 39.3 | 24.0
100 | 7.5 | 0.8
110 | 13.7 | 70.3
167 | 13.1 | 3.8
231 | 1.8 | 1.1
234 | 1.7 | 0.0
Total Dry Matter Intake (kg) | 27.8 | 29.5


<p>Let me know how you’d like to move forward!</p></body></html>

JoeWaddell:
> I also checked the `default_feed_2.json` file and saw that prices are missing for ingredients with these IDs: 15, 31, 42, 66, 197, and 210. Adding these prices and testing that diet as well could give more complete results.

Good catch! If we plan to keep that file as a good default for a slightly different set of UDR/composition, we should update the prices following the approach @KFosterReed so nicely laid out: I've opened an issue to that effect (since this set of IDs/prices has already been evaluated, best to keep these PRs smaller). https://github.com/RuminantFarmSystems/MASM/issues/2117

Otherwise, the differences in the NLP and UDR approach are just the nature of the two: since the NLP will generate the simplest ration possible within the bounds of the ration formulation constraints, average animal requirements of the pen in question, and these prices. We could further "diversify" the NLP rations by adding minimums and maximums, but that requires some more careful thought: https://github.com/RuminantFarmSystems/MASM/issues/1888

Likewise, if the composition of the automated ration still needs serious work that can't be solved by #1888, then we should still consider additional constraints!

Let me know if you (@tomhuhh) need anything else prior to approval!

tomhuhh (approved):
Thank you for creating the new issue #2117, @JoeWaddell! 

I am giving this PR update an approval. Thanks!

PR #1840, opened 2024-07-18
Utilize new functions in utility

### LINKED ISSUE #1669: [Input Manager] Utilize new utility functions
Two new utility functions `convert_flat_dict_to_nested_dict` and `deep_merge` are introduced to the code base. These can probably be used in Input Manager to both improve existing functionalities and code cleanliness and to introduce new features. Explore this possibility. 
Hint: take a look at the unit tests to see what these functions can and cannot do.

matthew7838:
Implements the new function `flatten_keys_to_nested_structure_nested_dict` in the IM.
### Context

Issue(s) closed by this pull request: closes #1669 

- [ ] Yes
- [x] No

### What

- The method` _prepare_data` will now use `flatten_keys_to_nested_structure_nested_dict` to return the data with the nested dictionary structure.
-  The tests have been updated for _prepare_data and deleted for `_set_nested_value`.
- Deprecated `_set_nested_value` in Input Manager.

### Why

Make use of the new methods and make the IM a lot cleaner.

### How

Cleans up the code by removing helper methods in IM and making use of more universally used Utility methods.

### Test plan

Ran a complete test suits on Input Manager.


Isgotkowitz (commented):
This is a good simplification of the code and from what I can tell it has no effect on the overall behavior. Unfortunately, it seems that adding a variable that is nested within a dictionary was not working to begin with.

Adding the following code to `SimulationEngine`'s init method will run the added lines in InputManager:
```python
    def __init__(self) -> None:
        """
        Initializes the simulation engine.
        """
        self.im = InputManager()
        self.time = Time()
        self._initialize_simulation()
        self.im.add_dict_variable_to_pool(
            "config.FIPS_county_code",
            {"FIPS_county_code": 55025},
            "config_properties",
            False
        )
```
On both the feature branch and `dev`, this code will crash because `prepare_data` does not return metadata properties in a way that works with `_validate_data` when it is called by `_add_variable_to_pool`. I am pretty sure that this should be a valid way to add data to the pool at runtime but I am 100% about that.

I'm not sure what the best path forward here is. I can approve if it's ok to leave the behavior as-is, just let me know

matthew7838:
> This is a good simplification of the code and from what I can tell it has no effect on the overall behavior. Unfortunately, it seems that adding a variable that is nested within a dictionary was not working to begin with.
> 
> Adding the following code to `SimulationEngine`'s init method will run the added lines in InputManager:
> 
> ```python
>     def __init__(self) -> None:
>         """
>         Initializes the simulation engine.
>         """
>         self.im = InputManager()
>         self.time = Time()
>         self._initialize_simulation()
>         self.im.add_dict_variable_to_pool(
>             "config.FIPS_county_code",
>             {"FIPS_county_code": 55025},
>             "config_properties",
>             False
>         )
> ```
> 
> _in progress_ On both the feature branch and `dev`, this code will crash because `prepare_data` does not return metadata properties in a way that works with `_validate_data` when it is called by `_add_variable_to_pool`. I am pretty sure that this should be a valid way to add data to the pool at runtime but I am 100% about that.
> 
> I'm not sure what the best path forward here is. I can approve if it's ok to leave the behavior as-is, just let me know

Let's talk about this during WT

Isgotkowitz (approved):
As far as I can tell, this is not changing any overall behavior.

However, I cannot get small pieces of data added to pool, even if they are dictionaries. This should be revisited, because it is currently not clear what runtime data additions should work and which should not. For example, the following does not work when added to the `__init__` routine of `SimulationEngine`:
```python
        self.im.add_dict_variable_to_pool(
            "animal.herd_initialization",
            {
                "initial_animal_num": 10000,
                "simulation_days": 5000
            },
            "animal_properties",
            False
        )
```

tests/test_input_manager.py
```diff
@@ -4814,7 +4790,7 @@ def mock_metadata_prepare_data() -> dict[Any, Any]:
 
 
 @pytest.mark.parametrize(
-    "variable_name,input_data,properties_blob_key," "expected_data,expected_metadata_properties",
+    "variable_name,input_data,properties_blob_key," "expected_data,expected_metadata_properties,nested",
```
ew3361zh:
```suggestion
    "variable_name,input_data,properties_blob_key," "expected_data,expected_metadata_properties,is_nested",
```

ew3361zh (approved):
Thanks for the test cases @Isgotkowitz. Matthew we're trying to add more specific test cases for how you ran the code and how we can duplicate your results (or try variations on the test cases you provided to ensure everything works) in the `Test Plan` section of the PR description.

I agree it seems like the issue is not specifically with the code added here which seems fine but tough to tell that it's working entirely as intended since one of the key uses isn't functioning as designed (we do have the unit tests which at least show the isolated behavior is working as designed).

Could you add an issue to address that `add_dict_variable_to_pool()` isn't currently functional, Matthew?

Overall nice work!

Issue #1803, opened 2024-07-05
[Manure] Remove manure treatment removal efficiencies


elle-andreen:
In the current Manure inputs (e.g. in `manure_management.json`), inputs exist to configure "removal efficiencies" `(..._removal_efficiency_for_treatment`) for specific nutrients in manure treatments. Per the discussion [in this issue](https://github.com/RuminantFarmSystems/MASM/issues/1031) and other comments from Kristan, these removal efficiencies and their functionality within the manure module should be deprecated. We have elected to move toward more process-based tracking of nutrient removal/loss from manure, e.g., updating nutrients according to estimated gas emissions.

elle-andreen:
Closed by the manure refresh, completed 6/20/2025!

Issue #1540, opened 2024-04-11
Changing VS removal efficiency of AD not resulting in difference in VS after AD  


tomhuhh:
Changing VS removal efficiency of AD did not result in difference in VS after AD.   

## Expected Behavior

Difference in VS removal efficiency results in variation in outflow of VS from AD.  

## Current Behavior

By changing the variable `volatile_solids_removal_efficiency_for_treatment` in the manure input file from 0 to 0.45 to 0.9, the VS outflow from AD does not change. 

Efficiency = 0 

![default_ad-degradable-solids-11-Apr-2024_Thu_15-27-27](https://github.com/RuminantFarmSystems/MASM/assets/52800053/00f5ca7e-2bf9-42de-8072-45120ef26118)

Efficiency = 0.45

![default_ad-degradable-solids-11-Apr-2024_Thu_15-24-21](https://github.com/RuminantFarmSystems/MASM/assets/52800053/3bd2c9a7-9534-4d38-976c-791a415691d5)

Efficiency = 0.9 

![default_ad-degradable-solids-11-Apr-2024_Thu_15-40-06](https://github.com/RuminantFarmSystems/MASM/assets/52800053/48d924c4-e93b-44a5-b8a3-d8713d5828ac)

## Steps to Reproduce

1. Go to the main branch, set up the manure scenario to be `5` in the animal input file, and adjust the variable `volatile_solids_removal_efficiency_for_treatment` in the manure input file from 0 to 0.45 to 0.9,
2. Use the graphic filter 

```
{
  "multiple": [
    {
      "type": "plot",
      "filters": [
        "AnaerobicDigestion_ManureTreatmentDailyOutput_Pen_0_CALF\\.liquid_manure_total_degradable_volatile_solids",
        "AnaerobicDigestion_ManureTreatmentDailyOutput_Pen_1_GROWING\\.liquid_manure_total_degradable_volatile_solids",
        "AnaerobicDigestion_ManureTreatmentDailyOutput_Pen_2_CLOSE_UP\\.liquid_manure_total_degradable_volatile_solids",
        "AnaerobicDigestion_ManureTreatmentDailyOutput_Pen_3_LAC_COW\\.liquid_manure_total_degradable_volatile_solids"
      ],
      "title": "AD Degradable Solids",
      "legend": [
        "Pen0",
        "Pen1",
        "Pen2",
        "Pen3"
      ]
    },
    {
      "type": "plot",
      "filters": [
[...]

Issue #1619, opened 2024-05-08
[Manure] Separate Manure Treatments from Storages


ew3361zh:
**Is your feature request related to a problem? Please describe.**
Recommend renaming the treatment classes that are storages to include the word “storage” or have some other way to easily distinguish them as treatments. Ensure current treatments that are actually storages can accept manure from handlers, separators, treatments, storages, or beddings.

**Describe the solution you'd like**
Separate storages and treatments into separate parent classes from which each of the child classes inherits. It would make more sense and help distinguish them from one another in ways that storages and treatments differ.

Current treatments that are actually storages:
Slurry Storage Underfloor
Slurry Storage Outdoor
Open Lots
Compost Bedded Pack Barn
Anaerobic Lagoon
Composting

Current treatments that are treatments:
Anaerobic Digestion

**Additional context**
When refactoring _pen_daily_update() and allowing for more flexibility among the components used and the order in which they are used, we will need to have enforceable guidelines laid out in each class as to which other types they can discharge to and accept manure from. Having storages and treatments distinguished would be helpful.

Isgotkowitz:
Switched anaerobic digestion and anaerobic lagoon around in the lists of actual treatments vs storages

allisterakun:
During grooming session:
@Isgotkowitz will come back and add more details about this issue.

Isgotkowitz:
### Current State
Manure management scenarios are configured like so:
```json
{
  "scenario_id": 0,
  "bedding_type": "sawdust",
  "manure_handler": "manual scraping",
  "manure_separator": "none",
  "manure_separator_after_digestion": "none",
  "manure_treatment": "anaerobic lagoon"
}
```

Manure goes from the separator (if there is one) to the treatment. Two of the treatments are compounds ("anaerobic digestion and lagoon" and "anaerobic digestion and lagoon with separator"), and under the hood these are really just two treatments. The manure goes from the digester to the lagoon, and optionally through a separator in between.

### Desired State
The treatments and storages should be specified separately in a manure management scenario, like so:
```json
{
  "scenario_id": 0,
  "bedding_type": "sawdust",
  "manure_handler": "manual scraping",
  "manure_separator": "none",
  "manure_treatment": "anaerobic digestion"
  "manure_separator_after_treatment": "none",
  "manure_storage": "anaerobic lagoon"
}
```
The directionality should be preserved - meaning that manure will still flow in the following order: separator -> treatment -> secondary separator -> storage.

[...]

PR #1843, opened 2024-07-18
Update Anaerobic digestion

### LINKED ISSUE #1370: Modify TAN of manure after anaerobic digestion
The N in manure that undergoes anaerobic digestion is known to be modified by microbes during the digestion process. In general, anaerobic digestion results in a greater proportion of manure N as ammoniacal nitrogen, due to mineralization of organic N. This can increase down-stream emissions of NH3 during storage of digestate (manure that has been anaerobically digested) as TAN can volatilize to ammonia.

The IFSM utilizes a blanket 15% increase in TAN in anaerobic digestate, compared to manure prior to digestion. [This study](https://www.sciencedirect.com/science/article/pii/S0048969719340367) also reports some real-world observations of N and other nutrient conversions during anaerobic digestion. 

As a starting point, we could consider implementing a basic multiplier to adjust digestate TAN, similar to IFSM.

**7/1/2024 update:** This change is being implemented as part of the general anaerobic digestion SME review/revisions. A formal PR does not yet exist - still waiting on SME review.

### LINKED ISSUE #1371: Add leakage emissions from anaerobic digesters
Though anaerobic digesters are specifically designed to generate and capture methane produced from anaerobic fermentation of manure and other substrates, like any closed system, leaks can and do occur. Methane leakage is generally considered the sole GHG emission from digesters, and is not currently quantified in RuFaS.

The 2023 USDA Quantifying Greenhouse Gas Fluxes in Agriculture and Forestry method utilizes methods leakage factors (fraction of total methane generated that leaks from the digester) based on the digester type and construction to predict digester leakage: 
image
These values are originally from the Clean Development Mechanism [Project and leakage emissions from anaerobic digesters](https://cdm.unfccc.int/methodologies/PAmethodologies/tools/am-tool-14-v2.pdf) document. 

We should implement basic MCF-type values in RuFaS to reflect GHG emissions in the form of methane from anaerobic digesters.

### LINKED ISSUE #1417: Update anaerobic digestion outputs
Currently the anaerobic digestion method does not output methane in units of kg methane, it instead reports methane volume, biogas, and biogas energy content. 

We would like to get more explicit definitions of the biogas output variables and add a storage methane output in units of kg CH4. 

P.S. The current definition of biogas is kg/d of CH4 generated from anaerobic digestion, converted from methane_generation_volume. 

## Expected Behavior

1. More explicit definitions of the biogas output variables reflected in the variable name as well as the doc strings. 
2. Add storage methane as an output of anaerobic digestion in the units of kg CH4. 

## Current Behavior

Currently the anaerobic digestion method does not output methane in units of kg methane, it instead reports methane volume, biogas, and biogas energy content. 

## Possible Solution

Update the outputs from anaerobic digestion as described above. 

## Steps to Reproduce

1.
3.
4.
5.

### LINKED ISSUE #1832: [Manure] Updates to anaerobic digestion treatment
The SME review of the anaerobic digestion submodule has recently been completed, and some suggested changes to anaerobic digestion are ready to be implemented. 

These changes apply to the "anaerobic digestion and lagoon" treatment but focus on updates to anaerobic digestion portion, rather than the anaerobic lagoon (the storage that holds manure [digestate] after it has passed through the digester).

Several issues exist already that are related to/covered in this enhancement:
Issue #1370 - related to TAN
Issue #1417  - related to output naming
Issue #1540 can also be closed when an eventual PR is merged.

For simplicity, this issue lists all changes to be made, including those noted above.

A branch has been created with the bulk of these changes, that could serve as a starting point for a PR ([test-AD-updates-v2])(https://github.com/RuminantFarmSystems/MASM/tree/test-AD-updates-v2)

1. The "Chen and Hashimoto" method in calculator.py should be deprecated and replaced with a new function to determine methane generation volume. Some inputs and configurations, as well as constants and other functions, that are associated specifically with this function should be deprecated. This also necessitates changes in the actual manure treatment files (anaerobic_digestion.py and anaerobic_digestion_and_lagoon.py).

2. A new numerical input for digester leakage rate should be created, as part of the anaerobic digestion and lagoon configuration. The calculation of digester leakage should be implemented in anaerobic_digestion.py and included as an output in AnaerobicDigestion_ManureTreatmentDailyOutput.

3. Modify the daily update function in anaerobic_digestion.py to utilize a new method to update the daily output value of liquid manure total and volatile solids. (Issue #1540)

4. Implement a uniform percentage increase in liquid manure TAN in anaerobic digestion daily output. 

5. Modify naming of some variables for improved clarity and ease of interpretation (#1417)

**Additional context**
Additional context and specifics will be provided on new functions, constants, descriptions, etc. The design doc for anaerobic digestion is also in the process of being updated. 

**Target Deadline**
Ideally these changes could be implemented within 2-4 weeks.

**Task List**
- [x] #1370 
- [x] #1417 
- [x] #1540 
- [x] Deprecate Chen-Hashimoto method, replace with new function.
- [x] Create numerical input for digester leakage rate.

Isgotkowitz:
### Context

Issue(s) closed by this pull request: closes #1832, closes #1370

- [ ] Yes
- [x] No

### What

This PR makes several updates to the Anaerobic Digestion and Anaerobic Lagoon modules, and several related spots in the code.

The changes include:
- Applying an automatic increase of Total Ammoniacal Nitrogen during the anaerobic digestion process (#1370)
- Swapping the term "biogas" for "methane" when referring to what is captured by the digester (#1417) edit: this change is not happening in this PR any longer, as it is not backwards compatible.
- Deprecating the Chen-Hashimoto method of methane generation estimation in the digester.
- Adding an input that allows users to control how much methane leakage there is in the digestion process.

### Why

These are just some improvements that were suggested/asked for to improve the Anaerobic digestion process.

### How

Constants have been added and removed in two of the manure constants files as equations are replaced. Methane generation, leakage, and energy content are now calculated with `calculate_CSTR_methane_volume`, `calculate_digester_methane_leakage`, and `calculate_methane_energy_content`, respectively, in `GasEmissionsCalculator`.

The process of calculating methane generation in `AnaerobicDigestion` has been modified so that it now accounts for methane leakage while calculating the final amounts of methane and methane-energy.

The metadata and manure module input have been updated so that the leakage fraction of methane from digesters can be controlled by the user, and the `ManureTreatmentConfig` class has been extended to handle this new input.

`ManureTreatmentDailyOutput` has been updated to include leaked methane and methane energy content (instead of biogas energy content).

### Test plan

Unit tests have been added and modified as appropriate to test all code changes. The outputs for anaerobic digestion have been updated to ensure that all methane generation and volatile solids loss is happening as expected.

Note that the new input files will be removed before merging, they are currently just here for testing purposes.

### Other
#1540/#1803 are not addressed in this PR. I think (but still need to confirm) that #1540 will be closed as unplanned and the issue will be addressed by #1803 instead, which I think will be large enough to merit its own PR.

I love that volatile solids loss is called volatile solids "destruction" in the code, it is very cathartic.

Edit: in order to maintain backwards-compatibility in this PR, the changes to variable names/outputs have been removed. So biogas is still biogas, instead of being changed to methane:/ i.e. this PR _does not_ fix #1417


RUFAS/routines/manure/manure_treatments/anaerobic_digestion.py
```diff
@@ -74,37 +77,51 @@ def _calc_anaerobic_digestion_daily_output(
...
         # MS.3.B.2
         minimum_digester_volume = daily_final_manure_volume * self.config.hydraulic_retention_time
         # MS.3.B.3
         top_cover_volume = minimum_digester_volume * self.config.top_cover_volume_fraction
 
-        new_daily_output.biogas = methane_generation_volume * GasEmissionConstants.AD_METHANE_DENSITY
+        new_daily_output.liquid_manure_total_ammoniacal_nitrogen = min(
+            self._current_manure_treatment_daily_input.liquid_manure_total_ammoniacal_nitrogen * ManureConstants.AD_TAN_INCREASE_FACTOR, 
+            self._current_manure_treatment_daily_input.liquid_manure_nitrogen
+        )
```
Isgotkowitz: @elle-andreen is this the blanket increase in TAN that is called for by #1370?

elle-andreen: Yes! This is the blanket increase - the `AD_TAN_INCREASE_FACTOR` constant should be equal to 1.60 to result in a blanket 60% increase in TAN content.

Isgotkowitz: Awesome, thank you

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
```diff
@@ -156,18 +156,23 @@ class ManureTreatmentDailyOutput(LiquidManurePortionProtocol):
...
-    biogas_energy_content: float
-        Energy content of biogas, MJ/m^3.
-    biogas_energy_content_unit: MeasurementUnits
-        Unit for biogas_energy_content.
+    methane_generation_mass: float
+        Mass of methane produced from anaerobic digestion, kg.
+    methane_generation_mass_unit: MeasurementUnits
+        Unit for methane_generation_mass.
+    methane_energy_content: float
+        Energy content of methane, MJ/m^3.
```
Isgotkowitz: @elle-andreen these units should be MJ not MJ/cubic meter right?

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
```diff
@@ -163,7 +163,7 @@ def daily_update(
                     pen.manure.manure_total_ammoniacal_nitrogen - housing_ammonia_emission,
                 )
             ),
-            liquid_manure_nitrogen=pen.manure.nitrogen,
+            liquid_manure_nitrogen=max(0.0, pen.manure.nitrogen - housing_ammonia_emission),
```
elle-andreen: This is my doing rather than Ed's, but this change is definitely outside the scope of this PR. It _is_ something that should be changed, it just doesn't belong here 😊 I can pull it out and either pin it to the issue for doing the manure management chain mass balance, or create a mini PR for it. It's only downstream effect on emissions would be a slight drop in N2O, because N2O is estimate from manure N, rather than TAN.

RUFAS/routines/manure/manure_treatments/anaerobic_digestion_and_lagoon.py
```diff
@@ -84,13 +84,27 @@ def _daily_update_helper(self) -> ManureTreatmentDailyOutput:
...
+        )
         self._accumulated_output.liquid_manure_total_degradable_volatile_solids = (
             self._anaerobic_lagoon._accumulated_output.liquid_manure_total_degradable_volatile_solids
         )
         self._accumulated_output.liquid_manure_total_non_degradable_volatile_solids = (
             self._anaerobic_lagoon._accumulated_output.liquid_manure_total_non_degradable_volatile_solids
         )
 
-        anaerobic_lagoon_daily_output.biogas = self.anaerobic_digestion_daily_output.biogas
+        anaerobic_lagoon_daily_output.methane_generation_mass = (
```
elle-andreen: I believe this variable should not be passed to the anaerobic lagoon treatment file (i.e., it should not appear in the ManureTreatmentDailyOutput for the anaerobic lagoon). This methane is strictly from the digester and should only be reported in the anaerobic digestion output, it is confusing to have it in the lagoon output because you end up with both `methane_generation_mass` and `storage_methane`.

RUFAS/routines/manure/manure_treatments/manure_treatment_configs.py
```diff
@@ -40,6 +40,9 @@ class ManureTreatmentConfig:
         Temperature set point for the anaerobic digestion.
     anaerobic_digestion_temperature_celsius:
         Temperature of the anaerobic digestion.
+    digester_methane_leakage_fraction : float, default 0.0
```
elle-andreen, line 43: Other variables in this list (dictionary?) don't have 'float' or 'default' included - are the other variables incomplete, or is the new addition incorrect? And, does `default` here need to be the actual default value or is `0.0` just the value format?

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -779,43 +779,57 @@ def _convert_temperature_celsius_to_kelvin(cls, temperature_celsius: float) -> f
...
+        methane_volume : float
+            Methane generation volume, m^3.
+
+        Returns
+        -------
+        float
+            Methane energy content, MJ.
 
         """
         return methane_volume * GasEmissionConstants.AD_METHANE_DENSITY * GasEmissionConstants.METHANE_ENERGY_DENSITY
```
elle-andreen: I did not realize this originally but I think we can just pass `captured_methane_generation_mass` here instead of methane volume, since we are just multiplying by the same density value (e.g. `methane_volume * GasEmissionConstants.AD_METHANE_DENSITY` here is equal to `captured_methane_generation_mass`).

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
elle-andreen: Correct, it's MJ, good question. We are multiplying captured methane mass (kg) by the energy density constant, which is actually in MJ/kg (rather than m3 as it read originally). So we end up with a MJ value.

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
```diff
@@ -296,16 +301,18 @@ class ManureTreatmentDailyOutput(LiquidManurePortionProtocol):
...
-    # different unit from docMeasurementUnitsing m^3 vs m^3/day?
-    biogas: float = 0.0  # biogas production per day (m3/day)
-    biogas_unit: MeasurementUnits = MeasurementUnits.CUBIC_METERS_PER_DAY
+    methane_generation_mass: float = 0.0
+    methane_generation_mass_unit: MeasurementUnits = MeasurementUnits.KILOGRAMS_PER_DAY
 
-    biogas_energy_content: float = 0.0  # biogas energy content (MJ/m3)
-    biogas_energy_content_unit: MeasurementUnits = MeasurementUnits.MEGAJOULES_PER_CUBIC_METER
+    methane_energy_content: float = 0.0
+    methane_energy_content_unit: MeasurementUnits = MeasurementUnits.MEGAJOULES_PER_CUBIC_METER
```
elle-andreen: Related to your Q above - this should be MJ, not MJ/kg

input/data/manure/manure_management.json
```diff
@@ -297,20 +297,21 @@
...
       "sludge_accumulation_period": 1.0,
-      "sludge_accumulation_volume_fraction": 0.03,
+      "sludge_accumulation_volume_fraction": 0.0,
       "top_cover_volume_fraction": 0.2,
       "evaporation_fraction": 0.02,
       "anaerobic_digestion_temperature_set_point": 37.5,
       "anaerobic_digestion_temperature_celsius": 37.5,
-      "manure_cover": "N/A"
+      "manure_cover": "N/A",
+      "digester_methane_leakage_fraction": 0.005
```
elle-andreen: The leakage fraction default here is different from the default given in the metadata properties (0.01) - is that intentional? I also want to check with Kristan and others on this value before we merge. It's one little value but leakage is pretty controversial in the real world!

elle-andreen:
This is look great so far Ed. I left some comments in the code, and the comments below are moreso for other folks and for future reference:

- I like the naming scheme you went with - `captured_methane_generation_volume` for total methane minus leakage, and `total_methane_generation_volume` for the actual total amount. If it's not problematic, I would actually propose using these names in the output - I think they make the interpretation clearer! This PR intrinsically changes the inputs, so we can also confirm that output renaming is acceptable. 
- Heating input energy: I debated taking this out but ultimately I think we should leave it for now. I expect it will be an important part of EEE for anaerobic digestion and I don't see any big problems with the function, it does include manure moisture content as a variable but I think moisture vs solids content at the point of an anaerobic digester (e.g. before a lot of evaporation occurs in storage, that we are not accounting for) is reasonably correct for now. Addressing moisture content at different points in the manure module is on the improvements list anyway. 
- Evaporated water: when I send the design doc for SMEs to review, I will ask if this is something we should have. I don't know where this equation came from, but also it seems that the volume from evaporated water is not being subtracted from total manure volume anyway, so it's not having any effect. 
- (Just a reminder) Remove all changes from `manure_management.json` before marking as ready for review, with the exception of the addition of the leakage fraction input of course

elle-andreen:
In the [google doc with calculation details](https://docs.google.com/document/d/1om-czbamAUBlSAGegGuNhlSUny6EWygnwFSLF6wWkx0/edit?usp=sharing), I have added a sample output to help people reviewing this follow how calculations work. Hopefully it's helpful!

That being said, outputs are being calculated as expected! We still have a couple accessory things to change or confirm should be changed, but the core function is there. 

Also just saw:
>I love that volatile solids loss is called volatile solids "destruction" in the code, it is very cathartic.

It is a good term, the usual term in the literature is reduction or destruction, and the VS are literally being destroyed (degraded into a new chemical form) so 😄

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
Isgotkowitz: I recommend creating a small PR to address exactly this - in fact, it is a really good practice to make very small PRs. I will undo the changes here

input/data/manure/manure_management.json
Isgotkowitz: No, I just wanted to try a value that I hoped was realistic - I will change it to 0.01

RUFAS/routines/manure/gas_emissions/calculator.py
Isgotkowitz: Implemented

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
Isgotkowitz: Updated

RUFAS/routines/manure/manure_treatments/manure_treatment_configs.py
Isgotkowitz: The other variables are incomplete. The things I added are in accordance with Numpy's [class docstring guideline](https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring).

0.0 is the default - in the `ManureTreatmentConfig` class. This default is not necessarily the same as the default defined in the metadata properties. A better solution would probably be to have no default for `digester_methane_leakage_fraction` in `ManureTreatmentConfig`.

input/data/manure/manure_management.json
Isgotkowitz: Also, this is an important distinction, that value is an input not a default. If a user were to not include the key `"digester_methane_leakage_fraction"`, then the Input Manager would pull a default value from the metadata properties, if provided.

Isgotkowitz:
> This is look great so far Ed. I left some comments in the code, and the comments below are moreso for other folks and for future reference:
> 
> * I like the naming scheme you went with - `captured_methane_generation_volume` for total methane minus leakage, and `total_methane_generation_volume` for the actual total amount. If it's not problematic, I would actually propose using these names in the output - I think they make the interpretation clearer! This PR intrinsically changes the inputs, so we can also confirm that output renaming is acceptable.

Thanks:) and I agree, I think it would make the outputs easier to understand. However the changes to outputs would be different in nature to the changes being made to the inputs here. The input changes are backwards compatible, i.e. someone with an existing manure management input file would not have to change it all for it to work with this new code. But a filter for getting digester methane outputs would likely have to be rewritten to work with the new naming scheme. We should talk to Pooya and/or Kristan about this. 

> * Heating input energy: I debated taking this out but ultimately I think we should leave it for now. I expect it will be an important part of EEE for anaerobic digestion and I don't see any big problems with the function, it does include manure moisture content as a variable but I think moisture vs solids content at the point of an anaerobic digester (e.g. before a lot of evaporation occurs in storage, that we are not accounting for) is reasonably correct for now. Addressing moisture content at different points in the manure module is on the improvements list anyway.

I am not sure what the heating input energy is, which spot in the code is this located?

> * Evaporated water: when I send the design doc for SMEs to review, I will ask if this is something we should have. I don't know where this equation came from, but also it seems that the volume from evaporated water is not being subtracted from total manure volume anyway, so it's not having any effect.

Is this equation changed at all in the code?

> * (Just a reminder) Remove all changes from `manure_management.json` before marking as ready for review, with the exception of the addition of the leakage fraction input of course

Will do.

RUFAS/routines/manure/manure_treatments/anaerobic_digestion.py
```diff
@@ -60,6 +61,8 @@ def _calc_anaerobic_digestion_daily_output(
             The daily output from anaerobic digestion.
 
         """
+        assert self._current_manure_treatment_daily_input is not None
```
Isgotkowitz, line 64: Added this just to make the number of Mypy errors go down.

input/data/manure/manure_management.json
elle-andreen: That's a good point! I always forget that the defaults come from the metadata properties, not the default input files.

elle-andreen:
> The input changes are backwards compatible, i.e. someone with an existing manure management input file would not have to change it all for it to work with this new code. But a filter for getting digester methane outputs would likely have to be rewritten to work with the new naming scheme. We should talk to Pooya and/or Kristan about this.

This is true! It will definitely be on the list of things to confirm before this merges. It is certainly not an essential change, it could stay as-is without a problem.

> I am not sure what the heating input energy is, which spot in the code is this located?

It is calculated in the `anaerobic_digestion.py` file using the `self._calc_specific_input_energy` function. It was not modified in any way by this PR though, I am mostly noting it so that the SMEs remember to revisit it down the line.

> Is this equation (Evaporated water) changed at all in the code?

Nope, it should be unchanged by this PR

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
```diff
@@ -156,18 +156,23 @@ class ManureTreatmentDailyOutput(LiquidManurePortionProtocol):
...
-    biogas: float
-        Amount of biogas produced, m^3.
-    biogas_unit: MeasurementUnits
-        Unit for biogas.
-    biogas_energy_content: float
-        Energy content of biogas, MJ/m^3.
-    biogas_energy_content_unit: MeasurementUnits
-        Unit for biogas_energy_content.
+    methane_generation_mass : float, default 0.0
+        Mass of methane produced from anaerobic digestion, kg.
```
tomhuhh: Would it help improve the clarity of the variable names by modifying it to 'AD_methane_generation_mass'? Likewise for the other variables. Thanks!

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -779,46 +779,60 @@ def _convert_temperature_celsius_to_kelvin(cls, temperature_celsius: float) -> f
...
-                / (
-                    hydraulic_retention_time * GasEmissionConstants.SPECIFIC_GROWTH_RATE
-                    + GasEmissionConstants.CHEN_HASHIMOTO_KINETIC_CONSTANT_KCH
-                    - 1
-                )
-            )
-            * manure_total_degradable_volatile_solids
-            * GeneralConstants.GRAMS_TO_KG
-        )
+        return GasEmissionConstants.ACHIEVABLE_METHANE_EMISSION * manure_total_volatile_solids
```
tomhuhh, line 806: What is the reference for this equation? 

The unit for `GasEmissionConstants.ACHIEVABLE_METHANE_EMISSION` is kg, while the unit of what is returned from `calculate_CSTR_methane_volume` is m^3 as indicated in the doc string. This creates a discrepancy. Either the doc string and the variable name need to be updated or the equation requires a unit conversion. Thanks!

tomhuhh (changes requested):

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
Isgotkowitz: That's a great question - is anaerobic digestion the only manure processor component that can or that should record methane generated like this?

RUFAS/routines/manure/gas_emissions/calculator.py
Isgotkowitz: Nice catch Haowen, I will talk to Elle about this

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
elle-andreen: Yes, anaerobic digesters are the only component with methane generation being estimated via the `calculate_CSTR_methane_volume` method and reporting the methane generation mass and volume variables. In terms of outputs, this variable should only appear in the anaerobic digestion daily output, not any other manure treatment, so I think that should largely avoid confusion. That being said, if you (@tomhuhh) feel like it would be helpful to have the names changed in the actual code to help avoid confusing these anaerobic digestion-specific methane variables with other methane variables, that would be fine with me!

RUFAS/routines/manure/gas_emissions/calculator.py
elle-andreen: Great questions! First: this equation came directly from the SME's, they did not provide a specific reference. It is more of a generalization/expert interpretation of the agreed-upon constant conversion constant of manure VS to CH4 (240 L or 0.24 m3 of CH4 per kg manure VS). 

Second, for the `GasEmissionConstants.ACHIEVABLE_METHANE_EMISSION` constant: other than in the new CSTR methane estimation function in this PR, this constant is being called in calculator.py by the unused `methane_emission_from_anaerobic_lagoon` function, as well as in `ifsm_methane_emission` (used in open lots, CBPB). I double checked those design docs and I believe the original units of the constant (kg CH4/kg VS) are incorrect should be updated to m3 CH4/kg VS.

Isgotkowitz: I will update the constant to m^3 / kg

Isgotkowitz: @elle-andreen should all effected methods be changed so that they are returning cubic meters of methane, or should the equations be modified so that they convert the volume of methane to kg?

tomhuhh: Thanks for the great discussion here! I think we should make a note in the doc string that this equation comes from communication with subject matter experts and add a point of contact. I do not know about you guys, but I can easily forget this level of details. If we have a question about it down the road, it will cost us extra time and efforts to remember what exactly happened here and who we should contact. Thanks!

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
tomhuhh: Thank you for the clarification, Elle! I think choosing clear, descriptive variable names is essential for writing understandable and maintainable code. An example is that we recently changed the naming for enteric methane from 'methane' to 'enteric_methane' in the animal module code to avoid confusion with the variables in the manure module. Similarly, some variables in the anaerobic digestion submodule have already been using names containing 'AD' such as `AD_TAN_INCREASE_FACTOR`, `AD_carbon_dioxide`, `AD_VS_destruction`, etc. Thank you!

Isgotkowitz: By changing the variable name, I assume we will change outputs - probably in a way that would require someone's filters to be rewritten. Is this ok, or would this be considered too much of a change to make right now?

Isgotkowitz:
@elle-andreen @tomhuhh while updating the units for the `ACHIEVABLE_METHANE_EMISSION` constant I found an inconsistency in the `METHANE_CONVERSION_FACTOR` (located in GasEmissionConstants). It states that it is used to convert cubic meters of methane to kg of methane, but also that it is "unitless". This is wrong - should I also update it in this PR?

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -779,46 +779,60 @@ def _convert_temperature_celsius_to_kelvin(cls, temperature_celsius: float) -> f
...
+        """
+        Calculates the mass of methane lost from a digester.
 
-        Args:
-            methane_volume: Methane generation volume, m^3.
+        Parameters
+        ----------
+        generated_methane_mass : float
+            Amount of methane generated within the digester, kg.
+        digester_methane_leakage_fraction : float
```
matthew7838, line 818: I think we should also decide if we want to use the current format for units or the ('unit') format we had in S&C. It's not a big problem but the earlier we can have a conclusion the less we have to modify later on.

RUFAS/routines/manure/gas_emissions/calculator.py
matthew7838: Might be beneficial to know which SME's equation it's from or at least stating it.

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -779,46 +779,60 @@ def _convert_temperature_celsius_to_kelvin(cls, temperature_celsius: float) -> f
         return temperature_celsius + 273.15
 
     @classmethod
-    def methane_volume_via_Chen_equation(
-        cls, manure_total_degradable_volatile_solids: float, hydraulic_retention_time: int
-    ) -> float:
-        """Calculates CH4 generation volume using the Chen-Hashimoto equation.
+    def calculate_CSTR_methane_volume(cls, manure_total_volatile_solids: float) -> float:
+        """Calculates CH4 generation volume of anaerobic digestion in a continuously-stirred tank reactor.
```
matthew7838:
```suggestion
        """
        Calculates CH4 generation volume of anaerobic digestion in a continuously-stirred tank reactor.
```

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -779,46 +779,60 @@ def _convert_temperature_celsius_to_kelvin(cls, temperature_celsius: float) -> f
...
+        Calculates the mass of methane lost from a digester.
 
-        Args:
-            methane_volume: Methane generation volume, m^3.
+        Parameters
+        ----------
+        generated_methane_mass : float
+            Amount of methane generated within the digester, kg.
+        digester_methane_leakage_fraction : float
+            Fraction of generated methane that escapes as leakage.
```
matthew7838:
```suggestion
            Fraction of generated methane that escapes as leakage (unitless).
```

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -779,46 +779,60 @@ def _convert_temperature_celsius_to_kelvin(cls, temperature_celsius: float) -> f
...
+        -------
+        float
+            Mass of methane lost as leakage, kg.
+
+        """
+        return generated_methane_mass * digester_methane_leakage_fraction
+
+    @classmethod
+    def calculate_methane_energy_content(cls, methane_mass: float) -> float:
+        """Calculates energy content of methane generated in a digester.
```
matthew7838:
```suggestion
        """
        Calculates energy content of methane generated in a digester.
```

RUFAS/routines/manure/manure_treatments/anaerobic_digestion.py
```diff
@@ -74,40 +77,61 @@ def _calc_anaerobic_digestion_daily_output(
...
+        )
+        total_methane_generation_mass = total_methane_generation_volume * GasEmissionConstants.AD_METHANE_DENSITY
+        AD_carbon_dioxide = (
+            total_methane_generation_volume * GasEmissionConstants.AD_CARBON_DIOXIDE_TO_METHANE_RATIO
+        ) * GasEmissionConstants.AD_CARBON_DIOXIDE_DENSITY
+        AD_VS_destruction = total_methane_generation_mass + AD_carbon_dioxide
+
+        new_daily_output.liquid_manure_total_solids = (
+            self._current_manure_treatment_daily_input.liquid_manure_total_solids - AD_VS_destruction
+        )
```
matthew7838, line 101: What's the purpose of "()" ?

tests/test_manure_module/test_gas_emissions_calculator.py
```diff
@@ -434,45 +433,36 @@ def test_convert_tempC_to_tempK() -> None:
...
-                hydraulic_retention_time * GasEmissionConstants.SPECIFIC_GROWTH_RATE
-                + GasEmissionConstants.CHEN_HASHIMOTO_KINETIC_CONSTANT_KCH
-                - 1
-            )
-        )
-        * VS_total
-        * GeneralConstants.GRAMS_TO_KG
-    )
+
+    expected = GasEmissionConstants.ACHIEVABLE_METHANE_EMISSION * VS_total
```
matthew7838: If these are constants, maybe we can pass in manually calculated values for expected?

matthew7838 (approved):
LGTM! The coding looks good, but some minor comments and things that I think are a good place to clear things up up sort it. Overall, impressive work to manage so many file changes.

tomhuhh:
> @elle-andreen @tomhuhh while updating the units for the `ACHIEVABLE_METHANE_EMISSION` constant I found an inconsistency in the `METHANE_CONVERSION_FACTOR` (located in GasEmissionConstants). It states that it is used to convert cubic meters of methane to kg of methane, but also that it is "unitless". This is wrong - should I also update it in this PR?

It sounds like an easy fix. I would include the change in this PR. Thanks for catching that Ed! Good work!

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
tomhuhh: That is a good point, Ed! My understanding is that the variable name `methane_generation_mass` is already a modification from the original variable name `biogas`. This means that the filter needs to be changed either way.  Thanks!

RUFAS/routines/manure/manure_treatments/manure_treatment_configs.py
elle-andreen: I'm a little fuzzy on what purpose defaults in the configs serve - if there was no default in `ManureTreatmentConfig` and the user also did not supply a value in the inputs, would this cause the simulation to fail? Or just take the metadata default value?

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
elle-andreen: Haowen is correct, we're already proposing to change the variable name, so we may as well change it to exactly what we want. I think we can go ahead and add the AD prefix to all variables that are specific to anaerobic digestion. In `manure_treatment_daily_output.py`, this is all variables from line 159 to 191 (although some of these, like top cover volume and minimum digester volume, may end up being removed).

Just to add a note here as well regarding naming and other changes to outputs - right now we are making changes to this PR according to the "ideal" without regard to maintain backward-compatibility. We are waiting on specific guidance from Kristan and Pooya on how exactly to handle these types of proposed changes. The likely outcome is that we will create a branch off of this branch that contains only the "internal" changes to preserve backward-compatibility, and preserve the rest of the changes in this PR to be merged at a later date.

RUFAS/routines/manure/gas_emissions/calculator.py
elle-andreen: Agreed, we can add info on the source of the function directly to the docstring here (I'll push a commit with this @Isgotkowitz). I can also ask the SMEs if they have one or more references where this method was used.

elle-andreen:
> @elle-andreen @tomhuhh while updating the units for the `ACHIEVABLE_METHANE_EMISSION` constant I found an inconsistency in the `METHANE_CONVERSION_FACTOR` (located in GasEmissionConstants)


(Just pasting this here to preserve the info) The constant `METHANE_CONVERSION_FACTOR` is related to what % of VS are converted to methane I believe, it's not a units conversion. The constant `METHANE_FACTOR` is actually where the m3 -> kg conversion takes place, and its units are okay. The constants are definitely named confusingly but other than that there is no calculation or units issue that I can see!

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -779,46 +779,60 @@ def _convert_temperature_celsius_to_kelvin(cls, temperature_celsius: float) -> f
...
+    def calculate_CSTR_methane_volume(cls, manure_total_volatile_solids: float) -> float:
+        """Calculates CH4 generation volume of anaerobic digestion in a continuously-stirred tank reactor.
 
-        Args:
-            manure_total_degradable_volatile_solids: total volatile solids, kg.
-            hydraulic_retention_time: hydraulic retention time, days.
+        Parameters
+        ----------
+        manure_total_volatile_solids : float
+            Total volatile solids, kg.
```
elle-andreen, line 790:
```suggestion
This function originates from personal communications with subject matter experts Wei Liao (liaow@msu.edu) and April Leytem (april.leytem@usda.gov). The equation is a simplification of the IPCC Tier II estimate of CH4 emissions from anaerobic digesters, where CH4 generated in the digester is assumed to be equivalent to the amount of manure volatile solids loaded per day, multiplied by the generally-accepted methane potential value for dairy manure (240 L CH4 per kg of manure volatile solids).

File: RUFAS/routines/manure/manure_treatments/anaerobic_digestion.py (hunk 16 above)
Comment (Isgotkowitz) [reply RUFAS/routines/manure/manure_treatments/anaerobic_digestion.py:101]: It's just formatting. `new_daily_output.liquid_manure_total_solids = self._current_manure_treatment_daily_input.liquid_manure_total_solids - AD_VS_destruction` is too long to fit on a single line.

File: RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py (hunk 10 above)
Comment (Isgotkowitz) [reply RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py]: >Just to add a note here as well regarding naming and other changes to outputs - right now we are making changes to this PR according to the "ideal" without regard to maintain backward-compatibility. We are waiting on specific guidance from Kristan and Pooya on how exactly to handle these types of proposed changes. The likely outcome is that we will create a branch off of this branch that contains only the "internal" changes to preserve backward-compatibility, and preserve the rest of the changes in this PR to be merged at a later date.

I can make these changes if that is what is desired for this PR. However we should consider whether this is an effective development strategy to employ. I think there are a number of problems with it.

First, making a lot of changes but only intending to actually merge some of them by creating another PR with a subset of those changes is a fraught process. It requires making the same changes not once but twice, which is not efficient but more importantly, opens up a window for those changes to be made incorrectly.

Second, after the changes that are possible are merged in, we're left with a PR that might easily be left open for a very long time. This is bad for several reasons: it takes time and effort to keep that PR from getting stale (potentially a lot of time and effort, because there might be significant changes merged in which require fixing a lot of merge conflicts). Also, the longer the PR is open, the harder it will be to remember exactly what and why it has the changes it does. Which will increase the amount of time needed to keep it updated.

In my opinion, a better strategy would be to write an issue specifically for the changes that we are unsure about, get those changes greenlit with Pooya and/or Kristan, then implement those changes. That way, the time and effort to actually make the changes only happens after we know that we are ready to accept the changes. If it turns out that we don't want to make the changes, then we just close the issue without having spent much time working on it. This agrees with what is stated in the code review Wiki: "Making a PR means that the code is ready to be merged as-is. A PR has to have any temporary and test files and pieces of code removed, and a clean and ready-to-merge submission is needed."

File: RUFAS/routines/manure/gas_emissions/calculator.py (hunk 11 above)
Comment (Isgotkowitz) [reply RUFAS/routines/manure/gas_emissions/calculator.py:806]: Thanks Elle!

File: RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py (hunk 10 above)
Comment (elle-andreen) [reply RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py]: I agree that it's not a great option, but I'm just relaying what has been conveyed to me. From my understanding, we should receive specific guidance on what to do with this specific PR, and also for these cases going forward, so I would not change anything until we hear from Kristan/Pooya (but I would definitely share your comment with them!). The request to make these changes to anaerobic digestion and the work on this branch predated the initiation of the new branching strategy, and I think if the branching strategy change plus absolute requirement for backward compatibility had been apparent at that time, we would have gone about this PR differently. I expect this will not be the only case where there are good but maybe not immediately urgent reasons to change inputs/outputs, I think we're just waiting on technical guidance for how to handle that

File: RUFAS/routines/manure/gas_emissions/calculator.py (hunk 12 above)
Comment (Isgotkowitz) [reply RUFAS/routines/manure/gas_emissions/calculator.py:818]: I agree. For now though I will keep it consistent with the rest of the units in this file.

File: tests/test_manure_module/test_gas_emissions_calculator.py (hunk 17 above)
Comment (Isgotkowitz) [reply tests/test_manure_module/test_gas_emissions_calculator.py]: Yes, that would be much better.

**Comment by Isgotkowitz** [comment]
> > @elle-andreen @tomhuhh while updating the units for the `ACHIEVABLE_METHANE_EMISSION` constant I found an inconsistency in the `METHANE_CONVERSION_FACTOR` (located in GasEmissionConstants)
> 
> (Just pasting this here to preserve the info) The constant `METHANE_CONVERSION_FACTOR` is related to what % of VS are converted to methane I believe, it's not a units conversion. The constant `METHANE_FACTOR` is actually where the m3 -> kg conversion takes place, and its units are okay. The constants are definitely named confusingly but other than that there is no calculation or units issue that I can see!

Sorry, you're right, I was unsure of the units on `METHANE_FACTOR`, not `METHANE_CONVERSION_FACTOR`. I do think though that the docstring for `METHANE_FACTOR` is incorrect, if it converts cubic meters of methane to kg methane then its units are $m^3$ / kg, not unitless.

File: RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py (hunk 10 above)
Comment (Isgotkowitz) [reply RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py]: Ok, thank you @elle-andreen. I will hold off on making more changes for now. Do you want to reach out to Kristan or Pooya about this PR, or would it be better if I asked about it?

Also, if it would be helpful, I could go ahead and split up this PR so that there is an additional PR, that one with backwards-compatible changes and this one with non-backwards compatible changes?

File: RUFAS/routines/manure/manure_treatments/anaerobic_digestion.py [hunk 19]
```diff
@@ -74,40 +77,61 @@ def _calc_anaerobic_digestion_daily_output(
...
+            methane_leakage / GasEmissionConstants.AD_METHANE_DENSITY
+        )
+        captured_methane_generation_mass = total_methane_generation_mass - methane_leakage
+        captured_methane_energy_content = GasEmissionsCalculator.calculate_methane_energy_content(
+            methane_mass=captured_methane_generation_mass
         )
         new_daily_output.heating_input_energy = heating_input_energy
         new_daily_output.evaporated_water = self.config.evaporation_fraction * daily_final_manure_volume
-        new_daily_output.biogas_energy_content = biogas_energy_content
+        new_daily_output.methane_energy_content = captured_methane_energy_content
```
elle-andreen:
```suggestion
        new_daily_output.biogas_energy_content = captured_methane_energy_content
```
Reverting output name change but preserve new internal calculation of biogas methane content

RUFAS/routines/manure/manure_treatments/anaerobic_digestion.py
```diff
@@ -74,40 +77,61 @@ def _calc_anaerobic_digestion_daily_output(
...
         new_daily_output.minimum_digester_volume = minimum_digester_volume
         new_daily_output.top_cover_volume = top_cover_volume
-        new_daily_output.methane_generation_volume = methane_generation_volume
+        new_daily_output.methane_leakage_mass = methane_leakage
+        new_daily_output.methane_generation_volume = captured_methane_generation_volume
+        new_daily_output.methane_generation_mass = captured_methane_generation_mass
```
elle-andreen:
```suggestion
        new_daily_output.biogas = captured_methane_generation_mass
```
Remove the output name change but preserve the new value of the output.

RUFAS/routines/manure/manure_treatments/anaerobic_digestion_and_lagoon.py
```diff
@@ -84,13 +84,23 @@ def _daily_update_helper(self) -> ManureTreatmentDailyOutput:
...
+        )
+        self._accumulated_output.liquid_manure_total_volatile_solids = (
+            self._anaerobic_lagoon._accumulated_output.liquid_manure_total_volatile_solids
+        )
         self._accumulated_output.liquid_manure_total_degradable_volatile_solids = (
             self._anaerobic_lagoon._accumulated_output.liquid_manure_total_degradable_volatile_solids
         )
         self._accumulated_output.liquid_manure_total_non_degradable_volatile_solids = (
             self._anaerobic_lagoon._accumulated_output.liquid_manure_total_non_degradable_volatile_solids
         )
```
elle-andreen, line 105:
```suggestion

anaerobic_lagoon_daily_output.biogas = self.anaerobic_digestion_daily_output.biogas
```
Reverting removal of this output to maintain backwards compatibility

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
```diff
@@ -156,18 +156,23 @@ class ManureTreatmentDailyOutput(LiquidManurePortionProtocol):
...
+    methane_generation_mass_unit: MeasurementUnits, default MeasurementUnits.KILOGRAMS_PER_DAY
+        Unit for methane_generation_mass.
+    methane_energy_content : float, default 0.0
+        Energy content of methane, MJ.
+    methane_energy_content_unit: MeasurementUnits, default MeasurementUnits.MEGAJOULES
+        Unit for methane_energy_content.
```
elle-andreen:
```suggestion
   biogas: float, default 0.0
        Mass of methane produced from anaerobic digestion, kg.
   biogas_unit: MeasurementUnits, default MeasurementUnits.KILOGRAMS_PER_DAY
        Unit for methane_generation_mass.
    biogas_energy_content : float, default 0.0
        Energy content of methane, MJ.
   biogas_energy_content_unit: MeasurementUnits, default MeasurementUnits.MEGAJOULES
        Unit for methane_energy_content.
```
Reverting output name changes to maintain backwards compatibility

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
```diff
@@ -296,16 +301,18 @@ class ManureTreatmentDailyOutput(LiquidManurePortionProtocol):
     solid_manure_daily_mass: float = 0.0
     solid_manure_daily_mass_unit: MeasurementUnits = MeasurementUnits.KILOGRAMS
 
-    # different unit from docMeasurementUnitsing m^3 vs m^3/day?
-    biogas: float = 0.0  # biogas production per day (m3/day)
-    biogas_unit: MeasurementUnits = MeasurementUnits.CUBIC_METERS_PER_DAY
+    methane_generation_mass: float = 0.0
+    methane_generation_mass_unit: MeasurementUnits = MeasurementUnits.KILOGRAMS_PER_DAY
```
elle-andreen:
```suggestion
   biogas: float = 0.0
    biogas_unit: MeasurementUnits = MeasurementUnits.KILOGRAMS_PER_DAY
```
Reverting output name changes to maintain backwards compatibility

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
```diff
@@ -296,16 +301,18 @@ class ManureTreatmentDailyOutput(LiquidManurePortionProtocol):
...
 
-    biogas_energy_content: float = 0.0  # biogas energy content (MJ/m3)
-    biogas_energy_content_unit: MeasurementUnits = MeasurementUnits.MEGAJOULES_PER_CUBIC_METER
+    methane_energy_content: float = 0.0
+    methane_energy_content_unit: MeasurementUnits = MeasurementUnits.MEGAJOULES
```
elle-andreen:
```suggestion
   biogas: float = 0.0
   biogas_unit: MeasurementUnits = MeasurementUnits.KILOGRAMS_PER_DAY
```
Reverting output name changes to maintain backwards compatibility

RUFAS/routines/manure/constants_and_units/gas_emission_constants.py
```diff
@@ -56,26 +56,29 @@ class GasEmissionConstants:
...
-    (methane : carbon dioxide).
+    Unit conversion factor for carbon dioxide generated from anaerobic digestion at 1 atm of pressure and 37.5C
+    (kg/:math:`m^3`).
     """
 
-    AD_METHANE_TO_METHANE_CARBON_DIOXIDE_RATIO: float = 2.86
+    AD_CARBON_DIOXIDE_TO_METHANE_RATIO: float = 4 / 6
     """
-    The mass conversion factor from methane to methane and carbon dioxide in anaerobic digestion (kg CH4 / kg CO2).
+    The volumetric ratio of carbon dioxide to methane generated during anaerobic digestion :math:`m^3`
```
Isgotkowitz: Would the units here be cubic meters CO2 to cubic meters CH4?

RUFAS/routines/manure/manure_treatments/manure_treatment_configs.py
Isgotkowitz: > I'm a little fuzzy on what purpose defaults in the configs serve

Honestly they don't serve much of a purpose at all, if all the defaults were removed I would expected the behavior to be the same for the most part.

> if there was no default in `ManureTreatmentConfig` and the user also did not supply a value in the inputs, would this cause the simulation to fail?

Not necessarily, it depends on the input. If there was a default for that input in the metadata then that default would be used and the simulation would run normally. If there was no default in the metadata then the simulation would crash, but that crash would occur while parsing the inputs. In either case though, it wouldn't really matter if there was a default in `ManureTreatmentConfig` or not

RUFAS/routines/manure/constants_and_units/gas_emission_constants.py
elle-andreen: Yes, I think that would be the more technically correct unit! (m^3 CO2 : m^3 CH4)

Isgotkowitz: Will update it

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
```diff
@@ -296,16 +301,18 @@ class ManureTreatmentDailyOutput(LiquidManurePortionProtocol):
...
-    # different unit from docMeasurementUnitsing m^3 vs m^3/day?
-    biogas: float = 0.0  # biogas production per day (m3/day)
-    biogas_unit: MeasurementUnits = MeasurementUnits.CUBIC_METERS_PER_DAY
+    biogas: float = 0.0
+    biogas_unit: MeasurementUnits = MeasurementUnits.KILOGRAMS_PER_DAY
 
-    biogas_energy_content: float = 0.0  # biogas energy content (MJ/m3)
-    biogas_energy_content_unit: MeasurementUnits = MeasurementUnits.MEGAJOULES_PER_CUBIC_METER
+    biogas_energy_content: float = 0.0
+    biogas_energy_content_unit: MeasurementUnits = MeasurementUnits.KILOGRAMS_PER_DAY
```
elle-andreen:
```suggestion
    biogas_energy_content_unit: MeasurementUnits = MeasurementUnits.MEGAJOULES
```

Isgotkowitz: Nice catch

elle-andreen (approved):
Thanks for your hard work on this Ed!

Because this PR includes changes to inputs and outputs (though backwards-compatibility is maintained), for others' reference, here is a table of the input and output changes from this PR. The actual table can be accessed [here](https://docs.google.com/document/d/1om-czbamAUBlSAGegGuNhlSUny6EWygnwFSLF6wWkx0/edit?usp=sharing).


image

tomhuhh:
Great work Elle and Ed! I tested with manure scenarios 5 through 7, before and after the changes suggested in this PR and the results are summarized below. 

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-1e05ef04-7fff-0d96-1206-769607413276">
  | AD updated | main
-- | -- | --
5: Manure trt storage CH4(kg) | 7.86 | 8.64
5: Biogas (m^3/day) | 26.48 | 19.07
6: Manure trt storage CH4(kg) | 4.96 | 7.29
6: Biogas (m^3/day) | 26.86 | 16.16
7: Manure trt storage CH4(kg) | 2.51 | 6.03
7: Biogas (m^3/day) | 35.40 | 21.34

</b>

The results agree with my expectations. Generally, with the changes proposed, more methane (biogas) is generated during the anaerobic digestion step compared to previously, and less methane (storage_methane) is generated during the anaerobic lagoon step. 

One final observation is that I noticed some inconsistency in terms of how biogas is reported. It seems that biogas is also listed as an attribute of ManureTreatmentDailyOutput in addition to as an attribute of anaerobic digestion. This makes sense because we can consider anaerobic digestion and lagoon as a manure treatment. However, if that is the case, we should also output biogas_energy_content, methane_generaton_volume, and methane_leakage_mass as attributes of ManureTreatmentDailyOutput as well. Currently they are output as zero. 

As my final observation has minimal impact on the final outcome of the model, so I decided to approve this PR. My suggestion is to have it fixed before the merge. Thanks!

tomhuhh (approved):

Isgotkowitz:
> Great work Elle and Ed! I tested with manure scenarios 5 through 7, before and after the changes suggested in this PR and the results are summarized below.
> 
>  	AD updated	main
> 5: Manure trt storage CH4(kg)	7.86	8.64
> 5: Biogas (m^3/day)	26.48	19.07
> 6: Manure trt storage CH4(kg)	4.96	7.29
> 6: Biogas (m^3/day)	26.86	16.16
> 7: Manure trt storage CH4(kg)	2.51	6.03
> 7: Biogas (m^3/day)	35.40	21.34
> The results agree with my expectations. Generally, with the changes proposed, more methane (biogas) is generated during the anaerobic digestion step compared to previously, and less methane (storage_methane) is generated during the anaerobic lagoon step.
> 
> One final observation is that I noticed some inconsistency in terms of how biogas is reported. It seems that biogas is also listed as an attribute of ManureTreatmentDailyOutput in addition to as an attribute of anaerobic digestion. This makes sense because we can consider anaerobic digestion and lagoon as a manure treatment. However, if that is the case, we should also output biogas_energy_content, methane_generaton_volume, and methane_leakage_mass as attributes of ManureTreatmentDailyOutput as well. Currently they are output as zero.
> 
> As my final observation has minimal impact on the final outcome of the model, so I decided to approve this PR. My suggestion is to have it fixed before the merge. Thanks!

Thank you for taking to the time to do such a detailed analysis @tomhuhh. I will look into why the biogas_energy_content, methane_generation_volume and leakage mass are 0 and figure out a fix.

elle-andreen:
> > One final observation is that I noticed some inconsistency in terms of how biogas is reported. It seems that biogas is also listed as an attribute of ManureTreatmentDailyOutput in addition to as an attribute of anaerobic digestion. This makes sense because we can consider anaerobic digestion and lagoon as a manure treatment. However, if that is the case, we should also output biogas_energy_content, methane_generaton_volume, and methane_leakage_mass as attributes of ManureTreatmentDailyOutput as well. Currently they are output as zero.
> > As my final observation has minimal impact on the final outcome of the model, so I decided to approve this PR. My suggestion is to have it fixed before the merge. Thanks!
> 
> Thank you for taking to the time to do such a detailed analysis @tomhuhh. I will look into why the biogas_energy_content, methane_generation_volume and leakage mass are 0 and figure out a fix.

@tomhuhh I may be wrong, but my understanding is that the ManureTreatmentDailyOutput values for pens assigned to the anaerobic digestion and lagoon combo treatment are from only the anaerobic lagoon, not both the digester and lagoon, as the digester output values are reported separately in their own output. I think the repetition of the biogas variables/values in both the anaerobic digestion and anaerobic lagoon outputs is confusing and I worry that adding these values where there were previously 0's may cause confusion for FARM ES. If not for needing to keep all changes backward-compatible, in this PR we would have removed the biogas output value from the anaerobic lagoon daily output I believe. I realize it's not ideal, but my recommendation would be to leave those outputs alone for now, then resolve or remove them once we are able to make non-backwards-compatible changes again, but I am open to your thoughts! We do also have a PR to separate manure treatments and storages (issue #1619) which will likely result in the removal of this type of output from the set of anaerobic lagoon daily outputs.

tomhuhh:
> > > One final observation is that I noticed some inconsistency in terms of how biogas is reported. It seems that biogas is also listed as an attribute of ManureTreatmentDailyOutput in addition to as an attribute of anaerobic digestion. This makes sense because we can consider anaerobic digestion and lagoon as a manure treatment. However, if that is the case, we should also output biogas_energy_content, methane_generaton_volume, and methane_leakage_mass as attributes of ManureTreatmentDailyOutput as well. Currently they are output as zero.
> > > As my final observation has minimal impact on the final outcome of the model, so I decided to approve this PR. My suggestion is to have it fixed before the merge. Thanks!
> > 
> > 
> > Thank you for taking to the time to do such a detailed analysis @tomhuhh. I will look into why the biogas_energy_content, methane_generation_volume and leakage mass are 0 and figure out a fix.
> 
> @tomhuhh I may be wrong, but my understanding is that the ManureTreatmentDailyOutput values for pens assigned to the anaerobic digestion and lagoon combo treatment are from only the anaerobic lagoon, not both the digester and lagoon, as the digester output values are reported separately in their own output. I think the repetition of the biogas variables/values in both the anaerobic digestion and anaerobic lagoon outputs is confusing and I worry that adding these values where there were previously 0's may cause confusion for FARM ES. If not for needing to keep all changes backward-compatible, in this PR we would have removed the biogas output value from the anaerobic lagoon daily output I believe. I realize it's not ideal, but my recommendation would be to leave those outputs alone for now, then resolve or remove them once we are able to make non-backwards-compatible changes again, but I am open to your thoughts! We do also have a PR to separate manure treatments and storages (issue #1619) which will likely result in the removal of this type of output from the set of anaerobic lagoon daily outputs.

Hello Elle! I think this is a great idea. Let's write an issue for it and merge this PR for now. 

@Isgotkowitz Do you mind keeping a record of this issue? Thanks!

Isgotkowitz:
@tomhuhh pointed out that anaerobic lagoons are reporting biogas when they are combined with anaerobic digesters, which is why it would be strange that lagoons would not also report biogas energy content, methane generation, and methane leakage. It seemed to us that it would be best to remove this copying of biogas so that anaerobic lagoons never report having any biogas, but this should be done in a follow-up PR so that we can take our time to clear this change with FARM-ES.

PR #1854, opened 2024-07-23
Properly distribute root residue & nutrients through soil profile

### LINKED ISSUE #1811: Crop root biomass distribution by root growth patterns
Crop roots often grow more near the surface and then have less root structure as the soil depth increases. We would like to improve the RuFaS representation of the distribution of the root biomass among the soil layers. 

Currently RuFaS predicts the total root biomass and then distributes the total root biomass between the soil layers based only on soil layer depth (i.e. each soil layer's proportion of the total soil profile) rather than the expected root biomass distribution among layers. 

The method for the root distribution is described in this paper: https://www.sciencedirect.com/science/article/pii/S0378429016300399?via%3Dihub

EDIT: This method is slightly more involved than I originally thought so I started a design doc: https://docs.google.com/document/d/1iMSzJ_HXmAgOAC9d8apqzoRnQDxl9o0EkdGf0SB8Kxk/edit?usp=sharing

We could leave it as it is but results in not enough nutrient update from upper layers and too much nutrient deposition in lower layers on plant death

**Additional context**

**Target Deadline**
July 23

Isgotkowitz:
This PR modifies how root biomass and the nutrients contained within are distributed throughout the soil profile when a crop dies/is killed.

### Context

Issue(s) closed by this pull request: closes #1811 

- [ ] Yes
- [x] No

### What

The three main changes made to achieve proper root mass and nutrient distribution are:
- Adding routines and parameters to the Crop submodule which facilitate more accurate mass distributions based on the Crop's biology and state.
- Modifying the way that root mass is transferred from the Crop submodule to the Carbon Cycling submodule.
- Fixing a bug that allowed crop's roots to extend past the bottom of the deepest soil layer.

### Why

As stated in the issue, the previous method for distributing root mass and nutrients was overly simple, and was resulting in too much N, P and C being partitioned into lower levels of the soil.

Before, the Carbon Cycling submodule had implemented the same mass distribution logic as the Crop module. With the change in mass distribution logic, I felt it would be beneficial to consolidate the code so that there was only one place where mass from roots is distributed into the soil profile.

### How

The equation that calculates the proportion of root biomass contained above a certain soil depth has been implemented in the new method `_calculate_root_mass_distribution` in `crop_management.py`. This new method is now utilized in `_distribute_residue_nutrients`, which now handles partitioning root biomass between the different soil layers in addition to N and P.

The logic in `residue_partition.py` has been changed so that it only transfers the mass from the `plant_residue` attribute to the metabolic and structural litter pools, and has had a few methods removed as they are no longer needed. To support this new system of transferring residue mass, `soil_data.py` has had several attributes removed as they should no longer be used. The attributes removed include `plant_root_residue`, `plant_surface_residue`, `total_residue`, and `crop_root_depth`. Their use has been replaced by `plant_residue` in `layer_data.py`. `dormancy.py` has been updated to account for this change as well.

To fix the bug that allowed crops' roots to grow past the bottom of the soil, I changed `_make_supported_crop` in `field.py` to bound the maximum root depth, and warn users when this happens.

### Test plan

All unit tests have been updated and run for added and modified code - with the exception of some small changes to how and which variables are recorded. I made these changes to make testing easier (more on this below), but if people would like to keep these changes then I can update tests accordingly.

To test that the new distribution functionality was working correctly and the mass balance was preserved, I used the below filter to verify that there was not a loss of biomass, N or P when transferring crop residue to the soil. Note that I only tried this out on the default scenario, I haven't used it on any of the pilot farms yet.
```json
{
    "multiple": [
        {
            "name": "C",
            "filters": [
                "^(?!.*vadose)FieldDataReporter.send_daily_variables.structural_litter_amount.*",
                "^(?!.*vadose)FieldDataReporter.send_daily_variables.metabolic_litter_amount.",
                "CropManagement._record_yield.harvest_yield.*"
            ],
            "variables": ["yield_residue"],
            "expand_data": true
        },
        {
            "name": "N",
            "filters": [
                "^(?!.*vadose)FieldDataReporter.send_daily_variables.fresh_organic_nitrogen_content.*",
                "^(?!.*vadose)FieldDataReporter.send_daily_variables.active_organic_nitrogen_content.*",
                "CropManagement._record_yield.harvest_yield.*"
            ],
            "variables": ["residue_nitrogen"],
            "expand_data": true
        },
        {
            "name": "P",
            "filters": [
                "^(?!.*vadose)FieldDataReporter.send_daily_variables.labile_inorganic_phosphorus_content.*",
                "CropManagement._record_yield.harvest_yield.*"
            ],
            "variables": ["residue_phosphorus"],
            "expand_data": true
        }
    ]
}
```

### Updates
This PR has changed a lot through the review process. Please read through the comments on this PR to see all the changes that have been made in this PR and an updated filter with which the changes have been tested/verified.


Isgotkowitz:
Update: while doing some initial testing, I found an upstream bug. When a Crop's maximum rooting depth is allowed to be deeper than the soil profile it is growing in - which screws up the root biomass distribution equation that is implemented in this equation. I will fix this bug, then continue the initial testing.

RUFAS/routines/field/crop/crop_management.py
```diff
@@ -392,40 +391,73 @@ def _distribute_residue_nutrients(self, soil_data: SoilData, root_residue_mass:
...
 
         for layer in soil_data.soil_layers[1:]:
-            layer_fraction = (
-                (layer.bottom_depth - layer.top_depth) / self.data.root_depth
-                if layer.bottom_depth <= self.data.root_depth
-                else max(0.0, (self.data.root_depth - layer.top_depth) / self.data.root_depth)
-            )
+            root_frac_to_top_depth = self._calculate_root_mass_distribution(layer.top_depth)
+            root_frac_to_bottom_depth = self._calculate_root_mass_distribution(layer.bottom_depth)
+            layer_fraction = root_frac_to_bottom_depth - root_frac_to_top_depth
```
allisterakun: Assuming the soil layers are consecutive, the bottom depth of layer 1 is the same as the top depth of layer 2. This means that the root fraction at the bottom of layer 1 is the same as the root fraction at the top of layer 2.

My point is that is there a way to use the root fraction at the bottom depth of the last layer as the root fraction 
at the top depth of the current layer except for the first layer, where this number should be 0) so we don't have to do the same calculation twice.

RUFAS/routines/field/manager/field_data_reporter.py
```diff
@@ -26,11 +26,12 @@ class FieldDataReporter:
     def __init__(self, fields: List[Field]):
         self.fields = fields
 
-    def send_daily_variables(self) -> None:
+    def send_daily_variables(self, time) -> None:
```
allisterakun:
```suggestion
    def send_daily_variables(self, time: Time) -> None:
```

RUFAS/routines/field/soil/carbon_cycling/residue_partition.py
```diff
@@ -111,16 +111,14 @@ def partition_residue(self, rainfall: float) -> None:
         layer.plant_dry_matter_residue_amount = 0
 
         layer.weighted_residue_dry_matter_lignin_fraction = self._determine_weighted_residue_dry_matter_lignin_fraction(
-            layer.soil_dry_matter_residue_amount, self.data.plant_root_residue
+            layer.soil_dry_matter_residue_amount, 0.0
```
allisterakun, line 114: I'm just wondering, why are we hard-coding it to 0.0 here?

allisterakun (changes requested):
Great job, Ed! It generally looks good. 
Despite the little crash course, I still have limited knowledge about soil.
I left three comments, mostly just clarification questions.
I requested changes because I see that the Pytest is crashing; otherwise, the code itself looks good.

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: Good point! I had not thought about this. The soil layers are definitely supposed to be consecutive, so I think you are right - we should be able to eliminate some of these calculations.

RUFAS/routines/field/soil/carbon_cycling/residue_partition.py
Isgotkowitz: Short answer: because `plant_root_residue` should always have been 0 when this method, `partition_residue` was being called. So the easiest way to remove `plant_root_residue` without touching anything else in the code was to just hardcode its value to 0.

Long answer: before this PR, the `plant_root_residue` pool was only being used to transfer mass from the crop module to the soil module. After that transfer completed, `plant_root_residue` was reset to 0. The method this code is in, `partition_residue`, runs daily and is not responsible for transferring any of that mass and is only responsible for degrading the structural and metabolic litter pools (despite what the name and docstring suggest!). So when `partition_residue` runs, `plant_root_residue` was only ever 0. Thus the easiest path to removing this value was to hardcode it to 0.

This is not a good solution - but there is such a large amount of technical debt in the carbon cycling module that my strategy with it was to remove as much as I could without changing any of the overall functionality.

Isgotkowitz:
> Great job, Ed! It generally looks good. Despite the little crash course, I still have limited knowledge about soil. I left three comments, mostly just clarification questions. I requested changes because I see that the Pytest is crashing; otherwise, the code itself looks good.

Thanks @allisterakun. I modified how output is collected so that the outputs from C&S would be easier to evaluate - but I haven't updated the testing for them, which is why Pytest is crashing. I will leave these changes in here to (hopefully) make it easier for @rawalashmita to evaluate the outputs. If we decide they are useful enough to keep I will update the unit tests, otherwise I will roll back these changes.

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: Implemented!

rawalashmita:
Looking at the output data, soil phosphorus (P) levels did not make sense by layer. P concentrations are almost always greater toward the surface and decreased with depth. Is this a particular field, rotation? Need to know what crop was being simulated and if there were any inputs during the simulation (i.e, manure or fertilizer) as it will help make sense of the root biomass data. Thank you!

KFosterReed:
> Looking at the output data, soil phosphorus (P) levels did not make sense by layer. P concentrations are almost always greater toward the surface and decreased with depth. Is this a particular field, rotation? Need to know what crop was being simulated and if there were any inputs during the simulation (i.e, manure or fertilizer) as it will help make sense of the root biomass data. Thank you!

All the information about the crop, soil type, manure + fertilizer applications, and other field managment practices can be found in the input files! There are also records of the applications/plantings/tillages etc reported in the outputs for confirmation but I would suggest first becoming familiar with model inputs to know what is being simulated in each field.

KFosterReed:
So this is working about 90% as I would expect! The equations and parameters are all translated correctly. The fractions that are allocated to each soil make sense and are applied correctly to the nutrient partitioning so the different nutrients are distributed according to the fractions that are estimated by the updated `calculate_root_mass_distribution` function. 

Where I ran into some trouble was when I tested it on crops that would be expected to have different root structures and depths and found that they had almost identical fractions allocated to each layer (corn and alfalfa in this case). This led me to look at the root depths of each crop and I noticed that both crops are maxing out at the same value (2000 mm in this case), instead of at their respective maximum rooting depths (2000 mm and 3000 mm for corn and alfalfa respectively). I am going to dig a little deeper and create an issue about this root depth bug in particular but I think we have two options for this PR: 

1. wait until the simulated crop root depth is fixed
2. replace the RuFaS predicted root depth with the maximum rooting depth published by Fan et al.

Isgotkowitz:
> So this is working about 90% as I would expect! The equations and parameters are all translated correctly. The fractions that are allocated to each soil make sense and are applied correctly to the nutrient partitioning so the different nutrients are distributed according to the fractions that are estimated by the updated `calculate_root_mass_distribution` function.
> 
> Where I ran into some trouble was when I tested it on crops that would be expected to have different root structures and depths and found that they had almost identical fractions allocated to each layer (corn and alfalfa in this case). This led me to look at the root depths of each crop and I noticed that both crops are maxing out at the same value (2000 mm in this case), instead of at their respective maximum rooting depths (2000 mm and 3000 mm for corn and alfalfa respectively). I am going to dig a little deeper and create an issue about this root depth bug in particular but I think we have two options for this PR:
> 
> 1. wait until the simulated crop root depth is fixed
> 2. replace the RuFaS predicted root depth with the maximum rooting depth published by Fan et al.

@KFosterReed thank you for looking into this.

KFosterReed:
Since the maximum simulated rooting depth for a crop is curently intentionally capped at the bottom of the soil profile, I suggest we move forward with this PR and investigate if we want to change the root depth maximum. My guess is that we will want to change it because the total soil profile depth is a user defined variable and just because someone is only interested in tracking what happens in the top 200 cm (or 15 cm) of the soil profile, doesn't mean the crop roots should be limited to that depth. Changing this might have more implications for how the vadose zone is treated and nutrient uptake is simulated so I don't want to add that to this PR. 

However, to acheive the goal of having a more realistic distribution of root biomass and nutrients, I would like to change the root depth used in this new method to the fixed `max_root_depth` crop attribute. This means that the root depth will extend beyond the bottom of most currently simulated Layer 3 depths and we should allocate the remainder of the root biomass and nutrients to the vadose zone

Isgotkowitz:
> Since the maximum simulated rooting depth for a crop is curently intentionally capped at the bottom of the soil profile, I suggest we move forward with this PR and investigate if we want to change the root depth maximum. My guess is that we will want to change it because the total soil profile depth is a user defined variable and just because someone is only interested in tracking what happens in the top 200 cm (or 15 cm) of the soil profile, doesn't mean the crop roots should be limited to that depth. Changing this might have more implications for how the vadose zone is treated and nutrient uptake is simulated so I don't want to add that to this PR.
> 
> However, to acheive the goal of having a more realistic distribution of root biomass and nutrients, I would like to change the root depth used in this new method to the fixed `max_root_depth` crop attribute. This means that the root depth will extend beyond the bottom of most currently simulated Layer 3 depths and we should allocate the remainder of the root biomass and nutrients to the vadose zone

Roger that, I will make those changes.

Isgotkowitz:
@KFosterReed I'm sorry, just to confirm, the depth of the roots should _not_ be capped or affected at all by the depth of the soil profile?

KFosterReed:
> @KFosterReed I'm sorry, just to confirm, the depth of the roots should _not_ be capped or affected at all by the depth of the soil profile?

Hey Ed! My suggestion at this point is not to change the simulated depth of the roots (`crop.data.root_depth`) at this point but to replace the use of this simulated variable in `calculate_root_mass_distribution()` with `max_root_depth` until we have decided how to and updated the methods for limiting the `root_depth` variable. 

My thought behind this is, as you pointed out, `root_depth` has several other impacts on model outcomes beyond this method. Allowing the root biomass to be distributed up to the maximum rooting depth has known and limited implications for now and is probably closer to keeping all the root biomass in the higher layers if the expeted depth is deeper.

Isgotkowitz:
Update: I have modified how the root distribution is calculated. It uses the `max_root_depth` instead of `root_depth`, but the max root depth values that were in RuFaS previously, not the ones from Fang et al. The depth that roots grow in RuFaS is no longer affected by the depth of the soil profile, at all. If it would be good to have a warning or error logged when this occurs, I can add that in. Otherwise I will just write up an issue for it.

I took a brief look at the outputs, and my very quick and dirty attempt at a mass balance indicated that the code was functioning properly. The filter I used for this is attached below. Note that biomass goes into the plant residue pool of the vadose zone the same as other soil layers, but unlike other soil layers, that mass stays in that pool instead of being divided into the structural and metabolic litter pools.
```json
{
    "multiple": [
        {
            "name": "C",
            "filters": [
                "^(?!.*vadose)FieldDataReporter.send_daily_variables.structural_litter_amount.*",
                "^(?!.*vadose)FieldDataReporter.send_daily_variables.metabolic_litter_amount.",
                ".*plant_residue.*vadose.*",
                "CropManagement._record_yield.harvest_yield.*"
            ],
            "variables": ["yield_residue"]
        },
        {
            "name": "N",
            "filters": [
                "^(?!.*vadose)FieldDataReporter.send_daily_variables.fresh_organic_nitrogen_content.*",
                "FieldDataReporter.send_daily_variables.active_organic_nitrogen_content.*",
                "CropManagement._record_yield.harvest_yield.*"
            ],
            "variables": ["residue_nitrogen"]
        },
        {
            "name": "P",
            "filters": [
                "FieldDataReporter.send_daily_variables.labile_inorganic_phosphorus_content.*",
                "CropManagement._record_yield.harvest_yield.*"
            ],
            "variables": ["residue_phosphorus"]
        }
    ]
}
```
CC @KFosterReed @rawalashmita @allisterakun

KFosterReed (approved):
I ran this with both alfalfa and and corn and it now partitions the roots between layers, including the vadose zone, for both types of crops that have different rooting depths!

Great work Ed!

KFosterReed:
Also, for those interested here is how it is partitioning the nutrients between the layers

**For Corn:** 

- **Surface Fraction:**  0.6280029651948561
- **Layer 0 root frac:**  0.09065795799969945
- **Layer 1, bottom:  230.0**,  root frac:  0.5372387372563443
- **Layer 2, bottom:  890.0,**   root frac:  0.2800965697807434
- **Layer 3, bottom:  2000.0,**   root frac:  0.09200673496321288
- **Layer 4, bottom:**  10000000  root frac:  0.0

**For Alfalfa:** 

- **Surface Fraction**:  0.3340332001742058
- **Layer 0 root frac**:  0.08267668624953185
- **Layer 1,  bottom:  230.0**,  root frac:  0.4490465227999927
- **Layer 2,  bottom:  890.0,**  root frac:  0.30429388963238424
- **Layer 3, bottom:  2000.0,**  root frac:  0.11589289184400464
- **Layer 4, bottom:  10000000,**  root frac:  0.04809000947408659

RUFAS/routines/field/crop/crop_management.py
```diff
@@ -392,39 +390,72 @@ def _distribute_residue_nutrients(self, soil_data: SoilData, root_residue_mass:
...
+        surface_layer.labile_inorganic_phosphorus_content += subsurface_phosphorus * root_frac_to_bottom_depth
+
+        root_frac_to_top_depth = root_frac_to_bottom_depth
+        layers_to_iterate_over = soil_data.soil_layers[1:] + [soil_data.vadose_zone_layer]
+        for layer in layers_to_iterate_over:
+            root_frac_to_bottom_depth = self._calculate_root_mass_distribution(layer.bottom_depth)
+            layer_fraction = root_frac_to_bottom_depth - root_frac_to_top_depth
+            layer.plant_residue += subsurface_residue * layer_fraction
             layer.active_organic_nitrogen_content += subsurface_nitrogen * layer_fraction
             layer.labile_inorganic_phosphorus_content += subsurface_phosphorus * layer_fraction
```
allisterakun: The formulas for calculating the `plant_residue`, `fresh/active_organic_nitrogen_content`, and `labile_inorganic_phosphorus_content` follow the same pattern and are repeated three times. What do you think of extracting it into a method?

RUFAS/routines/field/soil/soil_data.py
```diff
@@ -258,19 +250,18 @@ class SoilData:
...
     plant_residue_metabolic_fraction: float = 0
-    total_residue: float = 0
-    crop_root_depth: float = 0
     crop_yield_nitrogen: float = 0
 
     @property
-    def all_residue(self) -> float:  # TODO: not currently used.
-        """amount of total plant residue, above and below-ground, on the field (kg/ha)"""
-        return self.plant_surface_residue + self.plant_root_residue
+    def all_residue(self) -> float:
```
allisterakun:
```suggestion
    def total_residue(self) -> float:
```
The word "all" can be confused with the Python keyword `all()`.

allisterakun (approved):
Looks great, Ed! Nice work! I left some minor comments hope you can take a look!

RUFAS/routines/field/soil/soil_data.py
Isgotkowitz: Good call!

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: Good suggestion! I have extracted this into a method called `_add_yield_to_layer`, testing now.

rawalashmita:
Thanks, Ed, for working on it!! I have a couple of comments. Sorry I had to wait before posting my comments on the issue. The delay had to do with a lot of things, a big one was my learning curve about model structures and PR review process itself along with other technical issues. Anyways, in the output data, based on my understanding, the amount of C, N, P from the corn grain residue in 2013 is in the ballpark. However, the amount of N should be higher on the surface layer (layer 0). However, it is seen that N distribution in the second layer (layer 1) is higher throughout the whole simulation days similar to soil organic carbon trends with depth. Similarly, labile inorganic P should be higher at surface depth and should decrease with depths. The pattern that we are observing here is that labile P is highest in layer 1, followed by layers 2 and 3, and lowest in layer 0 throughout the whole simulation days. I discussed this with other SME as well to make sure I am reading these correctly, and they had the similar outlook on the results.

I understand the beauty of simulation modeling is that you can simulate things that don't happen in real life. However, I am wondering about potential impacts on soil carbon from planting alfalfa each year since that requires soil tillage which I assume you're capturing in the model? My naive attempt to look through the codes in crop_managament.py couldn't find it though. Tillage would affect CNP distribution. Also, alfalfa has large tap roots that establish over time so planting each year (rather than treating as perennial from year 2 onward) may impact simulated results for soil C and residue amounts. But I think the code def kill(self) -> None, kills the plant, preventing it from growing further, and converts all biomass to residue. Wondering how alfalfa residue was generated if it was harvested? Understanding these might help provide better solutions to the matter as needed.

PR #1865, opened 2024-07-25
Temporary replacement of residue_composition_factor

### LINKED ISSUE #1810: Temporary Replacement of Soil N Mineralization Rate factor calculation method with fixed value
The current method to calculate the soil N mineralization rate depends on the C:N ratio of the soil and is calculated with function `calculate_decay_rate_constant` in `mineralization_decomp.py`. The SME team needs to do some more research to determine what the best way to implement the functions that support this method. While this method is being investigated by the SME team, we would like to remove the effect of the soil C:N and soil C:P ratios which is called the `residue_composition_factor`. 

## Status

- The values of the C:N and C:P ratios in layers below Layer 0 are very high (above 2,000 and up to 30,000 in some scenaorios). This causes the residue composition factor to be very small (on the order of 10^-20) which in turn makes the mineralization decay rate to be very very small as well. 

## Intention
 
We want to replace the process based method for the effect of the soil C, N, and P on the decomposition rate factor with a fixed reasonable value while we evaluate the methods that represent this effect . 

- What is the expected outcome?

A N decompositon rate that does not depend on the composition of the soil but still fluctuates with water and temperature

- Why is this change necessary?

The Soil and Crop SME team needs to be able to assess the general accuracy of the mineralization from this soil N pool while we investigate how to represent the effect of soil composition on this process

## Request

The specific requst is the following: 

- In the `mineralize_and_decompose_nitrogen` function in `mineralization_decomp.py`, replace the call to `calculate_nutrient_cycling_residue_composition_factor` with a fixed value of 1 while further evaluation and analysis is completed. 
- Please do not remove the `calculate_nutrient_cycling_residue_composition_factor` method as we hope to re-instate it soon

### Target deadline

Request to be completed by July 31st

matthew7838:
This PR removes the effect of the soil C:N and soil C:P ratios which is called the `residue_composition_factor`.
### Context

Issue(s) closed by this pull request: closes #1810

- [ ] Yes
- [x] No

### What

- The original method of `calculate_nutrient_cycling_residue_composition_factor` is kept with no quality assessment so that those temporary unused calculations won't raise flake8 errors.
- Now `calculate_nutrient_cycling_residue_composition_factor` will simply return 1 to mute its effects.
- Variable names changed to bring more clarity, `root_term` is now factor_square_root_term and `fresh_organic_residue_mineralization_rate` is now `fresh_organic_residue_decay_rate`.

### Why

We want to replace the process based method for the effect of the soil C, N, and P on the decomposition rate factor with a fixed reasonable value 1 while SMEs evaluate the methods that represent this effect.

### How

Setting the return value to a constant and clarify(simplify) the names.

### Test plan

1. Ran with updated test suits.
2. Trust our great SMEs

### DIFF
--- a/RUFAS/routines/field/soil/nitrogen_cycling/mineralization_decomp.py
+++ b/RUFAS/routines/field/soil/nitrogen_cycling/mineralization_decomp.py
@@ -166,8 +166,9 @@ class MineralizationDecomposition:
         return exp(inner_term)
 
     @staticmethod
-    def _calculate_nutrient_cycling_residue_composition_factor(carbon_nitrogen_ratio: float,
-                                                               carbon_phosphorus_ratio: float) -> float:
+    def _calculate_nutrient_cycling_residue_composition_factor(
+        carbon_nitrogen_ratio: float, carbon_phosphorus_ratio: float
+    ) -> float:
         """
         Calculates the residue composition factor for use in computing the decay rate constant.
 
@@ -191,10 +192,10 @@ class MineralizationDecomposition:
         The values of the constant used to determine the nitrogen and phosphorus terms are 25 and 200, respectively.
 
         """
-        nitrogen_term = MineralizationDecomposition._calculate_nutrient_term_for_residue_composition_factor( # noqa
+        nitrogen_term = MineralizationDecomposition._calculate_nutrient_term_for_residue_composition_factor(  # noqa
             carbon_nitrogen_ratio, 25
         )
-        phosphorus_term = MineralizationDecomposition._calculate_nutrient_term_for_residue_composition_factor( # noqa
+        phosphorus_term = MineralizationDecomposition._calculate_nutrient_term_for_residue_composition_factor(  # noqa
             carbon_phosphorus_ratio, 200
         )
         # temporary fix
--- a/RUFAS/routines/field/soil/nitrogen_cycling/mineralization_decomp.py
+++ b/RUFAS/routines/field/soil/nitrogen_cycling/mineralization_decomp.py
@@ -166,8 +166,9 @@ class MineralizationDecomposition:
         return exp(inner_term)
 
     @staticmethod
-    def _calculate_nutrient_cycling_residue_composition_factor(carbon_nitrogen_ratio: float,
-                                                               carbon_phosphorus_ratio: float) -> float:
+    def _calculate_nutrient_cycling_residue_composition_factor(
+        carbon_nitrogen_ratio: float, carbon_phosphorus_ratio: float
+    ) -> float:
         """
         Calculates the residue composition factor for use in computing the decay rate constant.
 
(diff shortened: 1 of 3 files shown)


RUFAS/routines/field/soil/nitrogen_cycling/mineralization_decomp.py
```diff
@@ -190,17 +192,18 @@ def _calculate_nutrient_cycling_residue_composition_factor(carbon_nitrogen_ratio
...
 
         """
-        nitrogen_term = MineralizationDecomposition._calculate_nutrient_term_for_residue_composition_factor(
+        nitrogen_term = MineralizationDecomposition._calculate_nutrient_term_for_residue_composition_factor(  # noqa
             carbon_nitrogen_ratio, 25
         )
-        phosphorus_term = MineralizationDecomposition._calculate_nutrient_term_for_residue_composition_factor(
+        phosphorus_term = MineralizationDecomposition._calculate_nutrient_term_for_residue_composition_factor(  # noqa
             carbon_phosphorus_ratio, 200
         )
```
Isgotkowitz: If adding a noqa is strictly necessary, then it is important to specify what error the noqa is being used to ignore. So in this case, it should be `# noqa: F841`.

RUFAS/routines/field/soil/nitrogen_cycling/mineralization_decomp.py
```diff
@@ -190,17 +192,18 @@ def _calculate_nutrient_cycling_residue_composition_factor(carbon_nitrogen_ratio
...
-        return min(nitrogen_term, phosphorus_term, 1.0)
+        # temporary value
+        return 1
```
Isgotkowitz: This comment needs more detail. What was the context for this temporary value being added? What issue has been written up to ensure that this temporary value will be replaced with a permanent solution?

Isgotkowitz (approved):
Left comments on two things that need to be updated, otherwise looks great.

> 2. Trust our great SMEs
Lol yes, my favorite test strategy.

RUFAS/routines/field/soil/nitrogen_cycling/mineralization_decomp.py
matthew7838: That's helpful, thanks.

RUFAS/routines/field/soil/nitrogen_cycling/mineralization_decomp.py
```diff
@@ -233,5 +236,5 @@ def _calculate_decay_rate_constant(
         in the SWAT Input .BSN file (see "RSDCO" on page 101) and SWAT Input CROP.DAT file (see "RSDCO_PL" on page 205).
 
         """
-        root_term = (temp_factor * water_factor) ** 0.5
-        return fresh_organic_residue_mineralization_rate * composition_factor * root_term
+        factor_square_root_term = (temp_factor * water_factor) ** 0.5
```
rawalashmita: It might be beneficial to change water factor to moisture factor

rawalashmita: It might be beneficial to change composition factor to residue composition factor just to be clear and consistent.

rawalashmita (approved):
Looks good to me. Great Job, Matthew!

KFosterReed (approved):
I compared the fresh organic N mineralization rates before and after this PR and the change accomplishes what was expected! Kudos for the additional improvements in variable naming. Well done team!

PR #1866, opened 2024-07-25
Updating animal module metadata minimums

JoeWaddell:
### Context

- [ ] Yes
- [x] No

### What

Following Sensitivity Analysis runs, some increases in the metadata minimums seem necessary to help guide users in how this can/should be used. Some were artificially low, e.g. outside of what's biologically reasonable.

### Why

Some values were encountered in sensitivity analyses that caused the simulation to crash, or other unexpected behavior to occur.

### Test plan

This needs some additional evaluation to ensure that the new suggested minimums aren't too high: notably some of the minimum birth weights, as those could be reduced still, to model exceptionally small animals...biologically possible, though not necessarily likely.


Isgotkowitz (approved):
LGTM. But could you please add a little detail to the PR description? How did you find these new minimums, how do you know they're valid, etc. They're straightforward enough that they don't need much of an explanation, but having that PR description makes it possible for someone days/months/years down the line to look at these changes in the git history, find this PR, and read through how you arrived at these minimums.

YijingGong (commented):
Hi Joe, I think it looks good to me. But similar to Ed, I am wondering the source of the minimum values. I think the 1 is intuitive - to make sure the animals have the lactation process and weaning process. But what about the 8 kg of calving weights?
Is that associated with the sensitivity analysis, you found out the >8kg of calving weight do not crash the simulation?

JoeWaddell:
> But what about the 8 kg of calving weights?
> Is that associated with the sensitivity analysis, you found out the >8kg of calving weight do not crash the simulation?

That choice was a little arbitrary: it was meant to align with the default standard deviation, so it wouldn't produce errors (as seen in https://github.com/RuminantFarmSystems/MASM/pull/1887).

That said, if you think it's best, we can set that minimum to 1 so that it is in line with other minimums outside of what is biologically possible/likely, and the tagged PR #1887 would help alleviate potential errors if someone sets the average birth weight lower than the default standard deviation.

YijingGong:
> > But what about the 8 kg of calving weights?
> > Is that associated with the sensitivity analysis, you found out the >8kg of calving weight do not crash the simulation?
> 
> That choice was a little arbitrary: it was meant to align with the default standard deviation, so it wouldn't produce errors (as seen in #1887).
> 
> That said, if you think it's best, we can set that minimum to 1 so that it is in line with other minimums outside of what is biologically possible/likely, and the tagged PR #1887 would help alleviate potential errors if someone sets the average birth weight lower than the default standard deviation.

I don't think you will like it if many of the simulations produce error messages and stops, and it's hard to analyze that way. So I'm trying to think if there are ways to do the SA without setting the metadata minimum input as 8. Do you think it would make more sense 1) to set the minimum bound of calving birth as 8 kg for the SA problem, or 2) to modify in the code to treat negative birth weight as 1kg?

JoeWaddell:
> I don't think you will like it if many of the simulations produce error messages and stops, and it's hard to analyze that way. So I'm trying to think if there are ways to do the SA without setting the metadata minimum input as 8. Do you think it would make more sense 1) to set the minimum bound of calving birth as 8 kg for the SA problem, or 2) to modify in the code to treat negative birth weight as 1kg?

1) There are many places in which the sensitivity analysis setup requires some careful thought without setting a minimum, and I've started collecting those in a document that will be useful for eventual cross-validation, as well: https://docs.google.com/document/d/1rgybPfwpjjVweUVThyE1CJKReKtCdP8WvOP9s75dtMc/edit?usp=sharing

2) So in https://github.com/RuminantFarmSystems/MASM/pull/1887, I outlined precisely that as a potential alternative/addition to the error message! The change here was to raise it from the current minimum of 0, so I can reduce it from 8 to 1, if you think that's a better/less arbitrary decision. The change in 1887 (however we decide to resolve it) should handle the cases in which the user puts too high of a SD relative to the AVG.

Thanks for giving this some good thought!

YijingGong:
Sounds good to me then! I would think setting it to 1 makes it look less magical, and later limiting them in SA bounds if necessary.

YijingGong (approved):

PR #1870, opened 2024-07-29
Partition denitrified nitrates between N2O and N2

### LINKED ISSUE #1712: Field Nitrous Oxide emission estimates incorrectly attribute all denitrified nitrogen to Nitrous Oxide
The process of denitrification in the soil produces both nitrous oxide (a greenhouse gas) and nitrogen gas (also known as N_2, which does not contribute to the greenhouse effect). Right now we incorrectly attribute all of the denitrified soil nitrogen to nitrous oxide emission so we need to add a method to:

1. Estimate the ratio of denitrified nitrogen that is emitted as nitrous oxide. 
2. Apply that fraction to the methods that predict and report nitrous oxide emissions from the soil

## Expected Behavior

Nitrous oxide should be a variable fraction of denitrified nitrogen that can range from (roughly) 0.05 to 0.80. This fraction is dependent on 3 things: 

- Soil nitrate content (kg/ha)
- Soil pH
- and total soil carbon content (kg/ha)

## Current Behavior

All denitrified nitrogen is attributed to nitrous oxide emission

## Possible Solution

- Create a function to estimate the ratio of denitrified nitrogen that is partitioned to nitrous oxide within `denitrification.py` following the equations/logic to be provided in more detail in a google doc soon. 
- When defining the `layer.nitrous_oxide_emissions` soil layer attribute in the `denitrify()` method of the nitrogen_cycling directory, set it equal to `nitrified_nitrates/(1+nitrous_oxide_ratio)`
- In addition, the above 'nitrified_nitrates' is misnamed. References in variables and documentation notes for 'dentrification.py' should be 'denitrification' or 'denitrified'.

@KevinP-B or @KFosterReed will provide a more detailed description of the method for estimating the ration via a google doc linked to this issue soon but for now, a screenshot of the SWAT documentation is provided below. 

## Context (Environment)
We are currently over predicting nitrous oxide emissions by quite  a lot. This is the largest source of greenhouse gases in the field and it is important that we estimate this correctly. 

## Detailed Description

Proposed solution details can be found in the google doc
https://docs.google.com/document/d/1HWEU_hv2UPsiupqZjc_SkfYCExFuALZ7qVyZnlKa2Lc/edit?usp=sharing
## Target Deadline

This is an urgent task. May 31 deadline is very important to acheive.

Isgotkowitz:
Divides denitrified nitrates into two types of emissions, $N_2O$ and $N_2$, instead of having all denitrified nitrates leave the system as $N_2O$.

### Context

Issue(s) closed by this pull request: closes #1712

Note: the Mypy error count is increased, but this is due to a lot of type hints in `SoilData` and `LayerData` being incorrect/inappropriate. I felt that it would be better to leave them alone for now instead of adding more to this PR.

### What

Same as title.

### Why

By not accounting at all for $N_2$ emissions, $N_2O$ is systematically overestimated in RuFaS. Because $N_2O$ is such an important GHG, it is important we are as accurate as possible in our estimations of it.

### How

Modifies the `Denitrification` class to use routines from the SWAT-GHG add on to partition denitrified nitrates between $N_2O$ and $N_2$. This partitioning depends on several factors, soil nitrate content, carbon respiration, water content and pH. To account for pH, a new (non-essential) input has been added to the soil input file.

### Test plan

Updated the denitrification test suite to make sure that it properly calculates and utilizes the new routines.

Also, ran the default scenario on both this branch and `dev` with the below filter. I checked that the mass balance was preserved (i.e. $N_2O$ + $N_2$ on this branch = $N_2O$ on `dev`).
```json
{
    "multiple": [
        {
            "name": "_",
            "filters": ["(?!.*annual)Field.*nitrous_oxide_emissions.*"]
        },
        {
            "name": "_",
            "filters": [".*dinitrogen_emissions.*"]
        }
    ]
}
```


rawalashmita:
partitioning_factor = min(nitrate_effect, carbon_effect) * moisture_effect * pH_effect
The first part of the RN2/N2O equation is taking the minimum value for nitrate-N flux. If it is indeed doing this, is it possible it's using zero since the minimum value for nitrate-N flux is likely zero on at least one day of the simulation. Then as us say if it's zero it causes the function (RN2/N2O) to be zero. Just a thought. If it is the case, could we assign zero values some nominal value like the nitrate-N concentration detection limit or something that makes sense to do just so it's not zero and then we can back that out for annual fluxes. 

Secondly, I would also look into the units. 
Fr(NO3) is the effect of soil NO3 (kg/ha) and is equal to---(in the documentation)
nitrate_content_grams = nitrate_content * GeneralConstants.KG_TO_GRAMS (In the codes/model)
        fractional_term = atan(pi * 0.01 * (nitrate_content_grams - 190)) / pi
        return 1 - (0.5 + fractional_term) * 25
When you change kg into grams and keeping all other components same we might not obtain the correct numbers. Can we look into it?

KFosterReed:
Hey team! I was going to review this but there are over 1,500 files changed so I'm not sure where to start... seems like something went awry somewhere? @Isgotkowitz any thoughts on this?

Isgotkowitz:
> Hey team! I was going to review this but there are over 1,500 files changed so I'm not sure where to start... seems like something went awry somewhere? @Isgotkowitz any thoughts on this?

Yes, something did go awry. I will fix it right now, sorry about this Kristan. Edit: fixed as far as I can tell.

Isgotkowitz:
> Secondly, I would also look into the units. Fr(NO3) is the effect of soil NO3 (kg/ha) and is equal to---(in the documentation) nitrate_content_grams = nitrate_content * GeneralConstants.KG_TO_GRAMS (In the codes/model) fractional_term = atan(pi * 0.01 * (nitrate_content_grams - 190)) / pi return 1 - (0.5 + fractional_term) * 25 When you change kg into grams and keeping all other components same we might not obtain the correct numbers. Can we look into it?

In the [SWAT-GHG routines paper](https://www.sciencedirect.com/science/article/pii/S136481521630929X), $FR(NO_3)$ is defined by equation (8), and the equation in the code is exactly the same as (8). That paper doesn't state the units of $FR(NO_3)$ super explicitly. Under equation (4) it states "where $Fd(NO_3)$ and $Fd(C)$ have units of g N ha−1 d−1", and that was the closest I could find for an explanation of the units of $NO_3$ - so I assume they are working with grams per ha in that paper.

I suspect the real units issue is because of where/how they drew that equation from [Parton et al, 1996](https://agupubs.onlinelibrary.wiley.com/doi/10.1029/96GB01455). That paper has exactly the same equation as the SWAT paper, but Parton's units are micrograms of N per gram (see figure 5 b).

Isgotkowitz:
Gonna move this to blocked - until that nitrate equation gets figured out, there isn't a way to move this PR forward. Edit: I mean issue.

RUFAS/routines/field/soil/nitrogen_cycling/denitrification.py
```diff
@@ -48,22 +48,35 @@ def denitrify(self) -> None:
...
                 layer.nitrate_content,
                 layer.denitrification_rate_coefficient,
                 layer.nutrient_cycling_temp_factor,
                 layer.soil_overall_carbon_fraction,
             )
-            layer.nitrate_content -= nitrified_nitrates
-            layer.nitrous_oxide_emissions = nitrified_nitrates
-            layer.annual_nitrous_oxide_emissions_total += nitrified_nitrates
+
+            nitrate_effect = self._calculate_nitrate_effect(layer.nitrate_content)
```
KevinP-B: I'm not sure of the code-specific naming conventions, but it might be worth considering explicitly stating that these effects are on denitrified N partitioning, i.e. nitrate_denitrification_partitioning_effect. I defer to those who code to find the optima between intuitive naming and unwieldly variable names.

RUFAS/routines/field/soil/nitrogen_cycling/denitrification.py
```diff
@@ -113,3 +126,174 @@ def _calculate_denitrification_amount(
...
+        References
+        ----------
+        .. [1] Parton, W. J., et al. "Generalized model for N2 and N2O production from nitrification and
+           denitrification." Global biogeochemical cycles 10.3 (1996): 401-412.
+
+        """
+        nitrate_content_grams = nitrate_content * GeneralConstants.KG_TO_GRAMS
+        fractional_term = atan(pi * 0.01 * (nitrate_content_grams - 190)) / pi
+
+        return (0.5 - fractional_term) * 25
```
KevinP-B: Even if inefficient, I'd be more comfortable expressing this explicitly as it appears in parton 1996 to ensure we have an easier time tracking the provenance of the error in this equation, ie y=(1-(0.5+fractional_term))*25.

RUFAS/routines/field/soil/nitrogen_cycling/denitrification.py
```diff
@@ -113,3 +126,174 @@ def _calculate_denitrification_amount(
...
+        float
+            Amount of nitrous oxide emissions (kg / ha).
+
+        References
+        ----------
+        .. Wagena, Moges B., et al. "Development of a nitrous oxide routine for the SWAT model to assess greenhouse gas
+           emissions from agroecosystems." Environmental modelling & software 89 (2017): 131-143.
+
+        """
+        nitrates = denitrified_nitrates * GeneralConstants.KG_TO_GRAMS
```
KevinP-B, line 314: I believe the n2o equation should be the total denitrified nitrogen (what we used to call n2o emissions) over 1+partitioning_factor.

KevinP-B (commented):
Several comments on naming and equation accuracy to be addressed. Happy to chat specifics.

RUFAS/routines/field/soil/nitrogen_cycling/denitrification.py
Isgotkowitz: > I defer to those who code to find the optima between intuitive naming and unwieldly variable names.

Me too

RUFAS/routines/field/soil/nitrogen_cycling/denitrification.py
Isgotkowitz: Good call

RUFAS/routines/field/soil/nitrogen_cycling/denitrification.py
Isgotkowitz: Can you please expand in this? I am pretty sure this is what the code is doing, `denitrified_nitrates` is the total amount of N that RuFaS denitrifies.

RUFAS/routines/field/soil/nitrogen_cycling/denitrification.py
```diff
@@ -113,3 +131,174 @@ def _calculate_denitrification_amount(
         denitrification_factor = 1 - exponential_term
         bounded_denitrification_factor = max(min(1.0, denitrification_factor), 0.0)
         return nitrate_content * bounded_denitrification_factor
+
+    def _calculate_nitrate_effect(self, nitrate_content: float) -> float:
```
allisterakun, line 148: Given that the method above `_calculate_denitrification_amount()` is a static method, what do you think of making all of the new methods static for consistency?

allisterakun (approved):
LGTM! Just one minor comment of potentially making the new methods static

RUFAS/routines/field/soil/nitrogen_cycling/denitrification.py
Isgotkowitz: Ideally it would actually be the other way around, where all the subroutines in this method are instance methods instead of all of them being static methods. As I understand it, static methods have a slight performance penalty that instance methods do not have so they should not be used unless they are needed in a very specific case (like `task` in Task Manager).

I like your point about consistency, it would be best to have no more static methods in this module. However for the sake of keeping this PR focused and reasonably sized, I will not convert them here. However, I think we should include converting static methods to instance methods as part of the improvements that we make to base modules in the Crop and Soil module.

RUFAS/routines/field/soil/nitrogen_cycling/denitrification.py
```diff
@@ -113,3 +131,174 @@ def _calculate_denitrification_amount(
...
+        References
+        ----------
+        .. [1] Parton, W. J., et al. "Generalized model for N2 and N2O production from nitrification and
+           denitrification." Global biogeochemical cycles 10.3 (1996): 401-412.
+
+        """
+        nitrate_content_grams = nitrate_content * GeneralConstants.KG_TO_GRAMS
+        fractional_term = atan(pi * 0.01 * (nitrate_content_grams - 190)) / pi
+
+        return (1.0 - (0.5 + fractional_term)) * 25
```
KFosterReed, line 176: Looking at the equation from the literature/documentation in the issue, it seems like there is an extra parentheses and this should be:

`1.0 - (0.5 + fractional_term) * 25`

@KevinP-B , do you agree?

RUFAS/routines/field/soil/nitrogen_cycling/denitrification.py
```diff
@@ -113,3 +131,174 @@ def _calculate_denitrification_amount(
...
+
+    def _calculate_carbon_effect(self, carbon_respiration: float) -> float:
```
KFosterReed, line 178: Is the carbon input for this supposed to be the carbon respiration/emissions or the total soil carbon content?

RUFAS/routines/field/soil/nitrogen_cycling/denitrification.py
KFosterReed: Ok, nevermind, after some more testing and plotting of the equation itself, I see that this was a typo and it should be as it is `(1.0 - (0.5 + fractional_term)) * 25`. However, looking at the units of the original Parton paper (ug NO3/g) with values that range from 0-350 ug/g, I am guessing that we are not putting in the correct units for Nitrate content? The values for the nitrate content inputs in the simulation I ran ranged from 50-250 kg/ha which are much larger than the range of ug/g in the Parton paper

RUFAS/routines/field/soil/nitrogen_cycling/denitrification.py
KFosterReed: Ignore this comment as well. I see in the Parton paper that it should be C respiration!

RUFAS/routines/field/soil/nitrogen_cycling/denitrification.py
```diff
@@ -48,22 +54,47 @@ def denitrify(self) -> None:
...
             )
-            layer.nitrate_content -= nitrified_nitrates
-            layer.nitrous_oxide_emissions = nitrified_nitrates
-            layer.annual_nitrous_oxide_emissions_total += nitrified_nitrates
+
+            nitrate_concentration_mg_per_kg = LayerData.determine_soil_nutrient_concentration(
+                layer.nitrate_content, layer.bulk_density, layer.layer_thickness, field_size
+            )
+
+            # Milligrams per kilogram is equivalent to micrograms per gram
```
Isgotkowitz, line 76: I want this to be explicitly noted because I can imagine someone getting really tripped up on the units here before understanding this, but I am not sure an inline comment is the best way to go about noting this.

KFosterReed (approved):
This is a HUGE improvement team! The partitioning factor is now in the range of the expected values which leads to N2O emissions that are betwen 20-85% of the total denitrified nitrogen and are much much much closer to the expected emissions levels.

Great work team!

RUFAS/routines/field/soil/nitrogen_cycling/denitrification.py
Isgotkowitz: This equation has been modified so that it works in micrograms N per gram of soil, as specified in the literature.

PR #1872, opened 2024-07-29
ration formulation error messages

### LINKED ISSUE #1859: Add two errors to ration_driver.py
There are multiple ways in which ration formulation can fail.

One is that lactation could be reduced to the point that animals are no longer lactating. This lower limit is set for the user defined ration, but not yet in place for the automated formulation.

For non lactating cows: we have no way to reduce the requirements, and so reattempts are impossible. Right now, automated formulation falls back on the previous ration. However, if there is no previous ration, we need to handle this. #577 

This is currently in #1835, but will remove it such that it is siloed in its own separate PR, since it's outside the original scope of that PR.

For the case where lactation is reduced beyond a minimum: we should add logic to raise an error if lactation would be dropped below 0 (or some other more biologically reasonable minimum).

For cases where we are missing a previous ration: something like the below should work.
image

Alternatively something like:
```
        if solution is None:
            if pen.ration:
                return pen.ration, ration_vals
            else:
                om.add_error(
                    "First ration formulation error",
                    "Unable to formulate ration, and there is no previous ration to use.",
                    info_map,
                )
                raise
```

Note: in any case, this needs to apply to both the automated ration formulation process (the bulk of `formulate_ration` in ration_driver.py), and also the user-defined ration formulation (`get_used_defined_ration`)

JoeWaddell:
### Context

Issue(s) closed by this pull request: closes #1859

Note that this is merging to main, in case it's easier for FARM ES to merge in from there.

- [ ] Yes
- [X] No

### What

Logic to handle ration formulation errors that can't be resolved.

### Why

Necessary to stop simulation when ration formulation is impossible. 

### Test plan

Tested against rations that will never succeed. Raised and error and stopped while loop. New animal module constant created, so user has direct control over what the allowable minimum average milk production is for a pen.

### DIFF
--- a/RUFAS/routines/animal/ration/ration_driver.py
+++ b/RUFAS/routines/animal/ration/ration_driver.py
@@ -93,7 +93,7 @@ class RationManager:
                     "function": cls.formulate_ration.__name__,
                 }
                 num_reattempts += 1
-                if num_reattempts > 40:
+                if pen.avg_milk < 20 or num_reattempts > 100:
                     om.add_error(
                         "Ration formulation error.",
                         "Catastrophic ration formulation error: can't formulate, too many formulation attempts."
--- a/RUFAS/routines/animal/ration/ration_driver.py
+++ b/RUFAS/routines/animal/ration/ration_driver.py
@@ -94,10 +94,12 @@ class RationManager:
                 }
                 num_reattempts += 1
                 if num_reattempts > 40:
-                    om.add_error("Ration formulation error.",
-                                 "Catastrophic ration formulation error: can't formulate, too many formulation attempts."
-                                 + f" Check failed_constraint_summary_for_pen_{pen.id}",
-                                 info_map)
+                    om.add_error(
+                        "Ration formulation error.",
+                        "Catastrophic ration formulation error: can't formulate, too many formulation attempts."
+                        + f" Check failed_constraint_summary_for_pen_{pen.id}",
+                        info_map,
+                    )
                     raise
                 constraints_failed_list = []
                 failed_constraints = ration_optimizer.find_failed_constraints(
@@ -156,10 +158,12 @@ class RationManager:
         elif pen.ration != {}:
             return pen.ration, ration_vals
         else:
-            om.add_error("Ration formulation error.",
-                         "Catastrophic ration formulation error: can't formulate, no previous ration available."
-                         + f" Check failed_constraint_summary_for_pen_{pen.id}",
-                         info_map)
+            om.add_error(
+                "Ration formulation error.",
+                "Catastrophic ration formulation error: can't formulate, no previous ration available."
+                + f" Check failed_constraint_summary_for_pen_{pen.id}",
+                info_map,
+            )
             raise
 
     @staticmethod
--- a/RUFAS/routines/animal/ration/ration_driver.py
+++ b/RUFAS/routines/animal/ration/ration_driver.py
@@ -83,24 +83,24 @@ class RationManager:
         )
         # Reduction of milk production estimate process to achieve feasible solution
         num_reattempts: int = 0
+        info_map = {
+            "class": "RationManager",
+            "function": cls.formulate_ration.__name__,
+        }
 
         # TODO: Put AnimalCombination enum in a separate file and use it here instead of hardcoding the names
         # GitHub Issue #793
         if pen.animal_combination.name in ["LAC_COW"]:
             while not solution.success:
-                info_map = {
-                    "class": "RationManager",
-                    "function": cls.formulate_ration.__name__,
-                }
                 num_reattempts += 1
-                if pen.avg_milk < 20 or num_reattempts > 100:
+                if pen.avg_milk < 15:
                     om.add_error(
                         "Ration formulation error.",
                         "Catastrophic ration formulation error: can't formulate, too many formulation attempts."
                         + f" Check failed_constraint_summary_for_pen_{pen.id}",
                         info_map,
                     )
-                    raise
+                    raise RuntimeError
                 constraints_failed_list = []
                 failed_constraints = ration_optimizer.find_failed_constraints(
                     solution.x, ration_optimizer.cow_constraints, ration_config
@@ -160,11 +160,11 @@ class RationManager:
         else:
             om.add_error(
                 "Ration formulation error.",
-                "Catastrophic ration formulation error: can't formulate, no previous ration available."
+                "Catastrophic ration formulation error: can't formulate; no previous ration available."
                 + f" Check failed_constraint_summary_for_pen_{pen.id}",
                 info_map,
             )
-            raise
+            raise RuntimeError
 
     @staticmethod
     def calc_milk_average(pen: Pen) -> float:
--- a/RUFAS/routines/animal/ration/ration_driver.py
+++ b/RUFAS/routines/animal/ration/ration_driver.py
@@ -88,7 +88,17 @@ class RationManager:
         # GitHub Issue #793
         if pen.animal_combination.name in ["LAC_COW"]:
             while not solution.success:
+                info_map = {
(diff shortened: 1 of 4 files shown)


krb78 (approved):
Catastrophic! Haha, very convincing Joe. It looks good and appears to function properly :) Is it too much for us to add advice to errors like this? I know when I get errors for the first time, I don't quite have my head wrapped around why. What do you think about having an error FYI on the wiki or something? Is that a crazy idea? That way users can search their error and say "oh, looks like I should try to add an additional feed ingredient" or "the problem is my weather file had weather that was too cold to grow a crop".

JoeWaddell:
> Catastrophic! Haha, very convincing Joe. It looks good and appears to function properly :) Is it too much for us to add advice to errors like this? I know when I get errors for the first time, I don't quite have my head wrapped around why. What do you think about having an error FYI on the wiki or something? Is that a crazy idea? That way users can search their error and say "oh, looks like I should try to add an additional feed ingredient" or "the problem is my weather file had weather that was too cold to grow a crop".

I agree! The more detail and suggestions for error avoidance, the better. I also think these kinds of issues that people often run into (for example with ration formulation: feed prices, ingredient lists being too small, etc), that a FAQ on the wiki would be great!

So in short: I'll make the error message longer, to let the user know that (for both error cases) they have too restrictive of a feed list for the animals in the failing pen.

ew3361zh:
Do you have some example rations that will never succeed that I can use for testing?

JoeWaddell:
> Do you have some example rations that will never succeed that I can use for testing?

The simplest way is to remove feed items from the `input/data/feed/default_feed.json` in some groups, such that they only have a couple random items available to attempt ration formulation. That should for sure make it fail!

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -145,13 +159,22 @@ def formulate_ration(
                     dict(info_map, **{"units": fail_summary_units}),
                 )
 
-        if solution is not None:
+        if solution.success and solution is not None:
```
ew3361zh:
```suggestion
        if solution is not None and solution.success:
```

To avoid a potential error of checking for an attribute first if solution is None.

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -88,12 +88,26 @@ def formulate_ration(
...
 
         # TODO: Put AnimalCombination enum in a separate file and use it here instead of hardcoding the names
         # GitHub Issue #793
         if pen.animal_combination.name in ["LAC_COW"]:
             while not solution.success:
                 num_reattempts += 1
+                if pen.avg_milk < AnimalModuleConstants.MINIMUM_AVG_PEN_MILK:
+                    om.add_error(
+                        "Ration formulation error",
+                        "Catastrophic ration formulation error: can't formulate, too many formulation attempts."
```
ew3361zh: Is this the correct message? Maybe I'm misunderstanding the influence of the # of attempts on line 101 but  should it instead be specific to the ration formulation causing the `pen.avg_milk` to go below the `MINIMUM_AVG_PEN_MILK`?

JoeWaddell: Good call: it's technically correct (since repeated attempts will drop the average milk production), but I'll update the message to something easier to interpret.

ew3361zh (approved):
This looks good and caught the error in a helpful and informative way (though it was suprisingly resilient - I had to go through a couple rounds of removing random feeds!).

I left a couple suggestions/small feedback but my remaining question is around the use of "catastrophic" which does come off as pretty extreme. To me it reads like it will be a massive amount of work to fix from the user end (even despite the specific instructions) and might dissuade me from trying to address it.

Maybe would suggest "critical" or something slightly less alarming instead?

Also, I'm not totally clear on whether the changes to all the data and metadata files is intentional and fully vetted. But obviously please be sure those are supposed to be included with the rest of the changes here.

Nice work, Joe!

JoeWaddell:
> the use of "catastrophic" which does come off as pretty extreme. To me it reads like it will be a massive amount of work to fix from the user end (even despite the specific instructions) and might dissuade me from trying to address it.
> 
> Maybe would suggest "critical" or something slightly less alarming instead?

Good suggestion: I agree something like "critical" would be better.

> Also, I'm not totally clear on whether the changes to all the data and metadata files is intentional and fully vetted. But obviously please be sure those are supposed to be included with the rest of the changes here.

They are not! Some sort of git error - it's cropped up on another PR as well, so I will resolve that prior to merge.

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -88,12 +88,26 @@ def formulate_ration(
...
                 num_reattempts += 1
+                if pen.avg_milk < AnimalModuleConstants.MINIMUM_AVG_PEN_MILK:
+                    om.add_error(
+                        "Ration formulation error",
+                        "Critical ration formulation error: milk production has dropped too low."
+                        + f" Check failed_constraint_summary_for_pen_{pen.id} to see what caused formulation to fail."
+                        + " Possible solution is to provide additional feed ingredients to"
+                        + f" {pen.animal_combination.name}.",
+                        info_map,
+                    )
```
PooyaHekmati: Not a big deal for this PR as we are not going to hit this string multiple times a run; but:
1. Generally speaking, it is better to break an f-string across multiple lines like this rather using `+` to concatenate.
2. Also, it is better to leave the white space at the end of the last line instead of the beginning of the next line.
3. The name `"Ration formulation error"` is redundant as we already know the class and function names and we know this is an error (we are adding an error after all). We can increase the information density by shuffling the words like this.
```suggestion
                    om.add_error(
                        "milk production too low",
                        (
                           f"Check failed_constraint_summary_for_pen_{pen.id} to see what caused formulation to fail. "
                           f"Possible solution is to provide additional feed ingredients to {pen.animal_combination.name}."
                         ),
                        info_map,
                    )
```

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -88,12 +88,26 @@ def formulate_ration(
...
+                    raise RuntimeError
```
PooyaHekmati: Why is this a `RuntimeError` and not a `ValueError`?
[ValueError](https://docs.python.org/3/library/exceptions.html#ValueError)

RUFAS/routines/animal/ration/ration_driver.py
```diff
@@ -145,13 +155,22 @@ def formulate_ration(
...
             return pen.ration, ration_vals
+        else:
+            om.add_error(
+                "Ration formulation error",
+                "Critical ration formulation error: can't formulate; no previous ration available."
+                + f" Check failed_constraint_summary_for_pen_{pen.id} to see what caused formulation to fail."
+                + f" Possible solution is to provide additional feed ingredients to {pen.animal_combination.name}.",
+                info_map,
+            )
+            raise RuntimeError
```
PooyaHekmati: Same as above

RUFAS/routines/animal/ration/ration_driver.py
JoeWaddell: Excellent point: I think it could work as a ValueError, since it's technically valid, but outside what we think is reasonable.

RUFAS/routines/animal/ration/ration_driver.py
JoeWaddell: In this case I can also see the logic in considering it a ValueError, since it's missing a value (`pen.ration`)

RUFAS/routines/animal/ration/ration_driver.py
JoeWaddell: Great suggestion, point well taken! Updated accordingly.

PooyaHekmati (approved):

PR #1863, opened 2024-07-25
Debug out of range error for c2f7 inputs

matthew7838:
Fixing list index out of range for a specific set of inputs.
### Context

Issue(s) closed by this pull request: closes #1564

- [x] The [changelog](https://docs.google.com/spreadsheets/d/1U-3igxdstHTFb6k5dVcE5-x9IL0r6q18eUiwwytKZyA/edit#gid=0) has been updated.
      
How big are the changes in the PR? Would you nominate it for a major version upgrade?
- [ ] Yes
- [x] No

### What

- `RationManager.formulate_ration` and `get_user_defined_ration` will now take an argument sim_day
- `Animall_manager` will pass in the simulation day
- The simulation day will be used directly, instead of retrieving from body_weight_history.

### Why

This error will completely stop the simulation run which is troublesome for pilot testing.

### How

Instead of retrieving simulation days from body weight history, we will pass int the simulation day from the animal manager that's calling the method.

### Test plan

- Run with default inputs and it's still working.
- **Test the changes on [PR#1853](https://github.com/RuminantFarmSystems/MASM/pull/1853) but review this branch**

PR #1875, opened 2024-07-30
1863 Replica

PooyaHekmati:
This PR is a replica of PR #1863. PR 1863 was merged into `dev` branch, this PR merges into `main`. The reason for this duplication is that we need the changes from 1863 in `main`, but we don't want to bring other changes with it.


KFosterReed (approved):
I tested this with the FARM-ES set of input files and the out of range index error is fixed and no other sneaky input / ouput changes! 

Thanks Pooya!

PR #1883, opened 2024-08-02
Reporting failed tasks in TM

### LINKED ISSUE #1856: Report generator crashes when given too many Nones
image

Running the `sobol` branch using the fractional factorial method (@JoeWaddell may have local changes that mean it won't replicate the error - before tackling this issue, have me sync).

## Current Behavior

Simulation claims to have run successfully, but doesn't! Using the debugger, it seems to be a None type creeping in to the `data` field(s). 

## Possible Solution

The attempts I made so far were just making the existing code less elegant: I think it's upstream of this method, as it seems instead of an empty list, it's getting a None type object instead.

## Steps to Reproduce

1.
2.
3.
4.

matthew7838:
This issue has variated a lot from the planning. Turns out the error did get caught. This PR is no longer related to the original issue #1856 and focuses on enhancing task manager error reports.
### Context

Issue(s) closed by this pull request: closes #1856 (Already closed)

- [ ] Yes
- [x] No

### What

- `Task()` will now return the task ID when an error occurs.
- `_run_tasks` will retrieve the returns and log the details of the totaled failed tasks in the simulation during the multiprocessing.

### Why

Ensures that an easier way to identify the failed tasks.

### How

Logging the total failed tasks' task_id to the OM.

### Test plan

There's a line of commented `raise ValueError` in TM, activate that line, run with default and check the log.


RUFAS/task_manager.py
```diff
@@ -391,14 +398,15 @@ def task(args: Dict[str, Any], produce_graphics: bool, metadata_depth_limit: int
         except Exception as e:
             info_map.update(args)
             output_manager.add_error(
-                "Failed to finish the task",
+                "Failed to finish the task, continuing to the next task",
```
Isgotkowitz: I think this error name should be left as-is. If there are more tasks waiting to be started then it is true, but if all the other tasks are already finished or have already been started then this message is false - and `task` only operates on a single task so it doesn't have any idea which case this is.

RUFAS/task_manager.py
```diff
@@ -391,14 +398,15 @@ def task(args: Dict[str, Any], produce_graphics: bool, metadata_depth_limit: int
...
                 f"Failed to recover from error: {e}; traceback: {traceback.format_exc()}",
                 info_map,
             )
             output_manager.dump_all_nondata_pools(args["logs_directory"], args["exclude_info_maps"], "block")
             output_manager.add_log(
                 "Early termination", "Unexpected early termination. Please see logs for details.", info_map
             )
+            return task_id
```
Isgotkowitz: Instead of returning the `task_id`, I think it would be better to return the `output_prefix` - or possibly both. The task ID labels the task out of the total number of tasks that will be run ("1/1", "2/10", "3/100" etc.) Whereas the output prefix will be more descriptive (and more static too, I believe) - this is "default" or "animal_sensitivity_analysis" or whatever the user defines in the tasks file.

RUFAS/task_manager.py
```diff
@@ -317,7 +324,7 @@ def call_handler(
         handler(args, input_manager, output_manager, task_id, produce_graphics)
 
     @staticmethod
-    def task(args: Dict[str, Any], produce_graphics: bool, metadata_depth_limit: int | None) -> None:
+    def task(args: Dict[str, Any], produce_graphics: bool, metadata_depth_limit: int | None) -> Any:
```
Isgotkowitz: The return type should be `str` - the type of the task_id (and output_prefix, if you decide to go that route) will always be strings.

RUFAS/task_manager.py
```diff
@@ -301,8 +301,15 @@ def _run_tasks(
             self.task, produce_graphics=produce_graphics, metadata_depth_limit=metadata_depth_limit
         )
         results = self.pool.imap(task_with_args, single_run_args)
+        failed = []
         for _ in results:
```
Isgotkowitz:
```suggestion
        for result in results:
```

RUFAS/task_manager.py
```diff
@@ -301,8 +301,15 @@ def _run_tasks(
...
-            pass
+            if _ is not None:
+                failed.append(_)
+
+        if len(failed) > 0:
+            info_map = {"class": TaskManager.__name__, "function": TaskManager._run_tasks.__name__}
+            om = OutputManager()
+            om.add_log("Not all tasks are successful", f"There are tasks failed, the tasks are:{failed}", info_map)
```
Isgotkowitz:
```suggestion
            om.add_log("Not all tasks are successful", f"There are tasks failed, the tasks are: {failed}", info_map)
```
I suggest "Task(s) failed", "Failed tasks are: {failed}" - I think this would be a bit more readable and a little less redundant. Also, I think it would be better to have this be an error rather than a log.

RUFAS/task_manager.py
```diff
@@ -301,8 +301,15 @@ def _run_tasks(
             self.task, produce_graphics=produce_graphics, metadata_depth_limit=metadata_depth_limit
         )
         results = self.pool.imap(task_with_args, single_run_args)
+        failed = []
         for _ in results:
-            pass
+            if _ is not None:
+                failed.append(_)
```
Isgotkowitz: Maybe log the task as soon as it fails too, instead of waiting for all tasks to finish then writing the logs.

RUFAS/task_manager.py
```diff
@@ -391,14 +398,15 @@ def task(args: Dict[str, Any], produce_graphics: bool, metadata_depth_limit: int
         except Exception as e:
             info_map.update(args)
             output_manager.add_error(
-                "Failed to finish the task",
+                "Failed to finish the task, continuing to the next task",
                 f"Failed to recover from error: {e}; traceback: {traceback.format_exc()}",
                 info_map,
             )
             output_manager.dump_all_nondata_pools(args["logs_directory"], args["exclude_info_maps"], "block")
```
Isgotkowitz, line 405: This call to `dump_all_nondata_pools` often crashes for me - which is a problem because if this crashes here it won't return the task_id, and the crashed simulation won't be properly handled. If you want to try and fix this in this PR it would be greatly appreciated, but if not let me know and I will write up an issue for this.

Here is how the code crashes for me when running the default task:
image

RUFAS/task_manager.py
```diff
@@ -567,9 +575,12 @@ def _handle_herd_init_tasks(
...
         """Handler for all methods related to herd initialization."""
-        args["init_herd"] = True
-        TaskManager.handle_herd_initializaition(args, output_manager)
-        TaskManager.handle_post_processing(args, input_manager, output_manager, task_id)
+        try:
+            args["init_herd"] = True
+            TaskManager.handle_herd_initializaition(args, output_manager)
+            TaskManager.handle_post_processing(args, input_manager, output_manager, task_id)
+        except Exception:
+            pass
```
Isgotkowitz: Why does this need to be wrapped in a try-except block? If an exception occurs during the herd initialization process, does it need to be handled explicitly?

Isgotkowitz (commented):
This is an excellent start to better error handling, Matthew. I found a few small things that I think can be improved.

Also, there is a method call in the except block of `task` that often crashes on me and prevents the code from benefiting from the error handling you've introduced here. I think we should figure out a game plan for how that will addressed before this gets merged.

RUFAS/task_manager.py
matthew7838: Good point. The problem is that sometimes the next task takes a long time or the next message pops up and does not make sequential sense due to imap multi-processing, making it stuck in the state with this message for a few minutes(which confused me in the beginning). Are there any ways to better describe this and let the user know the WHOLE thing is still running?

RUFAS/task_manager.py
matthew7838: I can look into it but I think this should be a new issue. This PR has altered so much already.

RUFAS/task_manager.py
matthew7838: Nice catch, this is early stage changes that we no longer need. Thanks.

RUFAS/task_manager.py
matthew7838: There are other empty returns, which should be kept empty for now.

RUFAS/task_manager.py
matthew7838: Will output_prefix be the same for every task?

Isgotkowitz: It might be, but it is the user's responsibility to make sure they have the tasks named appropriately. And I think most users will use different output prefixes for the tasks. If it is a multi-run simulation, then distinct output prefixes will be generated by task manager.

If you're worried about it not being clear enough though, include both.

RUFAS/task_manager.py
Isgotkowitz: Then the return type is `str | None`, not Any.

RUFAS/task_manager.py
Isgotkowitz: I suggest including the output prefix and task id in the error message, so that it is clear which task failed. And hopefully it will make it clearer that only that task failed, while others have not.

RUFAS/task_manager.py
ew3361zh: Yea I think we just need to create the output folder for logs in a different (earlier) place to ensure there's a place to save them but I'm fine with this not being added here as well.

RUFAS/task_manager.py
ew3361zh: I like including both for clarity.

RUFAS/task_manager.py
Isgotkowitz: Roger that, I'll write up the new issue

Isgotkowitz (approved):
Nice job Matthew. I suggest using the format `f"{output_prefix} ({task_id})"` when reporting failures, I think it reads a little better. No idea why the number of Mypy errors just went up. I looked at the errors present in Task Manager and I couldn't see how any of them were related to your changes.

Before merging, I think some tests should be written to check the new functionality. I think it would be good to at least write a test that checks errors are handled appropriately by `task`. Ideally there would be a test for `_run_tasks` too, but that might take considerably more effort - I have no idea what a good test for stuff like partial and imap would look like.

ew3361zh (approved):
LGTM!

Issue #1864, opened 2024-07-25
Manure manager failed to recover from error: complex number in place of integer


JoeWaddell:
## Current Behavior

Some simulations from sensitivity analysis runs on `sobol` branch encounter an unrecoverable error.

## Steps to Reproduce

Simulations were run using the following task:
[sobol_animal_tasks.json](https://github.com/user-attachments/files/16379026/sobol_animal_tasks.json)

The summary of failures is below: this one seems restricted to the runs in which the simulation time is 9e24 (e.g., simulation never finished, this was an analytical placeholder). So it's not the first run, meaning it's harder to diagnose/debug.
[sensitivity_analysis global animal FF run summaries.xlsx](https://github.com/user-attachments/files/16379031/sensitivity_analysis.global.animal.FF.run.summaries.xlsx)

Example error file below:
[test-9-animal_S256 run 013_errors_15-Jul-2024_Mon_18-19-26.json](https://github.com/user-attachments/files/16379019/test-9-animal_S256.run.013_errors_15-Jul-2024_Mon_18-19-26.json)

ew3361zh:
I was able to recreate this error by creating calves with negative birthweights in the `heiferII` `_initialize_pregnancy_parameters()`.

In speaking with @JoeWaddell it seems like there are 2 main scenarios to be concerned about:

1. We don't want this error during normal simulations. There should be a catch to zero-out (or another calf-weight minimum?) negative calf-weights that are found.
2. We may want to allow this to happen up to a certain threshold for sensitivity analysis. That threshold is tbd. In this case we may want to allow a certain % of calves to be born with negative weights to not impact normal distributions.

There is a separate error that occurs for setting negative birthweights for cows (in `_handle_successful_conception()`) which should be addressed as well.

Calf weights are determined in `heifeirII._calculate_calf_birth_weight()` so we could ensure a minimum weight there.

Joe, does this sound correct so far? Anything to add?

ew3361zh:
After discussion with Allister and Pooya, I'm pushing resolution of this until after the Animal module refactor is completed since it touches on herd generation and reproduction (and possibly other sub-modules.

ew3361zh:
After doing some testing found that this issue is still present given the inputs outlined above.

It stems from a comparison between a complex number and a float in the `nasem_requirements_calculator.py` function `_calculate_growth_requirements()`:
```python
elif animal_type in [
            AnimalType.HEIFER_I,
[...]

PR #1887, opened 2024-08-05
Calf_birth_weight_error

JoeWaddell:
### Context

- [ ] Yes
- [x] No

### What

Added an error message in cases that the birth weight of an animal is less than or equal to 0. This can happen in rare cases when the input `birth_weight_avg_ho` (or the one for jerseys, e.g. `_je` suffix) is within two (e.g. the value designated as `STDI` in `animal_constants.py`) standard deviations of the average birth weight (as input as `birth_weight_avg_ho`/`_je`).

### Why

This is necessary because it causes a cascading effect that isn't immediately obvious - e.g. no errors are raised within the animal module, but suddenly some animals have a complex value for a body weight, which wreaks havoc all the way in the Manure module, and crashes the simulation. 

### How

All I've done here is add an error with detail as to what's happened, and raised a ValueError to stop the simulation.

An alternative solution would be to institute that the average birth weight is an absolute minimum, though in that case I'd flag this as at a warning - if not, still an error - each time this happens, since it would result in a lopsided birth weight distribution if, for example, half the animals born were be <= 0, as the actual distribution would be nowhere near  the average/stdev expected. The most likely outcome however is that the user would have made the standard deviation wide enough such that the occasional value would be invalid, which wouldn't unreasonably deviate from the desired distribution.

### Test plan

If the user increases the `birth_weight_std_ho` to one within half the `birth_weight_avg_ho` value, they can elicit this error. Otherwise it should not be raised.


JoeWaddell:
From previous tests: I _think_ this (e.g. negative birth weights) was the cause of #1864 (far downstream). @ew3361zh will try to replicate

Isgotkowitz (commented):
This is a nice fix! I think the error message could be improved a little, and this new erroring functionality should also be tested.

CC'ing @allisterakun b/c this change should get carried over to the refreshed animal module.

allisterakun:
Will do!

PR #1890, opened 2024-08-06
Add organic bedding

### LINKED ISSUE #1642: Add proportion of organic bedding to manure non-degradable volatile solids pool
Currently organic bedding is not added to the mass of manure that is tracked in the manure management system. There is a lot of uncertainty around this flow and its impact on emissions but we need to start tracking them. 

I would like a method that adds a fraction of the organic bedding types (sawdust, manure solids, straw) to the non-degradable volatile solids in the reception pit. This fraction would be equal to `(1-`"bedding_cleaned_fraction"`)`, where the bedding cleaned fraction is a user input to the `"bedding_configs"`

Alternatives are to:

1.  ignore the organic bedding like we are doing right now
2.  split the organic bedding that ends up in manure into the degradable and non-degradable volatile solid pools. However, there is not a lot of information about how we would do this so I don't think it is a viable option right now. 

**Additional context**
Methane generated from manure storage is impacted by the type of bedding and there is concensus that organic bedding contributes to methane generation. Although there is uncertainty about the magnitude of this contribution, it is still important for us to start to represent this based on the best available knowledge we have right now. 

Also, before we can implement this, we need to identify default estimates for `"bedding_cleaned_fraction"` for each of the organic bedding types. 

**Target Deadline**
It would be great to have this implemented before the end of May.

Isgotkowitz:
Add organic bedding material into the non-degradable volatile solids pool of manure when it leaves a pen.

### Context

Issue(s) closed by this pull request: closes #1642 

- [ ] Yes
- [x] No

### What

When organic bedding is used for animals, some of the bedding gets removed from the pen and mixed in with the manure that is transported out of that pen.

### Why

Organic bedding has the potential to degrade, which means it has the potential to contribute to GHG emissions. Also, it helps us move towards a more accurate tracking of mass in the manure module.

### How

All `Beddings` must now implement `calc_organic_bedding_mass_added_to_manure`. For beddings that use organic materials (sawdust, straw, etc.) this just multiplies the mass of bedding material used for the animals on a given day by the `bedding_cleaned_fraction` to calculate how much bedding actually left the pen. For non-organic bedding (sand and no bedding), the amount of organic bedding material returned is always 0.

`ManureHandler`s are (currently) responsible for handling how the bedding effects the manure, so `ManureHandler.daily_update` orchestrates the addition of organic matter into the non-degradable volatile solids pool of manure before it is passed along to the `ReceptionPit`. `ManureHandlerDailyOutput` also explicitly reports how much organic bedding was added every day. 

### Test plan

The test suite for manure beddings and handlers has been updated so that it tests new functionality and is passing.

I tried looking at the results with the following csv filter, and it seems that the total non-degradable volatile solids  in the Manure Handler daily output are close to the sums of organic bedding added and the non-degradable solids that come from the animal module, but not the same. However I am not really sure if this correct or not.
```txt
ManureHandlerDailyOutput_Pen_0_CALF.liquid_manure_total_non_degradable_volatile_solids
ManureHandlerDailyOutput_Pen_0_CALF.organic_bedding_added_to_manure
AnimalModuleInputToManureModule_Pen_0_CALF.non_degradable_volatile_solids
```


RUFAS/routines/manure/beddings/bedding_classes.py
```diff
@@ -233,6 +251,10 @@ def calc_total_bedding_mass(self, num_animals: int) -> float:
         """
         return num_animals * self.bedding_mass_per_day
 
+    def calc_organic_bedding_mass_added_to_manure(self, bedding_mass: float) -> float:
+        """Calculates the amount of organic bedding mass added to manure (kg)."""
+        return self.bedding_cleaned_fraction * bedding_mass
```
krb78, line 256: why add in a cleaned_fraction? why not just use bedding_mass_per_day? Isn't bedding_mass_per_day the amount of bedding added per day? We don't accumulate bedding in stalls (or let it diminish), so what is added = what is cleaned.

krb78 (approved):
I think it looks great, Ed. It's a nice simple PR and works. I had an annoying comment that might mess up your flow, but it seems to me that we've added an input we don't actually need.

RUFAS/routines/manure/beddings/bedding_classes.py
Isgotkowitz: >why add in a cleaned_fraction? why not just use bedding_mass_per_day? 

Unfortunately I am not the best person to ask about this, the reason I implemented it like this was because the issue called for it pretty explicitly. To get more details on this I would ask @KFosterReed.

> Isn't bedding_mass_per_day the amount of bedding added per day? We don't accumulate bedding in stalls (or let it diminish), so what is added = what is cleaned.

bedding_mass_per_day is defined (in docstrings) as the "Quantity of bedding required per animal per day (:math:`kg/animal/day`).", so I think you're right that it means it is the amount added per day - but I'm not 100% sure.

ew3361zh (approved):
LGTM! I would defer to SMEs on the results looking appropriate but no issues on the code side.

Isgotkowitz:
Going to merge this. I am not 100% confident that the non-degradable volatile solids pool is behaving exactly the way we want it to, but this is just a first implementation that will be expanded and refined as the mass balance for the manure module is refined.

PR #1891, opened 2024-08-06
Update feed emissions - 1868 Replica 

PooyaHekmati:
This is a replica of PR 1868 to bring the new emission files to the `main` branch.


PooyaHekmati:
@jadamchick tested this and ran it with the default herd and with one of the pilot testing herds (c2f11); ran smoothly and the outputs look reasonable.

PR #1892, opened 2024-08-07
Cap and flare modification

### LINKED ISSUE #1857: Add Representation of 'Cap and Flare' to outdoor manure storage/treatment systems
This manure management strategy covers long term liquid and slurry storage systems to capture the methane produced by the manure and then burns the methane to convert it to CO2 (which, in this case is not a greenhouse gas). Although, in theory, this practice could eliminate all of GHG emissions from the methane produced during storage, in practice, the burning or 'flaring' is not 100% effective and some methane still escapes depending on the weather and seal of the cover. This technology was on our original V1 requirements and we would like to add a simple representation of this method that is a modifier of our current `slurry storage outdoor` and `anaerobic lagoon` methods. 

I propose a relatively simple modification to the existing methods for estimating methane emissions from outdoor slurry storage and anaerobic lagoons which will reduce the current estimates of methane emission from these systems by the reported average methane destruction efficiency of 81%. I think we can achieve this with the following steps:

1. Add an input option to the current `"manure_cover"` input field within the `"manure_treatment_configs"` section of the `manure_management` input file. The new option could be `"cover and flare"` to indicate that this storage system has a cover as well as a flare. 
2. Add either a new input or a manure module constant that is the `methane_destruction_efficiency` with a default value of 81%
3. In the methane update methods in `slurry_storage_outdoor.py` and `anaerobic_lagoon.py` use the new `cover and flare` input to indicate that when this input is provided, the methane predicted by the `methane_emission_from_slurry_storage` is multiplied by `(1- methane_destruction_efficiency/100)` before it is recorded in the `daily_output.storage_methane`
4. Add a new daily output from the manure treatment methods that is something like `storage_methane_avoided` and if the `cover and flare` input is provided, set this new output equal to the  `methane_emission*methane_destruction_efficiency/100`

There are many places that this 'methane destruction' could take place. We could, for example, implement the partitioning of the storage methane into the avoided/burned methane and the methane emission within the `methane_emission_from_slurry_storage` method in `calculator.py`. Alternatively, these could be implemented in the manure treatment files as suggested above. A third option would be to apply these calculations in the EEE module, although this seems like the least appropriate to me because it is a modification of a manure management practice and thus a change to the biophysical outcomes, not the way these outcomes are summarized.

**Additional context**
I have attached a white paper by April Leytem where she references the average methane destruction efficiency value I am proposing. 

In the future, we would like to make the methane destruction efficiency value a function of weather and specific manure storage attributes but we do not know enough about these relationships to define the underlying equations to be able to implement this at this time.  

[Cap and Flare White Paper - Final.docx](https://github.com/user-attachments/files/16365180/Cap.and.Flare.White.Paper.-.Final.docx)

**Target Deadline**
August 30th if feasible

matthew7838:
Add representation of 'Cap and Flare' to outdoor manure storage/treatment systems.
### Context

Issue(s) closed by this pull request: closes #1857

- [ ] Yes
- [x] No

### What

- Add an input option to the current "manure_cover" input field within the "manure_treatment_configs" section of the manure_management input file. The new option could be "cover and flare" to indicate that this storage system has a cover as well as a flare.
- Added a manure module constant that is the methane_destruction_efficiency with a default value of 81%
- Added conditional statements so the methane predicted by the methane_emission_from_slurry_storage is multiplied by (1- methane_destruction_efficiency/100) when the newly added cover and flare were used.
- Add a new daily output called storage_methane_avoided.

### Why

This technology was on our original V1 requirements and we would like to add a simple representation of this method that is a modifier of our current slurry storage outdoor and anaerobic lagoon methods.

### Test plan

- SMEs to confirm simulation results as intended


RUFAS/routines/manure/constants_and_units/manure_constants.py
```diff
@@ -71,3 +71,8 @@ class ManureConstants:
     """
     The fraction of total solids in Liquid manure. Temporary placeholder until upstream solids tracking is fixed.
     """
+
+    METHANE_DESTRUCTION_EFFICIENCY = 81
+    """
+    The fraction of methane destruction efficiency rate.
```
Isgotkowitz:
```suggestion
    The percentage of methane destroyed on a daily(?) basis in systems using a cap and flare.
```
This constant is a percentage, not a fraction. Also, I am not sure that this is exactly what/how this constant used.

RUFAS/routines/manure/manure_treatments/anaerobic_lagoon.py
```diff
@@ -115,6 +115,12 @@ def _daily_update_helper(self) -> ManureTreatmentDailyOutput:
         methane_loss, methane_emission_from_degradable_volatile_solids = self._update_methane_emission(
             self._accumulated_output
         )
+
+        if self.config.manure_cover == "cover and flare":
+            daily_output.storage_methane_avoided = methane_loss * ManureConstants.METHANE_DESTRUCTION_EFFICIENCY / 100
+            methane_loss = methane_loss * (1 - ManureConstants.METHANE_DESTRUCTION_EFFICIENCY / 100)
```
Isgotkowitz:
```suggestion
            daily_output.storage_methane_avoided = methane_loss * ManureConstants.METHANE_DESTRUCTION_EFFICIENCY * GeneralConstants.PERCENTAGE_TO_FRACTION
            methane_loss = methane_loss * (1 - ManureConstants.METHANE_DESTRUCTION_EFFICIENCY * GeneralConstants.PERCENTAGE_TO_FRACTION)
```

RUFAS/routines/manure/manure_treatments/anaerobic_lagoon.py
```diff
@@ -115,6 +115,12 @@ def _daily_update_helper(self) -> ManureTreatmentDailyOutput:
         methane_loss, methane_emission_from_degradable_volatile_solids = self._update_methane_emission(
             self._accumulated_output
         )
+
+        if self.config.manure_cover == "cover and flare":
```
Isgotkowitz:
```suggestion
        if self.config.manure_cover == ManureCoverEnum.COVER_AND_FLARE.value:
```
Ideally, `self.config.manure_cover` would be an enum element and not a string, and would be compared directly to the `ManureCoverEnum`.

RUFAS/routines/manure/manure_treatments/manure_treatment_daily_output.py
```diff
@@ -321,6 +321,9 @@ class ManureTreatmentDailyOutput(LiquidManurePortionProtocol):
     solid_manure_carbon_decomposition: float = 0.0
     solid_manure_carbon_decomposition_unit: MeasurementUnits = MeasurementUnits.KILOGRAMS
 
+    storage_methane_avoided: float = 0.0
+    storage_methane_avoided_unit: MeasurementUnits = MeasurementUnits.KILOGRAMS
+
```
Isgotkowitz: Docstring needs to be updated. Also, I suggest the name `storage_methane_burned` instead - "avoided" is kinda unclear.

RUFAS/routines/manure/manure_treatments/slurry_storage_outdoor.py
```diff
@@ -313,6 +314,14 @@ def _daily_update_helper(self) -> ManureTreatmentDailyOutput:
             accumulated_liquid_manure_total_non_degradable_volatile_solids=(
                 self._accumulated_output.liquid_manure_total_non_degradable_volatile_solids),
         )
+
+        daily_output.storage_methane = methane_loss
+
+        if self.config.manure_cover == "cover and flare":
+            daily_output.storage_methane_avoided = methane_loss * ManureConstants.METHANE_DESTRUCTION_EFFICIENCY / 100
+            methane_loss = methane_loss * (1 - ManureConstants.METHANE_DESTRUCTION_EFFICIENCY / 100)
+            daily_output.storage_methane = methane_loss
```
Isgotkowitz: This code perfectly duplicates the new code in `AnaerobicDigestion`, which means that it should be extracted out into a method that can be used by both storages.

Isgotkowitz (commented):
This is a good start to introducing cap and flare, but I noticed several things that do not follow best practices for clean code. Left comments on several things, please let me know if you have any questions.

RUFAS/routines/manure/constants_and_units/manure_constants.py
matthew7838: I am not sure of the best description either. @KFosterReed do you have any suggestions?

RUFAS/routines/manure/manure_treatments/slurry_storage_outdoor.py
matthew7838: Not sure where would be the best place to extract this. Do you have any class in mind?

Isgotkowitz: My best educated guess is that it should be placed at the lowest interface/module/class that is shared by both Anaerobic Digestion and Slurry Storage Outdoor. So in this case I think the best bet would be `BaseManureTreatment`.

RUFAS/routines/manure/constants_and_units/manure_constants.py
KFosterReed: Ed's suggestion is great. I just removed 'daily' because although we are applying it daily, it is not exclusively a daily estimate

KFosterReed:
[C+F_test-AL_errors_14-Aug-2024_Wed_10-04-35.json](https://github.com/user-attachments/files/16615227/C%2BF_test-AL_errors_14-Aug-2024_Wed_10-04-35.json)

When I tried to test this, the simulation is failing because it is producing a key error from the nitrous oxide emissions methods that also require the manure cover inputs but don't have an option for the `'cover and flare'` input. Please let me know when this is corrected

KFosterReed:
Alright! This works as expected! The only thing missing is that there is no data added to the new output variable `storage_methane_burned` so it is always 0. If this can be updated so that it is actually reporting the destroyed methane then it is ready to go!

KFosterReed:
> Alright! This works as expected! The only thing missing is that there is no data added to the new output variable `storage_methane_burned` so it is always 0. If this can be updated so that it is actually reporting the destroyed methane then it is ready to go!

Update! I realized there was just a mismatch in the names for the variables addeded to the manure treatment outputs and the variables in the treatment files where the data was being assigned! I fixed it and now it works fully as expected! Nicely done Matthew

KFosterReed:
And here is an example of the summary outputs for slurry storage (SS) and anaerobic lagoons (AL) with and without cap and flare (C+F):

image

KFosterReed (approved):

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
```diff
@@ -382,3 +383,23 @@ def _adjust_accumulated_output(self, manure_treatment_daily_output: ManureTreatm
...
+            The amount of methane lost (kg).
+
+        Returns
+        -------
+        tuple[float, float]
+            The amount of storage methane burned and the adjusted methane loss (kg).
+
+        """
+        storage_methane_burned = methane_loss * ManureConstants.METHANE_DESTRUCTION_EFFICIENCY / 100
+        adjusted_methane_loss = methane_loss * (1 - ManureConstants.METHANE_DESTRUCTION_EFFICIENCY / 100)
```
```suggestion
        storage_methane_burned = methane_loss * ManureConstants.METHANE_DESTRUCTION_EFFICIENCY *  GeneralConstants.PERCENTAGE_TO_FRACTION
        adjusted_methane_loss = methane_loss * (1 - ManureConstants.METHANE_DESTRUCTION_EFFICIENCY *  GeneralConstants.PERCENTAGE_TO_FRACTION)
```

RUFAS/routines/manure/manure_treatments/anaerobic_lagoon.py
```diff
@@ -115,6 +115,12 @@ def _daily_update_helper(self) -> ManureTreatmentDailyOutput:
         methane_loss, methane_emission_from_degradable_volatile_solids = self._update_methane_emission(
             self._accumulated_output
         )
+
+        if self.config.manure_cover == ManureCoverEnum.COVER_AND_FLARE.value:
+            daily_output.storage_methane_burned, daily_output.storage_methane = self.calculate_cover_and_flare_methane(
+                methane_loss
+            )
+
```
Isgotkowitz, line 125: Storage methane is overwritten on line 128, that line needs to be moved up above this function the same as in slurry storage outdoor.

Isgotkowitz (changes requested):
There is a bug in anaerobic lagoon. Also, how is testing going to be handled? Will it be done in this PR or in a follow up?

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
matthew7838: Thanks, I did not update the old code and used the commit on GitHub, so I copied the old code.

Isgotkowitz (approved):
LGTM, nice job Matthew

Issue #1848, opened 2024-07-22
[OM] Output manager initialized as global all across the codebase(1/2)


matthew7838:
Most OMs in the code are global in the current codebase.
## Expected Behavior

Should be initialized in the constructor or in the static methods.

## Current Behavior

Most of them are global variables, which is not a good practice and makes testing harder.
## Possible Solution

Initialize them in the constructor or in the static methods.
## Steps to Reproduce

1. Browse different files.
2. Look for OM's declared outside the class.
3.
4.

PR #1908, opened 2024-08-08
Remove `_get_latitude` from `Weather` module

### LINKED ISSUE #1851: [Weather] Remove unnecessary calls to `_get_latitude` in `Weather`
The way that `_get_latitude` is used in the `Weather` module should be simplified.

## Status

The method `_get_latitude` is called in two different places in the `Weather` module, `get_conditions_series` and `get_current_day_conditions`. These method calls are redundant, because `_get_latitude` produces the same result every time it is called.

## Intention

By making a single call to `_get_latitude` instead of a call every single time that a Current Day Conditions instance is retrieved, the number of method calls to both `_get_latitude` and Input Manager's `get_data` can be reduced significantly. Additionally, if there was no latitude provided to a simulation, then the number of warnings will be significantly reduced as well.

## Request

Call `_get_latitude` in the `init` routine of `Weather`, and store the latitude as an instance attribute.

Isgotkowitz:
Removes the method `_get_latitude` from the `Weather` module.

### Context

Issue(s) closed by this pull request: closes #1851 

- [ ] Yes
- [x] No

### What

Removes the `_get_latitude` method from the `Weather` module.

### Why

This method is being removed for a couple reasons. One of them was that it was being called too much which resulted in too many calls being made to Input Manager's `get_data`, and in some cases too many warnings being logged.

Another issue with this method is that it was hiding configuration from caller methods. This could be very problematic if two different caller functions needed to gather current day conditions for two different latitudes.

### How

The two methods in `Weather` that gather `CurrentDayConditions` now allow caller functions to explicitly pass in the latitude that will be used to calculate the daylengths. However this input is not required (i.e. can be `None`) and `None` is the default argument for the latitude. If `None` is passed as the latitude, then the daylength(s) returned will be `None`. The calls to `get_current_day_conditions` in `FieldManager` have been updated to pass in the latitude of each field.

### Test plan

Updated and ran the unit test suite to ensure code had expected behavior.

**Update**: I had not updated any of the outputs before getting this reviewed, nor cleared any changes to outputs with FARM-ES. This PR has been updated so that it will record the daylength used each field (because each field has it's own latitude, and thus potentially a different daylength).  I have also remove the `daylength` recorded by Weather, because this daylength would have always been None as `Weather` is not tied specifically to any location. This approach is still pending final approval that it will be kosher for FARM-ES, this PR will not be merged until this is confirmed.

I have tested these changes by running the default simulation with the following filter, verifying that daylength is only reported by fields and not by `Weather`: 
```json
{
    "multiple": [
        {
            "name": " ",
            "filters": ["Weather.*", ".*daylength"],
            "variables": [".*"]
        }
    ]
}
```


RUFAS/weather.py
```diff
@@ -113,10 +115,12 @@ def get_current_day_conditions(self, time: Time) -> CurrentDayConditions:
...
-        latitude = self._get_latitude()
-        daylength = CurrentDayConditions.determine_daylength(
-            time.current_julian_day, latitude, time.current_calendar_year
-        )
+        if latitude:
+            daylength = CurrentDayConditions.determine_daylength(
+                time.current_julian_day, latitude, time.current_calendar_year
+            )
+        else:
+            daylength = None
```
ew3361zh, line 123: Previously if there was no latitude the model would default to using a latitude of Madison, WI which, I believe would always result in there being a daylength being available.

Can you explain a little further why this change to have daylengths of None and latitudes of None is desired over the previous behavior?

Isgotkowitz: > Previously if there was no latitude the model would default to using a latitude of Madison, WI which, I believe would always result in there being a daylength being available.

Yes, that is correct.

> Can you explain a little further why this change to have daylengths of None and latitudes of None is desired over the previous behavior?

This is desired because it is more explicit. If the daylength doesn't matter at all to any of the Animal module routines (for example), then it is good to reflect this with the value of daylength being `None`, and by extension having no latitude kept in the Animal module.

ew3361zh: I think that makes sense.
What do you think then about the third part of our grooming session notes for this issue?
`Hairy situations: latitude is used to get day-length. Right now C&S is the only module that uses it. But other modules may need this information. How to ensure they are using the correct latitude and not a default?`

Isgotkowitz: This is why I like the use of `None` when no latitude is passed. If a module doesn't have any latitude information but it tries to use the daylength from the CurrentDayConditions it has, it will get a NoneType error (or it should, if it doesn't I don't think there is anything the Weather module can do about that). Otherwise, if a module does have a latitude that it wants to get a daylength for, it is expressly the responsibility of that module to use the correct latitude.

If a module needs the daylength that doesn't need it right now, that means something is changing in that module. Part of that change will have to be getting and keeping a latitude to use, and it will be the responsibility of that module and the developers working on it to make sure that it is correct and not a default.

ew3361zh: Thanks, Ed, sounds good to me. I think that would match with the longer-term goal of the latitude information being derived from `Field`.

ew3361zh (approved):

RUFAS/weather.py
Isgotkowitz: On the other hand, I could see someone making the case that `None` shouldn't even be provided as a default latitude/daylength for any module. I can see the benefits of this. It wouldn't give modules that use `CurrentDayConditions` the option to mishandle `None` as the daylength, because daylength simply wouldn't be an attribute of `CurrentDayConditions`. There would be a separate routine somewhere else that gets the daylength and that would be its express purpose.

This wouldn't be a bad way to go, but I see a few downsides. The first being that it would not be as clean, it is very simple and convenient to have the daylength packaged with all the other conditions for a single day. Second, when a module changes from not needing daylength to needing it, it would need another method call and place to track that daylength instead of just adding a parameter to the existing method call it has for `CurrentDayConditions` (if it has one, I guess).

JoeWaddell: For the Animal module, we don't have anything in place yet that requires latitude/daylength, but I can imagine cases in which we might want _a_ value. However, I can't imagine cases where we'd want a coordinate for each pen (like we do with field, for more precision in growing degree hours), so I imagine info we can glean from the `FIPS_county_code` is enough.

JoeWaddell (approved):
Changes make sense, and looks like a clean implementation. Left a single comment, no changes required.

Isgotkowitz:
> For the Animal module, we don't have anything in place yet that requires latitude/daylength, but I can imagine cases in which we might want _a_ value. However, I can't imagine cases where we'd want a coordinate for each pen (like we do with field, for more precision in growing degree hours), so I imagine info we can glean from the `FIPS_county_code` is enough.

You raise an interesting point here @JoeWaddell, how to deal with location data when we have multiple sources of it. If the `FIPS_county_code` is specified for a simulation, then how do we check that it isn't _too_ far away from other locations? Maybe a farmer keeps their animals in one location, but buys their feed in another because that makes more sense. Or maybe their barn and main set of fields are in one county, but they have a satellite set of fields several counties over. It makes sense that the latitude might be outside of the strict bounds inherent in the FIPS code, but how far outside is too outside?

RUFAS/weather.py
Isgotkowitz: When you say "glean from `FIPS_county_code`", do you mean get a latitude out of the FIPS for use in calculating the daylength?

JoeWaddell: That was my thought: assuming the center of the county is close enough for the needs in some/many cases.

JoeWaddell:
> Maybe a farmer keeps their animals in one location, but buys their feed in another because that makes more sense. Or maybe their barn and main set of fields are in one county, but they have a satellite set of fields several counties over.

Excellent question: I think what's in place now likely covers the majority of circumstances, where the FIPS code would be assumed to be for the facilities, and the field lat/longs would handle edge cases where some fields are outside the county...but we could make it more explicit.

Isgotkowitz:
@ew3361zh @JoeWaddell I realized that I had not considered the output implications of this PR, I need to speak with someone who can be aware/communicate this change with FARM-ES before I merge.

Edit: spoke with @krb78, and outlined the following solution with her. Remove `daylength` from the `Weather` outputs, and have each field report the `daylength` it uses.

RUFAS/weather.py
```diff
@@ -93,14 +92,17 @@ def __init__(self, weather_file: dict, time: Time):
...
-    def get_current_day_conditions(self, time: Time) -> CurrentDayConditions:
+    def get_current_day_conditions(self, time: Time, latitude: float | None = None) -> CurrentDayConditions:
         """
         Creates a CurrentDayConditions object containing all the weather conditions on the current day.
 
         Parameters
         ----------
         time: Time
             Time object containing the current time of the simulation.
+        latitude : float | None, default None
```
ew3361zh, line 103:
```suggestion
        latitude : float | None, default=None
```
Very small thing here and not necessary to implement but it seems like some places we have `default=xyz` and other places we're doing `default xyz`. Seems like Numpy counts them as equivalent so only need to decide if we care that they all look the same internally.

ew3361zh (approved):
Assuming the outputs are as desired for Kaitlyn's approval, I'm good with these changes as is. Nice work, Ed.

RUFAS/weather.py
Isgotkowitz: I have noticed the same thing. Since Numpy seems ambivalent about how its done, I think it isn't a big deal if it is inconsistent in RuFaS. Though it would be nice to have them all be the same.

Isgotkowitz:
> Assuming the outputs are as desired for Kaitlyn's approval, I'm good with these changes as is. Nice work, Ed.

Thanks Niko. I checked in with Kaitlyn this morning and she said this approach was ok.

RUFAS/routines/field/manager/field_manager.py
```diff
@@ -49,10 +50,11 @@ def __init__(self, manure_manager: ManureManager, feed_manager: FeedManager):
             "function": "__init__",
         }
         self.im = InputManager()
+        self.om = OutputManager()
```
JoeWaddell, line 53: Why is output manager being assigned to the Field Manager like this?

JoeWaddell (approved):

RUFAS/routines/field/manager/field_manager.py
Isgotkowitz: Great question, a couple of reasons. First, we want to move away from instantiating the Output Manager outside of classes. Second, since it gets used in multiple different instance methods it saves us some time to instantiate Output Manager once and reuse that same instance in every method that needs it.

Isgotkowitz: #1848 is dedicated to correcting this across the code base.

PR #1916, opened 2024-08-12
Modify digester methane leakage variable name

elle-andreen:
### Context
A new output was recently created in PR #1843 representing the small fraction of methane produced in a digester that escapes the gas capture system and leaks out into the atmosphere (`methane_leakage_mass`, units = kg CH4).

With the upcoming FARM ES integration, though not an ideal long-term variable name, Kristan requested we rename this variable to `storage_methane` (also with units = kg CH4) so that it is picked up by existing filters, as other manure treatments utilize this variable name to represent methane emitted into to the atmosphere. 

Issue(s) closed by this pull request: N/A

- [ ] Yes
- [X] No


elle-andreen:
Note related to closing this PR (from Kristan): because of the report filter aggregations methods, we can still easily provide FARM ES with a single value for the total aggregated storage emissions, so there is no need to make this change!

PR #1919, opened 2024-08-13
Add Units to Report Filter Constants

### LINKED ISSUE #1787: [Report Generator] Associate Report Filter Constants with Enum

ew3361zh:
As a follow up to PR #1798, this PR takes that established unit-adding logic and applies it to constants in report filters. It does so by providing logic to try to align report constants with RuFaS `GeneralConstants` which themselves are mapped to specific units.

### Context

Issue(s) closed by this pull request: closes #1787

- [ ] Yes
- [x] No

### What

Updates RG to check report filter constants against established RUFAS GeneralConstants.
If there is a match and the user has opted to add units to their reports, this will take the matched constant and search the established `CONSTANTS_TO_UNITS` map which has my best attempt at providing a `MeasurementUnit` unit for each constant.
Once the unit for the constant is established, the logic from #1798 which parses and combines and simplifies units will apply that to the constant and whatever data is being aggregated with the constant.

### Why

This will provide units for reports aggregated with constants.

### How

- Builds the `CONSTANTS_TO_UNITS` dictionary map which has a unit for each constant.
- Adds logic to the RG._validate_constants() method to try and match the constant name provided by the user in the report filter with one of the GeneralConstants.
- If no match is found, `unit_not_found` will be the default unit provided. If a different default behavior here would be preferred, please suggest some ideas (I thought `unitless` might be a good default but since `unitless` is an established unit, this might be deceiving to the user and the results).
- Adds a function to normalize both the user-provided constant name and the GeneralConstant enum attribute to attempt to match regardless of case and spaces/underscores.

### Test plan

- Unit testing was updated to return coverage to previous level.
- Run with the provided test filter or some variation similar:
```
{
  "multiple": [
    {
      "name": "% milk fat",
      "filters": ["AnimalModuleReporter.report_life_cycle_manager_data.herd_milk_fat_percent"],
      "constants": {
          "FRACTION_TO_PERCENTAGE": 100
      },
      "slice_start": -365,
      "display_units": true
  },
  {
    "name": "Milk fat, kg",
    "filters": [
        "AnimalModuleReporter.report_life_cycle_manager_data.herd_milk_fat_kg"
    ],
    "slice_start": -365,
    "display_units": true
},
  {
    "name": "weighted fat value",
    "cross_references": [
        "Milk fat, kg.*"
    ],
    "constants": {
        "MILK_FAT_WEIGHT": 12.2
    },
    "display_units": true
}
  ]
}
```


RUFAS/general_constants.py
```diff
@@ -53,10 +56,35 @@ class GeneralConstants:
...
+        KG_TO_CUBIC_METERS: MeasurementUnits.CUBIC_METERS,
+        YEAR_LENGTH: MeasurementUnits.DAYS,
+        LEAP_YEAR_LENGTH: MeasurementUnits.DAYS,
+        SECONDS_PER_DAY: MeasurementUnits.SECONDS,
+        WATER_DENSITY_KG_PER_LITER: MeasurementUnits.KILOGRAMS,
+        WATER_DENSITY_KG_PER_M3: MeasurementUnits.CUBIC_METERS,
+        PROTEIN_TO_NITROGEN: MeasurementUnits.UNITLESS,
+        NITROGEN_TO_PROTEIN: MeasurementUnits.UNITLESS,
+        MILK_FAT_WEIGHT: MeasurementUnits.UNITLESS,
+        EARTH_ANGULAR_VELOCITY: MeasurementUnits.UNITLESS,
```
Isgotkowitz: The value we have for Earth's angular velocity is actually radians per hour, which I don't think we have in `MeasurementUnits`. I doubt it is going to be used anywhere, but would still be good to have it be correct.

RUFAS/general_constants.py
```diff
@@ -53,10 +56,35 @@ class GeneralConstants:
...
+        GRAMS_TO_KG: MeasurementUnits.KILOGRAMS,
+        KG_TO_GRAMS: MeasurementUnits.GRAMS,
+        LITERS_TO_CUBIC_METERS: MeasurementUnits.CUBIC_METERS,
+        CUBIC_METERS_TO_LITERS: MeasurementUnits.LITERS,
+        KG_TO_CUBIC_METERS: MeasurementUnits.CUBIC_METERS,
+        YEAR_LENGTH: MeasurementUnits.DAYS,
+        LEAP_YEAR_LENGTH: MeasurementUnits.DAYS,
+        SECONDS_PER_DAY: MeasurementUnits.SECONDS,
+        WATER_DENSITY_KG_PER_LITER: MeasurementUnits.KILOGRAMS,
+        WATER_DENSITY_KG_PER_M3: MeasurementUnits.CUBIC_METERS,
```
Isgotkowitz: These conversion constants all have some sort of compound unit, for example the unit of `YEAR_LENGTH` is days per year, not days. I'm not sure how this actually fits into logic of units, but if they need to have the units listed here to work then it should be explicit about that.

RUFAS/general_constants.py
```diff
@@ -53,10 +56,35 @@ class GeneralConstants:
     # Biochemistry-related
     PROTEIN_TO_NITROGEN = 0.16
     NITROGEN_TO_PROTEIN = 6.25
+    MILK_FAT_WEIGHT = 12.2
```
Isgotkowitz, line 60: Docstring needs to be updated with this value. Also I'm curious - is it really unitless?

RUFAS/units.py
```diff
@@ -33,6 +34,8 @@ class MeasurementUnits(Enum):
     KILOGRAMS_PER_ANIMAL = "kg/animal"
     KILOGRAMS_PER_DAY = "kg/day"
     KILOGRAMS_PER_HECTARE = "kg/ha"
+    KILOMETERS = "km"
+    LITERS = "L"
```
Isgotkowitz, line 49: Should this be "L" or "l"?

RUFAS/report_generator.py
```diff
@@ -870,19 +871,22 @@ def _add_constants_to_report_data(
             return
 
         try:
-            self._validate_constants(report_data, constants_config)
+            validated_constants_config = self._validate_constants(
+                report_data, constants_config, filter_content.get("display_units", False)
```
Isgotkowitz: I think it would be good to validate display units as well.

RUFAS/report_generator.py
```diff
@@ -917,15 +924,34 @@ def _validate_constants(
...
+            if not isinstance(name, str) or len(name) == 0:
+                raise ValueError(f"Constant name {name} must be a string and cannot be empty.")
 
             if not isinstance(value, (int, float)):
                 raise ValueError(f"Constant value {value} must be a number.")
 
+            if display_units:
+                normalized_provided_name = self._normalize_constant_name(name)
+                matching_constant_value = None
+                for attribute in dir(GeneralConstants):
```
Isgotkowitz: This `dir` yields a lot more than just the constants in GeneralConstants, it also includes a lot of things like `__dir__`, `__hash__`, `__eq__` etc. Even though it seems unlikely someone would ever name their constants something that matches one of those builtins, it would still be good to remove those options from the list - possibly with a regex filter similar to "__.*__"

RUFAS/report_generator.py
```diff
@@ -917,15 +924,34 @@ def _validate_constants(
...
+                    normalized_attribute_name = self._normalize_constant_name(attribute)
+                    if normalized_attribute_name == normalized_provided_name:
+                        matching_constant_value = getattr(GeneralConstants, attribute)
+                        break
+
+                unit_for_constant = GeneralConstants.CONSTANTS_TO_UNITS.get(matching_constant_value, "unit_not_found")
+                constant_with_units = f"{name}_({unit_for_constant})"
```
Isgotkowitz: If the associated units cannot be found, should a warning be logged about it? I tried reports with a constant that wasn't in the mapping but there was no record of that in the outputs, and no message about it logged.

Isgotkowitz (commented):
I am not sure exactly what the limitations of the units handling are currently, I haven't worked with them very much. However I tried with the following report, and I didn't see the units conversion I expected (the output was still in kg / ha instead of g / ha). If this is expected then it can be ignored.
```json
{
    "multiple": [
        {
            "name": "test units",
            "filters": [".*nitrate.*layer='1.*"],
            "constants": {
                "KG_TO_GRAMS": 1000.0
            },
            "horizontal_first": false,
            "display_units": true
        }
    ]
}
```
Other than that, I just noticed a few more minor things regarding the units of constants and docstring stuff.

ew3361zh:
> I am not sure exactly what the limitations of the units handling are currently, I haven't worked with them very much. However I tried with the following report, and I didn't see the units conversion I expected (the output was still in kg / ha instead of g / ha). If this is expected then it can be ignored.
> 
> ```json
> {
>     "multiple": [
>         {
>             "name": "test units",
>             "filters": [".*nitrate.*layer='1.*"],
>             "constants": {
>                 "KG_TO_GRAMS": 1000.0
>             },
>             "vertical_aggregation": "average",
>             "horizontal_aggregation": "product",
>             "horizontal_first": false,
>             "display_units": true
>         }
>     ]
> }
> ```
> 
> Other than that, I just noticed a few more minor things regarding the units of constants and docstring stuff.

Good call, these constants converting one unit to another should be in conversion-friendly format so this `KG_TO_GRAMS` should have a MeasurementUnit of `GRAMS_PER_KILOGRAM` = `g/kg`

RUFAS/general_constants.py
ew3361zh: I can update that.

RUFAS/general_constants.py
ew3361zh: Good call! Updating these all to be compound units. Currently the unit simplification in the MeasurementUnits class will reduce like units in numerators/denominators so this would be handled.

RUFAS/general_constants.py
ew3361zh: I believe so according to a conversation I had with Julie.

ew3361zh: Also not entirely sure this should be a GeneralConstant but it can be easily removed if desired.

RUFAS/units.py
ew3361zh: According to the internet, americans prefer the capital `L` so as not to potentially confuse it with an uppercase `i`.

RUFAS/report_generator.py
ew3361zh: Right now we're not validating any of the report filter attributes. For GG, we have a function dedicated to that purpose. I think it would be worth doing in a separate PR. I can write up an issue for it (and maybe at a grooming session we can weigh in on the necessity of this given the future changes coming to post-processing).

RUFAS/report_generator.py
ew3361zh: I can add it. I wish it were easier to log warnings in RG.

RUFAS/general_constants.py
```diff
@@ -53,10 +57,35 @@ class GeneralConstants:
...
     # Earth related data
     EARTH_ANGULAR_VELOCITY = 0.2618
 
     # Fractions and Percentages
     PERCENTAGE_TO_FRACTION = 0.01
     FRACTION_TO_PERCENTAGE = 100.0
+
+    CONSTANTS_TO_UNITS = {
+        "MM_TO_M": MeasurementUnits.METERS_PER_MILLIMETER,
+        "CM_TO_MM": MeasurementUnits.CENTIMETERS_PER_MILLIMETER,
```
JoeWaddell: This should be `MILLIMETERS_PER_CENTIMETER`, no?

JoeWaddell:
```suggestion
        "CM_TO_MM": MeasurementUnits.MILLIMETERS_PER_CENTIMETER,
```

RUFAS/general_constants.py
```diff
@@ -53,10 +57,35 @@ class GeneralConstants:
...
+        "MM_TO_CM": MeasurementUnits.MILLIMETERS_PER_CENTIMETER,
```
JoeWaddell:
```suggestion
        "MM_TO_CM": MeasurementUnits.CENTIMETERS_PER_MILLIMETER,
```

RUFAS/report_generator.py
```diff
@@ -859,25 +862,84 @@ def _add_constants_to_report_data(
...
         try:
             self._validate_constants(report_data, constants_config)
         except ValueError:
             raise
 
+        if filter_content.get("display_units", False):
+            constants_config, event_logs = self._add_units_to_constants(constants_config)
+
         max_length = max([len(lst) for lst in report_data.values()])
+        print(constants_config)
```
JoeWaddell: reminder this is here

JoeWaddell (commented):
Sending some of my in-progress comments while still reviewing

JoeWaddell (approved):
Works as expected from SME perspective. Left some notes earlier on some of the docstrings that I think are backwards. Great to see more support for units! Nice work

RUFAS/units.py
```diff
@@ -10,47 +10,67 @@ class MeasurementUnits(Enum):
...
     HECTARE = "ha"
-    HOURS = "hours"
-    INJECTIONS = "injections"
+    HOURS = "hour"
+    INJECTIONS = "injection"
+    KILOMETERS_PER_METER = "km/m"
     KILOGRAMS = "kg"
     KILOGRAMS_CARBON_DIOXIDE_PER_KILOGRAM_DRY_MATTER = "kg CO2 / kg DM"
     KILOGRAMS_PER_ANIMAL = "kg/animal"
+    KILOGRAMS_CARBON_DIOXIDE_EQ = "kg CO2-eq"
```
Isgotkowitz, line 42: I think it would be good to not abbreviate EQ here, I assume it is equivalent but I think it would be better to list that explicitly.

Isgotkowitz (approved):
Found one spot where I think an abbreviation should be expanded, and I agree with Joe's comment that it should be millimeters/centimeter. Got the filter working and it is outputing g / ha for me, so LGTM. Nice job Niko!

RUFAS/general_constants.py
ew3361zh: Good catch - my brain kept reversing the orders of these.

RUFAS/units.py
ew3361zh: This one was added by Kristan - I just re-alphabetized it. I can check with her and update it if she wants to while updating another unit for Kaitlyn next sprint (there's an issue established for it).

PR #1932, opened 2024-08-15
Handle NaNs sent to RG Aggregator functions

### LINKED ISSUE #1896: [RG]Handling NaN cases in aggregators
When a data series makes its way to the aggregation functions in the report generator, we don't do any data cleaning and attempt to perform the requested operation. If there are any entries in the data series that are not compatible with the requested math, we crash. A more graceful handling is desired.

This is a follow-up to issue #1856 . 

## Expected Behavior

Instead of crashing, return `0` or `None` or `NaN` etc...

## Current Behavior

See #1856 .

## Possible Solution

Either of these:
A. We can use libraries like SciPy or NumPy to perform data cleaning before performing the actual aggregation.
B. We can perform the aggregation and if the result is not of a valid type, return `0` or `None` or `NaN` etc....
C. [preferred] Raise an exception when something goes wrong in the aggregation and let the caller catch and handle it by populating the result with `0` or `None` or `NaN` etc... . This approach is preferred as it avoid code duplication.

ew3361zh:
Stemming out of the original issue #1856, this PR seeks to properly handle NaNs when sent to the RG aggregator functions.

### Context

Issue(s) closed by this pull request: closes #1896

- [ ] Yes
- [x] No

### What

Accounts for and handles NaNs in the aggregator functions of ReportGenerator.

### Why

Previously if there were NaNs in the data sent to RG, the simulation crashed (which we don't want to happen).

### How

Using the Numpy nan library, the aggregator functions all take NaNs into account and avoid crashing the code when they're encountered.

### Test plan

Testing is updated. I will try to provide some test cases.


RUFAS/report_generator.py
```diff
@@ -35,22 +37,25 @@ def division_aggregator(data: List[float]) -> float | None:
...
     float
-        The result of dividing the first number by each subsequent number.
-        Returns None if the list is empty or has only one element.
+        The result of dividing the first number by each subsequent number,
+        ignoring NaN values. Returns None if the list is empty or has only one
+        valid element (non-NaN).
     """
-    if len(data) < 2:
+    if len(data) < 2 or np.isnan(data[0]):
         return None
```
allisterakun: During group review:
Since we're skipping all the other None values in the list, please consult Pooya or Kristan about skipping the leading Nans until a non-Nan is found.

RUFAS/report_generator.py
```diff
@@ -35,22 +37,25 @@ def division_aggregator(data: List[float]) -> float | None:
     Returns
     -------
     float
-        The result of dividing the first number by each subsequent number.
-        Returns None if the list is empty or has only one element.
+        The result of dividing the first number by each subsequent number,
+        ignoring NaN values. Returns None if the list is empty or has only one
+        valid element (non-NaN).
```
allisterakun: Update the return to reflect the case of returning None (if the first element is None)

RUFAS/report_generator.py
```diff
@@ -115,15 +118,18 @@ def subtraction_aggregator(data: list[float]) -> float | None:
...
     -------
     float
-        The result of subtracting each subsequent number from the first number.
-        Returns None if the list is empty or has only one element.
+        The result of subtracting each subsequent number from the first number,
+        ignoring NaN values. Returns None if the list is empty or has only one
+        valid element (non-NaN).
     """
-    if len(data) < 2:
+    if len(data) < 2 or np.isnan(data[0]):
```
allisterakun: Same comment as above for the division.

allisterakun (changes requested):
During Group review:
1. Found duplicate protections against None and Nan. We are filtering out the None and Nan values before most of the aggregation operations. We are also adding the ability to handle None and Nan data in the actual aggregator functions.
2. Some decisions to consult about the division and subtraction operations: see comments below.

Consider:
1. Remove all the None and Nan filtering before the aggregation operations and just let the aggregator functions to handle these scenrios.
This would implement option A in issue #1896: 

> A. We can use libraries like SciPy or NumPy to perform data cleaning before performing the actual aggregation.
2. User option C in issue #1896, which is the preferred option:

> C. [preferred] Raise an exception when something goes wrong in the aggregation and let the caller catch and handle it by populating the result with 0 or None or NaN etc... . This approach is preferred as it avoid code duplication.

Also, come up with a plan for when to use 0 or None or NaN when these issues come up.

ew3361zh:
@matthew7838 @allisterakun updates summary:

- Removed NaN and None handling from aggregator functions and apply_vertical/apply_horizontal functions
- Added a wrapper function to check incoming data being sent to aggregators to be scanned for Nones and NaNs
- Report error to OM if there is a None or NaN in a data set and return None for the data
- Updated unit testing
- For testing, I added a line into the new `_handle_aggregator_errors()` method which would add a NaN or None into the data and saw the appropriate return (see pic for example)

Screenshot 2024-08-21 at 11 53 43 AM

Continuing to work on reducing mypy error count but the new approach can be reviewed in the meantime.

RUFAS/report_generator.py
```diff
@@ -547,6 +548,57 @@ def _update_key(key: str) -> str:
             updated_key = f"{key}_ver_agg"
         return updated_key
 
+    def _aggregate_units(
```
ew3361zh, line 573: This function is unchanged, I just moved it to be closer to similar functionality in the class.

RUFAS/report_generator.py
```diff
@@ -750,96 +809,108 @@ def _apply_horizontal_aggregation(
...
+        aggregator : Callable[[List[float], float, str], float]
             The aggregation function to be used.
 
         Returns
         -------
-        tuple[str, list[dict[str, str | dict[str, str]]]
-            The expected units of aggregating the report data using the accompanying aggregator function and
-            the event logs.
+        Dict[str, List[float]]
+            The vertically aggregated data as a dictionary of lists.
```
Isgotkowitz: This return type does not match the type hints.

RUFAS/report_generator.py
```diff
@@ -750,96 +809,108 @@ def _apply_horizontal_aggregation(
         if len(set(lengths)) != 1:
             raise ValueError("Can't aggregate data with different lengths")
         max_length = max(lengths)
-        aggregated_data: List[float] = []
+        aggregated_data: list[float] = []
+        event_logs: list[dict[str, str | dict[str, str]]] = []
         for i in range(max_length):
             temp_data = [report_data[key][i] for loop_key in loop_list for key in report_data if loop_key in key]
             non_null_data_points = list(filter(lambda x: x is not None, temp_data))
```
Isgotkowitz: Should this line be removed? It seems that handle_aggregration_errors will handle the data if there is a None passed

RUFAS/report_generator.py
```diff
@@ -750,96 +809,108 @@ def _apply_horizontal_aggregation(
...
-            second_key_numerator_units, second_key_denominator_units = MeasurementUnits.extract_units(second_key)
-            combined_numerator, combined_denominator, event_logs = self._combine_units(
-                first_key_numerator_units,
-                first_key_denominator_units,
-                second_key_numerator_units,
-                second_key_denominator_units,
-                aggregator_key,
-                simplify_units,
+        for key, data in report_data.items():
+            non_null_data_points = list(filter(lambda x: x is not None, data))
```
Isgotkowitz: I think there should be no filtering of Nones from the data, because `_handle_aggregation_errors` has the functionality to deal with Nones in the data, and it seems problematic to me to have to different places handling the same error condition in two different ways.

RUFAS/report_generator.py
```diff
@@ -750,96 +809,108 @@ def _apply_horizontal_aggregation(
...
 
         Returns
         -------
-        Dict[str, List[float]]
-            The vertically aggregated data as a dictionary of lists.
+        tuple[dict[str, list[float | None] | float | None], dict[str, str | dict[str, str]]]
+            The resulting aggregated data and the aggregation logs to be returned to OutputManager.
+            Returns None and an error message if the data contains None or NaN values.
         """
+        if any(x is None or math.isnan(x) for x in data):
```
Isgotkowitz: Might also be good to check for Numpy nan type too? Edit: seems that math.isnan catches numpy nans, so nevermind

Isgotkowitz: Alternatively, it could be checked that all data is numerical, to make sure no strings, bools, etc. are making it into the aggregation function

RUFAS/report_generator.py
```diff
@@ -750,96 +809,108 @@ def _apply_horizontal_aggregation(
...
+            info_map = {
+                "class": self.__class__.__name__,
+                "function": self._handle_aggregation_errors.__name__,
+            }
+            aggregation_error = {
+                "error": "ReportGenerator aggregation error",
+                "message": f"Encountered unaggregatable values in {key} data, returning None instead.",
+                "info_map": info_map,
+            }
```
Isgotkowitz: When I try aggregating some data that contains a lot of Nones or nans, this error seems to print for once for each of the None or Nan values that I am trying to aggregate? Is this intended, or would it be possible to only have it print once for a single data set that has aggregation attempted on it?

RUFAS/report_generator.py
ew3361zh: Yes! Thank you. I think I've been hitting the git stash and not enough of the git apply.

RUFAS/report_generator.py
ew3361zh: Agreed

RUFAS/report_generator.py
ew3361zh: Interesting. I will address that.

RUFAS/report_generator.py
ew3361zh: Good call on the bools, strings. NaNs count as floats oddly enough so I'll still need to specifically check for those.

RUFAS/report_generator.py
ew3361zh: Could you share what your testing process was here so I can try to replicate? I am still getting one error per set of data that contains any number of Nones/NaNs/strings/bools

RUFAS/report_generator.py
```diff
@@ -83,7 +84,7 @@ def sd_aggregator(data: List[float]) -> float:
         The standard deviation of the input numbers.
     """
     mean = average_aggregator(data)
-    return (sum((x - mean) ** 2 for x in data) / len(data)) ** 0.5 if data else 0
+    return (sum((x - mean) ** 2 for x in data) / len(data)) ** 0.5 if data else 0.0
```
ew3361zh, line 87: addresses mypy warning.

RUFAS/report_generator.py
```diff
@@ -555,40 +631,41 @@ def _combine_units(
...
         ValueError
             If the operation is addition or subtraction and the units are not the same.
         """
-        event_logs: list[dict[str, str | dict[str, str]]] = []
         info_map = {
             "class": ReportGenerator.__class__.__name__,
             "function": ReportGenerator._combine_units.__name__,
         }
-
+        event_log = {}
```
allisterakun: Is there a reason we're removing the type hint for `event_logs`?

RUFAS/report_generator.py
```diff
@@ -750,99 +838,109 @@ def _apply_horizontal_aggregation(
...
+        event_logs: list[dict[str, str | dict[str, str]]] = []
+        for key, data in report_data.items():
+            vertically_aggregated_data, aggregation_log = self._handle_aggregation_errors(aggregator, data, key)
+            if aggregation_log:
+                event_logs.append(aggregation_log)
+            aggregated_data[key] = [vertically_aggregated_data]
+        return aggregated_data, event_logs
 
-    def _apply_vertical_aggregation(
+    def _handle_aggregation_errors(
```
allisterakun: Maybe a better name for this function. Because it handles the aggregation errors and performs the aggregation.

allisterakun (approved):
It generally looks good; great job, Niko!
Just two minor comments.
Also, since you're introducing a new function in the PR, it'd be nice to mention it in the PR description.

RUFAS/report_generator.py
ew3361zh: I think what happened here is that this changed from a list of event logs to a single event log. I can add the type hints for this new version.

RUFAS/report_generator.py
ew3361zh: how about just `_handle_aggregation()`?

RUFAS/report_generator.py
Isgotkowitz: I added the following to Simulation Engine, then wrote a report filter to try and aggregate `none_var` (and same thing with Nan, could also be used to test bool, str, etc):
```python
    def simulate(self) -> None:
        """
        Executes the simulation.
        """

        # -- snip -- 
        EEEManager.estimate_all()
        info_map.update({"units": MeasurementUnits.UNITLESS})
        for i in range(100):
            self.om.add_variable("none_var", None, info_map)
        # -- snip --
```

ew3361zh: Hmm, I added this then used this report filter: 

```
{
      "name": "Ed Data",
      "filters": [
          ".*none_var.*"
      ],
      "vertical_aggregation": "sum"
  }
```

and only got one error:
Screenshot 2024-08-26 at 4 10 10 PM


Are you still seeing the multiple errors?

Isgotkowitz: Ooh I bet it was because I was using horizontal aggregation, that would make way more sense

RUFAS/report_generator.py
```diff
@@ -750,99 +838,105 @@ def _apply_horizontal_aggregation(
...
         max_length = max(lengths)
-        aggregated_data: List[float] = []
+        aggregated_data: list[float] = []
+        event_logs: list[dict[str, str | dict[str, str]]] = []
         for i in range(max_length):
             temp_data = [report_data[key][i] for loop_key in loop_list for key in report_data if loop_key in key]
-            non_null_data_points = list(filter(lambda x: x is not None, temp_data))
-            aggregated_data.append(aggregator(non_null_data_points))
+            horizontally_aggregated_data, aggregation_log = self._handle_aggregation(
+                aggregator, temp_data, next(iter(report_data))
```
Isgotkowitz: Is `next(iter(...))` used here because there should only ever be one key in "report_data"?

Isgotkowitz (approved):
LGTM, thank you for addressing comments. Had one random question about a piece of the code. Tried this out with some filters and found no difference between dev and this branch

RUFAS/report_generator.py
ew3361zh: It's an interesting question, I should have brought it up - I added the key to this function definition to be able to track where the potentially bad data (with nones, strings, nans, etc) were coming from. So it's only for error logging.

In the case of horizontal aggregation I was honestly not quite sure how to handle this because it's the only case of all the uses of this function where we're getting multiple keys there (sending each row's data to be aggregated). 

But trying to then track which key's data belongs to which key and then tracking that back to Output Manager to log became way too complicated. Especially since the other callers of this function have one key.

I could send the list of the keys for the data and add it as a string to the error message. We don't have access to the report filter at this point to know the title of the report which could be another reference point for users to go back and track down the bad data (but then they'd have to know their filter well enough to know all the vars it's capturing if they're using Regex).

ew3361zh: I will hold off on merging this until we've come to an agreed solution. But will add the list of all the keys of the horizontally aggregated data as a string for error logging.

So for example for this filter:
```
{
            "name": "Heifers",
            "filters": ["AnimalModuleReporter.report_daily_animal_population.num_(heiferIs|heiferIIs|heiferIIIs)"],
            "horizontal_aggregation":"sum",
            "horizontal_first": true,
            "slice_start": -365
           }
```

if I add some bad data to the data to be aggregated, it would print out this error:
`[26-Aug-2024_Mon_17-54-45.921209][ERROR][default] ReportGenerator aggregation error. Encountered unaggregatable values in ['AnimalModuleReporter.report_daily_animal_population.num_heiferIs', 'AnimalModuleReporter.report_daily_animal_population.num_heiferIIs', 'AnimalModuleReporter.report_daily_animal_population.num_heiferIIIs'] data, returning None instead.`

Isgotkowitz: Sounds good to me - my only suggestion would be to make the log message prettier/more readable, if this is doable without too much additional effort. Something like
`[26-Aug-2024_Mon_17-54-45.921209][ERROR][default] ReportGenerator aggregation error. Encountered unaggregatable values in variable(s): 'AnimalModuleReporter.report_daily_animal_population.num_heiferIs', 'AnimalModuleReporter.report_daily_animal_population.num_heiferIIs', 'AnimalModuleReporter.report_daily_animal_population.num_heiferIIIs', returning None instead.`

PR #1935, opened 2024-08-15
Update GH Action for 2nd Job to Get Latest Commit

### LINKED ISSUE #1862: [GH Actions] Second GitHub Action does not use latest commit
Our current GitHub Actions workflow is not running correctly when the formatter fixes linting errors

## Expected Behavior

The job `format_code` should run and commit any changes made by the Black formatter. Then the `lint_test_type_check` job should run on the commit made by `format_code` (if there is one).

## Current Behavior

`lint_test_type_check` runs on the commit that triggered the workflow run, even when the workflow made a commit with formatting.

## Steps to Reproduce

1. Make a new branch
2. Add a trivial error which will be caught by Flake8 and which could be fixed by running Black. (I like to add an extra blank line in a Python file - this usually does the trick).
3. Commit the error
4. Push the branch and make a PR for it.
5. Wait for the GitHub Action to finish running.
6. Click on the "Details" of the workflow run, and see that even though Black reformatted the code, Flake8 found an error.

## Possible Implementation

I found this issue: https://github.com/actions/checkout/issues/461, which has a fix and _should_ work for us.

ew3361zh:
Corrects the second job of the GH action to get the latest commit to the PR after the first job has pushed the Black-formatted changes.

### Context

Issue(s) closed by this pull request: closes #1862

- [ ] Yes
- [x] No

### What

Updates the second GH Action job to correctly find the latest commit to the PR its running on.

### Why

Currently the second GH Action job is grabbing the original commit to the PR which triggered the GH actions to run. The issue with this is that the commit from the first job where the code is formatted with Black is ignored. Therefore linting issues in the PR that were corrected by Black in the first job are still found by the second job and the Action can potentially fail.

### How

The commit made by the `Format code with Black and commit changes` step of the first job is now explicitly output by the first job. This makes it easier for the second job to find that commit.

### Test plan

Branch off this PR and make a small flake8 error to the code (an extra space at the end of a line e.g.) and commit it to the new PR. The first job should use Black to fix this error and commit it. The second job should now find this last commit and successfully pass all of its steps (most importantly the flake8 check).

I also tested this with a commit that changed a part of the code that Black does not format (docstrings) and this was still able to find and match the most recent commit.

I've added a debugging check step `Compare with pushed commit SHA from format_code job` in the second job that checks the commit hash from the format code with Black step in the first job against the commit hash searched for in the second job as the "latest commit". These should match. If they do not, we know we will have an issue.

### DIFF
--- a/.github/workflows/combined_format_lint_test_mypy.yml
+++ b/.github/workflows/combined_format_lint_test_mypy.yml
@@ -109,27 +109,6 @@ jobs:
           persist-credentials: false
           fetch-depth: 0
 
-      - name: Fetch latest commit SHA
-        id: fetch-latest-sha
-        run: |
-          BRANCH_NAME="${{ github.head_ref || github.ref_name }}"
-          echo "Branch name: $BRANCH_NAME"
-          LATEST_SHA=$(git rev-parse origin/"$BRANCH_NAME")
-          echo "Latest SHA: $LATEST_SHA"
-          echo "LATEST_SHA=$LATEST_SHA" >> $GITHUB_ENV
-      
-      - name: Compare with pushed commit SHA from format_code job
-        run: |
-          echo "Commit hash from format_code job: ${{ needs.format_code.outputs.pushed_commit_sha }}"
-          echo "Commit hash fetched in this job: $LATEST_SHA"
-
-          if [ "${{ needs.format_code.outputs.pushed_commit_sha }}" = "$LATEST_SHA" ]; then
-            echo "The commit hashes match."
-          else
-            echo "The commit hashes do not match."
-            exit 1
-          fi
-
       - name: Checkout the latest commit
         uses: actions/checkout@v4
         with:
--- a/.github/workflows/combined_format_lint_test_mypy.yml
+++ b/.github/workflows/combined_format_lint_test_mypy.yml
@@ -85,6 +85,16 @@ jobs:
           git remote set-url origin https://${GITHUB_TOKEN}@github.com/${{ github.repository }}.git
           CLEAN_BRANCH_NAME=${BRANCH_NAME/refs\/heads\//}
           git push -f origin HEAD:$CLEAN_BRANCH_NAME
+      
+      - name: Capture and save the latest commit SHA after Black formatting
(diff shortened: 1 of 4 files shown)


.github/workflows/combined_format_lint_test_mypy.yml
```diff
@@ -106,11 +117,23 @@ jobs:
...
+          else
+            echo "The commit hashes do not match."
+            exit 1
+          fi
 
       - name: Checkout the latest commit
         uses: actions/checkout@v4
         with:
-          ref: ${{ steps.fetch-latest-sha.outputs.sha }}
+          ref: ${{ needs.format_code.outputs.pushed_commit_sha }}
```
Isgotkowitz, line 115: Since this action is now using `pushed_commit_sha` from the last job, isn't the "Fetch latest commit SHA" job no longer necessary? I think it is not necessary to compare it with the commit from the last job, because with your changes we can be very confident that the right commit is ultimately being collected and used.

RUFAS/report_generator.py
```diff
@@ -729,7 +729,7 @@ def _apply_horizontal_aggregation(
         Parameters
         ----------
         report_data : Dict[str, List[float]]
-            The data pool to be aggregated, structured as a dictionary of lists.
+            The data pool to be aggregated; structured as a dictionary of lists.
```
Isgotkowitz: Does this need to be reverted?

Isgotkowitz (approved):
LGTM, tried it out on a dummy branch and it worked as expected. Nice job Niko!

I do have one suggestion, which is to remove the double check from the second job that compares the commit hashes. However this is a take it or leave it suggestion.

.github/workflows/combined_format_lint_test_mypy.yml
ew3361zh: I was thinking it could be helpful during the review process but will remove before merging in.

allisterakun (approved):
Looks great, good work, Niko! I don't have further issues besides Ed's comment!

I tried it on a dummy PR, and it works as expected. See the screenshot below!
image

Issue #1625, opened 2024-05-08
Disentangle the Manure and Crop & Soil modules


Isgotkowitz:
The connection between the Manure module and the Crop and Soil module should be disentangled so that they meet the requirements outlined in #1622.

## Suggested Solution

The method `_check_manure_application_schedule` in `Field` should be moved from `manage_field` into its' own wrapper method (or just made public), and should return a `NutrientRequestResults` containing requested manure (if any) to `FieldManager`. This request is then returned by `FieldManager` to `SimulationEngine`, which passes it to `ManureManager`'s `request_nutrients` method. That method will return `NutrientRequestResults` which should be passed back into `FieldManager`. It might be easiest to pass the `NutrientRequestResults` into `FieldManager` through `daily_update_routine`, which then routes the different manure applications to different `Fields` using a mapping based on the field's name.

The implementations of `NutrientRequest` and `NutrientRequestResults` should be moved into a new directory that is not in either the Crop and Soil module or the Manure module. A new subdirectory may be needed for this.

The Crop and Soil module will request manure from the `FieldManureSupplier` (instead of the `ManureManager`) when no animals are simulated, and this behavior must be preserved. One of the easiest ways of doing this might be to simply move the `FieldManureSupplier` into the Manure module and have the `ManureManager` manage the logic how manure applications are created.

## Current Behavior

The Crop and Soil module's `FieldManager` takes the Manure module as an argument at initialization and keeps it as an attribute of `Field` instances when animals are simulated. When a manure application needs to be created (and animals are being simulated), the Manure Manager is called to construct a manure application in `Field._execute_manure_application`.

Isgotkowitz:
Grooming session: @adhikaripradip75 we (the dev team) would like to get your input on how the Manure and Crop & Soil modules should be connected.

adhikaripradip75:
Hello Ed, Thanks for question.
Today I am at Marshfield and the internet connection is poor and I am not
able to access slack properly.
By the way what is coming out from the manure module: solid manure (kg),
liquid manure (L) and slurry too ??.

There will be a couple of options for the manure that is coming out of the
dairy to manure storage and storage to the cropping fields.
1. 100% applied to the fields.
2. Many dairy producers export portion to the third party and apply
the remaining portion in the crop fields.
3. Some dairy imports manure too.
4. 100% exported to the third party

Many dairy farmers have multiple crop fields, they only apply manure to the
[...]

PR #1940, opened 2024-08-16
Move `FieldManureSupplier` to Manure Module

Isgotkowitz:
Title

### Context

Issue(s) closed by this pull request: lays the groundwork for #1625 

- [ ] Yes
- [x] No

### What

The `FieldManureSupplier` is being moved from the Crop and Soil module to the Manure module - but importantly, it does not change how it is used by the Crop and Soil module, this will happen in a follow up PR.

### Why

When the Crop/Soil and Manure modules are disentangled, all manure requests and fulfillments should be passed through the same mechanism - in SimulationEngine. Moving the FieldManureSupplier in the Manure module ensures this will always happen, regardless of whether or not animals are being simulated.

### How

Using `git mv`, the FieldManureSupplier (and its tests) has been moved to the Manure module. It is set up to be used by the Manure Manager in the same way it is used in the Field Manager. All import statements for it have been updated.

### Test plan

The unit test suite for the Manure manager has been updated and written to ensure that it functions as expected.

Also the below filter has been used on both this branch and dev branch to ensure that the behavior of manure request fulfillments remains the same: 
```json
{
    "multiple": [
        {
            "name": "apps",
            "filters": [
                ".*manure_application\\.field.*"
            ]
        }
    ]
}
```

## Update
I have completely removed the `FieldManureSupplier` from the Crop and Soil module, and all manure that is requested by the Crop and Soil module gets routed through `ManureManager.request_nutrients`. This was achieved by removing `_get_manure_supplier` from `FieldManager`, as well as all import statements of `FieldManureSupplier` in the Crop and Soil module.  I realized this was an important change to make because otherwise there would have been duplicate code for deciding when to use `FieldManureSupplier`, and that is not a state the code should ever be left in. The test strategy remains the same, I ran the default simulation on this branch and on `dev` with `simulate_animals` set to false and they had the same manure applications applied - as expected!


RUFAS/routines/manure/manure_manager.py
```diff
@@ -9,6 +9,7 @@
 from RUFAS.time import Time
 
 from ...data_structures.pen_manure_data import PenManureData
+from ...output_manager import OutputManager
```
ew3361zh:
```suggestion
from RUFAS.output_manager import OutputManager
```

I believe we want to get away from these relative import statements when possible.

ew3361zh:
I think this looks good, Ed. Had one comment about an import statement.

My only hesitation to approve is confusion on whether this is supposed to close out #1625 (it says "lays the groundwork for 1625"). If not, are you working on a second PR that will address the rest of it this sprint?

Isgotkowitz:
> I think this looks good, Ed. Had one comment about an import statement.
> 
> My only hesitation to approve is confusion on whether this is supposed to close out #1625 (it says "lays the groundwork for 1625"). If not, are you working on a second PR that will address the rest of it this sprint?

No, this PR should not close out #1625. Yes, I will be working on that second PR soon! I thought it better to start with just this, because moving the FieldManureSupplier is a pretty easy change to make in a very self contained way. I haven't started the second PR yet, but if you would prefer to see the final result before approving this I can move it up my list.

RUFAS/routines/manure/manure_manager.py
Isgotkowitz: Hmm we should ask @PooyaHekmati about this, I was under the impression they were a best practice but I'm not sure.

ew3361zh:
> > I think this looks good, Ed. Had one comment about an import statement.
> > My only hesitation to approve is confusion on whether this is supposed to close out #1625 (it says "lays the groundwork for 1625"). If not, are you working on a second PR that will address the rest of it this sprint?
> 
> No, this PR should not close out #1625. Yes, I will be working on that second PR soon! I thought it better to start with just this, because moving the FieldManureSupplier is a pretty easy change to make in a very self contained way. I haven't started the second PR yet, but if you would prefer to see the final result before approving this I can move it up my list.

No need. This write-up here is sufficient. Just wanted to make sure it wasn't closing the issue with some outstanding pieces to be completed. Thanks, Ed.

ew3361zh (approved):

RUFAS/routines/manure/manure_manager.py
PooyaHekmati: @Isgotkowitz I'm curious to know the reason.

Isgotkowitz: @PooyaHekmati in the case that the path to module is modified (something gets renamed), the number of import statements that have to be modified is minimized.

PooyaHekmati: I don't think so. The relative path to the module is different depending on where you are importing it from. Therefore, when it needs modification, multiple things have to be modified if the paths are relative. But if all the paths to OM are absolute, then it is as easy as bulk replacement. 
Also, this is anti-pattern, which is bad.

Isgotkowitz: Noted! I will make this absolute then, thanks Pooya

RUFAS/routines/manure/manure_manager.py
```diff
@@ -85,10 +87,18 @@ class ManureManager:
         A dictionary that maps an animal pen's id to a ManureSeparator object.
     manure_treatments : Dict
         A dictionary that maps an animal pen's id to a Treatment object.
+    animals_are_simulated : bool
+        Records whether animals are being simulated.
```
allisterakun: I remember that this variable is called `simulate_animals` in the simulation engine. Do we want to rename it to keep consistency?

RUFAS/simulation_engine.py
```diff
@@ -173,6 +173,9 @@ def _initialize_simulation(self) -> None:
 
         self.animal_manager = AnimalManager(animal_class_config, self.feed, self.weather, self.time)
         all_pen_manure_data = self.animal_manager.collect_pen_manure_data()
-        self.manure_manager = ManureManager(all_pen_manure_data, self.weather, self.time, manure_class_config)
+        are_animals_simulated = self.im.get_data("config.simulate_animals")
```
allisterakun:
```suggestion
        are_animals_simulated: bool = self.im.get_data("config.simulate_animals")
```

allisterakun (approved):
Looks good, nice work, Ed! Quite straight-forward changes of import statements. Left two simple comments, otherwise looks good!

RUFAS/routines/manure/manure_manager.py
Isgotkowitz: Good idea, I will put that in. Food for thought though, maybe `simulate_animals` would be better as the parameter name and `animals_are_simulated` would be more appropriate as the attribute name. The reason is that `simulate_animals` sounds like an imperative i.e. it is telling the code whether or not to do something, while `animals_are_simulated` sounds more reactive, like it is passively accepting information. In the context of the manure module it makes a little more sense (to me) that it would passively accept whether or not animals are simulated, because it has no control over whether animals are simulated.

RUFAS/routines/field/manager/field_manager.py
```diff
@@ -56,10 +55,8 @@ def __init__(self, manure_manager: ManureManager, feed_manager: FeedManager):
         if not fields:
             self.om.add_warning("No field input files.", "No fields will be simulated.", info_map)
 
-        manure_supplier = self._get_manure_supplier(manure_manager)
-
         for field in fields:
-            new_field = self._setup_field(field, manure_supplier, feed_manager)
+            new_field = self._setup_field(field, manure_manager, feed_manager)
```
ew3361zh, line 59: Curious about the choice of using `manure_manager` here and `manure_supplier` everywhere else.

RUFAS/routines/manure/manure_manager.py
```diff
@@ -117,6 +129,15 @@ def __init__(
...
         self._manure_nutrient_manager = ManureNutrientManager()
+        self.simulate_animals = simulate_animals
+        if not self.simulate_animals:
+            info_map = {"class": self.__class__.__name__, "function": "__init__"}
+            OutputManager().add_log(
+                "Animals not being simulated",
+                "Manure for field applications will be created by the FieldManureSupplier",
+                info_map,
+            )
+        self._field_manure_supplier = FieldManureSupplier()
```
ew3361zh: What do you think about indenting 140 (`self._field_manure_supplier = FieldManureSupplier()`) so it only gets instantiated `if not self.simulate_animals` is true?

No reason to create it if it's not needed.

RUFAS/routines/field/manager/field_manager.py
Isgotkowitz: No good reason in particular - I just updated it to consistently use `manure_manager` instead of `manure_supplier`. This wasn't a significant change, but did add a lot of lines changed because a lot of keyword args had to be updated in tests

allisterakun (approved):
Good job, Ed! LGTM.

I tested setting `simulate_animals` to false, and the program behaved as expected. 
All unit tests are appropriately updated to cover all the changes made in this PR:
```
Name                                                                Stmts   Miss  Cover   Missing
-------------------------------------------------------------------------------------------------
.\RUFAS\routines\field\field\field.py                                 426      1    99%   1238
.\RUFAS\routines\field\manager\field_manager.py                       153      0   100%
.\RUFAS\routines\manure\field_manure_supplier.py                       35      0   100%
.\RUFAS\routines\manure\manure_manager.py                             167      0   100%
.\RUFAS\simulation_engine.py                                           80      3    96%   83, 152, 155
.\tests\soil_crop_tests\field_tests\test_field.py                     762      2    99%   1028-1029
.\tests\soil_crop_tests\field_tests\test_tillage_application.py        70      0   100%
.\tests\soil_crop_tests\manager_tests\test_field_data_reporter.py     139      0   100%
.\tests\soil_crop_tests\manager_tests\test_field_manager.py           186      0   100%
.\tests\test_manure_module\test_field_manure_supplier.py               14      0   100%
.\tests\test_manure_module\test_manure_manager.py                     338      0   100%
.\tests\test_simulation_engine.py                                     151      0   100%
-------------------------------------------------------------------------------------------------
TOTAL                                                                2521      6    99%
```

RUFAS/routines/manure/manure_manager.py
Isgotkowitz: Good idea. I'm not sure if this is a best practice to only init an attribute sometimes, but in this case it seems pretty safe and no of the static analysis tools have a problem with it, so I'll go for it.

PR #1943, opened 2024-08-20
Fixing global OM (1/2)

### LINKED ISSUE #1848: [OM] Output manager initialized as global all across the codebase(1/2)

matthew7838:
This PR removes the global initialization of OM in files outside routines, feed, and soil.
### Context

Issue(s) closed by this pull request: closes #1848 

- [ ] Yes
- [x] No

### What

- Removes global OM initialization.
- Initiates OM in static methods or constructors
- Update the tests' mockings.

### Why

- We want to avoid global variables and performance won't be much affected since OM is singleton.

### How

- Removes the global OM.

### Test plan

- Updates and runs test suites on modified code.

### DIFF
--- a/RUFAS/routines/field/soil/snow.py
+++ b/RUFAS/routines/field/soil/snow.py
@@ -3,7 +3,6 @@ from typing import Optional
 
 from RUFAS.routines.field.soil.soil_data import SoilData
 from RUFAS.current_day_conditions import CurrentDayConditions
-from RUFAS.output_manager import OutputManager
 
 
 class Snow:
--- a/RUFAS/routines/field/soil/snow.py
+++ b/RUFAS/routines/field/soil/snow.py
@@ -5,8 +5,6 @@ from RUFAS.routines.field.soil.soil_data import SoilData
 from RUFAS.current_day_conditions import CurrentDayConditions
 from RUFAS.output_manager import OutputManager
 
-om = OutputManager()
-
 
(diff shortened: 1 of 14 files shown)


RUFAS/routines/field/soil/snow.py
```diff
@@ -43,6 +41,7 @@ class Snow:
     def __init__(self, soil_data: Optional[SoilData] = None, field_size: Optional[float] = None):
         """object that tracks all soil variable throughout the simulation"""
         self.soil_data = soil_data or SoilData(field_size=field_size)
+        self.om = OutputManager
```
Isgotkowitz: Since the output manager isn't used at all in this module, I don't think it needs to be present at all.

Isgotkowitz (approved):
One minor comment, overall LG8TM!

allisterakun (approved):
Straight-forward changes, LGTM! Good job, Matthew! 
Just one minor thing:
Could you update the PR title as it updates the global OM, not the global IM?

PR #1944, opened 2024-08-20
Update Sphinx GitHub Action

### LINKED ISSUE #1886: [GH Actions] Update the Sphinx GitHub Action
The Sphinx GitHub Action needs to be updated so that it will run correctly.

## Status

The Sphinx GitHub Action is currently disabled and has not been run in a while. As a result, it no longer runs when triggered manually.

## Intention

A working Sphinx GitHub Action that can be triggered manually.

## Request

Work on the Sphinx GitHub Action until it runs properly again.

### Additional Information
It should be investigated when exactly this action should be run automatically. We may want it run automatically on any PR that is going from `test` to `main` but only run on any other PR if it is manually triggered, for example.

Isgotkowitz:
Updates the Sphinx workflow so that it can be manually triggered.

### Context

Issue(s) closed by this pull request: closes #1886 

Wiki:
- [x] Updated Wiki (will do this when I merge/am just about to merge)

### What

Updates the Sphinx GitHub Actions workflow so that it will not run automatically on PRs that are going into main. Instead devs will have the option to trigger it manually.

Also removes the steps from the Action that are not strictly related to rebuilding the html documentation.

### Why

In accordance with our versioning policy, we only want the documentation updated at specific times (usually when there is a major update). Having the workflow triggered manually makes this easier to realize.

The automatic removal and regeneration of .rst files has been removed, because this was removing .rst files that cannot be automatically regenerated.

### How

Removed the `push` and `pull_request` values, this prevents the workflow from running automtically. Also remove the tasks which remove rst files and regenerate them - see why section for more details.

### Test plan

Create a duplicate of this branch, and make a PR with it that was pointed towards `main` branch. Go to the "Actions" tab (same level as Issues, PRs, etc.), then click "Build Sphinx" in the left-hand column of the page. Hit "Enable workflow" (if you try this it's a good idea to mention that you are doing this in Slack, and don't forget to disable it when you're done testing). Now, hit "Run workflow", select target branch for rebuilding documentation, then hit the green "Run workflow".

Then the docs are rebuilt! Navigate back to the PR associated with the target branch, and see that the workflow updates the documentation and commits it. Then pull the branch so that the documentation changes can be seen locally. Open up the `index.html` page of the documentation (`docs/_build/html/index.html`) in a web browser and poke around the docs. To me they looked mostly correctly, in that it had both the code documentation and some scientific documentation. There are probably some updates that need to be made to the .rst files, but I think those should be made when we get a better understanding of what we are looking for in the documentation.


ew3361zh (approved):
Great job, Ed!

As far as I could tell this worked great and had the new updated structure of RUFAS which is wonderful to have.

Scientific documentation looked pretty official and scientific.

I do think the actual launch with this documentation would benefit from an exhaustive review to ensure things look as they should but as far as this script being a tool to help us achieve that goal, this seems good to go.

I think updating the Sphinx actions Wiki would be helpful.

matthew7838 (approved):
LGTM tested it and the updated docs seem right. Sorry for taking so long.

PR #1945, opened 2024-08-20
AnimalGenetics implementation

### LINKED ISSUE #1696: [Animal][AnimalGenetic] Genetic Value assignment for animals loaded from input JSON file and Replacement animals entering the herd
Blocked by #1695. 

Assign genetic values for animals loaded from the input JSON file and replacement animals entering the herd.

Follow the instructions in the [AnimalPrescibeFix](https://docs.google.com/document/d/1f231Koa7dBlsBTAXD6L1qBhZ_F6ndl9uy63CSeud4h8/edit#heading=h.lf57q7ulfzse) document to implement the logic for genetic value assignment for animals loaded from the input JSON file and replacement animals entering the herd.

Implement the function to find out a future/past value based on the current data using linear regression estimation. Utilize this function to estimate the mean value required by the genetic value assignment calculation if the birth year of the animal is not represented in the current data.

**Target Deadline**
By when should this issue be done.

### LINKED ISSUE #1697: [Animal][AnimalGenetic] Calculate Genetic Value for Newborn Calves
Blocked by #1695 & #1696. 

Calculate the genetic value for the newborn calves during the simulation.

Follow the instructions in the [AnimalPrescibeFix](https://docs.google.com/document/d/1f231Koa7dBlsBTAXD6L1qBhZ_F6ndl9uy63CSeud4h8/edit#heading=h.lf57q7ulfzse) document to implement the logic for calculating the genetic value of the newborn calves during the simulation.

Utilize the linear regression estimation function to estimate the mean value required by the genetic value assignment calculation if the birth year of the animal is not represented in the current data.

**Target Deadline**
By when should this issue be done.

allisterakun:
This PR implements the assignment of the net merit values to all animals in RuFaS.
# Context

Issue(s) closed by this pull request: closes #1696 and #1697.

- [ ] Yes
- [x] No

# What

### Net Merit
The Net Merit input is located at `input/data/animal/animal_genetics/NetMerit_HO.csv`. This CSV file has three columns: `["year_month", "average", "std"]`. Every row is an entry that contains a mean and standard deviation pair for a single month. The entries range from 2006-02 to 2024-08, with three entries per year. 

This input file is constructed from multiple percentile lookup tables, one for each entry. Each percentile table contains the net merit value for every integer percentile from 0-100. These tables are first combined into a CSV with the columns representing the different entries and rows indicating the corresponding percentile net merit values: 
[NetMerit_percentile_HO.csv](https://github.com/user-attachments/files/16773970/NetMerit_percentile_HO.csv)

For each entry, I generated a random population of 1,000,000 values from the percentile lookup table and found the mean and standard deviation to represent the population. The final [NetMerit_HO.csv](https://github.com/user-attachments/files/16774043/NetMerit_HO.csv) is constructed by combining the population summary for all entries.

In the `AnimalGenetics` class, this data is loaded from the InputManager and saved as a class property `AnimalGenetics.net_merit` in the structure of `dict[str, dict[str, dict[str, float]]]`:
```JSON
{
  "breed": {
    "%Y-%m": {
      "average": float,
      "std": float
    }
  }
}
```
where
- "breed" is the breed of the animal, e.g., "HO".
- "%Y-%m" represents the month of the entry, e.g., "2008-08".

#### Base Change
The CDCB reevaluates and adjusts the net merit value to keep it within a reasonable range by changing the baseline every five years. Therefore, to realign the past data with the current values, we change every value to match the oldest value and shift all data back into the reasonable range. Here is a lookup table for the base change date and value:
```JSON
{
    datetime.date.fromisoformat('2020-04-01'): 231,
    datetime.date.fromisoformat('2014-12-01'): 184,
    datetime.date.fromisoformat('2010-01-01'): 132,
    datetime.date.fromisoformat('2005-01-01'): 0
}
```

This action is performed after loading the net merit data from the `InputManager` in `AnimalGenetics.__init__()`. The logic is written in the `AnimalGenetics.net_merit_base_change()` function.

#### Fill Gap
The input net merit data only has three entries per year; therefore, we need to fill in the gap between entries using linear approximation. This action is performed after loading the net merit data from the `InputManager` in `AnimalGenetics.__init__()`. The logic is written in the `AnimalGenetics.net_merit_fill_gap()` function.

### PTA values for Top Semen Listing
The top semen listing data is queried from the [CDBC database for Top Animal Listing](https://webconnect.uscdcb.com/#/top-animal-listing). There are only 17 available data points from 2019-04 to 2024-08. After the base change action, we used linear approximation to fill in the gap and extend the data to cover every month from 2006 to 2024. 

### Animal Entering the Herd
This is for animals loaded from the saved data during simulation initialization and for animals bought during the simulation.
The value is a random value generated from a normal distribution with a mean and a standard deviation value.
We first get the current date (the simulation start date or the date the animal is bought) and the animal's age. From there, we can figure out the animal's birthdate. Then, we can find the corresponding mean and standard deviation from the net merit data. Finally, with the mean and standard deviation, we can return the generated random value drawn from the normal distribution.
This logic is written in `AnimalGenetics.assign_net_merit_value_to_animals_entering_herd()`.

### Newborn Calf
This is for the newborn calves born during the simulation.
The value is a random value generated from a normal distribution with a mean and a standard deviation value.
With the newborn calf's birthdate, we can find the top listing semen value for the corresponding month and the standard deviation for the net merit value. Then, we can calculate the mean and standard deviation used to construct the normal distribution:
#### mean
mean = top_listing_semen_PTA + dam_net_merit_value
#### standard deviation
standard_deviation = sqrt(Mendelian sampling variance) = sqrt(1/2 * (net_merit_std) **2)

This logic is written in `AnimalGenetics.assign_net_merit_value_to_newborn_calf()`.

### Animal.net_merit
I have added the net merit value as a new attribute to the animal classes and as one of the fields in the arguments to initialize an animal object.
### Animal.birth_date
I have changed the `Animal.birth_date` type from an integer to a string to represent the calendar date when the animal was born.
### HerdFactory
`HerdFactory` is updated to assign a net merit value when loading the animals from input or generating animals with simulation.
### LifeCycle
The `LifeCycleManager._handle_new_born()` is updated to calculate and assign the net merit value for the newborn calf.
### Simulation Dates
I have changed the default simulation date to range from 2012:244 to 2018:264 because of the constraint of the net merit data.
The current default simulation start date is in 2009. Multiple cows in the saved `animal_population.json` input are over 2000 days old; the oldest cow is 2282 days old. This would put the birthdate for those cows in 2003. However, the earliest net merit data is in 2006-02.

# Test plan

All affected unit tests are updated, and I have written unit tests for the new `AnimalGenetics` class. 
@YijingGong will help with the Subject Matter Expert evaluation.

# Utility Scripts/Functions
Script to generate a population of 1,000,000 data points from the percentile lookup CSV and generate the population summary (also graphs for all entries)
```python
import random
from pathlib import Path

import numpy as np
import pandas as pd
import matplotlib.pyplot as plt
import scipy
import seaborn as sns

result = pd.DataFrame()
res_summary = pd.DataFrame(columns=["year_month", "average", "std"])

im = InputManager()
net_merit_csv = Path("NetMerit_percentile_HO.csv")
net_merit = im._load_data_from_csv(net_merit_csv)
population_graph_path = "population_graph/"

for date in net_merit.keys():
    if date == "percentile":
        continue
    rand = [random.randint(0, 100) for _ in range(1000000)]
    rand_nm = [net_merit[date][100 - rand_percentile] for rand_percentile in rand]

    mean = np.mean(rand_nm)
    std = np.std(rand_nm)

    result.insert(len(result.columns), date, rand_nm)
    res_summary.loc[len(res_summary.index)] = [date, mean, std]

    fig, axes = plt.subplots(3, 1, sharex=True)
    fig.set_figheight(8)
    fig.set_figwidth(12)

    sns.histplot(rand_nm, ax=axes[0], bins=101, kde=True)
    axes[0].set_title("Population Distribution")
    xmin, xmax = min(rand_nm), max(rand_nm)
    x = np.linspace(xmin, xmax, 101)
    p = scipy.stats.norm.pdf(x, mean, std)
    scaled_p = [val * 25000000 for val in p]
    axes[0].plot(x, scaled_p, 'r')

    axes[1].hist(net_merit[date], bins=101)
    axes[1].set_title("Percentile Distribution")

    axes[2].plot(x, p, 'r')
    axes[2].set_title("PDF")
    plt.savefig(population_graph_path + date+'.png')
    plt.close()

result.to_csv("rand_nm_population.csv", index=False)
res_summary.to_csv("rand_nm_summary.csv", index=False)

```

Script to change all birth_date entry in `animal_population.json` from `0` to `""`
```python
import json

from typing import Any


def update_birth_date(data_dict: dict[str, Any]) -> dict[str, Any]:
    data_dict.update(birth_date="")
    return data_dict


with open('input/data/animal/animal_population.json') as json_file:
    data: dict[str, list[dict[str, Any]]] = json.load(json_file)

data["calves"] = list(map(update_birth_date, data["calves"]))
data["heiferIs"] = list(map(update_birth_date, data["heiferIs"]))
data["heiferIIs"] = list(map(update_birth_date, data["heiferIIs"]))
data["heiferIIIs"] = list(map(update_birth_date, data["heiferIIIs"]))
data["cows"] = list(map(update_birth_date, data["cows"]))
data["replacement"] = list(map(update_birth_date, data["replacement"]))

with open('input/data/animal/animal_population2.json', "w") as outfile:
    json.dump(data, outfile)
```

Functions to plot and save the net merit data in `AnimalGenetics`
```python
    def plot_net_merit(self, filename: str, width: float, title: str) -> None:
        year_months = list(self.net_merit["HO"].keys())
        means = [self.net_merit["HO"][year_month]["average"] for year_month in self.net_merit["HO"].keys()]
        stds = [self.net_merit["HO"][year_month]["std"] for year_month in self.net_merit["HO"].keys()]

        plt.title(title)
        plt.figure(figsize=(width, 7))
        plt.errorbar(year_months, means, fmt='.', yerr=stds)
        plt.xticks(rotation=90)
        plt.savefig(filename)
        plt.close()

    def save_net_merit_data_to_csv(self) -> None:
        year_months = list(self.net_merit["HO"].keys())
        df_net_merit = pd.DataFrame(columns=["year_month", "average", "std"])
        for i in range(len(year_months)):
            year_month = year_months[i]
            df_net_merit.loc[i] = [year_month,
                                   self.net_merit["HO"][year_month]["average"],
                                   self.net_merit["HO"][year_month]["std"]]
        df_net_merit.to_csv("updated_net_merit_HO.csv", index=False)
```


RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,153 @@
...
+import matplotlib.pyplot as plt
+import datetime
+
+base_change_lookup_table = {
+    datetime.date.fromisoformat('2020-04-01'): 369,
+    datetime.date.fromisoformat('2014-12-01'): 253,
+    datetime.date.fromisoformat('2010-01-01'): 140,
+    datetime.date.fromisoformat('2005-04-01'): 107,
+    datetime.date.fromisoformat('2000-04-01'): 0
+}
```
YijingGong, line 22: Probably I confused you, I know these values are from the pdf that I shared with you. But they are from another source, which is temporary. And I put the updated values here: https://docs.google.com/spreadsheets/d/1T0SPjUQ1jAWafi3XkbAFzo0c32Zk3mvYv56TvbhA7wY/edit?gid=782996058#gid=782996058, along with the changing time. Could you please update them?

YijingGong: Reference (in case any documentation is needed): VanRaden et al., 2010; VanRaden et al., 2014; Norman et al., 2020
accessed from the following websites:
https://aipl.arsusda.gov/reference/base2010.htm

https://aipl.arsusda.gov/reference/base2014.htm

https://uscdcb.com/wp-content/uploads/2020/02/Norman-et-al-Genetic-Base-Change-April-2020-FINAL_new.pdf

YijingGong: This change will affect 1) semen linear regression, 2) fill gaps of the cows' NM. But we shouldn't need to rerun the million sampling, right?

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,153 @@
...
+    def net_merit_fill_gap(
+            original_net_merit: dict[str, dict[str, dict[str, float]]]
+    ) -> dict[str, dict[str, dict[str, float]]]:
+        expanded_net_merit: dict[str, dict[str, dict[str, float]]] = {}
+        monthly_increase_lookup = {
+            2005: 140 / 60,
+            2010: 253 / 60,
+            2015: 369 / 60,
+            2020: (538 - 396) / 24
+        }
```
YijingGong: should we use the base_change_lookup_table here, instead of hard-coding?

RUFAS/routines/animal/genetics/animal_genetics.py
YijingGong: Maybe this will fix the weird jump in 2019-Dec

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,153 @@
...
+        return Utility.generate_random_number(average, std)
 
-    def assign_net_merit_value_to_newborn_calf(self, time: Time, dam_net_merit_value: float) -> float:
-        pass
-        return 0.0
+    def assign_net_merit_value_to_newborn_calf(self, time: Time, breed: str, dam_net_merit_value: float) -> float:
+        birth_year_month = str(time.current_calendar_year) + "-" + str(time.current_month).zfill(2)
+        semen_PTA: float = self.top_semen[breed][birth_year_month]
+        semen_net_merit = semen_PTA * 2
+        average_net_merit = (semen_net_merit + dam_net_merit_value) / 2
```
YijingGong: I took a look at the CDCB website again and realize that they always report the PTA (predicted transmitting ability) as the measurement of NM, not EBV (estimated breeding value). While EBV = 2 x PTA. I told you to multiply 2 here.  I was wrong. 
So here it should be:
calf net merit = semen PTA (from the semen NM data) + dam PTA (from the cow NM data) 
Nothing needs to be multiplied 2, and nothing needs to divided by 2.

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,153 @@
...
+        plt.close()
 
-    def assign_net_merit_value_to_animals_entering_herd(self, birth_date: str) -> float:
-        pass
-        return 0.0
+    def assign_net_merit_value_to_animals_entering_herd(self, birth_date: str, breed: str) -> float:
+        birth_year_month = birth_date[:7]
+        average = self.net_merit[breed][birth_year_month]["average"]
+        std = self.net_merit[breed][birth_year_month]["std"]
+        return Utility.generate_random_number(average, std)
```
YijingGong, line 202: This looks like a function generation a value based on Normal distribution. This name is a bit confusing - giving me the impression of it is just generating a random number.

YijingGong (commented):
I left some comments based on just looking at the codes. I'm not sure how can I run and see if everything works well. Do you have any suggestions?

RUFAS/routines/animal/animal_typed_dicts.py
```diff
@@ -48,6 +48,7 @@ class AnimalBaseInitArgsTypedDict(TypedDict):
     pen_history: NotRequired[List]
     conceptus_weight: NotRequired[float]
     calf_birth_weight: NotRequired[float]
+    nei_merit: NotRequired[float]
```
Isgotkowitz: Could you please add a docstring that includes `nei_merit`?

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,211 @@
...
+from RUFAS.util import Utility
+
+import datetime
+
+base_change_lookup_table = {
+    datetime.date.fromisoformat("2020-04-01"): 231,
+    datetime.date.fromisoformat("2014-12-01"): 184,
+    datetime.date.fromisoformat("2010-01-01"): 132,
+    datetime.date.fromisoformat("2005-01-01"): 0,
+}
```
Isgotkowitz: Could you please add a d
```suggestion
BASE_CHANGE_LOOKUP_TABLE = {
    datetime.date.fromisoformat("2020-04-01"): 231,
    datetime.date.fromisoformat("2014-12-01"): 184,
    datetime.date.fromisoformat("2010-01-01"): 132,
    datetime.date.fromisoformat("2005-01-01"): 0,
}
```
Could you please add an explanation of what this table is and its source(s)?

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,211 @@
...
+        Parameters
+        ----------
+        original_net_merit: dict[str, dict[str, dict[str, float]]]
+            The original net merit data.
+
+        Returns
+        -------
+        dict[str, dict[str, dict[str, float]]]
+            The net merit data after base change.
+
```
Isgotkowitz: Based on what the link says would the units of these be lifetime $?

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,211 @@
...
+base_change_lookup_table = {
+    datetime.date.fromisoformat("2020-04-01"): 231,
+    datetime.date.fromisoformat("2014-12-01"): 184,
+    datetime.date.fromisoformat("2010-01-01"): 132,
+    datetime.date.fromisoformat("2005-01-01"): 0,
+}
 
 
 class AnimalGenetics:
     def __init__(self) -> None:
```
Isgotkowitz: Could you please add a class docstring with an attributes section? I think it will be really important/helpful to document how net_merit and top_semen are structured and what their units are

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,211 @@
...
-        # read and save a lookup table for top-semen and net-merit percentile
-        pass
+        """Get the net merit and top listing semen data from InputManager, save them as class attributes, and perform
+        base change and fill the gap for the net merit data."""
+        im = InputManager()
+        net_merit_HO: dict[str, list[str | float]] = im.get_data("animal_net_merit")
```
Isgotkowitz, line 43: How would this code change when more than just holsteins are simulated? Would it be possible to write it in a way that allows for easier extension of the code?

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,211 @@
...
+        top_listing_semen_HO: dict[str, list[str | float]] = im.get_data("animal_top_listing_semen")
+        self.net_merit: dict[str, dict[str, dict[str, float]]] = {
+            "HO": {
+                net_merit_HO["year_month"][i]: {"average": net_merit_HO["average"][i], "std": net_merit_HO["std"][i]}
+                for i in range(len(net_merit_HO["year_month"]))
```
Isgotkowitz, line 49: Is there a better variable name that could be used than i?

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,211 @@
...
+        original_net_merit: dict[str, dict[str, dict[str, float]]]
+            The original net merit data.
+
+        Returns
+        -------
+        dict[str, dict[str, dict[str, float]]]
+            The net merit data after filling the gap in between entries.
+        """
+        expanded_net_merit: dict[str, dict[str, dict[str, float]]] = {}
+        monthly_increase_lookup = {2005: 140 / 60, 2010: 253 / 60, 2015: 369 / 60, 2020: (538 - 396) / 24}
```
Isgotkowitz: I think this lookup dictionary or at least parts of it should be extracted into constants. It is very magic numbery right now

RUFAS/routines/animal/genetics/animal_genetics.py
Isgotkowitz: For the internal representation of net merit, what would you think about making the time key a full `datetime.date` instead of a string in the form of "YYYY-MM" (the day would always be set to 1 in the date instance)? There is a lot of string parsing in the year-month time representations, and I think it would be good to get away from that if possible.

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,211 @@
...
+        Notes
+        -----
+        With the birthdate and the breed of the animal, this function first looks up the top listing semen value.
+        The mean for the net merit value of the newborn can then be calculated as the sum of the top listing semen
+        value and the net merit value of the dam.
+        The standard deviation is the square root of the Mendelian sampling variance, which is simply half of the
+        population variance.
+        """
+        birth_year_month = str(time.current_calendar_year) + "-" + str(time.current_month).zfill(2)
+        semen_PTA: float = self.top_semen[breed][birth_year_month]
```
Isgotkowitz: Would it be worth it to catch a KeyError here, in case a breed is being used that isn't supported by the genetics module?

RUFAS/routines/animal/life_cycle/animal_base.py
```diff
@@ -50,7 +50,7 @@ def __init__(self, args: AnimalBaseInitArgsTypedDict):
         """
         self.id = args["id"]
         self.breed = args["breed"]
-        self.birth_date = args["birth_date"]
+        self.birth_date: str = args["birth_date"]
```
Isgotkowitz, line 54: Would it be better to store birth date as a datetime.date?

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,211 @@
...
+        return Utility.generate_random_number(average, std)
+
+    def assign_net_merit_value_to_newborn_calf(self, time: Time, breed: str, dam_net_merit_value: float) -> float:
+        """
+        This function calculates the net merit value for the newborn calves.
+
+        Parameters
+        ----------
+        time: Time
+            The Time instance that contains the birthdate of the newborn calf.
```
Isgotkowitz, line 212: Does the time instance contain the birthdate of the calf, or is it assumed that this method is being called on the day of a calf's birth so the Time instance currently holds the calf's birthday?

input/metadata/properties/default.json
```diff
@@ -865,9 +865,8 @@
           "pattern": "^(HO|JE)$"
         },
         "birth_date": {
-          "type": "number",
-          "description": "Birth date.",
-          "minimum": 0
+          "type": "string",
+          "description": "Birth date."
```
Isgotkowitz, line 869: Should there be a pattern for validating this, something like `"\b\d{4}-\d{2}-\d{2}\b"`

input/metadata/properties/default.json
```diff
@@ -1496,10 +1515,60 @@
...
       }
     }
   },
+  "animal_net_merit_properties": {
+    "year_month": {
+      "type": "array",
+      "description": "The birth year and month of the animal.",
+      "properties": {
+        "type": "string",
+        "description": "The birth year and month of the animal."
```
Isgotkowitz, line 1533: There should be a pattern to make sure these dates are correctly formatted, as well as a specification of the formatting in the description.

Isgotkowitz (commented):
This is an awesome start to AnimalGenetics Allister. However I have a lot of questions, and probably will have more as I learn more about genetics. The things I left comments on fall generally into 2 categories:
1. More documentation. This includes more explanations of stuff like net merit (I still really don't understand what it is or what it represents) as well as more units the in the existing documentation. This applies to both code and non-code (i.e. metadata).
2. More use of `datetime.date` when working with dates, instead of using strings. I'm not sure how this will fit into the science though.

Also I am curious, why make these changes in `routines` instead of `biophysical`?

RUFAS/routines/animal/genetics/animal_genetics.py
allisterakun: It's the later situation. This method will be called only on the day of giving birth; therefore, the birth date for the calf is the current date, which can be extracted from the Time instance.

RUFAS/routines/animal/life_cycle/animal_base.py
allisterakun: I will look into the complexity of changing the type of the birthdate.

RUFAS/routines/animal/genetics/animal_genetics.py
allisterakun: This table is actually supposed to be dynamically compiled. It's just one of the logic is a little complex. I will update this!

allisterakun:
> Also I am curious, why make these changes in `routines` instead of `biophysical`?

Because at the time of implementation, the `biophysical` folder is not set up yet.
I will incorporate this into the `biophysical` folder

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,211 @@
...
+        -----
+        With the birthdate and the breed of the animal, this function first looks up the top listing semen value.
+        The mean for the net merit value of the newborn can then be calculated as the sum of the top listing semen
+        value and the net merit value of the dam.
+        The standard deviation is the square root of the Mendelian sampling variance, which is simply half of the
+        population variance.
+        """
+        birth_year_month = str(time.current_calendar_year) + "-" + str(time.current_month).zfill(2)
+        semen_PTA: float = self.top_semen[breed][birth_year_month]
+        average_net_merit = semen_PTA + dam_net_merit_value
```
YijingGong: We will need to rename some of these. 
Net merit ($) is a type of breeding value measure. Predicted Transmitting Ability (PTA) is half of the breeding value.
As Kristan mentioned that it's better to have generalized naming, I think this equation could be something like: average_PTA = semen_PTA + dam_PTA
At this point, it's ok. I will keep an eye on this in our next version (and also see what people say tomorrow).

YijingGong: Also in the PR documentation. You really did a great job in documenting! Just a note that we may need to change some of it later.

YijingGong (approved):
Great job! Everything meets my expectations, if not exceeds. The PR description is also very clear and well-written.

RUFAS/routines/animal/genetics/animal_genetics.py
Isgotkowitz: > This method will be called only on the day of giving birth; therefore, the birth date for the calf is the current date, which can be extracted from the Time instance.

This should go in the docstring (or something very similar). With the current docstring, someone might interpret it as the time instance passed magically has the birthdate of the calf regardless of what day the function is actually called on.

RUFAS/routines/animal/animal_typed_dicts.py
```diff
@@ -48,6 +48,8 @@ class AnimalBaseInitArgsTypedDict(TypedDict):
     pen_history: NotRequired[List]
     conceptus_weight: NotRequired[float]
     calf_birth_weight: NotRequired[float]
+    net_merit: NotRequired[float]
+    """The net merit value that represents the genetic value of the animal, USD."""
```
Isgotkowitz: Instead of adding an inline comment, could you start a new docstring for the `AnimalBaseInitArgsTypedDict` with this description in the `Attributes` section? I don't think that this style of documentation is bad, but I think (and hope) that eventually we will have full docstrings for all the structures in RuFaS, and having them already started will nudge people towards adding to them as more work is done to the module.

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,248 @@
...
+from RUFAS.util import Utility
+
+import datetime
+
+"""
+The base change time and value in the net merit value.
+
+References
+----------
+CDCB website
```
Isgotkowitz, line 14: Instead of just listing the Reference as "CDCB website", I think it would be better to list the 3 links that Yijing put in her comment:
- https://uscdcb.com/wp-content/uploads/2020/02/Norman-et-al-Genetic-Base-Change-April-2020-FINAL_new.pdf
- https://aipl.arsusda.gov/reference/base2014.htm
- https://aipl.arsusda.gov/reference/base2010.htm

Isgotkowitz: Edit: I just saw that these links are listed in the `initialize_class_variables` method, but I think it would not be a bad idea to list them here as well.

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,248 @@
...
+    @staticmethod
+    def assign_net_merit_value_to_animals_entering_herd(birth_date: str, breed: str) -> float:
+        """
+        This function calculates the net merit value for animals entering the herd, either during initialization or
+        for animals bought during the simulation.
+
+        Parameters
+        ----------
+        birth_date: str
+            The birthdate of the animal in the format "%Y-%m-%d".
```
Isgotkowitz:
```suggestion
            The birthdate of the animal in the format "YYYY-MM-DD".
```
If this is the full format of the dates I think it would be better to specify it fully.

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,248 @@
...
+        Notes
+        -----
+        With the birthdate and the breed of the animal, this function first looks up the top listing semen value.
+        The mean for the net merit value of the newborn can then be calculated as the sum of the top listing semen
+        value and the net merit value of the dam.
+        The standard deviation is the square root of the Mendelian sampling variance, which is simply half of the
+        population variance.
+        """
+        birth_year_month = str(time.current_calendar_year) + "-" + str(time.current_month).zfill(2)
+        semen_PTA: float = AnimalGenetics.top_semen[breed][birth_year_month]
```
Isgotkowitz: Instead of `semen_PTA` I would consider making this variable name `semen_predicted_transmitting_ability`. It is substantially longer, but is much clearer.

RUFAS/routines/animal/genetics/animal_genetics.py
Isgotkowitz: Edit: after seeing how `strftime` is used I understand why it is written as "%Y-%m-%d" here. But because this is documentation and not code, I think using the above would be better because it is more readable and consistent with how date formats are documented elsewhere.

input/metadata/properties/default.json
```diff
@@ -947,6 +950,11 @@
         },
         "events": {
           "type": "string"
+        },
+        "net_merit": {
+          "type": "number",
+          "description": "The net merit value.",
```
Isgotkowitz, line 956:
```suggestion
          "description": "The net merit value (USD).",
```

input/metadata/properties/default.json
```diff
@@ -1496,10 +1515,60 @@
...
+        "description": "The birth year and month of the baby calf."
+      }
+    },
+    "estimated_PTA": {
+      "type": "array",
+      "description": "The estimated PTA value for the top listing semen.",
+      "properties": {
+        "type": "number",
+        "description": "The estimated PTA value for the top listing semen."
+      }
```
Isgotkowitz: PTA should be expanded here as well, either in the input name or description or both.

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,248 @@
...
+
+    """
+        The Top Listing Semen lookup table.
+        {
+            breed: {
+                "year_month": float
+            }
+        }
+        """
+    top_semen: dict[str, dict[str, float]] = {}
```
Isgotkowitz:
```suggestion
    Attributes
    ----------
    net_merit : dict[str, dict[str, dict[str, float]]]
        Lookup table of Net Merit averages and standard deviations (lifetime USD), separated by breed and time.
    top_semen : dict[str, dict[str, float]]
        Lookup table of Top Listing Semen estimated Predicted Transmitting Ability, separated by breed.

    """
    net_merit: dict[str, dict[str, dict[str, float]]] = {}
    top_semen: dict[str, dict[str, float]] = {}
```

Isgotkowitz (approved):
Overall I think this code looks really good Allister! All of my comments were readability-related, and very minor (a lot of docstring-related formatting things). There are still a number of unresolved comments from my first review, but I think the only ones worth really addressing would be the validation patterns for the dates in the metadata.

Also, could you please add some more explanations about the changes to input data? I think not modifying the existing barnyard weather file and instead creating a new file that contains the weather data for the simulation would be better, and the origin of that data should be listed in the PR description. Also, I understand that simulations need to start after a certain date because of the availability of genetic value data. But I think there would be better ways to handle a start date that falls outside of the range than stopping the simulation, especially when animals are not being simulated. Maybe just assign the earliest known net merit value and raise an error to users alerting them that the actual net merit value could not be found?

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -1,15 +1,248 @@
+import numpy as np
+
+from RUFAS.input_manager import InputManager
 from RUFAS.time import Time
+from RUFAS.util import Utility
+
+import datetime
```
Isgotkowitz:
```suggestion
from datetime import date, timedelta, ...
```

PR #1948, opened 2024-08-21
Adding residue tracker

### LINKED ISSUE #481: Track residue content in `LayerData`
## Problem
Currently, there is no attribute in `LayerData` to track the residue content of a soil layer.

But equations in SWAT (namely in section 3:1.2.5) expect to know how much residue there is in a soil layer, and use this value to estimate the amount of carbon in the soil layer. Also in the named section, it calculates a decay rate of the residue in the soil layer and reduces the residue content based on that amount. 

## Temporary Solution
Currently, `MineralizationDecomp` (in the `NitrogenCycling` module, implements section 3:1.2.5) uses the `LayerData` attribute `total_soil_carbon_amount`. There are two potential problems with this:
1. There may be some carbon in the soil that is not from residue, and we don't want that to be included.
2. It is fair game for `MineralizationDecomp` to reduce residue in the soil, but having it reduce carbon would be bad practice (because carbon additions and reductions should be limited to the `CarbonCycling` module)

## Solution
Implement and maintain a residue tracker in `LayerData`

matthew7838:
Utilize a residue tracker property method in the layer data pool that sums structural and metabolic litter to track the residue carbon and utilize it in the mineralization and decomposition.

### Context

Issue(s) closed by this pull request: closes #481 

- [ ] Yes
- [x] No

### What

- A property method that adds structural_litter_amount and metabolic_litter_amount is created in `layer_data.py`.
- In `mineralization_decomp.py`, the calculation of carbon_nitrogen_ratio and carbon_phosphorus_ratio will use the new property method instead of `total_soil_carbon_amount`.
- Tests updated and added.

### Why

There may be some carbon in the soil that is not from residue, and we don't want that to be included.

### How

Having a tracker just to track the litter amount.

### Test plan

1. Have SMEs' second look at the approach and issue.
2. Run with complete test suits.

**NOTE**
I will remove the related TODO once ready to merge.


RUFAS/routines/field/soil/layer_data.py
```diff
@@ -990,6 +990,14 @@ def carbon_emissions(self) -> float:
...
         return self.active_carbon_to_slow_loss + self.slow_carbon_co2_lost_amount + self.passive_carbon_co2_lost_amount
 
+    @property
+    def carbon_residue_amount(self) -> float:
+        """
+        Tracks the total amount of carbon residue
+
+        """
+        return self.metabolic_litter_amount + self.structural_litter_amount
+
```
rawalashmita, line 1003: I see you added carbon residue amount as a sum of metabolic and structural residue amount (kg/ha). Makes sense to me.

RUFAS/routines/field/soil/nitrogen_cycling/mineralization_decomp.py
```diff
@@ -43,12 +43,12 @@ def mineralize_and_decompose_nitrogen(self) -> None:
             return
 
         carbon_nitrogen_ratio = self._calculate_residue_nutrient_ratio(
-            self.data.soil_layers[0].total_soil_carbon_amount,
+            self.data.soil_layers[0].carbon_residue_amount,
             self.data.soil_layers[0].fresh_organic_nitrogen_content,
```
rawalashmita, line 47: Then you replaced total soil carbon amount by carbon residue amount for the numerator of the equation while calculating carbon:nitrogen ratio. SWAT considers the 58% of the organic residue as the amount of carbon in the residue. However, when you are adding the total amount of carbon as the sum of structural and metabolic carbon in the residue instead of having to calculate the amount of carbon from organic residue, it makes sense to me to replace (0.58 * residue(layer)) by structural+metabolic carbon content in the residue. We should make sure the model calculates that for each soil layers though.

rawalashmita: The units also make sense as the amount of carbon which is in the numerator is in kgC/ha and the amount of nitrogen in the denominator is in kgN/ha. Hence, the C:N ratio will be unitless. However, SWAT only accounts for fresh organic N and NO3 (nitrate) in the residue and ignores NH4 content. Based on the law of parsimony, I think we can also ignore the fact that NH4 was being ignored as one of the forms of inorganic nitrogen in the residue in SWAT equation :D. Anyways. it looks good to me!

RUFAS/routines/field/soil/layer_data.py
```diff
@@ -990,6 +990,14 @@ def carbon_emissions(self) -> float:
         """
         return self.active_carbon_to_slow_loss + self.slow_carbon_co2_lost_amount + self.passive_carbon_co2_lost_amount
 
+    @property
+    def carbon_residue_amount(self) -> float:
+        """
+        Tracks the total amount of carbon residue
```
Isgotkowitz:
```suggestion
        Tracks the total amount of carbon residue in a soil layer (kg / ha).
```

Isgotkowitz (approved):
LGTM, nice job Matthew.

KFosterReed (approved):
This is a nice update. The resulting C:N and C:P ratios varied between 5 and 50 and the mineralization decomposition rates varied between 0.01 and 0.04 which I think are in the right range and a big improvement from where we were at!

Nice work team :)

PR #1949, opened 2024-08-21
Adjust check on manure treatment emptying from 1 to 0 remainder

### LINKED ISSUE #1403: Manure not emptied properly when storage period is one day
The manure pit is not emptied correctly with the current methodology of emptying the pit at the end of each storage time period according.  

## Expected Behavior

When the storage time period is set to 1, it should empty on a daily basis. 

## Current Behavior

When the storage time period is set to 30 days. It is updated correctly. 

Storage period 30 days

When the storage time period is set to 1 day. It does not work. 

Storage period 1 day

## Steps to Reproduce

1. Set `storage_time_period` as `1` for the manure treatment you select. 
2.
3.
4.

ew3361zh:
Currently the manure treatment `storage_time_period` is set to empty when there is a remainder of 1 when checked against the simulation day. This means it will never empty when the storage time period is 1 day (x/1 will never have a remainder of 1). This PR attempts to fix that by putting the remainder from that check of the `sim_day` against the `storage_time_period` to 0 which means it is now able to empty with a `storage_time_period` of 1.

### Context

Issue(s) closed by this pull request: closes #1403

- [ ] Yes
- [x] No

### What

In the base_manure_treatment this updates the check on the remainder for `self._sim_day % self.storage_time_period` from 1 to 0.

### Why

This will allow a remainder for a `storage_time_period` of 1 day to match the expected remainder for that day and it will therefore properly empty.

### How

`if self._sim_day % self.storage_time_period == 1:` -> `if self._sim_day % self.storage_time_period == 0:`

Very straightforward change.

### Test plan

Looking at LAC_COW pen manure treatment results (which use anaerobic lagoon for manure storage).

use this filter to get the accumulated liquid manure solids for the pen:
```
{
    "type": "plot",
    "filters": [
      "^Accumulated_ManureTreatmentDailyOutput_Pen_3_LAC_COW.liquid_manure_total_solids.*"
    ],
    "title": "LAC_COW Volatile Solids",
    "variables": [".*"]
}
```

Run 2 simulations:
In the first, set the `anaerobic lagoon` `storage_time_period` to 30 (meaning it will empty every 30 days).

It will look like this with an expected dip every 30 days when it's emptied:
Screenshot 2024-08-21 at 4 31 10 PM

In the second, set set the `anaerobic lagoon` `storage_time_period` to 1 (meaning it will empty every day).

It will look like this with no dips (because it's emptied every day) but also not linear growth.
Screenshot 2024-08-21 at 4 34 43 PM


allisterakun (approved):
LGTM!

tomhuhh (approved):
Thanks, Niko!

The PR was clear and easy to follow, particularly the test plan using the graphic filter. Following your instructions, I was able to achieve the same results. For ease of testing, I recommend adjusting the simulation duration to 365 days instead of 4 or 5 years. I'm glad it was a straightforward fix in the code.

When duration is set to 30 days for anaerobic lagoon. 
default_lac_cow-volatile-solids-25-Aug-2024_Sun_16-46-08

When duration is set to 1 day for anaerobic lagoon. 
default_lac_cow-volatile-solids-25-Aug-2024_Sun_16-50-37

PR #1950, opened 2024-08-22
Partition volatile solids destruction between degradable and non-degradable pools in Anaerobic Digestion

### LINKED ISSUE #1951: [Manure] Simulation crashes because of negative degradable volatile solids mass
There will be a crash in the manure module with the right combination of manure management scenario and pen.

## Expected Behavior

The simulation should not have any masses that go negative, and thus should not crash because of it.

## Current Behavior

The anaerobic digester will take away too much mass from the degradable volatile solids pool, causing it to go negative. This results in a crash in the anaerobic lagoon when the manure finally reaches it.

## Possible Solution

Proportionally distribute the mass of volatile solids destroyed between the degradable and non-degradable volatile solids pools.

## Steps to Reproduce

1. Switch to `dev` or `main` branch.
2. In the default animal file, change the manure management scenario for pen 0 to scenario 5.
3. Run the default simulation.

Isgotkowitz:
Fixes bug that results in negative amount of degradable volatile solids and crash.

### Context

Issue(s) closed by this pull request: closes #1951 

- [ ] Yes
- [x] No

### What

See top sentence.

### Why

Volatile solids destruction is based on the total amount of volatile solids in the manure, but previously this destruction only removed solids from the degradable volatile solids pool. If the proportion of degradable volatile solids to non-degradable volatile solids was low enough, this would result in a negative amount of degradable volatile solids and crash the simulation.

### How

Instead of taking mass away from just the degradable volatile solids pool, volatile solids destruction is now partitioned proportionally between the degradable and non-degradable volatile solids pool.

### Test plan

Updated the test suite to ensure that the expected amounts of degradable and non-degradable volatile solids were present after running the anaerobic digester's daily update routine.

Also, used manure management scenario 5 for pen 0 of the default simulation to check that it ran without crashing (this crashes on the stable branch).

Update: the mypy error count is going up with this bug fix. The reason for that is because the `_current_manure_treatment_daily_input` attribute of `AnaerobicDigestion` is typed as `LiquidManurePortionProtocol` or `None`. Obviously it isn't None, but it would be outside the scope of this PR to modify that for the Manure treatment base class.


tomhuhh (approved):
I tested the PR by updating the manure scenario to 5 (sawdust + ADL) for all the pens. The simulation runs smoothly. I think this is good to go. Thanks Ed for your effort!

Isgotkowitz:
> I tested the PR by updating the manure scenario to 5 (sawdust + ADL) for all the pens. The simulation runs smoothly. I think this is good to go. Thanks Ed for your effort!

@tomhuhh thank you for reviewing this so quickly! @elle-andreen and were discussing this, and decided to wrap these solids adjustments in a method that would bound the pools, and log them to the OM when found. So there have been significant changes to the code from since you reviewed.

tomhuhh:
Setting the bounds to avoid negative values makes sense to me but I think it is almost unlikely that all the volatile solids are destroyed at the anaerobic digester step. Nevertheless, I tested it again and the code still runs smoothly. Thanks Ed and Elle!

elle-andreen:
I ran this with all organic bedding methods, all looks good to me! 

Some context and explanation for others' reference: A PR (#1890) recently modified how bedding solids are handled - they are now added to the manure non-degradable volatile solids (ndVS) fraction of manure, whereas they previously were added only to total solids. With the addition of bedding to the ndVS fraction, this naturally also causes an increase in manure total VS. The anaerobic digestion treatment methane estimation method, unlike other components of the manure module, is based on **total VS** in manure loaded into the digester, without regard to degradable or non-degradable VS. However, we do not currently have good information about:
1. What proportion of digester CH4 production originates from breakdown of degradable vs. non-degradable VS
2. How degradable or methanogenic bedding solids are, i.e., should they be lumped into the manure VS pool in the context of anaerobic digestion and if so what is the degradable vs. non-degradable proportion

Because of this uncertainty we previously assumed that **all destroyed VS in anaerobic digestion were from the degradable VS fraction**. However, when quantities of bedding VS are very high in relation to the quantity of degradable VS (which only come from animal feces, not from bedding at this time), this can produce estimated VS destruction in the digester exceeding the quantity of manure degradable VS (i.e. non-permissible negative numbers), which results in an error. **Additionally, even when conditions do not cause an error, the degradable VS fraction can be greatly reduced, resulting in artificially low methane emissions in the subsequent manure storage lagoon as degradable VS are what drive CH4 production from storages. For example, once the changes in PR 1890 were applied and bedding was added to ndVS, using default values with manure scenario 5 (sawdust, manual scrape, anaerobic digestion + lagoon), this results in about 30 kg less CH4 emission per animal per year.** 

What the changes in this PR effectively accomplish is ensuring a **consistent proportion** of degradable VS is destroyed in anaerobic digestion, and thus makes it to the manure storage. This PR 'distributes' the VS destruction between the dVS and ndVS fractions according to their proportion in incoming manure, and results in _each of the 3 VS fractions_ (total, degradable, and non-degradable) all being reduced by the same percentage (~42.5%, if you do the math). This accomplishes a reasonably similar reduction in total and degradable VS as before PR 1890 merged, and also prevents degradable VS from becoming negative. 

This is a good temporary fix, but, we will need to follow up with SME's on how to refine this method in the context of anaerobic digestion, and on the use of degradable and non-degradable VS in general. This change will ostensibly result in underestimation of the amount/proportion of dVS destroyed in anaerobic digestion, and so lagoons after digesters will have slightly greater methane emissions than prior to 1890 and this PR. Under the same default conditions with manure scenario 5, **a digestate lagoon on this branch will produce about 10-15 kg more CH4 per animal per year (across animal classes) compared to a digestate lagoon where bedding is NOT added to the ndVS fraction**. I think this is more or less reasonable given that bedding is represents additional fermentable substrate going into the lagoon, but should be evaluated further at a later time (perhaps rolled into general discussion about degradable and non-degradable VS).

elle-andreen (approved):
Thanks for your quick work on this Ed!

KFosterReed:
@PooyaHekmati flagging that this will need to get pulled into main and shared with Vyla.

Isgotkowitz:
> @PooyaHekmati flagging that this will need to get pulled into main and shared with Vyla.

Roger that, will have the tests updated asap

tests/test_manure_module/test_manure_treatments.py
```diff
@@ -2455,6 +2440,51 @@ def test_calc_manure_heat_capacity() -> None:
     assert actual_manure_heat_capacity == approx(expected_manure_heat_capacity)
 
 
+@pytest.mark.parametrize(
+    "destruction,degradable_vol_sols,nondegradable_vol_sols,non_vol_solids,expected_total_solids,"
+    "expected_total_vol_sols,expected_degradable_vol_sols,expected_nondegradable_vol_sols,is_error",
+    [
+        (100.0, 80.0, 20.0, 20.0, 20.0, 0.0, 0.0, 0.0, False),
+        (30.0, 40.0, 20.0, 0.0, 30.0, 30.0, 20.0, 10.0, False),
+        (110.0, 10.0, 90.0, 100.0, 90.0, -10.0, 0.0, 0.0, True),
```
Isgotkowitz: While writing the test for the error case I realized that the logic does not handle the error case in a sensible way. I am not sure if it would be better to leave it as-is or come up with a better fix.

Isgotkowitz: Update: modified the fix so that this erroneous behavior is avoided.

Isgotkowitz:
Update: @tomhuhh and @elle-andreen I have modified the solution since the two of you reviewed it. It has the same logic as before and should produce the same results that you would have seen, however I wanted to make you two aware of this incase either of you would like to re-review it.

The change has combined the separate checks of the non-degradable and degradable pools into a single check to make sure there are not too many volatile solids being degraded. This can be done because we know that the degradations will effect each pool proportionally, so one pool is losing too many volatile solids then we know the other would as well.

RUFAS/routines/manure/manure_treatments/anaerobic_digestion.py
```diff
@@ -172,6 +175,61 @@ def _calc_specific_input_energy(self, average_temperature_celsius, moisture_cont
         )
         return heating_input_energy
 
+    def _recalculate_solids_after_destruction(
+        self, volatile_solids_destruction: float, manure_output: ManureTreatmentDailyOutput
+    ) -> None:
```
allisterakun:
```suggestion
    def _recalculate_solids_after_destruction(
        self, volatile_solids_destruction: float, manure_output: ManureTreatmentDailyOutput
    ) -> ManureTreatmentDailyOutput:
```

RUFAS/routines/manure/manure_treatments/anaerobic_digestion.py
```diff
@@ -172,6 +175,61 @@ def _calc_specific_input_energy(self, average_temperature_celsius, moisture_cont
...
+            )
+            manure_output.liquid_manure_total_volatile_solids = (
+                self._current_manure_treatment_daily_input.liquid_manure_total_volatile_solids
+                - volatile_solids_destruction
+            )
+
+        manure_output.liquid_manure_total_solids = (
+            self._current_manure_treatment_daily_input.liquid_manure_total_solids - volatile_solids_destruction
+        )
+
```
allisterakun, line 237:
```suggestion
            self._current_manure_treatment_daily_input.liquid_manure_total_solids - volatile_solids_destruction
        )
        return manure_output

```

RUFAS/routines/manure/manure_treatments/anaerobic_digestion.py
```diff
@@ -172,6 +175,61 @@ def _calc_specific_input_energy(self, average_temperature_celsius, moisture_cont
         )
         return heating_input_energy
 
+    def _recalculate_solids_after_destruction(
```
allisterakun, line 178: This function takes in a local variable (within the scope of `_calc_anaerobic_digestion_daily_output()`), `new_daily_output` as `manure_output`, and modifies the object without returning anything. It relies on the side effect to update the `new_daily_output` variable in the caller function. Please consider returning the updated `manure_output` object instead.

RUFAS/routines/manure/manure_treatments/anaerobic_digestion.py
```diff
@@ -96,19 +111,7 @@ def _calc_anaerobic_digestion_daily_output(
...
-            self._current_manure_treatment_daily_input.liquid_manure_total_volatile_solids - AD_VS_destruction
-        )
-        new_daily_output.liquid_manure_total_degradable_volatile_solids = (
-            self._current_manure_treatment_daily_input.liquid_manure_total_degradable_volatile_solids
-            - AD_VS_destruction
-        )
-        new_daily_output.liquid_manure_total_non_degradable_volatile_solids = (
-            self._current_manure_treatment_daily_input.liquid_manure_total_non_degradable_volatile_solids
-        )
+        self._recalculate_solids_after_destruction(AD_VS_destruction, new_daily_output)
```
allisterakun:
```suggestion
        new_daily_output = self._recalculate_solids_after_destruction(AD_VS_destruction, new_daily_output)
```

allisterakun (approved):
LGTM! Tested with the recommended scenarios mentioned in the PR description and works as expected.
I left one minor comment about returning the updated manure output in the new function.

Isgotkowitz:
> LGTM! Tested with the recommended scenarios mentioned in the PR description and works as expected. I left one minor comment about returning the updated manure output in the new function.

Thanks @allisterakun. Good suggestion about having the object returned, instead of relying on side effects. Will implement that rn

PR #1956, opened 2024-08-24
Certain combination of gestation length and days_in_preg_when_dry causes RuFaS to crash.

### LINKED ISSUE #1880: [Animal] Some combinations of gestation length and days in preg when dry can cause crash
## Expected Behavior

No crashing!

## Current Behavior

Currently in `cow.py` and specifically `get_bw_change`, the lines
```
bodyweight_tissue = self.tissue_changed / (
                    self.gestation_length - AnimalBase.config["days_in_preg_when_dry"]
                )
``` 
cause a problem if the `self.gestation_length` and `AnimalBase.config["days_in_preg_when_dry"]` are equal.

Such a case can happen given `gestation_length` is set in `_calculate_gestation_length` from `heiferII.py`, which use `AnimalBase.config["avg_gestation_len"]` and `AnimalBase.config["std_gestation_len"]`

## Possible Solution

Simplest fix is wrapping the equation in a try except block
```
            try:
                bodyweight_tissue = self.tissue_changed / (
                    self.gestation_length - AnimalBase.config["days_in_preg_when_dry"]
                )
            except:
                bodyweight_tissue = self.tissue_changed
```

However I'm uncertain this is ideal, as we could instead elect a more typical value in place of the denominator, e.g. the average for animals in their dry period.

## Steps to Reproduce

Modify the `default_animal.json` such that:
`"days_in_preg_when_dry": 282,` and `"avg_gestation_len": 320,`

Isgotkowitz:
Fixes an issue where cows entering the herd have prematurely dried off.

### Context

Issue(s) closed by this pull request: closes #1880 

- [ ] Yes
- [x] No

### What

Modifies the initialization routine for adding animals to the herd so that pregnant cows are only marked as dry if they have not reached the dry-off day of pregnancy specified by the user. A warning is logged when this is done to a cow.

Note that some input files are modified as well - this is for testing purposes, I will change them back before merging this PR.

### Why

This change prevents the crash described in #1880, where gestation length is the same as the number of days a cow is pregnant before it dries off. It prevents this by setting cows to be milking, which means that their days in milk is greater than 0, which prevents the path of execution from going to the place in the code where the code crashes.

### How

In `initialize_herd` in `LifeCycleManager`, a subroutine checks all cows to make sure that they are still milking if they have not reached the specified dry-off day of pregnancy yet. If the routine finds that the cow is dry but has not reached the user-specified day of pregnancy when the cow is supposed to dry off, then the cow's milking status is set to true and her days in milk are calculated based on when she last calved.

### Test plan

Wrote a new unit test to verify that a cow's milking status will be changed when the cow's attributes to not match what would be expected with the user-defined management attributes.

Modified the inputs in `default_animal.json` as specified in the issue, and ran the default simulation to verify that the crash no longer occurs. Also read through the warnings to verify that some animals were having their milking attributes changed (we expect this in order for the crash to be prevented).

I did see that some animals have a really high "days_in_milk" after the correction, with one animal nearing 400 days. I assume this is ok because 1) cases like this seem pretty rare 2) this makes sense if it takes a cow a really long time to get pregnant, and 3) this only happens to animals entering through at initialization, it should be the case that most or all of these animals have left the herd after a couple years.


Isgotkowitz:
Actions are currently failing because `default_task.json` has been modified. When this PR is ready to be merged I will revert all inputs back, and the Actions should pass at that point.

RUFAS/routines/animal/life_cycle/life_cycle.py
```diff
@@ -234,8 +235,42 @@ def _get_animals(self, animal_type: Type[GenericAnimal]) -> List[GenericAnimal]:
...
+
+        Returns
+        -------
+        list[Cow]
+            The list of Cow instances with their milking attributes checked and corrected.
+
+        """
+        for cow in cows:
+            if not cow.is_pregnant:
+                continue
```
allisterakun, line 260: This line is not covered by unit test.

```
PS D:\Work\Cornell\MASM> coverage report -m .\RUFAS\routines\animal\life_cycle\life_cycle.py .\tests\animal_module_tests\test_animal_life_cycle.py
Name                                                    Stmts   Miss  Cover   Missing
-------------------------------------------------------------------------------------
.\RUFAS\routines\animal\life_cycle\life_cycle.py          459     33    93%   260, 314-356, 758, 820, 1051-1052
.\tests\animal_module_tests\test_animal_life_cycle.py    1151      4    99%   1192-1194, 1637
-------------------------------------------------------------------------------------
TOTAL                                                    1610     37    98%

```

allisterakun (approved):
I changed "days_in_preg_when_dry" to 282 and kept "avg_gestation_len" at 320, as described in issue #1880. 
The simulation runs without problems with the new correction logic for cows' milking attributes. I tried removing the new logic, and the simulation crashed as expected. 
Confirming that the fix works as expected.
Great job, Ed! LGTM!

Since the GitHub Actions are broken at the moment due to modification to default_task.json, I ran MyPy on my local machine and confirmed that no new MyPy errors are introduced in this PR.
However, I found that line 260 in `life_cycle.py` is not covered by testing, please update the test to cover this if possible.

KFosterReed:
I tested this by making a new animal population file with a very early dry off periord using the default animal json file but changing `"days_in_preg_when_dry"` to 150. Then I ran the simulation for 1 year once with the same input used for the herd initialization and again with  `"days_in_preg_when_dry": 300`. 

The program didn't crash AND the resulting average herd DIM reflected the changes. I also printed out the individual animal DIM and saw the differences one would expect based on the inputs but nothing crazy.

RUFAS/routines/animal/life_cycle/life_cycle.py
Isgotkowitz: Just added a unit test case which will hit that line, thanks Allister

KFosterReed:
Also got the expected warnings! 

image

KFosterReed (approved):
Works as expected!

Isgotkowitz:
@KFosterReed and @allisterakun thank you for your very thorough reviews!

PR #1957, opened 2024-08-26
 [Time] Change type annotations in Time module

### LINKED ISSUE #1894: [Time] Change type annotations in `Time` module
Type hints for the attributes of `Time` are not quite correct.

## Status

Currently, a lot of the attributes for `Time` are typed as `datetime` the module.

## Intention

Lower the number of Mypy errors by typing `Time` attributes as `datetime.datetime` or `datetime.date`.

## Request

Change the type annotations from `datetime` to `datetime.datetime` or `datetime.date` in the `Time` module.

### Additional Information

Between `datetime.datetime` and `datetime.date`, it is my opinion that `datetime.date` would be more appropriate to use in RuFaS. This is because the `date` has a resolution of days, whereas `datetime` has a resolution of microseconds which would be overkill for RuFaS.

allisterakun:
### Context

Issue(s) closed by this pull request: closes #1894 

- [ ] Yes
- [x] No

### What

Update all time-related type hints in the Time class

### Why

To properly assign a type hint.

### How

Update all time-related type hints in the Time class to be `datetime.datetime`.

### Test plan

No unit tests are affected, the Mypy error count goes down by 56.

### DIFF
--- a/RUFAS/time.py
+++ b/RUFAS/time.py
@@ -17,10 +17,8 @@ class Time:
         self.im = InputManager()
         self.config_data: Dict[str, str | int | bool] = self.im.get_data("config")
 
-        start_date_str: str = self.config_data["start_date"]
-        end_date_str: str = self.config_data["end_date"]
-        self.start_date: datetime.datetime = datetime.datetime.strptime(start_date_str, "%Y:%j")
-        self.end_date: datetime.datetime = datetime.datetime.strptime(end_date_str, "%Y:%j")
+        self.start_date: datetime.datetime = datetime.datetime.strptime(str(self.config_data["start_date"]), "%Y:%j")
+        self.end_date: datetime.datetime = datetime.datetime.strptime(str(self.config_data["end_date"]), "%Y:%j")
 
         self.current_date: datetime.datetime = self.start_date
         self.simulation_length_days: int = (self.end_date - self.start_date).days
--- a/RUFAS/time.py
+++ b/RUFAS/time.py
@@ -9,17 +9,20 @@ from RUFAS.util import Utility
 
 
 class Time:
-    def __init__(self):
+    def __init__(self) -> None:
         """
         This object is responsible for creating and tracking time in the simulation.
         """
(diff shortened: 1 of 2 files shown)


RUFAS/time.py
```diff
@@ -9,17 +9,18 @@
...
         """
         self.om = OutputManager()
         self.im = InputManager()
         self.config_data: Dict[str, str | int | bool] = self.im.get_data("config")
-        self.start_date: datetime = datetime.datetime.strptime(self.config_data["start_date"], "%Y:%j")
-        self.end_date: datetime = datetime.datetime.strptime(self.config_data["end_date"], "%Y:%j")
 
-        self.current_date: datetime = self.start_date
+        self.start_date: datetime.datetime = datetime.datetime.strptime(str(self.config_data["start_date"]), "%Y:%j")
+        self.end_date: datetime.datetime = datetime.datetime.strptime(str(self.config_data["end_date"]), "%Y:%j")
```
allisterakun, line 21: Is this necessary? We know that `self.config_data["end_date"]` is a string, but by the previous definition (`self.config_data: Dict[str, str | int | bool]`), MyPy is not sure if `self.config_data["end_date"]` is a string, integer or boolean.

Isgotkowitz: What exactly are you referring to when you say "this"? Do you mean putting `str(...)` around `self.config_data["end_date"]`? If so, I am not sure. My advice is to do whatever it takes to make Mypy pipe down, even if it seems a little redundant.

Isgotkowitz (approved):
LGTM

RUFAS/time.py
allisterakun: Yes, this is what I meant (putting `str()` around the already-a-string entities).

Isgotkowitz: Gotcha, I think it is not strictly necessary but it is helpful to have.

ew3361zh (approved):
Very nice!

PR #1958, opened 2024-08-27
Deprecated update classes in pen function

### LINKED ISSUE #1208: Deprecate `update_classes_in_pen`
Deprecate use of `update_classes_in_pen` once AnimalCombination enum is fully implemented. May depend on PR #793

0luchi:
### Context

Issue(s) closed by this pull request: closes #1208 

- [ ] Yes
- [x] No

### What

- This PR deprecates `update_classes_in_pen` and all tests it covers since animal_combination covers this scenario and makes the function irrelevant.
- The PR also deprecates `classes_in_pen` and replaces it with `animal_combination`

### Why

This is important because `animal_combination` does the function of `classes_in_pen` and if a pen has an animal combination, then that already represents the classes in the pen.

### How

- I removed the `update_classes_in_pen` function and removed every instance of where it was called. 
- I also replaced all instances of `classes_in_pen` with `animal_combination` and updated all the test cases.

### Test plan

No tests were added but test cases were updated to reflect changes


RUFAS/routines/animal/pen.py
```diff
@@ -529,7 +519,10 @@ def calc_daily_walking_dist(self):
         """
         Sets the daily walking distance for the cows in the pen (if any).
         """
-        if "Cow" in self.classes_in_pen:
+        if (
+            self.animal_combination == AnimalCombination.LAC_COW
+            or self.animal_combination == AnimalCombination.CLOSE_UP
+        ):
```
Isgotkowitz: With if statements like this, it makes it often makes the code cleaner if the condition is extracted into its own line and named explicitly:
```suggestion
cows_are_in_pen = self.animal_combination in {AnimalCombination.LAC_COW, AnimalCombination.CLOSE_UP}
if cows_are_in_pen:
```

Isgotkowitz: Also, this change does not appear to be explicitly related to deprecating `update_classes_in_pen`, why is this needed?

Isgotkowitz (commented):
This is a good start to deprecating the update animals classes in pen, however it seems like there is more work to do. If `update_classes_in_pen` is removed, is the attribute `classes_in_pen` ever updated? Since `classes_in_pen` is still being used, I assume it has to be.

0luchi:
@Isgotkowitz  Thanks for looking at my PR. Concerning your comment, https://github.com/RuminantFarmSystems/MASM/pull/1958#pullrequestreview-2266383661 the issue related to it required `update_classes_in_pen` to be deprecated because `animal_combination`  has been implemented. So, I don’t think it needs to be updated anymore. Maybe @JoeWaddell would know more since he created the issue?

JoeWaddell:
> @Isgotkowitz Thanks for looking at my PR. Concerning your comment, [#1958 (review)](https://github.com/RuminantFarmSystems/MASM/pull/1958#pullrequestreview-2266383661) the issue related to it required `update_classes_in_pen` to be deprecated because `animal_combination` has been implemented. So, I don’t think it needs to be updated anymore. Maybe @JoeWaddell would know more since he created the issue?

Thanks for the tag! This issue I created from an old comment in the code from someone else - I think during one of the earlier animal manager refactors in 2022/23?

So agreed with Ed: it's a good start, and this PR will require replacing all instances of the `classes_in_pen` attribute with the appropriate pen attribute (`animal_combination`), and use of `AnimalType` checks instead of direct checks to strings (e.g. `AnimalType.LAC_COW` or `AnimalType.DRY_COW` instead of `Cow`)

RUFAS/routines/animal/life_cycle/pen_history.py
```diff
@@ -1,9 +1,10 @@
 from typing import TypedDict
+from ....enums import AnimalCombination
 
 
 class PenHistory(TypedDict):
     """
-    A class to represent the history of a pen on a farm.
+    A class  to represent the history of a pen on a farm.
```
Isgotkowitz:
```suggestion
    A class to represent the history of a pen on a farm.
```

RUFAS/routines/animal/life_cycle/animal_base.py
```diff
@@ -9,6 +9,7 @@
 from RUFAS.routines.animal.ration.amino_acid import EssentialAminoAcidRequirements
 from RUFAS.routines.animal.life_cycle.lactation_curve import LactationCurve
 from RUFAS.time import Time
+from ....enums import AnimalCombination
```
Isgotkowitz:
```suggestion
from RUFAS.enums import AnimalCombination
```
Relative pathing should be avoided if not at the same level or one higher, i.e. `from .enums ... ` or `from ..enums ...` would be fine but `from ...enums ...` should be turned into a full path instead. I am still not 100% sure about the best practices related to relative pathing or the logic underlying these guidelines, but this is what I have been told.

RUFAS/routines/animal/life_cycle/pen_history.py
```diff
@@ -1,9 +1,10 @@
 from typing import TypedDict
+from ....enums import AnimalCombination
```
Isgotkowitz: Same comment on relative pathing here

RUFAS/routines/animal/life_cycle/pen_history.py
```diff
@@ -16,12 +17,12 @@ class PenHistory(TypedDict):
         The end date of the pen's usage.
     pen : int
         The id of the pen that the animal is in.
-    classes_in_pen : list[str]
-        The classes of animals that have been in the pen.
+    animal_combination : list[str]
```
Isgotkowitz:
```suggestion
    animal_combination : AnimalCombination
```

Isgotkowitz (commented):
Found a few minor things that I commented on, but overall this looks great! I don't expect there would be any changes to outputs from this PR, and while I did not thoroughly check this I did check that the list of variable names produced is the same as on `dev` and it was, which makes me reasonably confident that nothing changed in the outputs.

Before approving I just want to confirm with @allisterakun that it is ok to make changes to the `animal` directory that is in the `routines` directory, instead of `biophysical`. Once that happens I will come back and approve.

Isgotkowitz (commented):
Code looks good to me! Once all the Flake8 errors are worked out and the new changelog (`changelog.md` in the codebase) is updated I will come back and approve.

RUFAS/routines/animal/pen.py
```diff
@@ -529,15 +513,14 @@ def calc_daily_walking_dist(self):
         """
         Sets the daily walking distance for the cows in the pen (if any).
         """
-        if "Cow" in self.classes_in_pen:
+        cows_are_in_pen = self.animal_combination in {AnimalCombination.LAC_COW, AnimalCombination.CLOSE_UP}
+        if cows_are_in_pen:
```
JoeWaddell: Here you might want to simply make the check directly, instead of creating a local variable, since it'll conform better to logic elsewhere in the codebase.

Also out of curiosity, is there a reason to use a set instead of a list?

RUFAS/routines/manure/pen_manure/manure_manager_pen.py
```diff
@@ -130,6 +128,6 @@ def exposed_manure_surface_area_from_pen_type(self) -> float:
 
         exposed_manure_surface_area = exposed_manure_surface_area_by_pen_type[self.pen_type]
 
-        if "LacCow" in self.classes_in_pen:
+        if AnimalCombination.LAC_COW == self.animal_combination:
```
JoeWaddell: Reversing the logic here might read better, but it's the same functionality. Just thinking ahead in case there's more branching logic in the future, since checking `self.animal_combination` against lists might come up.


```suggestion
        if self.animal_combination == AnimalCombination.LAC_COW:
```

tests/animal_module_tests/test_animal.py
```diff
@@ -30,6 +30,7 @@
 )
 from RUFAS.routines.animal.ration.ration_optimizer import RationOptimizer
 from RUFAS.routines.animal.ration.user_defined_ration import UserDefinedRationManager
+from RUFAS.enums import AnimalCombination
```
JoeWaddell: duplicate import, see line 11

RUFAS/routines/manure/pen_manure/manure_manager_pen.py
```diff
@@ -4,6 +4,7 @@
 from RUFAS.routines.manure.pen_manure.pen_manure import PenManure
 from ....enums import AnimalCombination
 from ....data_structures.pen_manure_data import PenManureData
+from RUFAS.routines.animal.animal_types import AnimalType
```
JoeWaddell: unused import

JoeWaddell (changes requested):
Looks good! Just commented on a few things to tidy up

RUFAS/routines/animal/pen.py
0luchi: No reason in particular

KFosterReed:
@JoeWaddell and/or @allisterakun  can you confirm that this branch/PR are OK to delete?

PR #1964, opened 2024-08-29
Preventing `dump_all_nondata_pools` crash in except block of `task()`

### LINKED ISSUE #1923: [Task Manager] Call to `OutputManager`'s `dump_all_nondata_pools` is crashing
When a task raises an exception in the `task` method of `TaskManager`, the call to `dump_all_nondata_pools` crashes in the except block.

## Expected Behavior

It should not crash. It should dump the logs correctly and allow the rest of the except block to run.

## Current Behavior

The code crashes when dumping all nondata logs after handling an exception.
image

## Possible Solution

Add a check to see if a file exists before trying to dump logs into it, if it doesn't then create it before dumping the logs into it.

## Steps to Reproduce

1. Raise an error within the `try` block of the method `task` in `TaskManager`.
2. Run `python main.py -v errors`
3. Look for the crash pictured above.

## Context (Environment)

Not being able to review the logs when the code crashes in a big problem. It can make it much more difficult to find the problem causing the crash to occur. Even worse, if a user randomly seeds a simulation that crashes and they didn't have logs printed to stdout, it could make that simulation nearly impossible to reproduce.

Isgotkowitz:
Prevents crash from occurring by ensuring directory is created before trying to create a file in said directory.

### Context

Issue(s) closed by this pull request: closes #1923 

- [ ] Yes
- [x] No

### What & Why & How

The crash described in the issue occurred because the `logs` directory in the `output` directory is not always present before the OutputManager's `dump_all_nondata_pools` method is called. When this is the case, the OutputManager attempts to write log files to a directory that doesn't exist and this is a big no-no. The solution is pretty simple, add a call to `create_directory` in `dump_all_nondata_pools`. This method is idempotent, so it runs gracefully whether or not the logs directory has already been created.

### Test plan

Raised `ValueError`s at different points within the `try` block of `task` in `TaskManager`, and ran the simulation with the `-c` flag. Checked the simulation crashed gracefully with all logs, warnings, and errors being dumped into the `logs` directory.

### DIFF
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -1539,6 +1539,7 @@ class OutputManager(object):
         """
         Dumps all non-data pools into the given path to a directory.
         """
+        self.create_directory(path)
         self.dump_variable_names_and_contexts(path, exclude_info_maps, format_option)
         self.dump_logs(path)
         self.dump_warnings(path)
--- a/tests/test_output_manager.py
+++ b/tests/test_output_manager.py
@@ -887,6 +887,7 @@ def test_dump_all_nondata_pools(mocker: MockerFixture) -> None:
     # Arrange
     output_manager = OutputManager()
     path = "dummy_path"
+    patch_create_dir = mocker.patch.object(output_manager, "create_directory")
     patch_for_dump_errors = mocker.patch.object(output_manager, "dump_errors")
     patch_for_dump_warnings = mocker.patch.object(output_manager, "dump_warnings")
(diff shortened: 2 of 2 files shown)


ew3361zh (approved):
Excellent, worked as intended, nice and straight-forward fix.

allisterakun (approved):
LGTM! 
I deleted the logs folder and removed the new line to create the logs directory; I couldn't recreate the crash due to the missing logs folder. 
With the creat directory line added, the simulation finishes without crashing, the logs directory is created as expected, and all logs are correctly dumped.

PR #1968, opened 2024-08-30
[Process] Add Markdown changelog to the repository

### LINKED ISSUE #1967: [Process] Switch changelog to being tracked by git
The current changelog is often updated before a PR is merged, and this can be problematic. So this issue is requesting a better system for managing a changelog.

## Status

The current changelog is a Google sheet. PR authors indicate they have updated it by clicking a box in their PR description. There is no automation to check that the PR is actually merged, or to remove the entry if the PR is closed for some reason without merging.

## Intention

Introduce tracking and automation into the changelog updating process so that they change log is more correct.

## Request

Create a new changelog file that is tracked by git on GitHub. It should have fields for the PR #, the impact area, a description, the version # associated with the change.

Isgotkowitz:
Adds a changelog to the repository so that it will only be updated when a PR is actually merged.

### Context

Issue(s) closed by this pull request: closes #1967 

- [ ] Yes
- [x] No

### What

Adds a changelog to the repository written in Markdown. The changelog will have an entry for each PR that gets merged, with each entry containing the version # associated with a particular PR, a link to said PR, the impact area, and a short description of changes.

Updates the PR description template so that it tells authors and reviewers where the new changelog is.

Updates the actions to block the merge if the change log has not been updated.

### Why

The current Google doc changelog is not always totally insync with the progress being made on GitHub. Authors will sometimes add entries for PRs that won't be merged until several days later, sometimes longer. This makes the changelog a less accurate document.

### How

Added a markdown file, titled it "Changelog", put instructions and example entry in, and an entry for this PR. Updates the actions to block the merge if the change log has not been updated.

### Test plan

The action is active on this PR itself, and it is working.

### Options for Alternative Designs and Extensions
An alternative way to maintain the changelog could be to have it written in ReStructuredText and link it to the documentation generated by Sphinx. This would look really pretty and be nicely integrated with our documentation, however it would be less developer-friendly to make edits to the change log.

### Update
The GitHub Action has been modified to fail if the changelog is not in the list of changed files and comment that this did not happen to the user. It does this by capturing whether or not the changelog is in the list of changed files in the first job, storing that as an output and then using it at the end of the second job. My reasoning is that we don't want get the changed files twice, but the job should always run through the formatting and linting steps before stopping because a PR author might reasonably want their code formatted, Flaked and Mypy'd before they update the changelog.

This change can be tested by making a new branch, pushing it to GitHub, and making a PR with it. That PR will pass the action if the changelog is still in the codebase, but if it is deleted then the action will fail and write a comment to the PR.

The changelog itself has been updated so that it will have authors indicate whether the changes are nominated for a major version change.

The PR description no longer has the reminder and checkbox for the changelog, because the Action takes over that role.


Isgotkowitz:
Update: realized the instructions for what to put in each entry were not specific enough, copied over instructions from old changelog

PooyaHekmati (changes requested):
We said that we want to remove the changelog from PR description template and have actions blocking the merge unless the changelog is updated. We also need the nomination for major version update.

Isgotkowitz:
@PooyaHekmati have addressed your comment. But I am curious about how the nomination for major version update works in our semantic versioning system. If a major change is introduced, wouldn't that major change be the reflected by bumping up the version # that is in the changelog entry? Meaning the change wouldn't need to be nominated for a major version update, because it would have already been agreed to when the PR gets merged.

PR #1977, opened 2024-09-06
Enhance emission.py

### LINKED ISSUE #1982: [EEE] emission.py should be better enhanced before adding unit tests.
emission.py has a lot of long and repetitive code, it would make the code itself and testing a lot more simpler and cleaner if we can have some of it cleaned up and extracted.
## Steps to Reproduce

1.
2.
3.
4.

matthew7838:
Clean up and enhance `emission.py`. Feel free to give more suggestions on what other things to extract/simplify.
### Context

Issue(s) closed by this pull request: closes #1982

### What

Extract and refactor code to reduce the complexity of methods.

### Why

For cleaner codes and easier testings.

### Test plan

1. Run with default data on both `dev` and `emission-enhance`.
2. Filter outputs with emission keywords.
3. Use comparison tools to compare.


RUFAS/routines/EEE/emissions.py
```diff
@@ -366,7 +324,7 @@ def _calculate_homegrown_feed_emissions(
                 aggregated_manure_apps[field],
                 aggregated_manure_requests[field],
             )
-            crops_with_emissions.extend(crops)
+            (0.0).extend(crops)
```
Isgotkowitz: What is 0.0? Why is it being extended?

RUFAS/routines/EEE/emissions.py
```diff
@@ -501,33 +466,90 @@ def _calculate_emissions_by_field(
...
-                fertilizer_applications["potassium"]
-                * fraction_of_total_mass_grown
-                * EMBEDDED_POTASSIUM_FERTILIZER_EMISSIONS_FACTOR
-            )
-            crop["manure_nitrogen_used"] = manure_applications["nitrogen"] * fraction_of_total_mass_grown
-            crop["manure_nitrogen_requested"] = manure_requests["nitrogen"] * fraction_of_total_mass_grown
 
         return feeds_grown
+
+    def filtered_results(
```
Isgotkowitz: This method should be renamed so that it starts with a verb i.e. `filter_results`.

RUFAS/routines/EEE/emissions.py
```diff
@@ -501,33 +466,90 @@ def _calculate_emissions_by_field(
...
+        Helper methods to process the filters for requests/applications.
+
+        Parameters
+        ----------
+        filters: dict[str, Any]
+            A filter to collect the desired data from OutputManager.
+        date_cutoff: datetime
+            Date that indicates the cutoff of occurrence.
+        occurance_time_name: Tuple[str, str]
+            How the key is named for the year and day from the data retrieved from OM.
```
Isgotkowitz: Instead of making this a tuple, why not just make the year and day keys be separate inputs? It would be much more explicit than having the year and day in a tuple and relying on ordering

```suggestion
        filters : dict[str, Any]
            A filter to collect the desired data from OutputManager.
        date_cutoff : datetime
            Date that indicates the cutoff of occurrence.
        occurance_time_name : Tuple[str, str]
            How the key is named for the year and day from the data retrieved from OM.
```

RUFAS/routines/EEE/emissions.py
```diff
@@ -501,33 +466,90 @@ def _calculate_emissions_by_field(
...
+        return list(
+            filter(
+                lambda app: Time.convert_year_jday_to_date(app[year], app[day]) >= date_cutoff,
+                processed_data,
+            )
+        )
+
+    @staticmethod
+    def manure_aggregation(
+        aggregated_manure: dict[str, dict[str, float]], manure_actions: list[dict[str, Any]]
```
```suggestion
        aggregated_manure: dict[str, dict[str, float]], manure_operations: list[dict[str, Any]]
```

RUFAS/routines/EEE/emissions.py
```diff
@@ -501,33 +466,90 @@ def _calculate_emissions_by_field(
...
+    ) -> None:
+        """Helper methods for _calculate_homegrown_feed_emissions"""
```
Isgotkowitz: This method should have a full docstring

RUFAS/routines/EEE/emissions.py
```diff
@@ -501,33 +466,90 @@ def _calculate_emissions_by_field(
...
+    def calculate_crop_emissions_and_fertilizer_usage(
+        crop: dict[str, Any],
+        field_emissions: dict[str, float],
+        fertilizer_applications: dict[str, float],
+        manure_applications: dict[str, float],
+        manure_requests: dict[str, float],
+        total_dry_mass_per_ha_grown: int,
+        field_size: float,
+    ) -> None:
+        """Helper method to deal with crop emissions and fertilizer related process."""
```
Isgotkowitz: This method is responsible for assigning emissions, fertilizer used, and other sustainability metrics to a particular crop based on how much of it was grown. This needs to be reflected in the docstring.

RUFAS/routines/EEE/emissions.py
```diff
@@ -501,33 +466,90 @@ def _calculate_emissions_by_field(
...
+        year, day = occurance_time_name
+        return list(
+            filter(
+                lambda app: Time.convert_year_jday_to_date(app[year], app[day]) >= date_cutoff,
+                processed_data,
+            )
+        )
+
+    @staticmethod
+    def manure_aggregation(
```
Isgotkowitz: This method name should start with a verb i.e. `aggregate_manure_operations`

RUFAS/routines/EEE/emissions.py
```diff
@@ -501,33 +466,90 @@ def _calculate_emissions_by_field(
...
+        date_cutoff: datetime
+            Date that indicates the cutoff of occurrence.
+        occurance_time_name: Tuple[str, str]
+            How the key is named for the year and day from the data retrieved from OM.
+
+        Returns
+        -------
+        list[dict[str, Any]]
+            Filtered result of the retrieved data.
+        """
```
Isgotkowitz, line 549:
```suggestion
        
        """
```

RUFAS/routines/EEE/emissions.py
```diff
@@ -91,84 +84,53 @@ def _gather_homegrown_feeds_and_fertilizer_apps(
...
+            "filters": ["Time.(day|calendar_year)"],
+            "slice_start": SLICE_START,
+            "slice_end": SLICE_END,
+        }
+        date_variables = self.om.filter_variables_pool(time_filter)
+        day_cutoff = date_variables["Time.day"]["values"][0]
+        year_cutoff = date_variables["Time.calendar_year"]["values"][0]
+        date_cutoff = Time.convert_year_jday_to_date(year_cutoff, day_cutoff)
+
         crop_filter = {
```
Isgotkowitz: What about wrapping all of these filters (crop, fertilizer, and both manures) into some data structure that allows you to iterate over them?

RUFAS/general_constants.py
```diff
@@ -27,6 +27,14 @@ class GeneralConstants:
         PERCENTAGE_TO_FRACTION: The conversion factor from a percentage to a fraction.
         FRACTION_TO_PERCENTAGE: The conversion factor from a fraction to a percentage.
 
+        EMBEDDED_NITROGEN_FERTILIZER_EMISSIONS_FACTOR:
+         Constants for calculating the embedded emissions of synthetic nitrogen. Reference IPCC 2021.
+        EMBEDDED_PHOSPHORUS_FERTILIZER_EMISSIONS_FACTOR:
+         Constants for calculating the embedded emissions of synthetic PHOSPHORUS. Reference GWP 100.
+        EMBEDDED_POTASSIUM_FERTILIZER_EMISSIONS_FACTOR:
+         Constants for calculating the embedded emissions of synthetic POTASSIUM. Reference BASF's Eco-efficiency
+         analysis tool.
```
Isgotkowitz: These constants need units. Also, there is nothing wrong with putting them in the general constants file but my two cents is that this isn't necessary, because these are not general constants - they are very specific to the emissions module.
```suggestion
        EMBEDDED_NITROGEN_FERTILIZER_EMISSIONS_FACTOR : float
            Constants for calculating the embedded emissions of synthetic nitrogen. Reference IPCC 2021.
        EMBEDDED_PHOSPHORUS_FERTILIZER_EMISSIONS_FACTOR : float
            Constants for calculating the embedded emissions of synthetic phosphorus. Reference GWP 100.
        EMBEDDED_POTASSIUM_FERTILIZER_EMISSIONS_FACTOR : float
            Constants for calculating the embedded emissions of synthetic potassium. Reference BASF's Eco-efficiency
            analysis tool.
```

RUFAS/routines/EEE/emissions.py
matthew7838: This is part of the original code, I'm not quite sure. I can look into it during the testing PR.

matthew7838: Update: It should be fixed.

RUFAS/general_constants.py
matthew7838: I do have this debate in mind, thanks for your suggestions and I do think it makes sense.

RUFAS/routines/EEE/emissions.py
matthew7838: Thanks a lot for the clarifications.

changelog.md
```diff
@@ -22,6 +22,7 @@ v0.9.1
 ### Next Version Updates
 
 - [1968](https://github.com/RuminantFarmSystems/MASM/pull/1968) - [minor change] [Changelog] Move the changelog from a Google sheet into a markdown document in the repository.
+- [1977](https://github.com/RuminantFarmSystems/MASM/pull/1977) - [minor change] [EEE] Improve emission.py codebase and clarity.
 - [1948](https://github.com/RuminantFarmSystems/MASM/pull/1948) - [minor change] [Soil and Crop] Utilize a residue tracker property method in the layer data pool that sums structural and metabolic litter.
-- [1973](https://github.com/RuminantFarmSystems/MASM/pull/1973) - [minor change] [OutputManager] Avoid global variable usage of output manager [2/2].
+- [1973](https://github.com/RuminantFarmSystems/MASM/pull/1948) - [minor change] [OutputManager] Avoid global variable usage of output manager [2/2].
```
Isgotkowitz: The changelog entry pointing for 1973 should not be pointing to the PR for 1948.

RUFAS/routines/EEE/emissions.py
```diff
@@ -140,35 +106,47 @@ def _gather_homegrown_feeds_and_fertilizer_apps(
...
+        ]
 
-        filtered_fert_apps = list(
-            filter(
-                lambda app: Time.convert_year_jday_to_date(app["year"], app["day"]) >= date_cutoff, processed_fert_apps
-            )
-        )
+        results = []
+        for filter_def in filters:
+            filtered_data = self.filter_results(filter_def, date_cutoff, *filter_def["date_fields"])
```
ew3361zh: Just curious to hear about why you didn't make this function private like most of the others in this class. Is the idea that it could be extended for use outside emissions?

RUFAS/routines/EEE/emissions.py
Isgotkowitz: There should be an extra line here between the end of the returns section and the end of the docstring in order for it to comply with the Numpy style guide.

RUFAS/routines/EEE/emissions.py
```diff
@@ -140,35 +106,47 @@ def _gather_homegrown_feeds_and_fertilizer_apps(
...
-        )
+            results.append(filtered_data)
 
-        filtered_manure_requests = list(
-            filter(
-                lambda app: Time.convert_year_jday_to_date(app["year"], app["day"]) >= date_cutoff,
-                processed_manure_requests,
-            )
-        )
+        filtered_yields, filtered_fert_apps, filtered_manure_apps, filtered_manure_requests = results
```
ew3361zh: Just a thought...I don't know what the future of this module entails but returning this tuple of lists of dicts seems like it could be built out better to accommodate future extension. Maybe a dictionary of lists of dicts?

e.g.

```
def _gather_homegrown_feeds_and_fertilizer_apps(
    self,
) -> dict[str, list[dict[str, Any]]]:
    ...
    return {
        "homegrown_feeds": filtered_yields,
        "fertilizer_applications": filtered_fert_apps,
        "manure_applications": filtered_manure_apps,
        "manure_requests": filtered_manure_requests,
    }
```

And then up in `estimate_emissions()` it would look like:

```
results = self._gather_homegrown_feeds_and_fertilizer_apps()
self._calculate_purchased_feed_emissions(results["homegrown_feeds"])
self._calculate_homegrown_feed_emissions(
    results["homegrown_feeds"],
    results["fertilizer_applications"],
    results["manure_applications"],
    results["manure_requests"]
)
```

RUFAS/routines/EEE/emissions.py
```diff
@@ -140,35 +106,47 @@ def _gather_homegrown_feeds_and_fertilizer_apps(
...
+                "date_fields": ("year", "day"),
+            },
+            {
+                "name": "Manure Requests",
+                "description": "Collects all manure requests that occurred in the simulation.",
+                "filters": ["Field._record_manure_application\\.manure_request\\.field='.*'"],
+                "variables": [".*"],
+                "date_fields": ("year", "day"),
+            },
+        ]
```
Isgotkowitz: Instead of having this be a list, would you consider having it be a dictionary instead? The benefit is that it would be possible to explicitly map the what the filter is related to instead of having it be implicit through the ordering.

RUFAS/routines/EEE/emissions.py
```diff
@@ -140,35 +106,47 @@ def _gather_homegrown_feeds_and_fertilizer_apps(
...
-            filter(
-                lambda app: Time.convert_year_jday_to_date(app["year"], app["day"]) >= date_cutoff, processed_fert_apps
-            )
-        )
+        results = []
+        for filter_def in filters:
+            filtered_data = self.filter_results(filter_def, date_cutoff, *filter_def["date_fields"])
+            if filter_def["name"] == "Homegrown Feeds":
+                for crop in filtered_data:
+                    crop["total_dry_yield"] = crop["dry_yield"] * crop["field_size"]
```
Isgotkowitz: Instead of having this if statement on every single filter that it iterates over, it would be more efficient to have it operate on just the `filtered_yields` outside of this for loop.

RUFAS/routines/EEE/emissions.py
```diff
@@ -344,16 +322,10 @@ def _calculate_homegrown_feed_emissions(
...
-            aggregated_manure_apps[field_name]["nitrogen"] += app["nitrogen"]
-            aggregated_manure_apps[field_name]["phosphorus"] += app["phosphorus"]
+        self.aggregate_manure_operations(aggregated_manure_apps, manure_applications)
 
         aggregated_manure_requests = {key: {"nitrogen": 0.0, "phosphorus": 0.0} for key in all_fields}
-        for request in manure_requests:
-            field_name = request["field_name"]
-            aggregated_manure_requests[field_name]["nitrogen"] += app["nitrogen"]
-            aggregated_manure_requests[field_name]["phosphorus"] += app["phosphorus"]
+        self.aggregate_manure_operations(aggregated_manure_requests, manure_requests)
```
Isgotkowitz: There is still a lot of overlap between these aggregations of manure requests and applications that could be put into a method. They both create dictionaries of dictionaries that contain amounts of nitrogen and phosphorus. Same with the aggregation of fertilizer, except that it contains an extra key, potassium.

RUFAS/routines/EEE/emissions.py
```diff
@@ -410,7 +382,7 @@ def _calculate_homegrown_feed_emissions(
                 }
                 self.om.add_variable(f"homegrown_{crop_type}_emissions", emissions_info, info_map)
 
-    def _collect_target_soil_characteristics(self, field_names: list[str]) -> dict[str, float]:
+    def _collect_target_soil_characteristics(self, field_names: list[str]) -> dict[str, dict[str, Any]]:
```
ew3361zh, line 372: It seems like you could do something similar with this function as you did refactoring `_gather_homegrown_feeds_and_fertilizer_apps()` grouping the filters and looping through to add them to the soil_data.

ew3361zh (commented):
I think this looks good, I have a few questions and one piece where I think you could do a similar refactor as you've already done elsewhere in the class to eliminate duplicate code.

RUFAS/routines/EEE/emissions.py
```diff
@@ -531,3 +510,54 @@ def _calculate_emissions_by_field(
...
+        aggregated_manure: dict[str, dict[str, float]]
+            Data for tracking the amount of aggregated manure.
+        manure_operations: list[dict[str, Any]]
+            The manure operations that will aggregate manure.
+
+        """
+        for app in manure_operations:
+            field_name = app["field_name"]
+            aggregated_manure[field_name]["nitrogen"] += app["nitrogen"]
+            aggregated_manure[field_name]["phosphorus"] += app["phosphorus"]
```
Isgotkowitz: A couple of things about this method:
1. Why is it static? I am just curious
2. The top level description in the docstring needs to be more specific. It shouldn't say that this is a helper method, it should say what it helps with.
3. The docstring for `aggregated_manure` should be something more like "Tracks totals amounts of nutrients in the manure operations by field".
4. The docstring for `manure_operations` should be something more like "The manure operations that will be aggregated by field". Right now it makes it should like the operations will be acting upon something instead of them being acted upon.
5. This method should return `aggregated_manure`. Right now it is relying it being modified as a side effect of `aggregate_manure_operations`, this is an anti-pattern.

RUFAS/routines/EEE/emissions.py
```diff
@@ -531,3 +510,54 @@ def _calculate_emissions_by_field(
             crop["manure_nitrogen_requested"] = manure_requests["nitrogen"] * fraction_of_total_mass_grown
 
         return feeds_grown
+
+    def filter_results(
+        self, filters: dict[str, Any], date_cutoff: datetime, year: str, day: str
+    ) -> list[dict[str, Any]]:
+        """
+        Helper methods to process the filters for requests/applications.
```
Isgotkowitz: This should state what the methods helps with, not that it is a helper.

RUFAS/routines/EEE/emissions.py
```diff
@@ -531,3 +510,54 @@ def _calculate_emissions_by_field(
...
+
+        Parameters
+        ----------
+        filters: dict[str, Any]
+            A filter to collect the desired data from OutputManager.
+        date_cutoff: datetime
+            Date that indicates the cutoff of occurrence.
```
```suggestion
            Date before which all data collected from the Output Manager is filtered out.
```

RUFAS/routines/EEE/emissions.py
```diff
@@ -531,3 +510,54 @@ def _calculate_emissions_by_field(
             crop["manure_nitrogen_requested"] = manure_requests["nitrogen"] * fraction_of_total_mass_grown
 
         return feeds_grown
+
+    def filter_results(
+        self, filters: dict[str, Any], date_cutoff: datetime, year: str, day: str
```
Isgotkowitz: I think it would be better to name `year` and `day` as `year_key` and `day_key` instead.

Isgotkowitz:
Could you please add more detail to your testing section of the PR? Which simulations did you run to verify that the outputs were unchanged?

matthew7838:
> Could you please add more detail to your testing section of the PR? Which simulations did you run to verify that the outputs were unchanged?

Added, default run is what I used.

RUFAS/routines/EEE/emissions.py
matthew7838: Also replying to @Isgotkowitz, yes I should make them private. I make them static/public at first thinking that there might be possible extensions and forgot to change them to private once I eliminated the option.

RUFAS/routines/EEE/emissions.py
matthew7838: Thank you so much for the docstring correction, I was so confused on what to put in.

RUFAS/routines/EEE/emissions.py
matthew7838: I agree that seems like a more maintainable approach. Thanks.

RUFAS/routines/EEE/emissions.py
matthew7838: @ew3361zh Two out of the four filters are handled quite differently. Maybe it could improve extendability, but not much simplification IMO. Do you have a second thought?

Isgotkowitz:
> > Could you please add more detail to your testing section of the PR? Which simulations did you run to verify that the outputs were unchanged?
> 
> Added, default run is what I used.

Ideally the test plan should be reproducible, it currently is not. It says that outputs with the emissions keyword were filtered out, but this is not precisely reproducible. What was the filter you used to get the outputs you compared? Also it says that the outputs on the dev branch and this branch were compared, but it does not specify what the results of that comparison were. Were the results exactly the same with zero differences? Were the results a little different but the differences were all in an acceptable tolerance? Did the result of this comparison match what the expected result of the comparison was?

Also, I starting filtering some outputs and found that the default simulation does not fully run the emissions estimation routines (on both `emission-enhance` and `dev`). Using the below filter, it seems that it was not capturing the emissions associated with any farm grown feeds:
```json
{
    "multiple": [
        {
            "name": "Actual Purchased Feeds",
            "filters": [
                ".*actual_purchased_feed_totals.*"
            ],
            "variables": [".*"]
        },
        {
            "name": "Actual Purchased Feed Emissions",
            "filters": [
                ".*actual_purchased_feed_emissions.*"
            ],
            "variables": [".*"]
        },
        {
            "name": "Actual Land Use Change Feed Emissions",
            "filters": [".*actual_land_use_change_feed_emissions.*"],
            "variables": [".*"]
        },
        {
            "name": "Homegrown Feed Emissions",
            "filters": [".*homegrown_.*_emissions.*"]
        }
    ]
}
```
Edit: removed the homegrown feed totals because RuFaS did like it for some reason.

RUFAS/routines/EEE/emissions.py
ew3361zh: I think I'd try to eliminate duplicate code and loop through the filters similar to how you did in `_gather_homegrown_feeds_and_fertilizer_apps()`. Then there is duplicate code for how ammonia and no2 emissions are summed and duplicate code for how the 2 carbon stocks are summed. I think it would be extendible but also reduce repetition and follow a little closer to the refactoring you did in the rest of the class.

Isgotkowitz:
Update to my last comment: to get the emissions routines to more fully run, use the ARL simulation and set the end year of the simulation to be 2017 instead of 2019. Using the above filter, there is a difference between the requested manure nitrogen for corn grain between this branch and dev

RUFAS/routines/EEE/emissions.py
```diff
@@ -559,6 +531,72 @@ def _partition_applied_crop_fertilizer_emissions(
...
+        operations : list[dict[str, Any]]
+            A list of application dictionaries. Each dictionary contains the field name and nutrient amounts.
+            Example: {"field_name": "Field1", "nitrogen": 50.0, "phosphorus": 20.0}
+
+        all_fields : list[str]
+            A list of field names to aggregate data for, usually combining fields with crops and fields with
+            applications.
+
+        nutrients : list[str]
+            A list of nutrient keys to aggregate (e.g., ["nitrogen", "phosphorus", "potassium"]).
```
```suggestion
        operations : list[dict[str, Any]]
            A list of application dictionaries. Each dictionary contains the field name and nutrient amounts.
            Example: {"field_name": "Field1", "nitrogen": 50.0, "phosphorus": 20.0}
        all_fields : list[str]
            A list of field names to aggregate data for, usually combining fields with crops and fields with
            applications.
        nutrients : list[str]
            A list of nutrient keys to aggregate (e.g., ["nitrogen", "phosphorus", "potassium"]).
```

Isgotkowitz (approved):
Following up on my last comment, I think the calculations for manure requested on dev were incorrect, so this PR introduces a fix. Nice job Matthew! This PR is going to set us up well for implementing unit testing, which is going to be a huge improvement for this module.

RUFAS/routines/EEE/emissions.py
matthew7838: Niko, I've updated the emission PR, the extraction of the two carbon related function does not makes much sense since it still requires two cases to assign start and end value which does not eliminate the duplicate code and I don't think there's much extension possibility in future. I still extract the ammonia and nitrogen to eliminate some duplicate code and improve extensibility.

ew3361zh (approved):
Nice work. Tested this with soil emissions and compared with dev and results were holding consistent. Think it will be really helpful to move onto testing and make sure these functions are behaving as expected with different inputs.

KFosterReed (approved):
I tested this update on 4 pilot testing farms (c1f1, c1f3, c2f4, and c2f6) and the update to manure N in particular is working as expected.

PR #1979, opened 2024-09-06
Partition Synthetic Fert Emissions by Crop

### LINKED ISSUE #1753: Fertilizer Emissions for specific crops
We have worked through a process of aggregating emissions for each specific crop (yay!) but when there are two crops grown in 1 calendar year we had to come up with a solution for handling the emissions. We decided (in June) that we would allocate these emissions by % dry matter. Therefore, if there is a corn silage - triticale double crop and corn silage makes up 90% of the total mass of dry matter harvested off the field, corn silage will receive 90% of the emissions. This is fine logic, except when it comes to emissions associated with synthetic fertilizers. Synthetic fertilizers are applied to a specific crop and all emissions from this fertilizer need to go to the specific crop. In our above example, if nitrogen fertilizer is applied to corn silage but no fertilizer is applied to triticale, all of the emissions associated with the production of that synthetic fertilizer (embedded emissions) should go to corn, not be spread in the 90:10 ratio. 

## Expected Behavior
For synthetic fertilizer applications, they should be associated with a specific crop, not parsed out by % DM. 

## Current Behavior
Currently being allocated across crops by % DM

## Possible Solution
While % DM is a great solution for our other emissions for right now, perhaps using a date cutoff for embedded fertilizer emissions could be a solution. If corn silage is planted on day 135 and harvested on 270 and triticale is planted on 274 and harvest on 130 (of the following year), we could use the date of the fertilizer application (for example 155) to be mean it is for corn silage. There will be some times where it is applied at day 134 (just before planting) which could mess up my idea and the logic :( but I trust someone is smarter than me at this. 

## Context (Environment)
While not the largest portion of our footprints, it is an important one when we grow crops on dairy farms. This would just make us more accurate in our calculations. 

## Possible Implementation
not sure I know the difference between the implementation and the solution :( 

## Target Deadline
by Nov 2024 :)

ew3361zh:
This PR updates the partitioning of emissions associated with synthetic fertilizers from being by dry yield mass to being by crop planting and harvesting schedules.

### Context

Issue(s) closed by this pull request: closes #1753

### What

Within `EmissionsEstimator._calculate_emissions_by_field()`, there is new logic to partition emissions associated with synthetic fertilizer by crop planting and harvesting schedule. The logic is as follows:

1. If the fertilizer application date was between a crop planting and crop harvesting dates, all the fertilizer emissions will be associated with that crop. The logic is such that in the future, if RUFAS were to allow the planting of multiple crops in the same field at the same time and the fertilizer application date overlapped for multiple crops, the emissions would be split evenly between those crops.
2. If the fertilizer application date does not fall between any crop planting and crop harvesting dates, the emissions will be applied to the next crop planted in the field.
3. If there is no next crop planted, the emissions will not be associated with any crops and a warning will be logged to OM notifying the user that those emissions were not associated with any crop.

Emissions for the whole field and emissions associated with manure applications have been left as they were.

NOTE: currently the emissions not associated with any crop because the application date falls outside any of the above criteria is still associated with the field total emissions. If this is not desired behavior, I can adjust.

### Why

This method of emissions partitioning is more accurate than our current system of partitioning it by dry yield.

### How

- If there is no dry yield, there will be no emissions for any crops.
- Crops in the field are then sorted by planting year and date.
- For each fertilizer application, we check whether its date falls between any of the planting and harvesting schedules for the crops grown in the field.
- If there are any crops whose dates fulfill those requirements, we partition the emissions for that application accordingly.
- If there are no crops whose dates fulfill those requirements, we cycle through the list of crops and check which (if any) of the crops' planting dates fall soonest after the fertilizer application date. If there is a crop that fulfills this requirement, we partition all the fertilizer emissions to this crop.
- If there are no crops whose dates fulfill either of the previous two criteria, we log a warning to OM notifying the user that we didn't associated any of that fertilizer's emissions with any particular crop.

### Test plan

I've included some testing input files given to me by Kaitlyn which have Corn Grain and Triticale Silage plantings. 
In this simulation:
- Corn grain is harvested right before the first fertilizer application is made. 
- Triticale is planted the day after the first fertilizer application is made.
- The second fertilizer application occurs directly after the first Triticale harvest occurs.
- The second corn grain is planted directly after the second fertilizer application is made BUT the harvest of this corn grain happens after the end of the simulation so it does not count towards emissions totals.

So in the end, the triticale will get all the emissions from the first fertilizer application and the emissions from the second fertilizer application are not applied to any crop and a warning is logged to OM.

Playing around with these dates to compare how the emissions are partitioned is recommended.
Here is the output filter I used to look at emissions for these crops:
```
{
  "multiple": [
{
    "name": "Homegrown Feed Totals",
    "filters": [".*_calculate_total_homegrown_feed_amounts_by_crop_type.homegrown_feed_totals.*"]
},
{"name": "Homegrown Feed Emissions - Corn Grain",
    "filters": [".*homegrown_feed_emissions.homegrown_corn_grain_emissions.*"],
    "variables": [
        "nitrous_oxide_emissions",
        "ammonia_emissions",
        "carbon_stock_change",
        "nitrogen_fertilizer_used",
        "nitrogen_fertilizer_embedded_CO2_emissions",
        "phosphorus_fertilizer_used",
        "phosphorus_fertilizer_embedded_CO2_emissions",
        "potassium_fertilizer_used",
        "potassium_fertilizer_embedded_CO2_emissions",
        "manure_nitrogen_used",
        "manure_nitrogen_requested"
    ],
    "vertical_aggregation": "sum"
},
{"name": "Homegrown Feed Emissions - Triticale Silage",
    "filters": [".*homegrown_feed_emissions.homegrown_triticale_silage_emissions.*"],
    "variables": [
        "nitrous_oxide_emissions",
        "ammonia_emissions",
        "carbon_stock_change",
        "nitrogen_fertilizer_used",
        "nitrogen_fertilizer_embedded_CO2_emissions",
        "phosphorus_fertilizer_used",
        "phosphorus_fertilizer_embedded_CO2_emissions",
        "potassium_fertilizer_used",
        "potassium_fertilizer_embedded_CO2_emissions",
        "manure_nitrogen_used",
        "manure_nitrogen_requested"
    ],
    "vertical_aggregation": "sum"
}
  ]
}
```


RUFAS/routines/EEE/emissions.py
```diff
@@ -475,32 +468,34 @@ def _calculate_emissions_by_field(
...
-                crop["phosphorus_fertilizer_used"] = 0.0
-                crop["phosphorus_fertilizer_embedded_CO2_emissions"] = 0.0
-                crop["potassium_fertilizer_used"] = 0.0
-                crop["potassium_fertilizer_embedded_CO2_emissions"] = 0.0
-                crop["manure_nitrogen_used"] = 0.0
-                crop["manure_nitrogen_requested"] = 0.0
             return feeds_grown
 
-        for crop in feeds_grown:
+        sorted_crops = sorted(feeds_grown, key=lambda crop: (crop["planting_year"], crop["planting_day"]))
```
Isgotkowitz, line 499: I recommend converting planting year and planting day to a datetime.date instance and have those be used to sort things. I'm not sure that using tuples here wouldn't work but using date instances makes it much more robust.
```suggestion
        sorted_crops = sorted(feeds_grown, key=lambda crop: (crop["planting_year"], crop["planting_day"]))
```

RUFAS/routines/EEE/emissions.py
```diff
@@ -509,25 +504,107 @@ def _calculate_emissions_by_field(
...
+            crop_harvest_date = self._get_date(crop["harvest_year"], crop["harvest_day"])
+
+            if crop_planting_date <= fertilizer_application_date <= crop_harvest_date:
+                applied_crops.append(crop)
+        return applied_crops
+
+    @staticmethod
+    def _get_date(year: int, day: int) -> date:
+        """Helper function to convert year and day to a date object."""
+        return date(year, 1, 1) + timedelta(days=day - 1)
```
Isgotkowitz: There is a method for doing this in the `Time` class

RUFAS/routines/EEE/emissions.py
```diff
@@ -332,14 +333,6 @@ def _calculate_homegrown_feed_emissions(
...
 
         fields_with_crops = set(grouped_feeds.keys())
-        fields_with_fertilizer_apps = {app["field_name"] for app in fertilizer_applications}
-        all_fields = list(fields_with_fertilizer_apps | fields_with_crops)
-        aggregated_fertilizer_apps = {key: {"nitrogen": 0.0, "phosphorus": 0.0, "potassium": 0.0} for key in all_fields}
-        for app in fertilizer_applications:
-            field_name = app["field_name"]
-            aggregated_fertilizer_apps[field_name]["nitrogen"] += app["nitrogen"]
-            aggregated_fertilizer_apps[field_name]["phosphorus"] += app["phosphorus"]
-            aggregated_fertilizer_apps[field_name]["potassium"] += app["potassium"]
```
Isgotkowitz, line 342: I think the fertilizer applications still need to be grouped by field. Otherwise a fertilizer application from one field could be assigned to a crop that was grown in another

Isgotkowitz (commented):
This is a great start to more accurate partitioning of emissions. However there is a bug that is introduced when simulating more than one field, see comment below. 

To reproduce the bug, make a copy of the field blob in the metadata (and name it something like "field_2"), remove the summation from the filter then rerun the simulation and compare the results to the original. The amounts of fertilizer used for each harvest of the crop will be doubled (which is the buggy behavior) though oddly enough the amounts of embedded emissions seem to have stayed correct.

RUFAS/routines/EEE/emissions.py
ew3361zh: Can you expand a little on what converting these to datetime.date objects would do to make this more robust and in what cases we might want to have this robustness available here? I don't think it would be too difficult to do but it seems like an added step/overhead of object creation for a list-sort that's working as is.

RUFAS/routines/EEE/emissions.py
ew3361zh: Ah gotcha, I think the purpose of that grouping step got lost in the weeds for me (pause for laughter).

ew3361zh:
Changes addressing @Isgotkowitz's feedback:
- Time class jday conversion function used and get_date() method removed.
- field name added as filter for calculating fertilizer applications/emissions by field in calculate_emissions_by_field - this should resolve the issue of having fertilizer from multiple fields separated properly.

1 outstanding question is if converting the planting date to a datetime.date object for sorting the list of crops in a field is worth the overhead.

krb78:
> This is a great start to more accurate partitioning of emissions. However there is a bug that is introduced when simulating more than one field, see comment below.
> 
> To reproduce the bug, make a copy of the field blob in the metadata (and name it something like "field_2"), remove the summation from the filter then rerun the simulation and compare the results to the original. The amounts of fertilizer used for each harvest of the crop will be doubled (which is the buggy behavior) though oddly enough the amounts of embedded emissions seem to have stayed correct.

Good catch, Ed! But this is close Niko! When a single field is run, it works great with all sorts of iterations of changes (changing the date of the config, the dates of application, and changing crop harvest types).

RUFAS/routines/EEE/emissions.py
```diff
@@ -509,25 +506,107 @@ def _calculate_emissions_by_field(
...
+                                fertilizer_application["nitrogen"] * EMBEDDED_NITROGEN_FERTILIZER_EMISSIONS_FACTOR
+                            )
+                            next_crop["phosphorus_fertilizer_used"] += fertilizer_application["phosphorus"]
+                            next_crop["phosphorus_fertilizer_embedded_CO2_emissions"] = (
+                                fertilizer_application["phosphorus"] * EMBEDDED_PHOSPHORUS_FERTILIZER_EMISSIONS_FACTOR
+                            )
+                            next_crop["potassium_fertilizer_used"] += fertilizer_application["potassium"]
+                            next_crop["potassium_fertilizer_embedded_CO2_emissions"] = (
+                                fertilizer_application["potassium"] * EMBEDDED_POTASSIUM_FERTILIZER_EMISSIONS_FACTOR
+                            )
```
matthew7838: Would it be better to extract this part into a helper method? I think that could makes things cleaner.

matthew7838 (approved):
LGTM, I have small suggestions. Otherwise, things are working well and logically.

RUFAS/routines/EEE/emissions.py
Isgotkowitz: To me, sorting a list of tuples is not as intuitive as sorting a list of dates. So if there is a strange edge case that would mess with the sorting routine, I expect it would happen with something less intuitive like tuples than it would with dates, especially because dates have a library dedicated to making sure they can captured and used correctly.

This was my gut reaction to reading this code, in part because I introduced a serious bug into a different part of this module by trying to compare dates more directly with numbers than with date objects. I looked up how Python sorts tuples just now and I think your method will be bug free.

> I don't think it would be too difficult to do but it seems like an added step/overhead of object creation for a list-sort that's working as is.

This is a very good point, I think either way would be fine but sticking with the tuples would have that edge of being faster.

RUFAS/routines/EEE/emissions.py
```diff
@@ -509,25 +506,114 @@ def _calculate_emissions_by_field(
...
+        self,
+        fertilizer_application: dict[str, Any],
+        sorted_crops: list[dict[str, Any]],
+        fertilizer_application_date: datetime.date,
+    ) -> bool:
+        """
+        Applies fertilizer to the next available crop after harvest and before the next planting.
+        Returns True if fertilizer was applied, False otherwise.
+        """
+        for i, crop in enumerate(sorted_crops):
```
```suggestion
        for index, crop in enumerate(sorted_crops):
```

RUFAS/routines/EEE/emissions.py
```diff
@@ -509,25 +506,114 @@ def _calculate_emissions_by_field(
...
+            crop_harvest_date = Time.convert_year_jday_to_date(crop["harvest_year"], crop["harvest_day"])
+            if i + 1 < len(sorted_crops):
```
Isgotkowitz: It would be good to extract the condition for this if statement into its own boolean variable.

RUFAS/routines/EEE/emissions.py
```diff
@@ -509,25 +506,114 @@ def _calculate_emissions_by_field(
...
+        Parameters
+        ----------
+        sorted_crops : list[dict[str, Any]]
+            The list of crops in the field sorted by planting date.
+        fertilizer_application_date : date
+            The date of the fertilizer applciation.
+
+        Returns
+        -------
+        list
```
```suggestion
        list[dict[str, Any]]
```

RUFAS/routines/EEE/emissions.py
```diff
@@ -509,25 +506,114 @@ def _calculate_emissions_by_field(
...
+                        fertilizer_application["nitrogen"] * EMBEDDED_NITROGEN_FERTILIZER_EMISSIONS_FACTOR
+                    )
+                    next_crop["phosphorus_fertilizer_used"] += fertilizer_application["phosphorus"]
+                    next_crop["phosphorus_fertilizer_embedded_CO2_emissions"] = (
+                        fertilizer_application["phosphorus"] * EMBEDDED_PHOSPHORUS_FERTILIZER_EMISSIONS_FACTOR
+                    )
+                    next_crop["potassium_fertilizer_used"] += fertilizer_application["potassium"]
+                    next_crop["potassium_fertilizer_embedded_CO2_emissions"] = (
+                        fertilizer_application["potassium"] * EMBEDDED_POTASSIUM_FERTILIZER_EMISSIONS_FACTOR
+                    )
```
```suggestion
                    next_crop["nitrogen_fertilizer_used"] += fertilizer_application["nitrogen"]
                    next_crop["nitrogen_fertilizer_embedded_CO2_emissions"] += (
                        fertilizer_application["nitrogen"] * EMBEDDED_NITROGEN_FERTILIZER_EMISSIONS_FACTOR
                    )
                    next_crop["phosphorus_fertilizer_used"] += fertilizer_application["phosphorus"]
                    next_crop["phosphorus_fertilizer_embedded_CO2_emissions"] += (
                        fertilizer_application["phosphorus"] * EMBEDDED_PHOSPHORUS_FERTILIZER_EMISSIONS_FACTOR
                    )
                    next_crop["potassium_fertilizer_used"] += fertilizer_application["potassium"]
                    next_crop["potassium_fertilizer_embedded_CO2_emissions"] += (
                        fertilizer_application["potassium"] * EMBEDDED_POTASSIUM_FERTILIZER_EMISSIONS_FACTOR
                    )
```
The emissions should be increased in the same fashion as the nutrient amounts, not overwritten

RUFAS/routines/EEE/emissions.py
```diff
@@ -509,25 +506,114 @@ def _calculate_emissions_by_field(
...
+            crop["nitrogen_fertilizer_used"] += fertilizer_application["nitrogen"] * split_factor
             crop["nitrogen_fertilizer_embedded_CO2_emissions"] = (
-                fertilizer_applications["nitrogen"]
-                * fraction_of_total_mass_grown
-                * EMBEDDED_NITROGEN_FERTILIZER_EMISSIONS_FACTOR
+                fertilizer_application["nitrogen"] * split_factor * EMBEDDED_NITROGEN_FERTILIZER_EMISSIONS_FACTOR
             )
-            crop["phosphorus_fertilizer_used"] = fertilizer_applications["phosphorus"] * fraction_of_total_mass_grown
+            crop["phosphorus_fertilizer_used"] += fertilizer_application["phosphorus"] * split_factor
             crop["phosphorus_fertilizer_embedded_CO2_emissions"] = (
```
```suggestion
            crop["phosphorus_fertilizer_embedded_CO2_emissions"] += (
```

RUFAS/routines/EEE/emissions.py
```diff
@@ -509,25 +506,114 @@ def _calculate_emissions_by_field(
...
-                fertilizer_applications["phosphorus"]
-                * fraction_of_total_mass_grown
-                * EMBEDDED_PHOSPHORUS_FERTILIZER_EMISSIONS_FACTOR
+                fertilizer_application["phosphorus"] * split_factor * EMBEDDED_PHOSPHORUS_FERTILIZER_EMISSIONS_FACTOR
             )
-            crop["potassium_fertilizer_used"] = fertilizer_applications["potassium"] * fraction_of_total_mass_grown
+            crop["potassium_fertilizer_used"] += fertilizer_application["potassium"] * split_factor
             crop["potassium_fertilizer_embedded_CO2_emissions"] = (
```
```suggestion
            crop["potassium_fertilizer_embedded_CO2_emissions"] += (
```

RUFAS/routines/EEE/emissions.py
```diff
@@ -509,25 +506,114 @@ def _calculate_emissions_by_field(
...
+        self, fertilizer_application: dict[str, float], applied_crops: list[dict[str, Any]]
+    ) -> None:
+        """
+        Partitions synthetic emissions from fertilizer applications to crop(s) whose planting and harvesting dates
+        encompass the fertilizer application date.
+        """
+        for crop in applied_crops:
+            split_factor = 1 / len(applied_crops)
+            crop["nitrogen_fertilizer_used"] += fertilizer_application["nitrogen"] * split_factor
             crop["nitrogen_fertilizer_embedded_CO2_emissions"] = (
```
```suggestion
            crop["nitrogen_fertilizer_embedded_CO2_emissions"] += (
```

RUFAS/routines/EEE/emissions.py
Isgotkowitz: Instead of passing the list of unfiltered fertilizer applications into `_calculate_emissions_by_field` along with the field name and filtering in that method, what about filtering them here (where the old code is) and passing just the filtered list into `_calculate_emissions_by_field`? I think the advantage of that is all the aggregations by field name would be kept in the same place.

> (pause for laughter)

Nice

RUFAS/routines/EEE/emissions.py
```diff
@@ -509,25 +506,114 @@ def _calculate_emissions_by_field(
...
+        list
+            A list of the crops whose planting and harvesting dates encompass the fertilizer application date.
+        """
+        applied_crops = []
+
+        for crop in sorted_crops:
+            crop_planting_date = Time.convert_year_jday_to_date(crop["planting_year"], crop["planting_day"])
+            crop_harvest_date = Time.convert_year_jday_to_date(crop["harvest_year"], crop["harvest_day"])
+
+            if crop_planting_date <= fertilizer_application_date <= crop_harvest_date:
```
```suggestion
            if crop_planting_date <= fertilizer_application_date < crop_harvest_date:
```
If a fertilizer application occurred on the same day that a crop was harvested, I think it would be more reasonable to assume that the fertilizer application should not be assigned to that crop and instead to the next. @krb78 what do you think?

Isgotkowitz (commented):
This is working much better, but unfortunately I found another bug - when multiple fertilizer applications are assigned to the same crop, the amount of nutrients goes up as expected but the amounts of embedded emissions associated with those nutrients do not. I think this is because the embedded emissions amounts are assigned rather than used to increase the amount already in a crop - changing this is suggested in some of comments I left

krb78:
Replying via email, sorry guys, but yes, if fertilizer is applied the day
the crop is harvested, it should go toward the next crop’s footprint.
Hopefully not a scenario we will see much, but good to catch.

Kaitlyn Briggs, DVM, MBA
(607)-386-0040

ew3361zh:
> Instead of passing the list of unfiltered fertilizer applications into _calculate_emissions_by_field along with the field name and filtering in that method, what about filtering them here (where the old code is) and passing just the filtered list into _calculate_emissions_by_field? I think the advantage of that is all the aggregations by field name would be kept in the same place.

My reasoning here was 2 fold - 
First that we're no longer aggregating the fertilizer applications in the same way we are for manure so it's not as aligned with that filtering process as it was in the previous arrangement. 
Second is that it seemed unnecessarily complicated to set up logic to create a new data structure that solely rearranges the same information (since it's no longer serving that aggregation function) while really only filtering out a small handful of key-value pairs. It opens the door for error and produces unnecessary overhead to try to align the structure.

I do understand the appeal of aligning those structures but it almost to me seemed helpful to not have them align since they serve different purposes and represent different ways of partitioning emissions down the road.

Isgotkowitz (approved):
I think this provides the desired functionality and covers the expected case well. However I tried intercropping (two crops in the same field at the same time) and it seemed like the nutrients and emissions from fertilizer were only attached to one of the crops, instead of split between the two.

krb78 (approved):
I APPROVE! I tried this with all sorts of iterations of dates fertilizer is applied and whether crops are harvested or just killed and it worked as it should :)

PR #1980, opened 2024-09-09
Add warning to _determine_clay_silt_ratio_factor

### LINKED ISSUE #1978: Error prevents simulation of soil with Zero clay and Zero silt
Soil is made up of a combination of sand, silt, and clay. Farming on soil that is 100% sand is not realistic but there are some use cases where we would like to simulate these outcomes anyway. Right now, if a soil has 0 clay and 0 silt, an error is produced in `soil_erosion.py` by the method `determine_clay_silt_ratio_factor()`. 

## Expected Behavior
The model should allow users to simulate soils with 0 clay and 0 silt. 

## Possible Solution

In the method `determine_clay_silt_ratio_factor()`, instead of producing the error  `if silt_fraction == 0 and clay_fraction == 0`, produce a warning that says "Warning: both silt and clay fractions are 0 which will lead to unreliable predictions of erosion and soil emissions" and return a value of 1 for the `clay_silt_factor`
## Steps to Reproduce

1. Run the default scenario with ` "clay_fraction": 0,
            "silt_fraction": 0,` as inputs in default_soil.json
2.
3.
4.

## Context (Environment)

FARM-ES users sometimes like to protect the identity and location of the farms and so do not provide the correct coordinates for those farms. This means that sometimes the soil inputs are not values for soils where people actually farm. They know it will affect the accuracy of the results but it should be their decision to make, not ours.

matthew7838:
Instead of throwing an error, add a warning and return 1 when silt and clay are zero in `_determine_clay_silt_ratio_factor` of `soil_erosion.py`.
### Context

Issue(s) closed by this pull request: closes #1978

### What

Instead of throwing an error, add a warning and return 1 when silt and clay are zero in `_determine_clay_silt_ratio_factor` of `soil_erosion.py`.

### Why

FARM-ES users sometimes like to protect the identity and location of the farms and so do not provide the correct coordinates for those farms. This means that sometimes the soil inputs are not values for soils where people actually farm. They know it will affect the accuracy of the results but it should be their decision to make, not ours.

### Test plan

- Test suites created and run.
- Have SME's second look.


RUFAS/routines/field/soil/soil_erosion.py
```diff
@@ -168,9 +169,20 @@ def _determine_clay_silt_ratio_factor(silt_fraction: float, clay_fraction: float
         SWAT Theoretical documentation eqn. 4:1.1.7
 
         """
+        om = OutputManager()
```
Isgotkowitz: It would be more efficient to have the Output Manager instantiated inside the if statement, because then it will only be instantiated when it is needed.

RUFAS/routines/field/soil/soil_erosion.py
```diff
@@ -168,9 +169,20 @@ def _determine_clay_silt_ratio_factor(silt_fraction: float, clay_fraction: float
...
-            raise ValueError("Cannot have fractions of silt and clay in the soil both be 0")
-        return (silt_fraction / ((clay_fraction) + (silt_fraction))) ** 0.3
+            info_map = {
+                "class": SoilErosion.__name__,
+                "function": SoilErosion._determine_clay_silt_ratio_factor.__name__,
+            }
+            om.add_warning(
+                "Silt and clay fractions are 0",
+                "Both silt and clay fractions are 0 which will lead to unreliable predictions of erosion and soil"
+                " emissions",
```
Isgotkowitz: When raising a warning or error it is good to not just list the cause but also the effect. For example:
```suggestion
                "Silt and clay fractions in the soil are 0, which will lead to unreliable predictions of erosion and soil"
                " emissions",
                "It is assumed that the ratio of clay to silt in the soil layer will not have any effect on the amount of erosion from the soil.",
```

RUFAS/routines/field/soil/soil_erosion.py
```diff
@@ -168,9 +169,20 @@ def _determine_clay_silt_ratio_factor(silt_fraction: float, clay_fraction: float
...
         """
+        om = OutputManager()
         if silt_fraction == 0 and clay_fraction == 0:
-            raise ValueError("Cannot have fractions of silt and clay in the soil both be 0")
-        return (silt_fraction / ((clay_fraction) + (silt_fraction))) ** 0.3
+            info_map = {
+                "class": SoilErosion.__name__,
+                "function": SoilErosion._determine_clay_silt_ratio_factor.__name__,
+            }
+            om.add_warning(
```
Isgotkowitz: Instead of raising this warning in the method `_determine_clay_silt_ratio_factor`, it would be better to raise it somewhere else where it will only be raised once for a soil profile (possibly post init in SoilData). By raising it every single time that this method runs, this warning will be logged on a daily basis if the silt and clay fractions are 0 resulting in the same warning being raised over and over again. It also dilutes other warnings that are generated by RuFaS.

Isgotkowitz (approved):
Prevents RuFaS from crashing and logs a warning as requested in the issue. However, I think there are some improvements that could make this solution a bit more efficient and user-friendly. These are listed in the comments.

Also, I would consider logging an error instead of a warning. Having zero clay and silt in a soil profile indicates a pretty unrealistic soil profile and deserves an error instead of a warning IMO. However I think the SMEs should have final say over this.

RUFAS/routines/field/soil/soil_erosion.py
matthew7838: Good suggestion.

RUFAS/routines/field/soil/soil_erosion.py
matthew7838: It is explicitly asked for a warning in the issue, I think when this case occurs the user knows what they're doing. The reason we have this issue to begin with is due to some farm privacy issue in FARM-ES.

RUFAS/routines/field/soil/soil_erosion.py
matthew7838: Again, I'm a little mixed about this since the user should know what they're doing at this point, but adding details is never bad.

RUFAS/routines/field/soil/soil_erosion.py
KFosterReed: I think Ed's point about not raising it in the method is a good one that I did not consider when I made the issue - we don't need a warning for every day and layer, just one is enough I think

matthew7838: Sorry I misread the comment, now looking back at it again I agree. I will do so.

ew3361zh (approved):
LGTM

PR #1983, opened 2024-09-10
Cache GH Action Dependencies between jobs.

### LINKED ISSUE #1561: [GH Actions] Adjust and remove the second repo checkout and dependencies installation
Two of the most time-consuming steps in the GitHub workflows are the repo checkout step and the installation of dependencies.

If there is no easy way to share those two steps between the two jobs - format-code and lint-test-type-check, we may have to combine everything into one big job. 

The expectation is to save about 1 minute every time (from 4.5 minutes down 3-3.5 minutes).

ew3361zh:
Updates the GH action to cache the dependencies installed for the first job to be accessed by the second job.

### Context

Issue(s) closed by this pull request: closes #1561

### What

Uses GH actions/cache@v3 to cache the dependencies installed from the first job and then accesses those dependencies for quicker install for the second job.

### Why

This will hopefully speed up the time it takes to run our GH actions both saving time and the action run minutes of which we have a limited number.

### How

Added the GH step using actions/cache@v3 before dependencies are installed in both jobs of the action which is the order of operations according to the GH documentation.

### Test plan

Going to run it a few times and compare the time difference in running the action with and without caching. Additionally the test will be whether the second job can successfully find the cached dependencies.


ew3361zh:
Closing this - results of research into time saving GH action changes are documented in accompanying issue #1561.

PR #1985, opened 2024-09-11
Extract Input data to CSV file

### LINKED ISSUE #1961: Extract Input data to CSV file
I would like to be able to extract the input data from multiple difference input files into a CSV file. This would be similar to the current metadata properties file but instead of the metadata information in the collumns next to the variable names, each column would have the input values for those variables. 

It would be great to have a new task that will enable this to be done repeatedly in the future but the goal would be to pass a set of metadata files and get a CSV file that has the the input file names, the variable names in that input file, and the values for each of those variables. 

Note that this does not need to include the values in any of the input csv files like the feed library, weather files, or feed emissions data

If this can't be made into a task, it would be great to just have the data from the pilot testing farms extracted. 

**Additional context**

**Target Deadline**
Is it possible to have it done by the end of next week (Sept 6)? :)

allisterakun:
### Context

Issue(s) closed by this pull request: closes #1961

### What

This PR extracts all the input data values (for all input variables from JSON input files) for each task and combines them into a single CSV file for side-by-side comparison.

### Why

To directly compare the difference of input values for different tasks.

### How

1. Introduced two variables for task_manager_metadata:
    ```JSON
    "export_input_data_to_csv": {
      "type": "bool",
      "description": "Select this flag to export the input data pool into a CSV file for all tasks.",
      "default": false
    },
    "input_data_csv_export_path": {
      "type": "string",
      "description": "The path to save the exported input data CSV file.",
      "default": "output/saved_input_data/"
    },
    "input_data_csv_import_path": {
      "type": "string",
      "description": "If user wish to compare the input data with previous runs, the user can pass in the previously saved input data CSV path. The result will be combined together for a side-by-side comparison.",
      "default": ""
    },
    ```
2. Introduced a new property, `csv_report_generation_list`, in `InputManager` to keep track of all the file blobs that we are interested in exporting. It is simply the file blob key for all the JSON input files.
3. Introduced a new function, `export_pool_to_csv()` in `InputManager` to export all the interested variables and their values into a CSV.
4. In `TaskManager.handle_post_processing()`, if the `"export_input_data_to_csv"` is true, it will call `Utility.combine_saved_input_csv()` to combine all the exported CSVs for every single task into one single CSV for a direct comparison.
5. Two functions, `flatten_dictoionary()` and `combine_saved_input_csv()`, were introduced in `Utility` to provide the functionalities needed for this task.

### Test plan

1. Updated all affected tasks.
2. Write new unit tests for the three newly added functions.
3. Tested on the default_metadata and barnyard_metadata, the output is as expected.


input/data/tasks/default_task.json
allisterakun, line 1: Changes to this file will be reverted before merging. This is just for testing.

KFosterReed:
This works pretty well so far! I tested it on several pilot farms (even with crops now!) and it collects all the inputs into an organized format that can easily be used for either recording and further analysis. One question I do have is that if there is already a 'saved_input_csv' file in the file path provided as an input, running it again will add new collumns. I think this is a great idea because wanted to ask for thoughts on the labeling of these collumns. Right now, it is using the user provided output prefix, which is great and works really well when the provided prefixes are different but if the user duplicates them, it looks like it appends a suffix '_x', '_y', and so on.

It seems like an OK first step and there isn't an obvious alternative without going to a full time stamp which I don't think we want? Any thoughts on this front?

matthew7838 (approved):
LGTM, tried running things according to test plans and it seems right on my side.

allisterakun:
> This works pretty well so far! I tested it on several pilot farms (even with crops now!) and it collects all the inputs into an organized format that can easily be used for either recording and further analysis. One question I do have is that if there is already a 'saved_input_csv' file in the file path provided as an input, running it again will add new collumns. I think this is a great idea because wanted to ask for thoughts on the labeling of these collumns. Right now, it is using the user provided output prefix, which is great and works really well when the provided prefixes are different but if the user duplicates them, it looks like it appends a suffix '_x', '_y', and so on.
> 
> It seems like an OK first step and there isn't an obvious alternative without going to a full time stamp which I don't think we want? Any thoughts on this front?

I guess we can modify the code to make this behavior more predictable:
Before adding a new entry to the existing 'saved_input_csv' file, I can check if a column with the current output prefix name already exists. If not, we will proceed as usual. If it does, we will append a suffix `_1` and check if `output_prefix_1` already exists. If it also exists, we will keep increasing the number until we don't have a prior entry. 
Does this sound like a plan?

KFosterReed:
> > This works pretty well so far! I tested it on several pilot farms (even with crops now!) and it collects all the inputs into an organized format that can easily be used for either recording and further analysis. One question I do have is that if there is already a 'saved_input_csv' file in the file path provided as an input, running it again will add new collumns. I think this is a great idea because wanted to ask for thoughts on the labeling of these collumns. Right now, it is using the user provided output prefix, which is great and works really well when the provided prefixes are different but if the user duplicates them, it looks like it appends a suffix '_x', '_y', and so on.
> > It seems like an OK first step and there isn't an obvious alternative without going to a full time stamp which I don't think we want? Any thoughts on this front?
> 
> I guess we can modify the code to make this behavior more predictable: Before adding a new entry to the existing 'saved_input_csv' file, I can check if a column with the current output prefix name already exists. If not, we will proceed as usual. If it does, we will append a suffix `_1` and check if `output_prefix_1` already exists. If it also exists, we will keep increasing the number until we don't have a prior entry. Does this sound like a plan?

That does sound like a good plan!

KFosterReed:
I tried running the updates with this task file: 
image

but did not get the compiled file as I expected. Instead I got a directory called `saved_pool_working_folder` and individual csvs for each of the 3 metadata in the task files: 
image

I also tried this with the `"SIMULATION_SINGLE_RUN"` task type and it produced the same thing. I am guessing this is user error but not sure what I'm doing wrong

allisterakun:
> I tried running the updates with this task file: image

I set the "input_data_csv_import_path" to an empty string and it fixed the issue.
Because this is the first run for extracting the input, we don't have a valid previous CSV result to import and combine with.

KFosterReed:
Hmmm, I tried it with an empty string and without the argument `input_data_csv_import_path` at all and got the same thing as above :/ I am going to try it on another machine and see if it works there

allisterakun:
> Hmmm, I tried it with an empty string and without the argument `input_data_csv_import_path` at all and got the same thing as above :/ I am going to try it on another machine and see if it works there

I pushed the changes just now when I replied with the comment. Did you also update the branch?

KFosterReed:
> > Hmmm, I tried it with an empty string and without the argument `input_data_csv_import_path` at all and got the same thing as above :/ I am going to try it on another machine and see if it works there
> 
> I pushed the changes just now when I replied with the comment. Did you also update the branch?

Sorry about that! I didn't realize you made changes. Now it works as expected! Could you please update the Task Manager Wiki with descriptions for how the new inputs/methods work? (Alternatively, you could make an issue to remember the Wiki needs to be updated if you don't have time to do it right away). 

Great work! Thanks Allister!

KFosterReed (approved):
This now works as expected!

allisterakun:
> > > Hmmm, I tried it with an empty string and without the argument `input_data_csv_import_path` at all and got the same thing as above :/ I am going to try it on another machine and see if it works there
> > 
> > 
> > I pushed the changes just now when I replied with the comment. Did you also update the branch?
> 
> Sorry about that! I didn't realize you made changes. Now it works as expected! Could you please update the Task Manager Wiki with descriptions for how the new inputs/methods work? (Alternatively, you could make an issue to remember the Wiki needs to be updated if you don't have time to do it right away).
> 
> Great work! Thanks Allister!

The [TaskManager Wiki](https://github.com/RuminantFarmSystems/MASM/wiki/Task-Manager#:~:text=for%20Sensitivity%20Analysis.-,export_input_data_to_csv,-%3A%20A%20boolean%20value) has been updated

PR #1986, opened 2024-09-11
Biophysical folder setup for Animal module

### LINKED ISSUE #1965: Bio-physical Setup
Allister will make the biophysical folder and the new structure for the animal module.

allisterakun:
### Context

Issue(s) closed by this pull request: closes #1965 

Creates the folders and files for the redesigned animal module.


Isgotkowitz (approved):
Sets up the new animal folder correctly. The issue does not specify this, but does this also need a copy of the manure folder? I thought it might because we are also trying to introduce really big changes to the manure module?

allisterakun:
> Sets up the new animal folder correctly. The issue does not specify this, but does this also need a copy of the manure folder? I thought it might because we are also trying to introduce really big changes to the manure module?

Yes, but this issue is only for the animal module setup, we will set up the other modules as we go. I only included one file, feed.py from the feed module b/c it is imported to the ration submodule in animal.

ew3361zh (approved):

PR #1989, opened 2024-09-11
Assigning correct ration attribute when non lac cow formulation fails

JoeWaddell:
### Context

- [ ] Yes
- [x] No

### What

When ration formulation fails for non lactating cow pens, the previous formulation interval's ration should be used. See the related issue below for other heifer ration formulation improvements that need to be considered in other PRs. rehttps://github.com/RuminantFarmSystems/MASM/issues/577

However: currently the ration selected is one that is summed for the entire pen, not the "per animal" ration.

Also note: 
The attribute `pen.ration` should perhaps be renamed for clarity in the future, but this is outside the scope of this quick fix. Opening #1990 to handle this related concern.

### Why

This error causes a rapid increase in phosphorus requirements, since this massively increases the amount fed to each animal (which is used in the calculation of `p_maint_feces` in the `phosphorus_rqmts` method.

### Test plan

Before/after comparison of growing animal pens captured in pilot test farm c1f3. Used report filter below, provided by @KFosterReed 

```
{
    "multiple": [

        {
            "name": "Phos Diet Amount",
            "filters": ["AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_._GROWING"],
          "variables": ["phosphorus"],
            "slice_start": -365
           },
        {
            "name": "Phos Req NASEM",
            "filters": ["AnimalModuleReporter.report_ration_interval_data.avg_rqmts_pen_._GROWING"],
            "variables": ["P_req", "P_req_process", "avg_BW"],
            "slice_start": -365
           },{
            "name": "Phos Diet Amount-Daily",
            "filters": ["AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_._GROWING"],
          "variables": ["phosphorus"]
           },
        {
            "name": "Phos Req NASEM-Daily",
            "filters": ["AnimalModuleReporter.report_ration_interval_data.avg_rqmts_pen_._GROWING"],
            "variables": ["P_req", "P_req_process", "avg_BW"]
           },{
            "name": "Avg Bodyweight",
            "filters": ["AnimalModuleReporter.report_ration_interval_data.avg_rqmts_pen_._GROWING"],
            "variables": ["P_req", "P_req_process"]
           }

    ]
}
```

### DIFF
--- a/RUFAS/biophysical/animal/ration/ration_driver.py
+++ b/RUFAS/biophysical/animal/ration/ration_driver.py
@@ -143,8 +143,8 @@ class RationManager:
             return ration, ration_vals
         # safeguard if scipy SLSQP bounds error still occurs after many iterations
         # using previous cycles ration for this pen
-        elif pen.ration != {}:
-            return pen.ration, ration_vals
+        elif pen.ration_per_animal != {}:
+            return pen.ration_per_animal, ration_vals
         else:
             om.add_error(
                 "No previous ration available",
--- a/RUFAS/biophysical/animal/ration/ration_driver.py
+++ b/RUFAS/biophysical/animal/ration/ration_driver.py
@@ -66,7 +66,7 @@ class RationManager:
         Returns
         -------
         Dict[str, float]
-            Formulated ration.
+            Formulated ration for a single animal.
         Dict[str, float]
             Summary of ration content.
         """
@@ -80,10 +80,10 @@ class RationManager:
         # Use grouping scenario to find the type of each animal in pen
         req.set_requirements(pen, animal_grouping_scenario, False)
         if udrm.use_user_defined_ration:
-            ration, ration_vals = cls.get_user_defined_ration(
+            ration_per_animal, ration_vals = cls.get_user_defined_ration(
                 req, pen, available_feeds, animal_grouping_scenario, sim_day
             )
-            return ration, ration_vals
+            return ration_per_animal, ration_vals
 
         previous_ration = None
         if hasattr(pen, "ration_per_animal"):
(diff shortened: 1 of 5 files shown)


KFosterReed:
This is a straightforward bug fix and works as expected now! Thanks Joe!

KFosterReed (approved):

RUFAS/biophysical/animal/ration/ration_driver.py
```diff
@@ -143,8 +143,8 @@ def formulate_ration(
             return ration, ration_vals
```
Isgotkowitz:
```suggestion
            return ration_per_animal, ration_vals
```

RUFAS/biophysical/animal/ration/ration_driver.py
```diff
@@ -143,8 +143,8 @@ def formulate_ration(
...
         # safeguard if scipy SLSQP bounds error still occurs after many iterations
         # using previous cycles ration for this pen
-        elif pen.ration != {}:
-            return pen.ration, ration_vals
+        elif pen.ration_per_animal != {}:
+            return pen.ration_per_animal, ration_vals
```
Isgotkowitz:
```suggestion
        elif previous_ration:
            return previous_ration, ration_vals
```
Might be better to reuse the variable that has ration per animal. Or, if `pen.ration_per_animal` is kept it might be good to check that the previous ration is present in pen (and not None), this is accomplished earlier in the code with `hasattr`.

RUFAS/biophysical/animal/ration/ration_driver.py
Isgotkowitz, line 1: For `formulate_ration`, it would be good to 1) update the "Returns" section of the docstring to indicate specifically that the first dictionary returned is the ration for an individual animal and 2) double check that is actually the case for all the return statements in `formulate_ration`.

Isgotkowitz (approved):
Made some suggestions for how the code could be improved, but in this case it might be best to stick with the minimal changes necessary.

RUFAS/biophysical/animal/ration/ration_driver.py
JoeWaddell: Great suggestion, changed the return from `make_ration_from_solution` accordingly.

PR #2001, opened 2024-09-16
[E2E Testing] Updates E2ETestResultsComparer so that it works on Windows machines

Isgotkowitz:
Fixes the `E2ETestResultsComparer` so it runs on Windows machines.

### Context

Issue(s) closed by this pull request: closes # N/A

### What

When the E2ETestResultsComparer runs, it tried to construct file paths using "/" as the file path separator. This works on Unix-like systems (Linux and MacOS), but not Windows.

### Why

End-to-end testing needs to run properly on both Windows and Unix-like machines.

### How

The target file path is sanitized with `re.escape` before looking for matching file paths.

### Test plan

Ran the test suite on both Linux and Windows machines to check that they are running properly. Also ran end to end testing on both Linux and Windows systems to make sure that it was passing (as it was expected to be).

The command to run end-to-end testing currently is `python main.py -p input/metadata/end_to_end_testing_tm_metadata.json`.

### DIFF
--- a/RUFAS/e2e_test_results_comparer.py
+++ b/RUFAS/e2e_test_results_comparer.py
@@ -8,7 +8,6 @@ from RUFAS.output_manager import OutputManager
 from RUFAS.units import MeasurementUnits
 
 import json
-import re
 
 
 ResultPathType = namedtuple("ResultPaths", ["domain", "expected_results_path", "actual_results_path"])
@@ -46,8 +45,8 @@ class E2ETestResultsComparer:
             )
             path_to_actual_results = None
             for path in json_output_path.iterdir():
-                escaped_path = re.escape(str(json_output_path / path_set.actual_results_path))
-                is_a_match = re.match(f"{escaped_path}.*", str(path))
+                actual_results_base_path = Path(path_set.actual_results_path)
+                is_a_match = path.name.startswith(actual_results_base_path.name)
                 if is_a_match:
                     path_to_actual_results = path
                     break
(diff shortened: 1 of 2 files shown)


RUFAS/e2e_test_results_comparer.py
```diff
@@ -46,7 +46,8 @@ def compare_actual_and_expected_test_results(json_output_path: Path) -> None:
             )
             path_to_actual_results = None
             for path in json_output_path.iterdir():
-                is_a_match = re.match(f"{json_output_path}/{path_set.actual_results_path}.*", str(path))
+                escaped_path = re.escape(str(json_output_path / path_set.actual_results_path))
+                is_a_match = re.match(f"{escaped_path}.*", str(path))
```
allisterakun: Can the two paths be converted into Path objects and compared using functionalities from the Path library instead of relying on regex matching?

allisterakun (approved):
I have a comment about a possible alternative fix; otherwise, it looks good!

I tested it on my Windows machine, and the fix works as expected. When I tried reverting the changes to what they were before, the program crashed with an error, as expected.

RUFAS/e2e_test_results_comparer.py
Isgotkowitz: Good thought @allisterakun, this is possible and it works great on both Windows and Linux

matthew7838 (approved):
LGTM, great idea to implement this by utilizing the path object.

Issue #1879, opened 2024-08-01
[Manure] Manure storage N2O emissions fix


elle-andreen:
## Expected Behavior
Estimated N2O emissions from manure storage that were implemented in PR #1203 (also see issue #1199 for the original request and info on the equations). Per that issue, this method is a strictly empirical approach drawn from IPCC (2019) Tier 1 recommendations that assumes a fixed fraction of **manure N entering the manure treatment system each day** that will be lost as N2O-N. 

## Current Behavior
A function exists (`_calc_empirical_nitrogen_loss_from_nitrous_oxide_emission`) in the slurry storage underfloor, slurry storage outdoor, anaerobic lagoon, and anaerobic digestion and lagoon manure treatment files to calculate N2O-N emissions. Currently, the input value for manure N in this equation is the daily output value for the manure treatment liquid manure N. This is incorrect - the daily output value for these manure treatments currently includes a subtraction of manure storage ammonia N emissions, which should not be happening. Functionally, this results in an underestimation of manure storage N2O emissions. The function should be based on the daily entry of manure N into manure storage and should take the daily **input** value of manure N, not the output. 

## Possible Solution
This is a very easy fix - in the treatment files mentioned above, plus the anaerobic digestion treatment file (the numerical outcome is correct in this treatment, but the function should still be changed for consistency), modify the manure N input value for `_calc_empirical_nitrogen_loss_from_nitrous_oxide_emission` to be `daily_input.liquid_manure_nitrogen`, instead of `daily_output.liquid_manure_nitrogen`. 

![image](https://github.com/user-attachments/assets/ce550f3f-1673-41af-a2c5-0022fd5f1fba)
Example from `anaerobic_lagoon.py` of the function currently.

I also wonder, because `_calc_empirical_nitrogen_loss_from_nitrous_oxide_emission` is reused across multiple files, if it should be relocated to the `calculator.py` file where other manure gas emissions equations are located. Or maybe there was a compelling reason not to do that!

elle-andreen:
![image](https://github.com/user-attachments/assets/fd576ec6-a9c9-4119-b2e9-ba50f9170132)

[...]

PR #2000, opened 2024-09-16
remove optional type hint

matthew7838:
Remove the optional type hint from current_day_condition.py.
### Context

Issue(s) closed by this pull request: closes #1939 

### What

Several attributes in CurrentDayConditions are typed as Optional[...] when they should not be. The annual_mean_air_temperature will remain optional with the default None value because it will be later calculated and assigned in weather.py after all the mean air temperature has been loaded to initiate each day's current_day_condition. Therefore, placeholders are needed. Let me know if it's better to remove the optional but keep the default `None`.

### Why

The type hints should accurately reflect which types we expect variables to have in RuFaS.

### How

Remov the type hints.

### Test plan

- Updated and run the test suite.
- Run with default, barnyard, and available simulation metadata, and nothing is broken.

PR #2002, opened 2024-09-16
Use Barn Temp To Calculate Manure Handler Emisssions

### LINKED ISSUE #1905: [Manure] Add bounds to barn indoor air temperature
Air temperature is a component of equations determining methane and ammonia emissions from manure indoors (i.e. housing emissions). Currently, air temperature in the barn is assumed to be equal to air temperature outside the barn. However, barn temperature is more regulated than outdoor temperature, and equations were not developed using temperature extremes like those found outdoors. Implementing simple minimum and maximum bounds would ensure barn temperatures (and thus housing emissions) stay within realistic ranges:

If outdoor temp < 5 C, barn temp = 5 C
If outdoor temp > 5 but < 30 C, barn temp = outdoor temp 
If outdoor temp > 30 C, barn temp = 30 C

ew3361zh:
Adds functionality to convert ambient air temperature from current day conditions into a more accurate representation of barn temperature to more accurately model conditions where the manure handlers operate.

### Context

Issue(s) closed by this pull request: closes #1905

### What

Adds a function to the GasEmissionsCalculator class to convert air temperature to the approximate corresponding barn temperature. This is currently only used by ManureHandler subclasses but will be needed by a couple storages as well (in issue #1879) so it makes sense to not encapsulate it in the ManureHandler base class.

Barn temperature is also added as an output to ManureHandlerDailyOutput for reporting purposes.

### Why

The conversion of the air temperature to barn temperature creates a more accurate temperature to use for the calculation of emissions in ManureHandlers.

### How

The function to convert air temperature to the corresponding estimated barn temperature was set up in GasEmissionsCalculator.

### Test plan

Unit testing is updated. Will need outputs confirmation from SME to ensure they now reflect closer to expected results.


RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -1343,3 +1343,24 @@ def empirical_nitrogen_loss_from_nitrous_oxide_emission(
         """
 
         return emission_factor_kg_nitrous_oxide_N_per_kg_manure_N * manure_nitrogen_kg_N_per_day
+
+    @staticmethod
+    def adjust_air_temperature(air_temp: float) -> float:
```
elle-andreen: Completely optional suggestion, I would name this function something like "get_barn_air_temperature" or something that is a bit more specific to barn temperature.

ew3361zh: I like that.

RUFAS/routines/manure/manure_handlers/milking_parlor.py
```diff
@@ -22,7 +22,7 @@ def __init__(
         minutes_spent_per_milking=7.0,
         wash_water_use_rate=20.0,
         fresh_water_use_rate=10.0,
-    ):
+    ) -> None:
```
ew3361zh, line 25: Addresses a mypy error of no return specified for this class init function (referenced in manure_handler_classes.py).

elle-andreen:
- Strictly functionally speaking, barn temp is being restricted as it should. This in turn impacts housing emissions, since temp is a predictor variable for housing ammonia and methane emissions.
- I ran 3 weather scenarios (WI, northeast, and TX) using defaults for ration (with user input), manure management scenarios for each pen type, etc., everything except weather, to see how this PR affects manure emissions compared to current behavior. 
- **For closeup and growing animals, the housing emissions response is a bit outsized** because stall #'s are not in line with animal numbers in this specific example (my bad). So my comments here are focusing on lactating cow emissions.
- Housing methane: colder climates like WI and the NE will have slightly higher (13-14%) housing CH4 emissions with this method. On a net basis, this change equates to **less than 3 additional grams per cow per day** of housing methane. 
- Housing CO2 and storage methane are virtually unchanged between branches. 
- Housing ammonia: colder climates will have 6-7% greater housing ammonia, or **2-5 additional grams of ammonia N per cow per day**. 2-5 g is less than 1% of the total manure N excreted by these cows. This does result in a downstream decrease in storage N emissions, but, it is small, less than 2 g of additional ammonia N emission per cow per day. With 500 lactating cows, that would equate to something like 1 additional kg/cow/year of storage NH3-N.
- The greater impact in colder-temp situations vs. hotter temps is because the upper limit (30 C) here is more conservative than the lower temp limit (5 C). This is probably best for now; heat abatement and temperature-humidity index (THI) values can vary a lot between farms and will be more important in hot-weather regions, so let's not stress the upper limit too much until we have more responsive heat abatement modeling. For cold temperatures, our lower limit is more likely to be relatively accurate due to animal body heat warming the barn, and also farms not wanting their waterers and things to freeze indoors. 

All in all, I think this change sets us up very well code-wise to hopefully implement more precise barn temperature estimation in the future, and is a step in the right direction without changing emissions too drastically.

image

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -1343,3 +1343,24 @@ def empirical_nitrogen_loss_from_nitrous_oxide_emission(
...
         return emission_factor_kg_nitrous_oxide_N_per_kg_manure_N * manure_nitrogen_kg_N_per_day
+
+    @staticmethod
+    def determine_barn_air_temperature(air_temperature: float) -> float:
+        """Adjusts the air temperature to be more reflective of the range found inside a barn.
+
+        Parameters
+        ----------
+        air_temperature : float
+            The air temperature.
```
Isgotkowitz:
```suggestion
            The air temperature (degrees C).
```

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -1343,3 +1343,24 @@ def empirical_nitrogen_loss_from_nitrous_oxide_emission(
...
+
+        Returns
+        -------
+        float
+            The barn temperature.
+        """
```
Isgotkowitz:
```suggestion
            The barn temperature (degrees C).

        """
```

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -1343,3 +1343,24 @@ def empirical_nitrogen_loss_from_nitrous_oxide_emission(
         """
 
         return emission_factor_kg_nitrous_oxide_N_per_kg_manure_N * manure_nitrogen_kg_N_per_day
+
+    @staticmethod
+    def determine_barn_air_temperature(air_temperature: float) -> float:
+        """Adjusts the air temperature to be more reflective of the range found inside a barn.
```
Isgotkowitz:
```suggestion
        """Determines the ambient inside barn temperature based on the outdoor air temperature.
```
I think it would be good to emphasize that a new value is being calculated here, instead of it being possible to interpret this as recalculating an existing temperature.

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -1343,3 +1343,24 @@ def empirical_nitrogen_loss_from_nitrous_oxide_emission(
...
+        -------
+        float
+            The barn temperature.
+        """
+        if air_temperature < 5:
+            return 5.0
+        elif 5 <= air_temperature <= 30:
+            return air_temperature
+        else:
+            return 30.0
```
Isgotkowitz: Numpy's `clip` method could be used here, which would eliminate the need for the if-else block. Might not necessarily be the best option here though, the if-else block is better if there is any chance a more dynamic range might be implemented or if warnings would need to be logged here.

RUFAS/routines/manure/gas_emissions/calculator.py
Isgotkowitz, line 1370: Is there an explanation or reference that could be put in this docstring? On their own, the numbers 5 and 30 feel very magic numbery.

RUFAS/routines/manure/manure_handlers/manure_handler_daily_output.py
```diff
@@ -103,10 +103,14 @@ class ManureHandlerDailyOutput(LiquidManurePortionProtocol):
...
+    air_temp: float
         Temperature of the current day, C.
-    tempC_unit: MeasurementUnits
-        Unit for tempC
+    air_temp_unit: MeasurementUnits
+        Unit for air_temp.
+    barn_temp: float
+        Temperature in the barn, C.
+    barn_temp_unit: MeasurementUnits
+        Unit for barn_temp.
```
Isgotkowitz:
```suggestion
    air_temp_unit: MeasurementUnits, default MeasurementUnits.DEGREES_CELSIUS
        Unit for air_temp.
    barn_temp: float
        Temperature in the barn, C.
    barn_temp_unit: MeasurementUnits, default MeasurementUnits.DEGREES_CELSIUS
        Unit for barn_temp.
```

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
```diff
@@ -92,7 +92,7 @@ def __init__(
         self.config = manure_handler_config
         self.milking_parlor = MilkingParlor()
 
-    def _get_current_day_average_temperature_in_celsius(self) -> float:
+    def _get_current_day_average_temperature_in_celsius(self) -> float | None:
```
Isgotkowitz: I assume Mypy flagged this as an error, but with #2000 I think it should go away.

Isgotkowitz (approved):
Nice job Niko, changes are straightforward and well-documented. Left improvements for a couple things, please address them as you see fit.

There are two points that I think are somewhat important and should be considered before merging. The first (in comment below) is that bounds of 5 and 30 currently feel like magic numbers, it would be good to either document them explicitly or make them user inputs.

The second is that this inference of barn temp is very important because it directly modifies user input to estimate impacts on the farm while being somewhat buried in the code. I think the worst case scenario would be barn temp gets modeled in the Animal module too at a later date, we forget this barn temp method is here and then there are two different methods in two different parts of the codebase calculating the same thing. Reporting the barn temp to OM is a good step to alleviating this problem, but maybe it would be good to do more? Making the bounds a user-input would help, adding a warning that always runs once would also help.

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
ew3361zh: #2000 leaves `mean_air_temperature` as optional "because it will be later calculated and assigned in weather.py after all the mean air temperature has been loaded to initiate each day's current_day_condition."

RUFAS/routines/manure/manure_handlers/manure_handler_daily_output.py
ew3361zh: None of the other type hints for measurement units in ManureHandlerDailyOutput is specified so I may go through and add those here for consistency.

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
Isgotkowitz: That is the annual_mean_air_temperature, this one is the daily value which was changed to not be optional.

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -1343,3 +1343,24 @@ def empirical_nitrogen_loss_from_nitrous_oxide_emission(
...
+        Parameters
+        ----------
+        air_temperature : float
+            The air temperature (°C).
+
+        Returns
+        -------
+        float
+            The barn temperature (°C).
+        """
```
elle-andreen, line 1370: The following can be added in terms of context/citation:

"Between 5 and 30 C, barn temperature is assumed to be equal to outdoor air temperature. This function assumes that barn temperature does not drop below 5 C or increase above 30 C. These bounds were suggested by manure SMEs and are supported by barn temperature ranges reported in Bucklin et al. (FL, upper limit; https://doi.org/10.13031/2013.28851). The lower bound (5 C) suggested by SMEs was based on general industry standards/conditions."

elle-andreen:
Thanks for your thoughtful comments @Isgotkowitz, I added some further information on where this method originated from that can be added to the docstring. You are correct that we may at some point have reason to determine barn temperature in the animal module - this would be probably in the context of determining heat stress/heat abatement for cows. I would not consider this method to be necessarily modifying a user input however. The user is asked to provide ambient outdoor temperature information, not barn temperature information specifically. In many cases barn temperature tracks along with outdoor temperature, but is often not the same. I do think we should investigate the possibility of adding inputs for barn temperature minimum/maximum in the future, we could use the bounds in this PR as the default for those inputs. This PR is intended to be more of an intermediate step forward, until we can invest the time into determining how feasible this information would be as an input.

changelog.md
```diff
@@ -27,13 +27,13 @@ v0.9.1
...
 - [1784](https://github.com/RuminantFarmSystems/MASM/pull/1784) - [minor change] [Task Manager][Feed Storage] Adds end-to-end testing routine for the Feed Storage module which can be extended to test other parts of RuFaS.
 - [1998](https://github.com/RuminantFarmSystems/MASM/pull/1998) - [minor change] [GitHub Actions] Adds if statement to make sure there is a PR to comment on before attempting to post a comment.
 - [2000](https://github.com/RuminantFarmSystems/MASM/pull/2000) - [minor change] [CurrentWeatherConditions] Removed optional type hints.
 - [1989](https://github.com/RuminantFarmSystems/MASM/pull/1989) - [minor change] [Animal] Method formulate_ration now correctly returns pen.ration_per_animal instead of pen.ration.
 - [2001](https://github.com/RuminantFarmSystems/MASM/pull/2001) - [minor change] [E2E Testing] Changes E2ETestResultsComparer so that it constructs file paths properly on Windows machines.
-
+- [2002](https://github.com/RuminantFarmSystems/MASM/pull/2002) - [minor change] [Manure] Adds conversion of air temperature to barn temperature for use in calculations of emissions associated with Manure Handlers.
```
elle-andreen: I would go with Ed's phrasing here: "Adds a method to determine the ambient inside barn temperature based on the outdoor air temperature for use in calculations of emissions associated with Manure Handlers."

elle-andreen (approved):
I left a few final suggestions based on Ed's comments, but in general this looks good! Thanks Niko!

RUFAS/routines/manure/gas_emissions/calculator.py
ew3361zh: We can use the clip method for now - wouldn't be too hard to switch it back to an if-else block later if we needed to do more in this function.

RUFAS/routines/manure/gas_emissions/calculator.py
ew3361zh: Added Elle's comments from below.

PR #2004, opened 2024-09-17
Implement Animal digest

### LINKED ISSUE #1927: [Animal][DigestiveSystem] DigestiveSystem Refactor Implementation [3/3]
Follow the Design in this Google Doc: [Digestive System](https://docs.google.com/document/d/1vY7EoFhiwWhCO9f_BNn01K0c_DohBlbZqdEEIBXgHG4/edit?usp=sharing) to implement the refactor of the DigestiveSystem submodule.

matthew7838:
This PR implements AnimalDigest submodule logic according to the design doc and creates unit tests for the module.
### Context

Issue(s) closed by this pull request: closes #1927 

### What

This PR refactors the animal digestion code in the animal module. The original code has a digestion-related calculation logic for dry/ lactating cows, heifers,s, and calves. Now they're extracted into three classes, `EntericMethaneCalculator`, `ManureExcretionCalculator`, and `MethaneMitigationCalculator`. Finally, they will be called in the `DigestiveSystem` class for daily updates depending on the animal types passed in from the properties.

### Why

To resolve the entanglement issue in Animal module.

### How

Separate methods in to different classes.

### Test plan

Write necessary unit tests to bring the coverage for the animal growth submodule to 100%.

Unit tests cover all methods, and all mock test cases run error-free. At this point, we are not able to run it in a simulation, because the animal module in the new `biophyscial` module is yet to be set up.


RUFAS/biophysical/animal/animal_properties/general_properties.py
```diff
@@ -41,13 +41,29 @@ class GeneralProperties:
     id: int
```
Isgotkowitz, line 55: Could the attributes `breed` and `gender` please be updated to reflect that there are now enums for these fields, and that gender should be sex?

RUFAS/biophysical/animal/animal_properties/milk_production_properties.py
```diff
@@ -55,3 +57,4 @@ class MilkProductionProperties:
     wood_m: float
     wood_n: float
     milk_production_history: list[MilkProductionRecord]
+    daily_milk_production: float
```
Isgotkowitz: This attribute is already in `GeneralProperties`, and it should only be in one place.

RUFAS/biophysical/animal/animal_properties/nutrient_properties.py
```diff
@@ -23,3 +24,5 @@ class NutrientProperties:
     ration_phosphorus_concentration: float
     phosphorus_for_gestation: float
     phosphorus_for_gestation_required_for_calf: float
+    fecal_phosphorus: float
+    urine_phosphorus_required: float
```
Isgotkowitz, line 30: Could you please start a docstring for this class? It doesn't have to document all the attributes, just the ones you added would be enough to get it started.

RUFAS/biophysical/animal/animal_properties/nutrient_properties.py
Isgotkowitz, line 1: I see a number of attributes in this class that are already in `GeneralProperties`. They should be removed from this class.

RUFAS/biophysical/animal/digestive_system/enteric_methane_calculator.py
```diff
@@ -0,0 +1,331 @@
...
+
+class EntericMethaneCalculator:
+    @staticmethod
+    def calf_methane(methane_model: str | None, body_weight: float) -> float:
+        """
+        Calculates the amount of methane emission for calf.
+
+        Parameters
+        ----------
+        methane_model: str
```
Isgotkowitz:
```suggestion
        methane_model: str | None
```
I think it would be good to note in this docstring (probably for this parameter) that if the methane_model is none then the methane emissions are set as zero

RUFAS/biophysical/animal/digestive_system/enteric_methane_calculator.py
```diff
@@ -0,0 +1,331 @@
...
+                    NDF_concentration,
+                    EE_concentration,
+                    starch_concentration,
+                    methane_mitigation_method,
+                    methane_mitigation_additive_amount,
+                )
+
+            methane_emission = (
+                methane_yield
+                * (1 + methane_yield_reduction / GeneralConstants.FRACTION_TO_PERCENTAGE)
```
Isgotkowitz:
```suggestion
                * (1 + methane_yield_reduction * GeneralConstants.PERCENTAGE_TO_FRACTION)
```
All of the constants in `GeneralConstants` are/should be multiplicative

RUFAS/biophysical/animal/digestive_system/methane_mitigation_calculator.py
```diff
@@ -0,0 +1,49 @@
...
+        ----------
+        NDF_concentration : float
+            Concentration of neutral detergent fiber (NDF) in the ration.
+        EE_concentration : float
+            Concentration of ether extract (EE) in the ration.
+        starch_concentration : float
+            Concentration of starch in the ration.
+        methane_mitigation_method: str
+            Methane mitigation method used to reduce enteric methane emissions, including "3-NOP", "Monensin",
+            "EssentialOils", and "Seaweed".
```
Isgotkowitz: If the methane mitigation method is Essential Oils or Seaweed, then the methane is never going to be mitigated at all. Should there be a warning logged if a user tries to use one of these methods to let them know that they are not gonna be effective at all? Or maybe we just remove these options entirely?

@KFosterReed - if I remember your practice presentation correctly, you said that there wasn't enough information in the literature to model the mitigation from Essential Oils and Seaweed? If this is the case then I think it would be better to remove these options entirely

changelog.md
```diff
@@ -42,8 +42,7 @@ v0.9.2
...
 - [1784](https://github.com/RuminantFarmSystems/MASM/pull/1784) - [minor change] [Task Manager][Feed Storage] Adds end-to-end testing routine for the Feed Storage module which can be extended to test other parts of RuFaS.
 - [1998](https://github.com/RuminantFarmSystems/MASM/pull/1998) - [minor change] [GitHub Actions] Adds if statement to make sure there is a PR to comment on before attempting to post a comment.
-- [2000](https://github.com/RuminantFarmSystems/MASM/pull/2000) - [minor change] [CurrentWeatherConditions] Removed optional type hints.
-- [1989](https://github.com/RuminantFarmSystems/MASM/pull/1989) - [minor change] [Animal] Method formulate_ration now correctly returns pen.ration_per_animal instead of pen.ration.
 - [2001](https://github.com/RuminantFarmSystems/MASM/pull/2001) - [minor change] [E2E Testing] Changes E2ETestResultsComparer so that it constructs file paths properly on Windows machines.
+- [2004](https://github.com/RuminantFarmSystems/MASM/pull/2004) - [minor change] [Animal] Implement the animal digestive system in the biophysical folder.
```
Isgotkowitz: There should be no removals of changelog entries. Also, the entry for this PR needs to go in the "Next Version Updates" section.

RUFAS/biophysical/animal/digestive_system/enteric_methane_calculator.py
```diff
@@ -0,0 +1,331 @@
+import numpy
```
Isgotkowitz: Instead of importing the whole module, what about just importing `exp` because that is the only function that is needed for this method?

RUFAS/biophysical/animal/digestive_system/enteric_methane_calculator.py
```diff
@@ -0,0 +1,331 @@
...
+        """
+        methane_emission = 0.0
+        if methane_model:
+            # Default: IPCC Tier 2
+            CP_concentration = nutrient_concentrations["CP"]
+            EE_concentration = nutrient_concentrations["EE"]
+            NDF_concentration = nutrient_concentrations["NDF"]
+            ASH_concentration = nutrient_concentrations["ash"]
+            soluble_residue = (
+                (GeneralConstants.FRACTION_TO_PERCENTAGE - ASH_concentration)
```
Isgotkowitz: The use of this constant does not make sense in this line

RUFAS/biophysical/animal/digestive_system/enteric_methane_calculator.py
```diff
@@ -0,0 +1,331 @@
...
+
+        Returns
+        -------
+        float
+            Amount of methane emission for heifer (g/day).
+
+        """
+        methane_emission = 0.0
+        if methane_model:
+            # Default: IPCC Tier 2
```
Isgotkowitz: What is the default here? What are the alternatives? How are they set?

RUFAS/biophysical/animal/animal_properties/general_properties.py
```diff
@@ -41,13 +41,29 @@ class GeneralProperties:
...
     is_pregnant: bool
     mature_body_weight: float
-    nutrients: list[str]
+    nutrients: dict[str, float]
+    nutrient_concentrations: dict[str, float]
     ration_formulation = {"objective": 0.00}
     sold: bool
     sold_at_day: int
     wean_weight: float
+    metabolizable_energy_intake: float
```
Isgotkowitz, line 63: For the attributes that you are adding/modifying, could you please add them to the docstring?

RUFAS/biophysical/animal/digestive_system/digestive_system.py
```diff
@@ -1,2 +1,110 @@
...
+        cls.METHANE_MITIGATION_ADDITIVE_AMOUNT = (animal_config)["methane_mitigation"][
+            "methane_mitigation_additive_amount"
+        ]
+
+    @staticmethod
+    def process_digestion(
+        general_properties: GeneralProperties,
+        animal_nutrient_property: NutrientProperties,
+        milk_production_properties: MilkProductionProperties,
+    ) -> tuple[float, float, AnimalManureExcretions]:
```
Isgotkowitz: I don't think this method conforms to the pattern laid out for the animal refactor. Instead of returning floats for the enteric methane and excreted phosphorus, it should store them in the appropriate properties structure and return the modified properties

RUFAS/biophysical/animal/digestive_system/enteric_methane_calculator.py
```diff
@@ -0,0 +1,331 @@
+import numpy
+
+from RUFAS.biophysical.animal.digestive_system.methane_mitigation_calculator import MethaneMitigationCalculator
+from RUFAS.general_constants import GeneralConstants
+
+
+class EntericMethaneCalculator:
+    @staticmethod
+    def calf_methane(methane_model: str | None, body_weight: float) -> float:
```
Isgotkowitz:
```suggestion
    def calculate_calf_methane(methane_model: str | None, body_weight: float) -> float:
```
These method names should start with verbs

RUFAS/biophysical/animal/digestive_system/enteric_methane_calculator.py
```diff
@@ -0,0 +1,331 @@
...
+            Amount of methane emission for heifer (g/day).
+
+        """
+        methane_emission = 0.0
+        if methane_model:
+            # Default: IPCC Tier 2
+            CP_concentration = nutrient_concentrations["CP"]
+            EE_concentration = nutrient_concentrations["EE"]
+            NDF_concentration = nutrient_concentrations["NDF"]
+            ASH_concentration = nutrient_concentrations["ash"]
```
Isgotkowitz: I think these variable names should be changed, so that they are not abbreviated (CP -> crude_protein) and use snake case (ASH -> ash)

RUFAS/biophysical/animal/digestive_system/enteric_methane_calculator.py
```diff
@@ -0,0 +1,331 @@
...
+        """
+        dry_matter_intake = nutrient_amounts["dm"]
+        ASH_concentration = nutrient_concentrations["ash"]
+        ADF_concentration = nutrient_concentrations["ADF"]
+        CP_concentration = nutrient_concentrations["CP"]
+        NDF_concentration = nutrient_concentrations["NDF"]
+        EE_concentration = nutrient_concentrations["EE"]
+        starch_concentration = nutrient_concentrations["starch"]
+        methane_emission = 0.0
+        if methane_model == "Mutian":  # [A.3E.C.1]
```
Isgotkowitz: I think these in-line references should be removed and references should go in the docstring

Isgotkowitz (changes requested):
Left comments on a bunch of different things, but this PR is headed in the right direction. The most important things I noticed were 1) attributes being duplicated across different properties, and 2) the main routine for digestive system should be modifying and returning one or more of the properties dataclasses. Also, this PR is pretty massive. Maybe it would be better to implement something like the manure excretion changes in a PR stacked on top of this one?

RUFAS/biophysical/animal/digestive_system/enteric_methane_calculator.py
matthew7838: Looking at the code, I don't think there's any alternatives for now. So it will always be the IPCC Tier 2 method, which is the calculations below, I will remove these inline comments to prevent confusion and also move the references as mentioned in the comments below.

RUFAS/biophysical/animal/digestive_system/enteric_methane_calculator.py
matthew7838: Is there a reference available for me to understand these abbreviations?

Isgotkowitz: Might be somewhere in the animal scientific documentation (https://3.basecamp.com/3486446/buckets/5296287/vaults/7108677681)? If you can't find it there, the asking the SMEs would be a good way to go

RUFAS/biophysical/animal/animal_properties/general_properties.py
```diff
@@ -16,36 +30,36 @@ class GeneralProperties:
...
-    breed: Enum
+    breed: Breed
     culled: bool
     daily_growth: float
     days_born: int
     days_in_preg: int
     events: AnimalEvents
     days_in_milk: int
     dry_off_day_of_pregnancy: int
-    daily_milk_produced: float
```
Isgotkowitz, line 36: `daily_milk_produced` is used in `AnimalPhosphorus`, so removing it from `GeneralProperties` will cause those routines to fail.

RUFAS/biophysical/animal/animal_properties/general_properties.py
```diff
@@ -5,6 +5,20 @@
...
 
+class Breed(Enum):
+    """Enum indicating the breed of the animal."""
+
+    HO = "Holstein"
+    JE = "Jersey"
+
+
+class Sex(Enum):
+    """Enum indicating the gender of the animal."""
```
Isgotkowitz:
```suggestion
    """Enum indicating the sex of the animal."""
```

RUFAS/biophysical/animal/animal_properties/general_properties.py
```diff
@@ -16,36 +30,37 @@ class GeneralProperties:
         Number of days the animal has been milking for in its current lactation.
     dry_off_day_of_pregnancy : int
         Number of days into pregnancy when a lactating cow stops milking.
-    daily_milk_produced : float
-        Milk production of the animal on a single day (kg).
-    is_milking
```
Isgotkowitz: These attributes should not be deleted

RUFAS/biophysical/animal/digestive_system/enteric_methane_calculator.py
```diff
@@ -0,0 +1,337 @@
...
+        Parameters
+        ----------
+        body_weight: float
+            Body weight of the current cow (kg).
+        methane_model: str
+            Methane model used for methane emission calculations, including "Mutian", "Mills", "IPCC".
+        is_lactating: bool
+            Indicator of cow's lactating status.
+        milk_fat: float
+            Milk fat (from animal input), % of milk.
```
Isgotkowitz:
```suggestion
            Milk fat, % of milk.
```
Right now this value comes from the animal input, but this may not always be the case. And because the source of this value doesn't change how this method works, I don't think it is necessary to include it.

RUFAS/biophysical/animal/digestive_system/enteric_methane_calculator.py
```diff
@@ -0,0 +1,337 @@
...
+                "dm": "kg/animal",
+                "CP": "percent of DM",
+                "ADF": "percent of DM",
+                "NDF": "percent of DM",
+                "lignin": "percent of DM",
+                "ash": "percent of DM",
+                "phosphorus": "percent of DM",
+                "potassium": "percent of DM",
+                "N": "percent of DM",
+            }
```
Isgotkowitz:
```suggestion
        Notes
        -----
        The dry matter ("dm") unit is kg per animal. Crude protein ("CP"), ADF, NDF, lignin, ash, phosphorus, potassium, and nitrogen ("N") are all percentages of dry matter.

```
I think it is not a good practice to put python in the docstring, unless it is in the Examples section

RUFAS/biophysical/animal/animal_properties/general_properties.py
```diff
@@ -16,36 +30,39 @@ class GeneralProperties:
         Number of days the animal has been milking for in its current lactation.
     dry_off_day_of_pregnancy : int
         Number of days into pregnancy when a lactating cow stops milking.
+    metabolizable_energy_intake : float
+        Metabolizable energy intake, Mcal/kg dry matter.
     daily_milk_produced : float
         Milk production of the animal on a single day (kg).
-    is_milking
```
Isgotkowitz, line 21: Why is this taken out?

RUFAS/biophysical/animal/digestive_system/digestive_system.py
```diff
@@ -1,2 +1,115 @@
...
+from RUFAS.biophysical.animal.digestive_system.manure_excretion_calculator import ManureExcretionCalculator
+from RUFAS.data_structures.animal_manure_excretions import AnimalManureExcretions
+from RUFAS.input_manager import InputManager
+
+
 class DigestiveSystem:
-    pass
+    METHANE_MODEL: str
+    METHANE_MITIGATION_METHOD: str
+    METHANE_MITIGATION_ADDITIVE_AMOUNT: float
```
Isgotkowitz, line 31: Could docstrings please be added for these class vars?

Also, for stuff like the model and mitigation method would it be possible to convert them to enums? This doesn't have to happen in this PR, but it would be a nice change to make after the animal refactor if it is possible.

RUFAS/biophysical/animal/digestive_system/manure_excretion_calculator.py
```diff
@@ -0,0 +1,762 @@
...
+        nutrient_concentrations = nutrient_conc
+        dry_matter_intake = nutrient_amounts["dm"]
+        crude_protein_concentration = nutrient_concentrations["CP"]
+        potassium_concentration = nutrient_concentrations["potassium"]
+
+        # Soluble residue
+        # Dietary percentage of soluble residues, % DM, in the note of [A.3B.C.2]
+
+        # Total urine, kg [A.3B.A.1]
+        urine = 9.0
```
Isgotkowitz: It would be good to declare urine as a module constant.

RUFAS/biophysical/animal/digestive_system/methane_mitigation_calculator.py
Isgotkowitz: @matthew7838 please follow up with Kristan and see if this is the case

tests/animal_module_tests/digestive_system/test_digestive_system.py
```diff
@@ -0,0 +1,252 @@
...
+    mock_emission = mocker.patch.object(EntericMethaneCalculator, "calculate_calf_methane", return_value=15.3)
+    mock_manure = mocker.patch.object(
+        ManureExcretionCalculator, "calculate_calf_manure", return_value=(3, expected_excretions)
+    )
+
+    DigestiveSystem.METHANE_MODEL = "dummy model"
+    DigestiveSystem.METHANE_MITIGATION_METHOD = "dummy_method"
+    DigestiveSystem.METHANE_MITIGATION_ADDITIVE_AMOUNT = 16
+
+    observed_statistics, observed_excretions = DigestiveSystem.process_digestion(
```
Isgotkowitz: Instead of using the term "observed", it is conventional to use the term "actual"

tests/animal_module_tests/digestive_system/test_enteric_methane_calculator.py
```diff
@@ -0,0 +1,320 @@
...
+    NDF_concentration = nutrient_concentrations["NDF"]
+    ASH_concentration = nutrient_concentrations["ash"]
+
+    expected_gross_energy_concentration = (
+        0.263 * CP_concentration
+        + 0.522 * EE_concentration
+        + 0.198 * NDF_concentration
+        + 0.160 * ((100 - ASH_concentration) - NDF_concentration - CP_concentration - EE_concentration)
+    )
+    expected = (0.065 * expected_gross_energy_concentration * dry_matter_intake) / 0.05565
```
Isgotkowitz: Duplicating logic being tested inside the body of the test is not a good practice. It is better to have the expected result be a parameter of the test case

RUFAS/data_structures/animal_manure_excretions.py
```diff
@@ -60,4 +58,3 @@ class AnimalManureExcretions(TypedDict):
     phosphorus: float
     phosphorus_fraction: float
     potassium: float
-    enteric_methane_g: float
```
Isgotkowitz: Instead of changing this Animal Manure excretions, I suggest making a duplicate of this module, using the duplicate in `biophysical/` and modifying that duplicate. This is because this data type is shared between the Animal and Manure modules, and if it is changed here that change should be propagated to all other parts of the codebase that use it, which I don't think we are ready for.

Isgotkowitz (commented):
I think this is looking really good and getting much closer to being finished Matthew! A few things I noticed:
- There is still an unresolved comment from my first review about removing some methane mitigation methods because there is no support for them in RuFaS. I tagged Kristan in the comment, but I think you will have to follow up with her to get an answer
- `AnimalManureExcretions` is modified, which is good, but not all uses of it have been modified which is problematic. There is a more detailed comment about this below
- In a lot of the tests, the logic being tested is duplicated. This is not a good testing practice

RUFAS/biophysical/animal/animal_properties/general_properties.py
matthew7838: Should properties be listed with attributes in docstrings? My IDE is giving me warnings.

RUFAS/biophysical/animal/digestive_system/digestive_system.py
matthew7838: I agree, I will add the docstrings for the sake of easy transformation into enums in the future.

RUFAS/biophysical/animal/digestive_system/methane_mitigation_calculator.py
KFosterReed: Hi Ed and Matthew! Sorry I missed this earlier and thanks for pinging me again! Yes, you're correct that we are going to wait to add essential oils and seaweed as actual options in our methane mitigation set. I think there is a possibility that we will get essential oils figured out in the next ~ 6 months but am I correct that it is 'easy enough' to add back when the time comes?

RUFAS/biophysical/animal/animal_properties/general_properties.py
Isgotkowitz: Yes, they should, but they should not have any description in the class docstring: https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring

RUFAS/biophysical/animal/digestive_system/methane_mitigation_calculator.py
Isgotkowitz: > but am I correct that it is 'easy enough' to add back when the time comes?

Yes, I think so. If essential oils are that close, it may be better to leave that option in and just take out seaweed

RUFAS/biophysical/animal/digestive_system/manure_excretion_calculator.py
matthew7838: The total urine is actually different for each animal. I don't think it makes sense to declare it as module constants here.

RUFAS/biophysical/animal/digestive_system/digestive_system.py
```diff
@@ -1,2 +1,129 @@
...
+                animal_nutrient_property.fecal_phosphorus,
+                animal_nutrient_property.urine_phosphorus_required,
+                general_properties.nutrients,
+                general_properties.nutrient_concentrations,
+            )
+            statistics["methane_emission"] = methane_emission
+            statistics["phosphorus_excreted"] = phosphorus
+            return statistics, excretion
+
+        else:
```
JoeWaddell: Minor point, but should we explicitly check that the animal is a Cow, in cases that additional/custom animal stages are added in the future and not caught here? E.g. explicitly check cow, then have an else that logs and error/warning if it doesn't fall under any of the explicit cases?

matthew7838: Good point. I will address this change.

RUFAS/biophysical/animal/data_types/animal_manure_excretions.py
```diff
@@ -0,0 +1,60 @@
+from typing import TypedDict
+
+
+class AnimalManureExcretions(TypedDict):
```
Isgotkowitz, line 4: Can you please write up an issue for syncing up this `AnimalManureExcretions` with the one in `data_structures/`, and add a TODO to this file noting that eventually the `AnimalManureExcretions` will have to be replaced with this one?

RUFAS/biophysical/animal/digestive_system/digestive_system.py
```diff
@@ -1,2 +1,141 @@
...
+        """
+        Handles an animal's daily digest updates.
+
+        Parameters
+        ----------
+        general_properties: GeneralProperties
+            Animal properties that are general or are used to determine many animal outcomes.
+        animal_nutrient_property: AnimalGrowthProperties
+            Animal properties that are related to animal nutrients.
+        milk_production_properties: ReproductionProperties
```
Isgotkowitz:
```suggestion
        milk_production_properties: MilkProductionProperties
```

RUFAS/biophysical/animal/digestive_system/enteric_methane_calculator.py
```diff
@@ -0,0 +1,310 @@
...
+            starch_to_acid_detergent_fiber_concentration_ratio = (
+                -0.0011 * starch_concentration / acid_detergent_fiber_concentrations
+            )
+            temp = -(starch_to_acid_detergent_fiber_concentration_ratio + 0.0045) * metabolizable_energy_intake * 4.184
+            methane_emission = 45.98 * (1 - exp(temp)) / 0.05565
+
+        elif methane_model == "IPCC":  # IPCC
+            # Calculating gross energy concentration (Moraes et al. 2014)
+            soluble_residue = (
+                GeneralConstants.FRACTION_TO_PERCENTAGE
```
Isgotkowitz, line 238: I like the use of a constant here instead of just `100`. But in this case I think it would be better to define a new constant, `ONE_HUNDRED_PERCENT`, instead of using a constant that has the same value but is supposed to be used in a different way. This is because it is very confusing to have a conversion factor being subtracted from. Also, if you define a new constant for 100 percent, could you please update `_calculate_dry_cow_manure` and `calculate_heifer_methane` to use the same constant?

RUFAS/biophysical/animal/digestive_system/enteric_methane_calculator.py
Isgotkowitz, line 1: What do you think about renaming `_calculate_lactating_cow_manure` and `_calculate_dry_cow_manure` to `_calculate_lactating_cow_methane` and `_calculate_dry_cow_methane`, respectively?

RUFAS/biophysical/animal/digestive_system/manure_excretion_calculator.py
```diff
@@ -0,0 +1,727 @@
...
+        # Degradable volatile solids, kg/day [A.3A.A.5]
+        degradable_volatile_solids = 0.9 * total_volatile_solids
+
+        # Non-degradable volatile solids, kg/day [A.3A.A.6]
+        non_degradable_volatile_solids = total_volatile_solids - degradable_volatile_solids
+
+        # Nitrogen excretion
+        # Amount of nitrogen excreted by the calf, kg [A.3A.B.1]
+        manure_nitrogen = (
+            112.55 * dry_matter_intake * (crude_protein_concentration / 100)
```
Isgotkowitz:
```suggestion
            112.55 * dry_matter_intake * (crude_protein_concentration * GeneralConstants.PERCENTAGE_TO_FRACTION)
```

RUFAS/biophysical/animal/digestive_system/manure_excretion_calculator.py
```diff
@@ -0,0 +1,727 @@
...
+            Amount of phosphorus required for urine production (g).
+        nutrient_amounts: Dict[str, float]
+            Amounts of nutrients in pen ration, calculated per animal, see Notes section for units.
+        nutrient_concentrations: Dict[str, float]
+            Concentrations of nutrients in pen ration, calculated per animal, percentages.
+
+        Returns
+        -------
+        float
+        Total amount of phosphorus excreted by the given animal (g).
```
Isgotkowitz:
```suggestion
            Total amount of phosphorus excreted by the given animal (g).
```

RUFAS/biophysical/animal/digestive_system/manure_excretion_calculator.py
```diff
@@ -0,0 +1,727 @@
...
+        # Non-degradable volatile solids, kg [A.3A.A.6]
+        non_degradable_volatile_solids = total_volatile_solids - degradable_volatile_solids
+
+        # Nitrogen in liquid and solid manure, kg [A.3B.B.1]
+        manure_nitrogen = (
+            15.1
+            + 0.83
+            * (dry_matter_intake * GeneralConstants.KG_TO_GRAMS)
+            * (crude_protein_concentration * GeneralConstants.PROTEIN_TO_NITROGEN)
+            / GeneralConstants.FRACTION_TO_PERCENTAGE
```
Isgotkowitz:
```suggestion
            * GeneralConstants.PERCENTAGE_TO_FRACTION
```

RUFAS/biophysical/animal/digestive_system/manure_excretion_calculator.py
```diff
@@ -0,0 +1,727 @@
...
+        ) * GeneralConstants.GRAMS_TO_KG
+
+        # Nitrogen excretion in feces, kg [A.3B.B.2]
+        fecal_nitrogen = (
+            0.345
+            + 0.317
+            * (dry_matter_intake * GeneralConstants.KG_TO_GRAMS)
+            * (crude_protein_concentration * GeneralConstants.PROTEIN_TO_NITROGEN)
+            / GeneralConstants.FRACTION_TO_PERCENTAGE
```
Isgotkowitz:
```suggestion
            * GeneralConstants.PERCENTAGE_TO_FRACTION
```

RUFAS/biophysical/animal/digestive_system/manure_excretion_calculator.py
```diff
@@ -0,0 +1,727 @@
...
+        # Nitrogen concentration in urinary urea, g urea-N/L [A.3G.B.2]
+        urine_urea_nitrogen_concentration = -1.16 + 0.86 * urinary_nitrogen_concentration
+
+        # Total ammoniacal nitrogen in the manure slurry, kg
+        manure_total_ammoniacal_nitrogen = urine_nitrogen
+
+        # Amount of potassium excreted, g [A.3B.B.4]
+        potassium = (
+            dry_matter_intake
+            * (potassium_concentration / GeneralConstants.FRACTION_TO_PERCENTAGE)
```
Isgotkowitz:
```suggestion
            * (potassium_concentration * GeneralConstants.PERCENTAGE_TO_FRACTION)
```

RUFAS/biophysical/animal/digestive_system/manure_excretion_calculator.py
```diff
@@ -0,0 +1,727 @@
...
+        # Amount of feces and urine excreted daily by the growing heifer, kg [A.3E.A.4]
+        total_manure_excreted = fecal_water + fecal_solids + urine
+
+        # Total manure nitrogen, kg [A.3E.B.1]
+        manure_nitrogen = (
+            20.3
+            + 0.654
+            * (dry_matter_intake * GeneralConstants.KG_TO_GRAMS)
+            * (crude_protein_concentration * GeneralConstants.PROTEIN_TO_NITROGEN)
+            / GeneralConstants.FRACTION_TO_PERCENTAGE
```
Isgotkowitz:
```suggestion
            * GeneralConstants.PERCENTAGE_TO_FRACTION
```

RUFAS/biophysical/animal/digestive_system/manure_excretion_calculator.py
```diff
@@ -0,0 +1,727 @@
...
+        urinary_nitrogen_concentration = (urine_nitrogen * GeneralConstants.KG_TO_GRAMS) / urine
+        # Nitrogen concentration in urinary urea, g urea-N/L [A.3G.B.2]
+        urine_urea_nitrogen_concentration = -1.16 + 0.86 * urinary_nitrogen_concentration
+
+        # Total ammoniacal nitrogen in the manure slurry, kg
+        manure_total_ammoniacal_nitrogen = urine_nitrogen
+
+        # Amount of potassium excreted, g [A.3E.B.3]
+        potassium = (
+            7.21 * dry_matter_intake + 15944 * potassium_concentration / GeneralConstants.FRACTION_TO_PERCENTAGE - 164.5
```
Isgotkowitz:
```suggestion
            7.21 * dry_matter_intake + 15944 * potassium_concentration * GeneralConstants.PERCENTAGE_TO_FRACTION - 164.5
```

RUFAS/biophysical/animal/digestive_system/manure_excretion_calculator.py
```diff
@@ -0,0 +1,727 @@
...
+        # Total solids excretion
+        # Amount of dry material excreted by the dry cow, kg [A.3F.A.3]
+        total_solids = 0.178 * dry_matter_intake + 2.733
+
+        # Organic matter intake, kg [A.2.A.3]
+        dry_matter_intake = max(dry_matter_intake, AnimalModuleConstants.MINIMUM_DMI_DRY)
+        organic_matter_intake = (
+            dry_matter_intake
+            * (GeneralConstants.FRACTION_TO_PERCENTAGE - ash_concentration)
+            / GeneralConstants.FRACTION_TO_PERCENTAGE
```
Isgotkowitz:
```suggestion
            * GeneralConstants.PERCENTAGE_TO_FRACTION
```

RUFAS/biophysical/animal/digestive_system/manure_excretion_calculator.py
```diff
@@ -0,0 +1,727 @@
...
+
+        # Total solids excretion
+        # Amount of dry material excreted by the dry cow, kg [A.3F.A.3]
+        total_solids = 0.178 * dry_matter_intake + 2.733
+
+        # Organic matter intake, kg [A.2.A.3]
+        dry_matter_intake = max(dry_matter_intake, AnimalModuleConstants.MINIMUM_DMI_DRY)
+        organic_matter_intake = (
+            dry_matter_intake
+            * (GeneralConstants.FRACTION_TO_PERCENTAGE - ash_concentration)
```
Isgotkowitz, line 569:
```suggestion
            * (GeneralConstants.FRACTION_TO_PERCENTAGE - ash_concentration)
```
Same comment on subtracting from `FRACTION_TO_PERCENTAGE` as in `enteric_methane_calculator` module.

RUFAS/biophysical/animal/digestive_system/manure_excretion_calculator.py
```diff
@@ -0,0 +1,727 @@
...
+        # Non-degradable volatile solids, kg [A.3A.A.6]
+        non_degradable_volatile_solids = total_volatile_solids - degradable_volatile_solids
+
+        # Nitrogen in liquid and solid manure, kg [A.3B.B.1]
+        manure_nitrogen = (
+            15.1
+            + 0.83
+            * (dry_matter_intake * GeneralConstants.KG_TO_GRAMS)
+            * (crude_protein_concentration * GeneralConstants.PROTEIN_TO_NITROGEN)
+            / GeneralConstants.FRACTION_TO_PERCENTAGE
```
Isgotkowitz, line 594:
```suggestion
            * GeneralConstants.PERCENTAGE_TO_FRACTION
```

RUFAS/biophysical/animal/digestive_system/manure_excretion_calculator.py
```diff
@@ -0,0 +1,727 @@
...
+        ) * GeneralConstants.GRAMS_TO_KG
+
+        # Nitrogen excretion in feces, kg [A.3B.B.2]
+        fecal_nitrogen = (
+            0.345
+            + 0.317
+            * (dry_matter_intake * GeneralConstants.KG_TO_GRAMS)
+            * (crude_protein_concentration * GeneralConstants.PROTEIN_TO_NITROGEN)
+            / GeneralConstants.FRACTION_TO_PERCENTAGE
```
Isgotkowitz:
```suggestion
            * GeneralConstants.PERCENTAGE_TO_FRACTION
```

RUFAS/biophysical/animal/digestive_system/manure_excretion_calculator.py
```diff
@@ -0,0 +1,727 @@
...
+        # Nitrogen concentration in urinary urea, g urea-N/L [A.3G.B.2]
+        urine_urea_nitrogen_concentration = -1.16 + 0.86 * urinary_nitrogen_concentration
+
+        # Total ammoniacal nitrogen in the manure slurry, kg
+        manure_total_ammoniacal_nitrogen = urine_nitrogen
+
+        # Amount of potassium excreted, g [A.3B.B.4]
+        potassium = (
+            dry_matter_intake
+            * (potassium_concentration / GeneralConstants.FRACTION_TO_PERCENTAGE)
```
Isgotkowitz:
```suggestion
            * (potassium_concentration * GeneralConstants.PERCENTAGE_TO_FRACTION)
```

Isgotkowitz (approved):
Left a bunch of minor comments, mostly related to the use of constants. But overall LGTM! Nice job working through all these equations Matthew

RUFAS/biophysical/animal/digestive_system/enteric_methane_calculator.py
JoeWaddell: Agreed: especially when the documentation winds up on a website accessible to a wider userbase (e.g. non programmers), it'll help folks know where we're sourcing our logic for the equations in each method.

It is hard to keep track of when/where the references are used, so some simple description of where they currently are could be included, e.g. `Mutian calculation: [A.3E.C.1]`

RUFAS/biophysical/animal/digestive_system/digestive_system.py
```diff
@@ -1,2 +1,141 @@
...
+                general_properties.daily_milk_produced,
+                animal_nutrient_property.fecal_phosphorus,
+                animal_nutrient_property.urine_phosphorus_required,
+                general_properties.nutrients,
+                general_properties.nutrient_concentrations,
+            )
+            statistics["methane_emission"] = methane_emission
+            statistics["phosphorus_excreted"] = phosphorus
+            return statistics, excretion
+        else:
```
JoeWaddell, line 132: This statement probably needs a return or a raised error at this stage, to tamp down the mypy error

RUFAS/biophysical/animal/digestive_system/digestive_system.py
```diff
@@ -1,2 +1,141 @@
...
+            info_map = {
+                "class": DigestiveSystem.__name__,
+                "function": DigestiveSystem.process_digestion.__name__,
+            }
+            om.add_error(
+                "Unsupported animal type",
+                f"Supported animal types are cow, heifer and calf. Got {general_properties.animal_type}",
```
JoeWaddell: This might be overengineering, but do we have a dynamic list you could use here to show what classes are supported? e.g. maye a set of the `AnimalType`s found inside the `AnimalGroupingScenario`?

On second thought: the more explicit thing might be to just add a list of the `general_properties.animal_type`s hardcoded above, since those would necessarily need to be updated in cases that would cause this error

JoeWaddell (approved):
Nice job!

Issue #1924, opened 2024-08-15
[Animal] Improve the `LactationCurve` module 


Isgotkowitz:
With the introduction of the Lactation Curve module introduced in #1461, RuFaS got a powerful new tool for predicting milk yields.

## Status

When #1461 is merged in, RuFaS will be able to set lactation curve parameters specifically to a farm based on its location, management practices, observed milk yield, and some other factors. However the code for this was written quickly, and has some areas which could be improved.

## Intention

Lower the runtime and improve the code quality of the `LactationCurve` module.

## Request

There are several areas of `LactationCurve` that could be improved.
1. Better logging. After the final base lactation curve parameters have been calculated, they should be reported to the OutputManager as variables. When lactation curve parameters are adjusted based on the annual milk yield of a farm, this is logged. We should log the alternative too - when lactation curve parameters are not adjusted because the is no annual milk yield input, this should be logged as well.
2. Cleaner code. Currently the `LactationCurve` module has individual attributes that references lactation curve parameters for parities 1, 2 and 3 directly, and a dictionary that maps parities to those attributes. This is roundabout and should be simplified, so that there is only the `parity_to_parameter_mapping` and that holds the lactation curve parameters directly.
3. Replace the use of Scipy's `quad` in `calc_305_day_milk_yield` with a vectorized integration utilizing Numpy. `quad` is noticeably slow, and vectorized integration would significantly speed up this method. See additional information section for more details.
4. In `_fit_wood_l_param_to_milk_yield`, an optimal value for Wood's l parameter is found by brute-forcing the search space. A more straight forward way to find the best l value would be to use a method like Scipy's `optimize.root_scalar`.

Note for # 3 and 4: these changes should only be made if they actually speed up the code _and_ SMEs approve any tradeoffs in precision. Verifying these things will take some experimentation.

### Additional Information

This is the code that ChatGPT gave me for vectorized integration with Numpy:
```python
    def vectorized_integration(self, f, a, b, args=(), num_points=10000):
        """
        Vectorized integration using NumPy.

        Parameters
        ----------
        f : callable
            The function to integrate.
        a : float
            The start of the integration interval.
        b : float
            The end of the integration interval.
        args : tuple
            Additional arguments to pass to the function.
        num_points : int
            Number of points to use for the integration.

        Returns
        -------
        float
            The result of the integration.
        """
        # Generate an array of points between a and b
        x = np.linspace(a, b, num_points)
        
        # Evaluate the function at these points
[...]

PR #2005, opened 2024-09-17
[Animal] Refactor milk production logic

### LINKED ISSUE #1925: [Animal] Clean up `Cow` code that is related to lactation
With the introduction of PR #1461, a lot of lactation related code has moved from the `Cow` class to the `LactationCurve` class. As a results, some things need to be updated in the `Cow` class.

## Status

With the introduction of the `LactationCurve` class, a lot of code in the `Cow` class has become duplicative or unnecessary. It should be removed from the class.

## Intention

Clean up and simplify the code in `Cow`. This will hopefully make it easier to comprehend and refactor.

## Request

Please address the following areas of `Cow` in `cow.py`:
1. Remove the `breed_index` and `breed_parity` attributes. These are no longer used.
2. Change `calculate_daily_milk_produced` to use the implementation of Wood's equation implemented in `get_milk_yield_values_wood_curve` in `LactationCurve`.
3. Remove `determine_param_value` and use `Utility.generate_random_number` instead.

### Additional Information

It may be better to wait on this issue until after the Animal module refactor is completed. My two cents is that it may help the refactor go more smoothly if `Cow` is in a better state before the refactor starts.

### LINKED ISSUE #1930: [Animal][MilkProduction] MilkProduction Refactor Implementation [3/3]
Follow the Design in this Google Doc: [Milk Production](https://docs.google.com/document/d/1YYP-ynNp16SF2fRSOXZhG4BOd7UuDhpx4WQN2b_qQV0/edit) to implement the refactor of the MilkProduction submodule.

Isgotkowitz:
Refactors milk production logic for the Animal module and places it in the `biophysical` module.

### Context

Issue(s) closed by this pull request: closes #1930

Also closes #1925 - it doesn't actually remove unnecessary code from `Cow`, but that unnecessary code is being left behind.

### What

This PR starts the refactor of milking-related code in the animal module, re-implementing the original logic of the Animal module while improving code cleanliness and adhering to the new design pattern for modules. Also adds `numba` as a dependency to optimize the code.

### Why

The various subroutines within the Animal module are very entangled, by separating them into more dedicated routines we will be able to significantly improve the code - in terms of both clarity and performance.

### How

`MilkProduction` has been added in the new `milk` directory, with logic from the old `Cow` module and a little from `LactationCurve` added. The animal properties `MilkProductionProperties` and `GeneralProperties` have been created and modified, respectively, to support the operations of `MilkProduction`. The `LactationCurve` module has been copied into `biophysical` and one method from it has been moved into `MilkProduction` (more of this will happen later as part of #1924.

In order to accelerate the code, Numba is added to the dependencies and the `njit` decorator is used. `njit` tells Python to compile the code when it runs the method. This works great for staticmethods that contain only math operations (including builtins, Numpy, and Scipy) on basic types (i.e. numbers). I didn't do any precise benchmarking, but after adding `njit` to the method that calculates daily milk yield the time necessary to run a handful of integration tests dependent on that method decreased from ~9 second to ~3 seconds. To see this yourself, compare running these two commands:
```
pytest tests/animal_module_tests/life_cycle/test_lactation_curve.py -k test_fit_wood_l_param_to_milk_yield
```
```
pytest tests/animal_module_tests/milk_tests/test_lactation_curve.py -k test_fit_wood_l_param_to_milk_yield
```
### Test plan

A test suite has been written to verify the behavior of the new code. I tried to strike the right balance between thoroughness and not being overly picky, I'm sure our testing will evolve as we get deeper into the animal refactor. Note that code decorated with `@njit` is tested, but the `coverage` tool misses this.

### Update
I have made the following changes since the first review of this PR:
- Changed names of some milk-related variables in `GeneralProperties` and updated their usage in `AnimalPhosphorus`. I commented the old names but didn't give an explanation, if you would like one please let me know.
- Changed the name of the last 305 day milk production variable to the name suggested by Kaitlyn.
- Added lactose and crude protein content attributes to `MilkProductionProperties`. I realized it would just be good to do this so that there would be a `_percent` and `_content` variable for all the nutritive properties of milk.
  - Sidenote, it turns out that lactose was being reported incorrectly from `Cow` in the `AnimalModuleReporter`. The variables was the percentage of lactose in milk, but the variable was reported to OM as kilograms.
- Added subroutine to calculate the nutrient content of milk and added the njit decorator to it. 

Another Update: while checking the test suite, I found that two of the methods decorated with `njit` actually ran slower! So I removed the decorator from those methods, made small changes to how they behave, and updated the test suite. My apologies for the multiple updates y'all!


RUFAS/biophysical/animal/data_types/milk_production_record.py
```diff
@@ -0,0 +1,24 @@
+from typing import TypedDict
+
+
+class MilkProductionRecord(TypedDict):
```
Isgotkowitz, line 4: This reimplements `MilkProductionHistory` from `RUFAS/routines/animal/life_cycle/cow.py`.

RUFAS/biophysical/animal/milk/lactation_curve.py
Isgotkowitz, line 1: Only thing changed in this file is that I moved `get_milk_yield_values_wood_curve` into `MilkProduction` (renamed it to `calculate_daily_milk_production`), and import `MilkProduction` so that method can be used.

RUFAS/biophysical/animal/milk/milk_production.py
```diff
@@ -0,0 +1,166 @@
...
+        milking_properties = MilkProduction._update_milking_history(milking_properties, general_properties, time)
+
+        if general_properties.days_in_milk == 305:
+            milk_history = [record["milk_production"] for record in milking_properties.milk_production_history[-305:]]
+            milking_properties.latest_305_day_milk_production = np.sum(milk_history)
+
+        return milking_properties, general_properties
+
+    @staticmethod
+    @njit
```
Isgotkowitz, line 116: The `njit` decorator throws off mypy and coverage, but I think the tradeoff in performance will be well worth it.

tests/animal_module_tests/general_property_values.py
Isgotkowitz, line 1: Instead of placing this data directly in `milk_tests/test_milk_production.py` I put it here, with the intention that it will be imported by other testing modules and used there as well.

tests/animal_module_tests/milk_tests/test_lactation_curve.py
Isgotkowitz, line 1: File is the same as it is in routines, save that it doesn't have the test for method that was moved to `MilkProduction`.

tests/animal_module_tests/milk_tests/test_milk_production.py
```diff
@@ -0,0 +1,203 @@
...
+        wood_l=29.0,
+        wood_m=0.25,
+        wood_n=0.003,
+        milk_production_history=[],
+    )
+
+
+@pytest.fixture
+def general_properties() -> GeneralProperties:
+    return GeneralProperties(**LAC_COW)
```
Isgotkowitz: Dict unpacking adds a ton of Mypy errors. It is super annoying and in this case I am not sure it is worth bending over backwards to make mypy happy.

tests/animal_module_tests/milk_tests/test_milk_production.py
```diff
@@ -0,0 +1,203 @@
...
+
+    assert pytest.approx(actual) == expected
+
+
+@pytest.mark.parametrize("milk,reduction,variance,expected", [(30.0, -4.0, 2.0, 28.0), (28.0, 0.0, -1.0, 27.0)])
+def test_adjust_milk_production(milk: float, reduction: float, variance: float, expected: float) -> None:
+    """Test that milk production is varied correctly."""
+    actual = MilkProduction._adjust_milk_production(milk, variance, reduction)
+
+    assert actual == expected
```
Isgotkowitz: The lines covered by these tests appear uncovered in the report by coverage. This is almost certainly because these mehtod are decorated with njit.

RUFAS/biophysical/animal/animal_properties/general_properties.py
```diff
@@ -17,18 +32,22 @@ class GeneralProperties:
     days_born: int
     days_in_preg: int
     events: AnimalEvents
+    days_in_milk: int
+    dry_off_day_of_pregnancy: int  # Old name: days_in_preg_when_dry, used to be in AnimalBase.config
```
krb78: the name change is great!

RUFAS/biophysical/animal/animal_properties/milk_production_properties.py
```diff
@@ -1,6 +1,50 @@
...
+
+    Attributes
+    ----------
+    true_protein_content : float
+        Amount of true protein in the milk produced (kg).
+    fat_content : float
+        Amount of fat in the milk produced (kg).
+    milk_production_reduction : float
+        Amount of milk daily production is reduced by (kg).
+    latest_305_day_milk_production : float
```
krb78: oo not to be picky here, but in the industry when we are referring to M305 or ME305 it's a correction factor that tries to level the playing field for all cows no matter how long or short they are in lactation, the month of calving, their parity, etc. that basically says "if cow Y and cow X were both only milking for 305 days, how much milk would they make". GIven your definition, and the fact that I assume you truly want the most recent 305 days of milk production, can we change this name? Maybe something like "milk_production_last_305days" I know it doesn't seem that different, but there is a really common set of codes in dairy management lingo that are L305 or P305 standing for the last lactation's M305.

RUFAS/biophysical/animal/milk/milk_production.py
```diff
@@ -0,0 +1,166 @@
...
+        milk_production: float, milk_production_variance: float, milk_production_reduction: float
+    ) -> float:
+        """
+        Randomly adjusts the milk production on a specific day.
+
+        Parameters
+        ----------
+        milk_production : float
+            Unvaried milk production (kg).
+        variance : float
```
allisterakun:
```suggestion
        milk_production_variance : float
```

RUFAS/biophysical/animal/milk/milk_production.py
```diff
@@ -0,0 +1,166 @@
...
+
+    @staticmethod
+    @njit
+    def calculate_daily_milk_production(day_of_milk: int, l_param: float, m_param: float, n_param: float) -> float:
+        """
+        Calculates the milk yield on the given day using Wood's lactation curve.
+
+        Parameters
+        ----------
+        day_of_milk : int
```
allisterakun: If I understand correctly, this parameter is the same as the `days_in_milk` property. Why don't we rename it to `days_in_milk` to avoid confusion?

RUFAS/biophysical/animal/animal_properties/milk_production_properties.py
```diff
@@ -1,6 +1,50 @@
...
+        Sum of milk production over the last 305 days of the cow's lactation (kg).
+    crude_protein_percent : float
+        Percentage of cow's milk that is crude protein by weight. TODO: double check the "by weight" part.
+    true_protein_percent : float
+        Percentage of cow's milk that is true protein. TODO: how do true protein and crude protein need to be compared in order to make sure they're valid.
```
krb78: this is a really good point - I don't actually know which one we were referring to when we built RuFaS... because it is PMO specific whether cooperatives report true or total / crude protein. We typically say total protein is ~5% of the N in milk so total / crude protein = true protein * 1.05

tests/animal_module_tests/general_property_values.py
```diff
@@ -0,0 +1,29 @@
+from RUFAS.biophysical.animal.data_types.animal_types import AnimalType
+from RUFAS.biophysical.animal.data_types.animal_events import AnimalEvents
+
+
+LAC_COW = {
```
krb78: This is a silly question, but are these just testing parameters?

tests/animal_module_tests/milk_tests/test_milk_production.py
```diff
@@ -0,0 +1,200 @@
...
+
+@pytest.fixture
+def milking_properties() -> MilkProductionProperties:
+    return MilkProductionProperties(
+        true_protein_content=1.5,
+        fat_content=2.0,
+        milk_production_reduction=0.3,
+        latest_305_day_milk_production=0.0,
+        crude_protein_percent=3.2,
+        true_protein_percent=3.0,
```
krb78, line 24: Are we asking folks to update both of these or do they update in tandem? If the latter, did you find another equation that wasn't my simple one about crude protein = 1.05*true protein?

tests/animal_module_tests/milk_tests/test_milk_production.py
```diff
@@ -0,0 +1,200 @@
...
+    general_properties: GeneralProperties,
+    time: Time,
+    days_in_milk: int,
+    expected_days_in_milk: int,
+    expected_305_day_production: float,
+) -> None:
+    """Tests perform_daily_milking_update when a cow is milking."""
+    general_properties.days_in_milk = days_in_milk
+    milking_properties.milk_production_history = [
+        MilkProductionRecord(simulation_day=1, days_in_milk=100, milk_production=1.0, days_born=500) for _ in range(305)
```
krb78, line 127: is this saying that we are only capturing milk production for the last 305 days of lactation? given the earlier definition, I think perhaps, but just checking that it's not presuming every lactation will only last 305 days.

tests/animal_module_tests/milk_tests/test_milk_production.py
```diff
@@ -0,0 +1,200 @@
...
+    milking_properties, general_properties = MilkProduction.perform_daily_milking_update(
+        milking_properties, general_properties, time
+    )
+
+    assert general_properties.days_in_milk == expected_days_in_milk
+    assert general_properties.estimated_daily_milk_produced == milk_produced
+    assert milking_properties.true_protein_content == 1.2
+    assert pytest.approx(milking_properties.fat_content) == 1.4
+    assert milking_properties.milk_production_history[-1] == expected_record
+    assert milking_properties.latest_305_day_milk_production == expected_305_day_production
```
krb78: I should know this... but is the expected_305_day_production in RuFaS the: 1) expected milk production in the first 305 days of lactation, 2) the "apples to apples" comparison of every cow corrected to a 305 day lactation, or 3) the expected production of the next 305 days, regardless of if she will milk all 305 days or just 100 more days? (or perhaps its none of those haha)

RUFAS/biophysical/animal/animal_properties/milk_production_properties.py
```diff
@@ -1,6 +1,50 @@
...
+    fat_percent : float
+        Percentage of cow's milk that is fat by weight. TODO: double check the "by weight" part.
```
krb78: not sure if this is a question for me, but fat % in milk and protein % in milk is calculated by grams protein/grams of milk... so 3.5% fat, if a cow was making 100kg of milk (which would be very impressive and on the far far extreme of realistic haha) would be 3.5kg of fat.

krb78 (commented):
Hi Ed! I left a bunch of comments. I'm not 100% sure if the TO DO's were things you wanted me to comment on or what. I think my comments are mostly just in our definitions and I apologize you are getting bombarded with them, because likely you are just the "messenger" and I should look for them elsewhere.

tests/animal_module_tests/general_property_values.py
Isgotkowitz: Not silly at all. Yes, these are just dummy values that make testing more convenient

RUFAS/biophysical/animal/animal_properties/milk_production_properties.py
Isgotkowitz: > GIven your definition, and the fact that I assume you truly want the most recent 305 days of milk production,

I'm not 100% sure what we actually want here, I carried over the logic from the previous implementation but it isn't necessarily the best or most useful for RuFaS.

With the way that the code is written, "latest_305_day_milk_production" (and it's predecessor "latest_milk_production_305days") are measures of how much milk a cow _actually_ produced, not much she would produce. So for the first 304 days of a cow's lactation, this value is 0.0 because it can't be calculated. This is good because it precisely accounts for the milking variance and production reduction of a cow, but it also means that two cows cannot be compared unless they have both been milking for at least 305 days.

RUFAS/biophysical/animal/animal_properties/milk_production_properties.py
Isgotkowitz: This could definitely be problematic in RuFaS. The crude protein content of milk is defined by the constant `MILK_CRUDE_PROTEIN` and is set at 3.2%. But the true protein percent is set directly by the user ("milk_protein_percent" in the Animal input file). The value in the default Animal config is also 3.2%, and that is the default in the metadata as well.

Also, I'm gonna guess that "mPrt" is an abbreviation for "milk Protein true". This isn't a helpful observation, it just feels really good to probably know what mPrt means.

tests/animal_module_tests/milk_tests/test_milk_production.py
Isgotkowitz: I left more details in the above comment, but that is something we should figure out.

tests/animal_module_tests/milk_tests/test_milk_production.py
Isgotkowitz: It isn't presuming that every lactation will last for 305 days. This is setting up a dummy milk production history to make sure that when the 305 day milk production is calculated, it only ever calculates and sets it on the 305th day of a cow's lactation.

tests/animal_module_tests/milk_tests/test_milk_production.py
Isgotkowitz: The variable (`milking_properties.latest_305_day_milk_production`) is the amount of milk that a cow produced in her first 305 days of her current lactation, so most similar to 2 but a little different. `expected_305_day_production` is named that way for testing purposes; it is the value we, the software testers, expect to `milking_properties.latest_305_day_milk_production` to have.

RUFAS/biophysical/animal/animal_properties/milk_production_properties.py
Isgotkowitz: Yes, this is a question intended for you as well as any other SMEs that see it. I wanted to confirm that it is by weight/mass, instead of something else (like volume, yuck 🥴)

RUFAS/biophysical/animal/milk/milk_production.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+
+    FAT_PERCENT = None
+    TRUE_PROTEIN_PERCENT = None
+
+    @classmethod
+    def set_milk_quality(cls, fat_percent: float, true_protein_percent: float) -> None:
+        """Sets user-defined milk qualities."""
+        cls.FAT_PERCENT = fat_percent,
+        cls.TRUE_PROTEIN_PERCENT = true_protein_percent
```
Isgotkowitz: @allisterakun I added these values as class "constants", which are intended to be set once and because of this I didn't include getters but tried to document them appropriately. Python is very encouraging of direct access so I don't think they are necessary at this point.

Isgotkowitz:
To summarize the discussion I had with @krb78 yesterday regarding a few key points of this PR:
- The variable `latest_305_day_milk_production`/`latest_milk_production_305days` is problematic. In the code, this value is calculated as the amount of milk that a cow produced in the first 305 days of her current lactation. This value _is only_ calculated after she has been lactating for 305 days, and as soon as she dries off it is reset to 0. This means that the value is not particularly helpful in comparing the productivity of two cows, which is what the real-world equivalent of this value would probably be used for.
- Crude protein content and true protein content of milk need to be checked against each other in some way. It is problematic to allow them to get outside of the ratio of 1.05 true protein = crude protein without alerting the user that this is happening at all. Currently the crude protein content is hardcoded in a constant with a value of 3.2, while the true protein content is a user input.

The above points are issues, however fixing them is outside the scope of this refactor. I will write up issues for addressing both of these once this refactor/the animal refactor is finished.

Edit: issues #2012 and #2013 have been written up for these two points

RUFAS/biophysical/animal/milk/milk_production.py
```diff
@@ -0,0 +1,213 @@
...
+        nutrient_percentage : float
+            Percentage of nutrient in the milk.
+
+        Returns
+        -------
+        float
+            Amount of nutrient contained in the milk (kg).
+
+        """
+        return milk * nutrient_percentage * 0.01
```
Isgotkowitz: Using 0.01 here instead of the constant defined in `GeneralConstants` is bad, however if `GeneralConstants` is referenced in this equation then Numba won't be able to understand it.

Isgotkowitz: Upon some benchmarking of this code I have discovered that adding `njit` actually makes equations this simple go slower

RUFAS/biophysical/animal/animal_properties/milk_production_properties.py
krb78: Ok, Ed, I thought about this a lot. I think we should just keep the current calculation and we should edit the definition and name of the parameter accordingly. I suggest: "current_lact_305_day_milk_produced". I think this will distinguish it enough from the industry M305 or ME305.

RUFAS/biophysical/animal/animal_properties/milk_production_properties.py
krb78: I think we discussed on Thursday that perhaps having it such that a user could enter either would be a good "down the road" option and have them linked such that only one needs to be changed.

krb78 (approved):

allisterakun (approved):
This PR looks good. Great work, Ed!

Before merging the PR, please:
1. delete all the temporary inline comments
2. write a test for the new cls method in the MilkProduction class

PR #2006, opened 2024-09-17
Adjust Storage Temp To Calculate Manure Storage Emissions

### LINKED ISSUE #1906: [Manure] Implement basic method to determine stored liquid manure temperature
The temperature of manure in storage is a key factor in determining the rate of biological and physicochemical activity that leads to manure composition changes and GHG emissions. Currently, the temperature of manure in storage is assumed to be equal to the ambient air temperature, which is a deviation from reality. 

For example, this [publication ](https://www.sciencedirect.com/science/article/pii/S0168169921002519)reports stored manure temperatures in clay pit vs. concrete tank storages located in Virginia across ~2 years. Compared to ambient air, manure temp was less variable (lower SD), reached lower maximums and higher minimums, and was greater (warmer) on average. This makes logical sense - a large, semi-solid mass will not change in temperature as quickly as ambient air, will generate some heat through biological activity, and is unlikely to reach extreme high or low temperatures because physics. 

image

A simple step toward more realistic estimation of stored manure temperature could be to implement upper and lower bounds, e.g., > 0 C, < 35 C (ideally these values should be sourced from published data or guidelines). Additionally, we could adopt the IPCC's methodology for estimating manure temperature of using the monthly average temperature. Another option is using a running previous-10-day average value as used in IFSM. 

In the future, as models for manure temperature improve, we should continually revisit opportunities for improvement of this estimation in RuFaS. While the suggestions here are an advancement over the current behavior, they do not take into account important factors like manure storage type/location, weather events, temperature gradients from surface to bottom, and other factors that influence temperature.

ew3361zh:
Adds functionality to convert ambient air temperature from current day conditions into a more accurate representation of manure storage temperature to more accurately model conditions of how manure storages operate.

### Context

Issue(s) closed by this pull request: closes #1906

### What

Adds a function to the base manure storage class to take in the ambient air temperature and adjust it to be within the approximate range temperatures will be within the storage. This is for outdoor storages.

For SlurryStorageUnderfloor, the functionality added in #2002 to adjust ambient air temperature to the approximate barn temperature is used.

### Why

These temperature adjustments are made prior to calculating the emissions from those storages. The adjusted temperatures more accurately reflect conditions within the respective storages which will lead to closer approximations of actual emissions generated.

### How

The function to convert air temperature to the corresponding estimated storage temperature was set up in the base manure storage class and implemented in SSO and AL. The previously set-up function to estimate barn temperature is used for SSU.

### Test plan

Unit testing is updated. Will need outputs confirmation from SME to ensure they now reflect closer to expected results.

### Note
When this PR is merged in, I will create a follow up issue to revisit this temperature estimation for continual improvement opportunities and better-documented guidelines as noted by @elle-andreen in issue #1906.


RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
```diff
@@ -273,6 +275,22 @@ def _daily_update_helper(self) -> ManureTreatmentDailyOutput:
...
         pass
 
+    @staticmethod
+    def _determine_outdoor_storage_temperature(air_temperature: float) -> float:
+        """Determines the temperature of the manure storage for outdoor storages.
+
+        Parameters
+        ----------
+        air_temperature : float
+            The current day's ambient air temperature.
```
```suggestion
            The current day's ambient air temperature (degrees C).
```

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
```diff
@@ -273,6 +275,22 @@ def _daily_update_helper(self) -> ManureTreatmentDailyOutput:
         """
         pass
 
+    @staticmethod
+    def _determine_outdoor_storage_temperature(air_temperature: float) -> float:
+        """Determines the temperature of the manure storage for outdoor storages.
```
```suggestion
        """Determines the temperature of the manure storage for outdoor liquid and slurry storages.
```
Would this be more accurate? It seems that this method is not being used for solid outdoor storages

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
```diff
@@ -273,6 +275,22 @@ def _daily_update_helper(self) -> ManureTreatmentDailyOutput:
...
+
+        Parameters
+        ----------
+        air_temperature : float
+            The current day's ambient air temperature.
+
+        Returns
+        -------
+        float
+            The estimated temperature of the manure storage.
```
```suggestion
            The estimated temperature of the manure storage (degrees C).
```

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
Isgotkowitz: Also, is this the temperature of the manure in the storage or is it the temperature of the air that is contained within the storage?

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
```diff
@@ -273,6 +275,22 @@ def _daily_update_helper(self) -> ManureTreatmentDailyOutput:
...
+        """
```
Isgotkowitz, line 279: I would also recommend leaving a note or reference in this docstring just to make sure that there is some information about where this method comes from.

Isgotkowitz (approved):
Looks great, another very clean manure improvement. I trust that @elle-andreen or another manure SME will review the outputs to make sure that the changes are having the intended effects on GHG emissions outputs. My comments are very similar to #2002 - just a few suggestions on how the docstring for the new temperature calculation method could be improved

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
elle-andreen: I agree with Ed's comment that this method applies presently to liquid manure, and I think it's unlikely to be also applied to solid manure at some point (i.e. I think solid manure may use a different method), so let's go ahead and specify.
This is the temperature of manure _in_ storage - we should specify that as well.

ew3361zh: Thanks, Elle, what do you think about:

`Determines the temperature of the manure in outdoor liquid and slurry storages.`

vs.

`Determines the temperature of the liquid manure in outdoor storages.`

vs.

`Determines the temperature of the liquid manure in outdoor liquid and slurry storages.`

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
ew3361zh: @elle-andreen should we cite [this reference](https://www.sciencedirect.com/science/article/abs/pii/S0168169921002519) here?

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -10,11 +10,11 @@
 
 class GasEmissionsCalculator:
     @classmethod
-    def methane_emission_from_slurry_storage(
+    def calculate_methane_emission_from_slurry_storage(
         cls,
         accumulated_liquid_manure_total_degradable_volatile_solids: float,
         accumulated_liquid_manure_total_non_degradable_volatile_solids: float,
-        temp: float = GasEmissionConstants.DEFAULT_SLURRY_STORAGE_TEMPERATURE,
+        storage_temperature: float = GasEmissionConstants.DEFAULT_SLURRY_STORAGE_TEMPERATURE,
```
elle-andreen: I know this default is overwritten by another method (i.e. this default is not actually used, the value from the function to determine stored manure temp is) but what is the purpose of this default value here functionally-speaking? To have a value for BaseManureTreatment to utilize? I ask because this default being here is potentially pretty confusing. Unless someone had other information, I think someone looking at this function would assume that there is NOT a function to determine temperature and the temperature value in this & other functions is just a constant value (I definitely did when I first saw it). Is there something else we can put here to help future code-readers understand that there's actually a different value being used here?

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -54,9 +54,9 @@ def methane_emission_from_slurry_storage(
             Total degradable volatile solids in manure (kg).
         accumulated_liquid_manure_total_non_degradable_volatile_solids: float,
             Total non-degradable volatile solids in manure (kg).
-        temp : float
-            Temperature in Celsius (:math:`^\\circ C`). Default is set to 20 degrees Celsius. This value is
-            listed as :attr:`DEFAULT_SLURRY_STORAGE_TEMPERATURE` in :class:`GasEmissionConstants`.
+        storage_temperature : float
+            Temperature of the storage in Celsius (:math:`^\\circ C`). Default is set to 20 degrees Celsius.
```
elle-andreen:
```suggestion
            Temperature of the manure in storage in Celsius (:math:`^\\circ C`). Default is set to 20 degrees Celsius.
```

RUFAS/routines/manure/gas_emissions/calculator.py
elle-andreen: If it's not too much trouble, I would modify this variable name from `storage_temperature` to `stored_manure_temperature` to be more precise.

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -10,11 +10,11 @@
 
 class GasEmissionsCalculator:
     @classmethod
-    def methane_emission_from_slurry_storage(
+    def calculate_methane_emission_from_slurry_storage(
```
elle-andreen: We can make this non-specific to slurry storage, as this method is currently called by the anaerobic lagoon storage files, and lagoons do not strictly hold slurry manure. I also modified the function name to match the format of the ammonia emission equation. I did not make this suggestion for all instances, I figure this change is better suited for 'replace all instances' in an IDE!

```suggestion
    def calculate_liquid_storage_methane(
```

RUFAS/routines/manure/gas_emissions/calculator.py
```diff
@@ -448,13 +448,13 @@ def calculate_housing_ammonia_emission(
         return max(0.0, total_ammonia_loss)
 
     @classmethod
-    def storage_ammonia_emission(
+    def calculate_storage_ammonia_emission(
```
elle-andreen:
```suggestion
    def calculate_liquid_storage_ammonia_emission(
```
If we're already changing the name, specifying the function is applicable to liquid storages is helpful because both liquid and solid manure storages calculate ammonia emissions, but call different methods in `calculator.py`. If you're just looking at calc.py, it can be confusing what is for what.

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
elle-andreen: I think option 1 is the best! This method applies to both liquid and slurry manure (although as you probably know, they're not really being formally differentiated in RuFaS right now. The distinction is still helpful though.)

RUFAS/routines/manure/gas_emissions/calculator.py
ew3361zh: Sounds good!

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
elle-andreen: We can use the below:

References

This function clamps stored manure temperature to between 0 and 35 °C. Between 0 and 35 °C, outdoor stored liquid manure temperature is assumed to be equal to ambient air temperature. These bounds were based on personal communication and recommendations from A. Leytem (april.leytem@usda.gov) and A. VanderZaag (andrew.vanderzaag@AGR.GC.CA). These bounds are also support by work from Genedy and Ogejo, 2021 (https://doi.org/10.1016/j.compag.2021.106234) who observed similar minimum and maximum liquid manure temperatures in outdoor clay pit and concrete tank manure storages. 
```

File: RUFAS/routines/manure/manure_treatments/slurry_storage_underfloor.py [hunk 9]
```diff
@@ -62,15 +62,16 @@ def calc_methane_emission(
             (kg :math:`CH_4`/day).
 
         """
-        temperature_celsius = self._get_current_day_average_temperature_celsius()
+        air_temperature = self._get_current_day_average_temperature_celsius()
+        barn_temperature = GasEmissionsCalculator.determine_barn_air_temperature(air_temperature)
```
elle-andreen, line 70: Could we note here or in the docstring above that barn temperature is being calculated and used here, because the slurry in indoors/under the barn? I'm not sure where the right place for that would be, or if that's appropriate to have right in the code.

RUFAS/routines/manure/gas_emissions/calculator.py
ew3361zh: My memory of this was that at the time we wrote this to accommodate not having the storage temperature available from whatever storage device was using this calculation (and wanting to use a standard default in those cases).

So the purpose here is if a temperature isn't supplied by the calling class, we have this default temperature which we'd use in its place to ensure the calculation could still occur.

Generally defaults are just there to provide a value when that value may not be available for some reason. The default will always get overridden by a value supplied to the function but it allows it to still operate if the calling class doesn't have that information available.

And I think it's generally understood in pythonic code that a default value supplied like this will be overridden if another value is supplied by the caller so I don't necessarily think we need to specify that here.

If it seems detrimental to have it here and you can't think of any cases where it might be useful, we can remove the default and then the function would just require the caller to enter a value there.

elle-andreen: That makes sense! But, I think what adds to my personal confusion is that it seems like only this method for methane is using this default in the way you described, unless I'm missing something? Should all the functions that are calling for a stored manure temperature be referencing this default?

RUFAS/routines/manure/manure_treatments/slurry_storage_underfloor.py
ew3361zh: Will add it to the docstrings as a note.

RUFAS/routines/manure/gas_emissions/calculator.py
ew3361zh: I think it would make sense to do an all or nothing approach here. So either they should all have a default available or none of them should. 
Just depends on which you think is better. 
If it seems like we're going to readily have storage temperatures available (or are comfortable that were it not available that people would consistently use the same default constant), then I think we should remove it from all of them. Even if it's just this one function.
I think this probably makes the most sense. It would force any user to specify the stored manure temp and if someone proposes using the default, it would be reviewed by folks before it got merged into dev/main.

elle-andreen (commented):
(Left this by mistake - see my approval comment!)

elle-andreen (approved):
Thanks for your work on this Niko! Updated manure storage emissions under a variety of scenarios are included in the attached file. In general, this PR increases increase methane emissions by 0 - 10%, depending on storage type and animal type, with the greatest change in slurry storage underfloor storages (as expected). Storage N2O emissions will not change. Ammonia emission increases can be larger (up to 50%) but are dependent on herd size and other factors. Even where percentage increases are large, because a relatively small amount of excreted manure N is lost as ammonia, the additional quantity of N loss to NH3-N is still relatively small. We did have April weigh in on these changes - she was fully on board in terms of this PR being a helpful step forward in terms of improving temperature and emissions estimates, and felt like our outcomes across scenarios were still within realistic ranges. 

[RuFaS storage temp comparison.xlsx](https://github.com/user-attachments/files/17625307/RuFaS.storage.temp.comparison.xlsx)

PR #2009, opened 2024-09-18
Time enhancement

### LINKED ISSUE #1922: [Feed Storage] [Time] Use alternative to `Time` in Feed Storage module when tracking dates in order to remove `deepcopy`
Instances of `Time` are used to track dates in the Feed Storage module, and these instances created using `deepcopy`. This slows down the code dramatically when processing degradations in the Feed Storage module.

## Status

`HarvestedCrop`s use instances of the `Time` module to track dates - when the crop was harvested, stored, and last degraded. The instances are created from other `Time` instances using `deepcopy`, so that the time is preserved and doesn't keep moving forward as the main `Time` instance is incremented.

## Intention

Make the code run faster by switching to a way of tracking time that does not require the use of `deepcopy`.

## Request

Instead of using `Time` instances to track dates, use `datetime.date` instances. This is just a suggested solution though, it will need to be thoroughly tested to ensure this is a viable solution.

PooyaHekmati:
### Context

Issue(s) closed by this pull request: closes #1922 

### What

Modifies the init of Time so that a time object can be created without needing config from IM. This in turn allows us to avoid deep copies used for time management.

### Why

To reduce the computation cost by avoiding unnecessary operations.

### How

Modifies `Time.__init__` to receive args.
Reduces import statement loads by avoid loading the entire module.
Removes deep copies from feed storage.

### Test plan

1- test suit
2- SME review
3- Profiling thanks to @JoeWaddell 
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 [![Flake8](https://img.shields.io/badge/Flake8-passed-brightgreen)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
 [![Pytest](https://img.shields.io/badge/Pytest-passed-brightgreen)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
 [![Coverage](https://img.shields.io/badge/Coverage-95%25-brightgreen)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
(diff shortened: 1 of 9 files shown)


ew3361zh (approved):
really nice, straight-forward and understandable change that has a huge impact. Nice work, Pooya!

Worked as intended - much speedier simulation runtimes.

JoeWaddell (approved):
Nice work! The changes to the code make sense, and from what I've reviewed generate identical results (both looking at outputs and tinkering with the debugger) - and with the added benefit of greatly reduced runtime - so it's an easy approval.

PR #2010, opened 2024-09-18
Update Input for Calculating N2O Emissions in Manure Treatments

### LINKED ISSUE #1879: [Manure] Manure storage N2O emissions fix

ew3361zh:
The input for calculating N2O emissions for storages was using manure daily_output which neglects the manure storage N removed for ammonia emissions earlier in the daily_update_helper() methods. It should be using the daily_input to better represent the total N in the manure that would contribute to those N2O emissions.

### Context

Issue(s) closed by this pull request: closes #1879

### What

There are 2 main things happening:
1. Updating the input for calculating empirical nitrogen loss from N2O emissions to be the daily manure input to the storage instead of the daily output (i.e. updating the amount of N in the liquid manure from post ammonia N removal to pre ammonia N removal).
2. Reworking the wrapper function for the GasEmissionsCalculator's `calculate_empirical_nitrogen_loss_from_nitrous_oxide_emission()` method that was located in the base manure treatment class. This function was effectively doing 2 things - getting the emissions factor based on storage type and cover status and then calling the `calculate_empirical_nitrogen_loss_from_nitrous_oxide_emission()` method with that information. I've updated this function to now just get the emissions factor and then we separately call the `calculate_empirical_nitrogen_loss_from_nitrous_oxide_emission()` method within each daily_update_helper() method.

Additionally renamed the `calculate_empirical_nitrogen_loss_from_nitrous_oxide_emission()` method to start with a verb (calculate).

### Why

The correct amount of N to use for the N2O emissions loss calculation is the total amount prior to any being removed for other emissions loss calculations.

### How

The references to the liquid_manure_nitrogen used for calculating N2O emissions have all been updated to use the daily_input amount rather than the daily_output amount.

The original base manure treatment class function that served as a 2-fold wrapper for getting the emissions factor and then calling the GasEmissionsCalculator function to calculate the N2O emissions has been reworked to now be a method just to get the emissions factor. The GasEmissionsCalculator N2O emissions calculation is called directly from the manure treatment `daily_update_helper()` methods.

### Test plan

Unit testing is updated.
Tested for results using this filter:
```
{
  "multiple": [
{
    "name": "Nitrous Oxide Emissions",
    "filters": [".*Pen_3.*storage_nitrous_oxide.*"],
    "vertical_aggregation": "sum"
}
  ]
}
```
But need confirmation from @elle-andreen that the results are as expected (saw a jump in nitrous oxide in the feature branch when compared to dev branch but not sure if it's the expected amount).

Screenshot 2024-09-22 at 9 48 32 PM


elle-andreen:
image

Some notes on emission changes from this PR:
- These values are from a ~850 lactating cow herd, storage type and manure management scenario for each pen are listed. This PR only affects outputs from the liquid storages. 
- Housing ammonia is not being subtracted from manure total N at the moment, so there is technically no N loss or gain between the animals and entering storage in these scenarios, except for the Closeup pen in this example, which has a screw press. So, in the branch outputs, the % of excreted N lost to N2O for all pens is equal to 0.50%. This is what we expect to see, as 0.50% is the N2O emission factor for liquid storages. Once ammonia N is subtracted correctly from total N for housing emissions.
- This PR will increase annual N2O emissions by 9-19%. This should not be weather dependent as the storage N2O function is based purely on storage type and manure N content. The numeric change is small though, and equates to about 70-80 g of N2O per cow per year (with slurry storage or anaerobic lagoon). This will likely decrease some whenever housing ammonia N subtraction is fixed. 
- Because more N is lost to N2O, there is less N to be lost to ammonia. The change is very slight though (< 0.1%) as NH3-N loss is much larger than N2O. 
- Side note, there is no storage ammonia emission from calves because all their ammoniacal N is being lost in housing at the moment. But, because housing ammonia N is not being subtracted from _total_ N, calves still have N2O emissions. This is also on the list to fix. 
- This change should not affect how storage ammonia emissions are calculated. Tthis change updates the _daily_ manure treatment N value, which is then added each day into the accumulated N value, and storage ammonia is being determined based on the accumulated N & TAN values.

elle-andreen (approved):
The PR is functioning as expected and all looks good to me! Thanks also for improving the actual N2O calculation function itself as well.

RUFAS/routines/manure/manure_treatments/base_manure_treatment.py
```diff
@@ -291,44 +290,29 @@ def calc_ammonia_emission(self, *args, **kwargs) -> float:
...
             The type of cover for the manure. Options are: cover, crust, no cover, and N/A.
-        manure_nitrogen_kg_N_per_day
-            The amount of manure nitrogen entering the manure treatment and storage system (kg N/day).
 
         Returns
         -------
         float
-            The empirical nitrogen loss from nitrous oxide emission (kg N/day).
+            Nitrous oxide emission factor (kg Nitrous Oxide N/kg manure N) for different manure treatment and storage
+            systems.
```
matthew7838: I'd like to suggest putting units at the end of sentence so it's more unified with other modules.

matthew7838 (approved):
LGTM, nice job Niko.

Issue #1878, opened 2024-07-31
[E2E Testing][SimEngine] Remove temporary fixes used to enable End-to-End testing in `SimulationEngine`


Isgotkowitz:
Remove code that was placed in `SimulationEngine` in order to make End-to-end testing possible on the Feed Storage module.

## Status

In #1784, code was added to the Simulation Engine to make the Feed Manager simulate the degradation of crops. This was necessary because the Feed Storage and Animal modules were not connected at the time, so even though crops may have been added to the Feed Manager by the Crop and Soil module, they were not being degraded as the simulation progressed.

## Intention

Clean up `SimulationEngine` by removing temporary fixes.

## Request

Remove all code related that is explicitly related to end-to-end testing in the `SimulationEngine`. Also remove the file `end_to_end_testing_inputs.json`.

### Additional Information

This issue is blocked until the Animal and Feed Storage modules are connected (#1782)

allisterakun:
Grooming session: @Isgotkowitz will come back and un-block this once #1782 is completed.

allisterakun:
Grooming session: @PooyaHekmati has taken #1782. Once he confirms the Animal and Feed Storage are fully connected, we will remove the temporary fix.

PR #2014, opened 2024-09-20
[Feed Storage] Evaluate and correct Feed Storage End-to-end testing results

### LINKED ISSUE #1996: [E2E Testing][Feed Storage] Verify expected outputs in the end-to-end testing routine of Feed Storage.
## Status

With the completion of PR #1784, end-to-end testing capabilties were added to RuFaS for the Feed Storage module. But the review process did not include a thorough evaluation of the outputs that were being collected from the Feed Storage module, which means that even though end-to-end testing is present it may not actually be testing Feed Storage properly.

## Intention

Review the outputs that are both expected and actually gathered from the Feed Storage module to ensure that end-to-end testing is thoroughly testing the Feed Storage module, and that the Feed Storage module is actually working correctly.

## Request

Work with SMEs (particularly @KevinP-B) to ensure that the end-to-end testing outputs for the Feed Storage module thoroughly test the modules capabilities and that the Feed Storage module is functioning properly.

Isgotkowitz:
Modifies the end-to-end testing inputs for Feed Storage and fixes some of the routines after reviewing outputs with SME.

### Context

Issue(s) closed by this pull request: closes #1996 

### What

Inputs for the Feed Storage's end-to-end testing routine are changed so that they look more like fresh crops that are about to go into storage, instead of looking crops that have already been stored for a while. After changing these inputs, the results from end-to-end testing were thoroughly reviewed and revealed some bugs in the Feed Storage routines.

### Why

During initial implementation of end-to-end testing, there was not a thorough review of the outputs being produced by the Feed Storage module. So issue #1996 was written so that we could take the time to tweak the Feed Storage end-to-end testing inputs and make them more realistic. 

### How

Bugs Fixed:
- Dry mass of hay was going negative, which was causing all sorts of issues. Fixed by playing with the numbers and finding that there were some fractions that should have been percents (arrrrrrggggghhh!!!!)
- Hay and Baleage were not losing any moisture. Fixed by forcing hay to lose moisture until it reaches 12% moisture by mass after 30 days of storage. Baleage loses moisture until it reaches 50% moisture. Currently that number is a default, but should be a user input (#2038)
- Silage and Baleage were losing some nutrients which should not have been lost, and not losing others that should have been lost. Fixed by changing which nutrient concentrations get recalculated after mass is lost.

### Test plan

Updated and expanded the test suite to verify correct behavior. Then ran end-to-end testing with 
```sh
python main.py -p input/metadata/end_to_end_testing_tm_metadata.json
```
To evaluate these results results, I spent a lot of time with the following CSV and worked through it several times with @KevinP-B to make sure that the masses of stored feeds seemed reasonable.
[end-to-end-testing_saved_variables_csv_feed_storage_filter.txt_21-Oct-2024_Mon_16-34-57.csv](https://github.com/user-attachments/files/17466979/end-to-end-testing_saved_variables_csv_feed_storage_filter.txt_21-Oct-2024_Mon_16-34-57.csv)

### Open Questions
- Unprotected hay loses a *lot* of crude protein - 40% of dry matter lost is CP, which quickly leads to CP content zeroing out if the initial content is low enough. This could be indicative of something problematic or off, or it could be a normal outcome of the Feed Storage routines. -- Will write up an issue to address this separately, this PR hits the most important issues.


RUFAS/routines/feed_storage/feed_manager.py
```diff
@@ -189,7 +189,8 @@ def purchase_feed(self) -> None:
     def setup_stored_feeds(self, feeds_info: dict[str, dict[str, str | float]], time: Time) -> None:
         """Sets up HarvestedCrops for the Feed Manager to degrade, if running end-to-end testing."""
         reusable_values = feeds_info["reusable_values"]
-        reusable_values.update({"harvest_time": time, "storage_time": time})
+        time_copy = Time(start_date=time.start_date, end_date=time.end_date, current_date=time.current_date)
+        reusable_values.update({"harvest_time": time_copy, "storage_time": time_copy})
```
Isgotkowitz, line 193: This change only effects end-to-end testing. If a copy is not made of the `Time` instance that is passed into this method, then the instance held in the `HarvestedCrop` will mutate every time `SimulationEngine` updates and this derails the degradation routines. Also this adds to the Mypy error count. Unfortunately there is no easy way to make it go away, so I think it best to wait until this code is removed (which will happen as part of #1878)

input/data/end_to_end_testing/end_to_end_inputs.json
Isgotkowitz, line 1: The new values here make the crops look more like what we would see in crops that are coming out of the field, where the old values were more like what the crops would look like after having been in storage for a long time.

RUFAS/routines/field/crop/crop_management.py
```diff
@@ -263,11 +263,13 @@ def _store_harvested_crop(self, time: Time, field_size: float, feed_manager: Fee
         It is assumed that the wet yield is recorded in kg / ha for crops, but stored in the FeedManager as kg.
 
         """
+        harvest_time = Time(start_date=time.start_date, end_date=time.end_date, current_date=time.current_date)
+        storage_time = Time(start_date=time.start_date, end_date=time.end_date, current_date=time.current_date)
```
Isgotkowitz, line 267: new copies of `Time` need to be made instead of the same instance that is passed into this method, because the instance passed into this method is modified by simulation engine on a daily basis and this will throw off the Feed Storage routines if it is used here.

RUFAS/routines/feed_storage/storage.py
```diff
@@ -170,17 +175,23 @@ def process_degradations(self, weather: Weather, time: Time) -> None:
...
-                crop.sugar, self.sugar_loss_coefficient, gaseous_dry_matter_loss, crop.dry_matter_mass
+            crop.lignin = self.recalculate_nutrient_percentage(
+                crop.lignin, self.lignin_loss_coefficient, gaseous_dry_matter_loss, crop.dry_matter_mass
+            )
+            crop.ash = self.recalculate_nutrient_percentage(
+                crop.ash, self.ash_loss_coefficient, gaseous_dry_matter_loss, crop.dry_matter_mass
             )
 
-            crop.last_time_degraded = copy.deepcopy(time)
+            crop.last_time_degraded = Time(time.start_date, time.end_date, time.current_date)
```
Isgotkowitz, line 194: This deep copy is removed because deep copying is bad! and really slows down the routines.

RUFAS/simulation_engine.py
Isgotkowitz, line 1: Ignore the changes in this file for now, changing the degradation interval length is only for observing the outputs with increased granularity

input/data/end_to_end_testing/e2e_json_feed_storage_filter.json
Isgotkowitz, line 1: This file contains the expected results for E2E testing that I updated by copying and pasting the new results, hence the massive changes

docs/feed_storage.rst
Isgotkowitz, line 1: Updated the "scientific" docs in this file, they need a more thorough rework at some point

RUFAS/routines/feed_storage/baleage.py
```diff
@@ -26,6 +37,23 @@ def __init__(self, capacity: float = float("inf")):
         ]
         self.bale_density: float = 0
 
+    def process_degradations(self, weather: Weather, time: Time) -> None:
```
ew3361zh, line 40: More of a design question but why did you choose to extend this function rather than write a separate moisture degradation function that calls the base class `process_degredations()`?

RUFAS/simulation_engine.py
```diff
@@ -211,4 +211,3 @@ def _initialize_simulation(self) -> None:
         if self.is_end_to_end_test_run:
             end_to_end_testing_inputs = self.im.get_data("end_to_end_testing_inputs")
             self.feed_manager.setup_stored_feeds(end_to_end_testing_inputs, self.time)
-            self.feed_manager.process_degradations(self.weather, self.time)
```
ew3361zh, line 214: Why is this removed?

RUFAS/routines/feed_storage/storage.py
```diff
@@ -382,6 +393,87 @@ def _get_conditions(
 
         return conditions
 
+    def _process_moisture_loss(self, time: Time, loss_period: int, final_moisture_percentage: float) -> None:
```
ew3361zh, line 396: What do you think about the verb here? Is this function doing any processing? It's maybe a little conflicting with the `_calculate_moisture_loss()` as far as the docstrings - maybe use `aggregate_moisture_losses()` or something that matches closer to the docstrings description but also separates it from the other `_calculate_moisture_loss()`

RUFAS/routines/field/crop/crop_management.py
ew3361zh: Do these need to be separate new instances of Time or could there be one updated instance shared between the two?

RUFAS/routines/feed_storage/baleage.py
Isgotkowitz: We want `process_degradations` to be the entry point for executing all losses on all `Storage` descendants - whether that is dry matter, moisture, or nutrients.

RUFAS/simulation_engine.py
Isgotkowitz: I think I changed this because I was using a different E2E feed storage degradation interval (1 day instead of 15). Having this first `process_degradations` call was a way to get a clear view of crop before any degradations had happened. It doesn't change the effectiveness of the tests, and eventually was going to be removed anyways as part of #1878

RUFAS/routines/feed_storage/storage.py
Isgotkowitz: The verb could definitely be better. What I really want it to convey is that it is actually removing moisture from crops, as opposed to just calculating how much moisture was lost.

Isgotkowitz: So I changed the docstring to make it clearer that `_process_moisture_loss` is different from `_calculate_moisture_loss`. However I kept the name of `_process_moisture_loss`, I think this is good not because the verb is the best but because it makes the naming more consistent with the naming of other methods in Feed Storage. Methods that process loss actually deduct mass from crops and record it, and methods that calculate loss just crunch the numbers for what should be lost.

ew3361zh: Very nice job with the updated docstrings, I think that makes it a lot clearer and makes the process verb a better fit imo.

RUFAS/routines/field/crop/crop_management.py
ew3361zh: Approving but wanted to get your thoughts on this question from the initial review @Isgotkowitz

ew3361zh (approved):
Changes looked good to me, well documented and understandable.
Tested locally and results were as expected.
Nice work!

One remaining question/clarification on the new instances of Time created in crop_management.py.

RUFAS/routines/field/crop/crop_management.py
Isgotkowitz: My bad! I think there could be one updated instance between the two, but if we are going to use the same `Time` instance for both I think we should reconsider why even need both of these variables at all. So for now, I think it is better to keep them separate, but we should revisit this a later date

Isgotkowitz:
> Changes looked good to me, well documented and understandable. Tested locally and results were as expected. Nice work!
> 
> One remaining question/clarification on the new instances of Time created in crop_management.py.

@ew3361zh thanks:), just replied

KevinP-B (approved):
This was a big one. Major goals accomplished and new sticking points (CP in hay, finetuning moisture loss) make sense for subsequent PR. Awesome work!

Isgotkowitz:
> This was a big one. Major goals accomplished and new sticking points (CP in hay, finetuning moisture loss) make sense for subsequent PR. Awesome work!

Thank you @KevinP-B :)

PR #2017, opened 2024-09-23
[Animal] Improve lactation curve module

### LINKED ISSUE #1924: [Animal] Improve the `LactationCurve` module

Isgotkowitz:
Improves the implementation of the `LactationCurve` module by cleaning up code and changing class structure.

### Context

Issue(s) closed by this pull request: closes #1924

### What

Instead of having `LactationCurve` operate as an instance that can be used to generate lactation curve parameters, it now has class methods and variables that are used to set and access lactation curve parameters. This makes access to the parameters easier because a single instance of the class doesn't have to be carefully managed, and more consistent because many of the other classes in the new Animal module will also be using class methods and variables.

### Why

The implementation of `LactationCurve` was a little hasty, so this PR refines the original implementation. It also makes the class design more consistent with how other base classes will be patterned in the animal module.

### How

All methods have been converted to class methods, and variables that need to be kept (base lactation parameters and their standard deviations) are now held in class variables. As requested in the issue, the variables holding the lactation parameters have been simplified so that they are less duplicative and there is better logging of activities and outputs to the OutputManager.

The method for calculating the 305 day milk production has been moved into the `MilkProduction` module, it was moved to the module because it does a better job of segregating responsibility between managing the lactation curve parameters specifically and calculating milk production.

I tried using `root_scalar` and vectorized integration, but found that root scalar did not work and vectorized integration provided no performance benefit. I suspect this is because `quad` is actually quite fast and the bottleneck was actually the daily calculation of milk production, which was accelerated with Numba in a different PR.

### Test plan

Unit tests have been updated to ensure that the behavior of the updated module has remained the same.


RUFAS/biophysical/animal/milk/lactation_curve.py
```diff
@@ -62,33 +62,31 @@ class LactationCurve:
 
     Attributes
     ----------
-    om : OutputManager
+    _om : OutputManager
```
ew3361zh, line 59: Why is OM a private attribute?

RUFAS/biophysical/animal/milk/lactation_curve.py
```diff
@@ -407,7 +384,7 @@ def _fit_wood_l_param_to_milk_yield(
 
         for l_param_error in np.arange(LOWER_BOUND, UPPER_BOUND, STEP_SIZE):
             l_param_varied = max(0.0, l_param + l_param_error)
-            varied_305_day_milk_yield = self.calc_305_day_milk_yield(l_param_varied, m_param, n_param)
+            varied_305_day_milk_yield = MilkProduction.calc_305_day_milk_yield(l_param_varied, m_param, n_param)
             milk_yield_difference = abs(varied_305_day_milk_yield - milk_yield)
             if milk_yield_difference < smallest_diff:
                 smallest_diff = milk_yield_difference
```
ew3361zh: ChatGPT thought this could be better optimized - was that explored at all?

Currently, the brute-force search over the range [-20, 20] with a step size of 0.01 results in over 4000 iterations. This can be quite inefficient. Instead, we can use the scipy.optimize.minimize function, which performs gradient-based or derivative-free optimization. This can quickly converge to the optimal solution without needing to test every point in the range.

ew3361zh (approved):
This looks good for what it's set out to accomplish. Seems like there's further room to continue improving the code - made a couple suggestions but overall it looked good.

RUFAS/biophysical/animal/milk/lactation_curve.py
Isgotkowitz: It was, but I only tried with Scipy's `root_scalar` and it did not work. I will try with `minimize` and see if it works better.

RUFAS/biophysical/animal/milk/lactation_curve.py
Isgotkowitz: It is an attribute so that it can be re-used in subroutines, and it is private because this reference to the OutputManager should only be used within `LactationCurve`. If the attribute wasn't private, it would open the door to other modules using the OM reference here (`LactationCurve.om.add_variable...`) which would not be a clean way to have other modules accessing the Output Manager.

RUFAS/biophysical/animal/milk/lactation_curve.py
Isgotkowitz: I tried using `minimize` and got it to work, and timed it against the original on the test cases. Interestingly, it performs about the same on some cases but wayyy faster on others.

Something within Scipy's `quadpack` implementation is causing a bunch of deprecation warnings being generated by `minimize`. These warnings could be caught in the same way that they are caught in `RationOptimizer`, it just wouldn't be the prettiest.

I will check in with Haowen and Yijing about this change, it changes the results very slightly (by 1/100th to be exact). Even if it winds up being a no-go, I think it would be too big of an issue because this function should only be getting called once in a simulation anyway

RUFAS/biophysical/animal/milk/lactation_curve.py
```diff
@@ -62,33 +62,31 @@ class LactationCurve:
...
 
     """
 
-    def __init__(self, time: Time) -> None:
+    _om = OutputManager()
+    _parity_to_parameter_mapping: dict[int, dict[str, float]] = {}
+    _parity_to_std_dev_mapping: dict[int, dict[str, float]] = {}
+
+    @classmethod
+    def set_lactation_parameters(cls, time: Time) -> None:
```
matthew7838, line 73: Missing docstrings

matthew7838 (approved):
LGTM, caught a missing docstring. Also, is the extra mypy error unpreventable?

RUFAS/biophysical/animal/milk/lactation_curve.py
ew3361zh: We should probably update the other attribute references to om (and im) to be private then to match this.

allisterakun (commented):
Question: For all the `cls` methods other than `set_latacion_parameters()`, what are your thoughts on making them static methods instead of class methods?

RUFAS/biophysical/animal/milk/lactation_curve.py
Isgotkowitz: Agreed. Although I think doing that should not be a high priority issue

RUFAS/biophysical/animal/milk/lactation_curve.py
Isgotkowitz: Good catch

Isgotkowitz:
> Question: For all the `cls` methods other than `set_latacion_parameters()`, what are your thoughts on making them static methods instead of class methods?

There are four methods which could be converted to static as-is, all of the other methods use class variables. I think making them static methods would not be a bad idea, but there are a couple things that come to mind:
1. If we leave a mix of class methods and static methods, is that ok or would it be better to prioritize more consistent use of staticmethods? I like consistency, but this might be one of those cases where we don't gain anything by making the code super consistent.
2. If we prioritize converting class methods to static methods, it could mean more instantiations of the OutputManager. This wouldn't be a big deal, because these all the class methods that use this routine are only supposed to run once at the beginning the simulation but it would still be good to keep in mind.

Edit: also, I think the method `get_wood_parameters` should stay a class method as well, because this is the only method that really needs to access those class vars besides `set_lactation_parameters`

RUFAS/biophysical/animal/milk/lactation_curve.py
```diff
@@ -402,15 +384,12 @@ def _fit_wood_l_param_to_milk_yield(
...
-        for l_param_error in np.arange(LOWER_BOUND, UPPER_BOUND, STEP_SIZE):
-            l_param_varied = max(0.0, l_param + l_param_error)
-            varied_305_day_milk_yield = self.calc_305_day_milk_yield(l_param_varied, m_param, n_param)
-            milk_yield_difference = abs(varied_305_day_milk_yield - milk_yield)
-            if milk_yield_difference < smallest_diff:
-                smallest_diff = milk_yield_difference
-                l_param_best_fit = l_param_varied
+
+        def objective(l_param_varied: float) -> float:
+            return abs(MilkProduction.calc_305_day_milk_yield(l_param_varied, m_param, n_param) - milk_yield)
```
Isgotkowitz: Pull this function out and make it its own thing

RUFAS/biophysical/animal/milk/lactation_curve.py
```diff
@@ -55,40 +52,43 @@ class LactationCurve:
...
 
     """
 
-    def __init__(self, time: Time) -> None:
+    _om = OutputManager()
+    _parity_to_parameter_mapping: dict[int, dict[str, float]] = {}
+    _parity_to_std_dev_mapping: dict[int, dict[str, float]] = {}
+
+    @classmethod
+    def set_lactation_parameters(cls, time: Time) -> None:
```
YijingGong, line 73: Not clear to me what is cls? Seeing it in the description as well, but any meaning of this abbreviation？

YijingGong (approved):
Hi Ed, it looks good to me. I was trying to test & print myself to see if the results from the new method matches the values produced from the original method. But I got lost - there seems two lactation_curve.py (one under milk folder, and one under life_cycle folder), and there's no cohort farm animal file on this branch as well. 

But it sounds to me you tested it and they differ at the level of 0.01, is that correct? If that is the case, I won't be worry about it. And the use of scipy.optimize minimize makes sense to me. 

Another thing that catch my attention is that the milking cow fraction is not a user input now, instead, it is now in the lactation_curve_adjustment_inputs.json. I am not sure, maybe this file is subject to changes as well, but to me it looks like they are more of a constant type of thing. To me the "milking_cow_fraction" should be something like "parity_fractions", where users can input values, if they don't, we use default values. It's probably not a big deal, but could you explain me why it is decided to be separate from the "parity_fractions"?

RUFAS/biophysical/animal/milk/lactation_curve.py
Isgotkowitz: When a method is decorated with `@classmethod`, it is the convention (possibly the rule) for the first parameter to be named `cls`. This `cls` acts similarly to `self` in regular instance methods, except it gives access to the entire class instead of just a single instance

Isgotkowitz:
Thank you for reviewing this Yijing:)

> Hi Ed, it looks good to me. I was trying to test & print myself to see if the results from the new method matches the values produced from the original method. But I got lost - there seems two lactation_curve.py (one under milk folder, and one under life_cycle folder), and there's no cohort farm animal file on this branch as well.
> 

With the big refactor of the animal module, we are leaving all the old stuff in `routines` and implementing the new modules in `biophysical`, which is why I only modified the `lactation_curve.py` file in `biophysical`. The reason there are no cohort farms is because they are not kept on the `dev` branch.

> But it sounds to me you tested it and they differ at the level of 0.01, is that correct? If that is the case, I won't be worry about it. And the use of scipy.optimize minimize makes sense to me.
> 

Yes, that is the greatest difference I observed in the test cases. And it just has to do with how Numpy's `arange` works, which we are moving away from.

> Another thing that catch my attention is that the milking cow fraction is not a user input now, instead, it is now in the lactation_curve_adjustment_inputs.json. I am not sure, maybe this file is subject to changes as well, but to me it looks like they are more of a constant type of thing. To me the "milking_cow_fraction" should be something like "parity_fractions", where users can input values, if they don't, we use default values. It's probably not a big deal, but could you explain me why it is decided to be separate from the "parity_fractions"?

This value is not a constant so users can modify it the same as they would the parity fractions. However I agree, it seems weird and inconvenient that it is in the lactation adjustments input and not the general animal input file. I cannot remember why it ended up there, I may have just been going too quickly to think through it properly. This would not be a difficult change to make in terms of changing the code, but we have to be very careful and deliberate when changing model inputs. I can make a note that this should be changed when we change the other inputs for the Animal refactor.

Isgotkowitz:
Update: while refining this PR, I found that a ton of warnings were being raised by Scipy as a result of calling `minimize`. They were all buried pretty deep in Scipy (something to do with how quadpack uses Numpy). This is the following error:
`{message : DeprecationWarning('Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)'), category : 'DeprecationWarning', filename : '/home/isgotkowitz/rufas/env/lib/python3.10/site-packages/scipy/integrate/_quadpack_py.py', lineno : 606, line : None}`

I remember a similar (the exact same??) issue coming up with `minimize` in the ration formulation routines (see `ration_optimizer.py`), so I used the same strategy to capture these warnings. Pretty sure this isn't a big issue, but will check with the rest of the dev team before merging this and see if a re-review is necessary.

tomhuhh (commented):
Thank you Ed for streamlining the workflow for the lactation curve functionalities!

PR #2022, opened 2024-09-24
[Manure] Remove DefaultEnum from Manure Module

### LINKED ISSUE #1170: Remove the default enum classes in the manure module to make use of the defaults in IM
Most of the enum classes in the manure module are extended from the `DefaultEnum` class. The original motivation for that class was to have a default option in the case of a failed name-lookup. Now that the input manager has a way to specify the default values, we should revisit all these enum classes and see if we can integrate those user-defined default values into the manure module.

ew3361zh:
Remove all references to custom Enum class DefaultEnum from Manure Module.

#### Important Note: I had to make some small changes in the GH Action because an error arose when I deleted the default_enum files. 
1. There is a new piece to the `Get all added or modified files` step which gets all the modified files except for the deleted files (`MODIFIED_FILES_ONLY`).
2. I kept the `ALL_CHANGED_FILES` var which will gather all the deleted files as well as added and modified files. This is still useful for the `Check for Unauthorized Changes` step because we don't want a protected file to be deleted and then slip under the radar because it's not a modified or added file.
3. Black now runs on the `MODIFIED_FILES_ONLY` so it won't error when trying to format a file that no longer exists.
4. `Check for Unauthorized Changes` now runs on `ALL_CHANGED_FILES` which includes deleted files as well as modified and added files.

### Context

Issue(s) closed by this pull request: closes #1170

### What

Removes the final reference to the use of the DefaultEnum class in Composting class.
Updates unit testing to reflect this change.
Removes DefaultEnum class and associated unit testing.

### Why

This class is no longer needed because InputManager is ensuring default values for all classes using the DefaultEnum previously.

### How

Deleted the class and testing file associated with DefaultEnum.
Updated the init function in Composting to read the config setting for which CompostingType is specified by the input and assigns it directly.

### Test plan

Unit testing updated to reflect new changes.
Ran simulation using one of the composting manure scenarios (so update one or several of the `default_animal.json` `pen_information` scenarios to use a `manure_management_scenario` that runs with `composting` - 11, 12, or 13 currently). No issues with the simulation running and composting being implemented.
I also tried changing the `composting_type` in the `manure_treatment_configs` to a nonsense word (forcing IM to use the default) and then to one of the other composting enum options other than `intensive windrow`.
All scenarios ran successfully.


changelog.md
```diff
@@ -45,4 +45,5 @@ v0.9.2
...
 - [2001](https://github.com/RuminantFarmSystems/MASM/pull/2001) - [minor change] [E2E Testing] Changes E2ETestResultsComparer so that it constructs file paths properly on Windows machines.
 - [2002](https://github.com/RuminantFarmSystems/MASM/pull/2002) - [minor change] [Manure] Adds a method to determine the ambient inside barn temperature based on the outdoor air temperature for use in calculations of emissions associated with Manure Handlers.
-- [2009](https://github.com/RuminantFarmSystems/MASM/pull/2009) - [minor change] [Time][Feed Storage] Modifies the init of Time so that a time object can be created without needing config from IM. This in turn allows us to avoid deep copies used for time management.
\ No newline at end of file
+- [2009](https://github.com/RuminantFarmSystems/MASM/pull/2009) - [minor change] [Time][Feed Storage] Modifies the init of Time so that a time object can be created without needing config from IM. This in turn allows us to avoid deep copies used for time management.
+- [2022](https://github.com/RuminantFarmSystems/MASM/pull/2022) - [minor change] [Manure] Removes DefaultEnum class and all references to it in the Manure module.
```
Isgotkowitz: I believe this should go in the section of the changelog above this, labeled "Next Version Updates"

Isgotkowitz (approved):
LGTM! Nice job Niko

matthew7838 (commented):
LGTM

matthew7838 (approved):

changelog.md
ew3361zh: Ah, thanks, I did not see that section.

PR #2023, opened 2024-09-25
[Manure] Fix to how organic bedding dry solids are calculated in manure handlers

### LINKED ISSUE #1975: [Manure] Bedding volatile solids addition fix
## Expected Behavior
PR #1890 added solids from organic bedding cleaned out of stalls to the non-degradable volatile solids fraction of manure. This occurs in the manure handler. The quantity of mass added from bedding should be the **dry solids** in the bedding (e.g. total mass * bedding DM), not the entire bedding mass (the beddings are 90% dry solids by default, this is the `bedding_dry_matter_content` input). 

## Current Behavior
Currently, the total mass of bedding (dry solids plus water) is being added to the non-degradable VS fraction of manure at the manure handler step. 

## Solution
In `bedding_classes.py`, on line 210, there is a function `calc_total_bedding_dry_solids` to calculate the dry solids in bedding. The value returned by that function is what should be used for `bedding_mass` in the function to determine organic bedding mass added to manure non-degradable VS (`def calc_organic_bedding_mass_added_to_manure` in `manure_handler_classes.py`).

## Steps to Reproduce
Utilize a manure management scenario with any of the organic beddings (i.e. anything but sand). Compare the change in manure handler non-degradable VS to the quantity of animal-excreted manure non-degradable VS, plus the mass of bedding solids cleaned per day.

## Target Deadline
This does not result in a huge numerical impact on manure emissions and composition, but should be addressed sooner than later, and should be a simple fix!

ew3361zh:
Small fix to how organic bedding dry solids are calculated in manure handlers.

### Context

Issue(s) closed by this pull request: closes #1975

### What

Adjusts the `total_organic_bedding_mass_added` obtained from calculating the organic bedding mass added to manure by the bedding dry matter content fraction. This better represents that only dry-solids should be added to the total of `non_degradable_volatile_solids` reported in the ManureHandlerDailyOutput.

### Why

The quantity of mass added from bedding should be the dry solids in the bedding (e.g. total mass * bedding DM fraxction), not the entire bedding mass.

### How

Once the `total_organic_bedding_mass` is calculated, it's now multiplied by the bedding's `bedding_dry_matter_content` to better approximate the dry solids portion.

### Test plan

Unit testing is updated. Exact amounts and whether the change reflects expected results needs to be confirmed by SME.

### DIFF
--- a/RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
+++ b/RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
@@ -159,11 +159,12 @@ class BaseManureHandler:
         if bedding:
             total_bedding_mass = bedding.calc_total_bedding_mass(pen.num_animals)
             total_organic_bedding_mass_added = bedding.calc_organic_bedding_mass_added_to_manure(total_bedding_mass)
+            organic_bedding_dry_solids = total_organic_bedding_mass_added * bedding.bedding_dry_matter_content
         else:
             total_bedding_mass = 0.0
-            total_organic_bedding_mass_added = 0.0
+            organic_bedding_dry_solids = 0.0
 
-        non_degradable_volatile_solids = pen.manure.non_degradable_volatile_solids + total_organic_bedding_mass_added
+        non_degradable_volatile_solids = pen.manure.non_degradable_volatile_solids + organic_bedding_dry_solids
 
         daily_output = ManureHandlerDailyOutput(
             simulation_day=sim_day,
@@ -188,7 +189,7 @@ class BaseManureHandler:
             cleaning_water_volume=self.calc_cleaning_water_volume_in_main_barn(pen.num_animals),
             total_bedding_volume=bedding.calc_total_bedding_volume(pen.num_animals) if bedding is not None else 0.0,
             total_bedding_mass=total_bedding_mass,
-            organic_bedding_added_to_manure=total_organic_bedding_mass_added,
+            organic_bedding_added_to_manure=organic_bedding_dry_solids,
             total_water_volume_in_milking_parlor=(
                 self.milking_parlor.calc_total_water_volume_used_in_milking_parlor(pen.num_lactating_cows)
             ),
--- a/RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
+++ b/RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
@@ -157,6 +157,7 @@ class BaseManureHandler:
             organic_bedding_dry_solids = total_organic_bedding_mass_added * bedding.bedding_dry_matter_content
         else:
             total_bedding_mass = 0.0
+            total_organic_bedding_mass_added = 0.0
             organic_bedding_dry_solids = 0.0
 
         non_degradable_volatile_solids = pen.manure.non_degradable_volatile_solids + organic_bedding_dry_solids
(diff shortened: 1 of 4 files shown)


RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
```diff
@@ -154,11 +154,12 @@ def daily_update(
         if bedding:
             total_bedding_mass = bedding.calc_total_bedding_mass(pen.num_animals)
             total_organic_bedding_mass_added = bedding.calc_organic_bedding_mass_added_to_manure(total_bedding_mass)
+            organic_bedding_dry_solids = total_organic_bedding_mass_added * bedding.bedding_dry_matter_content
         else:
             total_bedding_mass = 0.0
-            total_organic_bedding_mass_added = 0.0
```
elle-andreen, line 159: Does this need to stay in because we still have `total_organic_bedding_mass_added` in the above `if` statement?

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
```diff
@@ -183,7 +184,7 @@ def daily_update(
             cleaning_water_volume=self.calc_cleaning_water_volume_in_main_barn(pen.num_animals),
             total_bedding_volume=bedding.calc_total_bedding_volume(pen.num_animals) if bedding is not None else 0.0,
             total_bedding_mass=total_bedding_mass,
-            organic_bedding_added_to_manure=total_organic_bedding_mass_added,
+            organic_bedding_added_to_manure=organic_bedding_dry_solids,
```
elle-andreen: This was actually correct originally! We do still want to output the total organic bedding mass (inclusive of water) here, not just the dry solids. I don't think at this time we necessarily need to output the organic bedding dry solids separately.

ew3361zh: Ah gotcha - easy enough to put back in. In that case, the answer to your earlier question is that yes, we need to leave the zeroed out version of `total_organic_bedding_mass_added` in the else statement. Thanks for clarifying!

RUFAS/routines/manure/manure_handlers/manure_handler_classes.py
ew3361zh: Based on your other comment, yes, we should leave the original `total_organic_bedding_mass_added` in. I will change it back.

changelog.md
```diff
@@ -27,8 +27,10 @@ v0.9.2
 
 - [2021](https://github.com/RuminantFarmSystems/MASM/pull/2021) - [minor change] [Animal] **Adds dependency**. Creates a new `MilkProduction` submodule and several support submodules to calculate milk production.
 - [2022](https://github.com/RuminantFarmSystems/MASM/pull/2022) - [minor change] [Manure] Removes DefaultEnum class and all references to it in the Manure module.
+- [2023](https://github.com/RuminantFarmSystems/MASM/pull/2023) - [minor change] [Manure] Fixes calculation of organic bedding dry solids for manure handlers.
```
elle-andreen:
```suggestion
- [2023](https://github.com/RuminantFarmSystems/MASM/pull/2023) - [minor change] [Manure] Fixes calculation of organic bedding dry solids added to manure in manure handler.
```

elle-andreen (approved):
Left one final suggestion on the changelog verbiage, but this looks good to go otherwise. Thanks Niko!

allisterakun (approved):
Simply and straight-forward changes. LGTM, thanks, Niko!

PR #2025, opened 2024-09-27
Sorted imports in main.py, setup.py, RUFAS, and tests

Isgotkowitz:
Sorts imports in the codebase.

Note that if you look through the changed files, you will see some relative imports that should be absolute. I decided not to fix any of them, because this PR is already massive and I didn't want to start throwing in more changes. If converting the no-no relative imports to absolute is a priority, please leave a comment and I would be happy to get them out of the codebase in a separate PR.

### Context

Issue(s) closed by this pull request: closes # doesn't close anything, this was an Action Item from retro.

### What

Sorts the imports in the repo, pretty self-explanatory.

### Why

Makes imports easier on the eyes.

### How

`isort --profile black --line-length 120 main.py setup.py RUFAS tests`

### Test plan

Ran the test suite to verify that everything is still passing, and ran the default scenario and checked that outputs are unchanged.

### DIFF
--- a/RUFAS/biophysical/animal/milk/milk_production.py
+++ b/RUFAS/biophysical/animal/milk/milk_production.py
@@ -1,3 +1,4 @@
+import numpy as np
 from numba import njit
 
 from RUFAS.biophysical.animal.animal_constants import DRY
@@ -8,8 +9,6 @@ from RUFAS.biophysical.animal.data_types.milk_production_record import MilkProdu
 from RUFAS.general_constants import GeneralConstants
 from RUFAS.time import Time
 from RUFAS.util import Utility
-
-import numpy as np
 
 
 class MilkProduction:
--- a/RUFAS/routines/EEE/EEE_manager.py
+++ b/RUFAS/routines/EEE/EEE_manager.py
@@ -1,5 +1,5 @@
-from .emissions import EmissionsEstimator
 from ...output_manager import OutputManager
+from .emissions import EmissionsEstimator
 
 
 class EEEManager:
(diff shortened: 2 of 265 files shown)


RUFAS/biophysical/animal/milk/milk_production.py
```diff
@@ -1,3 +1,4 @@
+import numpy as np
 from numba import njit
```
allisterakun, line 2: While at it, do we also want to change these inputs to only input the necessary functions instead of the entire module?

RUFAS/routines/EEE/EEE_manager.py
```diff
@@ -1,5 +1,5 @@
-from .emissions import EmissionsEstimator
 from ...output_manager import OutputManager
```
allisterakun, line 1: And relative imports?

allisterakun (approved):
Thank you, Ed; it looks much cleaner now!
I have two comments on whether we want to spend time fixing the relative imports and only import the necessary functions instead of the entire module.
Otherwise, LGTM!

Isgotkowitz:
> Thank you, Ed; it looks much cleaner now! I have two comments on whether we want to spend time fixing the relative imports and only import the necessary functions instead of the entire module. Otherwise, LGTM!

Thank you @allisterakun. I think fixing both relative imports and the importing of unnecessary functions are both important and should be addressed. But because this PR touches so many files and it doesn't seem feasible for reviewers to check all files, I didn't want to start adding in manual changes on top of isort. I'm not sure though, maybe it would be better to get all the imports fixed at once because the changes are pretty trivial.

ew3361zh (approved):
LGTM!

PR #2031, opened 2024-10-04
[Formatting] Fix relative imports

Isgotkowitz:
Removes relative imports that import modules above the module where the import is happening.

### Context

Issue(s) closed by this pull request: none, just cleaning up the codebase

### What

Relative imports that start with one `.` are ok, relative imports that start with more than one `.` are bad. This PR removes all instances of the latter from the codebase.

### Why

TBH I'm still not 100% sure why a relative import of a module that is above another module is bad, but I'm guessing it just generally makes the code more obfuscated and less easily readable.

### How

`grep "^from \.\.*." -r RUFAS/ tests/ main.py setup.py` to find all relative imports in the codebase. Changed them to use the absolute path if they started with more than one dot. Then ran `isort --profile black --line-length 120 RUFAS/ tests/ main.py setup.py` to resort all the import statements.

### Open Question
- Would it be a good idea to add `isort` to `requirements.txt` and configure it in `pyproject.toml` but not add it to the GitHub Action so that anyone can run isort whenenver they feel like it without having to worry about configuring it correctly?

### Test plan

Ran default simulation on this branch and `dev`. Compared outputs from both simulations to make sure they were identical. Also reran the test suite to make sure that it is still passing


ew3361zh:
Some chatGPT context for when relative imports are bad:

When Relative Imports Are Bad:

    Larger Projects:
        In large projects with complex structures, relative imports can become confusing, especially as you move between deeply nested modules. It becomes harder to trace the location of the module being imported.
        Absolute imports make the structure more understandable by providing the full path to the module, which helps developers unfamiliar with the project understand it more quickly.

    Refactoring Issues:
        Relative imports can cause issues when refactoring. If the directory structure of the project changes, relative imports may break, whereas absolute imports remain valid as long as the module path exists somewhere in the Python path.

    Testing and External Execution:
        Relative imports can complicate running scripts directly or when running tests, especially outside the package root. If you run a file that relies on relative imports, Python may not know how to resolve them without being run as part of the package. This often forces you to adjust the PYTHONPATH, which can lead to confusion and brittle setups.

    Cross-Package Imports:
        If your project is organized into multiple packages or modules that interact with each other, relative imports can create tight coupling between them. Absolute imports make the code more modular and easier to decouple.


Best Practices:

    Use absolute imports when:
        Your project has a complex structure or is growing.
        You want clearer imports that are less dependent on the specific file hierarchy.
        You want easier cross-module and cross-package imports.

    Use relative imports when:
        You are working within a well-defined package and only importing modules that are very close to each other.
        You want to avoid hard-coding absolute paths within a package, especially in smaller projects or packages meant for reuse.

ew3361zh (approved):
beautiful

matthew7838 (approved):
LGTM

PR #2033, opened 2024-10-14
Consolidate methods in IM

### LINKED ISSUE #1489: [IM] Consolidate methods for adding data to the Input Manager at runtime
The various methods for adding or replacing input to the Input Manager at runtime should be consolidated into a single method.

## Expected Behavior

There should be one method available for adding input to the Input Manager at runtime (i.e. mid-simulation), and this method should be capable of handling all different types of data that RuFaS needs to add to the Input Manager during runtime.

## Current Behavior

Two different methods are present in the Input Manager for adding data during runtime, `add_dict_variable_to_pool` and `add_tabular_data_to_pool`, which respectively add dicts and lists to the Input Manager.

## Possible Solution

Consolidate `add_dict_variable_to_pool` and `add_tabular_data_to_pool` into a single method. Also ensure that this new method will be capable of adding all different types of data that RuFaS needs to add to the Input Manager during runtime.

Note that this issue is strongly tied to #1273.

matthew7838:
This PR will make adding runtime variables to only use one method.
### Context

Issue(s) closed by this pull request: closes #1489

### What

The method `add_tabular_data` will now be removed as its adding dictionary part duplicates with the `add_dict_variable_to_pool` method and the adding List part functions under the assumption that the List tabular data can be preprocessed into dictionary variables before adding it in runtime.

### Why

We should not have multiple methods of handling adding runtime variables, let alone the current state of only having one type to support.

### How

- Removed `add_tabular_data`.
- `add_dict_variable_to_pool` changed to `add_runtime_variable_to_pool`.

### Test plan

1. Run with complete test suites that reflect the change of naming.
2. Run with default metadata with no errors.


RUFAS/biophysical/animal/herd_factory.py
```diff
@@ -367,7 +367,7 @@ def initialize_herd(self) -> None:
             self.pre_animal_population = self._initialize_herd_from_data()
 
         self.post_animal_population = self._random_sample_with_replacement()
-        self.im.add_dict_variable_to_pool(
+        self.im.add_variable_to_pool(
```
Isgotkowitz:
```suggestion
        self.im.add_runtime_variable_to_pool(
```

changelog.md
```diff
@@ -60,4 +60,3 @@ v0.9.2
 - [2001](https://github.com/RuminantFarmSystems/MASM/pull/2001) - [minor change] [E2E Testing] Changes E2ETestResultsComparer so that it constructs file paths properly on Windows machines.
 - [2010](https://github.com/RuminantFarmSystems/MASM/pull/2010) - [minor change] Updates liquid manure nitrogen input source used to calculate N2O emissions for manure storages.
 - [2002](https://github.com/RuminantFarmSystems/MASM/pull/2002) - [minor change] [Manure] Adds a method to determine the ambient inside barn temperature based on the outdoor air temperature for use in calculations of emissions associated with Manure Handlers.
-- [2009](https://github.com/RuminantFarmSystems/MASM/pull/2009) - [minor change] [Time][Feed Storage] Modifies the init of Time so that a time object can be created without needing config from IM. This in turn allows us to avoid deep copies used for time management.
```
Isgotkowitz: This changelog entry should not be deleted.

tests/soil_crop_tests/manager_tests/test_field_manager.py
```diff
@@ -41,7 +41,7 @@ def input_manager_original_method_states(
     """Fixture to store original methods of InputManager"""
     return {
         "get_data": mock_input_manager.get_data,
-        "add_dict_variable_to_pool": mock_input_manager.add_dict_variable_to_pool,
+        "add_runtime_variable_to_pool": mock_input_manager.add_runtime_variable_to_pool,
```
Isgotkowitz:
```suggestion
```
Not sure why this is even in here because it isn't used at all, I think it would be good to remove it.

tests/test_input_manager.py
```diff
@@ -2287,7 +2286,7 @@ def test_add_variable_to_pool_eager_termination(
         ("var3", {"a": "A", "b": 2, "c": True}, "key3"),
```
Isgotkowitz, line 1580: Could you please add a test case (or cases) where list data is added? I think some of the cases from the `test_add_tabular_variable_to_pool` could be reused here.

Isgotkowitz (commented):
Code LGTM. There are a few small things that need to be addressed before approval, but they should be pretty easy to take care of

changelog.md
matthew7838: Thanks for catching this again, I keep accidentally pressing the accept yours button on Pycharm and this happens.

tests/test_input_manager.py
matthew7838: I am not sure what you mean here, adding a list will throw a type error which is tested in `test_add_runtime_variable_to_pool_type_error` .

matthew7838: Follow up, the method you mentioned in the comments was not modified. The modified method is `add_runtime_variable_to_pool`.

Isgotkowitz: `add_runtime_variable_to_pool` should be capable of adding data of any type to the pool at runtime, it just has to be packaged in a dictionary. `add_tabular_variable_to_pool` accepted both data that was just in a list or lists that were packaged in dictionaries.

Consider these two test cases from `test_add_tabular_variable_to_pool`:
1. `("var3", [0.0, 1.1, 2.2], "key3")`
2. `("var4", {"a": [1, 2, 3], "b": ["a", "b", "c"], "c": [0.0, 1.1, 2.2]}, "key4")`

1 shouldn't pass, because the data is just a list. But 2 should, because the list data is packed in a dictionary. Does this make more sense?

Isgotkowitz (approved):
Nice!

ew3361zh:
This mostly seems good but how was this tested (beyond unit tests) to show that it's successfully able to add variables at runtime? Are there variables added at runtime for the default simulation?

ew3361zh:
Also should this be closing #1489? It's marked as closing #1335 which is the parent issue. Or is it both?

matthew7838:
> This mostly seems good but how was this tested (beyond unit tests) to show that it's successfully able to add variables at runtime? Are there variables added at runtime for the default simulation?

Unfortunately no, there are not really many options to test it right now besides unit tests.

matthew7838:
> Also should this be closing #1489? It's marked as closing #1335 which is the parent issue. Or is it both?

Good catch.

ew3361zh:
> > This mostly seems good but how was this tested (beyond unit tests) to show that it's successfully able to add variables at runtime? Are there variables added at runtime for the default simulation?
> 
> Unfortunately no, there are not really many options to test it right now besides unit tests.

I just noticed that this function (an its predecessor) is used in herd initialization. Running the herd init task to trigger this use still works which is a good sign.

ew3361zh (approved):
LGTM!

PR #2035, opened 2024-10-17
Animal Reproduction Implementation

### LINKED ISSUE #1997: [Animal][Reproduction ] Reproduction Refactor Implementation [3/3]
Follow the Design in this Google Doc: [Milk Production](https://docs.google.com/document/d/16lsM66imhzHiBc_6oo-YsPKOK3MiMtbtSkg5OnWTXio/edit?usp=sharing) to implement the refactor of the MilkProduction submodule.

allisterakun:
This PR implements the Reproduction submodule logic and necessary unit tests.

### Context

Issue(s) closed by this pull request: closes #1997 

### What

This PR refactors the animal reproduction-related code in the animal module. The original logic was in each animal class (HeiferII and Cow). It is extracted from the animal classes, put into a stand-alone submodule, and added to the new `biophysical` module.

### Why

The various subroutines within the Animal module are very entangled; by separating them into more dedicated routines, we can significantly improve the code - clarity and performance.

### How

While extracting the logic from the animal classes, we also broke down many complex methods to improve the code clarity.

### Test plan

Write necessary unit tests to bring the coverage for the reproduction submodule to  92%. The exception is for the `determine_future_cull_date()` and `determine_future_death_date()` functions; I believe that there may be some incorrect logic that would lead to unintended results.

Unit tests cover all methods (but the two mentioned above), and all mock test cases run error-free. At this point, we are not able to run it in a simulation because the animal module in the new `biophysical' module has yet to be set up.


RUFAS/biophysical/animal/reproduction/reproduction.py
```diff
@@ -1,2 +1,2867 @@
...
+                        < AnimalConfig.death_day_probability[i + 1]
+                ):
+                    death_lower_limit = AnimalConfig.death_day_probability[i]
+                    death_upper_limit = AnimalConfig.death_day_probability[i + 1]
+                    death_time_lower_limit = AnimalConfig.death_day_probability[i]
+                    death_time_upper_limit = AnimalConfig.death_day_probability[i + 1]
+            n = (death_time_upper_limit - death_time_lower_limit) / (death_upper_limit - death_lower_limit)
+            print(round(
+                death_time_lower_limit + n * (death_date_random - death_lower_limit) + reproduction_outputs.days_born
+            ))
```
Isgotkowitz: Does this need to be printed?

RUFAS/biophysical/animal/reproduction/reproduction.py
```diff
@@ -1,2 +1,2867 @@
...
+        reference_rate : float
+            Reference rate for comparison.
+
+        Returns
+        -------
+        bool
+            True if the randomized rate is less than the reference rate, False otherwise.
+        """
+
+        return random() < reference_rate
```
Isgotkowitz, line 574: Does this still need to be moved to util?

allisterakun: I think it should

RUFAS/biophysical/animal/reproduction/reproduction.py
allisterakun: Will remove!

RUFAS/biophysical/animal/data_types/reproduction_outputs.py
```diff
@@ -0,0 +1,107 @@
...
+    future_cull_date: int
+    future_death_date: int
+    net_merit: float
+    phosphorus_for_gestation_required_for_calf: float
+
+    animal_level_statistics: AnimalReproductionStatistics
+    herd_level_statistics: HerdReproductionStatistics
+
+    newborn_calf_config: dict
+    new_calf_born: bool = False
```
Isgotkowitz: Instead of having two regular attributes (`newborn_calf_config` and `new_calf_born`), what would you think about making `new_born_calf` an `@property` method that is true if the `newborn_calf_config` is not an empty dictionary, false if it is? I think this would deduplicate the data a little, making it impossible to be in a situation where there is a configuration for a newborn calf but `new_calf_born` is false and vice versa.

RUFAS/biophysical/animal/data_types/reproduction_inputs.py
```diff
@@ -0,0 +1,26 @@
...
+
+@dataclass
+class ReproductionInputs:
+    animal_type: AnimalType
+    body_weight: float
+    breed: Breed
+    cull_reason: str
+    days_born: int
+    days_in_pregnancy: int
+    days_in_milking: int
```
Isgotkowitz:
```suggestion
    days_in_milk: int
```
This keeps it consistent with the `GeneralProperties`.

RUFAS/biophysical/animal/data_types/preg_check_config.py
```diff
@@ -1,7 +1,7 @@
 from typing import TypedDict
 
 
-class PregCheckConfig(TypedDict, total=False):
+class PregnancyCheckConfig(TypedDict, total=False):
```
Isgotkowitz, line 4: Nice, love to see it

RUFAS/biophysical/animal/reproduction/reproduction.py
```diff
@@ -1,2 +1,2864 @@
...
+
+        Returns
+        -------
+        ReproductionOutputs
+            Updated reproduction outputs for the animal.
+        """
+        reproduction_outputs = ReproductionOutputs(
+            **asdict(reproduction_inputs),
+            animal_level_statistics=AnimalReproductionStatistics(),
+            herd_level_statistics=HerdReproductionStatistics(),
```
Isgotkowitz, line 187: If I understand correctly, each animal will have a new `HerdReproductionStatistics` initialized for it, and then I assume that instance will be returned back up to `HerdManager` which will combine all the HerdReproductionStatistics into one instance which actually represents the herd. Would it be possible to avoid having multiple `HerdReproductionStatistics`?

RUFAS/biophysical/animal/reproduction/reproduction.py
```diff
@@ -1,2 +1,2864 @@
...
+        )
+
+        if reproduction_outputs.animal_type == AnimalType.HEIFER_II:
+            reproduction_outputs = self.heiferII_reproduction_update(
+                reproduction_outputs, time
+            )
+        else:
+            reproduction_outputs = self.cow_reproduction_update(
+                reproduction_outputs, time
+            )
```
Isgotkowitz: Based on this if block, it seems that it will be the `HerdManager`'s responsibility to make sure that only cows and heiferIIs go through this routine. But if this is the case, I think it would be better to split this module up into two: cow reproduction and heiferII reproduction. `HerdManager` would be responsible for sending the heifers into the heifer repro module and cows into the cow repro module.

RUFAS/biophysical/animal/data_types/reproduction_inputs.py
```diff
@@ -0,0 +1,26 @@
+from dataclasses import dataclass
+
+from RUFAS.biophysical.animal.data_types.animal_enums import Breed
+from RUFAS.biophysical.animal.data_types.animal_events import AnimalEvents
+from RUFAS.biophysical.animal.data_types.animal_types import AnimalType
+
+
+@dataclass
+class ReproductionInputs:
```
Isgotkowitz: Instead of having `ReproductionInputs`, what about just passing in the existing properties that hold the attributes to be used in `Reproduction` and putting them directly from their sources into `ReproductionOutputs`? The `ReproductionInputs` is nice because you can pack it into `ReproductionOutputs` with just that `**asdict` call, but I don't think it outweighs the cons to having this extra layer of duplication.

RUFAS/biophysical/animal/reproduction/reproduction.py
```diff
@@ -1,2 +1,2864 @@
...
+
+        return reproduction_outputs
+
+    def heiferII_reproduction_update(self,
+                                     reproduction_outputs: ReproductionOutputs,
```
Isgotkowitz: Could `ReproductionOutputs` be renamed to something else, maybe `ReproductionRelatedAttributes`? It is pretty confusing to see `ReproductionOutputs` as both the input and the output.

Isgotkowitz (commented):
Left some comments more related to the structure and organization of the code - this is gonna be a big step forward for the RuFaS repro code!

One thing that would really help (IMO) would be to put all the cow-related repro code into one module and heifer-related repro code into another. These two modules could be called by `Reproduction` or directly by `HerdManager`, but either way it would make the code a lot more digestible.

RUFAS/biophysical/animal/reproduction/reproduction.py
```diff
@@ -1,2 +1,2769 @@
...
+    delivery, pregnancy checks, estrus detection, and pregnancy management.
+
+    Parameters
+    ----------
+    heifer_reproduction_program : HeiferReproductionProtocol, optional
+        The reproduction program for heifers, by default None.
+    heifer_reproduction_sub_program : HEIFER_REPRODUCTION_SUB_PROTOCOLS, optional
+        The sub-program for heifer reproduction, by default None.
+    cow_reproduction_program : CowReproductionProtocol, optional
+        The reproduction program for cows, by default None.
```
KFosterReed, line 44: I think the definition for the cow reproduction sub protocol is missing

RUFAS/biophysical/animal/reproduction/reproduction.py
```diff
@@ -1,2 +1,2769 @@
...
+
+    body_weight_at_calving: float
+
+    repro_state_manager: ReproStateManager
+
+    def __init__(
+            self,
+            heifer_reproduction_program: HeiferReproductionProtocol = None,
+            heifer_reproduction_sub_program: HEIFER_REPRODUCTION_SUB_PROTOCOLS = None,
+            cow_reproduction_program: CowReproductionProtocol = None,
```
KFosterReed, line 115: I think I am confused about how the cow reproduction protocol is supposed to work... is it still in development since the list is still empty in the repro_protocol_misc.py file?

RUFAS/biophysical/animal/reproduction/reproduction.py
allisterakun: I will look into this!

RUFAS/biophysical/animal/reproduction/reproduction.py
allisterakun: This was originally empty for the `COW_REPRO_PROTOCOLS` dictionary: https://github.com/RuminantFarmSystems/MASM/blob/dev/RUFAS/routines/animal/life_cycle/repro_protocol_misc.py

RUFAS/biophysical/animal/reproduction/reproduction.py
```diff
@@ -1,2 +1,2769 @@
...
+            herd_level_statistics=HerdReproductionStatistics(),
+            newborn_calf_config={}
+        )
+
+        if reproduction_data_stream.animal_type == AnimalType.HEIFER_II:
+            reproduction_data_stream = self.heiferII_reproduction_update(
+                reproduction_data_stream, time
+            )
+        else:
+            reproduction_data_stream = self.cow_reproduction_update(
```
KFosterReed, line 196: I'm probably missing something here but is there something that prevents the AnimalType from being Calf or HeiferI? If not, then maybe we need an ifelse statement instead?

allisterakun: This will be handled in the `Animal` class where only the HeiferII or the Cow and end up here. But I will add an extra layer of safeguard just in case!

RUFAS/biophysical/animal/reproduction/reproduction.py
```diff
@@ -1,2 +1,2769 @@
...
+            reproduction_data_stream.events.add_event(
+                reproduction_data_stream.days_born,
+                time.simulation_day,
+                f"{animal_constants.SETTING_REPRO_PROGRAM_NOTE} from "
+                f"{self.heifer_reproduction_program} "
+                f"to {AnimalConfig.heifer_reproduction_program}",
+            )
+            self.heifer_reproduction_program = HeiferReproductionProtocol(AnimalConfig.heifer_reproduction_program)
+
+        if reproduction_data_stream.days_born >= AnimalConfig.heifer_breed_start_day:
```
KFosterReed, line 240: If I'm following this logic correctly, then it will call the repro protocol updates even if the animal is pregnant. Can we add a check on L 240 to include a check if the animal is not pregnant then move the pregnancy updates outside of this if statement to avoid calling these routines for pregnant animals?

RUFAS/biophysical/animal/reproduction/reproduction.py
```diff
@@ -1,2 +1,2769 @@
...
+            elif self.heifer_reproduction_program == HeiferReproductionProtocol.TAI:
+                reproduction_data_stream = self.execute_heifer_tai_protocol(reproduction_data_stream, time.simulation_day)
+            elif self.heifer_reproduction_program == HeiferReproductionProtocol.SynchED:
+                reproduction_data_stream = self.execute_heifer_synch_ed_protocol(reproduction_data_stream, time.simulation_day)
+            else:
+                raise ValueError(f"Invalid heifer repro program: {self.heifer_reproduction_program}")
+
+            if reproduction_data_stream.days_born == self.ai_day:
+                reproduction_data_stream = self._perform_ai(reproduction_data_stream, time.simulation_day)
+            elif reproduction_data_stream.is_pregnant:
```
KFosterReed, line 252: In combination with the suggestion above, I think moving this elif to the left/at the same level as ` if reproduction_data_stream.days_born >= AnimalConfig.heifer_breed_start_day:` on L 240 then we can avoid calling the repro protocol updates for pregnant animals... I think?

RUFAS/biophysical/animal/reproduction/reproduction.py
KFosterReed: Interesting...

RUFAS/biophysical/animal/reproduction/reproduction.py
```diff
@@ -1,2 +1,2769 @@
...
+                reproduction_data_stream.days_in_pregnancy += 1
+                reproduction_data_stream = self.heifer_pregnancy_update(reproduction_data_stream, time.simulation_day)
+        return reproduction_data_stream
```
KFosterReed, line 255: After reading through the heifer_pregnancy_update, it is not clear to me what triggers the heifer giving birth. I think a good strategy would be to follow more or less the same "order of operations" in both the heifer and cow repro updates. 

So like the cow update, we could start the heifer repro update by checking if they are pregnant. Then if they are pregnant and preg_day = gestation length --> give birth, if not, do the pregnancy updated. From there we can move on to the other repro protocols

KFosterReed: Ok, I understand now, the heifer calving routine doesn't occur inside the reproduction routine - it occurs in animal.py. In this case, I think it might actually make more sense for the cow calving methods to be in animal.py?

KFosterReed: I am actually still confused about how/where heifer calving occurs :/ In animal.py the `transition_heiferIII_to_cow` function changes the animal type to a DRY_COW. This is not correct because a Heifer III should become a cow when she calves and thus would become a lactating cow. Unless, the plan is for her to calve as a cow the following day? Either way, it is not clear how this is supposed to happen

RUFAS/biophysical/animal/reproduction/reproduction.py
jadamchick: It seems that the repro_protocol_misc.py is only for the very specific case of what happens when estrus detection fails to occur for heifer synchED protocols ..?

RUFAS/biophysical/animal/reproduction/reproduction.py
jadamchick: I'm not sure if the code needs cow_reproduction_sub_program (maybe?) but as far as user inputs go, they are specified in terms of Presynch, Ovsynch, Resynch -- i.e. there is no sub_program input for cow repro. FWIW

RUFAS/biophysical/animal/reproduction/reproduction.py
```diff
@@ -1,2 +1,2769 @@
...
+                                                                 ReproductionDataStream, AnimalReproductionStatistics,
+                                                                 HerdReproductionStatistics)
+
+from RUFAS.biophysical.animal.reproduction.hormone_delivery_schedule import HormoneDeliverySchedule
+from RUFAS.biophysical.animal.reproduction.repro_protocol_misc import InternalReproSettings
+from RUFAS.biophysical.animal.reproduction.repro_state_manager import ReproStateManager
+from RUFAS.routines.animal.life_cycle.animal_events import AnimalEvents
+from RUFAS.time import Time
+
+HEIFER_REPRODUCTION_SUB_PROTOCOLS = Union[HeiferTAISubProtocol, HeiferSynchEDSubProtocol]
```
jadamchick, line 28: I see that in repro_protocol_enums.py, the HeiferTAISubProtocol(Enum) already includes all four subprotocols (2 for TAI, and 2 for SynchED). Just curious if that is necessary? it seems redundant to me that it is already grouping the synchED together with the TAI as well as listing SynchED separate and then joining them here.

RUFAS/biophysical/animal/reproduction/reproduction.py
```diff
@@ -1,2 +1,2769 @@
...
+
+        return reproduction_data_stream
+
+    def _perform_ai(
+            self,
+            reproduction_data_stream: ReproductionDataStream,
+            simulation_day: int
+    ) -> ReproductionDataStream:
+        """
+        Perform artificial insemination (AI) on the heifer.
```
jadamchick, line 1168: I would add "Perform AI on the heifer **_or cow_**" to be super clear that the same method is used for both.

allisterakun:
https://whimsical.com/reproduction-NNGKQwaX3Gqnim5zhbnpsn

RUFAS/biophysical/animal/data_types/reproduction_io.py
```diff
@@ -0,0 +1,149 @@
...
+class AnimalReproductionStatistics:
+    """
+    Animal-level reproduction-related statistical properties.
+
+    Attributes
+    ----------
+     ED_days: int
+        The number of days the animal has been in the ED program.
+    estrus_count: int
+        The number of estrus during the ED program.
```
krb78, line 38: Is this referring to the number of times an individual animal is found in estrus during the ED program? If so, could this be made a little more clear - perhaps "the number of estrus events for an individual animals enrolled in the ED program". Alternatively, if it is the total number of estrus events for all animals, this should be clarified as well.

RUFAS/biophysical/animal/data_types/reproduction_io.py
```diff
@@ -0,0 +1,149 @@
...
+    GnRH_injections: int
+        The number of GnRH injections.
```
krb78, line 40: Same comment here. Is it for an individual animal or the total number of GnRH injections over a certain time period?

RUFAS/biophysical/animal/data_types/reproduction_io.py
```diff
@@ -0,0 +1,149 @@
...
+        The number of CIDR injections.
+    semen_number: int
+        number of straws of semen used
+    AI_times: int
+        The number of times that artificial injections are performed.
+    breeding_to_pregnancy_time: int
+        The number of days it takes for the animal to get pregnant from breeding start.
+    pregnancy_diagnoses: int
+        The number of pregnancy diagnoses.
+    calving_to_pregnancy_time: int
```
krb78, line 53: In the dairy world, this is known as "days open" - the number of days a cow has between calving and becoming pregnant.

RUFAS/biophysical/animal/data_types/reproduction_io.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+@dataclass
+class HerdReproductionStatistics:
+    """
+    Herd-level reproduction-related statistical properties.
+
+    Attributes
+    ----------
+    num_ai_performed: int
```
krb78, line 76: Should this name indicate that it is heiferIIs only as the definition suggests?

RUFAS/biophysical/animal/data_types/reproduction_io.py
```diff
@@ -0,0 +1,149 @@
...
+    AI_times: int = 0
+    breeding_to_pregnancy_time: int = 0
+    pregnancy_diagnoses: int = 0
+    calving_to_pregnancy_time: int = 0
+
+
+@dataclass
+class HerdReproductionStatistics:
+    """
+    Herd-level reproduction-related statistical properties.
```
krb78, line 72: Herd indicates across heifers and cows, correct? I just want to make sure that for all of the following we are very clear whether or not the definition is for the herd (cows + heifers) or just for one of the two.

RUFAS/biophysical/animal/data_types/reproduction_io.py
```diff
@@ -0,0 +1,149 @@
...
+    Attributes
+    ----------
+    num_ai_performed: int
+        The number of times AI was performed across all heiferIIs.
+    num_successful_conceptions: int
+        The number of successful conceptions out of all AI performed.
+    num_ai_performed_in_ED: int
+        The number of times AI was performed in the ED protocol.
+    num_successful_conceptions_in_ED: int
+        The number of successful conceptions out of all AI performed in the ED.
```
krb78, line 83: is this duplicative with the input "conception rate"? This is giving a count, but is it meant to purely cross-check that the conception rate inputted resulted in correct outputs?

RUFAS/biophysical/animal/reproduction/hormone_delivery_schedule.py
```diff
@@ -40,47 +41,47 @@ class HormoneDeliverySchedule:
...
             0: {"deliver_hormones": ["GnRH"]},
             7: {"deliver_hormones": ["PGF"]},
             10: {"deliver_hormones": ["GnRH"]},
             16: {"set_presynch_end": True},
         },
-        CowReproProtocolEnum.Presynch_G6G.value: {
+        CowPreSynchSubProtocol.Presynch_G6G.value: {
             0: {"deliver_hormones": ["PGF"]},
             2: {"deliver_hormones": ["GnRH"]},
             8: {"set_presynch_end": True},
```
krb78, line 82: Ok question: can the OvSynch GnRH day 0 occur on the same day as the "set_presynch_end"? if not, this needs to be changed to day 7.

allisterakun:
Closing this PR as the Animal Refresh has been merged by PR #2235.

PR #2039, opened 2024-10-21
[Animal] Unit tests for AnimalNutrients

### LINKED ISSUE #2037: [Animal] Add unit tests for AnimalNutrients class
There is currently no unit testing for the AnimalNutrients class.

Please add a full test-suite for the functions in this module/class:

RUFAS > biophysical > animal > animal_nutrients > animal_phosphorus.py

ew3361zh:
Adds unit testing covering the new functions in AnimalNutrients class.

### Context

Issue(s) closed by this pull request: closes #2037

### What

1. Unit tests are added for all the functions in animal_nutrients.py.
2. All references to AnimalPhosphorus are updated to AnimalNutrients.

### Why

Testing was not added when this new class was originally created and merged into dev.

### How

Added a unit test for each function in animal_nutrients.py.

### Test plan

Running coverage on animal_nutrients.py and ensuring as close 100% coverage as possible.
`get_dry_matter_intake()` currently doesn't do anything except return 0.0. I don't know what that function will eventually do so I have `pass` as the only thing this test is doing right now.


tests/animal_module_tests/nutrient_tests/test_animal_nutrients.py
```diff
@@ -0,0 +1,354 @@
...
+    result = AnimalNutrients.perform_daily_phosphorus_update(mock_animal_status, mock_general_properties)
+
+    mock_get_dmi.assert_called_once()
+    mock_calc_phosphorus_requirements.assert_called_once_with(mock_general_properties, mock_animal_status, 10.0)
+    mock_calc_total_phosphorus.assert_called_once_with(mock_animal_status)
+
+    assert result == mock_animal_status
+
+
+def test_get_dry_matter_intake() -> None:
```
matthew7838: What's the reason for passing this?

tests/animal_module_tests/nutrient_tests/test_animal_nutrients.py
```diff
@@ -0,0 +1,354 @@
...
+    assert mock_phosphorus_status.phosphorus_excess_in_diet == expected_excess_in_diet
+    assert mock_phosphorus_status.phosphorus_reserves == pytest.approx(expected_reserves, 1e-3)
+    assert mock_phosphorus_status.total_phosphorus_in_animal == pytest.approx(expected_total_phosphorus, 1e-3)
+
+
+@pytest.mark.parametrize(
+    "animal_type, dry_matter_intake, expected_loss",
+    [
+        # Case 1: Animal is a calf or heifer (if condition)
+        (AnimalType.CALF, 10.0, 0.0008 * 10.0 * GeneralConstants.KG_TO_GRAMS),
```
matthew7838: More of a suggestion as I'm also not really sure what would count as duplicating logic. However, with the relatively simple calculations here it might be good to just enter the manually calculated expected result?

tests/animal_module_tests/nutrient_tests/test_animal_nutrients.py
```diff
@@ -0,0 +1,354 @@
...
+@pytest.mark.parametrize(
+    "animal_type, endogenous_loss, growth_phosphorus, gestation_phosphorus, milk_phosphorus,"
+    "urine_production_phosphorus, expected_absorbed_phosphorus",
+    [
+        # Case 1: Animal is a dry cow or lactating cow (includes milk phosphorus)
+        (AnimalType.DRY_COW, 0.5, 1.0, 2.0, 3.0, 0.2, 0.2 + 0.5 + 1.0 + 2.0 + 3.0),
+        (AnimalType.LAC_COW, 0.4, 0.9, 1.8, 2.5, 0.1, 0.1 + 0.4 + 0.9 + 1.8 + 2.5),
+        # Case 2: Animal is a heifer (does not include milk phosphorus)
+        (AnimalType.HEIFER_II, 0.3, 0.8, 1.5, 2.0, 0.05, 0.05 + 0.3 + 0.8 + 1.5),
+        (AnimalType.HEIFER_III, 0.2, 0.7, 1.2, 1.5, 0.04, 0.04 + 0.2 + 0.7 + 1.2),
```
matthew7838: Same suggestion for these simple additions.

tests/animal_module_tests/nutrient_tests/test_animal_nutrients.py
```diff
@@ -0,0 +1,354 @@
...
+        (AnimalType.CALF, True),
+        (AnimalType.HEIFER_I, True),
+        (AnimalType.HEIFER_II, True),
+        (AnimalType.HEIFER_III, True),
+        # Cases where the animal is not a calf or heifer
+        (AnimalType.DRY_COW, False),
+        (AnimalType.LAC_COW, False),
+    ],
+)
+def test_is_calf_or_heifer(animal_type: AnimalType, expected_result: bool) -> None:
```
matthew7838: Not sure if it's the scope of this PR, I recall the name of this method will be changed. Maybe we can reflect it here? or wait for the PR that change its name both should work.

tests/animal_module_tests/nutrient_tests/test_animal_nutrients.py
```diff
@@ -0,0 +1,354 @@
...
+    monkeypatch.setattr(AnimalNutrients, "_calculate_absorbed_phosphorus", MagicMock(return_value=absorbed_phosphorus))
+    monkeypatch.setattr(
+        AnimalNutrients, "_calculate_animal_phosphorus_requirement", MagicMock(return_value=absorbed_phosphorus)
+    )
+
+    result = AnimalNutrients._calculate_phosphorus_requirements(
+        mock_general_properties, mock_phosphorus_status, dry_matter_intake
+    )
+
+    assert result.phosphorus_endogenous_loss == endogenous_loss
```
matthew7838: I think this part missed the assertion of method calls. What's the reason to not use MockerFixture here?

matthew7838 (commented):
Great work, Niko.  I have some suggestions for some changes.

tests/animal_module_tests/nutrient_tests/test_animal_nutrients.py
ew3361zh: As noted in the PR description, there is no current functionality in that method (it just returns 0.0) because we're not sure how/where dry matter intake will be calculated. So this is just a placeholder test until that functionality is implemented.

tests/animal_module_tests/nutrient_tests/test_animal_nutrients.py
ew3361zh: Yea I wasn't sure since this seems like such an arbitrary calculation (with magic numbers despite the reference in the docstrings). I feel like keeping the calculation visible in the test case is helpful for understanding what the function is doing and testing it properly (because then if one of those "magic numbers" was changed, it might be more easy to identify why the test suddenly wasn't passing).

I'm up for pushback on this one though because it looks a little cumbersome. If you/others feel strongly enough, we can just replace it with the pre-calculated expected loss.

tests/animal_module_tests/nutrient_tests/test_animal_nutrients.py
ew3361zh: Same reply from above as me.

tests/animal_module_tests/nutrient_tests/test_animal_nutrients.py
ew3361zh: Right, I couldn't recall what the feedback was on that. What was the suggestion on what it should be changed to?

tests/animal_module_tests/nutrient_tests/test_animal_nutrients.py
ew3361zh: This was a nightmare yesterday - I ended up figuring out it was something with these being static functions within the same class where mocking them using MockerFixture was interfering with other test results (it was the next test, `test_calculate_phosphorus_for_growth` that kept returning the same result regardless of the test case inputs). So I found the pytest monkeypatch which is essentially a context manager that prevents inter-test conflicts and mocks. I didn't run into it on any of the other tests but definitely something to watch out for in these new Animal classes where all the methods are static.

tests/animal_module_tests/nutrient_tests/test_animal_nutrients.py
matthew7838: I totally get your point, and I feel the same way. I'm countering some of the same dilemmas on animal digest PR. Maybe we can discuss this a bit in the WT.

tests/animal_module_tests/nutrient_tests/test_animal_nutrients.py
matthew7838: Is there a way to assert method calls with monkeypatch? I amwondering why this issue didn't pop up in other places tho.

tests/animal_module_tests/nutrient_tests/test_animal_nutrients.py
matthew7838: Nope, we should check with Allister or Pooya.

tests/animal_module_tests/nutrient_tests/test_animal_nutrients.py
Isgotkowitz: You guys make some valid points, but I think that ultimately the test suite is more robust if the results are precomputed. We should definitely talk about this in WT!

tests/animal_module_tests/nutrient_tests/test_animal_nutrients.py
Isgotkowitz: I think Pooya's FB was that this method could return true if an animal is a calf or a heifer, or is return true if it is a calf and false if its a heifer - the method name is ambiguous.

A better name would be `is_not_cow` - but this name isn't ideal because it has that negation in there. Maybe an even better solution would be to change the method to be `is_cow`, return true if it is a cow and false if it is a calf or heifer, and modify the if statement to negate `is_cow`

RUFAS/biophysical/animal/animal_nutrients/animal_nutrients.py
```diff
@@ -11,18 +11,18 @@
 HEIFERS_AND_COWS = [AnimalType.DRY_COW, AnimalType.LAC_COW, AnimalType.HEIFER_II, AnimalType.HEIFER_II]
 
 
-class AnimalPhosphorus:
+class AnimalNutrients:
```
Isgotkowitz: Would you consider renaming this class and module to just nutrients, instead of animal nutrients?

Isgotkowitz (commented):
I think this PR looks great Niko! Only commenting for now because I would also like to discuss how to approach the test suite, but once we do that in work together and figure out a game plan I will come back and approve

RUFAS/biophysical/animal/animal_nutrients/animal_nutrients.py
allisterakun: second that

ew3361zh: I would. We do currently have a ManureNutrients class though and I'd just want to be sure we wouldn't see a potential conflict with a future PlantNutrients/SoilNutrients/etcNutrients class

tests/animal_module_tests/nutrient_tests/test_animal_nutrients.py
ew3361zh: I checked and you have to create mock methods to be able to check for mocker method calls. Not sure how to get around this.

ew3361zh:
Updates:
- AnimalNutrients -> Nutrients (along with file and test-names)
- Uncalculated test case values are now pre-calculated
- `_is_calf_or_heifer()` is now `_is_cow()` and associated logic is adjust accordingly (i.e. reversed)

allisterakun (approved):
Great job, Niko!

The nutrients are now fully covered (besides that dummy method):

```
Name                                                           Stmts   Miss  Cover   Missing
--------------------------------------------------------------------------------------------
.\RUFAS\biophysical\animal\nutrients\nutrients.py                 78      1    99%   32
.\tests\animal_module_tests\nutrient_tests\test_nutrients.py     103      0   100%
--------------------------------------------------------------------------------------------
TOTAL                                                            181      1    99%

```

Just a couple of things before merging:
1. Please remove `from __future__ import annotations` in `nutrients.py`
2. Can we remove all the nutrients-related files (`animal_nutrients.py` and `animal_nutrients_status.py`) in the `routines` folder as they are no longer used?

RUFAS/routines/animal/animal_nutrients_daily_update/animal_nutrients.py
Isgotkowitz, line 1: Could you please explain why modules in `routines/` need to be changed, not just the ones in `biophysical`?

tests/animal_module_tests/nutrient_tests/test_nutrients.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+@pytest.mark.parametrize(
+    "phosphorus_intake, phosphorus_requirement, initial_reserves, phosphorus_for_gestation, phosphorus_for_growth,"
+    "expected_excess_in_diet, expected_reserves, expected_total_phosphorus",
+    [
+        # Case 1: phosphorus_intake < phosphorus_requirement (if condition)
+        (5.0, 8.0, -5.0, 2.0, 1.0, 0.0, -8.0, 100.0),
+        # Case 2: phosphorus_intake >= phosphorus_requirement and reserves < 0 (elif condition)
+        (10.0, 8.0, -5.0, 2.0, 1.0, 2.0, 0.7 * 2.0 - 5.0, 104.4),
```
Isgotkowitz:
```suggestion
        (10.0, 8.0, -5.0, 2.0, 1.0, 2.0, -3.6, 104.4),
```

Isgotkowitz (approved):
Had one question about why some files in `routines/` were being changed and one minor suggestion, otherwise LGTM! Nice job Niko

RUFAS/routines/animal/animal_nutrients_daily_update/animal_nutrients.py
ew3361zh: I think Allister is right that the routines versions could just be removed. I think I initially changed both to be consistent in case there was some other reason we were keeping both files around.

Issue #2016, opened 2024-09-20
[Animal] Coordinate Barn Temp Function with Animal Module


ew3361zh:
There is a function (`determine_barn_air_temperature()`) that determines the estimated barn temp when it's sent an ambient air temp. This function is in GasEmissionsCalculator because it's used for both manure handlers and storages. This function will be needed by the animal module for animal requirements and the animal health sub-module.

## Expected Behavior

There should only be one function to estimate barn temperature from air temperature. The Animal module will need access to this function so it should be somewhere that makes sense for all its current uses (and the future uses we know about).

## Current Behavior

The function resides in GasEmissionsCalculator as it's only used currently by 2 processes in the manure module - handlers and storages.

## Possible Solution

Find a location for this function that suits all the places it's needed.

matthew7838:
Grooming session: To be discussed during Nov4 meeting.

allisterakun:
Grooming session: @allisterakun will bring this up at the next animal module meeting on 11/26.

matthew7838:
Grooming session: @allisterakun will bring this up at the next animal module meeting on 12/24 or next year.

allisterakun:
Note: The status is the same as of 5/15/25.
I have brought this up at the animal module meeting:
1. This issue has a wishful priority.
2. There is no estimate on the t-shirt size for this issue because we still need more SME input. Two parts of the animal module could potentially need this barn temp function:
  1. Animal energy requirement calculation:
    1. We already have some temperature estimation logic for animal energy requirement calculation
    2. This function is only for estimating the barn temperature. If we were to use this function to estimate the barn temperature better, we would need some further logic to handle the temperature for other structure types (e.g., open lot).
  3. Animal Health
    1. Animal health is still in the early phase of planning

elle-andreen:
7/7/2025: This function now exists in the processor base class in the manure module post manure refresh. It still should likely be moved to a more widely accessible location as described in the original issue.

elle-andreen:
SME grooming 3/27/26: This issue is still wishful/low priority. See the comment from Allister above on where the barn temp function might integrate well in the Animal module. Though opportunities exist for integration, the effect of barn temperature on animal requirements is expected to be relatively minimal so this issue remains low priority. However, this issue is worth keeping as a reminder that if something like heat stress, ventilation, etc. were implemented in the Animal module, the existing barn temp function or a singular, new function should be used, rather than using different logic for the same concept.

Issue #1337, opened 2024-02-26
Unable to specify feed milk replacer (203) rather than whole milk (202) in calf ration 


jadamchick:
## Expected Behavior
If a farm uses milk replacer for calves, expect to be able to specify feed 203 for a calf ration ingredient

## Current Behavior
The simulation terminates early with the following error: 
[...]

PR #2043, opened 2024-10-23
Calf ration implementation

### LINKED ISSUE #1137: Refactoring Calf Ration

### LINKED ISSUE #2517: [Animal][Ration] Replace Dummy Values for Calf Calcium and Phosphorus Requirement Calculation
Address the TODO in `biophysical/animal/animal.py` in line 2248:
```python
            # TODO: do not use dummy values for calf calcium and phosphorus requirements.
```

JoeWaddell:
### Context

Issue(s) closed by this pull request: more completely closes https://github.com/RuminantFarmSystems/MASM/issues/1137 
and https://github.com/RuminantFarmSystems/MASM/issues/1337 , also should close https://github.com/RuminantFarmSystems/RuFaS/issues/2517 and https://github.com/RuminantFarmSystems/MASM/issues/2016 for calves

### What

Adds logic to re-implement previously developed calf ration formulation and requirement calculations. 

This will work for both the default user defined ration behavior, and automated "modes", but note that neither utilize the minimize method, like we do for all other animal classes. A) The user defined method gets the pen's average dry matter intake requirement and then applies the percentages in the feed input file. B) The automated version will use the ration as formulated by the equations now housed in calc_intake in  CalfRationManager

### Why

Calves were backburnered for a while, and had been restricted to using only precisely two inputs, and would crash if users tried selecting other feed IDs.

### More context and associated problems
This will require thought around what to do with the "milk_type" input in the animal input JSON, which could be deprecated (see comments by Kristan and Joe)

Default scenario values give only a rough total of 1kg per animal in the calf pen, whereas the diet we hardcoded on main is 3kg. This needs evaluation as to whether A) I've made a mistake in this translation of the old methods to current inputs, or if we're B) intentionally 3X overfeeding by default, or C) the methods themselves need refinement.

Related: given that these are process based methods independent of NASEM/NRC, consideration should be given to how the user may want to handle calf ration formulation independent of the user-defined ration formulation for other animal types, e.g. give some pre-forumlated ration for calves, but use automated formulation methods for all other groups. This level of granularity may not be necessary, but should get input from other SMEs involved in calf nutrition.

### How

Iterates over all calves in a calf pen, calculates their intake, then sums those and updates pen attribute for `ration_per_animal` to reflect total divided by the number of animals in the pen.

### Test plan

Currently lacks unit tests for new methods.

Needs more testing with multiple, more complex calf rations.

See "More context and associated problems" and section, as this needs evaluation to see if we're underfeeding with these new methods, and how temperature will/can affect the equations used here.

You can see the basic output using the following graph filter:
```
{
      "type": "plot",
      "filters": [
        ".*ration_per_animal_for_pen_0_CALF"
      ],
      "title": "Ration per animal for pen 0",
      "variables": [".*"],
      "omit_legend_prefix": true
}
```


tomhuhh (commented):
Good job Joe! I did not spot major issues with the code. I noticed that in the calculation of the intake of starters, the original code did not consider the dry matter content of the starter while it considered the dry matter content of whole milk and milk replacer. Also, I could not find the scientific documentation for calf ration calculations on Basecamp so I cannot verify the equation.  A quick search on Google with these key words and the parameters also did not result in the information I was looking for. I would like to learn what your and others' thoughts are on whether or not to consider the dry matter content of starters. Thanks!

tomhuhh:
Adding a screenshot of the code that I was referring to: 

Screenshot 2024-11-11 at 9 24 59 AM

tomhuhh:
Also, I would like to document that with the current version and the default input files, if I set the user defined ration to be true, the resulted daily dry matter intake per day is 84.4% to 102.2% of the previously fixed dry matter intake estimates (6 kg/head/d).

RUFAS/routines/animal/animal_manager.py
```diff
@@ -1118,8 +1118,34 @@ def _handle_pen_ration(self, feed: Feed, pen: Pen) -> None:
 
         while "status" not in ration_per_animal or ration_per_animal["status"].lower() != "optimal":
             if pen.animal_combination == AnimalCombination.CALF:
-                ration_per_animal = CalfRationManager.optimize()
-                ration_vals = {"ME_total": 0}
+                individual_calf_rations = []
+                wean_day = AnimalBase.config["wean_day"]
+                wean_length = AnimalBase.config["wean_length"]
+                if 202 in pen_specific_feed_data["feed_id"]:
```
KFosterReed: Does this If-else statement eliminate the need for the user input for `"milk_type"`? If so, can we delete it while keeping the model backwards compatible for input files that have this input? Or, maybe we just make an issue about user input adjustments that need to be made?

JoeWaddell: Correct! The intent here was that it could streamline the feed inputs a little bit...with the user simply selecting the feed IDs for calves as they do for the other animal classes, since it's currently a little redundant.

We could still keep this input, but we'd have to think through the logic of when (or if) it should ever override the user defined feeds.

If we do deprecate it: to maintain backwards compatibility (especially with FARM ES) what I thought is that we could keep the option in the metadata, and simply not let it get to the codebase itself. I think we could/should  capture it and flag a warning that the input has been deprecated, and the user should specify the feeds in the feed input JSON (we may want to think about a process for any input field deprecation we do in the future).

KFosterReed:
Here is the [link to the calf ration documentation](https://docs.google.com/document/d/1r1cCl9qiYQ3vE-yk_tbTN9oYFVxYFVVH/edit?usp=sharing&ouid=106545401981506568145&rtpof=true&sd=true)

KFosterReed:
After a little more testing this is looking good! The multiple replacer/starter options work as expected but note that to test it I made a dummy feed library where I just duplicated the existing milk replacer and starter options. 

When I tested the UDR method, it was weird that the average DMI/calf/day was so much larger than when using a non user defined option. A little debugging showed a small typo that when fixed has the DMI for UDR equal to the non-UDR methods. 

Here is a link with some of the scenarios I tested: [Calf Ration Scenarios](https://docs.google.com/spreadsheets/d/1-ZjdbwYehgPRmOv6_kAx1ux026QM2gAZZU4GzVTd-bg/edit?usp=sharing)

I am going to write up a few issues related to reducing the hard coded feed ids and improving the calf diet formulation in general but I think this is good to go as it is!

RUFAS/biophysical/animal/ration/ration_manager.py
```diff
@@ -388,8 +388,8 @@ def resolve_target_dmi(cls, animal_combination: AnimalCombination, pen: Any) ->
         option = cls.get_intake_option(animal_combination)
 
         if option is IntakeOption.PREDICT_DMI:
-            if animal_combination is AnimalCombination.CALF:
-                return float(cls.CALF_DRY_MATTER_INTAKE)
+            # if animal_combination is AnimalCombination.CALF:
+            #     return float(cls.CALF_DRY_MATTER_INTAKE)
             return float(pen.average_nutrition_requirements.dry_matter)
```
tomhuhh, line 391: These two lines are commented out here but existing tests still assert it exists (tests/.../test_ration_manager.py:426-432). Maybe worth checking if this is intentional. Thanks!

RUFAS/biophysical/animal/ration/calf_ration_manager.py
tomhuhh, line 187: Here seems to be a mismatch with the newly implemented logic where it follows for more options for milk replacer (203 to 207) and start grains (213 to 218). 

> lines 9-11: constants define only WHOLE_MILK_ID = 202, MILK_REPLACER_ID = 203, STARTER_ID = 216.
> lines 184-186: calculate_intake() looks up only whole milk 202, replacer 203, and starter 216.

It might be worth to clean up to avoid conflicts.

tomhuhh (commented):
Thanks Joe for taking the initiative on this effort! I left a couple comments. Thanks!

PR #2045, opened 2024-10-28
[Documentation] Update docs

Isgotkowitz:
Updates the documentation generated by Sphinx so that we can access to it.

Note to reviewers: it will probably be really inefficient and ineffective to review this PR by looking at the files changed. Instead it is more effective to pull the branch locally, and click around the finished html documentation in your browser.

### Context

Issue(s) closed by this pull request: closes # - it doesn't close any 😭

### What

Regenerates the HTML documentation with Sphinx.

### Why

It would just be nice to have updated HTML docs to look at for referencing Feed Storage stuff, and encouraging people to think about what their ideal scientific documentation looks like.

### How

Ran the sphinx action - [Read all about it here, folks!](https://github.com/RuminantFarmSystems/MASM/wiki/Using-Sphinx)

### Test plan

Navigated to the finished HTML documentation in my browser and looked at it to make sure that everything looked about right. If you're unsure about how to do this, please let me know and I'll be happy to walk you through it.


PooyaHekmati (approved):

ew3361zh (approved):
LGTM - checked out a random small sampling of the newly generated HTML files and seemed very normal. Thanks, Ed!

PR #2046, opened 2024-10-28
[DCA] Put DCA in RuFaS

### LINKED ISSUE #682: Identify discrepancies between Crop and Soil Meta data and DCA inputs

### LINKED ISSUE #968: keeping metadata and DCA schema in sync

### LINKED ISSUE #1171: [DCA] Incorporate Data Collection App into RuFaS
The `SchemaGenerator` (initially implemented in #1022) is a first step in automating the process of keeping the metadata properties and the Data Collection App in sync. There are a number of improvements and features that should be made to `SchemaGenerator` first before this process can be considered fully automated. This issue includes those improvements and features as tasks.

- [x] **Not implemented - Unnecessary.** Add a command line option so that the metadata file being read from can be set by the user.
- [x] **Not implemented - Unnecessary.** Add a command line option so that location where schemas are written to can be set by the user.
- [ ] Modify the SchemaGenerator methods so that the grid columns of an input can be set in metadata properties.
- [ ] Modify the SchemaGenerator methods so that a `"title"` key could be included in the metadata properties and used in the input schema. This will go a long way towards improving the usability of the auto-generated schemas in the DCA.
- [x] Modify the way schema files are written so that they are `.js` files instead of `.json`, and they set a variable with the schema contained in a variable. This will allow autogenerated schemas to be used as drop-in replacements for the schema files in the DCA (no copy-pasting required).
- [x] Change the `@staticmethods` in the SchemaGenerator to instance methods, and modify how the input data types are mapped to schema generation methods.

Isgotkowitz:
Integrates the Data Collection App into the RuFaS repo.

_Reviewers_: I recommend only reviewing the `.py` files in GitHub, and using your local git client and browser to try out the Data Collection App. If you're unsure of how to do this, please let me know and I can help you.

### Context

Issue(s) closed by this pull request: closes #1171, closes #968, closes #682

### What

This PR establishes the Data Collection as part of the RuFaS repo, by putting it in its own directory (`DataCollectionApp/`) at the same level as `RUFAS/`.

A task type is added to `TaskManager` which will update the Data Collection App so that any input files generated it will be compatible with the metadata properties.

The module `DataCollectionAppUpdater` is added to `RUFAS/`, and handles the updating of the Data Collection App when called by `TaskManager`.

### Why

Keeping the Data Collection App close to RuFaS makes them much easier to integrate, and by integrating so that updates can be made easily and conveniently we set ourselves up to make the Data Collection App a much more useful tool in the RuFaS ecosystem.

### How

The `DataCollectionAppUpdater` provides a one-stop shop for updating the DCA. When its main routine is called, it grabs the metadata properties from the `InputManager` and traverses them to create a dictionary. That dictionary is translated into JSON and then inserted into a Javascript file which is written into the DCA's `schema` directory.

The `DataCollectionAppUpdater` also rewrites the import scripts and list of available schema in DCA's `index.html` so that the changes in schema will be usable in the DCA. 

The original `index.html` has been modified so that it will save the data from different schemas as individual JSON files -- translation: if you enter Feed data, Animal data, and Config data in the DCA you will get a file for each. `DataCollectionAppUpdater` adds an extra field to each schema for collecting the filename, so that when each file is saved it will be with a name that the data enterer added.

### Test plan

Wrote a unit test suite for `DataCollectionAppUpdater` and the new method in `TaskManager`.

In addition to updating and expanding the unit test suite, I ran the task dedicated to updating the DataCollectionApp then checked the result of those updates - meaning I opened up the DCA and played around with it to make sure it worked right. I highly encourage reviewers to do the same. The default metadata has been modified so that you can run `python main.py` and it will update the DCA for you (I will change this back before merging).


input/metadata/task_manager_metadata.json
Isgotkowitz, line 1: I will change this back before merging.

PooyaHekmati:
Do we need to have those document files?

Isgotkowitz:
> Do we need to have those document files?

Great question. I think we only _need_ the user guide, because it is the only PDF that is actually accessible through the DCA interface (as far as I can tell). The others might be helpful though, so maybe we should keep them and follow up on them in a separate PR?

PooyaHekmati:
> > Do we need to have those document files?
> 
> Great question. I think we only _need_ the user guide, because it is the only PDF that is actually accessible through the DCA interface (as far as I can tell). The others might be helpful though, so maybe we should keep them and follow up on them in a separate PR?

I don't see why we need to introduce files to the codebase whose purpose is not known.

Isgotkowitz:
> > > Do we need to have those document files?
> > 
> > 
> > Great question. I think we only _need_ the user guide, because it is the only PDF that is actually accessible through the DCA interface (as far as I can tell). The others might be helpful though, so maybe we should keep them and follow up on them in a separate PR?
> 
> I don't see why we need to introduce files to the codebase whose purpose is not known.

Ok, they're out of there.

KFosterReed:
I was able to produce the Data Collection App using this PR in the RuFaS repo so that's great! However, the intro page to the different data collection pages doesn't seem to be behaving properly for me. I opened it on Chrome and Microsoft Edge and the "Data Entry" just shows up as text and not a button that can be used to start data entry: 

image


I will try on my personal computer as well to see if it is a USDA thing
** Update: Also does not work on my personal computer

Isgotkowitz:
> I was able to produce the Data Collection App using this PR in the RuFaS repo so that's great! However, the intro page to the different data collection pages doesn't seem to be behaving properly for me. I opened it on Chrome and Microsoft Edge and the "Data Entry" just shows up as text and not a button that can be used to start data entry:
> 
> image
> 
> I will try on my personal computer as well to see if it is a USDA thing ** Update: Also does not work on my personal computer

Crud, I will see if I can reproduce this on my computer.

Did you try running `python main.py` on this branch and seeing if anything changed in the DCA?

KFosterReed:
> > I was able to produce the Data Collection App using this PR in the RuFaS repo so that's great! However, the intro page to the different data collection pages doesn't seem to be behaving properly for me. I opened it on Chrome and Microsoft Edge and the "Data Entry" just shows up as text and not a button that can be used to start data entry:
> > image
> > I will try on my personal computer as well to see if it is a USDA thing ** Update: Also does not work on my personal computer
> 
> Crud, I will see if I can reproduce this on my computer.
> 
> Did you try running `python main.py` on this branch and seeing if anything changed in the DCA?

I did! That is how I got the HTML file that produced the image pasted above :/

Isgotkowitz:
> > > I was able to produce the Data Collection App using this PR in the RuFaS repo so that's great! However, the intro page to the different data collection pages doesn't seem to be behaving properly for me. I opened it on Chrome and Microsoft Edge and the "Data Entry" just shows up as text and not a button that can be used to start data entry:
> > > image
> > > I will try on my personal computer as well to see if it is a USDA thing ** Update: Also does not work on my personal computer
> > 
> > 
> > Crud, I will see if I can reproduce this on my computer.
> > Did you try running `python main.py` on this branch and seeing if anything changed in the DCA?
> 
> I did! That is how I got the HTML file that produced the image pasted above :/

Good news: I reproduced the issue. Bad news: its a Windows thing 😭

Isgotkowitz:
@KFosterReed should be working for you now

KFosterReed:
So far I just looked at the Animal Module page but it looks great! The save and validate functionalities work well and the layout is pretty easy to follow (ignoring some improvements we should make on the metadata properties for the Animal Module). 

I would like to look through the rest of the input files as well as try to run a simulation with the json files produced but so far so good!

Isgotkowitz:
> So far I just looked at the Animal Module page but it looks great! The save and validate functionalities work well and the layout is pretty easy to follow (ignoring some improvements we should make on the metadata properties for the Animal Module).
> 
> I would like to look through the rest of the input files as well as try to run a simulation with the json files produced but so far so good!

Strangely enough, it doesn't actually produce json files that can be fed right into RuFaS. It produces .txt files which contains a big list of all the inputs, which I believe was the same behavior as the old DCA.

I'm not fully sure what the vision for the DCA is, but I assume it would be better if it turned each individual entry into a standalone JSON

elle-andreen:
- I noticed some inputs are excluded from DCA (random and set seed in configs, for example). Is this intentional and where/how is it dictated which metadata variables are and are not included in the DCA schema? 
- I updated some descriptions in the metadata properties and they were correctly assimilated into the corresponding DCA descriptions 👍 
- I'm not sure if this is possible, but if a variable has no default supplied (e.g., the parity fraction variables) in the metadata properties, I think it would be helpful if the input box in the DCA was blank instead of being 0. I think it helps prevent situations where, if the user doesn't overwrite it, a 0 value will cause a failure or will be overwritten by another value unknown to the user. Maybe this falls under a future DCA improvement PR and this this PR. 
- I would be interested to get @KFosterReed's opinion but I think the PDF documents that were previously linked should still remain in the repo here. They are resources with a specific function to help users understand and translate inputs into the DCA. If they're not kept here in the repo and we still want to use them, they would have to be hosted elsewhere (Basecamp? Box?) and then linked somehow in the DCA. I think this is asking for accessibility issues and wrong link problems. But, maybe we want all DCA documentation elsewhere with the scientific/technical documentation.

The automatic updating of the content based on the metadata is super helpful and will come in handy. Some variables in the metadata need defaults and descriptions but that is a problem for another PR. The DCA itself also needs work to be easily usable but I think the specific goals of this PR have been accomplished!

DataCollectionApp/index.html
```diff
@@ -0,0 +1,219 @@
...
+
+                    saveJsonFile(jsonData, fileName, 'application/json');
+                });
+                alert(`${data.length} JSON files have been saved.`);
+            } else {
+                var jsonData = JSON.stringify(data, null, 2);
+                saveJsonFile(jsonData, 'data.json', 'application/json');
+                alert("Cannot parse data as an array. Saving all data in a single file instead.");
+            }
+        });
```
Isgotkowitz, line 204: This function differs from the old DCA. By parsing the data as an array, it is able to separate out the data from different input schemas and save them as individual JSONs

Isgotkowitz:
> * I noticed some inputs are excluded from DCA (random and set seed in configs, for example). Is this intentional and where/how is it dictated which metadata variables are and are not included in the DCA schema?

Right now, the DCA only works with the lowest level input files. You can think of the inputs sort of like a stack: at the top is the Task Manager metadata, then the next level is the tasks, then the regular metadata file, then the base input files (Animal, Feed, Field, etc.).

Right now this is intentional, because the metadatas and tasks files rely heavily on having correct file paths. I think a good next step would be to tweak some of the input schemas so that we can generate that lower level metadata automatically for the user, then allow them to write the tasks they want and link it to the metadata that was generated.

> * I updated some descriptions in the metadata properties and they were correctly assimilated into the corresponding DCA descriptions 👍

:grin:

> * I'm not sure if this is possible, but if a variable has no default supplied (e.g., the parity fraction variables) in the metadata properties, I think it would be helpful if the input box in the DCA was blank instead of being 0. I think it helps prevent situations where, if the user doesn't overwrite it, a 0 value will cause a failure or will be overwritten by another value unknown to the user. Maybe this falls under a future DCA improvement PR and this this PR.

I think this should be a future DCA improvement. It is definitely possible, but it would be good to make that change in a more isolated way so that we know exactly which inputs it effects and none of them are screwed up.

Isgotkowitz:
@KFosterReed and @elle-andreen this is ready for a re-review. The main thing that has changed is when the inputs are downloaded, they are individual JSON files. Also there is an input field that requires users to give a name which is used to name the corresponding JSON file. As Elle pointed out, there is still a lot of work to be done but I think it would be good to write up separate issues for each.

Isgotkowitz:
List of improvements to make after this PR:
- Add feature that automatically generates metadata containing all the input files.
  - Restructuring Field and all related inputs will make this feasible.
- Add Tasks as kind of input file that DCA can be used for.
- ~Don't automatically add 0's in input fields that don't have defaults.~
- ~Make null a valid input type for nullable inputs.~
- Improve messages when inputs are invalid
- Write Wiki for using the DCA
- Add boolean field to each property called "data_collection_app_compatible". When iterating through properties, use that field to determine whether or not to write a DCA schema for it, not the hard-coded list currently in `DataCollectionAppUpdater`
- Allow users to select from drop down of feed names instead of typing feed numbers.

RUFAS/data_collection_app_updater.py
```diff
@@ -0,0 +1,421 @@
...
+    "config_properties",
+    "crop_schedule_properties",
+    "feed_properties",
+    "fertilizer_schedule_properties",
+    "field_properties",
+    "manure_management_properties",
+    "manure_schedule_properties",
+    "tillage_schedule_properties",
+    "tractor_dataset_properties",
+]
```
PooyaHekmati, line 34: Why is this hardcoded? It has to come from an actual metadata file that IM has read and processed.

RUFAS/data_collection_app_updater.py
```diff
@@ -0,0 +1,421 @@
...
+        self._om.add_log("Schema generation starting", "Creating new schemas from metadata properties.", info_map)
+
+        Utility.empty_dir(SCHEMA_DIRECTORY_PATH)
+
+        properties: dict[str, Any] = self._im.meta_data["properties"]
+
+        schema_paths = []
+        for key in properties.keys():
+            if key not in PROPERTIES_TO_CREATE_SCHEMA_FOR:
+                continue
```
PooyaHekmati, line 106: Why?

RUFAS/data_collection_app_updater.py
```diff
@@ -0,0 +1,421 @@
...
+    "fertilizer_schedule_properties",
+    "field_properties",
+    "manure_management_properties",
+    "manure_schedule_properties",
+    "tillage_schedule_properties",
+    "tractor_dataset_properties",
+]
+
+"""Placeholder for inserting schema import scripts in index.html."""
+SCHEMA_SCRIPT_TAG_PLACEHOLDER: str = "    <!-- Spot where schema import scripts go -->"
```
PooyaHekmati:
```suggestion
SCHEMA_SCRIPT_TAG_PLACEHOLDER: str = "    <!-- Schema imports go here-->"
```

DataCollectionApp/template
```diff
@@ -0,0 +1,210 @@
...
+<html lang="en">
+
+<head>
+    <meta charset="utf-8" />
+    <title>RuFaS Data Collection</title>
+    <script src="./js/jsoneditor.min.js"></script>
+    <script src="./js/ace.js"></script>
+    <script src="./js/assign.js"></script>
+    <script src="./js/lz-string.min.js"></script>
+    <!-- Spot where schema import scripts go -->
```
PooyaHekmati:
```suggestion
    <!-- Schema imports go here-->
```

RUFAS/data_collection_app_updater.py
```diff
@@ -0,0 +1,434 @@
...
+        if maximum is not None:
+            schema["maximum"] = maximum
+        if default is not None:
+            schema["default"] = default
+        if description is not None:
+            schema["options"]["infoText"] = description
+        if nullable:
+            schema["type"] = ["number", "null"]
+        else:
+            schema["type"] = "number"
```
PooyaHekmati:
```suggestion
        schema["minimum"] = minimum or "null"
        schema["maximum"] = maximum or "null"
        schema["default"] = default or "null"
        schema["options"]["inputAttributes"]["placeholder"] = default or minimum or maximum or "null"
        schema["options"]["infoText"] = description or "null"
        if nullable:
            schema["type"] = ["number", "null"]
        else:
            schema["type"] = "number"
```

RUFAS/data_collection_app_updater.py
```diff
@@ -0,0 +1,434 @@
...
+        nullable = input_properties.get("nullable")
+
+        if default is not None:
+            schema["default"] = default
+        if description is not None:
+            schema["options"]["infoText"] = description
+        if nullable:
+            schema["type"] = ["boolean", "null"]
+        else:
+            schema["type"] = "boolean"
```
PooyaHekmati: same as bove

RUFAS/data_collection_app_updater.py
```diff
@@ -0,0 +1,434 @@
...
+        default = input_properties.get("default")
+        pattern = input_properties.get("pattern")
+        description = input_properties.get("description")
+        nullable = input_properties.get("nullable")
+
+        if default is not None:
+            schema["default"] = default
+        if description is not None:
+            schema["options"]["infoText"] = description
+        if pattern is not None:
```
PooyaHekmati: same as above

RUFAS/data_collection_app_updater.py
```diff
@@ -0,0 +1,434 @@
...
+        When a string input is taken, often it is to select from a preset group of options. This method is designed to
+        derive those options from the Regex pattern that is used to validate it.
+
+        """
+        pattern = "\\^\\(.*\\)\\$"
+        is_valid_pattern = bool(re.match(pattern, input_pattern))
+        if not is_valid_pattern:
+            raise ValueError(f"'{input_pattern}' is not a valid pattern. Cannot create list of valid options.")
+
+        unsplit_list = input_pattern[2:-2]
```
PooyaHekmati: why?

RUFAS/data_collection_app_updater.py
```diff
@@ -0,0 +1,434 @@
...
+                schema["pattern"] = pattern
+        if nullable:
+            schema["type"] = ["string", "null"]
+        else:
+            schema["type"] = "string"
+
+        return schema
+
+    @staticmethod
+    def _get_list_of_options(input_pattern: str) -> list[str]:
```
PooyaHekmati: this method is cryptic, maybe add examples to docstrings.
Also, why is it static?

PooyaHekmati (changes requested):

PooyaHekmati:
> List of improvements to make after this PR:
> 
> * Add feature that automatically generates metadata containing all the input files.

The DCA schema is supposed to be generated from a metadata; how can the output not match the input metadata?

> * Add Tasks as kind of input file that DCA can be used for.

I don't understand this sentence.

> * Don't automatically add 0's in input fields that don't have defaults.

My comments on the PR should resolve this.

> * ~Make null a valid input type for nullable inputs.~
> * Improve messages when inputs are invalid

RUFAS/data_collection_app_updater.py
Isgotkowitz: Can you please be more specific

PooyaHekmati: Why is `PROPERTIES_TO_CREATE_SCHEMA_FOR` needed in the first place?

RUFAS/data_collection_app_updater.py
Isgotkowitz: The pattern ensures that the string starts with "^(" and ends with ")$". This snippet removes the "^(" and ")$" before the string is split into substrings which are used to create the dropdown in the DCA

RUFAS/data_collection_app_updater.py
Isgotkowitz: The properties are pulled from the IM, which have been read and processed. As the properties are iterated over, this list is used to decided whether a DCA schema needs to be updated for that property.

This makes sense for a lot of properties, because the inputs associated with them contain so much data that it isn't helpful or feasible to collect them with the DCA. The properties that schema are not generated for are:
- animal_population_properties
- animal_net_merit_properties
- animal_top_listing_semen_properties
- lactation_properties
- economic_properties
- emissions_properties
- feed_emissions_properties
- user_feeds_properties
- NRC_Comp_properties
- NASEM_Comp_properties
- weather_properties
- EEE_constants_properties
- end_to_end_test_result_path_properties
- end_to_end_testing_input_properties

RUFAS/data_collection_app_updater.py
Isgotkowitz: Examples added. It is static because the logic of the method is entirely self-contained, it doesn't need any references to self and it doesn't make any other parts of the code cleaner to have it be an instance method

RUFAS/data_collection_app_updater.py
PooyaHekmati: I think this explanation merits a comment in the code to explain why 2 and -2 are there.

RUFAS/data_collection_app_updater.py
PooyaHekmati: The method is inconsistent with the rest of the methods. In the absence of good reasons to decide on the static-ness of a method, consistency is a good reason to rely on. Therefore, I don't think this needs to be static.

RUFAS/data_collection_app_updater.py
PooyaHekmati: This approach makes it harder to maintain. A better solution is to have metadata that only contains inputs that can be collected by DCA. Alternatively, you can add a new field to the metadata itself to indicate should a properties be added to DCA or not. This approach is not only more explicit and maintainable, it also facilitates the generation of metadata that is functional (the data either comes directly from the IM or is populated as user input.)

Isgotkowitz: > Alternatively, you can add a new field to the metadata itself to indicate should a properties be added to DCA or not.

I like this idea -- my first thought for field name is `data_collection_app_compatible`. For this approach though, would it be ok to implement it in a follow-up PR? To prevent the new field name from making the metadata illegal, `DataValidator` has to be modified, and given the size of this PR already I think it would be better to make that change separately

PooyaHekmati: Sure.
Also, remove the generated schemas from this PR before merging and add the schema directory to git ignore.

RUFAS/data_collection_app_updater.py
Isgotkowitz: Roger that, a comment has been added

RUFAS/data_collection_app_updater.py
Isgotkowitz: See comment above -- but in a nutshell, it is to prevent the DCA from getting schemas for inputs that aren't DCA appropriate. This method will be rewritten so that it doesn't rely on a hardcoded list though

RUFAS/data_collection_app_updater.py
Isgotkowitz: I did not implement this exact suggestion, but I did change the code in what I think is the spirit of this suggestion.

Min and max can't have "null" as their values, it screws up how inputs are validated. For the infoText, it can just be whatever the "description" is, JSON-Editor handles it well.

Instead of putting in the default, min, or max as the placeholder, I made them empty strings. While I was working with this I found that the grey placeholder values were hard to distinguish visually from blue, actual inputs, and that a blank box was more striking. I'm happy with the blank box, but an even better solution might be to have a string that says something like "Add numerical input"

Isgotkowitz:
> > List of improvements to make after this PR:
> > 
> > * Add feature that automatically generates metadata containing all the input files.
> 
> The DCA schema is supposed to be generated from a metadata; how can the output not match the input metadata?

Each schema corresponds to a single input file. The metadata is a single file with paths to a bunch of different input files. When a user puts information into the DCA and hits "Save Result", they get a bunch of individual input files, but no metadata with references to all those individual input files.

> > * Add Tasks as kind of input file that DCA can be used for.
> 
> I don't understand this sentence.

One of the input files contains the tasks which are given to TaskManager. There is no DCA schema for collecting that type of input in this PR

RUFAS/data_collection_app_updater.py
PooyaHekmati: I appreciate your feedback about the placeholders and your efforts to improve the user experience. That said, the decision to use the placeholders as they were originally specified was intentional, as empty strings defeat the purpose of having placeholders. Additionally, the text at the top of the DCA page already explains the meaning of blue and grey colors, ensuring users can differentiate between them.

Please proceed with implementing the placeholders as instructed. If there’s a need to revisit this feature in the future, we can address it as a team. Also, as part of this task, please add placeholders to other data types as well.

PooyaHekmati:
> > > List of improvements to make after this PR:
> > > 
> > > * Add feature that automatically generates metadata containing all the input files.
> > 
> > 
> > The DCA schema is supposed to be generated from a metadata; how can the output not match the input metadata?
> 
> Each schema corresponds to a single input file. The metadata is a single file with paths to a bunch of different input files. When a user puts information into the DCA and hits "Save Result", they get a bunch of individual input files, but no metadata with references to all those individual input files.

Normally, I would insist that this needs to be done in this PR as it was part of the plan. However, knowing that you are going to have another PR to remove that hardcoded list of properties, I think it makes more sense to implement this there.

elle-andreen:
- There are some metadata-to-DCA translation issues with the manure separator and treatment Name inputs (they're appearing as a drop-down instead of a text box). I did not see any other obvious instances of this in other schema though.
- For Field Properties, just to confirm, are field properties supposed to be provided _per field_ or just one set of properties that apply to your whole farm? I ask because there is no "Add new element" type option in that page, but I'm guessing that users do not need more than one set of Field Properties inputs
- There are still variable names that are being translated from the metadata to the DCA correctly, but don't make sense in the DCA context, e.g. under Manure Manage Scenarios, the variable "Bedding Type" would be easier to understand if it was named "Bedding _Name_", but "Bedding Type" is in fact its current name. It's not the end of the world but I'm curious to know whether we will address these types of things by changing the variable name in the metadata, or by changing how it appears in the DCA. I don't need an answer now, just something to consider for future DCA improvements!
- I really like the 'file name' feature, and the JSONS saved and ran properly for me 😁 

One improvement to add to the list, if desired: for feeds, I think I recall the old DCA was able to reference the feed library via a dropdown selection, so users could choose the feed by name, not by number. Either way, that would be great to have. 

Overall, this looks great to me and I will approve notwithstanding any further changes from my or others' comments!

Isgotkowitz:
> * There are some metadata-to-DCA translation issues with the manure separator and treatment Name inputs (they're appearing as a drop-down instead of a text box). I did not see any other obvious instances of this in other schema though.

Nice catch, I'll find a fix for this.

> * For Field Properties, just to confirm, are field properties supposed to be provided _per field_ or just one set of properties that apply to your whole farm? I ask because there is no "Add new element" type option in that page, but I'm guessing that users do not need more than one set of Field Properties inputs

Yes, the field properties require field inputs to be written per field. Users can specify multiple fields, but would need separate data entries for each field. This is a function of the current metadata properties structure, I am going to write up an issue proposing a change to the input structure which would result in a single input file for all field configurations.

> * There are still variable names that are being translated from the metadata to the DCA correctly, but don't make sense in the DCA context, e.g. under Manure Manage Scenarios, the variable "Bedding Type" would be easier to understand if it was named "Bedding _Name_", but "Bedding Type" is in fact its current name. It's not the end of the world but I'm curious to know whether we will address these types of things by changing the variable name in the metadata, or by changing how it appears in the DCA. I don't need an answer now, just something to consider for future DCA improvements!

Another good catch, the information pop-up helps a little with this but it is still pretty confusing. The thing is that the appearance of the DCA is dictated by the metadata, so we may need to add an extra field to the metadata in order to address this if we can't change the actual structure of the inputs.

> * I really like the 'file name' feature, and the JSONS saved and ran properly for me 😁

Noice, me too.

> One improvement to add to the list, if desired: for feeds, I think I recall the old DCA was able to reference the feed library via a dropdown selection, so users could choose the feed by name, not by number. Either way, that would be great to have.

That definitely sounds like it would make things a lot easier. I'm in favor of adding this in a follow-up PR, because I think it will require some significant additions to the metadata and/or code.

> Overall, this looks great to me and I will approve notwithstanding any further changes from my or others' comments!

👍

RUFAS/data_collection_app_updater.py
Isgotkowitz: Placeholders have been implemented

PooyaHekmati (approved):
[LGTM](https://lgtm.lol/i/809)

Isgotkowitz:
@elle-andreen I think the Manure inputs have been fixed without breaking any of the other dropdowns.

elle-andreen (approved):
All looks good to me now, thanks Ed! Looking forward to having this functionality onboard and to future DCA updates!

Isgotkowitz:
> List of improvements to make after this PR:
> 
> * Add feature that automatically generates metadata containing all the input files.

#2122

>   * Restructuring Field and all related inputs will make this feasible.

#2121

> * Add Tasks as kind of input file that DCA can be used for.

#2123 

> * ~Don't automatically add 0's in input fields that don't have defaults.~
> * ~Make null a valid input type for nullable inputs.~
> * Improve messages when inputs are invalid

#2124 

> * Write Wiki for using the DCA
> * Add boolean field to each property called "data_collection_app_compatible". When iterating through properties, use that field to determine whether or not to write a DCA schema for it, not the hard-coded list currently in `DataCollectionAppUpdater`

#2125

> * Allow users to select from drop down of feed names instead of typing feed numbers.

#2126

PR #2047, opened 2024-10-29
Extract send daily variable methods

### LINKED ISSUE #1818: [Crop & Soil] Break down main methods in field data reporter
Split the two main reporter methods in FieldDataReporter into smaller, more focused reporter methods.

Screenshot 2024-07-16 at 10 34 56 AM

The above diagram illustrates the proposed breakdown of send_daily_variables() and send_annual_variables() methods.
#### send_daily_variables() submethods	
- send_field_daily_variables() sends daily variables specific to the field. E.g., current_residue, transpiration, max_transpiration, etc.
- send_soil_daily_variables() handles sending variables related to the soil. E.g., water_evaporated, eroded_sediment, accumulated_runoff, etc.
- send_vadose_zone_layer_daily_variables() is responsible for sending daily variables for the vadose zone layer. E.g., active_organic_nitrogen_content, stable_organic_nitrogen_content, nitrate_content, fresh_organic_nitrogen_content, etc.
- send_soil_layer_daily_variables() handles sending daily variables for each soil layer. E.g., temperature, percolated_water, water_content, etc.
- send_crop_daily_variables() sends daily variables related to the crops. E.g., root_depth, biomass, usable_light, biomass_growth_max, etc.
#### send_annual_variables() submethods
- send_field_annual_variables() sends annual variables specific to the field. Currently, there is only one variable — annual_irrigation_water_use_total.
- send_soil_annual_variables() is responsible for sending annual variables related to the soil. E.g., annual_water_content_change, annual_nitrates_content_change, annual_soil_evaporation_total, etc.
- send_soil_layer_annual_variables() sends annual variables for each soil layer. E.g., annual_nitrous_oxide_emissions_total, annual_ammonia_emissions_total, annual_decomposition_carbon_CO2_lost, etc.

matthew7838:
These PR extracts send daily and annual methods into more specific field, soil, and layer-specific sub-methods.

### Context

Issue(s) closed by this pull request: closes #1818

### What

1. The methods will now be extracted as the diagram's plan in PR #1818.
2. Annual and daily tests have been simplified to stay closer to the purpose of unit testing, removing checks of multiple calls on the same method since we should only check that when each method is called it's performing correctly. Multiple call result checks should be performed at unit tests at higher hierarchy caller methods.

### Why

Original annual and daily reporting methods are too long and make it harder to track down and test.

### How

Extracted as the diagram's plan in PR #1818.

### Test plan

1. Full coverage test suits including the extracted methods run on field reporter.
Note: I only choose 4 variables for each method to test if it's being correctly reported to OM. Let me know if that sounds reasonable, I don't think it makes sense to assert all 40-60ish variables in the pool.


RUFAS/routines/field/manager/field_data_reporter.py
```diff
@@ -27,966 +29,1009 @@ def __init__(self, fields: List[Field]):
...
+
+            self.send_vadose_zone_layer_daily_variables(field)
+
+            for index, layer in enumerate(field.soil.data.soil_layers):
+                self.send_soil_layer_daily_variables(layer, index, field.field_data.name)
+            for crop in field.crops:
+                self.send_crop_daily_variables(crop, field.field_data.name)
+
+    def send_annual_variables(self) -> None:
+        """Sends annual variables to the output manager."""
```
ew3361zh: Small point but what do you think about adding something specific about which module these are variables from in the docstrings for this and `send_daily_variables()`?

RUFAS/routines/field/manager/field_data_reporter.py
```diff
@@ -27,966 +29,1009 @@ def __init__(self, fields: List[Field]):
...
+        for field in self.fields:
+            self.send_field_annual_variables(field)
+
+            self.send_soil_annual_variables(field)
+
+            for index, layer in enumerate(field.soil.data.soil_layers):
+                self.send_soil_layer_annual_variables(layer, field.field_data.name, index)
+
+    def send_crop_daily_variables(self, crop: Crop, field_name: str) -> None:
+        """Sends crop related daily variables."""
```
ew3361zh:
```suggestion
        """Sends crop related daily variables to output manager."""
```

RUFAS/routines/field/manager/field_data_reporter.py
```diff
@@ -27,966 +29,1009 @@ def __init__(self, fields: List[Field]):
...
-                field.soil.data.profile_stable_organic_nitrogen_total,
-                dict(info_map, **{"units": MeasurementUnits.KILOGRAMS_PER_HECTARE}),
-            )
-            self.om.add_variable(
-                "profile_fresh_organic_nitrogen_total",
-                field.soil.data.profile_fresh_organic_nitrogen_total,
-                dict(info_map, **{"units": MeasurementUnits.KILOGRAMS_PER_HECTARE}),
-            )
+    def send_vadose_zone_layer_daily_variables(self, field: Field) -> None:
+        """Sends vadose zone layer related daily variables."""
```
ew3361zh:
```suggestion
        """Sends vadose zone layer related daily variables to output manager."""
```

RUFAS/routines/field/manager/field_data_reporter.py
```diff
@@ -27,966 +29,1009 @@ def __init__(self, fields: List[Field]):
...
-                    crop.data.residue_nitrogen,
-                    dict(info_map, **{"units": MeasurementUnits.KILOGRAMS_PER_HECTARE}),
-                )
-                self.om.add_variable(
-                    "residue_phosphorus",
-                    crop.data.residue_phosphorus,
-                    dict(info_map, **{"units": MeasurementUnits.KILOGRAMS_PER_HECTARE}),
-                )
+    def send_field_daily_variables(self, field: Field) -> None:
+        """Sends field related daily variables."""
```
ew3361zh:
```suggestion
        """Sends field related daily variables to output manager."""
```

RUFAS/routines/field/manager/field_data_reporter.py
```diff
@@ -27,966 +29,1009 @@ def __init__(self, fields: List[Field]):
...
+        )
+        # confirm unit
+        self.om.add_variable(
+            "days_into_watering_interval",
+            field.field_data.days_into_watering_interval,
+            dict(info_map, **{"units": MeasurementUnits.DAYS}),
+        )
+
+    def send_soil_layer_annual_variables(self, layer: LayerData, field_name: str, index: int) -> None:
+        """Sends layer related annual variables."""
```
ew3361zh:
```suggestion
        """Sends layer related annual variables to output manager."""
```

RUFAS/routines/field/manager/field_data_reporter.py
```diff
@@ -27,966 +29,1009 @@ def __init__(self, fields: List[Field]):
...
+        )
 
-            nitrates_content_change = field.soil.data.profile_nitrates_total - field.soil.data.initial_nitrates_total
-            self.om.add_variable(
-                "annual_nitrates_content_change",
-                nitrates_content_change,
-                dict(info_map, **{"units": MeasurementUnits.KILOGRAMS_PER_HECTARE}),
-            )
+    def send_field_annual_variables(self, field: Field) -> None:
+        """Sends field related annual variables."""
```
ew3361zh: Same as others

RUFAS/routines/field/manager/field_data_reporter.py
```diff
@@ -27,966 +29,1009 @@ def __init__(self, fields: List[Field]):
...
+            "annual_irrigation_water_use_total",
+            field.field_data.annual_irrigation_water_use_total,
+            dict(info_map, **{"units": MeasurementUnits.MILLIMETERS}),
+        )
 
-            # ----------------------------adding layer data
-            for index, layer in enumerate(field.soil.data.soil_layers):
-                info_map["suffix"] = "field='" + field.field_data.name + "',layer='" + str(index) + "'"
+    def send_soil_annual_variables(self, field: Field) -> None:
+        """Sends soil related annual variables."""
```
ew3361zh: Same comment as others

ew3361zh:
Really nice work, Matthew! I think this is looking close to ready for me. I left a few comments on docstrings that could be a little more specific as to where they're sending variables data.

> Note: I only choose 4 variables for each method to test if it's being correctly reported to OM. Let me know if that sounds reasonable, I don't think it makes sense to assert all 40-60ish variables in the pool.

It does seem cumbersome to assert all variables in the pool but what if one of those variables gets accidentally removed from being reported to OM (we're doing lots of refactoring, refreshing, and moving things around), will that get missed here? i.e. is there a good case for asserting all the variables so that we'd notice if one got accidentally removed?

matthew7838:
> Really nice work, Matthew! I think this is looking close to ready for me. I left a few comments on docstrings that could be a little more specific as to where they're sending variables data.
> 
> > Note: I only choose 4 variables for each method to test if it's being correctly reported to OM. Let me know if that sounds reasonable, I don't think it makes sense to assert all 40-60ish variables in the pool.
> 
> It does seem cumbersome to assert all variables in the pool but what if one of those variables gets accidentally removed from being reported to OM (we're doing lots of refactoring, refreshing, and moving things around), will that get missed here? i.e. is there a good case for asserting all the variables so that we'd notice if one got accidentally removed?

I think that having a check for the total amount of calls is not the safest way but still works to a certain degree in checking this case.

ew3361zh:
> > Really nice work, Matthew! I think this is looking close to ready for me. I left a few comments on docstrings that could be a little more specific as to where they're sending variables data.
> > > Note: I only choose 4 variables for each method to test if it's being correctly reported to OM. Let me know if that sounds reasonable, I don't think it makes sense to assert all 40-60ish variables in the pool.
> > 
> > 
> > It does seem cumbersome to assert all variables in the pool but what if one of those variables gets accidentally removed from being reported to OM (we're doing lots of refactoring, refreshing, and moving things around), will that get missed here? i.e. is there a good case for asserting all the variables so that we'd notice if one got accidentally removed?
> 
> I think that having a check for the total amount of calls is not the safest way but still works to a certain degree in checking this case.

I wonder if this in combination with C&S e2e testing would be adequate?

matthew7838:
> > > Really nice work, Matthew! I think this is looking close to ready for me. I left a few comments on docstrings that could be a little more specific as to where they're sending variables data.
> > > > Note: I only choose 4 variables for each method to test if it's being correctly reported to OM. Let me know if that sounds reasonable, I don't think it makes sense to assert all 40-60ish variables in the pool.
> > > 
> > > 
> > > It does seem cumbersome to assert all variables in the pool but what if one of those variables gets accidentally removed from being reported to OM (we're doing lots of refactoring, refreshing, and moving things around), will that get missed here? i.e. is there a good case for asserting all the variables so that we'd notice if one got accidentally removed?
> > 
> > 
> > I think that having a check for the total amount of calls is not the safest way but still works to a certain degree in checking this case.
> 
> I wonder if this in combination with C&S e2e testing would be adequate?

That's a good question I can bring it up to Pooya on my 1-1 in a bit.

ew3361zh (approved):
This is looking good, Matthew, nice work!

Curious to hear about the follow up from your check-in with Pooya on how to best ensure that C&S is reporting out all the details we expect. I do think some combination of unit/e2e testing could do the job for us.

allisterakun (approved):
Looks great! Nice job, Matthew! A lot more cleaner now!

Isgotkowitz:
@matthew7838 this will change the names of variables that are reported from the Crop and Soil module -- has this change been coordinated with the team so that people know they will have to update their filters?

matthew7838:
> @matthew7838 this will change the names of variables that are reported from the Crop and Soil module -- has this change been coordinated with the team so that people know they will have to update their filters?

No, can I merge now? or should I wait for their confirmation?

Isgotkowitz:
> > @matthew7838 this will change the names of variables that are reported from the Crop and Soil module -- has this change been coordinated with the team so that people know they will have to update their filters?
> 
> No, can I merge now? or should I wait for their confirmation?

I strongly recommend mentioning it on slack and waiting at least an hour or two for people to see it. You may not get a firm confirmation, but if it is going to throw a big monkey wrench into things someone will probably let you know

PR #2048, opened 2024-10-29
[Animal] Add E2E Testing For Animal Module

### LINKED ISSUE #1992: [E2E Testing][Animal] Add end-to-end testing coverage for the Animal module
## Status

The end-to-end testing routine for RuFaS currently does not cover the Animal module

## Intention

End-to-end testing should cover the Animal module in order for maintainers to more easily verify that correct behavior is maintained in when reviewing changes.

## Request

Add the following files to the directory `input/data/end_to_end_testing`:
- `e2e_json_animal_filter.json`. This file should contain a filter for gathering actual results from the animal module, as well as contain the expected animal-related results. Use `e2e_json_feed_storage_filter.json` as an example for how to write this file.
- `e2e_comparison_animal_differences.json`. This file should contain a filter that will gather the results of end-to-end testing the Animal module. Use `e2e_comparison_feed_storage_differences.json` as an example for how to write this file.
- `end_to_end_testing_result_paths.json`. This file already exists, but an entry should be added that specifies the Animal domain, the path to expected outputs, and the pattern used to find the actual outputs from the Animal module. Use the existing entry for Feed Storage as an example for how to write this.

ew3361zh:
Adds Animal module (domain) end-to-end testing expected results, associated filter files, and updates FeedStorage expected results.

Note: `expected_results` sections are minified to save on space/memory. Initially I was thinking it might be necessary to extract results for different parts of the animal module separately to accommodate the large sets of results for Animal domain but minifying the whole set made them much more manageable in size. So I followed suit and minified the `expected_results` section for FeedStorage as well.

### Context

Issue(s) closed by this pull request: closes #1992

### What

Adds testing results, filters, and configs for Animal module end-2-end testing.

### Why

Continued expansion of e2e testing results covering all modules/domains.

### How

1. Initial animal e2e filter files were set up as outlined in the associated issue.
2. The start and end dates for the e2e testing config had to be updated to 2013-2014 because the previous dates were causing net merit errors.
3. Additionally the e2e testing config now has `simulate_animals` set to `true` to get actual animal results when e2e testing runs.
4. e2e testing was run with the animal filter set to `["AnimalModuleReporter.*", "AnimalManager.*"]` and the results were collected and updated in the `expected_results` section of the animal filter file.
5. e2e testing expected results for FeedStorage were updated based on results of the new date range for the e2e config.

### Test plan

e2e testing was run as outlined in the issue and there were no found differences between the expected and the actual results for both the FeedStorage domain and the Animal domain.


RUFAS/e2e_test_results_comparer.py
```diff
@@ -62,7 +62,7 @@ def compare_actual_and_expected_test_results(json_output_path: Path) -> None:
                 filter_and_results = json.load(e_to_e_results)
                 expected_results = filter_and_results["expected_results"]
 
-            diff = DeepDiff(expected_results, actual_results, ignore_order=True, verbose_level=2)
+            diff = DeepDiff(expected_results, actual_results, ignore_order=True, verbose_level=2, significant_digits=8)
```
ew3361zh, line 65: Reduces chances of differences found between actual and expected results and still is a very close match.

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -1061,7 +1061,7 @@ def _record_animal_events(
         """
 
         info_map = {
-            "class": AnimalModuleReporter.__class__.__name__,
+            "class": AnimalModuleReporter.__name__,
```
ew3361zh, line 1064: These were not reporting correctly to OM. Removing the `__class__` from the value allows it to report `AnimalModuleReporter` as we would hope.

input/data/config/end_to_end_testing_config.json
```diff
@@ -1,9 +1,9 @@
 {
-    "start_date": "2010:20",
-    "end_date": "2011:20",
+    "start_date": "2013:20",
+    "end_date": "2014:20",
```
ew3361zh, line 3: Date range updated to account for a date range that would work with net merit requirements.

input/data/config/end_to_end_testing_config.json
```diff
@@ -1,9 +1,9 @@
...
     "random_seed": 42,
     "set_seed": true,
-    "simulate_animals": false,
+    "simulate_animals": true,
```
ew3361zh, line 6: This needed to be true in order to get results from the animal module/domain.

RUFAS/e2e_test_results_comparer.py
Isgotkowitz: Nice!

RUFAS/routines/animal/animal_module_reporter.py
Isgotkowitz: This is a good catch! However, since this is changing the names of outputs, could you please clear it with @JoeWaddell, @krb78, and/or other members of the Animal team to make sure that everyone will be able to update their filters accordingly?

changelog.md
```diff
@@ -21,6 +21,7 @@ A **changelog** is a structured record of changes made to the codebase over time
 
 ### Current version
 - [1945](https://github.com/RuminantFarmSystems/MASM/pull/1945) - [minor change] [Animal] Implements Animal Genetics submodule.
+- [2048](https://github.com/RuminantFarmSystems/MASM/pull/2048) - [minor change] [Animal][Testing] Adds end to end testing for Animal domain.
```
Isgotkowitz: This entry should be added in the "Next Version Updates" section.

Isgotkowitz (commented):
The code here LGTM! But before this is merged I think the Animal SMEs/team should be alerted to the the change in how animal results are reported (there is a comment on this)

Something that isn't a sticking point on this PR is that whenever expected results for E2E testing are added or updated, a SME should check them to make sure there are no bugs that popped up in them. This isn't very important right now, because E2E testing is still in its infancy and not part of our development pipeline yet. Just wanted to mention it so we can keep it in mind going forward.

Isgotkowitz (approved):
Noice

matthew7838 (approved):
LGTM, got it running and passing. Really helpful for my CS e2e PR!

PR #2050, opened 2024-10-30
[Crop&Soil][Feed] Disentangle Crop & Soil and Feed Storage modules

### LINKED ISSUE #1626: Disentangle the Crop & Soil and Feed Storage modules
The connection between the Crop and Soil module and the Feed Storage module should be disentangled so that it meets the requirements outline in #1622.

## Suggested Solution

When a crop is harvested in `CropManagement`, it should return the harvested crop up to `Field`. `Field` should return that crop (and any other crops that are harvested on that day) to `FieldManager`.`FieldManager` should return all harvested crops to `SimulationEngine` through the `daily_update_routine`, and then `SimulationEngine` should pass all those crops to `FeedManager.receive_crop`.

The initialization routines of `FieldManager` and `Field` should be modified so that they do not import or use the `FeedManager` at all.

Because the `HarvestedCrop` data structure is shared by both Crop and Soil and Feed Storage, it should be moved somewhere that is not in either of these modules (please use the `git mv` command or an equivalent if you can for this - it helps to keep the Git history easier to comprehend and trace through). I suggest creating a new subdirectory at the level of `RUFAS` (if there isn't one when this issue is addressed) dedicated to containing data structures that are used to transfer information between biophysical modules.

## Current Behavior

The Feed Storage module's `FeedManager` is passed to the Crop and Soil module's `FieldManager` at initialization, and is kept as an attribute by every `Field` instance. When a crop is harvested (in the `CropManagement` class), a `HarvestedCrop` instance is created and passed to `FeedManager.receive_crop`.

Isgotkowitz:
Disentangles the Crop&Soil and Feed Storage modules.

### Context

Issue(s) closed by this pull request: closes #1626 

### What

This PR "disentangles" C&S and Feed Storage so that they
1. Do not import any modules from the other
2. Only pass information to each other by giving information to `SimulationEngine` and having it pass that information to the final destination
3. All data structures used to pass information through `SimulationEngine` are defined in the `data_structures/` directory, which is at the same level as `SimulationEngine`

### Why

Disentangling these modules provides better modularity and isolation between components of RuFaS, making them easier to observe, debug, and improve

### How

Before, the `FeedManager` was passed into the Crop and Soil module and kept as an attribute of `Field`. When a crop was harvested, `FeedManager` was passed into that crop and it would call `receive_crop` directly.

Now, when `Field` harvests a crop, `Crop` will return all the information needed for the harvest to be stored to `Field`, which will collect all the harvests from the current day. All harvests are passed up to `FieldManager`, then to `SimulationEngine`, which puts every harvest in `FeedManager` using `receive_crop`.

### Test plan

All unit tests have been updated so that they are now passing. To be 100% transparent, each test was updated just enough to make it pass, not necessarily to verify that every changed facet of the method being tested was working correctly. Testing all new functionality "perfectly" would take a lot of effort, and would not increase my confidence in the correctness by very much. But if this is not ok, please let me know and I will do a better job with the tests!

To verify that the simulation was still working correctly, I ran the default simulation on this branch and `dev` and compared the results. They were unchanged, and this gives me a lot of confidence that the disentanglement did not introduce any additional bugs into the code.


RUFAS/data_structures/crop_soil_feed_storage_connection.py
Isgotkowitz, line 1: I talked with Pooya about how to organize all the stuff that is used to communicate between biophysical modules. The conclusion we came to (which is subject to change) is that:
- All data types/structures/enums should go in the `data_structures/` directory
- Each connection between two biophysical modules (Animal -> Manure, Manure -> Crop&Soil, etc.) will have a dedicated .py file for the data structures used to communicate between them.
- The name of the .py file will be `module_one_module_two_connection.py` where "module_1" is the name of the module sending information, and "module_2" is the name of the module receiving the information.
- Every data structure/enum/type used to communicate information between two modules will go in the .py file named for that connection.

RUFAS/routines/feed_storage/feed_manager.py
```diff
@@ -22,21 +20,19 @@
...
-    PROTECTED_TARPED = ProtectedTarped
-    UNPROTECTED = Unprotected
-    BALEAGE = Baleage
-    DRY = Dry
-    HIGH_MOISTURE = HighMoisture
-    BUNKER = Bunker
-    PILE = Pile
-    BAG = Bag
+# Maps each storage type to its respective class.
+STORAGE_TYPE_TO_CLASS_MAP: dict[StorageType, Storage] = {
```
Isgotkowitz: Mypy doesn't like this type hint, it doesn't recognize any of the `Storage` subclasses as `Storage`s - each entry adds an error, and this accounts for the bulk of the Mypy error count increase

RUFAS/data_structures/crop_soil_feed_storage_connection.py
```diff
@@ -203,3 +302,7 @@ def _calculate_total_sensible_heat_generated(self) -> float:
             104 * moisture_frac**2.18 * self.bale_density**0.5 + 5.72 * moisture_frac**1.23 * self.bale_density**0.94
         )
         return heat
+
+
+"""Type alias that is used throughout the Crop module when sending information back to Simulation Engine."""
+HarvCropStorageType = tuple[HarvestedCrop, StorageType]
```
ew3361zh:
```suggestion
HarvestedCropStorageType = tuple[HarvestedCrop, StorageType]
```

RUFAS/data_structures/crop_soil_feed_storage_connection.py
```diff
@@ -203,3 +302,7 @@ def _calculate_total_sensible_heat_generated(self) -> float:
             104 * moisture_frac**2.18 * self.bale_density**0.5 + 5.72 * moisture_frac**1.23 * self.bale_density**0.94
         )
         return heat
+
+
+"""Type alias that is used throughout the Crop module when sending information back to Simulation Engine."""
```
ew3361zh: Can you add something about why these two need to be grouped together like this? Maybe a brief explanation of what the components are could help explain why they're together.

RUFAS/routines/feed_storage/feed_manager.py
ew3361zh: ChatGPT suggested this which fixed it for me:
```suggestion
STORAGE_TYPE_TO_CLASS_MAP: dict[StorageType, Type[Storage]] = {
```

I also had to change line 15 to:
`CROP_TO_STORAGE_MAPPING: Dict[CropCategory, List[Type[Storage]]] = {`

ew3361zh (commented):
This looks good - I have a few suggestions - I think it's worth trying to fix the MyPy errors because I don't think it would be too difficult (see comments).

I wondered about the data structure name's clarity - maybe it's fine, I just wanted to bring it up - possibly worth discussing in the wt? Does it need more information (just a `to` perhaps)? Condense the module names to be single strings of letters e.g. `cropsoil_feedstorage.py`? 

I think my vote might be for a `to` to be `crop_soil_to_feed_storage.py`.

I would like to see what filter you were using to test the results between the dev and feature branch though if you could share that.

RUFAS/data_structures/crop_soil_feed_storage_connection.py
Isgotkowitz: :sob: I was hoping to keep it shorter, but "Harv" does not look good

RUFAS/data_structures/crop_soil_feed_storage_connection.py
Isgotkowitz: Sure!

RUFAS/routines/feed_storage/feed_manager.py
Isgotkowitz: Nice, thank you Niko! I used the lower case `type` instead, just so that I wouldn't have to add any more imports but the end result is the same

Isgotkowitz:
> This looks good - I have a few suggestions - I think it's worth trying to fix the MyPy errors because I don't think it would be too difficult (see comments).

It was not, thank you for the suggestions!

> I wondered about the data structure name's clarity - maybe it's fine, I just wanted to bring it up - possibly worth discussing in the wt? Does it need more information (just a `to` perhaps)? Condense the module names to be single strings of letters e.g. `cropsoil_feedstorage.py`?
> 
> I think my vote might be for a `to` to be `crop_soil_to_feed_storage.py`.

Definitely worth discussing in the wt! I am not attached to any particular naming conventions as long as we're consistent with them, though whatever we come up I think we should clear it with @PooyaHekmati.

> I would like to see what filter you were using to test the results between the dev and feature branch though if you could share that.

Sure! Two are from the sensitivity analysis, one is a FARM-ES filter, and another is related to the emissions estimator. I think between all of them they constitute a pretty comprehensive check. But because there are no outputs coming from the Feed Storage module (and E2E testing is down) there isn't an easy way to check the Feed Storage side of things.

First sensitivity analysis one - [report_S30.json](https://github.com/user-attachments/files/17620548/report_S30.json)
First sensitivity analysis two - [report_S365.json](https://github.com/user-attachments/files/17620555/report_S365.json)
FARM-ES one - [report_scratch.json](https://github.com/user-attachments/files/17620558/report_scratch.json)
Feed Emissions post-processing one - [report_scratch_2.json](https://github.com/user-attachments/files/17620559/report_scratch_2.json)

RUFAS/data_structures/crop_soil_feed_storage_connection.py
ew3361zh: 100% agree - I didn't want to have to suggest it.

PooyaHekmati:
> > Condense the module names to be single strings of letters e.g. `cropsoil_feedstorage.py`?

I'm not going to pretend that I have read the PR and the whole convo here, but I can say that `cropsoil_feedstorage.py` does not adhere to PEP 8.

RUFAS/simulation_engine.py
```diff
@@ -132,7 +132,9 @@ def _daily_simulation(self) -> None:
         self.animal_manager.daily_updates(self.feed, self.weather, self.time)
         all_pen_manure_data = self.animal_manager.collect_pen_manure_data()
         self.manure_manager.daily_update(all_pen_manure_data, self.animal_manager.simulation_day)
-        self.field_manager.daily_update_routine(self.weather, self.time)
+        harvested_crops = self.field_manager.daily_update_routine(self.weather, self.time)
+        for crop in harvested_crops:
+            self.feed_manager.receive_crop(*crop)
```
ew3361zh: Do you have any thoughts long-term for what is handled here in `SimulationEngine._daily_simulation()` as we start using it to pass data to and from the different biophysical modules?
I don't know how many things will need to be passed but thinking about this since this PR introuces logic that starts sorting through harvested crops. And based on the disentanglement I'm working on for Manure and C&S, nutrient requests will need to be assessed for each field each day, sent to the manure module and then nutrient request responses from the manure module will need to be looped through and sent back to C&S.
I could see this function quickly getting a little out of hand.
I don't know that this needs to be rearranged but thought it was good to be thinking about it at this stage if we're not already.

Isgotkowitz: Since we don't totally have the disentangled and refreshed design worked out, I don't have any concrete ideas about what should be done with `_daily_simulation`. But I agree, we should be starting to give it some thought.

Even though this method may get a little out-of-hand, it will still be much easier for someone to understand what data is passed between the biophysical modules by reading one big method as opposed to combing through the entire codebase. So I'm not too concerned about an out-of-hand `_daily_simulation`.

I think it would be good to plan on doing all the disentangling and refreshing, letting `_daily_simulation` get as complicated as it needs to be (within reason), and then reassessing how we can refactor `SimulationEngine` more generally.

One possibility that I think is worth considering would be combining `_daily_simulation`, `_annual_simulation`, and `_run_simulation_main_loop` into one method, containing a single for loop which iterates over every single day in the simulation. This method would be a bit of a monster, but from a performance standpoint it would be better because every call to a subroutine incurs an overhead penalty.

RUFAS/routines/feed_storage/feed_manager.py
```diff
@@ -22,21 +20,19 @@
...
-    PROTECTED_WRAPPED = ProtectedWrapped
-    PROTECTED_TARPED = ProtectedTarped
-    UNPROTECTED = Unprotected
-    BALEAGE = Baleage
-    DRY = Dry
-    HIGH_MOISTURE = HighMoisture
-    BUNKER = Bunker
-    PILE = Pile
-    BAG = Bag
+# Maps each storage type to its respective class.
```
matthew7838: I don't think this inline comment will be necessary.

RUFAS/data_structures/crop_soil_feed_storage_connection.py
```diff
@@ -203,3 +302,11 @@ def _calculate_total_sensible_heat_generated(self) -> float:
...
         )
         return heat
+
+
+"""
+Type alias that is used throughout the Crop module when sending information back to Simulation Engine. The HarvestedCrop
+contains the mass and nutritional information that was harvested, and StorageType indicates how the harvested yield will
+be stored by the Feed Manager.
+"""
+HarvestedCropStorageType = tuple[HarvestedCrop, StorageType]
```
matthew7838: Just a suggestion, would utilizing a named tuple be more accessible?

RUFAS/routines/field/manager/field_manager.py
```diff
@@ -87,9 +88,12 @@ def daily_update_routine(self, weather: Weather, time: Time) -> None:
                 "units": MeasurementUnits.HOURS,
             }
             self.om.add_variable("daylength", current_conditions.daylength, info_map)
-            field.manage_field(time, current_conditions=current_conditions)
+            crops = field.manage_field(time, current_conditions=current_conditions)
```
matthew7838:
```suggestion
            harvested_crops = field.manage_field(time, current_conditions=current_conditions)
```
I think this makes the whole harvesting process easier to track in the S&C module.

tests/soil_crop_tests/crop_tests/test_crop.py
```diff
@@ -238,15 +237,12 @@ def test_manage_crop_harvest(mocker: MockerFixture) -> None:
     field_size = 10.0  # hectares
     mock_time = mocker.Mock(spec=Time)
     mock_soil_data = mocker.Mock(spec=SoilData)
-    mock_feed_manager = mocker.Mock(spec=FeedManager)
 
     manage_harvest_mock = mocker.patch.object(crop._crop_management, "manage_harvest")
 
-    crop.manage_crop_harvest(mock_harvest_op, field_name, field_size, mock_time, mock_soil_data, mock_feed_manager)
+    crop.manage_crop_harvest(mock_harvest_op, field_name, field_size, mock_time, mock_soil_data)
```
matthew7838: I agree with not extending the tests too much, but I think a method that changes from doesn't return to having a return value, we might still want to test that with the bare minimum. In this case, perhaps just mock a really simple return value of `manage_harvest` and make sure we did get that result.

matthew7838 (approved):
Looks great, Ed. I have a few comments.

1. For the naming of the connection, I think adding the `to` suggested by Niko is a lot more clear.
2. I agree with the testing direction but caught somewhere that I think should be improved fairly easily.

RUFAS/routines/feed_storage/feed_manager.py
Isgotkowitz: It is more of a one-liner docstring, but tomayto, tomahto. I will change this one (and the docstring above it) to use the """ comment notation instead, so it looks more like a docstring than an in line comment.

Also will improve this docstring, because you're right it is unnecessary as written.

RUFAS/data_structures/crop_soil_feed_storage_connection.py
Isgotkowitz: Yes, I think it would. It would make the splat operator (*) unnecessary in SimulationEngine, and would be way easier to mock.

RUFAS/routines/field/manager/field_manager.py
Isgotkowitz: agreed

tests/soil_crop_tests/crop_tests/test_crop.py
Isgotkowitz: Good call

RUFAS/data_structures/crop_soil_feed_storage_connection.py
```diff
@@ -1,9 +1,108 @@
...
+    """
+
+    SMALL_GRAIN = "Small grain"
+    CORN = "Corn"
+    SOY = "Soy"
+    GRASS = "Grass"
+    ALFALFA = "Alfalfa"
+
+
+class CropType(Enum):
```
ew3361zh: I'm curious about the decision to move `CropType`, `CropCategory`, and `StorageType` into this shared data_structure file. I saw your explanation from above but what if these structures are needed down the line to be sent from C&S to Animal or EEE or some other module? It seems like tying these to a specific data sharing direction or process isn't very future proof.

RUFAS/routines/field/manager/field_manager.py
ew3361zh: seconding this suggestion.

RUFAS/routines/field/crop/crop_management.py
```diff
@@ -67,19 +65,28 @@ def manage_harvest(
...
+            yield.
+
         """
         self.determine_harvest_index()
 
+        crop = None
         if harvest_op in (HarvestOperation.HARVEST_KILL, HarvestOperation.HARVEST_ONLY):
             self.cut_crop(collected_fraction=self.data.harvest_efficiency)
-            self._store_harvested_crop(time, field_size, feed_manager)
+            crop = self._get_harvested_crop(time, field_size)
```
ew3361zh: There are a couple other places where it's still crop/crops and I think it would be clearer for it to be harvested_crops whenever possible.
```suggestion
            harvested_crop = self._get_harvested_crop(time, field_size)
```

RUFAS/simulation_engine.py
```diff
@@ -132,7 +132,9 @@ def _daily_simulation(self) -> None:
         self.animal_manager.daily_updates(self.feed, self.weather, self.time)
         all_pen_manure_data = self.animal_manager.collect_pen_manure_data()
         self.manure_manager.daily_update(all_pen_manure_data, self.animal_manager.simulation_day)
-        self.field_manager.daily_update_routine(self.weather, self.time)
+        harvested_crops = self.field_manager.daily_update_routine(self.weather, self.time)
+        for crop in harvested_crops:
```
```suggestion
        for harvested_crop in harvested_crops:
```
¯\_(ツ)_/¯ maybe doesn't make a huge difference but why not.

ew3361zh (commented):
Looking really good, Ed! Once you address the comments, I'd like to run the simulation with the provided filters one more time for good luck but then should be good to approve.

RUFAS/data_structures/crop_soil_feed_storage_connection.py
ew3361zh: Maybe leave them where they were but the new named tuple structure is created and housed in the `crop_soil_to_feed_storage_connection.py` file? Then any other structures that also need to be sent from C&S to FeedStorage could be housed there as well and if there needs to be data that includes CropType that needs to be sent to a different module, it won't need to pull it from this `crop_soil_to_feed_storage_connection.py` file.

RUFAS/simulation_engine.py
ew3361zh: Also, it's just one line but I was talking to Pooya today about my disentanglement PR and he mentioned the level of abstraction wasn't consistent when I started iterating through all the fields in field_manager and I wondered if this too should maybe be done with this for loop. Maybe a process_harvested_crops() function type thing.

Isgotkowitz: We should discuss this more, I would like to flush out what the inconsistency is in the levels of abstraction.

RUFAS/routines/field/manager/field_manager.py
Isgotkowitz: In this particular instance, I changed `crops` to `newly_harvested_crops`, because they are the arg to the `extend` call on `harvested_crops`. But I have changed every other spot in the Field module to use `harvested_crops` as the name.

RUFAS/simulation_engine.py
Isgotkowitz: Following up: the suggested code is in `SimulationEngine`.

For the second part, we discussed and my cents is that none of the attributes in any of the managers (`FieldManager`, `FeedManager`, etc.) should be interacted with directly. Only the interface of the managers should be engaged with in `SimulationEngine`.

RUFAS/data_structures/crop_soil_feed_storage_connection.py
Isgotkowitz: Documenting discussion: 

> what if these structures are needed down the line to be sent from C&S to Animal or EEE or some other module?

For now `CropType`, `CropCategory` and `StorageType` are only used in the CropSoil - Feed Storage connection, so keeping them together makes things cleaner. But if they get used in more than one connection, we could put them in standalone files in `data_structures/`

> Maybe leave them where they were but the new named tuple structure is created and housed in the `crop_soil_to_feed_storage_connection.py` file?

The issue with leaving them where they were is that Crop and Soil would still be importing something from Feed Storage, which would be entangling the two modules.

tests/soil_crop_tests/crop_tests/test_crop.py
Isgotkowitz: Updated tests so that they are checking the returned values are correct.

Isgotkowitz:
@matthew7838 and @ew3361zh thank you for your FB! I have updated the names of variables so "harvested_crops" is used as consistently as possibly throughout the CropSoil and SimulationEngine files. Have also updated the test suite, so that the methods which now return `HarvestedCropStorageType`s have their return values tested.

ew3361zh (approved):
Looks good to me, Ed, nice work!
Ran this with the provided filters (which seemed adequate for comparison of the affected parts of the outputs) and also ran e2e testing and saw no differences.

I didn't see a response to one of Matthew's comments though - could you address this or comment on leaving things as they are?

> 1. For the naming of the connection, I think adding the `to` suggested by Niko is a lot more clear.

Isgotkowitz:
> Looks good to me, Ed, nice work! Ran this with the provided filters (which seemed adequate for comparison of the affected parts of the outputs) and also ran e2e testing and saw no differences.
> 
> I didn't see a response to one of Matthew's comments though - could you address this or comment on leaving things as they are?
> 
> > 1. For the naming of the connection, I think adding the `to` suggested by Niko is a lot more clear.

Yes, I will look into the whether or not adding "to" will adhere to PEP 8.

Isgotkowitz:
Checked with ChatGPT, `crop_soil_to_feed_storage_connection.py` does adhere to PEP 8 so I will change the name. Thank you for reminding me @ew3361zh!

PR #2057, opened 2024-11-07
[Manure][Crop & Soil] Disentangles Manure and Crop & Soil Modules

### LINKED ISSUE #1625: Disentangle the Manure and Crop & Soil modules

ew3361zh:
Reorganizes data passing between Manure and Crop & Soil modules to be through SimulationEngine

Note: Branched off #2050
Second Note: the large file-change count is primarily due to import statement updates resulting from moving `NutrientRequests` and `NutrientRequestResults` into the data_structures directory.

### Context

Issue(s) closed by this pull request: closes #1625

### What

- Moves data passing mechanisms between Crop & Soil and Manure modules to SimulationEngine 
- This facilitates Crop and Soil and Manure modules no longer directly interacting nor needing to know about one another.

### Why

Helps in the overarching effort to disentangle all biophysical modules from one another.

### How

- Reorganizes mechanism by which manure is requested by Fields so that requests and responses are all gathered via SimulationEngine's daily update prior to the FieldManager daily update being called. This is primarily done in the new SimulationEngine function `generate_daily_manure_applications()`.
- Creates new named-tuple data structures (`ManureEventNutrientRequest` and `ManureEventNutrientRequestResults`) for data passed between manure and crop & soil modules.
- Moves previously existing components of new data structures into `data_structures` directory.
- Removes all references to ManureManager from Crop & Soil module (including testing).

### Test plan

- Unit tests were updated.
- Simulation was run on both dev branch and feature branch to compare results of manure applications. Results were identical.
- Here is the filter used:
```
{
    "multiple": [
        {
        "name": "Field Manure Applications",
        "filters": [
            "Field._record_manure_application.*"
        ]
        }
    ]
}
```


README.md
```diff
@@ -1,7 +1,7 @@
 [![Flake8](https://img.shields.io/badge/Flake8-passed-brightgreen)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
 [![Pytest](https://img.shields.io/badge/Pytest-passed-brightgreen)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
-[![Coverage](https://img.shields.io/badge/Coverage-95%25-brightgreen)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
-[![Mypy](https://img.shields.io/badge/Mypy-3535%20errors-red)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
+[![Coverage](https://img.shields.io/badge/Coverage-96%25-brightgreen)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
+[![Mypy](https://img.shields.io/badge/Mypy-3531%20errors-red)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
```
Isgotkowitz: Nice!

RUFAS/routines/field/manager/field_manager.py
```diff
@@ -463,3 +467,16 @@ def _setup_soil_layer(field_size: float, top_depth: float, initial_residue: floa
...
+        """
+        Checks list of ManureEvents, sends all that occur today to another method to be executed.
+
+        Parameters
+        ----------
+        time : Time
+            Object containing the current year and day of the simulation.
+
+        """
+        manure_requests = field._check_manure_application_schedule(time)
```
Isgotkowitz: I think `_check_manure_application_schedule` should be renamed to `check_manure_application_schedule`. The underscore at the beginning indicates that it is "private", and should only be accessed within `Field`. So the difference would not be functional at all, but would keep the code more in line with Python naming conventions.

tests/soil_crop_tests/field_tests/test_field.py
```diff
@@ -78,15 +92,45 @@ def test_manage_field() -> None:
...
+    # Act
+    field.manage_field(mocked_time, mocked_weather, mock_manure_applications)
+
+    # Assert
     field._check_fertilizer_application_schedule.assert_called_once_with(mocked_time)
-    field._check_manure_application_schedule.assert_called_once_with(mocked_time)
+    field._execute_manure_application.assert_called_once_with(
+        requested_nitrogen=10,
+        requested_phosphorus=5,
+        requested_manure_type="TypeA",
```
Isgotkowitz, line 123: Good one

tests/soil_crop_tests/field_tests/test_field.py
```diff
@@ -78,15 +92,45 @@ def test_manage_field() -> None:
     field._check_crop_harvest_schedule = MagicMock()
     field._remove_dead_crops = MagicMock()
     field._reset_crop_field_coverage_fractions = MagicMock()
+    field._execute_manure_application = MagicMock()
```
Isgotkowitz:
```suggestion
    execute_manure_application = mocker.patch.object(field, "_execute_manure_application")
```
Mypy is really doesn't like assigning things to a method, even though it works fine.

tests/soil_crop_tests/field_tests/test_field.py
```diff
@@ -244,81 +288,96 @@ def test_check_fertilizer_application_schedule(
     field._execute_fertilizer_application.assert_has_calls(expected_execution_calls)
 
 
+def test_check_manure_application_schedule() -> None:
+    """Tests that ManureEvents are correctly checked and converted to ManureEventNutrientRequests when scheduled."""
```
Isgotkowitz, line 289: I think this test would be a good candidate for writing more of an integration test than a pure unit test, meaning no mocking of `_filter_events` or `_create_manure_request`. The input vars would be the current date, and expected `ManureEventNutrientRequest`s, with the list of manure events held by the field being defined within the body of the test. I don't think the test as it is right now isn't comprehensive, but having just those simple inputs and outputs with few mocks would make the test more readable.

RUFAS/data_structures/events.py
Isgotkowitz, line 1: What would you think about moving `ManureEvent` and `BaseFieldManagementEvent` to `data_structures`, but keeping the other events within the Crop and Soil module? To me it seems like there is no right answer here, because separating classes linked through inheritance doesn't feel great, but neither does putting unshared data structures into `data_structures`.

RUFAS/simulation_engine.py
```diff
@@ -142,6 +144,30 @@ def _daily_simulation(self) -> None:
...
+        Returns
+        -------
+        dict[str, list[ManureEventNutrientRequestResults]]
+            A dictionary containing the ManureEvents and corresponding NutrientRequestResults for each field in
+            the simulation.
+        """
+        manure_applications: dict[str, list[ManureEventNutrientRequestResults]] = {}
+        for field in self.field_manager.fields:
+            manure_events_requests = self.field_manager.check_manure_schedules(field, self.time)
+            manure_applications[field.field_data.name] = []
```
Isgotkowitz: Instead of `SimulationEngine` reaching into the attributes of `field_manager`, could the name of the requesting field be encapsulated in the data structure(s) passed between Field Manager and Manure Manager.

My reasoning is that with the current setup, `SimulationEngine` is much more dependent on how things are implemented in the Crop and Soil module than if the field name was being passed around in the manure requests and responses.

RUFAS/data_structures/crop_soil_to_manure_connection.py
```diff
@@ -0,0 +1,19 @@
+from typing import NamedTuple, Optional
+
+from RUFAS.data_structures.events import ManureEvent
+from RUFAS.data_structures.nutrient_request import NutrientRequest
+from RUFAS.data_structures.nutrient_request_results import NutrientRequestResults
```
Isgotkowitz: I suggest putting nutrient request and request_results in this file as well.

Isgotkowitz (commented):
Tested it out, found no changes between this branch and `dev`. Also tried out multiple manure requests on the same day, worked as expected. Bunch of suggestions for minor improvements.

The one thing that I think we need to figure out before this PR is merged is how much `SimulationEngine` interacts with the attributes inside of the different managers, there is a comment with more detail below

tests/soil_crop_tests/field_tests/test_field.py
ew3361zh: Yea, there are a bunch of these in the test file, wasn't sure if this was the PR to make those changes but I can certainly start with this one!

RUFAS/data_structures/crop_soil_to_manure_connection.py
allisterakun: Second that

RUFAS/routines/field/field/field.py
```diff
@@ -900,28 +893,59 @@ def _check_tillage_schedule(self, time: Time) -> None:
...
+        """
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self._create_manure_request.__name__,
+            "suffix": f"field='{self.field_data.name}'",
+            "date": {"year": event.year, "day": event.day},
+        }
+        if event.nitrogen_mass == event.phosphorus_mass == 0.0:
+            log_message = "Tried to apply manure with no nitrogen or phosphorus requested."
+            self.om.add_log("manure_application_log", log_message, info_map)
```
allisterakun: Should this be a warning?

RUFAS/simulation_engine.py
allisterakun: I definitely agree with this point! This information should be packaged in C&S module and passed to `SimulationEngine`

allisterakun (commented):
Generally looks good! Confirmed that it works as expected. Just some minor comments, and I do second Ed's point on moving the `generate_daily_manure_applications()` function from `SimulationEngine` into `FieldManager`

tests/soil_crop_tests/field_tests/test_field.py
ew3361zh: Good idea!

RUFAS/data_structures/events.py
ew3361zh: Yea, I had the same thought (mostly the latter part). I think it would be confusing to split them up though. I'm curious what @allisterakun thinks on this one.

RUFAS/routines/field/field/field.py
ew3361zh: Hmm, I believe that was a log before (just code I moved to a separate function) - maybe this is not an unexpected/unwanted occurrence but just one we're trying to keep track of when it happens? @Isgotkowitz Do you agree this should stay a log?

RUFAS/data_structures/events.py
ew3361zh: Do either of you know if we have any examples of this practice (splitting up child classes from their parent) in the model right now?

tests/soil_crop_tests/field_tests/test_field.py
Isgotkowitz: I think that would be better! Consistency is great, but if someone sees the correct way in the code I think they will be more inclined to follow it and fix more of the mocks if they're working on the tests

RUFAS/data_structures/events.py
Isgotkowitz: There are some in Feed Storage. `Storage` is the parent, it's children `Baleage`, `Silage`, `Hay`, and `Grain` are all in different files. But to me this seems different, because the organization is kept consistent: all child classes of `Storage` are in their own files, all grandchild classes of `Storage` are kept with their parent class.

RUFAS/routines/field/field/field.py
Isgotkowitz: No, I think a warning would be better. The inputs currently don't match the manure characteristics requested (users can request N, P, and K but manure applications only work with N and P). So it isn't wholly unexpected that a user would request manure with no N or P

ew3361zh:
Updates in response to feedback:
- Moves field name into the construction of both the `ManureEventNutrientRequest` and `ManureEventNutrientRequestResults` objects and it's now passed in at their creation.
- Moves NutrientRequestResults into the same file with NutrientRequests and updated the imports.
- Removed the private underscore from `check_manure_application_schedule()`
- Changed log to warning when nitrogen and phosphorus are both not requested when trying to make a manure request.

I did end up leaving the events class as is. Let me know if either of you has strong feeling about moving them. I'm a little undecided.

RUFAS/data_structures/nutrient_request.py
```diff
@@ -42,3 +44,79 @@ def __post_init__(self) -> None:
...
+    """Fraction of manure that is dry matter, between 0 and 1 (unitless). Default to 0."""
+
+    def __post_init__(self) -> None:
+        """
+        Validate the dataclass fields.
+
+        Returns
+        -------
+        None
+
```
Isgotkowitz:
```suggestion
```

RUFAS/data_structures/nutrient_request.py
```diff
@@ -1,6 +1,8 @@
+from __future__ import annotations
 from dataclasses import dataclass, fields
 
 from RUFAS.routines.manure.manure_treatments.manure_types import ManureType
```
Isgotkowitz: `ManureType` needs to be housed in `data_structures` as well.

RUFAS/data_structures/nutrient_request_results.py
Isgotkowitz, line 1: This file should be deleted.

RUFAS/simulation_engine.py
```diff
@@ -142,6 +144,28 @@ def _daily_simulation(self) -> None:
...
         self._advance_time()
 
+    def generate_daily_manure_applications(self) -> dict[str, list[ManureEventNutrientRequestResults]]:
+        """Requests nutrients from the manure manager for each field in the simulation.
+
+        Returns
+        -------
+        dict[str, list[ManureEventNutrientRequestResults]]
+            A dictionary containing the ManureEvents and corresponding NutrientRequestResults for each field in
+            the simulation.
```
Isgotkowitz:
```suggestion

    def generate_daily_manure_applications(self) -> list[ManureEventNutrientRequestResults]:
        """Requests nutrients from the manure manager for each field in the simulation.

        Returns
        -------
       list[ManureEventNutrientRequestResults]
            List containing the ManureEvents and corresponding NutrientRequestResults to be applied to fields.
```

Isgotkowitz (commented):
This is looking better, but there are still a few things that need to be addressed. The first is the Nutrient Request and Nutrient Request Results have been combined, but old file for Nutrient Request Results is still present. I'm also curious if you are still going to move the request and request results into the manure to crop soil connection file. I think this is still the better place for these structures to go, but this isn't a blocking issue.

More importantly, I missed the `ManureType` in my first review. Because it is used in both Manure and CropSoil, it needs to be housed somewhere in `data_structures` as well.

ew3361zh:
Updates:
- `crop_and_soil_to_manure_connection.py` --> `manure_to_crop_and_soil_connection.py` (the manure request result is the actual data being passed, the request from c&s is only asking for the data so the name should reflect that flow).
- moved `ManureTypes` Enum into data_structures dir as it is used by both C&S and Manure
- Combined `ManureEventNutrientRequest`, `ManureEventNutrientRequestResults`, `NutrientRequest`, and `NutrientRequestResults` into the `manure_to_crop_and_soil_connection.py` file.

RUFAS/data_structures/manure_to_crop_soil_connection.py
```diff
@@ -1,7 +1,49 @@
...
-
-import math
+from typing import NamedTuple, Optional
 from dataclasses import dataclass, fields
+from RUFAS.data_structures.events import ManureEvent
+from RUFAS.data_structures.manure_types import ManureType
+import math
+
+
+@dataclass(kw_only=True, frozen=True)
```
allisterakun, line 8: Just curious, is there a specific restriction that we're imposing for freezing the dataclass object after initialization? If so, should we go through the codebase to see if any other data classes fall under this restriction?

RUFAS/routines/field/field/field.py
```diff
@@ -21,10 +25,8 @@
     TillageEvent,
 )
 from RUFAS.routines.field.soil.soil import Soil
-from RUFAS.routines.manure.manure_manager import ManureManager
-from RUFAS.routines.manure.manure_nutrients.nutrient_request import NutrientRequest
-from RUFAS.routines.manure.manure_nutrients.nutrient_request_results import NutrientRequestResults
-from RUFAS.routines.manure.manure_treatments.manure_types import ManureType
+from RUFAS.data_structures.manure_to_crop_soil_connection import NutrientRequest, NutrientRequestResults
+from RUFAS.data_structures.manure_types import ManureType
```
allisterakun: Maybe run the sorting action on the import statements

RUFAS/routines/field/manager/tillage_schedule.py
```diff
@@ -1,7 +1,7 @@
 from typing import List
 
 from RUFAS.routines.EEE.enums import TillageImplement
```
allisterakun, line 3: Just curious, why is this `TillageImplement` in EEE?

allisterakun (approved):
Looks much better now, great job, Niko!
Left two questions of curiosity, and one suggestion that maybe run the sorting action for the import statements given all the modifications to the imports

RUFAS/data_structures/manure_to_crop_soil_connection.py
ew3361zh: This was code Loi had originally written, maybe there was some concern with this data class getting updated/changed in the process when it shouldn't have been? I think it could be helpful if there are other dataclasses that we want to be immutable but I think it would be fairly low priority (just my opinion).

RUFAS/routines/field/field/field.py
ew3361zh: Good call!

RUFAS/routines/field/manager/tillage_schedule.py
ew3361zh: I believe this is because the main data RUFAS needs from TillageImplements is the costs associated with purchasing/maintaining/repairing them and the emissions they produce. The actual tillage process doesn't seem to depend on much other than the depth to which a particular implement reaches.

RUFAS/data_structures/manure_to_crop_soil_connection.py
Isgotkowitz: I agree that it should be low priority, but as we improve the codebase we should keep an eye out for more dataclasses where this should be applied.

RUFAS/routines/field/manager/tillage_schedule.py
Isgotkowitz: Nice catch, I didn't remember this at all. Tillage implements needs to be moved to `data_structures` as well for RuFaS to be disentangled.

RUFAS/data_structures/manure_to_crop_soil_connection.py
```diff
@@ -1,7 +1,49 @@
-from __future__ import annotations
-
-import math
+from typing import NamedTuple, Optional
 from dataclasses import dataclass, fields
+from RUFAS.data_structures.events import ManureEvent
+from RUFAS.data_structures.manure_types import ManureType
```
Isgotkowitz, line 4: Maybe `ManureType` should go in this file as well?

changelog.md
```diff
@@ -46,7 +46,7 @@ v0.9.2
 - [2048](https://github.com/RuminantFarmSystems/MASM/pull/2048) - [minor change] [Animal][Testing] Adds end to end testing for Animal domain.
 - [1809](https://github.com/RuminantFarmSystems/MASM/pull/1809) - [minor change] [InputManager] Extract the validation methods from InputManager.
 - [2045](https://github.com/RuminantFarmSystems/MASM/pull/2045) - [minor change] [Documentation] Regenerates the Sphinx HTML documentation.
-- [2050](https://github.com/RuminantFarmSystems/MASM/pull/2050) - [minor change] [Crop & Soil][Feed Storage] Disentangle the Crop & Soil and Feed Storage modules.
+- [2057](https://github.com/RuminantFarmSystems/MASM/pull/2057) - [minor change] [Crop & Soil][Manure] Disentangle the Crop & Soil and Manure modules.
```
Isgotkowitz, line 50:
```suggestion
- [2050](https://github.com/RuminantFarmSystems/MASM/pull/2050) - [minor change] [Crop & Soil][Feed Storage] Disentangle the Crop & Soil and Feed Storage modules.
- [2057](https://github.com/RuminantFarmSystems/MASM/pull/2057) - [minor change] [Crop & Soil][Manure] Disentangle the Crop & Soil and Manure modules.
```

Isgotkowitz (approved):
Changelog needs a small correction and I have a suggestion for `ManureType`, but other than that LGTM! Nice job Niko!

RUFAS/routines/field/manager/tillage_schedule.py
ew3361zh: I can write up an issue for disentangling EEE from other modules.

RUFAS/data_structures/manure_to_crop_soil_connection.py
ew3361zh: Hmm, I think I might leave it as is in its own file since it's used for things other than that communication between C&S and Manure (including a ton of stuff specifically in the Manure module - `manure_manager.py` - where it's tracking that information for different manure components which doesn't feel like it has much direct tie to C&S).

So I think because it has a good percent of use in contexts outside sharing information between C&S and Manure that it should stay in its own file.

But it may be something that makes more sense to move there in the future.

PR #2059, opened 2024-11-11
[Constants][Units][Crop & Soil] Updates Constants Used by C&S Module to GeneralConstants

### LINKED ISSUE #1954: [GeneralConstants] [Crop & Soil] Update All References in C&S to use GeneralConstants
Currently we have at least 2 sets of constants that are duplicative - the RUFAS/general_constants.py and the RUFAS/routines/field/crop_and_soil_constants.py (there is also a set of constants in the manure module).

## Expected Behavior

We should just be using one set of constants for the entire model.

## Current Behavior

We are using at least 2-3 sets of constants. Having multiple sets of constants is duplicative and also can lead to an issue of different values and different constant names and potential conflicts/inconsistencies. 

## Possible Solution

1. Migrate the constants in C&S to the GeneralConstants class.
2. Ensure there are no duplicates within GeneralConstants.
3. Update all references to the C&S Constants to the analagous GeneralConstant.
4. Manure module has a lot of specific manure-related constants. Would like to discuss with dev team whether they should go into GeneralConstants or stay in their own class but just be moved from where they are now.

ew3361zh:
Finds all the references in the Crop and Soil module to internal C&S constants and updates them to use the GeneralConstants class versions. Also moved some missing constants that were only present in the C&S Constants file to GeneralConstants.

### Context

Issue(s) closed by this pull request: closes #1954

### What

1. Updates all references to constants in C&S module to GeneralConstants.
2. Updates GeneralConstants with missing constant info that was only in C&S module constants file.
3. Updates the GeneralConstants class `CONSTANTS_TO_UNITS` dict to ensure there are units for all the constants in GeneralConstants class.
4. Updates MeasurementUnits to ensure there are units for all the constants in GeneralConstants `CONSTANTS_TO_UNITS` dict.

### Why

We had duplicate code and missing constants between these two files. It's important to be using one reference for all constants and to ensure there are standardized units for each of those constants.

### How

1. Cross referenced all constants between the two files and made sure all the C&S constants were present in or updated to GeneralConstants.
2. Searched references to each constant in the C&S constants file and updated those references to the GeneralConstant version.
3. Updated GeneralConstants Units conversion dict with new constants.
4. Added corresponding MeasurementUnits for the new constants added to GeneralConstants.
5. Updated unit testing for GeneralConstants and MeasurementUnits classes.

### Test plan

Updated unit testing.
Ran simulation and pytest suite successfully.


RUFAS/routines/field/field/field_data.py
```diff
@@ -1,8 +1,8 @@
 from dataclasses import dataclass
 from typing import Optional
 
+from RUFAS.general_constants import GeneralConstants
```
allisterakun, line 4: Just a thought about only importing the constants that we need (as before) instead of the entire GeneralConstants. I don't think it makes a huge performance difference, but maybe this is a better practice?

RUFAS/routines/field/field/manure_application.py
```diff
@@ -1,6 +1,6 @@
 from typing import Dict, Optional
 
-from RUFAS.routines.field.crop_and_soil_constants import KILOGRAMS_TO_GRAMS, SQUARE_CENTIMETERS_TO_HECTARES
+from RUFAS.general_constants import GeneralConstants
```
allisterakun, line 3: Same

tests/soil_crop_tests/soil_tests/carbon_cycling_tests/test_carbon_cycle.py
```diff
@@ -17,7 +14,9 @@
 )
 def test_determine_soil_volume(layer_thickness: float, field_size: float) -> None:
     """Checks that soil volume was calculated correctly"""
-    expected = (layer_thickness * field_size * HECTARES_TO_SQUARE_MILLIMETERS) * CUBIC_MILLIMETERS_TO_CUBIC_METERS
+    expected = (
```
allisterakun: As we discussed before, maybe make this are-calculated value, instead of calculating on the fly.

allisterakun: This is not probably outside the scope of this PR, but maybe one step at a time?

allisterakun (approved):
LGTM! Left some comments for maybe better practice.

RUFAS/routines/field/field/field_data.py
ew3361zh: Yea what do you think the cutoff number of constants used would be to import the whole class?

ew3361zh: I tried this and it doesn't seem to work to just import the constants we need (kept getting a PyLance can't resolve error)

tests/soil_crop_tests/soil_tests/carbon_cycling_tests/test_carbon_cycle.py
ew3361zh: I can do a few of these, it's pretty easy.

RUFAS/routines/field/field/manure_application.py
ew3361zh: Same issue here - not working to do an individual import for a single GeneralConstant.

matthew7838 (approved):
LGTM, have some thoughts on Allister's comments.

PR #2061, opened 2024-11-12
[DataValidator] Removing OM usage from the data validator

### LINKED ISSUE #2052: [DataValidator] Remove the use of OM in DataValidator
## Expected Behavior

The `DataValidator` should not be dependent on the import of `OutputManager`.

## Current Behavior

The `DataValidator` imports OM and uses OM to report errors in multiple places, this makes the results in circular imports of OM and prevents the usability of DataValidator in OM.

## Possible Solution

For every call to OM in the validator, we want to make a variable and save the content of those calls, just like `event_logs` in the report generator. The caller of the class is responsible for passing `event_logs`'s contents to the OM. Also, make methods non-static since we will now store data of the validator and need to create an instance to use it.

## Steps to Reproduce

1. Branch to extract-validation branch.
2. Find `data_validator .py`.
3.
4.

matthew7838:
Remove the usage of OM in Datavalidator to enable its use in OM.

### Context

Issue(s) closed by this pull request: closes #2052

### What

1. No more usage/imports of OM in the data validator.
2. logs/errors/warnings will now be saved in the instance variable called event_logs, the caller will have to deal with the om and the saved logs/errors/warnings.
3. Since we are now storing/modifying data for each instance of DV, methods will now be non-static.
4. Tests have been modified to reflect the changes. 

### Why

It prevents circular import when using the data validator in OM.

### How

Removed OM usage.

### Test plan

Run with default tasks and outputs are the same, and run and pass with modified tests, the amount of warnings/errors/logs remains the same. I would think that's promising enough for the content of this PR.
The test plan right now is something worth discussing, it is currently not that concrete due to a few reasons.
1. The original coverage is lacking, however, most methods' changes are covered and run with the test modification in this PR.  Should I cover them all in this PR? or should that be a separate issue(I personally think this is cleaner/better)?
2. I will not add tests for the route log method since it's already in OM it's exactly the same, and it should be extracted in the future.


RUFAS/data_validator.py
```diff
@@ -201,21 +199,24 @@ def validate_properties(metadata: Dict[str, Any], metadata_depth_limit: int) ->
...
 
             if depth > metadata_depth_limit:
-                om.add_error(
-                    "Max metadata depth exceeded.",
-                    f"Metadata depth exceeds maximum allowed depth of {metadata_depth_limit} at path {path}",
-                    info_map,
+                self.event_logs.append(
+                    {
+                        "error": "Max metadata depth exceeded.",
+                        "error message": f"Metadata depth exceeds maximum allowed depth"
```
allisterakun:
```suggestion
                        "error_message": f"Metadata depth exceeds maximum allowed depth"
```
For consistency with the rest of the codebase

RUFAS/data_validator.py
```diff
@@ -201,21 +199,24 @@ def validate_properties(metadata: Dict[str, Any], metadata_depth_limit: int) ->
...
+                        f" of {metadata_depth_limit} at path {path}",
+                        "info map": info_map,
```
allisterakun:
```suggestion
                        "info_map": info_map,
```

RUFAS/data_validator.py
```diff
@@ -1037,7 +1146,7 @@ def _string_type_validator(
                 f"Variable: '{variable_path_str}' has value: {data_value}, is type: "
                 f"{type(data_value)}. {properties_violation_message}"
             )
-            om.add_warning(warning_name, warning_message, info_map)
+            self.event_logs.append({"warning": warning_name, "warning message": warning_message, "info_map": info_map})
```
allisterakun: How is this key "info_map" and the rest are "info map"

RUFAS/input_manager.py
```diff
@@ -86,11 +87,11 @@ def start_data_processing(self, metadata_path: Path, eager_termination: bool = T
             True if data is valid, otherwise False.
         """
         self._load_metadata(metadata_path)
-        valid, message = DataValidator.validate_metadata(self.__metadata, VALID_INPUT_TYPES, ADDRESS_TO_INPUTS)
+        valid, message = self.data_validator.validate_metadata(self.__metadata, VALID_INPUT_TYPES, ADDRESS_TO_INPUTS)
         if not valid:
             raise ValueError(message)
         self._load_properties()
-        valid, message = DataValidator.validate_properties(self.__metadata, self.metadata_depth_limit)
+        valid, message = self.data_validator.validate_properties(self.__metadata, self.metadata_depth_limit)
```
allisterakun, line 94: Are you adding these messages to the OM at some point?

RUFAS/data_validator.py
```diff
@@ -179,8 +176,10 @@ def get_modifiable_at_runtime(cls) -> List["Modifiability"]:
 class DataValidator:
     """This class is will be utilized to validate all types of data across RuFas codebase."""
 
-    @staticmethod
-    def validate_properties(metadata: Dict[str, Any], metadata_depth_limit: int) -> Tuple[bool, str]:
+    def __init__(self) -> None:
+        self.event_logs: list[dict[str, str | dict[str, str]]] = []
```
allisterakun, line 180: I think it is worth to define a `TypedDict` here for the event_log entries.

allisterakun (changes requested):
Requested change on this PR because:
1. Inconsistent key names in the event log array. Some of them are using "info_map" while the others are "info map".
    1. I think it is worth to create a `TypedDict` for the event log dictionary entires to prevent the inconsistency.
    2. Also, I think we should use snake case for all keys in event log dictionary for consistency with the rest of the code base.
5. After returning all of the logs/warnings/errors, are you adding them to the OM somewhere? I don't think I see them being processed in IM

RUFAS/data_validator.py
matthew7838: good catch

RUFAS/input_manager.py
matthew7838: It's being raised in the error thrown. Those messages are stored in the dv instances and should be handled in the IM.

RUFAS/data_validator.py
matthew7838: This sounds like a good idea.

matthew7838:
> 2. After returning all of the logs/warnings/errors, are you adding them to the OM somewhere? I don't think I see them being processed in IM

Update: On second thought, I should modify this in this PR.

RUFAS/data_validator.py
matthew7838: After some try, I think this is not the best option since one of the keys is constantly changing.

RUFAS/data_validator.py
```diff
@@ -179,8 +176,10 @@ def get_modifiable_at_runtime(cls) -> List["Modifiability"]:
...
+
+    def validate_properties(self, metadata: Dict[str, Any], metadata_depth_limit: int) -> Tuple[bool, str]:
```
ew3361zh:
```suggestion
    def validate_properties(self, metadata: dict[str, Any], metadata_depth_limit: int) -> tuple[bool, str]:
```
I think it's preferred practice to use the native python lowercase type hints where possible (doesn't yet work with Any).

RUFAS/input_manager.py
```diff
@@ -1340,3 +1344,36 @@ def export_pool_to_csv(self, output_prefix: str, output_path: Path) -> None:
         except OSError as e:
             self.om.add_error("Save CSV failure.", f"Unable to save to {output_path} because of {e}.", info_map)
             raise e
+
+    def _route_logs(self, log_pool: list[dict[str, str | dict[str, str]]]) -> None:
```
ew3361zh, line 1348: I think the best thing to do would be to use the OM version of this function. I don't see a need to duplicate the code here. Did you try making that OM function public and calling it from IM?

RUFAS/input_manager.py
```diff
@@ -86,14 +87,15 @@ def start_data_processing(self, metadata_path: Path, eager_termination: bool = T
...
+        valid, message = self.data_validator.validate_metadata(self.__metadata, VALID_INPUT_TYPES, ADDRESS_TO_INPUTS)
         if not valid:
             raise ValueError(message)
         self._load_properties()
-        valid, message = DataValidator.validate_properties(self.__metadata, self.metadata_depth_limit)
+        valid, message = self.data_validator.validate_properties(self.__metadata, self.metadata_depth_limit)
         if not valid:
             raise ValueError(message)
         is_input_data_valid = self._populate_pool(eager_termination)
+        self._route_logs(self.data_validator.event_logs)
```
ew3361zh, line 99: Should this happen before the ValueError is raised so that the logs can still be added to the OM? I think the way we have things set up the logs/errors/warnings pool still gets dumped even with a runtime error so that would probably be helpful to have these in that case.

I see that populate_pool also calls on data_validator and you will want to avoid double-adding the logs so maybe route_logs is only called in the `if not valid:` block.

RUFAS/input_manager.py
```diff
@@ -1082,6 +1085,7 @@ def add_runtime_variable_to_pool(
                 properties_blob_key=properties_blob_key,
                 eager_termination=eager_termination,
             )
+            self._route_logs(self.data_validator.event_logs)
```
ew3361zh: Should route_logs() be called in the else statement here as well?

ew3361zh (commented):
Nice work, Matthew, this seems like it's a good start to this issue. I left a few comments that I think need to be addressed before approving.

One side-note - the `info_map` function reference on line 930 for `check_modifiability()` in IM is incorrect. It has `"function": self._add_variable_to_pool.__name__,`. Could you fix it here?

RUFAS/input_manager.py
matthew7838: Talked with Pooya about this, there will be an issue regarding this that will use the OM version for both the report generator and IM. This implementationwill be a short temporary place holder.

RUFAS/input_manager.py
matthew7838: Great suggestion and good catch.

RUFAS/input_manager.py
matthew7838: yes

RUFAS/data_validator.py
```diff
@@ -322,19 +341,23 @@ def _metadata_number_validator(key_path: list[str], value: dict[str, Any]) -> Tu
...
 
         if default is not None and not isinstance(default, (int, float)) and not has_no_default:
-            om.add_error(
-                "Invalid metadata default number value.",
-                f"Invalid 'default' for '{key_path}': Expected a number but got {type(default)}.",
-                info_map,
+            self.event_logs.append(
+                {
+                    "error": "Invalid metadata default number value.",
+                    "message": f"Invalid 'default' for '{key_path}': Expected a number but " f"got {type(default)}.",
```
Isgotkowitz:
```suggestion
                    "message": f"Invalid 'default' for '{key_path}': Expected a number but got {type(default)}.",
```

RUFAS/data_validator.py
```diff
@@ -322,19 +341,23 @@ def _metadata_number_validator(key_path: list[str], value: dict[str, Any]) -> Tu
...
 
         if default is None and not nullable:
-            om.add_error(
-                "Invalid metadata default number value.",
-                f"Invalid 'default' for '{key_path}': Value is not nullable and default is 'None'.",
-                info_map,
+            self.event_logs.append(
+                {
+                    "error": "Invalid metadata default number value.",
+                    "message": f"Invalid 'default' for '{key_path}': " f"Value is not nullable and default is 'None'.",
```
Isgotkowitz:
```suggestion
                    "message": f"Invalid 'default' for '{key_path}': value is not nullable and default is 'None'.",
```

RUFAS/data_validator.py
```diff
@@ -415,52 +449,65 @@ def _metadata_string_validator(key_path: list[str], value: dict[str, Any]) -> Tu
...
         nullable = value.get("nullable", False)
         if default is None and not nullable:
-            om.add_error(
-                "Invalid metadata default string value.",
-                f"Invalid 'default' for '{key_path}': Value is not nullable and default is 'None'",
-                info_map,
+            self.event_logs.append(
+                {
+                    "error": "Invalid metadata default string value.",
+                    "message": f"Invalid 'default' for '{key_path}': Value is not nullable and" f" default is 'None'",
```
Isgotkowitz:
```suggestion
                    "message": f"Invalid 'default' for '{key_path}': Value is not nullable and default is 'None'",
```

RUFAS/data_validator.py
```diff
@@ -415,52 +449,65 @@ def _metadata_string_validator(key_path: list[str], value: dict[str, Any]) -> Tu
...
 
         if default is not None and not has_no_default and not isinstance(default, str):
-            om.add_error(
-                "Invalid metadata default string value.",
-                f"Invalid 'default' for '{key_path}': Expected a string but got {type(default)}",
-                info_map,
+            self.event_logs.append(
+                {
+                    "error": "Invalid metadata default string value.",
+                    "message": f"Invalid 'default' for '{key_path}': Expected a string but" f" got {type(default)}",
```
Isgotkowitz:
```suggestion
                    "message": f"Invalid 'default' for '{key_path}': Expected a string but got {type(default)}",
```

RUFAS/data_validator.py
```diff
@@ -415,52 +449,65 @@ def _metadata_string_validator(key_path: list[str], value: dict[str, Any]) -> Tu
...
         pattern = value.get("pattern")
         if pattern is not None and not isinstance(pattern, str):
-            om.add_error(
-                "Invalid metadata string properties pattern.",
-                f"Invalid 'pattern' for '{key_path}': Expected a string but got {type(pattern)}",
-                info_map,
+            self.event_logs.append(
+                {
+                    "error": "Invalid metadata string properties pattern.",
+                    "message": f"Invalid 'pattern' for '{key_path}': Expected a string but got" f" {type(pattern)}",
```
Isgotkowitz:
```suggestion
                    "message": f"Invalid 'pattern' for '{key_path}': Expected a string but got {type(pattern)}",
```

RUFAS/data_validator.py
```diff
@@ -489,67 +534,78 @@ def _metadata_bool_validator(key_path: list[str], value: dict[str, Any]) -> Tupl
...
         nullable = value.get("nullable", False)
         if default is None and not nullable:
-            om.add_error(
-                "Invalid metadata default bool value.",
-                f"Invalid 'default' for '{key_path}': Value is not nullable and default is 'None'",
-                info_map,
+            self.event_logs.append(
+                {
+                    "error": "Invalid metadata default bool value.",
+                    "message": f"Invalid 'default' for '{key_path}': Value is not nullable and" f" default is 'None'",
```
Isgotkowitz:
```suggestion
                    "message": f"Invalid 'default' for '{key_path}': Value is not nullable and default is 'None'",
```

RUFAS/data_validator.py
```diff
@@ -489,67 +534,78 @@ def _metadata_bool_validator(key_path: list[str], value: dict[str, Any]) -> Tupl
...
             return False, error_message
         if default is not None and not isinstance(default, bool) and not has_no_default:
-            om.add_error(
-                "Invalid metadata default bool value.",
-                f"Invalid 'default' for '{key_path}': Expected a bool but got {type(default)}",
-                info_map,
+            self.event_logs.append(
+                {
+                    "error": "Invalid metadata default bool value.",
+                    "message": f"Invalid 'default' for '{key_path}': Expected a bool but got" f" {type(default)}",
```
Isgotkowitz:
```suggestion
                    "message": f"Invalid 'default' for '{key_path}': Expected a bool but got {type(default)}",
```

RUFAS/input_manager.py
```diff
@@ -1340,3 +1346,36 @@ def export_pool_to_csv(self, output_prefix: str, output_path: Path) -> None:
...
+                    and isinstance(log["info_map"], dict)
+                ):
+                    self.om.add_log(log["log"], log["message"], log["info_map"])
+            elif "warning" in log:
+                if (
+                    isinstance(log["warning"], str)
+                    and isinstance(log["message"], str)
+                    and isinstance(log["info_map"], dict)
+                ):
+                    self.om.add_warning(log["warning"], log["message"], log["info_map"])
```
Isgotkowitz, line 1379: It would be good to check for the case when the log is not a log, warning, or error. Also if any error, warning, or log doesn't have the correct keyset.

RUFAS/input_manager.py
```diff
@@ -1340,3 +1346,36 @@ def export_pool_to_csv(self, output_prefix: str, output_path: Path) -> None:
...
+            A list of log, warning, and error dictionaries containing all the components needed
+            to log the information to the appropriate pool.
+        """
+        for log in log_pool:
+            if "error" in log:
+                if (
+                    isinstance(log["error"], str)
+                    and isinstance(log["message"], str)
+                    and isinstance(log["info_map"], dict)
+                ):
```
Isgotkowitz, line 1364:
```suggestion
                is_correct_type: bool = isinstance(log["error"], str) and isinstance(log["message"], str) and isinstance(log["info_map"], dict)
                is_keyset_correct: bool = log.keys() == ["error", "message", "info_map"]
                if is_correct_type and is_keyset_correct:
```

README.md
```diff
@@ -1,6 +1,6 @@
 [![Flake8](https://img.shields.io/badge/Flake8-passed-brightgreen)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
 [![Pytest](https://img.shields.io/badge/Pytest-passed-brightgreen)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
-[![Coverage](https://img.shields.io/badge/Coverage-96%25-brightgreen)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
+[![Coverage](https://img.shields.io/badge/Coverage-95%25-brightgreen)](https://github.com/RuminantFarmSystems/MASM/actions/workflows/combined_format_lint_test_mypy.yml)
```
Isgotkowitz, line 3: I believe this drop in coverage is a result of a drop in coverage of the Input Manager. Please update the test suite so that it maintains the same coverage level.

matthew7838: This is due to the routing method not being tested, it's the exact same as the one in OM. There will be a separate issue that cleans them and make use of the one in OM across the codebase regarding these log/error/warning routing. I also agree with your above comments and I'll address them in that issue/PR.

Isgotkowitz: Can you please write up that issue and mention this PR in it?

Isgotkowitz (approved):
LGTM, thank you for responding to FB Matthew.

README.md
matthew7838: Will do!

ew3361zh:
I think this looks good - I'm a little concerned that I'm getting pretty different numbers of warnings and logs between dev and this branch. Maybe it's just what I merged into dev this morning (the C&S-Manure disentanglement PR). But I will also test out a few other scenarios where we'd expect the same number but after intentionally erroring out.

dev:
Screenshot 2024-11-19 at 10 15 33 AM

om-remove branch:
Screenshot 2024-11-19 at 10 16 00 AM

ew3361zh (approved):
Nice work, Matthew - counts of warnings and logs are now matching on this branch and dev. Everything else looks good as well!

allisterakun (approved):
LGTM! Nice work!

PR #2063, opened 2024-11-14
[Feed Storage] Setting up bale diameter input

### LINKED ISSUE #1960: [Feed Storage] Determine how bale size should be set and implement it
The way that bale size is determined for a hay bale is not specified in the design doc, and the code currently uses a default value as a temporary solution.

## Status

The Feed Storage design doc specifies:

Hay has the following attributes:
...
- Bale_size
  - (enum) diameter of the bale in meters
    - 1.2, 1.5, 1.8

It however does not specify how this bale size should be calculated or determined.

## Intention

Find the methodology for how the bale size is set so that it can be set properly, instead of using a default.

## Request

Read the Feed Storage design doc, communicate with the original authors as need be, and either verify or determine the methodology for how bale size should be set.

Also, if determined to be appropriate, have `bale_size` return an enum element instead of a float.

matthew7838:
Adds the ability for the user to input for bale diameter.
### Context

Issue(s) closed by this pull request: closes #1960

### What

Added an input file for the bale size, and its corresponding properties.

### Why

The user can now input their own data of bale size by having the data in the `input/data/feed_management` folder.

### How

- Removed the hard-coded value of bale size in hay.py
- Added a bale size input with the originally removed hardcoded value.
- Added the file to the metadata.
- Added properties of bale size to the default property.
- Updated unit tests.

### Test plan

- Run with updated unit tests.
- **Intentionally** left a print statement in the constructor of hay.py. Run the simulation with default metadata but use ARL_rotation for the crop. The value should be printed out upon Hay initialization. I will the print before merge.


RUFAS/routines/feed_storage/hay.py
```diff
@@ -5,6 +5,7 @@
 from RUFAS.weather import Weather
 
 from .storage import Storage
+from ...input_manager import InputManager
```
Isgotkowitz:
```suggestion
from RUFAS.input_manager import InputManager
```
Relative imports should only be used to import modules from the same directory.

RUFAS/routines/feed_storage/hay.py
```diff
@@ -46,6 +47,9 @@ class Hay(Storage):
     """
 
     def __init__(self, capacity: float = float("inf")) -> None:
+        im = InputManager()
+        self.bale_size: float = im.get_data("feed_management.bale_diameter")
+        print(self.bale_size)
```
Isgotkowitz:
```suggestion
```

RUFAS/routines/feed_storage/hay.py
```diff
@@ -46,6 +47,9 @@ class Hay(Storage):
     """
 
     def __init__(self, capacity: float = float("inf")) -> None:
+        im = InputManager()
+        self.bale_size: float = im.get_data("feed_management.bale_diameter")
```
Isgotkowitz: Why not put this in the argument to the constructor? That way the Input Manager can be queried for this data once in a simulation and the bale size held in `FeedManager` or something, instead of calling the IM every time a new Hay is initiated.

Another way to go would be to make the bale size a class constant, but I think that option would not leave as open to extension.

changelog.md
```diff
@@ -20,6 +20,7 @@ A **changelog** is a structured record of changes made to the codebase over time
 ## Changelog Entries
 
 ### Current version
+- [1945](https://github.com/RuminantFarmSystems/MASM/pull/1945) - [minor change] [Animal] Implements Animal Genetics submodule.
```
```suggestion
```

changelog.md
```diff
@@ -52,6 +53,7 @@ v0.9.2
 - [2057](https://github.com/RuminantFarmSystems/MASM/pull/2057) - [minor change] [Crop & Soil][Manure] Disentangle the Crop & Soil and Manure modules.
 - [2059](https://github.com/RuminantFarmSystems/MASM/pull/2059) - [minor change] [Crop & Soil][GeneralConstants] Update references to constants in Crop and Soil module to use GeneralConstants.
 - [1866](https://github.com/RuminantFarmSystems/MASM/pull/1866) - [minor change] [Metadata] Adds minimums for some Animal module inputs in metadata.
+- [2063](https://github.com/RuminantFarmSystems/MASM/pull/1809) - [minor change] [Feed] Enable user input for bale size (diameter).
```
```suggestion
- [2063](https://github.com/RuminantFarmSystems/MASM/pull/1809) - [minor change] [Feed] Enable users to specify the diameter of hay bales.
```

tests/test_feed_storage_module/test_hay.py
```diff
@@ -165,14 +165,17 @@ def test_calculate_additional_dry_matter_loss(
     expected: float,
 ) -> None:
     """Tests _calculate_additional_dry_matter_loss in Hay."""
+    im = InputManager()
+    mocker.patch.object(im, "get_data", return_value=size)
+    hay = Hay()
```
Isgotkowitz, line 170: Instead of changing this test and making it inconsistent with the rest of Feed Storage test suite, I suggest just setting the bale size directly.
```suggestion
    hay.bale_size = size
```

RUFAS/routines/feed_storage/hay.py
```diff
@@ -5,6 +5,7 @@
 from RUFAS.weather import Weather
 
 from .storage import Storage
+from ...input_manager import InputManager
 
 """
 This final moisture percentage that expected to be contained in a hay crop. References Feed Storage Scientific
```
Isgotkowitz, line 11: Please remove `DEFAULT_BALE_DIAMETER` and the associated TODO

input/data/feed_management/bale_size.json
```diff
@@ -0,0 +1,3 @@
+{
+  "bale_diameter": 1.5
```
Isgotkowitz:
```suggestion
  "hay_bale_diameter": 1.5
```
I recommend making this more specific, because it applies only to hay and not to baleage

input/data/feed_management/bale_size.json
Isgotkowitz, line 1: Could you please make this filename more generic, and inline with the other default input files, something like `default_feed_management.json`?

input/metadata/properties/default.json
```diff
@@ -5842,5 +5842,14 @@
         "type": "number"
       }
     }
+  },
+  "feed_storage_input_properties": {
+    "bale_diameter": {
+        "type": "number",
+        "description": "Diameter of the bale.",
```
```suggestion
        "description": "Diameter of hay bales (meters).",
```

input/metadata/properties/default.json
```diff
@@ -5842,5 +5842,14 @@
...
+        "minimum": 0,
```
Isgotkowitz: I don't think 0 is an acceptable minimum. @KevinP-B what is the smallest bale diameter you have seen or heard of?

Isgotkowitz (commented):
This is a good start Matthew, but there are a number of things that either need to be cleaned up or improved before this PR is ready to be merged.

input/metadata/properties/default.json
matthew7838: After discussion, the new min and max will be 0.5 and 4.

RUFAS/routines/feed_storage/hay.py
matthew7838: The bale size should not be a class constant; room for extension is suggested. For this comment's first part, I think leaving the data extraction here for now is better. Perhaps we can come back to how we initialize different types of storage in the feed manager when other storages also implemented their user-defined inputs.

changelog.md
```diff
@@ -53,6 +53,7 @@ v0.9.2
 - [2057](https://github.com/RuminantFarmSystems/MASM/pull/2057) - [minor change] [Crop & Soil][Manure] Disentangle the Crop & Soil and Manure modules.
 - [2059](https://github.com/RuminantFarmSystems/MASM/pull/2059) - [minor change] [Crop & Soil][GeneralConstants] Update references to constants in Crop and Soil module to use GeneralConstants.
 - [1866](https://github.com/RuminantFarmSystems/MASM/pull/1866) - [minor change] [Metadata] Adds minimums for some Animal module inputs in metadata.
+- [2063](https://github.com/RuminantFarmSystems/MASM/pull/1809) - [minor change] [Feed] Enable users to specify the diameter of hay bales.
```
```suggestion
- [2063](https://github.com/RuminantFarmSystems/MASM/pull/2063) - [minor change] [Feed] Enable users to specify the diameter of hay bales.
```

input/metadata/default_metadata.json
```diff
@@ -45,7 +45,7 @@
     "crop": {
       "title": "Crop data",
       "description": "Crop selection and detailed rotation schedules.",
-      "path": "input/data/crop/default_rotation.json",
+      "path": "input/data/crop/ARL_rotation.json",
```
Isgotkowitz: Does this need to be reverted before merging?

input/metadata/properties/default.json
```diff
@@ -5842,5 +5842,14 @@
...
       }
     }
+  },
+  "feed_storage_input_properties": {
+    "hay_bale_diameter": {
+        "type": "number",
+        "description": "Diameter of hay bales (meters).",
+        "minimum": 0.5,
+        "maximum": 4,
+        "default": 2.5
```
Isgotkowitz, line 5852: How did you decide to have the the diameter default to 2.5?

Isgotkowitz (approved):
Thank you for addressing feedback. Found a few minor things that need to be fixed before merging, but otherwise looks good to me.

input/metadata/default_metadata.json
matthew7838: Yes, forgot to mention that, thanks, Ed.

input/metadata/properties/default.json
matthew7838: Kevin

input/data/feed_management/default_feed_management.json
```diff
@@ -0,0 +1,3 @@
+{
+  "hay_bale_diameter": 1.5
```
ew3361zh, line 2: Any reason why this shouldn't match the default in the metadata properties (2.5)?

tests/test_feed_storage_module/test_hay.py
ew3361zh: Curious about this as well.

ew3361zh (approved):
Looks good! I had one question about the default matching the metadata properties default value and then a follow up to make sure Ed's comment in the test section was addressed but otherwise good to go.

input/data/feed_management/default_feed_management.json
matthew7838: It's intended to keep the code running in its original state, the primary focus for now is only to allow user input availability, further value reevaluation can happen when SME's think it's needed.

PR #2064, opened 2024-11-14
[Task Manager][Docker] Add system resource usage monitoring to `TaskManager`

Isgotkowitz:
`TaskManager` will monitor the system resources (CPU, memory) being used by RuFaS when it is running inside a Docker container. If the system resource usage indicates something went wrong with RuFaS, `TaskManager` will stop all simulations and raise an error.

### Context

Issue(s) closed by this pull request: closes # n/a

Notes to reviewers:
- There are still a number of TODOs in this PR. I would like any and all input on these TODOs.

### What

When RuFaS is run inside a Docker container, `TaskManager` will monitor how much CPU and memory are being used. If the usage is abnormally low over a length of time then `TaskManager` will stop all the simulations that are running and raise an error.

### Why

The Vyla-FARMES team found that RuFaS running a certain set of simulations inside of a Docker container would lead to something going wrong with RuFaS, and the simulations would basically stop running. When this happened there was no way that RuFaS could stop itself and indicate that something was wrong to the container. This means that RuFaS would sit idle until the container "timed out" (in this case it ran for 24 hours and then would turn itself off).

### How

When `TaskManager` is initialized, it attempts to find whether or not it is running inside of a Docker container. If it is, then it communicates with the Docker daemon to find which specific container is running RuFaS and saves that container's ID. RuFaS uses the `docker` package to do this, so this package has been added to `requirements.txt`.

In order for `TaskManager` to find the right Docker container, it uses an environment variables that is set in the docker-compose.yml file used to deploy the container. This requires

When `TaskManager` starts executing the tasks it was provided in `_run_tasks`, a new thread is spawned if RuFaS is in a Docker container. This new thread runs the `_monitor_resources_for_stall` method, which checks the container's resource usage periodically. If the resource usage is below a certain threshold and has been below that threshold several times in a row then `TaskManager` considers RuFaS to be stalled. When this happens, `TaskManager` raises an error and stops all simulations.

### Test plan

To test that this fix properly detects stalls, it has been run in the docker container the Vyla team uses to test RuFaS to ensure that it properly detects stalls and halts simulations.

To test that the changes in this PR do not change the model's outputs or have a significant impact on performance, the results between this branch and `main` have been compared to make sure that they are the same. Also, the runtime of RuFaS simulations on this branch and `main` have been compared to ensure that they are reasonably close.

Runtime of a single default simulation:
|Environment\Resource Usage|Monitored|Unmonitored|
|--|--|--|
|**Local**                              |57 sec|58 sec|
|**Containerized**                |1 min 43 sec|1 min 33 sec|

This test was very barebones! IMO more thorough comparisons should be done

-- Detailed test conditions and results in progress --

### DIFF
--- a/RUFAS/task_manager.py
+++ b/RUFAS/task_manager.py
@@ -3,13 +3,20 @@ from functools import partial
 import multiprocessing
 import numpy
 from pathlib import Path
+from threading import Thread, Event
 import random
+from time import sleep
+from psutil import Process
+import os
+import docker
+from re import match
 from SALib.sample import ff as fractional_factorial_sampler
 from SALib.sample import saltelli as saltelli_sampler
 import traceback
 from typing import Any, Dict, List, Tuple, Callable
 
 from RUFAS.e2e_test_results_comparer import E2ETestResultsComparer
+from RUFAS.general_constants import GeneralConstants
 from RUFAS.input_manager import InputManager
 from RUFAS.output_manager import OutputManager, LogVerbosity
 from RUFAS.routines.animal.life_cycle.herd_factory import HerdFactory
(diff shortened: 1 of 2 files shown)


Isgotkowitz:
Update! Found a bug when running with more tasks than workers, will start debugging now.

Isgotkowitz:
Closing this PR after discussion with @PooyaHekmati. Instead of trying to monitor container properties at the level of RuFaS (which is inside the container), the container properties should be managed with capabilities that Docker provides.

PR #2066, opened 2024-11-14
[OutputManager] Improve data origins handling and reporting

### LINKED ISSUE #1433: [OM] Improve the handling of data origins info part 2

ew3361zh:
Adds handling in OM for data origins info and enables user control over the degrees to which that data is reported out in JSON output filters.

### Context

Issue(s) closed by this pull request: closes #1433

### What

Allows for user-specified reporting of data-origins info for variables added to OutputManager.
Updates formatting and handling of that data-origins data.
As noted in the PR description, this only applies to JSONs - data-origins info isn't reported out in graphs, csvs, or reports (and wouldn't necessarily make sense to do so in any of those formats).
The key vocabulary words are:
- `detailed_values` - where the information about the data origins will be located in the JSON output file.
- `True Origin` - this is the function which originally sent the data to its intermediary data reporter class. Helpful to know where the variable was originally created/modified/added.
- `Report Origin` - this is the name of the reporting class that actually sent the variable data to OutputManager (i.e. the intermediary class and function like `AnimalModuleReporter.report_ration_interval()`).

Within each report filter, the user can select one of the following to set the "origin_label" to:
1. "true and report origins" - Indicates that both the true origin and report origin should be included.
2. "true origin" - Indicates that only the true origin should be included.
3. "report origin" - Indicates that only the report origin should be included.
4. "none" - (default) Indicates that no origin information should be included.

**NOTE: The issue originally mentioned decoupling data-origins from the `exclude_info_maps` flag and possibly the `values` portion of variable data being added to OM. Neither of these seemed practical nor intuitive/logical as far as using this feature or what the expectation is when opting for `exclude_info_maps` to be `true`.**

Therefore, if a user marks `exclude_info_maps` as `true` in their task, this feature will not be available.

### Why

The origins-data was previously added to the OM but was not well-handled. This PR adds more user-control over reporting this data out.

### How

- Adds Enum OriginLabel in OM which specifies the different levels of reporting data-origins data out.
- Removes command-line arg for control over turning this feature off and on (replaced with an Enum that allows for greater specificity between individual reports from the same simulation).
- Updates formatting and addition of the data-origins data by OM based on the setting specified by the user in their JSON output filter.

    
### Test plan

**NOTE: currently data origins info is only added for a select group of variables in the Animal module. This is not intentional, this data origins info will eventually need to be added for the other parts of the model. Setting up the capacity for OM to handle data-origins info properly is the focus of this PR.**

To test, I created a test filter that showed all the different options that can be specified by OriginLabel and compared the results to ensure it was mapping properly with how that data-origins info is added to the info maps of those variables. Here is my filter but you can also do a project-wide search for `data_origins` being present in a variable's info-map and create a different filter to test further if you like.

Reminders: 
1. This only works with JSON output files so your test filter should start with `json_`.
2. You're looking at `detailed_values` output in that output file.

```
{
    "multiple": [
                {
                    "name": "Animal None",
                    "filters": [
                        "AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_.*"
                    ],
                    "origin_label": "none"
                    },
                    {
                        "name": "Animal True and Report Origins",
                        "filters": [
                            "AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_.*"
                        ],
                        "origin_label": "true and report origins"
                        },
                        {
                            "name": "Animal True Origin",
                            "filters": [
                                "AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_.*"
                            ],
                            "origin_label": "true origin"
                            },
                            {
                                "name": "Animal Report Origin",
                                "filters": [
                                    "AnimalModuleReporter.report_ration_interval_data.ration_nutrient_amount_pen_.*"
                                ],
                                "origin_label": "report origin"
                                }
    ]
}
```


RUFAS/output_manager.py
```diff
@@ -236,7 +254,7 @@ def setup_pool_overflow_control(
             f"{self.available_memory} Bytes.\n"
         )
 
-        if save_chunk_threshold_call_count > 0:
+        if save_chunk_threshold_call_count and save_chunk_threshold_call_count > 0:
```
ew3361zh: Addresses mypy error.

RUFAS/routines/animal/animal_module_reporter.py
```diff
@@ -275,6 +275,7 @@ def report_ration_interval_data(cls, pen: Pen, feed: Feed, simulation_day: int)
                 "DMIest_requirement": MeasurementUnits.KILOGRAMS,
                 "avg_BW": MeasurementUnits.KILOGRAMS,
                 "avg_milk_production_reduction_pen": MeasurementUnits.KILOGRAMS_PER_ANIMAL,
+                "avg_essential_amino_acid_requirement": MeasurementUnits.GRAMS_PER_DAY,
```
ew3361zh, line 278: These units were not added for this variable prior and this was discovered in this PR because it was causing an error where the data origins code was looking for the corresponding units.

RUFAS/output_manager.py
```diff
@@ -336,7 +354,9 @@ def add_variable(self, name: str, value: Any, info_map: Dict[str, Any], first_in
         self.add_variable_call += 1
         units = info_map.get("units")
         if units is None:
-            raise KeyError("'units' was not found in info_map for call to 'add_variable()'")
+            raise KeyError(f"'units' was not found in info_map for call to 'add_variable()' for {name}.")
+        if isinstance(units, dict) and len(info_map.get("units")) != len(value) and value != {}:
```
ew3361zh, line 360: This code is added here to ensure there are units added for nested variables added to OM pool. Previously a variable without units could sneak in without raising an error if it was nested within a dictionary.

RUFAS/output_manager.py
```diff
@@ -73,6 +73,32 @@ def __str__(self) -> str:
         return self.value[:-1].upper()
 
 
+class OriginLabel(Enum):
+    """
+    An enumeration representing the different labels for data origins when generating json output files.
+    Attributes
```
```suggestion
    An enumeration representing the different labels for data origins when generating json output files.

    Attributes
```

RUFAS/output_manager.py
```diff
@@ -631,6 +662,9 @@ def dict_to_file_json(self, data_dict: Dict[str, Any], path: Path, minify_output
         minify_output_file : bool
             Boolean flag indicating whether to minify the output JSON file.
 
+        origin_label : OriginLabel
```
```suggestion
        origin_label : OriginLabel, default OriginLabel.None
```

RUFAS/output_manager.py
```diff
@@ -1178,8 +1338,8 @@ def _parse_filtered_variables(
 
         Returns
         -------
-        Dict[str, OutputManager.pool_element_type]
-            Dictionary containing data from the filtered pool of data, with data from within dictionaries unpacked and
+        dict[str, OutputManager.pool_element_type]
+            dictionary containing data from the filtered pool of data, with data from within dictionaries unpacked and
```
```suggestion
            Dictionary containing data from the filtered pool of data, with data from within dictionaries unpacked and
```

RUFAS/output_manager.py
```diff
@@ -73,6 +73,32 @@ def __str__(self) -> str:
...
+        Indicates that only the true origin should be included.
+    REPORT_ORIGIN : str
+        Indicates that only the report origin should be included.
+    NONE : str
+        Indicates that no origin information should be included.
+
+    Notes
+    -----
+    NONE is the default setting.
+
```
```suggestion
```
I suggest removing this because NONE isn't actually the default in the context of this Enum, it is only the default in contexts (methods, classes, etc.) which use it as a default. And those are the places where it should be noted explicitly as a default.

Because it is used as the default in more than one context, one way to implement it as a default that is really obvious and consistent is to declare a module constant `DEFAULT_ORIGIN_LABEL = OriginLabel.NONE`, and use that module constant everywhere in OM.

RUFAS/output_manager.py
Isgotkowitz:
```suggestion
        if save_chunk_threshold_call_count is not None and save_chunk_threshold_call_count > 0:
```
I think it would be better to make this condition a little more explicit, I had to stop and think through it for a sec before I understood what it was doing.

RUFAS/output_manager.py
```diff
@@ -708,51 +757,154 @@ def _add_detailed_values(self, data_dict: Dict[str, Any]) -> Dict[str, Any]:
...
+        Parameters
+        ----------
+        origin_label : OriginLabel
+            The origin label specifying the format of the detailed values string.
+            It can be one of the following:
+            - OriginLabel.TRUE_AND_REPORT_ORIGINS: Include both true origin and report origin.
+            - OriginLabel.TRUE_ORIGIN: Include only the true origin.
+            - OriginLabel.REPORT_ORIGIN: Include only the report origin.
+            - OriginLabel.NONE: Include no origin information.
```
```suggestion
```
I don't think it is necessary to list all the options of OriginLabel as an input, I think it is clear enough that it can be any element of that Enum

Isgotkowitz (commented):
I think the code looks great. ~However I couldn't figure out how to make the feature work -- I changed `include_default_values` to true in the default config file, and all four outputs from the example filter you included were exactly the same. Did I miss a step?~ Edit: I was on the wrong branch 😭 my bad

Isgotkowitz (approved):
Works as expected, nice job Niko.

RUFAS/output_manager.py
ew3361zh: I like the idea of a module constant for that default but as I started looking into implementing it, I feel like it's not necessarily interchangeable everywhere we're using OriginLabel.NONE in OM and I'm not sure adds clarity to what's happening at each of those places. I will remove the NOTES section though, that makes sense.

Isgotkowitz: Sounds good

tests/animal_module_tests/test_animal_reporter.py
```diff
@@ -277,34 +277,18 @@ def test_report_ration_interval_data(animal_manager_fixture: AnimalManager, mock
...
-
-        assert om.variables_pool[
-            f"AnimalModuleReporter.report_ration_interval_data.ration_per_animal_for_pen_{i}_combo{i}"
-        ]["values"] == [test_data["formatted_ration"]]
-
-        assert om.variables_pool[
-            f"AnimalModuleReporter.report_ration_interval_data.ration_supply_report_for_pen_{i}_combo{i}"
-        ]["values"] == ["ration_supply_report"]
+    mock_ration_report.assert_called()
+    mock_add_variable.assert_called()
```
matthew7838: No need to go into full detail, but maybe testing a bit more information like the times it's called?

tests/animal_module_tests/test_animal_reporter.py
```diff
@@ -447,35 +431,13 @@ def test_report_life_cycle_manager_data(mocker: MockerFixture) -> None:
...
-    assert om.variables_pool["AnimalModuleReporter.report_life_cycle_manager_data.num_cow_for_parity_greater_than_3"][
-        "values"
-    ] == [400]
-    assert om.variables_pool["AnimalModuleReporter.report_life_cycle_manager_data.calving_to_preg_time_greater_than_3"][
-        "values"
-    ] == [400]
-    assert om.variables_pool["AnimalModuleReporter.report_life_cycle_manager_data.avg_age_for_calving_greater_than_3"][
-        "values"
-    ] == [400]
+    mock_add_variable.assert_called()
```
matthew7838: As above.

tests/animal_module_tests/test_animal_reporter.py
```diff
@@ -447,35 +431,13 @@ def test_report_life_cycle_manager_data(mocker: MockerFixture) -> None:
...
 
 
 # Test cases
```
matthew7838:
```suggestion
```

tests/test_output_manager.py
```diff
@@ -3326,3 +3345,38 @@ def test_setup_pool_overflow_control_user_define_max_memory_usage_percentage(
...
+        # Case 4: Valid 'origin_label' value
+        ({"origin_label": "true and report origins"}, OriginLabel("true and report origins"), False),
+    ],
+)
+def test_get_origin_label(
+    filter_content: dict[str, Any], expected_label: OriginLabel, expected_error: bool, mocker: MockerFixture
+) -> None:
+    """
+    Unit test for the _get_origin_label() method in OutputManager class.
+    """
```
matthew7838, line 3368: More of a preference/suggestion. I would sperate testing failing and successful cases into two different methods.

matthew7838 (approved):
Great PR, Niko. The changes and functionality look fine and are working on my side.  Left a few minor suggestions.

tests/animal_module_tests/test_animal_reporter.py
ew3361zh: Good call!

tests/test_output_manager.py
ew3361zh: I agree with you. I got some pushback on something similar before but maybe we should revisit it as a dev team and discuss a unified practice going forward.

PR #2067, opened 2024-11-14
Interpolated emissions

JoeWaddell:
### Context

- [ ] Yes
- [x] No

### What

In short: we put together an algorithm to interpolate missing values (e.g., 0s) by using the surrounding FIPS codes (limited to the same state).

New files on this branch to merge to `dev`:

`input/data/EEE/full_feeds_land_use_change_emissions_July2024_interpolated.csv`
`input/data/EEE/full_feeds_land_use_change_emissions_July2024_interpolated_regional_average.csv`

`input/data/EEE/full_feeds_emissions_July2024_interpolated.csv`
`input/data/EEE/full_feeds_emissions_July2024_interpolated_regional_average.csv`

Note there are only two new suffixes, which indicate which files they amend. 
`_interpolated` indicates we used the algorithm as described in the "How" section below, though this still has the places where regional values weren't available, leaving more 0s. Note: this is now only found in Basecamp: https://3.basecamp.com/3486446/buckets/5296287/vaults/8102308869
`_interpolated_regional_average` uses the methodology described in the "Additional questions we resolved" section below to fill the gaps in the `_interpolated` sheet.
   The two `interpolated_regional_average` sheets have replaced the default emissions sheets in all metadatas.

I've included all iterations for posterity, to ensure the user is using precisely the dataset they'd like. However, if we think this is unnecessary, we can archive the ones we don't want to use on basecamp.

### Why

The feed emissions csv uses data from 2017 census data, meaning some regions were not evaluated because they did not have dairy production. As time goes on, more of those gaps will likely need to be filled.

### How

For each feed, for each FIPS code, if there is a 0 for a value, we look for the FIPS codes in the surrounding 50km within the same state, and then report the average value. When looking for neighbor values, we don’t include A) those with 0s, or B) these newly created “interpolated” values (so there’s no averaging averages, which could lead to runaway homogenization of missing data). 
    If there are no values within 50km: we expand the search radius by 50km, and repeat as necessary, until all FIPS codes in the state are exhausted. We have a second report to show which feed/FIPS combinations were 0s, and the distances + values used to compute the interpolated average.

The code used is now found in `helpful_scripts/emissions_interpolation`

### Additional questions we resolved
For feeds in which no value is found in a state: we used a regional average, so there is now a second set of new sheets, with regional averages (note: region-wide 0s meant that we had to lump both New England and Great Lakes into Northeast, and Upper Midwest as equivalent to Northern Plains).

Other considerations were A) use the algorithm without restricting the results to the state boundaries, or B) cross that bridge when we come to it on a case by case basis.

### Test plan

Kristan and I have been checking the new files, as well as those below, which document where the changes were made, and the distances and values used in the averages for each interpolated value.

The files below can also be found archived on Basecamp: 
https://3.basecamp.com/3486446/buckets/5296287/vaults/8102308869
[full_feeds_emissions_July2024_nozeroes_detailed.csv](https://github.com/user-attachments/files/17757286/full_feeds_emissions_July2024_nozeroes_detailed.csv)

[full_feeds_land_use_change_emissions_July2024_nozeroes_detailed.csv](https://github.com/user-attachments/files/17757285/full_feeds_land_use_change_emissions_July2024_nozeroes_detailed.csv)


krb78 (approved):
easy peasy lemon squeezy! Thanks for doing this, Joe! :)

PooyaHekmati (approved):

PR #2069, opened 2024-11-18
[Soil and Crop] Flatten the units in SC module

### LINKED ISSUE #1548: [Crop & Soil] Flatten all instances of nested units within Crop & Soil Module
There are a handful of instances in the model where the units for variables added to OM are nested within a second dictionary within their respective units dictionary (see image for example):

Screenshot 2024-04-15 at 1 49 48 PM

These should be flattened out to be the same level as the other units in the units dictionary. This will make testing in OM to ensure units added are members of the `MeasurementUnits` enum easier.

## Expected Behavior

When units are added to info maps accompanying variables added to the OM, they should only be one level deep (i.e. no nesting).

## Current Behavior

There are a handful of units, I believe exclusively within the C&S module, that have units in nested dictionaries.

## Possible Solution

Remove the nested dictionary of units and add the units at the same level of depth as the other non-nested units.

Testing will likely need to be updated.

matthew7838:
This PR flattens instances of nested units within Crop & Soil Module.

### Context

Issue(s) closed by this pull request: closes #1548

### What

Date units are flattened out at the same level as other units. 

### Why

Date units should be flattened out at the same level as other units to keep things easier to read and consistent across the codebase. 

### How

Extract nested unit dictionaries.

### Test plan

Run with updated tests, outputs are working as expected.


RUFAS/routines/field/field/field.py
```diff
@@ -1097,7 +1097,8 @@ def _record_planting(
         units = {
             "crop": MeasurementUnits.UNITLESS,
             "heat_scheduled_harvest": MeasurementUnits.UNITLESS,
-            "date": {"year": MeasurementUnits.CALENDAR_YEAR, "day": MeasurementUnits.ORDINAL_DAY},
+            "planting_year": MeasurementUnits.CALENDAR_YEAR,
+            "planting_day": MeasurementUnits.ORDINAL_DAY,
```
Isgotkowitz:
```suggestion
            "year": MeasurementUnits.CALENDAR_YEAR,
            "day": MeasurementUnits.ORDINAL_DAY,
```
I think it would be better to just call these the year and day, because they're in `_record_planting` I think it is clear that they are the planting year and date.

RUFAS/routines/field/field/field.py
```diff
@@ -1110,7 +1111,8 @@ def _record_planting(
         value = {
             "crop": species,
             "heat_scheduled_harvest": heat_scheduled_harvest,
-            "date": {"year": year, "day": day},
+            "planting_year": year,
+            "planting_day": day,
```
Isgotkowitz:
```suggestion
            "year": year,
            "day": day,
```

Isgotkowitz (approved):
LGTM

There are a number info maps in `Field` that nest the year and day inside of the date key -- would you consider flattening these in this PR as well? They can be found by searching for "date" in `field.py`. IMO this would be a good change to package with the flattening of nested units because it keeps the recording of dates more consistent in `Field`.

ew3361zh:
Yea I think my question is related to Ed's - can you outline the process for how you searched for these nested units in the PR description?

matthew7838:
> Yea I think my question is related to Ed's - can you outline the process for how you searched for these nested units in the PR description?

`\{[^{}]*\{[^{}]*\}[^{}]*\}` This is the regex pattern that I used.

ew3361zh (approved):

PR #2070, opened 2024-11-18
[Soil and Crop] Restructure Schedule.py

### LINKED ISSUE #1820: [Crop & Soil] Restructure Schedule class and its subclasses
#### Methods in the Schedule class are general enough to be moved to the Utility class:
1. elongate_list() lengthens a list to match the length of a reference list.
2. repeat_pattern() repeats a pattern of numbers based on the specified skip and repeat parameters.
3. validate_pattern_parameters() validates a set of parameters against a set of validation rules for a given schedule.
#### In Schedule subclasses:
1. The generate_*_events() methods in the Schedule subclasses share some common logic so a base class like EventGenerator would be helpful.
- The EventGenerator class will contain a common method generate_events() that generates events based on the provided event class, event arguments, pattern years, etc.
2. The _validate_*_parameters() follow the same pattern of checking a set of parameters against a set of validation rules, so it is possible to create a common method _validate_parameters() in the base Schedule class to perform the actual validation.
- Each _validate_*_parameters() method in the Schedule subclasses can focus on preparing the necessary parameters and validation rules and pass them to the _validate_parameters() method in the base class.
3. _determine_if_all_non_negative_values() - checks that all values in a list are >= 0. Could be moved to Utility class.

Example refactoring:
Screenshot 2024-07-16 at 11 26 24 AM

matthew7838:
Resturucure common methods to parent `schedule.py` or utility class.
### Context

Issue(s) closed by this pull request: closes #1820

### What

- `elongate_list() `, `repeat_pattern()`, `validate_pattern_parameters()`, `_determine_if_all_non_negative_values() ` from the original method have been moved to the utility class.
- The core logic of validate_parameters() and method generation have been moved to the parent schedule class. 
- Each subclass will have its own event_generator that prepares the parameters needed.
- Each sub-module validation will use the main logic validation and add more module-specific validation.
- Added a fraction validation method to the utility method.

### Why

The original schedules are really repetitive and some parts are general enough to be moved up the hierarchy.

### How

Cleans the code and generalizes the logic.

### Test plan

Run with modified tests and default metadata confirming no changes to the output. Filtered out FieldDataReporter reported outputs to check.


RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,48 +98,145 @@ def _validate_years(years: List[int]) -> bool:
...
+                f"Provided parameters are: {', '.join(f'{key}={value}' for key, value in kwargs.items())}. "
+                f"Lengths are: {lengths}."
+            )
+        return True
+
+    def validate_parameters(
+        self,
+        non_negative_parameters: list[tuple[str, list] | None],
+        fraction_parameters: list[tuple[str, list] | None],
+    ) -> None:
```
Isgotkowitz: To simplify this a little while utilizing OO features, there are a couple different things that could be done:
- Make `validate_parameters` in `Schedule` an `@abstractmethod`
- Add an `@classmethod` to `Schedule` called `_validate_parameters`, which has the general functionality that every `Schedule` subclass' `validate_parameters` will need.

RUFAS/data_structures/events.py
```diff
@@ -164,12 +167,12 @@ class TillageEvent(BaseFieldManagementEvent):
...
-        tillage_depth: float,
-        incorporation_fraction: float,
-        mixing_fraction: float,
-        implement: TillageImplement,
         year: int = 1,
         day: int = 160,
+        tillage_depth: float = None,
+        incorporation_fraction: float = None,
+        mixing_fraction: float = None,
+        implement: TillageImplement = None,
```
Isgotkowitz: I don't think these parameters should have default arguments, especially "None" arguments because they are not valid.

RUFAS/data_structures/events.py
```diff
@@ -132,6 +134,7 @@ def __init__(
     def __eq__(self, other):
         """Overrides the equality operator for HarvestEvent objects."""
         if isinstance(other, HarvestEvent):
+            # print(super().__eq__(other))
```
ew3361zh: remove print statement

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,48 +98,161 @@ def _validate_years(years: List[int]) -> bool:
...
+        all_events = list(zip(all_years, all_days, *repeated_attributes))
+        if heat_scheduled_harvest:
+            all_events[:] = [harvest for harvest in all_events if harvest[2] in FINAL_HARVEST_OPERATIONS]
+        result = [event_class(*additional_attributes, *event) for event in all_events]
+
+        return result
+
     @staticmethod
-    def _repeat_pattern(pattern: List[int], skip: int = 0, repeat: int = 0) -> List[int]:
+    def validate_depths(depths: List[float]) -> bool:
```
ew3361zh:
```suggestion
    def validate_depths(depths: list[float]) -> bool:
```

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,48 +98,161 @@ def _validate_years(years: List[int]) -> bool:
...
+        days: list[int],
+        name: str,
+    ) -> None:
+        """
+        General validations for schedule parameter.
+
+        Parameters
+        ----------
+        non_negative_parameters: list[tuple[str, list]]
+            A list of parameters wrapped in a tuple containing their names and values that should be non-negative.
```
ew3361zh:
```suggestion
            A list of tuples containing parameter names and associated non-negative values.
```

RUFAS/routines/field/manager/crop_schedule.py
```diff
@@ -187,22 +158,20 @@ def generate_planting_events(self) -> List[PlantingEvent]:
...
+                self.planting_skip,
+                self.pattern_repeat,
+                False,
             )
-            planting_events.append(new_planting_event)
-        return planting_events
+        )
 
-    def generate_harvest_events(self) -> List[HarvestEvent]:
+    def generate_harvest_events(self) -> list:
```
ew3361zh:
```suggestion
    def generate_harvest_events(self) -> list[HarvestEvent]:
```

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,48 +98,161 @@ def _validate_years(years: List[int]) -> bool:
...
+            List of tillage depths to be validated.
 
         Returns
         -------
-        List[int]
-            The full repeated pattern of numbers.
+        bool
+            True if all tillage depths are valid, False otherwise.
+
+        Notes
```
ew3361zh: Not sure this notes section is necessary. Seems evident from what the function does.

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,48 +98,161 @@ def _validate_years(years: List[int]) -> bool:
...
+    ) -> None:
+        """
+        General validations for schedule parameter.
+
+        Parameters
+        ----------
+        non_negative_parameters: list[tuple[str, list]]
+            A list of parameters wrapped in a tuple containing their names and values that should be non-negative.
+        fraction_parameters: list[tuple[str, list]]
+            A list of parameters wrapped in a tuple containing their names and values that should be fractions.
```
ew3361zh:
```suggestion
            A list of tuples containing parameter names and associated values that should be fractions.
```

RUFAS/util.py
```diff
@@ -644,3 +645,134 @@ def combine_saved_input_csv(
         result_df.to_csv(output_csv_path, index=False)
 
         shutil.rmtree(saved_csv_working_folder)
+
+    @staticmethod
+    def elongate_list(list_to_elongate: List[Any], reference_list_length: int) -> List[Any]:
```
ew3361zh:
```suggestion
    def elongate_list(list_to_elongate: list[Any], reference_list_length: int) -> list[Any]:
```

RUFAS/util.py
```diff
@@ -644,3 +645,134 @@ def combine_saved_input_csv(
...
+        """
+        Takes a list and lengthens it to match the length of the reference list, if the original length was 1.
```
ew3361zh, line 651: Is this a general enough function to apply to list-lengthening outside of schedules? It seems like this is specific to situations in the schedule class as it is now but maybe I'm not fully understanding other parts of the code where this could be helpful.

RUFAS/util.py
```diff
@@ -644,3 +645,134 @@ def combine_saved_input_csv(
...
+        list.
+
+        """
+        if len(list_to_elongate) != 1:
+            return list_to_elongate
+        elongated_list = list_to_elongate * reference_list_length
+        return elongated_list
+
+    @staticmethod
+    def repeat_pattern(pattern: List[int], skip: int = 0, repeat: int = 0) -> List[int]:
```
ew3361zh:
```suggestion
    def repeat_pattern(pattern: list[int], skip: int = 0, repeat: int = 0) -> list[int]:
```

RUFAS/util.py
```diff
@@ -644,3 +645,134 @@ def combine_saved_input_csv(
...
+        """
+        Takes a pattern of numbers and repeats the pattern of differences between the numbers for a specified number of
+        repetitions, skipping over specified gaps between repetitions.
```
ew3361zh: I was having trouble understanding what this function does (both in the code and the docstring description) so I asked chatgpt to rewrite the docstring and I think it's a helpful improvement:

`Extends a pattern of numbers by repeating it a specified number of times. The pattern's differences
    between consecutive numbers are calculated and used for repetition, with an optional gap (skip)
    added between each repetition.`

RUFAS/util.py
```diff
@@ -644,3 +645,134 @@ def combine_saved_input_csv(
...
+        Examples
+        --------
+        >>> repeat_pattern([1, 3, 5], 1, 2)
+        [1, 3, 5, 7, 9, 11, 13, 15, 17]
+
+        >>> repeat_pattern([1, 3, 5], 0, 1)
+        [1, 3, 5, 6, 8, 10]
+
+        >>> repeat_pattern([2, 3, 7], 3, 2)
+        [2, 3, 7, 11, 12, 16, 20, 21, 24]
```
ew3361zh: Could you expand these examples? I think they could use some context/description for what the intention is and how the result happens from the inputs.

RUFAS/util.py
```diff
@@ -644,3 +645,134 @@ def combine_saved_input_csv(
...
+        differences_index = 0
+        number_of_new_values = range(repeat * len(pattern))
+        for _new_value in number_of_new_values:
+            full_pattern.append(full_pattern[-1] + differences[differences_index])
+            differences_index += 1
+            differences_index %= len(pattern)
+        return full_pattern
+
+    @staticmethod
+    def validate_pattern_parameters(name: str, pattern_skip: int, pattern_repeat: int) -> None:
```
ew3361zh: I'm not sure this is generic enough to warrant being in util.py. It feels pretty specific to schedule-related stuff.

ew3361zh (commented):
Good steps taken here, Matthew. There's a lot more to this than I initially thought. 

I think there is some good work here - I think it's not clear that all the functions moved to Utility are generic enough to warrant being there. They need to be made more generic if they really belong there. The schedule process also still has some confusing elements to it. I think in particular the `repeat_pattern()` function needs to be structured more clearly.

Also please do address the MyPy error count.

RUFAS/util.py
matthew7838: With the combination of pattern validation, I agree that it is better to stay in schedule class.

RUFAS/util.py
matthew7838: This one I agree.

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,17 +99,173 @@ def _validate_years(years: List[int]) -> bool:
...
+        Extends a pattern of numbers by repeating it a specified number of times. The pattern's differences between
+        consecutive numbers are calculated and used for repetition, with an optional gap (skip) added between each
+        repetition.
+
+        Parameters
+        ----------
+        pattern : List[Union[int, float]]
             The pattern to be repeated.
-        skip : int
+        skip : Union[int, float]
```
ew3361zh: These don't match the param type hints.
As a side note, I think we're moving away from using Union and doing something like `int | float` instead.

RUFAS/util.py
ew3361zh: I'm confused, are you moving this one back to Schedule class?

RUFAS/routines/field/manager/schedule.py
```diff
@@ -163,15 +277,24 @@ def _repeat_pattern(pattern: List[int], skip: int = 0, repeat: int = 0) -> List[
 
         Examples
         --------
-        >>> repeat_pattern([1, 3, 5], 1, 2)
+        >>> Schedule.repeat_pattern([1, 3, 5], 1, 2)
         [1, 3, 5, 7, 9, 11, 13, 15, 17]
+        First pattern [1, 3, 5] -> repeat the pattern with the difference of 2.
+        Second repeated pattern [7, 9, 11] -> With a gap of 1 from last element of the last pattern (5).
```
ew3361zh: Okay, I think this was part of my confusion. A gap of 0 just means the next year/date. There is no way to have the schedule repeat starting in the same year which makes sense now that I think about it.

Thanks for this more detailed explanation, it helps a lot!

RUFAS/routines/field/manager/schedule.py
```diff
@@ -163,15 +277,24 @@ def _repeat_pattern(pattern: List[int], skip: int = 0, repeat: int = 0) -> List[
...
         [2, 3, 7, 11, 12, 16, 20, 21, 24]
 
+        >>> Schedule.repeat_pattern([1, 2, 3], 0, 1)
+        [1, 2, 3, 4, 5, 6]
+
+        >>> Schedule.repeat_pattern([1, 2, 3], 1, 1)
+        [1, 2, 3, 5, 6, 7]
+
         """
         differences = [skip + 1]
```
ew3361zh, line 318:
```suggestion
        if not pattern or repeat <= 0:
            return pattern
        differences = [skip + 1]
```

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,17 +99,173 @@ def _validate_years(years: List[int]) -> bool:
         """
         return all(0 < years[index] <= years[index + 1] for index in range(0, len(years) - 1))
 
+    def generate_events(
+        self,
+        years: list[int],
+        days: list[int],
+        additional_attributes: Optional[list[Any]],
+        additional_attributes_events: list[list[Any]],
+        event_class: Any,
```
Isgotkowitz, line 106:
```suggestion
        event_class: BaseFieldManagementEvent,
```

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,17 +99,173 @@ def _validate_years(years: List[int]) -> bool:
...
+        Returns
+        -------
+        bool
+            True if all tillage depths are valid, False otherwise.
+
+        """
+        return all(depth > 0.0 for depth in depths)
+
+    @staticmethod
+    def validate_equal_lengths(header: str, **kwargs: Any) -> bool:
```
Isgotkowitz, line 197:
```suggestion
    def validate_equal_lengths(header: str, **kwargs: dict[str, Any]) -> bool:
```

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,17 +99,173 @@ def _validate_years(years: List[int]) -> bool:
...
+        """
+        Validates that all provided iterables have the same length.
+
+        Parameters
+        ----------
+        header: str
+            Error header when needs to throw an error.
```
Isgotkowitz:
```suggestion
            Error header when for when an error is raised.
```

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,17 +99,173 @@ def _validate_years(years: List[int]) -> bool:
...
+        Extends a pattern of numbers by repeating it a specified number of times. The pattern's differences between
+        consecutive numbers are calculated and used for repetition, with an optional gap (skip) added between each
+        repetition.
+
+        Parameters
+        ----------
+        pattern : List[int | float]
             The pattern to be repeated.
-        skip : int
+        skip : int | float
```
Isgotkowitz:
```suggestion
        skip : int
```

RUFAS/routines/field/manager/schedule.py
```diff
@@ -186,3 +311,20 @@ def _repeat_pattern(pattern: List[int], skip: int = 0, repeat: int = 0) -> List[
...
+        ------
+        ValueError
+            If the skip is < 0.
+            If the repeat is < 0.
+
+        """
+        if pattern_skip < 0:
+            raise ValueError(f"'{name}': expected pattern skip to be >= 0, received '{pattern_skip}'.")
+        if pattern_repeat < 0:
+            raise ValueError(f"'{name}': expected pattern repeat to be >= 0, received '{pattern_repeat}'.")
```
Isgotkowitz:
```suggestion
```
I think it would be ok to delete this method. Checking individual inputs in isolation is done by the IM, so it would be good to take this out.

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,17 +99,173 @@ def _validate_years(years: List[int]) -> bool:
...
+        all_events = list(zip(all_years, all_days, *repeated_attributes))
+        if heat_scheduled_harvest:
+            all_events[:] = [harvest for harvest in all_events if harvest[2] in FINAL_HARVEST_OPERATIONS]
+        result = [event_class(*additional_attributes, *event) for event in all_events]
+
+        return result
+
     @staticmethod
-    def _repeat_pattern(pattern: List[int], skip: int = 0, repeat: int = 0) -> List[int]:
+    def validate_depths(depths: list[float]) -> bool:
```
Isgotkowitz:
```suggestion
    def validate_positive_values(values: list[float]) -> bool:
```
By changing this method to not be specific to depths, it makes clearer that it could be applicable to any values that need to be positive.

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,17 +99,173 @@ def _validate_years(years: List[int]) -> bool:
...
+    def validate_equal_lengths(header: str, **kwargs: Any) -> bool:
+        """
+        Validates that all provided iterables have the same length.
+
+        Parameters
+        ----------
+        header: str
+            Error header when needs to throw an error.
+        kwargs : list of iterables
+            The iterables to check for length equality.
```
Isgotkowitz:
```suggestion
    def validate_equal_lengths(header: str, **kwargs: dict[str, Any]) -> bool:
        """
        Validates that all provided lists of values have the same length.

        Parameters
        ----------
        header: str
            Error header when needs to throw an error.
        kwargs : dict[str, Any]
            Key word arguments where the key is the name of the parameter and the value is a list of values.
```

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,17 +99,173 @@ def _validate_years(years: List[int]) -> bool:
...
+
+        Returns
+        -------
+        bool
+            True if all lengths are equal.
+
```
Isgotkowitz, line 212: An examples section would help make this method clearer

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,17 +99,173 @@ def _validate_years(years: List[int]) -> bool:
...
+        Parameters
+        ----------
+        years : List[int]
+            List of years for the schedule.
+        days : List[int]
+            List of days for the schedule.
+        additional_attributes : List[List]
+            Additional general attributes for the events (e.g., crop reference).
+        additional_attributes_events : List[List]
+            Additional attributes for each of the events (e.g., nitrogen_mass, phosphorus_mass, etc.).
```
Isgotkowitz:
```suggestion
        general_attributes: Optional[list[Any]],
        event_specific_attributes: list[list[Any]],
        event_class: Any,
        pattern_skip: int,
        pattern_repeat: int,
        heat_scheduled_harvest: bool,
    ) -> list[Any]:
        """
        Generic method to generate application events.

        Parameters
        ----------
        years : List[int]
            List of years for the schedule.
        days : List[int]
            List of days for the schedule.
        general_attributes : List[List]
            Attributes that apply to all events.
        event_specific_attributes : List[List]
            Attributes which are specific to individual events.
```

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,17 +99,173 @@ def _validate_years(years: List[int]) -> bool:
...
+        list
+            List of instantiated event objects.
+
+        """
+        all_years = self.repeat_pattern(years, pattern_skip, pattern_repeat)
+        all_days = days * (pattern_repeat + 1)
+        repeated_attributes = [attr * (pattern_repeat + 1) for attr in additional_attributes_events]
+        all_events = list(zip(all_years, all_days, *repeated_attributes))
+        if heat_scheduled_harvest:
+            all_events[:] = [harvest for harvest in all_events if harvest[2] in FINAL_HARVEST_OPERATIONS]
```
Isgotkowitz: Having this logic in `Schedule` is problematic, because it is specific to only one type of `BaseFieldManagementEvent`. Instead of having this filtering here, it should be embedded within `generate_harvest_events`.

Isgotkowitz (commented):
I'm sorry for taking so long to get to this Matthew, I think I started a review and then forgot to submit it.

I think this PR is looking a lot better. However there is some logic specific to `HarvestEvent` which is in the parent class `Schedule`, and this design is problematic.

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,15 +97,210 @@ def _validate_years(years: List[int]) -> bool:
...
+
+        Parameters
+        ----------
+        values : List[float]
+            List of tillage depths to be validated.
+
+        Returns
+        -------
+        bool
+            True if all tillage depths are valid, False otherwise.
```
Isgotkowitz:
```suggestion
        Checks that values passed are greater than 0.

        Parameters
        ----------
        values : List[float]
            List of values to be validated.

        Returns
        -------
        bool
            True if all values are greater than 0, False otherwise.
```

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,15 +97,210 @@ def _validate_years(years: List[int]) -> bool:
...
+
+        """
+        return all(depth > 0.0 for depth in values)
```
Isgotkowitz:
```suggestion
        return all(value > 0.0 for value in values)
```

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,15 +97,210 @@ def _validate_years(years: List[int]) -> bool:
...
+            if not Utility.validate_fractions(parameter):
+                raise ValueError(
+                    f"'{name}': " + f"expected all {parameter_name} to be in"
+                    f" range [0.0, 1.0], "
+                    f"received '{parameter}'."
+                )
+
     @staticmethod
-    def _repeat_pattern(pattern: List[int], skip: int = 0, repeat: int = 0) -> List[int]:
+    def repeat_pattern(pattern: List[int], skip: int = 0, repeat: int = 0) -> list[int]:
```
Isgotkowitz:
```suggestion
    def repeat_pattern(pattern: List[int | float], skip: int = 0, repeat: int = 0) -> list[int]:
```

Isgotkowitz (approved):
Looks good Matthew, produces the same results and errors as `dev`.

RUFAS/routines/field/manager/schedule.py
ew3361zh: Curious how this resolved? It looks like it's still type-hinted as `Any`.

RUFAS/routines/field/manager/schedule.py
ew3361zh: Same here - is this staying in because you guys talked it through in person? If so it would be helpful to leave a note here as to why this method wasn't deleted/what that conversation was.

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,15 +97,210 @@ def _validate_years(years: List[int]) -> bool:
...
+        event_class: Any,
+        pattern_skip: int,
+        pattern_repeat: int,
+    ) -> list[Any]:
+        """
+        Generic method to generate application events.
+
+        Parameters
+        ----------
+        years : List[int]
```
ew3361zh:
```suggestion
        years : list[int]
```

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,15 +97,210 @@ def _validate_years(years: List[int]) -> bool:
...
+            List of years for the schedule.
+        days : List[int]
```
ew3361zh:
```suggestion
        days : list[int]
```

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,15 +97,210 @@ def _validate_years(years: List[int]) -> bool:
...
+            List of days for the schedule.
+        additional_attributes : List[List]
```
ew3361zh:
```suggestion
        additional_attributes : Optional[list[Any]]
```

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,15 +97,210 @@ def _validate_years(years: List[int]) -> bool:
...
+            Additional general attributes for the events (e.g., crop reference).
+        additional_attributes_events : List[List]
```
ew3361zh:
```suggestion
        additional_attributes_events : list[list[Any]]
```

RUFAS/routines/field/manager/schedule.py
```diff
@@ -141,15 +97,210 @@ def _validate_years(years: List[int]) -> bool:
...
+            Additional attributes for each of the events (e.g., nitrogen_mass, phosphorus_mass, etc.).
+        event_class : class
```
ew3361zh: Should this be:
```suggestion
        event_class : BaseFieldManagementEvent
```
?
If not it should at least match the type hint in the params section

RUFAS/routines/field/manager/schedule.py
matthew7838: We talked about this in WT. I will note this down next time. It's causing a lot of problems with mypy, and the same is true for other hinting suggestions.

RUFAS/routines/field/manager/schedule.py
matthew7838: It was removed, sorry I forgot to resolve it here.

ew3361zh (approved):
Looks good. Left some comments about docstrings but they're minor. Nice work!

RUFAS/routines/field/manager/schedule.py
ew3361zh: It's still showing up for me on this branch.

matthew7838: Sorry, I think I was branch-hopping too much.

RUFAS/routines/field/manager/schedule.py
matthew7838: Good catch, thanks.

PR #2084, opened 2024-11-22
[Crop & Soil] Break Up CropData

### LINKED ISSUE #1815: [Crop & Soil] Restructure Crop and CropData classes.
## Problem

A common pattern in the crop and soil submodules is the use of catch-all dataclasses, i.e., CropData and SoilData, that contain all the relevant attributes, and then those attributes get updated by some modifying functions. This approach keeps all the data together in one place, making it easier to access and update. However, those classes are large and harder to maintain. One tricky thing is that multiple classes can modify the same data, possibly leading to buggy behaviors and race conditions.

## Proposed Solution

We should break up the CropData class into smaller, more focused dataclasses and then use composition to build it back up. So now a class like NitrogenIncorporation should interact only with a subset of the relevant data in a CropData object. This helps reduce the extent of the coupling and should be done with the single responsibility principle in mind.

The following diagrams illustrate these proposed changes.

Crop class decomposition
The Crop class will be composed of various subcomponents, including CropData, GrowthConstraints, BiomassAllocation, WaterDynamics, WaterUptake, NitrogenIncorporation, PhosphorusIncorporation, HeatUnits, LeafAreaIndex, RootDevelopment, CropManagement, and Dormancy.

Screenshot 2024-07-15 at 5 29 39 PM

The CropData class can be further broken down into smaller dataclasses such as CropIdentification, CropGrowth, CropNutrients, CropWater, and CropYield.

CropData class decomposition

- CropIdentification contains crop species, name, and plant category information, etc.
- CropGrowth holds data related to leaf area index, biomass, root depth, etc.
- CropNutrients stores information about nitrogen and phosphorus levels among others.
- CropWater includes water stress and water uptake data.
- CropYield contains harvest index and yield residue information.

Screenshot 2024-07-15 at 5 32 57 PM

Each subcomponent of the Crop class interacts with a specific subset of the CropData class, promoting the single responsibility principle and reducing coupling.
The grow_crop() method in the Crop class orchestrates the interactions between the subcomponents and the CropData class.

## Alternative Solution to breaking down CropData into smaller dataclasses
Instead of creating more data classes, we identify the attributes in CropData that are only used by one crop component, and make them attributes of that component directly.

ew3361zh:
Moves attributes of CropData that are only used in other classes out of CropData and into those specific classes.

### Context

Issue(s) closed by this pull request: closes #1815


ew3361zh:
Being covered in PR #2090

PR #2089, opened 2024-11-26
[Output Manager] Make the log routing usable by other modules.

### LINKED ISSUE #2082: [Output Manger] Make use of the log routing method across codebase
There's a function in the route log that allows saved errors/warnings/logs to be correctly routed and reported by the OM. Right now that method is not public and a few classes have the copy of the exact same method.

## Expected Behavior

OM should have a public routing method that allows usage across the codebase.

## Current Behavior

Routing is an OM private method.

## Possible Solution

1. Make it a `_route_log` static method.
2. Utilize it on the report generator and IM.
3. Implement the suggestions in PR #2061 by Ed.

## Steps to Reproduce

1.
4.
5.
6.

matthew7838:
Using the log routing function from OM in different modules.
### Context

Issue(s) closed by this pull request: closes #2082

### What

- Incorporated more concrete validation for the routing method mentioned in PR #2061
- The routing method is now public and being used across different modules.
- Updated unit tests.

### Why

This allows more flexibility and modularity for data validation.

### Test plan

Run with updated unit tests and make sure that the amount of warnings/logs/errors maintains the same.

### DIFF
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -1503,7 +1503,7 @@ class OutputManager(object):
                         filter_content["graph_details"]["metadata_prefix"] = self.__metadata_prefix
                         self.create_directory(graphics_dir)
                     log_pool = report_generator.generate_report(filter_content, filtered_pool)
-                    self._route_logs(log_pool)
+                    self.route_logs(log_pool)
                 else:
                     self._route_save_functions(
                         filter_file,
@@ -1562,7 +1562,7 @@ class OutputManager(object):
                     log_pool = graph_generator.generate_graph(
                         filtered_pool, filter_content, filter_file, graphics_dir, produce_graphics
                     )
-                    self._route_logs(log_pool)
+                    self.route_logs(log_pool)
                 except Exception as e:
                     self.add_error("graph generation exception", str(e), info_map)
             else:
@@ -1614,7 +1614,7 @@ class OutputManager(object):
         file_path = save_path / file_name
         self.dict_to_file_json(filtered_pool, file_path, origin_label=origin_label)
 
-    def _route_logs(self, log_pool: list[dict[str, str | dict[str, str]]]) -> None:
+    def route_logs(self, log_pool: list[dict[str, str | dict[str, str]]]) -> None:
         """Takes logs from other classes and routes them to the appropriate pools in
         Output Manager.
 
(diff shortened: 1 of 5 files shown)


tests/test_output_manager.py
```diff
@@ -1099,7 +1099,7 @@ def output_manager_original_method_states(
         "clear_output_dir": mock_output_manager.clear_output_dir,
         "is_file_in_dir": mock_output_manager.is_file_in_dir,
         "create_directory": mock_output_manager.create_directory,
-        "_route_logs": mock_output_manager._route_logs,
+        "_route_logs": mock_output_manager.route_logs,
```
Isgotkowitz:
```suggestion
        "route_logs": mock_output_manager.route_logs,
```

Isgotkowitz (commented):
Compared the logs from this branch to those on `dev`, no changes except for timestamps! Nice job Matthew!

Isgotkowitz (approved):

RUFAS/output_manager.py
```diff
@@ -1624,28 +1624,59 @@ def _route_logs(self, log_pool: list[dict[str, str | dict[str, str]]]) -> None:
...
+                else:
+                    self.add_warning(
+                        "Wrong format for adding warning.",
+                        f"Unable to add warning with the format: {log}",
+                        info_map,
+                    )
+            else:
+                self.add_warning(
+                    "Unsupported event key for output manager",
+                    f"Output manager can add logs, errors and warning. Given event contains the key {log.keys()}",
```
```suggestion
                    f"Output manager can add logs, errors and warnings. Given event contains the key {log.keys()}",
```
Also curious why it's referred to here as an `event key` here? I don't believe we use that terminology anywhere in OM and it might be confusing to see this warning message.

I'm not sure how in depth you want to get with the warning message (like which of the keys of the log is incorrect) but I think it's good to provide the log.keys that was sent to this function and maybe have something saying the first key has to be `log`, `warning`, or `error`, the second key has to be `message`, and the third key has to be `info_map`).

RUFAS/output_manager.py
```diff
@@ -1624,28 +1624,59 @@ def _route_logs(self, log_pool: list[dict[str, str | dict[str, str]]]) -> None:
...
                 ):
                     self.add_warning(log["warning"], log["message"], log["info_map"])
+                else:
+                    self.add_warning(
+                        "Wrong format for adding warning.",
+                        f"Unable to add warning with the format: {log}",
+                        info_map,
+                    )
+            else:
+                self.add_warning(
```
ew3361zh: I tried changing one of the keys to a log message in data_validator from a valid key to an invalid key and this warning was never reached, the simulation just crashed.

change was on line 251 of data_validator.py and I changed "message" to "msg".

So I think this might need to be a try-except instead?

matthew7838: Great catch!

PR #2093, opened 2024-11-27
[Testing] Increase Tolerance for E2E testing results 

### LINKED ISSUE #2058: [Testing] Adopt a tolerance allowance for E2E testing result differences
Add a tolerance calculation between expected and actual results of end to end testing to better quantify the differences and help decide whether that is considered failing.

## Expected Behavior

By adding a tolerance calculation on top of the current deepdiff `significant_digits` setting, we could reduce microscopic differences causing end-2-end-testing to fail.

## Current Behavior

Currently E2E testing expected and actual results are compared using deepdiff's internal `significant_digits` setting. We're noticing differences between (at least) different operating systems because of floating point precision and rounding differences. This causes end to end testing to fail on differences between expected and actual results that are in the millionths and ten millionths which is unhelpful.

## Possible Solution

```
diff = DeepDiff(data1, data2, significant_digits=8)

# Apply custom tolerance filter to ignore differences below a threshold
tolerance = 1e-7  # RUFAS custom tolerance setting
filtered_diff = {k: v for k, v in diff.items() if abs(v) > tolerance}
```

ew3361zh:
Adds a tolerance setting for comparing E2E testing expected and actual results.

### Context

Issue(s) closed by this pull request: closes #2058

### What

Increases the tolerance of the comparison between actual and expected results to allow for more lenient comparisons (though still very precise). Specifically allows for a higher tolerance comparing the `old_value` and `new_value` in the diff result `changed_values` section. If the difference is below this tolerance level, the difference will be removed from the diff object reported to OM meaning it won't cause E2E testing to fail. The tolerance can be set within the input file for end-to-end testing.

Also updated:
1. OM testing updated to 100% covering some E2E testing aspects in OM that were previously uncovered.
2. Updated the results for Animal and Crop & Soil `expected_results` because both had values added/removed.
3. Updated E2E JSON output filenames for comparisons of results (actual to expected): the comparison file will now have the structure `end-to-end-testing_comparison_domain_results_date`.

### Why

This is an attempt to account for floating point rounding differences we're seeing with outputs between different operating systems, hardware, etc. These are small differences that distract from the actual comparisons we're looking for between actual and expected results of end-to-end testing.

### How

A few things were implemented here:
1. Increasing the tolerance level for the deepdiff process itself from 8 significant digits to 6.
2. Implementing a new 3 part functionality that recursively looks through the `values_changed` portion of the deepdiff result and applies a relative tolerance to see whether the difference found is "significant" enough to warrant reporting to OM.
3. Each domain now has its own `tolerance` setting in the `end_to_end_testing_result_paths.json` file.
4. If the difference is not found to be significant by this tolerance, the function will remove this result from the diff result which means it won't be reported to OM.
5. If all the results are found to be insignificant at this tolerance level, they will all be removed and E2E testing will be deemed successful.

#### The tolerance calculation overview:
It will return False if the difference is not significant and True if it is significant. False tells the system to ignore that change in reporting out the e2e testing results. True tells the system to report it in the e2e testing results.
reference = abs(old_value) if abs(old_value) > 0 else 1
difference = abs(new_value - old_value)
return difference > tolerance * reference

#### Examples:
Not exceeding tolerance (i.e. not a significant change from expected to actual results, we don't want this reported by e2e testing results comparison):
old_value = 100, new_value = 105, tolerance = 0.05
reference = abs(100)  # 100
difference = abs(105 - 100)  # 5
return 5 > 0.05 * 100  # False

Exceeding tolerance (i.e. a change that is significant that we'd want reported by e2e testing comparison):
old_value = 100, new_value = 120, tolerance = 0.05
reference = abs(100)  # 100
difference = abs(120 - 100)  # 20
return 20 > 0.05 * 100  # True

Smaller values:
old_value = 0, new_value = 0.001, tolerance = 0.00001 (showing how this would work if the old_value is 0)
reference = abs(0) if abs(0) > 0 else 1  # 1
difference = abs(0.001 - 0)  # 0.001
return 0.001 > 0.00001 * 1  # True

Zero tolerance:
old_value = 10, new_value = 10.001, tolerance = 0 (case if the user wants to adjust to have no tolerance from old_values)
reference = abs(10)  # 10
difference = abs(10.001 - 10)  # 0.001
return 0.001 > 0 * 10  # True

### Test plan

This was run with the dev versions of expected results in the e2e input files where my Mac's floating point errors were causing failure of E2E testing for C&S and Animal modules. While it didn't pass entirely, it did remove the vast majority of the differences in the `values_changed` section. The other results would have been eliminated at a tolerance of `1e-2`.

Then I ran this with the updated results on the feature branch and it was successful with the results I got from updating the `expected_results` section of the e2e input files for C&S and Animal.

I think other folks should do the same but perhaps more importantly on the feature branch if your results on dev are working as things are right now.


changelog.md
```diff
@@ -56,6 +56,8 @@ v0.9.2
 - [2069](https://github.com/RuminantFarmSystems/MASM/pull/2069) - [minor change] [Crop & Soil] Flatten the units in SC module.
 - [2076](https://github.com/RuminantFarmSystems/MASM/pull/2076) - [minor change] [Crop & Soil] Removes TODOs from the Crop & Soil module.
 - [1785](https://github.com/RuminantFarmSystems/MASM/pull/1785) - [minor change] [TaskManager] Adds sobol and morris samplers, updates argument names.
+- [2086](https://github.com/RuminantFarmSystems/MASM/pull/2086) - [minor change] [Animal] Updates default parity fraction.
```
```suggestion
```

RUFAS/e2e_test_results_comparer.py
```diff
@@ -94,3 +97,79 @@ def _get_test_result_paths() -> list[ResultPathType]:
...
+        ----------
+        change : dict[str, float | str]
+            A dictionary representing a change with "old_value" and "new_value".
+        tolerance : float
+            The threshold for considering a difference as significant.
+
+        Returns
+        -------
+        bool
+            True if the change is significant, False otherwise.
```
Isgotkowitz: I think a note or something should be added to clarify that this check is only applied to numerical values. Any changes in non-numerical values are considered significant.

RUFAS/e2e_test_results_comparer.py
```diff
@@ -9,7 +9,8 @@
 from RUFAS.output_manager import OutputManager
 from RUFAS.units import MeasurementUnits
 
-ResultPathType = namedtuple("ResultPaths", ["domain", "expected_results_path", "actual_results_path"])
+ResultPathType = namedtuple("ResultPathType", ["domain", "expected_results_path", "actual_results_path"])
+DEFAULT_TOLERANCE = 1e-3
```
Isgotkowitz: What would you think about making the tolerance a part of `ResultPathType` so that it can be specific to different parts of the model?

RUFAS/e2e_test_results_comparer.py
```diff
@@ -94,3 +97,79 @@ def _get_test_result_paths() -> list[ResultPathType]:
...
+        Returns
+        -------
+        bool
+            True if the change is significant, False otherwise.
+        """
+        if isinstance(change, dict) and "old_value" in change and "new_value" in change:
+            old_value = change["old_value"]
+            new_value = change["new_value"]
+            if isinstance(old_value, (int, float)) and isinstance(new_value, (int, float)):
+                difference = abs(new_value - old_value)
```
Isgotkowitz, line 135: Instead of using an absolute tolerance approach, what about using a relative one? So if the change in values is > %1, then it would be considered significant.

I think this approach would be better because magnitude of RuFaS outputs varies a lot, with some in the single digits and others with many orders of magnitude.

Isgotkowitz (commented):
Still comparing the E2E results from this branch and `dev`, but leaving this review because I have some ideas that may enhance the tolerance feature

Isgotkowitz (commented):
Looking at the differences that are found, especially in the Crop and Soil module E2E test, I think that a lot of them would go away if a relative tolerance was used. Looking at the Animal tests, it seems like something is really out of wack because it seems there are a lot of values out of order.

ew3361zh:
Updates:
- Added `tolerance` to ResultPathType structure in metadata and e2e test input files. Allows for control over tolerance level at the domain level.
- Changed `is_significant` tolerance from absolute tolerance to a symmetrical relative tolerance which takes into account both values being compared.
- Comparison results are now passing for me on all 3 domains currently in e2e testing.

input/metadata/properties/default.json
```diff
@@ -5684,6 +5684,12 @@
         "actual_results_path": {
           "type": "string",
           "description": "Path to the file with the actual results for the specified domain."
+        },
+        "tolerance": {
+          "type": "number",
+          "description": "The tolerance used to compare expected and actual results for changed values found in the e2e results comparison.",
```
```suggestion
          "description": "The relative tolerance used to compare expected and actual results for changed values found in the e2e results comparison.",
```

RUFAS/e2e_test_results_comparer.py
```diff
@@ -91,6 +93,95 @@ def _get_test_result_paths() -> list[ResultPathType]:
...
+        relative to the "old_value". If the "old_value" is zero, a fallback reference value of 1 is used
+        to ensure the tolerance comparison remains meaningful.
+        """
+        if isinstance(changes, dict) and "old_value" in changes and "new_value" in changes:
+            old_value = changes["old_value"]
+            new_value = changes["new_value"]
+            if isinstance(old_value, (int, float)) and isinstance(new_value, (int, float)):
+                reference = abs(old_value) if abs(old_value) > 0 else 1
+                difference = abs(new_value - old_value)
+                return difference > tolerance * reference
```
Isgotkowitz:
```suggestion
                return difference > tolerance * GeneralConstants.PERCENT_TO_FRACTION * reference
```
What would you think of adding this, so that a 1% tolerance would be expressed as 1 instead of 0.01? I think it would make the tolerance easier to understand and set for someone who is just reading the description.

Isgotkowitz (commented):
Still waiting on the Animal E2E test to finish running so I can look at the results, but I just want to confirm the behavior of the tolerance and then I think it will be an easy approval

RUFAS/e2e_test_results_comparer.py
ew3361zh: I'm not sure I understand. If we used this conversion factor we'd also need to do a comparable adjustment on the other side of the equation otherwise we're throwing off the comparison.

Isgotkowitz: I think the conversion would be done by the user when specifying the tolerance. So if they had 0.01 as the tolerance before, they would rewrite the tolerance to be 1.0 and the conversion factor would turn it back into 0.01

ew3361zh: Ah, gotcha. I guess I could see it either way. Do you feel strongly it should be a %? I will implement it. Seems like more frequently tolerance is referred to as a % in general.

ew3361zh:
Updates:
- Changed tolerance input to be a percent rather than a decimal and updated the calculation in `is_significant` to convert it to a decimal.
- After discussion with Pooya, he said he was comfortable with 3 significant digits as the guideline for the deepdiff comparison so I've updated that as well.
- Results were all still passing as expected with the more lenient tolerance/significant digits in place.

Isgotkowitz (approved):
E2E testing runs and completes, but still fails for the Animal and Crop & Soil domains.

matthew7838 (approved):
All passed on my side.

ew3361zh:
After meeting with @PooyaHekmati, holding off on merging this PR in until we can get the results consistent and passing across all the current domains (Animal, C&S, Feed Storage).

Will be creating issues to investigate what's happening further and then will link them back to this PR.

Issue #548, opened 2023-06-01
[Crop Soil] Change management of `LayerData` objects in `SoilData`


Isgotkowitz:
### Current
`SoilData` maintains a list of `LayerData` objects in a list. The top soil layer is at index 0, the bottom one at index $n$ (when $n$ soil layers are present). This necessitates soil layers being accessed directly, i.e. getting the top layer would be `data.soil_layers[0]`. This is somewhat problematic because it means modules can access `LayerData` in ways that may lead to buggy or otherwise undesirable behavior.

### Solution
Create a dedicated class to manage the list of `LayerData`'s in a `SoilData`. This class should have implement standard CRUD functionalities, as well as provide convenient methods for accessing specific layers, such as the top and bottom, possibly also the vadose zone layer.

This idea was put forth by Loi in PR #536

Isgotkowitz:
An appropriate data structure for managing soil layers may be a linked list.

Isgotkowitz:
This issue should not be started until pilot testing on the Crop and Soil module has wrapped up and SMEs give the go-ahead to start working on the Crop and Soil module.

KFosterReed:
Groomed March 27, 2026: This is another older issue with a suggestion on how to improve the architecture of the soil methods. I don't have enough knowledge to understand how big of an issue this is so am marking it as a low priority

PR #2120, opened 2024-12-03
[Crop and Soil] Throw NitrogenIncoporation error to OM

### LINKED ISSUE #2074: [Crop & Soil] Log errors to OM before raising `ValueErrors` in `NitrogenIncorporation`
If invalid values are found in the `NitrogenIncorporation` module, `ValueErrors` are reported immediately.

## Status

If invalid values are found in the `NitrogenIncorporation` module, `ValueErrors` are reported immediately.

## Intention

Implement better, more thorough logging of errors in the `NitrogenIncorporation` module.

## Request

Log errors to the OM before raising exceptions in the `NitrogenIncorporation` module. The errors should have clear, detailed explanations and should include all information that may be helpful in the info map.

matthew7838:
Errors in NitrogenIncoporation should also be logged to the OM before crashing the simulation. **Please do give suggestions on how the error messages can be improved if needed.**
### Context

Issue(s) closed by this pull request: closes #2074

### What

For all places that throw errors, now there's a call to OM's add_error and log with detail messages before throwing. Update all tests that test the error case to assert the call of om's add_error.

### Why

Adds clarity for users to understand the failure.

### How

Log details and provides the exact case of the failure to the user in the OM log.

### Test plan

Running updated test suits should be enough.


RUFAS/routines/field/crop/nitrogen_incorporation.py
```diff
@@ -396,7 +397,19 @@ def determine_nutrient_shape_parameters(
             If half_mature_heat_fraction equals mature_heat_fraction.
 
         """
+        info_map = {
+            "class": NitrogenIncorporation.__class__.__name__,
+            "function": NitrogenIncorporation.determine_nutrient_shape_parameters.__name__,
+        }
+        om = OutputManager()
         if mature_heat_fraction == half_mature_heat_fraction:  # leads to divide by 0
+            om.add_error(
```
Isgotkowitz:
```suggestion
        if mature_heat_fraction == half_mature_heat_fraction:  # leads to divide by 0
            info_map = {
                "class": NitrogenIncorporation.__class__.__name__,
                "function": NitrogenIncorporation.determine_nutrient_shape_parameters.__name__,
            }
            om = OutputManager()
            om.add_error(
```
Since the OM is only used within this if block, it is very slightly more efficient to instantiate the OM in it.

RUFAS/routines/field/crop/nitrogen_incorporation.py
```diff
@@ -396,7 +397,19 @@ def determine_nutrient_shape_parameters(
...
+                "Equal half_mature_heat_fraction and mature_heat_fraction.",
+                f"half_mature_heat_fraction must not equal mature_heat_fraction,"
+                f" received mature_heat_fraction is {mature_heat_fraction},"
+                f" half_mature_heat_fraction is {half_mature_heat_fraction}",
+                info_map,
```
Isgotkowitz:
```suggestion
                "A crop's half mature heat fraction and mature heat fraction are equal.",
                f"Half mature heat fraction and mature heat fraction are both {mature_heat_fraction}, this results in a divide by zero error."
                info_map,
```
You don't have to use this exact suggestion, but keep in mind that it is good to write both the cause and effect in error and warning messages. Ideally, the info map would also contain information like the name of the crop and simulation date when this occurs. But that can't be done without changing the method args or making it non-static, neither of which would be worth while here IMO.

RUFAS/routines/field/crop/nitrogen_incorporation.py
```diff
@@ -481,24 +498,64 @@ def _determine_shape_log(
             or emergence_nitrogen_fraction < 0
             or emergence_nitrogen_fraction > 1
         ):
+            om.add_error(
+                "Invalid nitrogen_fraction, heat_fraction, mature_nitrogen_fraction,"
+                " or emergence_nitrogen_fraction.",
+                f"One or more of the following fractions are not between 0 and 1,"
+                f" nitrogen_fraction: {nitrogen_fraction}, heat_fraction: {heat_fraction},"
+                f" mature_nitrogen_fraction: {mature_nitrogen_fraction},"
+                f" emergence_nitrogen_fraction: {emergence_nitrogen_fraction}.",
```
Isgotkowitz:
```suggestion
                "Received invalid fractional value",
                f"All following values must be in the range (0, 1), received {nitrogen_fraction=}, {heat_fraction=}, {mature_nitrogen_fraction=}, {emergence_nitrogen_fraction=}.",
```
I suggest trying to shorten this message a little, I think it helps make the code more readable.

Isgotkowitz (approved):
Left some suggestions for how the error message could be improved, and I also suggest putting any relevant invalid values in the info maps as well as the error messages. But overall looks good, nice job Matthew!

RUFAS/routines/field/crop/nitrogen_incorporation.py
```diff
@@ -481,24 +497,61 @@ def _determine_shape_log(
...
             )
             raise ValueError(frac_error_msg)
-        if emergence_nitrogen_fraction == mature_nitrogen_fraction:  # leads to divide by zero
+        if emergence_nitrogen_fraction == mature_nitrogen_fraction:
+            om.add_error(
+                "Invalid value pair for emergence_nitrogen_fraction and mature_nitrogen_fraction.",
+                f"Equal emergence_nitrogen_fraction and mature_nitrogen_fraction will lead to zero"
+                f" division error,"
+                f"emergence_nitrogen_fraction and mature_nitrogen_fraction are both."
+                f" {emergence_nitrogen_fraction}",
```
ew3361zh: I liked the wording better on the half mature and mature heat fraction error messages and would suggest using that same format here.
```suggestion
                "A crop's emergence_nitrogen_fraction and mature_nitrogen_fraction are equal.",
                f"The emergence_nitrogen_fraction and mature_nitrogen_fraction are both"
                f" {emergence_nitrogen_fraction}, this results in a divide by zero error.",
```

RUFAS/routines/field/crop/nitrogen_incorporation.py
```diff
@@ -481,24 +497,61 @@ def _determine_shape_log(
...
+                info_map,
+            )
             raise ValueError("emergence_nitrogen_fraction must not be equivalent to mature_nitrogen_fraction")
-        if nitrogen_fraction == emergence_nitrogen_fraction:  # leads to divide by zero
+        if nitrogen_fraction == emergence_nitrogen_fraction:
+            om.add_error(
+                "Invalid value pair for nitrogen_fraction and emergence_nitrogen_fraction.",
+                f"Equal nitrogen_fraction and emergence_nitrogen_fraction will lead to zero division"
+                f" error, nitrogen_fraction and emergence_nitrogen_fraction are both."
+                f" {emergence_nitrogen_fraction}.",
```
ew3361zh: I would also rewrite this one in the same fashion as the half mature/mature heat fraction style. Just my preference though.

RUFAS/routines/field/crop/nitrogen_incorporation.py
```diff
@@ -481,24 +497,61 @@ def _determine_shape_log(
...
+                info_map,
+            )
             raise ValueError("nitrogen_fraction must not be equivalent to emergence_nitrogen_fraction")
-        if nitrogen_fraction == mature_nitrogen_fraction:  # leads to log(0)
+        if nitrogen_fraction == mature_nitrogen_fraction:
+            om.add_error(
+                "Invalid value pair for nitrogen_fraction and mature_nitrogen_fraction.",
+                f"Equal nitrogen_fraction and mature_nitrogen_fraction will lead to log(0) calculation"
+                f" error, nitrogen_fraction and mature_nitrogen_fraction are both."
+                f" {mature_nitrogen_fraction}.",
```
ew3361zh: Same comment as the others.

ew3361zh: I would rewrite this one similar to the other suggestions.

RUFAS/routines/field/crop/nitrogen_incorporation.py
```diff
@@ -481,24 +497,61 @@ def _determine_shape_log(
...
             raise ValueError("nitrogen_fraction must not be equivalent to mature_nitrogen_fraction")
-        if (
-            nitrogen_fraction > emergence_nitrogen_fraction or nitrogen_fraction == emergence_nitrogen_fraction
-        ):  # leads to ln(-y) or divide by 0
+        if nitrogen_fraction > emergence_nitrogen_fraction:
+            om.add_error(
+                "Invalid value pair for nitrogen_fraction and emergence_nitrogen_fraction.",
+                f"nitrogen_fraction must be less than emergence_nitrogen_fraction to avoid ln(-y) "
+                f"calculations, nitrogen_fraction is {nitrogen_fraction}, emergence_nitrogen_fraction is "
+                f"{emergence_nitrogen_fraction}.",
```
ew3361zh: Similar rewrite suggestion to the others.

RUFAS/routines/field/crop/nitrogen_incorporation.py
```diff
@@ -481,24 +497,61 @@ def _determine_shape_log(
...
+                info_map,
+            )
             raise ValueError("nitrogen_fraction must be less than emergence_nitrogen_fraction")
-        if nitrogen_fraction == 0:  # leads to ln(0)
+        if nitrogen_fraction == 0:
+            om.add_error(
+                "Invalid nitrogen_fraction.", "nitrogen_fraction must be greater than 0, received 0.", info_map
```
ew3361zh: Does nitrogen_fraction need to be greater than zero or just not equal to zero?

RUFAS/routines/field/crop/nitrogen_incorporation.py
```diff
@@ -481,24 +497,61 @@ def _determine_shape_log(
...
+            )
             raise ValueError("nitrogen_fraction must be greater than 0")
         if heat_fraction == 0:
+            om.add_error("Invalid heat_fraction.", "heat_fraction must be greater than 0, received 0", info_map)
```
ew3361zh: Similar comment - the code seems to indicate that the heat fraction should only not equal 0, nothing about whether greater or lesser than 0 is good or bad.

RUFAS/routines/field/crop/nitrogen_incorporation.py
```diff
@@ -508,6 +561,17 @@ def _determine_shape_log(
 
         # additional check
         if denominator > 1:  # leads to log(-y)
+            om.add_error(
+                "Invalid value pair for nitrogen_fraction and mature_nitrogen_fraction or"
+                " emergence_nitrogen_fraction and mature_nitrogen_fraction.",
```
ew3361zh: Small issue but maybe reword to allow for the case where both value pairs are invalid.

RUFAS/routines/field/crop/nitrogen_incorporation.py
```diff
@@ -644,6 +708,14 @@ def determine_deepest_accessible_layer(root_depth: float, layer_bounds: List[flo
 
         """
         if root_depth < 0.0:
+            info_map = {
+                "class": NitrogenIncorporation.__class__.__name__,
+                "function": NitrogenIncorporation.determine_deepest_accessible_layer.__name__,
+            }
+            om = OutputManager()
+            om.add_error(
+                "Invalid root depth.", f"Negative root depth is invalid, provided root depth is {root_depth}.", info_map
```
ew3361zh:
```suggestion
                "Invalid root depth.", f"Root depth must be >= 0, provided root depth is {root_depth}.", info_map
```

RUFAS/routines/field/crop/nitrogen_incorporation.py
```diff
@@ -694,13 +766,28 @@ def determine_layer_nutrient_uptake_potential(
...
         sorted_boundaries = layer_bounds.copy()
         sorted_boundaries.sort()
         if sorted_boundaries != layer_bounds:
+            om.add_error(
+                "Invalid layer boundaries order.",
+                f"Boundaries must be in ascending order (deeper layers follow shallower ones),"
+                f" received {layer_bounds}.",
+                info_map,
+            )
             raise ValueError("boundaries must be in ascending order (deeper layers follow shallower ones)")
```
ew3361zh, line 777: I'm not exactly sure why there needs to be a valueerror raised here or even a check on this. It seems like the layer_bounds should just be sorted if that's what is needed here? If the layer_bounds should already be sorted prior to arriving at this function, that should be checked earlier.

RUFAS/routines/field/crop/nitrogen_incorporation.py
```diff
@@ -815,7 +912,19 @@ def determine_layer_nutrient_uptake(
...
         # ensure all list inputs are the same length
+        info_map = {
+            "class": NitrogenIncorporation.__class__.__name__,
+            "function": NitrogenIncorporation.determine_layer_nutrient_uptake.__name__,
+        }
+        om = OutputManager()
         if len(layer_uptake_potentials) != len(layer_demands) or len(layer_uptake_potentials) != len(layer_nutrient):
+            om.add_error(
+                "Invalid layer_potential, layer_demand, and layer_nitrate length.",
+                "layer_potential, layer_demand, and layer_nitrate length does not have equal length,"
```
ew3361zh:
```suggestion
                "layer_potential, layer_demand, and layer_nitrate length do not have equal length,"
```

ew3361zh (commented):
Nice work. Made some suggestions on wording and have a couple questions about the necessity of certain warnings. Will approve when these have all been addressed.

RUFAS/routines/field/crop/nitrogen_incorporation.py
matthew7838: It's a fraction so I'm pretty sure it also has to be positive, do you think changing the assertion to <= 0 would be better? I think it's logically the same so not much preference from me.

ew3361zh: Yea I think that would make sense to me if a negative nitrogen_fraction would also produce an error.

RUFAS/routines/field/crop/nitrogen_incorporation.py
matthew7838: This could be related to #548 and it's best to keep this check here for now to at least have a minimum check on the sorting somewhere in the code.

ew3361zh (approved):

PR #2128, opened 2024-12-03
[Animal] Improve culling-related metadata descriptions

### LINKED ISSUE #2056: Improve culling -data properties descriptions
## Status

The input parameters definitions/metadata properties file descriptions for the animal -> from_literature -> culling inputs are accurate but not clear enough for a user to understand what is required without a deeper understanding of the RuFaS life cycle model. 

This leads to confusion in model use and hinders the utility of the data collection app.

## Intention

I would like the definitions of culling input parameters provided in the metadata properties file to be more easily understood and enable a model user to provide the correct information without doing a deep dive into the innerworkings of the life cycle module

## Request

My request is to update the culling input parameter metadata properties

### Target deadline

How about a month from now? December 6th

YijingGong:
### Context

Issue(s) closed by this pull request: closes #2056 

### What

Update metadata description, to be more specific, make animal -> from_literature -> culling inputs metadata descriptions more descriptive and understandable to users. 

### Why

Because the current culling-related metadata descriptions are not clear enough for a general user to understand what is required without a deeper understanding of the RuFaS life cycle model, which leads to confusion in model use and hinders the utility of the data collection app.

### How

I added more descriptions and explanations about what each culling input means. But I am not sure whether talking about the cumulative distribution function (CDF) makes it more confusing for general users or makes it more understandable. 

### Test plan

No testing plan because this metadata update does not change the code function.


PooyaHekmati (approved):
[LGTM](https://a.storyblok.com/f/252228/59787a8d48/english-grammar-help-clearer-or-more-clear.jpg/m/762x0)

jadamchick:
Hi Yijing, this looks really great. What do you think about changing or at least clarifying that everywhere it says "cull" it means "sale" rather than "sale or death"? it's a great opportunity to be more precise in the language used. We don't need to update the actual code, but could clarify in the metadata descriptions. 
I'm happy to go through and make suggested updates if you think it's a good idea

YijingGong:
> Hi Yijing, this looks really great. What do you think about changing or at least clarifying that everywhere it says "cull" it means "sale" rather than "sale or death"? it's a great opportunity to be more precise in the language used. We don't need to update the actual code, but could clarify in the metadata descriptions. I'm happy to go through and make suggested updates if you think it's a good idea

That is a great idea! I just by default think culling is removing animals while they are alive. But since here we are trying to make the description more descriptive and understandable to users, I think it is the best place to say that explicitly to increase the clarity. Please go ahead and update them in this branch, or leave more specific comments and I can update as well, whichever you prefer.

jadamchick:
> That is a great idea! I just by default think culling is removing animals while they are alive. But since here we are trying to make the description more descriptive and understandable to users, I think it is the best place to say that explicitly to increase the clarity. Please go ahead and update them in this branch, or leave more specific comments and I can update as well, whichever you prefer.

Yes, that makes sense! The possible confusion is because the term "culled" can also refer to any animal that has left the herd, whether dead or alive.
After reading through again, your updates have definitely helped to clarify things; I added only a few minor changes.

YijingGong:
@jadamchick 
I just merged the most updated `dev` branch into this one. I think this PR is ready to be merged to `dev`. One more approval is needed. Could you take a look and approve it, or leave further comments.

jadamchick (approved):

PR #2132, opened 2024-12-11
[FeedStorage] Fixing nutrient losses in feed storage E2E tests 

### LINKED ISSUE #2044: [Feed Storage] Investigate nutrient losses in Feed Storage E2E testing routine
The nutrient loss routines of the Feed Storage module need to be carefully evaluated and in some cases corrected.

## Status

With the correction of the Feed Storage degradation routines in #2014, the E2E testing results were evaluated and updated to better test the Feed Storage module. But the updates mainly focused on making sure that stored crops lost mass correctly, and not necessarily that they lost nutrients correctly.

## Intention

Ensure that nutrients are lost from stored crops as time goes by and they degrade.

## Request

Evaluate the nutrient compositions of feeds that are stored as part of the Feed Storage E2E testing routine, and confirm that they are correct and robust by evaluating the actual results of E2E testing with SMEs.

A particular aspect that needs to be looked into is the loss of crude protein in Unprotected Hay. The crude protein mass goes to zero before the hay is finished losing dry matter, so a number of warnings are generated because the Hay is attempting to lose crude protein even though it has none left. Please determine whether this indicates a bug, or is reasonable behavior based on inputs and routines. If this behavior is found to be reasonable, please consider changing how the warning is raised or removing it entirely, because the current warning is very unspecific and repetitive.

### Additional Information

The lead SME who should be worked with for this issue is @KevinP-B

matthew7838:
This PR evaluates the feed storage outputs in the e2e testing, especially the crude protein loss was reaching 0 too fast.
### Context

Issue(s) closed by this pull request: closes #2044

### What

- Recalculate the CP loss coefficient every time when we try to use it across all kinds if hay storages.
- The adjustment is based on the fraction of CP content in dry matter.

### Why

- This will correctly address the coefficient.

### How

- The coefficient will be much smaller and align with the actual scientific dicumentation.

### Test plan

1. Outputs approved by Kevin.
2. Rerun the e2e testing and make sure the CP content in the outputs is not reaching 0.


changelog.md
Isgotkowitz, line 1: The entry for 2081 should not be changed.

Isgotkowitz (commented):
I think this is almost good to go Matthew. There is one removed changelog entry that needs to be added back in, and the expected end-to-end testing results for feed storage need to be updated.

ew3361zh:
Same comments as Ed. Looks good, once the FeedStorage e2e results are updated this should be good to go.

ew3361zh (approved):
LGTM!

Isgotkowitz (approved):
LGTM

PR #2133, opened 2024-12-11
[Animal] Handle date out of range in Genetics module

### LINKED ISSUE #2053: [Animal] Lack of genetic data causes simulation to crash if start date is too early
RuFaS crashes in a not so graceful way if the start date is too early.

## Expected Behavior

RuFaS should be able to start and end on any arbitrary dates, assuming that there is enough weather data to support that simulation and the end date is after the start date.

## Current Behavior

If the simulation starts before Tuesday, May 22 of 2012 (day 122), RuFaS crashes with the error:
```
[01-Nov-2024_Fri_15-19-16.525102][ERROR][default] Failed to finish task: 1/1 with output prefix: default. Failed to recover from error: '2006-01'; traceback: Traceback (most recent call last):
  File "/home/isgotkowitz/rufas/RUFAS/task_manager.py", line 406, in task
    TaskManager.call_handler(
  File "/home/isgotkowitz/rufas/RUFAS/task_manager.py", line 340, in call_handler
    handler(args, input_manager, output_manager, task_id, produce_graphics)
  File "/home/isgotkowitz/rufas/RUFAS/task_manager.py", line 665, in _handle_simulation_engine_run_tasks
    TaskManager.handle_single_simulation_run(args, output_manager)
  File "/home/isgotkowitz/rufas/RUFAS/task_manager.py", line 451, in handle_single_simulation_run
    TaskManager.handle_herd_initializaition(args, output_manager)
  File "/home/isgotkowitz/rufas/RUFAS/task_manager.py", line 440, in handle_herd_initializaition
    herd_factory.initialize_herd()
  File "/home/isgotkowitz/rufas/RUFAS/routines/animal/life_cycle/herd_factory.py", line 409, in initialize_herd
    self.pre_animal_population = self._initialize_herd_from_data(time)
  File "/home/isgotkowitz/rufas/RUFAS/routines/animal/life_cycle/herd_factory.py", line 301, in _initialize_herd_from_data
    cows = list(
  File "/home/isgotkowitz/rufas/RUFAS/routines/animal/life_cycle/herd_factory.py", line 256, in _init_animal_from_data
    net_merit=AnimalGenetics.assign_net_merit_value_to_animals_entering_herd(
  File "/home/isgotkowitz/rufas/RUFAS/routines/animal/genetics/animal_genetics.py", line 200, in assign_net_merit_value_to_animals_entering_herd
    average = AnimalGenetics.net_merit[breed][birth_year_month]["average"]
KeyError: '2006-01'

[01-Nov-2024_Fri_15-19-16.551328][ERROR][Task Manager] Task(s) failed. Failed task(s) and output prefix are: ['default (1/1)']
```

## Possible Solution

Use the earliest known data for any dates from before the actual data begins, and log an error telling the user that this happened.

## Steps to Reproduce

1. Set weather data of default simulation to be `swat_weather.json`
2. Set the start date to be "2012:122" (or any date before then)
3. Run the simulation

## Detailed Description

@allisterakun and @YijingGong will be the best people to consult when implementing a solution to this problem.

Isgotkowitz:
Prevents the Animal Genetics module from crashing if there isn't enough genetic data to support animals in the simulations.

### Context

Issue(s) closed by this pull request: closes #2053 

### What

If a simulation was started early enough or ended late enough, RuFaS would attempt to assign genetic data to newborn calves or animals entering the herd with data that was not provided in the input. This resulted in a KeyError that would crash the simulation.

To fix this, when the Genetics model is initialized it grabs the earliest and latest dates of the available data. It then checks that the date being used for animal is between the earliest and latest dates, and clamps it if it is not.

### Why

RuFaS should just never crash in general. Moreover, if it is possible continue the simulation while logging an error, then it should be done.

Also, RuFaS is about much more than modeling animal genetics, so if someone wants to run a simulation and doesn't care about the genetic outcomes of the herd, they should be able to run it and then ignore the genetic outcomes. 

### How

When the animal genetics module is "initialized", four class attributes are set: the earliest and latest dates of net merit data, and the earliest and latest dates of top semen data. Whenever net merit is assigned to a newborn calf or a cow entering the herd, the birth date of that animal is checked to see if it is between the earliest and latest dates of genetic data available for either net merit or top semen.

If the birth date is before the earliest date or after the latest date, then the data from the earliest or latest date is used, respectively. An error is also logged to the OutputManager, alerting the user that animal's birthdate was not in the available data and the date that was used instead.

### Test plan

Updated the unit test suite, keeping test coverage at 100%.

Ran simulations from 2000-2004 and 2030-2034 to test that they correctly pulled the earliest and latest genetic data, respectively. Also ran the default simulation to verify that none of the animal birthdates were out of range.

### Note to reviewers
I went with the simplest approach I could think of to solve this problem. In some ways this is good, but there are some drawbacks:
- Tons of errors. Users can get flooded with animal genetics errors, and which could potentially obscure more important errors.
- Bogus genetics outcomes. Genetic outcomes will be crap if the data doesn't support the animals' birthdates. I figured this was ok because anyone who is interested in the genetic outcomes will want to only be running simulations in the supported date range.

An alternative approach would be to extrapolate the genetic for dates outside the provided range as they come up.


RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -22,6 +23,10 @@
 }
 
 
+"""The only breed currently supported in this module is Holsteins."""
+SUPPORTED_BREED: str = "HO"
```
Isgotkowitz, line 27: I felt a little bad rewriting "HO" so many times, so I thought it would be better to hardcode and document it.

input/data/weather/barnyard_100yr.csv
Isgotkowitz, line 1: Ed: revert this before merging

input/metadata/default_metadata.json
Isgotkowitz, line 1: Ed: revert this before merging

tests/animal_module_tests/genetics/test_animal_genetics.py
```diff
@@ -667,14 +667,19 @@ def test_initialize_class_variables(mocker: MockerFixture) -> None:
     mock_im_init = mocker.patch("RUFAS.input_manager.InputManager.__init__", return_value=None)
     mock_im_get_data = mocker.patch(
         "RUFAS.input_manager.InputManager.get_data",
-        side_effect=[{"year_month": [], "average": [], "std": []}, {"year_month": [], "estimated_PTA": []}],
+        side_effect=[
+            {"year_month": ["2006-01", "2006-12"], "average": [-1000.0, -900.0], "std": [100.0, 110.0]},
+            {"year_month": ["2007-03", "2008-01"], "estimated_PTA": [-620.0, -610.0]},
```
Isgotkowitz, line 673: Dummy data, just to test that the earliest and latest dates for net merit and top semen are found correctly.

input/data/weather/barnyard_100yr.csv
Isgotkowitz, line 1: Ed: revert this change before merging

input/metadata/default_metadata.json
Isgotkowitz, line 1: Ed: revert this change before merging.

RUFAS/routines/animal/genetics/animal_genetics.py
YijingGong: Just thinking of the expandable principle, if we later have 'JE' would be easy to add?

YijingGong (commented):
Looks good to me! I think sencitically it makes more sense to extrapolate, but since the submodule is under construction, and no one is actually using it, the current solution is sufficient to prevent the program collapse. 

 I tested several start and end dates for the simulation (1995 to 1997, 2000 to 2003, and 2020 to 2030). I am not familiar with weather data (it is using barnyard_100yr.csv in the default_metadata.json), but when I ran with 
```
"start_date": "1995:244", and 
    "end_date": "1997:243"
```
, it crashed due to the weather data error: 
```
[13-Dec-2024_Fri_16-00-46.434198][ERROR][default] Failed to finish task: 1/1 with output prefix: default. Failed to recover from error: 'Attempted to get weather conditions for day: 366, year: 1996.'; traceback: Traceback (most recent call last):
  File "/Users/yijinggong/Desktop/MASM/RUFAS/weather.py", line 124, in get_current_day_conditions
    self.weather_data[time.current_date].daylength = daylength
KeyError: datetime.datetime(1996, 12, 31, 0, 0)
```
Same for `"start_date": "2020:244",
    "end_date": "2030:243",`, maybe I missed something?
Another thing I noticed is that it is called an error: `[13-Dec-2024_Fri_16-04-10.435437][ERROR][default] Animal birthdate out of range for animal genetics data. No top semen data for 2003-08, using data from closest available date: 2006-01`, but eventually it shows `Finished task: TASK_MANAGER with 0 error(s), 113 warning(s), and 18 log(s).`, maybe it's confusing to users? But it looks like an output manager issue.

Isgotkowitz:
> Looks good to me! I think sencitically it makes more sense to extrapolate, but since the submodule is under construction, and no one is actually using it, the current solution is sufficient to prevent the program collapse.

Awesome! I will write up follow up issue for finding a more permanent solution.

> I tested several start and end dates for the simulation (1995 to 1997, 2000 to 2003, and 2020 to 2030). I am not familiar with weather data (it is using barnyard_100yr.csv in the default_metadata.json), but when I ran with
> 
> ```
> "start_date": "1995:244", and 
>     "end_date": "1997:243"
> ```
> 
> , it crashed due to the weather data error:
> 
> ```
> [13-Dec-2024_Fri_16-00-46.434198][ERROR][default] Failed to finish task: 1/1 with output prefix: default. Failed to recover from error: 'Attempted to get weather conditions for day: 366, year: 1996.'; traceback: Traceback (most recent call last):
>   File "/Users/yijinggong/Desktop/MASM/RUFAS/weather.py", line 124, in get_current_day_conditions
>     self.weather_data[time.current_date].daylength = daylength
> KeyError: datetime.datetime(1996, 12, 31, 0, 0)
> ```
> 
> Same for `"start_date": "2020:244", "end_date": "2030:243",`, maybe I missed something? Another thing I noticed is that it is called an error: `[13-Dec-2024_Fri_16-04-10.435437][ERROR][default] Animal birthdate out of range for animal genetics data. No top semen data for 2003-08, using data from closest available date: 2006-01`, but eventually it shows `Finished task: TASK_MANAGER with 0 error(s), 113 warning(s), and 18 log(s).`, maybe it's confusing to users? But it looks like an output manager issue.

The issue with `barnyard_100yr.csv` is that it is missing the leap days for a lot of years. I added a few days with dummy data, but they only work for 2000-2004 and 2030-2034. 

`Finished task: TASK_MANAGER with 0 error(s), 113 warning(s), and 18 log(s).` -- This message indicates that Task Manager was able to run without any errors, but the task (in this case, the default simulation) didn't even finish.

RUFAS/routines/animal/genetics/animal_genetics.py
Isgotkowitz: Great question. I think so, it seems like the only tricky thing might be expanding the inputs. But I'm sure we can find a way to do it with minimal code changes when the time comes

RUFAS/routines/animal/genetics/animal_genetics.py
```diff
@@ -22,6 +23,10 @@
 }
 
 
+"""The only breed currently supported in this module is Holsteins."""
```
matthew7838, line 26: Do we want this kind of message as a docstring? I think this docstring should explain the variable and have this as a note or use inline comments.

tests/animal_module_tests/genetics/test_animal_genetics.py
```diff
@@ -1191,3 +1214,31 @@ def test_assign_net_merit_value_to_newborn_calf(
...
+    om = OutputManager()
+    mock_add_error = mocker.patch.object(om, "add_error")
+
+    actual = AnimalGenetics._clamp_birth_year_month_in_data_range(birth_year_month, is_for_net_merit)
+
+    assert actual == expected_year_month
+    if expect_error:
+        mock_add_error.assert_called_once()
+    else:
+        mock_add_error.assert_not_called()
```
matthew7838: I think it's fine to leave it like this if you think it's too much extra work, but I think it's clearer and easier to expand if the unit tests are divided into one function testing all the possible passing cases and one testing all the possible failing cases.

matthew7838 (approved):
LGTM left a few comments on potential improvements.

YijingGong (approved):
Works for the mentioned time range of 2000-2004 and 2030-2034 (due to the weather data limitation). Just approved.

RUFAS/routines/animal/genetics/animal_genetics.py
Isgotkowitz: I'm not sure what you mean. Are you saying that this docstring should be expanded, or put somewhere else?

tests/animal_module_tests/genetics/test_animal_genetics.py
Isgotkowitz: Sure thing! Good suggestion Matthew

allisterakun (approved):
LGTM! Just one minor suggestion to consider: maybe we should add a check to make sure the class variables are initialized in the new `_clamp_birth_year_month_in_data_range()` function (given that it will be called every time we're trying to retrieve a net_merit / top_semen value)

Isgotkowitz:
> LGTM! Just one minor suggestion to consider: maybe we should add a check to make sure the class variables are initialized in the new `_clamp_birth_year_month_in_data_range()` function (given that it will be called every time we're trying to retrieve a net_merit / top_semen value)

What do you think should happen if the class variables are not set? I think the only option would be to log an error then crash, which wouldn't be all that different from what it would do now -- just crashing with a NoneType error

Isgotkowitz:
After some discussion, decided not to check if the class variables are set in `_clamp_birth_year_month_in_data_range`. If that function is reached before the class variables have been initialized with the dedicated setter, that indicates a bug in the code and crashing out will make us fix it.

An alternative would be call the setter if it has not been called already, if it is decided at a later date that this would be a better way to go then this can be implemented.

Isgotkowitz:
Not sure why the README on `dev` doesn't have the same number of Mypy errors as what is found in the Action on this branch. But it is only going up by 3, and none of them are on lines changed in this PR so I think it is ok to merge. Hopefully the issue will fix itself.

Issue #2130, opened 2024-12-05
[Animal] Refresh the calculation of animal nutritive requirements


Isgotkowitz:
The methods for calculating animal nutritive requirements need to be updated in order for them to be compatible with the refreshed animal module.

## Status

The old calculations of animal requirements in `animal/ration/animal_requirements.py` have not been updated yet to work with the refreshed animal code in `biophysical`.

## Intention

Have the code that calculates animal requirements be written in a cleaner, more maintainable way which will utilize new data structures.

## Request

Rewrite the code that calculates animal nutritive requirements to work with the refreshed `Animal` class and submodules.

PR #2141, opened 2024-12-18
[Animal] Refresh NASEM animal requirements calculator

Isgotkowitz:
Re-implements the NASEM equations for calculating animal requirements.

### Context

Issue(s) closed by this pull request: addresses #2130 

### What

Kicks off the implementation of animal requirements' calculations, calculating the amounts of energy and nutrients that animals need based on their life stage, lactation status, pregnancy stage, etc.

There are two different methodologies for calculating animal requirements, NASEM and NRC. The NASEM methodology is implemented in this PR, and the NRC methodology will be implemented in a follow up PR.

### Why

The requirements of animal determine how it is fed, and if what it is fed supports all the growing, maintenance, and other activities that the animal engages in.

### How

~An abstract base class, `NutritionRequirementsCalculator` defines the interface which both the NASEM and NRC requirements calculations will have to meet. It contains mostly abstract `@classmethod`s, which does a nice job of defining an interface but ensuring that any subclasses will function as calculators.~
Scratch the above, I found that it made Mypy throw a fit. `NutritionRequirementsCalculator` will still be the baseclass for NASEM and NRC requirements classes and will have one function that is shared. But it will not have an interface, which is too bad but still an improvement over the old animal code.

`NASEMRequirementsCalculator` implements the interface of `NutritionRequirementsCalculator`, with the equations being copied and pasted from `ration/animal_requirements.py`.

### Test plan

A partial unit test suite has been written for `NASEMRequirementsCalculator`. After the animal refresh reaches the point where full simulations can be run with the refreshed animal module, the test suite will be revisited and filled out.

### DIFF
--- a/RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
+++ b/RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
@@ -672,7 +672,7 @@ class NASEMRequirementsCalculator(NutritionRequirementsCalculator):
     def _calculate_activity_energy_requirements(
         cls,
         body_weight: float,
-        housing: float,
+        housing: str,
         distance: float,
     ) -> float:
         """
--- a/RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
+++ b/RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
@@ -95,7 +95,7 @@ class NASEMRequirementsCalculator(NutritionRequirementsCalculator):
         lactation_requirement = cls.calculate_lactation_energy_requirements(
             animal_type, milk_fat, milk_true_protein, milk_lactose, milk_production
         )
-        dry_matter_intake = cls.calculate_dry_matter_intake(
+        dry_matter_intake = cls._calculate_dry_matter_intake(
             body_weight,
             mature_body_weight,
             days_in_milk,
@@ -621,7 +621,7 @@ class NASEMRequirementsCalculator(NutritionRequirementsCalculator):
         return max(phosphorus_requirement, AnimalModuleConstants.MINIMUM_PHOSPHORUS)
 
     @classmethod
-    def calculate_dry_matter_intake(
+    def _calculate_dry_matter_intake(
         cls,
         body_weight: float,
         mature_body_weight: float,
@@ -641,10 +641,10 @@ class NASEMRequirementsCalculator(NutritionRequirementsCalculator):
             Body weight (kg)
         mature_body_weight : float
             Mature body weight (kg)
-        days_in_milk : int
+        days_in_milk : int | None
             Days in milk (days)
         lactating : bool
-            Physiological condition (conditional)
+            True if animal is lactating, else false.
         net_energy_lactation : float
             Net energy for lactation
         parity : int
@@ -673,9 +673,7 @@ class NASEMRequirementsCalculator(NutritionRequirementsCalculator):
 
         """
         if lactating:
-            parity_adjustment_factor = 0
-            if parity > 1:
-                parity_adjustment_factor = 1
+            parity_adjustment_factor = 1 if parity > 1 else 0
             dry_matter_intake_estimate = (
                 (3.7 + parity_adjustment_factor * 5.7)
                 + 0.305 * net_energy_lactation
--- a/RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
+++ b/RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
@@ -694,10 +694,9 @@ class NASEMRequirementsCalculator(NutritionRequirementsCalculator):
 
         Notes
         -----
-        Note that both NRC and NASEM calculations use distance walked in kilometers, hence the unit conversion in the
-        code itself. Activity requirement (net_energy_activity) is proportional to body weight and daily walking
-        distance. Grazing system and hilly topography will cost additional energy. Grazing is not implemented yet in the
-        current version of code.
+        NASEM calculations use distance walked in kilometers, hence the unit conversion. Activity requirement
+        (net_energy_activity) is proportional to body weight and daily walking distance. Grazing system and hilly
+        topography will cost additional energy. Grazing is not implemented yet in the current version of code.
 
         This is separate because it must be calculated after grouping due to pen input args and cannot be used
         individually on an animal.
--- a/RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
+++ b/RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
@@ -125,7 +125,7 @@ class NASEMRequirementsCalculator(NutritionRequirementsCalculator):
             milk_production,
             parity,
         )
-        phosphorus_requirement = cls.calculate_phosphorus_requirement(
+        phosphorus_requirement = cls._calculate_phosphorus_requirement(
             body_weight,
             mature_body_weight,
             animal_type,
@@ -536,7 +536,7 @@ class NASEMRequirementsCalculator(NutritionRequirementsCalculator):
         return max(calcium_requirement, AnimalModuleConstants.MINIMUM_CALCIUM)
 
     @classmethod
-    def calculate_phosphorus_requirement(
+    def _calculate_phosphorus_requirement(
         cls,
         body_weight: float,
         mature_body_weight: float,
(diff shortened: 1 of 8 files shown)


RUFAS/biophysical/animal/nutrients/nutrition_requirements_calculator.py
```diff
@@ -0,0 +1,98 @@
...
+    def _calculate_growth_energy_requirements(cls, **kwargs: dict[str, Any]) -> tuple[float, float, float]:
+        pass
+
+    @classmethod
+    @abstractmethod
+    def _calculate_pregnancy_energy_requirements(cls, **kwargs: dict[str, Any]) -> float:
+        pass
+
+    @classmethod
+    def _calculate_lactation_energy_requirements(
```
Isgotkowitz, line 10: NASEM and NRC have the same methodology for calculating this, so putting the logic for this in the base class deduplicates code.

RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
```diff
@@ -0,0 +1,719 @@
...
+            growth_req = 0.0
+
+        if day_of_pregnancy is None:
+            pregnancy_req: float = 0.0
+        else:
+            pregnancy_req = 0.02456 * exp((0.05581 - 0.00007 * day_of_pregnancy) * day_of_pregnancy) - 0.02456 * exp(
+                (0.05581 - 0.00007 * (day_of_pregnancy - 1)) * (day_of_pregnancy - 1)
+            ) * (body_weight / 715)
+
+        lactation_req: float = (0.295 + 0.239 * milk_true_protein) * milk_production
```
Rebecavb, line 502: In the NASEM book equation for Lactation requirement there is not a final multiplication times milk production. The equation in the book is: Milk Ca = 0.295 (±0.73) + 0.239 (±0.029) × milk true protein percent

RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
```diff
@@ -0,0 +1,719 @@
...
+            pregnancy_req = (
+                (
+                    0.02743 * exp((0.05527 - 0.000075 * day_of_pregnancy) * day_of_pregnancy)
+                    - 0.02743 * exp((0.05527 - 0.000075 * (day_of_pregnancy - 1)) * (day_of_pregnancy - 1))
+                )
+                * body_weight
+                / 715
+            )
+
+        lactation_req = milk_production * (0.49 + 0.13 * milk_true_protein)
```
Rebecavb, line 583: The book also includes another lactation phosphorous requirement in case the milk protein is unknown (Milk protein is unknown: Milk yield (kg/d) × 0.90). 

Just a comment in case you want to consider adding that one too

RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
Isgotkowitz: Do you have a link to where this is in the book? Edit: I think this multiplication is present, equation 20-376 in the NASEM book has "(0.295 + 0.239 × MlkNP_Milk) × An_Mlk Prod" on page 455.

RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
Isgotkowitz: This is good to know! I think the milk protein percentage will always be known, either because the user sets it or it is calculated, so we should be good.

RUFAS/biophysical/animal/nutrients/nutrition_requirements_calculator.py
```diff
@@ -0,0 +1,102 @@
...
+class NutritionRequirementsCalculator(ABC):
+    """
+    Defines the interface for animal nutritional requirements calculation, and logic shared between the NASEM and NRC
+    equations.
+    """
+
+    @classmethod
+    @abstractmethod
+    def calculate_requirements(cls, **kwargs: dict[str, Any]) -> NutritionRequirements:
+        pass
```
Isgotkowitz: I went with the pattern of adding abstract classmethods in `NutritionRequirementsCalculator` to indicate to readers/maintainer that it is an interface, and should followed accordingly by both `NASEMRequirementsCalculator` and `NRCRequirementsCalculator`.

But as I understand it, unimplemented abstractmethods will prevent a class from being instantiated but not from classmethods on that class from being used. So this "interface" will not behave as strictly as I would like.

RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
Rebecavb: NASEM book equation 7-4.
Link:  https://www.ncbi.nlm.nih.gov/books/NBK600600/#pz125-4

Rebecavb (approved):

RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
Isgotkowitz: The unit of the result in equation 7-4 is g / kg of milk. So in order for us to get the total amount of Ca required, we would need to multiply the result of that equation by the milk production in kilograms.

RUFAS/biophysical/animal/nutrients/nutrition_requirements_calculator.py
Isgotkowitz: Scratch this, Mypy hated it so I'm going with a simpler approach.

RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
```diff
@@ -0,0 +1,718 @@
...
+        parity: int,
+        calving_interval: int | None,
+        milk_fat: float,
+        milk_true_protein: float,
+        milk_lactose: float,
+        milk_production: float,
+        housing: str,
+        distance: float,
+        lactating: bool,
+        ndf_percentage: float,
```
Isgotkowitz, line 38: These data types are carried over from the original code, which had a lot of Mypy errors. Because so much of this code is copied and pasted, a lot of these errors have been carried over. I think this is acceptable for right now, as we are pushing so hard to get the refresh to a place where the animal module is running. But after this, we can come back and fix this code in the context of the entire refreshed animal module.

RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
```diff
@@ -0,0 +1,718 @@
...
+            metabolizable_protein=protein_requirement,
+            calcium=calcium_requirement,
+            phosphorus=phosphorus_requirement,
+            dry_matter=dry_matter_intake,
+            activity_energy=activity_requirement,
+            essential_amino_acids=essential_amino_acids,
+        )
+
+    @classmethod
+    def _calculate_maintentance_energy_requirements(
```
allisterakun:
```suggestion
    def _calculate_maintenance_energy_requirements(
```
Typo, please also fix all references

RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
```diff
@@ -0,0 +1,718 @@
...
+        cls, body_weight: float, mature_body_weight: float, day_of_pregnancy: int | None, days_in_milk: int | None
```
allisterakun, line 170: Is either `day_of_pregnancy` or `days_in_milk` ever None? Correct me if I'm wrong, but I think the parent method `calculate_requirements()` is always called with keyword names specified.

allisterakun (approved):
Generally looks good! Just a typo and a question.

RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
Isgotkowitz: Even with keyword args specified, they could still be None -- if the animal isn't pregnant or isn't in milk.

More importantly though, these type annotations are holdovers from the old animal code, and I think we should revisit them and the code that generates these values once we get everything connected and running in the refreshed animal module.

RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
Isgotkowitz: Good catch, thanks Allister

PR #2142, opened 2024-12-18
[Animal] Refresh NRC animal requirements calculator

Isgotkowitz:
Reimplements the NRC animal requirements calculations.

### Context

Issue(s) closed by this pull request: addresses #2130 

### What

Refreshes the calculations of an animal's requirements based on the NRC methodology, following up on #2141.

### Why

NRC provides the other methodology RuFaS uses to calculate the energy and nutrition requirements of an animal. Having these requirements calculated determines what an animal should be fed, and if a ration satisfies the energy and nutrient demands of an animal.

### How

Reimplemented the NRC energy and requirements calculations in the class `NRCRequirementsCalculator`, which follows the `NutritionRequirementsCalculator` interface. The code and docstrings are almost entirely copied and pasted from `ration/animal_requirements.py`. I made some minor changes to references, variable names, and formatting to zhush up the code a bit, but otherwise it is entirely the same.

### Test plan

Because the code is entirely copied and pasted, and the time crunch on the refresh, unit tests will be implemented at a later date. If there are any issues in the NRC calculations, they will be caught and swiftly resolved while testing the refreshed animal module.

### DIFF
--- a/RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
+++ b/RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
@@ -35,10 +35,10 @@ class NRCRequirementsCalculator(NutritionRequirementsCalculator):
         TDN_percentage: float,
     ) -> NutritionRequirements:
         """Calculates energy and nutrition requirements for an animal using the NRC methodology."""
-        maintenance_requirement, conceptus_weight, calf_birth_weight = cls.calculate_maintentance_energy_requirements(
+        maintenance_requirement, conceptus_weight, calf_birth_weight = cls._calculate_maintentance_energy_requirements(
             body_weight, mature_body_weight, day_of_pregnancy, body_condition_score_5, previous_temperature, animal_type
         )
-        growth_requirement, average_daily_gain, shrunk_body_weight = cls.calculate_growth_energy_requirements(
+        growth_requirement, average_daily_gain, shrunk_body_weight = cls._calculate_growth_energy_requirements(
             body_weight,
             mature_body_weight,
             conceptus_weight,
@@ -47,7 +47,7 @@ class NRCRequirementsCalculator(NutritionRequirementsCalculator):
             calving_interval,
             average_daily_gain_heifer,
         )
-        pregnancy_requirement = cls.calculate_pregnancy_energy_requirements(day_of_pregnancy, calf_birth_weight)
+        pregnancy_requirement = cls._calculate_pregnancy_energy_requirements(day_of_pregnancy, calf_birth_weight)
         lactation_requirement = cls.calculate_lactation_energy_requirements(
             animal_type, milk_fat, milk_true_protein, milk_lactose, milk_production
         )
@@ -114,7 +114,7 @@ class NRCRequirementsCalculator(NutritionRequirementsCalculator):
         )
 
     @classmethod
-    def calculate_maintentance_energy_requirements(
+    def _calculate_maintentance_energy_requirements(
         cls,
         body_weight: float,
         mature_body_weight: float,
@@ -125,6 +125,7 @@ class NRCRequirementsCalculator(NutritionRequirementsCalculator):
     ) -> tuple[float, float, float]:
         """
         Calculates energy requirement for maintenance, conceptus weight, and calf birth weight according to NRC (2001).
+
         Parameters
         ----------
         body_weight : float
@@ -139,37 +140,39 @@ class NRCRequirementsCalculator(NutritionRequirementsCalculator):
             Adjustment for previous temperature
         animal_type : AnimalType
             A type or subtype of animal specified in the AnimalType enum
+
         Returns
         -------
         tuple[float, float, float]
             Net energy requirement for maintenance (Mcal/day), conceptus weight (kg), and calf birth weight (kg)
+
         Notes
         -----
         Energy requirements for activity are not included within calculations for maintenance.
+
         References
         ----------
         .. [1] National Research Council, "Nutrient Requirements of Dairy Cattle, 7th edition." National Academic Press,
         Chapter 2 "Energy",pp. 18-25, 2001.
+
         """
         calf_birth_weight = mature_body_weight * 0.06275 if day_of_pregnancy else 0.0
         conceptus_weight = 0.0
         if day_of_pregnancy and day_of_pregnancy > 190:
             conceptus_weight = (18 + (day_of_pregnancy - 190) * 0.665) * (calf_birth_weight / 45)
+
         if animal_type in [AnimalType.LAC_COW, AnimalType.DRY_COW]:
             net_energy_maintenance = 0.08 * (body_weight - conceptus_weight) ** 0.75
-        elif animal_type in [
-            AnimalType.HEIFER_I,
-            AnimalType.HEIFER_II,
-            AnimalType.HEIFER_III,
-        ]:
+        elif animal_type in [AnimalType.HEIFER_I, AnimalType.HEIFER_II, AnimalType.HEIFER_III]:
             body_condition_score_9 = (body_condition_score_5 - 1) * 2 + 1
             net_energy_maintenance = (body_weight - conceptus_weight) ** (0.75) * (
                 0.086 * (0.8 + (body_condition_score_9 - 1) * 0.05)
             ) + 0.0007 * (20 - previous_temperature)
+
         return net_energy_maintenance, conceptus_weight, calf_birth_weight
 
     @classmethod
-    def calculate_growth_energy_requirements(
+    def _calculate_growth_energy_requirements(
         cls,
         body_weight: float,
         mature_body_weight: float,
@@ -182,6 +185,7 @@ class NRCRequirementsCalculator(NutritionRequirementsCalculator):
         """
         Calculates energy requirement for growth, average daily gain and estimate of shrunk body weight according to NRC
         (2001).
+
         Parameters
         ----------
         body_weight : float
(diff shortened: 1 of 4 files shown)


RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
```diff
@@ -0,0 +1,646 @@
...
+            Average daily gain (grams per day)
+
+        Returns
+        -------
+        tuple[float, float, float]
+            Net energy requirement for growth (Mcal/d), average daily gain (g/d), equivalent shrunk body weight (kg).
+
+        References
+        ----------
+        Animal Scientific Documentation [A.Cow.A.7]-[A.Cow.A.14], [A.Heifer.A.8]-[A.Heifer.A.15]
```
Isgotkowitz, line 264: I didn't want to take out the references to scientific docs, but I also didn't want to leave them as in-line comments. This was my compromise

RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
```diff
@@ -0,0 +1,647 @@
...
+        maintenance = (
+            0.3 * (body_weight - conceptus_weight) ** 0.6
+            + 4.1 * (body_weight - conceptus_weight) ** 0.5
+            + (
+                dry_matter_intake_estimate * GeneralConstants.KG_TO_GRAMS * 0.03
+                - 0.5 * (bacteria_estimate / 0.68 - bacteria_estimate)
+            )
+            + 0.4 * 11.8 * dry_matter_intake_estimate / 0.67
+        )
+
```
KFosterReed, line 414: The documentation has one difference in the parameters for this equation and says it should be: 
` - 0.5 * (bacteria_estimate / 0.8 - bacteria_estimate)
            )
            + 0.4 * 11.8 * dry_matter_intake_estimate / 0.67
        )`

@tomhuhh could you please double check with the NRC book?

RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
```diff
@@ -0,0 +1,647 @@
...
+        float
+            Phosphorus requirement (grams per day)
+
+        References
+        ----------
+        .. [1] National Research Council, "Nutrient Requirements of Dairy Cattle, 7th edition." National Academic Press,
+            Chapter 6 "Minerals",pp. 109-118. 2001.
+
+        """
+        growth: float = (1.2 + 4.635 * mature_body_weight**0.22 * body_weight ** (-0.22)) * (average_daily_gain / 0.96)
```
KFosterReed, line 552: Is this syntax correct? In the other methods, the equations are written as:

`growth = (1.2 + ...` but here they are written as `growth: float = (1.2 ...`

Just checking to see if this was intentional

RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
tomhuhh: The documentation seems to be correct. The denominator should be 0.8 instead of 0.68 according to the NRC book. See the screenshot below. 

Screenshot 2024-12-29 at 6 00 49 PM

KFosterReed: Thanks Haowen!

Isgotkowitz: Updated! Thank you both

RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
Isgotkowitz: Yes, this is intentional. For whatever reason, Mypy (the program which checks that code uses the correct types) doesn't think the result of this equation will be a float and yells at us. But adding `: float ...` tells Mypy explicitly that we are confident the result of this equation will be a float.

KFosterReed (approved):
Nice work Ed!

RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
```diff
@@ -0,0 +1,647 @@
...
+        milk_lactose: float,
+        milk_production: float,
+        housing: str,
+        distance: float,
+        previous_temperature: float | None,
+        net_energy_diet_concentration: float,
+        days_born: float,
+        TDN_percentage: float,
+    ) -> NutritionRequirements:
+        """Calculates energy and nutrition requirements for an animal using the NRC methodology."""
```
matthew7838: Not sure if these are shortened due to time limitations, or else this should have an attribute and return section.

RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
```diff
@@ -0,0 +1,647 @@
...
+            net_energy_maintenance = 0.08 * (body_weight - conceptus_weight) ** 0.75
+        elif animal_type in [AnimalType.HEIFER_I, AnimalType.HEIFER_II, AnimalType.HEIFER_III]:
+            body_condition_score_9 = (body_condition_score_5 - 1) * 2 + 1
+            net_energy_maintenance = (body_weight - conceptus_weight) ** (0.75) * (
+                0.086 * (0.8 + (body_condition_score_9 - 1) * 0.05)
+            ) + 0.0007 * (20 - previous_temperature)
+
+        return net_energy_maintenance, conceptus_weight, calf_birth_weight
+
+    @classmethod
```
matthew7838, line 225: Any specific reason to use a class method? I see this class more like a utility so maybe static methods make more sense to me.

RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
```diff
@@ -0,0 +1,647 @@
...
+        if animal_type in [AnimalType.LAC_COW]:
+            fat_corrected_milk_kg = (0.4 * milk_production) + (15 * milk_fat * (milk_production / 100))
+            dry_matter_intake_estimate: float = (0.372 * fat_corrected_milk_kg + 0.0968 * body_weight**0.75) * (
+                1 - exp(-0.192 * ((days_in_milk / 7) + 3.67))
+            )
+        elif animal_type in [AnimalType.DRY_COW]:
+            dry_matter_intake_estimate = ((1.97 - 0.75 * exp(0.16 * (day_of_pregnancy - 280))) / 100) * body_weight
+        else:
+            if days_born and days_born > 365:
+                value_to_use = 0.1128
```
matthew7838, line 637: This naming is confusing.

RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
```diff
@@ -0,0 +1,647 @@
...
+    ) -> float:
+        """
+        Calculates the net energy for activity requirement portion of the energy requirements.
+
+        Parameters
+        ----------
+        body_weight : float
+            Body weight (kg)
+        housing : str
+            Housing type (Barn or Grazing)
```
matthew7838:
```suggestion
            Body weight (kg).
        housing : str
            Housing type (Barn or Grazing).
```

RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
```diff
@@ -0,0 +1,647 @@
...
+            pregnancy = 0.0
+
+        if animal_type in [AnimalType.LAC_COW]:
+            maintenance: float = 1 * dry_matter_intake_estimate + 0.002 * body_weight
+            lactation: float = 0.9 * milk_production
+            phosphorus_requirement: float = growth + pregnancy + lactation + maintenance
+        elif animal_type in [AnimalType.HEIFER_I, AnimalType.HEIFER_II, AnimalType.HEIFER_III, AnimalType.DRY_COW]:
+            maintenance = 0.8 * dry_matter_intake_estimate + 0.002 * body_weight
+            phosphorus_requirement = growth + pregnancy + maintenance
+
```
matthew7838, line 570: Should there be a catch-all or error-handling case for other types like a calf?

RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
```diff
@@ -0,0 +1,647 @@
...
+        mature_body_weight : float
+            Mature body weight (kg)
+        day_of_pregnancy : int
+            Day of pregnancy (days)
+        body_condition_score_5 : int
+            Body condition score (score from 1 to 5)
+        previous_temperature : float
+            Adjustment for previous temperature
+        animal_type : AnimalType
+            A type or subtype of animal specified in the AnimalType enum
```
matthew7838:
```suggestion
        body_weight : float
            Body weight (kg).
        mature_body_weight : float
            Mature body weight (kg).
        day_of_pregnancy : int
            Day of pregnancy (days).
        body_condition_score_5 : int
            Body condition score (score from 1 to 5).
        previous_temperature : float
            Adjustment for previous temperature.
        animal_type : AnimalType
            A type or subtype of animal specified in the AnimalType enum.
```

matthew7838 (commented):
This looks pretty solid. I have a few questions as mentioned in the comments below.

1.  What is the reason for making these functions a class method instead of a static one?
2.  I know these are copy-pasted, but there are missing periods and mismatched units (some are in parenthesis, and some are not) across the file.
3. There are methods like `_calculate_protein_requirement` that do not handle any else cases like if the animal type is a calf. Is this intentional or should there be some default/error handling for this?

RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
Isgotkowitz: Added!

RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
Isgotkowitz: Only in the public method, `calculate_requirements` because it has to access other methods in the class. And making all other methods in the class `classmethod`s avoids people reading the code and wondering why there is a mix of static and class methods. Otherwise class and static methods would be pretty much interchangeable in this case.

RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
Isgotkowitz: I agree, but all of these routines are copied and pasted from the old animal module so I didn't go through to make any significant improvements on the implementation of the routines.

I think it would be better to move these routines through as-is right now, and we can revisit them once the refreshed animal module is up and running.

RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
Isgotkowitz: Good question! The NASEM and NRC requirements modules only work for non-calf animal types, so it will be the responsibility of a higher level module or function to make sure that the right animal type is routed to the right requirements calculation module.

This isn't a perfect way to structure the code, but it is the same as what was working before and I think it would be better to get the animal refresh running before we start working on improvements to stuff like this.

Isgotkowitz:
@matthew7838 thank you for your review! You brought up some good points on things that can be improved and I addressed some of them, but for others I think it would be better to work on them after the refreshed animal code is up and running. Especially given the time crunch on the refresh and that these routines are mostly being copied over from the old animal module.

RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
matthew7838: That sounds good. Out of curiosity, since static `calculate_requirements` can also access all other static methods, what's the reason not to make them all static? I asked because some parts of the codebase make them all static in similar cases.

Isgotkowitz: It is a super minor/nit-picky thing that I learned about recently, and doesn't have much to do with the actual functionality of the code. But...

When you have a staticmethod, the logic should be entirely self-contained, i.e. it should not have any references to the class that contains it. but if a method is a class method, it has access to all the other class and static methods of the class that contains it through the `cls` parameter.

Take the following code, which would work.
```python
class FooBar:
    
    @staticmethod
    def Foo():
        return FooBar.Bar()
    
    @staticmethod
    def Bar():
        # return some value
```

Even though the above works, it doesn't technically adhere to the intended usage of static and class methods. This would be a more "correct" way of implementing it:
```python
class FooBar:
    
    @classmethod
    def Foo(cls):
        return cls.Bar()
    
    @staticmethod
    def Bar():
        # return some value
```

matthew7838 (approved):
LGTM and lessons learned. Thanks, Ed!

PR #2143, opened 2024-12-18
[GraphGenerator] Add User Option for Setting Time on X-axis for Graphs

### LINKED ISSUE #912: [GraphGenerator] Make Time an option for the x-axis of graphics generated by `GraphGenerator`

ew3361zh:
Adds an user option to set the x-axis in graphs to use calendar dates instead of a simulation day counter.

### Context

Issue(s) closed by this pull request: closes #912

### What

Adds an optional boolean flag in graph filters `use_calendar_dates` to set whether they want the x-axis of their graphs to be calendar dates in the format of ordinal date (e.g. 001/2015, 229/2019). The default setting is `False` which will default the graph to plot using simulation day counter (i.e. what it currently does on dev).

### Why

This feature was requested from SMEs to have an option to see the data presented in a different manner which they said would be more helpful for results analysis.

### How

The `use_calendar_dates` flag is checked for and then sent to the GG's `_draw_graph()` method.
Within `draw_graph()`, the date range of the data to graph is determined based on the start date of the simulation from the Time class instance sent to GG.
A new helper function (`get_x_values()`) is set up within `_draw_graph()` which returns the x-axis values to match whether the user wants calendar dates or not.
Data is then plotted much the same way it was before and the only other difference is if the user does want calendar dates on the x-axis, some additional graph formatting is done to make the graph more readable.

### Test plan

Create a graph filter and try a version with `use_calendar_dates` set to `True` and then try it without that flag in the filter file.
The `True` version should have calendar dates across the x-axis.
Every other version should have the simulation day counter as before.

my example filter file:
`graph_tester.json`

```
{
    "type": "plot",
    "filters": [
      ".*ammonia_emissions.*"
    ],
    "title": "Ammonia Emissions",
    "use_calendar_dates": true
}
```
Screenshot 2024-12-18 at 12 06 47 PM

And then the default version without specifying the `use_calendar_dates` flag or setting it to `False`:
Screenshot 2024-12-18 at 12 10 37 PM

### DIFF
--- a/RUFAS/graph_generator.py
+++ b/RUFAS/graph_generator.py
@@ -9,6 +9,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 import numpy.typing as npt
 from matplotlib.figure import Axes, Figure
+import matplotlib.dates as mdates
 
 from RUFAS.util import Utility
 
@@ -553,18 +554,39 @@ class GraphGenerator:
         """
         if graph_type not in MATPLOTLIB_PLOT_FUNCTIONS:
             raise ValueError(f"Unsupported graph type: {graph_type}")
+        simulation_start_date = self.time.start_date
+        # prepare dates for x-axis
+        num_days = max(len(values) for values in data.values())  # Number of days based on data length
+        dates = [simulation_start_date + datetime.timedelta(days=i) for i in range(num_days)]
         plot_function = MATPLOTLIB_PLOT_FUNCTIONS[graph_type]
+        fig, ax = plt.subplots()
         if graph_type in TUPLE_BASED_FUNCTIONS:
             values_tuple = tuple(data[variable] for variable in selected_variables)
-            indices_list = list(range(len(values_tuple[0])))
-            plot_function(indices_list, values_tuple)
+            # indices_list = list(range(len(values_tuple[0])))
+            # plot_function(indices_list, values_tuple)
+            ax.xaxis_date()
+            plot_function(dates, values_tuple)
         else:
             for value in data.values():
                 if not mask_values:
-                    plot_function(value)
+                    # plot_function(value)
+                    plot_function(dates, value)
                 else:
                     indices, masked_values = self._mask_values(value)
-                    plot_function(indices, masked_values)
+                    masked_dates = [dates[i] for i in indices]
+                    # plot_function(indices, masked_values)
+                    plot_function(masked_dates, masked_values)
+        # Customize the x-axis date format
+        ax.xaxis.set_major_formatter(mdates.DateFormatter('%j/%Y'))  # Day of year / Year format
+        plt.xticks(rotation=45)  # Rotate x-axis labels for better readability
+
+        # Add labels and title
+        plt.xlabel("Calendar Day")
+        plt.ylabel("Values")
+        plt.title("Simulation Data Over Time")
+
+        plt.tight_layout()
+        plt.show()
 
     def _mask_values(self, values: list[Any]) -> tuple[npt.NDArray[Any], npt.NDArray[np.float32]]:
         """
--- a/RUFAS/graph_generator.py
+++ b/RUFAS/graph_generator.py
@@ -225,7 +225,8 @@ class GraphGenerator:
             mask_values = graph_details.get("mask_values", False)
             use_calendar_dates = graph_details.get("use_calendar_dates", False)
             self._draw_graph(
-                graph_details["type"], prepared_data, list(prepared_data.keys()), ax, mask_values, use_calendar_dates
+                graph_details["type"], prepared_data, list(prepared_data.keys()), ax, mask_values, use_calendar_dates,
+                graph_details.get("slice_start", None), graph_details.get("slice_end", None)
             )
             if graph_details.get("title"):
                 corrected_graph_title = Utility.remove_special_chars(graph_details.get("title", "Untitled graph"))
@@ -534,6 +535,8 @@ class GraphGenerator:
         ax: Axes,
         mask_values: bool = False,
         use_calendar_dates: bool = False,
+        slice_start: int | None = None,
+        slice_end: int | None = None,
     ) -> None:
         """
         Draw the graph based on the provided graph type and data.
@@ -549,10 +552,14 @@ class GraphGenerator:
         mask_values : bool, default False
             Whether data that will be plotted with non-tuple based functions should be masked to remove None or NaN
             values.
-        ax : Axes, default None
+        ax : Axes
             The matplotlib Axes object to plot the graph on.
         use_calendar_dates : bool, default False
             Whether to use calendar dates on the x-axis.
+        slice_start : int, default None
+            The starting index of the data to plot.
+        slice_end : int, default None
+            The ending index of the data to plot.
 
         Raises
         ------
@@ -561,14 +568,27 @@ class GraphGenerator:
         """
         if graph_type not in MATPLOTLIB_PLOT_FUNCTIONS:
             raise ValueError(f"Unsupported graph type: {graph_type}")
-        dates_in_data_range = [
-            self.time.start_date + datetime.timedelta(days=i) for i in range(max(len(v) for v in data.values()))
-        ]
+        max_data_length = max(len(v) for v in data.values())
+        if slice_start is not None:
+            if slice_start < 0:
+                slice_start = max(0, self.time.simulation_length_days + slice_start)
+            if slice_end is None or slice_end > self.time.simulation_length_days:
+                slice_end = self.time.simulation_length_days
+            elif slice_end < 0:
+                slice_end = max(0, self.time.simulation_length_days + slice_end)
+            dates_in_data_range = [
+                self.time.convert_simulation_day_to_date(i) for i in range(slice_start, slice_end)
+            ]
+        else:
+            dates_in_data_range = [
+                self.time.start_date + datetime.timedelta(days=i) for i in range(max_data_length)
+            ]
         plot_function = MATPLOTLIB_PLOT_FUNCTIONS[graph_type]
 
-        def get_x_values(values_length: int) -> list[int]:
-            """Get appropriate x-axis values based on user choice."""
-            return dates_in_data_range[:values_length] if use_calendar_dates else list(range(values_length))
+        get_x_values : Callable[[int], list[int]] = (
+            lambda values_length: dates_in_data_range[:values_length] if use_calendar_dates else
+            list(range(values_length))
+        )
 
         if graph_type in TUPLE_BASED_FUNCTIONS:
             values_tuple = tuple(data[variable] for variable in selected_variables)
--- a/RUFAS/graph_generator.py
+++ b/RUFAS/graph_generator.py
@@ -226,8 +226,9 @@ class GraphGenerator:
 
             mask_values = graph_details.get("mask_values", False)
             use_calendar_dates = graph_details.get("use_calendar_dates", False)
-            self._draw_graph(graph_details["type"], prepared_data, list(prepared_data.keys()), mask_values, ax,
-                             use_calendar_dates)
+            self._draw_graph(
+                graph_details["type"], prepared_data, list(prepared_data.keys()), mask_values, ax, use_calendar_dates
+            )
             if graph_details.get("title"):
                 corrected_graph_title = Utility.remove_special_chars(graph_details.get("title"))
                 graph_details["title"] = corrected_graph_title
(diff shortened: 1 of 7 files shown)


RUFAS/graph_generator.py
```diff
@@ -526,9 +531,11 @@ def _log_non_numerical_data(
     def _draw_graph(
         self,
         graph_type: str | list[str],
-        data: Dict[str, List[int | float]],
-        selected_variables: Optional[List[str]] = None,
```
ew3361zh, line 530: This list won't be optional ever since we're pulling out variables to graph based on the filter and using them as `selected_variables` even if the user doesn't specify `selected_variables` in the output filter.

RUFAS/graph_generator.py
```diff
@@ -91,7 +92,6 @@
 FIGURE_SETTERS: Dict[str, FUNCTION_TYPE] = {
     "align_labels": Figure.align_labels,
     "canvas": Figure.set_canvas,
-    "constrained_layout": Figure.set_constrained_layout,
```
ew3361zh, line 94: This along with the other Figure import below were not recognized as viable imports from matplotlib.Figure by MyPy.

RUFAS/graph_generator.py
```diff
@@ -553,18 +561,34 @@ def _draw_graph(
...
+                    plot_function(x_values, masked_values)
+                else:
+                    x_values = get_x_values(len(value))
+                    plot_function(x_values, value)
+        if use_calendar_dates:
+            ax.xaxis.set_major_formatter(mdates.DateFormatter("%j/%Y"))
+            plt.xlabel("Calendar Day")
+            plt.xticks(rotation=45)
+        else:
+            plt.xlabel("Simulation Day")
```
ew3361zh, line 616: Initially thought about adding the units to the y-axis automatically but remembered there are times where the variables graphed can have different units. This is why we decided to add the units to the variables in the legend instead (ages ago).

RUFAS/graph_generator.py
Isgotkowitz: Huh interesting.

RUFAS/graph_generator.py
```diff
@@ -553,18 +561,34 @@ def _draw_graph(
...
+                if mask_values:
                     indices, masked_values = self._mask_values(value)
-                    plot_function(indices, masked_values)
+                    x_values = get_x_values(len(indices))
+                    plot_function(x_values, masked_values)
+                else:
+                    x_values = get_x_values(len(value))
+                    plot_function(x_values, value)
+        if use_calendar_dates:
+            ax.xaxis.set_major_formatter(mdates.DateFormatter("%j/%Y"))
```
Isgotkowitz: Instead of using the day of year (1 - 365/366), I think it would be better to use a format that specifies the month and day of the month. I'm not sure why DOY was used initially for so many things in RuFaS, but I think most users would prefer a more standard date format instead.

RUFAS/graph_generator.py
```diff
@@ -553,18 +561,34 @@ def _draw_graph(
...
         if graph_type not in MATPLOTLIB_PLOT_FUNCTIONS:
             raise ValueError(f"Unsupported graph type: {graph_type}")
+        dates_in_data_range = [
+            self.time.start_date + datetime.timedelta(days=i) for i in range(max(len(v) for v in data.values()))
+        ]
         plot_function = MATPLOTLIB_PLOT_FUNCTIONS[graph_type]
+
+        def get_x_values(values_length: int) -> list[int]:
+            """Get appropriate x-axis values based on user choice."""
+            return dates_in_data_range[:values_length] if use_calendar_dates else list(range(values_length))
```
Isgotkowitz: I like the deduplication of code, but I don't think functions within functions are kosher from a clean code standpoint. Maybe it would work as a lambda function?

RUFAS/graph_generator.py
```diff
@@ -525,10 +528,12 @@ def _log_non_numerical_data(
 
     def _draw_graph(
         self,
-        graph_type: str | list[str],
-        data: Dict[str, List[int | float]],
-        selected_variables: Optional[List[str]] = None,
+        graph_type: str,
+        data: dict[str, list[int | float]],
+        selected_variables: list[str],
+        ax: Axes,
```
Isgotkowitz, line 544:
```suggestion
        ax: Axes | None,
```
Does this need to be None as well?

Isgotkowitz (commented):
This is an excellent start to having dates on the x-axis Niko! I left some comments on more minor things.

There are some edge cases that should be accounted for though, either by warning the users or adjusting the dates displayed.

The first is if the data is sliced in some way, the first date and last dates on the axis do not necessarily match up with the start and end dates of the simulation. To exemplify this, try:
```json
{
    "type": "plot",
    "filters": [
      ".*ammonia_emissions.*"
    ],
    "title": "Ammonia Emissions",
    "slice_start": -365,
    "slice_end": -1,
    "use_calendar_dates": true
}
```
I think this edge case can be addressed in the code.

Another edge case that I think can't be addressed in the code is if data is non-consecutive i.e. there are one or more days in between values. If this can't be addressed in the code, I think it would be good to at least warn users that this is one of the limitations of the graph generator.

RUFAS/graph_generator.py
ew3361zh: My original thought was to create this function outside the function but it was so specific that it seemed like nesting it was simpler and cleaner. I did read up on nested functions and couldn't find anything explicitly saying they were a bad practice or weren't within the guidelines of clean code.

Do you have any references on the matter?

I personally think lambda functions are less readable/understandable than regular functions but it is just my preference.

RUFAS/graph_generator.py
ew3361zh: If this is a reference to the docstring param notes, I removed the `default None` from the Axes section. Or were you seeing a warning in VS Code/MyPy somewhere about it? I think because of the way it's set up in the parent function, it wouldn't ever be None.

ew3361zh:
> Another edge case that I think can't be addressed in the code is if data is non-consecutive i.e. there are one or more days in between values. If this can't be addressed in the code, I think it would be good to at least warn users that this is one of the limitations of the graph generator.

@Isgotkowitz Do you have an example filter for this by any chance?

RUFAS/graph_generator.py
ew3361zh: Curious what @JoeWaddell's opinion on this is as well. I'm not sure I have strong feelings one way or another.

RUFAS/graph_generator.py
Isgotkowitz: > Do you have any references on the matter?

No, it just throws me for a loop whenever I see `def` nested inside a `def`. But I think you make a good case for using a nested function

Isgotkowitz:
> > Another edge case that I think can't be addressed in the code is if data is non-consecutive i.e. there are one or more days in between values. If this can't be addressed in the code, I think it would be good to at least warn users that this is one of the limitations of the graph generator.
> 
> @Isgotkowitz Do you have an example filter for this by any chance?

An example would be the ration formulation, because that data is only collected once per ration interval. I'll work on putting together a filter for it.

Edit: so the ration data that I could filter out is padded, as far as I could tell. AFAIK there isn't any data like this currently coming out of RuFaS, but it is important that we keep this in mind as something that would complicate setting the dates on the x-axis correctly

RUFAS/graph_generator.py
Isgotkowitz: Gotcha, it was a reference to the docstring. I didn't see any warnings about it

ew3361zh:
Updates/responses to feedback:
1. Handles and properly assigns dates on x-axis to sliced data.
2. Adds function to Time class to convert a slice start or slice end to its corresponding simulation day.
3. Adds date formatting option to graph filter:
- `day_of_year` which would look like 234/2022
- `day_month_year` which would look like 23/12/2024
- default is `day_month_year` as requested (sort of) by @Isgotkowitz and @JoeWaddell

To test, here's a twofer example filter:
```
{
  "type": "plot",
  "filters": [
    ".*daily.*ammonia_emissions.*"
  ],
  "title": "Ammonia Emissions",
  "slice_start": -365,
  "slice_end": -1,
  "use_calendar_dates": true,
  "date_format": "day_month_year"
}
```

RUFAS/graph_generator.py
```diff
@@ -554,17 +581,68 @@ def _draw_graph(
...
+
+        Returns
+        -------
+        matplotlib.dates.DateFormatter
+            A `DateFormatter` instance for the specified format.
+
+        """
+        format_mapping = {
+            "day_of_year": "%j/%Y",
+            "day_month_year": "%d/%m/%Y",
```
Isgotkowitz, line 638: European-style, I like it.

What would you think of making this an enum and putting it in a higher level module? My thought is that we may want to expand the range of options for date formats, and there may be other, higher-level modules which the user will want to configure to use a specific date format.

Also, for the graph generator, what about making it possible for users to select one of the pre-configured options (like the ones you have here) or inputting their own custom date format (i.e. a string with letters and percent symbols wherever they want)? This could be a good opportunity to dry-run an input like this, before we start trying to do something like this with dates in the inputs.

Isgotkowitz (commented):
Nice! I think this looks awesome Niko. I left some suggestions about how the date feature could be improved, but those can wait for future expansion if you think they'd be too much to include in this PR.

Not approving right now because of two things: it looks like Matthew's PR got mixed up in here, and also the filter in your comment isn't working for me unfortunately. I get the error message "[23-Dec-2024_Mon_17-20-00.281182][ERROR][default] Error plotting 'Ammonia Emissions' data set. Unforeseen error max() arg is an empty sequence when trying to graph data." and no graph is printed

ew3361zh:
> Not approving right now because of two things: it looks like Matthew's PR got mixed up in here, 

Rebased and Matthew's stuff is gone.

> and also the filter in your comment isn't working for me unfortunately. I get the error message "[23-Dec-2024_Mon_17-20-00.281182][ERROR][default] Error plotting 'Ammonia Emissions' data set. Unforeseen error max() arg is an empty sequence when trying to graph data." and no graph is printed

I think the github markdown formatting replaced a `*` in the filter with italics style...should have done it as code, apologies! (editor's note: original version above is fixed as well)

Here is the correct version:

```
{
  "type": "plot",
  "filters": [
    ".*daily.*ammonia_emissions.*"
  ],
  "title": "Ammonia Emissions",
  "slice_start": -365,
  "slice_end": -1,
  "use_calendar_dates": true,
  "date_format": "day_month_year"
}
```

RUFAS/graph_generator.py
ew3361zh: > What would you think of making this an enum and putting it in a higher level module? My thought is that we may want to expand the range of options for date formats, and there may be other, higher-level modules which the user will want to configure to use a specific date format.

Yea, I can write this up as an issue after this is merged if that works for you.

> Also, for the graph generator, what about making it possible for users to select one of the pre-configured options (like the ones you have here) or inputting their own custom date format (i.e. a string with letters and percent symbols wherever they want)? This could be a good opportunity to dry-run an input like this, before we start trying to do something like this with dates in the inputs.

I feel like custom date format could be an option we put in the enum. I find date formatting pretty unintuitive and always have to look these up so I could see some people struggling with it but there are enough resources available online that I think it wouldn't be too difficult to create the format they want.

Isgotkowitz: Works for me!

Isgotkowitz (approved):
LGTM! Nice job Niko

RUFAS/graph_generator.py
JoeWaddell: We talked on Slack, but to summarize: I like both options! Some inputs assume the day of the year (crop especially), but that's also not as easily readable to a general audience as having the actual months. The later (in this PR) addition to allow for both is excellent, since each serves different needs!

JoeWaddell (changes requested):
Looks great! Just holding off because I can't get the `day_month_year` option to work: it's only using the `day_of_year`. Tried both default, and explicitly calling both separately. Looking through the code again...

UPDATE: false alarm, it works! Weird git issue on my end - had to switch branches, then pull again...it hadn't pulled in all changes (despite pulling it in from remote first thing this morning...)

JoeWaddell (approved):

PR #2144, opened 2024-12-18
[Animal] Refresh code for comparing a ration's nutrient supply against an animal's nutrient demand

### LINKED ISSUE #2130: [Animal] Refresh the calculation of animal nutritive requirements

Isgotkowitz:
Refreshes code for comparing nutrition supply against nutrition demand.

### Context

Issue(s) closed by this pull request: closes #2130

### What

Compares the nutrients supplied by a ration against those required by animal. A flag indicates if the nutrients adequately supplied the demand, as well as a report containing all surpluses and deficiencies of nutrients that were supplied.

### Why

In order to compare a ration against an animal's requirements, the nutrients in the ration have to be compared against the animal's requirements in a very specific way. The result of this comparison decides whether the animal's outcomes (and requirements) need to be lowered, or if the ration needs to be reformulated.

### How

The evaluation functions were taken from the constraint functions in `ration_driver.py`, and placed in a dedicated class called `NutritionEvaluator`. This class takes in a `NutritionRequirements`, a `NutritionSupply`, and a flag indicating whether the animal being evaluated is a cow or not. After comparing the supply and requirements, a flag is returned indicating whether the supply is sufficient to meet the requirements and a report of the differences between the supply and demand.

### Test plan

The scaffolding of a unit test suite is provided, but will be filled in at later point. If there are any bugs in this code, they will be caught when the refreshed animal module undergoes heavy scrutiny.

### DIFF
--- a/RUFAS/biophysical/animal/animal_module_constants.py
+++ b/RUFAS/biophysical/animal/animal_module_constants.py
@@ -62,7 +62,7 @@ class AnimalModuleConstants:
     """Milk lactose content, percentage."""
 
     DMI_CONSTRAINT_FRACTION: float = 0.20
-    """The +/- fraction of DMI estimated allowed for ration formulation"""
+    """The +/- fraction of DMI estimated allowed for ration formulation."""
 
     MINIMUM_DMI: float = 1.0
     """Minimum estimated DMI instituted for all animals, kg/day"""
@@ -85,13 +85,13 @@ class AnimalModuleConstants:
     """Minimum calcium estimate, g/day"""
 
     MINIMUM_NDF: float = 15.0
-    """Minimum percentage of neutral detergent fiber (NDF) that a ration's dry matter must be."""
+    """Minimum percentage of a ration's dry matter that must be neutral detergent fiber (NDF)."""
 
     MAXIMUM_NDF: float = 45.0
-    """Maximum percentage of neutral detergent fiber (NDF) that a ration's dry matter must be."""
+    """Maximum percentage of a ration's dry matter that must be neutral detergent fiber (NDF)."""
 
     MINIMUM_FAT: float = 7.0
-    """Minimum percentage of fat that a ration's dry matter must be."""
+    """Minimum percentage of a ration's dry matter that must be fat."""
 
     MILK_REDUCTION_KG: float = 0.25
     """Milk reduction amount for each failed ration optimization attempt, kg"""
--- a/RUFAS/biophysical/animal/animal_module_constants.py
+++ b/RUFAS/biophysical/animal/animal_module_constants.py
@@ -84,11 +84,11 @@ class AnimalModuleConstants:
     MINIMUM_CALCIUM: float = 0.0
     """Minimum calcium estimate, g/day"""
 
-    MINIMUM_NDF: float = 15.0
-    """Minimum percentage of a ration's dry matter that must be neutral detergent fiber (NDF)."""
+    MINIMUM_RATION_NDF: float = 15.0
+    """Minimum percentage of a pen's ration's dry matter that must be neutral detergent fiber (NDF)."""
 
-    MAXIMUM_NDF: float = 45.0
-    """Maximum percentage of a ration's dry matter that must be neutral detergent fiber (NDF)."""
+    MAXIMUM_RATION_NDF: float = 45.0
+    """Maximum percentage of a pen's ration's dry matter that must be neutral detergent fiber (NDF)."""
 
     MINIMUM_FAT: float = 7.0
     """Minimum percentage of a ration's dry matter that must be fat."""
--- a/RUFAS/biophysical/animal/animal_module_constants.py
+++ b/RUFAS/biophysical/animal/animal_module_constants.py
@@ -61,8 +61,8 @@ class AnimalModuleConstants:
     MILK_LACTOSE: float = 4.85
     """Milk lactose content, percentage."""
 
-    DMI_CONSTRAINT_PERCENT: float = 0.20
-    """The +/- percentage of DMI estimated allowed for ration formulation"""
+    DMI_CONSTRAINT_FRACTION: float = 0.20
+    """The +/- fraction of DMI estimated allowed for ration formulation"""
 
     MINIMUM_DMI: float = 1.0
     """Minimum estimated DMI instituted for all animals, kg/day"""
@@ -83,6 +83,15 @@ class AnimalModuleConstants:
 
     MINIMUM_CALCIUM: float = 0.0
     """Minimum calcium estimate, g/day"""
+
+    MINIMUM_NDF: float = 15.0
+    """Minimum percentage of neutral detergent fiber (NDF) that a ration's dry matter must be."""
+
+    MAXIMUM_NDF: float = 45.0
+    """Maximum percentage of neutral detergent fiber (NDF) that a ration's dry matter must be."""
+
+    MINIMUM_FAT: float = 7.0
+    """Minimum percentage of fat that a ration's dry matter must be."""
 
     MILK_REDUCTION_KG: float = 0.25
     """Milk reduction amount for each failed ration optimization attempt, kg"""
--- a/RUFAS/biophysical/animal/animal_module_constants.py
+++ b/RUFAS/biophysical/animal/animal_module_constants.py
@@ -94,7 +94,7 @@ class AnimalModuleConstants:
     """Minimum percentage of a pen's ration's dry matter that must be NDF from forages (percent)."""
 
     MINIMUM_RATION_FAT: float = 7.0
-    """Minimum percentage of a ration's dry matter that must be fat."""
+    """Minimum percentage of a ration's dry matter that must be fat (percent)."""
 
     MILK_REDUCTION_KG: float = 0.25
     """Milk reduction amount for each failed ration optimization attempt, kg"""
--- a/RUFAS/biophysical/animal/animal_module_constants.py
+++ b/RUFAS/biophysical/animal/animal_module_constants.py
@@ -84,11 +84,14 @@ class AnimalModuleConstants:
     MINIMUM_CALCIUM: float = 0.0
     """Minimum calcium estimate, g/day"""
 
-    MINIMUM_RATION_NDF: float = 15.0
-    """Minimum percentage of a pen's ration's dry matter that must be neutral detergent fiber (NDF)."""
+    MINIMUM_RATION_NDF: float = 25.0
+    """Minimum percentage of a pen's ration's dry matter that must be neutral detergent fiber (NDF) (percent)."""
 
     MAXIMUM_RATION_NDF: float = 45.0
-    """Maximum percentage of a pen's ration's dry matter that must be neutral detergent fiber (NDF)."""
+    """Maximum percentage of a pen's ration's dry matter that must be NDF (percent)."""
+
+    MINIMUM_RATION_FORAGE_NDF: float = 15.0
+    """Minimum percentage of a pen's ration's dry matter that must be NDF from forages (percent)."""
 
     MINIMUM_FAT: float = 7.0
     """Minimum percentage of a ration's dry matter that must be fat."""
--- a/RUFAS/biophysical/animal/animal_module_constants.py
+++ b/RUFAS/biophysical/animal/animal_module_constants.py
@@ -93,7 +93,7 @@ class AnimalModuleConstants:
     MINIMUM_RATION_FORAGE_NDF: float = 15.0
     """Minimum percentage of a pen's ration's dry matter that must be NDF from forages (percent)."""
 
-    MINIMUM_FAT: float = 7.0
+    MINIMUM_RATION_FAT: float = 7.0
     """Minimum percentage of a ration's dry matter that must be fat."""
 
     MILK_REDUCTION_KG: float = 0.25
--- a/RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
+++ b/RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
@@ -169,7 +169,11 @@ class NutritionEvaluationResults:
     def is_valid_heifer_ration(self) -> bool:
         """True if evaluated supply meets requirements for heifers, else false."""
         non_negative_fields = {
-            self.maintenance_energy, self.growth_energy, self.calcium, self.phosphorus, self.forage_ndf_percent
+            self.maintenance_energy,
+            self.growth_energy,
+            self.calcium,
+            self.phosphorus,
+            self.forage_ndf_percent,
         }
         valid_non_negative_fields = all([field >= 0.0 for field in non_negative_fields])
 
--- a/RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
+++ b/RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
@@ -25,7 +25,7 @@ class NutritionRequirements:
     phosphorus : float
         Phosphorus requirement calculated with either the NASEM or NRC methodologies (g).
     secondary_phosphorus : float
-        Phosphorus requirement calculated with the dedicated animal phosphorus submodule (g).
+        Phosphorus requirement calculated by the dedicated animal phosphorus submodule (g).
     dry_matter : float
         Dry matter intake requirement (kg).
     activity_energy : float
--- a/RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
+++ b/RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
@@ -23,7 +23,9 @@ class NutritionRequirements:
     calcium : float
         Calcium requirement (g).
     phosphorus : float
-        Phosphorus requirement (g).
+        Phosphorus requirement calculated with either the NASEM or NRC methodologies (g).
+    secondary_phosphorus : float
+        Phosphorus requirement calculated with the dedicated animal phosphorus submodule (g).
     dry_matter : float
         Dry matter intake requirement (kg).
     activity_energy : float
@@ -40,6 +42,7 @@ class NutritionRequirements:
     metabolizable_protein: float
     calcium: float
     phosphorus: float
+    secondary_phosphorus: float
     dry_matter: float
     activity_energy: float
     essential_amino_acids: EssentialAminoAcidRequirements
--- a/RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
+++ b/RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
@@ -135,9 +135,8 @@ class NutritionEvaluationResults:
     ndf_percent : float
         Surplus or deficit of neutral detergent fiber (NDF) percentage in a ration. If NDF percentage is within
         acceptable bounds, this value will be 0.0.
-    ndf_percent : float
-        Surplus or deficit of neutral detergent fiber (NDF) percentage supplied by forages in a ration. If forage NDF
-        percentage is within acceptable bounds, this value will be 0.0.
+    forage_ndf_percent : float
+        Surplus or deficit of neutral detergent fiber (NDF) percentage supplied by forages in a ration.
     fat_percent : float
         Surplus or deficit of fat percentage in a ration. If fat percentage is within acceptable bounds, this value will
         be 0.0.
@@ -169,7 +168,9 @@ class NutritionEvaluationResults:
     @property
     def is_valid_heifer_ration(self) -> bool:
         """True if evaluated supply meets requirements for heifers, else false."""
-        non_negative_fields = {self.maintenance_energy, self.growth_energy, self.calcium, self.phosphorus}
+        non_negative_fields = {
+            self.maintenance_energy, self.growth_energy, self.calcium, self.phosphorus, self.forage_ndf_percent
+        }
         valid_non_negative_fields = all([field >= 0.0 for field in non_negative_fields])
 
         return valid_non_negative_fields and self._are_clamped_values_acceptable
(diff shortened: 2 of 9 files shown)


RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
```diff
@@ -23,7 +23,9 @@ class NutritionRequirements:
     calcium : float
         Calcium requirement (g).
     phosphorus : float
-        Phosphorus requirement (g).
+        Phosphorus requirement calculated with either the NASEM or NRC methodologies (g).
+    secondary_phosphorus : float
+        Phosphorus requirement calculated with the dedicated animal phosphorus submodule (g).
```
Isgotkowitz: This dedicated phosphorus submodule is `biophysical/animal/nutrients/nutrients.py`. The phosphorus requirement from that module will be added into this data structure by `Animal`, and I will write up an issue to renamed the submodule and class.

tests/animal_module_tests/nutrient_tests/test_nutrition_evaluator.py
```diff
@@ -0,0 +1,61 @@
...
+    """Tests that calcium supplied in ration is correctly compared against an animal's calcium requirement."""
+    pass
+
+
+def test_check_phosphorus_supplied() -> None:
+    """Tests that phosphorus supplied in ration is correctly compared against an animal's phosphorus requirement."""
+    pass
+
+
+def test_check_protein_supplied() -> bool:
```
Isgotkowitz:
```suggestion
def test_check_protein_supplied() -> None:
```

RUFAS/biophysical/animal/animal_module_constants.py
```diff
@@ -61,8 +61,8 @@ class AnimalModuleConstants:
     MILK_LACTOSE: float = 4.85
     """Milk lactose content, percentage."""
 
-    DMI_CONSTRAINT_PERCENT: float = 0.20
-    """The +/- percentage of DMI estimated allowed for ration formulation"""
+    DMI_CONSTRAINT_FRACTION: float = 0.20
+    """The +/- fraction of DMI estimated allowed for ration formulation"""
```
ew3361zh:
```suggestion
    """The +/- fraction of DMI estimated allowed for ration formulation."""
```

RUFAS/biophysical/animal/animal_module_constants.py
```diff
@@ -84,6 +84,15 @@ class AnimalModuleConstants:
     MINIMUM_CALCIUM: float = 0.0
     """Minimum calcium estimate, g/day"""
 
+    MINIMUM_NDF: float = 15.0
+    """Minimum percentage of neutral detergent fiber (NDF) that a ration's dry matter must be."""
```
ew3361zh:
```suggestion
    """Minimum percentage of a ration's dry matter that must be neutral detergent fiber (NDF)."""
```

RUFAS/biophysical/animal/animal_module_constants.py
```diff
@@ -84,6 +84,15 @@ class AnimalModuleConstants:
...
+
+    MAXIMUM_NDF: float = 45.0
+    """Maximum percentage of neutral detergent fiber (NDF) that a ration's dry matter must be."""
```
ew3361zh:
```suggestion
    """Maximum percentage of a ration's dry matter that must be neutral detergent fiber (NDF)."""
```

RUFAS/biophysical/animal/animal_module_constants.py
```diff
@@ -84,6 +84,15 @@ class AnimalModuleConstants:
...
+
+    MINIMUM_FAT: float = 7.0
+    """Minimum percentage of fat that a ration's dry matter must be."""
```
ew3361zh:
```suggestion
    """Minimum percentage of a ration's dry matter that must be fat."""
```

RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
ew3361zh:
```suggestion
        Phosphorus requirement calculated by the dedicated animal phosphorus submodule (g).
```

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+            "phosphorus": cls._check_phosphorus_supplied,
+            "protein": cls._check_protein_supplied,
+            "ndf": cls._check_neutral_detergent_fiber_supplied,
+            "fat": cls._check_fat_content,
+            "dry_matter": cls._check_dry_matter_intake,
+        }
+        cow_energy_nutrition_checkers = heifer_energy_nutrition_checkers + {
+            "total_energy": cls._check_total_energy_supplied,
+            "lactation": cls._check_lactation_energy_supplied,
+        }
```
ew3361zh: Dictionaries can't be added together, I think you have to use the ** operator.

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+
+        checkers = cow_energy_nutrition_checkers if is_cow else heifer_energy_nutrition_checkers
+        results = {name: method(requirements, supply) for name, method in checkers.items()}
+
+        evaluation = NutritionEvaluationResults(
+            total_energy=results.get("total_energy"),
+            maintenance_energy=results["maintentance"],
```
ew3361zh:
```suggestion
            maintenance_energy=results["maintenance"],
```
Which brings up another point - should these all just be results.get()s?

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+
+        total_energy_requirement = sum(
+            [
+                requirements.lactation_energy,
+                requirements.growth_energy,
+                requirements.maintenance_energy,
+                requirements.activity_energy,
+                requirements.pregnancy_energy,
+            ]
+        )
```
ew3361zh: Are all of these different energy types guaranteed to use the same unit?

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+            ndf_percent=results["ndf"],
+            fat_percent=results["fat"],
+            dry_matter=results["dry_matter"],
+        )
+
+        is_valid_ration = evaluation.is_valid_cow_ration if is_cow else evaluation.is_valid_heifer_ration
+        return is_valid_ration, evaluation
+
+    @classmethod
+    def _check_total_energy_supplied(requirements: NutritionRequirements, supply: NutritionSupply) -> float:
```
ew3361zh:
```suggestion
    def _calculate_total_energy_supplied(requirements: NutritionRequirements, supply: NutritionSupply) -> float:
```
Seems more like a calculation than a check (and matches the docstring definition better which I think is an accurate description).

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+                requirements.maintenance_energy,
+                requirements.activity_energy,
+                requirements.pregnancy_energy,
+            ]
+        )
+
+        return energy_supplied - total_energy_requirement
+
+    @classmethod
+    def _check_activity_maintenance_energy_supplied(
```
ew3361zh:
```suggestion
    def _calculate_activity_maintenance_energy_supplied(
```

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+        float
+            Difference between the maintenance energy supplied and the maintenance energy required (Mcal).
+
+        """
+        energy_requirement = requirements.activity_energy + requirements.maintenance_energy
+
+        return supply.maintenance_energy - energy_requirement
+
+    @classmethod
+    def _check_lactation_energy_supplied(requirements: NutritionRequirements, supply: NutritionSupply) -> float:
```
ew3361zh:
```suggestion
    def _calculate_lactation_energy_supplied(requirements: NutritionRequirements, supply: NutritionSupply) -> float:
```

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+        float
+            Difference between the lactation energy supplied and the lactation energy required (Mcal).
+
+        """
+        energy_requirement = requirements.lactation_energy + requirements.pregnancy_energy
+
+        return supply.lactation_energy - energy_requirement
+
+    @classmethod
+    def _check_growth_energy_supplied(requirements: NutritionRequirements, supply: NutritionSupply) -> float:
```
ew3361zh:
```suggestion
    def _calculate_growth_energy_supplied(requirements: NutritionRequirements, supply: NutritionSupply) -> float:
```
et al

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+        float
+            Difference between the phosphorus supplied and the phosphorus required (g).
+
+        """
+        requirement = max(requirements.phosphorus, requirements.secondary_phosphorus)
+
+        return supply.phosphorus - requirement
+
+    @classmethod
+    def _check_protein_supplied(requirements: NutritionRequirements, supply: NutritionSupply) -> bool:
```
ew3361zh:
```suggestion
    def _calculate_protein_supplied(requirements: NutritionRequirements, supply: NutritionSupply) -> float:
```

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+        """
+        fat_percentage = supply.fat_supply / supply.dry_matter * GeneralConstants.FRACTION_TO_PERCENTAGE
+
+        return fat_percentage - AnimalModuleConstants.MINIMUM_FAT
+
+    @classmethod
+    def _check_dry_matter_intake(requirements: NutritionRequirements, supply: NutritionSupply) -> float:
+        """Checks that the dry matter supplied meets the requirement."""
+        """
+        Calculates amount by which supplied dry matter under- or overshoots the required amount dry matter.
```
ew3361zh: Multiple docstrings/function definitions

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+        ----------
+        _ : NutritionRequirements
+            This argument is provided to keep the method signature uniform with other helper methods.
+        supply : NutritionSupply
+            Energy and nutrition supplied by a ration.
+
+        Returns
+        -------
+        float
+            Percentage by which supplied NDF under- or overshoots the required NDF range.
```
ew3361zh: My preference here would be to include the unit for uniformity's sake and reword if the redundancy is annoying.

ew3361zh (commented):
Nice work Ed, the modularity of this code compared to the old code is such a great step forward. I left some docstring/function name feedback and there's an issue in the main function with combining the dictionaries (95% sure).

Not sure if additional error handling is necessary and what here should be added to OM when it's up and running but it seems like a fair amount. Thoughts? Maybe that's better done at a later point?

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
Isgotkowitz: Good catch! Would you have any opposition to me using the pipe (`|`) operator instead though?

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
Isgotkowitz: Not necessarily. I deliberately used `.get` only for the attributes which are allowed to be `None` (total energy and lactation). I think this better reflects that these two attributes are different from the others.

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
Isgotkowitz: They _should_ use the same unit, I think in this context it is ok to assume that they do.

Also I realized I wrote an `@property` method for total_energy_requirement, so I will replace this code with that instead

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
Isgotkowitz: *Homer Simpson voice*: Doh!

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
Isgotkowitz: Even with the units added in I think it reads alright

Isgotkowitz (commented):
Nice catches Niko! Addressing comments now

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
ew3361zh: totally fine by me.

RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
```diff
@@ -23,7 +23,9 @@ class NutritionRequirements:
     calcium : float
         Calcium requirement (g).
     phosphorus : float
-        Phosphorus requirement (g).
+        Phosphorus requirement calculated with either the NASEM or NRC methodologies (g).
+    secondary_phosphorus : float
```
JoeWaddell: Reminder to carry over changes made in: https://github.com/RuminantFarmSystems/MASM/pull/2146

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
```diff
@@ -0,0 +1,306 @@
...
+
+        evaluation = NutritionEvaluationResults(
+            total_energy=results.get("total_energy"),
+            maintenance_energy=results["maintenance"],
+            lactation_energy=results.get("lactation"),
+            growth_energy=results["growth"],
+            calcium=results["calcium"],
+            phosphorus=results["phosphorus"],
+            metabolizable_protein=results["protein"],
+            ndf_percent=results["ndf"],
```
JoeWaddell: There needs to be a check for forage-specific NDF, see `forage_NDF_constraint` in `routines/animal/ration/ration_optimizer.py`

ew3361zh (approved):
Looks great, Ed. Tiny request to maybe update the test names to match the functions for now (even though they're not implemented)?

Isgotkowitz:
> Looks great, Ed. Tiny request to maybe update the test names to match the functions for now (even though they're not implemented)?

Yes, good call. I totally forgot about thsoe

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
```diff
@@ -0,0 +1,306 @@
...
+
+        """
+        heifer_energy_nutrition_checkers = {
+            "maintenance": cls._calculate_activity_maintenance_energy_supplied,
+            "growth": cls._calculate_growth_energy_supplied,
+            "calcium": cls._calculate_calcium_supplied,
+            "phosphorus": cls._calculate_phosphorus_supplied,
+            "protein": cls._calculate_protein_supplied,
+            "ndf": cls._calculate_neutral_detergent_fiber_supplied,
+            "fat": cls._calculate_fat_content,
```
JoeWaddell: For this and the dry matter check, should the `_supplied` suffix be used for consistency?

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
```diff
@@ -0,0 +1,306 @@
...
+
+        """
+        fat_percentage = supply.fat_supply / supply.dry_matter * GeneralConstants.FRACTION_TO_PERCENTAGE
+
+        return fat_percentage - AnimalModuleConstants.MINIMUM_FAT
+
+    @classmethod
+    def _calculate_dry_matter_intake(requirements: NutritionRequirements, supply: NutritionSupply) -> float:
+        """
+        Calculates amount by which supplied dry matter under- or overshoots the required amount dry matter.
```
JoeWaddell, line 312: Should be fat, not phosphorus

RUFAS/biophysical/animal/animal_module_constants.py
```diff
@@ -84,6 +84,15 @@ class AnimalModuleConstants:
     MINIMUM_CALCIUM: float = 0.0
     """Minimum calcium estimate, g/day"""
 
+    MINIMUM_NDF: float = 15.0
```
JoeWaddell: I think here it should be specified this is a minimum for the ration (pen level), and not requirement (animal level). I don't think we currently have overlap, but I could see a case in which a user could get confused thinking they're changing one and actually changing the other. Example would be if in the future we add a minimum protein requirement return, so individual animals have a requirement of at least X g protein, but the ration still could have a higher requirement (pen average/median/etc)

In short: adding a `_RATION` suffix, or otherwise indicating it's ration-specific would be a nice addition for clarity

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
```diff
@@ -0,0 +1,306 @@
...
+    NutritionRequirements,
+    NutritionSupply,
+    NutritionEvaluationResults,
+)
+from RUFAS.general_constants import GeneralConstants
+
+
+class NutritionEvaluator:
+    """Checks if energy and nutrients supplied in a ration satisfy the demand for an individual animal."""
+
```
JoeWaddell: This docstring (and elsewhere as applicable) could be generalized to avoid specifying this method to only apply to an individual animal, since the way most farms feed wouldn't be on an individual animal basis but a pen average (the way we currently do it in the code is to check the ration against the average/mean/etc requirements for a pen).

JoeWaddell (commented):
I really like how this is looking: going to be much easier for new users to read and interpret.

A few comments, but nothing major. Only outstanding concern is if the optimization methods can handle these returns (particularly that they're checking both mins and maxes simultaneously), but that's easy enough to handle when the optimizer is being addressed.

RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
Isgotkowitz: Thanks:)

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
Isgotkowitz: Yes, good idea!

RUFAS/biophysical/animal/animal_module_constants.py
Isgotkowitz: Is this the same for calcium and phosphorus too?

Isgotkowitz: And follow up question: if these are specific to the pen level and not the individual animal level, does that mean that a ration should be considered valid for an individual animal even if the NDF is outside the acceptable range for the pen level?

RUFAS/biophysical/animal/nutrients/nutrition_evaluator.py
Isgotkowitz: Added! Good catch Joe, that was a big miss.

Isgotkowitz:
Significant updates:
- Added calculation of the NDF supplied by forage in a ration.
- Added check for supply of forage NDF in the ration against minimum acceptable amount.
- Still need to verify how the check of nutrients differs when checking for an individual animal vs. checking for a pen average, and if further changes are needed to the code to accommodate any differences.
  - Update: there are differences between these, however the code does not have to be changed to accommodate these differences.

RUFAS/biophysical/animal/animal_module_constants.py
JoeWaddell: We met via zoom, but to add some context here for future reference: the NDF check would apply to the individual or pen level (e.g. the former in cases of rations being formulated on an an individual basis).

for calcium and phorphorus, those are specific to individual animal requirement returns, not ration-level optimized/formulation checks.

Isgotkowitz:
I'm not sure why there are 4 more errors in this branch, I checked them locally and the errors only appear to be in code that was added in other PRs.

RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
Isgotkowitz: Update: this other P requirement has been renamed to `process_based_phosphorus`.

JoeWaddell (approved):
Nice work handling the comments/changes, good progress here!

RUFAS/biophysical/animal/animal_module_constants.py
```diff
@@ -84,6 +84,18 @@ class AnimalModuleConstants:
...
+    """Minimum percentage of a pen's ration's dry matter that must be neutral detergent fiber (NDF) (percent)."""
+
+    MAXIMUM_RATION_NDF: float = 45.0
+    """Maximum percentage of a pen's ration's dry matter that must be NDF (percent)."""
+
+    MINIMUM_RATION_FORAGE_NDF: float = 15.0
+    """Minimum percentage of a pen's ration's dry matter that must be NDF from forages (percent)."""
+
+    MINIMUM_RATION_FAT: float = 7.0
+    """Minimum percentage of a ration's dry matter that must be fat."""
```
Isgotkowitz:
```suggestion
    """Minimum percentage of a ration's dry matter that must be fat (percent)."""
```

Issue #1782, opened 2024-06-26
Connect the `Animal` and `FeedStorage` modules


Isgotkowitz:
**Is your feature request related to a problem? Please describe.**
Currently, the new Feed Storage module and the Animal module are not connected, and the Animal module is still dependent on the old Feed module for feed-related information.

The connection between the old Feed module and the Animal module includes the `SimulationEngine`'s `Feed` instance being passed into the `AnimalManager`'s initialization method, and into `AnimalManager`'s `daily_updates` routine in `_daily_simulation`.

**Describe the solution you'd like**
The Animal module should be disconnected from the old Feed module, and be connected to the new Feed Storage module. 

The connection between the Animal and Feed Storage modules is very dependent on the interface of the Ration Formulator in the Animal module. Because the Animal module is currently undergoing a major refactor, this issue should wait until the interface is more stable before plans are laid out for the connection between the Animal and Feed Storage modules.

The connection between `Animal` and `FeedStorage` is outlined in this whimsical diagram: https://whimsical.com/ration-formulation-WZ4ij1WfGy2i9o4UTPNwPY

**Implementation Strategy**
Because of the complexity of the Animal and FeedStorage connection, the implementation of the connection will be broken into multiple tasks/PRs. The tasks listed here roughly correspond to a PR and are listed in chronological order.

- [x] Calculate animals requirements and get User-defined ration setup in refreshed Animal module.
- [x] Add Feed ID to `HarvestedCrop` and implement farm-grown feed addition and subtraction logic in `FeedStorage` module.
- [x] Put Feed IDs in Crop & Soil module, find dates of next crop harvests, and pass this info to `FeedStorage`.
- [x] Remove post-processing calculation of actual purchased feed in EEE module.

**Target Deadline**
By when should this issue be done.

Isgotkowitz:
Blocked by #1680

allisterakun:
Unblocking this issue: we will use some dummy logic for now and will update the logic after finalizing the ration formulation design.

Isgotkowitz:
Changing the t-shirt size to 1, and this issue will be closed when dummy logic is populated for the connection.

ew3361zh:
Only have the final piece of this left and similar to #2375, blocking this until we have a trusted FeedManager report that shows balance between feed requested/purchased/removed from storage.

PR #2146, opened 2024-12-19
[Animal] Implement refreshed user-defined ration code

Isgotkowitz:
Implement user-defined rations in the animal module.

### Context

Issue(s) closed by this pull request: addresses #1782 

### What

Changes the animal module to use the refreshed animal requirements code, which is used in the ration formulation process.

### Why

With the completion of #2130, the pieces are in place to formulate and evaluate rations for animals. Which we need in order to get full simulations running with the refreshed animal module.

### How

Changes are mainly in `Animal`, `Pen`, and `HerdManager` classes. 

`HerdManager` sets up the user-defined ration, and decides whether or not a `Pen` needs a ration reformulated.

`Pen` handles the ration formulation and evaluation, and updates animals with the new ration.

`Animal` has had the requirements method filled out, so that the requirements can be calculated as needed.

Odds and ends have been changed to make the whole thing work, please see in-code comments for specific explanations.

### Test plan

Unit tests have been updated where present. But this code will be tested by scrutinizing the results of running full simulations with different user-defined rations.


RUFAS/biophysical/animal/pen.py
```diff
@@ -304,23 +317,74 @@ def _set_animal_nutrient_values(
...
+                )
+
+                if self.average_milk_production < AnimalModuleConstants.MINIMUM_AVG_PEN_MILK:
+                    is_milk_production_sufficient = True
+                    break
+
+            animal.nutrition_supply = nutrition_supply
+            total_nutrient_evaluation_results += evaluation_result
+
+        self.average_nutrient_evaluation = total_nutrient_evaluation_results / len(self.animals_in_pen.values())
```
Isgotkowitz: @JoeWaddell sneak preview of what I would like to talk with you about. This is the heart of the user-defined ration code

RUFAS/enums.py
```diff
@@ -15,13 +15,10 @@ class AnimalCombination(Enum):
         Represents a pen with heiferIII and dry cows only.
     LAC_COW: int
         Represents a pen with lactating cows only.
-    GROWING_AND_CLOSE_UP: int
-        Represents a pen with all heifers and dry cows only.
```
Isgotkowitz, line 19: This is removed because there are rations defined for growing cows and close-up cows, but not both.

RUFAS/biophysical/animal/animal.py
```diff
@@ -561,4 +612,88 @@ def calculate_daily_walking_distance(self, vertical_dist_to_parlor: float, horiz
...
+            List of feeds available for ration formulation. Only needed for calf nutrition calculation.
+
+        Returns
+        -------
+        NutritionRequirements
+            The nutrition requirements for the animal.
+
+        """
+        if self.animal_type is AnimalType.CALF:
+            pass  # TODO: calves requirements are implemented in PR #2153, which is not merged yet.
```
Isgotkowitz:
```suggestion
            pass  # TODO: calves' requirements are implemented in PR #2153, which is not merged yet.
```

RUFAS/biophysical/animal/animal.py
```diff
@@ -163,6 +199,7 @@ def __init__(
         self.days_born = int(args.get("days_born"))
         self.birth_weight = float(args.get("birth_weight"))
         self.net_merit = args.get("net_merit", 0.0)
+        self.body_condition_score_5 = AnimalModuleConstants.DEFAULT_BODY_CONDITION_SCORE_5
```
Isgotkowitz, line 207: In the old implementation, the body condition score was set to 3 and never changed, and this is what this code mimics.

RUFAS/biophysical/animal/animal.py
```diff
@@ -171,19 +208,29 @@ def __init__(
...
         # remove sold and dead to use sold_at_day and dead_at_day instead
-        self.dry_matter_intake: float = 0.0
-
         self.events = AnimalEvents()
 
-        self.nutrient: dict[str, float] = {}
-        self.nutrient_concentrations: dict[str, float] = {}
-        self.nutrient_requirements: dict[str, float] = {}
+        self.nutrition_supply: NutritionSupply = NutritionSupply.make_empty_nutrition_supply()
+        self.previous_nutrition_supply: NutritionSupply | None = None
```
Isgotkowitz, line 219: The old DMI and nutrient attributes were holdovers, the `NutritionSupply` data structure will serve as the only data structure that contains nutritive information supplied to animals -- unless a reviewers knows a case in RuFaS which necessitates a different data structure.

RUFAS/biophysical/animal/animal.py
```diff
@@ -555,10 +622,109 @@ def calculate_daily_walking_distance(self, vertical_dist_to_parlor: float, horiz
...
+        if self.animal_type is AnimalType.CALF:
+            pass  # TODO: calves' requirements are implemented in PR #2153, which is not merged yet.
+
+        days_in_pregancy = self.days_in_pregnancy if self.is_pregnant else None
+        days_in_milk = self.days_in_milk if self.is_milking else None
+
+        if self.previous_nutrition_supply is None:
+            ndf_percentage = AnimalModuleConstants.DEFAULT_NDF_PERCENTAGE
+            tdn_percentage = AnimalModuleConstants.DEFAULT_TDN_PERCENTAGE
+            net_energy_diet_conc = AnimalModuleConstants.DEFAULT_NET_ENERGY_DIET_CONCENTRATION
```
Isgotkowitz, line 804: These values were pulled from the previous implementation - see `set_nutrient_rqmts` in `heiferI.py`, `heiferII.py` and `heiferIII.py`,

RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
```diff
@@ -93,6 +142,83 @@ class NutritionSupply:
...
+    adf_content: float
+    tdn_content: float
+    lignin_content: float
+    ash_content: float
+    potassium_content: float
+    nitrogen_content: float = field(init=False)
+
+    def __post_init__(self) -> None:
+        """Sets the nitrogen content of a ration based on the crude protein content."""
+        self.nitrogen_content = self.crude_protein * GeneralConstants.PROTEIN_TO_NITROGEN
```
Isgotkowitz: SMEs I have no idea if this is the right way to go forward here, seeking your input

RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
Isgotkowitz, line 1: For anyone unfamiliar with python dunder methods: they are a special set of methods that have double underscores before and after the name (like `__init__` and `__add__`). They allow the use of Python operators (like `()` to initialize new objects, or `+` to add two things together).

The `__add__` method allows two objects to be added together (`new_obj = obj1 + obj2`), and `__truediv__` allows an object to be divided (`halved_obj = ob1 / 2`). These methods are implemented on several of the data structures in this file because it makes it easier to calculate averages.

RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
```diff
@@ -80,6 +115,20 @@ class NutritionSupply:
...
+    tdn_content : float
+        Amount of Total Digestible Nutrients (TDN) in a ration (kg).
+    lignin_content : float
+        Amount of lignin in a ration (kg).
+    ash_content : float
+        Amount of ash in a ration (kg).
+    potassium_content : float
+        Amount of potassium in a ration (kg).
+    nitrogen_content : float
+        Amount of nitrogen in a ration (kg). This value is derived from the crude protein content.
```
Isgotkowitz: These values were added because it seems that they were recorded for rations previously, and it seemed to me to be better to put all nutritive values associated with a ration in the same place, instead of defining a new data structure.

RUFAS/biophysical/animal/nutrients/nutrition_supply_calculator.py
```diff
@@ -615,37 +623,23 @@ def _calculate_rumen_undegradable_protein_percentages(
         return rup_percentages
 
     @classmethod
-    def _calculate_neutral_detergent_fiber_content(cls, feeds: list[FeedInRation]) -> float:
+    def _calculate_nutritive_content(cls, feeds: list[FeedInRation], nutrient: str) -> float:
```
Isgotkowitz, line 626: This method was changed to make the code less duplicative.

RUFAS/biophysical/animal/nutrients/nutrition_supply_calculator.py
```diff
@@ -615,37 +623,23 @@ def _calculate_rumen_undegradable_protein_percentages(
...
-
-        Returns
-        -------
-        float
-            Total supply of fat in a ration (kg).
-
-        """
-        return sum([feed.amount * feed.info.EE * GeneralConstants.PERCENTAGE_TO_FRACTION for feed in feeds])
+        return sum(
+            [feed.amount * getattr(feed.info, nutrient) * GeneralConstants.PERCENTAGE_TO_FRACTION for feed in feeds]
```
Isgotkowitz, line 644: I don't like using `getattr`, I think it makes the code more obscure. However I couldn't think of a different way to implement this method in a clean, extendable way without using it.

RUFAS/biophysical/animal/pen.py
```diff
@@ -45,32 +55,9 @@ def __init__(
...
-            "ADF": 0.0,
-            "NDF": 0.0,
-            "lignin": 0.0,
-            "ash": 0.0,
-            "phosphorus": 0.0,
-            "potassium": 0.0,
-            "N": 0.0,
-        }
-        self.dry_matter_intake = 0.0
-        self.MEdiet = 0.0
```
Isgotkowitz, line 73: `NutritionSupply` has been expanded so that it contains all of these values. MEdiet is also captured in `NutritionSupply`, it is called `metabolizable` though.

RUFAS/biophysical/animal/animal.py
```diff
@@ -538,14 +588,20 @@ def update_pen_history(self, current_pen: int, current_day: int, animal_types_in
...
+                )
             )
         else:  # last_pen == curr_pen
             self.pen_history[-1]["end_date"] = current_day
             self.pen_history[-1]["animal_types_in_pen"] = list(animal_types_in_pen)
 
-    def calculate_daily_walking_distance(self, vertical_dist_to_parlor: float, horizontal_dist_to_parlor: float) -> float:
+    def set_daily_walking_distance(
+        self, vertical_dist_to_parlor: float, horizontal_dist_to_parlor: float
+    ) -> None:
```
Isgotkowitz: The walking distance is needed to estimate the activity energy requirement for animals, so it is implemented in this PR.

RUFAS/biophysical/animal/animal_module_constants.py
```diff
@@ -25,6 +31,15 @@ class AnimalModuleConstants:
     DEFAULT_NUM_STALLS_FOR_GROWING_AND_CLOSE_UP_PEN: int = 500
     """The default number of stalls to be created in a combined growing and close-up pen."""
 
+    DEFAULT_NET_ENERGY_DIET_CONCENTRATION: float = 1.0
+    """Default metabolizable energy density of a ration when this value is not known for an animal."""
+
+    DEFAULT_NDF_PERCENTAGE: float = 0.3
+    """Percentage of neutral detergent fiber in a ration when this value is not known for an animal."""
+
+    DEFAULT_TDN_PERCENTAGE: float = 0.7
```
Isgotkowitz, line 40: SMEs, do these seem like they are fractions and not percentages to you?

RUFAS/biophysical/animal/nutrients/nrc_requirements_calculator.py
```diff
@@ -102,14 +103,15 @@ def calculate_requirements(
             valine=0.0,
         )
 
-        return NutritionRequirementsCalculator(
+        return NutritionRequirements(
```
Isgotkowitz: Can't believe this one slipped through the cracks.

RUFAS/biophysical/animal/nutrients/nasem_requirements_calculator.py
Isgotkowitz, line 1: The `Nutrients` module in `biophysical/nutrients/nutrients.py` is dedicated to phosphorus stuff, and calculates its own phosphorus requirement that independent of the P requirement calculated by NASEM and NRC. When evaluating whether a ration meets the required amounts of nutrients, `NutritionEvaluator` looks at both of these numbers and uses the maximum.

RUFAS/biophysical/animal/pen.py
```diff
@@ -304,24 +302,100 @@ def _set_animal_nutrient_values(
...
+        )
+        ration_sufficient_for_milk_production = True
+        total_nutrition_supply: NutritionSupply = NutritionSupply.make_empty_nutrition_supply()
+        for animal in self.animals_in_pen.values():
+            while ration_sufficient_for_milk_production:
+                nutrition_supply: NutritionSupply = NutritionSupplyCalculator.calculate_nutrient_supply(
+                    feeds_used=available_feeds, ration=ration, body_weight=animal.body_weight
+                )
+
+                walking_distance = animal.daily_distance if self.animal_combination == AnimalCombination.LAC_COW else 0.0
```
Isgotkowitz, line 361: Correct me if I'm wrong, but I think the walking distance is only relevant to lactating cows (and not dry ones) because lactating cows have to walk to the milking parlor. I bring this up because `daily_distance` is calculated for both dry and lactating cows, but it may be better to just calculate this for lactating cows.

RUFAS/biophysical/animal/pen.py
```diff
@@ -304,24 +302,100 @@ def _set_animal_nutrient_values(
...
+                    break
+
+            animal.previous_nutrition_supply = animal.nutrition_supply
+            animal.nutrition_supply = nutrition_supply
+            total_nutrition_supply += nutrition_supply
+            total_nutrient_evaluation_results += evaluation_result
+
+        self.average_nutrient_evaluation = total_nutrient_evaluation_results / len(self.animals_in_pen.values())
+        self.average_nutrition_supply = total_nutrition_supply / len(self.animals_in_pen.values())
+        self.ration = ration
```
Isgotkowitz, line 392: @jadamchick this is what I have roughly envisioned for how the ration formulation process will interact with the milk production values. This only pertains to user-defined rations, but something similar will happen for optimized rations. Also note, this is by no means the final form of user-defined ration formulation.

CC @allisterakun

RUFAS/biophysical/animal/data_types/nutrition_data_structures.py
```diff
@@ -23,7 +24,9 @@ class NutritionRequirements:
     calcium : float
         Calcium requirement (g).
     phosphorus : float
-        Phosphorus requirement (g).
+        Phosphorus requirement calculated with either the NASEM or NRC methodologies (g).
+    secondary_phosphorus : float
```
JoeWaddell: I think we should avoid calling this secondary: since that might mislead a reader into thinking it's a supplemental P requirement, and not an entirely separate calculation. Delineating between the two with a prefix or suffix I think would work: `process_based` may be enough to indicate this is our RuFaS "in-house" calculation, and not the NRC or NASEM calculation. We should consider if there's other places where we use a similar alternative methodology, if we want to settle on a naming system for those calcs (e.g. in the calf ration, intake/requirements come from in-house calculations, also largely not from NRC/NASEM lit).

PR #2149, opened 2024-12-20
[Crop and Soil] Relocate TillageImplements Enum from EEE to data_structures dir

### LINKED ISSUE #2077: [EEE][Crop & Soil] Disentangle EEE from other modules
The connection between the EEE module and other modules should be disentangled so that they meet the requirements outlined in https://github.com/RuminantFarmSystems/MASM/issues/1622.

## Possible Solution

When working on issue #1625, it was discovered that the `TillageImplement` class was in the EEE module. There may be other data classes in there as well that are shared between multiple modules. Extract `TillageImplement` and any other data classes found that are used by EEE and another module and put them in the `data_classes` directory.

ew3361zh:
Moves the `TillageImplements` enum out of EEE to data_structures dir.

### Context

Issue(s) closed by this pull request: closes #2077

### What

Pretty simple moving the `TillageImplements` enum to the data_structures dir and updating the references to the class across the codebase.

### Why

This structure was residing in the EEE module but was used exclusively in the C&S module. It will be used by EEE in the future so the best location for it to match what we're doing with other similarly-used data structures is the move it to the data_structures dir.

### How

Relocated the class and updated the references.

### Test plan

Not much beyond running the simulation and the test suite.


Isgotkowitz (approved):
LGTM

allisterakun (approved):
LGTM

Issue #1827, opened 2024-07-16
[Crop & Soil] Enable users to apply purchased manure to meet shortfall in manure requests from C&S module


ew3361zh:
Requirement 3 in the [C&S requirements document](https://docs.google.com/document/d/1g7QWqYbb5Ai9NFHYctHrQAOU_vFL_XaUgB6Dx9rqJSg/edit) states that C&S:

- Must enable users to utilize manure (tracking its nutrients (C, N, P, and water) from the manure module or from user input (purchased manure) to apply on fields and impact crop growth.

Currently we allow users to add manure from manure requests but if those requests fall short of the nutrients required, we don't have an option for them to opt to purchase manure to make up for the shortfall.

That ideally would be an input setting that would be checked. Look into implementing this and if it seems doable, pursue it.

matthew7838:
Grooming session: @ew3361zh will follow up with @KFosterReed.

KFosterReed:
I would rank this as a medium to high priority and hopefully much more feasible with the disentanglement of the biophysical modules that are happening. For implementation, we could create a regional database of manure compositions using this site: http://manuredb.umn.edu/

allisterakun:
This is the sister issue of #1828 .
@Isgotkowitz will come back and explain the current state of the manure and CS connection. 
The dev team member who takes on this task should follow up with the SMEs to confirm future plans.

Isgotkowitz:
Manure-C&S connection:
- Users give Crop and Soil a schedule of manure applications they want executed. Each application contains a requested mass of N and P
- Crop and Soil sends the requested masses of N and P to the Manure module
- If animals are simulated, then Manure will attempt to fulfill requested N and P amounts with manure that is available in storage
- If animals are not simulated, then Manure fulfills the attempted amounts of N and P with manure that is artificially created by the `FieldManureSupplier` in the Manure module

Ed-itorial: even when animals are simulated, there should be a way to supplement nutrient shortfalls with "purchased" manure i.e. manure created out of thin air by the `FieldManureSupplier`.

ew3361zh:
> I would rank this as a medium to high priority and hopefully much more feasible with the disentanglement of the biophysical modules that are happening. For implementation, we could create a regional database of manure compositions using this site: http://manuredb.umn.edu/

[...]

PR #2150, opened 2024-12-23
[Crop and Soil] User control over synthetic fertilizer

### LINKED ISSUE #1828: [Crop & Soil] Allow users to have more control over whether synthetic fertilizers are used
#### Problem
In Field._execute_manure_application(), if there is unmet nitrogen/phosphorus demand after the manure application, the demand is attempted to be met by 1 of 2 synthetic fertilizer mixes:

1. An ONLY_NITROGEN_MIX if the unmet_phosphorus_demand == 0.0
2. A more optimized mix that attempts to apply the least amount of fertilizer based on the demanded nutrients.

These synthetic fertilizers are applied whether or not the user wants them and regardless of cost.

#### What needs to happen:
There should be a way for the user to set whether they want synthetic fertilizer applied at all to meet nutrient demand (as opposed to applying purchased manure perhaps - issue #1827). Perhaps this could be regulated by a certain cost/amount per application. And we eventually would like to be tracking the cost of the applied synthetic fertilizer though that likely fits better within the EEE module.

matthew7838:
This PR will allow user control over one of the three methods (manure/none/synthetic) by making `supplement_manure_nutrient_deficiencies` a user input.

### Context

Issue(s) closed by this pull request: closes #1828

### What

1. `supplement_manure_nutrient_deficiencies` is now removed from field-level inputs and moved into manure schedule inputs and is now a list to correspond to each manure event.
2. `supplement_manure_nutrient_deficiencies` is now a string with three options, none, manure, and synthetic fertilizers.
3. Overview of the new data flow: FieldManager `set_up_field` will call set up manure events and grab the new inputs of supplement methods into each event. The created field will be checked on its manure schedule before daily execution as before, but the new inputs of supplement methods will be passed along with the nutrient requests and request results.
4. Unit tests updated.

### Why

SMEs need more control over nutrient deficiency supplement methods and also need to address the fact that each manure event could use a different method to supplement nutrient deficiencies.

### How

This will allow using manure to supplement nutrient deficiencies and more user control. These changes also addressed SME's suggestions that each manure event could have a different method to supplement nutrient deficiencies.

### Test plan

1. Run and pass with updated unit tests.
2. Run with default metadata and the results have no change.
3. Modify the default manure schedule data so that some `supplement_manure_nutrient_deficiencies` values use the `synthetic fertilizer` option and the result reflects the changes. The correctness of the changed results cannot be verified at this point since no available simulation uses `synthetic fertilizer` option.


matthew7838:
@ew3361zh Have you tagged to keep you in the loop of this finalized approach.

RUFAS/data_structures/events.py
```diff
@@ -229,6 +230,7 @@ def __init__(
         nitrogen_mass: float,
         phosphorus_mass: float,
         manure_type: ManureType,
+        manure_supplement_method: ManureSupplementMethod,
```
ew3361zh, line 235: add definition to docstrings

RUFAS/data_structures/manure_supplement_methods.py
```diff
@@ -0,0 +1,21 @@
...
+    SYNTHETIC_FERTILIZER : str
+        Represents the usage of synthetic fertilizer supplements.
+    MANURE : str
+        Represents the usage of manure supplements.
+
+    """
+
+    NONE = "none"
+    SYNTHETIC_FERTILIZER = "synthetic fertilizer"
+    MANURE = "manure"
```
ew3361zh, line 21: what do you think about a `BOTH` option or `FERTILIZER_AND_MANURE` option? I'm not sure the normal farm practice and whether that never happens but maybe we want to leave it open as an option in case?

I say this right now as I'm working on #1827 and the current solution won't allow meeting both nutrient shortfalls solely via supplemental manure. So if the user had a manure that reached 75% N but 50% P, they may want to first use supplemental manure to get to 100% N but at that N:P ratio, they wouldn't get to 100% N, 100% P. So having a `BOTH` option (or whatever you think is a good name) would allow them to supplement with synthetic fertilizer after they'd exhausted the limit of supplemental manure that would allow them to reach 100% of one of the nutrients.

RUFAS/routines/field/field/field.py
```diff
@@ -914,7 +917,9 @@ def check_manure_application_schedule(self, time: Time) -> list[ManureEventNutri
         manure_requests: list[ManureEventNutrientRequest] = []
         for event in todays_manure_events:
             manure_request = self._create_manure_request(event)
-            manure_requests.append(ManureEventNutrientRequest(self.field_data.name, event, manure_request))
+            manure_requests.append(
+                ManureEventNutrientRequest(self.field_data.name, event, event.manure_supplement_method, manure_request)
```
ew3361zh: If we have the `event` already in the request, do we need to separately add the `event.manure_supplement_method`?

RUFAS/routines/field/manager/manure_schedule.py
```diff
@@ -65,6 +66,7 @@ def __init__(
         nitrogen_masses: List[float],
         phosphorus_masses: List[float],
         manure_types: List[ManureType],
+        manure_supplement_methods: List[ManureSupplementMethod],
```
ew3361zh:
```suggestion
        manure_supplement_methods: list[ManureSupplementMethod],
```

ew3361zh: Add def to dosctrings

ew3361zh (commented):
Looking good and on the right track. Left some comments and would be curious to see what you think. I like your approach here too. I think it adds good flexibility for a more precise application control from the user.

RUFAS/data_structures/manure_supplement_methods.py
matthew7838: I think this is a good point, but with the current structure, it makes adding the extra option easy. I think it's fine to add it when the usage of both is confirmed by the SMEs. If any time the usage of both is confirmed I am also fine with adding it here.

RUFAS/routines/field/field/field.py
matthew7838: I missed that, great suggestion!

RUFAS/data_structures/manure_supplement_methods.py
ew3361zh: Sounds good. Maybe having an SME review on this PR would be good for this reason.

RUFAS/data_structures/events.py
```diff
@@ -221,6 +222,8 @@ class ManureEvent(BaseFieldManagementEvent):
         Depth that manure is injected into the soil at (mm).
     surface_remainder_fraction : float
         Fraction of manure applied that remains on the soil surface (unitless).
+    manure_supplement_method : ManureSupplementMethod
+        The method that the event will use to supplement nutrient deficiencies.
```
ew3361zh: Small request to put this in the same order that the parameters are ordered in the init so it's easier to read (so based on the order in the init, this should come after manure_type in the docstring definitions).

RUFAS/routines/field/field/field.py
```diff
@@ -523,6 +525,7 @@ def _execute_manure_application(
         year: int,
         day: int,
         manure_supplied: NutrientRequestResults | None,
+        manure_supplement_method: ManureSupplementMethod,
```
ew3361zh, line 528: Add to docstring definitions

ew3361zh (commented):
This looks good, Matthew, just holding off because I'd like to see an example of this new feature working. I understand the default metadata works the same as before but the default is to not use this new enhancement. Could you provide an example that shows that this new feature works as expected?

matthew7838:
> This looks good, Matthew, just holding off because I'd like to see an example of this new feature working. I understand the default metadata works the same as before but the default is to not use this new enhancement. Could you provide an example that shows that this new feature works as expected?

These are in the tests, the tests did cover the synthetic fertilizer option since it's the only other option implemented right now.

ew3361zh:
> > This looks good, Matthew, just holding off because I'd like to see an example of this new feature working. I understand the default metadata works the same as before but the default is to not use this new enhancement. Could you provide an example that shows that this new feature works as expected?
> 
> These are in the tests, the tests did cover the synthetic fertilizer option since it's the only other option implemented right now.

When I run the default simulation, is the synthetic fertilizer option implemented?

matthew7838:
> > > This looks good, Matthew, just holding off because I'd like to see an example of this new feature working. I understand the default metadata works the same as before but the default is to not use this new enhancement. Could you provide an example that shows that this new feature works as expected?
> > 
> > 
> > These are in the tests, the tests did cover the synthetic fertilizer option since it's the only other option implemented right now.
> 
> When I run the default simulation, is the synthetic fertilizer option implemented?

No, and I don't think any simulation that we have right now utilizes that part. What I can think of that is most relevant to checking the synthetic fertilizer part is only the tests.

ew3361zh:
> > > > This looks good, Matthew, just holding off because I'd like to see an example of this new feature working. I understand the default metadata works the same as before but the default is to not use this new enhancement. Could you provide an example that shows that this new feature works as expected?
> > > 
> > > 
> > > These are in the tests, the tests did cover the synthetic fertilizer option since it's the only other option implemented right now.
> > 
> > 
> > When I run the default simulation, is the synthetic fertilizer option implemented?
> 
> No, and I don't think any simulation that we have right now utilizes that part. What I can think of that is most relevant to checking the synthetic fertilizer part is only the tests.

If a SME came and asked how to use this feature, how would they do it?

matthew7838:
> > > > > This looks good, Matthew, just holding off because I'd like to see an example of this new feature working. I understand the default metadata works the same as before but the default is to not use this new enhancement. Could you provide an example that shows that this new feature works as expected?
> > > > 
> > > > 
> > > > These are in the tests, the tests did cover the synthetic fertilizer option since it's the only other option implemented right now.
> > > 
> > > 
> > > When I run the default simulation, is the synthetic fertilizer option implemented?
> > 
> > 
> > No, and I don't think any simulation that we have right now utilizes that part. What I can think of that is most relevant to checking the synthetic fertilizer part is only the tests.
> 
> If a SME came and asked how to use this feature, how would they do it?

They should update the inputs just like how it as before the changes, the only difference is that now they should change `manure_schedule` files instead of `field` files.

ew3361zh:
> > > > > > This looks good, Matthew, just holding off because I'd like to see an example of this new feature working. I understand the default metadata works the same as before but the default is to not use this new enhancement. Could you provide an example that shows that this new feature works as expected?
> > > > > 
> > > > > 
> > > > > These are in the tests, the tests did cover the synthetic fertilizer option since it's the only other option implemented right now.
> > > > 
> > > > 
> > > > When I run the default simulation, is the synthetic fertilizer option implemented?
> > > 
> > > 
> > > No, and I don't think any simulation that we have right now utilizes that part. What I can think of that is most relevant to checking the synthetic fertilizer part is only the tests.
> > 
> > 
> > If a SME came and asked how to use this feature, how would they do it?
> 
> They should update the inputs just like how it as before the changes, the only difference is that now they should change `manure_schedule` files instead of `field` files.

I would suggest putting this in the "test plan" section. That section should normally include how you tested out the changes to the code by running simulations and how you showed that it worked in action (testing changes to particular input files and that the behavior/outputs are as you expect them to be from running a simulation).

ew3361zh (approved):
Nice work, Matthew!

Approving because I think the most important parts have been implemented and are working well as best I can tell.

Similar to Matthew's comments, I was able to add some synthetic fertilizer by maxing out the nitrogen masses (all 50000000) in the `default_manure_schedule.json` and having `synthetic_fertilizer` as the option in the `"supplement_manure_nutrient_deficiencies"` field in that file. So the mechanism seems to work as intended to trigger synthetic fertilizer use or not and it properly sets up manure supplementation in the sister issue I'm working on.

Please address the outstanding comments before merging though:

1. Properly order docstring params for ManureEvent class to match order in the init.
2. Add manure_supplement_method to docstring definitions (also in proper order) for `Field._execute_manure_application()`.

RUFAS/routines/field/field/field.py
```diff
@@ -649,7 +652,7 @@ def _execute_manure_application(
         if unmet_nitrogen_demand == 0.0 and unmet_phosphorus_demand == 0.0:
             return
 
-        if not self.field_data.supplement_manure_nutrient_deficiencies:
+        if manure_supplement_method == ManureSupplementMethod.NONE:
             warning_name = "Nutrient deficient manure application"
             warning_message = (
                 f"Manure nitrogen deficient by {unmet_nitrogen_demand} kg, manure phosphorus "
```
ew3361zh, line 660: Should the block of code right below now be in an `elif manure_supplement_method == ManureSupplementMethod.SYNTHETIC_FERTILIZER:` block?

Because we only want this to be applied if the user has selected the SYNTHETIC_FERTILIZER option and here it would be applied as long as they don't select the NONE option.

So my thought is it should be:
```
...
elif manure_supplement_method == ManureSupplementMethod.SYNTHETIC_FERTILIZER:
    if unmet_nitrogen_demand > 0.0 and unmet_phosphorus_demand == 0.0:
        optimal_mix = self.ONLY_NITROGEN_MIX
    else:
            optimal_mix = self._determine_optimal_fertilizer_mix(
                unmet_nitrogen_demand,
                unmet_phosphorus_demand,
                self.available_fertilizer_mixes,
            )
        self._execute_fertilizer_application(
            optimal_mix,
            unmet_nitrogen_demand,
            unmet_phosphorus_demand,
            0,
            application_depth,
            surface_remainder_fraction,
            year,
            day,
       )
```

matthew7838: I agree! I was trying not to alter any logic here but this makes complete sense to be changed.

RUFAS/data_structures/manure_supplement_methods.py
KFosterReed: This is a great point Niko! If using the supplement with manure option, most cases would result in one nutrient still being deficient or over applied. For now, I would vote for over applied but having that third supplementation option to exactly meet the nutrient requirements would be great

KFosterReed (approved):
This is a nice step towards having more options and control over the nutrient supplements when there is not enough manure for application. All of the inputs are accepted and the synthetic fertilizer input functions as expected in that when a nutrient request exceeds the amount of that nutrient in the ManureNutrient pool, it triggers a fertilizer application on that day in an amount that sums to the original nutrient request. 

The inputs for "manure" and "synthetic fertilizer and manure" are accepted but do not result in any changes in nutrient application at this time but that will follow after we have the ability to apply manure from external sources :)

PR #2153, opened 2024-12-23
[Animal] Refresh Calf Ration Intake and Requirements calculations

### LINKED ISSUE #1137: Refactoring Calf Ration

### LINKED ISSUE #1337: Unable to specify feed milk replacer (203) rather than whole milk (202) in calf ration

Isgotkowitz:
Refreshes the routines for calculating calf nutrient intake and requirements, and integrates it into the refreshed routines for calculating animal requirements and rations.

### Context

Issue(s) closed by this pull request: closes #1137, closes #1337, addresses #2130

### What

Refreshes the `CalfRationManager` so that it can be utilized in the refreshed animal module. Also makes some of the calf ration stuff more explicit, which will hopefully help the team make more informed decisions about how to handle calf nutrition in the future, because it is really different from the rest of the animals' nutrition routines.

### Why

Calves need to eat to! And the calf nutrition routines have been put on the backburner for a while in RuFaS, so this refresh gives us the opportunity to reintegrate them into the model and re-evaluate how we can improve them.

### How

Refreshes the `CalfRationManager` by moving it to a new, more appropriately named file and removing the old one. Removes the code that provided a hard-coded calf ration.

Refactored to the two key methods of `CalfRationManager` so that there are fewer abbreviations and references are in the docstrings, not in-line comments.

Reimplemented the changes made in #2043 to `calc_intake`. There were some changes I did not carry over from #2043, namely the implementation of a user-defined ration and a ration optimizer for calf rations. This is because the refreshed `UserDefinedRationManager` will work for calf rations, and since only user-defined rations will be usable initially in the refreshed animal module I thought it better to implement optimized calf rations at later date.

Also integrated the new `CalfRationManager` into animal, so that it is used to calculate the requirements of calves.

### Test plan

A unit test suite has not been written in this PR, but because the refreshed animal module will undergo heavy testing we will catch if something isn't working right in the calf ration routines.


RUFAS/biophysical/animal/ration/calf_ration_manager.py
```diff
@@ -0,0 +1,245 @@
...
+from numpy import exp, log
+
+from RUFAS.data_structures.feed_storage_to_animal_connection import Feed
+from RUFAS.general_constants import GeneralConstants
+
+
+"""RuFaS IDs for supported calf feeds."""
+WHOLE_MILK_ID = 202
+MILK_REPLACER_ID = 203
+STARTER_ID = 216
```
Isgotkowitz:
```suggestion
from RUFAS.data_structures.feed_storage_to_animal_connection import RUFAS_ID, Feed
from RUFAS.general_constants import GeneralConstants


"""RuFaS IDs for supported calf feeds."""
WHOLE_MILK_ID: RUFAS_ID = 202
MILK_REPLACER_ID: RUFAS_ID = 203
STARTER_ID: RUFAS_ID = 216
```

RUFAS/biophysical/animal/ration/calf_ration_manager.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+class CalfMilkType(Enum):
+    """Calf milk types."""
+
+    WHOLE = "whole"
+    REPLACER = "replacer"
```
Isgotkowitz, line 19: Calves are very picky about their milk. If they have any whole milk at all in their diet, then 100% of their milk intake is whole. Otherwise it is 100% replacer. This enum and the class var that uses it was created to make this clearer

RUFAS/biophysical/animal/animal.py
```diff
@@ -617,16 +619,50 @@ def calculate_nutrition_requirements(
...
+                threonine=0.0,
+                tryptophan=0.0,
+                valine=0.0,
+                isoleucine=0.0,
+                arginine=0.0,
+                histidine=0.0,
+                leucine=0.0,
+                phenylalanine=0.0,
+            )
+            self.growth.daily_growth = calf_requirements["live_weight_change"]
```
Isgotkowitz: "live_weight_change" was originally the value that the calves grew by every day. Currently in the refreshed `Growth` module, calves grow by a static fraction of their wean weight every day. A follow up to the refresh would be to change this so that calves are growing according to this live weight change.

RUFAS/biophysical/animal/animal.py
```diff
@@ -617,16 +619,50 @@ def calculate_nutrition_requirements(
...
+                maintentance=calf_requirements["ne_maint"],
+                growth=calf_requirements["ne_gain"],
+                pregnancy=0.0,
+                lactation=0.0,
+                protein=calf_requirements["adp_maintentance"],
+                calcium=0.0,
+                phosphorus=0.0,
+                secondary_phosphorus=0.0,
+                activity=0.0,
+                essential_amino_acids=calf_essetial_amino_acids
```
Isgotkowitz: A lot of these are dummy values. I'm not confident in the maintenance, growth, or protein values. But! The dry matter is correct, and it is critical because it ensures that the `UserDefinedRationManager` will correctly scale the user-defined calf ration to the calf's dry matter intake

Isgotkowitz:
Wanted to rebase this off `dev`, but couldn't because the commit history was such a mess. Replaced by PR #2161, which implements the exact same changes.

PR #2155, opened 2025-01-01
[Git-Actions] Added default files and metadata to the protected paths

### LINKED ISSUE #931: Add metadata and input data files to protected files list in lint_flake8.yml
PR #911 adds a mechanism in the lint_flake8.yml that has a list of protected files which will trigger a GitHub linting Action failure if any changes have been made to any of those files.

At the time of the PR was merged, there were some key files which we thought would be good candidates for this list but which were under review and revision so adding them at that time wasn't practical.

Those files include:
default scenario metadata file
default scenario input data files

Other files can be added to this list as needed.

This issue serves as a reminder to add those files to the list of protected files once they are finished being revised.

matthew7838:
Adding default files to the protected paths list.
### Context

Issue(s) closed by this pull request: closes #931 

### What

All default inputs and no crop inputs were added to the protected lists.

### Why

We are confident that these default inputs are now stable, so they should be protected to avoid unauthorized changes.

### Test plan

Modified each of the added protected files and the GitHub action failed.


Isgotkowitz (approved):
Made a copy of this branch and tried out making changes to the unauthorized files, worked as expected. Nice job Matthew

.github/workflows/combined_format_lint_test_mypy.yml
```diff
@@ -42,8 +42,9 @@ jobs:
             BASE_BRANCH="dev"
             git fetch origin $BASE_BRANCH
           fi
-          ALL_CHANGED_FILES=$(git diff --name-only origin/$BASE_BRANCH HEAD)
-          echo "All Changed Files: $ALL_CHANGED_FILES"
+          ALL_CHANGED_FILES=$(git diff --name-only origin/$BASE_BRANCH HEAD | grep -v 'changelog.md' | tr '\n' ' ')
```
ew3361zh, line 45: Can I ask why you added the changelog here?

.github/workflows/combined_format_lint_test_mypy.yml
```diff
@@ -58,7 +59,24 @@ jobs:
...
+            "input/data/feed/default_feed_2.json"
+            "input/data/feed_management/default_feed_management.json"
+            "input/data/fertilizer_schedule/default_fertilizer_schedule.json"
+            "input/data/field/default_field.json"
+            "input/data/manure_schedule/default_manure_schedule.json"
+            "input/data/soil/default_soil.json"
+            "input/data/soil/default_soil_2.json"
+            "input/data/tillage_schedule/default_tillage_schedule.json"
           )
+          echo "All Changed Files: $ALL_CHANGED_FILES"
```
ew3361zh: Does echoing the changed files again here have a purpose? Looks like it's done on line 46 as well. Just curious.

matthew7838: Good catch, this is a print statement for debugging. I will remove it.

.github/workflows/combined_format_lint_test_mypy.yml
matthew7838: The .md file name violates some regex pattern that prevents it from being added to an environment variable. I did not bother looking into it for two reasons.

1. The change log will always be changed, saving it and letting the user know that this file was modified does not add much value to the information.
2. It's also certain that the change log does not need to be protected in the next task, so there's no value in adding it to the environment variable.

Let me know if this thought process makes sense.

ew3361zh (approved):
Nice work, Matthew - branched off this and tested successfully with a protected filepath.

Issue #2152, opened 2024-12-23
[Animal][Testing] Write unit test suite for refreshed animal nutrition modules


Isgotkowitz:
The refreshed animal nutrition modules need to have unit test suites written for them to verify and ensure correct calculation of nutrition requirements and supply.

## Status

The routines for calculating animal nutrition requirements and ration's nutrition supply are being refreshed in the following PRs:
- #2136
- #2138 
- #2141 
- #2142 
- #2144 
- #2161 

The new modules implemented in these PRs have a patchwork of unit test suites covering them.

## Intention

Bring test coverage of the new animal nutrition modules up to 100% to ensure that they work correctly and that they keep doing so.

## Request

Write unit and/or integration tests for all uncovered parts of the following modules:
- [ ] `NASEMRequirementsCalculator`
- [ ] `NRCRequirementsCalculator`
- [x] `NutritionRequirementsCalculator`
- [ ] `NutritionSupplyCalculator`
- [ ] `NutritionEvaluator`
- [ ] `CalfRationManager`

ew3361zh:
Completed with Animal Refresh.

PR #2131, opened 2024-12-09
[Animal] Refresh animal requirements and nutrients calculation

Isgotkowitz:
### Context

Issue(s) closed by this pull request: closes #2130

PR #2161, opened 2025-01-06
[Animal] Refresh Calf Ration Intake and Requirements calculations

### LINKED ISSUE #1137: Refactoring Calf Ration

### LINKED ISSUE #1337: Unable to specify feed milk replacer (203) rather than whole milk (202) in calf ration

Isgotkowitz:
Refreshes the routines for calculating calf nutrient intake and requirements.

### Context

Issue(s) closed by this pull request: closes #1137, closes #1337, addresses #2130

### What

Makes some of the calf ration stuff more explicit, which will hopefully help the team make more informed decisions about how to handle calf nutrition in the future, because it is really different from the rest of the animals' nutrition routines.

### Why

Calves need to eat to! And the calf nutrition routines have been put on the backburner for a while in RuFaS, so this refresh gives us the opportunity to reintegrate them into the model and re-evaluate how we can improve them.

### How

Refreshes the `CalfRationManager` by moving it to a new, more appropriately named file and removing the old one. Removes the code that provided a hard-coded calf ration.

Refactored to the two key methods of `CalfRationManager` so that there are fewer abbreviations and references are in the docstrings, not in-line comments.

Reimplemented the changes made in #2043 to `calc_intake`. There were some changes I did not carry over from #2043, namely the implementation of a user-defined ration and a ration optimizer for calf rations. This is because the refreshed `UserDefinedRationManager` will work for calf rations, and since only user-defined rations will be usable initially in the refreshed animal module I thought it better to implement optimized calf rations at later date.

### Test plan

A unit test suite has not been written in this PR, but because the refreshed animal module will undergo heavy testing we will catch if something isn't working right in the calf ration routines.


RUFAS/biophysical/animal/ration/calf_ration_manager.py
```diff
@@ -0,0 +1,266 @@
...
+        starter_me_intake = starter_metabolizable_energy * starter_intake
+        me_intake = milk_me_intake + starter_me_intake
+
+        milk_me_proportion = milk_me_intake / me_intake
+        starter_me_proportion = starter_me_intake / me_intake
+
+        milk_cp_intake = GeneralConstants.PERCENTAGE_TO_FRACTION * (
+            whole_milk.CP * whole_milk_intake + milk_replacer.CP * milk_replacer_intake
+        )
+        starter_cp_intake = GeneralConstants.PERCENTAGE_TO_FRACTION * starter.CP * starter_intake
```
JoeWaddell, line 239: If I'm reading this correctly (you did a nice job removing duplicate logic in `calc_intake` and `calc_requirements`! Just a fun challenge to sort out what's new and what's legacy code) this and `milk_cp_intake` aren't being assigned to the `animal_intake` return, so it won't be available when called via `calc_requirements`, and would cause the method to fail. Good call to put these equations here, they'll just need to be saved to the return

RUFAS/biophysical/animal/ration/calf_ration_manager.py
```diff
@@ -0,0 +1,266 @@
...
+
+        apparent_digestible_protein_allowable_gain = (
+            (apparent_digestible_protein_intake - maintentenance_apparent_digestible_protein)
+            * biological_value
+            / 0.188
+            * GeneralConstants.GRAMS_TO_KG
+        )
+        live_weight_change = min(energy_allowable_gain, apparent_digestible_protein_allowable_gain)
+
+        nutrient_requirements = {
```
JoeWaddell, line 130: Noting that the names and fields here overlap only partially with other animal types. Is the PR to lay out how we'll unify the Requirements objects across animal types? Or would that be better suited for another PR, like: https://github.com/RuminantFarmSystems/MASM/pull/2131

JoeWaddell (commented):
This looks good so far: nice work deduplicating logic and generally improving readability here!

Just a few open comments before approval, and one general question: we should keep #2043 open while waiting on the ration formulation refresh, correct? The reason being: this process-based in-house calf logic works backwards from the other animal types...namely that it calculates intake prior to reporting requirements, so the logic in 2043 reflects that. It could actually be added before the other animal types, since for calves the `calc_intake` methodology is essentially in going to be used in place of new formulation logic (as in 2043 it avoids current formulation logic), regardless. That said, the "calf exception for optimized formulation" likely merits its own PR.

RUFAS/biophysical/animal/ration/calf_ration_manager.py
Isgotkowitz: I think it would be better to suited to do this in a different PR, because it will take a lot of SME input to figure out how to adapt the calf ration stuff into the `NutritionSupply`, `NutritionRequirements`, and `NutritionEvaluationResults` structures.

My thought was that we can come up with a bandage solution in #2146/#2151 for the refreshed Animal module to run and correctly report calf stuff. After that, we can catch our breath and really sort out what is and isn't needed in `CalfRationManager`, and make any modifications necessary to both the CalfRationManager and nutrition data structures so that we can correctly handle calf ration stuff.

RUFAS/biophysical/animal/ration/calf_ration_manager.py
Isgotkowitz: Fixed! Nice catch Joe

RUFAS/biophysical/animal/ration/calf_ration_manager.py
```diff
@@ -0,0 +1,268 @@
...
+    ) -> dict[str, float]:
+        """
+        Calculate dietary intake and nutrient requirements for the calf.
+
+        Parameters
+        ----------
+        days_born : int
+            Number of days since the calf was born.
+        body_weight : float
+            Body weight of the calf (kg)
```
matthew7838:
```suggestion
            Body weight of the calf (kg).
```

RUFAS/biophysical/animal/ration/calf_ration_manager.py
```diff
@@ -0,0 +1,268 @@
...
+        Parameters
+        ----------
+        milk_type : CalfMilkType
+            The type of milk being fed to the calf.
+
+        """
+        cls.milk_type = milk_type
+
+    @classmethod
+    def calc_requirements(
```
matthew7838, line 44: This method does not need to access any properties or methods of the class, making this static might be a better fit.

matthew7838 (approved):
Looks great Ed. Left a small comment. Also, are there any issues tracking the missing unit tests?

RUFAS/biophysical/animal/ration/calf_ration_manager.py
Isgotkowitz: Good point! If it were just `calc_requirements` in this class, I agree it would be better as a staticmethod. But because there is also `set_milk_type` and `calc_intake`, which need to be classmethods, it would be better to make `calc_requirements` a classmethod so that the entire class is made up of classmethods.

It is not a hard rule that all methods in a class have to be the same type (instance, class or static), but I have gotten the feedback before that it is not good practice to mix and match method types in a class unless there is a very compelling reason to do so.

Isgotkowitz:
> Looks great Ed. Left a small comment.

Thanks Matthew. 

> Also, are there any issues tracking the missing unit tests?

Yes, #2152 is about bringing coverage back up on the unit test suite for the refreshed nutrition code. It didn't originally cover `CalfRationManager`, but I just updated it to. Thanks for bringing this up!

JoeWaddell:
Great work: I look forward to seeing this in action with the refreshed Animal module!

JoeWaddell (approved):

PR #2162, opened 2025-01-07
[DCA] Replace the usage of PROPERTIES_TO_CREATE_SCHEMA_FOR

### LINKED ISSUE #2125: [DCA] Set whether DCA schema should be created for an input file in the metadata properties
When determining which properties to create input schemas for, the `DataCollectionAppUpdater` should reference the metadata properties instead of using a hardcoded list.

## Status

When `DataCollectionAppUpdater` is updating the DCA schemas, it determines whether a schema should be generated for a properties by checking if it is in the module constant `PROPERTIES_TO_CREATE_SCHEMA_FOR`.

If there is a new property in the metadata properties and a schema needs to be generated for it, this list has to be modified.

## Intention

Streamline the process of changing which properties schema are created for.

## Request

Add a top-level boolean field to each property entry called `"data_collection_app_compatible"`. When `DataCollectionAppUpdater` is iterating over the properties it checks this property. If it is true, a schema is generated for it, false if it isn't.

matthew7838:
Each property entry will have a field to indicate DCA schema compatibility instead of using the hardcoded list in the DCA codebase.
### Context

Issue(s) closed by this pull request: closes #2125

### What

- Each property entry will now have a "data_collection_app_compatible" field that takes in a boolean type to indicate its compatibility with creating DCA schema.

- The usage of the hardcoded list `PROPERTIES_TO_CREATE_SCHEMA_FOR` will be deprecated.

### Why

This prevents the need to dig into the code and modify the list every time there's a change to compatibility.

### How

Each property will use its own boolean field for compatibility check.

### Test plan

Make changes to each property's "data_collection_app_compatible" field, then run the DCA update task and make sure that the available schema reflects the changes to the "data_collection_app_compatible" field.

### DIFF
--- a/RUFAS/data_collection_app_updater.py
+++ b/RUFAS/data_collection_app_updater.py
@@ -16,23 +16,6 @@ INDEX_PATH: Path = Path("").joinpath("DataCollectionApp", "index.html")
 """Path to the template for regenerating the Data Collection App's home page."""
 TEMPLATE_PATH: Path = Path("").joinpath("DataCollectionApp", "template")
 
-"""
-Metadata properties for which the Data Collection App will generate schema.
-If a property is not listed, its schema will not be generated.
-"""
-PROPERTIES_TO_CREATE_SCHEMA_FOR: list[str] = [
-    "animal_properties",
-    "config_properties",
-    "crop_schedule_properties",
-    "feed_properties",
-    "fertilizer_schedule_properties",
-    "field_properties",
-    "manure_management_properties",
-    "manure_schedule_properties",
-    "tillage_schedule_properties",
-    "tractor_dataset_properties",
-]
-
 """Placeholder for inserting schema import scripts in index.html."""
 SCHEMA_SCRIPT_TAG_PLACEHOLDER: str = "    <!-- Schema imports go here-->"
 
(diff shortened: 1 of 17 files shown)


RUFAS/input_manager.py
```diff
@@ -978,8 +980,7 @@ def _validate_data(
 
         """
         validated_data = {}
-        variable_properties_to_ignore = ["type", "description", "modifiability"]
-
+        variable_properties_to_ignore = ["type", "description", "modifiability", "data_collection_app_compatible"]
```
allisterakun: At this point, I feel like we should make this a file constant and bring it to the front of the file (and make the name all caps), so that it's easier to locate and more clear to user. What do you think?

RUFAS/data_collection_app_updater.py
```diff
@@ -101,11 +84,15 @@ def _rewrite_schemas(self) -> list[Path]:
         properties: dict[str, Any] = self._im.meta_data["properties"]
 
         schema_paths = []
+
         for key in properties.keys():
-            if key not in PROPERTIES_TO_CREATE_SCHEMA_FOR:
+            if not properties[key]["data_collection_app_compatible"]:
```
allisterakun:
```suggestion
            data_collection_app_compatible = properties[key].get("data_collection_app_compatible", False)
            if not data_collection_app_compatible:
```
So that if for som reason, the "data_collection_app_compatible" key is not specified in the metadata properties, it wouldn't crash the simulation.

RUFAS/data_collection_app_updater.py
```diff
@@ -101,11 +84,15 @@ def _rewrite_schemas(self) -> list[Path]:
...
                 continue
+            schema_dict = {
+                key: value for key, value in properties[key].items() if key != "data_collection_app_compatible"
```
allisterakun: The variable name `key` here is the same as the temporary variable `key` used for the for loop, and later this line `properties[key]`.
Please change the variable name to avoid confusion when reading

allisterakun (approved):
Mostly look good! Have some suggestions about some minor details. Otherwise LGTM!

Isgotkowitz (approved):
Tried turning the new property flag on and off, DCA updates as expected. Allister made some great suggestions, once those are addressed I think this PR will be good to go!

PR #2165, opened 2025-01-10
[Crop and Soil] Combine crop's nutrient uptakes

### LINKED ISSUE #450: [Crop & Soil] Combine crop's uptake of water, phosphorus, nitrogen, etc. into a single module.
## Problem

Within the crop module, phosphorus uptake (`PhosphorusIncorporation`), nitrogen uptake (`NitrogenIncorporation`), and water uptake (`WaterUptake`) all use identical methods and should be grouped into a single `NutrientUptake` class, with possible child classes for nutrient-specific alterations of the methods. 

Additionally, the new `SoilData` format for layers (`soil_layers`) requires some retooling of the methods for nutrient transfer. 

## Suggested solution

Below is a skeleton of such a class that would contain the common functions, which use identical equations in SWAT across the three nutrients:

```python
from typing import Optional

from SC_redesign.Crop_and_Soil.crop.crop_data import CropData
from SC_redesign.Crop_and_Soil.soil.soil_data import SoilData

"""This module implement the crop's uptake of nutrients (water, N, P, etc.) from the soil"""

class NutrientUptake:
    def __int__(self, crop_data: Optional[CropData], soil_data: Optional[SoilData]):
        self.crop_data = crop_data or CropData()
        self.soil_Data = soil_data or SoilData()

    def find_deepest_accessible_layer(self):
        pass

    def access_layers(self):
        pass

    def find_uptake_potentials(self):
        pass

    def find_demands(self):
        pass

    def determine_actual_uptake(self):
        pass

    def extract(self):
        pass


class WaterUptake(NutrientUptake):
    def __init__(self, crop_data: Optional[CropData], soil_data: Optional[SoilData]):
        super().__init__(crop_data, soil_data)


class NitrogenUptake(NutrientUptake):
    def __init__(self, crop_data: Optional[CropData], soil_data: Optional[SoilData]):
        super().__init__(crop_data, soil_data)


class PhosphorusUptake(NutrientUptake):
    def __init__(self, crop_data: Optional[CropData], soil_data: Optional[SoilData]):
        super().__init__(crop_data, soil_data)
```

matthew7838:
Combine the crop's Nitrogen, Phosphorus, and water uptake using inheritance.
### Context

Issue(s) closed by this pull request: closes #450 

### What

- Most nitrogen incorporation functions that were used across different uptakes have been moved to a new parent NutrientUptake class.
- Crop data will be passed in and initialized in the parent class, but soil data will remain only be passed when the main routine function is called.
- Extracted common logic across different nutrient uptakes into the parent class.
- [Visualization](https://whimsical.com/crop-nutrient-uptake-hierarchy-Xo6fwDpgbGdsTgT3UoR2Eb)

### Why

To disentangle the uptakes, it does not make sense for phosphorus and water uptake to depend on the function of the nitrogen uptake class.

### How

Common functions will be implemented at a higher-level parent class.

### Test plan

- Run with multiple metadata and ensure the same outputs.
- Passing updated unit tests.


RUFAS/routines/field/crop/nutrient_uptake.py
```diff
@@ -0,0 +1,752 @@
...
+        layer_demand = [sum(layer_delta[:i]) for i in range(len(layer_delta))]
+        return [max(val, 0) for val in layer_demand]
+
+    @staticmethod
+    def determine_potential_nutrient_uptake(
+        demand: float,
+        nutrient_start: float,
+        mature_nutrient_fraction: float,
+        max_growth: float,
+    ) -> float:  # pseudocode: C.5.B.3
```
Isgotkowitz:
```suggestion
    ) -> float:
```

RUFAS/routines/field/crop/nutrient_uptake.py
```diff
@@ -0,0 +1,752 @@
...
+        nutrient_distribution_parameter: float,
+    ) -> list[float]:
+        """
+        Calculates the potential nutrient uptake from each soil layer based on plant demand and root depth.
+
+        Parameters
+        ----------
+        layer_bounds : list[float]
+            A list of lower boundaries for each soil layer, in ascending order (i.e., increasing depths). Each entry
+            represents the depth to the bottom of the layer.
```
Isgotkowitz:
```suggestion
            A list of lower boundaries for each soil layer, in ascending order (i.e., increasing depths). Each entry
            represents the depth to the bottom of the layer (mm).
```

RUFAS/routines/field/crop/nutrient_uptake.py
```diff
@@ -0,0 +1,752 @@
...
+        ----------
+        SWAT 5:2.3.2, 5:2.3.3, 5:2.3.20, 5:2.3.21
+
+        Raises
+        ------
+        ValueError
+            If half_mature_heat_fraction equals mature_heat_fraction.
+
+        """
+        if mature_heat_fraction == half_mature_heat_fraction:  # leads to divide by 0
```
Isgotkowitz:
```suggestion
        if mature_heat_fraction == half_mature_heat_fraction:
```

RUFAS/routines/field/crop/nutrient_uptake.py
```diff
@@ -0,0 +1,752 @@
+from bisect import bisect
+from math import exp, log
+from typing import Optional
+
+from RUFAS.output_manager import OutputManager
+from RUFAS.routines.field.crop.crop_data import CropData
+
+
+class NutrientUptake:
```
Isgotkowitz: Could you please add a docstring for `NutrientUptake` which briefly mentions this class contains overlapping logic for nitrogen, phosphorus, and water uptake?

Isgotkowitz: Pretty minor thing, but is it possible that there is a better term to use for nitrogen, phosphorus and water uptake than "nutrient"? I'm not sure if water is correctly characterized by "nutrient", this would be a good thing to ask the SMEs

changelog.md
```diff
@@ -84,6 +84,7 @@ v0.9.2
 - [2141](https://github.com/RuminantFarmSystems/MASM/pull/2141) - [minor change] [Animal] Refreshes module which calculates the nutrition requirements of an animal with the NASEM (2021) methodology.
 - [2142](https://github.com/RuminantFarmSystems/MASM/pull/2142) - [minor change] [Animal] Refreshes module which calculates the nutrition requirements of an animal with the NRC (2001) methodology.
 - [2144](https://github.com/RuminantFarmSystems/MASM/pull/2144) - [minor change] [Animal] Refreshes module which evaluates if the nutrition supply of a ration satisfies the demand of animal or the average demand of a group of animals.
+- [2165](https://github.com/RuminantFarmSystems/MASM/pull/2165) - [minor change] [Crop and Soil] Disentangle crop's nutrient uptake classes.
```
Isgotkowitz: I suggest rewriting this entry to not use the word "disentangle". I don't think this is a bad way to characterize the changes made in this PR but we have developed a pretty specific meaning for disentangle in RuFaS, and I don't think that this PR is necessarily disentangling the nutrient uptake logic.

Maybe something like "Reorganize logic of crop's nutrient uptake routines"?

tests/soil_crop_tests/crop_tests/test_nutrient_uptake.py
```diff
@@ -0,0 +1,504 @@
...
+
+@pytest.mark.parametrize(
+    "halfheat,heatfrac,emerge,half,near,mature,should_fail",
+    [
+        (0.5, 1.0, 0.8, 0.6, 0.3, 0.2, False),  # start
+        (0.99, 1.0, 0.8, 0.6, 0.3, 0.2, False),  # half_heat close to mature heat
+        (0.01, 1.0, 0.8, 0.6, 0.3, 0.2, False),  # small half_heat
+        (0.5, 1.0, 0.8, 0.6, 0.20001, 0.2, False),  # near very close to mature
+        (0.286, 0.54, 0.522, 0.4, 0.1, 0.08, False),  # arbitrary
+        # Above tests are copied from old subroutine tests
```
Isgotkowitz:
```suggestion
```

tests/soil_crop_tests/crop_tests/test_nutrient_uptake.py
```diff
@@ -0,0 +1,504 @@
...
+    heatfrac: float,
+    emerge: float,
+    half: float,
+    near: float,
+    mature: float,
+    should_fail: bool,
+    mocker: MockerFixture,
+) -> None:
+    """check that the shape parameters are correctly calculated by determine_nshapes() and that errors were raised
+    correctly"""
```
Isgotkowitz:
```suggestion
    """Check that the shape parameters are correctly calculated by determine_nshapes() and that errors were raised
    correctly."""
```
Please make same changes to all docstrings in this class.

tests/soil_crop_tests/crop_tests/test_phosphorus_incorporation.py
```diff
@@ -133,37 +36,35 @@ def test_uptake_phosphorus(phosphates, depths):
...
     incorp.find_deepest_accessible_soil_layer = MagicMock(return_value=None)
     incorp.access_layers = MagicMock(return_value=[1, 2, 3])
-    NitrogenIncorporation.determine_layer_nutrient_uptake_potential = MagicMock(return_value=[3.25, 6.33, 7.10])
-    NitrogenIncorporation.determine_layer_nutrient_demands = MagicMock(return_value=[12, 15, 17])
-    NitrogenIncorporation.determine_layer_nutrient_uptake = MagicMock(return_value=[8.9, 9.9, 13.12])
-    NitrogenIncorporation.determine_layer_extracted_resource = MagicMock(return_value=[5.0, 4.0, 2.0])
-    incorp.extend_phosphate_uptakes_to_full_profile = MagicMock()
-    incorp.extract_phosphorus_from_soil_layers = MagicMock()
-    incorp.tally_total_phosphorus_uptake = MagicMock()
+    NutrientUptake.determine_layer_nutrient_uptake_potential = MagicMock(return_value=[3.25, 6.33, 7.10])
```
Isgotkowitz: This pattern of assigning a method to a `MagicMock` instance causes Mypy errors. Instead could you please use the pattern `determine_layer_uptake_potential = mocker.patch.object(NutrientUptake, "determine_layer_nutrient_uptake_potential", return_value=[3.25, 6.33, 7.10]) instead?

RUFAS/routines/field/crop/nutrient_uptake.py
```diff
@@ -0,0 +1,752 @@
...
+
+        References
+        ----------
+        SWAT 5:2.3.5, 5:2.3.23
+
+        """
+        return min(demand - nutrient_start, 4 * mature_nutrient_fraction * max_growth)
+
+    @classmethod
+    def determine_layer_extracted_resource(cls, requests: list[float], sources: list[float]) -> list[float]:
```
Isgotkowitz: I like that you're using class methods in places where it is more appropriate than using static methods. However, there are a number of different method types in this class. Could you please find a way to implement all the methods in this class so that they are the same method type? I think that in this case, there aren't any particular benefits or constraints that necessitate mixing different class types.

RUFAS/routines/field/crop/phosphorus_incorporation.py
```diff
@@ -2,9 +2,9 @@
 
 from RUFAS.routines.field.crop.crop_data import CropData
 from RUFAS.routines.field.crop.nitrogen_incorporation import NitrogenIncorporation
```
Isgotkowitz:
```suggestion
```
Phosphorus incorporation should not need to import Nitrogen Incorporation at all.

RUFAS/routines/field/crop/phosphorus_incorporation.py
```diff
@@ -121,36 +120,36 @@ def incorporate_phosphorus(self, soil_data: SoilData) -> None:
...
             )
         self.uptake_phosphorus(layer_phosphates, layer_depths)
         soil_data.set_vectorized_layer_attribute("labile_inorganic_phosphorus_content", layer_phosphates)
-        self.data.phosphorus = NitrogenIncorporation.determine_stored_nutrient(
-            self.total_phosphorus_uptake, self.data.phosphorus, 0
+        self.crop_data.phosphorus = self.determine_stored_nutrient(
+            self.total_phosphorus_uptake, self.crop_data.phosphorus, 0
         )
 
     def uptake_phosphorus(self, layer_phosphates: List[float], layer_depths: List[float]) -> None:
```
Isgotkowitz: There is a method called `uptake_nitrogen` in `NitrogenIncorporation` which is extremely similar to this method. I think they could probably have the overlapping logic added into `NutrientUptake` in some fashion.

Isgotkowitz (changes requested):
This is an excellent start to deduplicating the nutrient uptake logic Matthew! However, I think there are still a number of ways that the code could be improved before merging. Please see comments I have left on here.

A good overall goal to keep in mind would be to have `NitrogenIncorporation`, `PhosphorusIncorporation`, and `WaterUptake` adhere to the Liskov-Substitution Principle. And while thinking about this, consider how this could be emphasized with things that don't have strict rules according to the Liskov-Substitution Principle, like the naming conventions.

RUFAS/routines/field/crop/phosphorus_incorporation.py
matthew7838: I was hesitant because the uptake of water is a completely different thing. After a second thought, I think it is fine to extract this.

Isgotkowitz: Good point. It may be easy to extract the overlapping logic for N and P, but if it doesn't work for water it does n't work.

matthew7838: Another difficulty is that both methods heavily involve modifying attributes that only make sense to be in the Nitrogen and Phosphorus class.

Isgotkowitz: Right, but the Nitrogen and Phosphorus attributes mirror each other almost perfectly. So how could the attributes be restructured so that they they are agnostic to whether they are N or P related?

matthew7838: I initially wanted to do that but we can't move them just to NutrientUptake class but then water uptakes do not have the required variables to initialize. What do you think about another class called ElementUptake(not sure what's a good name) that inherits the NutrientUptake with the new agnostic variables and then the N and P uptakes inherits the  ElementUptake class?

RUFAS/routines/field/crop/crop.py
```diff
@@ -79,8 +79,8 @@ def __init__(self, crop_data: Optional[CropData] = None):
         self._biomass_allocation = BiomassAllocation(self._data)
         self._water_dynamics = WaterDynamics(self._data)
         self._water_uptake = WaterUptake(self._data)
-        self._nitrogen_incorporation = NitrogenIncorporation(self._data)
-        self._phosphorus_incorporation = PhosphorusIncorporation(self._data)
+        self._nitrogen_incorporation = NitrogenUptake(self._data)
```
ew3361zh: What do you think about renaming to `_nitrogen_uptake`/`phosphorus_uptake` to reflect the object name? I think that would be my preference to avoid confusion.

RUFAS/routines/field/crop/crop.py
```diff
@@ -103,7 +103,7 @@ def biomass_allocation(self) -> BiomassAllocation:
         return self._biomass_allocation
 
     @property
-    def nitrogen_incorporation(self) -> NitrogenIncorporation:
+    def nitrogen_incorporation(self) -> NitrogenUptake:
         """Provides access to the NitrogenIncorporation object."""
```
ew3361zh:
```suggestion
        """Provides access to the NitrogenUptake object."""
```
And also change the return if you end up changing the attribute name as well.

RUFAS/routines/field/crop/crop.py
```diff
@@ -123,7 +123,7 @@ def crop_management(self) -> CropManagement:
         return self._crop_management
 
     @property
-    def phosphorus_incorporation(self) -> PhosphorusIncorporation:
+    def phosphorus_incorporation(self) -> PhosphorusUptake:
         """Provides access to the PhosphorusIncorporation object."""
```
ew3361zh:
```suggestion
        """Provides access to the PhosphorusUptake object."""
```

RUFAS/routines/field/crop/non_water_uptake.py
```diff
@@ -0,0 +1,175 @@
...
+    Unmet nutrient demands by overlaying soil layers (kg/ha).
+nutrient_requests : Optional[float]
+    Nutrient requested from each soil layer (kg/ha).
+actual_nutrient_uptakes : Optional[List[float]]
+    Actual nutrient uptake from each soil layer (kg/ha).
+total_nutrient_uptake : Optional[float]
+    Total nutrient uptake by the plant (kg/ha).
+
+
+"""
```
ew3361zh: Is this supposed to be here?

ew3361zh:
I'm not sure where in the process this PR is as far as addressing existing comments but it feels like there is a lot of big rearrangements happening but I'm not sure I see how it makes the code cleaner (and it still seems like there's a lot of duplicative code in the P and N Uptake classes).
Maybe could you do a diagram of the new structure of these classes - how they're inheriting from one another? I feel like a visual could be very helpful for this one. And maybe we could discuss in a WT?

RUFAS/routines/field/crop/crop.py
matthew7838: Totally agree, thanks for catching this!

RUFAS/routines/field/crop/nitrogen_uptake.py
```diff
@@ -0,0 +1,357 @@
...
+        soil_data.set_vectorized_layer_attribute("nitrate_content", layer_nitrates)
+        total_accessible_nitrates = sum(self.access_layers(layer_nitrates))
+        self.try_fixation(total_accessible_nitrates, soil_water_factor)
+        self.crop_data.nitrogen = self.determine_stored_nutrient(
+            self.total_nutrient_uptake,
+            self.crop_data.nitrogen,
+            self.fixed_nitrogen,
+        )
+
+    # ---- member functions (setters, internal utility, call sub-routines) ----
```
Isgotkowitz:
```suggestion
```

RUFAS/routines/field/crop/phosphorus_uptake.py
```diff
@@ -0,0 +1,131 @@
...
+
+from RUFAS.routines.field.crop.crop_data import CropData
+from RUFAS.routines.field.crop.non_water_uptake import NonWaterUptake
+from RUFAS.routines.field.soil.soil_data import SoilData
+
+"""
+This module is based upon the 'Phosphorus Uptake" section (5:2.3.2) of of the SWAT model documentation
+
+Also: Since much of the functionality is similar to nitrogen (5:2.3.2), many of the static Nitrogen functions are
+called directly.
```
Isgotkowitz:
```suggestion
```

RUFAS/routines/field/crop/phosphorus_uptake.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+    References
+    ----------
+    'Phosphorus Uptake' section (5:2.3.2) of the SWAT.
+
+    Notes
+    -----
+    Since much of the functionality is similar to nitrogen (5:2.3.2), many of the static Nitrogen functions are
+    called directly.
```
Isgotkowitz:
```suggestion

    References
    ----------
    'Phosphorus Uptake' section (5:2.3.2) of the SWAT.

```

Isgotkowitz (commented):
This is looking better Matthew, but there are still some steps that could be taken to make this inheritance structure cleaner and less duplicative.

First, one minor suggestion I have is to change the naming structure. Instead of "NutrientUptake", I suggest "Uptake" as the name of the base class. "Uptake" would be inherited by "WaterUptake" and "NutrientUptake", and "NutrientUptake" would be inherited by "NitrogenUptake" and "PhosphorusUptake".

There are also some things that you could do that would help to make the interfaces of the methods more uniform (using the naming scheme suggested above):
- Add an `abstractmethod` to "Uptake" with the signature `def uptake(self, soil_data: SoilData) -> None:`. This will be the main access point to all the final uptake classes.
- Add a routine to "NutrientUptake" called "uptake_nutrients", which implements the overlapping logic between "incorporate_nitrogen" and "incorporate_phosphorus". I think this routine will have to take in a lot of values and spit out a lot, but it will really help to deduplicate the N and P routines.

matthew7838:
> This is looking better Matthew, but there are still some steps that could be taken to make this inheritance structure cleaner and less duplicative.
> 
> First, one minor suggestion I have is to change the naming structure. Instead of "NutrientUptake", I suggest "Uptake" as the name of the base class. "Uptake" would be inherited by "WaterUptake" and "NutrientUptake", and "NutrientUptake" would be inherited by "NitrogenUptake" and "PhosphorusUptake".
> 
> There are also some things that you could do that would help to make the interfaces of the methods more uniform (using the naming scheme suggested above):
> 
> * Add an `abstractmethod` to "Uptake" with the signature `def uptake(self, soil_data: SoilData) -> None:`. This will be the main access point to all the final uptake classes.
> * Add a routine to "NutrientUptake" called "uptake_nutrients", which implements the overlapping logic between "incorporate_nitrogen" and "incorporate_phosphorus". I think this routine will have to take in a lot of values and spit out a lot, but it will really help to deduplicate the N and P routines.

This is a good suggestion for the abstract method. However, I have confirmed with SMEs that water is considered a nutrient here, so it does not make sense to do the WaterUptake/NutrientUptake separation. The most sensible name I discussed with Pooya is the current naming convention, which is not ideal, but not a lot of options.

RUFAS/routines/field/crop/nutrient_uptake.py
```diff
@@ -0,0 +1,771 @@
...
+        -------
+        List[float]
+            Amount of nutrient mass taken up from each soil layer.
+
+        References
+        ----------
+        SWAT 5:2.3.1, 5:2.3.2 (see paragraphs below equations 5:2.3.8 and 5:2.3.26)
+
+        """
+        # ensure all list inputs are the same length
```
Isgotkowitz:
```suggestion
```

RUFAS/routines/field/crop/nitrogen_uptake.py
```diff
@@ -0,0 +1,328 @@
...
+        The symbiotic organisms that fix nitrogen exist at different densities depending upon the age of the plant. This
+        growth stage factor reflects the density and activity level of these symbiotic organisms relative to the plant's
+        growth stage.
+
+        References
+        ----------
+        SWAT 2:2.3.10 - 2:2.3.14
+
+        """
+        # piece-wise function:
```
Isgotkowitz:
```suggestion
```

RUFAS/routines/field/crop/nitrogen_uptake.py
```diff
@@ -0,0 +1,328 @@
...
+    References
+    ----------
+    'Nitrogen Uptake' section (5:2.3.1) of the SWAT model.
+
+    """
+
+    def __init__(
+        self,
+        crop_data: Optional[CropData] = None,
+        nutrient_distro_param: float = 10.0,
```
Isgotkowitz, line 79: Looked into this default, doesn't seem to be specified anywhere. Exact same thing for Phosphorus distribution parameter. Fortunately, according to SWAT this value shouldn't have too large an effect on nutrient uptake so it probably isn't a huge deal. But even so, this should likely be a user input.

I think that Clay introduced it into the code (#290 for Phosphorus, Nitrogen was earlier), used a value of 10.0 and it stuck.

SWAT uses this in equations 5:2.3.6 (for N) and 5:2.3.24 (for P) ([SWAT theoretical docs](https://swat.tamu.edu/media/99192/swat2009-theory.pdf)). This value is specified in SWAT's [.bsn input file](https://swat.tamu.edu/media/69311/ch04_input_bsn.pdf), see N_UPDIS and P_UPDIS. If no value is entered, then this value defaults to 20.

Will write up an issue for making this change.

RUFAS/routines/field/crop/non_water_uptake.py
```diff
@@ -0,0 +1,171 @@
...
+        self.layer_nutrient_potentials = layer_nutrient_potentials
+        self.unmet_nutrient_demands = unmet_nutrient_demands
+        self.nutrient_requests = nutrient_requests
+        self.actual_nutrient_uptakes = actual_nutrient_uptakes
+        self.total_nutrient_uptake = total_nutrient_uptake
+
+    def uptake_main_process(self, soil_data: SoilData, nutrient_name: str, soil_layer_attr: str) -> None:
+        """
+        Generic nutrient uptake routine that extracts nutrient from the soil and updates crop_data.
+        """
```
Isgotkowitz: Docstring here

RUFAS/routines/field/crop/non_water_uptake.py
```diff
@@ -0,0 +1,171 @@
...
+        self.actual_nutrient_uptakes = self.determine_layer_extracted_resource(
+            self.nutrient_requests, accessible_nitrates
+        )
+
+        self.extend_nutrient_uptakes_to_full_profile(self.actual_nutrient_uptakes)
+        self.extract_nutrient_from_soil_layers(layer_nutrient, self.actual_nutrient_uptakes)
+        self.total_nutrient_uptake = self.tally_total_nutrient_uptake(self.actual_nutrient_uptakes)
+
+    def shift_nutrient_time(self, nutrient: float) -> None:
+        """Copies the current nutrient value to previous_nitrogen (for use between time steps)."""
```
Isgotkowitz:
```suggestion
        """Copies the current nutrient value to previous_nutrient (for use between time steps)."""
```

RUFAS/routines/field/crop/non_water_uptake.py
```diff
@@ -0,0 +1,171 @@
+from abc import ABC
+from typing import Optional
+
+from RUFAS.routines.field.crop.crop_data import CropData
+from RUFAS.routines.field.crop.nutrient_uptake import NutrientUptake
+from RUFAS.routines.field.soil.soil_data import SoilData
+
+
+class NonWaterUptake(NutrientUptake, ABC):
```
Isgotkowitz:
```suggestion
class NonWaterUptake(NutrientUptake):
```
Non water uptake does not need to inherit the Abstract Base Class, because it doesn't have any unique abstract methods.

RUFAS/routines/field/crop/nutrient_uptake.py
```diff
@@ -0,0 +1,771 @@
...
+from bisect import bisect
+from math import exp, log
+from typing import Optional
+
+from RUFAS.output_manager import OutputManager
+from RUFAS.routines.field.crop.crop_data import CropData
+from RUFAS.routines.field.soil.soil_data import SoilData
+
+
+class NutrientUptake:
```
Isgotkowitz:
```suggestion
class NutrientUptake(ABC):
```

RUFAS/routines/field/crop/nutrient_uptake.py
Isgotkowitz, line 1: Many of the methods defined in `NutrientUptake` are only utilized in `NonWaterUptake`, so they need to be moved there. For the ones that do actually need to be in `NutrientUptake`, please make sure that the units listed are correct. This is a little tricky because the N and P units are in kg / ha, but water is in mm.

Isgotkowitz (commented):
This is looking better, but there are some changes that need to be made in order to better organize the new classes.

Re: the ABC stuff, this is the current behavior of the code:
```python
>>> from RUFAS.routines.field.crop.nutrient_uptake import NutrientUptake
>>> from RUFAS.routines.field.crop.non_water_uptake import NonWaterUptake
>>> NutrientUptake([1])
<RUFAS.routines.field.crop.nutrient_uptake.NutrientUptake object at 0x7ffb2a306fc0>
>>> NutrientUptake([1])
<RUFAS.routines.field.crop.nutrient_uptake.NutrientUptake object at 0x7ffb2a307530>
```
And this is what should be happening:
```python
>>> from RUFAS.routines.field.crop.non_water_uptake import NonWaterUptake
>>> from RUFAS.routines.field.crop.nutrient_uptake import NutrientUptake
>>> NutrientUptake([1])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Can't instantiate abstract class NutrientUptake without an implementation for abstract method 'uptake'
>>> NonWaterUptake([1])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Can't instantiate abstract class NonWaterUptake without an implementation for abstract method 'uptake'
```

tests/soil_crop_tests/crop_tests/test_water_uptake.py
```diff
@@ -44,14 +44,14 @@ def test_uptake_water(mock_crop_data: CropData, max_trans: float) -> None:
     """ensure that uptake_water can run without error"""
     # This patch is a quick fix for the mock from NitrogenIncorporation spilling over into this one.
     with patch(
-        "RUFAS.routines.field.crop.nitrogen_incorporation.NitrogenIncorporation." "determine_layer_nutrient_demands",
+        "RUFAS.routines.field.crop.nitrogen_uptake.NitrogenUptake." "determine_layer_nutrient_demands",
```
Isgotkowitz:
```suggestion
        "RUFAS.routines.field.crop.nitrogen_uptake.NitrogenUptake.determine_layer_nutrient_demands",
```

tests/soil_crop_tests/crop_tests/test_phosphorus_uptake.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def test_uptake(mocker: MockerFixture, mock_crop_data: CropData) -> None:
+    """Check that uptake() correctly called functions and variables were updated as expected."""
+    uptake = PhosphorusUptake(mock_crop_data)
+    soil = SoilData(field_size=10)
+    mock_main_uptake = mocker.patch.object(uptake, "uptake_main_process")
+    mock_determine_stored = mocker.patch.object(uptake, "determine_stored_nutrient", return_value=1)
+    uptake.uptake(soil)
+    mock_main_uptake.assert_called_once()
```
Isgotkowitz:
```suggestion
    mock_determine_stored = mocker.patch.object(uptake, "determine_stored_nutrient", return_value=1)
    
    uptake.uptake(soil)
    
    mock_main_uptake.assert_called_once()
```

RUFAS/routines/field/crop/nutrient_uptake.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @staticmethod
+    def tally_total_nutrient_uptake(actual_nutrient_uptakes: list[float]) -> float:
+        """
+        Determines total nutrient extracted from soil by summing actual uptake from each layer.
+
+        Parameters
+        ----------
+        actual_nutrient_uptakes : Optional[List[float]]
+            Actual nutrient uptake from each soil layer (kg/ha).
```
Isgotkowitz:
```suggestion
            Actual nutrient uptake from each soil layer (kg/ha or mm).
```

Isgotkowitz (approved):
Thank you for working through all the improvements to this PR, Matthew! It looks great and is a big step forward in introducing cleaner code and better organization to the Crop module.

RUFAS/routines/field/crop/phosphorus_uptake.py
```diff
@@ -0,0 +1,92 @@
...
+    potential_nutrient_uptake : Optional[float], default None
+        Potential phosphorus to be taken up by the plant under ideal circumstances for the current day (kg/ha).
+    actual_nutrient_uptakes : Optional[List[float]], default None
+        Actual phosphorus to be taken up by the plant from each soil layer (kg/ha).
+    layer_nutrient_potentials : Optional[float], default None
+        Potential phosphorus uptake from each soil layer (kg/ha).
+    unmet_nutrient_demands : Optional[float], default None
+        Unmet phosphorus demands by overlaying soil layers (kg/ha).
+    nutrient_requests : Optional[float], default None
+        Phosphorus requested from each soil layer (kg/ha).
```
ew3361zh, line 38: It looks like in NitrogenUptake these were added as Attributes as well. Do you want to add them here to match?

ew3361zh (approved):
Looks really good, Matthew. Well organized, thought out, and executed. Left one comment about matching docstrings and it looks like `non_water_uptake.py` has some lines missing (see attached photo). Please address those before merging but really well done overall!

Screenshot 2025-02-06 at 4 35 45 PM

PR #2172, opened 2025-01-21
[DCA] Add TaskManager tasks as input that can be collected with Data Collection App.

### LINKED ISSUE #2123: [DCA] Add "Tasks" as a type of input file that can be created with the DCA
Add "Tasks" as a type of input that can be created with the DCA.

## Status

Only a subset of base-level input files can be created with the DCA i.e., animal, manure, feed, etc. Task Manager "Task" files cannot be created with the DCA. 

## Intention

Improve the user experience by reducing how much interaction is needed with raw JSON files.

## Request

Modify the Task Manager so that the properties of "Task" inputs are passed when updating the DCA inputs, and have the DCA Updater in RuFaS create an input schema for tasks.

Isgotkowitz:
Makes TaskManager tasks an input that can be collected with the Data Collection App.

### Context

Issue(s) closed by this pull request: closes #2123

### What

Changes how the `TaskManager` handles the DCA update task, so that the DCA updater can add tasks as a type of input that can be collected with the Data Collection App.

### Why

Tasks are a type of input that many users will need to create that are specific to the simulations they want to run. By making them collectable with the DCA, we make the process of creating these inputs easier.

### How

Modified the `TaskManager` to collect and pass the metadata properties from its `InputManager` as one of the values in the dictionary passed to the workers which actually execute tasks. Also changed the DCA update routines to pass these properties into the DCA Updater class.

Modified the DCA Updater class to parse the properties passed to it in addition to the ones that are pulled from its `InputManager` instance.

### Test plan

Updated and ran the unit test suite to ensure correct behavior.

Also reran the DCA Update task (by changing `default_task.json` to `dca_update_task.json` in `task_manager_metadata.json`). Opened up the DCA and checked that tasks could be collected properly.

### DIFF
--- a/DataCollectionApp/index.html
+++ b/DataCollectionApp/index.html
@@ -18,7 +18,6 @@
     <script src="./schema/manure_management_schema.js"></script>
     <script src="./schema/field_schema.js"></script>
     <script src="./schema/tractor_dataset_schema.js"></script>
-    <script src="./schema/tasks_schema.js"></script>
     <link rel='stylesheet' id='theme-link'>
     <link rel='stylesheet' id='iconlib-link'>
     <style>
@@ -107,7 +106,7 @@
             "items": {
                 "type": "object",
                 "title": "Data Entry",
-                "anyOf": [config_schema, animal_schema, crop_schedule_schema, fertilizer_schedule_schema, manure_schedule_schema, tillage_schedule_schema, feed_schema, manure_management_schema, field_schema, tractor_dataset_schema, tasks_schema]
+                "anyOf": [config_schema, animal_schema, crop_schedule_schema, fertilizer_schedule_schema, manure_schedule_schema, tillage_schedule_schema, feed_schema, manure_management_schema, field_schema, tractor_dataset_schema]
             }
         }
         var data = {}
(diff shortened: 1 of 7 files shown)


DataCollectionApp/index.html
Isgotkowitz, line 1: These updates were made automatically by the DCA updater task.

input/metadata/properties/tasks_properties.json
```diff
@@ -29,7 +29,7 @@
         "type": "object",
         "task_type": {
           "type": "string",
-          "pattern": "HERD_INITIALIZATION|SIMULATION_SINGLE_RUN|SIMULATION_MULTI_RUN|SENSITIVITY_ANALYSIS|INPUT_DATA_AUDIT|END_TO_END_TESTING|POST_PROCESSING|COMPARE_METADATA_PROPERTIES|DATA_COLLECTION_APP_UPDATE",
+          "pattern": "^(HERD_INITIALIZATION|SIMULATION_SINGLE_RUN|SIMULATION_MULTI_RUN|SENSITIVITY_ANALYSIS|INPUT_DATA_AUDIT|END_TO_END_TESTING|POST_PROCESSING|COMPARE_METADATA_PROPERTIES|DATA_COLLECTION_APP_UPDATE)$",
```
Isgotkowitz, line 32: Without the `^(` and `)$`, this pattern will not be correctly parsed into a drop-down of options.

elle-andreen (approved):
The new "tasks" schema appears to be functioning as intended for me! I was able to regenerate the schemas and create a task JSON without a problem.

image

matthew7838 (approved):
Great work! Tried the changes and seems like it's working as intended on my side.

PR #2179, opened 2025-01-23
[C&S] Make crop growth and management configurations a user input

### LINKED ISSUE #1831: [Crop & Soil] Support modification and customization of growth and management parameters
Expand user control of modification and customization of the growth and management parameters (e.g. maximum leaf area index, harvest index, minimum harvest index, max root depth, radiation use efficiency, etc.) of the crop they select. A full list of crop growth params available here: https://swat.tamu.edu/media/69419/Appendix-A.pdf

Isgotkowitz:
Makes crops configurable to the user, so that they may adjust which crops can be grown in a simulation, how those crops grow, and how they are harvested.

### Context

Issue(s) closed by this pull request: closes #1831 

### What

The RuFaS crop model uses the same equations to determine how crops grow and are harvested, but different inputs to those equations depending on the variety of crop. So all that really differentiates alfalfa from corn from wheat etc. is a handful of different attributes. Previously those attributes were hardcoded into RuFaS, but this PR turns them over to the user.

### Why

Making the crop configurations a user input has two main benefits:
1. It makes the model much more powerful and flexible for farmers and researchers. Farmers can tune their crop configs to better model their farm. And researchers can grow any crop that they configure.
2. A sensitivity analysis of the crop module can now be done very easily, because the SA task type can now be applied to the crops.

### How

~Created a new input file that will hold all the crop configurations available in a simulation, `crop_configurations/default_crop_configurations.json`. There is also a new section of the metadata properties that defines how this input should be structured. The DCA has been updated as well, so that crop configurations can be collected with it.~
To reduce the size of this PR, all changes to inputs are made in a PR that is stacked below this one, #2196 

All old crop configuration files which contained child classes of `CropData` have been removed. A new factory class for making `CropData` instances has been created, called `CropDataFactory`. It replaces the old factory class, `CropSpeciesDataFactory`.

The fixed mapping between crop species and corresponding RuFaS Feed IDs has been removed from the Emissions Estimator, and the mapping is created as specified by the user in the configurations of crop files.

### Test plan

Updated the unit tests of all code that has changed, and wrote a unit test suite for `CropDataFactory`.

Ran end-to-end testing to verify that Crop and Soil results have not changed. Compared outputs between this branch and `dev` from the Emissions Estimator to make sure that was also functioning properly. There was no difference in output values, but the columns were in slightly different order.

~Opened up the DCA and checked out the Crop Configuration input to make sure it rendered correctly.~

Tested out the following scenarios to make sure that they failed with an error being raised:
- A crop rotation uses a crop configuration that does not exist.
- A crop configuration has an invalid combination of "crop_category" and "crop_type"
- Two crop configurations have the same "name"


RUFAS/data_structures/crop_soil_to_feed_storage_connection.py
```diff
@@ -88,6 +88,32 @@ class CropType(Enum):
     MEADOW_FESCUE = "Meadow Fescue"
 
 
+"""Mapping which defines the valid combinations of Crop Categories and Crop Types."""
+CROP_CATEGORY_TO_CROP_TYPE_MAPPING = {
```
Isgotkowitz, line 92: Moving this mapping outside HarvestedCrop makes it possible to determine if a user specified an invalid combination of crop category and type in a mapping.

RUFAS/routines/field/crop/crop_data.py
```diff
@@ -185,6 +185,14 @@ class CropData:
...
     max_root_depth : float, default 2000
         Maximum depth of roots in the soil (mm).
+    root_distribution_param_da: float
+        Empirical root distribution parameter d_a (mm).
+        Reference: Fan, Jianling, et al. "Root distribution by depth for temperate agricultural crops." Field Crops
+            Research 189 (2016): 68-74, table 1. Note that the value has been converted to mm.
+    root_distribution_param_c: float
+        Empirical root distribution parameter c (unitless).
+        Reference: Fan, Jianling, et al. "Root distribution by depth for temperate agricultural crops." Field Crops
+            Research 189 (2016): 68-74, table 1.
```
Isgotkowitz, line 195: I moved these values back into `CropData` because they differ between crop species, and keeping them in the same place as the other unique attributes of a crop allows `CropData` to be initialized with a single dictionary unpacking.

The alternative would be to keep these attributes in the Crop Management module, and have the Crop remove these values from the config dictionary and put them into crop management specifically, then unpack the rest of the values into CropData.

RUFAS/routines/field/crop/crop_configurations/__init__.py
Isgotkowitz, line 1: All of the data that was in `crop_configurations/` has been moved into the `default_crop_configurations.json`.

Isgotkowitz:
Note that the GitHub Actions are failing because the default input files are protected, and this PR modifies them.

RUFAS/routines/field/crop/crop_data.py
KFosterReed: I think this makes sense

Isgotkowitz:
Changed the Field Data Reporter to include the simulation day when reporting crop variables. This breaks the test suite, but makes it way easier to get outputs from the Crop module. Will either revert this change or update the unit test suite before merging.

Isgotkowitz:
Update: all daily variables reported out of the Crop and Soil module have the simulation day in their info map. The test suite for `FieldDataReporter` has been updated, so the entire test suite should be passing.

RUFAS/routines/field/crop/crop_data_factory.py
```diff
@@ -0,0 +1,200 @@
...
+
+        plant_category = PlantCategory(config["plant_category"])
+        storage_type = StorageType(config["storage_type"])
+
+        config["plant_category"] = plant_category
+        config["crop_category"] = crop_category
+        config["crop_type"] = crop_type
+        config["storage_type"] = storage_type
+
+        return CropConfiguration(**config)
```
Isgotkowitz: Mypy _does not_ like the dictionary unpacking operator (`**`). But in this case, I think the code is overall made much cleaner by it and this justifies its use.

KFosterReed (approved):
Using the debugger I looked to see if the changes in the crop data were correctly passed to the attributes of crop instances when I changed the in the default crop config files. I then made new crop config files and changed the following attributes: 
Test 1: 
Alfalfa silage - increase yield N fraction
Alfalfa hay - optimal harvest index increase
Corn grain - increase optimal temp
Corn silage - decrease max root depth


Test 2: 
Alfalfa hay - dry matter percent increase
Alfalfa silage - is_nitrogen_fixer - false
Corn grain - light use efficiency - 80
Corn silage- potential heat units decrease

They all appeared to have the expected impact on crop yield outcomes. This will be a really nice update for when we are ready to do the sensitivity analysis for the soil and crop module and will meet our stakeholder requests (ahead of schedule!!) to be able to simulate different crop species :) 

Well done Ed!

Isgotkowitz:
> Current Coverage: 95%
> 
> Mypy errors on 1831-user-crop-specifications branch: 3736 Mypy errors on dev branch: 3453 283 more errors on 1831-user-crop-specifications branch

Crud I will look into why the errors went up so much and why the coverage went down

Isgotkowitz:
Changes:
- Changed the Sample Crop Configuration used for testing from being a generic dictionary to being a `CropConfiguration` typed dict. This dramatically lowered the # of Mypy errors.
- Updated the test suites for Root Development and Field Manager to bring both back up to 100%.
- Updated the expected E2E testing results for C&S. E2E testing started failing, and I am reasonably confident that this is solely because "simulation_day" was added to the info maps of all daily output variables from `FieldDataReporter`.

In regards to the # of Mypy errors going up and coverage decreasing:
- I think the added Mypy errors are entirely due to the `CropData()` calls in based crop modules, because there are now some attributes in `CropData` without defaults. Instead of adding defaults for those attributes, I will write up an issue to remove all unnecessary/inappropriate defaults from `CropData` and have the `CropData()` calls removed from all crop process modules.
- Haven't done the math on the test coverage, but I think the decrease is because all the dataclasses that were removed from `crop_configurations/` had 100% test coverage so they were boosting the overall coverage. But now that they're removed total number of lines covered / total number of lines changes, even though the test suite is just as comprehensive as it was before.

ew3361zh:
>     * Updated the expected E2E testing results for C&S. E2E testing started failing, and I am reasonably confident that this is solely because "simulation_day" was added to the info maps of all daily output variables from `FieldDataReporter`.

I can confirm this - I pasted the expected results currently on dev and the differences generated between actual and expected on this feature branch were just the sim day added in the info maps (over and over and over and over again).

RUFAS/data_structures/crop_soil_to_feed_storage_connection.py
```diff
@@ -88,6 +88,32 @@ class CropType(Enum):
     MEADOW_FESCUE = "Meadow Fescue"
 
 
+"""Mapping which defines the valid combinations of Crop Categories and Crop Types."""
```
ew3361zh:
```suggestion
"""Maps the valid combinations of Crop Categories and Crop Types."""
```

RUFAS/routines/field/crop/crop_data_factory.py
```diff
@@ -0,0 +1,203 @@
...
+                }
+                err_name = "Duplicate crop configuration name."
+                err_msg = f"{name=} is used for more than one crop configuration."
+                cls._om.add_error(err_name, err_msg, info_map)
+                raise ValueError(f"{err_name} {err_msg}")
+            cls._crop_configurations[name] = crop_config
+
+    @classmethod
+    def _setup_crop_configuration(cls, config: dict[str, Any]) -> CropConfiguration:
+        """
```
ew3361zh: What do you think about renaming this to avoid confusion with the public method with the same name? Seems ripe for accidental confusion.

RUFAS/routines/field/crop/crop_data_factory.py
```diff
@@ -0,0 +1,203 @@
...
+    yield_phosphorus_fraction: float
+
+
+class CropDataFactory:
+    """
+    Manages and manufactures CropData instances using user-input crop configurations.
+
+    Attributes
+    ----------
+    _crop_configurations : dict[str, dict[str, CropConfiguration]]
```
ew3361zh:
```suggestion
    _crop_configurations : dict[str, CropConfiguration]
```
I believe it's correct in the attribute type hint but regardless they should match.

RUFAS/routines/field/crop/crop.py
```diff
@@ -387,77 +368,3 @@ def set_crop_planting_attributes(
...
-        ----------
-        **specs
-            an optional set of arguments, passed to CropData that customizes the crop species.
-
-        Details
-        -------
-        This method can be used to create a new ('unsupported') crop species/type.
-        """
-        crop_data = CropData(**specs)
-        return Crop(crop_data)
```
ew3361zh, line 463: Grateful these have moved on, I have done a few lengthy battles with them.

ew3361zh (approved):
Really nice work, Ed! Was passing (or correctly failing) the simulation testing scenarios I ran with it. Including adding a new custom crop fairly easily.

Left a few comments but nothing big that would prevent me from approving.

RUFAS/routines/field/crop/crop_data_factory.py
Isgotkowitz: Good catch

RUFAS/routines/field/crop/crop.py
Isgotkowitz: Me too

RUFAS/routines/field/crop/crop_data_factory.py
Isgotkowitz: I think that's a very good idea, will do.

Isgotkowitz: Gonna use "_manufacture_crop_configuration" because it's a factory class, and also "manufacture" is a fun verb.

Isgotkowitz:
> Really nice work, Ed! Was passing (or correctly failing) the simulation testing scenarios I ran with it. Including adding a new custom crop fairly easily.
> 
> Left a few comments but nothing big that would prevent me from approving.

@ew3361zh thank you for wading through all these changes, I really appreciate it.

PR #2180, opened 2025-01-24
[Documentation] Update references in project from Python 3.10 to 3.12

### LINKED ISSUE #2170: [Documentation] Python 3.12 Wiki/Welcome Message/E2E Testing Clean-Up
Following the closing of issues #2147 and #2148, there are some clean-up items to help transition to Python 3.12:

- [x] Upgrade workflow action to 3.12
- [ ] Find and share tutorials for upgrading to Python 3.12 with different IDEs (mainly PyCharm and VS Code)
- [x] Update setup.py
- [ ] Remove 3.10 (and any other Python version) references from Wiki and point to setup.py and requirements.txt
- [x] Add check to main.py to ensure minimum version 3.12 is being used by user and log warning/error if not.
- [x] Delete duplicate welcome message from start up

ew3361zh:
With the discovery that Python 3.12 is necessary for the most consistent and accurate outputs, it was necessary to update all references to ensure users have 3.12 installed.

### Context

Issue(s) closed by this pull request: closes #2170

### What

1. Updates Github actions to use Python 3.12
2. Migrates RUFAS from setting project metadata and build details in setup.py to the more modern tool `pyproject.toml` as defined by [PEP518](https://peps.python.org/pep-0518/) and [PEP621](https://peps.python.org/pep-0621/). 
3. Adds check to ensure user has at least `requires-python` version set in `pyproject.toml` installed and logs an error if they are using an older Python version.
4. Removes duplicate welcome message at startup.

Tasks to complete right before this is merged into dev:

- [x] Find and share tutorials for upgrading to Python 3.12 with different IDEs (mainly PyCharm and VS Code)
- [x] Update Wiki with pyproject.toml info (and remove all refs to setup.py) - especially in onboarding section

### Why

Currently at least Python 3.12 is necessary for getting consistent and accurate results from the RUFAS model.

### How

1. All references in GH actions to python 3.10 were searched for and updated to 3.12.
2. setup.py was removed and the data stored there was migrated to the existing pyproject.toml file.
3. Functionality was added to task_manager.py to get the user's python version and to check that against the minimum version required from the pyproject.toml file. **_Note: to load in data from a `.toml` file, we need to use the `tomllib` library. This is standard with Python 3.11 and higher. Prior to 3.11, a separate, non-standard `.toml` library is needed. Rather than add this as a separate dependency only needed by users prior to 3.11, I added a check and logged the error automatically if they are using a version older than 3.11._**
4. Printing credits/welcome was separated from printing the task_id in OM which allows us not to print the welcome message twice.

### Test plan

To test that the Python version check was working properly:
- Run it with Python version 3.12 or higher and look at the messages in the terminal.
- Download a version earlier than 3.11 (I used 3.10.9) and create a separate virtual environment specifically pointing to that version of Python (my command on a Mac was `python3.10 -m venv env_310`) and then go into that new env, install requirements.txt, and run the simulation again. You should see the error message warning people they may get unreliable results.


RUFAS/task_manager.py
```diff
@@ -111,14 +116,13 @@ def start(
             save_chunk_threshold_call_count=0,
             variables_file_path=Path(""),
             output_prefix="Task Manager",
-            version_number=RUFAS_VERSION,
             task_id="TASK MANAGER",
             is_end_to_end_testing_run=False,
         )
+        self.check_python_version()
```
ew3361zh, line 122: This needs to be called after `OM.run_startup_sequence()` to allow potential errors to print to terminal.

RUFAS/task_manager.py
```diff
@@ -180,6 +184,51 @@ def start(
...
+        Returns the current version of RUFAS.
+
+        Returns
+        -------
+        str
+            Version of RUFAS or "Unknown" if the person is using a Python version earlier than 3.11.
+        """
+        user_python_version = Version(f"{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}")
+        if user_python_version < Version("3.11"):
+            return "Unknown"
```
ew3361zh: Any better ideas for what to return here as the RUFAS version if the user has an older version of Python? We can't load the .toml file to get the most current version of RUFAS if they're on Python < 3.11 and I don't think we want to add another place to have to keep the RUFAS version number. I guess we could potentially get it from the changelog.md but that seems like a second source of truth that's more for internal uses anyway.

RUFAS/task_manager.py
```diff
@@ -180,6 +184,51 @@ def start(
...
+        with open(pyproject_file_path, "rb") as pyproject_file:
+            import tomllib
+
+            rufas_version = tomllib.load(pyproject_file)["project"]["version"]
```
ew3361zh: tomllib is only for Python >= 3.11 so importing this to check for what Python version RUFAS requires if they're not on 3.11 or later will cause an error.

RUFAS/task_manager.py
```diff
@@ -180,6 +184,51 @@ def start(
...
+        return str(rufas_version)
+
+    def check_python_version(self) -> None:
+        """
+        Checks if the Python version meets minimum version set in pyproject.toml.
+        """
+        user_python_version = Version(f"{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}")
```
ew3361zh, line 216: Don't necessarily need the `.micro` version info since we're only checking to the minor level but I added it here because it shouldn't cause any issues and would allow us to be more specific in the future with version checking if we wanted.

RUFAS/task_manager.py
```diff
@@ -180,6 +184,51 @@ def start(
...
+        if user_python_version < Version("3.11"):
+            self.output_manager.add_error(
+                "Python version",
+                f"User's Python version {user_python_version} is less than the minimum required version (3.11). "
+                "Results may not be accurate.",
+                {"class": TaskManager.__name__, "function": TaskManager.check_python_version.__name__},
+            )
+        else:
+            import tomllib
```
ew3361zh, line 218: Same as above, if we import tomllib globally and someone on using a version of Python earlier than 3.11 runs RUFAS, it will crash. There is an earlier non-native .toml file reading library but to add it would require an additional dependency which seems like it's not necessarily worth the cost since we know RUFAS needs at least 3.12.

tests/test_time.py
```diff
@@ -181,7 +181,7 @@ def test_str(mock_config: Dict[str, Any], mocker: MockerFixture) -> None:
     time = Time()
 
     for n in range(364):
-        assert time.__str__() == f"Year: {1}, Day: {2+n}. Simulation Day: {n}"
+        assert time.__str__() == f"Year: {1}, Day: {2 + n}. Simulation Day: {n}"
```
ew3361zh: marked as a flake8 error with the update to the latest flake8.

RUFAS/task_manager.py
Isgotkowitz: Maybe import tomllib in a try-except block, and if a module import error is raised then log an error to the user telling them their version of python is likely behind? That was just the first thing that came to my mind, but it really doesn't do much to help folks who are running a Python version older than 3.10.

RUFAS/task_manager.py
```diff
@@ -20,7 +23,7 @@
 from RUFAS.units import MeasurementUnits
 from RUFAS.util import Utility
 
-RUFAS_VERSION = "0.9.2"
+pyproject_file_path = Path("pyproject.toml")
```
Isgotkowitz:
```suggestion
PYPROJECT_FILE_PATH = Path("pyproject.toml")
```

RUFAS/task_manager.py
Isgotkowitz: I think you're right about not needing to check the micro version. But maybe it would be a good idea to log what the version is, regardless of whether or not it works for RuFaS? It might help to have it logged in case there is a change in behavior between micro version of 3.12

pyproject.toml
```diff
@@ -14,4 +28,8 @@ disallow_incomplete_defs = true # Require all variables to be explicitly typed
 show_column_numbers = true # Show column numbers in error messages
 strict = true
 exclude = ['^(?!.*(?:RUFAS|tests|main\.py|setup\.py)).*$']
-follow_imports = "silent"
\ No newline at end of file
+follow_imports = "silent"
+
+[tool.pytest.ini_options]
+addopts = "--strict-markers"
+testpaths = ["tests"]
```
Isgotkowitz: What do these do?

pyproject.toml
```diff
@@ -1,10 +1,24 @@
...
+name = "RUFAS"
+version = "0.9.2"
+description = "Ruminant Farm Systems Model"
+readme = "README.md"
+authors = [
+{"name" = "Kristan Foster Reed", "email" = "Kristan.Reed@usda.gov"},
+]
+requires-python = ">=3.12"
+
+
```
Isgotkowitz:
```suggestion

[project.urls]
"Homepage = "https://rufas.org"
"Source" = "https://github.com/RuminantFarmSystems/MASM"
```
I'm not 100% sure this is the right way to add urls, but I think it would be good to include them

RUFAS/task_manager.py
```diff
@@ -180,6 +184,51 @@ def start(
...
+
+            rufas_version = tomllib.load(pyproject_file)["project"]["version"]
+        return str(rufas_version)
+
+    def check_python_version(self) -> None:
+        """
+        Checks if the Python version meets minimum version set in pyproject.toml.
+        """
+        user_python_version = Version(f"{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}")
+        if user_python_version < Version("3.11"):
```
Isgotkowitz:
```suggestion
        if user_python_version <= Version("3.11"):
```
People need to be on 3.12 at least right?

Also, might be good to make "3.11" or Version("3.11") a module constant

Isgotkowitz (commented):
I think this is an awesome step forward, thank you for doing all this research on versioning and build systems Niko!

I left suggestions on a couple things but all are take it or leave. The only thing holding me back from approving is that I got different error messages with different unsupported versions of python. When running 3.10, I got:
`[28-Jan-2025_Tue_11-44-55.054910][ERROR][Task Manager] Python version. User's Python version 3.10.16 is less than the minimum required version (3.11). Results may not be accurate.` But with 3.11 the message was `[28-Jan-2025_Tue_11-42-18.909273][ERROR][Task Manager] Python version. User's Python version 3.11.11 is less than the minimum required version >=3.12. Results may not be accurate.`

> Find and share tutorials for upgrading to Python 3.12 with different IDEs (mainly PyCharm and VS Code)

I think upgrading the version of Python installed is an OS issue, not an IDE one. Which isn't a critical distinction, but I think it's worth keeping in mind.

Also, one thing that I am curious about: is it decided that all Python versions 3.12 and above are kosher, or _just_ 3.12? I tried with 3.13 and it passed E2E testing, so it is probably alright. But supporting all versions 3.12 and up is still a bigger commitment than just 3.12, or just 3.12 and 3.13.

RUFAS/task_manager.py
ew3361zh: I like the try-except block but none of the logging is sent to the terminal until OM.run_startup_sequence() is executed which happens after the credits are printed which this is needed for. We could print the credits AFTER the startup sequence is run I guess?

RUFAS/task_manager.py
ew3361zh: Yea sorry this is confusing but this if statement is to catch a bad version that we can't catch by checking against the pyproject.toml file setting which we can't check without the tomllib library which isn't available in python earlier than 3.11.

The message in the add_error should be corrected to have the minimum required version be 3.12 though, good catch.

allisterakun (commented):
It works as expected.
After some experimentations with different Python versions, I finally understood the following:
1. We will first check if the python version is >= 3.11 to make sure that we can read the toml file. 
    1. if not, the rufas version will be "Unknown" and an error is generated to inform the user the potential inaccurate result of the simulation and suggest the user to update python. 
2. Once we can read the toml file, we will get the required python version defined in the toml file (which is currently 3.12), and we will check against that.
    1. If the user python version is less than the required python version defined in the toml file, an error will be generated, with the same error message (only difference is the required version).

After trying other alternatives, which would require other third-party packages, and discussing with @ew3361zh, we agreed that the current implementation is the best solution at this moment.

Here are some minor suggestions:
1. For the first check (3.11), please clarify the error message to indicate that we are requiring Python 3.11 so that we can read the toml file. It'd be nice to differentiate the two error messages and give more useful information.
2. Consider making 3.11 a file constant instead of having it hard-coded inline.
3. The check for 3.11 (so that we can read the toml file) happens twice, in `get_rufas_version()` and `check_python_version()`. Please consider if it's possible to only check this information once (possibly loading the toml file and save it as local variable or something)

pyproject.toml
ew3361zh: the testpaths one is to direct pytest where the test files are found. I think it's supposed to speed up the testing process by not having pytest search the entire directory for test files.

The other one might have been an autocomplete with copilot because I don't recall adding it and when I just looked it up it seems like something we don't really encounter or need to worry about right now (preventing accidental typos in pytest decorator names???). I will remove it.

pyproject.toml
ew3361zh: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#urls

ew3361zh:
Thanks, Allister. Some responses to your fb:
>     1. For the first check (3.11), please clarify the error message to indicate that we are requiring Python 3.11 so that we can read the toml file. It'd be nice to differentiate the two error messages and give more useful information.

I went with 3.12 since that is the requirement to run python. After thinking about it I'm not sure people need to know the difference of whether they can open the .toml file because they're on 3.11. I think the key takeaway is that we need them to use at least 3.12.

>     2. Consider making 3.11 a file constant instead of having it hard-coded inline.

Done! May be subject to change after Pooya reviews but I think it's ok to try for now.

>     3. The check for 3.11 (so that we can read the toml file) happens twice, in `get_rufas_version()` and `check_python_version()`. Please consider if it's possible to only check this information once (possibly loading the toml file and save it as local variable or something)

I added the try-except block Ed suggested. Let me know if that works for you.

ew3361zh:
> I left suggestions on a couple things but all are take it or leave. The only thing holding me back from approving is that I got different error messages with different unsupported versions of python. When running 3.10, I got: `[28-Jan-2025_Tue_11-44-55.054910][ERROR][Task Manager] Python version. User's Python version 3.10.16 is less than the minimum required version (3.11). Results may not be accurate.` But with 3.11 the message was `[28-Jan-2025_Tue_11-42-18.909273][ERROR][Task Manager] Python version. User's Python version 3.11.11 is less than the minimum required version >=3.12. Results may not be accurate.`

Fixed this.

> > Find and share tutorials for upgrading to Python 3.12 with different IDEs (mainly PyCharm and VS Code)
> 
> I think upgrading the version of Python installed is an OS issue, not an IDE one. Which isn't a critical distinction, but I think it's worth keeping in mind.

Yea it's kind of three-fold: where to get 3.12, how to install it properly, and then how to switch your IDE's interpreter to be using that version to run RUFAS.

> Also, one thing that I am curious about: is it decided that all Python versions 3.12 and above are kosher, or _just_ 3.12? I tried with 3.13 and it passed E2E testing, so it is probably alright. But supporting all versions 3.12 and up is still a bigger commitment than just 3.12, or just 3.12 and 3.13.

That's a good point. I can make it 3.12, 3.13 in the pyproject file and then update the messaging accordingly.

ew3361zh:
Updates:
- After discussion with Pooya, we now raise an error if the user is not on a version of Python that RUFAS supports.
- He did want it to be easy enough to still comment that check out if the user wanted to run it on an older version of Python and knew what they were doing so currently a user could just comment out the call to check_python_version() in task_manager.start() and the rest of the model would run.
- Testing suite updated

RUFAS/task_manager.py
```diff
@@ -180,6 +185,60 @@ def start(
...
+            specifier = SpecifierSet(requires_python)
+            if user_python_version not in specifier:
+                raise RuntimeError(
+                    f"RUFAS requires Python {requires_python}, but you are using Python"
+                    f"{user_python_version}. Please upgrade or downgrade your Python version to "
+                    "match the required version range."
+                )
+        except ImportError:
+            raise RuntimeError(
+                f"RUFAS requires Python {str(MINIMUM_PYTHON_VERSION)} or later. " "Please upgrade your Python version."
```
Isgotkowitz:
```suggestion
                f"RUFAS requires Python {str(MINIMUM_PYTHON_VERSION)} or later. Please upgrade your Python version."
```

Isgotkowitz: I suggest unifying these error messages if possible. Maybe make it more generic with an instruction to look in `pyproject.toml` for the acceptable range of Python versions.

RUFAS/task_manager.py
```diff
@@ -20,7 +23,8 @@
 from RUFAS.units import MeasurementUnits
 from RUFAS.util import Utility
 
-RUFAS_VERSION = "0.9.2"
+PYPROJECT_FILE_PATH = Path("pyproject.toml")
+MINIMUM_PYTHON_VERSION = Version("3.12")
```
Isgotkowitz, line 27: I like the idea of hardcoding this here, just in case Python can't open the toml file it is nice to be specific about what the minimum version is in the error message. But since this is duplicating information from `pyproject.toml`, maybe it would be good for TaskManager to check that `MINIMUM_PYTHON_VERSION` always matches what is in `pyproject.toml` to prevent the very rare case that they get out of sync?

Isgotkowitz (commented):
Unfortunately when I tried this with Python 3.13 it didn't work, with this error message:
```
    "values": [
      "This terminal error occurred during runtime. Traceback (most recent call last):\n  File \"/home/isgotkowitz/rufas/RUFAS/task_manager.py\", line 224, in check_python_version\n    raise RuntimeError(\n    ...<3 lines>...\n    )\nRuntimeError: RUFAS requires Python >=3.12, <=3.13, but you are using Python3.13.1. Please upgrade or downgrade your Python version to match the required version range.\n\nDuring handling of the above exception, another exception occurred:\n\nTraceback (most recent call last):\n  File \"/home/isgotkowitz/rufas/main.py\", line 23, in main\n    task_manager.start(\n    ~~~~~~~~~~~~~~~~~~^\n        metadata_path=Path(cmd_arguments.path_to_metadata),\n        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n    ...<7 lines>...\n        metadata_depth_limit=cmd_arguments.metadata_depth_limit,\n        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n    )\n    ^\n  File \"/home/isgotkowitz/rufas/RUFAS/task_manager.py\", line 106, in start\n    self.check_python_version()\n    ~~~~~~~~~~~~~~~~~~~~~~~~~^^\n  File \"/home/isgotkowitz/rufas/RUFAS/task_manager.py\", line 240, in check_python_version\n    raise RuntimeError(f\"An unexpected error occurred while checking the Python version: {e}\")\nRuntimeError: An unexpected error occurred while checking the Python version: RUFAS requires Python >=3.12, <=3.13, but you are using Python3.13.1. Please upgrade or downgrade your Python version to match the required version range.\n"
    ]
```

Something I noticed is that when an unsupported Python version is used and a RunTime error is raised, nothing is printed to stdout. This is very wishful, so it may be more appropriate for a separate PR, but could that error get printed? Having this error printed to the errors file is good, but I can imagine that it might take someone a while to dig through files and directories to find it, instead of seeing it instantly in their terminal

ew3361zh:
> Unfortunately when I tried this with Python 3.13 it didn't work, with this error message:
> 
> ```
>     "values": [
>       "This terminal error occurred during runtime. Traceback (most recent call last):\n  File \"/home/isgotkowitz/rufas/RUFAS/task_manager.py\", line 224, in check_python_version\n    raise RuntimeError(\n    ...<3 lines>...\n    )\nRuntimeError: RUFAS requires Python >=3.12, <=3.13, but you are using Python3.13.1. Please upgrade or downgrade your Python version to match the required version range.\n\nDuring handling of the above exception, another exception occurred:\n\nTraceback (most recent call last):\n  File \"/home/isgotkowitz/rufas/main.py\", line 23, in main\n    task_manager.start(\n    ~~~~~~~~~~~~~~~~~~^\n        metadata_path=Path(cmd_arguments.path_to_metadata),\n        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n    ...<7 lines>...\n        metadata_depth_limit=cmd_arguments.metadata_depth_limit,\n        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n    )\n    ^\n  File \"/home/isgotkowitz/rufas/RUFAS/task_manager.py\", line 106, in start\n    self.check_python_version()\n    ~~~~~~~~~~~~~~~~~~~~~~~~~^^\n  File \"/home/isgotkowitz/rufas/RUFAS/task_manager.py\", line 240, in check_python_version\n    raise RuntimeError(f\"An unexpected error occurred while checking the Python version: {e}\")\nRuntimeError: An unexpected error occurred while checking the Python version: RUFAS requires Python >=3.12, <=3.13, but you are using Python3.13.1. Please upgrade or downgrade your Python version to match the required version range.\n"
>     ]
> ```

Updated the pyproject requires-python range to be `">=3.12, <3.14"` which caught 3.13.1 on my machine. Good catch!

> Something I noticed is that when an unsupported Python version is used and a RunTime error is raised, nothing is printed to stdout. This is very wishful, so it may be more appropriate for a separate PR, but could that error get printed? Having this error printed to the errors file is good, but I can imagine that it might take someone a while to dig through files and directories to find it, instead of seeing it instantly in their terminal

Definitely not difficult to do. I had this as the setup at one point - we just have to have OM run its startup sequence before we check a user's python version and the RUFAS version.

RUFAS/task_manager.py
ew3361zh: We can and I think I originally tried that but it became difficult to determine which specific error was occurring and what the user should specifically do to address the issue. Because this is such an important part of getting things to run, I think leaving the error messages as specific as possible would be the best route.

RUFAS/task_manager.py
ew3361zh: Great idea! I will work on something for this.

ew3361zh:
Updates:
- Moved `check_rufas_version()` and `check_python_version()` to come after TM's OM runs its startup_sequence which will enable any errors/warnings/logs logged in those functions to print out in the terminal in addition to being logged.
- Added a check to ensure the `MINIMUM_PYTHON_VERSION` hard-coded in TM matches the version minimum in the pyproject.toml file.

Isgotkowitz (approved):
Works as expected! Clear error messages when an invalid Python version is used, and when there is a mismatch between the minimum version in the Task Manager and `pyproject.toml`.

Also I like the addition of the testpaths for pytest, very convenient.

Great work Niko!

allisterakun (approved):
LGTM! Good job, Niko!

RUFAS/task_manager.py
```diff
@@ -181,6 +186,70 @@ def start(
...
         )
 
+    def get_rufas_version(self) -> str:
+        """
+        Returns the current version of RUFAS.
+
+        Returns
+        -------
+        str
+            Version of RUFAS or "Unknown" if the person is using a Python version earlier than 3.11.
```
PooyaHekmati:
```suggestion
            Version of RUFAS or "Unknown" if the version of Python is less than 3.11.
```

RUFAS/task_manager.py
```diff
@@ -181,6 +186,70 @@ def start(
...
+        """
+        try:
+            with open(PYPROJECT_FILE_PATH, "rb") as pyproject_file:
```
PooyaHekmati, line 199: Does it have to be binary?
```suggestion
            with open(PYPROJECT_FILE_PATH, "r") as pyproject_file:
```

RUFAS/task_manager.py
```diff
@@ -181,6 +186,70 @@ def start(
...
+                import tomllib
+
+                rufas_version = tomllib.load(pyproject_file)["project"]["version"]
+        except Exception as e:
+            self.output_manager.add_error(
+                "Error reading RUFAS version",
+                f"Unable to read RUFAS version from pyproject.toml file. {e}",
+                {"class": TaskManager.__name__, "function": TaskManager.get_rufas_version.__name__},
```
PooyaHekmati: Maybe? :)
```suggestion
                {"class": self.__class__.__name__, "function": self.get_rufas_version.__name__},
```

PooyaHekmati (approved):

RUFAS/task_manager.py
ew3361zh: It does. The tomllib load function expects a binary file (which is different from the third-party toml library that expects a normal text-mode file).

RUFAS/task_manager.py
ew3361zh: Good catch. I think I may have had this as a static function at some point in the process.

PR #2189, opened 2025-01-29
[Testing] Regenerate e2e expected results

### LINKED ISSUE #2140: [E2E Testing] Add feature to re-generate E2E testing filters
A feature should be added to assist in the maintenance of E2E testing filters

## Status

E2E testing requires two specialized types of filters to run an E2E test on RuFaS. One contains the filter that collects the actual results and holds the expected results. The other has a filters which collects the differences between the actual and expected results.

In order to update the expected results, the new expected results have to be copied and pasted manually into a JSON file that contains the expected results. This is a finicky and cumbersome process.

## Intention

Make the process of updating E2E results partially automated, so that updating the E2E testing process is less time and resource intensive.

## Request

Add a task type which will update the expected results in E2E testing filters.

### Additional Information

It is very important that there is still some degree of HIL (Human In the Loop) when updating the expected results of E2E testing. If there isn't, it is very likely that the E2E expected results will get updated without anyone actually verifying that they are valid. And if the expected results are invalid, it defeats the entire point of E2E testing.

ew3361zh:
Sets up a new task to automatically update E2E testing expected results if there are changes found from actual results.

### Context

Issue(s) closed by this pull request: closes #2140

### What

Updates the previous E2ETestResultsComparer to be E2ETestResultsHandler keeping the previous functionality but adding the ability to generate new e2e results, compare them to the existing expected results, and if they're different, update the existing expected results with the new actual results.

This is set up in a new task that will automatically generate a new set of expected results for each domain (c&s, animal, feed storage) and update the corresponding e2e domain filter files if the results differ from what's in the `expected_results` section of that filter file.

These files are located in input > data > end_to_end_testing and are named in this format: `e2e_json_(domain)_filter.json`

The newly produced expected results file will have an invalid string line written across the top that a user generating these new files will have to remove by hand. This will keep some element of human in the loop for these files being updated/regenerated.

I ran this on the 3 existing filter files to have the most up-to-date results which also include the last-updated date in the filter file (none of the `expected_results` sections themselves are different).

### Why

Makes the process of updating E2E results partially automated, so that updating the E2E testing process is less time and resource intensive.

### How

1. Renames E2ETestResultsComparer to be E2ETestResultsHandler and updates all references in the model.
2. Adds a new task in TaskManager that runs a simulation in similar fashion to the e2e testing task, and then triggers actions in the E2ETestResultsHandler to evaluate these new results against the existing expected results.
3. Adds 2 new key functions to this class: 
- `update_expected_test_results()` which checks the actual results generated from the simulation against the expected results in the filter file for each domain. If the results are different (checked with deepdiff), the new actual results will overwrite the existing expected results in the filter file, and the date of the last updated expected results will be updated.
- `_write_formatted_json()` which takes the new actual results intended to overwrite the existing expected results and formats them properly (minifying them) while still maintaining the normal structure for the rest of the file. It will also write a line that violates json format across the top of the updated filter file to ensure there is HIL on updating these files.
4. New input/metadata files added to allow easy task execution:
- `update_end_to_end_testing_tm_metadata.json` - pointer to the new task
- `update_end_to_end_expected_results.json` - the new task file
5. When the filter files are initially opened to be able to extract the `expected_results`, there is a backup of the file as it is right then. If there are no differences found or there's an error in trying to load the json data or save the new file, the backup file will be there to be available so you won't lose everything.

### Test plan

Running this task will evaluate the current expected_results in each of the e2e filter files. To see it in action:
1. Change something in the `expected_results` value section of 1 or 2 of the e2e_json_domain_filter.json files in the input > data > end_to_end_testing dir. You can change all 3 but by just changing 1 or 2, you can see how this task runs without changing one of the files as well since there are different results for each. For example in one of the files I even just deleted a word from the disclaimer:
``` 
"expected_results": {"DISCLAIMER":"Under construction, use the results with caution."...}
```
-->
```
"expected_results": {"DISCLAIMER":", use the results with caution."...}
```
2. Run the new update e2e results metadata task with this command: `python main.py -p input/metadata/update_end_to_end_testing_tm_metadata.json -c`
3. Watch the logs towards the end where the existing filter files' expected results are being evaluated.
4. Once the process is finished, go in and examine the changed filter files with expected results. **Don't forget to remove the line at the top of any newly updated files!**
5. Run e2e testing with your newly updated files to make sure they work: `python main.py -p input/metadata/end_to_end_testing_tm_metadata.json -c`


RUFAS/e2e_test_results_handler.py
```diff
@@ -0,0 +1,322 @@
...
+        ----------
+        file_path : Path
+            The path to the JSON file.
+        data : dict
+            The data to write to the JSON file.
+        """
+        key_order = ORDERED_EXPECTED_RESULTS_FILE_KEYS
+        missing_keys = [key for key in key_order if key not in data]
+        if missing_keys:
+            raise ValueError(f"Missing required keys in data: {missing_keys}")
```
matthew7838: Should we also log error to OM?

matthew7838 (approved):
Excellent work, Niko. I've tested it, and it's working as expected on my side, Please address the mypy errors before merging.

allisterakun (approved):
Great job, Niko! Tested on my Windows PC with both Python 3.11 and 3.12
E2E testing would get stuck at animal test result comparison for a while and eventually fail for crop & soil and the animal modules when running Python 3.11, as expected. 
E2E testing passes without issues for all the modules when running Python 3.12. Also tried modifying the expected results as suggested in the PR description.

PR #2191, opened 2025-01-29
[Documentation] Regenerate Sphinx documentation structure

Isgotkowitz:
Checks the formatting of the Sphinx documentation and makes changes to address its structure.

### Context

Issue(s) closed by this pull request: None, this was a Fixathon task.

### What

The Fixathon task was a pretty open-ended issue that was just about making sure the formatting of the documentation generated from docstrings was correct, but not about checking that the content of the documentation was correct.

I looked through the documentation that was generated, and while it didn't look as pretty as the documentation of Numpy or Numba, I didn't spot any formatting issues. However, what I noticed is that there were a number of RuFaS modules entirely missing when the html documentation was regenerated and some modules which had been removed from the code a long time ago were still listed.

### Why

I believe the idea behind the fixathon task was to make our documentation look good, and ensure that it was in a publishable state. I'm not sure if the extra and missing modules falls into the formatting category or the checking correctness category, but regardless it was a very glaring issue so it felt appropriate to fix it.

### How

When `make html` is run in the `docs/` directory, it regenerates the documentation of modules based on the contents of docstrings in the code. BUT the modules that are included, and how modules are linked to submodules in the documentation, is determined by the `.rst` files in the `docs/` directory. These `.rst` files are _not_ updated by the Makefile, so it is up to us to manually edit them to keep them up-to-date or ...

Use `sphinx-apidoc`. This is a Sphinx extension that automates the process of updating the .rst files, and was at one point incorporated into the Sphinx GitHub Action by @ew3361zh, although I forget why it was removed.

To regenerate the `.rst` files, I ran the following command from the root directory of the project:
`sphinx-apidoc -f -o docs/ . tests --maxdepth=1 --separate`
Let's breakdown what is going on here:
- `-f`: "force"; this command tells sphinx-apidoc to overwrite existing .rst files based on what it finds in the codebase.
- `-o docs/`: tells sphinx-apidoc where to write the new .rst files to.
- `.`: path to the modules and packages that we want .rst files generated/regenerated for. In this case, it is everything in the project's root directory.
- `tests`: path or paths that we _do not_ want .rst files generated/regenerated for. My guess is that we don't need or want the tests documented with Sphinx.
- `--maxdepth=1`: when the subpackages and submodules are listed, this determines whether their subpackages and submodules are also listed in the page. I felt a maxdepth of 1 made for the best looking documentation, but this is just a personal preference.
- `--separate`: instead of listing all the documentation of modules from the same directory next to each other, this arg gives each module its own page for its documentation. Again, I felt that this made for much better looking documentation, but that is a personal opinion.

**Note there are still some modules missing from the .rst files!!**. But this is because there is no `__init__.py` file in the directory where they are located. My two cents is that it would be better to wait until we get the current refreshes finished, make sure every Python directory we want in the docs has an `__init__.py` in it, then re-run the above `sphinx-apidoc` command.

### Test plan

Ran the Sphinx command above, moved to the `docs` directory, ran `make clean && make html`, then clicked through the newly generated html documentation.

I recommend reviewers do the same (the make command will be different on Windows), and to experiment with the `-f`, `--maxdepth`, and `--separate` commands to see how they change the documentation.


Isgotkowitz:
Best command so far:

`sphinx-apidoc -f -o docs/ . tests --maxdepth=1 --separate`

docs/modules.rst
```diff
@@ -1,11 +1,9 @@
-GitHub
-======
+RuFaS
+=====
```
Isgotkowitz, line 2: I don't think there is an easy way to set this to automatically be "RuFaS". I think the best way to get around this would be to write a script that executes `sphinx-apidoc` with a consistent set of arguments, then overwrites this line.

ew3361zh (approved):
Maybe I'm doing something wrong because the different options in the commands all seem like they're generating the same-looking documentation (all of which looks fine to me - a couple of dataclasses like ReceptionPitDailyOutput look a little cramped and poorly formatted in their inits but nothing egregious).

I agree with waiting until we've done the refreshes to add an init.py file to the places where we're missing documentation. I'm fine to approve as this is right now.

Isgotkowitz:
> Maybe I'm doing something wrong because the different options in the commands all seem like they're generating the same-looking documentation

Did you regenerate the HTML docs each time you regenerated the .rst files?

ew3361zh:
> > Maybe I'm doing something wrong because the different options in the commands all seem like they're generating the same-looking documentation
> 
> Did you regenerate the HTML docs each time you regenerated the .rst files?

I think so? Do you have screenshots of the differences you were seeing or could you describe what differences you were seeing? I could've just been looking too macro or too micro a level of detail.

Isgotkowitz:
> > > Maybe I'm doing something wrong because the different options in the commands all seem like they're generating the same-looking documentation
> > 
> > 
> > Did you regenerate the HTML docs each time you regenerated the .rst files?
> 
> I think so? Do you have screenshots of the differences you were seeing or could you describe what differences you were seeing? I could've just been looking too macro or too micro a level of detail.

Regenerated HTML docs with the .rst files in this PR:
image

.rst files regenerated with `maxdepth` set to 2:
image

.rst files regenerated without the `separate` argument:
image

Isgotkowitz:
Update: removed some `.rst` files for modules that have been deleted from the codebase and weren't overwritten, including:
- `RUFAS.routines.field.crop.crop_configurations.rst`
- `RUFAS.routines.manure.default_enum.rst`
- `setup.rst`

`modules.rst` was also updated because the setup module is no longer listed in it

matthew7838 (approved):
LGTM, used all suggested comments, and things are matching and looking much better. I think some of the content in the PR description should also be incorporated into the wiki.

Isgotkowitz:
> I think some of the content in the PR description should also be incorporated into the wiki.

Great idea, it should definitely go into the wiki

Isgotkowitz:
Unfortunately I was not able to get the module path removed from the links to different documentation pages. I tried using a custom template, but wasn't able to get it to work.

PR #2194, opened 2025-01-29
[E2E] Convert variable names

### LINKED ISSUE #2193: [E2E] Convert Variable Name
In `E2ETestResultsComparer`, create a function that reads a CSV, which contains the **original** and the **new** name of variables. This function should convert the variable names in the expected results, which are the original variable names, to the corresponding new name of the variable, so that the variable names match the actual results.

allisterakun:
Create the functionality to convert the variable names of a previous run results to match the new variable names of the new/modified run results according to a CSV lookup table.

### Context

Issue(s) closed by this pull request: closes #2193 

### What

In E2ETestResultsComparer, created a function that reads a CSV, which contains the original and the new names of variables. This function converts the variable names in the expected results, which are the original variable names, to the corresponding new name of the variable so that the variable names match the actual results.

### Why

To compare results if the output variable names have changed.

### Test plan

1. Change one of the variable names in any of the e2e testing expected results. I changed the first variable, `DISCLAIMER`, in `input/data/end_to_end_testing/e2e_json_animal_filter.json` to `DISCLAIME`.
2. Create a lookup CSV `input/var_name_conversion.csv`:

| Original| New |
| -------- | ------- |
| DISCLAIME | DISCLAIMER |

4. Add the above file path to `input/data/tasks/end_to_end_testing_task.json`:
```json
{
  "parallel_workers": 4,
  "tasks": [
    {
      "task_type": "END_TO_END_TESTING",
      "metadata_file_path": "input/metadata/end_to_end_testing_metadata.json",
      "output_prefix": "end-to-end-testing",
      "filters_directory": "input/data/end_to_end_testing",
      "log_verbosity": "logs",
      "exclude_info_maps": false,
      "random_seed": 42,
      "convert_variable_table_path": "input/var_name_conversion.csv"
    }
  ]
}
```
5. Run a normal e2e testing. It should finish without failing.
6. Also, I have tried to change the value `"convert_variable_table_path": "input/var_name_conversion.csv"` to a string that does not ends in ".csv". This would automatically be fixed to "", which would skip this conversion process.


RUFAS/e2e_test_results_handler.py
```diff
@@ -23,15 +24,21 @@ class E2ETestResultsHandler:
...
-        json_output_directory : Path
+        json_output_path : Path
             Path to which JSON outputs are written to.
+        convert_variable_table_path : Path
+            Path to the csv look up table to convert the variable names in the expected results to match the variable
+            names in the actual results.
 
+        Returns
+        -------
+        None
```
ew3361zh:
```suggestion
```

RUFAS/e2e_test_results_handler.py
```diff
@@ -86,6 +97,21 @@ def compare_actual_and_expected_test_results(json_output_path: Path) -> None:
             for comparison_type, difference in filtered_diff.items():
                 om.add_variable(comparison_type, difference, info_map)
 
+    @staticmethod
+    def _convert_expected_result_variable_names(
+        expected_results: dict[str, Any], conversion_csv_path: Path
+    ) -> dict[str, Any]:
+        converted_expected_results: dict[str, Any] = {}
```
ew3361zh, line 141: Add docstrings

RUFAS/e2e_test_results_handler.py
```diff
@@ -86,6 +97,21 @@ def compare_actual_and_expected_test_results(json_output_path: Path) -> None:
...
+        with open(conversion_csv_path, "r") as conversion_lookup_csv:
+            conversion_lookup_table: pd.DataFrame = pd.read_csv(conversion_lookup_csv, index_col=None)
+        for key, value in expected_results.items():
+            if key in list(conversion_lookup_table["Original"]):
+                new_key = conversion_lookup_table.loc[conversion_lookup_table["Original"] == key, "New"].iloc[0]
```
ew3361zh: I think because this dataset could be fairly large, it would be more efficient to convert the df into a dictionary since we're iterating over the whole thing.

RUFAS/e2e_test_results_handler.py
```diff
@@ -86,6 +97,21 @@ def compare_actual_and_expected_test_results(json_output_path: Path) -> None:
             for comparison_type, difference in filtered_diff.items():
                 om.add_variable(comparison_type, difference, info_map)
 
+    @staticmethod
+    def _convert_expected_result_variable_names(
+        expected_results: dict[str, Any], conversion_csv_path: Path
+    ) -> dict[str, Any]:
+        converted_expected_results: dict[str, Any] = {}
+        with open(conversion_csv_path, "r") as conversion_lookup_csv:
+            conversion_lookup_table: pd.DataFrame = pd.read_csv(conversion_lookup_csv, index_col=None)
```
ew3361zh: since this is reliant on specific formatting with specific columns, what do you think about adding some error handling to ensure the correct formatting?

RUFAS/e2e_test_results_handler.py
```diff
@@ -86,6 +97,21 @@ def compare_actual_and_expected_test_results(json_output_path: Path) -> None:
             for comparison_type, difference in filtered_diff.items():
                 om.add_variable(comparison_type, difference, info_map)
 
+    @staticmethod
+    def _convert_expected_result_variable_names(
+        expected_results: dict[str, Any], conversion_csv_path: Path
+    ) -> dict[str, Any]:
+        converted_expected_results: dict[str, Any] = {}
+        with open(conversion_csv_path, "r") as conversion_lookup_csv:
```
ew3361zh: do we need this if you're using pd.read_csv() the next line?

RUFAS/e2e_test_results_handler.py
```diff
@@ -86,6 +97,21 @@ def compare_actual_and_expected_test_results(json_output_path: Path) -> None:
...
+            conversion_lookup_table: pd.DataFrame = pd.read_csv(conversion_lookup_csv, index_col=None)
+        for key, value in expected_results.items():
+            if key in list(conversion_lookup_table["Original"]):
```
ew3361zh: What about adding some tracking if something has already been mapped so we don't get any accidental duplicates?

ew3361zh (commented):
Looks like a good start on this, Allister, I had some questions about the function doing the lookup/comparison which I think could help improve it.

I was wondering what the plan was to do the lookup table csv? I understand this function is intended to execute it once it's in place but it seems like the process of creating the initial comparison csv lookup table could be challenging.

RUFAS/e2e_test_results_handler.py
ew3361zh: e.g. conversion_dict = conversion_lookup_table.set_index("Original")["New"].to_dict()

RUFAS/e2e_test_results_handler.py
allisterakun: Following up on this:
I have put together a small script (see below) to check and report the duplicate mappings (both one original name to multiple new names and multiple original names to one new name). 
```python
from pathlib import Path
from typing import Any

import pandas as pd


def add_error(name: str, msg: str, info_map: dict[str, Any]) -> None:
    print(f"{name}: {msg}")

def find_duplicates(mapping: pd.DataFrame, column_to_check: str, result_column: str) -> dict[str, list[str]]:
    duplicates: dict[str, list[str]] = {}

    unique_variable_names = mapping[column_to_check].unique().tolist()
    all_variable_names = mapping[column_to_check].tolist()

    if len(all_variable_names) == len(unique_variable_names):
        return duplicates

    duplicate_variable_names: set[str] = set()
    for variable_name in all_variable_names:
        if variable_name in unique_variable_names:
            unique_variable_names.remove(variable_name)
        elif variable_name not in duplicate_variable_names:
            duplicate_variable_names.add(variable_name)
            corresponding_variable_names = mapping[mapping[column_to_check] == variable_name][result_column].tolist()
            duplicates[variable_name] = corresponding_variable_names
        else:
            continue
    return duplicates

def check_one_to_one_mapping(mapping: pd.DataFrame) -> None:
    info_map: dict[str, Any] = {}
    original_duplicates: dict[str, list[str]] = find_duplicates(
        mapping,
        column_to_check='Original',
        result_column='New'
    )
    new_duplicates: dict[str, list[str]] = find_duplicates(
        mapping,
        column_to_check='New',
        result_column='Original'
    )
    for original_name, new_names in original_duplicates.items():
        add_error(
            "Duplicate Mapping Error",
            f"Original variable name: '{original_name}' is mapping to multiple new variable names: {new_names}",
            info_map)

    for new_name, original_names in new_duplicates.items():
        add_error(
            "Duplicate Mapping Error",
            f"New variable name: '{new_name}' is mapped from multiple original variable names: {original_names}",
            info_map)


if __name__ == "__main__":
    file_path = Path("mapping.csv")
    
    mapping_df = pd.read_csv(file_path, index_col=None)
    check_one_to_one_mapping(mapping_df)
```
And I used this [mapping.csv](https://github.com/user-attachments/files/18802526/mapping.csv) table to test it.

 Original | New
-----|-----
 DISCLAIME | DISCLAIMER
 a | bb
 b | aa
 c | aa
 a | cc
 c | dd
 d | ee
 d | ff
 e | gg
 e | gg
 x | xx
 y | yy
 z | zz

Here is the printed output:

> D:\Work\Cornell\MASM\venv\Scripts\python.exe D:\Work\Cornell\MASM\check_duplicate.py 
Duplicate Mapping Error: Original variable name: 'a' is mapping to multiple new variable names: ['bb', 'cc']
Duplicate Mapping Error: Original variable name: 'c' is mapping to multiple new variable names: ['aa', 'dd']
Duplicate Mapping Error: Original variable name: 'd' is mapping to multiple new variable names: ['ee', 'ff']
Duplicate Mapping Error: Original variable name: 'e' is mapping to multiple new variable names: ['gg', 'gg']
Duplicate Mapping Error: New variable name: 'aa' is mapped from multiple original variable names: ['b', 'c']
Duplicate Mapping Error: New variable name: 'gg' is mapped from multiple original variable names: ['e', 'e']
> 
> Process finished with exit code 0

I decided to do the checking process in `pd.DataFrame` because we cannot have duplicate keys if we turn it into a Python dictionary. This means we will lose all the scenarios in which the same original key maps to multiple new variable names.
I plan to convert the DataFrame into a Python dictionary once we are sure of the one-to-one relationship.

Please review this and see if it addresses the potential issue. If so, I will move the code into E2ETestResultsHandler and update all the function calls, docstrings, and unit tests.

allisterakun: According to ChatGPT, I think this is a better and faster version:
```python
def find_duplicates_via_groupby(
    mapping: pd.DataFrame, group_col: str, list_col: str
) -> dict[str, list[str]]:
    """
    Groups the DataFrame by `group_col`, then collects all corresponding values
    from `list_col`. Returns a dictionary of
    {
        group_col_value: [list_col_value1, list_col_value2, ...]
    }
    only for those groups that contain more than one value (i.e., duplicates).
    """
    grouped: dict[str, list[str]] = mapping.groupby(group_col)[list_col].apply(list).to_dict()
    # Keep only entries where we have more than one mapped value.
    duplicates: dict[str, list[str]] = {
        group_val: mapped_vals
        for group_val, mapped_vals in grouped.items()
        if len(mapped_vals) > 1
    }
    return duplicates

```

RUFAS/e2e_test_results_handler.py
allisterakun: This makes sense. 
The time complexity of value lookup in a dictionary by a key is generally O(1) (worst case O(n) if many keys hash to the same bucket (due to collisions), which rarely happens).
Meanwhile, the time complexity for `new_key = conversion_lookup_table.loc[conversion_lookup_table["Original"] == key, "New"].iloc[0]` is O(n).

Will update!

RUFAS/e2e_test_results_handler.py
ew3361zh: I think this looks like a good solution to the issue.

Isgotkowitz (commented):
Tried out this feature, and it works as expected. Holding off on approving because I know there are some changes coming, and this PR appears to have the changes from some other PRs contained in it.

My only suggestion for improving this feature is minor, which is to make the table of original and new names an entry in the metadata (with some properties for validation, ofc).

allisterakun:
> Tried out this feature, and it works as expected. Holding off on approving because I know there are some changes coming, and this PR appears to have the changes from some other PRs contained in it.
> 
> My only suggestion for improving this feature is minor, which is to make the table of original and new names an entry in the metadata (with some properties for validation, ofc).

There are two things that need to be validated:
1. Column names
    1. There should be both a "Original" and a "New" column.
    2. A KeyError will be raised otherwise
3. One-to-one mapping
    1. Make sure that each original variable name maps to one and only one new variable name, and each new variable name is mapped from one and only one original variable name.
    2. A ValueError will be raised otherwise.

These checks seems pretty similar to the "validation" for report filters, which are handled upon retrieving rather than in the IM. Therefore, I have put the above validations in the `E2ETestResultsHandler` class.

ew3361zh (approved):
Excellent work, Allister! Tested this successfully with duplicate Original and New values and it correctly id'ed them and errored out.

Also tested changes across multiple e2e expected results files and it had no trouble working with each of them successfully.

I think the eventual CSV will be difficult to create but this sets up the comparison really well.

input/metadata/properties/tasks_properties.json
```diff
@@ -50,6 +50,12 @@
           "description": "The path to the second metadata properties JSON file to load and compare to the other metadata file loaded",
           "default": "input/metadata/properties/default.json"
         },
+        "convert_variable_table_path": {
+          "type": "string",
+          "pattern": "^.*.csv$",
+          "description": "The path to the CSV file that contains the previous and the new name of variables.",
+          "default": ""
```
Isgotkowitz:
```suggestion
          "default": null,
          "nullable": true,
```
Should this also be nullable? In this case I think it would make sense that E2E testing could run fine with or without this.

RUFAS/e2e_test_results_handler.py
```diff
@@ -86,6 +94,151 @@ def compare_actual_and_expected_test_results(json_output_path: Path) -> None:
...
+            in `expected_results` is not found in the mapping, it is preserved in the returned dictionary.
+
+        Raises
+        ------
+        KeyError
+            Raised if the conversion table CSV does not contain both 'Original' and 'New' columns.
+
+        ValueError
+            Raised if the conversion table CSV contains duplicate mappings for original variable names.
+        """
```
Isgotkowitz:
```suggestion
        Parameters
        ----------
        expected_results : dict[str, Any]
            A dictionary where the keys represent the original variable names and the values are
            the associated data.
        conversion_csv_path : Path
            The file path to the conversion CSV containing the mapping of original variable names
            to new variable names.

        Returns
        -------
        dict[str, Any]
            A dictionary with updated keys based on the conversion mappings from the CSV. If a key
            in `expected_results` is not found in the mapping, it is preserved in the returned dictionary.

        Raises
        ------
        KeyError
            Raised if the conversion table CSV does not contain both 'Original' and 'New' columns.
        ValueError
            Raised if the conversion table CSV contains duplicate mappings for original variable names.
        
        Notes
        -----
        Reads a CSV file containing mappings of original variable names to new variable names and applies
        these mappings to the keys in the given dictionary `expected_results`. The conversion table must
        contain two columns: 'Original' and 'New'. Ensures no duplicate mappings exist in the CSV and raises
        appropriate errors otherwise. Returns a dictionary with updated keys while preserving their associated
        values.

        """
```

RUFAS/e2e_test_results_handler.py
```diff
@@ -86,6 +94,151 @@ def compare_actual_and_expected_test_results(json_output_path: Path) -> None:
...
+        list_column_name : str
+            The column containing values that are mapped from 'group_column_name'.
+
+        Returns
+        -------
+        dict[str, list[str]]
+            A dictionary where each key is a duplicated entry from 'group_column_name',
+            and each value is a list of corresponding mapped values from 'list_column_name'.
+
+        """
```
Isgotkowitz:
```suggestion
        Identifies entries in a DataFrame where a single key maps to multiple values.

        Parameters
        ----------
        mapping : pd.DataFrame
            The DataFrame containing the mapping data. Must include the specified columns.
        group_column_name : str
            The column to be analyzed for duplicate mappings.
        list_column_name : str
            The column containing values that are mapped from 'group_column_name'.

        Returns
        -------
        dict[str, list[str]]
            A dictionary where each key is a duplicated entry from 'group_column_name',
            and each value is a list of corresponding mapped values from 'list_column_name'.

        Notes
        -----
        This method examines a DataFrame containing mappings between two columns and
        finds instances where a value in the 'group_column_name' column is associated
        with more than one unique value in the 'list_column_name' column.

        The result is a dictionary where:
          - Keys are the duplicated values from 'group_column_name'.
          - Values are lists of corresponding values from 'list_column_name'.

        """
```

RUFAS/e2e_test_results_handler.py
```diff
@@ -86,6 +94,151 @@ def compare_actual_and_expected_test_results(json_output_path: Path) -> None:
...
+            Raised if the conversion table CSV contains duplicate mappings for original variable names.
+        """
+        om: OutputManager = OutputManager()
+        info_map: dict[str, Any] = {
+            "class": E2ETestResultsHandler.__class__.__name__,
+            "function": E2ETestResultsHandler._convert_expected_result_variable_names.__name__,
+        }
+
+        converted_expected_results: dict[str, Any] = {}
+        df_conversion_lookup_table: pd.DataFrame = pd.read_csv(conversion_csv_path, index_col=None)
```
Isgotkowitz, line 142: Might be good to wrap this `pd.read_csv` call in some sort of try-except block, because reading in a file has a lot potential errors (particularly `FileNotFoundError`s). And having the try-except block would help it crash a bit more gracefully, or possibly let it continue without using the mapping.

Isgotkowitz (approved):
This looks good, nice job Allister! Left some comments that I think could make for minor improvements, but up to you for what is appropriate to implement at this time.

PR #2196, opened 2025-01-29
[C&S] Update inputs and metadata files to contain crop configurations

Isgotkowitz:
Modifies and expands RuFaS inputs in preparation for making crops configured in the inputs.

### Context

Issue(s) closed by this pull request: addresses #1831

### What

This PR prepares the inputs for PR #2179, which makes crops configurable for users. This PR is strictly limited to updating the Crop and Soil inputs, and does not actually change the functionality of the Crop and Soil module.

### Why

Modifying the input files causes the GitHub Actions to fail, which is bad. By isolating the changes that cause this failure to a single PR, we are able to more carefully check the changes to make sure that nothing is being broken.

### How

Copies the information differentiating different crop varieties from the dataclasses in `RUFAS/routines/field/crop/crop_configurations/` into a single JSON input file.

The metadata properties are updated so that the new crop configuration inputs will be accepted and validated when passed to a simulation.

Adds the crop configuration as an input to all the existing metadata files, so that the crop configuration information will be available in simulations.

Updated the DCA by running the DCA update task, so that crop configurations can be created with the DCA.

### Test plan

Ran the simulation, the test suite, and E2E testing to verify that nothing broke and there were no changes to outputs.

Ran the `./check_changes.sh` script (`./check_changes.bat` on Windows) to make sure that no Flake8 or Mypy errors were added.

Opened up the DCA and checked out the Crop Configurations input collection page to make sure it rendered correctly


DataCollectionApp/index.html
```diff
@@ -11,13 +11,15 @@
...
     <script src="./schema/crop_schedule_schema.js"></script>
+    <script src="./schema/crop_configuration_schema.js"></script>
     <script src="./schema/fertilizer_schedule_schema.js"></script>
     <script src="./schema/manure_schedule_schema.js"></script>
     <script src="./schema/tillage_schedule_schema.js"></script>
     <script src="./schema/feed_schema.js"></script>
     <script src="./schema/manure_management_schema.js"></script>
     <script src="./schema/field_schema.js"></script>
     <script src="./schema/tractor_dataset_schema.js"></script>
+    <script src="./schema/tasks_schema.js"></script>
```
Isgotkowitz, line 22: I didn't update `index.html` when I made tasks a type of input that can be collected with the DCA. I think that was probably a mistake, I reflecting on that I think it would be better to commit and track the changes that are made after running the DCA update task.

KFosterReed (approved):
This looks great Ed! Thanks for doing a good job with the metadata definitions!

ew3361zh (approved):
LGTM! Thanks for breaking this up, Ed. It was much more manageable this way.

PR #2203, opened 2025-02-03
[Time] Change "day" units from "simulation day" to "julian day"

### LINKED ISSUE #2184: Two different outputs for "day" have the same definition
There are these two outputs: 
`Time.day (simulation day)` and `Time.simulation_day (simulation day)`. One is the actual simulation day that is a continuous integer from the start of the simulation. The other is the Julian day. Having them both be defined as "simulation day is confusing. Time.day should be redefined as "Julian Day"

## Possible Solution

The output for `Time.day (simulation day)` is changed to `Time.day (Julian day)`

Isgotkowitz:
See title

### Context

Issue(s) closed by this pull request: closes #2184

### What

Again see title.

### Why

"simulation_day" is the incorrect unit for the julian day output.

### How

Added "JULIAN_DAY" to the units enum, and use it when recording time in the `Time` class.

### Test plan

Updated the unit test for the units class.

Checked the time output with filters `"Time.day"`, and `"Time.simulation_day"`, and verified that they both have the correct units right now.


allisterakun (approved):
LGTM!

ew3361zh (approved):

PR #2205, opened 2025-02-03
[Manure] Implement base `Processor` class

### LINKED ISSUE #2097: [Manure] Implement base `Processor` class
Implement the base `Processor` class. Use the "Developer Guidelines" in #2092.

Isgotkowitz:
Implements the base `Processor` class as part of the refresh.

### Context

Issue(s) closed by this pull request: closes #2097 

### What

Implements the base `Processor` class so that the rest of the refreshed Manure processors can inherit from it.

### Why

The interface of `Processor` will be inherited and implemented by every other processor in the Manure module, making it essential to implement the interface first before actually implementing the other processors.

### How

Implements `Processor` as an abstract class with two abstract methods: `receive_manure` and `process_manure`. It also implements logic that will be universal to all processors in `check_manure_stream_compatibility`.

### Test plan

Tests that `Processor` can't actually be instantiated, which is the correct behavior because no processor should be able to be instantiated unless it has `receive_manure` and `process_manure` implemented for it.


RUFAS/biophysical/manure/processor.py
```diff
@@ -0,0 +1,92 @@
...
+
+        """
+        is_valid_housing_emissions_calculator = (
+            True if (self.is_housing_emissions_calculator and manure_stream.pen_manure_data is not None) else False
+        )
+        is_valid_non_housing_emissions_calculator = (
+            True if (not self.is_housing_emissions_calculator and manure_stream.pen_manure_data is None) else False
+        )
+
+        return is_valid_housing_emissions_calculator ^ is_valid_non_housing_emissions_calculator
```
Isgotkowitz, line 97: `^` is the Python XOR operator, which is kind of overkill in this case.

tests/test_biophysical/test_manure/test_processor.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+from RUFAS.biophysical.manure.processor import Processor
+
+
+def test_processor_init_error() -> None:
+    """Test that base Processor class throws appropriate error when initialized."""
+    with pytest.raises(TypeError):
+        Processor(is_housing_emissions_calculator=True)
```
Isgotkowitz, line 8: Mypy doesn't like that I'm trying to instantiate `Processor` because it is an abstract class with missing implementations, but that is the whole point of this test 🙄

elle-andreen (approved):
This looks good!

Adding some relevant notes from Ed on the validation that this file does (`check_manure_stream_compatibility`) for future reference:

If a ManureStream is taken in by a Processor that _does_ calculate housing emissions, but the pen_manure_data field is none and housing emissions cannot be calculated, an exception will be raised. Likewise, if a ManureStream is taken in by a Processor that _doesn't_ calculate housing emissions and the pen_manure_data field is not None, it means that ManureStream has not been processed by another Processor before. So no housing emissions have been calculated for it, and since every Processor sets the pen_manure_data field to None after processing it will not be possible for housing emissions to be calculated on that ManureStream at any subsequent point in the chain (to prevent double-counting of housing emissions). 

Note that processors that calculate housing emissions will be handlers (though parlor cleaning handlers will also return 0 values as manure is assumed to be cleaned too rapidly to result in considerable emissions), as well as open lot and CBPB storages.

tests/test_biophysical/test_manure/test_processor.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+# TODO: test with a child or grandchild of Processor, either in #2098 or #2099
```
matthew7838, line 11: Which method is this comment for?

tests/test_biophysical/test_manure/test_processor.py
```diff
@@ -0,0 +1,14 @@
...
+# def test_check_manure_stream_compatibility() -> None:
+#     """Tests that ManureStreams are correctly checked for compatibility."""
+#     pass
```
matthew7838: If these are passed I don't think there's a need to comment these out.

RUFAS/biophysical/manure/processor.py
```diff
@@ -0,0 +1,92 @@
...
+from RUFAS.time import Time
+
+
+class Processor(ABC):
+    """
+    Base class for all manure processors.
+
+    Attributes
+    ----------
+    is_housing_emissions_calculator : bool
```
allisterakun, line 19: Maybe put this into __init__ docstring?
I think the variables here are all class variables instead of instance variables

RUFAS/biophysical/manure/processor.py
```diff
@@ -0,0 +1,92 @@
...
+    Methods
+    -------
+    receive_manure(manure: ManureStream) -> None
+        Entry point of manure into the processor.
+    process_manure(conditions: CurrentDayConditions, time: Time) -> dict[str, ManureStream]
+        Handles the daily operations for the processor.
+
+    """
+
+    def __init__(self, is_housing_emissions_calculator: bool) -> None:
```
allisterakun, line 31: Should processors also have an attribute that holds the manure in them?

allisterakun: Or not all processors will hold some manure in them?

allisterakun (approved):
LGTM! Left two comments.

matthew7838 (approved):
LGTM, left a few comments for clarifications. Great work.

tests/test_biophysical/test_manure/test_processor.py
Isgotkowitz: Good point

tests/test_biophysical/test_manure/test_processor.py
Isgotkowitz: `check_manure_stream_compatibility`

RUFAS/biophysical/manure/processor.py
Isgotkowitz: I think it might go in both. [Numpy's guide](https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring) says "An Attributes section, located below the [Parameters](https://numpydoc.readthedocs.io/en/latest/format.html#params) section, may be used to describe non-method attributes of the class:", which I interpret as both class and instance variables.

RUFAS/biophysical/manure/processor.py
Isgotkowitz: Great question, in reality not all processors hold manure. Handlers push manure around, while storages, digesters and some separators accumulate manure and hold it over multiple days.

But in RuFality, every processor will hold manure but I'm not sure that they will do so in the same manner. Like some may accumulate all the manures they receive into a single `ManureStream`, while others may hold multiple `ManureStream`s in a list or some other data structure before processing.

So for now, I'll leave out an attribute that tracks manure held by the Processor, but as we get deeper into Phase 2 we may identify that there should an attribute like that in `Processor`.

PR #2209, opened 2025-02-04
[Manure] Quick Fix for NoneType error when logging an empty manure request result

### LINKED ISSUE #2208: Empty Manure Request causes error resulting in simulation failure
Testing the pilot farms with the most recent updates in dev led to an error related to an empty manure nutrient request. 

## Expected Behavior

The manure applications should proceed without error. 

## Current Behavior

When running the pilot scenario `c1f1_wcrops_metadata.json`, an error is produced that 
## Steps to Reproduce

1. Checkout the pilot_testing_Dec2023 branch
2. Run c1f1_wcrops_metadata.json
3.
4.

## Context (Environment)

We are unable to run the pilot farms or develop end to end testing

## Detailed Description

This error is created and occurs at Julian day 301 year 2020: 

Image

I tried to recreated the error by changing the manure schedule in the default scenario but I wasn't able to do so. 

## Possible Implementation

No ideas
## Target Deadline

ASAP

ew3361zh:
Addresses error found when the manure requests are so great they tap out the on-farm manure and we're logging empty manure request results.

### Context

Issue(s) closed by this pull request: closes #2208

### What

When field makes a manure request and there is no available on-farm manure, `ManureNutrientManager` returns `None`. We still want to log this `ManureNutrientRequestResult` so are mimicking the behavior in `Field._execute_manure_application()` when there is no manure returned to try and apply to the field. This behavior is logging all 0s for the nutrients in the request result.

### Why

On the pilot farm branch we were seeing an error arise that hadn't come up on dev. The pilot farm being tested had multiple fields and was using up all the on-farm manure and None was eventually returned from `ManureNutrientManager`. Since we're now logging all nutrient request results by attribute, this was leading to an error because NoneTypes don't have any of the nutrient request result attributes.

### How

Added a check in the `ManureManager._record_manure_request_results()` function that if the ManureNutrientRequestResults is None, we're recording the result with all 0s for the attributes.

### Test plan

I think the best test plan is to try this out on the pilot testing branch as Kristan laid out in the issue. It's a small slice of code so I would copy it, switch to the pilot testing branch, paste it, and run the code with the pilot farm metadata Kristan specifies in the issue.


Isgotkowitz (approved):
LGTM, fix works as expected for the pilot farm simulation.

KFosterReed (approved):
This works as expected now! Thanks for the quick fix :)

PR #2211, opened 2025-02-05
[DCA] Improve messages informing user of invalid data.

### LINKED ISSUE #2124: [DCA] Improve invalid input messages in the Data Collection App
Improve the error messages generated by the DCA when invalid data is entered.

## Status

When invalid data is entered in the DCA, a red box appears under the data entry tabs with the message "Value must validate against at least one of the provided schemas". At the very bottom of the page, under "Validation", the only message that is ever displayed is:
```json
[
  {
    "path": "root.0",
    "property": "anyOf",
    "message": "Value must validate against at least one of the provided schemas",
    "errorcount": 1
  }
]
```

## Intention

Make the error message specific so that the cause can be easily identified and fixed.

## Request

Modify the DCA so that the error message given in the "Validation" box is specific, and guides the user towards valid input.

### Additional Information

Reference the [JSON-Editor](https://github.com/json-editor/json-editor) framework for how to do this. It should have a way to provide better error messages.

Also, ChatGPT knows about JSON-Editor so it can be asked framework-specific questions as well.

Isgotkowitz:
Improves the error messages in the DCA when invalid input is entered.

### Context

Issue(s) closed by this pull request: closes #2124

### What

When invalid data is entered into input fields in the DCA, the validation errors specify which tab ("Data Entry") and variable field the invalid data is in. See below for example:
image

### Why

Specific validation errors enable the user to quickly identify and fix any mistakes they have made, making the DCA a more user-friendly tool.

### How

Whenever there is a change to the DCA, the JSON-Editor uses its internal validation routine (`jsoneditor.validate()`) to check if the input data are valid against the available schema. If there are any errors present, then they are parsed by the new method `stringifyRelevantErrorObjects` so that irrelevant errors are filtered out and the objects which contain information about the error is converted to a user-friendly string.

Part of the string conversion is changing the index of the schema against which the data is checked to a recognizable name, for example changing `0` to `"Config Properties"`. This is done using a mapping between schema indexes and property names generated by the new method `mapSchemaIndexToPropertyNames`.

When validating data, the DCA now uses "oneOf" the schema instead of "anyOf" the schema i.e. that data entered in a tab should be valid against one and only one of the available schema, not any of them. This may present problems in the future, but for now I think it is a reasonable assumption (and Idk how to make this work without changing that).

### Test plan

Ran the `dca_update_task.json` in the default metadata then opened up the DCA in my browser.

Opened up multiple "Data Entry" tabs, and:
- made them different input types
- entered invalid data
- fixed invalid data
- switched the same Data Entry tab between different input types.

All while keeping an eye on the "Validation" box to make sure that the validation errors (if any) matched what I expected them to be. Also tried downloading input files to make sure that the downloads still work (they do).


RUFAS/data_collection_app_updater.py
```diff
@@ -147,7 +147,7 @@ def _rewrite_index_page(self, schema_paths: list[Path]) -> None:
 
         pattern_to_remove = r"\./schema/|\.js"
         schema_names = [re.sub(pattern_to_remove, "", name) for name in localized_schema_paths]
-        list_of_schema = f'"anyOf": {schema_names}'.replace("'", "")
+        list_of_schema = f'"oneOf": {schema_names}'.replace("'", "")
```
Isgotkowitz, line 150: This line gets written into `index.html`, which is why I am updating it here.

DataCollectionApp/template
Isgotkowitz, line 1: `template` is used to regenerate `index.html`, so changes need to be present in this file for them to persist in `index.html` when the DCA is updated.

elle-andreen (approved):
This worked as expected for me! Thanks Ed, this is helpful for validation. Some of the validation errors are not as useful as they maybe could be (e.g. matching a pattern that is non-sensical to someone without coding knowledge) but that's beyond the scope of this PR, and I think even just drawing attention to the error is very helpful.

matthew7838 (approved):

PR #2212, opened 2025-02-05
[E2E Testing] Update feed storage e2e results

Isgotkowitz:
Updates E2E testing expected results for the Feed Storage module.

### Context

Issue(s) closed by this pull request: none, follows up on changes made for #2184

### What

Updates the E2E expected results for Feed Storage module to accommodate for the change in units made by #2203

### Why

End to end testing was failing for the Feed Storage module, because the expected results had "simulation day" as the unit for `Time.day` instead of "julian day".

### How

Used the update E2E results task.

### Test plan

Opened up the old results from `dev` and the new results in this branch in `vimdiff`, and looked through results to confirm that the only differences were the units of `Time.day`.

Also reran E2E testing to make sure that it is now passing.

### Other
When I ran the E2E test results update task, the Animal results were updated as well. I looked into it, and it seemed that the only differences were insignificant floating point differences and order of cows.


ew3361zh:
> When I ran the E2E test results update task, the Animal results were updated as well. I looked into it, and it seemed that the only differences were insignificant floating point differences and order of cows.

Maybe it's because I made the deepdiff check in the e2e results update less flexible than the check for regular e2e testing (assuming default tolerance levels are used).

ew3361zh (approved):

matthew7838 (approved):
LGTM, changes confirmed, e2e tests passing.

allisterakun (approved):
LGTM

PR #2214, opened 2025-02-07
[Manure] Implement base manure `Storage` class for refresh

### LINKED ISSUE #2098: [Manure] Implement base `Storage` class
Implement the base `Storage` class. Use the "Developer Guidelines" in #2092.

Isgotkowitz:
Implements the base `Storage` class and logic that is shared between child storages (and other types of processors).

### Context

Issue(s) closed by this pull request: closes #2098 

### What

Implements the base manure `Storage` class, and some logic that is shared by multiple child classes (of both `Storage` and `Processor`).

### Why

Implementing the base `Storage` class clears the way forward for refreshing the actual manure storages.

### How

A base manure `Storage` class is created, and as much shared logic as I could find in `RUFAS/routines/manure/gas_emissions/calculator.py` is implemented in both the `Storage` and `Processor` classes.

Also, I added some things to `Processor` that I realized in hindsight will be needed by all of its children.

All of the new methods for emissions calculations were implemented as `classmethod`s, because Storage and Process still cannot be instantiated and the new routines would not be testable in this PR if they were implemented as instance methods.

### Test plan

Unit tests have been written for almost all new functionality added. Some of it could not be tested because `Storage` and `Processor` still cannot be instantiated, but I have updated the TODO for adding tests to them.

### Updates
@elle-andreen put together a list of overlapping logic from `BaseManureTreatment` that would also be appropriate to put in the base `Storage` class. I have implemented this logic and added unit tests as appropriate, including unit tests for `receive_manure` and `process_manure`.


RUFAS/biophysical/manure/processor.py
```diff
@@ -95,3 +106,159 @@ def check_manure_stream_compatibility(self, manure_stream: ManureStream) -> bool
         )
 
         return is_valid_housing_emissions_calculator ^ is_valid_non_housing_emissions_calculator
+
+    @classmethod
+    def _calculate_ammonia_emissions(
```
Isgotkowitz, line 114: This method combines the functionality of both `calculate_housing_ammonia_emission` and `calculate_liquid_storage_ammonia_emission` in `calculator.py`.

RUFAS/biophysical/manure/processor.py
```diff
@@ -95,3 +106,159 @@ def check_manure_stream_compatibility(self, manure_stream: ManureStream) -> bool
...
+        total_ammoniacal_nitrogen_per_meter = total_ammoniacal_nitrogen / surface_area
+        equilibrium_coefficient = cls._calculate_ammonia_equilibrium_coefficient(temp_kelvin, pH)
+        ammonia_loss_per_meter = (total_ammoniacal_nitrogen_per_meter * GeneralConstants.SECONDS_PER_DAY * density) / (
+            ammonia_resistance * manure_kilograms_per_square_meter * equilibrium_coefficient
+        )
+        total_ammonia_loss = min(ammonia_loss_per_meter * surface_area, total_ammoniacal_nitrogen)
+        return max(0.0, total_ammonia_loss)
+
+    @classmethod
+    def _calculate_ammonia_equilibrium_coefficient(cls, temp: float, pH: float) -> float:
```
Isgotkowitz, line 184: Reimplements `_equilibrium_coefficient` from `calculator.py`.

RUFAS/biophysical/manure/processor.py
```diff
@@ -95,3 +106,159 @@ def check_manure_stream_compatibility(self, manure_stream: ManureStream) -> bool
...
+        float
+            Equilibrium coefficient for the ammonia gas in the air (unitless).
+
+        """
+        henrys_ammonia_coefficient = cls._calculate_henry_law_coefficient_of_ammonia(temp)
+        ammonium_dissociation_coefficient = cls._calculate_dissociation_coefficient_of_ammonium(temp, pH)
+        return henrys_ammonia_coefficient * ammonium_dissociation_coefficient
+
+    @classmethod
+    def _calculate_henry_law_coefficient_of_ammonia(cls, temp: float) -> float:
```
Isgotkowitz, line 207: Reimplements `_henry_law_coefficient_of_ammonia` in `calculator`.py

RUFAS/biophysical/manure/processor.py
```diff
@@ -95,3 +106,159 @@ def check_manure_stream_compatibility(self, manure_stream: ManureStream) -> bool
...
+        Returns
+        -------
+        float
+            Henry's law coefficient of ammonia (unitless).
+
+        """
+        return 10 ** (1478 / temp - 1.69)
+
+    @classmethod
+    def _calculate_dissociation_coefficient_of_ammonium(cls, temp: float, pH: float) -> float:
```
Isgotkowitz, line 225: Reimplements `_dissociation_coefficient_of_ammonium` in `calculator.py`.

RUFAS/biophysical/manure/processor.py
```diff
@@ -95,3 +106,159 @@ def check_manure_stream_compatibility(self, manure_stream: ManureStream) -> bool
...
+        ammonia_loss_per_meter = (total_ammoniacal_nitrogen_per_meter * GeneralConstants.SECONDS_PER_DAY * density) / (
+            ammonia_resistance * manure_kilograms_per_square_meter * equilibrium_coefficient
+        )
+        total_ammonia_loss = min(ammonia_loss_per_meter * surface_area, total_ammoniacal_nitrogen)
+        return max(0.0, total_ammonia_loss)
+
+    @classmethod
+    def _calculate_ammonia_equilibrium_coefficient(cls, temp: float, pH: float) -> float:
+        """
+        Calculates the equilibrium coefficient for the :math:`NH_3` gas in the air for a given concentration of total
```
Isgotkowitz:
```suggestion
        Calculates the equilibrium coefficient for the ammonia gas in the air for a given concentration of total
```

RUFAS/biophysical/manure/processor.py
```diff
@@ -95,3 +106,159 @@ def check_manure_stream_compatibility(self, manure_stream: ManureStream) -> bool
...
+        Returns
+        -------
+        float
+            Dissociation coefficient of ammonium (unitless).
+
+        """
+        return 1 + 10 ** (0.09018 + 2729.9 / temp - pH)
+
+    @classmethod
+    def _calculate_digester_storage_nitrous_oxide_emissions(
```
Isgotkowitz: Reimplements `calculate_empirical_nitrogen_loss_from_nitrous_oxide_emission` in `calculator.py`.

RUFAS/biophysical/manure/processor.py
```diff
@@ -95,3 +106,159 @@ def check_manure_stream_compatibility(self, manure_stream: ManureStream) -> bool
...
+
+        """
+        henrys_ammonia_coefficient = cls._calculate_henry_law_coefficient_of_ammonia(temp)
+        ammonium_dissociation_coefficient = cls._calculate_dissociation_coefficient_of_ammonium(temp, pH)
+        return henrys_ammonia_coefficient * ammonium_dissociation_coefficient
+
+    @classmethod
+    def _calculate_henry_law_coefficient_of_ammonia(cls, temp: float) -> float:
+        """
+        Calculate Henry's law coefficient of ammonia.
```
Isgotkowitz:
```suggestion
        Calculate Henry's Law coefficient of ammonia.
```

RUFAS/biophysical/manure/processor.py
```diff
@@ -95,3 +106,159 @@ def check_manure_stream_compatibility(self, manure_stream: ManureStream) -> bool
...
+        total_ammoniacal_nitrogen: float,
+        volume: float,
+        density: float,
+        temperature: float,
+        ammonia_resistance: float,
+        surface_area: float,
+        pH: float,
+    ) -> float:
+        """
+        Calculate storage ammonia emissions for liquidmanure treatments.
```
Isgotkowitz:
```suggestion
        Calculate housing and liquid storage ammonia emissions.
```

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -0,0 +1,188 @@
...
+    Attributes
+    ----------
+    _received_manure : ManureStream
+        The total amount of manure received by a storage in a single day.
+    _stored_manure : ManureStream
+        The current amount of manure currently held by the storage.
+    _cover : StorageCover
+        The cover of the storage.
+    _storage_time_period : int
+        Interval between emptyings of the storage (days).
```
Isgotkowitz: I'm pretty sure the correct thing to do is not redocument parameters/attributes that are documented in the base class.

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -0,0 +1,188 @@
...
+            raise ValueError(error_message)
+
+        self._received_manure += manure
+
+    @classmethod
+    def _calculate_methane_emissions(
+        cls, volatile_solids: float, manure_temperature: float, is_degradable: bool
+    ) -> float:
+        """
+        Calculate the methane emission.
```
Isgotkowitz:
```suggestion
        Calculate the methane emitted from liquid manure storages.
```

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -0,0 +1,188 @@
...
+                "manure_stream": manure,
+            }
+            error_message = f"Processor {self.name} received an incompatible ManureStream."
+            self._om.add_error("invalid_manure_stream", error_message, info_map)
+            raise ValueError(error_message)
+
+        self._received_manure += manure
+
+    @classmethod
+    def _calculate_methane_emissions(
```
Isgotkowitz, line 236: This method reimplements `calculate_liquid_storage_methane` in `calculator.py` in a way that is less duplicative, but will require two method calls instead of one.

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -0,0 +1,188 @@
...
+            * arrhenius_exponent
+            * volatile_solids
+            * rate_correcting_factor
+            * GeneralConstants.GRAMS_TO_KG
+        )
+
+        return methane_emissions
+
+    @classmethod
+    def _calculate_arrhenius_exponent(cls, temp: float) -> float:
```
Isgotkowitz, line 277: Reimplements `_arrhenius_exponent` from `calculator.py`.

RUFAS/biophysical/manure/storage/storage_cover.py
```diff
@@ -0,0 +1,24 @@
+from enum import Enum
+
+
+class StorageCover(Enum):
```
Isgotkowitz, line 4: There used to be an N/A cover, but this was only used for Anaerobic Digesters which will no longer be a separator. Thus every separator will have a cover, and if it doesn't then `NO_COVER` should suffice.

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -0,0 +1,188 @@
...
+from RUFAS.util import Utility
+
+from .storage_cover import StorageCover
+
+
+"""
+Activation energy (joules per mole, J/mol). The activation energy is the minimum energy that must be available to
+molecules for a reaction to occur.
+"""
+ACTIVATION_ENERGY: float = 81_000.0
```
Isgotkowitz, line 19: These constants were pulled in from `GasEmissionsConstants`

tests/test_biophysical/test_manure/test_storage/test_storage.py
```diff
@@ -0,0 +1,53 @@
...
+
+from RUFAS.biophysical.manure.storage.storage import Storage
+from RUFAS.biophysical.manure.storage.storage_cover import StorageCover
+
+
+def test_storage_init_error() -> None:
+    """Test that Storage cannot be initialized because it does not fully implement the abstract base class."""
+
+    with pytest.raises(TypeError):
+        Storage(
```
Isgotkowitz: This call adds a Mypy error because Storage doesn't have all the abstractmethods of `Processor` implemented.

RUFAS/util.py
```diff
@@ -573,6 +573,11 @@ def generate_time_series(date: datetime.date, starting_offset: int, ending_offse
 
         return time_series
 
+    @staticmethod
+    def convert_celsius_to_kelvin(temperature: float) -> float:
+        """Converts a temperature in degrees Celsius to degrees Kelvin."""
+        return temperature + 273.15
```
allisterakun: Put 273.15 into `general_constants.py`

RUFAS/biophysical/manure/processor.py
```diff
@@ -95,3 +106,159 @@ def check_manure_stream_compatibility(self, manure_stream: ManureStream) -> bool
...
+        total_ammoniacal_nitrogen : float
+            Total ammoniacal nitrogen in manure (kg).
+        volume : float
+            Total volume of the manure produced by the animals in the storage area (m^3).
+        density : float
+            Density of the manure (kg / m^3).
+        total_solids : float
+            Total solids present in the manure (kg).
+        temperature : float
+            Ambient temperature around where manure is (degrees C).
```
elle-andreen:
```suggestion
            Temperature of the manure (degrees C).
```

RUFAS/biophysical/manure/processor.py
```diff
@@ -95,3 +106,159 @@ def check_manure_stream_compatibility(self, manure_stream: ManureStream) -> bool
...
+        ammonia_resistance : float
+            Resistance of ammonia transport to the atmosphere (siemens / meter).
+        surface_area : float
+            Total surface area of the manure storage (m^2).
+        pH : float
+            pH value for storage ammonia emission (pH).
```
elle-andreen:
```suggestion
            pH value of the manure (unitless).
```

RUFAS/biophysical/manure/processor.py
```diff
@@ -95,3 +106,159 @@ def check_manure_stream_compatibility(self, manure_stream: ManureStream) -> bool
...
+
+        Returns
+        -------
+        float
+            Storage ammonia emission (kg).
```
elle-andreen:
```suggestion
            Ammonia N emission from manure (kg).
```

RUFAS/biophysical/manure/processor.py
```diff
@@ -95,3 +106,159 @@ def check_manure_stream_compatibility(self, manure_stream: ManureStream) -> bool
...
+        """
+        Calculates the equilibrium coefficient for the ammonia gas in the air for a given concentration of total
+        ammoniacal nitrogen in the solution.
+
+        Parameters
+        ----------
+        temp : float
+            Manure solution temperature in Kelvin (K).
+        pH : float
+            Manure solution acidity (unitless).
```
elle-andreen:
```suggestion
            Manure solution pH (unitless).
```

RUFAS/biophysical/manure/processor.py
elle-andreen: Would it be appropriate to make the name and description of this function more general, so that it's not specific to digesters? My understanding is by putting it here, we make this method available to all processor types (which is reasonable as both digesters and storages will use this method)

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -0,0 +1,188 @@
...
+            potassium=0.0,
+            ash=0.0,
+            non_degradable_volatile_solids=0.0,
+            degradable_volatile_solids=0.0,
+            total_solids=0.0,
+            volume=0.0,
+            pen_manure_data=None,
+        )
+        self._cover = cover
+        self._storage_time_period = storage_time_period
```
elle-andreen, line 119: Is this where the new storage surface area input should be added for storages? (Multiple storage types will reference a surface area value)

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def receive_manure(self, manure: ManureStream) -> None:
+        """Receives manure and puts it in storage to be processed."""
+        is_compatible_manure = self.check_manure_stream_compatibility(manure)
```
elle-andreen, line 131: I may be forgetting something, but is this more or less a placeholder for where/how we will implement future restrictions on passing manure?

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -0,0 +1,188 @@
...
+            raise ValueError(error_message)
+
+        self._received_manure += manure
+
+    @classmethod
+    def _calculate_methane_emissions(
+        cls, volatile_solids: float, manure_temperature: float, is_degradable: bool
+    ) -> float:
+        """
+        Calculate the methane emitted from liquid manure storages.
```
elle-andreen:
```suggestion
        Calculates methane that is emitted from liquid manure storages by calculating emissions from the degradable and non-degradable volatile solids fractions. 
```

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -0,0 +1,188 @@
...
+
+        Parameters
+        ----------
+        volatile_solids : float
+            Mass of volatile solids that are emitting methane (kg). Can be degrdable or non-degradable.
```
elle-andreen:
```suggestion
            Mass of volatile solids that are emitting methane (kg). Can be degradable or non-degradable.
```

RUFAS/biophysical/manure/storage/storage_cover.py
elle-andreen: I think your brain may have substituted separator for storage here haha. But what do you mean by every storage will have a cover?

RUFAS/biophysical/manure/processor.py
Isgotkowitz: Instead of `(kg NH3-N)`, do you think it would be ok to just list the units as kg?

elle-andreen: > Instead of `(kg NH3-N)`, do you think it would be ok to just list the units as kg?

I modified to I think a better option! It's just important to note that this function is for ammonia **nitrogen** emission, not ammonia emission. There is an issue related to this but may as well fix low hanging fruit as we go.

RUFAS/biophysical/manure/processor.py
Isgotkowitz: More general as in it doesn't mention any specific processor type? Right now the title mentions both digesters and storages, but will be inherited by every type of Processor. So this name isn't ideal, but I worry that making it too generic will be make it harder to override in other child classes.

RUFAS/biophysical/manure/storage/storage.py
Isgotkowitz: Yes, this is where it should be. Thank you for catching this.

RUFAS/biophysical/manure/storage/storage.py
Isgotkowitz: I don't think so, all the cases I can think of right now should be covered by this. But as we get into the actual Storage children and post-refresh, it may be the case that this needs to be changed to be more flexible.

RUFAS/biophysical/manure/processor.py
elle-andreen: My brain either completely glossed over the "or storage" parts here, or I had not pulled the most recent changes or something. What's here is good!

RUFAS/biophysical/manure/storage/storage.py
Isgotkowitz: Implemented

Isgotkowitz:
Update: working on adding some more logic that overlaps between different storages

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -0,0 +1,335 @@
...
+            self._om.add_variable("emptied_manure_total_ammoniacal_nitrogen", self._stored_manure.water, info_map)
+            self._om.add_variable("emptied_manure_nitrogen", self._stored_manure.water, info_map)
+            self._om.add_variable("emptied_manure_phosphorus", self._stored_manure.water, info_map)
+            self._om.add_variable("emptied_manure_potassium", self._stored_manure.water, info_map)
+            self._om.add_variable("emptied_manure_ash", self._stored_manure.water, info_map)
+            self._om.add_variable("emptied_manure_non_degradable_volatile_solids", self._stored_manure.water, info_map)
+            self._om.add_variable("emptied_manure_degradable_volatile_solids", self._stored_manure.water, info_map)
+            self._om.add_variable("emptied_manure_total_volatile_solids", self._stored_manure.water, info_map)
+            self._om.add_variable("emptied_manure_total_solids", self._stored_manure.water, info_map)
+            self._om.add_variable("emptied_manure_mass", self._stored_manure.water, info_map)
```
ew3361zh: Not doing a full review but checking this code for what I'm working on with Separators and saw that all the var values being sent here (and below) are `self_stored_manure.water` and wanted to make sure it got caught.

RUFAS/biophysical/manure/processor.py
```diff
@@ -95,3 +109,159 @@ def check_manure_stream_compatibility(self, manure_stream: ManureStream) -> bool
...
+            Amount of nitrogen that was added to the digester or storage on the current day (kg).
+
+        Returns
+        -------
+        float
+            Nitrous oxide nitrogen emissions (kg).
+
+        """
+
+        return nitrous_oxide_emissions_factor * nitrogen_added
```
elle-andreen: Making a note here for future reference that processors that calculate nitrous oxide emissions (e.g. digesters and storages, at this time) will contain logic for determining N2O emission factor in their individual processor file, which is then passed to this (inherited) function. Emission factor logic that is based on whether the storage is covered (applicable to liquid storages) is inherited via the base Storage class. The individual logics to determine emission factors exist currently in base_manure_treatment.py, in the processor ("treatment") file itself, or in calculator.py.

RUFAS/biophysical/manure/storage/storage.py
Isgotkowitz: Nice catch, thanks Niko:)

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -0,0 +1,356 @@
...
+        if not is_compatible_manure:
+            info_map = {
+                "class": self.__class__.__name__,
+                "function": self.receive_manure.__name__,
+                "processor_name": self.name,
+                "manure_stream": manure,
+            }
+            error_message = f"Processor '{self.name}' received an incompatible ManureStream."
+            self._om.add_error("invalid_manure_stream", error_message, info_map)
+            print(error_message)
```
Isgotkowitz:
```suggestion
```

elle-andreen (approved):
This looks all good to me! Thanks Ed! Looking forward to moving on to the individual storage processors soon 💩

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -0,0 +1,355 @@
...
+class Storage(Processor):
+    """
+    Base manure Storage class.
+
+    Parameters
+    ----------
+    cover : StorageCover
+        What the storage will be covered with, if anything.
+    storage_time_period : int | None
+        How long manure is stored for before emptying the storage (days). None if the storage is never emptied.
```
allisterakun, line 76: If the storage is never emptied, would it make more intuitive to set the `storage_time_period` to `sys.maxsize` rather than `None`?

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -0,0 +1,355 @@
...
+            )
+            self._om.add_variable(
+                "emptied_manure_degradable_volatile_solids", self._stored_manure.degradable_volatile_solids, info_map_kg
+            )
+            self._om.add_variable(
+                "emptied_manure_total_volatile_solids", self._stored_manure.total_volatile_solids, info_map_kg
+            )
+            self._om.add_variable("emptied_manure_total_solids", self._stored_manure.total_solids, info_map_kg)
+            self._om.add_variable("emptied_manure_mass", self._stored_manure.mass, info_map_kg)
+            self._om.add_variable("emptied_manure_volume", self._stored_manure.volume, info_map_m3)
```
allisterakun: These output reporting function calls could be extracted into a function.

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -0,0 +1,355 @@
...
+        )
+        self._om.add_variable(
+            "accumulated_manure_degradable_volatile_solids", self._stored_manure.degradable_volatile_solids, info_map_kg
+        )
+        self._om.add_variable(
+            "accumulated_manure_total_volatile_solids", self._stored_manure.total_volatile_solids, info_map_kg
+        )
+        self._om.add_variable("accumulated_manure_total_solids", self._stored_manure.total_solids, info_map_kg)
+        self._om.add_variable("accumulated_manure_mass", self._stored_manure.mass, info_map_kg)
+        self._om.add_variable("accumulated_manure_volume", self._stored_manure.volume, info_map_m3)
```
allisterakun: These output reporting function calls could be extracted into a function.

RUFAS/biophysical/manure/storage/storage.py
allisterakun: Why are these constants pulled into `storage.py`? I assume they will also be used in other processors?

allisterakun (approved):
LGTM, Great job, Ed!
Left some take-or-leave comments, please take a look!

RUFAS/biophysical/manure/processor.py
Isgotkowitz: Thanks:) I think what is here is good for now, but a better approach may be to make the functional name more general, and override it in child classes to raise a NotImplemented Error for all child classes that don't emit nitrous oxide.

RUFAS/biophysical/manure/storage/storage.py
Isgotkowitz: As best I could tell, they are only used for routines that `Storage` utilizes. `GasEmissionsConstants` was a catch-all for gas-emissions related constants, even though some of them are only used in a spot or two.

RUFAS/biophysical/manure/storage/storage.py
Isgotkowitz: Good question, I will think about this and try out whether the modulus math would still work. I think `None` is good in that it is explicit, but using `maxsize` would make a lot of sense too.

RUFAS/biophysical/manure/storage/storage.py
Isgotkowitz: Good call, I will find a way to recycle the logic so that it can be used for the accumulated manure outputs too.

RUFAS/biophysical/manure/storage/storage.py
Isgotkowitz: Modulus math would work fine unless `simulation_day` is equal to `sys.maxsize`, but that is a pretty unreasonable edge case.

I am going to keep it as-is though because the `None` keeps the code more readable, it may not be as obvious what `sys.maxsize` means either when someone is reading the code or debugging values that are held.

RUFAS/biophysical/manure/storage/storage.py
Isgotkowitz: Extracted!

RUFAS/biophysical/manure/processor.py
elle-andreen: Hi @Isgotkowitz, very last-minute change to this PR - we realized recently that the N2O emissions factor is being used erroneously with digesters. There really should be no N2O emissions from this type of digester given that it is nearly gas-tight. I believe all this means for this PR is that this function could move from Processor down to Storage and can be called something like `_calculate_storage_nitrous_oxide_emissions`!

Isgotkowitz: @elle-andreen thanks for letting me know! That is gonna make the code much cleaner

PR #2221, opened 2025-02-07
[DCA] Add dropdown to the feed input fields

### LINKED ISSUE #2126: [DCA] Change feed input to have a drop-down of names instead of typed feed IDs
Change the schema for "Feed Properties" so that there is a drop-down of feed names.

## Status

When users configure a feed input file, including the feeds available for ration formulation with each group of animals, they have to type in the numerical ID that references the ID from a Feed library file.

## Intention

Make the Feed input creation process easier by having Feed names listed alongside the numeric feed IDs.

## Request

Change all Feed inputs which take in a typed numeric Feed ID to be a drop-down of options, with the Feed name and numeric ID listed in each option.

This drop-down must be preserved through the automated updates to the DCA schema.

### Additional Information

See the original DCA implementation to see what the final product should look like.

matthew7838:
Adds dropdown options for feed data inputs with it's corresponding id.
### Context

Issue(s) closed by this pull request: closes #2126

### What

- Added a function to gather all the options in available user feeds.
- Added another main routine to update the newly generated schema with dropdowns, the gathered available user feeds will be passed here and used to modify the schema.

### Why

User-friendliness.

### How

Users don't need to manually type the id.

### Test plan

Run the DCA update task and check out all the fields for feed data. Unit tests passed.


RUFAS/data_collection_app_updater.py
```diff
@@ -429,3 +433,99 @@ def _add_filename_input_field(self, schema: dict[str, Any]) -> dict[str, Any]:
...
+        """
+        script_dir = os.path.dirname(os.path.abspath(__file__))
+        js_path = os.path.join(script_dir, "..", "DataCollectionApp", "schema", "feed_schema.js")
+
+        js_path = os.path.normpath(js_path)
+
+        with open(js_path, "r", encoding="utf-8") as file:
+            js_content = file.read()
+        json_str = js_content.split("=", 1)[1].strip()  # Remove 'feed_schema ='
+        feed_schema = json.loads(json_str)  # Convert to Python dictionary
```
Isgotkowitz:
```suggestion
        feed_schema = json.loads(json_str)
```

RUFAS/data_collection_app_updater.py
```diff
@@ -429,3 +433,99 @@ def _add_filename_input_field(self, schema: dict[str, Any]) -> dict[str, Any]:
...
         return schema
+
+    @staticmethod
+    def gather_feed_data() -> dict[str, Any]:
+        """Gather the user feed data to update."""
+        file_path = os.path.join(os.path.dirname(__file__), "..", "input", "data", "feed", "user_feeds.csv")
+        df = pd.read_csv(file_path)
+        return {
+            "id": df["rufas_id"].tolist(),
+            "name": [f"{name} - {rufas_id}" for name, rufas_id in zip(df["Name"], df["rufas_id"])],
```
Isgotkowitz:
```suggestion
            "name": [f"{name} (description) - {rufas_id}" for name, description, rufas_id in zip(df["Name"], df["feed_description"], df["rufas_id"])],
```

RUFAS/data_collection_app_updater.py
```diff
@@ -429,3 +433,99 @@ def _add_filename_input_field(self, schema: dict[str, Any]) -> dict[str, Any]:
         }
         schema["properties"].update(filename_field)
         return schema
+
+    @staticmethod
+    def gather_feed_data() -> dict[str, Any]:
+        """Gather the user feed data to update."""
+        file_path = os.path.join(os.path.dirname(__file__), "..", "input", "data", "feed", "user_feeds.csv")
```
Isgotkowitz: Instead of putting this in the method, can you please make it a module constant with the same conventions/patterns used at the top of this file?

RUFAS/data_collection_app_updater.py
```diff
@@ -429,3 +433,99 @@ def _add_filename_input_field(self, schema: dict[str, Any]) -> dict[str, Any]:
...
+        Main routine to update the structure of feed schema.
+
+        Parameters
+        ----------
+        user_feed : dict[str, list]
+            The updated user feed data.
+
+        """
+        script_dir = os.path.dirname(os.path.abspath(__file__))
+        js_path = os.path.join(script_dir, "..", "DataCollectionApp", "schema", "feed_schema.js")
```
Isgotkowitz: Same comment about making this file path a module constant at the top of the file.

RUFAS/data_collection_app_updater.py
```diff
@@ -429,3 +433,99 @@ def _add_filename_input_field(self, schema: dict[str, Any]) -> dict[str, Any]:
...
+            The target properties to update the data with the dropdown menu data.
+        dropdown_data : dict[str, Any]
+            The data to be parsed into the schema.
+
+        """
+        for key, prop_value in properties.items():
+            if isinstance(prop_value, dict):
+                prop_value["enum"] = dropdown_data["id"]
+                prop_value.setdefault("options", {})["enum_titles"] = dropdown_data["name"]
+                break
```
Isgotkowitz: Both `modify_items_schema` and `update_first_property_with_enum` rely on side effects to update data, this is an anti-pattern.

Instead they should take in the data to be modified, modify it, and then return the modified data.

tests/test_data_collection_app_updater.py
```diff
@@ -1,5 +1,11 @@
+import json
+import os
+
+import pandas as pd
```
Isgotkowitz, line 3:
```suggestion
```

Isgotkowitz (commented):
Works as expected for the most part, I think with some minor changes this will be good to go.

However, one thing that is good to keep in mind is how this code could potentially be extended. In the original DCA, there was only one input schema that had this dropdown of options, but this is no longer the case -- the Crop Configuration input also has an input for RuFaS Feed IDs that would benefit from the same dropdown. Maybe there is a way that the metadata properties could be modified so that it would be possible to specify for any input that it should be converted to be a dropdown of feed inputs.

This would require a lot of work and isn't strictly necessary though, so don't feel like you have to take it on unless you are feeling extra motivated.

tests/test_data_collection_app_updater.py
matthew7838: Decided to go with not importing the function in the tests as it makes the mocking complicated.

matthew7838:
> Works as expected for the most part, I think with some minor changes this will be good to go.
> 
> However, one thing that is good to keep in mind is how this code could potentially be extended. In the original DCA, there was only one input schema that had this dropdown of options, but this is no longer the case -- the Crop Configuration input also has an input for RuFaS Feed IDs that would benefit from the same dropdown. Maybe there is a way that the metadata properties could be modified so that it would be possible to specify for any input that it should be converted to be a dropdown of feed inputs.
> 
> This would require a lot of work and isn't strictly necessary though, so don't feel like you have to take it on unless you are feeling extra motivated.

That's a great idea and I'll create an issue for that. I don't think I can power through that in this PR.

changelog.md
```diff
@@ -100,7 +100,8 @@ v0.9.2
 - [2210](https://github.com/RuminantFarmSystems/MASM/pull/2210) - [minor change] [OutputManager] Refactor tests in test_output_manager.py to all use pytest-mock.
 - [2205](https://github.com/RuminantFarmSystems/MASM/pull/2205) - [minor change] [Manure] Implements base manure Processor class.
 - [2212](https://github.com/RuminantFarmSystems/MASM/pull/2212) - [minor change] [E2E Testing] Updates Feed Storage E2E Testing expected results.
-- [2191](https://github.com/RuminantFarmSystems/MASM/pull/2191) - [minor change] [Documentation] Regenerate code documentation and structure.
+- [2221](https://github.com/RuminantFarmSystems/MASM/pull/2221) - [minor change] [DCA] Added dropdowns for feed storage options.
+
```
```suggestion
- [2191](https://github.com/RuminantFarmSystems/MASM/pull/2191) - [minor change] [Documentation] Regenerate code documentation and structure.
- [2221](https://github.com/RuminantFarmSystems/MASM/pull/2221) - [minor change] [DCA] Added dropdowns for feed storage options.

```

RUFAS/data_collection_app_updater.py
```diff
@@ -429,3 +443,112 @@ def _add_filename_input_field(self, schema: dict[str, Any]) -> dict[str, Any]:
...
+
+        Parameters
+        ----------
+        data : dict[str, Any]
+            The schema structure.
+        dropdown_data : dict[str, list]
+            The updated content in the dropdown.
+        skip_first : bool, default=True
+            Boolean indicators to help skip the first "properties" field
+
```
Isgotkowitz, line 493:
```suggestion

    Returns
    --------
    dict[str, Any]
        Input schema with updated dropdowns.
 
```

RUFAS/data_collection_app_updater.py
```diff
@@ -429,3 +443,112 @@ def _add_filename_input_field(self, schema: dict[str, Any]) -> dict[str, Any]:
...
+        """
+        new_properties = copy.deepcopy(properties)
+        for key, prop_value in new_properties.items():
+            if isinstance(prop_value, dict):
+                prop_value["enum"] = dropdown_data["id"]
+                prop_value["options"] = {**prop_value.get("options", {}), "enum_titles": dropdown_data["name"]}
+                break
+
+        return new_properties
```
Isgotkowitz:
```suggestion
        for prop_value in properties.values():
            if isinstance(prop_value, dict):
                prop_value["enum"] = dropdown_data["id"]
                prop_value["options"] = {**prop_value.get("options", {}), "enum_titles": dropdown_data["name"]}
                break

        return properties
```
The `deepcopy` isn't necessary (as far as I can tell) and should be avoided if at all possible because it is very resource intensive. Same for the deepcopy in `modify_items_schema`

tests/test_data_collection_app_updater.py
```diff
@@ -12,6 +16,176 @@ def dca_updater() -> DataCollectionAppUpdater:
...
+    input_schema: dict[str, Any],
+    expected_enum_location: list[str],
+    expected_enum_titles_location: list[str],
+    sample_dropdown_data: dict[str, list[Any]],
+) -> None:
+    """
+    Test modify_items_schema with multiple schema structures.
+
+    This updated test captures the returned modified schema (instead of expecting an in-place mutation)
+    and then traverses the returned dictionary along the expected paths.
```
Isgotkowitz:
```suggestion
    Test modify_items_schema with multiple schema structures.
```
I don't think this comment is necessary because when the code is merged and someone reads this, they aren't gonna know that it was initially implemented differently.

tests/test_data_collection_app_updater.py
```diff
@@ -12,6 +16,176 @@ def dca_updater() -> DataCollectionAppUpdater:
...
+
+    updated_schema = processor.modify_items_schema(input_schema, sample_dropdown_data)
+
+    enum_location = updated_schema
+    for key in expected_enum_location:
+        enum_location = enum_location[key]
+
+    enum_titles_location = updated_schema
+    for key in expected_enum_titles_location:
+        enum_titles_location = enum_titles_location[key]
```
Isgotkowitz: Instead of doing this processing, could you please define the expected result for each input schema and then check `updated_schema` with `assert updated_schema == expected_schema`? I think that would make this test a lot easier to understand.

RUFAS/data_collection_app_updater.py
```diff
@@ -25,6 +29,15 @@
...
 INPUT_PLACEHOLDER: str = "null"
 
+"""Directory to the available user feed inputs."""
+USER_FEED_PATH: str = os.path.join(os.path.dirname(__file__), "..", "input", "data", "feed", "user_feeds.csv")
+
+"""Path to all the DCA scripts"""
+SCRIPT_PATH = os.path.dirname(os.path.abspath(__file__))
+
+"""Path to the feed_schema.js schema file."""
+FEED_SCHEMA_PATH = os.path.join(SCRIPT_PATH, "..", "DataCollectionApp", "schema", "feed_schema.js")
```
Isgotkowitz:
```suggestion
"""Directory to the available user feed inputs."""
USER_FEED_PATH: Path = Path("").joinpath("input", "data", "feed", "user_feeds.csv")

"""Path to the feed_schema.js schema file."""
FEED_SCHEMA_PATH: Path = Path("").joinpath("DataCollectionApp", "schema", "feed_schema.js")
```
Using paths instead of strings didn't seem to break the code when I tried it out. I think it would be better to do it like this because it keeps it consistent with how the other module constant Paths are defined.

Also I recommend moving these up to be with the other module constants that end with `_PATH`, so that the paths and placeholders remain distinctly grouped.

Might also be able to get rid of the `os` import.

RUFAS/data_collection_app_updater.py
```diff
@@ -429,3 +443,112 @@ def _add_filename_input_field(self, schema: dict[str, Any]) -> dict[str, Any]:
...
+        }
+
+    def update_feed_schema(self, user_feed: dict[str, list[Any]]) -> None:
+        """
+        Main routine to update the structure of feed schema.
+
+        Parameters
+        ----------
+        user_feed : dict[str, list]
+            The updated user feed data.
```
Isgotkowitz:
```suggestion
            The user feeds input data parsed so that it can be used to setup dropdowns.
```

Isgotkowitz (commented):
This is looking better Matthew, but there are still some improvements that could be made.

Most are minor, but the changelog needs to be fixed and unless it is _really_ necessary `deepcopy` shouldn't be used.

RUFAS/data_collection_app_updater.py
```diff
@@ -429,3 +439,117 @@ def _add_filename_input_field(self, schema: dict[str, Any]) -> dict[str, Any]:
...
+            Modify the schema with dropdowns by updating the corresponding field with updated feed data.
+
+            Parameters
+            ----------
+            data : dict[str, Any]
+                The schema structure.
+            dropdown_data : dict[str, list]
+                The updated content in the dropdown.
+            skip_first : bool, default=True
+                Boolean indicators to help skip the first "properties" field
```
Isgotkowitz:
```suggestion
        Modify the schema with dropdowns by updating the corresponding field with updated feed data.

        Parameters
        ----------
        data : dict[str, Any]
            The schema structure.
        dropdown_data : dict[str, list]
            The updated content in the dropdown.
        skip_first : bool, default=True
            Boolean indicators to help skip the first "properties" field
```

tests/test_data_collection_app_updater.py
```diff
@@ -1,3 +1,7 @@
+import json
+import os
```
Isgotkowitz:
```suggestion
```
If `os` isn't used in the DCA Updater, does it need to be imported to test it?

RUFAS/data_collection_app_updater.py
```diff
@@ -429,3 +439,117 @@ def _add_filename_input_field(self, schema: dict[str, Any]) -> dict[str, Any]:
...
+        dropdown_data : dict[str, Any]
+            The data containing the dropdown values to insert.
+
+        Returns
+        -------
+        dict[Any, Any]
+            A new dictionary with the update applied.
+
+        """
+        new_properties = copy.copy(properties)
```
Isgotkowitz, line 548: I appreciate you taking the deepcopy out, but I am still unsure why any copying is necessary here. Can you please explain why the properties have to be copied before they are modified?

input/metadata/task_manager_metadata.json
```diff
@@ -3,7 +3,7 @@
     "tasks": {
       "title": "Task manager data",
       "description": "Configuration file for general simulation parameters.",
-      "path": "input/data/tasks/default_task.json",
+      "path": "input/data/tasks/dca_update_task.json",
```
Isgotkowitz: Don't forget to revert this before merging

RUFAS/data_collection_app_updater.py
matthew7838: Otherwise, this function will still rely on side effects from the dictionary. Isolating the original dictionary from side effects also makes the code easier to track when things go downhill.

Isgotkowitz: Gotcha, I'm sorry I should've reminded myself of what exactly constitutes a side effect before commenting. I had to have a chat with DeepSeek to have it correct me, and I think you were right to use `deepcopy` here instead of `copy`. Because this is a nested dictionary, this function is still relying on side effects because the dictionaries within `properties` is not being copied.

The reason I cautioned against using `deepcopy` is that it can easily slow down RuFaS if it is used in the wrong place (unfortunately I know from experience). But because this is such a specialized part of the codebase and isn't used at all in actual RuFaS simulations, I think it would be ok to use it here. Ultimately though I think it will be alright whichever way you choose to go, avoiding side effects is ideal but it's not the end of the world when the code is well written.

Isgotkowitz (approved):

Isgotkowitz:
Thank you for addressing my feedback Matthew!

RUFAS/data_collection_app_updater.py
```diff
@@ -429,3 +439,117 @@ def _add_filename_input_field(self, schema: dict[str, Any]) -> dict[str, Any]:
...
+        skip_first : bool, default=True
+            Boolean indicators to help skip the first "properties" field
+
+        Returns
+        --------
+        dict[str, Any]
+            Input schema with updated dropdowns.
+
+        """
+        new_data = copy.copy(data)
```
allisterakun:
```suggestion
        new_data = data.copy()
```
I believe we can just use the built-in copy method for dictionary, so that we don't have to import copy to use the generic copy method

allisterakun (approved):
LGTM! Tested out and it works as expected. Confirmed that `data_collection_app_updater.py` is 100% covered by unit test.
Great job, Matthew!

Issue #1941, opened 2024-08-19
[Report Generator] Validate Report Filter Content


ew3361zh:
**Is your feature request related to a problem? Please describe.**
Because report filters are entirely user-created, there is a lot of room for error - misspellings, mistypings, etc. Often when a filter is entered incorrectly, it can be challenging to diagnose by the user. 

**Describe the solution you'd like**
This report filter error diagnosis should be made easier for the user with better error handling and validation for RG filters. Validate the filter attributes user and the types for each filter attribute value.

**Describe alternatives you've considered**
N/A

**Additional context**
This is best suited for the next iteration of RG and so will be immediately blocked.

**Target Deadline**
Completed with the next iteration of RG.

allisterakun:
Grooming session:
- The keys in the filter JSON files are valid (no random unknown keys)
  - Generate a warning for unknown keys
- The required keys are all present
  - Generate an error if a required key is missing
- The values for each key have the correct type and are valid

This functionality should be part of the OutputManager and be called before the simulation starts.

ew3361zh:
Just providing a non-judgemental use-case here for context.
User got an error trying to figure out why a filter wasn't working:

<img width="584" alt="Image" src="https://github.com/user-attachments/assets/f34d73be-ad9d-4692-80c8-8ffb91d90dd9" />

The error message:
<img width="418" alt="Image" src="https://github.com/user-attachments/assets/3fe017b1-edac-450f-8f53-870d5d3f1fed" />

The resolution was that the RG key for listing your pattern matches is `filters`, not `filter`.

This message would be much more helpful to get before the full simulation runs.

ew3361zh:
Should look at validating other output filters as well (i.e. GG, OM's CSV and JSON filters).

allisterakun:
Special case:
In PR #2291, we are allowing users to pass in a date format string (e.g., "%m-%d-%Y"); this input also needs to be validated to make sure it's a valid date format for both `strftime()` and `strptime()` with meaningful % directives. This validation is currently in "util.py" as a function called "validate_date_format()":
```python
    @staticmethod
    def validate_date_format(date_format: str) -> bool:
        """
        Checks if date_format is a valid Python datetime format for both strftime() and strptime().
        Parameters
        ----------
        date_format : str
            The date format to be validated.
        Returns
        -------
        bool
            True if date_format is a valid Python datetime format; False otherwise.
        """
[...]

PR #2225, opened 2025-02-10
[OM] Adds ability for user to control significant digits of output

### LINKED ISSUE #2139: [OM] Add ability to control the number of significant digits in output
It would be a helpful feature for users and maintainers to control the number of significant digits in output variables.

## Status

When variables are added to the OM with `add_variable()` calls, there is no control over the number of significant digits that are included. The OM holds however many are given to it, and reports out that number of significant digits when creating report, JSON, and CSV outputs.

## Intention

Give users and maintainers control over how many significant digits are included in the output, so that the amount of information coming out of RuFaS can be more precisely controlled.

## Request

There are a couple possible ways that control over significant digits in outputs could be given:
1. Add an argument to `add_variable` that tells OM how many significant digits to keep from a value being passed to it.
2. Have users set a target number of significant digits they want in the config input file, and keep that number of significant digits every time that `add_variable` is called in a simulation.
3. Add a field to filter files which lets users specify how many significant digits they want included in the output files generated using that filter.

One or more of the above options may be implemented.

ew3361zh:
A new output filter field is added for users to specify if they want the data rounded in whatever format (json, csv, report, graph) their output filter is in.

### Context

Issue(s) closed by this pull request: closes #2139

### What

Adds a new optional field `data_significant_digits` to output filters users can use to set how many significant digits out they'd like the data to be presented in the reports/csvs/jsons/graphs.

### Why

Creates cleaner looking reports.

### How

Added a new util function `round_numeric_values_in_dict()` to execute this rounding function on numerical data and to ignore non-numerical data.
This util function is called in OM in the route-save-functions and separately in ReportGenerator so that any data being aggregated is not affected.
**NOTE:** Within OM, I added a tracker for RG functionality because of the edge case where someone may be using the `data_significant_digits` with one report and then cross-reference that report in another report. Results for this could vary slightly from expected results of aggregating this data without rounding the significant digits. If both of those cases are true separately or in the same filter file across all report generation done in the simulation, a warning will be logged to OM. I'm not sure much more could be done to prevent this.

### Test plan

Created separate filters for reports, graphs, csvs, and jsons and incorporated the new filter file field and examined the output. Outputs with this new field present rounded numerical data to the appropriate digit.

Here is an example filter file I basically ran for each of the output types except for GG where I added `type`: `plot`:
```
{
    "name": "ManRequest-Farm",
    "filters": [
      "ManureManager._record_manure_request_results.on_farm_manure.*"
    ],
    "variables": [".*"],
    "data_significant_digits": 2,
  }
```

### DIFF
--- a/RUFAS/graph_generator.py
+++ b/RUFAS/graph_generator.py
@@ -455,6 +455,7 @@ class GraphGenerator:
                 "is_aggregated_report_data",
                 "use_calendar_dates",
                 "date_format",
+                "data_significant_digits",
             ]
         )
         graph_filter_validation_logs: list[dict[str, str | dict[str, str]]] = []
@@ -481,7 +482,7 @@ class GraphGenerator:
                 graph_filter_validation_logs.append(
                     {
                         "warning": f"Can't plot data for {filter_key}",
-                        "message": f"Invalid filter file key '{filter_key}' does not match"
+                        "message": f"Invalid filter file key '{filter_key}' does not match "
                         "any optional keys. "
                         f"Please see Graph Generator wiki for a list of valid filter"
                         "keys.",
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -1460,6 +1460,9 @@ class OutputManager(object):
             f"exclude_info_maps flag set to {exclude_info_maps}",
             info_map,
         )
+        has_cross_references = False
+        has_data_significant_digits = False
+        cross_ref_significant_digits_warning = False
         list_of_filter_files = self._list_filter_files_in_dir(filters_dir_path)
         report_generator = ReportGenerator(self.time)
         if self.chunkification:
@@ -1490,6 +1493,11 @@ class OutputManager(object):
                     )
                     continue
 
+                if "cross_references" in filter_content:
+                    has_cross_references = True
+                if "data_significant_digits" in filter_content:
+                    has_data_significant_digits = True
+
                 filtered_pool: dict[str, OutputManager.pool_element_type] = {}
                 if "filters" in filter_content.keys():
                     filtered_pool = self.filter_variables_pool(filter_content)
@@ -1516,6 +1524,14 @@ class OutputManager(object):
                     )
             report_file_path = report_dir / self.generate_file_name(f"report_{filter_file}", "csv")
             if report_generator.reports:
+                if has_cross_references and has_data_significant_digits and not cross_ref_significant_digits_warning:
+                    self.add_warning(
+                        "Report Generation Warning",
+                        "Reports generated have both cross references and data significant digits. Results may be "
+                        "affected.",
+                        info_map,
+                    )
+                    cross_ref_significant_digits_warning = True
                 self.create_directory(report_dir)
                 self._dict_to_file_csv(report_generator.reports, report_file_path)
                 report_generator.clear_reports()
@@ -1538,7 +1554,20 @@ class OutputManager(object):
             "class": self.__class__.__name__,
             "function": self._route_save_functions.__name__,
         }
-
+        if "data_significant_digits" in filter_content:
+            filtered_pool = {
+                key: (
+                    Utility.round_numeric_values_in_dict(value, filter_content["data_significant_digits"])
+                    if isinstance(value, dict)
+                    else value
+                )
+                for key, value in filtered_pool.items()
+            }
+            self.add_log(
+                "Rounding Values",
+                f"Rounded values to {filter_content['data_significant_digits']} significant digits",
+                info_map,
+            )
         is_json = filter_file.startswith(self._filter_prefixes.get("json", "Better than a key error."))
         if is_json and self.is_first_post_processing:
             self.create_directory(json_dir)
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -1540,12 +1540,18 @@ class OutputManager(object):
         }
         if "data_significant_digits" in filter_content:
             filtered_pool = {
-                key: Utility.round_numeric_values_in_dict(value, filter_content["data_significant_digits"])
-                if isinstance(value, dict) else value
+                key: (
+                    Utility.round_numeric_values_in_dict(value, filter_content["data_significant_digits"])
+                    if isinstance(value, dict)
+                    else value
+                )
                 for key, value in filtered_pool.items()
             }
-            self.add_log("Rounding Values",
-                         f"Rounded values to {filter_content['data_significant_digits']} significant digits", info_map)
+            self.add_log(
+                "Rounding Values",
+                f"Rounded values to {filter_content['data_significant_digits']} significant digits",
+                info_map,
+            )
         is_json = filter_file.startswith(self._filter_prefixes.get("json", "Better than a key error."))
         if is_json and self.is_first_post_processing:
             self.create_directory(json_dir)
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -1460,6 +1460,10 @@ class OutputManager(object):
             f"exclude_info_maps flag set to {exclude_info_maps}",
             info_map,
         )
+        has_cross_references = False
+        has_data_significant_digits = False
+        cross_ref_reports: list[str] = []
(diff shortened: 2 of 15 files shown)


tests/test_graph_generator.py
```diff
@@ -1,16 +1,12 @@
...
 from pathlib import Path
 from typing import Any, Dict, List
-from unittest.mock import patch
-
 import numpy as np
 import pytest
 from freezegun import freeze_time
 from matplotlib import pyplot as plt
 import matplotlib.dates as mdates
-from mock.mock import MagicMock
```
ew3361zh, line 11: most of the changes to this file have to do with switching from the unittest.mock patch and mock.mock MagicMock imports to the pytest versions of each of these which reduces mypy errors.

RUFAS/output_manager.py
```diff
@@ -1516,6 +1524,14 @@ def save_results(  # noqa: C901
...
             report_file_path = report_dir / self.generate_file_name(f"report_{filter_file}", "csv")
             if report_generator.reports:
+                if has_cross_references and has_data_significant_digits and not cross_ref_significant_digits_warning:
+                    self.add_warning(
+                        "Report Generation Warning",
+                        "Reports generated have both cross references and data significant digits. Results may be "
+                        "affected.",
+                        info_map,
+                    )
+                    cross_ref_significant_digits_warning = True
```
Isgotkowitz: Instead of capturing this warning once and recording it, what would you think about recording it each time there is a report that both cross-references and limits the significant digits? In each warning, it would be good to include the report name and the name of the report file so that users would know exactly where they're getting this warning from. This way they can better assess whether or not it is something they should actually be worried about.

RUFAS/report_generator.py
```diff
@@ -457,8 +461,8 @@ def _perform_aggregations(
             report_data: dict[str, dict[str, list[Any]]] = filtered_pool
             if filter_content.get("display_units", False):
                 report_data = self._add_var_units(report_data)
-            report_data = {key: report_data[key]["values"] for key in report_data.keys()}
-            if not all(report_data[key] for key in report_data.keys()):
+            report_data = {key: report_data[key]["values"] for key in report_data}
+            if not all(report_data[key] for key in report_data):
```
Isgotkowitz, line 465: I'm curious why you decided to change this.

It is just my personal preference, but I like `for key in dictionary.keys()` better than `for key in dictionary` even though they both work fine. I think the former is more explicit, and makes it clearer that we intend to be iterating through a dictionary and more specifically the keys of the dictionary.

RUFAS/util.py
```diff
@@ -712,3 +712,29 @@ def validate_fractions(fractions: List[float]) -> bool:
...
+
+        """
+        return {
+            key: (
+                [round(x, significant_digits) for x in value]
+                if isinstance(value, list) and all(isinstance(x, (float, int)) for x in value)
+                else value
+            )
+            for key, value in data.items()
+        }
```
Isgotkowitz, line 759: This function has some interesting characteristics that I think are worth noting explicitly:
- If `significant_digits` is `None`, floats are converted to ints.
- If `significant_digits` is less than 0, it rounds to the nearest multiple of 10, 100, 1000, etc.

RUFAS/output_manager.py
```diff
@@ -1538,7 +1554,20 @@ def _route_save_functions(
             "class": self.__class__.__name__,
             "function": self._route_save_functions.__name__,
         }
-
+        if "data_significant_digits" in filter_content:
```
Isgotkowitz, line 1561: What about adding a check at the beginning of this if statement that checks to make sure that the type of "data_significant_digits" is either an int or None?

Isgotkowitz (approved):
Works as expected, nice job Niko. This is a very nifty feature.

I left some minor comments, apply them as you see fit

RUFAS/report_generator.py
ew3361zh: From what I read, yes, they both function the same but there is an additional cost of creating a view object when you use the .keys() and is about 10% slower on average. I think with newer versions of Python, dicts have been optimized to search through keys without creating the view object and it's the preferred approach.

I do see your point about the loss of clarity of what we're looping through but we do use the report_data[key]["values"] right before it so hopefully that would make the type clear from context.

RUFAS/util.py
ew3361zh: Will make a note, thanks for bringing this up.

RUFAS/output_manager.py
ew3361zh: I was thinking a good approach might be to address this in issue #1941 - I left a note that GG filters should be validated as well.

RUFAS/output_manager.py
ew3361zh: @Isgotkowitz I made an update that addresses this - would you mind taking a quick look to see if it's what you had in mind?

Isgotkowitz: Sure!

Isgotkowitz: I think it does the most important thing, which is to let the user know that their results may be affected and is specific about which results.

One thing I might do differently would be to make a separate warning for each result that is affected, instead of one report listing all the results that may have been affected. But that would be my personal preference, I don't think there is necessarily one right or wrong way to do things

RUFAS/report_generator.py
Isgotkowitz: Interesting! I had no idea, thanks for bringing this up

RUFAS/output_manager.py
Isgotkowitz: Good idea

RUFAS/graph_generator.py
```diff
@@ -481,7 +482,7 @@ def _validate_graph_filter(
                 graph_filter_validation_logs.append(
                     {
                         "warning": f"Can't plot data for {filter_key}",
-                        "message": f"Invalid filter file key '{filter_key}' does not match"
+                        "message": f"Invalid filter file key '{filter_key}' does not match "
```
matthew7838, line 485: Any reason for this extra space?

RUFAS/util.py
```diff
@@ -712,3 +712,34 @@ def validate_fractions(fractions: List[float]) -> bool:
...
+        Returns
+        -------
+        dict[str, any]
+            The dictionary with numeric values rounded to the specified number of significant digits.
+
+        Notes
+        -----
+        If significant_digits is None, floats are converted to ints.
+        If significant_digits is less than 0, it rounds to the nearest multiple of 10, 100, 1000, etc.
+        If significant_digits is 0, it rounds to the nearest integer.
```
matthew7838: I am a little confused, in what case would negative significant_digits make sense? I think it would be a nice clarification to have an example here and also have the case tested in the tests.

matthew7838 (commented):
Look great, Niko. Left a question in the comment.

RUFAS/output_manager.py
ew3361zh: I thought about that - part of the issue was noting which specific reports had the significant digits chopped and then connecting that back to how the report was referred to in the cross-reference section of the report referencing it. It got into a little bit of dicey string parsing and matching and I didn't feel confident it would be accurate enough unless I really put some effort in. This seemed like a simple and comprehensive enough substitute for that without taking too much resources to get it to log the necessary info.

RUFAS/graph_generator.py
ew3361zh: It came up when I was testing the message - the next line didn't have a space either so it was combining the last word of this sentence with the first word of the next sentence into a long nonsense word.

RUFAS/util.py
ew3361zh: I don't think anyone would actually use it but it's just the `ndigits` argument passed to the python built-in round() function. So if you give set significant digits to `-1`, it will start rounding to the left of the decimal instead of the right. I will try to spell that out a little more since python's documentation is a little thin on what happens there. I'm not sure anyone would use this feature but good to know full behavior.

matthew7838: Would that make a case for adding an error or warning?

ew3361zh: I'm not sure I understand. It's the default behavior of the built-in round function.

matthew7838 (approved):

PR #2230, opened 2025-02-14
[Manure] Fix manure metadata

### LINKED ISSUE #2204: [Manure] Missing Metadata in Bedding Configurations
## Expected Behavior

Each bedding configuration variable should include metadata that specifies:
- Units (e.g., kg, m³, percentage)
- Description (clear explanation of what the variable represents)
- Range of possible inputs (minimum and maximum acceptable values)

## Current Behavior

- The current bedding_configs section lacks metadata for most numerical variables.
- The variables bedding_mass_per_day, bedding_density, bedding_dry_matter_content, bedding_cleaned_fraction, bedding_carbon_fraction, bedding_phosphorus_content, and sand_removal_efficiency do not include:
  - Units of measurement.
  - Explanations of what each variable represents.
  - Acceptable input ranges.

Image

## Possible Solution

- Add metadata to each variable, specifying:
  - Units (e.g., bedding mass per day in kg/day, density in kg/m³, etc.).
  - Clear descriptions explaining the role of each variable.
  - Input validation (e.g., minimum and maximum values).

## Steps to Reproduce

1. Open the bedding configurations schema.
2. Observe the missing metadata for key numerical fields.
3. Compare with other well-documented sections of the model.

## Context (Environment)

This issue makes it difficult for users to interpret and correctly configure bedding parameters, potentially leading to incorrect inputs and inconsistent simulation results.

## Detailed Description

The current implementation of bedding_configs (as shown in the attached image) defines the structure of the configuration but lacks essential metadata for several variables. Without specifying units, descriptions, and valid input ranges, users may struggle to provide accurate values, leading to potential misconfigurations in RuFaS simulations.

## Possible Implementation

- Add a description field for each numerical variable.
- Specify units where applicable.
- Define valid input ranges for parameters like density, dry matter content, and carbon fraction.

elle-andreen:
### Context
Issue(s) closed by this pull request: closes #2204. 

Metadata descriptions should be complete and clear in order for users to understand the definition and context of each variable. Metadata for many Manure Module variables was missing, incomplete or out of date.

### What
- Added and updated descriptions and range values where necessary. For variables that can be utilized multiple times (e.g. if you define 3 handlers, `cleanings_per_day` can be defined for each handler) I have tried to ensure the description reflects that. E.g.,   `cleanings_per_day` for _this specific_ handler, not `cleanings_per_day` for _all_ handlers. 
- Modified the value of `"cover"` for slurry storage underfloor in `manure_management.json` to be `no cover`, instead of `crust`; generally a crust cannot form in an underfloor manure storage as manure is continually falling from above and disturbing any kind of surface crust.
- I have **not** modified the description of variables that are likely to be removed during the manure refresh (e.g., removal efficiencies for manure treatments).

### Why
Having complete descriptions, ranges and units in the metadata is essential to enabling users to understand input values and context. 

### Test plan
This PR is strictly related to metadata and default values. I ran the simulation to confirm the new range values did not cause any validation issues. End to end testing was run and passed successfully with no output changes.


input/metadata/properties/default.json
```diff
@@ -5149,29 +5149,48 @@
...
-          "type": "number"
+          "type": "number",
+          "description": "The bedding phosphorus content as a percent of total mass, on a wet weight basis (unitless).",
+          "minimum": 0,
+          "maximum": 1
         },
         "sand_removal_efficiency": {
-          "type": "number"
+          "type": "number",
+          "description": "The percentage of the fresh sand added each day that is assumed to be removed by a sand separator, settling lane, etc. immediately following manure removal from the barn (unitless)",
```
krb78: This is only the % of fresh sand added each day that is removed? I thought this would be the % removal of any type (fresh or added a while ago) of sand that ends up mixed with manure.

input/metadata/properties/default.json
```diff
@@ -5185,28 +5204,32 @@
...
         "manure_handler_type": {
           "type": "string",
-          "description": "Manure Handling general type -- Method for cleaning barn alleyways.",
+          "description": "The type of manure handler (method for cleaning barn alleyways).",
           "pattern": "^(flush system|alley scraper|manual scraping|tillage|harrowing)$"
         },
         "cleaning_water_use_rate": {
-          "type": "number"
+          "type": "number",
+          "description": "The rate (L per animal per day) of fresh, non-recycled water use by a specified manure handler (liters)",
```
krb78: If this is only fresh water, then how do we get total water if later on we only ask for % recycled? (I know this is just algebra but it seems overly complicated given the other inputs). It seems like this should be total water in the flush and then later on "cleaning_water_recycle_fraction" definition should stay and the remaining part is the clean water.

input/metadata/properties/default.json
```diff
@@ -5223,29 +5246,50 @@
...
+          "description": "The dry matter content (% DM) of the manure solids produced by a specified solid liquid separator (unitless).",
+          "minimum": 0.0,
+          "maximum": 1.0
         },
         "total_solids_removal_efficiency_for_separator": {
-          "type": "number"
+          "type": "number",
+          "description": "For a specified solid liquid separator, the percentage of total solids removed from incoming manure into the separated manure solids fraction (unitless).",
+          "minimum": 0.0,
+          "maximum": 1.0
```
krb78, line 5262: just out of curiosity, is this maximum actually 1? I've never gotten rufas to successfully run that high. It gives me an error saying that there isn't any manure.

krb78 (commented):
Just a few minor comments, Elle! Looks good :)

input/metadata/properties/default.json
```diff
@@ -5149,29 +5149,48 @@
...
         "bedding_density": {
-          "type": "number"
+          "type": "number",
+          "description": "The density of the bedding on a wet weight basis (kg/m3)",
+          "minimum": 0
         },
         "bedding_dry_matter_content": {
-          "type": "number"
+          "type": "number",
+          "description": "The percent dry matter of the bedding (unitless).",
```
tomhuhh: Instead of 'percent', which usually refers to, for example, 25%, I would use 'decimal fraction' for better clarity.

input/metadata/properties/default.json
```diff
@@ -5149,29 +5149,48 @@
...
+          "minimum": 0,
+          "maximum": 1
         },
         "bedding_cleaned_fraction": {
-          "type": "number"
+          "type": "number",
+          "description": "The percentage of the fresh bedding added daily that is removed from the housing area via cleaning or other processes (unitless).",
```
tomhuhh: Instead of 'percentage', which usually refers to, for example, 25%, I would use 'decimal fraction' for better clarity.

input/data/manure/manure_management.json
```diff
@@ -263,7 +263,7 @@
       "phosphorus_removal_efficiency_for_treatment": 0.0,
       "potassium_removal_efficiency_for_treatment": 0.0,
       "storage_time_period": 120,
-      "manure_cover": "crust"
+      "manure_cover": "no cover"
```
tomhuhh, line 266: Nice catch!

input/metadata/properties/default.json
```diff
@@ -5318,18 +5362,18 @@
...
         "last_compost_turning_or_addition": {
           "type": "number",
-          "description": "The number of days since the last turning or addition event (day)",
+          "description": "The number of days since the last turning or addition event for a specified compost storage (day)",
           "default": 1
         },
         "manure_cover": {
           "type": "string",
-          "description": "Indicates the presence or absence of a cover in the manure treatment or storage system. A cover referes to a human made cover where as a crust is naturally formed. A cover and flare refers to a storage system that is covered to capture and then destroy the methane produced with a flare. For the anaerobic diesgtion treatment, this property should be set to N/A.",
+          "description": "The type of cover present on a specified manure treatment or storage. A cover referes to a human made cover where as a crust is naturally formed. A cover and flare refers to a storage system that is covered to capture and then destroy the methane produced with a flare. For the anaerobic diesgtion treatment, this property should be set to N/A.",
```
tomhuhh: A typo here: 'A cover `refers` to a human made cover' instead of 'A cover `referes` to a human made cover'.

input/metadata/properties/default.json
```diff
@@ -5260,7 +5304,7 @@
         },
         "manure_treatment_type": {
           "type": "string",
-          "description": "Manure Treatment Methods -- Select the Manure Treatment Methods.",
+          "description": "Manure storage type.",
```
tomhuhh: Maybe 'manure `treatment and` storage type'? 'anaerobic digestion' in the list seems like a treatment type.

input/metadata/properties/default.json
elle-andreen: Thanks for catching this - you're correct, this value is actually **total** cleaning water regardless of source (recycled or fresh). Will fix this!

input/metadata/properties/default.json
elle-andreen: Good question. Purely theoretically, the max should be 1, but the simulation indeed crashes above ~0.95% TS removal efficiency unless you increase the percent_dry_solids value. It seems like if you remove too much solids, the volume of the wet mass removed becomes greater than the volume of incoming manure and results in a 0 or negative value for volume of the liquid fraction. When I do this calculation by hand though, if you set TS removal efficiency to 1.0 and leave percent dry solids at 0.35, that shouldn't be happening. It seems like this is a bug albeit not an urgent one for now, I will create an issue so we can check on it in the future. I think it could become more of a problem in the future with chaining multiple separators.

input/metadata/properties/default.json
```diff
@@ -5119,19 +5119,19 @@
         },
         "bedding_type": {
```
allisterakun, line 5120: I'd suggest renaming this variable to avoid confusion, as it is the user-defined name of the bedding configuration rather than the actual type.

input/metadata/properties/default.json
```diff
@@ -5149,29 +5149,48 @@
         },
```
allisterakun, line 5149: For all the processor detail configurations, do you think it's necessary to address in the "name" field that it will be used as the unique identifier for specifying processor configuration for each manure management scenario? 
And maybe also mention that there should be no duplicate names across all processors, no matter the type. (like a bedding config and a separator config cannot share the same name)

input/metadata/properties/default.json
```diff
@@ -5223,29 +5246,50 @@
...
         "manure_separator_type": {
           "type": "string",
-          "description": "Manure Separator Type -- The type of solid-liquid separator equipment to separate coarse fibrous solids/sand. ",
+          "description": "The type of mechanical solid-liquid separator.",
```
allisterakun, line 5249: Are we only having solid-liquid separators?

allisterakun (approved):
Generally, it looks good to me! Left some minor comments.

input/metadata/properties/default.json
elle-andreen: I completely agree with this suggestion but renaming it would require changing the variable name throughout the codebase, which, other than being outside the scope of this PR I think, is also outside my confidence in modifying that many use instances of this variable 😅 I think this is something that can be folded into the manure refresh however as the input structure will be changing completely.

input/metadata/properties/default.json
elle-andreen: At this time RuFaS doesn't track bedding accumulation in pens, I believe that is partly why the bedding_cleaned_fraction is set to 1 in manure_management.json. Generally pretty much all bedding ends up leaving the barn as part of the manure in reality, and that just happens in a single day in RuFaS vs. over a few days in reality. The literal way this works in the code is: total sand in manure post-sand removal = sand bedding mass per day x no. animals in pen x sand removal efficiency

input/metadata/properties/default.json
elle-andreen: Do you mean **mechanical** solid liquid separation, or something else? I specified mechanical here for now because the current separator method is assumed to be by mechanical separation (e.g. separating manure rapidly via machine), and we shouldn't be entering separation efficiencies here from other separators like settling basins for example. Non-mechanical separators often have GHG emissions, which are not accounted for with the current separator methods.

allisterakun: I remembered that we were talking about having sand separators at some point. I guess this technically would also fall under the solid-liquid separators.

tomhuhh (approved):
Looks great to me, thanks Elle!

PR #2233, opened 2025-02-17
[Manure] Implement base `Digester` class

### LINKED ISSUE #2099: [Manure] Implement base `Digester` class
Implement the base `Digester` class. Use the "Developer Guidelines" in #2092.

Isgotkowitz:
Implements the base manure `Digester` class.

### Context

Issue(s) closed by this pull request: closes #2099

### What

Adds the base class and test module for manure Digesters.

Also added some attributes to the base `Processor` class that I realized in hindsight would be needed by all child classes.

### Why

Digesters are one of the four child types of `Processor`, so this has to be implemented before the final digester child (Anaerobic Digester) is implemented.

### How

Creates a `digester/` directory with a `Digester` class that inherits from `Processor`. It does not implement any logic because in the near future there will only be 1 child digester class (`AnaerobicDigester`) and it is unlikely to share any logic with any other type of `Digester` that might be added to RuFaS at a later date. So all logic unique to `AnaerobicDigester` will go into that class specifically. This strategy was suggested by @elle-andreen.

### Test plan

Added a test to make sure that `Digester` cannot be initialized on its own. Other than that, there isn't really anything to test.

### DIFF
--- a/RUFAS/biophysical/manure/processor.py
+++ b/RUFAS/biophysical/manure/processor.py
@@ -3,6 +3,7 @@ from abc import ABC, abstractmethod
 from RUFAS.current_day_conditions import CurrentDayConditions
 from RUFAS.data_structures.animal_to_manure_connection import ManureStream
 from RUFAS.time import Time
+from RUFAS.output_manager import OutputManager
 
 
 class Processor(ABC):
@@ -11,13 +12,19 @@ class Processor(ABC):
 
     Parameters
     ----------
+    name : str
+        Unique identifier of the processor.
     is_housing_emissions_calculator : bool
         Indicates if a Processor calculates housing emissions.
 
     Attributes
     ----------
+    name : str
+        Unique identifier of the processor used to label outputs.
     is_housing_emissions_calculator : bool
         If true, processor will only accept ManureStreams with non-None PenManureData, if false then vice versa.
+    _om : OutputManager
+        Instance of the OutputManager.
 
     Methods
     -------
@@ -28,9 +35,11 @@ class Processor(ABC):
 
     """
 
-    def __init__(self, is_housing_emissions_calculator: bool) -> None:
+    def __init__(self, name: str, is_housing_emissions_calculator: bool) -> None:
         """Initializes a new Processor."""
+        self.name = name
         self.is_housing_emissions_calculator = is_housing_emissions_calculator
+        self._om = OutputManager()
 
     @abstractmethod
     def receive_manure(self, manure: ManureStream) -> None:
--- a/tests/test_biophysical/test_manure/test_digester/test_digester.py
+++ b/tests/test_biophysical/test_manure/test_digester/test_digester.py
@@ -1,0 +1,8 @@
+import pytest
+from RUFAS.biophysical.manure.digester.digester import Digester
+
+
+def test_processor_init_error() -> None:
+    """Test that base Digester class throws appropriate error when initialized."""
+    with pytest.raises(TypeError):
+        Digester(name="test digester", is_housing_emissions_calculator=False)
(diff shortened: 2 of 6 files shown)


tests/test_biophysical/test_manure/test_digester/test_digester.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+from RUFAS.biophysical.manure.digester.digester import Digester
+
+
+def test_processor_init_error() -> None:
+    """Test that base Digester class throws appropriate error when initialized."""
+    with pytest.raises(TypeError):
+        Digester(name="test digester", is_housing_emissions_calculator=False)
```
Isgotkowitz, line 8: This adds another Mypy error, because it doesn't like that we are trying to instantiate a class without the abstract methods implemented, even though that is the whole point of the test 🙄

tests/test_biophysical/test_manure/test_processor.py
```diff
@@ -5,10 +5,10 @@
 def test_processor_init_error() -> None:
     """Test that base Processor class throws appropriate error when initialized."""
     with pytest.raises(TypeError):
-        Processor(is_housing_emissions_calculator=True)
+        Processor(name="test processor", is_housing_emissions_calculator=True)
 
 
-# TODO: test with a child or grandchild of Processor, either in #2098 or #2099
+# TODO: test with a grandchild of Processor in #2102, #2103, #2104, or #2105
```
Isgotkowitz, line 11: It is going to have to be a grandchild of Processor that tests these methods, I will do so when I work on one of the issues listed here.

RUFAS/biophysical/manure/processor.py
```diff
@@ -28,9 +35,11 @@ class Processor(ABC):
 
     """
 
-    def __init__(self, is_housing_emissions_calculator: bool) -> None:
+    def __init__(self, name: str, is_housing_emissions_calculator: bool) -> None:
         """Initializes a new Processor."""
+        self.name = name
         self.is_housing_emissions_calculator = is_housing_emissions_calculator
+        self._om = OutputManager()
```
ew3361zh, line 42: Curious why you have the om attribute as private? I don't believe we're doing that anywhere else in the model since OM is a Singleton.

ew3361zh (approved):
Looks good!

RUFAS/biophysical/manure/processor.py
Isgotkowitz: There really isn't any functional reason for this, you're correct because it is a singleton it doesn't matter which OM instance is used when recording data or logs.

But I imagine that if I saw one `Processor` or the refreshed `ManureManager` using another `Processor`'s OM instance to record data I think that would be really weird and confusing. There is nothing about `Processor` that makes me think it should ever be another entity's proxy/gateway/interface to OutputManager, and making the OM attribute private makes that really explicit.

RUFAS/biophysical/manure/processor.py
```diff
@@ -28,9 +35,11 @@ class Processor(ABC):
...
 
     @abstractmethod
     def receive_manure(self, manure: ManureStream) -> None:
```
ew3361zh, line 45: As I'm working with the Separator implementation, I'm wondering if you wanted to add some safeguards around receive_manure() receiving None at this highest level?

Isgotkowitz: I think those safeguards should be the responsibility of `ManureManager`, which will be the entity handling the Processors. With the type annotation, it is pretty explicit that `receive_manure` should never be passed None as an arg and thus it is the calling function's responsibility to make sure that a `ManureStream` is passed.

Isgotkowitz: If a `None` somehow did make it in there, it would result in a NoneType error if `check_manure_stream_compatibility` was used to check it. So I think it is ok as is

elle-andreen (approved):
Looks good! For future reference: as mentioned in the PR description, I suggested just leaving the Digester base class empty for now because the current anaerobic digestion methods are based specifically on continuously stirred tank reactors (CSTRs). If we add other digester types, I strongly suspect they'll have different calculation methods. And if not, we'll move any overlapping methods "up" to the base class.

PR #2234, opened 2025-02-18
[Manure] Implement Refresh Base Manure Separators

### LINKED ISSUE #2100: [Manure] Implement base `Separator` class
Implement the base `Separator` class. Use the "Developer Guidelines" in #2092.

ew3361zh:
Create the refreshed base manure Separator class.

### Context

Issue(s) closed by this pull request: closes #2100

### What

1. Adds Separator base class which inherits from Processor class.
2. ~~Implements MachineSeparator class which inherits from Separator class.~~
3. ~~Adds Separator config class to encapsulate all efficiencies are present and packaged together.~~
4. Added a bool dunder method to ManureStream to allow using it as a bool.
5. Added defaults for ManureStream to allow empty instantiations for when no manure is in the separator and process_manure() is called on that separator.

### Why

Part of manure refresh process.

### How

Created the Separator class which inherits from the Processor class.
~~Additionally implemented a Separator config class to package all efficiencies together.~~
~~Implemented required inherited abstract methods including extending process_manure() to allow it to eventually implement multiple different separate_manure() child methods.~~
~~Created MachineSeparator class which implements roughly what our current separator classes are doing - basically applying separation efficiencies - and split the resulting output into liquid and solid manure streams in a dict.~~

### Test plan

Updated unit testing.


RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,133 @@
...
+    is_housing_emissions_calculator : bool
+        If true, processor will only accept ManureStreams with non-None PenManureData, if false then vice versa.
+
+    """
+
+    def __init__(self, config: SeparatorConfig, is_housing_emissions_calculator: bool = False) -> None:
+        """Initializes a new Separator."""
+        super().__init__(is_housing_emissions_calculator)
+        self.config: SeparatorConfig = config
+        self.held_manure: ManureStream | None = None
```
Isgotkowitz:
```suggestion
    def __init__(self, config: SeparatorConfig) -> None:
        """Initializes a new Separator."""
        super().__init__(is_housing_emissions_calculator=False)
        self.config: SeparatorConfig = config
        self.held_manure: ManureStream | None = None
```
I think we can be reasonably confident that there will never be a separator that calculates housing emissions.

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,133 @@
+from dataclasses import dataclass
+from RUFAS.biophysical.manure.processor import Processor
+from RUFAS.current_day_conditions import CurrentDayConditions
+from RUFAS.data_structures.animal_to_manure_connection import ManureStream
+from RUFAS.time import Time
+
+
+@dataclass
+class SeparatorConfig:
```
Isgotkowitz: Instead of packaging these into a separate data structure, I think they should just be attributes of the `Separator`.

I think having the `SeparatorConfig` adds an extra layer to the Separators that isn't really necessary.

RUFAS/biophysical/manure/separators/machine_separator.py
```diff
@@ -0,0 +1,64 @@
+from RUFAS.biophysical.manure.separators.separator import Separator, SeparatorConfig
+from RUFAS.data_structures.animal_to_manure_connection import ManureStream
+from RUFAS.routines.manure.constants_and_units.manure_constants import ManureConstants
```
Isgotkowitz: Instead of importing this from routines, could you please make a new `ManureConstants` file in `biophysical/manure` and add the constants there.

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,133 @@
...
+        Parameters
+        ----------
+        conditions : CurrentDayConditions
+            Current weather and environmental conditions that manure is being processed in.
+        time : Time
+            Time instance containing the simulations temporal information.
+
+        Returns
+        -------
+        dict
```
Isgotkowitz:
```suggestion
        dict[str, ManureStream]
```

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,133 @@
...
+                The solid portion of the separated manure.
+            - "liquid" : ManureStream
+                The liquid portion of the separated manure.
+
+        """
+        if not self.held_manure:
+            return {"solid": ManureStream(), "liquid": ManureStream()}
+        return self._separate_manure()
+
+    def _separate_manure(self) -> dict[str, ManureStream]:
```
Isgotkowitz: Instead of having a `_separate_manure` method, this logic should be in `process_manure` because the "processing" that a separator does is separating.

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,133 @@
...
+        dict
+            A dictionary containing:
+            - "solid" : ManureStream
+                The solid portion of the separated manure.
+            - "liquid" : ManureStream
+                The liquid portion of the separated manure.
+
+        """
+        if not self.held_manure:
+            return {"solid": ManureStream(), "liquid": ManureStream()}
```
Isgotkowitz:
```suggestion
            return {}
```
This isn't specified in the design doc (that's my bad), but I was thinking that if there is no manure to be passed on to the next Processor then it would be best to return an empty dict. I see two main benefits to this:
1. It is more explicit to someone looking at what is being passed between Processors that there is no mass being transferred, because they can see that the dict is empty instead of looking at the dict, then having to look at the attributes inside each `ManureStream`.
2. It saves a small amount of processing power, because the `ManureManager` will see that the dict is empty and just move on to the next Processor. Instead of passing empty `ManureStream`s on to the next Processor, which will still have to be added even though they do not have any mass.

RUFAS/data_structures/animal_to_manure_connection.py
```diff
@@ -164,6 +164,31 @@ def __add__(self, other: "ManureStream") -> "ManureStream":
...
+        """
+        Returns False if all nutrient, solids, and volume values are zero.
+        A manure stream is considered "empty" if it contains no mass or volume.
+
+        Returns
+        -------
+        bool
+            True if at least one attribute is non-zero, False otherwise.
+        """
+        return any(
```
Isgotkowitz: Should this also check that `PenManureData` is `None`? Also, instead of implementing this as the `bool` dunder method, maybe it would be better to implement it as an `@property` attribute, and name it something like "is_empty". I say this because it is not super intuitive what a "True" vs. "False" `ManureStream` would be, but "is_empty" is much more explicit.

RUFAS/data_structures/animal_to_manure_connection.py
```diff
@@ -122,17 +122,17 @@ class ManureStream:
...
+    ammoniacal_nitrogen: float = 0.0
+    nitrogen: float = 0.0
+    phosphorus: float = 0.0
+    potassium: float = 0.0
+    ash: float = 0.0
+    non_degradable_volatile_solids: float = 0.0
+    degradable_volatile_solids: float = 0.0
+    total_solids: float = 0.0
+    volume: float = 0.0
+    pen_manure_data: PenManureData | None = None
```
Isgotkowitz: I came up against the same issue when I was working on #2214, but I took a very different approach. Instead of adding defaults to `ManureStream`, I added the factory method `make_empty_manure_stream` which makes an empty `ManureStream`.

Each strategy has its pros and cons, adding defaults is cleaner but I think it is less robust than the factory method. I say this because having the defaults makes it easier for a `ManureStream` instantiator to mess up by specifying some attributes but not others. Like if it says that there is a certain amount of ash, nutrients, water, etc. but doesn't provide any value for volume, the resulting `ManureStream` will still be created but will be physically impossible/incorrect.

RUFAS/biophysical/manure/separators/machine_separator.py
```diff
@@ -0,0 +1,64 @@
...
+            A dictionary containing:
+            - "solid" : ManureStream
+                The solid portion of the separated manure.
+            - "liquid" : ManureStream
+                The liquid portion of the separated manure.
+        """
+        if not self.held_manure:
+            raise ValueError("Cannot separate manure when 'held_manure' is None.")
+
+        solid_manure = ManureStream(
```
Isgotkowitz: I appreciate you taking the initiative to make a separate subclass for Machine/Mechanical separators, but I think the refresh is not the right time to be adding this distinction. The only separators that need to be immediately supported are rotary screen and screw press, and since they both use the logic in the same logic it should be in the base `Separator`.

At some point in the future when more separator configurations need to be added, non-mechanical ones in particular, I think the structure you have started here will be used. But that structure should be implemented only after it has been designed and reviewed through the design doc process.

RUFAS/biophysical/manure/separators/machine_separator.py
```diff
@@ -0,0 +1,64 @@
...
+            nitrogen=self.held_manure.nitrogen * self.config.nitrogen_efficiency,
+            phosphorus=self.held_manure.phosphorus * self.config.phosphorus_efficiency,
+            potassium=self.held_manure.potassium * self.config.potassium_efficiency,
+            ash=self.held_manure.ash * self.config.ash_efficiency,
+            non_degradable_volatile_solids=self.held_manure.non_degradable_volatile_solids
+            * self.config.non_degradable_volatile_solids_efficiency,
+            degradable_volatile_solids=self.held_manure.degradable_volatile_solids
+            * self.config.degradable_volatile_solids_efficiency,
+            total_solids=self.held_manure.total_solids * self.config.total_solids_efficiency,
+            volume=self.held_manure.volume * ManureConstants.SOLID_MANURE_DENSITY,  # TODO: Check if this is correct
```
Isgotkowitz: I believe this is incorrect, I think volume should be equal the total mass of manure multiplied by the density, but @elle-andreen is better equipped to answer this.

RUFAS/biophysical/manure/separators/machine_separator.py
```diff
@@ -0,0 +1,64 @@
...
+            potassium=self.held_manure.potassium * (1 - self.config.potassium_efficiency),
+            ash=self.held_manure.ash * (1 - self.config.ash_efficiency),
+            non_degradable_volatile_solids=self.held_manure.non_degradable_volatile_solids
+            * (1 - self.config.non_degradable_volatile_solids_efficiency),
+            degradable_volatile_solids=self.held_manure.degradable_volatile_solids
+            * (1 - self.config.degradable_volatile_solids_efficiency),
+            total_solids=self.held_manure.total_solids * (1 - self.config.total_solids_efficiency),
+            volume=self.held_manure.volume * ManureConstants.LIQUID_MANURE_DENSITY,  # TODO: Check if this is correct
+        )
+        return {"solid": solid_manure, "liquid": liquid_manure}
```
Isgotkowitz: Before these `ManureStreams` get passed on to the next Processor, their attributes should be recorded as specified in this sheet: https://docs.google.com/spreadsheets/d/1LTLLNjA580pkIWm6ILqwrCDMb9903NnIV6GAw0W6l_8/edit?gid=0#gid=0 (my bad, I should have emphasized this better in the design doc).

The way I approached this to minimize code duplication in #2214 was to make the output prefix an attribute of the class, but that approach has yet to be reviewed.

Isgotkowitz (commented):
This is an excellent start to having the refreshed `Separator`s Niko! I think there are some things about the structure of the separators that can be simplified, and a few things that will be shared by separators that need to be added.

RUFAS/biophysical/manure/separators/separator.py
ew3361zh: I can definitely do this, my thought process was I wasn't sure whether a set of efficiencies was something that would apply to all separators or whether they would apply only to mechanical separators. e.g. for other separators is there an equation to determine an efficiency based on some other attributes of the separator (volume/capacity/chemical reaction/etc).

RUFAS/biophysical/manure/separators/separator.py
ew3361zh: yea this was only set up for the grandchild classes to allow them to implement different separation processes (if desired).

RUFAS/data_structures/animal_to_manure_connection.py
ew3361zh: I can add that code to my PR. I think I only did it to be able to return empty liquid and solid manure streams from the separator when there was no held manure to be separated (which we've replaced with returning an empty dict instead).

RUFAS/biophysical/manure/separators/machine_separator.py
ew3361zh: Done and done!

RUFAS/biophysical/manure/separators/separator.py
Isgotkowitz: If we get to that point with separators, it may be cleaner just to override the `process_manure` implementation in the grandchild classes

RUFAS/biophysical/manure/separators/separator.py
Isgotkowitz: Good question, you're probably onto something with that. But until there is actually a case where that happens, I think the best thing to do would be make the efficiencies attributes of the class.

Also, if there are some efficiencies that are used by some child classes but not others, maybe it would be better to only put them in or override them in the child classes somehow.

RUFAS/biophysical/manure/separators/machine_separator.py
```diff
@@ -0,0 +1,64 @@
+from RUFAS.biophysical.manure.separators.separator import Separator, SeparatorConfig
+from RUFAS.data_structures.animal_to_manure_connection import ManureStream
+from RUFAS.routines.manure.constants_and_units.manure_constants import ManureConstants
+
+
+class MachineSeparator(Separator):
+    """
+    A separator that uses a machine to separate solids from liquids.
```
elle-andreen:
```suggestion
    A separator that uses mechanical force to remove a portion of solids from manure.
```

RUFAS/biophysical/manure/separators/machine_separator.py
```diff
@@ -0,0 +1,64 @@
...
+        Configuration for the separator.
+    """
+
+    def __init__(self, config: SeparatorConfig) -> None:
+        """Initializes a new MachineSeparator."""
+        super().__init__(config)
+
+    def _separate_manure(self) -> dict[str, ManureStream]:
+        """
+        Separates the solids from the liquids in the manure.
```
elle-andreen:
```suggestion
        Separates a portion of solids from manure.
```

RUFAS/biophysical/manure/separators/machine_separator.py
```diff
@@ -0,0 +1,64 @@
...
+
+        solid_manure = ManureStream(
+            water=self.held_manure.water * self.config.water_efficiency,
+            ammoniacal_nitrogen=self.held_manure.ammoniacal_nitrogen * self.config.ammoniacal_nitrogen_efficiency,
+            nitrogen=self.held_manure.nitrogen * self.config.nitrogen_efficiency,
+            phosphorus=self.held_manure.phosphorus * self.config.phosphorus_efficiency,
+            potassium=self.held_manure.potassium * self.config.potassium_efficiency,
+            ash=self.held_manure.ash * self.config.ash_efficiency,
+            non_degradable_volatile_solids=self.held_manure.non_degradable_volatile_solids
+            * self.config.non_degradable_volatile_solids_efficiency,
```
elle-andreen:
```suggestion
            * self.config.total_volatile_solids_efficiency,
```
Just FYI, there are not/won't be separate efficiency values for degradable and non-degradable volatile solids, so we should just use the total volatile solids efficiency here & for degradable VS. Total volatile solids efficiency is an existing input for separators.

RUFAS/biophysical/manure/separators/machine_separator.py
```diff
@@ -0,0 +1,64 @@
...
+            degradable_volatile_solids=self.held_manure.degradable_volatile_solids
+            * self.config.degradable_volatile_solids_efficiency,
```
elle-andreen:
```suggestion
            * self.config.total_volatile_solids_efficiency,
```

RUFAS/biophysical/manure/separators/machine_separator.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+        liquid_manure = ManureStream(
+            water=self.held_manure.water * (1 - self.config.water_efficiency),
+            ammoniacal_nitrogen=self.held_manure.ammoniacal_nitrogen * (1 - self.config.ammoniacal_nitrogen_efficiency),
+            nitrogen=self.held_manure.nitrogen * (1 - self.config.nitrogen_efficiency),
+            phosphorus=self.held_manure.phosphorus * (1 - self.config.phosphorus_efficiency),
+            potassium=self.held_manure.potassium * (1 - self.config.potassium_efficiency),
+            ash=self.held_manure.ash * (1 - self.config.ash_efficiency),
+            non_degradable_volatile_solids=self.held_manure.non_degradable_volatile_solids
+            * (1 - self.config.non_degradable_volatile_solids_efficiency),
```
elle-andreen:
```suggestion
            * (1 - self.config.total_volatile_solids_efficiency),
```

RUFAS/biophysical/manure/separators/machine_separator.py
```diff
@@ -0,0 +1,64 @@
...
+            degradable_volatile_solids=self.held_manure.degradable_volatile_solids
+            * (1 - self.config.degradable_volatile_solids_efficiency),
```
elle-andreen:
```suggestion
            * (1 - self.config.total_volatile_solids_efficiency),
```

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,133 @@
...
+    phosphorus_efficiency : float
+        The efficiency of the separator in removing phosphorus from the manure.
+    potassium_efficiency : float
+        The efficiency of the separator in removing potassium from the manure.
+    ash_efficiency : float
+        The efficiency of the separator in removing ash from the manure.
+    non_degradable_volatile_solids_efficiency : float
+        The efficiency of the separator in removing non-degradable volatile solids from the manure.
+    degradable_volatile_solids_efficiency : float
+        The efficiency of the separator in removing degradable volatile solids from the manure.
```
elle-andreen:
```suggestion
    total_volatile_solids_efficiency : float
        The efficiency of the separator in removing volatile solids from the manure.
```
Same comment as in machine_processor.py. I think there are some other instances in this file that should be updated as well.

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,133 @@
...
+    potassium_efficiency: float
+    ash_efficiency: float
+    non_degradable_volatile_solids_efficiency: float
+    degradable_volatile_solids_efficiency: float
+    total_solids_efficiency: float
+
+
+class Separator(Processor):
+    """
+    A manure processor that separates solids from liquids.
```
elle-andreen:
```suggestion
    A manure processor that separates a portion of solids from manure.
```
(Sorry, this is me being pedantic, but what else are SMEs for 😅)

RUFAS/biophysical/manure/separators/machine_separator.py
ew3361zh: I think Ed wanted this grandchild class removed but good to have this description for if we need to implement it specifically in the future.

RUFAS/biophysical/manure/separators/separator.py
ew3361zh: your pedantry is a gift to the project :).

Isgotkowitz: > your pedantry is a gift to the project :).

I second this

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,270 @@
...
+    Attributes
+    ----------
+    is_housing_emissions_calculator : bool = False
+        A flag to indicate if the processor is a housing emissions calculator. Will always be false for a separator.
+    name : str
+        The name of the separator.
+    prefix : str
+        The prefix to be used for naming output variables.
+    water_efficiency : float
+        The efficiency of the separator in removing water from the manure.
```
ew3361zh: Replace with `percent_dry_solids`

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,270 @@
...
+        """
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self.process_manure.__name__,
+        }
+        if not self.held_manure:
+            self.om.add_variable("empty_separator_output", {}, {**info_map, "units": MeasurementUnits.UNITLESS})
+            return {}
+        solid_manure_water = self.held_manure.water * self.water_efficiency
+        solid_manure_total_solids = self.held_manure.total_solids * self.total_solids_efficiency
```
ew3361zh: start with calculation of solid_manure_total_solids
then divide this by percent_dry_solids
that yields the total_mass
derive water by subtracting solid_manure_total_solids from total_mass

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,270 @@
...
+        )
+        self.om.add_variable(
+            f"{solid_stream_name}.volume", solid_manure.volume, {**info_map, "units": MeasurementUnits.CUBIC_METERS}
+        )
+        self.om.add_variable(
+            f"{solid_stream_name}.mass", solid_manure.mass, {**info_map, "units": MeasurementUnits.KILOGRAMS}
+        )
+
+        liquid_manure_name = f"{self._prefix}.Liquid.{self._name}"
+        liquid_manure_water = self.held_manure.water * (1 - self.water_efficiency)
```
ew3361zh: calculate this based off subtracting solid water from held manure total water

ew3361zh:
Updates:
- New calculations of water, mass in solid stream and water in liquid stream after conversations with @elle-andreen
- Logging stream attributes to OM
- Clearing heald-manure within process_manure() after streams have been generated
- Moved ManureConstants to biophysical/manure and removed unused constants from the file. The old manure constants is still in place since it seemed unnecessary to be updating and maintaining both.
- Removed MachineSeparator grandchild class and moved that logic to Separator
- Removed associated functions with implementing MachineSeparator grandchild class (e.g. _separate_manure() function)
- Added `is_empty` property along with classmethod `make_empty_manure_stream()` to ManureStream following what @Isgotkowitz is doing in his Storage PR #2214

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,277 @@
...
+    Attributes
+    ----------
+    is_housing_emissions_calculator : bool = False
+        A flag to indicate if the processor is a housing emissions calculator. Will always be false for a separator.
+    name : str
+        The name of the separator.
+    prefix : str
+        The prefix to be used for naming output variables.
+    percent_dry_solids : float = 0.0
+        The percent dry content in manure solids.
```
elle-andreen:
```suggestion
        The dry matter content (percent DM) of separated manure solids.
```

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,277 @@
...
+class Separator(Processor):
+    """
+    A manure processor that separates a portion of solids from manure.
+
+    Parameters
+    ----------
+    name : str = ""
+        The name of the separator.
+    percent_dry_solids : float = 0.0
+        The percent dry content in manure solids.
```
elle-andreen:
```suggestion
        The dry matter content (percent DM) of separated manure solids.
```

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,277 @@
...
+            potassium=self.held_manure.potassium * self.potassium_efficiency,
+            ash=self.held_manure.ash * self.ash_efficiency,
+            non_degradable_volatile_solids=self.held_manure.non_degradable_volatile_solids
+            * self.volatile_solids_efficiency,
+            degradable_volatile_solids=self.held_manure.degradable_volatile_solids * self.volatile_solids_efficiency,
+            total_solids=solid_manure_total_solids,
+            volume=solid_manure_volume,
+            pen_manure_data=None,
+        )
+        solid_stream_name = f"{self._prefix}.Solids.{self._name}"
```
elle-andreen:
```suggestion
        solid_stream_name = f"{self._prefix}.SeparatedSolids.{self._name}"
```
This is the prefix that appears in the variable name in CSVs/reports, correct? "SeparatedSolids" would be helpful to just be more linguistically specific to solids generated from separators.

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,277 @@
...
+            {**info_map, "units": MeasurementUnits.KILOGRAMS},
+        )
+        self.om.add_variable(
+            f"{solid_stream_name}.volume", solid_manure.volume, {**info_map, "units": MeasurementUnits.CUBIC_METERS}
+        )
+        self.om.add_variable(
+            f"{solid_stream_name}.mass", solid_manure.mass, {**info_map, "units": MeasurementUnits.KILOGRAMS}
+        )
+
+        liquid_manure_name = f"{self._prefix}.Liquid.{self._name}"
```
elle-andreen:
```suggestion
        liquid_manure_name = f"{self._prefix}.SeparatedLiquid.{self._name}"
```

elle-andreen (approved):
Left a few suggestions on output names but this looks great to me Niko! Thanks for sticking through some separator-induced confusion. I will take a look at whatever comments Ed/another dev leaves when they review but I think this is in great shape.

RUFAS/biophysical/manure/separators/separator.py
ew3361zh: Correct, this prefix would be a part of the variable name. I will make the change on the SeparatedSolids. Totally agree that's clearer.

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,277 @@
...
+        self.held_manure: ManureStream | None = None
+        self.percent_dry_solids: float = percent_dry_solids
+        self.ammoniacal_nitrogen_efficiency: float = ammoniacal_nitrogen_efficiency
+        self.nitrogen_efficiency: float = nitrogen_efficiency
+        self.phosphorus_efficiency: float = phosphorus_efficiency
+        self.potassium_efficiency: float = potassium_efficiency
+        self.ash_efficiency: float = ash_efficiency
+        self.volatile_solids_efficiency: float = volatile_solids_efficiency
+        self.total_solids_efficiency: float = total_solids_efficiency
+        self.om = OutputManager()
```
Isgotkowitz:
```suggestion
```
Instead of having this, I think `self._om` could be used from `Processor` instead.

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,277 @@
...
+        )
+        self.om.add_variable(
+            f"{liquid_manure_name}.mass", liquid_manure.mass, {**info_map, "units": MeasurementUnits.KILOGRAMS}
+        )
+
+        self.om.add_variable(
+            f"{self._prefix}.{self._name}.simulation_day",
+            time.simulation_day,
+            {**info_map, "units": MeasurementUnits.SIMULATION_DAY},
+        )
```
Isgotkowitz: I think this should not be reported from any separator. The only output variable reporting the simulation day should be the dedicated one from `Time`'s reporting method, otherwise there will be multiple outputs that *should * have the same thing.

However, I think it would be really good to add `"simulation_day"` to the info map, so that data expansion can be applied to outputs from separators.

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,277 @@
...
+            * self.volatile_solids_efficiency,
+            degradable_volatile_solids=self.held_manure.degradable_volatile_solids * self.volatile_solids_efficiency,
+            total_solids=solid_manure_total_solids,
+            volume=solid_manure_volume,
+            pen_manure_data=None,
+        )
+        solid_stream_name = f"{self._prefix}.SeparatedSolids.{self._name}"
+        self.om.add_variable(
+            f"{solid_stream_name}.water", solid_manure.water, {**info_map, "units": MeasurementUnits.KILOGRAMS}
+        )
```
Isgotkowitz: For the reporting of the solid and liquid separated manure streams, would you write a utility function that reports a `ManureStream` to the output manager and can be used for both the solid and liquid manure streams? I think this would make the code quite a bit cleaner.

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,277 @@
...
+        self,
+        name: str = "",
+        percent_dry_solids: float = 0.0,
+        ammoniacal_nitrogen_efficiency: float = 0.0,
+        nitrogen_efficiency: float = 0.0,
+        phosphorus_efficiency: float = 0.0,
+        potassium_efficiency: float = 0.0,
+        ash_efficiency: float = 0.0,
+        volatile_solids_efficiency: float = 0.0,
+        total_solids_efficiency: float = 0.0,
```
Isgotkowitz:
```suggestion
        percent_dry_solids: float,
        ammoniacal_nitrogen_efficiency: float,
        nitrogen_efficiency: float,
        phosphorus_efficiency: float,
        potassium_efficiency: float,
        ash_efficiency: float,
        volatile_solids_efficiency: float,
        total_solids_efficiency: float,
```
I suggest not having any defaults for these values. If there are any defaults we want applied, I think they should be defined in the metadata properties and only in the metadata properties.

Alternatively, if we want certain defaults or hardcoded values for certain types of processors, then I think it would be best to put them in those the child classes.

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,277 @@
...
+        """
+        if self.held_manure is None:
+            self.held_manure = manure
+        else:
+            self.held_manure += manure
+
+    def process_manure(self, conditions: CurrentDayConditions, time: Time) -> dict[str, ManureStream]:
+        """
+        Executes the daily manure processing operations.
+        For this class, it separates a portion of solids from the manure and returns the solid and liquid portions.
```
Isgotkowitz:
```suggestion
        Executes the daily separation of solids from the manure and returns the solid and liquid portions.
```

changelog.md
```diff
@@ -108,6 +108,7 @@ v0.9.2
 - [2211](https://github.com/RuminantFarmSystems/MASM/pull/2211) - [minor change] [Data Collection App] Makes error messages specific and detailed when invalid input is entered into the Data Collection App.
 - [2229](https://github.com/RuminantFarmSystems/MASM/pull/2229) - [minor change] [Emissions] Bug fix for date comparisons in emissions.py.
 - [2230](https://github.com/RuminantFarmSystems/MASM/pull/2230) - [minor change] [Manure] Updates Manure metadata descriptions and ranges.
+- [2234](https://github.com/RuminantFarmSystems/MASM/pull/2234) - [minor change] [Manure] Implements base manure Separator and child MachineSeparator classes.
```
```suggestion
- [2234](https://github.com/RuminantFarmSystems/MASM/pull/2234) - [minor change] [Manure] Implements base manure Separator class.
```

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,277 @@
...
+        phosphorus_efficiency: float = 0.0,
+        potassium_efficiency: float = 0.0,
+        ash_efficiency: float = 0.0,
+        volatile_solids_efficiency: float = 0.0,
+        total_solids_efficiency: float = 0.0,
+    ) -> None:
+        """Initializes a new Separator."""
+        super().__init__(name=name, is_housing_emissions_calculator=False)
+        self._name: str = name
+        self._prefix: str = f"{self.__class__.__name__}"
```
Isgotkowitz:
```suggestion
```
I think it would be better to use the prefix that is defined in `Processor`, otherwise there will be inconsistent naming patterns in output from manure processors.

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,277 @@
...
+    total_solids_efficiency : float
+        The efficiency of the separator in removing total solids from the manure.
+    om : OutputManager
+        The output manager instance used to store and manage output data.
+
+    """
+
+    def __init__(
+        self,
+        name: str = "",
```
Isgotkowitz:
```suggestion
        name: str,
```
The name should _never_ be blank, it should always be defined by the user. If it is blank, it could disrupt the operations of `ManureManager` in a catastrophic way and the make outputs either really difficult or impossible to interpret.

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,277 @@
...
+        nitrogen_efficiency: float = 0.0,
+        phosphorus_efficiency: float = 0.0,
+        potassium_efficiency: float = 0.0,
+        ash_efficiency: float = 0.0,
+        volatile_solids_efficiency: float = 0.0,
+        total_solids_efficiency: float = 0.0,
+    ) -> None:
+        """Initializes a new Separator."""
+        super().__init__(name=name, is_housing_emissions_calculator=False)
+        self._name: str = name
```
Isgotkowitz:
```suggestion
```
I don't think it is necessary to have a `_name` attribute, it feels duplicative of the `name` in `Processor`.

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,277 @@
...
+            - "solid" : ManureStream
+                The solid portion of the separated manure.
+            - "liquid" : ManureStream
+                The liquid portion of the separated manure.
+
+        """
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": self.process_manure.__name__,
+        }
```
Isgotkowitz:
```suggestion
        info_map = {
            "class": self.__class__.__name__,
            "function": self.process_manure.__name__,
            "prefix": self._prefix,
        }
```
I think the prefix should be set in the info map. This will be the difference between `ManureSeparator.SeparatedSolids.YourProcessorNameHere.water` and `Separator.process_manure.ManureSeparator.SeparatedSolids.YourProcessorNameHere.water`, and I interpreted the [output specification](https://docs.google.com/spreadsheets/d/1LTLLNjA580pkIWm6ILqwrCDMb9903NnIV6GAw0W6l_8/edit?gid=0#gid=0) as asking for the former.

Also, I know that the output specification calls for `ManureSeparator.SeparatedSolids.YourProcessorNameHere.water`, but would it possible to do `ManureSeparator.YourProcessorNameHere.SeparatedSolids.water` instead? The only benefit is that this would make the variable name more uniform with the outputs from other Processors because it would be following the prefix set in `_prefix` in `Processor`.

@elle-andreen could you please weigh in on these?

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,277 @@
...
+            non_degradable_volatile_solids=self.held_manure.non_degradable_volatile_solids
+            * self.volatile_solids_efficiency,
+            degradable_volatile_solids=self.held_manure.degradable_volatile_solids * self.volatile_solids_efficiency,
+            total_solids=solid_manure_total_solids,
+            volume=solid_manure_volume,
+            pen_manure_data=None,
+        )
+        solid_stream_name = f"{self._prefix}.SeparatedSolids.{self._name}"
+        self.om.add_variable(
+            f"{solid_stream_name}.water", solid_manure.water, {**info_map, "units": MeasurementUnits.KILOGRAMS}
```
Isgotkowitz:
```suggestion
            f"{solid_stream_name}.manure_water", solid_manure.water, {**info_map, "units": MeasurementUnits.KILOGRAMS}
```
I believe all of these variable names need to be prefixed with `manure_`

Isgotkowitz (commented):
I think this looks really good Niko! I left a bunch of suggestions on things that I think could be improved, but these are mostly minor suggestions. The one thing that I think we need to hammer out before this PR is merged is the full names of output variables from separators, and that should be done in coordination with Elle. Once we get that figured out tho, I think it will be good to go!

Also, I apologize for taking so long to get the updates to `Processor` merged in, that is my bad for not doing a better job of planning.

RUFAS/biophysical/manure/separators/separator.py
elle-andreen: That's fine with me! I wrote up the output names long ago and they're open to modification for clarity or ease of implementation

RUFAS/biophysical/manure/separators/separator.py
elle-andreen: Yes, if being reported to output manager, they should be prefixed with "manure_"

ew3361zh:
Updates:
- Added function to log manure streams to OM
- Removed defaults for init params
- Utilized inherited attributes
- Updated var names sent to OM to have `manure_` before the actual name of the variable

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,221 @@
...
+            potassium=self.held_manure.potassium * self.potassium_efficiency,
+            ash=self.held_manure.ash * self.ash_efficiency,
+            non_degradable_volatile_solids=self.held_manure.non_degradable_volatile_solids
+            * self.volatile_solids_efficiency,
+            degradable_volatile_solids=self.held_manure.degradable_volatile_solids * self.volatile_solids_efficiency,
+            total_solids=solid_manure_total_solids,
+            volume=solid_manure_volume,
+            pen_manure_data=None,
+        )
+        solid_stream_name = f"{self.name}.SeparatedSolids"
```
Isgotkowitz:
```suggestion
        solid_stream_name = "SeparatedSolids"
```
The name of the processor is already included in the `_prefix`, so it is not necessary to include it here as well.

RUFAS/biophysical/manure/separators/separator.py
```diff
@@ -0,0 +1,221 @@
...
+            ash=self.held_manure.ash * (1 - self.ash_efficiency),
+            non_degradable_volatile_solids=self.held_manure.non_degradable_volatile_solids
+            * (1 - self.volatile_solids_efficiency),
+            degradable_volatile_solids=self.held_manure.degradable_volatile_solids
+            * (1 - self.volatile_solids_efficiency),
+            total_solids=liquid_manure_total_solids,
+            volume=liquid_manure_volume,
+            pen_manure_data=None,
+        )
+        liquid_stream_name = f"{self.name}.SeparatedLiquid"
```
Isgotkowitz:
```suggestion
        liquid_stream_name = "SeparatedLiquid"
```

Isgotkowitz (approved):
LG8TM! Nice job Niko, had two minor suggestions on the names of outputs but everything else looks awesome. Thanks for working on this!

PR #2236, opened 2025-02-18
[Testing] Patch: Automatically Update All E2E Expected Results

### LINKED ISSUE #2213: [Testing] E2E Testing Update Should Update All Expected Results
## Status

Currently E2E testing update will compare the existing e2e expected result for each domain with the newly generated set of results. If there is no difference, it will not update the file.

## Intention

In discussing with Pooya, the more appropriate behavior would be to update all the results regardless of differences. Whether or not differences were found should be logged though so the user is aware of the update. Perhaps even save the differences using the same mechanism that's in the main e2e function.

## Request

Remove the check to save based on whether differences exist and automatically overwrite the existing e2e expected results with the new expected results.

ew3361zh:
Quick patch to alter previous behavior where we would only update the expected results for e2e testing domains if there were differences found. This automatically updates all expected results regardless of differences found between expected and actual.

### Context

Issue(s) closed by this pull request: closes #2213

### What

1. Updates behavior in e2e update expected results task to now automatically overwrite all expected results with the actual results from the simulation run.
2. Saves any differences found between existing expected results and new actual results.

### Why

This process is cleaner and more understandable. When this task is run, all expected results will be overwritten and any differences found will be saved to the JSONs folder.

### How

Moved call to save new expected results out of if block where differences are found so that the existing expected results for each domain is overwritten with the newly generated actual results.
Differences between the existing expected results and actual results are still checked. If differences were found, they are saved to a JSON file.

### Test plan

1. Run the e2e update expected results script:
```bash
python main.py -p input/metadata/update_end_to_end_testing_tm_metadata.json -c
```
2. Note the output in the terminal when fresh expected results are being generated.
3. Despite e2e testing passing currently, there are some small-decimal differences found so e2e results will update for at least Animal and C&S modules.
4. Check the difference files in the output/JSONs folder (should be e.g. `CropAndSoil_update_diff.json`) to see the differences found.


allisterakun (approved):
LGTM! Work as expected!

matthew7838 (approved):
LGTM, also worked as expected on my side.

PR #2237, opened 2025-02-19
[Manure] Implement base handler class

### LINKED ISSUE #2101: [Manure] Implement base `Handler` class
Implement the base `Handler` class. Use the "Developer Guidelines" in #2092.

matthew7838:
This PR implements the base handler class for manure processors.

### Context

Issue(s) closed by this pull request: closes #2101 

### What

1. There is now a base class called `handler` which will have methods that will be used across **all** handlers.
2. There will be another class called `single_stream_handler` which inherits the base `handler` class that will handle methods and manure stream receiving for manual scraping, flush system, and alley scraper handler files. The mentioned handler types will inherit this class.
3. A placeholder ammonia emission calculation is added in `processor.py`, this will be implemented with PR #2214.
4. The om reporting emissions for methane and CO2 will be implemented (likely in `single_stream_handler` class) when manual scraping, flush system, and alley scraper handlers are implemented.
5. All functions have unit tests written.
6. The constructor will now take in an enum type of processor sub types, such values will be reported with the info map in the outputs.
7. Added an enum class for all types of sub-processes.

### Why

This will unblock phase 2 of manure refresh.

### How

The PR will set up base handler implementations and logic for all the child class handlers.

### Test plan

Unit tests have been written for all methods with 100% coverage. For now, it will only be a combination of SME review and unit tests.


RUFAS/biophysical/manure/handler.py
```diff
@@ -0,0 +1,280 @@
...
+        Unique identifier of the processor.
+    is_housing_emissions_calculator : bool
+        Indicates if a Processor calculates housing emissions.
+
+    Attributes
+    ----------
+    manure_stream : ManureStream
+        The ManureStream instance being checked for compatibility.
+    fresh_water_volume_used_for_milking : float
+        The amount of fresh water used for milking (L).
```
elle-andreen: This calculation should be performed only by parlor cleaning handlers, is there a specific reason it's here in the base class?

RUFAS/biophysical/manure/handler.py
```diff
@@ -0,0 +1,280 @@
...
+        if not super().check_manure_stream_compatibility(manure_stream):
+            return False
+        if (
+            manure_stream.pen_manure_data is not None
+            and manure_stream.pen_manure_data.pen_type is not None
+            and manure_stream.pen_manure_data.pen_type not in ["freestall", "tiestall"]
+        ):
+            self._om.add_error(
+                "Unsupported pen type.",
+                f"Handler only supports freestall and tiestall," f" received {manure_stream.pen_manure_data.pen_type}.",
```
elle-andreen:
```suggestion
                f"Handler should only be used with freestall and tiestall pen types," f" received {manure_stream.pen_manure_data.pen_type}.",
```

RUFAS/biophysical/manure/handler.py
matthew7838: This will make the extraction of process_manure logic more generic. For all other handlers this attribute will not be updated so it will process as 0 (not performed). Parlor cleaning handler will update the value before calling the base `process_manure` method so when `process_manure` method is called, it will do the processing with calculated `fresh_water_volume_used_for_milking` value.

RUFAS/biophysical/manure/handler.py
```diff
@@ -0,0 +1,280 @@
...
+            info_map = {"class": cls.__name__, "function": cls.determine_barn_area.__name__}
+            om.add_error("Invalid pen type.", f"Valid pen types are tiestall and freestall, got {pen_type}", info_map)
+            raise ValueError(f"Invalid pen type: {pen_type}")
+
+        is_lac_cow = animal_combination == AnimalCombination.LAC_COW
+        surface_area_per_stall = surface_area_table.get((pen_type, is_lac_cow), 0.0)
+
+        return num_stalls * surface_area_per_stall
+
+    def check_manure_stream_compatibility(self, manure_stream: ManureStream) -> bool:
```
elle-andreen: In processor.py and other files, functions with this same name are used to check the ManureStream's compatibility with the processor, but the nature of the check is different - in other places we're checking the processor's status as a valid housing emissions calculator, but here we're checking against pen type. I'm not sure if it's either good or bad practice to use the same function name to do the same overarching thing but in two different way? I think it could potentially be confusing, but there could be a good reason I'm not aware of for why we should keep the function name consistent!

matthew7838: You are right, I will coordinate this with other member's PR.

RUFAS/biophysical/manure/handler.py
```diff
@@ -0,0 +1,280 @@
...
+        Indicates if a Processor calculates housing emissions.
+
+    Attributes
+    ----------
+    manure_stream : ManureStream
+        The ManureStream instance being checked for compatibility.
+    fresh_water_volume_used_for_milking : float
+        The amount of fresh water used for milking (L).
+    ammonia_emission : float
+        The amount of ammonia emission (kg).
```
elle-andreen: I could be wrong here but should this be an attribute? This is a value determined each day for the manure passing through the handler, it's not an attribute of the handler itself.

RUFAS/biophysical/manure/handler.py
```diff
@@ -0,0 +1,281 @@
...
+        info_map_m3 = {"units": MeasurementUnits.CUBIC_METERS}
+        cleaning_water_volume = self.determine_cleaning_water_volume_in_main_barn(
+            self.manure_stream.pen_manure_data.num_animals,
+            self.config.cleaning_water_use_rate,
+            self.config.cleaning_water_recycle_fraction,
+        )
+        barn_temperature = self.determine_barn_temperature(conditions.mean_air_temperature)
+
+        total_cleaning_water_volume = (
+            cleaning_water_volume + self.fresh_water_volume_used_for_milking
```
elle-andreen: Where is the actual calculation of `fresh_water_volume_used_for_milking` happening or supposed to happen?

RUFAS/biophysical/manure/handler.py
```diff
@@ -0,0 +1,281 @@
...
+            )
+        }
+
+    @staticmethod
+    def determine_cleaning_water_volume_in_main_barn(
+        num_animals: int, cleaning_water_use_rate: float, cleaning_water_recycle_fraction: float
+    ) -> float:
+        """
+        Calculates the volume of fresh (non-recycled) cleaning water used for, and ultimately added to, a single manure
+         stream on a single simulation day by the manure handler.
```
elle-andreen:
```suggestion
         stream on a single simulation day by the manure handler. For parlor cleaning handlers, this water volume represents an optional parlor flush (separate from fresh water only cleaning water). For all other handler types, this water volume represents water use by handlers in the pen, such as a barn floor flush system.
```

RUFAS/biophysical/manure/handler.py
```diff
@@ -0,0 +1,281 @@
...
+        float
+            The volume of fresh (non-recycled) cleaning water (m^3).
+
+        """
+        return num_animals * (cleaning_water_use_rate * (1 - cleaning_water_recycle_fraction))
+
+    @staticmethod
+    def determine_barn_temperature(air_temp: float) -> float:
+        """
+        Calculates the barn temperature.
```
elle-andreen: For this function, and other functions as we're moving things around in the refresh, please consider carefully what you're pulling in vs. leaving out of the docstring. There is important reference information included for this specific function in `calculator.py` that we don't want to lose in the reference. That being said, I realize a lot of the docstrings in the current manure module are cluttered/formatted wrong/have unneeded text. I trust you guys' judgement on stuff like Parameters, Returns, etc. but please do pull in anything under Notes, References, or things like that. I will throw this in the dev team Slack as well to make others aware!

RUFAS/biophysical/manure/handler.py
matthew7838: This will happen in the parlor cleaning handler right before using this method.

RUFAS/biophysical/manure/handler.py
```diff
@@ -0,0 +1,281 @@
...
+        """
+        Calculates the volume of fresh (non-recycled) cleaning water used for, and ultimately added to, a single manure
+         stream on a single simulation day by the manure handler.
+
+        Parameters
+        ----------
+        num_animals : int
+            Number of animals.
+        cleaning_water_use_rate : float
+            The use rate of cleaning water (unitless).
```
elle-andreen: I think the units here would be L - technically it's liters per animal per day, but the function returns a value in L.

RUFAS/biophysical/manure/handler.py
```diff
@@ -0,0 +1,281 @@
...
+        cleaning_water_recycle_fraction : float
+            The fraction of cleaning water recycled (unitless).
+
+        Returns
+        -------
+        float
+            The volume of fresh (non-recycled) cleaning water (m^3).
```
elle-andreen:
```suggestion
            The volume of fresh (non-recycled) cleaning water (L).
```
Still in L here, we use this value in the function for `total_cleaning_water_volume` which applies the L to m3 conversion constant. I sort of bungled units relating to these water values as a whole in the design doc though, so that's my bad!

RUFAS/biophysical/manure/handler.py
```diff
@@ -0,0 +1,281 @@
...
+        )
+        barn_temperature = self.determine_barn_temperature(conditions.mean_air_temperature)
+
+        total_cleaning_water_volume = (
+            cleaning_water_volume + self.fresh_water_volume_used_for_milking
+        ) * GeneralConstants.LITERS_TO_CUBIC_METERS
+        self._om.add_variable("total_cleaning_water_volume", total_cleaning_water_volume, info_map_m3)
+        self._om.add_variable("barn_temperature", barn_temperature, info_map_c)
+
+        manure_water = self.manure_stream.water + cleaning_water_volume
```
elle-andreen:
```suggestion
        manure_water = self.manure_stream.water + (total_cleaning_water_volume * GeneralConstants.WATER_DENSITY_KG_PER_M3)
```
I had this wrong in the design doc, my apologies for that. I am updating this in the design doc now. Water in ManureStream is in kg, so added the density conversion constant here.

RUFAS/biophysical/manure/handler.py
```diff
@@ -0,0 +1,281 @@
...
+
+        manure_total_ammoniacal_nitrogen = max(0.0, self.manure_stream.ammoniacal_nitrogen - self.ammonia_emission)
+        nitrogen = self.manure_stream.nitrogen
+        phosphorus = self.manure_stream.phosphorus
+        potassium = self.manure_stream.potassium
+        ash = self.manure_stream.ash
+        non_degradable_volatile_solids = self.manure_stream.non_degradable_volatile_solids
+        degradable_volatile_solids = self.manure_stream.degradable_volatile_solids
+        volume = self.manure_stream.volume
```
elle-andreen:
```suggestion
        volume = self.manure_stream.volume + total_cleaning_water_volume 
```

RUFAS/biophysical/manure/processor.py
```diff
@@ -104,3 +104,16 @@ def check_manure_stream_compatibility(self, manure_stream: ManureStream) -> bool
         )
 
         return is_valid_housing_emissions_calculator ^ is_valid_non_housing_emissions_calculator
+
+    @classmethod
+    def _calculate_ammonia_emissions(
```
elle-andreen: I think Ed has already added this method in PR 2214 (although it hasn't merged), just a heads up

RUFAS/biophysical/manure/single_stream_handler.py
```diff
@@ -0,0 +1,168 @@
...
+        barn_temperature = self.determine_barn_temperature(conditions.mean_air_temperature)
+        surface_area = self.determine_barn_area(
+            self.manure_stream.pen_manure_data.animal_combination,
+            self.manure_stream.pen_manure_data.pen_type,
+            self.manure_stream.pen_manure_data.num_stalls,
+        )
+        self.ammonia_emission = self._calculate_ammonia_emissions(
+            self.manure_stream.ammoniacal_nitrogen,
+            self.manure_stream.volume,
+            990.0,
```
elle-andreen: 990 is the density value, can we pass the actual density constant here? (`ManureConstants.SLURRY_MANURE_DENSITY`)   It may also be good to check with the rest of the dev team on what the plan is for relocating constants. I see some constants have been moved into base classes, but that density value will be used throughout the model so I'm not sure if its proper place is in a new 'ManureConstants' file in the biophysical folder, or if it should go in `processor.py`, or something else.

RUFAS/biophysical/manure/single_stream_handler.py
```diff
@@ -0,0 +1,168 @@
...
+            barn_temperature,
+            self.determine_ammonia_resistance(barn_temperature),
+            surface_area,
+            7.7,
```
elle-andreen: Same comment as above, can we pass `GasEmissionConstants.DEFAULT_PH_FOR_HOUSING_AMMONIA` here instead of just the number?

RUFAS/biophysical/manure/single_stream_handler.py
```diff
@@ -0,0 +1,168 @@
...
+        )
+        return super().process_manure(conditions, time)
+
+    @classmethod
+    def determine_methane_emissions(
```
elle-andreen:
```suggestion
    def determine_housing_methane_emissions(
```
Since we have other functions that calculate methane, it's helpful to specify that these are _housing_ methane emissions. I did not suggest this same change for ammonia because the same base ammonia calculation is used for both storage and housing ammonia emissions.

RUFAS/biophysical/manure/single_stream_handler.py
```diff
@@ -0,0 +1,168 @@
...
+        float
+            Methane emission from manure (kg).
+
+        """
+        barn_area = cls.determine_barn_area(animal_combination, pen_type, num_stalls)
+
+        return max(0.0, 0.13 * barn_temperature) * barn_area / 1000
+
+    @classmethod
+    def determine_carbon_dioxide_emissions(
```
elle-andreen:
```suggestion
    def determine_housing_carbon_dioxide_emissions(
```

RUFAS/biophysical/manure/single_stream_handler.py
```diff
@@ -0,0 +1,168 @@
...
+        -------
+        float
+            Carbon dioxide emission from manure (kg).
+
+        """
+        barn_area = cls.determine_barn_area(animal_combination, pen_type, num_stalls)
+        return max(0.0, 0.0065 + 0.0192 * barn_temperature) * barn_area / 1000
+
+    @staticmethod
+    def determine_ammonia_resistance(temp: float, hsc: float = 260) -> float:
```
elle-andreen:
```suggestion
    def determine_housing_ammonia_resistance(temp: float, hsc: float = GasEmissionConstants.HOUSING_HSC) -> float:
```
If this is the wrong place to implement than HSC constant, please modify!

RUFAS/biophysical/manure/processor.py
matthew7838: We talked about this, this is just a placeholder.

RUFAS/biophysical/manure/handler.py
matthew7838: This is meant to store the calculation for the ammonia housing emission (which will be calculated using the method in processor.py), are we referring to the same thing? or should I clarify with the naming?

RUFAS/biophysical/manure/handler.py
matthew7838: Would it be better to rename it to something like animal_cleaning_water_use_amount? I think the word rate feels like some sort of fraction.

RUFAS/biophysical/manure/single_stream_handler.py
matthew7838: Seems like the right place to me.

RUFAS/biophysical/manure/handler.py
elle-andreen: Yeah I think "rate" is referring to that it's liters _per_ cow, but I am fully on board with renaming it to cleaning_water_use_amount! I would just drop "animal_", to me that sounds like we are cleaning the animals themselves 😅

RUFAS/biophysical/manure/handler.py
elle-andreen: We are talking about the same thing I think, the actual value for ammonia_emission in kg. I think the issue is me not knowing exactly what should and shouldn't be listed as an attribute. My understanding of what constitutes an attribute is that it's something about the processor that is more like a property of the processor, and this value is more of a "product" of the handler rather than something specific to the handler. I just don't recall seeing gas emission values listed as attributes of other base classes so was not sure here

matthew7838: I see where this is coming from now. The reason for this is simply the same as `fresh_water_volume_used_for_milking` because, unlike other emissions, ammonia emission is not solely being calculated in the `SingleStreamHandler` so initializing a default 0 is needed to also serve like an on/off switch between parlor cleaning and non-parlor cleaning handlers. However, your point is still valid and I will seek other possible solutions if possible. @Isgotkowitz I would like to know your opinion on this.

RUFAS/biophysical/manure/single_stream_handler.py
matthew7838: If it's throughout the model then it should still be in 'ManureConstants'.

RUFAS/biophysical/manure/handler.py
matthew7838: After a second look, this is an overwritten method and it's meant to be. This method will call the processor's stream check and the handler's check. Let me know if you need more clarification.

RUFAS/biophysical/manure/handler/single_stream_handler.py
```diff
@@ -0,0 +1,165 @@
+from typing import Any
+
+from RUFAS.biophysical.manure.handler.handler import Handler, HandlerConfig
+from RUFAS.current_day_conditions import CurrentDayConditions
+from RUFAS.data_structures.animal_to_manure_connection import ManureStream
+from RUFAS.routines.manure.constants_and_units.gas_emission_constants import GasEmissionConstants
+from RUFAS.routines.manure.constants_and_units.manure_constants import ManureConstants
```
Isgotkowitz: There should be no cross over between `routines` and `biophysical`.

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,255 @@
...
+
+    """
+
+    name: str
+    manure_handler_type: str
+    cleaning_water_use_amount: float
+    minutes_per_cleaning: int
+    cleanings_per_day: int
+    cleaning_water_recycle_fraction: float
+    use_parlor_flush: bool
```
Isgotkowitz:
```suggestion
```
I think that all attributes which configure a handler should be kept directly as instance attributes. This will keep the Handlers consistent with other Processor child classes.

RUFAS/biophysical/manure/handler/single_stream_handler.py
Isgotkowitz, line 1: I appreciate you trying to be very explicit about the difference between the milking parlor cleaner and all other handlers, but because the logic in single stream handler is shared between 3/4 (flush system, manual scraper, and alley scraper) of the available handlers a better approach would be to implement the shared logic of these 3 handlers in the base `Handler` class and have them inherit directly from `Handler`.

When the Parlor Cleaning handler is implemented, it can override any of the behaviors defined in `Handler` that do not apply. This approach will lessen the amount of code that needs to be implemented to realize the refreshed manure module, and will keep the implementation more consistent with the planned design.

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,255 @@
...
+        The custom configuration for the handlers.
+
+    """
+
+    def __init__(self, name: str, is_housing_emissions_calculator: bool, config: HandlerConfig):
+        super().__init__(name, is_housing_emissions_calculator)
+        self.manure_stream: ManureStream | None = None
+        self.fresh_water_volume_used_for_milking: float = 0.0
+        self.ammonia_emission: float = 0.0
+        self.config = config
```
Isgotkowitz:
```suggestion
    def __init__(self, name: str, is_housing_emissions_calculator: bool, config: HandlerConfig, is_parlor_cleaner: bool):
        super().__init__(name, is_housing_emissions_calculator)
        self.manure_stream: ManureStream | None = None
        self.fresh_water_volume_used_for_milking: float = 0.0
        self.ammonia_emission: float = 0.0
        self.config = config
        self._is_parlor_cleaner: bool = is_parlor_cleaner
```
I recommend adding this param/attribute combo and extending the behavior of `check_manure_stream_compatibility` so that handlers check whether or not the `StreamType` of `PenManureData` is compatible with the handler or not.

If `_is_parlor_cleaning` is True, then the `StreamType` must be `PARLOR` for the manure stream to be compatible and vice versa.

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,255 @@
...
+        The amount of fresh water used for milking (L).
+    ammonia_emission : float
+        The amount of ammonia emission (kg).
+    config : HandlerConfig
+        The custom configuration for the handlers.
+
+    """
+
+    def __init__(self, name: str, is_housing_emissions_calculator: bool, config: HandlerConfig):
+        super().__init__(name, is_housing_emissions_calculator)
```
Isgotkowitz:
```suggestion
    def __init__(self, name: str, config: HandlerConfig):
        super().__init__(name, is_housing_emissions_calculator=True)
```
For now, I think it is safe to assume that all handlers will be considered housing emissions calculators.

Isgotkowitz (dismissed):
This is a good start, but there are some pretty significant changes that need to be made. Namely, `SingleStreamHandler` does not need to be implemented, instead I think we should plan on overriding any inapplicable logic in the individual child classes.

RUFAS/biophysical/manure/handler/single_stream_handler.py
matthew7838: Addressed using a new design by creating sub-processor types enum and passing it down to the constructor.

RUFAS/biophysical/manure/handler/handler.py
matthew7838: This will be addressed in the parlor cleaning implementation PR.

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,256 @@
...
+        use_parlor_flush: bool,
+    ):
+        super().__init__(name, is_housing_emissions_calculator=True)
+        self.manure_stream: ManureStream | None = None
+        self.fresh_water_volume_used_for_milking: float = 0.0
+        self.ammonia_emission: float = 0.0
+        self.handler_type = handler_type
+        self.cleaning_water_use_amount = cleaning_water_use_amount
+        self.minutes_per_cleaning = minutes_per_cleaning
+        self.cleanings_per_day = cleanings_per_day
```
ew3361zh: It looks like from the specific Handler design doc that these two attributes are unused and could be removed. Was that considered?

RUFAS/biophysical/manure/handler/single_stream_handler.py
```diff
@@ -0,0 +1,204 @@
...
+        info_map = {"class": self.__class__.__name__, "function": self.receive_manure.__name__}
+        super().receive_manure(manure_stream)
+        if self.manure_stream is not None:
+            self._om.add_error(
+                "Multiple stream received.",
+                f"Non parlor handler should only receive one manure stream at a time,"
+                f" handler {self.name} already received a manure stream.",
+                info_map,
+            )
+            raise ValueError("Handler cannot receive multiple streams.")
```
ew3361zh:
```suggestion
            raise ValueError("Non-parlor handler cannot receive multiple streams.")
```

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,256 @@
...
+            Current weather and environmental conditions that manure is being processed in.
+        time : Time
+            Time instance containing the simulations temporal information.
+
+        Returns
+        -------
+        dict[str, ManureStream]
+            Mapping between classification of manure coming out of this processor to the ManureStream containing the
+            manure information. If the processor is a separator, the classifications are "solid" and "liquid". Otherwise
+            the only classification is "manure".
```
ew3361zh:
```suggestion
            manure information.
```

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,256 @@
...
+        ----------
+        conditions : CurrentDayConditions
+            Current weather and environmental conditions that manure is being processed in.
+        time : Time
+            Time instance containing the simulations temporal information.
+
+        Returns
+        -------
+        dict[str, ManureStream]
+            Mapping between classification of manure coming out of this processor to the ManureStream containing the
```
ew3361zh:
```suggestion
            Mapping between classification of manure coming out of this handler to the ManureStream containing the
```

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,256 @@
...
+            "simulation_day": time.simulation_day,
+            "handler_type": self.handler_type,
+        }
+        if self.manure_stream is None or self.manure_stream.pen_manure_data is None:
+            self._om.add_error(
+                "None type ManureStream.",
+                "The processed ManureStream or pen data of the manure stream is None type.",
+                info_map,
+            )
+            raise TypeError("TypeError: Handler tries to process 'NoneType' object ManureStream.")
```
ew3361zh:
```suggestion
            raise TypeError("TypeError: Handler tried to process 'NoneType' object ManureStream or PenManureData.")
```

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,256 @@
...
+        Indication for if a parlor flush is used in addition to routine parlor water cleaning with fresh water.
+
+    Attributes
+    ----------
+    manure_stream : ManureStream
+        The ManureStream instance being checked for compatibility.
+    fresh_water_volume_used_for_milking : float
+        The amount of fresh water used for milking (L).
+    ammonia_emission : float
+        The amount of ammonia emission (kg).
```
ew3361zh: I was initially confused why this was an attribute since it is calculated by calling the processor base class method `calculate_ammonia_emissions()`. I see now it's calculated in the child class and since it's a little computationally expensive, it's set as an instance attribute rather than recalculating it (and I'm assuming it's expected to be the same value at both points it's needed).

Do you think we could make it more explicit in this attribute description where/when this is calculated and set? I think that would be helpful for future developer work.

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,256 @@
...
+        cleaning_water_volume = self.determine_cleaning_water_volume_in_main_barn(
+            self.manure_stream.pen_manure_data.num_animals,
+            self.cleaning_water_use_amount,
+            self.cleaning_water_recycle_fraction,
+        )
+        barn_temperature = self.determine_barn_temperature(conditions.mean_air_temperature)
+
+        total_cleaning_water_volume = (
+            cleaning_water_volume + self.fresh_water_volume_used_for_milking
+        ) * GeneralConstants.LITERS_TO_CUBIC_METERS
```
ew3361zh: I can't tell from the handler design doc if it explicitly says to extract this into a function but it might be helpful and would make things easier to test.

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,256 @@
...
+
+        self._om.add_variable("total_cleaning_water_volume", total_cleaning_water_volume, info_map_m3)
+        self._om.add_variable("barn_temperature", barn_temperature, info_map_c)
+
+        manure_water = self.manure_stream.water + (
+            total_cleaning_water_volume * GeneralConstants.WATER_DENSITY_KG_PER_M3
+        )
```
ew3361zh: another one that might be helpful to extract into a short helper function to make it easier to test and more maintainable if this equation needs to get expanded/adapted for different handler types.

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,256 @@
...
+                phosphorus=phosphorus,
+                potassium=potassium,
+                ash=ash,
+                non_degradable_volatile_solids=non_degradable_volatile_solids,
+                degradable_volatile_solids=degradable_volatile_solids,
+                volume=volume,
+                total_solids=total_solids,
+                pen_manure_data=None,
+            )
+        }
```
ew3361zh: Is there a reason to create these variables vs assigning these directly in the creation of the return ManureStream object?

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def determine_cleaning_water_volume_in_main_barn(
+        self, num_animals: int, cleaning_water_use_amount: float, cleaning_water_recycle_fraction: float
+    ) -> float:
+        """
+        Calculates the volume of fresh (non-recycled) cleaning water used for, and ultimately added to, a single manure
+         stream on a single simulation day by the manure handler. For parlor cleaning handlers, this water volume
+          represents an optional parlor flush (separate from fresh water only cleaning water). For all other handler
+           types, this water volume represents water use by handlers in the pen, such as a barn floor flush system.
```
ew3361zh: I'd move `For parlor cleaning handlers...barn floor flush system` into a Notes section.

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,256 @@
...
+                ash=ash,
+                non_degradable_volatile_solids=non_degradable_volatile_solids,
+                degradable_volatile_solids=degradable_volatile_solids,
+                volume=volume,
+                total_solids=total_solids,
+                pen_manure_data=None,
+            )
+        }
+
+    def determine_cleaning_water_volume_in_main_barn(
```
ew3361zh: Could this be a static function as well?

RUFAS/biophysical/manure/handler/single_stream_handler.py
```diff
@@ -0,0 +1,204 @@
...
+HOUSING_HSC = 260.0  # s/m
+"""
+Default housing specific constant (s/m). This constant may be used in calculations
+related to the housing conditions for animals. Default is set to 260.0 s/m.
+"""
+
+
+class SingleStreamHandler(Handler):
+    """
+    Base class for all handlers that only accepts single manure stream at a time.
```
ew3361zh:
```suggestion
    Base class for all handlers that only accept a single manure stream at a time.
```

RUFAS/biophysical/manure/handler/single_stream_handler.py
```diff
@@ -0,0 +1,204 @@
...
+
+    Parameters
+    ----------
+    name : str
+        Unique identifier of the processor.
+    handler_type: ProcessorTypes
+        The class of sun type of manure handlers that this handler falls into.
```
ew3361zh:
```suggestion
        The manure handler sub-type class into which this handler is categorized.
```

RUFAS/biophysical/manure/handler/single_stream_handler.py
```diff
@@ -0,0 +1,204 @@
...
+    cleaning_water_use_amount : float
+        Amount of cleaning water used per animal per day (L).
+    minutes_per_cleaning : int
+        Number of minutes needed per animal per cleaning, minutes.
```
ew3361zh:
```suggestion
        Number of minutes needed per animal per cleaning (minutes).
```

RUFAS/biophysical/manure/handler/single_stream_handler.py
```diff
@@ -0,0 +1,204 @@
...
+    cleanings_per_day : int
+        Number of cleanings per day.
+    cleaning_water_recycle_fraction : float
+        Fraction of cleaning water that is from recycled (not fresh) water sources.
```
ew3361zh:
```suggestion
        Fraction of cleaning water that is from recycled (not fresh) water sources (unitless).
```
I think because there are units on a few of these that it might be helpful to add units on all of the numeric ones.

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,256 @@
...
+    Attributes
+    ----------
+    manure_stream : ManureStream
+        The ManureStream instance being checked for compatibility.
+    fresh_water_volume_used_for_milking : float
+        The amount of fresh water used for milking (L).
+    ammonia_emission : float
+        The amount of ammonia emission (kg).
+    handler_type: ProcessorTypes
+        The class of sun type of manure handlers that this handler falls into.
```
elle-andreen:
```suggestion
        The type of manure handler.
```

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,256 @@
...
+
+        self._om.add_variable("total_cleaning_water_volume", total_cleaning_water_volume, info_map_m3)
+        self._om.add_variable("barn_temperature", barn_temperature, info_map_c)
+
+        manure_water = self.manure_stream.water + (
+            total_cleaning_water_volume * GeneralConstants.WATER_DENSITY_KG_PER_M3
+        )
+
+        manure_total_ammoniacal_nitrogen = max(0.0, self.manure_stream.ammoniacal_nitrogen - self.ammonia_emission)
+        nitrogen = self.manure_stream.nitrogen
```
matthew7838, line 165: Sure, but I'll have to create an issue to track it.

RUFAS/biophysical/manure/handler/single_stream_handler.py
```diff
@@ -0,0 +1,204 @@
...
+                f"Non parlor handler should only receive one manure stream at a time,"
+                f" handler {self.name} already received a manure stream.",
+                info_map,
+            )
+            raise ValueError("Handler cannot receive multiple streams.")
+        self.manure_stream = manure_stream
+
+    def process_manure(self, conditions: CurrentDayConditions, time: Time) -> dict[str, ManureStream]:
+        """
+        Executes the daily manure processing operations.
```
ew3361zh: Since the inheritance is a little confusing, I think it could be helpful to outline that this sub class version of process_manure() focuses on emissions calculations and then calls the parent process_manure() which focuses on water calculations and returning the manure_stream.

RUFAS/data_structures/processor_types.py
```diff
@@ -0,0 +1,41 @@
+from enum import Enum
+
+
+class ProcessorTypes(Enum):
+    """
+    This is an Enum class that represents different types of manure processors.
+
```
ew3361zh: I think because we have some open discussions on this and it's really only being used for typing in this PR, I think it would be better/simpler to leave this out and just set the type hints in the class attributes as strings for now.

ew3361zh (commented):
Nice work Matthew. I have some clarifications/questions/comments before moving forward but it feels relatively close to done.

RUFAS/biophysical/manure/handler/handler.py
matthew7838: Will do. I think that's a good point.

RUFAS/biophysical/manure/handler/handler.py
matthew7838: Just for cleanness since some require simple intuitive additions, I would prefer to keep them out of the final return. This is more of a personal preference, if you think a direct assignment is better I am fine doing it.

RUFAS/biophysical/manure/handler/handler.py
matthew7838: Yes, good catch.

RUFAS/biophysical/manure/handler/handler.py
elle-andreen: You beat me to this comment Niko - these are indeed unused and can be removed!

RUFAS/biophysical/manure/handler/handler.py
elle-andreen: Can we change this function name to `determine_handler_cleaning_water_volume` ? The "main_barn" part of the original name is a little confusing because this function is also used by the parlor cleaning handler, which doesn't happen in the barn (where the cows are housed) per se

RUFAS/biophysical/manure/handler/handler.py
matthew7838: Resoved with a different approach as discussed.

RUFAS/biophysical/manure/handler/handler.py
matthew7838: Good catch, thanks!

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,325 @@
...
+    name : str
+        Unique identifier of the processor.
+    handler_type: str
+        The class of sun type of manure handlers that this handler falls into.
+    cleaning_water_use_amount : float
+        Amount of cleaning water used per animal per day (L).
+    minutes_per_cleaning : int
+        Number of minutes needed per animal per cleaning, minutes.
+    cleanings_per_day : int
+        Number of cleanings per day.
```
elle-andreen: I think these can be removed, since these inputs were removed

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,325 @@
...
+class Handler(Processor):
+    """
+    Base class for all handlers.
+
+    Parameters
+    ----------
+    name : str
+        Unique identifier of the processor.
+    handler_type: str
+        The class of sun type of manure handlers that this handler falls into.
```
elle-andreen:
```suggestion
        The type of manure handler.
```

RUFAS/biophysical/manure/handler/single_stream_handler.py
```diff
@@ -0,0 +1,157 @@
...
+    name : str
+        Unique identifier of the processor.
+    handler_type: str
+        The manure handler sub-type class into which this handler is categorized.
+    cleaning_water_use_amount : float
+        Amount of cleaning water used per animal per day (L).
+    minutes_per_cleaning : int
+        Number of minutes needed per animal per cleaning (minutes).
+    cleanings_per_day : int
+        Number of cleanings per day.
```
elle-andreen: I think these can also be removed

RUFAS/biophysical/manure/handler/single_stream_handler.py
```diff
@@ -0,0 +1,157 @@
...
+        barn_temperature : float
+            Temperature of the barn (Celsius).
+
+        Returns
+        -------
+        float
+            Carbon dioxide emission from manure (kg).
+
+        """
+        return max(0.0, 0.0065 + 0.0192 * barn_temperature) * manure_deposition_surface_area / 1000
```
elle-andreen: I'm not sure where the division by 1000 is coming from - the manure deposition surface area value at least should also be in m2, unless this is another units or other conversion that I'm missing

elle-andreen (approved):
This looks great to me Matthew! I had a very minor comment or two, and I also decided to remove my TODO suggestion and just add it to the list of items to be addressed post-refresh. Thanks for your work on this!

RUFAS/biophysical/manure/handler/handler.py
ew3361zh: Going to reiterate I think it would be helpful to have both possibilities for the type error here but also the TypeError in the actual message itself is redundant.
```suggestion
            raise TypeError("Handler tries to process 'NoneType' object ManureStream or PenManureData.")
```

RUFAS/biophysical/manure/handler/handler.py
```diff
@@ -0,0 +1,359 @@
...
+        emission_info_map: dict[str, Any] = {
+            "class": self.__class__.__name__,
+            "function": self.process_manure.__name__,
+            "prefix": self._prefix,
+            "simulation_day": time.simulation_day,
+            "units": MeasurementUnits.KILOGRAMS,
+            "handler type": self.handler_type,
+        }
+
+        self.manure_stream = None
```
ew3361zh, line 183: Is there a difference between resetting this to None vs. resetting it to an empty stream?

ew3361zh: Oh nevermind, I see that if it's not none then it's possible there is another stream being received by a single stream handler.

RUFAS/biophysical/manure/handler/single_stream_handler.py
```diff
@@ -0,0 +1,153 @@
...
+
+        """
+        if self.manure_stream is None or self.manure_stream.pen_manure_data is None:
+            info_map = {"class": self.__class__.__name__, "function": self.process_manure.__name__}
+            self._om.add_error(
+                "None type ManureStream.",
+                "The processed ManureStream or pen data of the manure stream is None type.",
+                info_map,
+            )
+            raise TypeError("TypeError: Handler tries to process 'NoneType' object ManureStream.")
```
ew3361zh:
```suggestion
            raise TypeError("Handler tries to process 'NoneType' object ManureStream or PenManureData.")
```

ew3361zh (approved):
Looks good, Matthew, I'd just fix those TypeErrors but otherwise I think it's ready to get merged!

PR #2240, opened 2025-02-19
[Animal] 3/9 Animal Refresh: Ration and Nutritions

allisterakun:
### Context

### What

This PR updates the code to:

- Propagate process-based phosphorus into the NASEM/NRC requirement calculations and include it in the `NutritionRequirements` object.
- Refine supply calculations by adding new fields (like `wet_matter`, `digestible_energy_supply`, `starch_supply`, etc.) to the `NutritionSupply` structure.

### Why

- Integrating “process-based phosphorus” ensures that a separate, model-driven calculation of P requirements is considered alongside existing nutrient standards.
- Expanding the `NutritionSupply` fields fosters more robust ration formulations (including byproducts, wet matter, digestible energy, amino acids, etc.).

### How

- `nasem_requirements_calculator.py`, `nrc_requirements_calculator.py`
    - Each adds a parameter `process_based_phosphorus_requirement` to its `calculate_requirements` function.
    - The returned `NutritionRequirements` object has `process_based_phosphorus` set accordingly.
- `nutrition_supply_calculator.py`
    - New fields on `NutritionSupply` (e.g., `wet_matter`, `starch_supply`, `byproduct_supply`, etc.) are computed by the new internal `_calculate_nutritive_content`, `_calculate_digestible_energy`, `_calculate_byproducts_supply` methods.
    - The `nutrient_standard` logic influences whether to use `DE` or `DE_Base`.
- `amino_acid.py`
    - Replaces the old `TypedDict` with a `@dataclass` version, `EssentialAminoAcidRequirements`, adding `__add__` and `__truediv__` overrides for easier summation or averaging of these requirements.

### Test plan

Run the new unit tests for all the ration and nutrients related codes.


jadamchick:
Considering yet-to-come updates to ration formulation, at this stage reviewers should check for these capabilities: 
- matches user-defined ration exactly
- dry matter intake doesn't have wonky dips / hits expected values
- milk production reduces appropriately when it isn't supported by the ration but does not bottom out in some infinite loop 

cc @JoeWaddell 

Also a note - use of the user-defined ration is hard-coded at this point, even though the default value of ration: user_input is set to False in the animal input json

see also the discussion & updates [here](https://github.com/RuminantFarmSystems/MASM/pull/2251) and [here](https://github.com/RuminantFarmSystems/MASM/issues/2254)

tomhuhh:
Thanks Allister for working on this and thanks Julie for the detailed instructions. 

I spotted an error right away. Please see the screenshot below. For the record, I used all the default files as inputs. The filter I used is 'AnimalModuleReporter.report_ration_interval_data.ration_per_animal_for_pen_*'. My branch name is 'animal_refresh_09_simulation_engine_HH'.

Screenshot 2025-03-02 at 1 59 31 PM

Also, for some reason, enteric methane is not in the output variable list. 

Finally, when I increased the herd size from 100-cow to 1000-cow, I encountered the following type error. 

Screenshot 2025-03-02 at 2 25 05 PM


Looking forward to learning more about the refreshed module. Thanks!

RUFAS/biophysical/animal/nutrients/nutrition_supply_calculator.py
```diff
@@ -58,12 +66,17 @@ def calculate_nutrient_supply(
         calcium = cls._calculate_calcium_supply(feeds)
         phosphorus = cls._calculate_phosphorus_supply(feeds)
         dry_matter_intake = sum([feed.amount for feed in feeds])
+        wet_matter = sum([feed.amount / (feed.info.DM * GeneralConstants.PERCENTAGE_TO_FRACTION) for feed in feeds])
         protein = cls._calculate_metabolizable_protein_supply(
             feeds, dry_matter_intake, actual_tdn_percentages, body_weight
         )
-        ndf_content = cls._calculate_neutral_detergent_fiber_content(feeds)
+        nutrients_to_calculate = ["NDF", "EE", "CP", "ADF", "TDN", "lignin", "ash", "potassium", "starch"]
```
ew3361zh: I think this should be a class-level variable.

RUFAS/biophysical/animal/nutrients/nutrition_supply_calculator.py
```diff
@@ -615,27 +638,31 @@ def _calculate_rumen_undegradable_protein_percentages(
         return rup_percentages
 
     @classmethod
-    def _calculate_neutral_detergent_fiber_content(cls, feeds: list[FeedInRation]) -> float:
+    def _calculate_nutritive_content(cls, feeds: list[FeedInRation], nutrient: str) -> float:
         """
-        Calculates the neutral detergent fiber (NDF) content of a ration.
+        Calculates the content of a specific nutrient ration.
```
ew3361zh:
```suggestion
        Calculates the content of a specific nutrient in a ration.
```

RUFAS/biophysical/animal/nutrients/nutrition_supply_calculator.py
```diff
@@ -645,21 +672,39 @@ def _calculate_forage_neutral_detergent_fiber_content(cls, feeds: list[FeedInRat
...
+            Total supply of nutrient in a ration (Mcal).
+
+        """
+        de_attribute = "DE_Base" if cls.nutrient_standard is NutrientStandard.NASEM else "DE"
+        return sum([feed.amount * getattr(feed.info, de_attribute) for feed in feeds])
+
+    @classmethod
+    def _calculate_byproducts_supply(cls, feeds: list[FeedInRation]) -> float:
+        """
+        Calculates amount of byproducts in a ration.
```
ew3361zh:
```suggestion
        Calculates the amount of byproducts in a ration.
```

RUFAS/biophysical/animal/ration/user_defined_ration_manager.py
```diff
@@ -0,0 +1,104 @@
...
+
+class UserDefinedRationManager:
+    """
+    Handles the initialization and management of user-defined animal rations.
+
+    Attributes
+    ----------
+    user_defined_rations : dict[AnimalCombination, dict[RUFAS_ID, float]]
+        Map of animal groupings to ration formulations, where the ration formulation is defined by a mapping from the
+        RuFaS ID of a feed to the percentage it makes up in a ration.
```
ew3361zh:
```suggestion
        A mapping of animal groupings to their respective ration formulations. 
    Each ration formulation is represented as a dictionary, where the key is the 
    RuFaS ID of a feed and the value is the percentage it contributes to the ration.
```

RUFAS/biophysical/animal/ration/user_defined_ration_manager.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        info_map = {"class": cls.__class__.__name__, "function": cls.set_user_defined_rations.__name__}
+
+        cls.user_defined_rations = {
+            AnimalCombination.CALF: {},
+            AnimalCombination.GROWING: {},
+            AnimalCombination.CLOSE_UP: {},
+            AnimalCombination.LAC_COW: {},
+        }
+        user_defined_ration_percentages = ration_config["user_defined_ration_percentages"]
```
ew3361zh, line 41: Is ration_config a list or a dict? Here it looks like it's a dict but the docstrings and type hints have it as a list of dicts.

RUFAS/biophysical/animal/ration/user_defined_ration_manager.py
```diff
@@ -0,0 +1,104 @@
...
+
+        """
+        info_map = {"class": cls.__class__.__name__, "function": cls.set_user_defined_rations.__name__}
+
+        cls.user_defined_rations = {
+            AnimalCombination.CALF: {},
+            AnimalCombination.GROWING: {},
+            AnimalCombination.CLOSE_UP: {},
+            AnimalCombination.LAC_COW: {},
+        }
```
ew3361zh:
```suggestion
        cls.user_defined_rations = {ac: {} for ac in AnimalCombination}
```
Not sure if we specifically don't want user_defined_rations for the other AnimalCombinations but this is cleaner if we do.

RUFAS/biophysical/animal/ration/user_defined_ration_manager.py
```diff
@@ -0,0 +1,104 @@
...
+                in user_defined_ration_percentages[combination.value]
+            }
+
+        invalid_ration_found: bool = False
+        for animal_combo, ration in cls.user_defined_rations.items():
+            total_percentage_of_ration = sum(ration.values())
+            info_map["ration"] = ration
+            info_map["animal_combination"] = animal_combo.value
+            info_map["units"] = MeasurementUnits.PERCENT
+            if abs(total_percentage_of_ration - 100.0) > 0.1:
```
ew3361zh: what is this threshold tolerance of 0.1? Should it be extracted? Maybe defined at a class level?

ew3361zh (approved):
Really nice work, Allister! Some suggestions I think would be good to address but nothing big enough to prevent me from approving.

RUFAS/biophysical/animal/ration/user_defined_ration_manager.py
allisterakun: Updated the type hint for ration_config to `dict[str, dict[str, list[dict[str, int | float]]]]`

jadamchick:
-	The inputs for ration: user input (false – in the animal .json) and UDR tolerance (0.10 – in the feed .json) are misleading, since the actual ration formulation on the refresh is not using user-defined input and allows 0.0 tolerance around it. 
-	At minimum, can we set these to reflect what is actually happening? (user input = true and tolerance = 0)
-	Ideally there should also be something to prevent the user from thinking that they can change these and expect a functional difference in the model?

tomhuhh:
Hi Allister,

Thank you for your patience. I completed some tests, and while most of the results matched my expectations, I did notice a few discrepancies that I'd like to discuss. Here’s a summary of what I did:

- Tested the default inputs on the refreshed branch.
- Ran the c2f8 feed input file along with other default inputs on the refreshed branch.
- Executed the default inputs on the dev branch with user input ration set to true.

And here are my observations:

1. On the refreshed branch, the inclusion rates perfectly align with the user-defined ratio.
2. For lactating cows, milk production is reduced appropriately. Using the default input file on the refreshed branch, I observed that the milk reduction stays within 0.5 kg/d, which meets our expectations.
3. Milk production reduction is also appearing as an output for close-up animals (close-up heifers and dry cows), even though they shouldn't have any milk production.
Screenshot 2025-03-23 at 9 41 03 AM

4. Dry matter intake behaves as expected, except during the first ration interval where lactating cows show much lower intake. I noticed this with both the default and c2f8 diets. Has this come up before? Could someone remind me of our previous conclusions?
Screenshot 2025-03-23 at 9 42 01 AM

5. The dry matter intake estimates for lactating cows on the refreshed branch are lower than those on the dev branch, which is expected since the refreshed branch results are more accurate.
Screenshot 2025-03-23 at 9 42 58 AM
Screenshot 2025-03-23 at 9 45 42 AM

6. Body weight estimates are quite similar across both branches, though the growing body weight on the refreshed branch is slightly higher than on the dev branch.
Screenshot 2025-03-23 at 9 44 15 AM

Lastly, just a friendly reminder that Julie's comments above might need to be addressed.

Looking forward to your thoughts on these points!

allisterakun:
> 3\. Milk production reduction is also appearing as an output for close-up animals (close-up heifers and dry cows), even though they shouldn't have any milk production.

We kept the milk production reduction record for a dry cow when she was a lactating cow; that's why this value is non-zero.



> 4\. Dry matter intake behaves as expected, except during the first ration interval where lactating cows show much lower intake. I noticed this with both the default and c2f8 diets. Has this come up before? Could someone remind me of our previous conclusions?

Right, @JoeWaddell, @jadamchick, and I noticed this issue a while ago: the ratio for the lac cow pen takes longer to reach a stable state. We could not figure out the reason at the moment.

> Lastly, just a friendly reminder that Julie's comments above might need to be addressed.

This was implemented on the 08 branch for feed storage here: https://github.com/RuminantFarmSystems/MASM/pull/2245/files#diff-c6cbc9020a9b639b28d5cc79411e7b96790470724dff7996340162eb8b38e941

tomhuhh:
Thanks, Allister. I'm approving the PR, but I'm still curious about these model behaviors and would love to keep the conversation going to learn more.

> We kept the milk production reduction record for a dry cow when she was a lactating cow; that's why this value is non-zero.

Could you help me understand the reasoning behind keeping this value for close-up animals and where it’s being used? I'm also curious whether the milk reduction is applied to individual cows or the entire pen. The alignment between the output (0.5, lactating cow column) and user-defined threshold of 0.5 suggests that every cow in the pen experiences reduction in milk production. However, in my view, if diets are formulated based on the pen's average requirements, then only the cows needing more energy and protein might be underfed. Thus, the output should be lower than 0.5, unless the diet does not meet the requirements of the animals in that pen at all. Maybe in that case, the user should get a warning that they need to enter a better user defined ration. This might be out of the scope of this work, but I am just curious about people's thoughts on this. 
 

> This was implemented on the 08 branch for feed storage here: https://github.com/RuminantFarmSystems/MASM/pull/2245/files#diff-c6cbc9020a9b639b28d5cc79411e7b96790470724dff7996340162eb8b38e941

Thanks for clarifying!

tomhuhh (approved):
Great work, Allister! Thanks again for your patience.

allisterakun:
> > We kept the milk production reduction record for a dry cow when she was a lactating cow; that's why this value is non-zero.
> 
> Could you help me understand the reasoning behind keeping this value for close-up animals and where it’s being used? I'm also curious whether the milk reduction is applied to individual cows or the entire pen. The alignment between the output (0.5, lactating cow column) and user-defined threshold of 0.5 suggests that every cow in the pen experiences reduction in milk production. However, in my view, if diets are formulated based on the pen's average requirements, then only the cows needing more energy and protein might be underfed. Thus, the output should be lower than 0.5, unless the diet does not meet the requirements of the animals in that pen at all. Maybe in that case, the user should get a warning that they need to enter a better user defined ration. This might be out of the scope of this work, but I am just curious about people's thoughts on this.
> 

Happy to explain! 

After a conversation with @JoeWaddell, we decided to use the pen average nutrient requirements for ration evaluation and reduce the milk on the pen level but keep the option open for individual animal reduction. We can do that by creating a function in the Animal class for milk reduction, which we call at the pen level for every animal in the pen.

### Animal.reduce_milk_production() -> bool
The is the function to reduce milk production in the Animal class:
```
    def reduce_milk_production(self) -> bool:
        """
        Attempts to reduce milk production.

        Returns
        -------
        bool
            True if the reduction was successful, False otherwise.

        """
        is_milk_reduction_too_high = (
            self.milk_production.milk_production_reduction + AnimalModuleConstants.MILK_REDUCTION_KG
        ) > AnimalConfig.milk_reduction_maximum
        if is_milk_reduction_too_high is True:
            return False
        self.milk_production.milk_production_reduction += AnimalModuleConstants.MILK_REDUCTION_KG
        return True
```
This function tries to reduce the milk production for an animal by `AnimalModuleConstants.MILK_REDUCTION_KG = 0.25` every time it gets called. It ensures the overall reduction afterward does not exceed the user-define maximum milk reduction `AnimalConfig.milk_reduction_maximum`. If it does exceed, this function will return False, indicating the maximum milk reduction has been reached for this animal.
One note here is that the milk reduction for each animal is never reset; that's why you're seeing non-zero milk reductions for dry cows.

### Pen.reduce_milk_production() -> bool
This is the function to reduce milk production for all animals in the pen:
```
    def reduce_milk_production(self) -> bool:
        """
        Attempts to reduce the milk production of all animals in the pen.

        Returns
        -------
        bool
            False if all animals in the pen have already reached the maximum reduction, True otherwise.
        """
        is_production_reduced: list[bool] = []
        for animal in self.animals_in_pen.values():
            is_production_reduced.append(animal.reduce_milk_production())
        return any(is_production_reduced)
```
This function will iterate through all animals in the pen to call their `reduce_milk_production()` function to reduce their milk production. The result from each animal's milk reduction attempt is collected into an array. After going through all the animals in the pen, this function will check the output array by calling `any()`  on the array. This is checking if we have successfully reduced the milk production for at least one animal in the pen, returning Ture. Otherwise, we will return False, indicating all animals in this pen have reached the user-defined milk reduction maximum. 

Hope this explanation helps!

PR #2241, opened 2025-02-19
[Animal] 4/9 Animal Refresh: Animal

allisterakun:
### Context

### What

This PR provides a comprehensive rework of the `Animal` class and its associated tests:

- Introduces extensive logic to manage lifecycle transitions for calves, heifers (I, II, III), and cows.
- Standardizes daily routines (growth, phosphorus requirements, reproduction, milking, digestion) in a single daily_routines flow.
    - The phosphorus requirements update and the manure excretion calculations are moved from the Pen class into Animal class.
- Adds robust property-based getters/setters to handle life-stage–dependent logic (e.g., `is_pregnant`, `days_in_pregnancy`, `is_milking`).
- Implements dedicated initialization pathways (`_initialize_newborn_calf`, `_initialize_calf_or_heiferI`, `_initialize_heiferII_or_heiferIII`, `_initialize_cow`) to cleanly configure animals.
- Replaces old static references with config-driven reproduction logic, bridging with new typed dictionaries (NewBornCalfValuesTypedDict, etc.).

### Why

- Facilitates clarity and maintainability by unifying lifecycle transitions, daily routine updates, and typed dictionary setups.
- Enhances test coverage .
- Removes any extraneous or scattered logic for the animal’s lifecycle, focusing it under one class with explicit daily methods.
- Improves the simulation by supporting transitional steps (e.g., HeiferIII -> Cow) with new checks (e.g., evaluate_heiferIII_for_cow).

### How

- Initialization Overhaul:
    - Condensed each life stage’s initialization details into separate private methods, called conditionally based on AnimalType.
    - For newborn calves, _assign_sex_to_newborn_calf references semen_type from AnimalConfig; integrates stillbirth and selling logic via random checks.
    - Heifer/cow paths include new repro setup (HeiferReproductionProtocol, CowReproductionProtocol), culling date calculators, and calving intervals.
- Lifecycle & Daily Routines:
    - `daily_routines`: aggregates `_daily_nutrients_update`, `_daily_digestive_system_update`, `daily_milking_update`, `daily_growth_update`, plus `reproduction_update` logic.
    - `animal_life_stage_update`: a dispatch method that chooses a specialized `_X_life_stage_update` method per type (calf, heifer, cow).
        - If the animal gives birth to a newborn calf, sets the appropriate cull/death day and reason.
- New Properties:
    - `days_in_milk`, `days_in_pregnancy`, `future_cull_date`, `is_milking`, etc., have specialized logic to protect consistency (e.g., calves can’t have days_in_pregnancy, automatically set to 0).
    - `sold` and `dead` properties reference `sold_at_day` and `dead_at_day` for consistent checks.

### Test plan

Run the unit tests for the Animal class


Isgotkowitz:
Updated the milk production reduction process so that it uses the reduction amount kept in `AnimalConfig`, instead of the newly-removed constant in `AnimalModuleConstants`

RUFAS/biophysical/animal/animal.py
```diff
@@ -1,2 +1,2409 @@
...
+                self.previous_nutrition_supply.ndf_supply / previous_dmi * GeneralConstants.FRACTION_TO_PERCENTAGE
+            )
+            tdn_percentage = (
+                self.previous_nutrition_supply.tdn_supply / previous_dmi * GeneralConstants.FRACTION_TO_PERCENTAGE
+            )
+            net_energy_diet_conc = (
+                self.previous_nutrition_supply.metabolizable_energy
+                / previous_dmi
+                * GeneralConstants.FRACTION_TO_PERCENTAGE
+            )
```
allisterakun: Currently, the previous ration plays a factor in the animal nutrition requirements calculation, which can in turn affect the final ration. 
If you want to evaluate the effect of previous ration, please comment out this block of code and use only the default values (on lines 2346-2349, within the `if self.previous_nutrition_supply is None:` statement).

allisterakun: See detailed discussion about this on PR #2251

RUFAS/biophysical/animal/animal.py
```diff
@@ -1,2 +1,2409 @@
...
+from RUFAS.data_structures.feed_storage_to_animal_connection import NutrientStandard, Feed
+from RUFAS.general_constants import GeneralConstants
+from RUFAS.time import Time
+
+
 class Animal:
-    pass
+    """
+    DO NOT USE THE PROPERTIES THAT START WITH '_'. INSTEAD, USE THE FUNCTIONS THAT ARE DECORATED WITH @property.
+
```
matthew7838: Should add another short line for animal description.

RUFAS/biophysical/animal/animal.py
```diff
@@ -1,2 +1,2409 @@
...
+    body_condition_score_5: float
+        The body condition score on a scale of 1 to 5, (unitless).
+    cull_reason: str
+        The reason for the animal to leave the herd.
+    body_weight_history: list[BodyWeightHistory]
+        The body weight history of the animal.
+    pen_history: list[PenHistory]
+        The pen history of the animal.
+    sold_at_day: int | None
+        The simulation day in which the animal was sold, (simulation day).
```
matthew7838: Not quite so sure that simulation day is a unit, if not perhaps we can remove it?
```suggestion
        The simulation day in which the animal was sold.
```

RUFAS/biophysical/animal/animal.py
```diff
@@ -1,2 +1,2409 @@
...
+        Returns
+        -------
+        None
+
+        """
+        cls.nutrient_standard = nutrient_standard
+
+    @classmethod
+    def setup_lactation_curve_parameters(cls, time: Time) -> None:
+        """
```
matthew7838: No class info is needed in this function, I think a static method is better. Is there a specific reason for this to be a class method?

RUFAS/biophysical/animal/animal.py
```diff
@@ -1,2 +1,2409 @@
...
+
+    @property
+    def days_in_pregnancy(self) -> int:
+        """
+        The total number of days an animal has been in pregnancy.
+
+        Notes
+        -----
+        - For animals of type CALF or HEIFER_I, the pregnancy duration is always considered to be zero.
+        - For all other types of animals, the value of `_days_in_pregnancy` is returned.
```
matthew7838: Nitpicky thing. I looked up the Numpy style guide and usually, the notes section comes at the end.

RUFAS/biophysical/animal/animal.py
```diff
@@ -1,2 +1,2409 @@
...
+        ----------
+        days_in_pregnancy : int
+            The number of days the animal has been in pregnancy.
+
+        Raises
+        ------
+        TypeError
+            If the animal type is either CALF or HEIFER_I.
+
+        """
```
matthew7838, line 231: You did have a return section for None in previous methods. This is not a big thing but I think it should be consistent across the file to either have a return section for None or not.

RUFAS/biophysical/animal/animal.py
```diff
@@ -1,2 +1,2409 @@
...
+        if self.animal_type in [AnimalType.CALF, AnimalType.HEIFER_I]:
+            raise TypeError()
+        self.reproduction.heifer_reproduction_program = heifer_reproduction_program
+
+    @property
+    def heifer_reproduction_sub_program(self) -> HeiferTAISubProtocol | HeiferSynchEDSubProtocol:
+        """
+        heifer_reproduction_sub_program property
```
matthew7838:
```suggestion
        heifer_reproduction_sub_program property.
```

RUFAS/biophysical/animal/animal.py
```diff
@@ -1,2 +1,2409 @@
...
+            If `AnimalConfig.semen_type` is not "conventional" or "sexed".
+
+        """
+        if AnimalConfig.semen_type == "conventional":
+            male_calf_rate = AnimalConfig.male_calf_rate_conventional_semen
+        elif AnimalConfig.semen_type == "sexed":
+            male_calf_rate = AnimalConfig.male_calf_rate_sexed_semen
+        else:
+            raise ValueError(f"Unexpected semen type: {AnimalConfig.semen_type}")
+        self.sex = Sex.MALE if random() < male_calf_rate else Sex.FEMALE
```
matthew7838: Should these need to add an error to the OM?

RUFAS/biophysical/animal/animal.py
```diff
@@ -1,2 +1,2409 @@
...
+
+        Returns
+        -------
+        NewBornCalfValuesTypedDict or None
+            A dictionary containing details related to a newly born calf if a calf is born during this update;
+            otherwise, None.
+
+        Returns
+        -------
+        None
```
matthew7838:
```suggestion
```

RUFAS/biophysical/animal/animal.py
```diff
@@ -1,2 +1,2409 @@
...
+                if (
+                    AnimalConfig.death_day_probability[i]
+                    <= death_date_random
+                    < AnimalConfig.death_day_probability[i + 1]
+                ):
+                    death_lower_limit = AnimalConfig.death_day_probability[i]
+                    death_upper_limit = AnimalConfig.death_day_probability[i + 1]
+                    death_time_lower_limit = AnimalConfig.cull_day_count[i]
+                    death_time_upper_limit = AnimalConfig.cull_day_count[i + 1]
+            n = (death_time_upper_limit - death_time_lower_limit) / (death_upper_limit - death_lower_limit)
```
matthew7838: Is there a better name for this?

RUFAS/biophysical/animal/animal.py
```diff
@@ -1,2 +1,2409 @@
...
+        if last_pen is None or last_pen != current_pen:
+            self.pen_history.append(
+                PenHistory(
+                    start_date=current_day,
+                    end_date=current_day,
+                    pen=current_pen,
+                    animal_types_in_pen=list(animal_types_in_pen),
+                )
+            )
+        else:  # last_pen == curr_pen
```
matthew7838:
```suggestion
        else:
```

RUFAS/biophysical/animal/animal.py
```diff
@@ -1,2 +1,2409 @@
...
+                self.body_weight,
+                AnimalConfig.wean_day,
+                AnimalConfig.wean_length,
+                available_feeds,
+                self.nutrient_standard,
+            )
+            calf_requirements = CalfRationManager.calc_requirements(
+                self.days_born, self.body_weight, previous_temperature, calf_intake
+            )
+            # TODO: do not use dummy values for calf calcium and phosphorus requirements - issue pending.
```
matthew7838, line 2204: Just a reminder to add the issue number if there's one.

RUFAS/biophysical/animal/animal.py
```diff
@@ -1,2 +1,2409 @@
...
+            requirements = NASEMRequirementsCalculator.calculate_requirements(
+                body_weight=self.body_weight,
+                mature_body_weight=self.mature_body_weight,
+                day_of_pregnancy=days_in_pregnancy,
+                body_condition_score_5=self.body_condition_score_5,
+                days_in_milk=days_in_milk,
+                average_daily_gain_heifer=self.growth.daily_growth,
+                animal_type=self.animal_type,
+                parity=self.calves,  # TODO: calves
+                calving_interval=self.calving_interval,  # TODO: calving interval
```
matthew7838: Need more clarity on these todos.

matthew7838 (commented):
Look good overall. Left a few comments regarding mostly regarding docstrings. However, I have two things to bring up.

1. Should we consider adding errors to OM in some cases? I think it's fine for setter to not do this but might want to consider doing this for other parts of the code when throwing errors.
2. If there are any references please add them in the docstrings.

jadamchick:
I ran the default herd with 1000 cows on both the refresh (09) branch and on Dev (with UDR=true and tolerance=0 to match the Refresh ration constraints). I summarized the results for the last year of the sim, ran each 4x, and averaged the results. 

1.	Parity structure, turnover, and heifer III sales/purchases 
-	The Refresh herd has an older parity structure (23% L4+ v. 16% on dev), more heiferIII’s sold and fewer bought, and lower turnover. This looks like it is all driven by lower number of deaths and non-production sales in the Refresh herd. Given the input risks by parity (of death, culling) and the number of cows per parity in each herd, it appears that Refresh is much closer to the expected number of death and sales. 
-	Possibly the refresh is addressing some issues that were buried in Dev?
-
image

summary results & filter attached: 
[report_pilotfarms_nocrops_18Feb2025.json](https://github.com/user-attachments/files/19220622/report_pilotfarms_nocrops_18Feb2025.json)

[12Mar_ dev vs refresh 1000cows 4runs.xlsx](https://github.com/user-attachments/files/19220623/12Mar_.dev.vs.refresh.1000cows.4runs.xlsx)

matthew7838 (approved):
LGTM from the dev side's POV.

jadamchick:
From Allister @allisterakun : 
There was an error on dev when calculating the future_death_date, that will always set the future death date to be the next day if an. animal is determined to die. We have fixed this in animal refresh. And I believe this is the cause for this behavior
To verify this, I fixed the error on dev, and run a simulation with 1000 cows (on dev), and the results look much closer to the refresh but some values are sort of in between, see report below. However, the parity structure doesn't seem to be affected.

[fixed_future_death_dev.csv](https://github.com/user-attachments/files/19247494/fixed_future_death_dev.csv)

From Julie: I think it is hard to differentiate between meaningful difference and random noise here. Maybe the death date scheduling causes a difference in the # of deaths in the last year of the simulation, bc animals who otherwise would only be scheduled to have a future death date, in the refresh, have already had their death event occur on Dev. It doesn't explain differences in cull rate. 

image

None of these differences in herd structure seem to be impacting important outputs (replacement herd size, milk allocation factor are similar even despite the heiferIII dynamics). 
Plan: Allister did a thorough review of all of the death and sold logic; Julie will do the same when reviewing for documentation. We'd also value the input of any other SME's for thoughts on this @krb78 @KFosterReed

KFosterReed (approved):
I read through this new file and stepped through several of the functions including the ones related to culling, death, and the life stage updates. All seem to be working correctly and with Julie's demographics thumbs up I think this is good to go!

PR #2246, opened 2025-02-19
[Animal] 9/9 Animal Refresh: SimulationEngine

allisterakun:
### Context

## What

This PR removes the old `Feed` class (and related `feed/__init__.py`) in favor of the newer, centralized `FeedManager`. It also integrates manure-handling and animal-lifecycle updates by:

- Deleting the obsolete `biophysical/feed/feed.py` file and references.
- Short-circuiting the manure manager routine when `manure_request_results` is None to prevent errors.
- Adjusting composting logic (`_calculate_dry_matter_loss`) to rename the methane parameter for clarity.
- Removing enteric methane fields in `PenManure` (no longer needed).
- Updating `SimulationEngine` to replace the deprecated `AnimalManager` + `Feed` usage with `HerdManager` + `FeedManager`.

## Why

These changes unify feed-related operations under a single manager, streamline the manure flow, and remove dead code. Specifically:

- The `Feed` class was largely replaced by FeedManager, so the old file was redundant.
- Certain manure routines needed minor guard checks (handling None returns) to prevent crashing.
- Enteric methane calculations and reporting now happen in the animal module, so we drop them from `PenManure '.
- The `SimulationEngine` daily loop now uses `HerdManager` and `FeedManager` exclusively, improving maintainability and clarity.

### Feed Removal
The `feed/` directory and the `Feed` class are deleted; references replaced with `FeedManager`.
### Manure Manager
Added a None check in `_record_manure_request_results` plus minor parameter renaming in `composting.py`.
### PenManure
Removed `enteric_methane_kg` and cleaned up the factory method (`get_instance`) to reference typed attributes.
### SimulationEngine
- New `_formulate_ration` method that calls `FeedManager` for degradations, obtains updated inventory, and triggers `HerdManager` ration formulation.
- Replaces all `animal_manager` usage with `herd_manager` calls in the daily simulation loop.
- Schedules periodic feed recalculations and ration intervals.

### Test plan

Wrote a complete unit test suite for `SimulationEngine`, which covers 100% of the code added in this PR.


RUFAS/routines/manure/manure_manager.py
```diff
@@ -374,6 +374,9 @@ def _record_manure_request_results(
         manure_source : str
             The source of the manure.
         """
+        # TODO: remove this when #2209 is merged
+        if manure_request_results is None:
+            return
```
ew3361zh: Should this be removed? #2209 is merged.

RUFAS/simulation_engine.py
```diff
@@ -123,27 +124,78 @@ def _run_simulation_main_loop(self) -> None:
...
+        if is_time_to_recalculate_max_daily_feeds is True:
+            crops_to_get_next_harvest_dates = [
+                crop for crop in self.feed_manager.crop_to_rufas_id.keys() if crop not in next_harvest_dates.keys()
+            ]
+            next_harvest_dates = self.field_manager.get_next_harvest_dates(crops_to_get_next_harvest_dates)
+            self.next_max_daily_feed_recalculation = self.time.current_date + self.max_daily_feed_recalculation_interval
+
+        if next_harvest_dates != {}:
+            total_inventory = self.feed_manager.get_total_inventory(
+                self.time.current_date.date(), self.weather, self.time)
```
ew3361zh, line 147: Can this call to get total_inventory be extracted from here along with line 168 and done once?

RUFAS/simulation_engine.py
```diff
@@ -123,27 +124,78 @@ def _run_simulation_main_loop(self) -> None:
...
-        self.animal_manager.daily_updates(self.feed, self.weather, self.time)
-        all_pen_manure_data = self.animal_manager.collect_pen_manure_data()
-        self.manure_manager.daily_update(all_pen_manure_data, self.animal_manager.simulation_day)
         manure_applications = self.generate_daily_manure_applications()
         harvested_crops = self.field_manager.daily_update_routine(self.weather, self.time, manure_applications)
+        next_harvest_dates: dict[str, date | None] = {}
         for harvested_crop in harvested_crops:
             self.feed_manager.receive_crop(harvested_crop.harvested_crop, harvested_crop.storage_type)
-        routines.daily_feed_routine(self.feed, self.field_manager, self.animal_manager)
+            if harvested_crop.harvested_crop.config_name not in next_harvest_dates.keys():
```
```suggestion
            if harvested_crop.harvested_crop.config_name not in next_harvest_dates:
```
Avoids creating unnecessary dict keys view object.

RUFAS/simulation_engine.py
```diff
@@ -123,27 +124,78 @@ def _run_simulation_main_loop(self) -> None:
...
+                crop_config_name = harvested_crop.harvested_crop.config_name
+                next_harvest_date = self.field_manager.get_next_harvest_dates([crop_config_name])
+                next_harvest_dates[harvested_crop.harvested_crop.config_name] = next_harvest_date.get(crop_config_name)
+
+        is_time_to_recalculate_max_daily_feeds = self.next_max_daily_feed_recalculation == self.time.current_date
+        if is_time_to_recalculate_max_daily_feeds is True:
```
```suggestion
        if is_time_to_recalculate_max_daily_feeds:
```

RUFAS/simulation_engine.py
```diff
@@ -123,27 +124,78 @@ def _run_simulation_main_loop(self) -> None:
...
+            crops_to_get_next_harvest_dates = [
+                crop for crop in self.feed_manager.crop_to_rufas_id.keys() if crop not in next_harvest_dates.keys()
+            ]
+            next_harvest_dates = self.field_manager.get_next_harvest_dates(crops_to_get_next_harvest_dates)
```
ew3361zh, line 142: I think there's a potential issue here since this will overwrite the next_harvest_dates dict entirely and it's being updated earlier on line 135. I don't know if that's intentional but if so maybe make it clearer that that is the intent. For example move the check on is_time_to_recalculate_max_daily_feeds and don't do any of the looping through the harvested crops.
Otherwise, maybe the intent is just to update the dict and add this stuff to what's already in the next_harvest_dates dict at this point? If so it should be switched to dict.update()

ew3361zh (approved):
Looks good, just a few suggestions!

RUFAS/simulation_engine.py
allisterakun: I think the total inventory here is the current inventory. If it is time to recalculate max daily feeds, we will purchase some feed for the next cycle according to the current total inventory.
The "total_inventory" in line 168 is the updated inventory (potentially after the making the purchase). So I think it's better to keep it this way

ew3361zh: Ok, sounds good. I think this is good for now.

jadamchick (approved):
Nice work Allister & team!!

PR #2247, opened 2025-02-19
[Crop & Soil] ManurePool ZeroDivisionError for Surface Machine Manure Application

### LINKED ISSUE #2199: [Crop & Soil][Manure] Divide By Zero Error in C&S ManurePool when Animals not simulated and manure schedule srf set to 1.
In testing some manure application code, a divide by zero bug was discovered that arises when you don't simulate animals and adjust some manure schedule inputs (exact scenarios laid out below).

## Expected Behavior

We should be able to run simulations successfully with or without animals being simulated.

## Current Behavior

It appears these settings (`application_depths` = 0 and `surface_remainder_fractions` = 1) will trigger a reset in both `application_depth` AND `surface_remainder_fraction` for a manure application or a fertilizer application.
If you set the inputs at the values above there is code to catch those specific settings in Field._execute_manure_application() and Field._execute_fertilizer_application() that looks like this:
```
invalid_depth_and_remainder_fraction = (application_depth == 0.0 and surface_remainder_fraction != 1.0) or (
                application_depth > 0.0 and surface_remainder_fraction == 1.0
            )

            error_name = "manure_application_error"
            if invalid_depth_and_remainder_fraction:
                self._record_nutrient_application_error(
                    application_depth, surface_remainder_fraction, error_name, year, day
                )
                application_depth = 0.0
                surface_remainder_fraction = 1.0
```
And that new combination of values for `application_depth` and `surface_remainder_fraction` leads to that divide by zero error downstream.
If you adjust the reset values in that error to what's in the default manure input files as default `application_depth `(150) and default `surface_remainder_fraction` (0.8), then the error downstream doesn't occur.

So it looks like we intentionally try to catch those values above but then reset them to values that cause other errors down the line in the `ManurePool` class `_determine_moisture_change()` function. I am not familiar with why those errors were set up in that way and if they're still necessary and/or accurate but it does seem worth exploring if that's still how we want to reset them and if we do want to reset them to those values, how can we avoid the divide-by-zero error in `ManurePool._determine_moisture_change()`.

## Steps to Recreate
1. In `default_config.json` set `simulate_animals` to `false`
2. In `default_manure_schedule.json` set the 4th from the bottom `surface_remainder_fractions` to `1` AND/OR set the 4th from the bottom `application_depths` to `0`.
3. Run a simulation.

example of step 2 in `default_manure_schedule.json` jic:
```
"application_depths": [
    150.0,
    150.0,
    150.0,
    150.0,
    150.0,
    150.0,
    150.0,
    150.0,
    150.0,
    150.0,
    150.0,
    0,
    150.0,
    150.0,
    150.0
  ],
  "surface_remainder_fractions": [
    0.80,
    0.80,
    0.80,
    0.80,
    0.80,
    0.80,
    0.80,
    0.80,
    0.80,
    0.80,
    0.80,
    1,
    0.80,
    0.80,
    0.80
  ],
```

allisterakun:
### Context

Issue(s) closed by this pull request: closes #2199 

### What

Fix the `ZeroDivisionError` described in issue #2199. 

### Why

The ZeroDivisionError occurs because the manure_applied_mass property of SoilData.machine_manure is not properly set. The machine_manure property is an instance of ManurePool, initialized using default_factory, which means manure_applied_mass defaults to 0.0.

Typically, manure_applied_mass is assigned a value at the end of ManureApplication.apply_machine_manure(), following sub-surface manure application. However, a conditional check before this application can cause the function to exit early if no sub-surface manure application defined by user-input. As a result, manure_applied_mass remains at its default 0.0, leading to the ZeroDivisionError when later referenced.

### How

To fix this issue, simply changing the order of operation to assign the `manure_applied_mass` before the condition check which could lead to early exit of the function.

### Test plan

1. The input has been modified to reproduce the "ZeroDivisionError". Simply run a simulation as usual, it should complete without any issues. This indicates the bug has been fixed.
2. Revert the changes in the input files (`default_config.json` and `default_manure_schedule.json`) to run an E2E test simulation. It should also pass without issues. This indicates that the changes made in this PR does not affect other variables.
3. Play with the values of `application_depth` and `suface_remainder_fraction` in `default_manure_schedule.json` to run a simulation. The simulation should complete with no issues under any circumstances.


input/data/config/default_config.json
allisterakun, line 1: The changes to this file will be reverted before merging.

input/data/manure_schedule/default_manure_schedule.json
allisterakun, line 1: The changes to this file will be reverted before merging.

KFosterReed (approved):
I tested this under several combinations of manure application methods:
- depth = 0, suface remainder fraction = 1
- depth = 0, suface remainder fraction = 0
- depth = 0, suface remainder fraction = 0.8
- depth >0 , suface remainder fraction = 1
- depth >0, suface remainder fraction = 0
- depth > 0, suface remainder fraction = 0.8

Alll ran and resulted in a successful manure application that was adjusted to a meaningful depth and surface remainder fraction given the user inputs

ew3361zh (approved):
Nice work Allister!

PR #2251, opened 2025-02-21
[Animal][Ration] Debugging lactating and close-up animal rations

Isgotkowitz:
This is a working branch made in order to debug some funky behavior of animals' rations brought to light by @jadamchick.

Action items that will start being addressed in this PR:
- [x] Remove `ration_driver.py`
- [x] Make and report user-friendly failed ration constraint report based on `NutritionEvaluationResults`.
  - Two variables are now reported
    - `"AnimalModuleReporter._report_average_nutrient_evalualtion_results.avg_eval_results_pen_{pen id}_{animal combination}"`
    - `"AnimalModuleReporter._report_average_nutrient_evalualtion_results.avg_eval_report_pen_{pen id}_{animal combination}"`
  - The "report" provides boolean values indicating whether the ration failed or passed for a certain nutritional requirement.
  - The "results" provides the differences between the nutrition supply and requirement. These values are confusing because some nutrient supplies just need to be above a certain threshold, and others need to be in a certain range.
  - The nutrients that need to be within in a certain range are metabolizable protein, NDF %, Fat %, and dry matter. If the value for these is 0, it means they were within the acceptable range. If they are less than zero, it means that they were beneath the minimum threshold, and vice versa.
  - This is still a very rough reporting of `NutritionEvaluationResults`, all suggestions for improvements are very much welcomed.
- [x] Find out why close-ups' ration is changing along w/ lactating animals' ration when the lactating cow rations are changes, and fix if necessary.
  - The Close-Up animal outcomes change when changes are made to the lactating animals' ration because the nutrition requirement calculations for an animal factor in the nutrient supply of the previous ration.
  - This effect can be verified by modifying `Animal.calculate_nutrition_requirements` so that the previous NDF %, TDN %, and net energy diet concentration values are hardcoded.
  - I believe Joe mentioned this might be the reason, but I couldn't find it here or in slack so it must have been in person.
- [x] Find out why larger milk reductions lead to higher lactation requirement for lactating cows, and fix if necessary.
  - Welp when the milk production reduction was applied, it was being added instead of subtracted from the amount of milk produced 😭
  - I have changed this and the net energy lactation requirement for lactating cows is much lower.

CC @allisterakun @JoeWaddell


jadamchick:
Thanks Ed! 
1. For the failed constraint ration report, maybe it makes sense to wait until the full ration formulation (i.e., optimization is possible)? I think that doing something very similar to what @JoeWaddell had set up in handle_failed_constraints() in the old RationManager would be great if possible. 

2. Close-up rations being impacted by lact cow ration changes: that makes sense. I would want input from @KFosterReed , @krb78  or someone else more ration-savvy to gauge whether it has meaningful impact or is fine to leave as-is? I will post another comment below summarizing my observations

3. Addition v. subtraction: great find!

jadamchick:
I ran this branch under 3 scenarios: 
1. default_feed.json
2. change lact cow ration to 100% feed 88 (grass hay); milk reduction max 0.5 kg; all other rations unchanged
3. change lact cow ration to 100% feed 88 (grass hay); milk reduction max 20.0 kg; all other rations unchanged

Observations: 
1. milk production reduction and directionality of NEl_requirement looks much better now! 
2. When lact cow ration changes: a) lact cow DMIest requirement does not change; b) close-up DMIest requirement does change

My understanding (as Ed said above) is that the close-up DMI change is because the previous supply impacts the current requirements. So close-ups who were on distinct lact cow rations have distinct close-up requirements. 

Questions: 
1. Why, under the same logic, does the lact cow requirement not change when the ration changes? 
2. Is this change to close-up ration significant enough to care about? it is roughly a 20% change in DMI even with no intentional change to close-up ration

image

Report filter attached


@Isgotkowitz Ed, my input on the tasks for this PR: 
- ration driver: check!
- constraint eval report: as I said above, maybe this makes sense to wait until can have one summary that is useful for when the optimizer is working and there are those iterations included as well; and would use the previous RationManager's handle_failed_constraints() as a model if possible
- ration changes: I'll make an issue to flag this for continued investigation, but since you've at least figured out why it is happening I don't think it needs to hold up this PR
- milk reduction going in the wrong direction: fixed! 
[report_milkreductioneval.json](https://github.com/user-attachments/files/18986993/report_milkreductioneval.json)

jadamchick (approved):

jadamchick:
Another note -- at @allisterakun's suggestion I ran the same scenarios on dev, with these results (I set ration: user_input to True, so it would use the UDR): 

image

Note that in this case, the close-up ration is unchanged, indicating that there is something different happening in the refresh version

Issue #2065, opened 2024-11-14
Report summary stats of animal population file used for herd initialization


KFosterReed:
**Is your feature request related to a problem? Please describe.**
It is helpful to know the characteristics of the animal population that is used to initialize the herd but we don't currently report any of this information to the output manager

**Describe the solution you'd like**
I would like to add to the summary stats that are created by the `get_herd_summary()` function in `animal_population.py` to include some more useful stats like average bodyweight for each animal class, average milk production, and the number of cows in each parity. Before resolving this issue we should poll other SMEs to see what additional summary stats would be useful

**Describe alternatives you've considered**
we could create a standalone task to summarize the animal population file

**Additional context**

**Target Deadline**
Soon after animal module refresh is complete

matthew7838:
This issue is blocked by #2041.

allisterakun:
@KFosterReed please advise the desired herd summary statistics.

KFosterReed:
Initial thoughts:

- average BW for each animal type
- distribution of ages for each animal type
- number of dry and lactating cows
- number of cows in each parity

@jadamchick @krb78 @JoeWaddell any others you can think of?

KFosterReed:
@tomhuhh also if you have any requests to add, please do!

allisterakun:
Just confirming that we are interested in the summary statistics for the final animal population used for the simulation. Is that correct?

tomhuhh:
> Just confirming that we are interested in the summary statistics for the final animal population used for the simulation. Is that correct?

Correct, that's my understanding as well. Regarding the statistics outputs for the initialized animal population, I would consider extending the request to include counts for all animal types (calves, growing heifers, close-up heifers, lactating cows, and dry cows) instead of just dry and lactating cows.

allisterakun:
Also, what does "distribution of ages for each animal type" mean? Is it simply the average age for each animal type?

JoeWaddell:
>what does "distribution of ages for each animal type" mean?

There's interest in knowing whether the ages approximate a uniform distribution, versus a normal distribution, for instance. That's going to be a little more challenging to summarize in a digestible way for most, but once I've pushed my local commits from #2252 I can help explore how we can institute this kind of check.

allisterakun:
[...]

Issue #2042, opened 2024-10-22
[Animal][HerdFactory] Understand and document HerdFactory at boundary conditions


ew3361zh:
After investigation/research from issue #1395, we have some guidelines for different boundary conditions which affect the success of herd initialization. However, there are multiple combinations of these factors which are still providing somewhat unpredictable results.

With HerdFactory being redesigned and updated in #2040 and #2041, further investigation is needed to understand how it behaves at these boundaries.

This issue will be marked as blocked by #2041.

allisterakun:
Will be closed by #2252

KFosterReed:
SME Grooming 7/24/2026 marking as closed by #2252

Issue #2041, opened 2024-10-22
[Animal][HerdFactory] HerdFactory Refresh


allisterakun:
As we refresh the animal module, we are moving away from the inheritance chain: `AnimalBase` -> `Calf` -> `HeiferI` -> `HeiferII` -> `HeiferIII` -> `Cow` to a single `Animal` class. The function logic in `HerdFactory` must also be updated to accommodate this structure change.

allisterakun:
Closing this issue as the Animal Refresh has been merged by PR #2235. The refreshed animal module should resolve this issue.

PR #2252, opened 2025-02-25
Herd_init_distributions

### LINKED ISSUE #2040: [Animal][HerdFactory] Update Herd Generation Methods
Currently the herd generation methods randomly select animals from within each animal class for herd generation. We would like to increase the factors that are considered when conducting this random selection so that the initial herd more closely represents the initial herd that the user wants to specify. Some example of these additional factors would be the number of dry and lactating cows, the number of cows in each parity, stage of lactation, and the age of the youngstock. 

We are working on a complete design doc: [linked here](https://docs.google.com/document/d/1-7thRsVcQBnooYt4lcVgtNfj_rm56bMYRIVoH_MIYMQ/edit?usp=sharing)

But an initial, incomplete description of the proposed solution is the following: 

We propose to add criteria or factors that inform which animals are selected by the herd generation methods. Some of these factors could be drawn from existing inputs and all of the additional factors should be applied in addition to the existing criteria that determine the number of animals in each animal class that are selected. 
For example, if the herd generation inputs are: 
"herd_information": {
		"calf_num": 10,
		"heiferI_num": 55,
		"heiferII_num": 45,
		"heiferIII_num_springers": 7,
		"cow_num": 100,
		"replace_num": 1000,
		"herd_num": 100,
		"breed": "HO",
		"parity_fractions": {
      "1": 0.35,
      "2": 0.25,
      "3": 0.40
    },
    "annual_milk_yield": 15207837
	},

The proposed updates would use the parity_fractions and the cow_num to inform the number of cows within each parity that are selected. So instead of selecting a 100 cows with a random distribution of parities, the new method would select 35 cows from the first lactation, 25 cows from the second lactation and 40 cows with 3 or greater parity. 

None yet

**Additional context**
With the current functionality the initialized herd can be substantially different from the desired herd composition and is part of the reason a 4 year simulation is required before a steady state herd can be achieved. This also presents a challenge and increases reliance on availability and accuracy of input data accuracy, specifically with respect to animal culling and reproductive management, which can be challenging to collect. Adding more granularity to the attributes of the animals that inform the herd initialization process will enable the randomly selected herd to more closely reflect the desired herd composition and reduce the time to achieve a steady state herd.

**Target Deadline**

January 2025

JoeWaddell:
### Context

Issue(s) closed by this pull request: closes #2040 and see also: #2041, #2042, and #2065

### What

See Design Doc here: https://docs.google.com/document/d/1-7thRsVcQBnooYt4lcVgtNfj_rm56bMYRIVoH_MIYMQ/edit?usp=sharing

### Why

New initialization inputs/categories will improve the degree to which simulated herd reflects real life herds.

### Needs
Additional subcategories, e.g. dry and lactating for each parity.

### Test plan

Needs additional unit tests.

Addition of a method to check the herd distribution throughout the simulation for each input category, if not reflected in the existing outputs (e.g. new subsets added in the future).


jadamchick:
Hey there,
I tried this out to see if it performs as expected (and as an excuse to get more familiar with initialization). 

**What I did**
I made a new branch as a copy of this one and then merged in dev to get the `report_animal_population_statistics` functionality (just had to manually resolve the conflict to keep calves>6 rather than calves>4 in the `_cows_update` method in `herd_factory.py`).

I ran a selection of Joe's test herds and added a couple of my own -- one with a fairly realistic cow demographic structure, and one with bad math (added up to 109%) just to compare how they perform. 

I used a csv filter of `^AnimalModuleReporter.report_animal_population_statistics.*`

**Results**
Default & default_10x (use_detailed_values = false for both) were as expected (Post demographics similar to pre)

However, any herd with use_detailed_values = true has: 
- Fewer cows than specified in cow_num
- By the exact number that are expected to be L5+ based on the parity_5_fraction 

I checked the csv output of `AnimalModuleReporter's report_daily_animal_population` directly on Joe's branch, and it shows the same (for example, 949 cows on day 1 of the simulation of the herd_tweaking scenario). -- just verifying that it wasn't something related to merging in dev on my branch

I didn't dig back into the code to troubleshoot, but it seems pretty consistent across different parity fractions & cow numbers that the L5+ cows are not being drawn as part of the herd. 

**A few more notes**
- I also tried "bad math" to have the parity fractions add up to 109%.. but since my excess was in the L5+ fraction, it was ignored. Based on the code, I expected it to build a herd of 1090 cows rather than 1000
- Any thoughts about how to reconcile these inputs with the lactation curve inputs asking for the same information? (parity_fractions in the animal.json and milking_cow_fraction in the lactation_curve_adjustment_inputs.json)
- To try to start playing with the stated purpose ("improve the degree to which simulated herd replicates real life herds"), I built a filter to report values over varying times in the simulation. Nothing earth-shattering, just showing how the simulation reverts to the equilibrium as a result of parameter values regardless of initial conditions. I'm working out a few bugs in the report filter but will share it here once I get those ironed out. I can also try to build it out with more variables (actual herd production or footprint?) if it would be of interest. For now, here are the summarized results for herd_tweaking scenario

image

RUFAS/biophysical/animal/animal_module_reporter.py
```diff
@@ -918,10 +918,14 @@ def report_life_cycle_manager_data(cls, herd_statistics: HerdStatistics, simulat
...
         parity_2 = herd_statistics.num_cow_for_parity["2"]
         parity_3 = herd_statistics.num_cow_for_parity["3"]
+        parity_4 = herd_statistics.num_cow_for_parity["4"]
+        parity_5 = herd_statistics.num_cow_for_parity["5"]
         parity_greater_than_3 = herd_statistics.num_cow_for_parity["greater_than_3"]
         om.add_variable("num_cow_for_parity_1", parity_1, dict(info_map, **{"units": MeasurementUnits.ANIMALS}))
         om.add_variable("num_cow_for_parity_2", parity_2, dict(info_map, **{"units": MeasurementUnits.ANIMALS}))
         om.add_variable("num_cow_for_parity_3", parity_3, dict(info_map, **{"units": MeasurementUnits.ANIMALS}))
+        om.add_variable("num_cow_for_parity_4", parity_4, dict(info_map, **{"units": MeasurementUnits.ANIMALS}))
+        om.add_variable("num_cow_for_parity_5", parity_5, dict(info_map, **{"units": MeasurementUnits.ANIMALS}))
```
jadamchick, line 968: Maybe helpful to add this distinction to report_animal_population_statistics() as well, so can easily see the breakdown for the initialized herd

input/metadata/properties/default.json
```diff
@@ -71,6 +71,51 @@
         "default": 100,
         "minimum": 6
       },
+      "cow_detailed_num": {
+        "type": "object",
+        "description": "detailed cow initialization distribution",
+        "use_detailed_values": {
+          "type": "bool",
+          "description": "True to use the detailed partity values in herd initalization.",
```
jadamchick: typo -- detailed **parity** values

jadamchick:
Thanks for the updated animal population -- with the population provided the cow numbers are now on target. (I did not yet try generating my own population on this branch since the updates) 

A couple minor comments: 
1. It looks like it can still be off by 1 in each subgroup due to rounding / partial cows; this caused me to have 998 instead of 1000 cows in one scenario. For example: 
* when % lact = 0.85 and % L2 = 0.35
* Lactating L2's: 0.35 * 1000 * 0.85 = 297.5 --> int(297.5) = 297
* Dry L2's: 0.35 * 1000 * 0.15  = 52.5 --> int(52.5) = 52
* 297 + 52 = 349 rather than 350 

That seems pretty minor to me, but just flagging that as a reason that the numbers might not match exactly 

2. If the inputted fractions don't add to 1, you wind up with a different number of total cows than the target cows (it just does the math individually for each subgroup without checking or reconciling with the total) -- maybe worth feedback to the user that this has happened? 

And here is a report filter that can be used / modified to see initial conditions and changes over the course of the simulation (pre and post population values, 1st month, 12th month, 1st year, 5th year). Where relevant (for summed values), it annualizes the monthly values (multiplies the 31-d value by 11.774 to get a 365-d equivalent). 

[report_inititalandslices_15Apr.json](https://github.com/user-attachments/files/19798176/report_inititalandslices_15Apr.json)


Here is the type of info reported (for the herd_tweak file): 
image

KFosterReed:
Thanks for the filter Julie! Super helpful! And with the updated rounding function, this is working just as I would expect!

I think this is looking really good! 

I tested scenarios in particular - one that reduced the animal bodyweight input and one that changed the age at first calving to see how those inputs would affect the differences in the animal population outputs over time. Not exactly a test of this method but a test of the desired impact of this method. Overall, year 5 is still different than year 1 for most herd demographic and production outcomes but it is still an improvement over the previous fully random methods. 

Here is a google doc with the outputs from Julie's filter if you want to check it out: [Herd Init Scenarios](https://docs.google.com/spreadsheets/d/1uQMirMHx4gjIi6X78spMybkVKnDTquoubta85ypWWds/edit?usp=sharing)

Once the inputs have been updated and Logiag is happy with it, I am ready to sign off on this!

JoeWaddell:
> Thanks for the updated animal population -- with the population provided the cow numbers are now on target. (I did not yet try generating my own population on this branch since the updates)
> 
> A couple minor comments:
> 
>     1. It looks like it can still be off by 1 in each subgroup due to rounding / partial cows; this caused me to have 998 instead of 1000 cows in one scenario. For example:
> 
> 
>     * when % lact = 0.85 and % L2 = 0.35
> 
>     * Lactating L2's: 0.35 * 1000 * 0.85 = 297.5 --> int(297.5) = 297
> 
>     * Dry L2's: 0.35 * 1000 * 0.15  = 52.5 --> int(52.5) = 52
> 
>     * 297 + 52 = 349 rather than 350
> 
> 
> That seems pretty minor to me, but just flagging that as a reason that the numbers might not match exactly

Thanks Julie! Kristan and I were able to fix this yesterday: it was an issue of my not realizing/remembering that `int` "rounds down" all decimals, so I just added a proper rounding of the floats prior to int conversion. That _mostly_ gets rid of the issue of cow numbers being a bit off - the other potential reason would be in cases where there are more decimals rounded down than up. That won't necessarily be a common occurrence, but we'll just have to make sure to alert users to this possibility. We could also add a warning, if everyone agrees it's a good idea.

I've also updated the inputs here: they're now merged with the parity fractions section, with `lactating_fraction` separate (we can tweak the name). I'm going to continue to tidy the logic so it's a bit easier to follow, and continue updating unit tests. Thanks again for comments and a good filter!

jadamchick:
> I've also updated the inputs here: they're now merged with the parity fractions section, with lactating_fraction separate (we can tweak the name). I'm going to continue to tidy the logic so it's a bit easier to follow, and continue updating unit tests. Thanks again for comments and a good filter!

Sounds great, and the animal.json file looks cleaner and easier to follow. 
One more comment on the inputs --  It looks as though the lactation_curve.py will need to be updated as well? my suggestion would be: 
a) remove `milking_cow_fraction` from lactation_curve_adjustment_inputs.json -- replace the reference in the code in lactation_curve.py with the new `lactating_fraction`
b) update `parity_3_percentage` in lactation_curve.py to be a sum of parity 3, 4, and 5 fractions (since the previous input was parity 3 or more)

And on that note, it might be even more clear if the input name is changed from` parity_5_fraction` to `parity_5_and_greater_fraction` or something similar.

RUFAS/biophysical/animal/data_types/animal_population.py
```diff
@@ -287,6 +314,46 @@ def get_replacement_cows(self) -> list[Animal]:
...
+    def cows_parity_4_not_milking(self) -> List[Animal]:
+        return [cow for cow in self.cows if cow.calves == 4 and not cow.is_milking]
+
+    @property
+    def cows_parity_5_milking(self) -> List[Animal]:
+        return [cow for cow in self.cows if cow.calves == 5 and cow.is_milking]
+
+    @property
+    def cows_parity_5_not_milking(self) -> List[Animal]:
+        return [cow for cow in self.cows if cow.calves == 5 and not cow.is_milking]
```
ew3361zh, line 361: There's a lot of repeated logic here - I might suggest having it in a single function where you pass in parity and the `is_milking` status.

Also, we're using the built-in `list` type-hint instead of importing it so it might look something like:

```
def get_cows_by_parity_and_milking(self, parity: int, is_milking: bool) -> list[Animal]:
    return [cow for cow in self.cows if cow.calves == parity and cow.is_milking == is_milking]
```

RUFAS/biophysical/animal/data_types/animal_population.py
```diff
@@ -173,7 +173,24 @@ class AnimalPopulation:
 
     def __post_init__(self) -> None:
         """Post init function to find the max id of all animals, and set the current_animal_id"""
-        all_animals = self.calves + self.heiferIs + self.heiferIIs + self.heiferIIIs + self.cows + self.replacement
+        all_animals = (
```
ew3361zh, line 182: Can you explain this a little? Is there overlap between `self.cows` and `self.cows_for_parity_xyz`? And why is it different from the `all_animals` in the `get_herd_summary()` method later in the class?

RUFAS/biophysical/animal/herd_factory.py
```diff
@@ -2,7 +2,7 @@
 import datetime
 import random
 from pathlib import Path
-from typing import Any
+from typing import Any, Dict, List
```
ew3361zh: Both Dict and List have built-in types (`dict` and `list`) that are recommended to use for more recent versions of Python.

RUFAS/biophysical/animal/herd_factory.py
```diff
@@ -505,27 +536,89 @@ def _random_sample_with_replacement_by_type(self, animal_type: str) -> list[Anim
...
+        else:
+            ANIMAL_NUM_KEY: dict[str, str] = {
+                "calf": "animal.herd_information.calf_num",
+                "heiferI": "animal.herd_information.heiferI_num",
+                "heiferII": "animal.herd_information.heiferII_num",
+                "heiferIII": "animal.herd_information.heiferIII_num_springers",
+                "cow": "animal.herd_information.cow_num",
+                "replacement": "animal.herd_information.replace_num",
+            }
+            animal_num = self.im.get_data(ANIMAL_NUM_KEY[animal_type])
```
ew3361zh: Not urgent here imo but I think this could be cleaned up/simplified to reduce duplication and increase clarity.

ew3361zh (approved):
LGTM, Joe, nice work! 

I left a couple questions but nothing that needs addressing here if it's working as intended for herd generation expectations. I do think the code could be cleaned up for clarity in a separate PR as we discussed.

RUFAS/biophysical/animal/data_types/animal_population.py
JoeWaddell: Great suggestion: I'll open an issue for this! Should be an easy change, but will want to be sure to have some good unit test coverage for it. Also we may want to add more tweaking to the suggested method (to include some amount of filtering based on age), so worth opening a separate Issue & subsequent PR. See Issue #2377

RUFAS/biophysical/animal/herd_factory.py
JoeWaddell: Also tagged in Issue #2377

RUFAS/biophysical/animal/data_types/animal_population.py
JoeWaddell: I've updated `get_herd_summary` to include up to parity 5. I also didn't want to entirely get rid of the .cows attribute in case it would be needed in the future, or is used elsewhere currently. I can also tag this in Issue #2377, in case it's now been made entirely redundant.

jadamchick (commented):
Looks good and seems to be working as expected. Two comments on things that don't affect function of this PR but I think would help make RuFaS less confusing -- maybe can flag for future updates if too much to take on now? 

a- change the name from "parity 5" to "parity 5 and greater" or "parity greater than 4" to make it clear that it includes parities 6, 7, etc. as well 

b- the lactation curve calculations reference a "milking_cow_fraction" (line 255 in lactation_curve.py) that is currently an input in lactation_curve_adjustment_inputs.json.  Since users are now prompted to input the actual proportion of the herd that is milking cows ("lactating_fraction"), can we update that lactation curve reference to be the new user-input value?

JoeWaddell:
> Looks good and seems to be working as expected. Two comments on things that don't affect function of this PR but I think would help make RuFaS less confusing -- maybe can flag for future updates if too much to take on now?

I think Issues makes the most sense, to keep the PR's objectives more narrowly focused.

> a- change the name from "parity 5" to "parity 5 and greater" or "parity greater than 4" to make it clear that it includes parities 6, 7, etc. as well

In most cases here, they are actually just the parity 5 animals, not parity 5+. In herd generation, for example, parity 6 or greater animals are strictly excluded. In some reporting methods I included the 3+ group as it already was, to not break filters, and also to lump together the higher parity animals.

> b- the lactation curve calculations reference a "milking_cow_fraction" (line 255 in lactation_curve.py) that is currently an input in lactation_curve_adjustment_inputs.json. Since users are now prompted to input the actual proportion of the herd that is milking cows ("lactating_fraction"), can we update that lactation curve reference to be the new user-input value?

I'll make the lactation tweak a separate issue, since making the lactating fraction user-defined has more implications than extending the existing parities, and could/should require the addition of more unit tests.

jadamchick (approved):
approved!

PR #2255, opened 2025-02-26
[Output Manager] Fixes Bug For Reporting CSV Units

### LINKED ISSUE #2224: [OutputManager] OM not properly recording units when saving to CSVs
## Expected Behavior

When saving to a CSV in OM, units should be able to be added to the column headers properly.

## Current Behavior

Currently it does not seem to work to add units to the columns and an error is logged because there is no subkey to obtain the correct units.

## Possible Solution

Look into the implementation of `OM._dict_to_csv_column_list()` and ensure this is properly functioning. This function is calling the `OM._get_units_substr()` function and that's where the error is occurring/being logged.

## Steps to Reproduce

1. Create a csv filter - I used a filter to collect data on manure request results logs:
```
{
    "name": "ManRequest-Farm",
    "filters": [
      "ManureManager._record_manure_request_results.on_farm_manure.*"
    ],
    "variables": [".*"]
}
```
2. Run a simulation.
3. Errors occur looking like this:
Image
4. The CSV is still saved properly with the values intact but the units aren't added to the variable names.

ew3361zh:
Accounts for reporting units properly to CSVs generated by OM when units are reported to OM as a dictionary.

### Context

Issue(s) closed by this pull request: closes #2224

### What

Fixes the error

### Why

Previously this was causing an error because there was no check within the `_get_units_substr()` method for if the variable name was in the units dictionary prior to checking if the subkey was None.

### How

Adds a check in `OM._get_units_substr()` before error is raised for if there is no `subkey`.
The check is for whether the `variable_name` is in the `units` dict. If it is, it will extract the correct units for that variable and return it.

### Test plan

This is the example CSV filter from the issue that wasn't properly reporting units:
```
{
    "name": "ManRequest-Farm",
    "filters": [
      "ManureManager._record_manure_request_results.on_farm_manure.*"
    ],
    "variables": [".*"]
}
```
Running a simulation with it now correctly reports the units to the CSV column headers.

I also wanted to make sure this wasn't interfering with the existing other possible calls to this `_get_units_substr()` method. This filter triggers the other pathway in `OM._dict_to_csv_column_list()` that calls it:
```
{
    "name": "sim eng",
    "filters": [
      ".*available_feeds_on_final_day.*"
    ]
  }
```

I would recommend running a simulation with at least both of these csv filters and checking the results. They were both working as expected for me.

### DIFF
--- a/RUFAS/output_manager.py
+++ b/RUFAS/output_manager.py
@@ -1002,6 +1002,9 @@ class OutputManager(object):
         if not isinstance(units, dict):
             return f" ({units})" if units else ""
 
+        if variable_name in units:
+            return f" ({units[variable_name]})"
+
         if subkey is None:
             self.add_error(
                 "units_subkey_missing",
--- a/changelog.md
+++ b/changelog.md
@@ -111,6 +111,7 @@ v0.9.2
 - [2236](https://github.com/RuminantFarmSystems/MASM/pull/2236) - [minor change] [Testing] Changes E2E expected results update task to automatically update all expected results regardless of whether the old and new expected results differ.
 - [2233](https://github.com/RuminantFarmSystems/MASM/pull/2233) - [minor change] [Manure] Adds a base manure Digester class.
 - [2247](https://github.com/RuminantFarmSystems/MASM/pull/2247) - [minor change] [Crop & Soil] Fix the ZeroDivisionError for surface machine manure application.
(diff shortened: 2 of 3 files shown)


matthew7838 (approved):
LGTM, outputs are looking as expected.

allisterakun (approved):
Works as expected! LGTM! Great work, Niko!

PR #2273, opened 2025-03-05
[Manure] Re-implement Anaerobic Digestion Logic

### LINKED ISSUE #1120: Replace hard-coded values in manure module with constants
There are a series of hard coded magical numbers in the manure module. Check with a SME on each of them and determine if they can be extracted as constants to the manure constants file.

If so, update their usage accordingly (i.e. replace the magic numbers with the constants).

- [ ] In MASM > RUFAS > routines > manure > manure_treatments > anaerobic_digestion.py there is a magic number (`4.0`) used to determine what to return from the `_bound_influent_temperature()` method:

```
def _bound_influent_temperature(cls, average_temperature_celsius: float) -> float:
        """Returns the max between the given average temperature and temperature bound.

        Args:
            average_temperature_celsius: Average daily temperature, C.

        """
        return max(average_temperature_celsius, 4.0)
```

- [ ] In MASM > RUFAS > routines > manure > manure_treatments > anaerobic_digestion.py there are several magic numbers in the return value of the `_calc_manure_heat_capacity()` method:
```
def _calc_manure_heat_capacity(
        cls, average_temperature_celsius: float, moisture_content: float
    ) -> float:
        """Returns heat capacity of manure.

        Args:
            average_temperature_celsius: Average daily temperature, C.
            moisture_content: A 0-1 decimal representing water content of manure.

        Returns:
            Heat capacity of manure, kJ /kg /C.

        """
        return (
            0.68298
            + 0.025662 * average_temperature_celsius
            + 0.01306 * moisture_content * 100
        )
```
- [ ] In MASM > RUFAS > routines > manure > manure_treatments > base_manure_treatment.py the calculation for `final_manure_volume` in the `_initialize_daily_output_during_update` method uses a magic number (1000):
```
final_manure_volume = (
            manure_treatment_daily_input.liquid_manure_daily_volume
            - (
                manure_treatment_daily_input.liquid_manure_total_solids
                * self.config.total_solids_removal_efficiency_for_treatment
            )
            / 1000.0
        )
```

### LINKED ISSUE #2102: [Manure] Re-implement `AnaerobicDigestion` logic
Re-implement the `AnaerobicDigestion` logic. Use the "Developer Guidelines" in #2092.

Isgotkowitz:
Re-implements the logic for modeling anaerobic digestion.

### Context

Issue(s) closed by this pull request: closes #2102, closes #1120

### What

Re-implements the logic for anaerobic digestion in the new `AnaerobicDigester` class, which is a child of the base `Digester` class.

Also adds in the previous defaults for an anaerobic digester into a new manure management configuration file, which will be used to track refreshed manure inputs.

### Why

This logic needs to be reimplemented in order for RuFaS to be able to simulate anaerobic digesters with the refreshed manure module.

### How

Fills in the logic for the `AnaerobicDigester` class, including `receive_manure` and `process_manure`.

`receive_manure` is extremely simple, it checks if manure received is compatible and if so, adds it to the manure that needs to be digested on the current day.

`process_manure` is a re-implementation of the daily logic from the pre-refresh `AnaerobicDigestion`. It takes all the manure that was received on the current day, processes all the physical and chemical changes to the held manure, records the changes and the changed manure, then passes all the manure on.

### Test plan

Wrote a complete unit test suite for `AnaerobicDigester`, which covers 100% of code added in this PR.

### DIFF
--- a/RUFAS/biophysical/manure/digester/anaerobic_digester.py
+++ b/RUFAS/biophysical/manure/digester/anaerobic_digester.py
@@ -257,6 +257,7 @@ class AnaerobicDigester(Digester):
         )
         self._om.add_variable("manure_nitrogen", self._manure_to_digest.nitrogen, info_map_kg)
         self._om.add_variable("manure_phosphorus", self._manure_to_digest.phosphorus, info_map_kg)
+        self._om.add_variable("manure_potassium", self._manure_to_digest.potassium, info_map_kg)
         self._om.add_variable("manure_ash", self._manure_to_digest.ash, info_map_kg)
         self._om.add_variable(
             "manure_non_degradable_volatile_solids", self._manure_to_digest.non_degradable_volatile_solids, info_map_kg
--- a/RUFAS/biophysical/manure/digester/anaerobic_digester.py
+++ b/RUFAS/biophysical/manure/digester/anaerobic_digester.py
@@ -56,7 +56,7 @@ class AnaerobicDigester(Digester):
 
     Attributes
     ----------
-    _received_manure : ManureStream
+    _manure_to_digest : ManureStream
         The total amount of manure received by an anaerobic digester in a single day.
     _temperature_set_point : float
         Temperature set point for the anaerobic digestion (degrees C).
@@ -82,7 +82,7 @@ class AnaerobicDigester(Digester):
         evaporation_fraction: float,
     ) -> None:
         super().__init__(name=name, is_housing_emissions_calculator=False)
-        self._received_manure: ManureStream = ManureStream.make_empty_manure_stream()
+        self._manure_to_digest: ManureStream = ManureStream.make_empty_manure_stream()
         self._temperature_set_point: float = temperature_set_point
         self._hydraulic_retention_time: int = hydraulic_retention_time
         self._top_cover_volume_fraction: float = top_cover_volume_fraction
@@ -94,11 +94,11 @@ class AnaerobicDigester(Digester):
         is_received_manure_valid = self.check_manure_stream_compatibility(manure)
         if is_received_manure_valid is False:
             raise ValueError(f"Anaerobic digester {self.name} received an invalid manure stream.")
-        self._received_manure += manure
+        self._manure_to_digest += manure
 
     def process_manure(self, conditions: CurrentDayConditions, time: Time) -> dict[str, ManureStream]:
         """Digests manure received on the current day."""
-        if self._received_manure.is_empty is True:
+        if self._manure_to_digest.is_empty is True:
             self._report_anaerobic_digester_outputs(
                 biogas=0.0,
                 biogas_energy_content=0.0,
@@ -112,36 +112,36 @@ class AnaerobicDigester(Digester):
             )
             return {}
 
-        moisture_fraction = self._received_manure.water / self._received_manure.mass
+        moisture_fraction = self._manure_to_digest.water / self._manure_to_digest.mass
         specific_input_energy = self._calculate_specific_input_energy(
             conditions.mean_air_temperature, moisture_fraction, self._temperature_set_point
         )
         heating_input_energy = (  # TODO: wrong unit conversion factor?
-            specific_input_energy * self._received_manure.volume * GeneralConstants.LITERS_TO_CUBIC_METERS
+            specific_input_energy * self._manure_to_digest.volume * GeneralConstants.LITERS_TO_CUBIC_METERS
         )
 
-        minimum_volume = self._received_manure.volume * self._hydraulic_retention_time
-        top_cover_volume = self._received_manure * self._top_cover_volume_fraction
+        minimum_volume = self._manure_to_digest.volume * self._hydraulic_retention_time
+        top_cover_volume = self._manure_to_digest * self._top_cover_volume_fraction
 
-        self._received_manure.ammoniacal_nitrogen = min(
-            self._received_manure.ammoniacal_nitrogen * TAN_INCREASE_FACTOR, self._received_manure.nitrogen
+        self._manure_to_digest.ammoniacal_nitrogen = min(
+            self._manure_to_digest.ammoniacal_nitrogen * TAN_INCREASE_FACTOR, self._manure_to_digest.nitrogen
         )
 
-        generated_methane_volume = self._calculate_CSTR_methane_volume(self._received_manure.total_volatile_solids)
+        generated_methane_volume = self._calculate_CSTR_methane_volume(self._manure_to_digest.total_volatile_solids)
         generated_methane_mass = generated_methane_volume * METHANE_DENSITY
         generated_carbon_dioxide_mass = (
             generated_methane_volume * CARBON_DIOXIDE_TO_METHANE_RATIO * CARBON_DIOXIDE_DENSITY
         )
         total_volatile_solids_destruction = generated_methane_mass + generated_carbon_dioxide_mass
-        self._received_manure = self._destroy_volatile_solids(total_volatile_solids_destruction)
+        self._manure_to_digest = self._destroy_volatile_solids(total_volatile_solids_destruction)
 
-        self._received_manure.volume -= total_volatile_solids_destruction / ManureConstants.SLURRY_MANURE_DENSITY
+        self._manure_to_digest.volume -= total_volatile_solids_destruction / ManureConstants.SLURRY_MANURE_DENSITY
 
         methane_leakage = self._calculate_methane_leakage(generated_methane_mass, self._methane_leakage_fraction)
         captured_methane_mass = generated_methane_mass - methane_leakage
         captured_methane_energy_content = self._calculate_methane_energy_content(captured_methane_mass)
 
-        evaporated_water = self._received_manure.volume * self._evaporation_fraction
+        evaporated_water = self._manure_to_digest.volume * self._evaporation_fraction
 
         self._report_anaerobic_digester_outputs(
             biogas=captured_methane_mass,
@@ -155,8 +155,8 @@ class AnaerobicDigester(Digester):
             time=time,
         )
 
-        manure_to_pass = replace(self._received_manure)
-        self._received_manure = ManureStream.make_empty_manure_stream()
+        manure_to_pass = replace(self._manure_to_digest)
+        self._manure_to_digest = ManureStream.make_empty_manure_stream()
         return {"manure": manure_to_pass}
 
     def _destroy_volatile_solids(self, total_volatile_solids_destruction: float, time: Time) -> ManureStream:
@@ -174,14 +174,14 @@ class AnaerobicDigester(Digester):
             Manure being processed by the anaerobic digester after volatile solids are removed.
 
         """
-        if self._received_manure.total_volatile_solids < total_volatile_solids_destruction:
+        if self._manure_to_digest.total_volatile_solids < total_volatile_solids_destruction:
             info_map = {
                 "class": self.__class__.__name__,
                 "function": self._destroy_volatile_solids.__name__,
                 "name": self.name,
                 "date": time.current_date.date(),
                 "simulation_day": time.simulation_day,
-                "total_volatile_solids": self._received_manure.total_volatile_solids,
+                "total_volatile_solids": self._manure_to_digest.total_volatile_solids,
                 "total_volatile_solids_destruction": total_volatile_solids_destruction,
             }
             err_name = f"Anerobic digester '{self.name}' attempted to destroy more volatile solids than available"
@@ -192,17 +192,17 @@ class AnaerobicDigester(Digester):
             non_degradable_volatile_solids = 0.0
         else:
             degradable_volatile_solids_frac = (
-                self._received_manure.degradable_volatile_solids / self._received_manure.total_volatile_solids
+                self._manure_to_digest.degradable_volatile_solids / self._manure_to_digest.total_volatile_solids
             )
 
-            degradable_volatile_solids = self._received_manure.degradable_volatile_solids - (
+            degradable_volatile_solids = self._manure_to_digest.degradable_volatile_solids - (
                 total_volatile_solids_destruction * degradable_volatile_solids_frac
             )
-            non_degradable_volatile_solids = self._received_manure.non_degradable_volatile_solids - (
+            non_degradable_volatile_solids = self._manure_to_digest.non_degradable_volatile_solids - (
                 total_volatile_solids_destruction * (1.0 - degradable_volatile_solids_frac)
             )
         return replace(
-            self._received_manure,
+            self._manure_to_digest,
             degradable_volatile_solids=degradable_volatile_solids,
             non_degradable_volatile_solids=non_degradable_volatile_solids,
         )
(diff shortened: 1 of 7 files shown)


RUFAS/biophysical/manure/digester/anaerobic_digester.py
```diff
@@ -0,0 +1,415 @@
...
+        self._evaporation_fraction: float = evaporation_fraction
+
+    def receive_manure(self, manure: ManureStream) -> None:
+        """Receives and stores manure to digested."""
+        is_received_manure_valid = self.check_manure_stream_compatibility(manure)
+        if is_received_manure_valid is False:
+            raise ValueError(f"Anaerobic digester {self.name} received an invalid manure stream.")
+        self._manure_to_digest += manure
+
+    def process_manure(self, conditions: CurrentDayConditions, time: Time) -> dict[str, ManureStream]:
```
Isgotkowitz, line 99: This method very closely follows `_calc_anaerobic_digestion_daily_output` in `anaerobic_digestion.py`.

RUFAS/biophysical/manure/digester/anaerobic_digester.py
```diff
@@ -0,0 +1,415 @@
...
+        total_volatile_solids_destruction = generated_methane_mass + generated_carbon_dioxide_mass
+        self._manure_to_digest = self._destroy_volatile_solids(total_volatile_solids_destruction)
+
+        self._manure_to_digest.volume -= total_volatile_solids_destruction / ManureConstants.SLURRY_MANURE_DENSITY
+
+        methane_leakage = self._calculate_methane_leakage(generated_methane_mass, self._methane_leakage_fraction)
+        captured_methane_mass = generated_methane_mass - methane_leakage
+        captured_methane_energy_content = self._calculate_methane_energy_content(captured_methane_mass)
+
+        evaporated_water = self._manure_to_digest.volume * self._evaporation_fraction
```
Isgotkowitz: Should this water be removed from `water` and `volume`?

RUFAS/biophysical/manure/digester/anaerobic_digester.py
```diff
@@ -0,0 +1,415 @@
...
+        effluent_temperature = cls._bind_influent_temperature(average_temp)
+        influent_heat_capacity = cls._calculate_manure_heat_capacity(average_temp, moisture_fraction)
+        anaerobic_digester_heat_capacity = cls._calculate_manure_heat_capacity(temperature_set_point, moisture_fraction)
+
+        average_heat_capacity = (influent_heat_capacity + anaerobic_digester_heat_capacity) / 2.0
+        heating_input_energy = average_heat_capacity * (temperature_set_point - effluent_temperature)
+        return heating_input_energy
+
+    @classmethod
+    def _bind_influent_temperature(cls, average_temp: float) -> float:
```
Isgotkowitz, line 324: Same as `bound_influent_temperature` in the pre-refresh `AnaerobicDigestion`.

input/data/manure/refreshed_manure_management.json
```diff
@@ -0,0 +1,15 @@
+{
+    "anaerobic_digesters":[
+        {
```
Isgotkowitz, line 3: This looks quite a bit different from the current anaerobic digestion configuration, and this has both pros and cons.

The big pro is that it strips the anaerobic digester inputs down to the absolute bare minimum number of inputs that it needs, making it easier to draw a straight line from the inputs to the code.

The con is that if this approach is followed for the rest of the manure module, every single final child class of `Processor` will require its own list in the manure management file, instead of having all storage, handler, separator, etc. configurations compiled into a single list.

Isgotkowitz: Personally, I feel the pro outweighs the con but I don't have to deal with writing many inputs.

RUFAS/biophysical/manure/digester/anaerobic_digester.py
elle-andreen: Theoretically yes, but it is not subtracted in the current module as you probably saw, so let's leave it be for now

RUFAS/biophysical/manure/digester/anaerobic_digester.py
```diff
@@ -0,0 +1,419 @@
...
+        heating_input_energy : float
+            Energy required to maintain the anaerobic digester's temperature (MJ).
+        methane_generation_volume : float
+            Volume of all methane generated during anaerobic digestion (m^3).
+        methane_leakage_mass : float
+            Mass of all methane generated during anaerobic digestion (m^3).
+        minimum_digester_volume : float
+            Minimum volume of manure allowed to be left in the anaerobic digester (m^3).
+        top_cover_volume : float
+            Volume of manure that is contained in the cover of the anaerobic digester (m^3).
```
elle-andreen:
```suggestion
            Headspace volume above manure inside digester where biogas collects (m^3).
```

RUFAS/biophysical/manure/digester/anaerobic_digester.py
```diff
@@ -0,0 +1,419 @@
...
+        amount of manure volatile solids loaded per day, multiplied by the generally-accepted methane potential value
+        for dairy manure (240 L CH4 per kg of manure volatile solids).
+
+        """
+        return ManureConstants.ACHIEVABLE_METHANE_EMISSION * total_volatile_solids
+
+    @classmethod
+    def _calculate_methane_leakage(cls, generated_methane_mass: float, leakage_fraction: float) -> float:
+        """
+        Calculates the mass of methane lost from an anaerobic digester.
```
elle-andreen:
```suggestion
        Calculates the mass of methane lost from an anaerobic digester as leakage.
```

RUFAS/biophysical/manure/digester/anaerobic_digester.py
```diff
@@ -0,0 +1,419 @@
...
+            methane_generation_volume=generated_methane_volume,
+            methane_leakage_mass=methane_leakage,
+            minimum_digester_volume=minimum_digester_volume,
+            top_cover_volume=top_cover_volume,
+            time=time,
+        )
+
+        digested_manure = replace(self._manure_to_digest)
+        self._manure_to_digest = ManureStream.make_empty_manure_stream()
+        return {"manure": digested_manure}
```
elle-andreen, line 160: Could you explain what is happening in lines 158-160?

RUFAS/biophysical/manure/digester/anaerobic_digester.py
```diff
@@ -0,0 +1,419 @@
...
+
+    def _destroy_volatile_solids(self, total_volatile_solids_destruction: float, time: Time) -> ManureStream:
```
elle-andreen, line 162: Is there a specific reason that this function is "up" in an earlier part of the file whereas the other main functions in the file (calc specific input energy, bind influent temp, calculate heat capacity, etc.) are after the _report_anaerobic_digester_outputs function?

RUFAS/biophysical/manure/manure_constants.py
```diff
@@ -3,6 +3,9 @@ class ManureConstants:
     A class to store constants for manure management.
     """
 
+    ACHIEVABLE_METHANE_EMISSION = 0.24
```
elle-andreen, line 6: Are we deprecating the separate gas emissions constants vs. manure constants? I am fine with this, just want to confirm!

input/data/manure/refreshed_manure_management.json
elle-andreen: Hmmmm this is a quandary best solved by group consensus. Is it possible to do something similar to the existing manure config inputs and have a single list of all applicable configs for each processor, but we just omit non-applicable configs from the default input file? To use refreshed handlers as an example:

The combined list would be:
```
      "name": 
      "manure_handler_type": 
      "cleaning_water_use_rate": 
      "minutes_per_cleaning": 
      "cleanings_per_day":
      "cleaning_water_recycle_fraction":
      "use_parlor_flush": "true
```
But the section/example for defining a parlor cleaning handler would be:
```
      "name": "main parlor wash",
      "manure_handler_type": "parlor cleaning",
      "cleaning_water_recycle_fraction": 0.80
      "use_parlor_flush": "true"
```
Whereas the example for an alley scraper would be:
```
      "name": "heifer scraper",
      "manure_handler_type": "alley scraper",
      "cleaning_water_use_rate": 5,
      "minutes_per_cleaning": 8,
      "cleanings_per_day": 2,
      "cleaning_water_recycle_fraction": 0.80
```

I will say, in the meantime, I think it is reasonable to place each processor's inputs here as we go, as you have done for AD, while we decide how to handle inputs as a whole.

RUFAS/biophysical/manure/digester/anaerobic_digester.py
```diff
@@ -0,0 +1,419 @@
...
+        Returns
+        -------
+        float
+            Mass of methane lost as leakage (kg).
+
+        """
+        return generated_methane_mass * leakage_fraction
+
+    @classmethod
+    def _calculate_methane_energy_content(cls, methane_mass: float) -> float:
```
allisterakun, line 408: Same as `GasEmissionsCalculator.calculate_methane_energy_content(cls, methane_mass: float) -> float` from routines

RUFAS/biophysical/manure/digester/anaerobic_digester.py
```diff
@@ -0,0 +1,419 @@
...
+        April Leytem (april.leytem@usda.gov). The equation is a simplification of the IPCC Tier II estimate of CH4
+        emissions from anaerobic digesters, where CH4 generated in the digester is assumed to be equivalent to the
+        amount of manure volatile solids loaded per day, multiplied by the generally-accepted methane potential value
+        for dairy manure (240 L CH4 per kg of manure volatile solids).
+
+        """
+        return ManureConstants.ACHIEVABLE_METHANE_EMISSION * total_volatile_solids
+
+    @classmethod
+    def _calculate_methane_leakage(cls, generated_methane_mass: float, leakage_fraction: float) -> float:
```
allisterakun, line 388: Same as `GasEmissionsCalculator.calculate_digester_methane_leakage(cls, generated_methane_mass: float, digester_methane_leakage_fraction: float) -> float:` from routines

RUFAS/biophysical/manure/digester/anaerobic_digester.py
```diff
@@ -0,0 +1,419 @@
...
+        Returns
+        -------
+        float
+            Heat capacity of manure in an anaerobic digester (kJ / kg / degrees C).
+
+        """
+        return 0.68298 + 0.025662 * temperature + 0.01306 * moisture_fraction * GeneralConstants.FRACTION_TO_PERCENTAGE
+
+    @classmethod
+    def _calculate_CSTR_methane_volume(cls, total_volatile_solids: float) -> float:
```
allisterakun, line 362: Same as `GasEmissionsCalculator.calculate_CSTR_methane_volume(cls, manure_total_volatile_solids: float) -> float:` from routines

RUFAS/biophysical/manure/digester/anaerobic_digester.py
```diff
@@ -0,0 +1,419 @@
...
+        moisture_fraction : float
+            Fraction of manure mass that is water (unitless).
+
+        Returns
+        -------
+        float
+            Heat capacity of manure in an anaerobic digester (kJ / kg / degrees C).
+
+        """
+        return 0.68298 + 0.025662 * temperature + 0.01306 * moisture_fraction * GeneralConstants.FRACTION_TO_PERCENTAGE
```
allisterakun, line 359: The old routines code for this function had a TODO that said to convert these numbers into constants. Are there any plans for that?

RUFAS/biophysical/manure/digester/anaerobic_digester.py
allisterakun: In the function above, we update the values of `self._manure_to_digest` for manure digestion.
At the end of it, we save the updated `self._manure_to_digest` values into a local variable called `digested_manure` (line 158).
Then we reset the `self._manure_to_digest` to an empty ManureStream (line 159).
Finally, we pack the saved result, `digested_manure`, into a dictionary and return it as the output (line 160).

RUFAS/biophysical/manure/digester/anaerobic_digester.py
allisterakun: The order of these functions does not matter as long as they are all in the same class.
If a class gets too large, we sort its functions alphabetically.

RUFAS/biophysical/manure/digester/anaerobic_digester.py
allisterakun: Confirmed that it follows almost exactly the same logic in `AnaerobicDigestion._calc_anaerobic_digestion_daily_output()`

RUFAS/biophysical/manure/digester/anaerobic_digester.py
```diff
@@ -0,0 +1,419 @@
...
+METHANE_DENSITY: float = 0.629
+
+"""Methane energy density (MJ / kg)."""
+METHANE_ENERGY_DENSITY: float = 55.0
+
+"""The lower temperature bound for manure flowing into an anaerobic digester (degrees C)."""
+MINIMUM_INFLUENT_TEMPERATURE = 4.0
+
+"""Factor by which total ammoniacal nitrogen content is increased by the anaerobic digestion process (unitless)."""
+TAN_INCREASE_FACTOR = 1.60
```
allisterakun, line 34: Question, instead of having these constants in a gas constants file, are we just putting them here?

allisterakun (approved):
LGTM, great job, Ed!

RUFAS/biophysical/manure/digester/anaerobic_digester.py
elle-andreen: So these lines are creating a new ManureStream that represents the manure leaving the digester, correct? I was a little confused since the lines below reporting to OM (under `_report_anaerobic_digester_outputs`) are all referencing manure_to_digest, which sort of made it seem like we were reporting the un-digested manure values (e.g. specs of manure coming into the digester, not manure leaving the digester). But I guess by that point _manure_to_digest is actually reflective of the manure leaving the digester/processor, right?

RUFAS/biophysical/manure/digester/anaerobic_digester.py
Isgotkowitz: Roger, roger

RUFAS/biophysical/manure/digester/anaerobic_digester.py
allisterakun: Yes, you are correct. At the end of this function, all "manure" represented in `self._manure_to_digest` is already digested and processed.

input/data/manure/refreshed_manure_management.json
Isgotkowitz: > Is it possible to do something similar to the existing manure config inputs and have a single list of all applicable configs for each processor, but we just omit non-applicable configs from the default input file?

Yes, if I follow you correctly this is definitely possible. However the big drawback to this approach would be that Input Manager would not be able to comprehensively validate each configuration. Meaning it would not be able to check whether an input in a configuration was unnecessary, or if a necessary input was missing from a configuration. Another drawback would be that the Input Manager would log a massive number of irrelevant warnings during the validation process, which would dilute the pool of warnings that would actually help the user figure out what is wrong with their input file.

RUFAS/biophysical/manure/digester/anaerobic_digester.py
Isgotkowitz: > So these lines are creating a new ManureStream that represents the manure leaving the digester, correct?

Yes, this is correct.

> I was a little confused since the lines below reporting to OM (under _report_anaerobic_digester_outputs) are all referencing manure_to_digest, which sort of made it seem like we were reporting the un-digested manure values (e.g. specs of manure coming into the digester, not manure leaving the digester).

Yeah the naming makes this confusing. At the start of the `process_manure` method, `_manure_to_digest` is an appropriate name because the manure has not been digested.

> But I guess by that point _manure_to_digest is actually reflective of the manure leaving the digester/processor, right?

Also correct, at that point the manure has been digested.

I will change the name to something that makes the code less confusing, like `_manure_in_digester`

RUFAS/biophysical/manure/digester/anaerobic_digester.py
Isgotkowitz: Everything that @allisterakun said is correct. There is a method to my madness though, but it is purely personal preference/style/whim. `__init__` always goes at the top, then the public instance methods*, then the private instance methods**, and then the class methods***.

\* a public instance method is a method whose name does _not_ start with `_`, and has `self` as the first argument.
** a private instance method is a method whose name _does_ start with `_` and has `self` as the first argument.
*** a class method is a method decorated with `@classmethod` and has a first argument of `cls`.

RUFAS/biophysical/manure/manure_constants.py
Isgotkowitz: I think we should. I am hoping that by moving all constants into the files where they are used or the closest shared file where they are used, we will get a really small manure constants file and there will be no need to separate the manure constants from the gas emissions constants.

RUFAS/biophysical/manure/digester/anaerobic_digester.py
Isgotkowitz: I inferred that the `moisture_fraction` is converted to a percentage, because it was multiplied by 100.

But I am reasonably confident that the other values in this method are straight out of scientific literature and were derived from some observed data, in which case they are only used for this equation and it would be better to keep them here instead of trying to come up with constants for them.

input/data/manure/refreshed_manure_management.json
elle-andreen: Agreed, I know this is already happening in the current module. We current have 4 processor categories, and 11 discrete processor types within/across those categories (although handlers and separators could possibly be condensed into 1 because the types within class are so similar, which would bring us to 7). I suppose one argument in favor of structuring inputs in this way is that it helps the user think of the manure chains in a more realistic way - if you have two separators and three storages, you need to define those and only those processors, instead of one for each pen. Handlers would violate this to some degree since we need a separate handler for every applicable pen, which is not always the reality, but we should definitely consider the pros and cons of both options.

RUFAS/biophysical/manure/digester/anaerobic_digester.py
Isgotkowitz: I have been. I think this makes the code cleaner because we don't get a bunch of different modules importing a big file full of constants when only a handful are needed and the rest are explicitly unrelated.

RUFAS/biophysical/manure/digester/anaerobic_digester.py
elle-andreen: Thanks both of you for the explanations. If it's not much of a hassle, I think a slightly different name would help SMEs out in understanding the code!

RUFAS/biophysical/manure/manure_constants.py
elle-andreen: Works for me! There's sort of a blurry line between what is gas emissions and what is just manure anyway.

RUFAS/biophysical/manure/digester/anaerobic_digester.py
Isgotkowitz: Not a hassle at all! Name has been changed to `_manure_in_digester`.

Isgotkowitz:
@elle-andreen and @allisterakun thank you for your reviews. I have changed `_manure_to_digest` to `_manure_in_digester`, which should make the `process_manure` method easier to follow.

elle-andreen (approved):
Looks great to me! Thanks for your careful work and also for @allisterakun's detailed comparison with previous methods.

PR #2280, opened 2025-03-13
[Manure] Adds separator configs to refresh manure mgmt json

### LINKED ISSUE #2109: [Manure] Re-implement `Separator` logic
Re-implement the `Separator` logic. Use the "Developer Guidelines" in #2092.

ew3361zh:
Builds off existing `refreshed_manure_management.json` starter input file and incorporates elements from discussion at 3/12/25 manure module meeting.

### Context

Issue(s) closed by this pull request: closes #2109

### What

Creates a `separator_configs` section of the `refreshed_manure_management.json` input file.
To match the new naming structure:
- Adds `digester_type` for `anaerobic_digester_configs` entry. 
- Adds `separator_type` for `separator_configs` entries.
- We could alternatively just have this field be `type`. Thoughts?

To help facilitate initialization of the proper processor:
- Adds `processor_type` for both to help with initialization.

To match the separator input json attributes with the refreshed separator attributes:
- Adds `ash_removal_efficiency` for both separator types.
- Adjusts efficiency names in json file to match refreshed efficiency names in Separator class for clarity.

**Moves `_log_manure_stream()` method into Processor base class as it will be implemented by multiple child and grandchild classes.**

### Why

Setting up initialization of the newly refreshed Separator class.

### How

Copy-pasted the separator configs from the current `manure_management.json` file and implemented the above changes.

### Test plan

No real test plan at this point as it's just adding configs that are not yet used.

### Notes
- At some point we will need to adjust the init methods to handle `processor_type` and each child class's `type`.

### DIFF
--- a/RUFAS/biophysical/manure/processor.py
+++ b/RUFAS/biophysical/manure/processor.py
@@ -125,7 +125,8 @@ class Processor(ABC):
         if manure_stream_dict.keys() != ManureStream.MANURE_STREAM_UNITS.keys():
             self._om.add_error(
                 "Manure Stream Keys Error",
-                f"Expected keys: {set(ManureStream.MANURE_STREAM_UNITS.keys())}, received: {set(manure_stream_dict.keys())}.",
+                f"Expected keys: {set(ManureStream.MANURE_STREAM_UNITS.keys())}, "
+                f"received: {set(manure_stream_dict.keys())}.",
                 info_map,
             )
             raise ValueError("Manure Stream must contain the same keys as manure_stream_units to properly log it.")
--- a/RUFAS/biophysical/manure/processor.py
+++ b/RUFAS/biophysical/manure/processor.py
@@ -103,7 +103,8 @@ class Processor(ABC):
         """
         pass
 
-    def _log_manure_stream(self, manure_stream: ManureStream | dict[str, float], stream_name: str, time: Time) -> None:
+    def _log_manure_stream(self, manure_stream: ManureStream | dict[str, float | None], stream_name: str, time: Time
+                           ) -> None:
         """
         Logs the manure stream data to Output Manager.
 
--- a/RUFAS/biophysical/manure/processor.py
+++ b/RUFAS/biophysical/manure/processor.py
@@ -6,7 +6,6 @@ from RUFAS.data_structures.animal_to_manure_connection import ManureStream
 from RUFAS.general_constants import GeneralConstants
 from RUFAS.time import Time
 from RUFAS.output_manager import OutputManager
-from RUFAS.units import MeasurementUnits
 from RUFAS.util import Utility
 
 
@@ -40,22 +39,6 @@ class Processor(ABC):
         Handles the daily operations for the processor.
 
     """
-
-    MANURE_STREAM_UNITS = {
-        "water": MeasurementUnits.KILOGRAMS,
-        "ammoniacal_nitrogen": MeasurementUnits.KILOGRAMS,
-        "nitrogen": MeasurementUnits.KILOGRAMS,
-        "phosphorus": MeasurementUnits.KILOGRAMS,
-        "potassium": MeasurementUnits.KILOGRAMS,
-        "ash": MeasurementUnits.KILOGRAMS,
-        "non_degradable_volatile_solids": MeasurementUnits.KILOGRAMS,
-        "degradable_volatile_solids": MeasurementUnits.KILOGRAMS,
-        "total_solids": MeasurementUnits.KILOGRAMS,
-        "volume": MeasurementUnits.CUBIC_METERS,
-        "mass": MeasurementUnits.KILOGRAMS,
-        "total_volatile_solids": MeasurementUnits.KILOGRAMS,
-        "pen_manure_data": None,
-    }
 
     def __init__(self, name: str, is_housing_emissions_calculator: bool) -> None:
         """Initializes a new Processor."""
@@ -139,10 +122,10 @@ class Processor(ABC):
             )
             raise ValueError("Manure stream must be a dictionary or a ManureStream instance to properly log it.")
 
-        if manure_stream_dict.keys() != self.MANURE_STREAM_UNITS.keys():
+        if manure_stream_dict.keys() != ManureStream.MANURE_STREAM_UNITS.keys():
             self._om.add_error(
                 "Manure Stream Keys Error",
-                f"Expected keys: {set(self.MANURE_STREAM_UNITS.keys())}, received: {set(manure_stream_dict.keys())}.",
+                f"Expected keys: {set(ManureStream.MANURE_STREAM_UNITS.keys())}, received: {set(manure_stream_dict.keys())}.",
                 info_map,
             )
             raise ValueError("Manure Stream must contain the same keys as manure_stream_units to properly log it.")
@@ -150,7 +133,7 @@ class Processor(ABC):
         for key, value in manure_stream_dict.items():
             if key != "pen_manure_data":
                 self._om.add_variable(
-                    f"{stream_name}.manure_{key}", value, {**info_map, "units": self.MANURE_STREAM_UNITS[key]}
+                    f"{stream_name}.manure_{key}", value, {**info_map, "units": ManureStream.MANURE_STREAM_UNITS[key]}
                 )
 
     def check_manure_stream_compatibility(self, manure_stream: ManureStream) -> bool:
--- a/RUFAS/biophysical/manure/processor.py
+++ b/RUFAS/biophysical/manure/processor.py
@@ -87,24 +87,24 @@ class Processor(ABC):
         """
         pass
 
-    def _log_manure_stream(
+    def _report_manure_stream(
         self, manure_stream: ManureStream | dict[str, float | None], stream_name: str, time: Time
     ) -> None:
         """
-        Logs the manure stream data to Output Manager.
+        Reports the manure stream data to Output Manager.
 
         Parameters
         ----------
         manure_stream : ManureStream | dict[str, float]
-            The manure stream to log. If a `ManureStream` instance is passed, it will be converted to a dictionary.
+            The manure stream to report. If a `ManureStream` instance is passed, it will be converted to a dictionary.
         stream_name : str
-            The name of the manure stream being logged.
+            The name of the manure stream being reported.
         time : Time
             The simulation time object.
         """
         info_map = {
             "class": self.__class__.__name__,
-            "function": self._log_manure_stream.__name__,
+            "function": self._report_manure_stream.__name__,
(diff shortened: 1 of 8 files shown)


input/data/manure/refreshed_manure_management.json
```diff
@@ -11,5 +13,33 @@
                 "storage": 1.0
             }
         }
+    ],
+    "separator_configs":[
+        {
+            "name": "rotary_screen",
+            "processor_type": "separator",
+            "separator_type": "rotary screen",
+            "percent_dry_solids": 0.20,
```
elle-andreen:
```suggestion
            "separated_solids_dry_matter": 0.20,
```
Just a change to improve the clarity of what this variable represents

input/data/manure/refreshed_manure_management.json
```diff
@@ -11,5 +13,33 @@
...
+            "ammoniacal_nitrogen_efficiency": 0.15,
+            "phosphorus_efficiency": 0.40,
+            "potassium_efficiency": 0.15,
+            "ash_efficiency": 0.10
+          },
+          {
+            "name": "screw_press",
+            "processor_type": "separator",
+            "separator_type": "screw press",
+            "percent_dry_solids": 0.35,
```
elle-andreen:
```suggestion
            "separated_solids_dry_matter": 0.35,
```

input/data/manure/refreshed_manure_management.json
```diff
@@ -11,5 +13,33 @@
                 "storage": 1.0
```
elle-andreen: As we talked about in the 3/12 refresh meeting, the inputs to configure the manure processor chain will mostly likely be in a separate file from the processor configuration inputs. I think we may as well remove this input from AD for now so that there's no confusion down the line when more inputs are brought into this file

input/data/manure/refreshed_manure_management.json
```diff
@@ -1,7 +1,9 @@
 {
-    "anaerobic_digesters":[
+    "anaerobic_digester_configs":[
         {
             "name": "anaerobic_digester",
+            "processor_type": "anaerobic digester",
```
elle-andreen: It's completely fine if it's too complicated to do so, but is it possible that the processor type could be inferred from the 'section' in which the specific config instance is provided? I.e., by providing this input under the "anaerobic digesters": array, is in inferable elsewhere in the code that this is a digester processor? Same applies to the other 3 processor classes. This is not problematic, just a little redundant, but I know we need the processor type for OutputManager purposes

elle-andreen (approved):
Pretty straightforward but important for getting the inputs set up more clearly! I made a few suggestions but in general this looks good to me and is in line with what we discussed as a group. Thanks Niko 😄

input/data/manure/refreshed_manure_management.json
```diff
@@ -1,7 +1,9 @@
...
+            "digester_type": "anaerobic digester",
```
elle-andreen:
```suggestion
            "digester_type": "CSTR or continuous mix",
```
Just making this more specific. We only have one type of digester at the moment but this is the specific type of digester that the processor represents.

matthew7838 (approved):
LGTM and I think it makes sense. I'll create a separate PR for handlers once this is merged.

input/data/manure/refreshed_manure_management.json
ew3361zh: That's fine by me. What do you think about updating this name in the Separator attributes to match this (right now it's `percent_dry_solids`? I think it would be helpful if they matched since the rest do.

input/data/manure/refreshed_manure_management.json
ew3361zh: I think that could work. My preference would be to make the processor category name match our expected enum name as closely as possible (ideally an exact match so e.g. instead of `"separator_configs"` it would be `"separator"` - I think we could infer from the file that these are all configs for the manure processors so adding `configs` is a little redundant anyway). But since we'll be pulling these configs from the input file, getting the next higher up level input shouldn't be difficult.

input/data/manure/refreshed_manure_management.json
ew3361zh: Would one or the other be better or should the full type really be `"CTSR or continuous mix"`?

input/data/manure/refreshed_manure_management.json
```diff
@@ -11,5 +13,33 @@
...
+            "processor_type": "separator",
+            "separator_type": "rotary screen",
+            "percent_dry_solids": 0.20,
+            "total_solids_efficiency": 0.35,
+            "volatile_solids_efficiency": 0.40,
+            "nitrogen_efficiency": 0.30,
+            "ammoniacal_nitrogen_efficiency": 0.15,
+            "phosphorus_efficiency": 0.40,
+            "potassium_efficiency": 0.15,
+            "ash_efficiency": 0.10
```
ew3361zh: @elle-andreen I just made up the number for `ash_efficiency` here since it didn't exist previously. Can you confirm for both separator types?

input/data/manure/refreshed_manure_management.json
elle-andreen: Very good point! We should definitely do that. I'm sorry to make this change at the inputs stage instead of in the actual separator relocation PR, I will try to note this sort of thing earlier going forward to we don't have to go back to already-merged files to change names

input/data/manure/refreshed_manure_management.json
elle-andreen: I originally included them both because they're used sort of interchangeably. But I think just "continuous mix" would be sufficient so the type option is not confusing!

input/data/manure/refreshed_manure_management.json
```diff
@@ -1,15 +1,39 @@
...
+            "name": "separator1",
+            "separator_type": "rotary screen",
+            "separated_solids_dry_matter": 0.20,
+            "total_solids_efficiency": 0.35,
+            "volatile_solids_efficiency": 0.40,
+            "nitrogen_efficiency": 0.30,
+            "ammoniacal_nitrogen_efficiency": 0.15,
+            "phosphorus_efficiency": 0.40,
+            "potassium_efficiency": 0.15,
+            "ash_efficiency": 0.10
```
elle-andreen:
```suggestion
            "ash_efficiency": 0.20
```
Thank you for confirming this! I should have provided this more explicitly. The default value (0.20) is the same for both types of separators.

allisterakun (approved):
It makes sense from my perspective, great job, Niko!

Just curious:
I know that we sort of agreed on making the metadata at the very end of manure refresh. But would this be a good opportunity to break that down into small chunks? Like we don't necessarily need to directly modify the actual metadata properties, but maybe just start a new json at some temporary location and building that chunk by chunk. Then at the end, we could just modify that a little bit and add/replace the manure section in the actual metadata properties.

input/data/manure/refreshed_manure_management.json
```diff
@@ -1,15 +1,39 @@
...
+            "name": "separator2",
+            "separator_type": "screw press",
+            "separated_solids_dry_matter": 0.35,
+            "total_solids_efficiency": 0.25,
+            "volatile_solids_efficiency": 0.30,
+            "nitrogen_efficiency": 0.30,
+            "ammoniacal_nitrogen_efficiency": 0.10,
+            "phosphorus_efficiency": 0.20,
+            "potassium_efficiency": 0.23,
+            "ash_efficiency": 0.10
```
elle-andreen:
```suggestion
            "ash_efficiency": 0.20
```

ew3361zh:
Summary of changes:
- Moves `_log_manure_stream()` method from Separator child class to Processor base class.
- Adds logging of `total_volatile_solids` and `mass` ManureStream properties to OM in this method.
- Extracts MANURE_STREAM_UNITS into ManureStream class as a class attribute.
- Updates `percent_dry_solids` Separator attribute to preferred `separated_solids_dry_matter` which also matches the input name convention.
- Updates `ash_efficiency` value input JSONs to accurate value.
- Updates unit testing.

ew3361zh:
> It makes sense from my perspective, great job, Niko!
> 
> Just curious: I know that we sort of agreed on making the metadata at the very end of manure refresh. But would this be a good opportunity to break that down into small chunks? Like we don't necessarily need to directly modify the actual metadata properties, but maybe just start a new json at some temporary location and building that chunk by chunk. Then at the end, we could just modify that a little bit and add/replace the manure section in the actual metadata properties.

I think this is a good idea but there got to be some extra stuff I needed to do in this PR which suddenly made it change a total of 350 lines so I think to keep it simple and not too expansive I may push this down the road to where we originally wanted to do it.

RUFAS/biophysical/manure/processor.py
```diff
@@ -86,6 +87,55 @@ def process_manure(self, conditions: CurrentDayConditions, time: Time) -> dict[s
...
+            manure_stream_dict["mass"] = ManureStream.mass
+        elif isinstance(manure_stream, dict):
+            manure_stream_dict = manure_stream.copy()
+        else:
+            self._om.add_error(
+                "Manure Stream Type Error",
+                "This function requires either a ManureStream instance or a dictionary.",
+                info_map,
+            )
+            raise ValueError("Manure stream must be a dictionary or a ManureStream instance to properly log it.")
```
elle-andreen: Just want to confirm since I am not able to 100% perfectly comprehend the code here, will this check cause any issues if a Processor does not receive/generate a ManureStream instance each day? Some processors may not receive ManureStream(s) every single day and so would have nothing to pass on either

ew3361zh: I'd be happy to go through the code with you if that would be helpful. But this would only get triggered if the function were called and the manure stream value sent to the function were not either a ManureStream object or a dictionary.

So this function doesn't have any knowledge of or responsibility for when it is being called. So far it's only implemented in `Separator` at a point in its `process_manure()` function when there is actual manure to be separated. We would definitely want to make sure wherever else it gets implemented that we're paying attention to what's being sent.

RUFAS/biophysical/manure/processor.py
```diff
@@ -86,6 +87,56 @@ def process_manure(self, conditions: CurrentDayConditions, time: Time) -> dict[s
         """
         pass
 
+    def _log_manure_stream(
```
allisterakun: One minor suggestion is to rename this function to report output manure stream (or something like that), because it's not logging the manure stream (either to the console or calling om.add_log()) but rather reporting some output.

allisterakun (approved):
LGTM! Good work, Niko!

matthew7838 (approved):
LGTM, one thing about the naming of input, I feel like using the name `type` is more generalized and might make things easier down the road, also it's already under `separator ` or `digester`, which I think the current naming is a bit redundant.

PR #2285, opened 2025-03-19
Re-implement SlurryStorageOutdoor and SlurryStorageUnderfloor

### LINKED ISSUE #2107: [Manure] Re-implement `SlurryStorageOutdoor` logic
Re-implement the `SlurryStorageOutdoor` logic. Use the "Developer Guidelines" in #2092.

### LINKED ISSUE #2108: [Manure] Re-implement `SlurryStorageUnderfloor` logic
Re-implement the `SlurryStorageUnderfloor` logic. Use the "Developer Guidelines" in #2092.

### LINKED ISSUE #2207: [Manure] Rename nitrous oxide and ammonia outputs and add new outputs representative of total mass
In the manure module, RuFaS reports kg of N2O nitrogen (storage_nitrous_oxide) and NH3 nitrogen (storage_ammonia) produced by manure storages or other components. These variable names are misleading as they represent the mass of _nitrogen_ in N2O or NH3, not the _total_ mass of N2O or NH3. We calculate and utilize mass of nitrogen lost to either of these gasses in order to track nitrogen through the module, but the total mass is what is relevant for greenhouse gas footprints, comparison to other models, etc. We should provide this value directly instead of relying on the user to be aware of and remember that they need to convert from a nitrogen to a total mass basis.

Steps to address this issue:

NITROUS OXIDE:
1. Create a new constant: GasEmissionConstants.N2ON_to_N2O_CONVERSION = 1.5711 
Description: Conversion factor for kg of total N2O emitted per kg of N2O-N emitted. Determined by taking the inverse of the proportion of N mass in N2O (0.6365 / 1 = 1.5711).

2. Rename existing instances of storage_nitrous_oxide to storage_nitrous_oxide_**N** (units = kg)

~~3. Add a _new_ output wherever storage_nitrous_oxide_N is included in daily output (to OutputManager) in the Manure module:
storage_nitrous_oxide (kg) = storage_nitrous_oxide_**N** * GasEmissionConstants.N2ON_to_N2O_CONVERSION~~ 

~~4. Update appropriate tests.~~

AMMONIA:
1. Create a new constant: GasEmissionConstants.NH3N_to_NH3_CONVERSION = 1.2159
Description: Conversion factor for kg of total NH3 emitted per kg of NH3-N emitted. Determined by taking the inverse of the proportion of N mass in N2O (0.8224 / 1 = 1.2159).

2. Rename existing instances of storage_ammonia to storage_ammonia_**N** (units = kg)

~~3. Add a _new_ output wherever storage_ammonia_N is included in daily output (to OutputManager) in the Manure module:
storage_ammonia (kg) = storage_ammonia_**N** * GasEmissionConstants.NH3N_to_NH3_CONVERSION~~

~~4. Update appropriate tests.~~

3/26/25 updates: After a conversation with @KFosterReed, we elected to just rename output names, rather than addition additional output variables to represent total mass of N2O or NH3. These values can easily be calculated by the user using output filters or other manual calculation.

allisterakun:
Re-implement SlurryStorageOutdoor and SlurryStorageUnderfloor
## Context

Issue(s) closed by this pull request: closes #2107, #2108.

### Re-implementation of SlurryStorageOutdoor and SlurryStorageUnderfloor
1. Adds two new classes, SlurryStorageOutdoor and SlurryStorageUnderfloor, each extending Storage logic for specific manure-processing scenarios.
    1. Subclasses Storage, overriding process_manure, calculating methane/ammonia/nitrous oxide, then reporting them alongside updated ManureStream totals or empties.
    2. Specifically handle precipitation (outdoor) vs. no precipitation (underfloor) storage.
2. New test files for SlurryStorageOutdoor and SlurryStorageUnderfloor.

### Adjustments to Reporting Functions
1. Changes a key in manure logging within processor.py—replacing "{stream_name}.manure_{key}" with "{stream_name}_manure_{key}".
2. Adjusts the `Storage.process_manure()` method to output only total GHG/ammonia/nitrous oxide fields (e.g., "storage_methane", "storage_ammonia", "storage_nitrous_oxide") instead of the entire manure breakdown. The full breakdown is now via `Processor._report_manure_stream()`.
3. Correspondingly revises the related tests under tests/test_biophysical/test_manure/.

## Test plan

1. Fixed the broken unit tests.
2. Wrote new unit tests for SlurryStorageOutdoor and SlurryStorageUnderfloor.


RUFAS/biophysical/manure/processor.py
```diff
@@ -105,7 +105,7 @@ def _report_manure_stream(
         info_map = {
             "class": self.__class__.__name__,
             "function": self._report_manure_stream.__name__,
-            "prefix": self._prefix,
+            "prefix": prefix,
```
ew3361zh: I don't think the changes you made here are necessary. Here's my breakdown (tagging @matthew7838 and @elle-andreen as well for a full discussion).

self._prefix is initialized in the Processor base class as:
`f"{self.__class__.__name__}.{self.name}"`

This would result in something like:
`Separator.niko_seperator`

When `_report_manure_stream()` adds a stream to the variable pool it has this format:
`f"{stream_name}.manure_{key}"`

Because you're also sending a prefix in the info map (currently in dev as `self._prefix`), the end result of the variable added to OM will be something like:
`Separator.niko_seperator.stream_name.manure_nitrogen`

The current notes from the weekly manure meeting have the desired format to be:
`Manure.ProcessorCategory.ProcessorType.ProcessorName.Variable`

So my above example SHOULD be (I believe):
`Manure.Separator.rotary_screen.niko_separator.stream_name.manure_nitrogen`

The changes we need to make to get there:
- Rework `self._prefix` in `Processor` to be `f"Manure.{self.__class__.__name__}.{self.type}.{self.name}"

And I think we then keep the `stream_name` variable as is in the `_report_manure_stream()` method and just send it the stream name of either `accumulated` or `received`.

Because in that case, if we fixed `self._prefix` as outlined above, `_report_manure_stream()` and sent `accumulated` or `received` as the `stream_name`, it would then add a variable like this:
`f"Manure.{self.__class__.__name__}.{self.type}.{self.name}.stream_name.manure_{key}"`

Which would get us both:
`Manure.Separator.rotary_screen.niko_separator.acculuated.manure_nitrogen`
and
`Manure.Separator.rotary_screen.niko_separator.received.manure_nitrogen`

Does that makes sense?

matthew7838: I agree with this and have similar changes to the prefix setup in the handler. However, in the notes, I think we want `Manure.Separator.rotary_screen.niko_separator.accumulated_manure_nitrogen` instead of `Manure.Separator.rotary_screen.niko_separator.accumulated.manure_nitrogen`. I think it's cleaner to combine the stream name (received/accumulated) into the final variable name.

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
```diff
@@ -0,0 +1,147 @@
...
+            ammonia_resistance=STORAGE_HSC,
+            surface_area=self._surface_area,
+            pH=DEFAULT_PH_FOR_AMMONIA,
+        )
+        stored_manure.ammoniacal_nitrogen = max(0.0, stored_manure.ammoniacal_nitrogen - storage_ammonia)
+        stored_manure.nitrogen = max(0.0, stored_manure.nitrogen - storage_ammonia)
+
+        nitrous_oxide_emissions = self._calculate_nitrous_oxide_emissions(
+            nitrous_oxide_emissions_factor=STORAGE_COVER_NITROUS_OXIDE_EMISSIONS_FACTOR_MAPPING[self._cover.value],
+            nitrogen_added=self._received_manure.nitrogen,
```
matthew7838: Wouldn't the `self._received_manure` be empty here since the super process method was called in the beginning?

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
```diff
@@ -0,0 +1,144 @@
...
+        is_housing_emissions_calculator: bool,
+        cover: StorageCover,
+        storage_time_period: int | None,
+        surface_area: float,
+        nitrous_oxide_emissions_factor: float,
+        capacity: float = inf,
+    ):
+        super().__init__(
+            name=name,
+            is_housing_emissions_calculator=is_housing_emissions_calculator,
```
ew3361zh:
```suggestion
            is_housing_emissions_calculator=False
```
I think this should be false for everything except Handlers and I'm not sure if we need to have it be an init param - I think we could just pass in False and not worry about it.

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
```diff
@@ -0,0 +1,144 @@
...
+The mass conversion factor from methane to methane and carbon dioxide emitted from stored manure, based on a molar
+ratio of 1:3 (methane : carbon dioxide).
+"""
+
+
+class SlurryStorageOutdoor(Storage):
+    def __init__(
+        self,
+        name: str,
+        is_housing_emissions_calculator: bool,
```
ew3361zh:
```suggestion
```

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def process_manure(self, current_day_conditions: CurrentDayConditions, time: Time) -> dict[str, ManureStream]:
+        if self._cover in [StorageCover.NO_COVER, StorageCover.CRUST]:
+            self._received_manure.volume += (
+                current_day_conditions.precipitation * GeneralConstants.MM_TO_M * self._surface_area
+            )
+            self._received_manure.water += (
+                current_day_conditions.precipitation
+                * GeneralConstants.MM_TO_M
+                * self._surface_area
```
ew3361zh: Small suggestion to not do the same calculation twice but it's pretty negligible:

precipitation_volume = current_day_conditions.precipitation * GeneralConstants.MM_TO_M * self._surface_area
self._received_manure.volume += precipitation_volume
self._received_manure.water += precipitation_volume * GeneralConstants.WATER_DENSITY_KG_PER_M3

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
```diff
@@ -0,0 +1,144 @@
...
+                * GeneralConstants.WATER_DENSITY_KG_PER_M3
+            )
+        received_manure = self._received_manure
```
ew3361zh: I think you need to do `copy(self._received_manure)` since the Storage.process_manure() will empty the `received_manure` after adding it to stored_manure and the non-empty version is needed for calculations and reporting below.

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
```diff
@@ -0,0 +1,144 @@
...
+        manure_to_return = super().process_manure(current_day_conditions, time)
+        stored_manure = manure_to_return["manure"] if manure_to_return else copy(self._stored_manure)
```
ew3361zh: I asked @elle-andreen on slack this morning about what we're doing on emptying day and whether we're calculating emissions even if the storage is emptied (which is what your current code does) or if we don't calculate emissions on emptying day because the storage has no manure in it. I am guessing your version is correct but just wanted to confirm.

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -160,59 +160,39 @@ def process_manure(self, _: CurrentDayConditions, time: Time) -> dict[str, Manur
         if self.is_overflowing is True:
             self.handle_overflowing_manure(time)
 
-        self._report_storage_outputs(self._accumulated_output_prefix, "accumulated", self._stored_manure, time)
+        self._report_manure_stream(self._stored_manure, "accumulated", time)
```
ew3361zh: Do we need to report the manure stream here? I think this would report the accumulated twice, once pre-emissions losses here and once post-emissions losses in the child `process_manure()`

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
```diff
@@ -0,0 +1,144 @@
...
+        if not manure_to_return:
+            self._stored_manure = copy(stored_manure)
+
+        self._report_manure_stream(stored_manure, "accumulated", time)
+        self._report_manure_stream(received_manure, "received", time)
+
+        self._report_storage_outputs(total_storage_methane, storage_ammonia, nitrous_oxide_emissions, time)
+        self._report_slurry_storage_outputs(storage_methane_burned, time)
+
+        return manure_to_return
```
ew3361zh: Just to confirm with @elle-andreen, what we're returning here is what would be passed on to the next processor (if there is one)? So a storage will only return something here if it is emptying day, is that right?

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -160,59 +160,39 @@ def process_manure(self, _: CurrentDayConditions, time: Time) -> dict[str, Manur
...
 
         return manure_to_be_returned
 
     def _report_storage_outputs(
-        self, info_map_prefix: str, variable_name_prefix: str, manure: ManureStream, time: Time
+        self, storage_methane: float, storage_ammonia: float, nitrous_oxide_emissions: float, time: Time
     ) -> None:
         """
         Reports attributes of a manure stream instance as directed from a manure storage.
```
ew3361zh: Maybe update to say it reports storage manure emissions or something like that.

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -160,59 +160,39 @@ def process_manure(self, _: CurrentDayConditions, time: Time) -> dict[str, Manur
...
-        )
-        self._om.add_variable(
-            f"{variable_name_prefix}_manure_total_volatile_solids", manure.total_volatile_solids, info_map_kg
-        )
-        self._om.add_variable(f"{variable_name_prefix}_manure_total_solids", manure.total_solids, info_map_kg)
-        self._om.add_variable(f"{variable_name_prefix}_manure_mass", manure.mass, info_map_kg)
-        self._om.add_variable(f"{variable_name_prefix}_manure_volume", manure.volume, info_map_m3)
+        self._om.add_variable("storage_methane", storage_methane, info_map)
+        self._om.add_variable("storage_ammonia", storage_ammonia, info_map)
+        self._om.add_variable("storage_nitrous_oxide", nitrous_oxide_emissions, info_map)
```
ew3361zh: Curious if you heard back from Elle on whether these should be formatted similar to the outputs for the manure stream?

i.e. should it be:
f"{stream_name}_manure_storage_methane", value, info_map}"

ew3361zh (commented):
Really great work, Allister. Left a few questions/comments but feels close to being ready!

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
allisterakun: @elle-andreen please confirm this.

RUFAS/biophysical/manure/storage/storage.py
allisterakun: @elle-andreen please confirm the naming conventions when reporting the emission variables

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
ew3361zh: From the design doc:

Screenshot 2025-03-21 at 11 05 44 AM

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
```diff
@@ -0,0 +1,139 @@
...
+    STORAGE_COVER_NITROUS_OXIDE_EMISSIONS_FACTOR_MAPPING,
+)
+from RUFAS.biophysical.manure.storage.storage_cover import StorageCover
+from RUFAS.current_day_conditions import CurrentDayConditions
+from RUFAS.data_structures.animal_to_manure_connection import ManureStream
+from RUFAS.general_constants import GeneralConstants
+from RUFAS.time import Time
+from RUFAS.units import MeasurementUnits
+
+SLURRY_MANURE_DENSITY = 990
```
elle-andreen, line 17: All 3 of these constants will also be used by slurry storage underfloor and anaerobic lagoon, not sure if that makes them good candidates to go into the base Storage file?

allisterakun: We did notice the issues regarding the manure constants, which were discussed in one of the dev team meetings. We decided to keep them in separate processor files until we finish all the processor implementations (basically finish phase 2). Then, one of us will pick up the task to organize them.

We came to this decision because as we implement one processor, we do not know if a constant would be used in some other processors we have yet to implement. Instead of putting a constant in processor A and moving it to some base processor file later when we implement processor B as we find it also uses the same constant, we would just do a cleanup at the very end after we're done with all the processor implementations.

elle-andreen: @allisterakun That sounds like a great plan to me!

RUFAS/biophysical/manure/storage/slurry_storage_underfloor.py
```diff
@@ -0,0 +1,117 @@
...
+            nitrous_oxide_emissions_factor=nitrous_oxide_emissions_factor,
+            capacity=capacity,
+        )
+
+    def process_manure(self, current_day_conditions: CurrentDayConditions, time: Time) -> dict[str, ManureStream]:
+        received_manure = copy(self._received_manure)
+        manure_to_return = super().process_manure(current_day_conditions, time)
+        stored_manure = manure_to_return["manure"] if manure_to_return else copy(self._stored_manure)
+
+        manure_temperature = self._determine_outdoor_storage_temperature(
```
elle-andreen: This is a little counter-intuitive but the temperature determination here uses the `determine_barn_temperature` function in Matthew's handler PR (because a slurry underfloor storage is by nature indoors). However you want to handle this is up to you, whether that is move the function to some kind of base class, or modify the `_determine_outdoor_storage_temperature` to include a condition for slurry underfloor, or something else!

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
```diff
@@ -0,0 +1,139 @@
...
+"""The density of slurry manure (kg/:math:`m^3`)."""
+
+STORAGE_HSC = 4.1
```
elle-andreen, line 20: Could we please rename this constant to "STORAGE_RESISTANCE" & update all its uses across all files? The name was incorrect in the existing manure module file

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -160,59 +171,41 @@ def process_manure(self, _: CurrentDayConditions, time: Time) -> dict[str, Manur
...
         Parameters
         ----------
-        info_map_prefix : str
-            Prefix for the info map.
-        variable_name_prefix : str
-            Prefix for the variable name.
-        manure : ManureStream
-            The manure stream instance to report.
+        storage_methane : float
+            The storage methane emission of the current day, (kg).
```
elle-andreen:
```suggestion
            The methane emitted from manure storage on the current day, (kg).
```

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -160,59 +171,41 @@ def process_manure(self, _: CurrentDayConditions, time: Time) -> dict[str, Manur
...
+        storage_ammonia_nitrogen : float
+            The storage ammonia emission of the current day, (kg).
```
elle-andreen:
```suggestion
            The ammonia emitted from manure storage on the current day, (kg).
```

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -160,59 +171,41 @@ def process_manure(self, _: CurrentDayConditions, time: Time) -> dict[str, Manur
...
+        storage_nitrous_oxide_nitrogen : float
+            The nitrous oxide emission of the current day, (kg).
```
elle-andreen:
```suggestion
            The nitrous oxide emitted from manure storage on the current day (kg).
```

elle-andreen (approved):
This looks good to me so far! I left a few minor comments and will review the dev reviewer feedback when Niko and/or Matthew are able to take a look. I know you know, but just a reminder too that we can remove the N2O and NH3 conversion factor constants, per the discussion today. Thank you Allister!!

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
```diff
@@ -0,0 +1,139 @@
...
+"""Housing specific constant for manure storage (s/m)."""
```
elle-andreen, line 21:
```suggestion
"""Resistance value utilized in calculation of ammonia emission from manure stored in slurry storage outdoor, slurry storage underfloor, or anaerobic lagoon (s/m)."""
```

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
matthew7838: I believe this is used across various processors, so I would suggest holding this change and leaving this as a note in #2299, so it can be found easily when working on #2299.

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
```diff
@@ -0,0 +1,244 @@
...
+            pH=DEFAULT_PH_FOR_AMMONIA,
+        )
+        self._manure_to_process.ammoniacal_nitrogen = max(
+            0.0, self._manure_to_process.ammoniacal_nitrogen - storage_ammonia_nitrogen)
+        self._manure_to_process.nitrogen = max(0.0, self._manure_to_process.nitrogen - storage_ammonia_nitrogen)
+        return storage_ammonia_nitrogen
+
+    def _apply_nitrous_oxide_emissions(self, received_manure: ManureStream) -> float:
+        """
+        Calculates nitrous oxide emissions from stored manure and accounts the nitrogen
```
elle-andreen:
```suggestion
        Calculates nitrous oxide emissions from stored manure and accounts for the nitrogen
```

input/data/manure/refreshed_manure_management.json
```diff
@@ -35,5 +35,25 @@
...
           }
+    ],
+    "storage": [
+      {
+        "name": "slurry_storage_outdoor_1",
+        "type": "slurry storage outdoor",
+        "capacity": 100,
+        "cover": "no cover",
+        "storage_time_period": 120,
+        "surface_area": 88.8,
```
elle-andreen: I'm not sure where would be appropriate to note that we'll need to come back and update this value once we know whether we can use ManureManager or something else to estimate surface area, or if we need to use another, more simpler function. Either way, I _think_ the value of all surface area input values should be null by default, and over-written by either the user or the internal function to determine surface area. I guess these could maybe be set to null now, but I'll leave that up to you Allister.

elle-andreen (approved):
This looks good to me! I think the standardization discussions you guys have been having are a major improvement. As you told me, functionality here is the same as it was in the previous version of this PR, but the added docstring and methods are very helpful in improving code readability within and between files! Thanks 🙂

input/data/manure/refreshed_manure_management.json
allisterakun: I will change it to null for now, and create an issue to revisit this after the implementation of ManureManager.

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
```diff
@@ -0,0 +1,244 @@
...
+        manure_to_return = super().process_manure(current_day_conditions, time)
+        self._manure_to_process = manure_to_return["manure"] if manure_to_return else copy(self._stored_manure)
+
+        manure_temperature = self._determine_outdoor_storage_temperature(
+            air_temperature=current_day_conditions.mean_air_temperature
+        )
+
+        storage_methane_burned, total_storage_methane = self._apply_methane_emissions(manure_temperature)
+        storage_ammonia_nitrogen = self._apply_ammonia_emissions(manure_temperature)
+        storage_nitrous_oxide_nitrogen = self._apply_nitrous_oxide_emissions(received_manure)
```
ew3361zh: With SSO, SSU, and AL, we kind of have 2 variables that could be considered for `_manure_to_process` (we're sending `received_manure` to this function for nitrous oxide emissions application in the same way we're doing with `stored_manure` but we've set `manure_to_process` to be the stand in for `stored_manure`). 

Is that confusing at all do you think?

RUFAS/biophysical/manure/storage/slurry_storage_underfloor.py
```diff
@@ -0,0 +1,225 @@
...
+
+        """
+        storage_nitrous_oxide_nitrogen = self._calculate_nitrous_oxide_emissions(
+            nitrous_oxide_emissions_factor=STORAGE_COVER_NITROUS_OXIDE_EMISSIONS_FACTOR_MAPPING[self._cover],
+            nitrogen_added=received_manure.nitrogen,
+        )
+        self._manure_to_process.nitrogen = max(0.0, self._manure_to_process.nitrogen - storage_nitrous_oxide_nitrogen)
+        return storage_nitrous_oxide_nitrogen
+
+    def _report_slurry_storage_underfloor_outputs(
```
ew3361zh: Can you say a little about why you created this function vs. calling the `_report_processor_output()` from `process_manure()` directly?

My main counter would be that adding a new output from `process_manure()` would now involve updating 4 separate things:
- the call to this function in `process_manure()` to include the new output
- the args section of this function
- the docstring of this function
- adding the call to self._report_processor_output() here

Whereas if it's just a direct call to `_report_processor_output()` in `process_manure()` it would be just adding 1 line of code.

allisterakun: That's true. the only reason I created this function was because I saw a large block of codes for reporting those outputs. But I think it's ok and I will move them back to `process_manure()`

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
allisterakun: I think I get your point. After taking a look, we are not really modifying any properties of the "received_manure". The only thing we really need is the `.nitrogen` property of the received manure. So I don't think that passing in the received manure into the N2O calculation function is relying on side effect. Also, I think we can change the method so that we pass in `received_manure.nitrogen` directly, instead of the entire object to further avoid confusions.

RUFAS/biophysical/manure/storage/slurry_storage_underfloor.py
allisterakun: Done!

RUFAS/biophysical/manure/storage/slurry_storage_outdoor.py
ew3361zh: Sounds good to me!

ew3361zh (approved):
Looks great, Allister!

tests/test_biophysical/test_manure/test_storage/test_slurry_storage_underfloor.py
```diff
@@ -0,0 +1,299 @@
...
+    expected_stored_manure = copy(stored_manure)
+    mock_calculate_ammonia_emissions = mocker.patch.object(
+        slurry_storage_underfloor, "_calculate_ammonia_emissions", return_value=(dummy_storage_ammonia := 1.23)
+    )
+    expected_stored_manure.ammoniacal_nitrogen = max(
+        0.0, expected_stored_manure.ammoniacal_nitrogen - dummy_storage_ammonia
+    )
+    expected_stored_manure.nitrogen = max(0.0, expected_stored_manure.nitrogen - dummy_storage_ammonia)
+
+    slurry_storage_underfloor._apply_ammonia_emissions((dummy_manure_temperature := 25.0))
```
matthew7838: The best practice to test these would be manually calculating the nutrient values and setting them in the expected manure stream object instead of applying the same logic from the code to the tests. For example, 
```
max(
        0.0, expected_stored_manure.ammoniacal_nitrogen - dummy_storage_ammonia
    )
```
It should be manually calculated and assigned to `expected_stored_manure.ammoniacal_nitrogen`.

matthew7838 (approved):
Great work, Allister. This is also really helpful for me to set up an open lot. Left a comment regarding testing that shouldn't be too much of a change.

input/data/manure/refreshed_manure_management.json
```diff
@@ -35,5 +35,25 @@
...
+    ],
+    "storage": [
+      {
+        "name": "slurry_storage_outdoor_1",
+        "type": "slurry storage outdoor",
+        "capacity": 100,
+        "cover": "no cover",
+        "storage_time_period": 120,
+        "surface_area": null,
+        "nitrous_oxide_emissions_factor": 0.25
```
elle-andreen: I don't think this value needs to be an input for either slurry storage - the emission factor should be determined based on the value of `manure_cover `, which I think (?) is happening in this PR already

input/data/manure/refreshed_manure_management.json
```diff
@@ -35,5 +35,25 @@
             "potassium_efficiency": 0.23,
             "ash_efficiency": 0.20
           }
+    ],
+    "storage": [
+      {
+        "name": "slurry_storage_outdoor_1",
+        "type": "slurry storage outdoor",
+        "capacity": 100,
```
elle-andreen: I think we can also make this value null for now, or even omit it from the default manure input? I believe it is not utilized yet and it could be misleading for users to think does make a difference in calculations
```suggestion
        "capacity": null,
```

PR #2293, opened 2025-03-24
make empty sold animal report by animal type

### LINKED ISSUE #2288: report_sold_animal_information_sort_by_sell_day and cascading report filter errors
## Expected Behavior

The the variable `AnimalModuleReporter.report_sold_animal_information_sort_by_sell_day.sold_calves_sold_weight` should be reported, even if empty, to avoid cascading report filter problems - the "milk allocation factor" field in the provided report filter provided fails because of this.

## Current Behavior

There's a problem where cross referencing a missing value causes an important output to be missing. Namely, it's the total NE milk and meat_hor_agg' value, which is used to generate the milk allocation factor, which is a critical output for the FARM reports.

I dig a bit of digging, and narrowed it down to .*AnimalModuleReporter.report_sold_animal_information_sort_by_sell_day.sold_calves_sold_weight.* not capturing anything...because it's a really small farm, and no calves are being sold. The nature of the reporting method means that the variable name is never in the output pool (unlike other variables, which would show up with a value of 0).

Additional context: this is on the FARM ES fork, so it's months behind `dev` - but I confirmed `dev` returns the same error. It can be elicited using the filter I've attached, and by making a couple tweaks to the default_animal.json  - I reduced the herd size a little, but the main thing is setting both the male_calf_rate_sexed_semen and male_calf_rate_sexed_semen to 0. Those two changes mean that calves won't get sold ever, so no calves enter the report_sold_animal_information_sort_by_sell_day reporting method...meaning we get a cascading series of filter errors (since they're all interdependent). Just the top ones here are critical - you can see the sold calf bw isn't generated, which then means the next 5 filters also fail.

## Possible Solution

Two options considered right now:
A) add report logic to allow some report fields to return a 0 when the user specifies, since in this case a 0 is "real" (there may be cases where lack of a value shouldn't assume 0)
B) "initialize" the variable(s) in report_sold_animal_information_sort_by_sell_day, so that there's at least a value of 0 to start with for each animal group. This requires some thought as to how we should initialize these - weight sold makes sense to report as 0, but setting the sell day is trickier. Should we set `_first_sell_event` and `_last_sell_event` both to 0 to indicate at the report stage that there were no actual sales?

Something resembling B is preferred by @KFosterReed (my explanation on Slack was slightly different, so please chime in if the added speculation here is off kilter!) and I (@JoeWaddell) am inclined to agree: in part because it adds a layer of complexity to an already complex report generator, but perhaps more importantly it could lead to incorrect assumptions/use on the side of the user.

## Steps to Reproduce

1. Use the provided animal json, and output filter, and then run default.

[report_FARMES_Oct2.json](https://github.com/user-attachments/files/19374988/report_FARMES_Oct2.json)

[default_animal.json](https://github.com/user-attachments/files/19374990/default_animal.json)

JoeWaddell:
### Context

Issue(s) closed by this pull request: closes #2288

### What

Added logic to send fake data to the `report_sold_animal_information_sort_by_sell_day` method, to avoid missing output variables at the Report/Graph/etc Generator steps.

### Why

If this is missing in the outputs, report generator fails to work properly, and fails to generate results that depend on initial calculations (see #2288).

### How

Here, I simply added a check to report a list of 1 animal, with a body_weight of 0, if no animals were in that group. This needs to be reconsidered, as it implies an animal sold (when updating the `sold_count` in the `report_sold_animal_information_sort_by_sell_day` method).

What might need to be done, instead, is supplying an empty list of animals, and modifying `report_sold_animal_information_sort_by_sell_day` to handle an empty list differently. 

The `sold_at_day` is also set here as day 0, which could also be misleading.

### Test plan

We need to think carefully about how to report the `body_weight` for an empty output, and issues it might cause if a user tries using the value in calculating a mean, or other ways a 0 could be invalid.


KFosterReed:
So I think this looks like a pretty elegant solution to me! I am not super worried about the about how it affects the `report_sold_animal_information_sort_by_sell_day` reports because 1) the total number of animals sold is still reported correctly in `AnimalModuleReporter.report_life_cycle_manager_data.sold_calf_num` and zeros are still recorded on days when animals are not sold in the `report_sold_animal_information_sort_by_sell_day.[animal_type]_sold_count` and `report_sold_animal_information_sort_by_sell_day.[animal_type]_sold_weight` variables. See the attached csv collected from the attached filter run with Joe's example animal json in [Issue 2288](https://github.com/RuminantFarmSystems/MASM/issues/2288). 

Overall, I think it is an improvement and we can probably think about other outputs that we should follow a similar path!

[default_saved_variables_csv_sold_animals.txt_24-Mar-2025_Mon_17-01-55.csv](https://github.com/user-attachments/files/19438745/default_saved_variables_csv_sold_animals.txt_24-Mar-2025_Mon_17-01-55.csv)
[csv_sold_animals.txt](https://github.com/user-attachments/files/19438750/csv_sold_animals.txt)

RUFAS/routines/animal/animal_module_reporter.py
allisterakun, line 1: I think you should change the `AnimalModuleReporter` under the `biophysical` folder instead.

allisterakun (changes requested):

allisterakun (approved):
LGTM!

jadamchick (approved):
Looks good! I did have to update the `stillbirth_rate` to zero along with the `male_calf_rate` to replicate a scenario in which zero calves were sold (since stillbirths get added to the sold list -- which prompted me to make an issue about that, [here](https://github.com/RuminantFarmSystems/RuFaS/issues/2484))

As far as I can tell, it seems like a workable solution. The `AnimalModuleReporter.report_life_cycle_manager_data.sold_calf_num` still reports zero when there are no calves sold, so it isn't falsely counting anything. Total calf body weight sold is also reported as zero, and average calf body weight sold is left blank (in my filter), so those seem to be behaving nicely.

Issue #2281, opened 2025-03-14
[Manure] Add Manure Processor Type Enum


ew3361zh:
From discussions in the weekly manure module refresh meetings:

1. Set up an Enum to map user input to desired manure processor.
2. Once a general approach to manure processor type (and processor sub-type e.g. a `screw_press` Separator) are established in input files, ensure that information is being properly used for reporting to OM.

Should be considered part of Phase 2 of #2092

PR #2306, opened 2025-03-26
Manure Refresh Patch

allisterakun:
This PR implements the changes requested per the dev team meeting on 3/26/2025. Specifically:
## New `Processor._report_processor_output()` function
This is a wrapper function for `self._om.add_variable()` and the necessary `info_map` in the base `Processor` class.
All of the children and grandchildren class can call this method to add a variable to the `OutputManager` by providing:
- variable_name : str
- variable_value : float
Currently, we are expecting the added variable to be a float, we can easily update this to accept `Any` value if needed.
- data_origin_funcion : str
The name of the function that reported the variable value.
- unit : MeasurementUnits
The unit for the reported variable.
- time : Time

All direct calls to `om.add_variable()` are updated to use this new method.

## New `Storage._manure_to_process` attribute
Create a new `_manure_to_process` attribute of type `ManureStream` in the base `Storage` class.
This property represents the manure that will be processed during the `process_manure()` method call. We introduce this variable to track that manure as emissions are being calculated and removed.

## Relocate `Handler. determine_barn_temperature()` and `Storage._determine_outdoor_storage_temperature()`
Both `Handler. determine_barn_temperature()` and `Storage._determine_outdoor_storage_temperature()` are relocated into the base `Processor` class.
`determine_barn_temperature()` is renamed to `_determine_barn_temperature` for consistency.

## Change all class methods into static methods
After discussion and careful evaluation, we have decided to change all class methods in the refreshed manure module into static methods, by changing the function decoration from `@classmethod` to `@staticmethod` and make any other necessary updates to all the references to those methods.

### Generic prefix
The base prefix is configured to be
```python
self._prefix = f"Manure.{base_class_name}.{self.__class__.__name__}.{self.name}"
```
where the `base_class_name` is defined as 
```python
base_class_name = self.__class__.__bases__[0].__name__
```
This format should apply to all the `Storage` and `Digester` classes. (there is only one `AnaerobicDigester` class under `Digester`).
Here is a full list of the resulting prefixes for all members in the `Storage` and `Digester` class.
- `AnaerobicDigester`: "Manure.Digester.AnaerobicDigester.my_anaerobic_digester"
- `AnaerobicLagoon`: "Manure.Storage.AnaerobicLagoon.my_anaerobic_lagoon"
- `SlurryStorageOutdoor`: "Manure.Storage.SlurryStorageOutdoor.my_slurry_storage_outdoor"
- `SlurryStorageUnderFloor`: "Manure.Storage.SlurryStorageOutdoor.my_slurry_storage_underfloor"
- `OpenLots`: "Manure.Storage.OpenLots.my_open_lots"
- `Composting`: "Manure.Storage.Composting.my_composting"
- `CompostBeddedPackBarn`: "Manure.Storage.CompostBeddedPackBarn.my_compos_bedded_pack_barn"

### Handler
The prefixes for all `Handler` classes are configured in the base `Handler` class as
```python
self._prefix = f"Manure.{self.__class__.__name__}.{self.handler_type}.{self.name}"
```
The `f"{self.handler_type}"` portion will be replaced with `f"{self.type.value}"` once the processor type enum is implemented in issue #2281.
Here is a full list of the resulting prefixes for all members of the `Handler` class (using "MANUAL_SCRAPER" as handler type for example):
- `Handler`: "Manure.Handler.MANUAL_SCRAPER.my_manual_scraper_handler"
- `SingleStreamHandler`: "Manure. SingleStreamHandler.MANUAL_SCRAPER.my_manual_scraper_handler"

### Separator
The prefix for the `Separator ` class is configured as
```python
self._prefix = f"Manure.{self.__class__.__name__}.{"separator_type"}.{self.name}"
```
where `"separator_type"` is a placeholder for the actual type of separator. This will be updated after the implementation of the processor type enum in issues #2281.
Here is the an example for resulting prefix:
- `Separator`: "Manure.Separator.separator_type.my_separator"

## Other Changes
- Updated `AnaerobicDigester._report_anaerobic_digester_outputs()` function to use `Processor._report_manure_stream()` to handle the reporting of `self._manure_in_digester` properties.
- Changed all variables with name `temp` or `average_temp` to `temperature` and `average_temperature`

## TODO
- Update the prefix for `Handler` and `Separator` after completing issue #2281.

## Test plan

All affected unit tests have been updated.

### DIFF
--- a/RUFAS/biophysical/manure/digester/anaerobic_digester.py
+++ b/RUFAS/biophysical/manure/digester/anaerobic_digester.py
@@ -108,7 +108,7 @@ class AnaerobicDigester(Digester):
                 methane_leakage_mass=0.0,
                 minimum_digester_volume=0.0,
                 top_cover_volume=0.0,
-                time=time,
+                simulation_day=time.simulation_day,
             )
             return {}
 
@@ -152,7 +152,7 @@ class AnaerobicDigester(Digester):
             methane_leakage_mass=methane_leakage,
             minimum_digester_volume=minimum_digester_volume,
             top_cover_volume=top_cover_volume,
-            time=time,
+            simulation_day=time.simulation_day,
         )
 
         digested_manure = replace(self._manure_in_digester)
--- a/RUFAS/biophysical/manure/digester/anaerobic_digester.py
+++ b/RUFAS/biophysical/manure/digester/anaerobic_digester.py
@@ -253,7 +253,7 @@ class AnaerobicDigester(Digester):
         self._report_manure_stream(self._manure_in_digester, "", simulation_day)
 
         self._report_processor_output(
-            "biogas",biogas, data_origin_function, MeasurementUnits.KILOGRAMS, simulation_day
+            "biogas", biogas, data_origin_function, MeasurementUnits.KILOGRAMS, simulation_day
         )
         self._report_processor_output(
             "methane_leakage_mass",
--- a/RUFAS/biophysical/manure/digester/anaerobic_digester.py
+++ b/RUFAS/biophysical/manure/digester/anaerobic_digester.py
@@ -221,7 +221,7 @@ class AnaerobicDigester(Digester):
         methane_leakage_mass: float,
         minimum_digester_volume: float,
         top_cover_volume: float,
-        time: Time,
+        simulation_day: int,
     ) -> None:
         """
         Reports manure that was digested and the amounts of different things that were lost in the anaerobic digestion
@@ -231,7 +231,7 @@ class AnaerobicDigester(Digester):
         ----------
         biogas : float
             Captured biogas (kg).
-        biogass_energy_content : float
+        biogas_energy_content : float
             Energy from captured biogas (MJ).
         evaporated_water : float
             Water that evaporated during anaerobic digestion (m^3)
@@ -245,61 +245,64 @@ class AnaerobicDigester(Digester):
             Minimum volume of manure allowed to be left in the anaerobic digester (m^3).
         top_cover_volume : float
             Headspace volume above manure inside digester where biogas collects (m^3).
+        simulation_day : int
+            The current simulation day.
 
         """
-        self._report_manure_stream(self._manure_in_digester, "", time)
+        data_origin_function = self._report_anaerobic_digester_outputs.__name__
+        self._report_manure_stream(self._manure_in_digester, "", simulation_day)
 
         self._report_processor_output(
-            "biogas", biogas, self._report_anaerobic_digester_outputs.__name__, MeasurementUnits.KILOGRAMS, time
+            "biogas",biogas, data_origin_function, MeasurementUnits.KILOGRAMS, simulation_day
         )
         self._report_processor_output(
             "methane_leakage_mass",
             methane_leakage_mass,
-            self._report_anaerobic_digester_outputs.__name__,
+            data_origin_function,
             MeasurementUnits.KILOGRAMS,
-            time,
+            simulation_day,
         )
         self._report_processor_output(
             "methane_generation_volume",
             methane_generation_volume,
-            self._report_anaerobic_digester_outputs.__name__,
+            data_origin_function,
             MeasurementUnits.CUBIC_METERS,
-            time,
+            simulation_day,
         )
         self._report_processor_output(
             "evaporated_water",
             evaporated_water,
-            self._report_anaerobic_digester_outputs.__name__,
+            data_origin_function,
             MeasurementUnits.CUBIC_METERS,
-            time,
+            simulation_day,
         )
         self._report_processor_output(
             "minimum_digester_volume",
             minimum_digester_volume,
-            self._report_anaerobic_digester_outputs.__name__,
+            data_origin_function,
             MeasurementUnits.CUBIC_METERS,
-            time,
+            simulation_day,
         )
         self._report_processor_output(
             "top_cover_volume",
             top_cover_volume,
-            self._report_anaerobic_digester_outputs.__name__,
+            data_origin_function,
             MeasurementUnits.CUBIC_METERS,
-            time,
+            simulation_day,
         )
         self._report_processor_output(
             "heating_input_energy",
             heating_input_energy,
-            self._report_anaerobic_digester_outputs.__name__,
+            data_origin_function,
             MeasurementUnits.MEGAJOULES,
-            time,
+            simulation_day,
         )
         self._report_processor_output(
             "biogas_energy_content",
             biogas_energy_content,
-            self._report_anaerobic_digester_outputs.__name__,
+            data_origin_function,
             MeasurementUnits.MEGAJOULES,
-            time,
+            simulation_day,
         )
 
     @staticmethod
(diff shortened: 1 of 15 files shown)


RUFAS/biophysical/manure/processor.py
```diff
@@ -224,22 +228,22 @@ def _calculate_ammonia_emissions(
...
+    @staticmethod
+    def _calculate_ammonia_equilibrium_coefficient(temperature: float, pH: float) -> float:
         """
         Calculates the equilibrium coefficient for the ammonia gas in the air for a given concentration of total
         ammoniacal nitrogen in the solution.
 
         Parameters
         ----------
-        temp : float
+        temperature : float
```
ew3361zh, line 245: nice, thank you

RUFAS/biophysical/manure/processor.py
```diff
@@ -290,4 +294,90 @@ def _calculate_dissociation_coefficient_of_ammonium(cls, temp: float, pH: float)
...
+        References
+        ----------
+        Between 5 and 30 C, barn temperature is assumed to be equal to outdoor air temperature.
+        This function assumes that barn temperature does not drop below 5 C or increase above 30 C.
+        These bounds were suggested by manure SMEs and are supported by barn temperature ranges
+        reported in Bucklin et al. (FL, upper limit; https://doi.org/10.13031/2013.28851).
+        The lower bound (5 C) suggested by SMEs was based on general industry standards/conditions.
+
+        """
+        return float(clip(air_temp, 5, 30))
```
ew3361zh:
```suggestion
        return float(clip(air_temp, 5.0, 30.0))
```
Functionally I don't think it makes a difference but suggesting this just to be consistent with the other method.

RUFAS/biophysical/manure/processor.py
```diff
@@ -290,4 +294,90 @@ def _calculate_dissociation_coefficient_of_ammonium(cls, temp: float, pH: float)
...
+
+    def _report_processor_output(
+        self, variable_name: str, variable_value: float, data_origin_function: str, unit: MeasurementUnits, time: Time
```
ew3361zh:
```suggestion
        self, variable_name: str, variable_value: float, data_origin_function: str, units: MeasurementUnits, time: Time
```

RUFAS/biophysical/manure/processor.py
```diff
@@ -290,4 +294,90 @@ def _calculate_dissociation_coefficient_of_ammonium(cls, temp: float, pH: float)
...
+        Parameters
+        ----------
+        variable_name : str
+            The name of the reported variable.
+        variable_value : str
+            The value of the reported gas emission value.
+        data_origin_function : str
+            The name of the function that reported the gas emission value.
+        unit : MeasurementUnits
+            The unit for the reported gas emission value.
```
ew3361zh:
```suggestion
        units : MeasurementUnits
            The units for the reported gas emission value.
```

RUFAS/biophysical/manure/processor.py
```diff
@@ -290,4 +294,90 @@ def _calculate_dissociation_coefficient_of_ammonium(cls, temp: float, pH: float)
...
+        time : Time
+            Time instance tracking the current time of the simulation.
+
+        """
+        info_map = {
+            "class": self.__class__.__name__,
+            "function": data_origin_function,
+            "prefix": self._prefix,
+            "simulation_day": time.simulation_day,
+            "units": unit,
```
ew3361zh:
```suggestion
            "units": units,
```

RUFAS/biophysical/manure/separator/separator.py
```diff
@@ -84,6 +84,8 @@ def __init__(
         self.ash_efficiency: float = ash_efficiency
         self.volatile_solids_efficiency: float = volatile_solids_efficiency
         self.total_solids_efficiency: float = total_solids_efficiency
+        # TODO: replace '{"separator_type"}' with '{self.type.value}' once we implement the processor type enum.
```
ew3361zh, line 87:
```suggestion
        # TODO: replace '{"separator_type"}' with '{self.type.value}' once we implement the processor type enum. Issue #2281
```

ew3361zh, line 87:
```suggestion
        # TODO: replace '{"separator_type"}' with '{self.type.value}' once we implement the processor type enum. Issue #2281
```

RUFAS/biophysical/manure/storage/storage.py
```diff
@@ -119,7 +118,7 @@ def __init__(
         self._storage_time_period = storage_time_period
         self._surface_area = surface_area
         self._nitrous_oxide_emissions_factor = nitrous_oxide_emissions_factor
-        self._accumulated_output_prefix = f"Accumulated{self._prefix}"
+        self._manure_to_process = ManureStream.make_empty_manure_stream()
```
ew3361zh, line 122: Add to attribute docstrings

RUFAS/biophysical/manure/processor.py
allisterakun: Little curious of why it should be plural. I thought this function is reporting just one variable, which should correspond to a single unit.

ew3361zh: I think of units as kind of like pants - it's a plural for both the singular and plural.

ew3361zh (approved):
LGTM, Allister, nice work. I think we may want to see if there's a way we can set the prefix universally from the Processor class once the Enum is set up but it might be that this is the best way. Thanks for jumping on this!

RUFAS/biophysical/manure/processor.py
```diff
@@ -290,4 +293,90 @@ def _calculate_dissociation_coefficient_of_ammonium(cls, temp: float, pH: float)
...
+        This function assumes that barn temperature does not drop below 5 C or increase above 30 C.
+        These bounds were suggested by manure SMEs and are supported by barn temperature ranges
+        reported in Bucklin et al. (FL, upper limit; https://doi.org/10.13031/2013.28851).
+        The lower bound (5 C) suggested by SMEs was based on general industry standards/conditions.
+
+        """
+        return float(clip(air_temperature, 5.0, 30.0))
+
+    def _report_processor_output(
+        self, variable_name: str, variable_value: float, data_origin_function: str, units: MeasurementUnits, time: Time
```
ew3361zh: before i forget, for this reporting function and the report stream function, can we change them both to receive just the simulation day rather than the whole `Time` object?

elle-andreen (approved):
These changes make sense to me. Thanks Allister!

RUFAS/biophysical/manure/digester/anaerobic_digester.py
```diff
@@ -384,16 +403,16 @@ def _calculate_CSTR_methane_volume(cls, total_volatile_solids: float) -> float:
         """
         return ManureConstants.ACHIEVABLE_METHANE_EMISSION * total_volatile_solids
 
-    @classmethod
-    def _calculate_methane_leakage(cls, generated_methane_mass: float, leakage_fraction: float) -> float:
+    @staticmethod
+    def _calculate_methane_leakage(generated_methane_mass: float, leakage_fraction: float) -> float:
```
matthew7838, line 407: I know this might be a bit out of the scope of this PR, but is there a reason these methods are set for internal use only, with the `_` in front of the method name?

matthew7838 (approved):
LGTM and makes sense to me as we discussed. Raised a small discussion in the comments.

RUFAS/biophysical/manure/digester/anaerobic_digester.py
allisterakun: I don't see any reason for these methods to be called from outside the process and its children/grandchildren class, so I'm just making them private.

